PR #13112, opened 2016-09-06
don't include date opened unless specified or on create

snopoke:
@biyeun @czue


czue:

PR #13102, opened 2016-09-05
Invalid XML is crashing sql-forms reindexing.

kaapstorm:
Context: http://manage.dimagi.com/default.asp?236666

@millerdev I have not reproduced this bug locally, but I can see from the comment in the code that this change replaces that this is something you have thought about. I don't want to fix this issue only to break things that have been working. Can you think of negative implications of this change?

cc @dannyroberts @biyeun cc @NoahCarnahan @orangejenny

PR #13114, opened 2016-09-06
form_data returns empty dictionary on XMLSyntaxError

kaapstorm:
Context: http://manage.dimagi.com/default.asp?236666

Instead of trying to return valid XML, as I tried in my first attempt in PR #13102, form_data catches when it's broken and returns an empty dictionary.

@snopoke cc @dannyroberts @biyeun @NoahCarnahan @orangejenny


snopoke:
:+1: on :white_check_mark:

Looked at this with @kaapstorm and this is in line with how it is being handled by couch forms.

dannyroberts:
awesome, thanks for taking this! 👍

PR #13116, opened 2016-09-06
child modules easy refs

snopoke:
http://manage.dimagi.com/default.asp?236529
@orangejenny @emord 

This fixes the specific issue but I'd still like to have a discussion about rethinking the structure of the schema.

### DIFF
--- a/corehq/apps/app_manager/util.py
+++ b/corehq/apps/app_manager/util.py
@@ -431,9 +431,13 @@ def get_session_schema(form):
     from corehq.apps.app_manager.suite_xml.sections.entries import EntriesHelper
     structure = {}
     datums = EntriesHelper(form.get_app()).get_datums_meta_for_form_generic(form)
-    datums = [d for d in datums if not d.is_new_case_id and d.case_type and d.datum.id == 'case_id']
+    datums = [
+        d for d in datums
+        if not d.is_new_case_id and d.case_type and d.requires_selection
+    ]
     if len(datums):
-        structure['case_id'] = {
+        session_var = datums[-1].datum.id
+        structure[session_var] = {
             "reference": {
                 "source": "casedb",
                 "subset": "case",
--- a/corehq/apps/app_manager/tests/test_util.py
+++ b/corehq/apps/app_manager/tests/test_util.py
@@ -193,6 +193,64 @@ class SchemaTest(SimpleTestCase):
             },
         })
 
+    def test_get_session_schema_for_child_module(self):
+        """
+            m0 - opens 'gold-fish' case.
+            m1 - has m0 as root-module, has parent-select, updates 'guppy' case, creates
+                 'pregnancy' subcases to guppy
+        """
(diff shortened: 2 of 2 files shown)


emord:
:+1: on :white_check_mark: 

The schema was more generic initially, but we had to cut down on the generic-ness of it once we found some case hierarchies that were not easily supported via `#case/case_type/property`. I think it was due to referrals having multiple possible parents, and at form building time, we would not be able to understand what case type was the parent.

When we moved to `#case/property` or `#case/parent/property` we intentionally made the decision to only support one case id at a time.

Feel free to grab me this week to discuss these changes, although I assume they will change once @millerdev continues his work on a better syntax for advanced modules

corehq/apps/app_manager/util.py
```diff
@@ -431,9 +431,13 @@ def get_session_schema(form):
...
     datums = EntriesHelper(form.get_app()).get_datums_meta_for_form_generic(form)
-    datums = [d for d in datums if not d.is_new_case_id and d.case_type and d.datum.id == 'case_id']
+    datums = [
+        d for d in datums
+        if not d.is_new_case_id and d.case_type and d.requires_selection
+    ]
     if len(datums):
-        structure['case_id'] = {
+        session_var = datums[-1].datum.id
+        structure[session_var] = {
```
millerdev: Will `session_var` ever be anything other than `"case_id"`? I wonder if that could break case hashtags in Vellum since Vellum basically has one case hashtag prefix `#case/` which means `instance('casedb')/cases/case[@case_id = instance('commcaresession')/session/data/case_id]/`? On looking I'm not sure if any of that is hard-coded in Vellum, but I seem to remember that some of it had been at one time.

[Documentation for the caseDB schema](https://github.com/dimagi/Vellum/blob/master/src/datasources.js) can be found in Vellum. There is a similar comment in HQ in the [docstring of `get_form_data_schema`](https://github.com/dimagi/commcare-hq/blob/master/corehq/apps/app_manager/views/formdesigner.py#L146)

emord: `case_id` is hardcoded in vellum as the place where we filter on. `instance('casedb')/casedb/case[@case_id=instance('commcaresession')/session/data/case_id]/index/parent"` but the actual variable from the session should be able to support any string, (but not multiple case ids)

snopoke: @millerdev yes the session var is sometimes something other than `case_id` e.g. in child modules. That's what this change is all about.

I have tested this locally and it works fine. 

Re the docs - they leave a lot unexplained e.g. what is `"key": "@case_type"` used for (the case type isn't even included in the schema anywhere except the name)? Do the subset ID's mean anything i.e. if I change `parent` to `mother` will it break things? How do relationships work - are the relationship names important?

millerdev: Agreed, the docs could be improved, and those are all good questions.

PR #13117, opened 2016-09-06
Update abt Madagascar report targets

NoahCarnahan:
@orangejenny @kaapstorm


NoahCarnahan:
(killed build because no test coverage)

PR #13121, opened 2016-09-06
don't use imports from __init__ that require models

nickpell:
django 1.9 fails hard if **init** imports models (even indirectly)

@calellowitz


calellowitz:
👍 on ✅

nickpell:
@dannyroberts @sravfeyn

nickpell:
✅

PR #13125, opened 2016-09-06
begin moving indirect model imports out of __init__.py

nickpell:
Needed for django 1.9, which fails hard if models.py is imported directly or indirectly in **init**.py

Doing a first pass with a5288.

@dannyroberts @sravfeyn

### DIFF
--- a/corehq/apps/reports/dispatcher.py
+++ b/corehq/apps/reports/dispatcher.py
@@ -1,20 +1,23 @@
-from django.utils.decorators import method_decorator
 from django.core.urlresolvers import reverse
 from django.http import Http404, HttpResponseRedirect, HttpResponseBadRequest
-from django.views.generic.base import View
+from django.utils.decorators import method_decorator
 from django.utils.translation import ugettext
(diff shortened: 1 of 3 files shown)


dannyroberts:
Related?

```
ImportError: No module named app-manager-testviews-domain
```
```
ImportError: No module named test-access-restrictions-domain
```
```
ImportError: No module named fandago
```

nickpell:
✅

corehq/apps/reports/dispatcher.py
```diff
@@ -84,7 +87,19 @@ def process(reports):
...
+
+        module_name = get_domain_module_map().get(domain)
+        if module_name is None:
+            custom_reports = ()
+        else:
+            module = __import__(module_name, fromlist=['reports'])
+            reports = getattr(module, 'reports')
+            custom_reports = process(getattr(reports, attr_name, ()))
+
+            # soon to be removed
```
emord: @nickpell I think that this can be removed now right?

nickpell: Looks like yes.

PR #13126, opened 2016-09-06
99DOTS case repeater

proteusvacuum:
This is a pretty big one:
- Refactor of repeaters to allow us to add custom repeaters
- Adds the 99DOTS register patient repeater (outlined [here](https://docs.google.com/document/d/1PH4EPcSqhsYP-lV-_4_V6C9AvkpaBmIEe5ID5U2mAlc/edit#heading=h.n5l7q7bz65k1)) -- Still need to add the "update patient" repeater, but thought I'd get feedback on this approach while I work on it.

The refactor does a few things:
- Defines repeaters in settings, so we can have different repeaters active on different environments, etc.
- Moves a bunch of state / properties back into their respective classes so that we only have one-way dependencies (i.e. the generic repeater code needs to know nothing about the enabled repeaters)
- Adds `handle_success` and `handle_failure` methods to repeat records so that we can carry out actions on success / failure (e.g. update a case property)
- Adds `available_for_domain` to repeaters so they can be turned on / off depending on logic
- Adds `allow_retries` to the repeater to turn on / off retries depending on some logic based on the response (i.e. we don't just hammer the endpoint if there is an error in the data that we are sending).

For eNikshay I also added the `ENikshayCaseStructureMixin`, which should help whenever case structures change in the app.

Code Buddy @emord 
FYI @czue

### DIFF
--- a/corehq/apps/domain/views.py
+++ b/corehq/apps/domain/views.py
@@ -2124,18 +2124,6 @@ class CaseSearchConfigView(BaseAdminProjectSettingsView):
         }
 
 
-class RepeaterMixin(object):
-
-    @property
-    def friendly_repeater_names(self):
-        return {
-            'FormRepeater': _("Forward Forms"),
-            'CaseRepeater': _("Forward Cases"),
-            'ShortFormRepeater': _("Forward Form Stubs"),
-            'AppStructureRepeater': _("Forward App Schema Changes"),
-        }
-
-
 class DomainForwardingRepeatRecords(GenericTabularReport):
     name = 'Repeat Records'
     base_template = 'domain/repeat_record_report.html'
@@ -2256,7 +2244,7 @@ class DomainForwardingRepeatRecords(GenericTabularReport):
         )
 
 
-class DomainForwardingOptionsView(BaseAdminProjectSettingsView, RepeaterMixin):
+class DomainForwardingOptionsView(BaseAdminProjectSettingsView):
     urlname = 'domain_forwarding'
     page_title = ugettext_lazy("Data Forwarding")
     template_name = 'domain/admin/domain_forwarding.html'
@@ -2281,7 +2269,7 @@ class DomainForwardingOptionsView(BaseAdminProjectSettingsView, RepeaterMixin):
         }
 
 
-class AddRepeaterView(BaseAdminProjectSettingsView, RepeaterMixin):
+class AddRepeaterView(BaseAdminProjectSettingsView):
     urlname = 'add_repeater'
     page_title = ugettext_lazy("Forward Data")
     template_name = 'domain/admin/add_form_repeater.html'
@@ -2308,7 +2296,7 @@ class AddRepeaterView(BaseAdminProjectSettingsView, RepeaterMixin):
 
     @property
     def page_name(self):
-        return "Forward %s" % self.friendly_repeater_names.get(self.repeater_type, "Data")
+        return self.repeater_class.friendly_name
 
     @property
     @memoized
--- a/corehq/apps/domain/views.py
+++ b/corehq/apps/domain/views.py
@@ -2129,10 +2129,10 @@ class RepeaterMixin(object):
     @property
     def friendly_repeater_names(self):
         return {
-            'FormRepeater': _("Forms"),
-            'CaseRepeater': _("Cases"),
-            'ShortFormRepeater': _("Form Stubs"),
-            'AppStructureRepeater': _("App Schema Changes"),
+            'FormRepeater': _("Forward Forms"),
+            'CaseRepeater': _("Forward Cases"),
+            'ShortFormRepeater': _("Forward Form Stubs"),
+            'AppStructureRepeater': _("Forward App Schema Changes"),
         }
 
 
--- a/corehq/apps/domain/views.py
+++ b/corehq/apps/domain/views.py
@@ -2309,7 +2309,11 @@ class AddRepeaterView(BaseAdminProjectSettingsView):
         try:
             return get_all_repeater_types()[self.repeater_type]
         except KeyError:
-            raise Http404("No such repeater {}. Valid types: {}".format(self.repeater_type, get_all_repeater_types.keys()))
+            raise Http404(
+                "No such repeater {}. Valid types: {}".format(
+                    self.repeater_type, get_all_repeater_types.keys()
+                )
+            )
 
     @property
     @memoized
--- a/corehq/apps/domain/views.py
+++ b/corehq/apps/domain/views.py
@@ -115,11 +115,12 @@ from dimagi.utils.web import get_ip, json_response, get_site_domain
 from corehq.apps.users.decorators import require_can_edit_web_users, require_permission
 from corehq.apps.repeaters.forms import GenericRepeaterForm, FormRepeaterForm
 from corehq.apps.repeaters.models import Repeater, FormRepeater, CaseRepeater, ShortFormRepeater, \
-    AppStructureRepeater, RepeatRecord, repeater_types, RegisterGenerator
+    AppStructureRepeater, RepeatRecord, RegisterGenerator
 from corehq.apps.repeaters.dbaccessors import (
     get_paged_repeat_records,
     get_repeat_record_count,
 )
+from corehq.apps.repeaters.utils import get_all_repeater_types
 from corehq.apps.repeaters.const import (
     RECORD_FAILURE_STATE,
     RECORD_PENDING_STATE,
@@ -544,7 +545,7 @@ def test_repeater(request, domain):
     url = request.POST["url"]
     repeater_type = request.POST['repeater_type']
     format = request.POST.get('format', None)
-    repeater_class = repeater_types[repeater_type]
+    repeater_class = get_all_repeater_types()[repeater_type]
     form = GenericRepeaterForm(
         {"url": url, "format": format},
         domain=domain,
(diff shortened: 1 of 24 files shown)


corehq/apps/domain/views.py
```diff
@@ -2308,15 +2301,15 @@ def repeater_type(self):
...
 
     @property
     @memoized
     def repeater_class(self):
         try:
-            return repeater_types[self.repeater_type]
+            return get_all_repeater_types()[self.repeater_type]
         except KeyError:
-            raise Http404()
+            raise Http404("No such repeater {}. Valid types: {}".format(self.repeater_type, get_all_repeater_types.keys()))
```
dimagimon: E501 line too long (123 > 115 characters)

corehq/apps/repeaters/utils.py
```diff
@@ -0,0 +1,9 @@
+from collections import OrderedDict
+from django.conf import settings
+
+from dimagi.utils.modules import to_function
+
+def get_all_repeater_types():
```
dimagimon: E302 expected 2 blank lines, found 1

proteusvacuum:
and @nickpell if you are interested in some eNikshay stuff

proteusvacuum:
MEOWTECH @dannyroberts

czue:
looks juicy. :lemon: :strawberry: :apple:

czue:
@benrudolph and @kaapstorm might also be interested since they have mucked around with repeaters and custom forwarders. 

I took a super cursory pass and looked ok.

> Adds allow_retries to the repeater to turn on / off retries depending on some logic based on the response (i.e. we don't just hammer the endpoint if there is an error in the data that we are sending).

I think `simple_post_with_cached_timeout` is also supposed to handle this problem (though it's a weird place for it). were you seeing problems with that or just wanted to make it more explicit and/or needed to use different logic?

custom/enikshay/tests/test_case_utils.py
```diff
@@ -57,5 +71,18 @@ def test_get_adherence_cases_between_dates(self):
 
     @run_with_all_backends
     def test_get_episode(self):
-        self.create_case_structure()
         self.assertEqual(get_open_episode_case_from_person(self.domain, 'person').case_id, 'episode')
+
+    @run_with_all_backends
```
snopoke: Do these need to run on both backends? Can we not just run the on SQL?

custom/enikshay/integrations/ninetyninedots/repeater_generators.py
```diff
@@ -0,0 +1,63 @@
...
+    get_person_case_from_occurrence,
+)
+
+
+class RegisterPatientPayload(jsonobject.JsonObject):
+    beneficiary_id = jsonobject.StringProperty(required=True)
+    phone_numbers = jsonobject.StringProperty(required=True)
+
+
+@RegisterGenerator(NinetyNineDotsRegisterPatientRepeater, 'case_json', 'JSON', is_default=True)
```
snopoke: I think the format (`case_json`) is supposed to be unique

snopoke: Never mind - it has to be unique per repeater type

custom/enikshay/integrations/ninetyninedots/repeaters.py
```diff
@@ -0,0 +1,32 @@
...
+
+    @classmethod
+    def get_custom_url(cls, domain):
+        from custom.enikshay.integrations.ninetyninedots.views import RegisterPatientRepeaterView
+        return reverse(RegisterPatientRepeaterView.urlname, args=[domain])
+
+    def allowed_to_forward(self, case):
+        # checks whitelisted case types and users
+        allowed_case_types_and_users = super(NinetyNineDotsRegisterPatientRepeater, self).allowed_to_forward(case)
+        enabled = case.dynamic_case_properties().get('dots_99_enabled') == 'true'
```
snopoke: Could extract `case.dynamic_case_properties()` to a local var since I think it's not the cheapest function call especially for larger cases.

custom/enikshay/integrations/ninetyninedots/repeaters.py
```diff
@@ -28,10 +33,16 @@ def allowed_to_forward(self, case):
...
         if response is not None and response.status_code == 500:
             return True
         return False
+
+
+def create_case_repeat_records(sender, case, **kwargs):
+    create_repeat_records(NinetyNineDotsRegisterPatientRepeater, case)
+
+case_post_save.connect(create_case_repeat_records, CommCareCase)
+case_post_save.connect(create_case_repeat_records, CommCareCaseSQL)
```
snopoke: Also possibly only need this for SQL cases unless there are existing Couch domains that need this feature.

snopoke:
Didn't do a detailed review but looks good. +1 to simplifying repeater registration

custom/enikshay/integrations/ninetyninedots/repeaters.py
proteusvacuum: Good point.

corehq/apps/domain/views.py
```diff
@@ -2124,18 +2125,6 @@ def page_context(self):
         }
 
 
-class RepeaterMixin(object):
-
-    @property
-    def friendly_repeater_names(self):
```
benrudolph: 👬

corehq/apps/repeaters/models.py
```diff
@@ -368,7 +398,6 @@ def __unicode__(self):
         return "forwarding forms to: %s" % self.url
 
 
-@register_repeater_type
```
benrudolph: 👏 wanted to remove this for a long time

corehq/apps/repeaters/models.py
```diff
@@ -549,34 +573,62 @@ def fire(self, max_tries=3, force_send=False):
...
+            return self.handle_response(response, post_info, tries)
+
+    def handle_response(self, response, post_info, tries):
+        if 200 <= response.status_code < 300:
+            return self.handle_success(response)
+        else:
+            return self.handle_failure(response, post_info, tries)
+
+    def handle_success(self, response):
+        """Do something once the repeater succeeds
```
benrudolph: nit: could be more descriptive

benrudolph:
yes, there are a couple things mitigating hammering an endpoint. one is the `simple_post_with_cached_timeout` which'll cache the response. also on a failed request, we exponentially back off "firing" the repeater so failed repeaters only get processed once over max days: https://github.com/dimagi/commcare-hq/blob/master/corehq/apps/repeaters/const.py#L4

benrudolph:
looks good, some much needed ❤️ for the repeater code

proteusvacuum:
Valid points about `simple_post_with_cached_timeout`, but I may have mis-described what `allow_retries` is doing. It actually determines whether to _requeue_ the repeater once the `simple_post` method has failed `max_tries` times. I can see scenarios where we wouldn't want this to happen. For instance, in the case of 99DOTS, if I am getting error codes in the `400` range, it means that there is something malformed with the data that I am sending, and no manner of retrying will change that.

benrudolph:
i had still thought that that logic was already in place, but it was most likely convoluted. in any case it's more clear now

benrudolph:
oh wait, now i see what you're saying. ok that's cool, so if we back off to 7 days, there are scenarios where we want to stop trying altogether

proteusvacuum:
I think this is still a bit confusing. 

To be clear, before this change:
- The post method tries 3 times in rapid-fire succession (without any exponential backoff)
- If all three of these failed, the repeat record is requeued with exponential backoff. Each time the requeued record fires, it tries 3 times in succession.

After this change:
- If `allow_immediate_retries` returns true, then try 3 times in succession.
- If this fails, and `allow_retries` returns true, then requeue with exponential backoff. 

Both those methods can now do logic on the `request`, so depending on the result of that, different actions could be taken (e.g. if there is a `404`, then we can send a registration message to the same endpoint, or something).

dannyroberts:
```======================================================================
FAIL: test_trigger (custom.enikshay.integrations.ninetyninedots.tests.test_repeaters.TestRegisterPatientRepeater)
  File "/mnt/commcare-hq-ro/corehq/util/test_utils.py", line 232, in inner
  File "/mnt/commcare-hq-ro/corehq/util/test_utils.py", line 205, in __call__
  File "/mnt/commcare-hq-ro/corehq/util/test_utils.py", line 220, in call_with_settings
    fn(*args, **kwargs)
  File "/mnt/commcare-hq-ro/corehq/util/test_utils.py", line 201, in fn_with_pre_and_post
    self.fn(*args, **kwargs)
  File "/mnt/commcare-hq-ro/custom/enikshay/integrations/ninetyninedots/tests/test_repeaters.py", line 83, in test_trigger
    self.assertEqual(1, len(self.repeat_records().all()))
AssertionError: 1 != 0
```

benrudolph:
👍 thanks for the clarification

corehq/apps/domain/views.py
dimagimon: E501 line too long (123 > 115 characters)

corehq/apps/repeaters/utils.py
dimagimon: E302 expected 2 blank lines, found 1

settings.py
```diff
@@ -1579,6 +1581,16 @@
...
+    'corehq.apps.repeaters.models.CaseRepeater',
+    'corehq.apps.repeaters.models.ShortFormRepeater',
+    'corehq.apps.repeaters.models.AppStructureRepeater',
+)
+
+REPEATERS = BASE_REPEATERS + LOCAL_REPEATERS
+
+
 
 STATIC_UCR_REPORTS = [
```
dimagimon: E303 too many blank lines (3)

proteusvacuum:
✔️

benrudolph:
💥

benrudolph:
we may want to keep an eye on the repeater queue celery monitor for a day or so

proteusvacuum:
💃

dannyroberts:
Awesome stuff @proteusvacuum!

PR #13127, opened 2016-09-06
Bump timeout

benrudolph:
not sure the root cause yet, but this at least fixes the test

@dannyroberts


benrudolph:
(danny was acting on my behalf on that merge)

PR #13097, opened 2016-09-03
add reindex_sql_forms_in_domain command

dannyroberts:
@benrudolph @snopoke 

I couldn't think of a better way to do this. It doesn't really work within the reindexer system; I made a number of attempts to do that, but it felt like this rabbit hole, so I pulled out the essence of what happens when you run a reindexer, and just wrote that for-loop.

I also just ran this on nufnip-baseline-1 and betterbirth-light-touch to fix the missing forms they were seeing.

PR #13128, opened 2016-09-06
respond to pr feedback on reindex_sql_forms_in_domain

dannyroberts:
from https://github.com/dimagi/commcare-hq/pull/13097

### DIFF
--- a/corehq/apps/hqcase/management/commands/reindex_sql_forms_in_domain.py
+++ b/corehq/apps/hqcase/management/commands/reindex_sql_forms_in_domain.py
@@ -1,3 +1,4 @@
+from dimagi.utils.chunked import chunked
 from django.core.management import BaseCommand
 
 from corehq.form_processor.backends.sql.dbaccessors import *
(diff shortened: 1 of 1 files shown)


dannyroberts:
@snopoke

dannyroberts:
cc: @nickpell @sravfeyn

corehq/apps/hqcase/management/commands/reindex_sql_forms_in_domain.py
```diff
@@ -1,17 +1,21 @@
...
-        for form in FormAccessorSQL.get_forms(doc_ids):
-            form_json = form.to_json(include_attachments=True)
-            txform = transform_xform_for_elasticsearch(form_json)
-            reindexer = get_sql_form_reindexer()
-            reindexer.doc_processor.process_bulk_docs([txform])
+        all_doc_ids = FormAccessorSQL.get_form_ids_in_domain_by_state(domain, state)
+        for doc_ids in chunked(all_doc_ids, 100):
+            print 'Reindexing doc_ids: {}'.format(','.join(doc_ids))
+            reindexer.doc_processor.process_bulk_docs([
+                transform_xform_for_elasticsearch(form.to_json(include_attachments=True))
```
snopoke: Since you're using the reindexer it will handle the transformation for you so you don't need `transform_xform_for_elasticsearch`

dannyroberts: Oh wow you're totally right. I was confusing the `doc_provider` (which doesn't know about the transform) and the `doc_processor` (which knows about both the doc provider and the transform).

This was wrong in my previous PR too. Thanks for catching, I'll change it

dannyroberts: I also decided to use `reindexer.reindex_accessor.doc_to_json` instead of hardcoding `form.to_json(...)`

nickpell:
looks good to me

PR #13129, opened 2016-09-07
Implement IETF advice for browser compatibility

kaapstorm:
Context: http://manage.dimagi.com/default.asp?236479

Implements [IETF advice](https://tools.ietf.org/html/rfc6266#appendix-D) for disastrous [browser compatibility](http://greenbytes.de/tech/tc2231/#attfnboth) for Content-Disposition header fields.

@dannyroberts @biyeun

### DIFF
--- a/corehq/apps/reports/tests/test_util.py
+++ b/corehq/apps/reports/tests/test_util.py
@@ -4,7 +4,7 @@ from elasticsearch.exceptions import ConnectionError
 from mock import Mock
 
 from corehq.apps.domain.shortcuts import create_domain
-from corehq.apps.reports.util import create_export_filter, safe_filename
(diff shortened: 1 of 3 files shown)


kaapstorm:
In Chrome, it gets the name of the export right, and goes with the ASCII version of the filename:
filename_in_chrome

In Firefox (on Linux -- default Linux font may be to blame here) the font fails to render the Gothic alphabet characters, but treats the filename the same as Chrome:
less_than_awesome

I was hoping we'd get the non-ASCII characters in the filename, because *progress* and *Western imperialism* and all, but I'm OK with letting go at this point.

kaapstorm:
:white_check_mark:

biyeun (approved):

PR #13130, opened 2016-09-07
Switched cbt filter to multi select

kkrampa:
@czue 
http://manage.dimagi.com/default.asp?217204#BugEvent.1211756


czue:
:+1: on :white_check_mark:

PR #13132, opened 2016-09-07
Add possibility to use POST requests in HQReportDataTables

kkrampa:
@czue Enabled this only for domains behind feature flag. What do you think about this solution?

### DIFF
--- a/corehq/apps/reports/static/reports/javascripts/config.dataTables.bootstrap.js
+++ b/corehq/apps/reports/static/reports/javascripts/config.dataTables.bootstrap.js
@@ -102,7 +102,7 @@ function HQReportDataTables(options) {
                 params.bProcessing = true;
                 params.sAjaxSource = {
                     url: self.ajaxSource,
-                    method: self.ajaxMethod
+                    method: self.ajaxMethod,
                 };
                 params.bFilter = $(this).data('filter') || false;
                 self.fmtParams = function (defParams) {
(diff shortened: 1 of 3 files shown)


corehq/apps/reports/static/reports/javascripts/config.dataTables.bootstrap.js
```diff
@@ -99,7 +100,10 @@ function HQReportDataTables(options) {
             if(self.ajaxSource) {
                 params.bServerSide = true;
                 params.bProcessing = true;
-                params.sAjaxSource = self.ajaxSource;
+                params.sAjaxSource = {
+                    url: self.ajaxSource,
+                    method: self.ajaxMethod
```
dimagimon: Missing trailing comma. (comma-dangle)

czue:
looks good to me and can't think of anything obvious this would break though am always a little nervous when touching core reports code. @kkrampa assume you have tested with the built in reports and both POSTs and GETs?

pinging @dannyroberts @biyeun @kaapstorm on interrupt in case something breaks related to this. I think ok to merge from my side once @kkrampa confirms on testing.

(is there any reason not to just switch to `POST` entirely?)

kkrampa:
@czue I checked this again and it seems to be working on my local machine. By entirely you mean all UCRs not only those behind feature flag?

czue:
I just meant with flag on and off

kkrampa:
There is no reason to do this so let me change this and test locally.

kkrampa:
@czue Updated

dannyroberts:
Seems fine, but just checking—what's the problem this resolves? I'm guessing the amount of data we need to send back to the server sometimes doesn't fit in a get param?

kkrampa:
@dannyroberts Yep, I've added UCR that has about 70 columns and it doesn't work with GET request.

czue:
> I'm guessing the amount of data we need to send back to the server sometimes doesn't fit in a get param?

yeah exactly. if reports have tons of columns it exceeds some limit

PR #13135, opened 2016-09-07
Convert core location updates to SQL

esoergel:
This on staging, open for review, but waiting on QA


esoergel:
http://manage.dimagi.com/default.asp?236915

esoergel:
Still waiting on the rest of QA before merge, but this is good to review.
@czue

corehq/apps/locations/forms.py
```diff
@@ -91,10 +92,14 @@ def __init__(self, location, bound_data=None, is_new=False, user=None,
...
-        kwargs['initial'] = dict(self.location._doc)
+        kwargs['initial'] = {
+            'parent_id': location.parent_location_id,
+            'name': location.name,
+            'site_code': location.site_code,
+            'external_id': location.external_id,
+        }
         if not self.is_new_location:
-            kwargs['initial']['location_type'] = self.location.location_type
+            kwargs['initial']['location_type'] = self.location.location_type.code
```
czue: why do most places use `name` but this one uses `code`?

corehq/apps/locations/forms.py
```diff
@@ -241,6 +246,12 @@ def clean_site_code(self):
     def clean_location_type(self):
         loc_type = self.cleaned_data['location_type']
 
+        try:
+            loc_type_obj = LocationType.objects.get(Q(code=loc_type)|Q(name=loc_type))
```
czue: i guess this is probably related to the above question (but seems a bit weird to me). why do we check both fields?

corehq/apps/locations/forms.py
esoergel: That's a great question... This should be name, I'll update it.

corehq/apps/locations/forms.py
esoergel: This one I think does make sense to check either - this form is used in bulk upload as well, which supports specifying location type either by name or code.

I have a [backlog tech-debt item](https://trello.com/c/664aSeWl/8-not-related-to-migration-make-sense-of-location-type-code-and-location-type-name-we-frequently-primarily-use-location-type-name-) to make sense of this all

corehq/apps/locations/forms.py
esoergel: oh, ha!  I'd already fixed it in another commit.

czue: nice, sorry i only made it halfway through this last night

corehq/apps/locations/forms.py
czue: interesting. doesn't the name uniqueness only based on the parent mean that this can also raise `LocationType.MultipleObjectsReturned`?

corehq/apps/locations/models.py
```diff
@@ -541,6 +541,12 @@ def location_type_object(self):
     def location_type_name(self):
         return self.location_type.name
 
+    @property
+    def sql_location(self):
+        # For backwards compatability
+        notify_of_deprecation("'sql_location' was just called on a sql_location.  That's kinda silly.")
```
czue: haha

corehq/apps/locations/forms.py
```diff
@@ -243,28 +242,35 @@ def clean_site_code(self):
 
         return site_code
 
+    def _get_allowed_types(self, parent):
```
czue: nit picky but might be easier if this took a `domain` and a `parent` and then got pulled out of the class or made static

corehq/apps/locations/forms.py
```diff
@@ -243,28 +242,35 @@ def clean_site_code(self):
...
             raise forms.ValidationError(_('The selected parent location cannot have child locations!'))
-        elif loc_type not in child_types:
-            raise forms.ValidationError(_('Location type not valid for the selected parent.'))
 
-        return loc_type
+        if not loc_type:
+            if len(allowed_types) == 1:
+                loc_type_obj = allowed_types[0]
+            else:
+                assert False, 'You must select a location type'
```
czue: i know this just moved, but are we using assertions for actual app logic? might want to change that to use custom exceptions (and if this text is user facing it should probably be translated). if the assertions are actual assertions then might just need to tweak the text a little to make it not sound like it's user-facing

corehq/apps/locations/forms.py
```diff
@@ -231,16 +232,22 @@ def clean_site_code(self):
 
         if site_code:
             site_code = site_code.lower()
+            if (SQLLocation.objects.filter(domain=self.domain,
+                                        site_code__iexact=site_code)
+                                .exclude(location_id=self.location.location_id)
+                                .exists()):
+                raise forms.ValidationError(
+                    'another location already uses this site code'
```
czue: translate

corehq/apps/locations/forms.py
```diff
@@ -267,7 +267,9 @@ def clean_location_type(self):
                 assert False, 'You must select a location type'
         else:
             try:
-                loc_type_obj = LocationType.objects.get(Q(code=loc_type) | Q(name=loc_type))
+                loc_type_obj = (LocationType.objects
+                                .filter(domain=self.domain)
```
czue: haha this is rather important

czue:
nice really excited about getting locations off couch

corehq/apps/locations/forms.py
esoergel: There are actually a bunch of untranslated strings on this class, just updated them.

corehq/apps/locations/forms.py
```diff
@@ -226,33 +230,57 @@ def clean_site_code(self):
...
+                loc_type_obj = allowed_types[0]
+            else:
+                raise forms.ValidationError(_('You must select a location type'))
+        else:
+            try:
+                loc_type_obj = (LocationType.objects
+                                .filter(domain=self.domain)
+                                .get(Q(code=loc_type) | Q(name=loc_type)))
+            except LocationType.DoesNotExist:
+                raise forms.ValidationError(_( "LocationType '{}' not found").format(loc_type))
```
dimagimon: E201 whitespace after '('

corehq/apps/locations/forms.py
esoergel: You're thinking of Locations, it looks like we don't actually enforce *any* uniqueness constraints on `LocationType`s...
That seems like a problem.  I just checked on prod, and it looks possible to add a unique_together on both `('domain', 'name')` and `('domain', 'code')`.  There's only one real domain that needs to be modified, and it looks like a mistake that there's an issue anyways.
```python
(LocationType.objects
 .values('domain', 'code')
 .order_by()
 .annotate(count=Count('pk'))
 .filter(count__gt=1))
```

As for this immediate issue, once Sravan releases the new bulk upload feature, `LocationForm` will no longer be used in bulk upload, and we can switch this to the unambiguous `location_type_pk`.

My preference for actions to take here are to leave this as-is, immediately add those `unique_together`s, and once we're no longer using this in bulk upload, switch to pk references.  How does that sound?

Sheel and I were also just talking about how the CRUD workflow around Locations needs a lot of love, and I think we'll soon do a redesign and rewrite.

esoergel:
Talbot has gone through the locations QA plan with this on staging, and there are no outstanding issues that seem related (there is [an issue](http://manage.dimagi.com/default.asp?237323) with fixtures assigned to locations not syncing properly, but that's also on prod).
This is no longer blocked on QA, so my understanding is the only outstanding issue is the [business about LocationType uniqueness](https://github.com/dimagi/commcare-hq/pull/13135#r78541055), which we may end up solving in a separate PR.

corehq/apps/locations/forms.py
czue: Sorry you're right I got confused between locations and types.

> My preference for actions to take here are to leave this as-is,
immediately add those unique_togethers, and once we're no longer using this
in bulk upload, switch to pk references. How does that sound?

sounds good to me, thanks!

esoergel:
:tada:

PR #13136, opened 2016-09-07
move custom reports above project reports

dannyroberts:
screen shot 2016-09-07 at 3 11 13 pm


czue:
nice been meaning to ask about this for ages

PR #13137, opened 2016-09-07
Bmb/deleted user fields

biyeun:
addresses several issues raised in this ticket
http://manage.dimagi.com/default.asp?236894

@dannyroberts @kaapstorm


amsagoff:
@esoergel fyi

esoergel:
@biyeun Does this make the data appear on save?  I'm guessing all that's necessary for that is for `form_valid` to redirect to the edit page rather than re-render the `get` method (at least, that's commonly the problem for that sort of issue).

PR #13139, opened 2016-09-07
explicitly list options with non-falsy defaults

nickpell:
Needed for django 1.8, since `LabelCommand.option_list` becomes empty.  See failed [test](https://travis-ci.org/dimagi/commcare-hq/jobs/157956695) for more info.

@snopoke Can you confirm that `require_only_option` still works as originally intended?

cc: @dannyroberts @sravfeyn


snopoke:
Testing the locally the list of options from Django is `settings`, `pythonpath`, `verbosity`, `traceback`, `no_color`.

I think a safer option may be to filter for known args of THIS command: https://github.com/dimagi/commcare-hq/pull/13145

PR #13140, opened 2016-09-07
sort domains by name in dropdown

dannyroberts:
Currently they're in scrambled order
screen shot 2016-09-07 at 4 26 09 pm


biyeun:
gonna merge because this looks like it's not part of any tests.

PR #13141, opened 2016-09-07
Add CaseAccessors to default imports in shell_plus

proteusvacuum:
@emord / anyone

Was finding myself pasting this in a lot locally when testing things out.


benrudolph:
👍 probably makes sense for form as well some day. no tests so just merging

PR #13142, opened 2016-09-07
check that potential None objects exist before validating against them in show_in_navigation

biyeun:
@dannyroberts @kaapstorm 

potentially the fix to http://manage.dimagi.com/default.asp?236664 though it might end up revealing the original issue. i can't seem to duplicate the problem locally. it appears to happen on production sporadically, so we'll see...


dannyroberts:
slow and steady 🐢 !

biyeun:
@dannyroberts ?!

PR #13143, opened 2016-09-07
fallback to get_by_name in project_info

biyeun:
This fixes a really nasty bug which i thought originally was my local configuration, but was actually a core issue with the project_info view in the exchange. apparently we've been inconsistently passing ids and names to the project_info reverse url method. oi.

http://manage.dimagi.com/default.asp?236825

@dannyroberts @kaapstorm

### DIFF
--- a/corehq/apps/appstore/views.py
+++ b/corehq/apps/appstore/views.py
@@ -2,6 +2,7 @@ import json
 from datetime import date
 from urllib import urlencode
 
+from couchdbkit import ResourceNotFound
 from django.contrib import messages
 from django.contrib.auth.decorators import login_required
 from django.core.urlresolvers import reverse
@@ -212,7 +213,10 @@ class ProjectInformationView(BaseCommCareExchangeSectionView):
     @property
     @memoized
     def project(self):
-        return Domain.get(self.snapshot)
+        try:
+            return Domain.get(self.snapshot)
+        except ResourceNotFound:
+            return Domain.get_by_name(self.snapshot)
 
     def dispatch(self, request, *args, **kwargs):
         if not can_view_app(request, self.project):


czue:
yuck and nice find. 

wondering if we can't fix the calls to reverse to be consistent? it feels like `self.snapshot` being either a name or an ID could lead to other problems and generally be harder to maintain.

kaapstorm:
Any reason why `snapshot_name = snapshot_id[:25]`? 

Shouldn't we also change `name` to `id` [here](https://github.com/dimagi/commcare-hq/blob/975fd0b8bcd6fe24a53bada27fa6dbfc7c7fa1b0/corehq/apps/domain/templates/domain/email/published_app_notification.txt#L3-L3) and [here](https://github.com/dimagi/commcare-hq/blob/dd9fbac389073a70ea35234cbe2360a55e44412c/corehq/apps/domain/templates/domain/email/published_app_notification.html#L2-L2)? I realise that doesn't help us with all those e-mails that have already been sent, but it might make things a little more consistent going forward.

biyeun:
@snopoke maybe simon might have an idea? this seems like a sql constraint to me...

biyeun:
✅  also

biyeun:
@kaapstorm I originally had changed `name` to `get_id`, but then I noted that many of the original usages were from way back see one from dan myung: https://github.com/dimagi/commcare-hq/blame/master/corehq/apps/domain/templates/domain/snapshot_settings.html#L139

with that in mind, I decided to address the issue at the view level, especially so that bad links out there would work again.

I think the issue here is why the `25` character constraint started being enforced

snopoke:
I'm not sure why the length is restricted to 25 specifically but we restrict it because there are SQL columns that store domain names and they have char limits (though I think the lowest is 100).

PR #13145, opened 2016-09-08
fix require_only_option for Django 1.8

snopoke:
@nickpell 

ref: https://github.com/dimagi/commcare-hq/pull/13139


nickpell:
looks good, thanks!

PR #13149, opened 2016-09-08
couch sql migrate progress + other tweaks

snopoke:
@calellowitz


corehq/apps/couch_sql_migration/diff.py
```diff
@@ -0,0 +1,183 @@
...
+
+
+def _filter_owner_id_diffs(diffs):
+    def _acceptable_owner_diff(diff):
+        return diff.path[0] == 'owner_id' and diff.old_value is None
+
+    return [
+        diff for diff in diffs
+        if not _acceptable_owner_diff(diff)
+        ]
```
dimagimon: E123 closing bracket does not match indentation of opening bracket's line

corehq/form_processor/submission_post.py
```diff
@@ -118,9 +118,8 @@ def _handle_basic_failure_modes(self):
 
     def _post_process_form(self, xform):
         self._set_submission_properties(xform)
-        if xform.is_submission_error_log:
-            found_old = scrub_meta(xform)
-            legacy_notification_assert(not found_old, 'Form with old metadata submitted', xform.form_id)
```
snopoke: Seems like this has been broken since we moved the code around. Not sure why it ended up under the `if`

corehq/apps/couch_sql_migration/diff.py
dimagimon: E123 closing bracket does not match indentation of opening bracket's line

corehq/form_processor/models.py
```diff
@@ -567,7 +567,7 @@ class CommCareCaseSQL(DisabledDbMixin, models.Model, RedisLockableMixIn,
     deleted_on = models.DateTimeField(null=True)
     deletion_id = models.CharField(max_length=255, null=True)
 
-    external_id = models.CharField(max_length=255)
+    external_id = models.CharField(max_length=255, null=True)
```
czue: why was this necessary?

corehq/apps/couch_sql_migration/constants.py
```diff
@@ -65,6 +66,10 @@ def _case_ignored_diffs():
         FormJsonDiff(
             diff_type=u'missing', path=(u'indices', u'[*]', u'doc_type'),
             old_value=u'CommCareCaseIndex', new_value=Ellipsis),
+        FormJsonDiff(diff_type=u'missing', path=(u'referrals',), old_value=[], new_value=Ellipsis),
+        FormJsonDiff(diff_type=u'missing', path=(u'location_',), old_value=[], new_value=Ellipsis),
+        FormJsonDiff(diff_type=u'type', path=(u'type',), old_value=None, new_value=u''),
```
czue: will be fun to see what this list looks like by the time things are done

corehq/apps/couch_sql_migration/couchsqlmigration.py
```diff
@@ -454,6 +454,10 @@ def _filter_form_diffs(doc_type, diffs):
     ]
     filtered = _check_deprecation_date(filtered, doc_type)
     filtered = _check_deletion_fields_date(filtered, doc_type)
+    filtered = [
+        diff for diff in filtered
+        if not (diff.diff_type == 'type' and diff.path == ('openrosa_headers', 'HTTP_X_OPENROSA_VERSION'))
```
czue: did we just stop saving these?

corehq/apps/couch_sql_migration/couchsqlmigration.py
```diff
@@ -492,6 +496,10 @@ def _filter_case_diffs(doc_type, diffs):
         if diff.path not in const.CASE_IGNORED_PATHS and diff not in const.CASE_IGNORED_DIFFS
     ]
     filtered_diffs = _check_deletion_fields_date(filtered_diffs, doc_type)
+    filtered_diffs = [
+        diff for diff in filtered_diffs
+        if diff.path[0] == 'owner_id' and diff.old_value is None
+    ]
```
czue: was the intention that only the `owner_id` ends up as a possible diff? why would that be and what's the point of the other checks?

corehq/apps/tzmigration/timezonemigration.py
```diff
@@ -193,3 +194,14 @@ def prepare_case_json(planning_db):
         planning_db.add_diffs('case', case.case_id, json_diff(case_json, case.to_json()))
 
     return planning_db
+
+
+def is_datetime(string):
```
czue: nit but this function might be better named `is_datetime_string` or something

corehq/apps/couch_sql_migration/diff.py
czue: nevermind, see this got fixed here

corehq/form_processor/backends/sql/update_strategy.py
```diff
@@ -111,6 +111,9 @@ def _apply_update_action(self, update_action):
             if key == 'location_id':
                 # special treatment of location_id
                 self.case.location_id = value
+            elif key == 'name':
+                # replicate legacy behaviour
+                self.case.name = value
```
czue: yuck

corehq/form_processor/models.py
snopoke: not completely necessary but I figured storing `None` as the empty value was better than storing an empty string

corehq/apps/couch_sql_migration/couchsqlmigration.py
snopoke: No, it's a type diff related to how the JSON is generated. In one case it's a Decimal and in the other it's a string

corehq/apps/couch_sql_migration/couchsqlmigration.py
snopoke: This is to cover legacy data where a case would have a blank owner ID if it never got set in the form. We've since made it so that the owner ID is set to the user who submitted the form (unless otherwise specified).

corehq/form_processor/backends/sql/update_strategy.py
snopoke: yea, this is a bit odd. The real case name property is called `case_name` so technically if you set the `name` property then you just get a dynamic case property called `name`.

I'm not sure if the old functionality was intentional but it works because we just set dynamic case props as attributes on the couch case object so setting `name` is the same as changing the case name property.

Either way I figured if a user is setting a property called `name` they are likely intending to set the case name.

corehq/apps/couch_sql_migration/constants.py
snopoke: updated with a more complete list now

corehq/apps/couch_sql_migration/diff.py
```diff
@@ -105,14 +115,29 @@ def filter_form_diffs(doc_type, diffs):
...
     filtered = _filter_date_diffs(filtered)
     return filtered
 
 
+def _filter_partial_diffs(diffs, partial_diffs_to_exclude):
+    """Filter out diffs that match a subset of attributes
+    :type partial_diffs_to_exclude: dict([(attr, value)...])
+    """
+    def _partial_match(diff):
+        print diff
```
czue: print

scent.py
```diff
@@ -13,8 +13,13 @@
...
 @file_validator
-def py_files(filename):
-    return filename.endswith('.py') and not os.path.basename(filename).startswith('.')
+def python_test_files(filename):
+    return (
+        filename.endswith('.py') or
+        filename.endswith('.json') or
+        filename.endswith('.xml') and
+        not os.path.basename(filename).startswith('.')
+    )
```
czue: this is amazing

czue:
no blockers, feel free to merge on :white_check_mark:

PR #13154, opened 2016-09-08
adding management command to restore es from snapshot

calellowitz:
@esoergel @benrudolph @czue @snopoke 
This management command can be used to restore elasticsearch from a snapshot. It wasnt clear to me whether it belonged in ansible or the hq repo (since we dont really manage scripts like this in ansible but the rest of the work is there). Happy to move to the ansible repo (maybe its own role like rolling restart?) if that seems cleaner.
Ansible work at https://github.com/dimagi/commcarehq-ansible/pull/524


corehq/apps/es/management/commands/restore_es_snapshot.py
```diff
@@ -0,0 +1,38 @@
+from django.core.management.base import BaseCommand, CommandError
+from datetime import datetime, timedelta
+from corehq.elastic import get_es_new
+from elasticsearch.client import SnapshotClient, IndicesClient
+from django.conf import settings
+
+class Command(BaseCommand):
```
dimagimon: E302 expected 2 blank lines, found 1

corehq/apps/es/management/commands/restore_es_snapshot.py
```diff
@@ -0,0 +1,38 @@
...
+
+    def handle(self, *args, **options):
+        print "Restoring ES indices from snapshot"
+        if len(args) != 1:
+            raise CommandError('Usage is restore_es_snapshot %s' % self.args)
+        date = self.process_arguments(args)
+        es = get_es_new()
+        self.close_indices(es)
+        self.restore_snapshot(es, date)
+
```
dimagimon: W293 blank line contains whitespace

snopoke:
Seems OK to have this live here. It is possible to restore only a single index or do we always have to restore all of them?

calellowitz:
The api supports restoring a single index. I can add an optional argument to restore only a specific index or indices.

corehq/apps/es/management/commands/restore_es_snapshot.py
```diff
@@ -0,0 +1,53 @@
...
+        snapshot_client = SnapshotClient(es)
+        env = settings.SERVER_ENVIRONMENT
+        repo_name = '{}_es_snapshot'.format(env)
+        snapshot_client.restore(repo_name,
+                                '{repo_name}_{year}_{month}_{day}'.format(
+                                    repo_name=repo_name, year=date.year,
+                                    month=date.month, day=date.day
+                                ),
+                                body={'indices': indices}
+        )
```
dimagimon: E124 closing bracket does not match visual indentation

calellowitz:
@snopoke 
Added option to restore specific indices

corehq/apps/es/management/commands/restore_es_snapshot.py
```diff
@@ -0,0 +1,55 @@
...
+        restore_date = (datetime.utcnow() - timedelta(days=days_ago))
+        return restore_date
+
+    @staticmethod
+    def get_indices(args):
+        if len(args) > 1:
+            indices = ','.join(args[1:])
+        else:
+            indices = '_all'
+        return indices
```
esoergel: Does this require the full index name, or are aliases acceptable?

corehq/apps/es/management/commands/restore_es_snapshot.py
```diff
@@ -0,0 +1,55 @@
...
+
+    def handle(self, *args, **options):
+        print "Restoring ES indices from snapshot"
+        if len(args) < 1:
+            raise CommandError('Usage is restore_es_snapshot %s' % self.args)
+        date = self.get_date(args)
+        indices = self.get_indices(args)
+        es = get_es_new()
+        client = self.get_client_and_close_indices(es, indices)
+        self.restore_snapshot(es, date, indices)
```
esoergel: Might want to verify that the snapshots exist before deleting the old indices

esoergel:
Looks good overall, but I'd add in some more sanity checks, things like:
> This will remove the following indices: ...
Restoring from backup made on <date>
Found snapshot <>
Are you sure you'd like to continue?

corehq/apps/es/management/commands/restore_es_snapshot.py
calellowitz: aliases work

corehq/apps/es/management/commands/restore_es_snapshot.py
```diff
@@ -0,0 +1,81 @@
+from django.core.management.base import BaseCommand, CommandError
+from datetime import date, timedelta
+from corehq.elastic import get_es_new
+from elasticsearch.client import SnapshotClient, IndicesClient
+from django.conf import settings
+from pillowtop.utils import get_all_pillow_instances
+from corehq.apps.hqadmin.models import ESRestorePillowCheckpoints
+from pillowtop.es_utils import get_index_info_from_pillow
+
+class Command(BaseCommand):
```
dimagimon: E302 expected 2 blank lines, found 1

calellowitz:
This should finally be good now. I realized we had to push back pillow checkpoints a bit ago and was trying to figure out how to do that.
After a discussion with @snopoke I decided to just do daily pillow checkpoint snapshots (based on cron job scheduling these should be a few hours before the es index snapshot occurs). I keep these for 3 days, which is the same length of time as the es snapshots.

calellowitz:
cc: @esoergel as well

corehq/apps/es/management/commands/restore_es_snapshot.py
```diff
@@ -0,0 +1,85 @@
...
+        env = settings.SERVER_ENVIRONMENT
+        repo_name = '{}_es_snapshot'.format(env)
+        snapshot_client.restore(repo_name,
+                                '{repo_name}_{year}_{month}_{day}'.format(
+                                    repo_name=repo_name, year=date.year,
+                                    month=date.month, day=date.day
+                                ),
+                                wait_for_completion=True,
+                                body={'indices': indices}
+        )
```
dimagimon: E124 closing bracket does not match visual indentation

calellowitz:
added the hold off on merge tag so i can do some further testing

corehq/apps/hqadmin/tasks.py
```diff
@@ -1,9 +1,20 @@
...
 def pillow_seq_store_task():
     pillow_seq_store()
+
+
+@periodic_task(run_every=crontab(hour=0, minute=0), queue='background_queue')
+def create_es_snapshot_checkpoints():
+    today = date.today()
+    three_days_ago = today - timedelta(days=3)
+    ESRestorePillowCheckpoints.get_pillow_snapshots()
+    ESRestorePillowCheckpoints.objects.filter(date_updated__lt=three_days_ago).delete()
```
czue: wonder if we want to make this a bit longer? it seems very low-cost to save, e.g. a month and sometimes we get into a situation with a cloudant rewind where we need something older than 3 days....

calellowitz: upped to 30 days. the 3 days was because thats how long we kept the snapshots but it seems like there is no downside to keeping more

calellowitz:
This is fully tested on staging so i removed the hold off on merge tag
@snopoke @czue 
The one major change to functionality i have made since then is that now all pillows are reset to the old checkpoints, not just those corresponding to restored indices since it does not seem like there is a good way to find all pillows that interact with each index. Im guessing they can catch up pretty quickly so it should not be a huge problem but I have not been able to test it on prod pillows

corehq/apps/es/management/commands/restore_es_snapshot.py
```diff
@@ -0,0 +1,78 @@
...
+    args = "days_ago <index_1> <index_2> ..."
+
+    def handle(self, *args, **options):
+        print "Restoring ES indices from snapshot"
+        if len(args) < 1:
+            raise CommandError('Usage is restore_es_snapshot %s' % self.args)
+        date = self.get_date(args)
+        indices = self.get_indices(args)
+        es = get_es_new()
+        client = self.get_client_and_close_indices(es, indices)
```
snopoke: Should add in a user confirmation here before going ahead, especially since the default is to close ALL indices.

corehq/apps/es/management/commands/restore_es_snapshot.py
```diff
@@ -0,0 +1,78 @@
...
+        for pillow in get_all_pillow_instances():
+            checkpoint = pillow.checkpoint
+            try:
+                checkpoint = ESRestorePillowCheckpoints.objects.get(checkpoint_id=checkpoint.checkpoint_id,
+                                                                    date_updated=date)
+                seq = checkpoint.seq
+            except ESRestorePillowCheckpoints.DoesNotExist:
+                seq = DEFAULT_EMPTY_CHECKPOINT_SEQUENCE
+
+            pillow.checkpoint.update_to(seq)
```
snopoke: I think this requires a pillow restart - would probably be good to tell the user that they need to stop all pillows prior to doing a restore and start them again afterwards.

corehq/apps/hqadmin/migrations/0007_esrestorepillowcheckpoint_datefield.py
```diff
@@ -0,0 +1,19 @@
...
+    dependencies = [
+        ('hqadmin', '0006_esrestorepillowcheckpoints'),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name='esrestorepillowcheckpoints',
+            name='date_updated',
+            field=models.DateField(),
+        ),
```
snopoke: Can combine this with the previous migration (each separate migration has a significant overhead)

corehq/apps/hqadmin/models.py
```diff
@@ -65,6 +68,21 @@ def get_by_pillow_name(cls, pillow_name):
         return store
 
 
+class ESRestorePillowCheckpoints(models.Model):
+    seq = models.TextField()
+    checkpoint_id = models.CharField(max_length=255, db_index=True)
+    date_updated = models.DateField()
+
+    @classmethod
+    def get_pillow_snapshots(cls):
```
snopoke: Since this doesn't return anything I'd rename to `create_pillow_checkpoing_snapshots` or event `snapshot_pillow_checkpoints`

snopoke (changes requested):
Generally looks good.

corehq/apps/hqadmin/migrations/0007_esrestorepillowcheckpoint_datefield.py
calellowitz: If this is already on staging is there a safe way to make it one migration without breaking staging deploys?

snopoke:
hmm, no. In that case fine to leave as is

calellowitz:
@snopoke updated

corehq/apps/hqadmin/models.py
```diff
@@ -74,7 +74,7 @@ class ESRestorePillowCheckpoints(models.Model):
     date_updated = models.DateField()
 
     @classmethod
-    def get_pillow_snapshots(cls):
+    def create_pillow_checkpoint_snaphots(cls):
```
snopoke: nit: typo `snaphots`

snopoke (approved):

corehq/apps/hqadmin/models.py
calellowitz: fixed

calellowitz:
✅

PR #13160, opened 2016-09-08
re-include pact reports

nickpell:
accidentally removed in https://github.com/dimagi/commcare-hq/pull/13125/commits/e435b340b92ac665ba0dbc48c70d8d85bb5dc779


nickpell:
self-merging in order to get fix deployed

PR #13161, opened 2016-09-08
fix 500s when custom reports are not found

nickpell:


nickpell:
self-merging and deploying

PR #13162, opened 2016-09-09
allow copying advanced forms

snopoke:
@calellowitz 

cc @phillipm

### DIFF
--- a/corehq/apps/app_manager/models.py
+++ b/corehq/apps/app_manager/models.py
@@ -2168,6 +2168,9 @@ class ModuleBase(IndexedSchema, NavMenuItemMediaMixin, CommentMixin):
     def uses_usercase(self):
         return False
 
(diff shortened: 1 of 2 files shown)


czue:
looks good on :+1:

corehq/apps/app_manager/models.py
```diff
@@ -2168,6 +2168,9 @@ def uses_media(self):
     def uses_usercase(self):
         return False
 
+    def add_insert_form(self, from_module, form, index=None, with_source=False):
+        raise IncompatibleFormTypeException()
```
czue: just for my own education why was this necessary?

snopoke: Not striclty necessary but this method wasn't implemented by all the module subclasses so this is just a safety

PR #13165, opened 2016-09-09
fix 500ing for ils-gateway domain

nickpell:
http://manage.dimagi.com/default.asp?237038


czue:

nickpell:
Tests passed

PR #13167, opened 2016-09-09
Dsr multi loc

sravfeyn:


corehq/apps/users/models.py
```diff
@@ -1,6 +1,7 @@
 from __future__ import absolute_import
 from datetime import datetime
 from dateutil.relativedelta import relativedelta
+import copy
```
dimagimon: F401 'copy' imported but unused

sravfeyn:
Hey @esoergel can you review this. There are still some todos, but this is in a good shape for staging

corehq/apps/users/models.py
```diff
@@ -382,6 +382,7 @@ class DomainMembership(Membership):
     override_global_tz = BooleanProperty(default=False)
     role_id = StringProperty()
     location_id = StringProperty()
+    assigned_location_ids = StringListProperty()
```
esoergel: Does this also apply to web users?

corehq/apps/users/forms.py
```diff
@@ -745,14 +751,17 @@ class CommtrackUserForm(forms.Form):
...
+            self.domain = kwargs['domain']
             del kwargs['domain']
         super(CommtrackUserForm, self).__init__(*args, **kwargs)
-        self.fields['primary_location'].widget = SupplyPointSelectWidget(domain=domain)
-        if Domain.get_by_name(domain).commtrack_enabled:
-            programs = Program.by_domain(domain, wrap=False)
+        self.fields['assigned_locations'].widget = SupplyPointSelectWidget(
+            domain=self.domain, multiselect=True, id='id_assigned_locations'
+        )
+        self.fields['primary_location'].widget = SupplyPointSelectWidget(domain=self.domain)
```
esoergel: Very nice!

corehq/apps/users/forms.py
```diff
@@ -768,13 +777,70 @@ def __init__(self, *args, **kwargs):
...
+            old_location_ids = domain_membership.assigned_location_ids
+            if set(location_ids) != set(old_location_ids):
+                if location_ids:
+                    user.assign_to_locations(self.domain, location_ids)
+                else:
+                    user.unassign_all_locations(self.domain)
+
+    def clean_assigned_locations(self):
+        # select2 (< 4.0) doesn't format multiselect for remote data as an array
+        #   but formats it as comma-seperated list, so we need to clean the returned data
```
esoergel: that's annoying

corehq/apps/users/forms.py
```diff
@@ -768,13 +777,70 @@ def __init__(self, *args, **kwargs):
...
+        from corehq.apps.locations.models import SQLLocation
+        from corehq.apps.locations.util import get_locations_from_ids
+
+        value = self.cleaned_data.get('assigned_locations')
+        if not isinstance(value, basestring) or value.strip() == '':
+            return []
+
+        location_ids = [location_id.strip() for location_id in value.split(',')]
```
esoergel: Any chance select2 could give empty values or use a trailing comma here?  If so, it might be prudent to wrap this in a `filter(None, ...)` to drop null values.

corehq/apps/users/models.py
```diff
@@ -1678,6 +1678,16 @@ def unset_location(self):
...
+    def assign_to_locations(self, location_ids):
+        self.assigned_location_ids = location_ids
+        self.get_domain_membership(self.domain).assigned_location_ids = location_ids
+        self.save()
+
+    def unassign_all_locations(self):
+        self.assigned_location_ids = []
+        self.get_domain_membership(self.domain).assigned_location_ids = []
+        self.save()
+
```
esoergel: If you made these methods accept `domain` as a parameter, you could use the same methods for both `WebUser`s and `CommCareUser`s (and probably stick it in `CouchUser`)

corehq/apps/users/forms.py
```diff
@@ -869,6 +869,15 @@ def clean_assigned_locations(self):
...
         return [location.location_id for location in locations]
 
+    def clean(self):
+        cleaned_data = super(CommtrackUserForm, self).clean()
+
+        primary_location_id = cleaned_data['primary_location']
+        assigned_location_ids = cleaned_data['assigned_locations']
+        if primary_location_id:
+            if primary_location_id not in assigned_location_ids:
+                raise ValidationError(_("Only one of the user's locations can be a primary location"))
```
esoergel: Is there way to attach this error to the `primary_location` form field? I believe as-is, it'll show up as a form-level error.

corehq/apps/users/management/commands/add_multi_location_property.py
```diff
@@ -0,0 +1,41 @@
...
+logger.setLevel('DEBUG')
+
+
+class Command(BaseCommand):
+    args = ""
+    help = ("")
+
+    def handle(self, *args, **options):
+        self.options = options
+        user_ids = self.get_user_ids()
```
esoergel: Quick plug for `with_progress_bar` - if you change this line to
```python
user_ids = with_progress_bar(self.get_user_ids())
```
you'll get a progress bar showing expected time remaining

corehq/apps/users/management/commands/add_multi_location_property.py
```diff
@@ -0,0 +1,41 @@
...
+from corehq.apps.es import UserES, users as user_filters
+from corehq.apps.users.models import CouchUser
+from corehq.util.couch import iter_update, DocUpdate
+
+
+logger = logging.getLogger('user_migration')
+logger.setLevel('DEBUG')
+
+
+class Command(BaseCommand):
```
esoergel: This command looks solid

corehq/apps/users/management/commands/add_multi_location_property.py
```diff
@@ -0,0 +1,41 @@
...
+        })
+        return DocUpdate(doc)
+
+    def get_user_ids(self):
+        res = (UserES()
+               .OR(user_filters.web_users(), user_filters.mobile_users())
+               .non_null('location_id')
+               .fields(['_id'])
+               .run().hits)
+        return [r['_id'] for r in res]
```
esoergel: You might want `exclude_source` and `doc_ids` here:
```python
def get_user_ids(self):
    res = (UserES()
           .OR(user_filters.web_users(), user_filters.mobile_users())
           .non_null('location_id')
           .exclude_source()
           .run())
    return res.doc_ids
```

corehq/apps/users/models.py
```diff
@@ -1659,62 +1659,88 @@ def set_location(self, location):
...
+            self.get_domain_membership(self.domain).assigned_location_ids.append(self.location_id)
+            self.user_data['commcare_location_ids'] = ','.join(self.assigned_location_ids)
+        self.save()
 
     def unset_location(self):
         """
-        Unset the primary location and remove all associated user data and cases
+        Resets primary location to next available location from assigned_location_ids.
+            If there are no more locations in assigned_location_ids,
+            then the primary location and user data are cleared
```
esoergel: This seems like confusing behavior, what's the idea behind it?  Would it be possible to remove this function entirely and instead use `unset_location_by_id` and something like `unset_all_locations` or `reset_locations([])`?

corehq/apps/es/users.py
```diff
@@ -109,7 +109,8 @@ def user_ids(user_ids):
     return filters.term("_id", list(user_ids))
 
 
-def location(location_id):
+def primary_location(location_id):
+    # by primary location
```
esoergel: actually I imagine most things currently using this filter would prefer to return users where any of their assigned locations match the location(s) passed to the filter.

That is, any users where `user.location_ids` intersects at all with `location_ids`

esoergel: looks like you added that already in a later commit.  I think the project health dashboard should also be updated to that filter, though.

corehq/apps/es/users.py
```diff
@@ -118,3 +119,14 @@ def primary_location(location_id):
...
+
+def location(location_id):
+    # by any assigned-location primary or not
+    return filters.OR(
+        filters.AND(mobile_users(), filters.term('assigned_location_ids', location_id)),
+        filters.AND(
+            web_users(),
+            filters.term('domain_memberships.assigned_location_ids', location_id)
+        ),
+    )
```
esoergel: Did you confirm that this works as expected when the passed in `location_id` is a list of location ids?  That is, will
```python
query.location(['123', '456'])
```
Match a user where
```python
location_ids = ['abc', '123']
```
I'm not sure how the `terms` filter compares two lists.

corehq/pillows/mappings/user_mapping.py
```diff
@@ -31,6 +31,7 @@
                                                 'type': 'string'}},
                            'type': 'multi_field'},
                 'location_id': {'index': 'not_analyzed', 'type': 'string'},
```
esoergel: shouldn't this name also be changed to `primary_location_id`?

corehq/apps/locations/fixtures.py
```diff
@@ -141,10 +141,9 @@ def get_all_locations_to_sync(user):
     else:
         all_locations = set()
 
-        user_location = user.sql_location
-        user_locations = set([user_location]) if user_location is not None else set()
```
esoergel: What if the user isn't assigned to a location?  Will that be short-circuited elsewhere?

corehq/apps/api/auth.py
```diff
@@ -0,0 +1,190 @@
...
+        return super(HQThrottle, self).should_be_throttled(identifier, **kwargs)
+
+
+class CustomResourceMeta(object):
+    authorization = ReadOnlyAuthorization()
+    authentication = LoginAndDomainAuthentication()
+    serializer = CustomXMLSerializer()
+    default_format = 'application/json'
+    throttle = HQThrottle(throttle_at=getattr(settings, 'CCHQ_API_THROTTLE_REQUESTS', 25),
+                          timeframe=getattr(settings, 'CCHQ_API_THROTTLE_TIMEFRAME', 15))
```
esoergel: Is this a bad merge or something?  What's this file doing here?

esoergel: nevermind, I see you later removed it.

esoergel:
I had a number of comments, but overall it looks good.  Lemme know when it's on staging and I'll play around with it.

corehq/apps/locations/fixtures.py
sravfeyn: The new substitute `user.sql_locations` returns empty list if there are not assigned locations

corehq/pillows/mappings/user_mapping.py
sravfeyn: I didn't change the property name on couch model, so it shouldn't change here, yeah?

corehq/apps/es/users.py
sravfeyn: Good point! Have just tested it and it works.

corehq/apps/es/users.py
sravfeyn: (done)

corehq/apps/users/management/commands/add_multi_location_property.py
sravfeyn: :+1

corehq/apps/users/forms.py
sravfeyn: Since this validation is related to two fields, the error can't be captured in a field specific clean method, that's why I had to put this here

corehq/apps/users/forms.py
sravfeyn: select2 doesn't pass empty values as the options itself won't be empty

corehq/apps/users/models.py
sravfeyn: Yes (?), as is the usual location_id. I took this as implied, should I ping Sheel for a spec?

corehq/apps/users/models.py
sravfeyn: Yeah, I don't like this duplicate code either, but it's hard to reuse it without using some hacky function/attribute-name inspect tooling, as the two methods differ, both in the way they get the attribute `assigned_location_ids` and the attributes that need to be updated. Keeping them separate is more cleaner.

corehq/apps/users/models.py
sravfeyn: The function is used by `unset_location_by_id` when the specified location is primary location. The idea behind falling back is that the user can never have non empty `assigned_location_ids` and null `location_id` (from the spec), in subsequent commits I have added that behaviour as a kwarg.

I agree the behaviour is confusing and I added a note saying that calling this directly should be avoided. One of the main reasons I kept is to not break old code.

corehq/apps/users/models.py
esoergel: Maybe good to confirm, but it's probably best to do it now, while web users' locations aren't used for much.

corehq/apps/users/models.py
esoergel: You could also have different methods on each class, but have them accept the same parameters, to avoid having to do `if user.is_commcare_user()` each time in the code where the updates are performed (although I'm not actually sure how often that happens).  IDK, your call.

corehq/apps/users/forms.py
esoergel: You can attach the error to a specific field, though.  I wasn't actually sure if that's possible, but it looks like Django supports that in this form:
```python
self.add_error('primary_location',
               _("Only one of the user's locations can be a primary location"))
```
https://docs.djangoproject.com/en/1.10/ref/forms/api/#django.forms.Form.add_error

corehq/apps/users/models.py
esoergel: Makes sense.  There's a lot of baggage with locations that forces situations like this.  You could prefix it with an underscore if you want to deter usage, but this is also okay.

corehq/pillows/mappings/user_mapping.py
esoergel: oh, you're right, I thought that name had changed.  Yeah, the two should match

esoergel:
Signing off on all changes up through b7e1256ed2cdf32d6fa445221bc5ed33649570ed "another test"

corehq/apps/users/models.py
sravfeyn: @sheelio Shouldn't web-users also have the multiple-locations field (they already have the location_id field)?

sheelio: Hey Sravan - I think we should design for web users having multiple locations (but I don't think the UI allows for it yet).  What is the lift to make that change for web users?

sravfeyn: I have implemented it already, but haven't exposed the UI as there was no UI for location currently. It will be easier to add a UI latter when needed.

corehq/apps/users/tests/test_location_assignment.py
```diff
@@ -0,0 +1,278 @@
...
+    (
+        web_user('a', ['a']),
+        web_user('a', ['a']),
+        False
+    ),
+    (
+        web_user('a'),
+        web_user('a', ['a']),
+        True
+    ),    (
```
dimagimon: E241 multiple spaces after ','

corehq/apps/users/forms.py
```diff
@@ -884,9 +884,11 @@ def clean(self):
         assigned_location_ids = cleaned_data['assigned_locations']
         if primary_location_id:
             if primary_location_id not in assigned_location_ids:
-                raise ValidationError(_("Only one of the user's locations can be a primary location"))
+                self.add_error('primary_location',
+                               _("Primary location can only be one of user's locations"))
         if assigned_location_ids and not primary_location_id:
-            raise ValidationError(_("Primary location can't be empty if user has any locations set"))
+            self.add_error('primary_location',
+                           _("Primary location can't be empty if user has any locations set"))
```
esoergel: :+1:

corehq/apps/users/util.py
```diff
@@ -171,3 +171,8 @@ def parts():
             yield u' "%s"' % html.escape(full_name)
 
     return safestring.mark_safe(''.join(parts()))
+
+
+def user_location_data(location_ids):
+    # Spec for 'commcare_location_ids' custom data field
+    return ' '.join(location_ids)
```
esoergel: Good call abstracting out this functionality.

corehq/apps/users/tests/test_location_assignment.py
```diff
@@ -194,3 +198,63 @@ def assertNonPrimaryLocation(self, expected):
...
+        web_user('a', ['a'])
+    ),
+    (
+        web_user('a', ['b']),
+        web_user('a', ['b', 'a'])
+    ),
+])
+def test_migration(self, user, expected):
+    actual = Command().migrate_user(user).doc
+    self.assertEqual(actual, expected)
```
esoergel: :star:

corehq/apps/users/management/commands/add_multi_location_property.py
```diff
@@ -24,22 +24,39 @@ def migrate_user(self, doc):
 
     def migrate_cc_user(self, doc):
 
-        if is_already_migrated(doc):
+        # skip if doesn't have location
+        if not doc['location_id']:
             return
 
+        # skip if already migrated
```
esoergel: I'm not sure this'd be a perfect fit for what you're doing here, but just FYI, if a document is unchanged after being sent through the migration function, `iter_update` won't save it.  So it might be doing essentially this same logic there anyways.
https://github.com/dimagi/commcare-hq/blob/d8262b2cc0ed5c876566f87e21c2fa5139fa129d/corehq/util/couch.py#L301-L301

If a non-trivial amount of time is spent in the `migrate_user` call, it's certainly better to skip here.

esoergel:
Reviewed up through 77db5c3 "correct filters to find users", looks good, nothing blocking there.  Tests are failing with:
```
FAIL: test_location_append (corehq.apps.users.tests.test_location_assignment.CCUserLocationAssignmentTest)
  File "/mnt/commcare-hq-ro/corehq/apps/users/tests/test_location_assignment.py", line 53, in test_location_append
    self.assertAssignedLocations(self.loc_ids)
  File "/mnt/commcare-hq-ro/corehq/apps/users/tests/test_location_assignment.py", line 105, in assertAssignedLocations
    self.assertListEqual(actual_ids, expected_location_ids)
AssertionError: Lists differ: [u'cbba0278dc40a0b5fd590760f71... != [u'cbba0278dc40a0b5fd590760f71...
First differing element 0:
u'cbba0278dc40a0b5fd590760f7186086 cbba0278dc40a0b5fd590760f7185a4f'
u'cbba0278dc40a0b5fd590760f7186086'
Second list contains 1 additional elements.
First extra element 1:
u'cbba0278dc40a0b5fd590760f7185a4f'
- [u'cbba0278dc40a0b5fd590760f7186086 cbba0278dc40a0b5fd590760f7185a4f']
?                                    ^
+ [u'cbba0278dc40a0b5fd590760f7186086', u'cbba0278dc40a0b5fd590760f7185a4f']
?                                    ^^^^^
```

sravfeyn:
Hey @esoergel QA has passed on this. This can be merged now. I will run the management command after a deploy, for now I added migration in the CommCareUser.wrap method as well, so that a delay in running management command doesn't affect existing setup

sravfeyn:
✅

corehq/apps/users/views/mobile/users.py
```diff
@@ -226,6 +226,10 @@ def page_context(self):
             'demo_restore_date': naturaltime(demo_restore_date_created(self.editable_user)),
             'hide_password_feedback': settings.ENABLE_DRACONIAN_SECURITY_FEATURES
         }
+        if self.commtrack_form.errors:
+            messages.error(self.request, _(
+                "There were some errors while saving user's locations. Please check 'Locations' tab"
+            ))
```
esoergel: :+1:

corehq/apps/users/models.py
```diff
@@ -1623,13 +1630,34 @@ def sql_location(self):
                 pass
         return None
 
+    @property
+    def sql_locations(self):
+        from corehq.apps.locations.models import SQLLocation
+        if self.assigned_location_ids:
+            try:
+                return SQLLocation.objects.filter(location_id__in=self.assigned_location_ids)
+            except SQLLocation.DoesNotExist:
```
proteusvacuum: @sravfeyn 
Noticed this when looking at https://github.com/dimagi/commcare-hq/pull/14019#discussion_r89429985
I don't think `filter` ever throws a `DoesNotExist`, but instead returns an empty queryset.

corehq/apps/users/models.py
```diff
@@ -1623,13 +1630,34 @@ def sql_location(self):
...
+                pass
+        return []
```
proteusvacuum: See: 
https://github.com/dimagi/commcare-hq/pull/14019#discussion_r89429985
I think this should be `SQLLocation.objects.none()` so this can be chainable with other queryset methods.

sravfeyn:
@proteusvacuum Fixed in https://github.com/dimagi/commcare-hq/pull/14020/files

PR #13168, opened 2016-09-09
Give a better error message

kaapstorm:
Context: http://manage.dimagi.com/default.asp?236239

Labelling "Hold off" because I'm not sure this the right approach. It could be putting on a bandaid instead of fixing the problem at the source. But it does allow the app to build without editing the form.

@orangejenny @dannyroberts @biyeun


kaapstorm:
Kicking build

kaapstorm:
:white_check_mark: but want to confirm this is the right approach.

kaapstorm:
Dropping "Hold off" because partner is waiting.

emord:
Does this fix the validation if you upload xml with `form="default"`? I thought that we just sent the raw xml to the form validator jar, but could be wrong.

kaapstorm:
It does fix the validation, but it doesn't change the form -- so the form will end up on the device with `form="default"`.

emord:
from @ctsims in the ticket:
 
>Clayton Sims [3:19 PM]  in an <itext> form="default" woudl be bad, IIRC.
 >[3:19] I don't think we treat it specially in any way, so it wouldn't actually be the default

 That sounds like not the correct behavior. We should remove it from the form totally and if we don't we should have a validation message saying the form won't work.

kaapstorm:
**Plan B**: Don't try to fix it. Just give the user a better error message:
better_error_msg

corehq/apps/app_manager/xform.py
```diff
@@ -19,6 +19,9 @@
 import re
 
 
+VALID_VALUE_FORMS = ('image', 'audio', 'video', 'video-inline')
```
emord: `markdown` is also valid

emord:
:+1: on :white_check_mark:

benrudolph:
would be very useful to include the form name for this error

kaapstorm:
I went ahead and did that ... and then noticed that on the Deploy page HQ already tells you what form the error is in:
error_on_deploy

And, of course, when the error is shown on the Form page (screenshot above) you already know what form the error is in -- it's the form you're looking at.

I've stashed the change. If you still think it's worth pushing, just let me know and I will. It looks like this:
error_with_form_name

kaapstorm:
:boot: :construction_worker:

benrudolph:
Sorry norm! Thanks for looking into it though

kaapstorm:
No worries @benrudolph. I thought it would be a good idea too.

PR #13034, opened 2016-08-29
Create location-based access whitelist

esoergel:
This PR sets up a framework for whitelisting parts of HQ via a view decorator.  Anything not explicitly decorated will be off-limits to someone with their access restricted.

There is a new user roles permission called "Full Organization Access (Allow role to access data from all locations)".  This is checked by default for all roles.  When a user is added to a role with this un-checked, they can only access whitelisted parts of HQ, and only if they're assigned to a location.

I switched around the specifics of the whitelist a couple times as I ran into issues with earlier implementations, so it may be easiest just to review in one chunk.  Sorry, I tried rebasing, but it was pretty complicated.

Things to look for in review:
- [ ] Middleware correctly blocks all non-whitelisted views(ie there aren't any loopholes)
- [ ] Toggles UI correctly displays only accessible pages in the sidebar and menu bar
- [ ] Whitelisted pages actually do restrict access where appropriate.
- [ ] Nothing about the above impacts unrestricted users (as of now, ALL existing users are unrestricted)
- [ ] Nothing about the above makes the site unusable for the restricted users.

@proteusvacuum @sravfeyn

I'm currently testing on staging.  Feel free to invite `yourname+test@ourdomain.com` on the `ethan-test` domain if you want to try it out.

PR #13169, opened 2016-09-09
Fix accepting domain invitation

gcapalbo:
http://manage.dimagi.com/default.asp?236664

Introduced with https://github.com/dimagi/commcare-hq/pull/13034

Looks like some of the new additions to [corehq.tabs.uitab](https://github.com/dimagi/commcare-hq/blob/e954aaddd5d9f6d1517b3ad16fb9d636d167a19a/corehq/tabs/uitab.py) create a dependency on request.project being present. request.project is normally added with the login_and_domain_required decorator which isn't used in the domain invitation view.

@benrudolph code buddy
fyi @esoergel @biyeun 

@biyeun should we revert https://github.com/dimagi/commcare-hq/pull/13098 and https://github.com/dimagi/commcare-hq/pull/13142 or are those still good changes to have?

### DIFF
--- a/corehq/apps/users/views/__init__.py
+++ b/corehq/apps/users/views/__init__.py
@@ -37,7 +37,8 @@ from corehq.apps.analytics.tasks import (
     track_existing_user_accepted_invite_on_hubspot,
 )
 from corehq.apps.analytics.utils import get_meta
-from corehq.apps.domain.decorators import (login_and_domain_required, require_superuser, domain_admin_required)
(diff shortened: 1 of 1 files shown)


biyeun:
@gcapalbo i think those are still good changes to have, since the method allows for `project` to be `None`

dannyroberts:
merging since I believe it's not tested

gcapalbo:
offlined with Ethan, a better solution is here: https://github.com/dimagi/commcare-hq/pull/13195

PR #13175, opened 2016-09-09
move cvsulive custom reports to reports.py

nickpell:


nickpell:
@sravfeyn @dannyroberts

PR #13183, opened 2016-09-09
use request.can_access_location

esoergel:
@gcapalbo

### DIFF
--- a/corehq/tabs/uitab.py
+++ b/corehq/tabs/uitab.py
@@ -66,12 +66,9 @@ class UITab(object):
         return self._request.get_full_path()
 
     @property
-    def restrict_access_by_location(self):
-        """Is this a web user who can't access project-wide data?"""
-        return (
-            self.couch_user and self.domain
-            and not self.couch_user.has_permission(self.domain, 'access_all_locations')
-        )
+    def can_access_all_locations(self):
+        """Is this a web user who can access project-wide data?"""
+        return getattr(self._request, 'can_access_all_locations', True)
 
     @property
     def dropdown_items(self):
@@ -80,7 +77,7 @@ class UITab(object):
 
     @property
     def filtered_dropdown_items(self):
-        if not self.restrict_access_by_location:
+        if self.can_access_all_locations:
             return self.dropdown_items
 
         filtered = []
@@ -100,7 +97,7 @@ class UITab(object):
     @property
     @memoized
     def filtered_sidebar_items(self):
-        if not self.restrict_access_by_location:
+        if self.can_access_all_locations:
             return self.sidebar_items
 
         filtered = []
@@ -127,7 +124,8 @@ class UITab(object):
         if not self.show_by_default and not self.is_active_tab:
             return False
 
-        if self.restrict_access_by_location and not self.filtered_dropdown_items:
+        if not self.can_access_all_locations and not self.filtered_dropdown_items:
+            # location-safe filtering makes this whole tab inaccessible
             return False
 
         try:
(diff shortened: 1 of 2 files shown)


corehq/tabs/uitab.py
```diff
@@ -127,7 +124,8 @@ def should_show(self):
         if not self.show_by_default and not self.is_active_tab:
             return False
 
-        if self.restrict_access_by_location and not self.filtered_dropdown_items:
```
esoergel: `self.restrict_access_by_location` was incorrectly evaluating to True because the users weren't yet members of the domain.  I've switched it to use `request.can_access_all_locations`, which handles that correctly.

That evaluating to True meant that `self.filtered_dropdown_items` to run.  Normally that runs in templates, which swallows the errors about `project == None`

Note that it doesn't error for logged-in users who are restricted.

corehq/tabs/uitab.py
```diff
@@ -66,12 +66,9 @@ def request_path(self):
...
     @property
-    def restrict_access_by_location(self):
-        """Is this a web user who can't access project-wide data?"""
-        return (
-            self.couch_user and self.domain
-            and not self.couch_user.has_permission(self.domain, 'access_all_locations')
-        )
+    def can_access_all_locations(self):
+        """Is this a web user who can access project-wide data?"""
+        return getattr(self._request, 'can_access_all_locations', True)
```
gcapalbo: i'm not sure it's good to default this setting to True, typically you'd default access to False right?

esoergel: Sorta.  The assumption is that if this isn't set, it's because this is some special case where it didn't go through middleware.  The middleware that sets this also defaults to True if it can't figure out the user's permissions, such as if the user isn't a member of the domain, or if the request doesn't have a domain.  I'm intentionally being pretty defensive - the idea is that certain users are explicitly marked as restricted when accessing pages associated with a domain.  If the users aren't marked, or we can't figure out whether they're marked, it falls back to the current behavior (which may be, for example, raising an exception in `login_and_domain_required`).

I do agree in principle with being stricter about this sort of thing, and I think this is important to get right, so I'd welcome suggestions about how to do this safely without failing on a bunch of obscure edge cases.

corehq/tabs/uitab.py
gcapalbo: I'm not sure I follow the part about how it used to swallow the errors.  If a template references a dynamic property and that dynamic property raises an exception, the exception is still raised.

Could it be the case that this template never referenced the dropdown or sidebar items, but now something elsewhere in the python code is referencing them on this request?

esoergel: I think you're right, there must be something excluding the menu bar from the invitation page, but I introduced this reference.  Now we only call `self.filtered_dropdown_items` when it's for sure a restricted user, which means there is a domain and the user is a member of the domain with a restricted role.

corehq/tabs/uitab.py
gcapalbo: i think if we default to True we're opening ourselves up to a vulnerability that's impossible to notice when it happens.

i'm not sure i understand the original code change well enough but is the main issue here that we just need should_show to return False, and bypass the filtered_dropdown_items and filtered_sidebar_items in this situation? if so is there anything else we can use to do that?

gcapalbo: (maybe check if the user is not a part of the domain?)

gcapalbo:
@esoergel so with https://github.com/dimagi/commcare-hq/pull/13195, do you still want to keep the rest of the commits in this PR or just the docstring?

esoergel:
I think still keep the rest of them.  I like the name change for consistency across code, and using the check that already happened in the middleware rather than reimplementing it here.

PR #13188, opened 2016-09-09
get indirect imports of models.py out of __init__.py for intrahealth

nickpell:


nickpell:
@sravfeyn @dannyroberts

PR #13189, opened 2016-09-09
move m4change CUSTOM_REPORTS from __init__.py to reports.py

nickpell:


nickpell:
@sravfeyn @dannyroberts

PR #13190, opened 2016-09-09
move opm CUSTOM_REPORTS to reports.py

nickpell:


nickpell:
@sravfeyn @dannyroberts

PR #13191, opened 2016-09-09
move mc CUSTOM_REPORTS to reports/__init__.py

nickpell:


nickpell:
@sravfeyn @dannyroberts

PR #13193, opened 2016-09-09
move CUSTOM_REPORTS for up_nrhm to reports/__init__.py

nickpell:


nickpell:
@sravfeyn @dannyroberts

PR #13197, opened 2016-09-11
Datetime picker

benrudolph:
@czue @wpride had some fun building out the date time question for new cloudcare (does not work on old cloudcare, but also doesn't break it). it also refactors the time and date entry to use the same library.

screen shot 2016-09-11 at 4 18 59 pm

screen shot 2016-09-11 at 5 26 46 pm
screen shot 2016-09-11 at 5 26 36 pm
screen shot 2016-09-11 at 5 26 42 pm

cross: https://github.com/dimagi/touchforms/pull/260
will, i couldn't figure out where to make the commcare-core PR against. we should standardize a branch for formplayer or something. here's the relevant commit: https://github.com/dimagi/commcare-core/commit/da4b5171678182050628ab8b86fcb701f9def989

all the prs are best read commit by commit

cc: @gcapalbo

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -53,7 +53,8 @@
     "moment": "2.12.0",
     "jquery-treetable": "^3.2.0",
     "css-grid-polyfill-binaries": "1.1.1",
-    "requirejs": "~2.3.1"
+    "requirejs": "~2.3.1",
+    "datetimepicker": "2.5.4"
   },
   "devDependencies": {
     "chai": "3.3.0",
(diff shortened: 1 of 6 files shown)


czue:
wow, looks awesome. does this need to block on a mobile library update?

benrudolph:
yeah it will, but it's a simple as an update to the formplayer submodule (we don't have to wait on a commcare release schedule)

emord:
@benrudolph was just looking at HQ's bootstrap-daterangepicker. Looks like it has an option for picking time as well http://www.daterangepicker.com/#config any thoughts on consolidating on one date/time picker across hq?

benrudolph:
nice good find. yep agree it would be good to consolidate (will put it on the backlog) and it looks pretty

PR #13199, opened 2016-09-12
Include other columns when not ignoring unspecified options

kaapstorm:
In form exports, if `split_column` was True and `ignore_unspecified_options` was False, columns would be omitted.

@dannyroberts @biyeun cc @benrudolph

### DIFF
--- a/corehq/apps/export/models/new.py
+++ b/corehq/apps/export/models/new.py
@@ -1616,7 +1616,8 @@ class SplitExportColumn(ExportColumn):
             return value
 
         if not isinstance(value, basestring):
-            return [None] * len(self.item.options) + [] if self.ignore_unspecified_options else [value]
+            unspecified_options = [] if self.ignore_unspecified_options else [value]
+            return [None] * len(self.item.options) + unspecified_options
 
         selected = OrderedDict((x, 1) for x in value.split(" "))
         row = []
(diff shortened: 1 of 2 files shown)


corehq/apps/export/models/new.py
```diff
@@ -1616,7 +1616,8 @@ def get_value(self, domain, doc_id, doc, base_path, split_column=False, **kwargs
             return value
 
         if not isinstance(value, basestring):
-            return [None] * len(self.item.options) + [] if self.ignore_unspecified_options else [value]
+            unspecified_options = [] if self.ignore_unspecified_options else [value]
+            return [None] * len(self.item.options) + unspecified_options
```
kaapstorm: Considering:

    >>> [None] * 5 + [] if True else ['foo']
    [None, None, None, None, None]

But:

    >>> [None] * 5 + [] if False else ['foo']
    ['foo']

So rather:

    >>> [None] * 5 + ([] if False else ['foo'])
    [None, None, None, None, None, 'foo']

kaapstorm:
Almost forgot. Context: http://manage.dimagi.com/default.asp?236455

corehq/apps/export/models/new.py
benrudolph: nice do you think you could add a minor test for this? https://github.com/dimagi/commcare-hq/blob/master/corehq/apps/export/tests/test_export_column.py#L39

kaapstorm:
:white_check_mark:

PR #13200, opened 2016-09-12
reference form attachments in case attachments instead of copying

snopoke:
@millerdev 
buddy @calellowitz

### DIFF
--- a/corehq/ex-submodules/casexml/apps/case/tests/test_multimedia.py
+++ b/corehq/ex-submodules/casexml/apps/case/tests/test_multimedia.py
@@ -2,7 +2,6 @@ from datetime import datetime, timedelta
 import uuid
 import os
 import hashlib
-from unittest.case import skip
 
 from django.conf import settings
 from django.template import Template, Context
@@ -304,7 +303,6 @@ class CaseMultimediaTest(BaseCaseMultimediaTest):
                                            sync_token=sync_log._id)
 
 
-@skip("Case attachments don't work with S3 API yet")
 class CaseMultimediaS3DBTest(BaseCaseMultimediaTest):
     """
     Tests new attachments for cases and case properties
--- a/corehq/ex-submodules/casexml/apps/case/tests/test_multimedia.py
+++ b/corehq/ex-submodules/casexml/apps/case/tests/test_multimedia.py
@@ -2,6 +2,7 @@ from datetime import datetime, timedelta
 import uuid
 import os
 import hashlib
+from unittest.case import skip
 
 from django.conf import settings
 from django.template import Template, Context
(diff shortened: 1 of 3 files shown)


corehq/ex-submodules/casexml/apps/case/tests/test_multimedia.py
```diff
@@ -300,3 +302,34 @@ def test_sync_log_invalidation_bug(self):
...
+            assert get_blob_db() is self.s3db, (get_blob_db(), self.s3db)
+
+    def tearDown(self):
+        self.s3db.close()
+        super(CaseMultimediaS3DBTest, self).tearDown()
+
+    @run_with_all_backends
+    def test_case_attachment(self):
+        single_attach = 'fruity_file'
+        xform, case = self._doCreateCaseWithMultimedia(attachments=[single_attach])
```
millerdev: This sends attachments to [`submit_form_locally`](https://github.com/dimagi/commcare-hq/blob/9a9de02b8bcbffb913ca0671e11cca1396ad20d5/corehq/ex-submodules/casexml/apps/case/tests/test_multimedia.py#L94-L100), which I would expect to work. I thought case attachments were working, but still stored in couch. Is that not true?

snopoke: Case attachments come from form attachments hence the call to
submit_form_locally
<https://github.com/dimagi/commcare-hq/blob/9a9de02b8bcbffb913ca0671e11cca1396ad20d5/corehq/ex-submodules/casexml/apps/case/tests/test_multimedia.py#L94-L100>.


The SQL backend uses the blobdb for all attachments (form and case).
Currently we duplicate form attachments for cases so one thing we could
consider is just referencing the same blob in both the form and case.

I also see that boto S3 has a `copy` API method which we could also use
rather than doing a read and then write.

millerdev: I'm trying out the copy API right now. Will have a PR soon, hopefully.

millerdev: https://github.com/dimagi/commcare-hq/pull/13203

corehq/ex-submodules/casexml/apps/case/tests/test_multimedia.py
```diff
@@ -300,3 +301,33 @@ def test_sync_log_invalidation_bug(self):
...
+    Tests new attachments for cases and case properties
+    Spec: https://github.com/dimagi/commcare/wiki/CaseAttachmentAPI
+    """
+
+    def setUp(self):
+        super(CaseMultimediaS3DBTest, self).setUp()
+        with trap_extra_setup(AttributeError, msg="S3_BLOB_DB_SETTINGS not configured"):
+            config = settings.S3_BLOB_DB_SETTINGS
+            self.s3db = TemporaryS3BlobDB(config)
+            assert get_blob_db() is self.s3db, (get_blob_db(), self.s3db)
```
millerdev: nit: the above two lines don't need to (and probably shouldn't) be in the `with trap_extra_setup` block.

millerdev:
👍 on ✅

snopoke:
:white_check_mark:

PR #13206, opened 2016-09-13
avoid AttributeError when custom module does not have any reports / r…

nickpell:
…eports.py

http://manage.dimagi.com/default.asp?237197

@dannyroberts @sravfeyn


czue:
looks good - guessing no tests? we might want to consider merge/deploy before hsph gets going today

nickpell:
I'd say safe to merge before tests.  Don't think there could be any coverage for this case.

PR #13207, opened 2016-09-13
Bump use sqlite backend

benrudolph:
@czue @wpride plenty of disk space on prod and 15gb is free on india (of 25gb) which i think is enough based on its low usage
cc: @gcapalbo

### DIFF
--- a/corehq/toggles.py
+++ b/corehq/toggles.py
@@ -794,7 +794,7 @@ TF_USES_SQLITE_BACKEND = PredictablyRandomToggle(
     'Use a SQLite backend for Touchforms',
     TAG_PRODUCT_PATH,
     [NAMESPACE_DOMAIN],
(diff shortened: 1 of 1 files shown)


czue:
big strides

corehq/toggles.py
```diff
@@ -794,7 +794,7 @@ def _commtrackify(domain_name, toggle_is_enabled):
     'Use a SQLite backend for Touchforms',
     TAG_PRODUCT_PATH,
     [NAMESPACE_DOMAIN],
-    0.1,
+    0.5,
```
gcapalbo: this means this will be turned on for 50% of all domains, right? am curious if we've run the sms/ivr test suites with this backend (it has to be run manually due to touchforms dependency)?

PR #13214, opened 2016-09-13
Care Bangladesh changes

kkrampa:
@czue 
http://manage.dimagi.com/default.asp?219730#BugEvent.1211758


kkrampa:
Also, I created case for reindex: http://manage.dimagi.com/default.asp?237285

PR #13215, opened 2016-09-13
delete SQL attachments with forms

snopoke:
@calellowitz 

cc @dimagi/scale-team @millerdev


corehq/form_processor/backends/sql/dbaccessors.py
```diff
@@ -224,13 +225,49 @@ def form_exists(form_id, domain=None):
             return result.form_exists
 
     @staticmethod
-    @transaction.atomic
```
emord: Why did you remove the atomic decorator? Would it make sense to move this to the block where forms are deleted?

snopoke: Removed it from here since there is only one SQL call so atomicity is implicit. 

We could add it to the `hard_delete_case_and_forms` function which is where this get's called from. Atomicity isn't respected in a sharded environment though since transactions only apply to the proxy DB.

millerdev:
👍 on ✅

millerdev:
can't merge

snopoke:
:white_check_mark:

PR #13217, opened 2016-09-13
Fix prime restore

proteusvacuum:
Prime restore was breaking because I renamed a method. 
@emord


emord:
:+1: on :white_check_mark: although I suppose this isn't tested

proteusvacuum:
✅ 
I'll write a test for this in a followup.

PR #13219, opened 2016-09-13
Show dropdown of location types for top level locations

emord:
introduced https://github.com/dimagi/commcare-hq/commit/2039cc27

http://manage.dimagi.com/default.asp?237275

Pretty ugly fix. knockout uses `undefined` by default, but the way we pass in the hierarchy the `parent_type` is interpreted as `null` Thought about changing it back to `==` but decided to be explicit


esoergel:
Nice, thanks for the fix!  Canceling build and merging since there are no relevant tests.  I confirmed locally that this fixes the issue.

PR #13024, opened 2016-08-29
Mk 235887 configure menu grid

mkangia:
http://manage.dimagi.com/default.asp?235887

PR #13230, opened 2016-09-14
Retain old values for cc-grid-menus

mkangia:

PR #13231, opened 2016-09-14
Revert "Mk 235887 configure menu grid"

orangejenny:
Reverts dimagi/commcare-hq#13024

Last-minute bug showed up, this shouldn't be deployed.

@mkangia / @benrudolph


mkangia:
Need to figure out fixes needed for:
Modules Menu Display: List
Forms Menu Display : Grid
results in Module menu to be grid as well.

also followup PR : https://github.com/dimagi/commcare-hq/pull/13230 to be integrated to avoid migration and retain old values.

orangejenny:
don't think it's worth waiting for tests on this one

PR #13232, opened 2016-09-14
Revert "Revert "Mk 235887 configure menu grid""

mkangia:
Reverts dimagi/commcare-hq#13231

http://manage.dimagi.com/default.asp?235887


mkangia:
@orangejenny @czue 
Sorry for the confusion here guys. Just checked that if we don't [Retain old values](https://github.com/dimagi/commcare-hq/pull/13232/commits/445ab3b874b3b45845c0243f1699b48a692d5316) it works as expected.
But the question is that [here](https://github.com/dimagi/commcare-hq/pull/13232/files#diff-a0cd4db8a4357e7a1f662e159786c735L350) we have changed the values for the fields and made it a part of the feature flag (along with [this](https://github.com/dimagi/commcare-hq/pull/13232/files#diff-30960a86c22284ec0bd8040403d4bce5L152)) . How does this effect existing applications? Is it ok to change the values without any data migration?

orangejenny:
I think there needs to be some kind of migration, or else support both the old and new values. The latter is probably easier, it might just be a matter of having `grid_display_for_all_modules` accept either "all" or "yes".

corehq/apps/app_manager/tests/test_grid_menus.py
```diff
@@ -1,15 +1,27 @@
 from django.test import SimpleTestCase
 from corehq.apps.app_manager.tests.app_factory import AppFactory
 from corehq.apps.app_manager.tests.util import TestXmlMixin
+from mock import patch
+
+
+def mock_toggle(slug, enable_users=None):
+    from toggle.models import Toggle
+    if enable_users is None: enable_users = []
```
dimagimon: E701 multiple statements on one line (colon)

corehq/apps/app_manager/tests/test_grid_menus.py
```diff
@@ -58,10 +70,153 @@ def test_use_grid_menus_is_false(self):
...
+        factory.app.get_module(0).put_in_root = True
+        factory.app.use_grid_menus = True
+        suite = factory.app.create_suite()
+        self.assertXmlPartialEqual(
+            '<partial><menu id="root" style="grid"><text><locale id="modules.m0"/></text>\
+            <command id="m0-f0"/></menu></partial>',
+            suite,
+            root_xpath
+        )
+
```
dimagimon: W391 blank line at end of file

mkangia:
@orangejenny @czue @amsagoff 
Finally got this working!!!! Had to abort the plan for using the existing fields only.
So overall the change is that we have added one more config "Forms Menu Display" for form views and setup at module setting page along with minor textual changes in existing configs.
This is deployed on staging now.

Please help with review and any text changes if needed.
screen shot 2016-09-20 at 1 50 01 am

Have added "Set Grid Menus as No for this to be active" since Grid Menus takes priority over these two. Can remove this text if its not needed or can change it to something better?

czue:
congrats! I'm going to let @orangejenny and @amsagoff lead feedback

corehq/apps/app_manager/models.py
```diff
@@ -5687,6 +5692,22 @@ def get_subcase_types(self, case_type):
...
 
+    @memoized
+    def grid_display_for_some_modules(self):
+        return self.grid_menu_toggle_enabled() and self.grid_form_menus == 'some'
+
+    @memoized
+    def grid_display_for_all_modules(self):
+        return self.grid_menu_toggle_enabled() and self.grid_form_menus == 'all'
+
+    def grid_menu_toggle_enabled(self):
```
orangejenny: The more standard way to check a toggle is `toggles.GRID_MENUS.enabled(self.domain)`

corehq/apps/app_manager/static/app_manager/json/commcare-app-settings.yaml
```diff
@@ -149,12 +149,25 @@
...
-  description: Display root menu as a grid instead of a list.
-  widget: bool
+  name: Modules Menu Display
+  description: Display root menu as a list or grid.
+  values: [false, true]
+  value_names: ['List', 'Grid']
   default: false
   since: "2.24"
   toggle: GRID_MENUS
+  foot_note: "Set Grid Menus as No for this to be active"
```
orangejenny: I don't think the text you're displaying as a footnote is really different than help text. I'd just add those explanations to `description` and remove the footnote logic altogether.

orangejenny: On second thought, for the `Set Grid Menus as No for this to be active` text specifically, you can remove it and instead use the `requires` property. If you look under "Java Phone General Settings" you'll see that "Auto-Sync Frequency" and "Unsynced Time Limit" are disabled if "Sync Mode" is set to "Push Only." This leads to a display like this:
screen shot 2016-09-20 at 5 33 37 pm

The `For per-module option...` text for the setting below is still needed, but can be moved into that setting's description.

orangejenny (changes requested):
Generally looks good, see inline comments for changes.

orangejenny:
I'm not sure it's a necessary to keep the `Grid Menus` toggle as a gateway for the other two, but defer to @amsagoff

amsagoff:
also think good to remove

orangejenny:
Hey Manish, sorry that some of the comments on this are conflicting a bit. To summarize my and Amelia's comments:
- Remove "Grid Menus" setting
- Update toggle logic
- Move "For per-module option..." text into the description and remove the foot_note logic.

corehq/apps/app_manager/models.py
mkangia: Awesome!

corehq/apps/app_manager/static/app_manager/json/commcare-app-settings.yaml
mkangia: `requires` .. nice hook!

corehq/apps/app_manager/models.py
mkangia: Nice! Would update this.

mkangia:
@orangejenny Thanks Jenny..
Have made the code changes.
Though, curious how the existing applications would be effected by removing Grid Menus setting.

corehq/apps/app_manager/tests/test_grid_menus.py
```diff
@@ -1,15 +1,22 @@
 from django.test import SimpleTestCase
 from corehq.apps.app_manager.tests.app_factory import AppFactory
 from corehq.apps.app_manager.tests.util import TestXmlMixin
+from corehq.util.test_utils import flag_enabled
 
 
-class GridMenuSuiteTests(SimpleTestCase, TestXmlMixin):
+def mock_toggle(slug, enable_users=None):
```
proteusvacuum: @mkangia : I don't think you need this function any more.

mkangia: Yup! You are right.

corehq/ex-submodules/toggle/models.py
```diff
@@ -43,6 +44,11 @@ def remove(self, item):
             self.enabled_users.remove(item)
             self.save()
 
+    @memoized
```
orangejenny: This function isn't used anymore

orangejenny (changes requested):

orangejenny:
> Though, curious how the existing applications would be effected by removing Grid Menus setting.

Yeah, this is a good point. The most forgiving thing to do would be a lazy migration: keep reading the old setting, and using that value to set the new settings. Isn't having the old setting on equivalent to having both of the new settings set to "Grid"? What do you think of implementing that?

corehq/ex-submodules/toggle/models.py
```diff
@@ -1,6 +1,7 @@
 from __future__ import absolute_import
 from datetime import datetime
 from dimagi.ext.couchdbkit import *
+from dimagi.utils.decorators.memoized import memoized
```
dimagimon: F401 'memoized' imported but unused

corehq/ex-submodules/toggle/models.py
mkangia: Removed!

mkangia:
@orangejenny i am not very sure about this but i guess the old builds of applications should keep working since they would have setup files retained from the time when the builds were created. Isn't it?

> Isn't having the old setting on equivalent to having both of the new settings set to "Grid"
Yup!

Lazy Migration. i won't prefer doing that. Adds code that we would not be sure about removing ever. Moreover the migrated fields are feature flags, we would need to add feature flag to all these domains as well. Defeats the point of Feature Flags i guess.

The reason i retained the existing config was to avoid surprises for users. If we are announcing the changes and users know then it should be cool.
Thats what i think from my limited context.

orangejenny:
Ohhh, I forgot this is all feature flagged. Yeah, don't worry about a migration. Good to merge!

>  If we are announcing the changes and users know then it should be cool.

You likely already know this, but this is only true for flagged features. In general we can't rely on users paying attention to announcements.

orangejenny (approved):

orangejenny:
Just realized this still has the "hold off on merge" tag, so I'm holding off, but this is good to go from my perspective.

mkangia:
@orangejenny 
> If we are announcing the changes and users know then it should be cool.
i was referring to removing the 'Grid Menus' option which is available by default. That would go missing now since we are removing it.

Wanted to be sure about removing the field/migration stuff and also do a final test on mobile once .. the tests are passing thought.. so the 'hold off on merge'.

mkangia:
@orangejenny just removed some unneeded changes done earlier. Tested as well!

orangejenny:
hurrah, it's merged!

PR #13235, opened 2016-09-14
WIP Missing value

benrudolph:
@czue @NoahCarnahan there was a bug reported about missing values for exports. i tried implementing but then got confused on what all the edge cases should be. is there documentation on when to show `---` vs. `''`? I'm guessing we should never return `None` as a value. one of the more confusing ones is how to handle multiple choice questions

cc: @gcapalbo

### DIFF
--- a/corehq/apps/export/const.py
+++ b/corehq/apps/export/const.py
@@ -52,3 +52,8 @@ PROPERTY_TAG_STOCK = 'stock'
 FORM_EXPORT = 'form'
 CASE_EXPORT = 'case'
 MAX_EXPORTABLE_ROWS = 100000
+
+# When a question is missing completely from a form/case this should be the value
+MISSING_VALUE = '---'
+# When a question has been answered, but is blank, this shoudl be the value
+EMPTY_VALUE = ''
--- a/corehq/apps/export/models/new.py
+++ b/corehq/apps/export/models/new.py
@@ -51,6 +51,8 @@ from corehq.apps.export.const import (
     USER_DEFINED_SPLIT_TYPES,
     PLAIN_USER_DEFINED_SPLIT_TYPE,
     DATA_SCHEMA_VERSION,
+    MISSING_VALUE,
+    EMPTY_VALUE,
 )
 from corehq.apps.export.exceptions import BadExportConfiguration
 from corehq.apps.export.dbaccessors import (
@@ -206,6 +208,8 @@ class ExportColumn(DocumentSchema):
             except ValueError:
                 # Unable to convert the string to a date
                 pass
+        if value is None:
+            value = MISSING_VALUE
         return value
 
     @staticmethod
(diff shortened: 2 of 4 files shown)


czue:
I thought that `---` meant it was not in the form at all, whereas `''` meant it was in the form but empty. however yeah, for things like multiple choice questions i have no idea what we should do

czue:
build failures
```

FAIL: Ensure dates are transformed for excel when `transform_dates` is set to True
  File "/mnt/commcare-hq-ro/corehq/apps/export/tests/test_get_export_file.py", line 302, in test_transform_dates
    u'rows': [[None], [couch_to_excel_datetime('2015-07-22T14:16:49.584880Z', None)]],
AssertionError: {u'My table': {u'headers': [u'Date'], u'rows': [[u'---'], [u'2015-07-22 14:16:49 [truncated]... != {u'My table': {u'headers': [u'Date'], u'rows': [[None], ['2015-07-22 14:16:49']] [truncated]...
  {u'My table': {u'headers': [u'Date'],
-                u'rows': [[u'---'], [u'2015-07-22 14:16:49']]}}
?                           ^^^^^^    -
+                u'rows': [[None], ['2015-07-22 14:16:49']]}}
?                           ^^^^

FAIL: test_export_transforms (corehq.apps.export.tests.test_get_export_file.ExportTest)
  File "/mnt/commcare-hq-ro/corehq/apps/export/tests/test_get_export_file.py", line 589, in test_export_transforms
    [u'2016-04-27'],  # offset by 3 since that's the mocked random offset
AssertionError: {u'My table': {u'headers': [u'DEID Date Transform column [sensitive]'], u'rows': [truncated]... != {u'My table': {u'headers': [u'DEID Date Transform column [sensitive]'], u'rows': [truncated]...
  {u'My table': {u'headers': [u'DEID Date Transform column [sensitive]'],
-                u'rows': [[u'---'], [u'2016-04-07'], [u'2016-04-27']]}}
?                           ^^^^^^
+                u'rows': [[None], [u'2016-04-07'], [u'2016-04-27']]}}
?                           ^^^^
 



```

czue:
might be a question for @sheelio or @amsagoff as to whether anyone knows what the "expected" behavior of these is. I suspect the only source of documentation is/was the code.

NoahCarnahan:
> I thought that --- meant it was not in the form at all, whereas '' meant it was in the form but empty. 
I thought it was something like this as well

benrudolph:
👍 thanks for the input guys. going to come up with something logical, and then document it

sheelio:
Have never seen "None" before.  --- usually means skipped in form or not a case property that was ever set for the case.  Blank means it was set to "".

benrudolph:
updated with blanks and dashes according to that convention

corehq/apps/export/const.py
```diff
@@ -52,3 +52,8 @@
 FORM_EXPORT = 'form'
 CASE_EXPORT = 'case'
 MAX_EXPORTABLE_ROWS = 100000
+
+# When a question is missing completely from a form/case this should be the value
+MISSING_VALUE = '---'
+# When a question has been answered, but is blank, this shoudl be the value
+EMPTY_VALUE = ''
```
czue: this is helpful

PR #13240, opened 2016-09-14
size() deprecated

orangejenny:
`size` was deprecated in 1.8 and removed in 3.0: https://api.jquery.com/size/

The [main migration guide](https://jquery.com/upgrade-guide/3.0/) doesn't call this out as a breaking change, which is a little bit concerning. Had a look through [other removed functions](https://api.jquery.com/category/removed/) and don't see anything else that's problematic for us.

@emord


emord (approved):

orangejenny:
tests passed, merging

PR #13241, opened 2016-09-14
Sort by server_date as backup

emord:
http://manage.dimagi.com/default.asp?236117#1218242

There is an old bug with commtrack forms in formbuilder where they report the date of a stock transaction, but not a time. This report had several transactions on the same date but received at different times. I don't think that this will necessarily be right all the time (like in the case of out of order submissions), but it will be much closer most of the time.

The bug in formbuilder has been resolved for a while by writing now() instead of today(), but we aren't overwriting old xml.

### DIFF
--- a/corehq/apps/reports/analytics/couchaccessors.py
+++ b/corehq/apps/reports/analytics/couchaccessors.py
@@ -140,7 +140,7 @@ def get_ledger_values_for_case_as_of(domain, case_id, section_id, as_of, program
     results = transactions.exclude(
         report__date__gt=as_of
     ).order_by(
-        'product_id', '-report__date'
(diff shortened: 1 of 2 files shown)


corehq/apps/reports/analytics/couchaccessors.py
```diff
@@ -140,7 +140,7 @@ def get_ledger_values_for_case_as_of(domain, case_id, section_id, as_of, program
     results = transactions.exclude(
         report__date__gt=as_of
     ).order_by(
-        'product_id', '-report__date'
+        'product_id', '-report__date', '-report__server_date'
```
esoergel: :+1: 
Hopefully this'll fix that transient commtrack bug too, as you suggested!

PR #13243, opened 2016-09-14
Make sure location_type names and codes are unique

esoergel:
@czue @orangejenny
This does three things:
- adds a db constraint that location types be unique by both name and code
- adds client-side validation of the same
- fails hard on the server if client validation broke

best read :blowfish: by :blowfish: 

Prod has only one domain for which this is an issue, which we're resolving manually.
I'll similarly verify staging and India before allowing this to be merged.  If the migration fails, it provides enough information to fix it.  Here's how to check for uniqueness:

``` python
from corehq.apps.locations.models import *
from django.db.models import *
(LocationType.objects
 .values('domain', 'name')
 .order_by()
 .annotate(count=Count('pk'))
 .filter(count__gt=1))
(LocationType.objects
 .values('domain', 'code')
 .order_by()
 .annotate(count=Count('pk'))
 .filter(count__gt=1))
```

@sheelio FYI

### DIFF
--- a/corehq/apps/locations/static/locations/ko/location_types.js
+++ b/corehq/apps/locations/static/locations/ko/location_types.js
@@ -80,6 +80,27 @@ hqDefine('locations/ko/location_types.js', function(){
                 }
             });
 
+            // Make sure name and code are unique
+            _.each(['name', 'code'], function (field) {
+                var counts_by_value = _.countBy(self.loc_types(), function (loc_type) {
+                    return loc_type[field]();
+                });
+                var duplicates = [];
+                _.each(counts_by_value, function (count, value) {
+                    if (count > 1) {
+                        duplicates.push(value);
+                        valid = false;
+                    }
+                });
+                _.each(self.loc_types(), function (loc_type) {
+                    var has_error = (field === 'name') ? loc_type.duplicate_name_error : loc_type.duplicate_code_error;
+                    has_error(false);
+                    if (_.contains(duplicates, loc_type[field]())) {
+                        has_error(true);
+                    }
+                });
+            });
+
             var top_level_loc = false;
             $.each(self.loc_types(), function(i, e) {
                 if (!e.parent_type()) {
@@ -166,6 +187,8 @@ hqDefine('locations/ko/location_types.js', function(){
         self.view = view_model;
 
         self.name_error = ko.observable(false);
+        self.duplicate_name_error = ko.observable(false);
+        self.duplicate_code_error = ko.observable(false);
 
         self.validate = function() {
             self.name_error(false);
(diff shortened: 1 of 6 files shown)


corehq/apps/locations/views.py
```diff
@@ -246,10 +247,16 @@ def mk_loctype(name, parent_type, administrative,
...
+                if prop not in loc_type:
+                    raise LocationConsistencyError("Missing an organization level property!")
             pk = loc_type['pk']
             if not _is_fake_pk(pk):
                 pks.append(loc_type['pk'])
+        if not (
+            len(loc_types) == len(set(lt['name'] for lt in loc_types))
+            and len(loc_types) == len(set(lt['code'] for lt in loc_types))
+        ):
+            raise LocationConsistencyError("'name' and 'code' are supposed to be unique")
```
czue: These errors just get handled as 500s yeah? since there is front-end validation that seems fine, we could update if it comes up for some race condition or something.

czue (approved):
looks ok to me once your'e confident on the migrations.

corehq/apps/locations/views.py
esoergel: Yeah, these will 500, which should alert us if something breaks in the JS validation

esoergel:
Okay, the queries I pasted above return no results on prod, staging, and india, so I think this is g2m.  @benrudolph assuming you're the deployer, I've manually made sure that the data won't cause this migration to fail, but you should be aware of this migration just in case, feel free to ping me if anything comes up.

esoergel:
(self-merging, since Cory approved)

corehq/apps/locations/static/locations/ko/location_types.js
```diff
@@ -80,6 +80,27 @@ hqDefine('locations/ko/location_types.js', function(){
...
+                });
+                var duplicates = [];
+                _.each(counts_by_value, function (count, value) {
+                    if (count > 1) {
+                        duplicates.push(value);
+                        valid = false;
+                    }
+                });
+                _.each(self.loc_types(), function (loc_type) {
+                    var has_error = (field === 'name') ? loc_type.duplicate_name_error : loc_type.duplicate_code_error;
```
orangejenny: It is a tiny bit weird that you're iterating over a theoretically generic set of `loc_types` when the error types are only names & codes, but I don't have a better suggestion that wouldn't be overly verbose.

esoergel: I don't think I follow, "theoretically generic"?

orangejenny: nm, i misread

PR #13246, opened 2016-09-15
fix domain-name normalize issue

sravfeyn:
http://manage.dimagi.com/default.asp?237438

### DIFF
--- a/corehq/tabs/uitab.py
+++ b/corehq/tabs/uitab.py
@@ -149,7 +149,11 @@ class UITab(object):
 
     @property
     def url_prefixes(self):
-        return [url_prefix_format.format(domain=self.domain)
+        # Use self._request.domain instead of self.domain to generate url-prefixes
+        # because the latter might have a normalized domain name which might not match the
+        # domain name mentioned in the URL. for example domain-name 'hq_test' is normalized to
+        # 'ht-test'
+        return [url_prefix_format.format(domain=self._request.domain)
                 for url_prefix_format in self.url_prefix_formats]
 
     def get_url_prefix_formats_suggestion(self):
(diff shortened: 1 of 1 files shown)


sravfeyn:
@nickpell

corehq/tabs/uitab.py
```diff
@@ -149,7 +149,11 @@ def url(self):
 
     @property
     def url_prefixes(self):
-        return [url_prefix_format.format(domain=self.domain)
+        # Use self._request.domain instead of self.domain to generate url-prefixes
+        # because the latter might have a normalized domain name which might not match the
+        # domain name mentioned in the URL. for example domain-name 'hq_test' is normalized to
+        # 'ht-test'
```
nickpell: small typo: ht-test vs hq-test

nickpell: I wonder if ```self.domain``` should be set to ```self._request.domain```?  If this is a problem here I'd be concerned that it might be a problem elsewhere too.

sravfeyn: Yeah, I thought about it, but replacing would break things that query the db by domain-name, (the normalized domain name is the name stored in db). Further this issue is a rare accident. Someone using hq_test instead of hq-test in URL is rare (as all the urls generated by HQ will have the latter).

nickpell: 👍

nickpell:
:+1: when :white_check_mark:

emord:
Test failures look related:

```
...
  File "/mnt/commcare-hq-ro/corehq/tabs/uitab.py", line 157, in url_prefixes
    for url_prefix_format in self.url_prefix_formats]
AttributeError: 'WSGIRequest' object has no attribute 'domain'
```

PR #13249, opened 2016-09-15
Simplify report submitting

benrudolph:
@emord @biyeun this simplifies the report submit button. I think there were race conditions and perhaps the upgrade in jquery brought them to light. In any case this works locally for all the major reports
http://manage.dimagi.com/default.asp?237437
cc: @gcapalbo

### DIFF
--- a/corehq/apps/reports/static/reports/javascripts/reports.async.js
+++ b/corehq/apps/reports/static/reports/javascripts/reports.async.js
@@ -47,10 +47,6 @@ var HQAsyncReport = function (o) {
 
         self.updateReport(true, window.location.search.substr(1), self.standardReport.filterSet);
 
-        // only update the report if there are actually filters set
-        if (!self.standardReport.needsFilters) {
-            self.standardReport.filterSubmitButton.addClass('disabled');
-        }
         self.filterForm.submit(function () {
             var params = hqImport('reports/javascripts/reports.util.js').urlSerialize(this);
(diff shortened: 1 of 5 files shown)


corehq/apps/reports/static/reports/javascripts/reports.async.js
```diff
@@ -160,7 +136,10 @@ var HQAsyncReport = function (o) {
                     self.hqLoading.fadeIn();
                 }
 
-            }
+            },
+            complete: function() {
+                self.standardReport.filterSubmitButton.button('reset');
```
benrudolph: this is needed for an async report to re-enable the apply button. there's some complicated logic above, but i don't think it's necessary. `reset` keeps all the styles from what i noticed (perhaps fixed in some version upgrade)

emord (approved):

biyeun:
@benrudolph looks good but there are merge conflicts

benrudolph:
updated, but think it makes sense to wait til monday to merge. i still like the changes that this PR makes to the apply button

gcapalbo:
@benrudolph is this good to merge now?

benrudolph:
i'm gonna punt on this for now. i don't have bandwidth to ensure this is correct for all the reports we have since the logic that was happening was a bit magical to me. it makes me nervous that there are a lot of reports that reuse this and it's hard to tell what exactly this did for them

PR #13252, opened 2016-09-15
upgrade jquery to 2.2.4

orangejenny:
@emord


emord (approved):
No way that this is tested, but will be paranoid anyways

PR #13253, opened 2016-09-15
don't use removeProp on native properties

orangejenny:
https://api.jquery.com/removeProp/ says "Do not use this method to remove native properties such as checked, disabled, or selected. This will remove the property completely and, once removed, cannot be added again to element. Use .prop() to set these properties to false instead."

@emord


benrudolph:
nice

emord:
Thanks for finding the explanation. Totally glanced over that. so I guess jquery 1.11 did this incorrectly

emord (approved):

emord:
going to deploy this to fix some reports

PR #13255, opened 2016-09-15
Translate returned module/form name

orangejenny:
http://manage.dimagi.com/default.asp?235640

@esoergel / @czue

### DIFF
--- a/corehq/apps/app_manager/views/forms.py
+++ b/corehq/apps/app_manager/views/forms.py
@@ -251,7 +251,7 @@ def _edit_form_attr(request, domain, app_id, unique_form_id, attr):
         if xform.exists():
             xform.set_name(name)
             save_xform(app, form, xform.render())
-        resp['update'] = {'.variable-form_name': form.name[lang]}
+        resp['update'] = {'.variable-form_name': trans(form.name, [lang], use_delim=False)}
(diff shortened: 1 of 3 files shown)


corehq/apps/app_manager/views/forms.py
```diff
@@ -251,7 +251,7 @@ def should_edit(attribute):
         if xform.exists():
             xform.set_name(name)
             save_xform(app, form, xform.render())
-        resp['update'] = {'.variable-form_name': form.name[lang]}
+        resp['update'] = {'.variable-form_name': trans(form.name, [lang], use_delim=False)}
```
esoergel: It looks like `trans(name, [lang], use_delim=False)` is essentially a more robust version of `name[lang]`, yeah?  I was confused about this change 'till I saw this bit from the source of `trans`:
```python
        if lang in name and name[lang]:
            return name[lang] + ("" if langs and lang == langs[0] else suffix(lang))
```

esoergel (approved):
:+1:

corehq/apps/app_manager/views/forms.py
orangejenny: Kind of - I think of it as a display version of `name[lang]`

PR #13256, opened 2016-09-15
Allow moving form to first position in non-basic modules

orangejenny:
http://manage.dimagi.com/default.asp?236226

None and zero have different meanings here.

@esoergel / @czue


esoergel (approved):

czue:
sneaky

PR #13259, opened 2016-09-15
Uneditable export columns

benrudolph:
@czue @NoahCarnahan this implement "uneditable" user defined columns. Thought here is that when we do a migration we can now just force add the columns for the existing exports. this'll hopefully get us closer to 0 domains on old exports. open for review, still need to write tests and hook it up to the migrate_exports command

cc: @gcapalbo

### DIFF
--- a/corehq/apps/export/models/new.py
+++ b/corehq/apps/export/models/new.py
@@ -1454,6 +1454,8 @@ class UserDefinedExportColumn(ExportColumn):
     data within the form. It should only be needed for RemoteApps
     """
 
+    is_editable = BooleanProperty(default=True)
+
     # On normal columns, the path is defined on an ExportItem.
     # Since a UserDefinedExportColumn is not associated with the
     # export schema, the path is defined on the column.
--- a/corehq/apps/export/static/export/js/models.js
+++ b/corehq/apps/export/static/export/js/models.js
@@ -320,6 +320,7 @@ hqDefine('export/js/models.js', function () {
         e.preventDefault();
         table.columns.push(new UserDefinedExportColumn({
             selected: true,
+            is_editable: true,
             deid_transform: null,
             doc_type: 'UserDefinedExportColumn',
             label: '',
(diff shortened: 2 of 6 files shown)


czue:
sorry, can you remind me what the rationale for this was (as opposed to user defined)?

in terms of behavior - these will show up in the UI and export but you can't change them? can you delete them? assume if you can then they would be gone forever at that point?

i think we probably talked about this I just can't remember

benrudolph:
correct, this will show up in the UI and you can't change them or delete them, it will look exactly like your previous export. i thought it'd be confusing if all of a sudden after the migration a bunch columns had text field inputs instead of the normal looking `form.question` in the column definition

czue:
> this will show up in the UI and you can't change them or delete them, it will look exactly like your previous export

ok, have you confirmed this with anyone on product? i don't know how bad/disruptive that would be but sounds like something we should at least flag

benrudolph:
no i haven't, but the alternative would be to just have those columns disappear. i thought the goal was to make the migration be as transparent as possible. what part do you think would be disruptive?

czue:
not disruptive so much as getting support tickets from people asking why they can't edit their exports.

is the alternative not to create an editable, manual column? i think I just don't understand this well enough to be useful, but if you feel good that you've thought it though and talked to any relevant stakeholders I think feel free to just ignore me

benrudolph:
ah i see where we're missing each other. the part that would be uneditable would be the path to the column not the label. there are 3 columns in the export, selected, question, and label. on a normal column selected and label are editable. on a user defined column, question is editable so that a user can specify where the data exists. a non -editable user defined column functions that exact same as a regular column except that we've already determined the path to the question and thus do not let users modify that (like normal columns)

czue:
gotcha, that seems totally fine then. sorry about causing churn.

benrudolph:
ok tests written, sorry for the delay. we're almost there. got a few more items to tick off before going big and converting everyone

czue:
:+1: from me. haven't looked at code closely so would prefer if @NoahCarnahan or @gcapalbo reviewed and merged

NoahCarnahan:
Looks good. One step closer!

PR #13261, opened 2016-09-16
restrict AppDiffView to only apps in domains user can access

dannyroberts:
@emord


dannyroberts:
@nickpell @sravfeyn buddy

nickpell (approved):

PR #13267, opened 2016-09-16
Also include fixtures at users' locations

esoergel:
http://manage.dimagi.com/default.asp?237323

### DIFF
--- a/corehq/apps/fixtures/tests.py
+++ b/corehq/apps/fixtures/tests.py
@@ -12,6 +12,7 @@ from corehq.apps.fixtures.models import FixtureDataItem, FixtureDataType, Fixtur
     FixtureItemField, FieldList
 from corehq.apps.fixtures.exceptions import FixtureVersionError
 from corehq.apps.fixtures.utils import is_identifier_invalid
+from corehq.apps.locations.tests.util import LocationHierarchyTestCase
 from corehq.apps.users.models import CommCareUser
 from corehq.apps.users.dbaccessors.all_commcare_users import delete_all_users
 from django.test import TestCase, SimpleTestCase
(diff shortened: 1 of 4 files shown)


esoergel:
@czue @orangejenny

orangejenny:
don't really have context for why this works, but it looks reasonable

esoergel:
The user should receive fixtures assigned to their location or any ancestor location.  If a fixture is assigned to, say Suffolk, a Boston user should also see it.  That currently works, but right now `loc_ids` doesn't include the user's location, so a user at Suffolk wouldn't see a fixture assigned to Suffolk.

orangejenny:
got it, thanks for explaining :)

orangejenny (approved):

corehq/apps/fixtures/tests.py
```diff
@@ -171,6 +172,105 @@ def test_get_indexed_items(self):
...
+            item_attributes=[],
+        )
+        data_type.save()
+
+        def make_data_item(location_name, cost):
+            """Make a fixture data item and assign it to location_name"""
+            data_item = FixtureDataItem(
+                domain=cls.domain,
+                data_type_id=data_type.get_id,
+                fields= {
```
dimagimon: E251 unexpected spaces around keyword / parameter equals

corehq/apps/fixtures/tests.py
```diff
@@ -171,6 +172,105 @@ def test_get_indexed_items(self):
...
+                            field_value=location_name,
+                            properties={},
+                        )]
+                    ),
+                },
+                item_attributes={},
+            )
+            data_item.save()
+
+            fixture_ownership = FixtureOwnership(
```
dimagimon: F841 local variable 'fixture_ownership' is assigned to but never used

esoergel:
@orangejenny Just added two more commits to future-proof this a bit.

orangejenny:
new commits look good, hurrah for tests

czue:
nice

PR #13268, opened 2016-09-16
Mobile worker login to dashboard

emord:
Redirect a mobile worker to the dashboard if they have access to reports.

http://manage.dimagi.com/default.asp?237396#1219436

cc @proteusvacuum

### DIFF
--- a/corehq/apps/reports/util.py
+++ b/corehq/apps/reports/util.py
@@ -441,21 +441,6 @@ def get_installed_custom_modules():
     return [import_module(module) for module in settings.CUSTOM_MODULES]
 
 
-def is_mobile_worker_with_report_access(couch_user, domain):
-    if not couch_user.is_commcare_user():
-        return False
-
-    if domain is not None:
-        role = couch_user.get_role(domain)
-        if role:
-            return (
-                role.permissions.view_reports or
-                len(role.permissions.view_report_list) > 0
-            )
-
-    return False
-
-
 def get_INFilter_element_bindparam(base_name, index):
     return '%s_%d' % (base_name, index)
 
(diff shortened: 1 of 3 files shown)


corehq/apps/reports/util.py
```diff
@@ -442,11 +442,18 @@ def get_installed_custom_modules():
 
 
 def is_mobile_worker_with_report_access(couch_user, domain):
-    return (
-        couch_user.is_commcare_user
-        and domain is not None
-        and Domain.get_by_name(domain).default_mobile_worker_redirect == 'reports'
-    )
+    if not couch_user.is_commcare_user():
```
czue: was this always wrong?

corehq/apps/reports/util.py
```diff
@@ -442,11 +442,18 @@ def get_installed_custom_modules():
...
+        return False
+
+    if domain is not None:
+        role = couch_user.get_role(domain)
```
czue: I think this might be able to just return `couch_user.can_view_any_reports(domain)`. that logic looks more complicated but I think it works and might be nicer to keep everything in one place.

corehq/apps/reports/util.py
proteusvacuum: might be good to add a test for this

proteusvacuum (commented):
Will wait for followup on cory's points.

corehq/apps/reports/util.py
emord: i looked for that and wondered why that method wasn't a thing. guess i passed over it https://github.com/dimagi/commcare-hq/pull/13268/commits/235aad1506d3d3cd64e2436ccb36a5c1d356c9bf

PR #13269, opened 2016-09-16
adding management command to force user 412

calellowitz:
This is a management command to aid in the process when users get into a bad sync state and we need to force a 412. The state hash for the new simplified sync logs is computed from `case_ids_on_phone` so this will make sure the state hash on hq does not match the phone one.
@czue and @proteusvacuum seem to have the most context on sync work
@snopoke code buddy
cc: @ctsims @amstone326


corehq/ex-submodules/casexml/apps/phone/management/commands/invalidate_sync_heads.py
```diff
@@ -0,0 +1,33 @@
...
+            reduce=False,
+            include_docs=True,
+        )
+
+        logs = []
+        for res in results:
+            log = SimplifiedSyncLog.wrap(res['doc'])
+            log.case_ids_on_phone = {'broken to force 412'}
+            logs.append(log)
+        SimplifiedSyncLog.bulk_save(logs)
```
dimagimon: W292 no newline at end of file

corehq/ex-submodules/casexml/apps/phone/management/commands/invalidate_sync_heads.py
```diff
@@ -0,0 +1,33 @@
...
+            endkey=[user_id, date],
+            descending=True,
+            reduce=False,
+            include_docs=True,
+        )
+
+        logs = []
+        for res in results:
+            log = SimplifiedSyncLog.wrap(res['doc'])
+            log.case_ids_on_phone = {'broken to force 412'}
```
czue: I wonder if we want to be more explicit about this (e.g. add a special property we can set to trigger this in restore code as opposed to using the hashing algorithm which is a bit of a hack). I think we would want to do that if we ended up using this type of mechanism more.

proteusvacuum: I agree with this. I don't think it would be too hard to add a `log.force_412` property and just check if that is true when the hash check is executed.

corehq/ex-submodules/casexml/apps/phone/management/commands/invalidate_sync_heads.py
```diff
@@ -0,0 +1,33 @@
...
+    Forces a 412 for a given user by creating bad state in the all synclogs
+    for the given user after the given date
+    """
+    args = "user_id date"
+
+    def handle(self, *args, **options):
+        if len(args) != 2:
+            raise CommandError("Usage is ./manage.py invalidate_sync_heads %s" % self.args)
+        user_id = args[0]
+        date = args[1]
```
proteusvacuum: what format is the date in here? I think this could use some validation, as I'd be wary of accidentally blowing away information on synclongs that we don't need to.

corehq/ex-submodules/casexml/apps/phone/management/commands/invalidate_sync_heads.py
```diff
@@ -0,0 +1,33 @@
...
+            "phone/sync_logs_by_user",
+            startkey=[user_id, {}],
+            endkey=[user_id, date],
+            descending=True,
+            reduce=False,
+            include_docs=True,
+        )
+
+        logs = []
+        for res in results:
```
proteusvacuum: Should we not do this only on the latest result? This seems like it will set this property on all synclogs for that specific date, but I think we only want the last successful one.

PR #13273, opened 2016-09-16
Avoid Location.get_db()

esoergel:
@czue @orangejenny

### DIFF
--- a/corehq/apps/locations/tests/test_location_sync.py
+++ b/corehq/apps/locations/tests/test_location_sync.py
@@ -42,7 +42,6 @@ class TestLocationSync(TestCase):
         cls.state = loc_types['state']
         cls.county = loc_types['county']
         cls.city = loc_types['city']
-        cls.db = Location.get_db()
 
     @classmethod
     def tearDownClass(cls):
@@ -68,7 +67,7 @@ class TestLocationSync(TestCase):
 
     def assertNumLocations(self, number):
         self.assertEqual(SQLLocation.objects.count(), number)
-        self.assertEqual(get_doc_count_by_type(self.db, 'Location'), number)
+        self.assertEqual(get_doc_count_by_type(Location.get_db(), 'Location'), number)
 
     def test_sync_couch_to_sql(self):
         mass = couch_loc("Massachusetts", self.state)
--- a/corehq/apps/locations/views.py
+++ b/corehq/apps/locations/views.py
@@ -502,14 +502,11 @@ def location_descendants_count(request, domain, loc_id):
 @can_edit_location
 @location_safe
 def unarchive_location(request, domain, loc_id):
-    # hack for circumventing cache
-    # which was found to be out of date, at least in one case
-    # http://manage.dimagi.com/default.asp?161454
-    # todo: find the deeper reason for invalid cache
-    loc = Location.get(loc_id, db=Location.get_db())
-    if loc.domain != domain:
+    try:
+        loc = SQLLocation.objects.get(domain=domain, location_id=loc_id)
+    except SQLLocation.DoesNotExist:
         raise Http404()
-    loc.unarchive()
+    loc.couch_location.unarchive()
     return json_response({
         'success': True,
         'message': _("Location '{location_name}' has successfully been {action}.").format(
--- a/corehq/apps/users/cases.py
+++ b/corehq/apps/users/cases.py
@@ -26,17 +26,21 @@ def get_wrapped_owner(owner_id):
             'CommCareUser': CommCareUser,
             'WebUser': WebUser,
             'Group': Group,
-            'Location': Location,
         }.get(doc_type)
 
-    for db in [user_db(), Location.get_db()]:
-        try:
-            owner_doc = db.get(owner_id)
-        except ResourceNotFound:
-            continue
-        else:
-            cls = _get_class(owner_doc['doc_type'])
-            return cls.wrap(owner_doc) if cls else None
+    try:
+        return Location.get(owner_id)
+    except ResourceNotFound:
+        pass
+
+    try:
+        owner_doc = user_db().get(owner_id)
+    except ResourceNotFound:
+        pass
+    else:
+        cls = _get_class(owner_doc['doc_type'])
+        return cls.wrap(owner_doc) if cls else None
+
     return None
 
 
--- a/corehq/apps/users/models.py
+++ b/corehq/apps/users/models.py
@@ -1689,8 +1689,6 @@ class CommCareUser(CouchUser, SingleMembershipMixin, CommCareMobileContactMixin)
                 "Attempting to access multiple locations for a user in a domain that does not support this."
             )
 
-        from corehq.apps.locations.models import Location
-
         def _get_linked_supply_point_ids():
             mapping = self.get_location_map_case()
             if mapping:
@@ -1700,12 +1698,11 @@ class CommCareUser(CouchUser, SingleMembershipMixin, CommCareMobileContactMixin)
         def _get_linked_supply_points():
             return SupplyInterface(self.domain).get_supply_points(_get_linked_supply_point_ids())
 
-        def _gen():
-            location_ids = [sp.location_id for sp in _get_linked_supply_points()]
-            for doc in iter_docs(Location.get_db(), location_ids):
-                yield Location.wrap(doc)
-
-        return list(_gen())
+        location_ids = [sp.location_id for sp in _get_linked_supply_points()]
+        return list(SQLLocation.objects
+                    .filter(domain=self.domain,
+                            location_id__in=location_ids)
+                    .couch_locations())
 
     def supply_point_index_mapping(self, supply_point, clear=False):
         from corehq.apps.commtrack.exceptions import (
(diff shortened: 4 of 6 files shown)


czue:
build is failing:
```
ERROR: test_report_runner (custom.ilsgateway.tests.test_report_runner.TestReportRunner)
  File "/mnt/commcare-hq-ro/custom/ilsgateway/tests/test_report_runner.py", line 105, in test_report_runner
    report_run(TEST_DOMAIN)
  File "/vendor/lib/python2.7/site-packages/celery/app/task.py", line 420, in __call__
    return self.run(*args, **kwargs)
  File "/mnt/commcare-hq-ro/corehq/util/decorators.py", line 122, in _inner
    return fn(*args, **kwargs)
  File "/mnt/commcare-hq-ro/custom/ilsgateway/tasks.py", line 73, in report_run
    populate_report_data(run.start, run.end, domain, run, strict=strict)
  File "/mnt/commcare-hq-ro/custom/ilsgateway/tanzania/warehouse/updater.py", line 255, in populate_report_data
    for org in chunk
  File "/vendor/lib/python2.7/site-packages/celery/canvas.py", line 362, in __call__
    return self.apply_async(args, kwargs)
  File "/vendor/lib/python2.7/site-packages/celery/canvas.py", line 251, in apply_async
    return _apply(args, kwargs, **options)
  File "/vendor/lib/python2.7/site-packages/celery/app/builtins.py", line 275, in apply_async
    return self.apply(args, kwargs, **options)
  File "/vendor/lib/python2.7/site-packages/celery/app/builtins.py", line 302, in apply
    last and (last.get(), ),
  File "/vendor/lib/python2.7/site-packages/celery/canvas.py", line 181, in apply
    return self.type.apply(args, kwargs, **options)
  File "/vendor/lib/python2.7/site-packages/celery/app/task.py", line 743, in apply
    request=request, propagate=throw)
  File "/vendor/lib/python2.7/site-packages/celery/app/trace.py", line 355, in eager_trace_task
    uuid, args, kwargs, request)
  File "/vendor/lib/python2.7/site-packages/celery/app/trace.py", line 253, in trace_task
    I, R, state, retval = on_error(task_request, exc, uuid)
  File "/vendor/lib/python2.7/site-packages/celery/app/trace.py", line 240, in trace_task
    R = retval = fun(*args, **kwargs)
  File "/mnt/commcare-hq-ro/custom/ilsgateway/tanzania/warehouse/updater.py", line 493, in process_non_facility_warehouse_data
    % (location.name, str(location.location_id), len(facs)))
TypeError: object of type 'itertools.imap' has no len()

ERROR: test_can_clear_locations (corehq.apps.commtrack.tests.test_locations.MultiLocationsTest)
  File "/mnt/commcare-hq-ro/corehq/apps/commtrack/tests/test_locations.py", line 210, in test_can_clear_locations
    self.assertEqual(len(user.locations), 0)
  File "/mnt/commcare-hq-ro/corehq/apps/users/models.py", line 1702, in locations
    return list(SQLLocation.objects
NameError: global name 'SQLLocation' is not defined
```

czue:
can you explain what this gets us towards? guessing you have a plan but at first blush it looks like it mostly just makes the code more complicated.

esoergel:
@czue Sure, in order to get off of couch locations, we have to replace anything that uses `Location.get_db()`, since that's (presumably) going to be using some generic couch code to get a Location document.  Assumiing we'll keep the `Location` class around, just no longer as a `Document` class, `Location.get()` is easy enough to implement as `SQLLocation.objects.get(location_id=...)`.  However, `db.get(location_id)` is more problematic.

That said, I'm certainly open to other alternatives.

czue:
>  Assuming we'll keep the `Location` class around, just no longer as a `Document` class

Surprised we would do this - are you saying we would still have `SQLLocation` and `Location` as two separate classes? Would we not want to reconcile them?

One suggestion I have is that instead of removing one particular dependency (`get_db`) we should strive to encapsulate any API that is dependent on couch to a small set of known methods (a-la dbaccessors). So in this PR when you get a single location, remove the couch dependencies from those apis (so e.g. make a new helper function and raise `LocationNotFound` rather than `ResourceNotFound`). That way the part of that code that depends on couch is clearly encapsulated and all the calling code should not need to be touched when couch locations are gone.

czue:
>  Assuming we'll keep the `Location` class around, just no longer as a `Document` class

Surprised we would do this - are you saying we would still have `SQLLocation` and `Location` as two separate classes? Would we not want to reconcile them?

One suggestion I have is that instead of removing one particular dependency (`get_db`) we should strive to encapsulate any API that is dependent on couch to a small set of known methods (a-la dbaccessors). So in this PR when you get a single location, remove the couch dependencies from those apis (so e.g. make a new helper function and raise `LocationNotFound` rather than `ResourceNotFound`). That way the part of that code that depends on couch is clearly encapsulated and all the calling code should not need to be touched when couch locations are gone.

esoergel:
@czue 
> Surprised we would do this - are you saying we would still have SQLLocation and Location as two separate classes? Would we not want to reconcile them?

I'd definitely like to reconcile them eventually, but I was thinking that we'd have an intermediate state where `Location` is a wrapper around `SQLLocation`.  My main concern with getting rid of it entirely is that half of the usages of `Location` are in `custom/`, so they're difficult to test and even run.  Although if we're confident in API compatability, I guess that shouldn't be a problem.

The reason I'm trying to remove `get_db` in particular (as opposed to say, `Location.get`) is because it allows you to query locations using means outside of the `Location` class (which is I think what you're saying).

I like the suggestion about a generic exception, though.  What do you think of this approach?
```python
class LocationNotFound(ResourceNotFound, SQLLocation.DoesNotExist):
    pass
```
That would make it less likely to break existing exception handling.

If you've got time when you come in town for the summit, it'd be great to have a sit-down to go over the approach for the remainder of this migration (I'm on interrupt, so I won't be doing much work on it 'till after then anyways).

czue:
that the exception approach looks convenient, although my personal opinion is to always make things explicit, which I think it does not do a great job of. when we did the scale migration the way we approached it was to define the set of APIs that we needed to support and then make sure that all calling code was updated to use those APIs. whereas I think the strategy you're proposing is more to continue to use the built in couch/django APIs but add some stuff that make it a little easier to be agnostic. personally I think we should strive to replace every instance of `Location.get` with something else rather than try and make that particular function backend-agnostic, and then we can just update the exception handling whenever we update the surrounding code to call the new accessors. That has the benefit of making it clear whether something has been "migrated" or not.

This is definitely subjective so might be easier to take to voice if that doesn't resonate (and also makes sense that all the custom code and lack of tests complicates things here). 

> I'd definitely like to reconcile them eventually, but I was thinking that we'd have an intermediate state where `Location` is a wrapper around `SQLLocation`. 

What's the benefit of this? If you need to wrap the functions anyways, could you not just implement them on `SQLLocation` and then always deal with `SQLLocation` objects?

esoergel:
I agree with getting rid of `Location.get`, and that should be pretty easy, since that's `grep`able.  This PR was to get rid of APIs that are difficult to replace outright with a SQLLocation version - that is, `iter_docs`, `get_document_or_404`, `get_docs`, and `db.get`.  After this PR (and another one to actually deprecate `get_db`), the only ways to access couch locations will be through methods on the `Location` class and through `locations_queryset.couch_locations()`.

I think this is essentially what you're describing about having supported APIs, except I've been thinking about having these accessors be methods on the class they return, so instead of replacing:
`get_couch_location_by_site_code` with `get_sql_location_by_site_code`
I've been thinking of replacing:
`Location.get_by_site_code` with `SQLLocation.objects.get_by_site_code`

[this](https://docs.google.com/document/d/11P4ngUibAHpq2sHAyMpzxGXi5i6s1NreOIe9OfCnx5k/edit#) is the high-level plan I've been working from.
And [here](https://trello.com/b/NGCEjsgK/locations-tech-debt-https-docs-google-com-document-d-11p4nguibahpq2shaympzxgxi5i6s1nreoie9ofcnx5k-edit) are specific actionables.

I'm not married to having `Location` be a wrapper around `SQLLocation`, basically my thoughts there were that we need to wean stuff of the `Location` class.  There are two ways to order that:

1. move things one-by-one, then drop couch and the sync code
2. drop couch and the sync code (and write wrap code instead), *then* move things one-by-one

I was thinking the one-by-one part would take a long time, particularly with the custom code, so it'd be best to get rid of the complex sync code earlier if possible.  The api compatibility is almost there though, so this might not actually be such an issue.

czue:
> [this (https://docs.google.com/document/d/11P4ngUibAHpq2sHAyMpzxGXi5i6s1NreOIe9OfCnx5k/edit#) is the high-level plan I've been working from.

funny i took this same position 9 months ago at the bottom of that doc. at least I'm consistent. 

> The api compatibility is almost there though, so this might not actually be such an issue.

Yeah this is what I was thinking. I guess my main concern is that not knowing whether the object you are working with is a couch location or a sql location is almost as bad from a tech debt perspective as having the locations in two databases and it seems easier to just do everything once than in two stages... but you're closer to it so I'm up for deferring to you if you're confident that's the easiest phasing and we won't just continue to be in a limbo/confusing state for ages.

Either way, do you want this change in or do you want to clean it up a bit based on this conversation (e.g. remove `Location.get`)?

esoergel:
@czue I'm ready to get back to this now.  We offlined about this, and what I remember from that was that I'd like to get this in now, then remove `Location.get` in a separate PR actually converting existing usages to SQL.

czue:
yep, that sounds right to me

PR #13281, opened 2016-09-19
make custom data in excel optional

sravfeyn:
@esoergel

### DIFF
--- a/corehq/apps/locations/bulk_management.py
+++ b/corehq/apps/locations/bulk_management.py
@@ -207,6 +207,7 @@ class LocationStub(object):
     @property
     @memoized
     def custom_location_data(self):
+        # This just compiles the custom location data, the validation is done in _custom_data_errors()
         if self.custom_data is self.NOT_PROVIDED or self.uncategorized_data is self.NOT_PROVIDED:
             # if either of these are not provided, then existing data should be updated, not overridden
             metadata = copy.copy(self.db_object.metadata)
--- a/corehq/apps/locations/bulk_management.py
+++ b/corehq/apps/locations/bulk_management.py
@@ -126,6 +126,7 @@ class LocationStub(object):
     A representation of location excel row
     """
     titles = LOCATION_SHEET_HEADERS
+    NOT_PROVIDED = 'NOT_PROVIDED_IN_EXCEL'
     meta_data_attrs = ['name', 'site_code', 'latitude', 'longitude', 'external_id']
 
     def __init__(self, name, site_code, location_type, parent_code, location_id,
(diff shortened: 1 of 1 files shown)


corehq/apps/locations/bulk_management.py
```diff
@@ -168,8 +169,19 @@ def from_excel_row(cls, row, index, location_type):
         longitude = row.get(cls.titles['longitude'])
```
sravfeyn: Didn't actually make lat/lng as optional since we expect users to just use the downloaded excel file which already contains these

sravfeyn:
@esoergel pinging again in case you have missed it

corehq/apps/locations/bulk_management.py
```diff
@@ -189,15 +201,23 @@ def lookup_old_collection_data(self, old_collection):
...
+            # if either of these are not provided, then existing data should be updated, not overridden
+            metadata = copy.copy(self.db_object.metadata)
+        else:
+            # if both of these are provided, then existing data should be overridden
+            metadata = {}
+        if self.custom_data != self.NOT_PROVIDED:
+            metadata.update(self.custom_data)
+        if self.uncategorized_data != self.NOT_PROVIDED:
+            metadata.update(self.uncategorized_data)
+        return metadata
```
esoergel: You might want to consider using `CustomDataEditor` for this.  I don't think there are any system fields in location metadata, but if there were, we'd need to preserve those fields - they shouldn't be editable via upload, and they shouldn't be overwritten.  Anyways, that class has a `get_data_to_save` method which would be useful here.

corehq/apps/locations/bulk_management.py
```diff
@@ -326,10 +346,11 @@ def validate_and_parse_stubs_from_excel(self):
 
         # all locations sheets should have correct headers
         location_stubs = []
+        optional_headers = [LOCATION_SHEET_HEADERS['custom_data'], LOCATION_SHEET_HEADERS['uncategorized_data']]
         for sheet_name, sheet_reader in sheets_by_title.items():
             if sheet_name != self.types_sheet_title:
-                actual = set(sheet_reader.fieldnames)
-                expected = set(LOCATION_SHEET_HEADERS.values())
+                actual = set(sheet_reader.fieldnames) - set(optional_headers)
+                expected = set(LOCATION_SHEET_HEADERS.values()) - set(optional_headers)
```
esoergel: I think these aren't necessarily all optional, you can specify some fields as required (see https://www.commcarehq.org/a/ethan-commtrack/settings/locations/fields/ for example).  It's probably fine to skip those titles for the header validation (as you've done here), but then I think we need the `CustomDataEditor.is_valid` thing to make sure that users aren't uploading invalid custom data.

I'd also be okay with skipping custom data stuff altogether for this first pass and just saying we don't yet support it through upload, then getting that fix in separately.

Sorry, I should've caught this in an earlier pass, but I didn't think of it.

corehq/apps/locations/bulk_management.py
sravfeyn: Going to fully respond tomorrow, but I am doing the validation already [here](https://github.com/dimagi/commcare-hq/blob/master/corehq/apps/locations/bulk_management.py#L598), this method assumes validation is over

esoergel: Gotcha, that makes sense.  In that case my only comment is, would it be possible to use the validator's `get_data_to_save` functionality here?

corehq/apps/locations/bulk_management.py
esoergel: Cool, so if they don't include a column for a required field, then it'll fail in a different place?

corehq/apps/locations/bulk_management.py
sravfeyn: So, I am using the bulk way of validating custom data as noted here. If there were no errors the data is good the way it is, so this method should return correct data, I will add a note that this assumes validation is done. Using `get_data_to_save` means making `CustomDataEditor` work in bulk way (it would have to fetch the definition for each row).

corehq/apps/locations/bulk_management.py
sravfeyn: These fields are the excel headers like `name, code, parent_code, lat` etc, not the custom location data fields.  These two lines make the excel headers `data, uncategorized_data` optional. (sorry if you mean something else)

esoergel: No, I think we're on the same page here.

PR #13282, opened 2016-09-19
Clarify inactive domain message

esoergel:
This message is shown when an active, logged-in user views a domain that hasn't yet been activated.  That is, a domain that was created by a new user who has never confirmed their email (or hasn't yet confirmed it).
http://manage.dimagi.com/default.asp?237559

@czue @orangejenny


orangejenny:
I'm not sure this is clarifying things that much, but it isn't hurting anything, either.

PR #13283, opened 2016-09-19
force manual linking for child module forms

snopoke:
http://manage.dimagi.com/default.asp?237546
@calellowitz


snopoke:
:white_check_mark:

PR #13285, opened 2016-09-19
Short-circuit context-lookups for remote apps

esoergel:
http://manage.dimagi.com/default.asp?237621
I believe the next block of context is irrelevant for remote apps, @dannyroberts, can you confirm?

At any rate, this allows the application page to load for remote apps


esoergel:
@czue @orangejenny

PR #13286, opened 2016-09-19
Pass in dashboard permissions

emord:
http://manage.dimagi.com/default.asp?237478#1220484

old exports were unable to create dashboards. @NoahCarnahan any reason why this was implemented as another parameter instead of as part of the `CustomExportView` (i'm not that familiar with KO patterns)


emord:
@esoergel can you review/merge this before next deploy as well

NoahCarnahan:
Thanks @emord. There is no particular reason I didn't add this to the first parameter of `CustomExportView.wrap`. I think I was under the impression that `CustomExportView.wrap` was wrapping a model passed from Django, and not an object that was being built in JS anyways.

PR #13287, opened 2016-09-19
set soft time limit in celery

benrudolph:
@gcapalbo how do you feel about setting a time limit for our celery tasks? something long, but just in case a task goes terribly this'll at least kill it (or optionally handle the exception)


gcapalbo:
i think that's a good idea, thanks. i wonder if this will still work if a child process is running without the parent worker process, or if we'll need a different way to take care of those situations.

either way, good to merge on :white_check_mark: (kicked the build to be safe, though this probably shouldn't affect tests)

benrudolph:
i think in the current situation we should never had a child process running without the parent worker process, so it should still work (if you're referring to the way we're spawning a second celery worker process, in general if a child process gets orphaned from its parent process i think we're out of luck)

PR #13289, opened 2016-09-19
Refactor debugger to not use second call

wpride:
We can include the instanceXml in every form response easily and cheaply,  so do so here. Saves us a second call to Formplayer and some callbacks between JS modules. LMK if you think there's a problem with doing it this way.

Also displays XML on form load instead of waiting for first answer

cross https://github.com/dimagi/formplayer/pull/139

### DIFF
--- a/corehq/apps/cloudcare/static/cloudcare/js/form_entry/webformsession.js
+++ b/corehq/apps/cloudcare/static/cloudcare/js/form_entry/webformsession.js
@@ -329,6 +329,9 @@ WebFormSession.prototype.answerQuestion = function(q) {
         },
         function(resp) {
             $.publish('session.reconcile', [resp, q]);
+            if (self.answerCallback !== undefined) {
+                self.answerCallback(self.session_id);
+            }
         });
 };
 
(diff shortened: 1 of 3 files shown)


corehq/apps/cloudcare/static/cloudcare/js/form_entry/webformsession.js
```diff
@@ -193,6 +222,7 @@ WebFormSession.prototype.handleSuccess = function(resp, callback) {
         self.lastRequestHandled = resp.seq_id;
 
         try {
+            self.displayInstanceXml(resp);
```
benrudolph: can we put this after the callback call? if this fails then it'll actually block rendering

corehq/apps/cloudcare/static/cloudcare/js/form_entry/webformsession.js
```diff
@@ -299,9 +329,6 @@ WebFormSession.prototype.answerQuestion = function(q) {
         },
         function(resp) {
             $.publish('session.reconcile', [resp, q]);
-            if (self.answerCallback !== undefined) {
-                self.answerCallback(self.session_id);
-            }
```
benrudolph: this'll break cloudcare debugger in old cloudcare

PR #13290, opened 2016-09-19
Make Spanish translations appear on HQ

esoergel:
Michelle is working on a proposal to get HQ translated into Spanish.  I noticed that we're already set up for translations on transifex, but we don't actually display those translations.  Turns out that's simply because `'es'` isn't in `settings.LANGUAGES`.  Ideally we'd only show options from `settings.LANGUAGES` in the dropdown where you select a language (since those are the only ones we'll actually sorta translate), but we currently need to keep the full list of every language because that's used in cloudcare for app translations.

Read the commits separately, the messages have more context.

### DIFF
--- a/settings.py
+++ b/settings.py
@@ -140,9 +140,9 @@ CSRF_SOFT_MODE = True
 
 MIDDLEWARE_CLASSES = [
     'corehq.middleware.NoCacheMiddleware',
-    'django.middleware.common.CommonMiddleware',
     'django.contrib.sessions.middleware.SessionMiddleware',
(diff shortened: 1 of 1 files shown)


esoergel:
@czue @orangejenny

orangejenny:
Slightly surprised that the middleware change isn't worth running tests for, is that overly paranoid?

esoergel:
Couldn't hurt, this won't be deployed until tomorrow anyways.

orangejenny:
thanks :)

PR #13292, opened 2016-09-19
don't worry about #.history.# table

benrudolph:
@czue @NoahCarnahan lobbying to just remove this table altogether. all the exports i've seen look like this:

screen shot 2016-09-19 at 6 19 41 pm

cc: @gcapalbo


czue:
sounds good to remove to me.

PR #13295, opened 2016-09-20
change misnomer

sravfeyn:
@nickpell


nickpell (approved):

nickpell:
kicked build

emord:
I think this should be changed in corehq/apps/hqwebapp/views.py as well https://github.com/dimagi/commcare-hq/pull/13268/files

Perhaps you hadn't pulled latest?

sravfeyn:
Good point, thanks! Updated

PR #13300, opened 2016-09-20
allow env specification

benrudolph:
@gcapalbo this'll allow us to send messages to the staging channel. specifically for when we initiate staging deploys


gcapalbo (approved):
nice

gcapalbo:
(merging as this doesn't affect build)

PR #13301, opened 2016-09-20
Fix title block for formplayer

benrudolph:
@wpride 
http://manage.dimagi.com/default.asp?237714
cc: @gcapalbo

### DIFF
--- a/corehq/apps/cloudcare/templates/cloudcare/formplayer_home.html
+++ b/corehq/apps/cloudcare/templates/cloudcare/formplayer_home.html
@@ -4,11 +4,7 @@
 {% load compress %}
 
 {% block title %}
-    {% if request.couch_user.is_web_user %}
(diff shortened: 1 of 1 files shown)


corehq/apps/cloudcare/templates/cloudcare/formplayer_home.html
```diff
@@ -4,11 +4,7 @@
 {% load compress %}
 
 {% block title %}
-    {% if request.couch_user.is_web_user %}
-        {% trans "Preview Form" %}
```
benrudolph: since web user's can now use cloudcare just like a mobile worker, i think it makes sense to remove this

wpride: trans "Nimbus"?

PR #13302, opened 2016-09-20
Always add known case updates

benrudolph:
@NoahCarnahan this ensures that we always add known case updates to the export

cc: @gcapalbo


NoahCarnahan (approved):
👍

benrudolph:
ok i think it's fair to merge this. the tests keep timing out and the previous commit they passed. the only change on the last commit is a bump in the version number

PR #13312, opened 2016-09-20
auto case update enhancements

calellowitz:
@gcapalbo cc: @esoergel (for final look at the form)
best reviewed 🐡


corehq/apps/data_interfaces/models.py
```diff
@@ -12,6 +14,15 @@
 ALLOWED_DATE_REGEX = re.compile('^\d{4}-\d{2}-\d{2}')
 AUTO_UPDATE_XMLNS = 'http://commcarehq.org/hq_case_update_rule'
 
+class PropertyTypeChoices(object):
```
dimagimon: E302 expected 2 blank lines, found 1

corehq/apps/data_interfaces/models.py
```diff
@@ -213,12 +252,21 @@ class AutomaticUpdateAction(models.Model):
...
+    property_value_type = models.CharField(max_length=15,
+                                           choices=PropertyTypeChoices.choices,
+                                           default=PropertyTypeChoices.EXACT)
+
     class Meta:
         app_label = "data_interfaces"
+
+
+# import signals
+from corehq.apps.data_interfaces import signals
```
dimagimon: F401 'signals' imported but unused

corehq/apps/data_interfaces/signals.py
```diff
@@ -0,0 +1,23 @@
...
+    """
+    from corehq.apps.data_interfaces.tasks import run_case_update_rules_on_save
+
+    if FormAccessors(case.domain).get_form(case.xform_ids[-1]).xmlns != AUTO_UPDATE_XMLNS \
+            and AUTO_CASE_UPDATE_ENHANCEMENTS.enabled(case.domain):
+        run_case_update_rules_on_save.delay(case.domain, case.type, case.case_id)
+
+
+case_post_save.connect(case_changed_receiver, CommCareCase)
+sql_case_post_save.connect(case_changed_receiver, CommCareCaseSQL)
```
dimagimon: W292 no newline at end of file

corehq/apps/data_interfaces/views.py
```diff
@@ -25,7 +25,7 @@
     AddCaseGroupForm, UpdateCaseGroupForm, AddCaseToGroupForm,
     AddAutomaticCaseUpdateRuleForm)
 from corehq.apps.data_interfaces.models import (AutomaticUpdateRule,
-    AutomaticUpdateRuleCriteria, AutomaticUpdateAction)
+                                                AutomaticUpdateRuleCriteria, AutomaticUpdateAction, PropertyTypeChoices)
```
dimagimon: E501 line too long (120 > 115 characters)

corehq/apps/data_interfaces/models.py
calellowitz: this is necessary for the signals to be registered

corehq/apps/data_interfaces/models.py
```diff
@@ -213,12 +253,21 @@ class AutomaticUpdateAction(models.Model):
...
```
dimagimon: F401 'signals' imported but unused

corehq/apps/data_interfaces/models.py
esoergel: I believe the preferred approach is to import in `AppConfig.ready`
https://docs.djangoproject.com/en/1.10/ref/applications/#django.apps.AppConfig.ready

corehq/apps/data_interfaces/models.py
dimagimon: F401 'signals' imported but unused

gcapalbo:
@calellowitz i haven't gotten a chance yet to review this but will do so

corehq/apps/data_interfaces/models.py
```diff
@@ -12,6 +14,15 @@
 ALLOWED_DATE_REGEX = re.compile('^\d{4}-\d{2}-\d{2}')
 AUTO_UPDATE_XMLNS = 'http://commcarehq.org/hq_case_update_rule'
 
+class PropertyTypeChoices(object):
+    EXACT = "Exact"
+    CASE_PROPERTY = "Case Property"
```
gcapalbo: i generally prefer these constants to be code-like, for example, `EXACT` and `CASE_PROPERTY`

corehq/apps/data_interfaces/models.py
```diff
@@ -49,18 +61,26 @@ def organize_rules_by_case_type(cls, rules):
 
     @classmethod
     def get_boundary_date(cls, rules, now):
-        min_boundary = min([rule.server_modified_boundary for rule in rules])
+        min_boundary = None
+        for rule in rules:
+            if not rule.server_modified_boundary:
```
gcapalbo: would be better if it's `not rule.filter_on_server_modified`

corehq/apps/data_interfaces/models.py
```diff
@@ -49,18 +61,26 @@ def organize_rules_by_case_type(cls, rules):
 
     @classmethod
     def get_boundary_date(cls, rules, now):
```
gcapalbo: should add a comment that says it returns None if any of the rules don't filter on server modified

corehq/apps/data_interfaces/models.py
```diff
@@ -76,25 +96,44 @@ def _rule_matches_case(self, case, now):
...
+        cases_to_update = defaultdict(dict)
         close = False
 
+        def _get_case_property_value(current_case, name):
+            return current_case.resolve_case_property(name)[0].value
+
+        def _add_update_property(name, value, current_case):
+            while name.startswith('parent/'):
+                name = name[7:]
+                current_case = current_case.parent
```
gcapalbo: we'll want to instead use `current_case.get_parent(identifier=DEFAULT_PARENT_IDENTIFIER)` (which returns a list that can be empty) and take the first one. `current_case.parent` assumes the relationship is normal parent/child without allowing for other relationships, like extension case relationships

corehq/apps/data_interfaces/models.py
```diff
@@ -76,25 +96,44 @@ def _rule_matches_case(self, case, now):
...
+            cases_to_update[current_case.case_id][name] = value
+
         for action in self.automaticupdateaction_set.all():
             if action.action == AutomaticUpdateAction.ACTION_UPDATE:
-                properties[action.property_name] = action.property_value
+                # break this out as helper function?
```
gcapalbo: i don't think it's necessary to break it out but i'd prefer if we don't leave this comment if that's what we decide

corehq/apps/data_interfaces/models.py
gcapalbo: it's preferable to add these constants to the class where they're used instead of creating a new object to hold them

corehq/apps/data_interfaces/models.py
```diff
@@ -76,25 +96,44 @@ def _rule_matches_case(self, case, now):
...
+                    value = action.property_value
+
+                if value != _get_case_property_value(case, action.property_name):
+                    _add_update_property(action.property_name, value, case)
             elif action.action == AutomaticUpdateAction.ACTION_CLOSE:
                 close = True
 
-        update_case(case.domain, case.case_id, case_properties=properties, close=close,
-            xmlns=AUTO_UPDATE_XMLNS)
+        for id, properties in cases_to_update.items():
```
gcapalbo: nice, thanks for adding the ability to update parent cases too

corehq/apps/data_interfaces/forms.py
```diff
@@ -191,66 +205,89 @@ def __init__(self, *args, **kwargs):
...
+                'update_property_value',
+                ng_model='update_property_value',
+            )
+        ]
+
+        if self.enhancements_enabled:
+            self.allow_updates_without_closing()
+        else:
+            # do not allow them to edit property_value_type
+            _update_property_fields.pop(1)
```
gcapalbo: i think it's better to conditionally insert rather than conditionally remove since if the order changes and we forget to update the index we end up removing the wrong element with this approach (instead of just inserting in the wrong place which isn't as bad)

corehq/apps/data_interfaces/forms.py
```diff
@@ -191,66 +205,89 @@ def __init__(self, *args, **kwargs):
...
+            ),
+            Div(
+                *_update_property_fields,
+                ng_show='showUpdateProperty()'
+            )
+        ]
+
+        if not self.enhancements_enabled:
+            # do not allow them to edit filter_on_server_modified
+            _basic_info_fields.pop(2)
```
gcapalbo: (same comment about insert vs pop)

corehq/apps/data_interfaces/forms.py
```diff
@@ -270,9 +307,12 @@ def __init__(self, *args, **kwargs):
...
+    def _valid_boundary(self, value):
+        return self.enhancements_enabled or (isinstance(value, int) and value > 30)
+
     def clean_server_modified_boundary(self):
         value = self.cleaned_data.get('server_modified_boundary')
-        if not isinstance(value, int) or value < 30:
+        if not self._valid_boundary(value):
             raise ValidationError(_("Please enter a whole number greater than "
                                     "or equal to 30."))
         return value
```
gcapalbo: we should make this return None if we're not filtering on server modified boundary, so that it gets set to None in the rule

corehq/apps/data_interfaces/signals.py
```diff
@@ -0,0 +1,23 @@
...
+from corehq.form_processor.signals import sql_case_post_save
+
+
+def case_changed_receiver(sender, case, **kwargs):
+    """
+    Spawns a task to run auto update rules tied to the given case.
+    """
+    from corehq.apps.data_interfaces.tasks import run_case_update_rules_on_save
+
+    if FormAccessors(case.domain).get_form(case.xform_ids[-1]).xmlns != AUTO_UPDATE_XMLNS \
```
gcapalbo: it would be better if the first check we did was the feature flag check, so that we can short circuit most of the time

corehq/apps/data_interfaces/signals.py
```diff
@@ -0,0 +1,23 @@
...
+from casexml.apps.case.signals import case_post_save
+
+from corehq.apps.data_interfaces.models import AUTO_UPDATE_XMLNS
+from corehq.form_processor.interfaces.dbaccessors import FormAccessors
+from corehq.toggles import AUTO_CASE_UPDATE_ENHANCEMENTS
+from corehq.form_processor.models import CommCareCaseSQL
+from corehq.form_processor.signals import sql_case_post_save
+
+
+def case_changed_receiver(sender, case, **kwargs):
```
gcapalbo: we should also wrap all of the code in this function into a try/except so that we never raise exceptions during the signal processing (django runs all signals synchronously). we can notify exception in the except clause

corehq/apps/data_interfaces/signals.py
gcapalbo: we should also move the FormAccessors check to the celery task itself. it's ok that we spawn a task for every case update, and it's better that we make the signal processing be as quick as possible.

corehq/apps/data_interfaces/tasks.py
```diff
@@ -111,3 +113,18 @@ def run_case_update_rules_for_domain(domain, now=None):
...
             rule.save()
+
+
+@task(queue='background_queue', acks_late=True, ignore_result=True)
+def run_case_update_rules_on_save(domain, case_type, case_id):
+    rules = AutomaticUpdateRule.by_domain(domain).filter(case_type=case_type)
+    case = CaseAccessors(domain).get_case(case_id)
+    now = datetime.utcnow()
+    key = 'case-update-on-save-case-{case}'.format(case=case_id)
+    with CriticalSection([key]):
```
gcapalbo: entering the critical section should be the first thing we do in this function

corehq/apps/data_interfaces/tasks.py
```diff
@@ -111,3 +113,18 @@ def run_case_update_rules_for_domain(domain, now=None):
...
+        for rule in rules:
+            stop_processing = rule.apply_rule(case, now)
+            rule.last_run = now
```
gcapalbo: i think it's better if we don't update the last_run property actually, that was more intended to be the last scheduled run (i.e., against all cases in the domain)

corehq/apps/data_interfaces/forms.py
gcapalbo: (never mind i see you addressed this in a later commit which is better than insert)

corehq/apps/data_interfaces/signals.py
gcapalbo: (i see you updated the task spawn criteria in a later commit so ignore my first comment. but we should still move the FormAccessors check to the task)

corehq/apps/data_interfaces/app_config.py
```diff
@@ -0,0 +1,17 @@
...
+
+from corehq.apps.data_interfaces.signals import case_changed_receiver
+
+
+class DataInterfacesAppConfig(AppConfig):
+    name = 'data_interfaces'
+
+    def ready(self):
+        case_post_save.connect(case_changed_receiver, CommCareCase)
+        sql_case_post_save.connect(case_changed_receiver, CommCareCaseSQL)
```
gcapalbo: i've never done it this way before but i can't tell from reading [this](https://docs.djangoproject.com/en/1.10/ref/applications/#django.apps.AppConfig.ready) whether we should be importing models from other modules here or not. to be safe i think we may want to just use a string label for the sender rather than the model class

corehq/apps/data_interfaces/app_config.py
```diff
@@ -0,0 +1,17 @@
...
+from casexml.apps.case.models import CommCareCase
+from casexml.apps.case.signals import case_post_save
+
+from corehq.form_processor.models import CommCareCaseSQL
+from corehq.form_processor.signals import sql_case_post_save
+
+from corehq.apps.data_interfaces.signals import case_changed_receiver
+
+
+class DataInterfacesAppConfig(AppConfig):
```
gcapalbo: actually, looks like an easy workaround is [here](https://docs.djangoproject.com/en/1.10/topics/signals/#preventing-duplicate-signals)

calellowitz: i couldnt get the string labels to work but I did move the imports into the ready function which seems to also be supported.

corehq/apps/data_interfaces/tasks.py
calellowitz: I had to keep the critical section after making sure the case hadnt been updated by this task, otherwise the second instantiation was hitting against the critical section until timing out.

calellowitz:
@gcapalbo I tried to address all the feedback. I added comments in the two places I did not.

corehq/apps/data_interfaces/models.py
```diff
@@ -114,13 +106,13 @@ def _get_case_property_value(current_case, name):
         def _add_update_property(name, value, current_case):
             while name.startswith('parent/'):
                 name = name[7:]
-                current_case = current_case.parent
+                # uses first parent if there are multiple
+                current_case = current_case.get_parent(identifier=DEFAULT_PARENT_IDENTIFIER)[0]
```
gcapalbo: this would raise an exception if there is no parent (should check that len() > 0)

corehq/apps/data_interfaces/tasks.py
gcapalbo: i think you're seeing that because you must have run the function synchronously right? if run asynchronously it should just make the second task block until it's allowed to process while the first finishes, which is what we want

corehq/apps/data_interfaces/tasks.py
```diff
@@ -111,3 +113,21 @@ def run_case_update_rules_for_domain(domain, now=None):
         for rule in rules:
             rule.last_run = now
             rule.save()
+
+
+@task(queue='background_queue', acks_late=True, ignore_result=True)
+def run_case_update_rules_on_save(case):
+    key = 'case-update-on-save-case-{case}'.format(case=case.case_id)
+    with CriticalSection([key]):
+        import ipdb; ipdb.set_trace()
```
dimagimon: E702 multiple statements on one line (semicolon)

gcapalbo:
@calellowitz this should be good now, i tested locally and made some fixes to this pr. just waiting for the build to pass on the second to last commit and then we can merge

PR #13317, opened 2016-09-21
Parse celery workers into ones we expect to be running and stopped

benrudolph:
@gcapalbo how do you feel about this for the updated check? it's not dependent on my other PR so can merge whenever you're happy with it

cc: @proteusvacuum

### DIFF
--- a/corehq/apps/hqadmin/service_checks.py
+++ b/corehq/apps/hqadmin/service_checks.py
@@ -141,7 +141,7 @@ def check_heartbeat():
                 bad_workers.append('* {} celery worker down'.format(hostname))
 
         for hostname in expected_stopped:
-            if not all_workers[hostname]:
+            if all_workers[hostname]:
                 bad_workers.append(
                     '* {} celery worker is running when we expect it to be stopped.'.format(hostname)
                 )
--- a/corehq/apps/hqadmin/service_checks.py
+++ b/corehq/apps/hqadmin/service_checks.py
@@ -22,6 +22,7 @@ from corehq.apps.es import GroupES
 from corehq.blobs import get_blob_db
 from corehq.elastic import send_to_elasticsearch
 from corehq.util.decorators import change_log_level
+from corehq.apps.hqadmin.utils import parse_celery_workers
 
 ServiceStatus = namedtuple("ServiceStatus", "success msg")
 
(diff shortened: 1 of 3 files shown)


benrudolph:
✔️

corehq/apps/hqadmin/utils.py
```diff
@@ -84,3 +85,32 @@ def get_task_html(detailed_stats, worker_name):
...
+    def _strip_timestamp(hostname):
+        return '.'.join(hostname.split('.')[:-1])
+
+    timestamped_workers = sorted(timestamped_workers, key=_strip_timestamp)
+
+    for hostname, group in groupby(timestamped_workers, _strip_timestamp):
+
+        sorted_workers = sorted(list(group), reverse=True)
+        expect_running.append(sorted_workers.pop(0))
+        expect_stopped.extend(sorted_workers)
```
gcapalbo: i'm not sure this will always work. this is still assuming that every worker we expect to be up will be returned by the flower api which i don't think is an assumption we want to make (for example, if flower is started when workers are already down). i think we might just need to explicitly list the workers we expect to be up.

benrudolph: i think i might be missing something, but this list is strictly workers we expect to be down. which are those that have a timestamp that is earlier than any worker with the same hostname and timestamp. most of the time this'll be empty. are you worried about the case where flower starts before the most recent timestamped worker will start?

benrudolph: ah sorry misread your comment. i think it's OK if flower starts before the workers do. flower will then pick up those workers in the status api as they come online

benrudolph:
(we should just quickly offline about this tomorrow)

corehq/apps/hqadmin/service_checks.py
```diff
@@ -133,9 +134,18 @@ def check_heartbeat():
...
-        for hostname, status in all_workers.items():
-            if not status:
+        expected_running, expected_stopped = parse_celery_workers(all_workers)
+
+        for hostname in expected_running:
+            if not all_workers[hostname]:
                 bad_workers.append('* {} celery worker down'.format(hostname))
+
+        for hostname in expected_stopped:
+            if not all_workers[hostname]:
```
gcapalbo: i just noticed this - we should remove the `not` here right?

benrudolph: oh shoot you're totally right

gcapalbo:
:+1: on :white_check_mark:

benrudolph:
One potential thing to flag is the fact we are string sorting numbers. I
don't think this matters since barring some crazy event, the epoch from one
celery worker to the next one should have the same number of digits and
thus should work to sort on

PR #13319, opened 2016-09-21
Catch improperly formatted search string query

esoergel:
@dannyroberts


esoergel:
http://manage.dimagi.com/default.asp?237756

esoergel:
bump
also @biyeun new buddy

PR #13322, opened 2016-09-21
Je/ucr es reports

emord:
This is a work in progress and littered with TODO's but I'm going offline starting tomorrow night and probably won't get back to this until after the tech summit, so throwing this up here so that it can be reviewed while its fresh.

That said, it's quite functional and can currently handle report builder generated reports for pie/bar charts, case/form list and data tables.

@czue @nickpell @NoahCarnahan


corehq/apps/userreports/reports/data_source.py
```diff
@@ -1,6 +1,7 @@
 from corehq.apps.userreports.models import DataSourceConfiguration, get_datasource_config
+from corehq.apps.userreports.es.data_source import ConfigurableReportEsDataSource
 from corehq.apps.userreports.sql.data_source import ConfigurableReportSqlDataSource
-from corehq.apps.userreports.util import get_table_name
+from corehq.apps.userreports.util import get_table_name, get_backend_id
```
dimagimon: F401 'get_table_name' imported but unused

corehq/apps/userreports/reports/data_source.py
```diff
@@ -46,10 +46,6 @@ def lang(self, lang):
         self.data_source.lang = lang
 
     @property
-    def aggregation_columns(self):
-        return self.data_source.aggregation_columns
```
czue: nice

corehq/apps/userreports/es/data_source.py
```diff
@@ -44,29 +44,25 @@ def __init__(self, domain, config_or_config_id, filters, aggregation_columns, co
...
-    # def column_configs(self):
-    #     return self._column_configs.values()
+    @property
+    def column_configs(self):
+        return self._column_configs.values()
 
     @property
     def table_name(self):
-        return get_table_name(self.domain, self.config.table_id)
+        # TODO make this the same function as the adapter
```
czue: not sure it matters but i'm not sure i understand this todo. is the adapter the `IndicatorAdapter`?

corehq/apps/userreports/es/columns.py
```diff
@@ -0,0 +1,26 @@
+from corehq.apps.reports.datatables import (
+    DataTablesColumn,
+    DataTablesHeader,
+    DTSortType,
+)
+
+
+class EsColumnConfig(object):
```
czue: does this need to be different from `SqlColumnConfig`? I guess the `.columns` are different types?

corehq/apps/userreports/es/data_source.py
```diff
@@ -168,50 +152,6 @@ def required_fields(self):
...
-    #     # for columns that end up being complex queries (e.g. aggregate dates)
-    #     # there could be more than one column ID and they may specify aliases
-    #     if column_id in self._column_configs:
-    #         return self._column_configs[column_id].get_query_column_ids()
-    #     else:
-    #         # if the column isn't found just treat it as a normal field
-    #         return [column_id]
+    @method_decorator(catch_and_raise_exceptions)
+    def get_total_row(self):
+        return []
```
czue: #todo?

corehq/apps/userreports/tests/test_report_aggregation.py
```diff
@@ -64,16 +66,69 @@ def _create_data_source(cls):
                 },
             ],
         )
+        # this is a hack to have both sql and es backends created in a class
+        # method. alternative would be to have these created on each test run
```
czue: didn't look with a fine tooth comb, but is the whole thing copy/pasted except the backend ID? seems like it would be more DRY to make these in a loop over `backend_id`s

czue (changes requested):
> it's quite functional and can currently handle report builder generated reports for pie/bar charts, case/form list and data tables.

This is pretty heroic. Well done.

In general it feels like there's a fair bit more copy/paste in this set of changes than the data source side. Is that because you think it was necessary to architect that way or didn't have the time to e.g. build out shared functions or a base class / api for the data sources/columns/etc. and just wanted to put it up for review?

Also note that I didn't review any detailed logic very closely for performance/ES usage/correctness (though hopefully tests catch the last one). Guessing we could do that all at once when things are more stable/fina.

corehq/apps/userreports/es/data_source.py
emord: yeah. was a todo for me to remember that this wasn't DRY

corehq/apps/userreports/es/columns.py
emord: Good point. I don't think it does. I basically ended up copy/pasting that entire file when I was trying to figure out distinct values in ES, so that came with it. Don't think I gave it too much thought since it was in `sql/columns.py`

corehq/apps/userreports/tests/test_report_aggregation.py
emord: Agreed. hoping to get rid of all the hacks before the final revision :)

emord:
> In general it feels like there's a fair bit more copy/paste in this set of changes than the data source side. Is that because you think it was necessary to architect that way or didn't have the time to e.g. build out shared functions or a base class / api for the data sources/columns/etc. and just wanted to put it up for review?

Yeah. there is a bit more copy paste than I would want for a final implementation. I think that's mostly the result of playing around with various reports and seeing what I was missing to get those to work. It also wasn't always obvious what what used outside of `userreports` and what was actually a private method.

> Also note that I didn't review any detailed logic very closely for performance/ES usage/correctness (though hopefully tests catch the last one).

I also haven't reviewed too closely for performance. Was there any set of tests or use cases you used for testing sql performance when originally creating ucrs?

corehq/apps/userreports/es/columns.py
```diff
@@ -0,0 +1,122 @@
...
+    :param column_config:
+    :param expansion_limit:
+    :return:
+    """
+    from corehq.apps.userreports.es.adapter import IndicatorESAdapter
+    adapter = IndicatorESAdapter(data_source_configuration)
+    query = adapter.get_query_object()
+    too_many_values = False
+
+    distinct_values = query.distinct_values(column_config.field, expansion_limit)
```
NoahCarnahan: I think the second argument to `distinct_values()` should be `expansion_limit + 1`? Otherwise I don't think `len(distinct_values)` could ever be greater than `expansion_limit`

corehq/apps/userreports/es/data_source.py
```diff
@@ -0,0 +1,234 @@
...
+                for col in [self._column_configs[sort_column_id]]
+            ]
+        elif self.column_configs:
+            col = self.column_configs[0]
+            return [(col.field, ASCENDING)]
+        return []
+
+    @property
+    def columns(self):
+        db_columns = [c for sql_conf in self.es_column_configs for c in sql_conf.columns]
```
NoahCarnahan: not a big deal but might want to change `sql_conf` to `es_conf` or something like that

NoahCarnahan (commented):
Looking good. I also didn't try to validate the ES usage, but happy to look more closely at that part if that would be helpful.

czue:
> I also haven't reviewed too closely for performance. Was there any set of tests or use cases you used for testing sql performance when originally creating ucrs?

No, not really at all. I think as long as we're doing things like pushing pagination and aggregation to the DB it should be at least equivalent to the postgres versions.

emord:
> No, not really at all. I think as long as we're doing things like pushing pagination and aggregation to the DB it should be at least equivalent to the postgres versions.

I brought this up when I was looking at paginating the other standard reports, but ES doesn't have an equivalent to `offset` or `start` when using aggregations. This is why I have two different queries in the `get_data` function right now.

I don't think that this is an issue since we limit the number of categories you can have in the graphs and things like case/form list do not use aggregations. It should only be an issue when a data table is being scrolled to x number of rows (I don't know what x is in this case, and it will likely depend on lots of other things). We should add analytics to these reports to see when someone attempts to scroll through x number of rows in the report so that we can monitor these sort of things.

emord:
I don't think that this PR is going to get merged in its current state. I'll be refactoring and pulling out pieces for easier review

PR #13323, opened 2016-09-21
Make ICDS Issue Tracker reports visible

sheelio:
To fix http://manage.dimagi.com/default.asp?237678


czue:
build fail not related

PR #13324, opened 2016-09-21
API throttling follow-up

esoergel:
@snopoke
Followup from the API rate limiting issues that happened recently
http://manage.dimagi.com/default.asp?237829
http://manage.dimagi.com/default.asp?237763
commit-by-commit will make the most sense


snopoke:
:boot: :construction:

snopoke (approved):

esoergel:
The accursed `150 != 50` failure claims another victim
:cry:

PR #13325, opened 2016-09-22
Abt custom report: Update Zambia target configurations again

NoahCarnahan:
@kaapstorm


NoahCarnahan:
self merging because no test coverage and client wants these pronto

PR #13333, opened 2016-09-22
removing broken combobox refresh

calellowitz:
http://manage.dimagi.com/default.asp?236728#1221807
The combobox was changed to a select2 a while ago but the js was never updated. The refresh step isnt necessary on a select2.
@NoahCarnahan @esoergel @sravfeyn cc: @gcapalbo


NoahCarnahan (approved):
👍

calellowitz:
✅

PR #13335, opened 2016-09-22
duplicate indicators in live domain

snopoke:
@czue @calellowitz 

The V2 app has now been copied into the live domain so the V2 indicator names need to work in the live domain.


czue:
build timed out but i think this is fine

PR #13338, opened 2016-09-22
use fast counting to get SQL doc counts

snopoke:
This is used by the reindexer's for progress calculations. It's a nice to have.
@czue

### DIFF
--- a/corehq/form_processor/backends/sql/dbaccessors.py
+++ b/corehq/form_processor/backends/sql/dbaccessors.py
@@ -69,6 +69,13 @@ class ReindexAccessor(six.with_metaclass(ABCMeta)):
     # TODO: implement this https://wiki.postgresql.org/images/3/35/Pagination_Done_the_PostgreSQL_Way.pdf
 
     @abstractproperty
+    def model_class(self):
+        """
+        :return: the django model class belonging to this reindexer
+        """
+        raise NotImplementedError
+
+    @abstractproperty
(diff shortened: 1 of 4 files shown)


corehq/util/doc_processor/sql.py
```diff
@@ -87,4 +85,16 @@ def get_document_iterator(self, chunk_size, event_handler=None):
         )
 
     def get_total_document_count(self):
-        return -1
+        return sum(
+            self.reindex_accessor.get_doc_count(from_db)
+            for from_db or 'default' in _get_db_aliases_to_query()
```
dimagimon: E901 SyntaxError: invalid syntax

corehq/form_processor/backends/sql/dbaccessors.py
```diff
@@ -101,13 +108,27 @@ def get_docs(self, from_db, startkey, last_doc_pk=None, limit=500):
         """
         raise NotImplementedError
 
+    def get_doc_count(self, from_db):
+        """Get the doc count from the given DB
+        :param from_db: The DB alias to query
+        """
+        sql_query = "SELECT reltuples FROM pg_class WHERE oid = '{}'::regclass"
```
snopoke: This relies on the autovacuum daemon which updates the stats regularly: https://wiki.postgresql.org/wiki/Count_estimate

This will return an estimate of the row count.

PR #13340, opened 2016-09-22
fix reindex queries

snopoke:
@czue 

I discovered a serious issue with these queries. Because the inner queries don't sort each query returns a somewhat arbitrary set of rows so the more pages you have the more rows you are likely to miss.

These queries should do index only scans: https://wiki.postgresql.org/images/3/35/Pagination_Done_the_PostgreSQL_Way.pdf

### DIFF
--- a/corehq/form_processor/backends/sql/dbaccessors.py
+++ b/corehq/form_processor/backends/sql/dbaccessors.py
@@ -66,8 +66,6 @@ def get_cursor(model):
 
 
 class ReindexAccessor(six.with_metaclass(ABCMeta)):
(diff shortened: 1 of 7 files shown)


czue (approved):
looks good. just for my own education what was the original purpose of the inner queries (sorry, didn't read the presentation closely - feel free to tell me to look there)?

czue:
kicked build

czue:
do we need to do a reindex?

snopoke:
Yes, this requires a reindex which I'll run.

snopoke:
To answer your question about the purpose of the inner queries - I wrote those before knowing about index only scans. I was trying to write an efficient query that didn't rely of 'OFFSET' (which is slow).

czue:
:+1: thanks

PR #13342, opened 2016-09-23
Inline edit widget: do nothing on blur

orangejenny:
http://manage.dimagi.com/default.asp?237858

@biyeun


czue:
nice, thanks for super fast turnaround! what was the original purpose of that block/design?

orangejenny:
Don't recall, but it was most likely based off of similar widgets. I believe this is how Trello card descriptions used to work - although they're now updated so that when you click away, the widget reverts to read-only and shows the old value, but also stores the new value and indicates that you have unsaved changes.

PR #13348, opened 2016-09-23
Make sure mobile select filter exists in report

esoergel:
http://manage.dimagi.com/default.asp?237873
@dannyroberts @orangejenny


orangejenny:
Don't have a lot of context, but code looks reasonable.

PR #13353, opened 2016-09-23
Enter key does search

wpride:


czue:
yes! :white_check_mark: when the lint error is addressed

PR #13354, opened 2016-09-23
help-message

sravfeyn:
@millerdev


orangejenny:
If this is J2ME-only, shouldn't it just move into the J2ME settings section?

sravfeyn:
Good point, I was UI blind! I will move it.

sravfeyn:
@orangejenny Updated

PR #13360, opened 2016-09-26
Clear breadcrumbs on back

benrudolph:
@wpride @biyeun this clears the breadcrumbs when people hit the back arrow to go to the main screen

cc: @proteusvacuum


wpride:
:+1: on ✅

PR #13364, opened 2016-09-26
validate app in direct_ccz

benrudolph:
@wpride this should catch the errors we were missing before. don't merge yet because we need to write code for formplayer to handle `error_html`

cc: @proteusvacuum


wpride:
@benrudolph  niceee

snopoke:
sorry, merged premeturely, will revert

PR #13374, opened 2016-09-27
move indirect models.py imports out of __init__.py

nickpell:
@proteusvacuum 

Needed for django 1.9

### DIFF
--- a/custom/enikshay/__init__.py
+++ b/custom/enikshay/__init__.py
@@ -1,1 +1,0 @@
-from custom.enikshay.integrations.ninetyninedots.repeater_generators import RegisterPatientPayloadGenerator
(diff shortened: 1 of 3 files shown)


proteusvacuum:
Hey, this change is causing the New "Register Patient Repeater" page to complain with 
```
KeyError: <class 'custom.enikshay.integrations.ninetyninedots.repeaters.NinetyNineDotsRegisterPatientRepeater'>
```

nickpell:
@proteusvacuum Is there another place we can put this?

proteusvacuum:
what is the requirement? just not have anything in `__init__.py`?

nickpell:
There can't be any imports of ```models.py``` in ```__init__.py```.  That's both direct and indirect imports.

proteusvacuum:
@nickpell I updated this. The `enikshay` app is never loaded in settings (because there is nothing there), so that code was never getting run.

nickpell:
Great, thanks!

PR #13380, opened 2016-09-28
return if already datetime for validate_phone_datetime

mkangia:
http://manage.dimagi.com/default.asp?237990

Follow up on : http://manage.dimagi.com/default.asp?233241

### DIFF
--- a/corehq/ex-submodules/casexml/apps/case/tests/test_extract_caseblocks.py
+++ b/corehq/ex-submodules/casexml/apps/case/tests/test_extract_caseblocks.py
@@ -1,7 +1,19 @@
-import uuid
-from django.test import SimpleTestCase
+import uuid, os
+from datetime import datetime, timedelta
+from django.test import SimpleTestCase, TestCase
+from django.template import Template, Context
+
+from dimagi.utils.parsing import json_format_datetime
+
+from corehq.apps.receiverwrapper.util import submit_form_locally
+from corehq.util.test_utils import TestFileMixin
+from corehq.form_processor.interfaces.dbaccessors import FormAccessors
+from casexml.apps.case.tests.util import delete_all_xforms
 from casexml.apps.case.const import CASE_ATTR_ID
 from casexml.apps.case.xform import extract_case_blocks
+
+CREATE_XFORM_ID = "6RGAZTETE3Z2QC0PE2DKM88MO"
+TEST_DOMAIN_NAME = 'test-domain'
 
 
 class TestExtractCaseBlocks(SimpleTestCase):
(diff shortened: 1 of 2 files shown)


mkangia:
@calellowitz

corehq/ex-submodules/casexml/apps/case/util.py
```diff
@@ -21,6 +21,9 @@
 
 
 def validate_phone_datetime(datetime_string, none_ok=False, form_id=None):
+    if isinstance(datetime_string, datetime.datetime):
```
czue: is there just a specific workflow where this is called with a datetime and not a string? just wondering if it could be handled in the calling code rather than by making this function more ambiguous about what types it expects.

mkangia: @czue its being called from only one place.. will try to locate the scenario which leads to this.

mkangia: @czue : its definitely happening from [here](https://github.com/dimagi/commcare-hq/blob/master/corehq/apps/reports/tasks.py#L298) since forms are fetched via FormAccessors resulting in already parsed `datetime.datetime` object and not a string.
I am not sure but i guess this was recently added as shared by Robin "I was testing to see the new multimedia export ....".

Returning the value if its already a datetime.datetime object seems ok even otherwise. Though the argument name `datetime_string` conflicts.
If we want to make this better we can either change [here](https://github.com/dimagi/commcare-hq/blob/master/corehq/ex-submodules/casexml/apps/case/xform.py#L373) to validate only if its a string or if its not a datetime.datetime object already (preferred since None and empty string need are handled differently).

I think we are missing test cases for this hence it was not noticed earlier.

czue: got it. so just to make sure I understand, the main issue is that when the form is being parsed for the first time during submission the datetimes are not converted from strings when they enter `_extract_case_blocks`, but once they are coming out of the database they are?

If that's right this approach sounds good to me. Might be worth a quick comment explaining it above, but otherwise this seems good to merge. Just holding off to make sure I've understood it.

czue (approved):

corehq/ex-submodules/casexml/apps/case/util.py
czue: I guess the other takeaway is that it would be good to have a test for this

mkangia:
@czue i am not sure if there are more places for form submission event to occur but when the form is submitted for the first time the datetime_string is a string only (tested this locally as well).

Tracing from the exception, in case when FormAccessors is used for this particular [method](https://github.com/dimagi/commcare-hq/blob/master/corehq/apps/reports/tasks.py#L286) to fetch forms it has datetime as a datetime object. I am not sure if thats the same for all places where forms are fetched.

corehq/ex-submodules/casexml/apps/case/tests/test_extract_caseblocks.py
```diff
@@ -1,8 +1,20 @@
-import uuid
-from django.test import SimpleTestCase
+import uuid, os
```
dimagimon: E401 multiple imports on one line

corehq/ex-submodules/casexml/apps/case/util.py
mkangia: @czue test case added. Should wait for a green build.

corehq/ex-submodules/casexml/apps/case/tests/test_extract_caseblocks.py
```diff
@@ -75,3 +87,40 @@ def test_blocks_in_repeat(self):
...
+        final_xml = self._formatXForm(CREATE_XFORM_ID, xml_data, {})
+        response, form, cases = submit_form_locally(
+            final_xml,
+            TEST_DOMAIN_NAME,
+            attachments={},
+            last_sync_token=None,
+            received_on=None
+        )
+        xform = FormAccessors(TEST_DOMAIN_NAME).get_form(form.get_id)
+        extract_case_blocks(xform.form_data)
```
czue: sweet. assume this line failed before the change?

czue (approved):

corehq/ex-submodules/casexml/apps/case/tests/test_extract_caseblocks.py
mkangia: Yup. This fails without the fix

PR #13385, opened 2016-09-29
bind to select2 on page load

mkangia:
http://manage.dimagi.com/default.asp?237988

@orangejenny


orangejenny:
sweet, thanks for tracking this down!

PR #13384, opened 2016-09-28
pillow retry fix off 2016 09 26 23.06 production deploy

snopoke:
@esoergel

PR #13388, opened 2016-09-29
Revert "pillow retry fix off 2016 09 26 23.06 production deploy"

gcapalbo:
Reverts dimagi/commcare-hq#13384
@esoergel @snopoke looks like this might be what caused the pillow retry issues. right now i've stopped the pillow retry enqueuing process, but that will restart if we deploy. so we shouldn't deploy unless we either revert this or fix it, i'll leave to you which makes the most sense.


corehq/couchapps/domain_shows/shows/domain_date.js
```diff
@@ -0,0 +1,56 @@
+function (doc, req) {
```
dimagimon: Parsing error: Unexpected token (

snopoke:
https://github.com/dimagi/commcare-hq/pull/13389

PR #13391, opened 2016-09-30
Fetch domain by name if not found by id when attempting to make a copy

mkangia:
http://manage.dimagi.com/default.asp?237994

Not sure yet why this is needed but seems like the needed fix.
domain fetched by id and by_name at places around this module/functionality.
Used hint from ProjectInformationView#project.

### DIFF
--- a/corehq/apps/hqwebapp/templatetags/hq_shared_tags.py
+++ b/corehq/apps/hqwebapp/templatetags/hq_shared_tags.py
@@ -517,6 +517,7 @@ class IncludeAwareNode(loader_tags.IncludeNode):
             include_count = context.render_context._includes = []
         return include_count
 
+
 @register.simple_tag(takes_context=True)
 def url_replace(context, field, value):
     """Usage <a href="?{% url_replace 'since' restore_id %}">
@@ -528,3 +529,9 @@ def url_replace(context, field, value):
     params = context['request'].GET.copy()
     params[field] = value
     return params.urlencode()
+
+
+@register.filter
+def view_pdb(element):
+    import ipdb; ipdb.set_trace()
+    return element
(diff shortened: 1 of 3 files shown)


corehq/apps/hqwebapp/templatetags/hq_shared_tags.py
```diff
@@ -528,3 +529,9 @@ def url_replace(context, field, value):
     params = context['request'].GET.copy()
     params[field] = value
     return params.urlencode()
+
+
+@register.filter
+def view_pdb(element):
+    import ipdb; ipdb.set_trace()
```
dimagimon: E702 multiple statements on one line (semicolon)

esoergel:
I made a big reply in the fogbugz ticket.  I don't think this is quite the right fix, but it's close.  See there for more.

mkangia:
@esoergel  Thanks Ethan. Will follow up...

mkangia:
@esoergel updated to use id for generating url instead of fetching domain by name.

corehq/apps/hqwebapp/templatetags/hq_shared_tags.py
esoergel: How does this guy work?

esoergel:
:+1: Thanks!

corehq/apps/hqwebapp/templatetags/hq_shared_tags.py
mkangia: @esoergel this can be used in view as a filter {{ some_element|view_pdb }} to debug the element. Once finished returns the element.

esoergel: :sunglasses:

PR #13395, opened 2016-09-30
prorate excess user unit cost when invoice is for a partial month

nickpell:
Needed for Oct 1 invoicing so we don't double charge clients for the full user cost if they have more than one subscription over the course of the invoicing period.

@esoergel @emord

### DIFF
--- a/corehq/apps/accounting/invoicing.py
+++ b/corehq/apps/accounting/invoicing.py
@@ -400,6 +400,22 @@ class LineItemFactory(object):
         except KeyError:
             raise LineItemError("No line item factory exists for the feature type '%s" % feature_type)
 
+    @property
+    @memoized
+    def is_prorated(self):
+        return not (
+            self.invoice.date_end.day == self._days_in_billing_period
+            and self.invoice.date_start.day == 1
+        )
+
+    @property
+    def num_prorated_days(self):
+        return self.invoice.date_end.day - self.invoice.date_start.day + 1
+
+    @property
+    def _days_in_billing_period(self):
+        return calendar.monthrange(self.invoice.date_end.year, self.invoice.date_end.month)[1]
+
 
 class ProductLineItemFactory(LineItemFactory):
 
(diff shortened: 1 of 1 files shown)


corehq/apps/accounting/invoicing.py
```diff
@@ -491,6 +491,17 @@ def unit_cost(self):
 class UserLineItemFactory(FeatureLineItemFactory):
 
     @property
+    def unit_cost(self):
+        non_prorated_unit_cost = super(UserLineItemFactory, self).unit_cost
+        if self.is_prorated:
+            return Decimal(
+                "%.2f" % round(
+                    non_prorated_unit_cost * self.num_prorated_days / self._days_in_billing_period, 2
```
esoergel: Are you sure this isn't ever doing integer division?  I think as long as `non_prorated_unit_cost` is a Decimal it should behave as expected.

esoergel (commented):
g2m if you're confident that's never integer division

corehq/apps/accounting/invoicing.py
nickpell: Added a ```1.0``` to be safe

esoergel (approved):

esoergel:
Build failure potentially related:
```
ERROR: test_manual_adjustment (corehq.apps.accounting.tests.test_forms.TestAdjustBalanceForm)
  File "/mnt/commcare-hq-ro/corehq/apps/accounting/tests/test_forms.py", line 25, in test_manual_adjustment
    original_balance = self.invoice.balance
AttributeError: 'NoneType' object has no attribute 'balance'
```

corehq/apps/accounting/invoicing.py
nickpell: turns out the ```1.0 *``` produced an unsupported operator error (mixing ```float``` and ```Decimal```) so I decided to just assert that ```non_prorated_unit_cost``` is of type ```Decimal```

nickpell:
tests pass, self-merging

PR #13398, opened 2016-09-30
Show notifications when someone enters or saves a form you have open

czue:
image

this is currently behind a feature flag though i'd be interested in input from @dimagi/product on what would be required in terms of UAT / blocking release

@snopoke buddy. might be easier all at once.


corehq/apps/app_manager/static/app_manager/js/app-notifications.js
```diff
@@ -0,0 +1,24 @@
+/* globals hqDefine moment */
+hqDefine('app_manager/js/app-notifications.js', function () {
+    function NotifyFunction(userId, $element) {
+        return function(msg) {
+            var msgObj = JSON.parse(msg);
+            // only show notifcations from other users
+            if (msgObj.user_id != userId) {
```
dimagimon: Expected '!==' and instead saw '!='. (eqeqeq)

corehq/apps/app_manager/static/app_manager/js/app-notifications.js
```diff
@@ -0,0 +1,24 @@
...
+                    $inner.html(dismissButton);
+                    $element.removeClass('hidden').append($inner);
+                } else {
+                    $inner = $($element.children()[0]);
+                }
+                $inner.append(moment(msgObj.timestamp).format('h:mm:ss a') + ': ' + msgObj.text + '<br>');
+            }
+        };
+
+    };
```
dimagimon: Unnecessary semicolon. (no-extra-semi)

corehq/apps/app_manager/views/notifications.py
```diff
@@ -0,0 +1,37 @@
...
+from dimagi.utils.parsing import json_format_datetime
+
+
+def notify_form_opened(domain, couch_user, app_id, unique_form_id):
+    message = _('This form has been opened for editing by {}.').format(couch_user.username)
+    notify_event(domain, couch_user, app_id, unique_form_id, message)
+
+
+def notify_form_changed(domain, couch_user, app_id, unique_form_id):
+    message = _('This form has been updated by {}. Reload the page to see the latest changes.').format(couch_user.username)
```
dimagimon: E501 line too long (123 > 115 characters)

corehq/apps/app_manager/views/notifications.py
```diff
@@ -0,0 +1,37 @@
...
+    if APP_BUILDER_NOTIFICATIONS.enabled(domain):
+        message = {
+            'domain': domain,
+            'user_id': couch_user._id,
+            'username': couch_user.username,
+            'text': message,
+            'timestamp': json_format_datetime(datetime.datetime.utcnow()),
+        }
+        message = RedisMessage(json.dumps(message))
+        RedisPublisher(facility=get_facility_for_form(domain, app_id, unique_form_id), broadcast=True).publish_message(message)
```
dimagimon: E501 line too long (127 > 115 characters)

corehq/apps/app_manager/templates/app_manager/partials/form_designer_requirejs_config.html
```diff
@@ -5,6 +5,7 @@
 define("jquery", [], function () { return window.jQuery; });
 define("jquery.bootstrap", ["jquery"], function () {});
 define("underscore", [], function () { return window._; });
+define("moment", [], function () { return window.moment; });
```
czue: I was running into some issues with moment and requirejs. Not sure if this was the right way to do this but it worked. @orangejenny do you own this?

settingshelper.py
```diff
@@ -172,12 +172,19 @@ def celery_failure_handler(task, exc, task_id, args, kwargs, einfo):
 
 
 def get_allowed_websocket_channels(request, channels):
+    from django.conf import settings
     from django.core.exceptions import PermissionDenied
+    from corehq.apps.users.models import CouchUser
     if request.user and request.user.is_authenticated() and request.user.is_superuser:
         return channels
     else:
+        couch_user = CouchUser.from_django_user(request.user)
```
czue: in [the docs for the framework we use](https://django-websocket-redis.readthedocs.io/en/latest/usage.html#safety-considerations) it says that this function shouldn't perform any blocking requests, though it doesn't really explain why. I'm not sure if this only works in development mode, but i figured I could test on prod and then decide what to do.

corehq/apps/app_manager/static/app_manager/js/app-notifications.js
```diff
@@ -0,0 +1,24 @@
+/* globals hqDefine moment */
+hqDefine('app_manager/js/app-notifications.js', function () {
+    function NotifyFunction(userId, $element) {
+        return function(msg) {
+            var msgObj = JSON.parse(msg);
+            // only show notifcations from other users
+            if (msgObj.user_id !== userId) {
+                var $inner;
+                // if nothing in the outer element, add the notification div and populate
```
orangejenny: This is pretty similar to what [notify_user](https://github.com/dimagi/commcare-hq/blob/master/corehq/apps/style/templates/style/base.html#L399-L407) does, can you use that instead?

orangejenny (commented):
Just looked over the javascript, not reviewing thoroughly.

corehq/apps/app_manager/static/app_manager/js/app-notifications.js
czue: nice, didn't know about that. hopefully i can

orangejenny:
This looks awesome! And so useful!

UX thoughts related to releasing this more widely:
- I can imagine users having false expectations: that they'll get a notification when the other person stops viewing the form; that there will be a notification displayed when they open the form if anyone else currently has it open. Should we try to mitigate these?
- Looks like the alert will get taller each time another person opens or saves the form. It's great that this will be easily noticed, but it might not be a great experience if the alert is repeatedly growing. Could do something like initially render the alert with the timestamp extra noticeable like this...
screen shot 2016-09-30 at 6 11 58 pm
...then change the timestamp to regular text after a few seconds. Then later if there's a new message, just update the existing message, but also again use the extra styling so it's eye-catching. Then again, maybe this is enough of an edge case that we don't care about it.
- If you close the alert and then another message comes in, the alert gets re-created, right? This sounds annoying if there are repeated messages. But, again, maybe this is an edge case.
- This kind of interacts with vellum's warning popup when you overwrite another user's changes. Might want to double-check that the language here and in that popup is similar.
- Should we filter out users who aren't on the project? I wonder if it'll be confusing (or creepy) for end users to be notified when someone from Dimagi is looking at their app for debugging purposes. Might be enough of an edge case not to matter.
- It would be great to do this on other complex pages eventually (form settings and module settings come to mind), though we'd then need to be careful that people don't expect it on every page ad assume that they're "safe" to edit a page when actually that page doesn't have this implemented. But that's a future concern, don't think it's an issue if vellum is the only page that has these notifications.

corehq/apps/app_manager/static/app_manager/js/app-notifications.js
czue: I looked at this and the code is way simpler, but the UI feels a bit clunkier (to me). Not sure whether you think better to do it this way? 

image

czue:
thanks for the great feedback on the wider release. won't respond specifically to comments yet, but will make a gdoc or something to track things. @dimagi/product this might be a good use case for the "provide framework/support for devs to do the product work on their own features" action item from the summit

corehq/apps/app_manager/static/app_manager/js/app-notifications.js
orangejenny: Hrm, yeah, it is uglier, I'd just been thinking of it from the code perspective.

It would actually be nice for `alert_user` to have an option to update an existing message rather than create a new one (the multiple alerts thing happens occasionally on other pages, and it's basically always ugly), and doesn't look like it'd be hard, but I haven't thought it all the way through, so understand if you don't want to deal with that in this PR.

corehq/apps/app_manager/static/app_manager/js/app-notifications.js
```diff
@@ -0,0 +1,14 @@
+/* globals hqDefine moment message_alert */
```
dimagimon: 'message_alert' is defined but never used (no-unused-vars)

corehq/apps/app_manager/static/app_manager/js/app-notifications.js
```diff
@@ -0,0 +1,14 @@
...
+hqDefine('app_manager/js/app-notifications.js', function () {
+    function NotifyFunction(userId) {
+        return function(msg) {
+            var msgObj = JSON.parse(msg);
+            // only show notifcations from other users
+            if (msgObj.user_id !== userId) {
+                var message = moment(msgObj.timestamp).format('h:mm:ss a') + ': ' + msgObj.text;
+                alert_user(message, 'info', true);
```
dimagimon: 'alert_user' is not defined. (no-undef)

corehq/apps/app_manager/static/app_manager/js/app-notifications.js
czue: wasn't so bad https://github.com/dimagi/commcare-hq/pull/13398/commits/48f2a195b8f1670f3b5ce8b3551d92f23702d7e9

benrudolph:
this is awesome! i've always thought about integrating websockets into the app manager. not related to this, but i thought it'd be cool if the case mgmt tab was able to auto update the list of form questions available (for when you open a form builder tab and the case mgmt tab)

⬆️ ⬇️ ↕️ ↖️ ↪️

corehq/apps/app_manager/views/notifications.py
```diff
@@ -0,0 +1,41 @@
...
+
+
+def notify_form_opened(domain, couch_user, app_id, unique_form_id):
+    message = _('This form has been opened for editing by {}.').format(couch_user.username)
+    notify_event(domain, couch_user, app_id, unique_form_id, message)
+
+
+def notify_form_changed(domain, couch_user, app_id, unique_form_id):
+    message = _(
+        'This form has been updated by {}. Reload the page to see the latest changes.'
```
proteusvacuum: I've seen other sites with this functionality have the `Reload` word be a link back to the page itself (or some javascript that reloads the page). Might be nice to add that at some point.

proteusvacuum:
This is sweet

corehq/apps/app_manager/static/app_manager/js/app-notifications.js
```diff
@@ -0,0 +1,14 @@
+/* globals hqDefine moment alert_user */
+hqDefine('app_manager/js/app-notifications.js', function () {
+    function NotifyFunction(userId) {
```
emord: Might be a good use of the notification API? https://developer.mozilla.org/en-US/docs/Web/API/Notifications_API/Using_the_Notifications_API

corehq/apps/app_manager/views/notifications.py
```diff
@@ -0,0 +1,41 @@
+import json
+import datetime
+from django.utils.translation import ugettext as _
+from ws4redis.publisher import RedisPublisher
```
emord: Do you think it would be worthwhile to explore moving our current websocket usage to channels now that it's an official django project? https://www.djangoproject.com/weblog/2016/sep/09/channels-adopted-official-django-project/

Obviously not actionable now or in the near future, but will be easiest now that we only have two (I think?) pages that use this

czue: nice, yeah that's a good idea

corehq/apps/app_manager/static/app_manager/js/app-notifications.js
orangejenny: nice, thanks!

corehq/apps/app_manager/static/app_manager/js/app-notifications.js
czue: that is pretty neat, though I'm not sure an out-of-browser experience is actually superior in this case since you might have a ton of things open and want to know what it was referring to. do you mean in addition to the page-level notification?

corehq/apps/app_manager/static/app_manager/js/app-notifications.js
czue: (fixed)

czue:
@emord @orangejenny @proteusvacuum curious if there are any blockers on this? i'd be interested to play with it for the TB app. not trying to push it out if there are things you want addressed, it was just a bit hard to tell from the comments and wasn't clear to me if anyone was waiting for something specifically.

proteusvacuum (approved):
LGTM

corehq/apps/app_manager/static/app_manager/js/app-notifications.js
emord: Yeah. I originally meant in addition to. your reasoning against it makes sense. I forgot that most people have lots of forms open at once. Although could be nice to be able to hit save, get a notification and realize you have duplicate tabs open.

emord (approved):

amsagoff:
Jenny, thanks for the design review!

Cory - this looks sweet. I think the next step to pushing this out would be
to address Jenny's feedback. I think that would mean, for each feedback
item, either:

1) implementing changes to address it, or
2) getting consensus with Jenny that it is not a blocker, or
3) if you can't get consensus, I'm happy to try to get some analytics that
will help or we can run some tests to try to get more information. This
might be hallway testing, A/B testing, etc.

Might also be worth doing a UAT pass before release, depending on how much
hallway/other testing we've done in the process.

Let me know how that sounds!

czue:
@amsagoff that sounds good. i'll offline a few other thoughts.

orangejenny:
To be clear, none of my points should block merge, they're only relevant to wider release, this is good to go from my perspective.

czue:
ok cool, i think this is :white_check_mark: then

settingshelper.py
snopoke: Can you not use `request.couch_user` which should have already been populated in the middleware?

settingshelper.py
```diff
@@ -172,12 +172,19 @@ def celery_failure_handler(task, exc, task_id, args, kwargs, einfo):
...
+        for domain in couch_user.get_domains():
+            if request.path.startswith('{}{}:'.format(settings.WEBSOCKET_URL, domain)):
+                return channels
```
snopoke: Should this list not be restricted to only channels for the current domain?

PR #10225, opened 2016-02-05
Drop breadcrumbs from form edit

orangejenny:
http://manage.dimagi.com/default.asp?216566

@esoergel / @sravfeyn

PR #9588, opened 2015-12-15
Remove breadcrumbs in form builder

emord:
@amsagoff do you want the :bread:crumbs removed now or should it be coupled with :banana: ?

buddy: @esoergel

PR #13399, opened 2016-09-30
drop no_header

orangejenny:
Was added in https://github.com/dimagi/commcare-hq/pull/9588 to remove breadcrumbs...this later broke and was fixed a different way in https://github.com/dimagi/commcare-hq/pull/10225 

@biyeun 

`?w=1`


czue:
nice I just spent 15 minutes debugging why my alerts weren't showing up.

orangejenny:
ugh, sorry, didn't think to ping you ...but yep, I was poking around in this code after reading your PR

czue (approved):

PR #13409, opened 2016-09-30
update groups urls

nickpell:


nickpell:
@kaapstorm

kaapstorm:
I do love these changes. Referring to views with strings is so Django 0.9. :-)

PR #13424, opened 2016-10-03
add legend to CommCare Translations tab

orangejenny:
http://manage.dimagi.com/default.asp?237755

@biyeun


biyeun:
@orangejenny i don't think tests cover this portion. kicked the build anyway...let me know if i should ignore and merge because otherwise looks ✅

orangejenny:
let's skip the build, given that there aren't tests, and then travis can get to work on other PRs.

orangejenny:
(killed build)

PR #13430, opened 2016-10-03
Statement Period filter should default to previous period

nickpell:
Want previous period, since invoices do not get created until the current period is over.

screen shot 2016-10-03 at 1 36 31 pm

code buddy @emord

### DIFF
--- a/corehq/apps/accounting/filters.py
+++ b/corehq/apps/accounting/filters.py
@@ -3,6 +3,10 @@ import datetime
 from dateutil.relativedelta import relativedelta
 
 from django.core.urlresolvers import reverse
+from django.utils.translation import ugettext_noop as _
+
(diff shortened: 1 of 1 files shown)


corehq/apps/accounting/filters.py
```diff
@@ -344,6 +357,22 @@ class StatementPeriodFilter(OptionalMonthYearFilter):
     slug = 'statement_period'
     label = _("Statement Period")
 
+    def selected_period(self):
```
emord: Is this a function that is called somewhere already?

nickpell: I'm overriding ```selected_period``` in ```OptionalMonthYearFilter```

emord (approved):

nickpell:
self-merging

PR #13431, opened 2016-10-03
fixing gir for unicode domains and countries

calellowitz:
@sravfeyn cc: @gcapalbo 
The default csv writer does not support unicode output and was breaking on a domain with country listed as `Côte d'Ivoire`. I used our existing util (which i believe is taken from the python docs) to fix

### DIFF
--- a/corehq/apps/hqadmin/views.py
+++ b/corehq/apps/hqadmin/views.py
@@ -3,13 +3,13 @@ from lxml import etree
 import HTMLParser
 import json
 import socket
-import csv
 from datetime import timedelta, date, datetime
 from collections import defaultdict, namedtuple, OrderedDict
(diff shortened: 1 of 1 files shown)


millerdev (approved):

corehq/apps/hqadmin/views.py
```diff
@@ -1001,7 +1001,7 @@ def get(self, request, *args, **kwargs):
             except (ValueError, ValidationError):
                 messages.error(
                     request,
-                    _("Enter a valid year-month. e.g. 2015-09 (for December 2015)")
+                    _("Enter a valid year-month. e.g. 2015-09 (for September 2015)")
```
gcapalbo: ha, nice catch

PR #13433, opened 2016-10-03
Assert rev in pillow processor

benrudolph:
@czue @snopoke this'll publish the doc rev to the changes feed and then assert that the revs are the same. curious to see if any of these pop up in the coming days. i think we only need it for the ElasticProcessor but i may be overlooking another place where it might be useful

http://manage.dimagi.com/default.asp?237983
http://manage.dimagi.com/default.asp?237943

cc: @proteusvacuum

### DIFF
--- a/corehq/apps/change_feed/tests/test_pillow.py
+++ b/corehq/apps/change_feed/tests/test_pillow.py
@@ -127,4 +127,4 @@ class TestElasticProcessorPillows(SimpleTestCase):
                 )
             )
         except DocumentMismatchError:
-            self.fail('Incorectly raise a document not found error for matching revs')
+            self.fail('Incorectly raise a DocumentMismatchError for matching revs')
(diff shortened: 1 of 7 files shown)


corehq/apps/change_feed/tests/test_pillow.py
```diff
@@ -1,4 +1,5 @@
 # coding=utf-8
+from mock import patch, MagicMock
```
dimagimon: F401 'MagicMock' imported but unused

czue:
looks good and no blockers on my side. wondering if we should apply this more broadly? I can't think of any instance where you wouldn't want to fail hard on out-of-date revisions and it seems like this would cause problems in UCRs and custom reports and probably most other things we use reports for...

benrudolph:
yep that sounds right to me

benrudolph:
@czue cleaned up and added some more checks

czue (approved):

PR #13434, opened 2016-10-03
Change default background color

benrudolph:
@biyeun this changes the color of the default plan type so that it is not white on white
http://manage.dimagi.com/default.asp?238625
screen shot 2016-10-03 at 5 10 54 pm
cc interrupt: @nickpell @mkangia

### DIFF
--- a/corehq/apps/style/static/style/less/_hq/label.less
+++ b/corehq/apps/style/static/style/less/_hq/label.less
@@ -13,6 +13,7 @@
   .border-bottom-radius(5px);
   padding: 10px;
   color: #ffffff;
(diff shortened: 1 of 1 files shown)


benrudolph:
killed build

benrudolph:
this gives a default to the plan which will prevent it from ever being white on white, but maybe we should come up with a color for the reseller plan? that was the original failure. there is no color for the reseller page on the home page:

https://www.commcarehq.org/pricing/

biyeun:
merging as this doesn't touch tests

nickpell:
> but maybe we should come up with a color for the reseller plan?

are we ok not doing this? @biyeun

PR #13436, opened 2016-10-03
fixing api to allow non superuser access

calellowitz:
@gcapalbo / anyone
This fixes a bug that in the domain list api, to allow people without superuser accounts to authenticate


gcapalbo:
look good. now that this is open to other users though i think it makes sense to move it out of hqadmin. maybe /accounts is a better place? do you know how often this api is being used right now and if we can move it?

calellowitz:
This was written mainly for what Clayton has termed the hub install (installing over a local network from one phone to others), so now is not being used at all. I do not think it is expected to be used publicly, but just in that context.

gcapalbo:
ok, we should probably move it out of hqadmin then, what do you think?

gcapalbo:
(not blocking, will merge and will let you move it in another pr if you agree)

PR #13438, opened 2016-10-03
Add toggle to blacklist submissions from a domain

esoergel:
@dannyroberts


dannyroberts:
The only thing to keep an eye on here is how this effects submission throughput. My guess is that a toggle lookup won't matter too much (and I think they're cached).

esoergel:
It's been a week, we could probably check now.  Where would you look to monitor submission timings?

PR #13444, opened 2016-10-04
ledger doc store fix for archived products / locations

snopoke:
Use unfiltered Django manager in doc store to prevent doc not found errors.
@czue 

Outstanding question: Should stock state for archived products and locations be removed from ES?

### DIFF
--- a/corehq/form_processor/document_stores.py
+++ b/corehq/form_processor/document_stores.py
@@ -106,4 +106,4 @@ class LedgerV1DocumentStore(DjangoDocumentStore):
         def _doc_gen_fn(obj):
             return obj.to_json()
 
-        super(LedgerV1DocumentStore, self).__init__(StockState, _doc_gen_fn)
+        super(LedgerV1DocumentStore, self).__init__(StockState, _doc_gen_fn, model_manager=StockState.include_archived)
--- a/corehq/form_processor/tests/test_ledgers.py
+++ b/corehq/form_processor/tests/test_ledgers.py
@@ -1,3 +1,4 @@
+import uuid
 from collections import namedtuple
 
 from django.conf import settings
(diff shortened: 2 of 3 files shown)


corehq/form_processor/document_stores.py
```diff
@@ -106,4 +106,4 @@ def __init__(self, domain):
         def _doc_gen_fn(obj):
             return obj.to_json()
 
-        super(LedgerV1DocumentStore, self).__init__(StockState, _doc_gen_fn)
+        super(LedgerV1DocumentStore, self).__init__(StockState, _doc_gen_fn, model_manager=StockState.include_archived)
```
dimagimon: E501 line too long (119 > 115 characters)

corehq/form_processor/tests/test_ledgers.py
```diff
@@ -211,3 +213,67 @@ def _assert_transactions(self, values, ignore_ordering=False):
...
+        self.addCleanup(toggle_product_archive)
+
+        self.assertTrue(SQLProduct.objects.get(code=self.product_a.code).is_archived)
+
+        state = stock_states[0]
+        store = LedgerV1DocumentStore(DOMAIN)
+        doc = store.get_document(state.id)
+        self.assertEquals(int(doc['_id']), state.id)
+        self.assertEquals(doc['case_id'], state.case_id)
+
```
dimagimon: W391 blank line at end of file

czue (approved):

czue:
> Should stock state for archived products and locations be removed from ES?

I don't have an obvious answer but my guess is maybe not removed but tagged as archived? I think maybe we can just do nothing until it comes up in a reports context though? @esoergel might also have an opinion

esoergel:
Tagged as archived (and filtered out in the default `LedgerES`) seems like a safe bet.

I also can't think of a good reason to keep them in elasticsearch, since presumably the only reason we'd query against them would be to restore them, which should be easy enough also in SQL.

PR #13446, opened 2016-10-04
Added revised Case Finding report and Treatment Outcome report

kkrampa:
@czue 

treatment outcome

case finding


custom/enikshay/reports/case_finding.py
```diff
@@ -0,0 +1,131 @@
...
+        return DataTablesHeader(
+            DataTablesColumn('Age'),
+            DataTablesColumn('0-4'),
+            DataTablesColumn('5-14'),
+            DataTablesColumn('15-24'),
+            DataTablesColumn('25-34'),
+            DataTablesColumn('35-44'),
+            DataTablesColumn('45-54'),
+            DataTablesColumn('55-64'),
+            DataTablesColumn(u'≥65'),
```
czue: nit: it might be nice if these columns were automatically generated from `AGE_RANGES`

custom/enikshay/__init__.py
```diff
@@ -1 +1,10 @@
 from custom.enikshay.integrations.ninetyninedots.repeater_generators import RegisterPatientPayloadGenerator
+from custom.enikshay.reports.case_finding import CaseFindingReport
+from custom.enikshay.reports.treatment_outcome import TreatmentOutcomeReport
+
+CUSTOM_REPORTS = (
```
czue: I think we are trying to move these into `reports.py` moving forwards to avoid too many imports in the `__init__` modules

custom/enikshay/reports/filters.py
```diff
@@ -0,0 +1,34 @@
+from django.core.urlresolvers import reverse
+
+from corehq.apps.locations.models import SQLLocation
+from corehq.apps.reports.filters.base import BaseMultipleOptionFilter
+
+
+class EnikshayLocationFilter(BaseMultipleOptionFilter):
```
czue: is there anything that makes this a custom filter as opposed to one that could be dropped into any report for locations?

custom/enikshay/reports/views.py
```diff
@@ -0,0 +1,45 @@
+from django.http.response import JsonResponse
+from django.utils.decorators import method_decorator
+from django.views.generic.base import View
+
+from corehq.apps.domain.decorators import login_and_domain_required
+from corehq.apps.locations.models import SQLLocation
+from corehq.apps.userreports.reports.filters.choice_providers import ChoiceQueryContext
+
+
+class LocationsView(View):
```
czue: do we not have anything like this already? i thought we had some other location APIs we might be able to use / modify. cc @esoergel

custom/enikshay/templates/enikshay/partials/report_content.html
```diff
@@ -0,0 +1,98 @@
...
+        {% else %}
+            {% trans "Please apply filters to view:" %} {{ report.report_title }}
+        {% endif %}
+        </h2>
+    </div>
+    <div class="panel-body-datatable">
+    {% block reporttable %}
+    {% if report.needs_filters %}
+        {% include 'reports/partials/description.html' %}
+    {% else %}
```
czue: is there a reason why none of this could be pulled out/shared with [tabular.html](https://github.com/dimagi/commcare-hq/blob/master/corehq/apps/reports/templates/reports/async/tabular.html#L73) ?

czue (commented):
thanks for putting this together!

no blockers but a couple comments i'll let you respond on before merge

custom/enikshay/reports/views.py
kkrampa: We have a similar view for UCR filters but it takes as argument ```report_id``` so I copied and modified it a little bit.

custom/enikshay/reports/filters.py
czue: sorry, probably an obvious question but why can't you use the UCR filter? do you mean the `LocationChoiceProvider`?

czue:
@proteusvacuum can you do a pass on this? just wondering if there are any issues/conflicts with the location filtering stuff you're working on.

custom/enikshay/reports/filters.py
proteusvacuum: Yeah, would be good to make this inherit from `LocationChoiceProvider` so that it also gets the location restriction features I'm currently working on (which are required for these reports)

custom/enikshay/templates/enikshay/partials/report_content.html
kkrampa: I changed it in 2acefe5 and a8be8b1

custom/enikshay/reports/views.py
proteusvacuum: Can you inherit from the view that we are currently using? I'm currently making changes to it so that it works with location restrictions. This view will get blocked by the location middleware otherwise.

proteusvacuum:
@kkrampa I'm currently working on restricting the locations in UCR reports so that people only see data based on their assigned locations. I should have something pushed later today and will ping you on it so you can see what I've done. Both the filter and the locations view that you've added here will need to be changed slightly (or inherit from the ones we currently use) so that this functionality will work.

custom/enikshay/reports/views.py
kkrampa: In UCRs this view https://github.com/dimagi/commcare-hq/blob/master/corehq/apps/userreports/views.py#L1220 is used and it's not Class Based View.

kkrampa:
@proteusvacuum So then, I will wait for your changes.

czue:
@kkrampa see https://github.com/dimagi/commcare-hq/pull/13476

kkrampa:
@proteusvacuum @czue 
It seems I only need to pass user here https://github.com/dimagi/commcare-hq/pull/13446/files#diff-8502d8a84c3efa0a2c8abd6efe412327R21 to make this work with @proteusvacuum changes

proteusvacuum:
@kkrampa, you'll also need to update https://github.com/dimagi/commcare-hq/pull/13446/files#diff-49ee5c39504653740388978724eec38dR20 to only include locations that are accessible to the user, I think. I think it might be better if you used the UCR `LocationChoiceProvider` filter in that case... I'm not sure I understand why it can't be used in this instance. 
You should also decorate all the new views you have with the `@location_safe` decorator.

custom/enikshay/reports/views.py
esoergel: @czue the closest thing I can think of is the `EmwfOptionsView`, but that also handles users and groups, so it probably doesn't make sense.  There's the tastypie location resource, but that doesn't currently accept a query parameter.

kkrampa:
@proteusvacuum I will do it.
What's the status of #13476? Are you going to merge this soon?

proteusvacuum:
@kkrampa thanks. That PR should be merged sometime early next week once @esoergel signs off on it.

custom/enikshay/reports/sqldata/case_finding_sql_data.py
```diff
@@ -0,0 +1,125 @@
...
+            generate_for_all_ranges('male', self.filters + [RawFilter("sex = 'm'")]) +
+            generate_for_all_ranges('female', self.filters + [RawFilter("sex = 'f'")]) +
+            generate_for_all_ranges('transgender', self.filters + [RawFilter("sex = 't'")]) +
+            generate_for_all_ranges('all', self.filters) +
+            [
+                DatabaseColumn(
+                    '',
+                    CountColumn(
+                        'doc_id',
+                        filters=self.filters + test_type_filter + [RawFilter("episode_type = 'Suspect'")],
```
proteusvacuum: should this be `suspect`?

settings.py
```diff
@@ -389,7 +389,8 @@
 
     'custom.icds_reports',
     'custom.enikshay.integrations.ninetyninedots',
-    'custom.pnlppgi'
+    'custom.pnlppgi',
+    'custom.enikshay'
```
proteusvacuum: does this mean we can remove the `custom.enikshay.integrations.ninetyninedots` line above?

custom/enikshay/reports/views.py
```diff
@@ -0,0 +1,37 @@
...
+from corehq.apps.domain.decorators import login_and_domain_required
+from corehq.apps.locations.permissions import location_safe
+from corehq.apps.userreports.reports.filters.choice_providers import ChoiceQueryContext, LocationChoiceProvider
+from custom.enikshay.reports.utils import Report
+
+
+class LocationsView(View):
+
+    @method_decorator(login_and_domain_required)
+    @method_decorator(location_safe)
```
proteusvacuum: @esoergel do you know if this will work with your new method decorator? Or should we wait for your PR to be merged?

proteusvacuum (changes requested):

proteusvacuum:
Thanks for the updates @kkrampa. I had a few more minor comments, but the crux of this looks great.

custom/enikshay/reports/views.py
esoergel: Looks like you just merged the other PR, so @kkrampa can pull in master and use that version.  I haven't tested this wrapped with `@method_decorator`, but it should work out of the box as a method decorator on a dispatch method or as a class decorator on a django view.

My preference would be:
```python
@location_safe
class LocationsView(View):
    ...
```
As is now done for the other class-based views.

custom/enikshay/reports/generic.py
```diff
@@ -0,0 +1,92 @@
...
+from corehq.apps.reports.sqlreport import SqlTabularReport, SqlData
+from corehq.apps.reports.standard import CustomProjectReport, DatespanMixin
+from corehq.apps.reports.util import get_INFilter_bindparams
+from corehq.apps.userreports.util import get_table_name
+from custom.enikshay.reports.filters import EnikshayLocationFilter
+from custom.utils.utils import clean_IN_filter_value
+
+TABLE_ID = 'episode'
+
+ReportConfig = namedtuple('ReportConfig', ['domain', 'locations_id', 'start_date', 'end_date'])
```
kkrampa: It wasn't supposed to be a stub but nonetheless, I've changed a name of this variable.

proteusvacuum (approved):

proteusvacuum:
:+1: on :white_check_mark:

kkrampa:
@proteusvacuum :white_check_mark:

PR #13448, opened 2016-10-04
Fix legacy multimedia export

benrudolph:
@czue @NoahCarnahan this'll properly match the schema's index to the question id (instead of the export's display value which can change)

cc: @proteusvacuum

### DIFF
--- a/corehq/apps/reports/tasks.py
+++ b/corehq/apps/reports/tasks.py
@@ -392,6 +392,18 @@ def _write_attachments_to_file(fpath, use_transfer, num_forms, forms_info, case_
                     DownloadBase.set_progress(build_form_multimedia_zip, form_number + 1, num_forms)
 
 
+def _convert_legacy_indices_to_export_properties(indices):
+    # Strip the prefixed 'form' and change '.'s to '-'s
+    return set(map(
+        lambda index: '-'.join(index.split('.')[1:]),
+        # Filter out any columns that are not form questions
+        filter(
+            lambda index: index and index.startswith('form'),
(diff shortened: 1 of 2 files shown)


corehq/apps/reports/tasks.py
```diff
@@ -402,9 +414,9 @@ def _get_export_properties(export_id, export_is_legacy):
         if export_is_legacy:
             schema = FormExportSchema.get(export_id)
             for table in schema.tables:
-                # - in question id is replaced by . in excel exports
-                properties |= {c.display.replace('.', '-') for c in
-                               table.columns if c.display}
+                properties |= _convert_legacy_indices_to_export_properties(
+                    map(lambda column: column.index, table.columns)
```
benrudolph: main change here is that we're using `column.index` instead of `column.display`

dannyroberts:
Not sure if related:

```======================================================================
FAIL: test_permissions_for_activation (corehq.apps.domain.tests.test_domain_transfer.TestTransferDomainViews)
  File "/mnt/commcare-hq-ro/corehq/apps/domain/tests/test_domain_transfer.py", line 167, in test_permissions_for_activation
    self.assertEqual(resp.status_code, 200)
AssertionError: 404 != 200
auditcare.middleware: DEBUG: Middleware is running in a test context, disabling monkeypatch
```

benrudolph:
fortunately (or unfortunately) i don't think it is. been breaking on other tests

NoahCarnahan (approved):
👍 on ✅

PR #13452, opened 2016-10-04
upgrade django-mptt to 0.8.6

nickpell:
Needed for django >1.8

http://django-mptt.github.io/django-mptt/upgrade.html

### DIFF
--- a/corehq/apps/locations/models.py
+++ b/corehq/apps/locations/models.py
@@ -364,7 +364,7 @@ class SQLLocation(SyncSQLToCouchMixin, MPTTModel):
 
     supply_point_id = models.CharField(max_length=255, db_index=True, unique=True, null=True)
 
-    objects = LocationManager()
+    objects = _tree_manager = LocationManager()
     # This should really be the default location manager
     active_objects = OnlyUnarchivedLocationManager()
 
--- a/requirements/requirements.txt
+++ b/requirements/requirements.txt
@@ -67,7 +67,7 @@ stripe==1.12.2
 django-compressor==1.6
 django-angular==0.7.16
 Pycco==0.3.0
-django-mptt==0.7.4
+django-mptt==0.8.6
 jsonpath-rw==1.4.0
 pip>=6.1.1
 git+git://github.com/smartfile/django-transfer.git@6e0dc94c3341c358fca8eb2bf74e23aee3983ec4


nickpell:
Test failure reproducible locally:
```
FAIL: test_archive (corehq.apps.commtrack.tests.test_locations.LocationsTest)
  File "/mnt/commcare-hq-ro/corehq/util/test_utils.py", line 222, in inner
  File "/mnt/commcare-hq-ro/corehq/util/test_utils.py", line 195, in __call__
  File "/mnt/commcare-hq-ro/corehq/util/test_utils.py", line 210, in call_with_settings
    fn(*args, **kwargs)
  File "/mnt/commcare-hq-ro/corehq/util/test_utils.py", line 191, in fn_with_pre_and_post
    self.fn(*args, **kwargs)
  File "/mnt/commcare-hq-ro/corehq/apps/commtrack/tests/test_locations.py", line 73, in test_archive
    original_count
AssertionError: 1 != 2
```

nickpell:
@esoergel Do you have ideas why this test might be failing?  After the child location is archived ```get_descendants``` of the parent returns an empty list when it's supposed to return a list containing the archived child.

nickpell:
Realized Ethan is offline ... @czue ?

nickpell:
Here's a lead:

on currrent:
```
>>> SQLLocation.objects
<corehq.apps.locations.models.LocationManager at 0x106222210>
>>> SQLLocation._tree_manager
<corehq.apps.locations.models.LocationManager at 0x10c6d3110>
```

On 0.8.6:
```
>>> SQLLocation.objects
<corehq.apps.locations.models.LocationManager at 0x106222210>
>>> SQLLocation._tree_manager
<corehq.apps.locations.models.OnlyUnarchivedLocationManager at 0x106222250>
```

nickpell:
Looks like the issue was introduced when the dependency changed: https://github.com/django-mptt/django-mptt/pull/404

nickpell:
Confirmed on staging that existing locations are unaffected, and that new locations can be added.

@czue Any other recommendations for QA, or do you think this is good to merge?

czue:
@nickpell i think we have decent test coverage. if you have verified most common locations workflows on staging (including reports and bulk updates) then I think this is probably ok to go.

esoergel:
Nick, I'd guess it has to do with the way mptt dynamically selects which model manager to use:
https://github.com/django-mptt/django-mptt/blob/c1994f41ba287e4a083e4e70035732397dc315a8/mptt/models.py#L357

It looks like all default mptt methods (like `get_descendants`) use `self._tree_manager` to perform the queries.  We have two querysets (one excludes archived locations).  It looks like mptt essentially selects one non-deterministically.  I think the best resolution would be to explicitly set `self._tree_manager` ourselves so we know which one we're using.  This could be done by overriding `SQLLocation.register`, calling `super`, and changing `_tree_manager` on the result.

I've also been planning to eventually change the default model manager (`objects`) to `OnlyUnarchivedLocationManager`, but that's outside of the scope of what you're doing here.  I'll table that, so I *think* `_tree_manager = LocationManager` would give feature parity with before the library upgrade.

esoergel:
@sravfeyn has also wrestled with `django-mptt` a bit and may be interested in this topic.

PR #13454, opened 2016-10-04
Restrict edits for read only reports

mkangia:
http://manage.dimagi.com/default.asp?237648

Once done we need to remove the copies of Static Reports created. Aware of icds-cas for now.

### DIFF
--- a/corehq/apps/userreports/ui/forms.py
+++ b/corehq/apps/userreports/ui/forms.py
@@ -85,14 +85,18 @@ class ConfigurableReportEditForm(DocumentFormBase):
                 'configured_charts',
                 'sort_expression',
             ),
-            hqcrispy.FormActions(
-                twbscrispy.StrictButton(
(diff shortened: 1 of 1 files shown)


mkangia:
@sheelio @czue @NoahCarnahan Review please.

czue:
> Once done we need to remove the copies of Static Reports created.

are you going to follow up on this? should we make a mgmt command or something?

mkangia:
@czue i will follow up on this. Not making a management command. Need to do it carefully so would delete them one at a time.

czue:
> Need to do it carefully 

I actually would have thought this might be an argument for a scripted approach. can we not just delete anything that has a conflicting ID with a static report?

mkangia:
We can. Still want to do it one at a time. Making sure at every step. Also may be trying to avoid writing a management command.

PR #13456, opened 2016-10-04
Revert "Revert "require name when copying app""

benrudolph:
Reverts dimagi/commcare-hq#13455

@orangejenny i think this legitimately broke the tests


orangejenny:
closing in favor of https://github.com/dimagi/commcare-hq/pull/13457

PR #13458, opened 2016-10-05
update django digest

nickpell:


nickpell:
self-merging submodule update

PR #13473, opened 2016-10-06
fetch snapshot by id for importing [skip ci]

mkangia:
@esoergel 
http://manage.dimagi.com/default.asp?238900

on the same lines as https://github.com/dimagi/commcare-hq/pull/13391/files


czue:
offlined with manish about [ci skip] but looks good

PR #13477, opened 2016-10-06
Br/mm export filtering

benrudolph:
@czue @NoahCarnahan this has mm downloads respect filters. also removes couch call to get form ids for old exports. noah any reason we didn't do this before? i can't remember if that was deliberate or not

cc: @orangejenny

### DIFF
--- a/corehq/apps/reports/analytics/esaccessors.py
+++ b/corehq/apps/reports/analytics/esaccessors.py
@@ -585,6 +585,4 @@ def scroll_case_names(domain, case_ids):
 def _get_attachment_dicts_from_form(form):
     if 'external_blobs' in form:
         return form['external_blobs'].values()
-    elif '_attachments' in form:
-        return form['_attachments'].values()
     return []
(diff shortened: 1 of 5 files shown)


corehq/apps/reports/analytics/esaccessors.py
```diff
@@ -541,14 +541,13 @@ def get_aggregated_ledger_values(domain, case_ids, section_id, entry_ids=None):
...
 
 def get_form_ids_having_multimedia(domain, app_id, xmlns, startdate, enddate):
-    # TODO: Remove references to _attachments once all forms have been migrated to Riak
     query = (FormES()
              .domain(domain)
              .app(app_id)
              .xmlns(xmlns)
              .submitted(gte=startdate, lte=enddate)
              .remove_default_filter("has_user")
-             .source(['_attachments', '_id', 'external_blobs']))
```
benrudolph: @millerdev confirming that all envs have been migrated, so this is obsolete

benrudolph:
probably best reviewed all at once

NoahCarnahan:
Looks good, thanks. I don't think it was intentional.

PR #13478, opened 2016-10-06
Export from riak to zip

emord:
This uses the existing migration framework to export a domain's attachments (except sql forms for now).

It uses a management command for importing from zip to an existing riak, because there will be a lot of buildup for it to be the input to the migration framework for something that will likely only be used once (progress loader, document provider, document processor...)

@millerdev @snopoke @czue

### DIFF
--- a/corehq/blobs/__init__.py
+++ b/corehq/blobs/__init__.py
@@ -19,11 +19,6 @@ def get_blob_db():
     return _db[-1]
 
 
-def get_blob_db_exporter(slug, domain):
-    from .migratingdb import MigratingBlobDB
-    return MigratingBlobDB(_get_zip_db(slug, domain), get_blob_db())
-
-
 def _get_s3_db(settings):
     from .s3db import S3BlobDB
     config = getattr(settings, "S3_BLOB_DB_SETTINGS", None)
@@ -37,11 +32,6 @@ def _get_fs_db(settings):
         reason = settings.SHARED_DRIVE_CONF.get_unset_reason("blob_dir")
         raise Error("cannot initialize blob db: %s" % reason)
     return FilesystemBlobDB(blob_dir)
-
-
-def _get_zip_db(slug, domain):
-    from .zipdb import ZipBlobDB
-    return ZipBlobDB(slug, domain)
 
 
 class BlobInfo(namedtuple("BlobInfo", ["identifier", "length", "digest"])):
--- a/corehq/blobs/__init__.py
+++ b/corehq/blobs/__init__.py
@@ -6,7 +6,7 @@ DEFAULT_BUCKET = "_default"
 _db = []  # singleton/global, stack for tests to push temporary dbs
 
 
-def get_blob_db(export=False, domain=None):
+def get_blob_db():
     if not _db:
         from django.conf import settings
         db = _get_s3_db(settings)
@@ -15,11 +15,13 @@ def get_blob_db(export=False, domain=None):
         elif getattr(settings, "BLOB_DB_MIGRATING_FROM_FS_TO_S3", False):
             from .migratingdb import MigratingBlobDB
             db = MigratingBlobDB(db, _get_fs_db(settings))
-        elif export:
-            from .migratingdb import MigratingBlobDB
-            db = MigratingBlobDB(_get_zip_db(domain), db)
         _db.append(db)
     return _db[-1]
+
+
+def get_blob_db_exporter(slug, domain):
+    from .migratingdb import MigratingBlobDB
+    return MigratingBlobDB(_get_zip_db(slug, domain), get_blob_db())
 
 
 def _get_s3_db(settings):
@@ -37,9 +39,9 @@ def _get_fs_db(settings):
     return FilesystemBlobDB(blob_dir)
 
 
-def _get_zip_db(domain):
+def _get_zip_db(slug, domain):
     from .zipdb import ZipBlobDB
-    return ZipBlobDB(domain)
+    return ZipBlobDB(slug, domain)
 
 
 class BlobInfo(namedtuple("BlobInfo", ["identifier", "length", "digest"])):
--- a/corehq/blobs/__init__.py
+++ b/corehq/blobs/__init__.py
@@ -6,7 +6,7 @@ DEFAULT_BUCKET = "_default"
 _db = []  # singleton/global, stack for tests to push temporary dbs
 
 
-def get_blob_db():
+def get_blob_db(export=False, domain=None):
     if not _db:
         from django.conf import settings
         db = _get_s3_db(settings)
@@ -15,6 +15,9 @@ def get_blob_db():
         elif getattr(settings, "BLOB_DB_MIGRATING_FROM_FS_TO_S3", False):
             from .migratingdb import MigratingBlobDB
             db = MigratingBlobDB(db, _get_fs_db(settings))
+        elif export:
+            from .migratingdb import MigratingBlobDB
+            db = MigratingBlobDB(_get_zip_db(domain), db)
         _db.append(db)
     return _db[-1]
 
@@ -32,6 +35,11 @@ def _get_fs_db(settings):
         reason = settings.SHARED_DRIVE_CONF.get_unset_reason("blob_dir")
         raise Error("cannot initialize blob db: %s" % reason)
     return FilesystemBlobDB(blob_dir)
+
+
+def _get_zip_db(domain):
+    from .zipdb import ZipBlobDB
+    return ZipBlobDB(domain)
 
 
 class BlobInfo(namedtuple("BlobInfo", ["identifier", "length", "digest"])):
--- a/corehq/blobs/management/commands/import_blob_zip.py
+++ b/corehq/blobs/management/commands/import_blob_zip.py
@@ -23,4 +23,4 @@ class Command(BaseCommand):
             bucket = '/'.join(filename.split('/')[:-1])
             identifier = filename.split('/')[-1]
             blob = cStringIO.StringIO(from_zip.read(filename))
-            to_db.put(blob, identifier, bucket)
+            to_db.copy_blob(blob, identifier, bucket)
--- a/corehq/blobs/management/commands/import_blob_zip.py
+++ b/corehq/blobs/management/commands/import_blob_zip.py
@@ -3,7 +3,7 @@ import zipfile
 
 from django.core.management.base import BaseCommand
 
-from corehq.blobs import get_blob_db
+from corehq.blobs import BlobInfo, get_blob_db
 
 USAGE = "Usage: ./manage.py import_blob_zip <zipname>"
 
@@ -23,4 +23,6 @@ class Command(BaseCommand):
             bucket = '/'.join(filename.split('/')[:-1])
             identifier = filename.split('/')[-1]
             blob = cStringIO.StringIO(from_zip.read(filename))
-            to_db.copy_blob(blob, identifier, bucket)
+            # copy_blob only needs the identifier
+            blob_info = BlobInfo(identifier=identifier, length="", digest="")
+            to_db.copy_blob(blob, blob_info, bucket)
--- a/corehq/blobs/management/commands/import_blob_zip.py
+++ b/corehq/blobs/management/commands/import_blob_zip.py
@@ -1,0 +1,26 @@
+import cStringIO
+import zipfile
+
+from django.core.management.base import BaseCommand
+
+from corehq.blobs import get_blob_db
+
+USAGE = "Usage: ./manage.py import_blob_zip <zipname>"
+
+
+class Command(BaseCommand):
+    help = USAGE
+
+    def add_arguments(self, parser):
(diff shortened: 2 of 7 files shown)


corehq/blobs/zipdb.py
```diff
@@ -0,0 +1,56 @@
...
+    """Blobs stored in zip file. Used for exporting a domain's blobs
+    """
+
+    def __init__(self, slug, domain):
+        self.zipname = 'export-{domain}-{slug}-blobs.zip'.format(domain=domain, slug=slug)
+
+    def put(self, content, basename="", bucket=DEFAULT_BUCKET):
+        identifier = self.get_identifier(basename)
+        path = self.get_path(identifier, bucket)
+        content = content.read()
```
emord: Need to read the whole file in memory because of limitations with `zipfile`

corehq/blobs/__init__.py
```diff
@@ -19,6 +19,11 @@ def get_blob_db():
     return _db[-1]
 
 
+def get_blob_db_exporter(slug, domain):
+    from .migratingdb import MigratingBlobDB
+    return MigratingBlobDB(_get_zip_db(slug, domain), get_blob_db())
```
millerdev: I'm pretty sure this should not be using `MigratingBlobDB` since the only operation called on this db is `copy_blob`, which should only affect the zip db.

corehq/blobs/migrate.py
```diff
@@ -299,6 +319,34 @@ def migrate(self, filename=None, reset=False, max_retry=2, chunk_size=100):
...
+
+    def migrate(self, filename=None, reset=False, max_retry=2, chunk_size=100):
+        if not self.domain:
+            raise MigrationError("Must specify domain")
+
+        doc_migrator = self.doc_migrator_class(self.slug, self.couchdb, filename, self.domain)
+
+        progress_logger = CouchProcessorProgressLogger(self.doc_types)
+
+        document_provider = CouchDocumentProvider(self.iteration_key, self.doc_types)
```
millerdev: Might be worth writing a new document provider that iterates over documents in the domain rather than all documents of the given types. Could also make a provider that can handle types from multiple couch dbs to allow a single export with all blobs rather than one for each couch db. Might also need a different progress logger implementation to give accurate progress output.

corehq/blobs/zipdb.py
```diff
@@ -0,0 +1,56 @@
...
+
+
+class ZipBlobDB(AbstractBlobDB):
+    """Blobs stored in zip file. Used for exporting a domain's blobs
+    """
+
+    def __init__(self, slug, domain):
+        self.zipname = 'export-{domain}-{slug}-blobs.zip'.format(domain=domain, slug=slug)
+
+    def put(self, content, basename="", bucket=DEFAULT_BUCKET):
```
millerdev: I think this should `raise NotImplementedError`; should only be using `copy_blob`.

corehq/blobs/zipdb.py
```diff
@@ -0,0 +1,56 @@
...
+from corehq.blobs.exceptions import BadName
+from corehq.blobs.interface import AbstractBlobDB, SAFENAME
+
+
+class ZipBlobDB(AbstractBlobDB):
+    """Blobs stored in zip file. Used for exporting a domain's blobs
+    """
+
+    def __init__(self, slug, domain):
+        self.zipname = 'export-{domain}-{slug}-blobs.zip'.format(domain=domain, slug=slug)
```
millerdev: What do you think of opening the zip file here in `__init__` and adding a `close` method (called by `BlobDbBackendExporter.processing_complete`) rather than open/close on each `copy_blob`. Seems like it might be useful to check if the export file already exists at the beginning of the export and refuse to modify it. Could add an `--append` flag to the management command if you want to override that sanity check, but I assume that wouldn't be the default case.

Also, do you think the export and import management commands should document in their help that they use this filename (in the current directory)? Might also be nice to add an option to specify this file path on the command line.

corehq/blobs/zipdb.py
```diff
@@ -0,0 +1,56 @@
...
+        return safejoin(bucket, identifier)
+
+
+def safejoin(root, subpath):
+    if not SAFENAME.match(subpath):
+        raise BadName(u"unsafe path name: %r" % subpath)
+    path = join(root, subpath)
+    if commonprefix([root + sep, path]) != root + sep:
+        raise BadName(u"invalid relative path: %r" % subpath)
+    return path
```
millerdev: This looks the same as the version in fsdb. Why not import `safejoin` from there?

corehq/blobs/migrate.py
```diff
@@ -269,6 +274,21 @@ def should_process(self, doc):
...
+class BlobDbBackendExporter(BlobDbBackendMigrator):
+
+    def __init__(self, slug, couchdb, filename=None, domain=None):
+        self.domain = domain
+        super(BlobDbBackendExporter, self).__init__(slug, couchdb, filename)
+
+    @property
+    @memoized
+    def db(self):
+        return get_blob_db_exporter(self.slug, self.domain)
```
millerdev: Why not set `self.db = get_blob_db_exporter(slug, domain)` in `__init__` above?

corehq/blobs/migrate.py
```diff
@@ -269,6 +274,21 @@ def should_process(self, doc):
         return doc.get("external_blobs")
 
 
+class BlobDbBackendExporter(BlobDbBackendMigrator):
```
millerdev: This class should override `_backup_doc` with a no-op. I think there's no need to write anything to the migration log for exports (especially not a full json dump of the doc).

millerdev (changes requested):
Nice! There are a few things that need some attention yet, but this is looking good.

What's your plan for new attachments that are written to the db between when the migration starts and when the restore happens?

millerdev:
I see there are no tests. Do you think this code is code will be used infrequently enough that it's not worth checking that it actually reads and writes the correct content?

corehq/blobs/management/commands/import_blob_zip.py
```diff
@@ -0,0 +1,26 @@
...
+
+    def handle(self, zipname, **options):
+        from_zip = zipfile.ZipFile(zipname)
+
+        to_db = get_blob_db()
+
+        for filename in from_zip.namelist():
+            bucket = '/'.join(filename.split('/')[:-1])
+            identifier = filename.split('/')[-1]
+            blob = cStringIO.StringIO(from_zip.read(filename))
```
emord: Yeah. You can't use `seek` in a `ZipFileExt`. I also looked at using tarfiles, but there are similar issues there

corehq/blobs/migrate.py
```diff
@@ -92,6 +92,7 @@
 import couchforms.models as xform
 from corehq.apps.app_manager.models import Application, RemoteApp
 from couchexport.models import SavedBasicExport
+from dimagi.utils.decorators.memoized import memoized
```
dimagimon: F401 'memoized' imported but unused

corehq/blobs/__init__.py
```diff
@@ -19,6 +19,11 @@ def get_blob_db():
     return _db[-1]
 
 
+def get_blob_db_exporter(slug, domain):
```
emord: https://github.com/dimagi/commcare-hq/pull/13478/commits/25637ae268f334eeb2d7aafafddec9ae9f89a186

corehq/blobs/zipdb.py
emord: https://github.com/dimagi/commcare-hq/pull/13478/commits/25637ae268f334eeb2d7aafafddec9ae9f89a186

corehq/blobs/management/commands/import_blob_zip.py
```diff
@@ -0,0 +1,26 @@
...
+            to_db.put(blob, identifier, bucket)
```
emord: https://github.com/dimagi/commcare-hq/pull/13478/commits/f5f6d4abad7c58eb92d6de11b666f9e3e94c1a9e

corehq/blobs/migrate.py
emord: https://github.com/dimagi/commcare-hq/pull/13478/commits/1000633a123822ae4b9c79e20799b6bb45a18a3e

corehq/blobs/__init__.py
emord: https://github.com/dimagi/commcare-hq/pull/13478/commits/55f06adcea7fa97659a32b0dde38e5e52a4abf06 & https://github.com/dimagi/commcare-hq/pull/13478/commits/b8ff4f485d46c608e28a6e71a8678797c78a6fde

I'm fine with either approach, but I'm sure why one would be better than hte other. I originally used `MigratingBlobDB` because you need access to both the zipfile "db" and the current riak instance. It does seem weird because we're not "migrating" though.

corehq/blobs/migrate.py
```diff
@@ -233,13 +234,17 @@ class BlobDbBackendMigrator(BaseDocMigrator):
...
         self.total_blobs = 0
         self.not_found = 0
         if not isinstance(self.db, MigratingBlobDB):
             raise MigrationError(
                 "Expected to find migrating blob db backend (got %r)" % self.db)
 
+    @property
+    @memoized
+    def db(self):
+        return get_blob_db()
```
emord: That was because I was inheriting the exporter from the migrator which would check the db against `MigratingBlobDB`. changed in  55f06ad & b8ff4f4

corehq/blobs/zipdb.py
emord: I did it this way, so that the chances of a corrupt zip file are very small in comparison to having a file open for a very long time. zipfiles don't provide any sort of `flush` command

emord: agreed about documentation and extra flags. can do that more easily in a follow up PR

emord:
> What's your plan for new attachments that are written to the db between when the migration starts and when the restore happens?

I'm not worried about this because for the icds migration we are stopping submissions and the app is frozen. Should revisit if this code is used again

corehq/blobs/migrate.py
emord: Interesting idea. I had not thought about having a `DomainDocProvider`

snopoke:
LGTM assuming there will be a followup PR to address some of the comments. @millerdev waiting for your approval.

corehq/blobs/migrate.py
emord: Ah thanks.I had started looking at implementing that myself. I don't think case attachments are a thing that is deployed yet, so probably won't need the case provider right?

corehq/blobs/migrate.py
```diff
@@ -269,6 +272,101 @@ def should_process(self, doc):
...
+
+    def process_doc(self, doc):
+        from corehq.form_processor.backends.sql.dbaccessors import FormAccessorSQL
+        from corehq.blobs import BlobInfo
+        from_db = get_blob_db()
+        for attachment in FormAccessorSQL.get_attachments(doc['_id']):
+            bucket = attachment.blobdb_bucket()
+            blob_id = attachment.blob_id
+            info = BlobInfo(identifier=blob_id, length=attachment.content_length,
+                            digest="md5="+attachment.md5)
```
dimagimon: E226 missing whitespace around arithmetic operator

corehq/blobs/migrate.py
```diff
@@ -274,16 +272,96 @@ def should_process(self, doc):
...
+        return True
+
+    def processing_complete(self, skipped):
+        super(SqlFormAttachmentExporter, self).processing_complete(skipped)
+        if self.not_found:
+            print("{} of {} blobs were not found in the old blob database. It "
+                  "is possible that some blobs were deleted as part of normal "
+                  "operation during the migration if the migration took a long "
+                  "time. However, it may be cause for concern if a majority of "
+                  "the total number of migrated blobs were not found."
```
millerdev: What do you think of moving this string to a global constant since it's used in a number of places (DRY)?

corehq/blobs/zipdb.py
```diff
@@ -54,3 +44,11 @@ def safejoin(root, subpath):
...
         raise BadName(u"invalid relative path: %r" % subpath)
     return path
+
+
+def get_blob_db_exporter(slug, domain):
+    return _get_zip_db(slug, domain)
+
+
+def _get_zip_db(slug, domain):
+    return ZipBlobDB(slug, domain)
```
millerdev: nit: could combine above two functions into
```py
def get_blob_db_exporter(slug, domain):
    return ZipBlobDB(slug, domain)
```

millerdev (changes requested):
Looking good. I think there's still one broken thing (noted in review comment).

corehq/blobs/migrate.py
snopoke: right

corehq/blobs/zipdb.py
emord: Thanks that's good to know. I was just looking at the zip file api. I'd be hesitant to rely on private implementation

PR #13481, opened 2016-10-06
Kill remote app creation button

orangejenny:
@benrudolph


orangejenny:
Double-checked that no one has this flag on in prod or india (not that it works anyway).

benrudolph:
sweet!
@czue this seems right, right?

czue:
yeah this seems good to me if no one has that flag enabled.

PR #13488, opened 2016-10-07
replace broken_build logic with a soft assert

orangejenny:
http://manage.dimagi.com/default.asp?238388

@benrudolph / @dannyroberts


corehq/apps/app_manager/views/download.py
```diff
@@ -289,21 +290,8 @@ def resolve_path(path):
...
-                        try:
-                            request.app.save()
-                        except ResourceConflict:
-                            # this really isn't a big deal:
-                            # It'll get updated next time a resource is request'd;
-                            # in fact the conflict is almost certainly from
-                            # another thread doing this exact update
-                            pass
+                    _assert = soft_assert('@'.join(['jschweers', 'dimagi.com']))
+                    _assert(False, 'Expected build resource %s not found' % path)
```
benrudolph: shouldn't we keep the logic around and remove when we're confident that the soft assert hasn't been fired?

dannyroberts: @benrudolph yeah given the background this ticket, I agree with @orangejenny. Basically I believe this check is no longer relevant ever since we started making builds more atomically years ago, and we found that it was needlessly flagging builds as broken during a riak outage.

benrudolph: gotcha thanks for the explanation. 👍 on ✔️

orangejenny:
removed related test

orangejenny:
tests passing now

PR #13489, opened 2016-10-07
fix bad merge on datetime

biyeun:
@benrudolph fixes bad merge that was messing up datetimes.

feels like this should be a hotfix?


benrudolph:
don't think prod has been dpeloyed yet

PR #12543, opened 2016-07-20
add an alternate flat locations fixture

czue:
This adds a feature flag that adds a second locations fixture in a flat format (which is often easier to work with in apps). 

The format might change a bit, but want to iterate on this with kriti for UATBC.

``` xml
<fixture id="commtrack:locations_v2" user_id="d7c3c4246b1126d6506522d4feb235df">
    <locations>
      <location id="dae90df587e610adb7baedf1d5038abf" state_id="dae90df587e610adb7baedf1d5039270" type="city">
        <name>Cape Town</name>
        <site_code>cape_town</site_code>
        <external_id/>
        <latitude/>
        <longitude/>
        <location_type>City</location_type>
        <supply_point_id>23b435953e894b8dbfc5ad0e03f0b81d</supply_point_id>
        <location_data/>
      </location>
      <location city_id="dae90df587e610adb7baedf1d5038abf" id="dae90df587e610adb7baedf1d503885a" state_id="dae90df587e610adb7baedf1d5039270" type="neighborhood">
        <name>Vredehoek</name>
        <site_code>vredehoek</site_code>
        <external_id/>
        <latitude>-33.9411000000</latitude>
        <longitude>18.4232000000</longitude>
        <location_type>Neighborhood</location_type>
        <supply_point_id>ad9dbae5bbd24a12b1c352e0b625ef03</supply_point_id>
        <location_data/>
      </location>
      <location id="dae90df587e610adb7baedf1d5039270" type="state">
        <name>Western Cape</name>
        <site_code>western_cape</site_code>
        <external_id/>
        <latitude/>
        <longitude/>
        <location_type>State</location_type>
        <supply_point_id>896373cca62c454f8d9fa41a544e7e6a</supply_point_id>
        <location_data/>
      </location>
    </locations>
  </fixture>
```

@esoergel @sheelio and @proteusvacuum probably interested. can review either way. tried to break up by :fish: but took a couple small detours.

PR #13491, opened 2016-10-07
Add locations fixture to the set of auto-detected instances

orangejenny:
http://manage.dimagi.com/default.asp?238363

@benrudolph

fyi @snopoke / @dannyroberts I can't think of a reason _not_ to auto-detect as many of the standard instances as we can think to add, but let me know if I'm missing something

### DIFF
--- a/corehq/apps/app_manager/suite_xml/post_process/instances.py
+++ b/corehq/apps/app_manager/suite_xml/post_process/instances.py
@@ -94,6 +94,13 @@ class EntryInstances(PostProcessor):
             if instance_name in unknown_instance_ids:
                 instances.add(Instance(id=instance_name, src='jr://fixture/{}'.format(instance_name)))
                 unknown_instance_ids.remove(instance_name)
+        instance_name = 'locations'
+        if instance_name in unknown_instance_ids:
+            if toggles.FLAT_LOCATION_FIXTURE.enabled(self.app.domain):
+                instances.add(Instance(id=instance_name, src='jr://fixture/{}'.format(instance_name)))
+            else:
+                instances.add(Instance(id=instance_name, src='jr://fixture/commtrack:{}'.format(instance_name)))
+            unknown_instance_ids.remove(instance_name)
         return instances, unknown_instance_ids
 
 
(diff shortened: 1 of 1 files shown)


corehq/apps/app_manager/suite_xml/post_process/instances.py
```diff
@@ -116,6 +116,7 @@ def __call__(self, fn):
 INSTANCE_BY_ID = {
     'groups': Instance(id='groups', src='jr://fixture/user-groups'),
     'reports': Instance(id='reports', src='jr://fixture/commcare:reports'),
+    'locations': Instance(id='locations', src='jr://fixture/commtrack:locations'),
```
czue: if we are adding this can we also add support for the [revised location fixture format](https://github.com/dimagi/commcare-hq/pull/12543)?

confusingly that is `jr://fixture/locations` - i don't actually know how the ID side of this dict is used, but we might want to design some kind of plan for that now before getting ourselves into trouble as I think the long term plan is to deprecate `commtrack:locations` and move to the flat format

orangejenny: Updated to respect the feature flag, so `instances('locations')` will reference the flat fixture if it's available, but there's no way to reference both the flat and hierarchical fixture. If the flat is eventually going to replace the hierarchical, I think that'll be fine.

czue (approved):

corehq/apps/app_manager/suite_xml/post_process/instances.py
```diff
@@ -94,6 +94,13 @@ def _add_custom_referenced_instances(self, instances, unknown_instance_ids):
             if instance_name in unknown_instance_ids:
                 instances.add(Instance(id=instance_name, src='jr://fixture/{}'.format(instance_name)))
                 unknown_instance_ids.remove(instance_name)
+        instance_name = 'locations'
+        if instance_name in unknown_instance_ids:
+            if toggles.FLAT_LOCATION_FIXTURE.enabled(self.app.domain):
+                instances.add(Instance(id=instance_name, src='jr://fixture/{}'.format(instance_name)))
+            else:
+                instances.add(Instance(id=instance_name, src='jr://fixture/commtrack:{}'.format(instance_name)))
+            unknown_instance_ids.remove(instance_name)
```
esoergel: neato!

PR #13499, opened 2016-10-07
update smsbackends urls

nickpell:
For compatibility with django 1.10


nickpell:
@gcapalbo cc: @benrudolph

PR #13507, opened 2016-10-09
Ensure explicit community subscription when downloading snapshot to a…

mkangia:
…n existing project

http://manage.dimagi.com/default.asp?236652

### DIFF
--- a/corehq/apps/appstore/views.py
+++ b/corehq/apps/appstore/views.py
@@ -286,6 +286,9 @@ def approve_app(request, snapshot):
 @login_required
 @retry_resource(3)
 def import_app(request, snapshot):
+    """
+    Download to existing project
+    """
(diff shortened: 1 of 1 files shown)


mkangia:
@nickpell i am not very sure about this but seems like we need to ensure subscriptions while copying to existing project as well.

corehq/apps/appstore/views.py
```diff
@@ -286,6 +286,9 @@ def approve_app(request, snapshot):
 @login_required
 @retry_resource(3)
 def import_app(request, snapshot):
```
nickpell: Since this function doesn't create a new project space, I don't think we need to assign a community subscription here.

mkangia:
Changes not needed.

PR #13508, opened 2016-10-09
Mk accounting walkthrough

mkangia:
Just some code improvements attempted when studying code base. Changes not tested yet. Relying to test cases.

### DIFF
--- a/corehq/apps/accounting/async_handlers.py
+++ b/corehq/apps/accounting/async_handlers.py
@@ -43,10 +43,10 @@ class FeatureRateAsyncHandler(BaseRateAsyncHandler):
 
     @property
     def create_response(self):
-        if Feature.objects.filter(name=self.name).count() > 0:
+        if Feature.objects.filter(name=self.name):
             raise AsyncHandlerError("Feature '%s' already exists, and likely already "
                                     "in this Software Plan Version." % self.name)
-        new_feature, _ = Feature.objects.get_or_create(
+        new_feature, _ = Feature.objects.create(
             name=self.name,
             feature_type=self.rate_type,
         )
@@ -66,10 +66,10 @@ class SoftwareProductRateAsyncHandler(BaseRateAsyncHandler):
 
     @property
     def create_response(self):
-        if SoftwareProduct.objects.filter(name=self.name).count() > 0:
+        if SoftwareProduct.objects.filter(name=self.name):
             raise AsyncHandlerError("Product '%s' already exists, and likely already "
                                     "in this Software Plan Version." % self.name)
-        new_product, _ = SoftwareProduct.objects.get_or_create(
+        new_product, _ = SoftwareProduct.objects.create(
             name=self.name,
             product_type=self.rate_type
         )
@@ -161,14 +161,15 @@ class Select2BillingInfoHandler(BaseSelect2AsyncHandler):
         from django_countries.data import COUNTRIES
         countries = sorted(COUNTRIES.items(), key=lambda x: x[1].encode('utf-8'))
         if self.search_string:
-            return filter(lambda x: x[1].lower().startswith(self.search_string.lower()), countries)
+            search_string = self.search_string.capitalize()
+            return filter(lambda x: x[1].startswith(search_string), countries)
         return countries
 
     @property
     def active_accounts_response(self):
         accounts = BillingAccount.objects.filter(is_active=True)
         if self.search_string:
-            accounts = accounts.filter(name__contains=self.search_string)
+            accounts = accounts.filter(name__icontains=self.search_string)
         return [(a.id, a.name) for a in accounts]
 
     @property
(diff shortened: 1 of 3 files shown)


corehq/apps/accounting/async_handlers.py
```diff
@@ -43,10 +43,10 @@ class FeatureRateAsyncHandler(BaseRateAsyncHandler):
 
     @property
     def create_response(self):
-        if Feature.objects.filter(name=self.name).count() > 0:
+        if Feature.objects.filter(name=self.name):
```
czue: fyi I think `.exists()` can be slightly more efficient. see [bool section below here](https://docs.djangoproject.com/en/1.10/ref/models/querysets/#when-querysets-are-evaluated)

corehq/apps/accounting/async_handlers.py
```diff
@@ -161,14 +161,15 @@ def country_response(self):
         from django_countries.data import COUNTRIES
         countries = sorted(COUNTRIES.items(), key=lambda x: x[1].encode('utf-8'))
         if self.search_string:
-            return filter(lambda x: x[1].lower().startswith(self.search_string.lower()), countries)
+            search_string = self.search_string.capitalize()
```
czue: this feels a little more brittle to me than just forcing both to be lowercase, even though it probably runs a tiny bit faster

corehq/apps/accounting/models.py
```diff
@@ -679,12 +679,9 @@ def new_rate(cls, feature_name, feature_type,
         feature, _ = Feature.objects.get_or_create(name=feature_name, feature_type=feature_type)
         rate = FeatureRate(feature=feature)
 
-        if monthly_fee is not None:
-            rate.monthly_fee = monthly_fee
-        if monthly_limit is not None:
-            rate.monthly_limit = monthly_limit
-        if per_excess_fee is not None:
-            rate.per_excess_fee = per_excess_fee
+        rate.monthly_fee = monthly_fee or rate.monthly_fee
```
czue: you noted that if any of these are falsey this might do the wrong thing (not assign when it should). I think this could be more trouble than it's worth

corehq/apps/hqwebapp/async_handler.py
```diff
@@ -14,15 +14,19 @@ class AsyncHandlerMixin(object):
     def handler_slug(self):
         return self.request.POST.get('handler')
 
-    def get_async_handler(self):
-        handler_class = dict([(h.slug, h) for h in self.async_handlers])[self.handler_slug]
-        return handler_class(self.request)
+    def get_async_handlers_mapping(self):
```
czue: if this is only used here I would prefix it with `_` to indicate it is private

corehq/apps/hqwebapp/async_handler.py
```diff
@@ -14,15 +14,19 @@ class AsyncHandlerMixin(object):
...
+        mapping = {}
+        for async_handler in self.async_handlers:
+            mapping[async_handler.slug] = async_handler
+        return mapping
```
czue: this whole function can be replaced with a dict comprehension I think
```python
return {async_handler.slug: async_handler for async_handler in self.async_handlers}
```

corehq/apps/hqwebapp/async_handler.py
```diff
@@ -14,15 +14,19 @@ class AsyncHandlerMixin(object):
...
 
     @property
     @memoized
     def async_response(self):
-        if self.handler_slug in [h.slug for h in self.async_handlers]:
-            return self.get_async_handler().get_response()
+        async_handlers_mapping = self.get_async_handlers_mapping()
+        if self.handler_slug in async_handlers_mapping.keys():
```
czue: `.keys()` is unnecessary

czue (changes requested):
some good stuff in here. I would think you want to run most of this code or convince yourself it is tested before it gets merged

corehq/apps/accounting/async_handlers.py
mkangia: Oh Nice!

corehq/apps/accounting/async_handlers.py
mkangia: All the countries are Capitalised. So thought would be better to just capitalise the search string instead of doing calling lower on both.
How is this brittle?

corehq/apps/accounting/models.py
mkangia: All of these are datetime objects. So assuming these should either be datetime values or None. and won't be [] or "" or False or other python's Falsy.

corehq/apps/hqwebapp/async_handler.py
mkangia: 👍

corehq/apps/hqwebapp/async_handler.py
mkangia: Naice!!!

corehq/apps/accounting/async_handlers.py
czue: You are relying on the fact that "All the countries are Capitalised" which may be true, but if they upgrade and change that, or if some new country doesn't capitalize for some weird reason, etc. then it would break. by coercing both to the same case you are making it more future-proof of any unexpected changes.

corehq/apps/accounting/models.py
czue: Aren't these [decimals and integers](https://github.com/dimagi/commcare-hq/blob/master/corehq/apps/accounting/models.py#L649-L655)?

corehq/apps/hqwebapp/async_handler.py
czue: iterating through keys of a dictionary is pretty standard/pythonic without the `keys()` call

PR #13509, opened 2016-10-10
Add boolean option for case actions.

proteusvacuum:
screenshot from 2016-10-09 20-48-56

Gives the option to use a "raw" boolean value as the condition to do form actions. I was a bit surprised that this didn't already exist, so this might just be my "developer brain" projecting onto Commcare.

This means you can have hidden values that resolve to a boolean (e.g. `1=1`), instead of coercing it to some value (`if(1=1, 'open_case', 'close_case')`), which I think will allow for more DRYness in hidden values. 

I was also considering adding an 'is false' option (which would resolve to `not(value)`), but wasn't sure if that was confusing for non-programmers.

@dimagi/product not sure if you have an opinion on this...

CC: @millerdev

### DIFF
--- a/corehq/apps/app_manager/models.py
+++ b/corehq/apps/app_manager/models.py
@@ -269,7 +269,7 @@ class FormActionCondition(DocumentSchema):
     type = StringProperty(choices=["if", "always", "never"], default="never")
     question = StringProperty()
     answer = StringProperty()
-    operator = StringProperty(choices=['=', 'selected'], default='=')
+    operator = StringProperty(choices=['=', 'selected', 'boolean_true'], default='=')
 
     def is_active(self):
         return self.type in ('if', 'always')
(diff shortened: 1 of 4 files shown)


amsagoff:
Hey Farid,

Nice! Does it detect if it's a boolean value automatically? Or is there
something the user needs to configure? (Trying to understand if this
changes the experience at all for people with string or choice values, if
not this is definitely a straight improvement!)

Thanks,
Amelia

proteusvacuum:
@amsagoff there is no detection of whether it is a boolean or not... This just adds an extra option to that drop down to make the options `is`, `selected` and `is true`. The `is` and `selected` options remain the same.

amsagoff:
Cool, thanks for clarifying. No opinions from this product person. Excited to see what else you have in store for us :)

PR #13510, opened 2016-10-10
Update form linking UI

proteusvacuum:
I was finding this UI really hard to use...

@millerdev 
## Before

screenshot from 2016-10-09 21-54-26
## After

screenshot from 2016-10-09 21-52-37

### DIFF
--- a/corehq/apps/app_manager/templates/app_manager/partials/form_workflow.html
+++ b/corehq/apps/app_manager/templates/app_manager/partials/form_workflow.html
@@ -42,9 +42,18 @@
 </div>
 
 <script type="text/html" id="form-link-template">
-    <div class="form-group">
+    <div class="panel panel-default">
+        <div class="panel-heading clearfix">
+            <a class="pull-right"
+               role="button"
+               href="#"
+               data-bind="click: $parent.onDestroyFormLink.bind($parent)">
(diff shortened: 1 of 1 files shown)


benrudolph:
nice making good cleanups! i've always wondered why the select dropdown for advanced modules looks like it's from windows 95

corehq/apps/app_manager/templates/app_manager/partials/form_workflow.html
```diff
@@ -54,10 +63,14 @@
...
-                <i class="fa fa-arrow-right"></i>
+        </div>
+        <div class="row">
+            <div class="col-sm-12 text-center">
+                <i class="fa fa-arrow-down"></i>
             </div>
-            <div class="col-sm-5">
+        </div>
+        <div class="row">
+            <div class="col-sm-11">
```
benrudolph: how come this isn't 12?

benrudolph:
👍 the only thing that's a bit odd is that the dropdown spacing doesn't match the xpath expression text box. definitely an improvement though so happy to merge

proteusvacuum:
The dropdown has padding on it that made it overflow the panel. I gave up
trying to remove it without adding inline styles.

PR #13511, opened 2016-10-10
Add zapier to standard plan

kkrampa:
@czue @nickpell 
http://manage.dimagi.com/default.asp?233476#BugEvent.1222407


czue (approved):
lg2m but will let @nickpell do a pass just in case. i checked the list of feature flags and they all look like test projects so I think fine not to worry about those.

nickpell:
looks good!

PR #13512, opened 2016-10-10
nested ES query fix

snopoke:
@sravfeyn 
Quoted from ES docs: "Note that any fields referenced inside the query must use the complete path (fully qualified)."

Not sure how this ever worked. Tested the query via ES HEAD and changing to the full path returns the currect results.


benrudolph:
🔥

PR #13513, opened 2016-10-10
simple progress reporting for reindexers

snopoke:
@sravfeyn


benrudolph:
does this already happen? when i run the reindexer i get output like: `Processed 12831206/12830054 of 12831000 documents in 8:14:56.858167 (0:00:00.476684 remaining)`

snopoke:
You only get that for some of the reindexers, others don't have anything.

PR #13517, opened 2016-10-10
Fix location meta

benrudolph:
@czue @NoahCarnahan this does two things.
1. Ensures that if there is a node that has `#text` we query for the `#text` node. For example `form.meta.location.#text` should still return the location node
2. Makes the location system property a proper GPS column so that it's splittable

screen shot 2016-10-10 at 11 12 40 am

http://manage.dimagi.com/default.asp?239006

cc: @orangejenny


NoahCarnahan:
👍

PR #13518, opened 2016-10-10
Only fire mismatch revs when revs are stale

benrudolph:
@snopoke how about something simple like this? it sort of handles revs that aren't in the format `[0-9]*-.*`, but not very well. i don't think we have to worry about that too much though

http://manage.dimagi.com/default.asp?239085

cc: @orangejenny

### DIFF
--- a/corehq/apps/change_feed/tests/test_pillow.py
+++ b/corehq/apps/change_feed/tests/test_pillow.py
@@ -111,6 +111,13 @@ class TestElasticProcessorPillows(SimpleTestCase):
             data_source_type='couch',
             data_source_name='test_commcarehq'
         )
+        newer_metadata = ChangeMeta(
+            document_id='test-id',
+            # Rev is lower than the rev in the fetched document and we should not throw an error
+            document_rev='2-me',
+            data_source_type='couch',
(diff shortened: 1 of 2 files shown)


corehq/apps/change_feed/tests/test_pillow.py
```diff
@@ -122,6 +128,16 @@ def test_mismatched_rev(self):
                 )
             )
 
+        with self.assertRaises(DocumentMismatchError):
```
czue: I might just be overlooking something but shouldn't there be a test that explicitly checks that something smaller does not raise an exception?

czue (approved):

corehq/apps/change_feed/tests/test_pillow.py
benrudolph: there's one that ensures that the same doesn't raise an exception `good_metadata` which checks that the equality check is working. it implies that the `2-me` would also work, but in the future if this ever changes it could be good to have. will add an additional case

snopoke:
:white_check_mark:

benrudolph:
thanks, nice cleanup

PR #13520, opened 2016-10-10
domain notifications

calellowitz:
This allows notifications to go out only to specific domains, using the existing notification framework in hq.
@emord you have the most context and my buddy is offline this week
cc: @amsagoff

### DIFF
--- a/corehq/apps/notifications/tests/test_notifications.py
+++ b/corehq/apps/notifications/tests/test_notifications.py
@@ -15,7 +15,7 @@ class NotificationTest(TestCase):
         self.user = User()
         self.user.username = 'mockmock@mockmock.com'
         self.user.save()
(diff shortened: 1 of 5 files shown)


corehq/apps/notifications/tests/test_notifications.py
```diff
@@ -14,34 +15,35 @@ def setUp(self):
         self.user = User()
         self.user.username = 'mockmock@mockmock.com'
         self.user.save()
+        self.couch_user = WebUser(username=self.user.username, domains=['test-dom'])
```
emord: I think this should be deleted in the `teardown`

calellowitz: that user is never saved to the db, so no need to delete

PR #13521, opened 2016-10-10
Get username instead of user_id in MM download

benrudolph:
@NoahCarnahan http://manage.dimagi.com/default.asp?233241 uses username instead of user_id in MM downloads

cc: @orangejenny

### DIFF
--- a/corehq/apps/reports/tasks.py
+++ b/corehq/apps/reports/tasks.py
@@ -370,9 +370,9 @@ def _expose_download(fpath, use_transfer, zip_name, download_id, num_forms):
 
 
 def _format_filename(form_info, question_id, extension, case_id_to_name):
-    filename = u"{}-user_{}-form_{}{}".format(
+    filename = u"{}-{}-form_{}{}".format(
         unidecode(question_id),
-        form_info['form'].user_id or 'unknown',
+        form_info['form'].get_data('form/meta/username') or form_info['form'].user_id or 'user_unknown',
         form_info['form'].form_id or 'unknown',
         extension
     )
--- a/corehq/apps/reports/tasks.py
+++ b/corehq/apps/reports/tasks.py
@@ -372,7 +372,7 @@ def _expose_download(fpath, use_transfer, zip_name, download_id, num_forms):
 def _format_filename(form_info, question_id, extension, case_id_to_name):
     filename = u"{}-{}-form_{}{}".format(
         unidecode(question_id),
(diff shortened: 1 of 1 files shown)


corehq/apps/reports/tasks.py
```diff
@@ -370,9 +370,9 @@ def _expose_download(fpath, use_transfer, zip_name, download_id, num_forms):
 
 
 def _format_filename(form_info, question_id, extension, case_id_to_name):
-    filename = u"{}-user_{}-form_{}{}".format(
+    filename = u"{}-{}-form_{}{}".format(
```
benrudolph: decided to drop the `user_` prefix (i had added it a week or two ago) because `user_brudolph@` is confusing and weird. it was useful when it was an ID.

corehq/apps/reports/tasks.py
```diff
@@ -370,9 +370,9 @@ def _expose_download(fpath, use_transfer, zip_name, download_id, num_forms):
...
         unidecode(question_id),
-        form_info['form'].user_id or 'unknown',
+        form_info['form'].get_data('form/meta/username') or form_info['form'].user_id or 'user_unknown',
```
snopoke: should change the order here so that we don't fetch the form XML from Riak unless we have to (unless we're already fetching it elsewhere)

snopoke (changes requested):

NoahCarnahan (approved):
lgtm. Thanks!

corehq/apps/reports/tasks.py
NoahCarnahan: I believe this or statement is using the actual value thats returned, not the truthiness of the statement, so order matters (i.e. we want the username if it exists, otherwise the user_id, otherwise "user_unknown")

snopoke: right, makes sense. In that case it may make sense to try and cache the username by user_id's so that you don't have to fetch the form XML for every form (unless you're already having to fetch it for some other reason)

benrudolph: ah yes we do fetch the form elsewhere. i had thought we did aggressive caching of the form.xml since it's pretty much a read only thing, but just looked and it doesn't seem like we don't cache beyond `memoized`. `memoized` in this case should be enough since we don't re-fetch form, but i'll still move it closer to the original call to get_xml

snopoke (approved):

PR #13523, opened 2016-10-10
Do not use TF sqlite backend toggle

benrudolph:
@czue @wpride good suggestion. this'll make it easier to turn it off. we should just be quick to the draw to turn it on for the disabled domains

cc: @orangejenny


czue (approved):
looks good on :white_check_mark: though we should make sure to coordinate updating this toggle for the listed domains immediately upon deploy. are you planning on coordinating that?

benrudolph:
yep i am

PR #13525, opened 2016-10-10
show app builder notifications to all superusers

czue:
I was thinking about this and I think this is one of those rare features that I don't feel bad about rolling out as superuser-only. my rationale is that it doesn't provide a notably different UX, just a layer of additional feedback on top of the existing UX so it doesn't have some of the problems we've had of, e.g. demos looking different on a dimagi account than otherwise. I also think that 95% of the time if some dimagis are app building, then most of the app-building is being done by only dimagis. additionally, making it superuser-only is required anyways right now because there is some tech debt work to make the websockets framework support custom domain-level permissions, so that work would need to be required to happen before we released widely regardless.

i'm keen to get this out in the wild and see if it adds value, and doing it this way explicitly internally makes sense to me as a way to build out the featureset since we are likely the most important users of this feature. curious if anyone has any objections to that?

cc @dimagi/product @orangejenny and @emord buddy

### DIFF
--- a/corehq/apps/app_manager/views/notifications.py
+++ b/corehq/apps/app_manager/views/notifications.py
@@ -19,17 +19,17 @@ def notify_form_changed(domain, couch_user, app_id, unique_form_id):
 
 
 def notify_event(domain, couch_user, app_id, unique_form_id, message):
-    message = {
+
+    message_obj = RedisMessage(json.dumps({
         'domain': domain,
         'user_id': couch_user._id,
         'username': couch_user.username,
         'text': message,
         'timestamp': json_format_datetime(datetime.datetime.utcnow()),
-    }
-    message = RedisMessage(json.dumps(message))
+    }))
     RedisPublisher(
         facility=get_facility_for_form(domain, app_id, unique_form_id), broadcast=True
-    ).publish_message(message)
+    ).publish_message(message_obj)
 
 
 def get_facility_for_form(domain, app_id, unique_form_id):
--- a/corehq/apps/app_manager/views/notifications.py
+++ b/corehq/apps/app_manager/views/notifications.py
@@ -19,7 +19,8 @@ def notify_form_changed(domain, couch_user, app_id, unique_form_id):
 
 
 def notify_event(domain, couch_user, app_id, unique_form_id, message):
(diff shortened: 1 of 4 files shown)


amsagoff:
Makes sense. Do you have a set of things you are trying to test/get
feedback on? Or just generally if this adds value?

Does it make sense to do a round of UAT, but maybe a bit more
value-targeted, to see if we can get any feedback right off the bat?

orangejenny:
Code looks fine, but app manager test is failing - `ERROR: test_basic_app (corehq.apps.app_manager.tests.test_views.TestViews)`

czue:
@amsagoff I think the main area of feedback I'd be interested on top of whether it adds value is other scenarios where this would be useful (e.g. configuring settings, moving stuff around, etc.). I think that is likely to come up if people see the feature and ask about it but imagine that could also be helped through an internal UAT process so happy to do that if you agree.

either way I think I would advocate for releasing it to superusers as-is before UAT just because I can't see much downside.

czue:
@orangejenny fixed build

orangejenny:
Last thought - do you think it'd be worth noting in the message that this is a superuser-only feature, so that it's obvious to FMs *not* to talk this feature up to external people?

corehq/apps/app_manager/views/notifications.py
```diff
@@ -20,18 +19,17 @@ def notify_form_changed(domain, couch_user, app_id, unique_form_id):
...
-            'user_id': couch_user._id,
-            'username': couch_user.username,
-            'text': message,
-            'timestamp': json_format_datetime(datetime.datetime.utcnow()),
-        }
-        message = RedisMessage(json.dumps(message))
-        RedisPublisher(
-            facility=get_facility_for_form(domain, app_id, unique_form_id), broadcast=True
-        ).publish_message(message)
+    message = '{} (<a href="https://confluence.dimagi.com/display/ccinternal/App+Builder+Notifications">what is this?</a>)'.format(message)
```
dimagimon: E501 line too long (139 > 115 characters)

czue:
@orangejenny that's a good idea. updated as follows:
image
which links to [here](https://confluence.dimagi.com/display/ccinternal/App+Builder+Notifications). does that seem reasonable?

orangejenny:
Nice, I like it.

orangejenny (approved):

PR #13526, opened 2016-10-10
Remove unreliable check

benrudolph:
@gcapalbo this does not work as expected. Sometimes flower will randomly say that the old process is up and sometimes down when the old process is running a worker to finish out a task

### DIFF
--- a/corehq/apps/hqadmin/service_checks.py
+++ b/corehq/apps/hqadmin/service_checks.py
@@ -134,17 +134,11 @@ def check_heartbeat():
         t = cresource.get("api/workers", params_dict={'status': True}).body_string()
         all_workers = json.loads(t)
         bad_workers = []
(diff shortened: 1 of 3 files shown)


gcapalbo:
hmm, strange. i kind of liked this check because it makes sure that we don't have workers around running old code that can still accept new tasks. i think it would be worth trying a different way to check before we remove this entirely. i tested locally, and using this seems to produce the behavior we want (no worker present in warm shutdown):

    from celery import Celery
    app = Celery()
    app.config_from_object(settings)
    app.control.ping()

if flower is the problem, i think i'd like to try this out before we remove the check entirely. what do you think?

benrudolph:
yeah i also liked the check. but flower was inexplicably returning that the queue was available for a short period of time and then saying that it wasn't. i couldn't reliably reproduce it either.

i think that's a good call, will play around with that today

PR #13534, opened 2016-10-11
target = parent for app preview

benrudolph:
@biyeun this'll ensure that when we hit a build error link in the preview app, we'll change the url of the parent window.

cc: @orangejenny

### DIFF
--- a/corehq/apps/app_manager/templates/app_manager/partials/build_errors.html
+++ b/corehq/apps/app_manager/templates/app_manager/partials/build_errors.html
@@ -317,5 +317,12 @@
                 {% trans "Errors Building an Application" %}
             </a>
         </p>
+        {% if in_iframe %}
+        <script>
+            $('#build-errors a').each(function(index) {
+                $(this).attr('target', '_parent');
+            });
+        </script>
+        {% endif %}
     </div>
 {% endif %}
--- a/corehq/apps/app_manager/views/cli.py
+++ b/corehq/apps/app_manager/views/cli.py
@@ -100,7 +100,8 @@ def direct_ccz(request, domain):
             'build_errors': errors,
             'domain': domain,
             'langs': langs,
-            'lang': lang
+            'lang': lang,
+            'in_iframe': True,
         })
         return json_response(
             {'error_html': error_html},


corehq/apps/app_manager/templates/app_manager/partials/build_errors.html
```diff
@@ -317,5 +317,12 @@
                 {% trans "Errors Building an Application" %}
             </a>
         </p>
+        {% if in_iframe %}
+        <script>
+            $('#build-errors a').each(function(index) {
+                $(this).attr('target', '_parent');
+            });
+        </script>
+        {% endif %}
```
benrudolph: didn't love using a script for this, but it would have been a whole bunch of copy paste `if iframe add target`

orangejenny: Could this go in a more general script like app_manager.js? Eventually we're going to be ditching inline javascript.

benrudolph: sadly no, because this is rendered as a partial, not part of the entire page (we just return the html)

orangejenny: I don't follow...as long as the js is included on every page that uses this partial, isn't that still fine? I'm thinking of something like 
```$(document).on("iframe.preview-phone-container a", "click", function() {
   document.location = $(this).attr("href");
});```

benrudolph: the example of `app_manager.js` wouldn't work because the partial rendered in preview app doesn't have that script. the partial in general is generated like this: https://github.com/dimagi/commcare-hq/blob/master/corehq/apps/app_manager/views/cli.py#L97

and then we return that html in a json call from formplayer. it's arguable a little less clear if you need to have another script loaded in order for that page to render correctly.

the example js code you gave i don't think would work since i _think_ document would refer to the iframe document and thus change the URL of the iframe, not the parent window. though definitely plausible that with a little hacking we could make that work. happy to quickly sync offline

orangejenny: Hmm...I see what you mean, though I still think eventually we're going to want to find a way around this. Yeah, let's catch up offline.

PR #13535, opened 2016-10-11
Parse celery pings

benrudolph:
@gcapalbo here's a version using celery pings. going to throw it up on staging

cc: @orangejenny

### DIFF
--- a/corehq/apps/hqadmin/utils.py
+++ b/corehq/apps/hqadmin/utils.py
@@ -90,11 +90,10 @@ def get_celery_stats():
 def parse_celery_pings(worker_responses):
     pings = {}
     for worker in worker_responses:
-        assert len(worker.keys()) == 1 and '@' in worker.keys()[0]
+        assert len(worker.keys()) == 1
 
         worker_fullname = worker.keys()[0]
-        worker_hostname = worker_fullname.split('@')[1]
-        pings[worker_hostname] = worker[worker_fullname].get('ok') == 'pong'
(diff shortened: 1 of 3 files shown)


corehq/apps/hqadmin/utils.py
```diff
@@ -87,6 +87,17 @@ def get_task_html(detailed_stats, worker_name):
...
 
 
+def parse_celery_pings(worker_responses):
+    pings = {}
+    for worker in worker_responses:
+        assert len(worker.keys()) == 1 and '@' in worker.keys()[0]
+
+        worker_fullname = worker.keys()[0]
+        worker_hostname = worker_fullname.split('@')[1]
+        pings[worker_hostname] = worker[worker_fullname]['ok'] == 'pong'
```
gcapalbo: we might want to use `worker[worker_fullname].get('ok')` just to be safe so we never fail hard

gcapalbo:
looks great, thanks. just one minor comment otherwise i think we're good to merge once it's tested on staging

corehq/apps/hqadmin/utils.py
benrudolph: 👍 https://github.com/dimagi/commcare-hq/pull/13535/commits/c545efdfee2ba691500351d0e881a26953e94941

gcapalbo:
good to merge from my end, let me know once you're comfortable with the staging test

benrudolph:
caught one dumb mistake, but should be good to go now

PR #13536, opened 2016-10-11
Allow manually modifying form link datums

proteusvacuum:
This shows a button that allows you to modify form link datums even if those form links can be "automatically" linked.
This allows for form linking between forms with the same case type, where the first form both loads and creates a case of the same type. This way you can choose the case you want to load (i.e. the case that is loaded, or the case that is created). 
@snopoke heads up
@millerdev

### DIFF
--- a/corehq/apps/app_manager/static/app_manager/js/form_workflow.js
+++ b/corehq/apps/app_manager/static/app_manager/js/form_workflow.js
@@ -108,6 +108,7 @@ hqDefine('app_manager/js/form_workflow.js', function() {
         self.autoLink = ko.observable();
         self.forms = workflow.forms || [];
         self.datums = ko.observableArray();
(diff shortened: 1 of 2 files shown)


millerdev:
@proteusvacuum: not sure if @snopoke got a notification on this since it looks like you @ mentioned his name in the PR description in an edit that happened after the original submission.

millerdev (approved):
Would it be hard to add tests for the JS code changes? Maybe they're too small to bother? Otherwise nothing stuck out to me, but I don't feel like I have much context here.

benrudolph:
there are actually tests for this exact file 😄 https://github.com/dimagi/commcare-hq/blob/master/corehq/apps/app_manager/static/app_manager/spec/form_workflow_spec.js

benrudolph:
though it does look like you might not get that many useful tests from your changes

PR #13539, opened 2016-10-12
do less work when updating domain stats

snopoke:
image

Currently ES gets pummeled once a day when this task runs. This should reduce the load significantly.
@sravfeyn @NoahCarnahan


czue:
nice.

benrudolph:
🙌

PR #13544, opened 2016-10-12
No random urls in bower

emord:
noticed this when looking through the IPs list

@benrudolph is there any upside to using the urls that I'm missing

buddy @czue

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -56,13 +56,13 @@
     "requirejs": "~2.3.1",
     "datetimepicker": "2.5.4",
     "DOMPurify": "0.8.3",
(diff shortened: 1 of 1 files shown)


benrudolph:
not exactly sure how the internals work, but would this mean that our `cchq` would need a public key on github?

emord:
@benrudolph I don't think so. we're already using similar notation https://github.com/dimagi/commcare-hq/blob/master/bower.json#L48

benrudolph:
cool, sounds good then

benrudolph (approved):

emord:
Closing this for now. Looks like sinon uses require which would have to be setup

PR #13546, opened 2016-10-12
First pass on cloudcare rename

benrudolph:
@czue @dimagi/product first pass on renaming cloudcare to "web apps." one fairly hard thing to change is the actual CloudCare setting in the app manager: https://github.com/dimagi/commcare-hq/blob/br/rename-cloudcare/corehq/apps/app_manager/static/app_manager/json/commcare-app-settings.yaml#L80. Cory wonder if you know any clever ideas to be able to toggle the name based on the flag. otherwise i think we'll have to do something pretty hacky. Haven't touched the billing or pricing too much either. Pricing will be hard in the sense that we may have users who haven't logged in yet so we don't know what their status is in regards to Old Cloudcare vs. New

cc: @orangejenny

### DIFF
--- a/corehq/apps/cloudcare/templates/cloudcare/config.html
+++ b/corehq/apps/cloudcare/templates/cloudcare/config.html
@@ -148,11 +148,10 @@
         <div id="cloudcare-app-settings-form" data-bind="with: applicationAccess">
             <label class="radio">
                 <input type="radio" value="false" data-bind="checked: restrict.JSON"/>
-                {% blocktrans %}Allow all mobile workers to see all {% endblocktrans %}
                 {% if request|is_new_cloudcare %}
-                web apps.
+                {% blocktrans %}Allow all mobile workers to see all web apps.{% endblocktrans %}
                 {% else %}
-                CloudCare {% trans "apps" %}.
+                {% blocktrans %}Allow all mobile workers to see all CloudCare apps.{% endblocktrans %}
                 {% endif %}
             </label>
             <label class="radio">
--- a/corehq/apps/cloudcare/templates/cloudcare/config.html
+++ b/corehq/apps/cloudcare/templates/cloudcare/config.html
@@ -138,17 +138,30 @@
 {% endblock %}
 
 {% block page_content %}
+    {% if request|is_new_cloudcare %}
+    <h2>{% trans 'Manage Web Apps Permissions' %}</h2>
+    {% else %}
     <h2>{% trans 'Manage CloudCare Permissions' %}</h2>
+    {% endif %}
     <section id="cloudcare-app-settings" style="display: none">
         <div data-bind="saveButton: saveButton"></div>
         <div id="cloudcare-app-settings-form" data-bind="with: applicationAccess">
             <label class="radio">
                 <input type="radio" value="false" data-bind="checked: restrict.JSON"/>
-                {% blocktrans %}Allow all mobile workers to see all CloudCare apps{% endblocktrans %}
+                {% blocktrans %}Allow all mobile workers to see all {% endblocktrans %}
+                {% if request|is_new_cloudcare %}
+                web apps.
+                {% else %}
+                CloudCare {% trans "apps" %}.
+                {% endif %}
             </label>
             <label class="radio">
                 <input type="radio" value="true" data-bind="checked: restrict.JSON"/>
+                {% if request|is_new_cloudcare %}
+                {% blocktrans %}Customize each mobile worker's access to web apps{% endblocktrans %}
+                {% else %}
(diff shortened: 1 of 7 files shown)


corehq/apps/export/templates/export/download_export.html
```diff
@@ -103,8 +103,17 @@
                         <a href="https://confluence.dimagi.com/display/commcarepublic/Deploy+and+Install+an+Application+on+CommCareHQ"
                            target="_blank">deploy</a>
                         your application and submit data from a phone. You may also
+                    {% endblocktrans %}
                         <a href="https://confluence.dimagi.com/display/commcarepublic/CloudCare+-+Web+Data+Entry"
-                           target="_blank">submit data via CloudCare</a>, depending on your project's
+                           target="_blank">{% trans "submit data via" %}
+                           {% if request|is_new_cloudcare %}
+                           web apps
```
orangejenny: same comment as in config.html

corehq/apps/cloudcare/templates/cloudcare/config.html
```diff
@@ -138,17 +138,30 @@
...
+    {% endif %}
     <section id="cloudcare-app-settings" style="display: none">
         <div data-bind="saveButton: saveButton"></div>
         <div id="cloudcare-app-settings-form" data-bind="with: applicationAccess">
             <label class="radio">
                 <input type="radio" value="false" data-bind="checked: restrict.JSON"/>
-                {% blocktrans %}Allow all mobile workers to see all CloudCare apps{% endblocktrans %}
+                {% blocktrans %}Allow all mobile workers to see all {% endblocktrans %}
+                {% if request|is_new_cloudcare %}
+                web apps.
```
benrudolph: yeah i'm still unclear if web apps is a proper noun or just regular language. @amsagoff thoughts on whether it should be translated?

orangejenny:
Does https://github.com/dimagi/commcare-hq/blob/master/corehq/apps/users/forms.py need to be updated in this pass?

benrudolph:
probably could. wish there was some language we could use that would work in both old and new. not a huge fan of adding a bunch of logic just to change the wording

orangejenny:
> not a huge fan of adding a bunch of logic just to change the wording

yeah, agreed :-/

orangejenny:
> toggle the name based on the flag

Maybe extend the `toggle` attribute to understand `!` and then have one copy of the setting with `toggle: USE_FORMPLAYER_FRONTEND` and a second with `toggle: !USE_FORMPLAYER_FRONTEND`? Kind of hacky, but could be worse.

amsagoff:
You can ignore billing and pricing for now. I'll also try to get everyone
moved over to web apps ASAP so we don't have to worry about long-term
maintenance here.

czue:
probably you guys thought about this, but was there a specific reason we decided not to just update the language before the wider release to make this simpler? i feel like we could do that first without the marketing work and it's not obvious to me that it would cause a major issue (most real cloudcare users only have the one tab anyways)

amsagoff:
Definitely worth weighing the tradeoffs if the lift here is more than
minimal (my impression is that it would be minimal). I am also very OK with
doing something smaller like renaming the app setting for everyone but
toggling what's in the top-level nav.

The failure mode is that users are confused about whether or not they are
on web apps vs. commcare, which makes communicating around upgrading more
confusing/difficult. Ideally the messaging we give the user is consistent
across our marketing materials and our product.

Ben, if this is proving tricky we can offline about the tradeoffs

benrudolph:
Ok agreed. I think for the tricky ones we can just use one verbiage (we do used Web Based Applications in some language so i don't think it'd be overly confusing) and then for the ones where it requires significant amount of hacking we can weight the visibility of the text

czue:
cool, sounds like you guys have a good plan forward ( @benrudolph to explicitly answer your question I don't know any non-hacky way to update that setting short of adding a new piece of framework which seems like overkill for this).

krsdimagi:
@orangejenny is this a fairly easy merge at this point?  Would be kind of nice to get into today's deploy for some rebranded newsletter updates.

benrudolph:
Hey Kevin I'll make some updates to this and then make sure it gets into
today's deploy

benrudolph:
ok i think this is fine for now, working getting the app setting to work correctly for new cloudcare folks. think we should go ahead and merge this though. @orangejenny i updated with your feedback

corehq/apps/export/templates/export/download_export.html
```diff
@@ -98,15 +98,31 @@
                     {% endblocktrans %}
                 </p>
                 <p>
+                     {% if request|is_new_cloudcare %}
```
emord: nit: looks like an extra space here

PR #13548, opened 2016-10-12
Organizations permissions

emord:
http://manage.dimagi.com/default.asp?239222

Limits editing structure of locations to editors of mobile works and editing levels to app editors

buddy @czue

### DIFF
--- a/corehq/apps/locations/views.py
+++ b/corehq/apps/locations/views.py
@@ -23,7 +23,6 @@ from soil.exceptions import TaskFailedError
 from soil.util import expose_cached_download, get_download_context
 
 from corehq import toggles
-from corehq.apps.app_manager.decorators import require_can_edit_apps
 from corehq.apps.commtrack.tasks import import_locations_async, location_lock_key, LOCK_LOCATIONS_TIMEOUT
 from corehq.apps.commtrack.util import unicode_slug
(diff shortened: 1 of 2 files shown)


czue:
@sheelio are we documenting these restrictions anywhere? seems complicated/non-intuitive enough that it would be good to do.

corehq/apps/locations/views.py
```diff
@@ -122,6 +123,7 @@ class LocationsListView(BaseLocationView):
 
     @use_jquery_ui
     @location_safe
+    @method_decorator(require_can_edit_commcare_users)
```
esoergel: @sheelio are you referring to the "Users" tab of an "Edit Location" page?  If so, I believe this check needs to go at the top of this method: https://github.com/dimagi/commcare-hq/blob/f806fa5e06fd86dc999ecdefec2cf3621da555f0/corehq/apps/locations/views.py#L602-L615

sheelio:
@czue 
WIll document the restrictions on the web user page.

corehq/apps/locations/views.py
esoergel: @sheelio and I offlined, and it sounds like this should go in `locations_access_required`, as no locations views should be accessible to those users.

PR #13550, opened 2016-10-12
Clear case type

orangejenny:
http://manage.dimagi.com/default.asp?239243

### DIFF
--- a/corehq/apps/app_manager/models.py
+++ b/corehq/apps/app_manager/models.py
@@ -883,7 +883,7 @@ class FormBase(DocumentSchema):
                 return self.validate_form()
         return self
 
-    def validate_for_build(self, validate_module=True):
+    def validate_for_build(self):
         errors = []
 
         try:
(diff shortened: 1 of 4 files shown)


orangejenny:
@snopoke 
cc @benrudolph

corehq/apps/app_manager/models.py
```diff
@@ -5560,7 +5558,7 @@ def validate_app(self):
             errors.extend(module.validate_for_build())
 
         for form in self.get_forms():
-            errors.extend(form.validate_for_build(validate_module=False))
+            errors.extend(form.validate_for_build())
```
snopoke: Looking at this again and where it's called I think this was added to prevent duplicate work. This is the only place this is called with `validate_module=False` and it's inside the `app.validate_app()` function which already does module validation. 

It's possible that the logic in `form.validate_for_build` and `module.validate_for_build` have somehow diverged so that calling `form.validate_for_build(validate_module=True)` does more module level validation than `module.validate_for_build()` but if that's the case then it should probably be fixed to be consistent.

I don't have any other concerns about this change other than the performance implications of doing the module level validation repeatedly for every form in the app.

snopoke (commented):
Not sure how this fixes the linked FB ticket.

orangejenny:
https://github.com/dimagi/commcare-hq/pull/13550/commits/0557e15822097005ee331fcf204610b0456dc505 is the commit that fixes the issue in the ticket, removing validation on both the client and the server.

Fair point about performance; having every form re-run module validation wouldn't be good. I reverted the commits that got rid of `validate_module` and instead updated modules to run case validation if they contain any registration forms (and therefore require a case).

PR #13551, opened 2016-10-12
Shorten generated blob identifiers

millerdev:
This should reduce the amount of RAM needed by the blob db backend. I have never once needed the filename that was previously embedded in the blob id, so I don't mind removing it.

Best to review commits individually. 51eb53d is refactoring only.

@snopoke @dannyroberts cc @proteusvacuum

### DIFF
--- a/corehq/blobs/interface.py
+++ b/corehq/blobs/interface.py
@@ -6,6 +6,7 @@ from uuid import uuid4
 
 from corehq.blobs import DEFAULT_BUCKET
 from corehq.blobs.exceptions import ArgumentError
+from corehq.blobs.util import random_url_id
 
 SAFENAME = re.compile("^[a-z0-9_./{}-]+$", re.IGNORECASE)
 NOT_SET = object()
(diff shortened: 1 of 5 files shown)


corehq/blobs/tests/test_fsdb.py
```diff
@@ -173,12 +137,60 @@ def test_empty_attachment_name(self):
...
+    @classmethod
+    def tearDownClass(cls):
+        cls.db = None
+        rmtree(cls.rootdir)
+        cls.rootdir = None
+
+    def test_put_with_colliding_blob_id(self):
+        # unfortunately can't do this on S3 because there is no way to
+        # reliably check if an object exists before putting it.
+        with patch("corehq.blobs.interface.random_url_id", new=lambda n:'not-unique'):
```
dimagimon: E231 missing whitespace after ':'

snopoke:
I assume this will only apply to new attachments?

millerdev:
Correct. We could write a migration to generate new ids for existing blobs if we wanted to save even more space.

corehq/blobs/interface.py
```diff
@@ -86,13 +87,28 @@ def copy_blob(self, content, info, bucket):
...
+        change this if we ever expect an object to have significantly
+        more than 1000 blobs (attachments). The probability of a
+        collision with a 64 bit ID is:
+
+        k = 1000
+        N = 2 ** 64
+        (k ** 2) / (2 * 2 ** 64) = 2.7e-14
+
+        which is somewhere near the probability of a meteor landing on
+        your house. For most objects the number of blobs present at any
```
benrudolph: 🌎 :comet:

PR #13552, opened 2016-10-12
Update case property to match app

proteusvacuum:
The case properties in the eNikshay domain have changed since this was written. Updating them to match reality. (No backwards compatibility issues as no one is using this feature yet).
@millerdev


millerdev (approved):

millerdev:
Kicked travis (timeout)

PR #13559, opened 2016-10-13
remove incompatible uses of RelatedManager

nickpell:
needed for django 1.9


nickpell:
@kaapstorm

kaapstorm:
Just one nit. Merge at will.

PR #13560, opened 2016-10-13
target parent on link click for app preview

benrudolph:
@orangejenny @biyeun better way of ensuring that we open the link on the parent page


orangejenny:
looks good to me

PR #13570, opened 2016-10-13
move google analytics out of analytics_all.html

orangejenny:
Move google analytics out of analytics_all.html so that prelogin can use analytics_all.html.

@calellowitz

### DIFF
--- a/corehq/apps/style/templates/style/base.html
+++ b/corehq/apps/style/templates/style/base.html
@@ -193,6 +193,7 @@
         {% block head %}
         {% endblock %}
 
+        {% include 'style/includes/analytics_google.html' %}
         {% include 'style/includes/analytics_all.html' %}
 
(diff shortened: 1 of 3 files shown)


calellowitz:
is it possible to just include `analytics_google` in prelogin?

orangejenny:
Re-reading through `google_analytics.html` this morning...I think so. The only minor point is that hq and prelogin create the tracking object slightly differently: HQ passes in username or "none", which ought to be fine for prelogin; and prelogin sets the cookieDomain `auto`, which I think should be fine for HQ.

Putting GA back in analytics_all does mean the GA code is now being run twice on prelogin. Fixing that in https://github.com/dimagi/commcarehq-prelogin/pull/99

calellowitz:
👍 on ✅

PR #13574, opened 2016-10-14
use explict imports for from django.conf.urls

nickpell:
`patterns` is deprecated, so I want to make it explicit that we're using it (and easier for removal)


nickpell:
@emord

PR #13579, opened 2016-10-14
Get current app in download views when target=save

esoergel:
@orangejenny


esoergel:
@amstone326 FYI

amstone326:
:tada: so excited

PR #13583, opened 2016-10-14
use passed args in migrate_multi

nickpell:
@snopoke 

Needed when faking a single migration.


snopoke:
@emord updated to include docs for args (+a bit extra)

PR #13584, opened 2016-10-14
Fix cmitfb migration for django 1.8

emord:
@orangejenny


emord:
@krsdimagi this is what will have to be deployed to continue the migration

orangejenny:
no test coverage, killed build

PR #13587, opened 2016-10-14
remove deprecated SubfieldBase

nickpell:
@biyeun


nickpell:
cc: @kaapstorm

kaapstorm:
Do you need to add `from_db_value` and `to_python` methods, as suggested [on SO](http://stackoverflow.com/a/36744699)?

nickpell:
It already has a ```to_python```.  This is used when deserializing data from a form.

```from_db_value``` is used when deserializing data from the DB, and ```TimeZoneField``` is never used that way in HQ.  I'd skip adding the function if it's never going to get used.

PR #13590, opened 2016-10-14
add SessionAuthenticationMiddleware to MIDDLEWARE

nickpell:
https://docs.djangoproject.com/en/1.10/topics/auth/default/#session-invalidation-on-password-change

@kaapstorm


kaapstorm:
Surprised this wasn't there before.

PR #13591, opened 2016-10-14
define TEMPLATES in settings.py

nickpell:
https://docs.djangoproject.com/en/1.10/ref/templates/upgrading/


nickpell:
@snopoke @millerdev Running successfully on staging

PR #13592, opened 2016-10-15
refresh form properties list

proteusvacuum:
I'm surprised refreshing the module settings page every time you make a change to the app wasn't more annoying for people.
This adds a refresh button that fetches new questions from the form that you are editing, likely in another tab.
## Regular modules

peek 2016-10-15 00-19
## Advanced modules (and conditionals)

peek 2016-10-15 00-45

@dimagi/product, @orangejenny  what do you think?
cc: @millerdev

### DIFF
--- a/corehq/apps/app_manager/static/app_manager/js/case-config-utils.js
+++ b/corehq/apps/app_manager/static/app_manager/js/case-config-utils.js
@@ -1,3 +1,4 @@
+/* globals alert_user */
 hqDefine('app_manager/js/case-config-utils.js', function () {
     return {
         getQuestions: function (questions, filter, excludeHidden, includeRepeat, excludeTrigger) {
@@ -42,6 +43,23 @@ hqDefine('app_manager/js/case-config-utils.js', function () {
                 }
             }
             return options;
+        },
+        refreshQuestions: function(questions_observable, url, moduleId, formId, event){
+            var $el = $(event.currentTarget);
+            $el.find('i').addClass('fa-spin');
+            $.get(
+                url,
+                {module_id: moduleId, form_id: formId}
+            ).success(function(data){
+                $el.addClass('btn-success').removeClass('btn-danger');
+                questions_observable(data);
+                $el.find('i').removeClass('fa-spin');
+            }).error(function(e){
+                $el.removeClass('btn-success').addClass('btn-danger');
+                $el.find('i').removeClass('fa-spin');
+                alert_user("Something went wrong refreshing your form properties. "
+                           + "Please refresh the page and try again", "danger");
+            });
         },
         filteredSuggestedProperties: function (suggestedProperties, properties) {
             var used_properties = _.map(properties, function (x) {
--- a/corehq/apps/app_manager/urls.py
+++ b/corehq/apps/app_manager/urls.py
@@ -21,7 +21,7 @@ from corehq.apps.app_manager.views import (
     edit_report_module, validate_module_for_build, commcare_profile, edit_commcare_profile, edit_commcare_settings,
     edit_app_langs, edit_app_attr, edit_app_ui_translations, get_app_ui_translations, rearrange, odk_qr_code,
     odk_media_qr_code, odk_install, short_url, short_odk_url, save_copy, revert_to_copy, delete_copy, list_apps,
-    direct_ccz, download_index, download_file, formdefs, release_manager,
+    direct_ccz, download_index, download_file, formdefs, release_manager, get_form_questions,
 )
 from corehq.apps.hqmedia.urls import application_urls as hqmedia_urls
 from corehq.apps.hqmedia.urls import download_urls as media_download_urls
@@ -54,6 +54,7 @@ app_urls = [
     url(r'^modules-(?P<module_id>[\w-]+)/forms-(?P<form_id>[\w-]+)/$',
         view_form, name='view_form'),
     url(r'^get_form_datums/$', get_form_datums, name='get_form_datums'),
+    url(r'^get_form_questions/$', get_form_questions, name='get_form_questions'),
     url(r'^modules-(?P<module_id>[\w-]+)/forms-(?P<form_id>[\w-]+)/source/$',
         form_designer, name='form_source'),
     url(r'^summary/$', AppSummaryView.as_view(), name=AppSummaryView.urlname),
--- a/corehq/apps/app_manager/static/app_manager/js/case-config-ui-2.js
+++ b/corehq/apps/app_manager/static/app_manager/js/case-config-ui-2.js
@@ -1,4 +1,4 @@
-/*globals $, COMMCAREHQ, ko, _, alert_user*/
+/*globals $, COMMCAREHQ, ko, _*/
 
 hqDefine('app_manager/js/case-config-ui-2.js', function () {
     "use strict";
@@ -126,21 +126,7 @@ hqDefine('app_manager/js/case-config-ui-2.js', function () {
         };
 
         self.refreshQuestions = function(url, moduleId, formId, event){
-            var $el = $(event.currentTarget);
-            $el.find('i').addClass('fa-spin');
-            $.get(
-                url,
-                {module_id: moduleId, form_id: formId}
-            ).success(function(data){
-                $el.addClass('btn-success').removeClass('btn-danger');
-                self.questions(data);
-                $el.find('i').removeClass('fa-spin');
-            }).error(function(e){
-                $el.removeClass('btn-success').addClass('btn-danger');
-                $el.find('i').removeClass('fa-spin');
-                alert_user("Something went wrong refreshing your form properties. "
-                           + "Please refresh the page and try again", "danger");
-            });
+            return caseConfigUtils.refreshQuestions(self.questions,url, moduleId, formId, event);
         };
         self.getAnswers = function (condition) {
             return caseConfigUtils.getAnswers(self.questions(), condition);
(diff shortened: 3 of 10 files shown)


corehq/apps/app_manager/static/app_manager/js/case-config-utils.js
```diff
@@ -43,6 +44,23 @@ hqDefine('app_manager/js/case-config-utils.js', function () {
...
+            var $el = $(event.currentTarget);
+            $el.find('i').addClass('fa-spin');
+            $.get(
+                url,
+                {module_id: moduleId, form_id: formId}
+            ).success(function(data){
+                $el.addClass('btn-success').removeClass('btn-danger');
+                questions_observable(data);
+                $el.find('i').removeClass('fa-spin');
+            }).error(function(e){
```
dimagimon: 'e' is defined but never used (no-unused-vars)

czue:
this is sweet

benrudolph:
this is pretty cool. curious if we should just be polling every 30 seconds or something so that people don't need to hit the refresh button

proteusvacuum:
huh, I didn't really think of doing that. Could that be confusing if you (or someone else) are actively editing an app and have things on that page change on you all the time?

benrudolph:
yeah not too sure. my thought is essentially you always want the change. i don't think there's a situation where you'd want to have the out dated data (in an ideal world it'd instantly reflect the current state of the app). also 30 seconds could be about the time it takes to edit a question id and go back to another tab to update a question. the thing that would be bad is if the dropdown disappears on you if you have it open and it gets refreshed

corehq/apps/app_manager/templates/app_manager/partials/case_config_shared.html
```diff
@@ -180,3 +182,12 @@
         </a>
     </div>
 </script>
+
+<script type="text/html" id="case-config:refresh-form-questions">
+    <div style="margin-right: 1em;"
```
millerdev: is `style=...` a thing we do frequently? Best practice is to put that in a CSS file, right?

millerdev (approved):
Looks good to me. I agree that it would be ideal for the user not to have to think about refreshing manually, although I'm not wild about polling every 30s. It would seem better to me to tie into the form modification/concurrent edit notification system that @czue implemented recently and update the data on this page each time a form in the module is modified. For now this seems better than what we had before, although I'd be interested to know if @orangejenny agrees.

corehq/apps/app_manager/templates/app_manager/partials/case_config_shared.html
orangejenny: Yeah, style should go in an external file. But for this case, you probably want a [http://getbootstrap.com/components/#btn-groups-toolbar](button toolbar) anyway, not a custom style.

corehq/apps/app_manager/urls.py
```diff
@@ -35,6 +35,7 @@
     url(r'^modules-(?P<module_id>[\w-]+)/forms-(?P<form_id>[\w-]+)/$',
         'view_form', name='view_form'),
     url(r'^get_form_datums/$', 'get_form_datums', name='get_form_datums'),
+    url(r'^get_form_questions/$', 'get_form_questions', name='get_form_questions'),
```
orangejenny: Minor, but seems like we more often put module & form id directly in the url.

orangejenny:
Hey, I'm not sold that this in an improvement in its current incarnation. It's a very prominent (location + color) button on a heavily trafficked page that solves a problem that I suspect non-tech-savvy users aren't even aware is a problem.

> I'm surprised refreshing the module settings page every time you make a change to the app wasn't more annoying for people.

It is. It is very annoying. Though a lot of the annoyance comes from the page being slow, so my vague plan for dealing with it has been to make this page smaller & faster by doing things like loading the separate tabs on demand rather than as one giant page load.

However, now we have the new app manager notifications, and I agree with Daniel that this could fit really well into those notifications. I could see an alert that pops up saying, "This form has been modified." and has a link to refresh questions. That could be a good experience, and I think it's not that far removed from what's already here. Thoughts?

proteusvacuum:
@orangejenny I really like this idea:
> I could see an alert that pops up saying, "This form has been modified." and has a link to refresh questions. That could be a good experience, and I think it's not that far removed from what's already here. Thoughts?

I'm not a huge fan of things changing underneath you on a page that you aren't looking at, (and having something invisibly updating at an interval I can't control could be even more frustrating than just refreshing the page). I think this is a great solution to those problems. 

I'll look into integrating this with the websockets framework.

emord:
> I'm not a huge fan of things changing underneath you on a page that you aren't looking at,

I think auto updating would be similar to having an async select2 input, which may be the nicest solution, but probably not trivial to put on this page. I could see having a warning if the form changed in some way that caused case management to become invalid (like deleting a question)

proteusvacuum:
> async select2 input, which may be the nicest solution
I think the async select wouldn't actually solve this because there are a few more places on that page that use the form questions (e.g. the conditional question options), and the validation of whether a question still exists or not. With the way I implemented this, you get all that 'for free'.

emord:
> I think the async select wouldn't actually solve this because there are a few more places on that page that use the form questions (e.g. the conditional question options), 

Yeah. Ideally those would also use an async select2

> and the validation of whether a question still exists or not. With the way I implemented this, you get all that 'for free'.

Agreed that this is the easiest way. I was thinking that it could be solved via:

> warning if the form changed in some way that caused case management to become invalid

(none of this is blocking, was just thinking about other solutions)

czue:
just fyi if you do go with the websockets thing then this has to be superuser only for now.

also would recommend trying to reuse the existing code as much as possible just because we are likely to change our websocket library soonish and so the fewer touch points we have the better.

amsagoff:
Just a thought - before we go deeper down this path - is there somewhere
less-obtrusive we could release this initially as a test? We could collect
analytics and either pull it from the product or keep it/improve it
depending on metrics like over 30% of people who click it once end up
clicking it >5 times.

One thought for a temporary home could be as a gray button (less obtrusive)
in the pane where the case properties are (so it's clearer that's what it's
refreshing)

proteusvacuum:
@amsagoff I added this behind a feature flag and also added analytics. 
I tried making this hook up to the the Websockets feature, but I found it much more confusing, and less discoverable than the button here. If we want to do that, it might require some more thought, maybe alongside how features like that would work in general with the rest of the app manager redesign.

Also, just adding the button on the "case properties" was somewhat misleading too, as the whole tab updates, so I left it where it was for now. I'd be happy to open this up to a better venue for design feedback than github comments if you think this isn't a good idea though. I figured I'd try and get it into the hands of FM's to get their feedback on the functionality so we can iterate from that maybe.

amsagoff:
Agree doing some hallway testing makes sense as a next step. Happy to help
with that.

proteusvacuum:
OK, cool. Removing the do not merge label so we can deploy it and get some feedback. ✅

proteusvacuum:
(new code buddy @nickpell )

proteusvacuum:
Bump

orangejenny:
Would be nice to add this to usercase config, too, probably in corehq/apps/app_manager/templates/app_manager/v1/partials/usercase_config.html

But that isn't blocking, going to merge.

benrudolph:
👏 nice push on getting this in!

krsdimagi:
Farid and I just did some hallway testing for this feature, and have the following thoughts and potential next steps:

Feedback:

1. Users unanimously appreciated the ability to more quickly refresh the case properties once they knew about the feature, and liked the error message when deleting a question from the form but not deleting the case property.
2. First instinct was generally to refresh the page - but on further prompting the refresh button was discoverable - consensus was that for a new user, this may be more discoverable than the page refresh anyway.
4. There was some disagreement on release - two users thought that it should be released widely to everyone, whereas two other though it might be good to keep it as an internal tool, as FMs may build apps with more tabs open than other users (although this is anecdotal and I'm not really sure we can say this is the case).
5. All users immediately wanted to enable the feature, and were interesting in thinking of other places similar functionality might save time and be useful - it is obviously valuable.

Potential next steps:

There are a couple considerations here:

1. Who do we release to?  Everyone, or just power users? 
2. Release the feature as-is
3. Improve/expand the feature and then release it - there were some suggestions for the feature, which included: including it on the case list/detail tabs, auto-checking for any changes that would require a refresh, and graying out the button until there were changes that would need a refresh to occur (users generally didn't like the idea of an auto-update, as this could be confusing in the app building workflow)
4. Decide it's not worth releasing (based on user feedback, I think this is definitely not the case)

proteusvacuum:
@amsagoff will defer to you on what you think we should do next. @krsdimagi do you have a recommendation of the different options you laid out?
I am leaning towards just removing the feature flag for now (or widely advertising the feature flag to FMs), and we can leave the more complex improvements for the app manager redesign. How does that sound?

amsagoff:
Thanks for the run-down @krsdimagi . I think we're trading off the benefits with the prominence in the UI for users who won't be able to discover/make use of this in their workflow. Overall it's feeling worth it, but I'm thinking there might be something we can do to make it a bit less prominent. Right now it looks about equal with "save" in terms of importance in the UI - wondering if it makes sense to make the color gray or something. Looking to @biyeun for input as the last blocker.

biyeun:
@amsagoff yeah, two green buttons isn't great. Gray or turquoise (`btn-info`) are my choices

PR #13593, opened 2016-10-15
move inline js out of js block

nickpell:
required when compressing in django 1.9

@kaapstorm


orangejenny:
Should the inline js now go in a `js-line` block?

nickpell:
@orangejenny updated: https://github.com/dimagi/commcare-hq/pull/13593/commits/48abce9303cbca44dd8f885ebeabde3a3754b09f

nickpell:
✅

PR #13594, opened 2016-10-15
mute internal notifications when subscription did not exist previously

nickpell:
http://manage.dimagi.com/default.asp?238566

@kaapstorm


kaapstorm:
Looks like the failing test just needs to be updated.

nickpell:
✅

PR #13606, opened 2016-10-17
Revert "Revert "partially supported SQL cloudcare API""

gcapalbo:
http://manage.dimagi.com/default.asp?239090
@snopoke everything looks ok to me except for the failing hard when footprint=True

Reverts dimagi/commcare-hq#13605

### DIFF
--- a/corehq/apps/cloudcare/api.py
+++ b/corehq/apps/cloudcare/api.py
@@ -150,7 +150,7 @@ class CaseAPIHelper(object):
                             case_list,
                             self.domain,
                             strip_history=self.strip_history,
-                        ).values()
+                        )
 
         return [CaseAPIResult(domain=self.domain, couch_doc=case, id_only=self.ids_only) for case in case_list]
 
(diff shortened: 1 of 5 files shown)


snopoke:
@NoahCarnahan @emord I'm at DSA summit from tomorrow. Could one of you look at this please. I think we should be able to use the [get_dependent_case_info](https://github.com/dimagi/commcare-hq/blob/3fc9712c88d55212b35512b78675194459dd85e5/corehq/ex-submodules/casexml/apps/phone/cleanliness.py#L205) function to get the list of case_ids.

corehq/apps/cloudcare/api.py
```diff
@@ -143,16 +150,16 @@ def _filter(res):
                             case_list,
                             self.domain,
                             strip_history=self.strip_history,
-                        ).values()
+                        )
```
dimagimon: E126 continuation line over-indented for hanging indent

emord:
@gcapalbo @snopoke https://github.com/dimagi/commcare-hq/pull/13606/commits/0bf69563d2f22d5a8bf0c8895d44a1425485b4aa

It's not clear to me whether this is the correct solution, but it seems to be a step in the right direction 

https://github.com/dimagi/commcare-hq/blob/revert-13605-revert-13597-sk/sql-cloudcare-api/corehq/apps/cloudcare/tests/test_api.py#L188-L209 do not pass on SQL domains. Are they also required?

snopoke:
reworked it a bit and now all tests pass on both except `testGetAllStripHistory` since SQL doesn't support strip history (but that seems like just an optimization).

gcapalbo:
@snopoke thanks, looks good. i'm just going to run a couple tests locally now before merging

gcapalbo:
(local testing passed)

PR #13607, opened 2016-10-17
Fix worker activity times in spanish

emord:
http://manage.dimagi.com/default.asp?239466#1233470

bad solution, but the google chart library doesn't support unicode axis labels. Ideally we'd just use a JS library for reports, but I think that's out of the scope for interrupt.

buddy @czue


czue (approved):
:+1:

czue:
build timed out. :boot:

PR #13609, opened 2016-10-17
Add get_sql_location for mobile workers

emord:
@esoergel http://manage.dimagi.com/default.asp?239572#1233519

buddy @czue

### DIFF
--- a/corehq/apps/users/models.py
+++ b/corehq/apps/users/models.py
@@ -1623,6 +1623,10 @@ class CommCareUser(CouchUser, SingleMembershipMixin, CommCareMobileContactMixin)
                 pass
         return None
 
+    @memoized
+    def get_sql_location(self, domain):
+        return self.sql_location
+
     def set_location(self, location):
         """
         Set the location, and all important user data, for


corehq/apps/users/models.py
```diff
@@ -1623,6 +1623,10 @@ def sql_location(self):
                 pass
         return None
 
+    @memoized
+    def get_sql_location(self, domain):
+        return self.sql_location
```
czue: @esoergel does this seem like the right approach?

czue (approved):

esoergel:
This seems fine as a quickfix approach.  It's more concerning that we have mobile workers making requests that we'd expect to come from web users.  Should `CommCareUser` be a subclass of `WebUser`?  I imagine we must have a bunch more of these incompatibilities.

emord:
fwiw I was only able to reproduce this issue when I assigned a mobile worker to a role that could edit mobile workers. There's a select2 on the mobile worker edit page.

It seems like a weird thing to do, but it's not something we specifically prohibit

czue:
> Should `CommCareUser` be a subclass of `WebUser`? 

I think this would probably create more confusion as they are fundamentally different in different ways (most notably in terms of domain membership) so would rather address whatever calling code doesn't realize that these are `CouchUser`s instead of the subclasses.

czue:
can we move the function up to `CouchUser`?

emord:
:white_check_mark:

esoergel:
@emord FYI Cory had marked this PR as approved, so it's legit for you to self-merge after the build passes.

PR #13611, opened 2016-10-17
Move archive and delete to SQLLocation

esoergel:
@sravfeyn @mkangia

(definitely wait on tests)


esoergel:
@calellowitz FYI

sravfeyn:
Good riddance!

I made a one comment, looks good otherwise

PR #13612, opened 2016-10-17
don't uninstall phonenumbers

emord:
@nickpell http://manage.dimagi.com/default.asp?239602

 ¯\_(ツ)_/¯

### DIFF
--- a/requirements/uninstall-requirements.txt
+++ b/requirements/uninstall-requirements.txt
@@ -1,5 +1,4 @@
 couchdbkit
-phonenumbers
 pycurl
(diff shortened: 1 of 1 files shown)


emord:
I'll play around with this more tomorrow, but it looks like `phonenumberslite` installs things in the `phonenumbers` directory and this removes it. Don't know why this works on prod, but not on staging yet

emord:
Still not sure why this is needed on staging, but works on prod/india/swiss.

That said it shouldn't be necessary anyways, so should be good to merge

PR #13614, opened 2016-10-18
upgrade twilio

nickpell:
http://manage.dimagi.com/default.asp?233526

https://github.com/twilio/twilio-python/blob/master/CHANGES.md

@gcapalbo This is needed to put Twilio prices on the SMS pricing page (re: email thread with subject "Twilio Gateway Fees").  Once tests pass, what do you recommend for QA?

### DIFF
--- a/requirements/requirements.txt
+++ b/requirements/requirements.txt
@@ -60,7 +60,7 @@ sqlagg==0.7.0
 django-redis==4.2
 redis==2.10.3
 # WeasyPrint==0.20.2
-twilio==3.6.4
+twilio==5.6.0
 django-countries==3.3
 reportlab==3.0
 stripe==1.12.2


gcapalbo:
@nickpell i think this is probably fine, i looked at the change log summary and seems like we should be ok. but i'll run a couple tests locally to confirm that and then i'll merge

gcapalbo:
@nickpell i tested locally and this looks good from my end.
@calellowitz is there anything you want to test regarding the two-factor stuff before we merge?

nickpell:
@calellowitz Let us know if any 2FA stuff needs testing here

calellowitz:
changelog looks fine on my end. If you want to test on staging you would need to deploy a localsettings change and set up the twilio backend with the right credentials. I dont have a strong opinion on whether that is necessary. i dont think it would break but if it does and someone only has sms two factor to get in then they could be locked out

nickpell:
@gcapalbo @calellowitz Tested 2FA on staging using the upgraded library, and it worked.

This is good to merge from me.

PR #13618, opened 2016-10-18
short term fix for not including dashes

emord:
@millerdev buddy @czue http://manage.dimagi.com/default.asp?239638

This will solve the issue of not being able to retrieve already saved attachments, but I will need to look deeper to make sure this is consistent between all forms. Offlined with @millerdev and it seems that ideally we would always use without dashes as it saves RAM

### DIFF
--- a/corehq/form_processor/models.py
+++ b/corehq/form_processor/models.py
@@ -423,7 +423,11 @@ class AbstractAttachment(DisabledDbMixin, models.Model, SaveStateMixin):
         try:
             blob = db.get(self.blob_id, self.blobdb_bucket())
         except (KeyError, NotFound, BadName):
-            raise AttachmentNotFound(self.name)
+            try:
+                # http://manage.dimagi.com/default.asp?239638
+                blob = db.get(self.blob_id, self.blobdb_bucket(True))
+            except (KeyError, NotFound, BadName):
+                raise AttachmentNotFound(self.name)
 
         if stream:
             return blob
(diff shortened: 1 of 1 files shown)


corehq/form_processor/models.py
```diff
@@ -440,12 +444,15 @@ def delete_content(self):
...
-    def blobdb_bucket(self):
+    def blobdb_bucket(self, remove_dashes=False):
         if self.blob_bucket is not None:
             return self.blob_bucket
         if self.attachment_id is None:
             raise AttachmentNotFound("cannot manipulate attachment on unidentified document")
-        return os.path.join(self._attachment_prefix, str(self.attachment_id))
+        attach_id = str(self.attachment_id)
+        if remove_dashes:
+            attach_id = attach_id.replace('-', '')
```
millerdev: You can get this with `self.attachment_id.hex` so
```py
if remove_dashes:
    attach_id = self.attachment_id.hex
else:
    attach_id = str(self.attachment_id)
```

corehq/form_processor/models.py
```diff
@@ -423,7 +423,11 @@ def read_content(self, stream=False):
         try:
             blob = db.get(self.blob_id, self.blobdb_bucket())
         except (KeyError, NotFound, BadName):
-            raise AttachmentNotFound(self.name)
+            try:
+                # http://manage.dimagi.com/default.asp?239638
+                blob = db.get(self.blob_id, self.blobdb_bucket(True))
```
millerdev: might want to check if `self.blob_bucket` is empty and `'-' in str(self.attachment_id)` before hitting riak again

corehq/form_processor/models.py
```diff
@@ -423,7 +423,14 @@ def read_content(self, stream=False):
         try:
             blob = db.get(self.blob_id, self.blobdb_bucket())
         except (KeyError, NotFound, BadName):
-            raise AttachmentNotFound(self.name)
+            if not self.blob_bucket and '-' in str(self.attachment_id):
+                try:
+                    # http://manage.dimagi.com/default.asp?239638
+                    blob = db.get(self.blob_id, self.blobdb_bucket(True))
```
czue: super nit-picky but for booleans i find it's nice to always call it as a kwarg so it's clear what the value is doing

czue (commented):
no blockers from me - will let @millerdev formally approve/merge

millerdev (approved):

PR #13619, opened 2016-10-18
Add docs for Location Permissions [ci skip]

esoergel:
@proteusvacuum @mkangia @sravfeyn @sheelio 
I added some introductory docs about how the location whitelist works.  Let me know if anything is unclear or if there are things I'm missing!

### DIFF
--- a/corehq/apps/locations/permissions.py
+++ b/corehq/apps/locations/permissions.py
@@ -1,3 +1,77 @@
+"""
+Location Permissions
+====================
+
+Normal Access
+-------------
+
+Location Types - Users who can edit apps on the domain can edit location types::
+Locations - Users who can edit mobile workers on the domain can edit locations
+
+
+Restricted Access and Whitelist
+--------------------------------
+
+Many large projects have mid-level users who should have access to a subset of
+the project based on the organization's hierarchy.
+
+This is handled by a special permission called "Full Organization Access" which
+is enabled by default on all user roles. To restrict data access based on a
+user's location, projects may create a user role with this permission disabled.
+
+This is checked like so::
+
+    user.has_permission(domain, 'access_all_locations')
+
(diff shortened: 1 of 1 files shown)


sravfeyn:
Nice, thanks for this!

proteusvacuum:
Nice! Is it worth putting this on readthedocs?

corehq/apps/locations/permissions.py
```diff
@@ -1,3 +1,77 @@
...
+be visible to the user. In this way, they only see menu and sidebar links that
+are accessible.
+
+To mark a view as location safe, you apply the ``@location_safe`` decorator to
+it. This can be applied directly to view functions, view classes, HQ report
+classes, or tastypie resources (see implentation and existing usages for
+examples).
+
+UCR and Report Builder reports will be automatically marked as location safe if
+the report contains a location choice provider.
```
mkangia: @esoergel 
Can you elaborate a bit on how this is achieved?

mkangia:
This is great!!! Thanks for compiling this @esoergel

corehq/apps/locations/permissions.py
```diff
@@ -1,3 +1,77 @@
...
+Many large projects have mid-level users who should have access to a subset of
+the project based on the organization's hierarchy.
+
+This is handled by a special permission called "Full Organization Access" which
+is enabled by default on all user roles. To restrict data access based on a
+user's location, projects may create a user role with this permission disabled.
+
+This is checked like so::
+
+    user.has_permission(domain, 'access_all_locations')
```
esoergel: That's a good point.  Farid and I just spoke about this and he pointed out that we already use the term "location" everywhere else in the code, so I think it makes sense to stick with that, so at least the code is consistent.

mkangia: And we can't change the permission name to 'Access All Locations'.

PR #13060, opened 2016-08-31
couch sql migrate

snopoke:
@czue

buddy @biyeun

Sorry this got a bit big:
- [ ] print migration stats
- [ ] save document diffs and allow printing them
- [ ] do timezone migration
- [ ] prevent form submissions during migration
- [ ] commit migration once complete
- [ ] migrate attachments & form operations
- [ ] copy unprocessed forms

PR #13623, opened 2016-10-18
attachment followup

emord:
@millerdev buddy @gcapalbo 

following up https://github.com/dimagi/commcare-hq/pull/13618

https://github.com/dimagi/commcare-hq/pull/13060/commits/1c8bda82f5f0487cd149c5589457bf37cd65f65b @snopoke added a `blob_bucket` to attachments during work on couch -> sql migrate but we don't actually write it for new sql forms. This PR saves that bucket when the attachment is created so that we don't run into a similar issue again.

I will have a followup PR to migrate old attachments to save their `blob_bucket`

### DIFF
--- a/corehq/form_processor/models.py
+++ b/corehq/form_processor/models.py
@@ -426,7 +426,7 @@ class AbstractAttachment(DisabledDbMixin, models.Model, SaveStateMixin):
             if not self.blob_bucket and '-' in str(self.attachment_id):
                 try:
                     # http://manage.dimagi.com/default.asp?239638
-                    blob = db.get(self.blob_id, self.blobdb_bucket(True))
+                    blob = db.get(self.blob_id, self.blobdb_bucket(remove_dashes=False))
                 except (KeyError, NotFound, BadName):
                     raise AttachmentNotFound(self.name)
             else:
@@ -447,7 +447,7 @@ class AbstractAttachment(DisabledDbMixin, models.Model, SaveStateMixin):
 
         return deleted
 
-    def blobdb_bucket(self, remove_dashes=False):
+    def blobdb_bucket(self, remove_dashes=True):
         if self.blob_bucket is not None:
             return self.blob_bucket
         if self.attachment_id is None:


corehq/form_processor/models.py
```diff
@@ -447,7 +447,7 @@ def delete_content(self):
 
         return deleted
 
-    def blobdb_bucket(self, remove_dashes=False):
+    def blobdb_bucket(self, remove_dashes=True):
```
emord: Reversed the logic here, because not having dashses decreases RAM use

corehq/form_processor/backends/sql/processor.py
```diff
@@ -33,6 +33,7 @@ def store_attachments(cls, xform, attachments):
                 attachment_id=uuid.uuid4(),
                 content_type=attachment.content_type,
             )
+            xform_attachment.blob_bucket = xform_attachment.blobdb_bucket()
```
millerdev: What do you think of moving this into `XFormAttachmentSQL.__init__`? Alternately, might make sense to move it into `AbstractAttachment.write_content`

millerdev (approved):

corehq/form_processor/backends/sql/processor.py
millerdev: On second thought, from a space perspective it does seem redundant to save this value in `blob_bucket` since it's simply `"form/" + self.attachment_id.hex`. It's more efficient (in terms of space) to not set this here since `blobdb_bucket()` now returns the correct value by default. Thoughts?

millerdev: > I don't know how postgres stores text fields...

If it's a "char" field then it's pre-allocated/fixed-length; a "varchar" requires 1 byte + length of string.

millerdev (approved):

snopoke:
In postures text and char fields are packed and not pre-alocated so leaving
it blank will save space (which is why it was this way to start with).

I'm not pro saving the bucket for all attachments, only ones that differ
from the default.

emord:
seems to be more in favor of not saving the bucket, so I removed those commits. Now this just reverses the logic so that the default for new forms will be the hex

corehq/form_processor/models.py
```diff
@@ -426,7 +426,7 @@ def read_content(self, stream=False):
             if not self.blob_bucket and '-' in str(self.attachment_id):
                 try:
                     # http://manage.dimagi.com/default.asp?239638
-                    blob = db.get(self.blob_id, self.blobdb_bucket(True))
+                    blob = db.get(self.blob_id, self.blobdb_bucket(remove_dashes=False))
                 except (KeyError, NotFound, BadName):
```
millerdev: To get rid of this `try`/`except` we will need a migration for attachments that were broken by the new UUID field type up until this PR gets deployed, correct?

emord: Yep. I'm going to work on that migration today. It's a little messy because to migrate sharded tables I need to write SQL functions :/

millerdev (approved):
Looks good. I'd like to see a migration and removal of the try/except as a follow-up to this.

PR #12611, opened 2016-07-26
PreFilter filters report results without configuration

kaapstorm:
A PreFilter is a report filter that is automatically applied with other ReportFilters (in the case of custom reports) or ReportAppFilters (in the case of mobile reports). The app builder is not prompted for filter configuration when the report is used in report modules.

@NoahCarnahan, @nickpell cc @czue, @sheelio

PR #13629, opened 2016-10-18
apply prefilters in mobile UCR

czue:
@kaapstorm as best i can tell this has just never worked as early as https://github.com/dimagi/commcare-hq/pull/12611/

I wish i had a test for this but it seemed like a fair amount of bootstrap work since none of the report fixture data is well tested

@dannyroberts buddy


kaapstorm:
Wow. Thanks @czue . I'm pretty sure I tested prefilters (manually) with mobile reports. Maybe I accidentally dropped the commit when I rebased. Thank you for catching and fixing.

PR #13630, opened 2016-10-19
Inferred schemas

benrudolph:
@czue @NoahCarnahan what do you guys think of this first pass on inferred schemas?

cc: @nickpell


benrudolph:
🐟

corehq/couchapps/inferred_schemas_by_case_type/views/view/map.js
```diff
@@ -0,0 +1,5 @@
+function(doc) {
```
dimagimon: Parsing error: Unexpected token (

corehq/apps/export/models/new.py
```diff
@@ -935,6 +935,7 @@ def put_item(self, path):
         item = ExportItem(
             path=path,
             label='.'.join(map(lambda node: node.name, path)),
+            last_occurrences={INFERRED_OCCURRENCE: INFERRED_OCCURRENCE}
```
benrudolph: this is somewhat ugly. open to ideas. i needed a way to express that this item didn't belong to an app / build version pair and that it was infered

NoahCarnahan: The first alternative that comes to mind is to add an `is_inferred_occurrence` flag to `ExportItem` instead. It makes things a little more verbose, but I think it is more obvious whats going on than having to know the format of the special `last_occurrences` values

If you don't do that, I think its worth adding an `is_inferred_occurrence` property to `ExportItem` as a way of sort of "documenting" what the special `last_occurrences` property means.

NoahCarnahan (commented):
Looks good! Suggested an alternative to the `last_occurrences` thing, but don't feel super strongly about it.

NoahCarnahan (approved):
👍

czue:
:+1: looks good

snopoke:
Slightly related - I noticed that usercase properties are not available in export unless they are used in the app. Would be good to add them in to the inferred schema: https://github.com/dimagi/commcare-hq/blob/08f2027123978d71af960fa5f9c175f7b4a35467/corehq/apps/callcenter/utils.py#L135-L135

benrudolph:
(@snopoke i've made a note of this)

PR #13633, opened 2016-10-19
get_template_from_string_removed

emord:
@nickpell 

https://docs.djangoproject.com/en/1.10/ref/templates/upgrading/#get-template-from-string


nickpell (approved):

emord:
killed build because this obviously has no ests

nickpell:
that makes sense, good to merge?

PR #13634, opened 2016-10-19
Chunk it

benrudolph:
@esoergel from on discussion on #hq-ops
 cc: @nickpell


benrudolph:
though reducing it 10 instead of 1 would be Ok considering it's been at 100

esoergel:
Sounds legit.

PR #13637, opened 2016-10-19
Release location restrictions

esoergel:
http://manage.dimagi.com/default.asp?237973

unpermitted-location-restriction


corehq/apps/users/views/__init__.py
```diff
@@ -458,6 +464,16 @@ def invitations(self):
 
     @property
     def page_context(self):
+        if (not self.can_restrict_access_by_location
+            and any(not role.permissions.access_all_locations
+                    for role in self.user_roles)
+        ):
```
dimagimon: E124 closing bracket does not match visual indentation

esoergel:
@sravfeyn @sheelio @kaapstorm

corehq/apps/users/views/__init__.py
```diff
@@ -458,6 +464,16 @@ def invitations(self):
...
+            messages.warning(self.request, _(
+                "This project has user roles that restrict data access by "
+                "organization, but the software plan no longer supports that. "
+                "Any users assigned to roles that are restricted in data access "
+                "by organization can no longer access this project.  Please "
+                "update the existing roles."))
```
sheelio: Hey Sravan - the checkbox should be set to "True" and disabled when you go back to the UI.  This means we're forcing the user to turn off the feature. 

Since the user is manually turning off the feature for a given role, we shouldn;t re-enable it after upgrade. 

@esoergel 
I'm finding it a little weird that checkbox=True means that the feature is disabled.  Do we want to flip that and change the wording?

esoergel: @sravfeyn I just attached a picture of what it looks like to the PR note

@sheelio I'm conflicted.  I agree that that's true if you think about it as a feature that you enable, but if you think of it as a permission, then it's weird for it to be negative.  I'd actually initially written it as a negative, but it felt weird to have a list of checkboxes enabling permissions, with one checkbox that removes permissions.  I'm happy to reconsider though, how would you want to phrase it?

esoergel:
:white_check_mark:

corehq/apps/users/views/__init__.py
```diff
@@ -422,18 +422,24 @@ def _fmt_result(domain, u):
...
             UserRole.by_domain(self.domain),
             key=lambda role: role.name if role.name else u'\uFFFF'
         ))
 
-        #  indicate if a role has assigned users, skip admin role
-        for i in range(1, len(user_roles)):
-            role = user_roles[i]
-            role.__setattr__('hasUsersAssigned',
-                             True if len(role.ids_of_assigned_users) > 0 else False)
+        for role in user_roles[1:]:
```
proteusvacuum: Nit: This is kind of a "magic number". Can you either comment why this is `[1:]` or add something in the loop to show that we are skipping the `AdminUserRole`.

corehq/apps/hqadmin/management/commands/cchq_prbac_bootstrap.py
```diff
@@ -95,6 +95,7 @@ def ensure_role(self, role, dry_run=False):
         Role(slug=privileges.ACTIVE_DATA_MANAGEMENT, name='Active Data Management', description=''),
         Role(slug=privileges.CUSTOM_REPORTS, name='Custom Reports', description=''),
         Role(slug=privileges.ROLE_BASED_ACCESS, name='Role-based Access', description=''),
+        Role(slug=privileges.RESTRICT_ACCESS_BY_LOCATION, name='Restrict Access By Location', description=''),
```
proteusvacuum: What is the description field used for here? Should we add one?

proteusvacuum (approved):

PR #13638, opened 2016-10-19
remove imports from tests/__init__.py

dannyroberts:
Thanks @czue. I found it's not just unnecessary, it also triple runs each test.


czue (approved):
:+1:

benrudolph:
> I found it's not just unnecessary, it also triple runs each test.
wow does that mean we should actively try and remove those asap?

czue:
I thought Daniel already removed them all but could be wrong

benrudolph:
ah you're right, he did

PR #13639, opened 2016-10-19
Migrate sql form attachments

emord:
@millerdev buddy @gcapalbo 

This will write `blob_bucket` attribute on any forms that aren't in the default bucket


corehq/form_processor/backends/sql/dbaccessors.py
```diff
@@ -461,6 +461,15 @@ def _get_form_ids_for_user(domain, user_id, is_deleted):
             results = fetchall_as_namedtuple(cursor)
             return [result.form_id for result in results]
 
+    @staticmethod
+    @transaction.atomic
+    def write_blob_bucket(attachment, bucket):
+        with get_cursor(XFormInstanceSQL) as cursor:
+            cursor.execute(
+                'SELECT write_blob_bucket(%s, %s::text)',
```
emord: Have to explictly cast this as `text` because cursor automatically type casts these based on fields of the model, but `bucket` is just a string being passed in

corehq/form_processor/management/commands/overwrite_blob_buckets.py
```diff
@@ -0,0 +1,51 @@
...
+        for form_id in form_ids:
+            form = FormAccessorSQL.get_form(form_id)
+            for attachment in form.get_attachments():
+                if attachment.blob_bucket:
+                    continue
+
+                bucket = attachment.blobdb_bucket(remove_dashes=False)
+                attach_id = str(attachment.attachment_id)
+                try:
+                    blob_db.get(attachment.blob_id, bucket)
```
millerdev: wonder if we should create a `blob_db.head(...)` method to check if a blob exists rather than fetching the whole blob?

corehq/form_processor/management/commands/overwrite_blob_buckets.py
```diff
@@ -0,0 +1,51 @@
...
+
+class Command(BaseCommand):
+    help = "Overwrites blob_bucket for attachments that aren't the default"
+
+    def handle(self, *args, **options):
+        possible_bad_forms = (
+            FormES()
+            .submitted(
+                gte=datetime.date(2016, 10, 14),
+                lte=datetime.datetime.now()
```
millerdev: this could have a hard end date when https://github.com/dimagi/commcare-hq/pull/13623 is deployed, correct?

millerdev (commented):
@snopoke might want to review this as well

corehq/form_processor/backends/sql/dbaccessors.py
snopoke: Surprised that you need to do this since the cursor doesn't know anything about the model here and the value being passed is just a string and not a field object or anything like that.

snopoke (changes requested):

corehq/form_processor/management/commands/overwrite_blob_buckets.py
millerdev: I'm writing a separate PR for this now.

millerdev: https://github.com/dimagi/commcare-hq/pull/13650

corehq/form_processor/interfaces/dbaccessors.py
```diff
@@ -157,6 +157,9 @@ def soft_delete_forms(self, form_ids, deletion_date=None, deletion_id=None):
     def soft_undelete_forms(self, form_ids):
         return self.db_accessor.soft_undelete_forms(self.domain, form_ids)
 
+    def write_blob_bucket(self, attachment):
```
emord: originally did it for consistency, but suppose we could totally remove this in a week https://github.com/dimagi/commcare-hq/pull/13639/commits/058e2df0938f72b32a52f4853b7cda56af3557d7

corehq/form_processor/management/commands/overwrite_blob_buckets.py
emord: was deployed last night https://github.com/dimagi/commcare-hq/pull/13639/commits/a81defc82211a9691f85237f71221165bdb9201f

corehq/form_processor/management/commands/overwrite_blob_buckets.py
emord: https://github.com/dimagi/commcare-hq/pull/13639/commits/2fa085ec0b6c93f667ea712268930b6e6d4f8ae8

corehq/form_processor/backends/sql/dbaccessors.py
```diff
@@ -461,6 +461,15 @@ def _get_form_ids_for_user(domain, user_id, is_deleted):
             results = fetchall_as_namedtuple(cursor)
             return [result.form_id for result in results]
 
+    @staticmethod
+    @transaction.atomic
+    def write_blob_bucket(attachment, bucket):
```
emord: Removed those others. Forgot about them as they weren't being used. https://github.com/dimagi/commcare-hq/pull/13639/commits/058e2df0938f72b32a52f4853b7cda56af3557d7

corehq/form_processor/backends/sql/dbaccessors.py
emord: From what I understand psycopg2/django infers the types from the fields in the django model. Because strings are not in the django model it's necessary to explicitly cast them. They have to be explicitly cast because the way we use sql functions there's not a direct mapping to a table that postgres can realize at query time

millerdev (approved):

PR #13641, opened 2016-10-19
Sanitize app, module, and form names

orangejenny:
http://manage.dimagi.com/default.asp?239736

@calellowitz

### DIFF
--- a/corehq/apps/style/static/style/ko/components/inline_edit.js
+++ b/corehq/apps/style/static/style/ko/components/inline_edit.js
@@ -79,6 +79,7 @@ hqDefine('style/ko/components/inline_edit.js', function() {
                     return;
                 }
 
+                self.value(self.value().replace(/<\/?script>/g, ''));
                 self.readOnlyValue = self.value();
                 var data = self.saveParams;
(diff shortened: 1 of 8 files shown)


calellowitz (approved):

corehq/apps/style/static/style/ko/components/inline_edit.js
```diff
@@ -79,6 +79,7 @@ hqDefine('style/ko/components/inline_edit.js', function() {
                     return;
                 }
 
+                self.value(self.value().replace(/<\/?script>/g, ''));
```
benrudolph: this isn't quite fool proof. you can still execute javascript without a script tag. fwiw ive used DOMPurify in cloudcare to guard against this: https://github.com/cure53/DOMPurify. it's not quite the same scenario, but you can customize it to not allow any html tags (which is what i think we want)

orangejenny:
@benrudolph thanks for the info, updated to use DOMPurify for the client and html.strip_tags for the server

benrudolph:
thanks for the changes 👍 on ✔️

benrudolph (approved):

PR #13643, opened 2016-10-19
Don't default to keyUp for text entry update

benrudolph:
@biyeun


benrudolph:
tests have passed locally, want to get in for hotfix

PR #13645, opened 2016-10-19
use new domain's name for creating subscription

emord:
http://manage.dimagi.com/default.asp?236652#1234995

This failed to set a subscription if the requested domain's name was taken and it was saved as a different domain

@nickpell buddy @gcapalbo


nickpell (approved):
Yikes.  Good catch.

PR #13646, opened 2016-10-19
Don't do vellum case management migration for forms unless they require a case

orangejenny:
...even if they have something in case_preload.

@emord 
cc @calellowitz

### DIFF
--- a/corehq/apps/app_manager/management/commands/migrate_app_to_cmitfb.py
+++ b/corehq/apps/app_manager/management/commands/migrate_app_to_cmitfb.py
@@ -48,7 +48,7 @@ class Command(BaseCommand):
             forms = [f for f in module.forms if f.doc_type == 'Form']
             for form in forms:
                 preload = form.actions.case_preload.preload
-                if preload:
+                if preload and form.requires == 'case':
                     xform = XForm(form.source)
                     xform.add_case_preloads(preload)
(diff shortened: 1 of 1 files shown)


emord:
Maybe we should have a log statement like "this app is in a weird state"?

orangejenny:
Are we going to take any action based on that? And we're not currently storing the output of any migrations done via the ui.

emord:
hmm. yeah I suppose we won't do anything about it

emord (approved):

orangejenny:
I'd be more in favor of clearing out case_preload if those apps are in an undesirable state...although that would be unfortunate for anyone who changed the case action, changed it back, and lost their old work...which is not a bad argument for *not* clearing case_preload on changing case action, maybe we're wrong for thinking of these apps as weird or as victims of a bug.

emord (changes requested):
Oh that's a good point. I don't think we can do this and not clear it out. 

With the way app manager currently does this if they ever changed it back I think those old properties would be written to the final build, and the user would have no way to change them. This isn't an issue if we take out the code that uses case_preload, but I don't think that's feasible at the moment, since we're in a state where only some apps are migrated

orangejenny:
updated

emord (approved):

PR #13649, opened 2016-10-20
don't 403 but process and ignore noauth submissions if in demo mode

sravfeyn:
@esoergel @kaapstorm 

Wait for tests


kaapstorm:
:+1: LGTM

PR #13651, opened 2016-10-20
always use intersection

emord:
http://manage.dimagi.com/default.asp?239542#1235567

@NoahCarnahan I think your solution is correct

buddy @gcapalbo


NoahCarnahan (approved):
👍 thanks

PR #13653, opened 2016-10-20
Async restore progress bar enhancements

benrudolph:
@proteusvacuum 
screen shot 2016-10-20 at 11 35 58 am

how about this? a lot of the notions that happen on the phone are fundamentally different in formplayer and make it very hard to replicate all the states. for example in formplayer, we attempt to do a restore on every request and do one if the restore isn't there. all this happens transparently to client so "communicating with the server" state is just the normal web request state. also at this time we have no hook into the progress that is made by formplayer in parsing the request, which i believe is "processing data from the server" state. definitely would be nice to have at some point, but right now it's less of an issue since a server is able to parse the xml much faster than a phone. will update to include when/if we get to that point

cc: @nickpell

### DIFF
--- a/corehq/apps/style/static/cloudcare/less/cloudcare/module.less
+++ b/corehq/apps/style/static/cloudcare/less/cloudcare/module.less
@@ -189,12 +189,6 @@
     font-size: 1.8rem;
   }
 }
-@media(max-width: 720px) {
-  .module-caselist-table-container {
-    width: 100%;
-    overflow-x: scroll;
-  }
-}
 
 #formplayer-progress {
   top: 0;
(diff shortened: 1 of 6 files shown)


corehq/apps/style/static/cloudcare/less/cloudcare/module.less
```diff
@@ -215,4 +209,37 @@
...
+      margin-bottom: 0px;
+    }
+    .subtext {
+      margin-top: 0px;
+    }
+    margin-bottom: 14px;
+  }
+}
+
+@media(max-width: 720px) {
```
proteusvacuum: nit, but should this be `@screen-sm-max`?

proteusvacuum (approved):
Looks great! Thanks for these changes.

corehq/apps/style/static/cloudcare/less/cloudcare/module.less
benrudolph: 👍

PR #13654, opened 2016-10-20
Fix bug loading saved reports with "last month" date filter

NoahCarnahan:
Fixes http://manage.dimagi.com/default.asp?239800

When selecting a saved report, a function on that page updates the url with the saved reports selected dates. The `getDateUrlFragment()` function was faulty, and also replicated logic on the python model unnecessarily. This PR replaces its usage with the value calculated in the model.

cc @proteusvacuum


proteusvacuum (approved):
Looks great! 👍 on ✅

PR #13657, opened 2016-10-20
superuser options -> support options

emord:
limit to support instead

buddy @gcapalbo


gcapalbo:
(build failure unrelated, merging since it's just template changes)

PR #13658, opened 2016-10-20
update views auditcare logs

emord:
Seems like two apps we should be logging? @krsdimagi 

buddy @gcapalbo

### DIFF
--- a/settings.py
+++ b/settings.py
@@ -740,6 +740,8 @@ AUDIT_MODULES = [
     'corehq.apps.userreports',
     'corehq.apps.data',
     'corehq.apps.registration',
+    'corehq.apps.hqadmin',
+    'corehq.apps.accounting',
     'tastypie',
 ]
 


settings.py
```diff
@@ -740,6 +740,8 @@
     'corehq.apps.userreports',
     'corehq.apps.data',
     'corehq.apps.registration',
+    'corehq.apps.hqadmin',
```
emord: `hqadmin` we can view user's restores

settings.py
```diff
@@ -740,6 +740,8 @@
...
+    'corehq.apps.accounting',
```
emord: `accounting` can make changes to domains subscriptions and stuff

gcapalbo (approved):

gcapalbo:
@emord were you waiting for @krsdimagi to confirm before we merge?

emord:
@gcapalbo I thought he may have some thoughts or why/if we should/shouldn't be logging these. feels fine to merge to me though?

krsdimagi:
@emord @gcapalbo hey, looks good to me.  I think I err on the side of logging more rather than less.  Good to merge.

PR #13661, opened 2016-10-20
Put some safeguards in on location restrictions

esoergel:
This links to a confluence page so the users can have some idea of what's wrong.  This also now notifies us if this happens.  I could see us turning this off, but it makes sense for us to be conservative and at least be aware of how frequently this happens.


esoergel:
@sravfeyn @kaapstorm

kaapstorm:
LGTM

mkangia:
Notification definitely worth it!! 👍

PR #13662, opened 2016-10-20
Invalidate future subscriptions on changing plan

emord:
http://manage.dimagi.com/default.asp?239772#1235014

@nickpell does this look like the right approach?

buddy @gcapalbo

### DIFF
--- a/corehq/apps/accounting/models.py
+++ b/corehq/apps/accounting/models.py
@@ -1305,16 +1305,6 @@ class Subscription(models.Model):
             **kwargs
         )
 
-        # changing a plan overrides future subscriptions
-        future_subscriptions = Subscription.objects.filter(
-            subscriber=self.subscriber,
-            date_end=None,
-        )
-        if date_end is not None:
-            future_subscriptions = future_subscriptions.filter(date_start__lt=date_end)
-        if future_subscriptions.count() > 0:
-            future_subscriptions.update(date_end=F('date_start'))
-
         new_subscription.save()
 
         new_subscription.set_billing_account_entry_point()
--- a/corehq/apps/accounting/models.py
+++ b/corehq/apps/accounting/models.py
@@ -1304,6 +1304,17 @@ class Subscription(models.Model):
             funding_source=(funding_source or FundingSource.CLIENT),
             **kwargs
         )
+
+        # changing a plan overrides future subscriptions
+        future_subscriptions = Subscription.objects.filter(
+            subscriber=self.subscriber,
+            date_end=None,
+        )
+        if date_end is not None:
+            future_subscriptions = future_subscriptions.filter(date_start__lt=date_end)
+        if future_subscriptions.count() > 0:
+            future_subscriptions.update(date_end=F('date_start'))
+
         new_subscription.save()
 
         new_subscription.set_billing_account_entry_point()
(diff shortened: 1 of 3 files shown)


corehq/apps/accounting/models.py
```diff
@@ -1304,6 +1304,17 @@ def change_plan(self, new_plan_version, date_end=None,
             funding_source=(funding_source or FundingSource.CLIENT),
             **kwargs
         )
+
+        # changing a plan overrides future subscriptions
+        future_subscriptions = Subscription.objects.filter(
+            subscriber=self.subscriber,
+            date_end=None,
```
nickpell: I think you can remove the ```date_end=None``` filter because we'd want to get rid of finite subscriptions as well.

nickpell (changes requested):
This looks good but I think we want to limit this to the self-service workflow.  For Ops, we should keep the existing date validation in place to avoid unintentionally removing future subscriptions.

Moving this code to https://github.com/dimagi/commcare-hq/blob/35aa6d01fd2451dbb4df565b2078211e9cda6c90/corehq/apps/domain/forms.py#L1372 should fix the self-service workflow while keeping the validation for the Ops workflow.

corehq/apps/domain/forms.py
```diff
@@ -1370,6 +1371,13 @@ def save(self, commit=True):
                 if not account_save_success:
                     return False
 
+                # changing a plan overrides future subscriptions
+                future_subscriptions = Subscription.objects.filter(
+                    subscriber=self.current_subscription.subscriber,
```
nickpell: this could error in the unlikely situation that ```current_subscription``` is None - might be safer / more future-proof to reference the domain's name directly

corehq/apps/domain/forms.py
```diff
@@ -1370,6 +1371,13 @@ def save(self, commit=True):
...
+                )
```
nickpell: still should filter ```date_start``` > today

corehq/apps/domain/forms.py
emord: https://github.com/dimagi/commcare-hq/pull/13662/commits/850aeb71368a5288203991c4aaf4e82584070bb4

corehq/apps/domain/forms.py
emord: https://github.com/dimagi/commcare-hq/pull/13662/commits/850aeb71368a5288203991c4aaf4e82584070bb4

corehq/apps/domain/forms.py
```diff
@@ -1370,6 +1371,14 @@ def save(self, commit=True):
                 if not account_save_success:
                     return False
 
+                # changing a plan overrides future subscriptions
+                future_subscriptions = Subscription.objects.filter(
+                    subscriber=self.domain,
+                    date_start__gte=datetime.date.today()
```
nickpell: one last change - this should be ```date_start__gt``` because otherwise we'll include the current subscription and its ```date_end``` should be set to today (handled by ```change_plan```) rather than its ```date_start```.

nickpell (changes requested):
looking good, one more change needed

nickpell (approved):

PR #13663, opened 2016-10-20
Don't close key value mapping modal on clicking backdrop

orangejenny:
http://manage.dimagi.com/default.asp?239866

@calellowitz


orangejenny:
ping

PR #13664, opened 2016-10-20
Fix location_id key errors

gcapalbo:
http://manage.dimagi.com/default.asp?239882
@sravfeyn fyi


sravfeyn:
Ugh, sorry. Thanks for fixing

PR #13665, opened 2016-10-20
Remove attachment workaround

emord:
buddy @gcapalbo 

Ran the attachment migration

I was unsure about removing the sql templates. I think because we directly reference the template in migrations, we'll always have to have the sql template hanging around?


gcapalbo:
looks ok to me, will let @millerdev sign off

re: the sql templates, yeah i guess if we wanted to remove them we'd have to make a migration that drops the stored routines, and then once we drop them everywhere we'd have to update the migrations to do a no-op instead of referencing them, and then they'd be safe to remove (but that seems like a lot of effort)

millerdev (approved):
Looks good to me. @snopoke might have a trick for the sql templates.

benrudolph:
i think what we usually do is just remove the reference of the template from migration and then delete the sql function. see: https://github.com/dimagi/commcare-hq/blob/br/async-fb/corehq/sql_proxy_accessors/migrations/0016_get_case_ids_in_domain_by_owners.py#L16 and corresponding commit: https://github.com/dimagi/commcare-hq/commit/3c3ab8c0

gcapalbo:
@benrudolph no reason other than just so we don't have unused routines hanging around

emord:
ran the migration on india, and moved to noop migrations

benrudolph (approved):
good from my end

PR #12401, opened 2016-07-11
Revert "Revert "Set vellum case management flag on app creation only""

orangejenny:
Re-implement https://github.com/dimagi/commcare-hq/pull/12381

@emord

PR #13666, opened 2016-10-20
Script to find forms broken by cmitfb migration

orangejenny:
http://manage.dimagi.com/default.asp?239756, etc.

@emord 
cc @calellowitz

### DIFF
--- a/corehq/apps/app_manager/management/commands/cmitfb_identify_broken_forms.py
+++ b/corehq/apps/app_manager/management/commands/cmitfb_identify_broken_forms.py
@@ -1,0 +1,48 @@
+import logging
+from datetime import datetime
+from django.core.management import BaseCommand
+from django.http import Http404
+
+from corehq.apps.app_manager.dbaccessors import get_app
+from corehq.apps.es.domains import DomainES
+from corehq.apps.es.apps import AppES
+
+
+logger = logging.getLogger('app_migration')
+logger.setLevel('DEBUG')
+
+
+class Command(BaseCommand):
+    help = '''
+        Identify forms that were migrated because they had preload values,
+        even though they did not require a case, which breaks the form.
+    '''
+
+    def handle(self, *args, **options):
+        # Find all apps using vellum case management
+        app_query = AppES().is_build(False).term('vellum_case_management', True).source(['domain', '_id'])
+
+        # Find all domains created after vellum case management was released,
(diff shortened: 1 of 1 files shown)


corehq/apps/app_manager/management/commands/cmitfb_identify_broken_forms.py
```diff
@@ -0,0 +1,48 @@
...
+        even though they did not require a case, which breaks the form.
+    '''
+
+    def handle(self, *args, **options):
+        # Find all apps using vellum case management
+        app_query = AppES().is_build(False).term('vellum_case_management', True).source(['domain', '_id'])
+
+        # Find all domains created after vellum case management was released,
+        # since their apps were never in the old world and therefore never migrated
+        epoch = datetime(year=2016, month=7, day=15)
```
emord: Is this the day cmitfb was release to new domains? It would be better to do the day after as it's possible that someone created a domain in the morning that was old world and ES begins dates at midnight

corehq/apps/app_manager/management/commands/cmitfb_identify_broken_forms.py
```diff
@@ -0,0 +1,48 @@
...
+        domain_query = DomainES().date_range('date_created', gt=epoch).source('name')
+        new_domains = {d['name']: 1 for d in domain_query.run().hits}
```
emord: nit: not sure why using a dictionary instead of a set

emord (commented):
Looks good depending on the answer to the date question

corehq/apps/app_manager/management/commands/cmitfb_identify_broken_forms.py
orangejenny: are set lookups constant time?

emord: yep

corehq/apps/app_manager/management/commands/cmitfb_identify_broken_forms.py
orangejenny: It's a bit afterwards. The initial release was early July, then the discovery the release was too wide, etc. I just set the date a couple days after https://github.com/dimagi/commcare-hq/pull/12401 was merged.

PR #13670, opened 2016-10-21
Added quarter filter to UCR

kkrampa:
@czue 
We need this filter in a couple of enikshay reports. Probably it's better to have this filter in enikshay custom module but I didn't find way to do it. What do you think about this solution?

quarter

### DIFF
--- a/corehq/apps/reports_core/filters.py
+++ b/corehq/apps/reports_core/filters.py
@@ -9,7 +9,7 @@ from corehq.util.dates import iso_string_to_date
 from dimagi.utils.dates import DateSpan
 from dimagi.utils.decorators.memoized import memoized
 from django.utils.translation import ugettext, ugettext_lazy as _
-
+from django.conf import settings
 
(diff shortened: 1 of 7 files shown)


corehq/apps/reports_core/filters.py
```diff
@@ -129,6 +129,52 @@ def filter_context(self):
         }
 
 
+class QuarterFilter(BaseFilter):
+    template = 'reports_core/filters/quarter_filter.html'
+
+    def __init__(self, name, label='Quarter', css_id=None):
```
czue: should this be lazy translated?

czue (changes requested):
This looks good to me. I think it's ok this to be here since I believe it would be generally useful for other projects/reports. Can you update the UCR docs to include this?

kkrampa:
@czue Updated

czue (approved):
:+1: thanks

kkrampa:
@czue :white_check_mark:

PR #13673, opened 2016-10-21
Data browser not showing in jQuery 3

orangejenny:
Breadcrumbs are no longer included on form designer, and in jQuery 3, `$('.breadcrumb-hq-section').outerHeight()` [returns undefined instead of null](https://jquery.com/upgrade-guide/3.0/#breaking-change-return-values-on-empty-sets-are-undefined), which screws up the math in vellum's `adjustToWIndow` and results in the data browser not being displayed.

@calellowitz


calellowitz (commented):
looks good once tests and merge conflicts are fixed

PR #13684, opened 2016-10-24
Set domain metadata for every phone entry that is available

gcapalbo:
@emord


emord (approved):

gcapalbo:
(approved, self-merging)

PR #13686, opened 2016-10-24
Add get_location() method to CommCareUser

NoahCarnahan:
http://manage.dimagi.com/default.asp?239995

@proteusvacuum @esoergel

### DIFF
--- a/corehq/apps/locations/views.py
+++ b/corehq/apps/locations/views.py
@@ -136,7 +136,7 @@ class LocationsListView(BaseLocationView):
         loc_restricted = self.request.project.location_restriction_for_users
         return self.can_access_all_locations and (
             not loc_restricted
-            or (loc_restricted and not self.request.couch_user.get_location(self.domain))
+            or (loc_restricted and not self.request.couch_user.get_sql_location(self.domain))
         )
 
     @property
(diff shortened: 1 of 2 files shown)


proteusvacuum:
I *think* we should just be using `get_sql_location` for all your location needs right now.

NoahCarnahan:
Ah didn't know that, thanks. This PR is to fix a usage here:
https://github.com/dimagi/commcare-hq/blob/aab7989081302e11e960c06ee418c9fe2458aa51/corehq/apps/locations/views.py#L139
@esoergel , should this usage be `user.get_sql_location()`?

Additionally, should `CommCareUser.location` be removed from that model, or renamed `_location`? or are there legitimate usages for it?

esoergel:
The usage you linked could be either one, actually (it's just checking whether or not the user has an assigned location), so I'd be down for the fix to be switching that to SQL.

To the other question, couch locations are deprecated, so we shouldn't be introducing new usages of them, but it's legitimate for old usages to still exist, or for minor changes to use couch locations in order to interoperate with existing couch-location functionality.

NoahCarnahan:
Good to know, thanks. Will update pr.

esoergel (approved):

PR #13687, opened 2016-10-24
Add ninetyninedots back to settings

proteusvacuum:
@NoahCarnahan 
I think this got removed in error when the enikshay reports were merged.


czue (approved):
oops sorry if that was my fault.

proteusvacuum:
No, I had suggested that it was maybe a good idea, I think we just didn't verify that it actually worked.

PR #13688, opened 2016-10-24
Inferred export migrate

benrudolph:
@czue @NoahCarnahan this allows us to migrate with inferred schemas. i made a mistake in my last PR and this PR corrects for it (will explain in comments)

### DIFF
--- a/corehq/apps/export/models/new.py
+++ b/corehq/apps/export/models/new.py
@@ -975,6 +975,7 @@ class InferredSchema(Document):
     group_schemas = SchemaListProperty(InferredExportGroupSchema)
     case_type = StringProperty(required=True)
     version = IntegerProperty(default=1)
+    last_app_versions = DictProperty()
 
     class Meta:
         app_label = 'export'
@@ -1061,14 +1062,15 @@ class ExportDataSchema(Document):
 
             current_schema.record_update(app.copy_of or app._id, app.version)
 
-        inferred_schema = current_schema._get_inferred_schema()
-        if inferred_schema:
-            current_schema = cls._merge_schemas(current_schema, inferred_schema)
-
         current_schema.domain = domain
         current_schema.app_id = app_id
         current_schema.version = DATA_SCHEMA_VERSION
         current_schema._set_identifier(identifier)
+
+        inferred_schema = current_schema._get_inferred_schema()
+        if inferred_schema:
+            current_schema = cls._merge_schemas(current_schema, inferred_schema)
+
         current_schema = cls._save_export_schema(
             current_schema,
             original_id,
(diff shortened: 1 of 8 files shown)


benrudolph:
cc: @nickpell

corehq/apps/export/models/new.py
```diff
@@ -559,10 +559,6 @@ def generate_instance_from_schema(cls, schema, saved_export=None):
         )
         group_schemas = schema.group_schemas
 
-        inferred_schema = instance._get_inferred_schema()
-        if inferred_schema:
-            group_schemas.extend(inferred_schema.group_schemas)
-
```
benrudolph: this was totally wrong, i needed to do this in the schema generation, not the instance generation so we can merge the schemas

corehq/apps/export/tests/test_export_conversion.py
```diff
@@ -352,37 +357,13 @@ def test_remote_app_conversion(self, _, __):
...
         self.assertIsNotNone(column)
         self.assertEqual(column.label, 'age')
         self.assertTrue(meta.is_remote_app_migration)
 
-    @mock.patch(
-        'corehq.apps.export.utils._is_remote_app_conversion',
-        return_value=True,
-    )
-    def test_remote_app_conversion_with_repeats(self, _, __):
-        instance, meta = self._convert_case_export('remote_app_repeats')
```
benrudolph: this was just wrong before. it was interpreting the saved export as a case one (even though it was supposed to be a form). it never really made any difference until we decided to handle columns we can't find differently. i.e. form export columns that couldn't be found are created as CustomUserExportColumns while case ones are added to the inferred schemas. Ah the complexity grows...

nickpell:
looks good, will let @czue or @NoahCarnahan merge

corehq/apps/export/models/new.py
```diff
@@ -975,6 +975,7 @@ class InferredSchema(Document):
     group_schemas = SchemaListProperty(InferredExportGroupSchema)
     case_type = StringProperty(required=True)
     version = IntegerProperty(default=1)
+    last_app_versions = DictProperty()
```
NoahCarnahan: Does this property get used?

benrudolph: in the merge schemas function it expects the schema interface to have this property (it will error without it). the other option is to add an if statement in the merge logic. i think it's ok to add because semantically it makes sense that this is an empty dictionary and within the merge function the if statement would basically be `if not last_app_versions: return {}` so it can continue with the merge

NoahCarnahan: ah makes sense. Might be nice to put that in a comment in the code. I imagine this being confusing to someone looking at it in the future.

benrudolph: 👍

benrudolph: https://github.com/dimagi/commcare-hq/pull/13688/commits/85cb6b38bb2a337ea29bfe014ea799d3f6ee54c9

PR #13689, opened 2016-10-24
style persistent case tile case property

orangejenny:
very minor.

before:
screen shot 2016-10-24 at 12 16 55 pm
after:
screen shot 2016-10-24 at 12 16 14 pm

@calellowitz / @proteusvacuum


proteusvacuum:
thanks.

PR #13690, opened 2016-10-24
Add ability to force convert columns in migrate_exports command

benrudolph:
@czue @NoahCarnahan allows us to force convert columns in the migrate exports command

cc: @nickpell


nickpell:
:+1: when :white_check_mark:

PR #13693, opened 2016-10-24
update Ops form to reduce confusion

nickpell:
Removes "debit" option since we can just add negative credit.
Update text to clarify what this form does.

Motivated by http://manage.dimagi.com/default.asp?237443

@benrudolph


nickpell:
✅

benrudolph:
👍

PR #13703, opened 2016-10-24
Ensure we convert daily saved exports property

benrudolph:
@NoahCarnahan @czue ...not sure how no one noticed this 😳 

cc: @nickpell


NoahCarnahan (approved):
lgtm

czue (approved):
free load off our servers!

PR #13704, opened 2016-10-24
remove care_benin custom reports

nickpell:
Domain is on a community plan and can't view the custom reports anyway.  We've confirmed with the partner that they don't need them anymore.

@benrudolph


czue (approved):
sweet! i think these were our first sql reports. :wave:

benrudolph:
nice!

nickpell:
self-merging approved PR

dannyroberts:
❤️ for UCR Custom Reports. This is the dream; we have arrived. 🌈

PR #13707, opened 2016-10-24
Directly download restore xml

proteusvacuum:
@NoahCarnahan 
Adds the ability to directly download a restore xml file (by adding `&download=true` to the url).

(FYI @phillipm this is useful for immediately saving restores to be used in the CLI)

### DIFF
--- a/corehq/apps/hqadmin/views.py
+++ b/corehq/apps/hqadmin/views.py
@@ -472,6 +472,12 @@ class AdminRestoreView(TemplateView):
             response, _ = self._get_restore_response()
             return response
 
+        download = request.GET.get('download') == 'true'
+        if download:
+            response, _ = self._get_restore_response()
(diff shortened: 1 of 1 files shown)


corehq/apps/hqadmin/views.py
```diff
@@ -472,6 +472,12 @@ def get(self, request, *args, **kwargs):
             response, _ = self._get_restore_response()
             return response
 
+        download = request.GET.get('download') == 'true'
+        if download:
+            response, _ = self._get_restore_response()
+            response['Content-Disposition'] = "attachment; filename={}-restore.xml".format(username)
```
czue: is there possibly a unicode issue here? (forget if we ever allow unicode usernames, though thought we might have at one point)

czue (approved):
no blockers

PR #13709, opened 2016-10-24
Fix SMS language dropdowns

orangejenny:
https://github.com/dimagi/langcodes/pull/8 means the SMS languages page needs to include select2.

@calellowitz / @gcapalbo


orangejenny:
ping, not having this merged is making jquery3 testing slightly more difficult

PR #13714, opened 2016-10-25
New UCR functions for monthly reports

sheelio:
@czue I added these based on our email discussion.  These are the functions to support the monthly UCR tables that Neal and I are working on.

### DIFF
--- a/custom/ucr_ext/expressions.py
+++ b/custom/ucr_ext/expressions.py
@@ -43,26 +43,6 @@ class RootPropertyNameSpec(JsonObject):
     datatype = StringProperty(required=False)
 
 
-class IterateFromOpenedDateSpec(JsonObject):
-    type = TypeProperty('ext_iterate_from_opened_date')
-
-
-class MonthStartSpec(JsonObject):
-    type = TypeProperty('ext_month_start')
-
-
-class MonthEndSpec(JsonObject):
-    type = TypeProperty('ext_month_end')
-
-
-class ParentIdSpec(JsonObject):
-    type = TypeProperty('ext_parent_id')
-
-
-class OpenInMonthSpec(JsonObject):
-    type = TypeProperty('ext_open_in_month')
-
-
 class GetCaseFormsByDateSpec(JsonObject):
     type = TypeProperty('ext_get_case_forms_by_date')
     case_id_expression = DefaultProperty(required=False)
@@ -95,7 +75,6 @@ def root_property_name(spec, context):
 
 
 def iterate_from_opened_date(spec, context):
-    IterateFromOpenedDateSpec.wrap(spec)
     spec = {
         'type': 'evaluator',
         'datatype': 'array',
@@ -173,7 +152,6 @@ def iterate_from_opened_date(spec, context):
 
 
 def month_start(spec, context):
-    MonthStartSpec.wrap(spec)
     spec = {
         'type': 'month_start_date',
         'date_expression': {
@@ -191,7 +169,6 @@ def month_start(spec, context):
 
 
 def month_end(spec, context):
-    MonthEndSpec.wrap(spec)
     spec = {
         'type': 'month_end_date',
         'date_expression': {
@@ -209,7 +186,6 @@ def month_end(spec, context):
 
 
 def parent_id(spec, context):
-    ParentIdSpec.wrap(spec)
     spec = {
         'type': 'nested',
         'argument_expression': {
@@ -245,7 +221,6 @@ def parent_id(spec, context):
 
 
 def open_in_month(spec, context):
-    OpenInMonthSpec.wrap(spec)
     spec = {
         'type': 'conditional',
         'test': {
--- a/custom/ucr_ext/expressions.py
+++ b/custom/ucr_ext/expressions.py
@@ -1,0 +1,431 @@
+from corehq.apps.userreports.expressions.factory import ExpressionFactory
+from jsonobject.base_properties import DefaultProperty
+from corehq.apps.userreports.specs import TypeProperty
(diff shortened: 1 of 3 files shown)


custom/ucr_ext/expressions.py
```diff
@@ -0,0 +1,431 @@
...
+
+    def configure(self, from_date_expression, to_date_expression):
+        self._from_date_expression = from_date_expression
+        self._to_date_expression = to_date_expression
+
+    def __call__(self, item, context=None):
+        from_date_val = transform_date(self._from_date_expression(item, context))
+        to_date_val = transform_date(self._to_date_expression(item, context))
+        if from_date_val is not None and to_date_val is not None:
+            return ((to_date_val.year - from_date_val.year)*12) + to_date_val.month - from_date_val.month
```
dimagimon: E226 missing whitespace around arithmetic operator

custom/ucr_ext/tests.py
```diff
@@ -0,0 +1,380 @@
...
+                "case_id_expression": {
+                    "type": "constant",
+                    "constant": self.test_case_id
+                }
+            }
+        })
+        self.assertEqual(7, expression({"some_field", "some_value"}, context))
+
+    def test_start_end(self):
+        context = EvaluationContext({"domain": self.domain, "start_date": "2015-03-01", "end_date": "2015-03-31"}, 0)
```
dimagimon: E501 line too long (117 > 115 characters)

sheelio:
@dannyroberts @calellowitz 
To review / merge (its for this UCR testing / investigation I'm doing with Neal).

custom/ucr_ext/tests.py
```diff
@@ -0,0 +1,382 @@
...
+from datetime import date, datetime
+from django.test import SimpleTestCase, TestCase
+from corehq.apps.userreports.expressions.factory import ExpressionFactory
+from corehq.apps.userreports.specs import EvaluationContext
+from corehq.apps.receiverwrapper.util import submit_form_locally
+from casexml.apps.case.mock import CaseStructure, CaseFactory
+from casexml.apps.case.tests.util import delete_all_cases, delete_all_xforms
+
+
+def _safe_text(input_value):
```
czue: why was this function necessary?

custom/ucr_ext/tests.py
```diff
@@ -0,0 +1,382 @@
...
+    if input_value is None:
+        return ''
+    try:
+        return str(input_value)
+    except:
```
czue: nit: should probably be `except Exception` to avoid trapping unexpected runtime things.

custom/ucr_ext/tests.py
```diff
@@ -0,0 +1,382 @@
...
+            "to_date_expression": {
+                "type": "property_name",
+                "property_name": "to",
+            },
+        })
+        self.assertEqual(0, expression({"from": "2015-06-03", "to": "2015-06-06"}))
+        self.assertEqual(1, expression({"from": "2015-06-29", "to": "2015-07-01"}))
+        self.assertEqual(12, expression({"from": "2015-06-29", "to": "2016-06-06"}))
+        self.assertEqual(11, expression({"from": "2015-06-29", "to": "2016-05-01"}))
+        self.assertEqual(15, expression({"from": "2015-06-29", "to": "2016-09-06"}))
```
czue: might be good to add a test for the case where "from" comes after "to" which would also serve to document the expected behavior

custom/ucr_ext/expressions.py
```diff
@@ -0,0 +1,431 @@
...
+        return None
+
+
+class RootPropertyNameSpec(JsonObject):
+    type = TypeProperty('ext_root_property_name')
+    property_name = StringProperty(required=True)
+    datatype = StringProperty(required=False)
+
+
+class IterateFromOpenedDateSpec(JsonObject):
```
czue: i can't tell how these classes that are just wrappers around the type are adding value. can they be removed?

custom/ucr_ext/expressions.py
```diff
@@ -0,0 +1,431 @@
...
+        },
+        'expression_if_false': {
+            'type': 'constant',
+            'constant': 'no'
+        }
+    }
+    return ExpressionFactory.from_spec(spec, context)
+
+
+def get_case_forms_by_date(spec, context):
```
czue: haha this is a monster

czue (commented):
I hadn't realized that a lot of what you were doing is just making aliases to make it convenient to generate complicated expressions. it somehow feels like there should be better framework / product support for that use case that doesn't mean adding python functions to the codebase every time we want to do this.

i don't think we need to build that now, but I am wondering if there's something we should change in the spec representation that might make it easier to apply that change some time in the future. for example, if, rather than making the `type` the mapping if the `type` was something like `"alias"` and then we keyed off a `"subtype"` property. having written that out I think it's just more complicated and not any easier to deal with moving forwards but wanted to at least pose the question.

had a couple other small comments/questions

czue:
also if we think we want these to be more widely available we should probably document them

czue:
last comment - impressive job on the tests!

sheelio:
@czue 
Some of these are just complex bits of expression that we could remove if we had named expressions that could call each other.

Others are more like functions (have inputs) which seem much more complicated to build outside of adding new expression types.

czue:
yeah i saw that distinction.

i've been thinking it would be worth allowing named expressions to call each other and was also considering something like a project-level bank of named expressions that could be used across data sources. both far out.

sheelio:
(i'm going to hold off on documenting since I'm not sure that we do want to use these widely yet)

czue (approved):
thanks for followup

PR #13715, opened 2016-10-25
Update styling to fit on one line

proteusvacuum:
@NoahCarnahan


snopoke:
nice, noticed that too during the demo

PR #13671, opened 2016-10-21
billing notification

calellowitz:
@nickpell cc: @orangejenny 
relevant spec https://docs.google.com/document/d/1WrR86HytZDuvlfsyFLdwlGxdvlht4hquwPVzlJOF1Eg/edit#

PR #13716, opened 2016-10-25
reduce confusing and duplicated code

nickpell:
get_default_plan actually returns a plan version, so let's rename the function to make it obvious.

and, let's stop recalculating the plan version from the returned plan version.

https://github.com/dimagi/commcare-hq/commit/cad9759c23b3871cef7f8e84cffa8586d2723abb is a random optimization I found

code buddy @benrudolph 

@calellowitz This affects https://github.com/dimagi/commcare-hq/pull/13671.  Sorry that every example suggests you have to do more work than you really need for getting the latest plan version.

### DIFF
--- a/corehq/apps/accounting/invoicing.py
+++ b/corehq/apps/accounting/invoicing.py
@@ -79,9 +79,7 @@ class DomainInvoiceFactory(object):
 
     @transaction.atomic
     def _ensure_full_coverage(self, subscriptions):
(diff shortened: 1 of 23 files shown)


emord:
Nice. That was definitely not obvious to me a week ago

benrudolph (approved):
wow this is a gem 💎

emord (approved):

nickpell:
self-merging approved PR

PR #13723, opened 2016-10-25
Je/ucr related stuff

emord:
pulling out some changes from here https://github.com/dimagi/commcare-hq/pull/13322

buddies @nickpell @benrudolph


corehq/apps/es/es_query.py
```diff
@@ -353,12 +353,20 @@ def pprint(self):
         """pretty prints the JSON query that will be sent to elasticsearch."""
         print self.dumps(pretty=True)
 
-    def sort(self, field, desc=False):
+    def sort(self, field, desc=False, reset_sort=True):
```
benrudolph: there're a bunch of tests for these i think and could be easy to make one for this too

benrudolph (approved):
looks good nothing blocking

PR #13725, opened 2016-10-25
fix ucr choice list 500ing

dannyroberts:
_get_report_filter can return a DynamicChoiceListFilter which isn't always picklable

http://manage.dimagi.com/default.asp?240153

### DIFF
--- a/corehq/apps/userreports/views.py
+++ b/corehq/apps/userreports/views.py
@@ -1220,7 +1220,6 @@ def data_source_status(request, domain, config_id):
     return json_response({'isBuilt': config.meta.build.finished})
 
 
-@quickcache(['domain', 'report_id', 'filter_id'], timeout=5 * 60)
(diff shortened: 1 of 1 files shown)


dannyroberts:
@proteusvacuum cc @czue

proteusvacuum:
still curious about what exactly was causing this, but looks good.

czue:
seems fine to me. @proteusvacuum was there a reason you originally added caching there and any concerns with it going away (assuming that was you)?

proteusvacuum:
Yeah, I was calling that method twice at some point, but then refactored so that it was only needed once  ([here](https://github.com/dimagi/commcare-hq/commit/18f1656d55c5f31c3416bf90ea58f4569787f369)), and forgot to remove the quickcache. The caching is no longer necessary, so no concerns about it being removed.

proteusvacuum (approved):

PR #13729, opened 2016-10-26
add get_descendants test for all cases; refetch parent before saving

sravfeyn:
http://manage.dimagi.com/default.asp?240047#1237655 @esoergel

### DIFF
--- a/corehq/apps/locations/bulk_management.py
+++ b/corehq/apps/locations/bulk_management.py
@@ -809,12 +809,21 @@ def save_locations(location_stubs, types_by_code):
 
         for child_stub in child_stubs:
             child = child_stub.db_object
-            child.parent = parent_location
             if child_stub.do_delete:
                 # keep track of to be deleted items to delete them in top-to-bottom order
                 to_be_deleted.append(child)
             elif child_stub.needs_save:
                 child.location_type = types_by_code.get(child_stub.location_type)
+                if parent_location:
+                    # refetch parent_location object so that mptt related fields are updated consistently,
+                    #   since we are saving top to bottom, parent_location would not have any pending
+                    #   saves, so this is the right point to refetch the object.
+                    child.parent = SQLLocation.objects.get(
+                        domain=parent_location.domain,
+                        site_code=parent_code
+                    )
+                else:
+                    child.parent = None
                 child.save()
             update_children(child_stub)
 
--- a/corehq/apps/locations/tests/test_bulk_management.py
+++ b/corehq/apps/locations/tests/test_bulk_management.py
@@ -464,6 +464,42 @@ class TestBulkManagement(TestCase):
             actual.append((attr, parent))
 
         self.assertEqual(set(actual), expected_locations)
+        self.assertMpttDescendants(expected_locations)
+
+    def assertMpttDescendants(self, pairs):
+        # Given list of (child, parent), check that for each location
+        # SQLLocation.get_descendants is same as calculated descendants
+
+        from collections import defaultdict
+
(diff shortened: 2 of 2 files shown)


corehq/apps/locations/bulk_management.py
```diff
@@ -809,12 +809,21 @@ def update_children(parent_stub):
...
         for child_stub in child_stubs:
             child = child_stub.db_object
-            child.parent = parent_location
             if child_stub.do_delete:
                 # keep track of to be deleted items to delete them in top-to-bottom order
                 to_be_deleted.append(child)
             elif child_stub.needs_save:
                 child.location_type = types_by_code.get(child_stub.location_type)
+                if parent_location:
+                    # refetch parent_location object so that mptt related fields are updated consistently,
```
sravfeyn: Didn't use any of mptt methods (`delay_mptt_updates`, `rebuild` and `partial_rebuild`) because it gets challenging to sync couch changes as these depend on mptt data (which won't be update since we delay mptt). We would need to redo couch changes separately again.

esoergel: Do you mean `lineage`?  Aside from that, I wouldn't expect `rebuild` to affect any of the fields we sync between the two.

corehq/apps/locations/bulk_management.py
```diff
@@ -809,12 +809,21 @@ def update_children(parent_stub):
...
+                    #   since we are saving top to bottom, parent_location would not have any pending
+                    #   saves, so this is the right point to refetch the object.
+                    child.parent = SQLLocation.objects.get(
+                        domain=parent_location.domain,
+                        site_code=parent_code
+                    )
```
esoergel: This has the potential to complicate things significantly, but it may be worth caching these lookups, since that'll be a lot of db hits.  Another approach could be to save these breadth-first, that is, level-by-level, then do a bulk get for locations at each level before moving on to the next level.

corehq/apps/locations/tests/test_bulk_management.py
```diff
@@ -837,3 +873,30 @@ def test_rearrange_locations(self):
...
+
+        result = self.bulk_update_locations(
+            FLAT_LOCATION_TYPES,
+            swap_parents,
+        )
+
+        self.assertEqual(result.errors, [])
+        self.assertLocationTypesMatch(FLAT_LOCATION_TYPES)
+        self.assertLocationsMatch(self.as_pairs(swap_parents))
+        self.assertCouchSync()
```
sravfeyn: Yeah of course. It was a TDD

corehq/apps/locations/bulk_management.py
sravfeyn: We are saving breadth-first. Even then, we have to refetch objects in upper levels, because when lower level objects are saved, mptt might update mptt-fields of upper level objects. This is also the reason why caching won't work, as the prefetched upper objects won't get the mptt updates that are necessary to keep mptt consistent.

The only efficient solution that I found out so far will need following two to happen
1. The mptt PR that supports bulk-save needs to be merged and we would need to upgrade to that mptt
2. Remove couch syncing. Otherwise we would need to write a bulk couch sync (I have also tried this in the beginning, but thought it's not worth)

esoergel: > when lower level objects are saved, mptt might update mptt-fields of upper level objects.

Ugh.  Those justifications sound good to me, I agree with your conclusions.  Good job on this PR, btw, this is a very satisfactory resolution to that issue!

PR #13737, opened 2016-10-26
lowering cache timeout for resource not found

calellowitz:
@snopoke @millerdev cc: @dannyroberts @orangejenny 
http://manage.dimagi.com/default.asp?240232#1238172
This cuts cache timeout when an attachment cant be found, under the assumption that resource not found errors are likely the blobdb acting up and we should keep looking if they want to try again, but keeps timeouts when the resource is found at 1 day.

### DIFF
--- a/corehq/apps/app_manager/models.py
+++ b/corehq/apps/app_manager/models.py
@@ -4068,8 +4068,8 @@ class LazyBlobDoc(BlobMixin):
     def __attachment_cache_key(self, name):
         return u'lazy_attachment/{id}/{name}'.format(id=self.get_id, name=name)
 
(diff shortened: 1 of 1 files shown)


corehq/apps/app_manager/models.py
```diff
@@ -4116,8 +4116,9 @@ def lazy_fetch_attachment(self, name):
                     if hasattr(e, 'response'):
                         del e.response
                     content = e
+                    self.__set_cached_attachment(name, content, timeout=60*5)
```
dimagimon: E226 missing whitespace around arithmetic operator

czue:
nice this is a good idea

czue (approved):

PR #13739, opened 2016-10-26
replacing couch lookup with es

calellowitz:
@benrudolph @dannyroberts 
http://manage.dimagi.com/default.asp?239907#1238223

### DIFF
--- a/corehq/apps/export/esaccessors.py
+++ b/corehq/apps/export/esaccessors.py
@@ -40,3 +40,8 @@ def get_ledger_section_entry_combinations(domain):
     ]
     query = LedgerES().domain(domain)
     return NestedTermAggregationsHelper(base_query=query, terms=terms).get_data()
+
+def get_case_name(case_id):
+    return (CaseES()
+            .case_ids([case_id])
+            .values('name'))
--- a/corehq/apps/export/esaccessors.py
+++ b/corehq/apps/export/esaccessors.py
@@ -41,6 +41,7 @@ def get_ledger_section_entry_combinations(domain):
     query = LedgerES().domain(domain)
     return NestedTermAggregationsHelper(base_query=query, terms=terms).get_data()
 
+
 def get_case_name(case_id):
     return (CaseES()
             .case_ids([case_id])
--- a/corehq/apps/export/transforms.py
+++ b/corehq/apps/export/transforms.py
@@ -1,6 +1,6 @@
 from django.core.cache import cache
 
-from corehq.apps.es import CaseES
+from corehq.apps.export.esaccessors import get_case_name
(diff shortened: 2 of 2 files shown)


corehq/apps/export/transforms.py
```diff
@@ -27,10 +27,7 @@ def _cached_case_id_to_case_name(case_id):
     ret = cache.get(key, NULL_CACHE_VALUE)
     if ret != NULL_CACHE_VALUE:
         return ret
-    try:
-        case = CommCareCase.get_lite(case_id)
-        ret = case['name'] if "name" in case else None
-    except ResourceNotFound:
-        ret = None
+    case = CaseES().case_ids([case_id]).values('name')
```
benrudolph: nit: this should probably go in an `esaccessor`

calellowitz:
@benrudolph fixed

corehq/apps/export/esaccessors.py
```diff
@@ -40,3 +40,8 @@ def get_ledger_section_entry_combinations(domain):
     ]
     query = LedgerES().domain(domain)
     return NestedTermAggregationsHelper(base_query=query, terms=terms).get_data()
+
+def get_case_name(case_id):
```
dimagimon: E302 expected 2 blank lines, found 1

corehq/apps/export/esaccessors.py
```diff
@@ -40,3 +40,8 @@ def get_ledger_section_entry_combinations(domain):
...
+    return (CaseES()
+            .case_ids([case_id])
+            .values('name'))
```
dimagimon: W292 no newline at end of file

corehq/apps/export/esaccessors.py
```diff
@@ -40,3 +40,9 @@ def get_ledger_section_entry_combinations(domain):
     ]
     query = LedgerES().domain(domain)
     return NestedTermAggregationsHelper(base_query=query, terms=terms).get_data()
+
+
+def get_case_name(case_id):
```
dannyroberts: another nit: I have an unwritten rule that anytime I look up a doc by its id, I also make sure it's in the domain I expect. Being strict about that makes you not have to rely on reasoning about the millions of ways (malicious or innocent) a case id might end up referenced in a form in the wrong domain, for example.

That said, I see this isn't something you added, so I'm happy to leave for another PR. But if it's easy, it'd be tighter if `_cached_case_id_to_case_name` required a `domain` argument and passed it in to this, and the search was limited by that domain.

dannyroberts: offlined, will prioritize merging this fix and let @calellowitz do this off-pr

PR #13742, opened 2016-10-26
Kill celery stale workers

benrudolph:
@gcapalbo i spent some time digging into the stale worker issue on swiss and staging. some of the workers on swiss just didn't respond to SIGTERM for some reason. i couldn't find any trace of the signal in strace either. eventually gave up and decided that we should just have a job to take care of it. this is a management command to kill the stale workers. if it works out alright, i think it would then make sense to move it to a cron task. what do you think?

cc: @emord @nickpell

### DIFF
--- a/corehq/apps/hqadmin/management/commands/kill_stale_celery_workers.py
+++ b/corehq/apps/hqadmin/management/commands/kill_stale_celery_workers.py
@@ -1,0 +1,33 @@
+import json
+
+from django.core.management.base import BaseCommand
+from django.conf import settings
+from celery import Celery
+from restkit import Resource
+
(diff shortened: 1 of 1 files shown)


benrudolph:
this hasn't been run yet on a live production machine, but it's safe to merge since it's a mgmt command. i'm just going to run it on staging until i've confirmed it does what i expect

gcapalbo:
this will just initiate a warm shutdown for the identified workers right?

if so, i think it would be better if we just incorporated this into the process as our way to shutdown the workers exclusively (instead of propagating the TERM signal) going forwards so that we don't have two ways of shutting down the workers.

PR #13745, opened 2016-10-27
Enikshay changes

kkrampa:
@czue First commit only removes unused fields from episode case and reports, all additions are in second commit.


custom/enikshay/ucr/tests/test_episode_datasource.py
```diff
@@ -137,58 +137,3 @@ def test_hiv_status(self):
         row = query[0]
 
         self.assertEqual(row.hiv_status, 'reactive')
-
-    @run_with_all_ucr_backends
```
kkrampa: I deleted fields used in these tests.

custom/enikshay/reports/filters.py
```diff
@@ -36,3 +42,58 @@ def filter_context(self):
         context = super(EnikshayLocationFilter, self).filter_context
         context['endpoint'] = self.pagination_source
         return context
+
+
+class QuarterFilter(BaseReportFilter):
```
czue: this is needed (in addition to [this one](https://github.com/dimagi/commcare-hq/blob/master/corehq/apps/reports_core/filters.py#L132)) because the report is rendered outside the UCR framework? I think it would be good to share more code between these two classes.

czue (dismissed):
Interested to get your take on code reuse between the two `QuarterFilter` classes and templates. let me know if you want some more guidance on that and I can try to provide some. At a minimum I think they could either call a bunch of shared functions, or maybe inherit from the same `MixIn` for some stuff?

custom/enikshay/templates/enikshay/partials/report_content.html
```diff
@@ -0,0 +1,63 @@
+{% load i18n %}
```
kkrampa: Currently, I don't have better idea than copy-paste those part of code from corehq

kkrampa:
@czue Updated. It seems build needs to be restarted

custom/enikshay/templates/enikshay/partials/report_content.html
czue: this is because of https://github.com/dimagi/commcare-hq/pull/13748 right? @calellowitz do you have any other ideas?

if this is the only solution that's dissatisfying but I think we can just move forwards with it. we should probably at least put a comment indicating the situation and reason and linking to the PR

czue (commented):
thanks for follow up and looks good - just want to see if @calellowitz has any ideas on the template

kkrampa:
@czue Any action from my side? Can we move this forward?

czue:
@kkrampa sorry about the delay

PR #13746, opened 2016-10-27
UCR distinct values in adapter

emord:
Pulling out  some more of the ES UCR stuff. Slightly modified so that I could remove some of the copy-paste.

I waffled over whether or not distinct values should go into the adapter, or into a separate db helper class, but decided adapter was good enough. @czue do you have any thoughts on that decision?

buddies @nickpell @benrudolph

### DIFF
--- a/corehq/apps/userreports/sql/__init__.py
+++ b/corehq/apps/userreports/sql/__init__.py
@@ -1,3 +1,3 @@
 from .adapter import get_indicator_table, IndicatorSqlAdapter, metadata, ErrorRaisingIndicatorSqlAdapter
-from .columns import get_expanded_column_config, ColumnConfig
+from .columns import get_expanded_column_config
 from .util import get_column_name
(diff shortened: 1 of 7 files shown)


corehq/apps/userreports/sql/__init__.py
```diff
@@ -1,3 +1,3 @@
 from .adapter import get_indicator_table, IndicatorSqlAdapter, metadata, ErrorRaisingIndicatorSqlAdapter
-from .columns import get_expanded_column_config, ColumnConfig
+from .columns import get_expanded_column_config
```
dimagimon: F401 'get_expanded_column_config' imported but unused

czue:
> I waffled over whether or not distinct values should go into the adapter, or into a separate db helper class, but decided adapter was good enough. @czue do you have any thoughts on that decision?

Agree this is good enough. Also doesn't immediately look like it would be too hard to change this decision later if we cared to.

czue:
looks good! thanks.

PR #12872, opened 2016-08-18
Add column info to UCR data API

NoahCarnahan:
Addresses [#234901](http://manage.dimagi.com/default.asp?234901).

Adds the `"columns"` property to the object returned by the `ConfigurableReportDataResource` api endpoint. e.g.

``` json
{
  "columns": [
    {
      "header": "Name",
      "slug": "column_agg"
    },
    {
      "expand_column_value": "VT",
      "header": "state-VT",
      "slug": "column_0-0"
    },
    {
      "expand_column_value": "MA",
      "header": "state-MA",
      "slug": "column_0-1"
    }
  ],
  "data": [
    {
      "column_0-0": 1,
      "column_0-1": 1,
      "column_agg": "Noah"
    },
    ...
  ],
  ...
}
```

@snopoke @gcapalbo  Best reviewed commit-by-commit.

PR #13747, opened 2016-10-27
Allow expanded columns to have transforms

emord:
@NoahCarnahan found this while refactoring. Comes from https://github.com/dimagi/commcare-hq/pull/12872 I don't think that this was intentional, but it doesn't allow expanded columns to have transforms.

buddies @nickpell @benrudolph


NoahCarnahan:
👍

PR #13748, opened 2016-10-27
Revert "Moved report content to separate template"

calellowitz:
@dannyroberts
http://manage.dimagi.com/default.asp?240301#1238708
The mvp chw report (and likely others) rely on being able to override parts of the reportcontent block which is impossible when that block is included as a separate template.


czue:
does `addtoblock` help with this problem at all?

@calellowitz also are there any implications of this change on TB reports? saw was introduced in https://github.com/dimagi/commcare-hq/pull/13446 and wasn't clear whether it was a required part of that change. cc @kkrampa @proteusvacuum

kkrampa:
It will break Enikshay reports. I will fix this tomorrow.

czue:
no one is using those reports yet so prioritizing this. @dannyroberts @calellowitz @kkrampa I'm still curious if we can use `addtoblock` to fix this (if there's not a better answer).

calellowitz:
@czue `addtoblock` does not play nicely with `include`. In particular blocks appended in that way cannot span from an included template to the parent that included it, so it doesnt seem like a great solution for either this mvp report or the tb ones as written. Its possible that after this pr the tb multireport could use addtoblock instead of the include at https://github.com/dimagi/commcare-hq/pull/13446/files#diff-41b47b752bffc5a8889e799454995de2R11 but there may still need to be code replication because I am not sure you can rerender the same inherited block over and over.

czue:
interesting - thanks for the explanation

PR #13750, opened 2016-10-27
Prefactor submit history report

esoergel:
@kaapstorm @sravfeyn 
I'm doing some work on this report, so here's a little prefactor work.  This should have no functional changes, it just converts the report to use ESQuery and cleans it up a bit.

Review :blowfish: by :tropical_fish: 
Comments on style are welcome.

This is currently on staging.

### DIFF
--- a/corehq/apps/data_interfaces/interfaces.py
+++ b/corehq/apps/data_interfaces/interfaces.py
@@ -17,7 +17,6 @@ from corehq.apps.reports.standard.cases.basic import CaseListMixin
 from corehq.apps.reports.standard.cases.data_sources import CaseDisplay
 from corehq.apps.reports.standard.inspect import SubmitHistoryMixin
 from corehq.apps.reports.filters.base import BaseSingleOptionFilter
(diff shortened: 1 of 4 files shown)


corehq/apps/data_interfaces/interfaces.py
```diff
@@ -183,12 +180,17 @@ def template_context(self):
...
 
-    def _es_xform_filter(self):
-        return self.mode.xform_filter
+    @property
+    def es_query(self):
+        query = super(BulkFormManagementInterface, self).es_query
+        if self.mode.mode_name == self.mode.RESTORE_MODE:
+            return query.only_archived()
+        else:
+            return query
```
sravfeyn: Doesn't this include archived forms?

esoergel: No, archived forms are excluded by default from all FormES queries, since they have a different doctype.  The `only_archived()` method removes that default filter and adds a filter to exclude non-archived forms.

PR #13752, opened 2016-10-27
restricting case owner from export to domain

calellowitz:
@dannyroberts


dannyroberts:
Looks good, can you merge master to fix the conflict?

PR #13754, opened 2016-10-27
remove dot notation for referring to view_app

nickpell:
@benrudolph @emord

### DIFF
--- a/corehq/apps/app_manager/views/view_generic.py
+++ b/corehq/apps/app_manager/views/view_generic.py
@@ -85,7 +85,7 @@ def view_generic(request, domain, app_id=None, module_id=None, form_id=None,
     if app and app.copy_of:
         # don't fail hard.
         return HttpResponseRedirect(reverse(
-            "corehq.apps.app_manager.views.view_app", args=[domain, app.copy_of]
+            "view_app", args=[domain, app.copy_of] # TODO - is this right?
         ))
 
     # grandfather in people who set commcare sense earlier
--- a/corehq/tabs/tabclasses.py
+++ b/corehq/tabs/tabclasses.py
@@ -596,7 +596,7 @@ class ProjectDataTab(UITab):
 
 
 class ApplicationsTab(UITab):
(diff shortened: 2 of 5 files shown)


corehq/apps/app_manager/views/view_generic.py
```diff
@@ -85,7 +85,7 @@ def view_generic(request, domain, app_id=None, module_id=None, form_id=None,
     if app and app.copy_of:
         # don't fail hard.
         return HttpResponseRedirect(reverse(
-            "corehq.apps.app_manager.views.view_app", args=[domain, app.copy_of]
+            "view_app", args=[domain, app.copy_of] # TODO - is this right?
```
dimagimon: E261 at least two spaces before inline comment

corehq/tabs/tabclasses.py
```diff
@@ -596,7 +596,7 @@ def dropdown_items(self):
 
 
 class ApplicationsTab(UITab):
-    view = "corehq.apps.app_manager.views.view_app"
+    view = "default_app"
```
emord: Why are a couple of these `default_app` instead of `view_app` ?

corehq/apps/app_manager/views/view_generic.py
nickpell: @dannyroberts I wasn't sure which url should be returned here - there are many different ones that are all handled by ```corehq.apps.app_manager.views.view_app```.  Do you have an idea?

corehq/tabs/tabclasses.py
nickpell: If you check out ```app_manager/urls.py``` there are many different urls that all point to the same view function ```corehq.apps.app_manager.views.view_app```.  ```default_app``` refers to ```/a/<domain>/apps/```.  This seems like an anti-pattern, but fixing that is out of scope for this set of changes.

emord (approved):
Looks good to me. Will let you decide on merging based on danny's response

PR #13755, opened 2016-10-28
return None if case_id empty

snopoke:
http://manage.dimagi.com/default.asp?240377
@calellowitz 
@dannyroberts 

Was resulting in this ES query:

```
{
  "query": {
    "filtered": {
      "filter": {
        "and": [
          {
            "terms": {
              "_id": [
                null
              ]
            }
          },
          {
            "match_all": {}
          }
        ]
      },
      "query": {
        "match_all": {}
      }
    }
  },
  "_source": [
    "name"
  ],
  "size": 1000000
}   
```


benrudolph:
@calellowitz any chance we can write a test for this case so we know it won't happen again?

PR #13758, opened 2016-10-28
fix middleware

benrudolph:
@wpride


wpride:
Tests pass locally and this is broken until merged, merging

benrudolph:
JS tests passed locally. will followup with some tests

PR #13761, opened 2016-10-28
Question comments in form summary

proteusvacuum:
Adds question comments to the form summary (both downloadable and on the web). 
screenshot from 2016-10-28 14-25-37

@NoahCarnahan 
FYI @emord (and @phillipm I think you wanted this at some point)

### DIFF
--- a/corehq/apps/app_manager/templates/app_manager/ng_partials/form_summary_view.html
+++ b/corehq/apps/app_manager/templates/app_manager/ng_partials/form_summary_view.html
@@ -127,6 +127,7 @@
                                 <span ng-show="showLabels">{{ getQuestionLabel(question) }}</span>
                                 {% endverbatim %}
                                 <span ng-hide="showLabels">{{ question.value }}</span>
+                                <span class="text-muted" ng-show="showComments"> &nbsp; {{ question.comment }}</span>
                                 <span ng-show="question.required" title="{% trans "This question is required" %}">*</span>
                                 <ol>
                                     <li ng-repeat="option in question.options">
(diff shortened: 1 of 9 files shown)


phillipm:
Dope! Thanks Farid!

emord (approved):

proteusvacuum:
@dannyroberts I was a bit confused by the `test_readable_formdata` tests... I ran the command that was mentioned in the comment in the `_test_corpus` method, but I'm not sure what the output really means, and those huge `yaml` and `json` files were really hard for me to verify if they were creating the correct output or not. It also seems like the output of that command changed more than I would have expected. Do you have any insight into this?

dannyroberts:
Ick. I think they changed so much largely because of arbitrary formatting considerations. (Like in the new format, it looks like properties are sorted alphabetically, but not in the old?) One thing that could make it much clearer to you what actually changed in this PR is to break https://github.com/dimagi/commcare-hq/pull/13761/commits/c8d3bcd0a985730409f2cb876919d88ef84a98b2 into two commits: one in which you just rerun the formatting command on the pre-PR code, and another where you run it again on the new code. The second diff I think would then be likely to contain only what actually changed, and not include so many simple formatting diffs.

dannyroberts:
My larger perspective on this kind of test is that it has plusses and minuses. An obvious minus is it's hard to look at the expected-output file and have any intuition as to whether it's correct. The plus (if we can work out less-volatile diffs) is that you can see what changed, which may be easier to validate is correct. With high-noise diffs, this main advantage gets wiped out, and the test no longer really proves or assures anything.

proteusvacuum:
@dannyroberts splitting up the commits was a great idea, thanks.

proteusvacuum:
✅

NoahCarnahan:
👍

PR #13762, opened 2016-10-28
fix fixture upload api by dumping to tempfile first

dannyroberts:
passing the django request file to get_workbook causes it to get "used up"
so the second time you do it it's raises an exception

http://manage.dimagi.com/default.asp?240371


dannyroberts:
@calellowitz cc @czue

calellowitz (approved):

PR #13764, opened 2016-10-28
Handle other exceptions with getting the payload

proteusvacuum:
I was seeing a bunch of issues when `get_payload` was throwing an exception, and the record was never being marked as failed. 
@NoahCarnahan


NoahCarnahan (approved):
looks good to me!

PR #13765, opened 2016-10-28
Signal to update inferred schema

benrudolph:
@NoahCarnahan @czue this adds a signal for the inferred schema to be updated. it also refactors the inferred schema generation a little and adds more database constraints on what can be saved to the schema. 🐟 or 🏠 

These properties were "inferred" from a user case save. kind of cool:
screen shot 2016-10-28 at 3 28 37 pm

cc: @emord @nickpell

### DIFF
--- a/corehq/apps/callcenter/utils.py
+++ b/corehq/apps/callcenter/utils.py
@@ -11,6 +11,7 @@ from corehq.apps.es.domains import DomainES
 from corehq.apps.es import filters
 from corehq.apps.hqcase.utils import submit_case_blocks
 from corehq.apps.locations.models import SQLLocation
+from corehq.apps.export.signals import added_inferred_export_properties
 from corehq.form_processor.interfaces.dbaccessors import CaseAccessors
 from corehq.util.quickcache import quickcache
 from corehq.util.timezones.conversions import UserTime, ServerTime
@@ -153,6 +154,13 @@ def _get_user_case_fields(commcare_user):
         'language': commcare_user.language or '',
         'phone_number': commcare_user.phone_number or ''
     })
+
+    added_inferred_export_properties.send(
+        'UserSave',
+        domain=commcare_user.domain,
+        case_type=USERCASE_TYPE,
+        properties=fields.keys()
+    )
     return fields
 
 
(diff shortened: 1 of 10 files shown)


corehq/apps/callcenter/utils.py
```diff
@@ -153,6 +154,13 @@ def valid_element_name(name):
         'language': commcare_user.language or '',
         'phone_number': commcare_user.phone_number or ''
     })
+
+    added_inferred_export_properties.send(
+        'UserSave',
+        domain=commcare_user.domain,
+        case_type=USERCASE_TYPE,
+        properties=fields.keys()
+    )
```
snopoke: 👍 thanks

snopoke (commented):
Looks good

NoahCarnahan (approved):
lgtm

corehq/apps/importer/tasks.py
```diff
@@ -78,6 +80,14 @@ def _submit_caseblocks(caseblocks):
...
                     row_number=caseblocks[0]._id
                 )
+            else:
+                properties = set().union(*map(lambda c: set(c.dynamic_case_properties().keys()), cases))
+                added_inferred_export_properties.send(
+                    'CaseImporter',
+                    domain=domain,
+                    case_type=case_type,
+                    properties=properties,
+                )
```
benrudolph: i don't think we do. the merge will handle the duplicates and i figured it'd add unnecessary complexity to try and filter out the ones that are already in the schema

benrudolph:
merging on approval. right now it's a django signal, but we can easily change to an async task if we think it's too much overhead

czue:
i'm way behind on code review, but what's with the duplicate `one` and `two` properties in the screenshot? 

assume just some weird thing you were testing but wanted to ask in case there is a bug where we don't properly resolve duplicate properties or something

benrudolph:
gah good eyes... this hasn't been deployed yet, so going to revert and figure out a solution. regular case properties get marked as `ScalarItem` which makes them not resolve with `ExportItem`.

benrudolph:
suppose tests only help when you write them for the real scenario

PR #13766, opened 2016-10-28
Simplify fixture upload interfaces and error handling

dannyroberts:
A lot of code moved around so commit-by-commit (🐡) is easiest

- make API and UI endpoints as similar as possible in the code they call and push all error handling in so it's shared
- reduce and simplify resulting code
- split up files within `corehq/apps/fixtures/upload/` and arrange more logically, mark private functions and classes


corehq/apps/users/views/mobile/users.py
```diff
@@ -902,7 +900,7 @@ def post(self, request, *args, **kwargs):
         upload = request.FILES.get('bulk_upload_file')
         try:
             self.workbook = WorkbookJSONReader(upload)
-        except (InvalidFileException, BadZipfile):
+        except InvalidExcelFileException:
```
dannyroberts: (fixed)

dannyroberts:
This test passes locally as well as when I run `./scripts/docker test corehq/apps/fixtures/tests/test_upload.py` locally, as well as on Cal's computer. Weird. Still looking into it.

dannyroberts:
Making it a zip file fixed it. Still very unclear why it was failing only on travis (and not on other linux setups like docker), but I'm certain enough that it's not indicative of a major bug that I don't think it needs to block.

dannyroberts:
✅

corehq/apps/fixtures/templates/fixtures/partials/fixture_upload_status_api.txt
```diff
@@ -0,0 +1,21 @@
...
+{% for msg in result.messages %}
+{{ msg }}
+{% endfor %}
+{% endif %}
+{% else %}
+{% trans "Upload failed! Details:" %}
+{% for e in result.errors %}
+{{ e }}
+{% endfor %}
+{% endif %}
```
esoergel: Some indenting would make this file easier to read.

esoergel (approved):
looks good!

dannyroberts:
by the way I forgot to highlight this—when we get an unexpected fixture upload error, it 500s and sends us a 500 email. The user experience is the same (it's an AJAX request), we just let it bubble up to our error reporting mechanism instead of swallowing it and acting like it's fine and failing silently. (I'd recently added a soft assert, but it wasn't sending useful info.)

PR #13768, opened 2016-10-31
refresh cachebuster to force reload toggles.js

dannyroberts:
Immediate fix for http://manage.dimagi.com/default.asp?240172

### DIFF
--- a/corehq/apps/hqwebapp/templatetags/hq_shared_tags.py
+++ b/corehq/apps/hqwebapp/templatetags/hq_shared_tags.py
@@ -235,7 +235,7 @@ def toggle_js_url(domain, username):
 def toggle_js_domain_cachebuster(domain):
     # to get fresh cachebusters on the next deploy
     # change the date below (output from *nix `date` command)
-    #   Thu Mar  3 16:21:30 EST 2016
+    #   Mon Oct 31 10:30:09 EDT 2016
     return random_hex()[:3]
 
 


dannyroberts:
@orangejenny no need to wait for tests. cc @czue

snopoke:
I did already try this here: https://github.com/dimagi/commcare-hq/compare/2016-10-25_19.33-production-deploy...sk/hotfix-cachebuster-2016-10-26 but it didn't seem to work

dannyroberts:
@snopoke, oh I assumed you'd deployed that temporarily and then it started happening again after the next deploy (since it wasn't on master). I can check that `quickcache` behaves the way I think it does.

dannyroberts:
Yeah, `quickcache` works the way I think it does, revaluating (and in `toggle_js_domain_cachebuster`'s case returning a new value) when even a comment within the function changes. That should make `<foo>` in `toggles.js?cachebuster=<foo>` change, which should make your browser treat it as a new resource—this is the same technique we use for versioning/cachebusting static files. Unless there's something deeper I missed in your comment on the ticket, the caching headers you referenced are intentional—unlike most dynamically generated content, we *do* want the browser to cache this and then control cachebusting through the url param. (In terms of client-side performance, having the file cached is pretty crucial.)

snopoke:
Makes sense though I definitely still experienced the issue after the hotfix was deployed and Kishan was still having to tell users to hard refresh. I've recorded what my current cachebusting params are for a few domains so I'll test again once this has been deployed to make sure they've changed.

dannyroberts:
Weird, yeah that's a good call. I'll do the same.

snopoke:
Comparing URL's from different domains (but logged with the same user) I'd expect that user cachebuster suffix would be the same but it's not, how come?

dannyroberts:
On two different domains, the urls I get are

```
https://www.commcarehq.org/a/droberts/toggles.js?username=droberts@dimagi.com&cachebuster=72f-c0b
https://www.commcarehq.org/a/droberts-1/toggles.js?username=droberts@dimagi.com&cachebuster=96f-c0b
```

You'll see the `c0b` is the same because it's the same user, but the `72f`/`96f` changes because it's a different domain.  That's all to be expected (since toggle.js has to be cached by user+domain). Are your cachebusters not sharing the second 3 letters?

dannyroberts:
My urls have been pretty consistently what I thought they'd be, and I specifically checked that first time after deploy at least one of them wasn't hitting the cache, so the change in cachebuster seemed to work in that case.

Oct. 31

pre-deploy:

```
https://www.commcarehq.org/a/droberts/toggles.js?username=droberts@dimagi.com&cachebuster=72f-c0b
https://www.commcarehq.org/a/droberts-1/toggles.js?username=droberts@dimagi.com&cachebuster=96f-c0b
```

post-deploy:

```
https://www.commcarehq.org/a/droberts/toggles.js?username=droberts@dimagi.com&cachebuster=b94-c0b
https://www.commcarehq.org/a/droberts-1/toggles.js?username=droberts@dimagi.com&cachebuster=af1-c0b
```

snopoke:
:+1: all good

PR #13769, opened 2016-10-31
A fix for preview app and some tests

benrudolph:
@wpride after removing your "auth as" in preview app, it would try and navigate you to `single_app/undefined` since an app id was never specified. made a function that should reliably get us the current app id as this difference between formplayer and preview has been pretty annoying. i also added a test for the session middlware. should make it clearer what it's doing too

cc: @emord @nickpell

### DIFF
--- a/corehq/apps/cloudcare/static/cloudcare/js/formplayer/app.js
+++ b/corehq/apps/cloudcare/static/cloudcare/js/formplayer/app.js
@@ -216,7 +216,7 @@ FormplayerFrontend.reqres.setHandler('getCurrentAppId', function() {
         user = FormplayerFrontend.request('currentUser'),
         appId;
 
-    appId = urlObject.appId
+    appId = urlObject.appId;
 
     if (appId) {
         return appId;
--- a/corehq/apps/cloudcare/static/cloudcare/js/formplayer/app.js
+++ b/corehq/apps/cloudcare/static/cloudcare/js/formplayer/app.js
@@ -191,6 +191,7 @@ FormplayerFrontend.on("start", function (options) {
         if (this.getCurrentRoute() === "") {
             if (options.phoneMode) {
                 appId = options.apps[0]['_id'];
+                user.previewAppId = appId;
 
                 FormplayerFrontend.trigger('setAppDisplayProperties', options.apps[0]);
                 FormplayerFrontend.trigger("app:singleApp", appId);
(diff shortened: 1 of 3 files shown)


corehq/apps/cloudcare/static/cloudcare/js/formplayer/spec/session_middleware_spec.js
```diff
@@ -0,0 +1,30 @@
...
+/* eslint-env mocha */
+
+describe('SessionMiddle', function() {
+    var Middleware = FormplayerFrontend.SessionNavigate.Middleware;
+
+    it('Should call middleware and apis with same arguments', function() {
+        var middlewareSpy = sinon.spy(),
+            result,
+            API = {
+                myRoute: sinon.spy(function(one, two, three) {
```
dimagimon: 'three' is defined but never used (no-unused-vars)

corehq/apps/cloudcare/static/cloudcare/js/formplayer/app.js
```diff
@@ -209,6 +210,25 @@ FormplayerFrontend.on("start", function (options) {
...
+
+    appId = urlObject.appId;
+
+    if (appId) {
+        return appId;
+    }
+
+    // If it's not in the URL, then we are either on the home screen of formplayer
+    // and there is no app selected, or we are in preview mode.
+    appId = user.previewAppId;
```
wpride: I'm probably missing something, but shouldn't the appId be set in preview mode?

benrudolph: what do you mean by set? it uses it in the route but doesn't save it anywhere. the id exists in the # url fragment. we could add it to the url in the url object util. is that what you mean? that might be nicer than attaching it the user

wpride: Gotcha, yeah that's what I was trying to convey. Fine to roll into a later refactor if we want

benrudolph: so i actually looked into doing that. and it doesn't quite work, because when you set the appId in the URL, the fragment changes from `#single_app/<app_id>` to `#{...json..}` which means when you refresh or go back, you go to the formplayer home screen instead of the single app view. we mit in the future by doing: `#single_app/<app_id>/{...json...}` but seems kind of hacky. another approach would be to drop the `single_app` route and just use it as a parameter in the json blob

wpride: Might be time to write a new front end! Man technology advances so fast these days...

benrudolph: 😵

PR #13777, opened 2016-10-31
Make form detail and edit views location safe.

esoergel:
@sravfeyn @kaapstorm 
This isn't actually useful on it's own, since there's no way to navigate to the form pages without reports working.  I'm getting put on interrupt though, so I figure this is a good stopping point 'till I can get back to it.
:blowfish: :x: :blowfish: should make sense.


esoergel:
@kaapstorm

kaapstorm:
LGTM. Failure looks related.

esoergel:
:white_check_mark:

PR #13778, opened 2016-10-31
fix case_id_to_case_name method signature

dannyroberts:
https://github.com/dimagi/commcare-hq/pull/13778
This was just an oversight I believe. @calellowitz to confirm

### DIFF
--- a/corehq/apps/export/transforms.py
+++ b/corehq/apps/export/transforms.py
@@ -16,8 +16,8 @@ def owner_id_to_display(owner_id, doc, domain):
     return cached_owner_id_to_display(owner_id)
 
 
-def case_id_to_case_name(case_id, doc):
-    return _cached_case_id_to_case_name(case_id)
+def case_id_to_case_name(case_id, doc, domain):
+    return _cached_case_id_to_case_name(case_id, domain)
 
 NULL_CACHE_VALUE = "___NULL_CACHE_VAL___"
 


calellowitz (approved):

dannyroberts:
I'm going to go ahead and hotfix deploy this (by cherrypicking to a hotfix branch on top of last deploy).

czue:
this makes sense to me, though is it clear to you how this way of being broken could produce the stack trace in http://manage.dimagi.com/default.asp?240515? it seems like the call 
`TRANSFORM_FUNCTIONS[self.item.transform](value, doc)`
would now fail because domain is not being provided.

dannyroberts:
@czue thanks for the catch. You're right that this wouldn't have fixed it. Given that this has been the way it is for years, I'm going to opt to revert https://github.com/dimagi/commcare-hq/pull/13752 instead of trying to patch it under pressure. It clearly just didn't go through enough testing/wasn't as simple as we thought it was.

dannyroberts:
https://github.com/dimagi/commcare-hq/pull/13778

dannyroberts:
https://github.com/dimagi/commcare-hq/pull/13780

benrudolph:
also just noting that this should be fairly straightforward to write a test for, especially not that it doesn't use couch. you can check out some example tests here for writing an export file: https://github.com/dimagi/commcare-hq/blob/br/hardware-settings-2/corehq/apps/export/tests/test_get_export_file.py#L476

PR #13782, opened 2016-10-31
Revert "Revert "Signal to update inferred schema""

benrudolph:
Reverts dimagi/commcare-hq#13781

### DIFF
--- a/corehq/apps/callcenter/utils.py
+++ b/corehq/apps/callcenter/utils.py
@@ -11,6 +11,7 @@ from corehq.apps.es.domains import DomainES
 from corehq.apps.es import filters
 from corehq.apps.hqcase.utils import submit_case_blocks
 from corehq.apps.locations.models import SQLLocation
+from corehq.apps.export.signals import added_inferred_export_properties
 from corehq.form_processor.interfaces.dbaccessors import CaseAccessors
 from corehq.util.quickcache import quickcache
 from corehq.util.timezones.conversions import UserTime, ServerTime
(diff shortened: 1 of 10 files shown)


benrudolph:
@czue @NoahCarnahan so this is kind of a sticky situation. all of our system properties for exports are defined with `ExportItem`, and all our regular schema items are defined with `ScalarItem`. When doctypes don't match, it doesn't merge items (as seen in the last PR). if someone were to do a case import, and have one of the columns map to a system property like case_id, then it'd merge fine. if it were another property like `my_random_case_prop` it would merge incorrectly because `ExportItem`s and `ScalarItem`s don't merge. a couple of workarounds

1. Switching all system properties to scalar types doesn't really work, because then we end up with duplicates in the export even if we rebuild the schema (we never really delete items from a person's export)

2. Remove `ScalarItem`. the only thing `ScalarItem` did was give a semantic meaning to the export item. for all intents and purposes it was the same as `ExportItem`. if we do this though we end up with duplicates of all items that were scalar items in existing exports. (we could write some sort of purge that would remove it upon opening an export)

3. we get tricksy with the way we find the column in the export when using the inferred schema. if the path matches a system property, it's an `ExportItem`, else it's a `ScalarItem`.

Curious if you guys have any thoughts. my gut says 3 is the easiest to do cleanly, but leaves us with tech debt. i would like to be in the world where this can't happen again and coalesce ExportItem and ScalarItem since they are the same thing

NoahCarnahan:
Thanks for laying out the options. 3 also sounds like the easiest approach to me. Do you want us to hold off on merging this PR until your comment is resolved?

benrudolph:
ok sounds good. yeah I'll add label

PR #13784, opened 2016-11-01
Fix possible issue with transform functions due to difference in argu…

mkangia:
…meets count

https://github.com/dimagi/commcare-hq/pull/13752/files#diff-0111f3dfa65bf39568e054504f4430afR11 should need this.

### DIFF
--- a/corehq/apps/export/models/new.py
+++ b/corehq/apps/export/models/new.py
@@ -222,7 +222,7 @@ class ExportColumn(DocumentSchema):
         if transform_dates:
             value = couch_to_excel_datetime(value, doc)
         if self.item.transform:
-            value = TRANSFORM_FUNCTIONS[self.item.transform](value, doc)
+            value = TRANSFORM_FUNCTIONS[self.item.transform](value, doc, domain)
         if self.deid_transform:
             try:
                 value = DEID_TRANSFORM_FUNCTIONS[self.deid_transform](value, doc)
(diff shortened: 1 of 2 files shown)


mkangia:
@calellowitz @dannyroberts Took a shot at this.

mkangia:
@calellowitz the original changes done were reverted due to some ES issues. So just keeping this PR to ensure the changes go through when you pick it up again. You can close/merge it then.

corehq/apps/export/models/new.py
```diff
@@ -222,7 +222,7 @@ def _transform(self, value, doc, transform_dates, domain):
         if transform_dates:
             value = couch_to_excel_datetime(value, doc)
         if self.item.transform:
-            value = TRANSFORM_FUNCTIONS[self.item.transform](value, doc)
+            value = TRANSFORM_FUNCTIONS[self.item.transform](value, doc, domain)
```
benrudolph: all of these signatures would need to change: https://github.com/dimagi/commcare-hq/blob/master/corehq/apps/export/const.py#L28. we'd also need need to make sure that old exports still work which i believe gets called here: https://github.com/dimagi/commcare-hq/blob/master/corehq/ex-submodules/couchexport/models.py#L572. there are some example tests that you could leverage in `test_get_export_file`

benrudolph (changes requested):

corehq/apps/export/models/new.py
mkangia: Thanks @benrudolph 
The signatures being called already have the argument added. It was missing when called.

i think @calellowitz would look into once he is back since the underlying changes were reverted.

benrudolph: 👍 missed those original changes, thanks

PR #13788, opened 2016-11-01
remaining SQL models for dump / load

snopoke:
This all the remaining SQL models except for `reminders` and `sms` models.

@emord 
buddy @biyeun


corehq/apps/dump_reload/sql/dump.py
```diff
@@ -19,6 +19,7 @@
     'form_processor.XFormOperationSQL': SimpleFilter('form__domain'),
     'form_processor.CommCareCaseSQL': SimpleFilter('domain'),
     'form_processor.CommCareCaseIndexSQL': SimpleFilter('domain'),
+    'form_processor.CaseAttachmentSQL': SimpleFilter('case__domain'),
```
emord: Do case attachments exist for icds? I thought that case attachments aren't released/used for performance reasons. (agreed it should be here for completeness though)

emord (approved):
Looks good to me

snopoke:
Correct, they aren't used but included here for completeness

PR #13789, opened 2016-11-01
UCR reports in ES

emord:
This is almost completely ES specific code. A big limitation right now to adding more tests is that ES doesn't support the percentage column type which most of the tests have right now. I think once that and the new calculated column have support, then it almost all (maybe even all) ucr tests will support both backends.

It could also be an interesting place to try out a scientist experiment where a data source could have a backing that's both postgres and ES, so that we can test out performance and correctness

buddies @nickpell @benrudolph


corehq/apps/userreports/es/columns.py
```diff
@@ -0,0 +1,34 @@
+from corehq.apps.es import filters
+from corehq.apps.es.aggregations import FilterAggregation
+from corehq.apps.reports.datatables import DataTablesColumn
```
dimagimon: F401 'DataTablesColumn' imported but unused

corehq/apps/userreports/reports/data_source.py
```diff
@@ -1,6 +1,7 @@
 from corehq.apps.userreports.models import DataSourceConfiguration, get_datasource_config
+from corehq.apps.userreports.es.data_source import ConfigurableReportEsDataSource
 from corehq.apps.userreports.sql.data_source import ConfigurableReportSqlDataSource
-from corehq.apps.userreports.util import get_table_name
+from corehq.apps.userreports.util import get_table_name, get_backend_id
```
dimagimon: F401 'get_table_name' imported but unused

corehq/apps/userreports/tests/utils.py
```diff
@@ -108,3 +111,12 @@ def pre_run_with_es_backend(fn, *args, **kwargs):
         ),
     ]
 )
+
+
+def mock_sql_backend():
```
emord: Created these mocks to get past some tests that will require the SQL backend, but don't actually save the data source config to the db

PR #13792, opened 2016-11-01
Shutdown by hostname command

benrudolph:
@gcapalbo this is what the command would look like to shutdown a worker by hostname. fine to merge, will do some testing on staging

### DIFF
--- a/corehq/apps/hqadmin/management/commands/shutdown_celery_worker_by_hostname.py
+++ b/corehq/apps/hqadmin/management/commands/shutdown_celery_worker_by_hostname.py
@@ -2,6 +2,8 @@ from django.core.management.base import BaseCommand
 from django.conf import settings
 from celery import Celery
 
+from corehq.apps.hqadmin.utils import parse_celery_pings
+
 
 class Command(BaseCommand):
     help = "Gracefully shutsdown a celery worker"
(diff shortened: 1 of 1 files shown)


benrudolph:
(killed build)

gcapalbo:
looks good. is there any output we can get from this to tell if the command is sent properly or not? i think it would be great if running this command gives a different exit code based on whether the operation was successful or not

benrudolph:
that's a good idea, i'll look into what it returns / if we can make it return something different

corehq/apps/hqadmin/management/commands/shutdown_celery_worker_by_hostname.py
```diff
@@ -11,3 +13,9 @@ def handle(self, hostname, *args, **options):
         celery = Celery()
         celery.config_from_object(settings)
         celery.control.broadcast('shutdown', destination=[hostname])
+        worker_responses = celery.control.ping(timeout=10, destination=[hostname])
+        pings = parse_celery_pings(worker_responses)
+        if hostname in pings:
+            print 'Did not shutdown worker'
+            exit(1)
+        print 'Successfully initiated warm shutdown'
```
benrudolph: i tested this out on staging and `shutdown` returns nothing in any situation. the ping response will return empty though if it's properly in warm shutdown

PR #13794, opened 2016-11-01
query on correct field

nickpell:
@emord @benrudolph 

Recommending merging before tests and deploying since the self-service subscription change workflow is broken.


benrudolph:
ok

benrudolph:
fine by me

PR #13796, opened 2016-11-01
Clear app cache when copying app

orangejenny:
http://manage.dimagi.com/default.asp?240441

@dannyroberts / @esoergel

### DIFF
--- a/corehq/apps/app_manager/views/apps.py
+++ b/corehq/apps/app_manager/views/apps.py
@@ -298,6 +298,7 @@ def copy_app(request, domain):
         else:
             app_id_or_source = app_id
 
+        clear_app_cache(request, domain)
         return copy_app_check_domain(request, form.cleaned_data['domain'], form.cleaned_data['name'],
                                      app_id_or_source)
     else:
--- a/corehq/apps/app_manager/views/apps.py
+++ b/corehq/apps/app_manager/views/apps.py
@@ -277,12 +277,6 @@ def app_source(request, domain, app_id):
 
 
 @require_can_edit_apps
-def copy_app_check_domain(request, domain, name, app_id_or_source):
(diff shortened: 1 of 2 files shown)


corehq/apps/app_manager/views/apps.py
```diff
@@ -277,12 +277,6 @@ def app_source(request, domain, app_id):
 
 
 @require_can_edit_apps
-def copy_app_check_domain(request, domain, name, app_id_or_source):
```
dannyroberts: Do you know why this used to be here as a view?

dannyroberts (approved):

corehq/apps/app_manager/views/apps.py
orangejenny: I don't. Looked at history, it's been its own view for at least a year, but not sure why.

dannyroberts: ok. `git log -S copy_app_check_domain` implies it was added in 67d05212fd447e3747ce4b24a1f48588de144e14.

It's super not explicit, but the intention was that `login_and_domain_required` be used to check that the user has login access to the domain `req.GET.get('domain')` as well as the domain passed in the url (/a/:domain:/...). It's not the most straightforward way to do it, but it is an important check, and I think the proposed change would remove that check. Do you agree?

orangejenny: Thanks for investigating further (and for telling me about `-S`, which I did not know or had forgotten about). Definitely agree it's an important check. Having `copy_app_check_domain` as a separate view when it's never called as a view seems misleading, so I just added a commit to put back the domain-checking logic but keep it inside `copy_app`. I think it's clearer, lmk if you find it more confusing than the original.

dannyroberts: Yeah, that's a lot clearer. Thanks!

PR #13800, opened 2016-11-01
add some recommendations for elastic query performance

dannyroberts:
Just opening up for collaboration. Is this a good place/format for this?

### DIFF
--- a/docs/elasticsearch.rst
+++ b/docs/elasticsearch.rst
@@ -1,56 +1,6 @@
 ElasticSearch
 =============
 
-Playing Nice with ES: Simple things to know to get reasonable performance
--------------------------------------------------------------------------
-
-Here are the most basic things to know if you want to get reasonable performance
-out of Elasticsearch.
-
-Prefer "get" to "search"
-========================
-
-Don't use search to fetch a doc or doc fields by doc id; use "get" instead.
-
-**Bad:**
-
-POST /hqcases_2016-03-04/case/_search
-{
-  "query": {
-    "filtered": {
-      "filter": {
-        "and": [{
-          "terms": {
-            "_id": [case_id]
-          }
-        }, {
-          "match_all": {}
-        }]
-      },
-      "query": {"match_all":{}}
-    }
-  },
-  "_source": ["name"],
-  "size":1000000
-}
-
-**Good:**
-
-GET /hqcases_2016-03-04/case/<case_id>?_source_include=name
-
-
-Prefer scroll queries
-=====================
-
-Use a scroll query when fetching lots of records.
-
-
-Prefer query to filter
-======================
-
-Don't use ``query`` when you could use ``filter`` if you don't need rank.
-
-
 Indexes
 -------
 We have indexes for each of the following doc types:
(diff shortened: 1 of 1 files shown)


dannyroberts:
@esoergel does this feel like a reasonable page to put it on, and a reasonable place in that page? Feel free to suggest a new place or to make changes yourself. In addition, can I get your eyes on my example and help coming up with examples in the other sections I added, if an example could be helpful? FYI @snopoke @orangejenny

docs/elasticsearch.rst
```diff
@@ -1,6 +1,56 @@
 ElasticSearch
 =============
 
+Playing Nice with ES: Simple things to know to get reasonable performance
```
emord: I was going to say that you could put this https://github.com/dimagi/commcare-hq/blob/master/docs/es_query.rst because it's linked at the bottom of this page, but that doesn't look like a very helpful doc. Maybe replace `Querying ElasticSearch` with this?

docs/elasticsearch.rst
```diff
@@ -1,6 +1,56 @@
...
+Prefer scroll queries
+=====================
+
+Use a scroll query when fetching lots of records.
+
+
+Prefer query to filter
+======================
+
+Don't use ``query`` when you could use ``filter`` if you don't need rank.
```
emord: You could move this section and point to something like https://github.com/dimagi/commcare-hq/blob/master/corehq/apps/es/es_query.py#L8

Seems like those docs could say something like "Only use `set_query` if you know what you are doing"

czue:
once we agree where this goes we should also add the "use `size(0)` when you're only doing aggregations thing" - otherwise it returns the entire result set with the result

docs/elasticsearch.rst
dannyroberts: Ha, unhelpful indeed. Maybe I'll delete that doc, and move this stuff to `Querying ElasticSearch` at the bottom of this doc, like you suggested.

dannyroberts: oh, good to know. I'm going to leave it as it is then. and just add to the bottom of `Querying ElasticSearch`.

docs/elasticsearch.rst
dannyroberts: Ok, adding comment on `set_query`. I didn't fully understand the first sentence in this comment.

dannyroberts:
I think I incorporated most of the feedback. Lemme know if I missed something!

docs/elasticsearch.rst
```diff
@@ -83,7 +83,74 @@ blow away the index and start over.
...
+    GET /hqcases_2016-03-04/case/<case_id>?_source_include=name
+
+
+Prefer scroll queries
+=====================
+
+Use a scroll query when fetching lots of records.
+
+
+Prefer query to filter
```
benrudolph: i think this should be `Prefer filter to query`

dannyroberts: Oh, oops!

czue (approved):
thanks for these Danny!

docs/elasticsearch.rst
```diff
@@ -83,7 +83,74 @@ blow away the index and start over.
...
+          "query": {"match_all":{}}
+        }
+      },
+      "_source": ["name"],
+      "size":1000000
+    }
+
+**Good:**::
+
+    GET /hqcases_2016-03-04/case/<case_id>?_source_include=name
```
NoahCarnahan: Might be worth giving some code samples instead of this format. ESQuery is my go-to way of interacting with ElasticSearch (and I think we encourage that, and discourage building the requests manually), so it might be worth pointing out that this is an exception to that rule, and showing which function should be used instead (is it `get_es_new().get(index, id)`?)

PR #13808, opened 2016-11-02
Warn on copying app if feature flags differ

orangejenny:
Just an idea after mulling over http://manage.dimagi.com/default.asp?240571 and some summit conversations, open to feedback.

@dannyroberts / @esoergel / @dimagi/product 

screen shot 2016-11-02 at 12 10 59 am


corehq/apps/domain/views.py
```diff
@@ -2525,6 +2528,20 @@ def page_context(self):
 
 @login_and_domain_required
 @require_superuser
+@require_GET
+def feature_flag_diff(request, domain):
+    params = json_request(request.GET)
+    other_domain = params.get('domain')
+    diff = []
+    if Domain.get_by_name(other_domain):
+        diff = [(t.slug, t.label) for t in all_toggles() if t.enabled(request.domain) and not t.enabled(other_domain)]
```
dimagimon: E501 line too long (118 > 115 characters)

czue:
this is awesome. 

what do you think about providing the options "Copy and enable all" and "Copy without enabling"? I imagine that in most cases the person wants to enable everything, though providing that does make it easier for people to turn on lots of flags which may be against our long-term philosophy of trying to reduce flag usage...

amsagoff:
If this isn't just superuser facing, we may want to change the text
("feature flag" isn't a term external users are exposed to)

sheelio:
I'm guessing we only want this for superusers (since feature flags don't make sense for normal users anyway?)

(Alternatively, we could have it always copy the feature flags if you're not a superuser and you copy an app from a project with feature flags enabled).

czue:
looks like this is already superuser only

orangejenny:
How about checkboxes to enable flags? I like the thought of making that easier but am wary of copying all flags because of the ones that control things like the new exports backend and new cloudcare.

czue:
checkboxes was actually my first thought, but I thought that might be asking you to do too much work. :)

amsagoff:
superuser only and checkboxes all the way!

corehq/apps/app_manager/templates/app_manager/v1/app_view.html
```diff
@@ -188,6 +188,48 @@
             }));
         });
     </script>
+    <script>
+        $("#copy form button").click(function() {
+            var $submit = $(this),
+                $form = $submit.closest("form"),
+                domain = $form.find("#id_domain").val(),
+                $modal = $("#copy-feature-flags");
+            {% if request|toggle_enabled:"SUPPORT" %}
```
esoergel: Shouldn't this be enabled for all dimagi users?  You could use `user.is_superuser`, since those are the users who can access feature flags.

corehq/apps/app_manager/templates/app_manager/v1/app_view.html
```diff
@@ -188,6 +188,48 @@
...
+                            _.each(links, function(link) {
+                                $ul.append($("<li/>").html(link));
+                            });
+                            $modal.one("click", ".btn-primary", function() {
+                                $(this).disableButton();
+                                $form.submit();
+                            });
+                            $modal.one("hide.bs.modal", function() {
+                                $submit.enableButton();
+                            });
```
esoergel: This is cool, I hadn't seen [`.one`](http://api.jquery.com/one/) before.

corehq/apps/domain/views.py
```diff
@@ -2525,6 +2528,21 @@ def page_context(self):
...
+def feature_flag_diff(request, domain):
+    params = json_request(request.GET)
+    other_domain = params.get('domain')
+    diff = []
+    if Domain.get_by_name(other_domain):
+        diff = [(t.slug, t.label) for t in all_toggles()
+                if t.enabled(request.domain) and not t.enabled(other_domain)]
+        diff.sort(cmp=lambda x, y: cmp(x[1], y[1]))
+        diff = ["<a href='{}'>{}</a>".format(reverse(ToggleEditView.urlname, args=[d[0]]), d[1]) for d in diff]
+    return json_response(diff)
```
esoergel: Probably not worth changing at this point, but you may be able to do this client-side with just [`toggles.js`](https://github.com/dimagi/commcare-hq/blob/f466e62b0754ca09c4dc0e3cc784cb0c9a31cd0f/corehq/apps/hqwebapp/views.py#L1119-L1126), which the user should already have cached in their browser.

esoergel (commented):
I had a couple minor, optional pieces of feedback, but implementation-wise, this looks g2m as-is.  From the PR discussion it looks like there'll be some changes though, so I'll hold off on approval for those.

corehq/apps/domain/views.py
orangejenny: Would be nice to do client-side, but toggles.js only knows about the current domain's flags, not the one you're copying to.

orangejenny:
Added checkboxes to selectively enable flags while coying. Also added feature previews, but kept it superuser-only - didn't think the cost of the extra step was worth the benefit for non-superusers, since there are relatively few feature previews.

screen shot 2016-11-03 at 5 06 27 pm

corehq/apps/domain/views.py
esoergel: You can access it for any domain the user has access to:
```python
In [1]: reverse('toggles_js', args=['mydomain'])
Out[1]: u'/a/mydomain/toggles.js'
```
Although I guess it's imported as a named module, so I could see that being tricky to get two copies in accessible formats.

orangejenny: Oh, I see what you mean. Makes sense, but think I'm going to leave it as-is.

esoergel: sounds good

czue:
(added the announce label, for next internal announcement, because this is sweet)

PR #13809, opened 2016-11-02
Check for duplicates only in product's domain

kkrampa:
http://manage.dimagi.com/default.asp?240599#1240383

@gcapalbo


czue (approved):
kicked the build

PR #13811, opened 2016-11-02
Enforce size 0 when using aggregations

emord:
https://github.com/dimagi/commcare-hq/pull/13800#issuecomment-257698720

@czue @esoergel @dannyroberts 

I don't think that there's much of a use case for having a list of things and an aggregation, but I could be missing something. Thoughts on enforcing this instead of relying on documentation?

### DIFF
--- a/corehq/apps/es/es_query.py
+++ b/corehq/apps/es/es_query.py
@@ -207,11 +207,15 @@ class ESQuery(object):
 
     def run(self, no_hits_with_aggs=True):
         """Actually run the query.  Returns an ESQuerySet object."""
-        query = self
-        if no_hits_with_aggs and query.uses_aggregations():
-            query = query.size(0)
+        query = self._clean_before_run(no_hits_with_aggs)
         raw = run_query(query.index, query.raw_query, debug_host=query.debug_host)
         return ESQuerySet(raw, deepcopy(query))
+
+    def _clean_before_run(self, no_hits_with_aggs=True):
+        query = deepcopy(self)
+        if no_hits_with_aggs and query.uses_aggregations():
+            query = query.size(0)
+        return query
 
     def scroll(self):
         """
--- a/corehq/apps/es/es_query.py
+++ b/corehq/apps/es/es_query.py
@@ -205,15 +205,15 @@ class ESQuery(object):
             size = sliced_or_int.stop - start
         return self.start(start).size(size).run().hits
 
-    def run(self, no_hits_with_aggs=True):
+    def run(self, include_hits=False):
         """Actually run the query.  Returns an ESQuerySet object."""
-        query = self._clean_before_run(no_hits_with_aggs)
+        query = self._clean_before_run(include_hits)
         raw = run_query(query.index, query.raw_query, debug_host=query.debug_host)
         return ESQuerySet(raw, deepcopy(query))
 
-    def _clean_before_run(self, no_hits_with_aggs=True):
+    def _clean_before_run(self, include_hits=False):
         query = deepcopy(self)
-        if no_hits_with_aggs and query.uses_aggregations():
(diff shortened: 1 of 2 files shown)


esoergel:
I can imagine some instances where that'd be useful, say a report that uses the list for the rows, but also wants a totals row at the bottom using data pulled from ES.  I can't think of anything offhand that does that though, and I can certainly imagine it's more likely that we DON'T want a list in aggregation queries.

We could default to `size(0)` unless you manually specify in some way that you DO want the list.

Either way, I think `query.run().hits` should raise an exception if you try to access it from a query that excluded results automatically.  That would make it obvious if someone is expecting the current behavior.

esoergel:
(test failures look real, btw)

corehq/apps/es/es_query.py
```diff
@@ -446,7 +452,11 @@ def doc_ids(self):
     @property
     def hits(self):
         """Return the docs from the response."""
-        return [self.normalize_result(self.query, r) for r in self.raw_hits]
+        raw_hits = self.raw_hits
+        if not raw_hits and self.query._aggregations and self.query._size == 0:
+            raise EsError("no hits, did you forget about no_hits_with_aggs?")
```
dimagimon: F821 undefined name 'EsError'

emord:
I added a way to override the default behavior. I hadn't thought about the total row but that makes sense. I think now reports handles those with separate requests.

esoergel (approved):
lgtm on :construction_worker_man:

emord:
Added a test. Briefly went through all uses of `aggregation` in HQ `ag -w --stats --python aggregation` (can ignore all of userreports) and they all used `size(0)`, so I don't think this will cause errors :fist: :deciduous_tree:

corehq/apps/es/es_query.py
```diff
@@ -205,10 +205,17 @@ def __getitem__(self, sliced_or_int):
             size = sliced_or_int.stop - start
         return self.start(start).size(size).run().hits
 
-    def run(self):
+    def run(self, no_hits_with_aggs=True):
```
dannyroberts: If anyone ever used this, I'd suggest a different name, like `include_hits=None` and then below `if include_hits is None: include_hits = query.uses_aggregations()` and `if not include_hits: query.size(0)`. That way the way you call it is `.run(include_hits=True)` (if using aggs, otherwise no need to call it).

On the other hand, it looks like it's not used anywhere. I know you and @esoergel talked about it, but if that's the case I'm tempted to YAGNI it.

corehq/apps/es/tests/test_esquery.py
```diff
@@ -204,3 +205,32 @@ def test_sort(self):
             {"timeStart": {"order": "asc"}},
         ]
         self.checkQuery(query.sort('timeStart', reset_sort=False), json_output)
+
+    def test_cleanup_before_run(self):
```
dannyroberts: Would be cool if the tests here were more along the lines of testing what query is sent through to ES (through a mock) so you didn't have to test the private interface `_clean_before_run`. Just a thought, not blocking

dannyroberts (approved):
Looks good to me, made a couple comments you can decide what do with

czue (approved):
nice this is a great call

corehq/apps/es/es_query.py
emord: https://github.com/dimagi/commcare-hq/pull/13811/commits/60c341f8592dacef8bdfb494ff792e952e287205

Thanks! I couldn't come  up with a good name

corehq/apps/es/tests/test_esquery.py
emord: Yeah. agreed it's not the best test, but I balked at writing up more test code than necessary since it would probably only get used right here.

PR #13812, opened 2016-11-02
Refactor formplayer frontend

benrudolph:
@wpride @biyeun this refactors the formplayer frontend. i didn't change any logic or anything, then main goal was to pull apart namespaces so they fit in the current world. some guiding principles:

1. `FormplayerFrontend.module('path.to.name'...)` should specify the file path so one can easily find the file that the module is referencing. I.E `FormplayerFrontend.Apps.Views` should live in `apps/views.js`

2. There are essentially 6 types of files in each folder:
`api.js` - this file denotes methods that connect directly to formplayer to fetch resources (it's a bit convoluted since we do some fetching with the models as well)
`controller.js` - links api calls to view renderings
`models.js` - models
`collections.js` - collections
`views.js` or `views` folder - views
`utils.js` - utility method pertaining to that logical module

i'll be working off this branch to avoid hellish merge conflicts, so i'd also appreciate you guys working off this if possible so we can get this in. i don't think we're doing too much churn on this part of the code so thinking it'll be OK. definitely best read commit by commit :fish:

cc: @NoahCarnahan


corehq/apps/cloudcare/static/cloudcare/js/formplayer/apps/models.js
```diff
@@ -0,0 +1,9 @@
+/*global FormplayerFrontend */
+
+FormplayerFrontend.module("Apps.Models", function (Models, FormplayerFrontend, Backbone) {
+    Models.App = Backbone.Model.extend({
+        urlRoot: "appSelects",
```
benrudolph: i copied this, but i don't actually think it does anything. didn't want to make code changes on this pass

PR #13815, opened 2016-11-02
analytics for live preview A/B test

orangejenny:
@amsagoff this does the following
- Adds `dimension6` and `dimension7` to google analytics: 6 is `yes` if it's the user's first 24 hours and 7 is `yes` if it's past the first 24 hours but within the first week. I'm assuming we can use this to track # visits to HQ within a new user's first day and first week. (Would be good if you or @krsdimagi can confirm this.)
- Adds a kiss event for form builder saves within the first 24 hours.
- Adds a kiss event each time a question is added in form builder in the first 24 hours.
- Adds a kiss and a google event for clicking `Preview Form` within the first 24 hours.

Did not add an event for app manager saves in general because there are quite a few pages in the ui where we save the app. They all go through `ApplicationBase.save` eventually, but by that point we no longer know who the user is.

@dannyroberts

### DIFF
--- a/corehq/apps/app_manager/templates/app_manager/v1/form_view_base.html
+++ b/corehq/apps/app_manager/templates/app_manager/v1/form_view_base.html
@@ -259,6 +259,9 @@
             $('#cloudcare-preview-url').click(function(e) {
                 ga_track_event('CloudCare', 'Click "Preview Form"');
                 analytics.workflow("Clicked Preview Form");
+                {% if request.couch_user.days_since_created == 0 %}
+                    analytics.workflow("Clicked Preview Form within first 24 hours");
+                {% endif %}
             });
         {% endif %}
         });
--- a/corehq/apps/app_manager/templates/app_manager/v1/form_view_base.html
+++ b/corehq/apps/app_manager/templates/app_manager/v1/form_view_base.html
@@ -260,8 +260,8 @@
                 ga_track_event('CloudCare', 'Click "Preview Form"');
                 analytics.workflow("Clicked Preview Form");
                 {% if request.couch_user.days_since_created == 0 %}
-                    ga_track_event('CloudCare', 'Click "Preview Form" within first 24 hours');
-                    analytics.workflow("Clicked Preview Form within first 24 hours");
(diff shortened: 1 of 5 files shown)


amsagoff:
:sunglasses:

czue:
not asking you to change anything but just noting in case you didn't know that you can workaround this issue:

> Did not add an event for app manager saves in general because there are quite a few pages in the ui where we save the app. They all go through `ApplicationBase.save` eventually, but by that point we no longer know who the user is.

by using the global request object which is kind of designed for a situation like this. see [here](https://github.com/dimagi/commcare-hq/blob/master/corehq/util/log.py#L214) for one example.

orangejenny:
That's perfect, indeed, I didn't know about `get_request`. Updated.

orangejenny:
@dannyroberts this is ready for review

corehq/apps/app_manager/templates/app_manager/v1/form_view_base.html
```diff
@@ -259,6 +259,10 @@
             $('#cloudcare-preview-url').click(function(e) {
                 ga_track_event('CloudCare', 'Click "Preview Form"');
                 analytics.workflow("Clicked Preview Form");
+                {% if request.couch_user.days_since_created == 0 %}
+                    ga_track_event('CloudCare', 'Click "Preview Form" within first 24 hours');
+                    analytics.workflow("Clicked Preview Form within first 24 hours");
```
dannyroberts: Not blocking, but I'm curious why we need to have these two very similar analytics calls—but I assume there's a good reason. Also, I imagine there's no reason the text on the two is slightly different?

- Click "Preview Form" within first 24 hours
- Clicked Preview Form within first 24 hours

dannyroberts (approved):

corehq/apps/app_manager/templates/app_manager/v1/form_view_base.html
orangejenny: Tweaked the names to make them match, for the sake of sanity. I don't remember why we agreed to have both a kiss event and a GA event - I think the major benefit of kiss here is being able to view the data as a funnel, and the benefit of GA is being able to look at the data by session.

PR #13816, opened 2016-11-02
fix sizing for usercase config select2s

orangejenny:


orangejenny:
no test coverage, killing build

PR #13818, opened 2016-11-02
Signal to update inferred take two

benrudolph:
@NoahCarnahan @czue Same thing but take two. I've better handled inferring ScalarItem and ExportItem and have written an integration test to ensure it merges


benrudolph:
screen shot 2016-11-02 at 3 39 30 pm
^^ and here's the new export with the correct merging

NoahCarnahan:
👍

czue:
:+1: thanks

PR #13821, opened 2016-11-02
Validate app profiling

benrudolph:
@dannyroberts working on speeding up build times for app preview. did a little profiling of the validate app and curious to get your thoughts (maybe we can go over it tomorrow together). On initial build of my app with redis flushed we get a flame graph that looks like this:
screen shot 2016-11-02 at 6 42 50 pm
It makes sense that the dominant factor here is running form translate.

On second run, you can see how much the caching helps as the form translate portion takes almost no time in comparison with the rest
screen shot 2016-11-02 at 6 44 43 pm
Here we see a few other fairly bulky calls and I added a caching to some of them. Here's what it looks like after that caching:

screen shot 2016-11-02 at 7 02 31 pm

also crucially missing is the absolute number of seconds that it takes. going to work on adding those tomorrow

cc: @NoahCarnahan

### DIFF
--- a/corehq/apps/app_manager/models.py
+++ b/corehq/apps/app_manager/models.py
@@ -777,6 +777,15 @@ class CommentMixin(DocumentSchema):
         return self.comment if len(self.comment) <= 500 else self.comment[:497] + '...'
 
 
+def _render_xform_vary_on(self, build_profile_id=None):
+    def _hash(val):
+        return hashlib.md5(val).hexdigest()
+    return [
+        _hash(self.source.encode('utf-8')),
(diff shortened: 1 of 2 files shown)


NoahCarnahan:
Test failures look real

snopoke:
coincidentally I briefly looked at this yesterday (though was looking at the `create_build_files` function which also take a long time).

Sweet flame graphs.

benrudolph:
oh nice, yeah there's definitely some low hanging caching i think we could do to significantly speed things up

PR #13822, opened 2016-11-03
Add ESQuery.get_ids() method

esoergel:
Note that many of these instances weren't excluding source, so they were retrieving full documents, then only using the ids.  This is a common enough use-case that it seems like it makes sense to abstract out, particularly to make sure we do it properly.

I also discovered that it looks like we're not actually excluding source when we say we do, so this is a moot point for now, but I'm working on a fix for that too (to be PR'd separately).  For now, just consider this API cleanup.
@emord

### DIFF
--- a/corehq/apps/hqadmin/reporting/reports.py
+++ b/corehq/apps/hqadmin/reporting/reports.py
@@ -401,7 +401,7 @@ def get_total_clients_data(domains, datespan, interval, datefield='opened_on'):
             .domain(domains)
             .filter({"ids": {"values": cases}})
             .opened_range(lt=datespan.startdate)
-            .size(0)).run().total
+            .count())
 
     return format_return_data(histo_data, cases_before_date, datespan)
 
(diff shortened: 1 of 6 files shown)


corehq/apps/hqadmin/reporting/reports.py
```diff
@@ -598,7 +598,7 @@ def get_domain_stats_data(domains, datespan, interval,
 
 def commtrack_form_submissions(domains, datespan, interval,
         datefield='received_on'):
-    mobile_workers = UserES().exclude_source().mobile_users().show_inactive().run().doc_ids
+    mobile_workers = UserES().mobile_users().show_inactive().get_ids()
```
emord: Seems like this should filter on domains, and use `get_mobile_users`... Not blocking, this file could use a once over for filtering by domains. I think we're returning lots of docs here that aren't needed

emord (approved):

corehq/apps/hqadmin/reporting/reports.py
esoergel: Whoa, yeah!  Looks like as-is this is pulling a list of all users on HQ...

PR #13823, opened 2016-11-03
Fixed UCR pre filter

kkrampa:
@emord 
It seems your changes broke pre-filter. I'm not sure if it's proper fix.


emord:
Thanks @kkrampa Sorry about that!

PR #13824, opened 2016-11-03
Unknown feature flags are actually experimental

kaapstorm:
"Unknown" feature flags are actually "Experimental" but lack the self-confidence to say so ... unless they're "Product Path" and just don't know it.

(I created this PR because the "Unknown" thing suddenly annoyed me more than before, and I figured I'd cautiously suggest the next step after @esoergel's work.)

@esoergel cc @calellowitz , @amsagoff , what do you reckon?


czue (approved):
lgtm, thanks for follow up

PR #13827, opened 2016-11-03
Filter by domains

emord:
https://github.com/dimagi/commcare-hq/pull/13822/files#r86357306

@esoergel

### DIFF
--- a/corehq/apps/hqadmin/reporting/reports.py
+++ b/corehq/apps/hqadmin/reporting/reports.py
@@ -812,14 +812,7 @@ def get_case_owner_filters(domains):
 
     mobile_user_ids = list(get_user_ids(True, domains))
 
-    def all_groups():
-        for domain in domains:
-            for group in Group.by_domain(domain):
-                yield group
-    group_ids = [
-        group._id for group in all_groups()
-    ]
-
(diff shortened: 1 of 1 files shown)


corehq/apps/hqadmin/reporting/reports.py
```diff
@@ -805,14 +807,14 @@ def get_submitted_users():
...
+def get_case_owner_filters(domains):
     result = {'terms': {}}
 
-    mobile_user_ids = list(get_user_ids(True))
+    mobile_user_ids = list(get_user_ids(True, domains))
 
     def all_groups():
-        for domain in Domain.get_all():
-            for group in Group.by_domain(domain.name):
+        for domain in domains:
```
emord: I think this was getting every group that existed in hq everytime someone loaded admin reports...

czue: ha, wow it does look that way. and every domain doc.

emord:
Ok. I think I could spend a couple of days making this file use ES better, but I'll leave that for another day. I think it should be a little nicer now

emord:
:white_check_mark:

esoergel:
awesome, thanks for the cleanup!

PR #13829, opened 2016-11-03
Fix knockout templates for jQuery 3

orangejenny:
Same issue as https://github.com/dimagi/commcare-hq/pull/13677

Grepped for places we use jQuery to create a div that uses a knockout template, looks like it's just these two files.

@dannyroberts

### DIFF
--- a/corehq/apps/app_manager/static/app_manager/js/graph-config.js
+++ b/corehq/apps/app_manager/static/app_manager/js/graph-config.js
@@ -44,7 +44,7 @@ hqDefine('app_manager/js/graph-config.js', function () {
 
             // Load the modal with the copy
             var $modalDiv = $(document.createElement("div"));
-            $modalDiv.attr("data-bind", "template: \'graph_configuration_modal\'");
+            $modalDiv.attr("data-bind", "template: 'graph_configuration_modal'");
 
             $modalDiv.koApplyBindings(graphViewModelCopy);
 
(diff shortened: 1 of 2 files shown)


orangejenny:
@dannyroberts could you review? jQuery 3 is ready to go.

corehq/apps/app_manager/static/app_manager/js/graph-config.js
```diff
@@ -43,7 +43,8 @@ hqDefine('app_manager/js/graph-config.js', function () {
             };
 
             // Load the modal with the copy
-            var $modalDiv = $('<div data-bind="template: \'graph_configuration_modal\'"></div>');
+            var $modalDiv = $(document.createElement("div"));
+            $modalDiv.attr("data-bind", "template: \'graph_configuration_modal\'");
```
dannyroberts: I think the escaping in `\'` is no longer necessary

corehq/apps/hqwebapp/static/hqwebapp/js/key-value-mapping.js
```diff
@@ -189,7 +189,8 @@ uiElement.key_value_mapping = function (o) {
     m.openModal = function () {
         // create a throw-away modal every time
         // lets us create a sandbox for editing that you can cancel
-        var $modalDiv = $('<div data-bind="template: \'key_value_mapping_modal\'"></div>');
+        var $modalDiv = $(document.createElement("div"));
+        $modalDiv.attr("data-bind", "template: \'key_value_mapping_modal\'");
```
dannyroberts: ditto

dannyroberts (approved):

dannyroberts:
Looks good to me (though might address the one comment if you agree)

corehq/apps/app_manager/static/app_manager/js/graph-config.js
orangejenny: indeed, updated

PR #13834, opened 2016-11-03
filter out actions with undefined server_date

nickpell:
http://manage.dimagi.com/default.asp?240618

@NoahCarnahan


corehq/ex-submodules/casexml/apps/case/models.py
```diff
@@ -375,7 +375,9 @@ def modified_since_sync(self, sync_log):
         if self.server_modified_on >= sync_log.date:
             # check all of the actions since last sync for one that had a different sync token
             return any(filter(
-                lambda action: action.server_date > sync_log.date and action.sync_log_id != sync_log._id,
+                lambda action: action.server_date is not None and
+                               action.server_date > sync_log.date and
+                               action.sync_log_id != sync_log._id,
```
dimagimon: E131 continuation line unaligned for hanging indent

PR #13836, opened 2016-11-04
Preview and single app modes

benrudolph:
@wpride (and @czue you may be interested) this separates out the concerns of the different views of nimbus. before we would conflate phoneMode with singleAppMode. This straightens that all out, pulls formplayer main route into its own view and allows for 2 new views on nimbus.

1. Live Preview - Just the regular phone with the most recent app
2. Regular Nimbus - All apps at either the latest build or latest starred depending on a feature flag
`http://localhost:8000/a/aspace/cloudcare/apps/v2/`
3. [new] Preview Nimbus - All apps at the most current version of the app
`http://localhost:8000/a/aspace/cloudcare/apps/v2/preview/`
4. [new] Single App Preview Nimbus - One app at the most current version
`http://localhost:8000/a/aspace/cloudcare/apps/v2/preview/<app_id>/`
looks like this:
screen shot 2016-11-03 at 11 33 50 pm

Aaand for a sneak preview at what I'm implementing next, I'm going to take the 4th option Single App Preview Nimbus, and allow people to "pop" out their Live Preview session. This'll make it easier for power users to use live preview in a bigger window. It could look something like this:

screen shot 2016-11-03 at 11 35 57 pm

(also added a home button to the breadcrumb bar)
🏠 there was a fair bit churn
cc: @NoahCarnahan

### DIFF
--- a/corehq/apps/cloudcare/views.py
+++ b/corehq/apps/cloudcare/views.py
@@ -236,7 +236,6 @@ class CloudcareMain(View):
 class FormplayerMain(View):
 
     preview = False
-    single_app = False
     urlname = 'formplayer_main'
 
     @use_datatables
@@ -263,7 +262,7 @@ class FormplayerMain(View):
             app_ids,
         )
         apps = filter(None, apps)
-        apps = filter(lambda app: app['cloudcare_enabled'], apps)
+        apps = filter(lambda app: app['cloudcare_enabled'] or self.preview, apps)
         apps = filter(lambda app: app_access.user_can_access_app(request.couch_user, app), apps)
         apps = sorted(apps, key=lambda app: app['name'])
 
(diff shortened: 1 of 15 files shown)


corehq/apps/cloudcare/views.py
```diff
@@ -231,6 +233,113 @@ def _get_case(domain, case_id):
             return render(request, "cloudcare/cloudcare_home.html", context)
 
 
+class FormplayerMain(View):
```
benrudolph: felt _so_ good to rip this out of `CloudcareMain`

benrudolph:
One thing to note, is that `/preview/` won't be that useful until we get formplayer doing correct app updates. right now if you open app 123, it will install that app. Then when you make a change to that app, formplayer doesn't correctly update the app to reflect the changes. This isn't an issue for the regular views because each build_id is new everytime you make a change

czue:
:heart_eyes:

benrudolph:
✅

NoahCarnahan (commented):
Looks ok to me, but might be worth some one with more context reviewing as well

wpride:
@NoahCarnahan thanks I'm reviewing too

PR #13839, opened 2016-11-04
MDR suspect report

kkrampa:
It's better to review both commits separately. @czue Any suggestions to index column?

### DIFF
--- a/custom/enikshay/ucr/data_sources/episode.json
+++ b/custom/enikshay/ucr/data_sources/episode.json
@@ -327,6 +327,74 @@
                 "column_id": "mobile_number"
             },
             {
+                "display_name": "person/tu",
+                "datatype": "string",
+                "type": "expression",
+                "is_primary_key": false,
+                "transform": {
+                },
+                "is_nullable": true,
+                "expression": {
+                    "value_expression": {
+                        "datatype": "string",
+                        "type": "property_name",
+                        "property_name": "tu_choice"
+                    },
+                    "type": "related_doc",
+                    "related_doc_type": "CommCareCase",
+                    "doc_id_expression": {
+                        "datatype": null,
+                        "type": "named",
+                        "name": "person_id"
+                    }
+                },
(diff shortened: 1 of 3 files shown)


czue:
@kkrampa the purpose of the `IndexColumn` is to just have a column in each row that increments a counter in the report? It feels pretty silly to me - can we potentially check with Erin if we can get rid of it? If we do need it I don't immediately have a better solution, though definitely don't love the tech debt that this one adds. I'll make a few other comments, but maybe we can add this report without the `IndexColumn` and then see what erin says?

corehq/apps/userreports/reports/util.py
```diff
@@ -23,3 +25,7 @@ def has_location_filter(view_fn, *args, **kwargs):
         if hasattr(filter_, 'choice_provider') else False
         for filter_ in report.ui_filters
     )
+
+
+def get_index_slug(domain):
+    return 'index{}'.format(hashlib.sha1(hashlib.sha1(domain).hexdigest()).hexdigest()[:8])
```
czue: why hash the domain twice?

corehq/apps/userreports/tests/test_columns.py
```diff
@@ -214,20 +215,35 @@ def test_getting_distinct_values(self):
             'apple',
             'banana',
             'blueberry'
-        ])
+        ], [{
+            "type": "expanded",
+            "field": 'my_field',
+            "display": 'my_field',
+            "format": "default",
+        }])
```
czue: looks like these got copy/pasted to a bunch of places. could we make it a class attribute to avoid that?

czue (changes requested):

kkrampa:
@czue I will ask Erin if we can drop this field.

kkrampa:
@czue I removed commit with index column and index column from report.

czue (approved):
thanks

kkrampa:
:white_check_mark:

PR #13844, opened 2016-11-06
represent blank dimagi_contact in one way

nickpell:
http://manage.dimagi.com/default.asp?240619

Resolve inconsistency between ```None``` and ```''``` for blank ```dimagi_contact``` according to django's convention: https://docs.djangoproject.com/en/1.10/ref/models/fields/#django.db.models.Field.null

### DIFF
--- a/corehq/apps/accounting/models.py
+++ b/corehq/apps/accounting/models.py
@@ -320,7 +320,7 @@ class Currency(models.Model):
 
 DEFAULT_ACCOUNT_FORMAT = 'Account for Project %s'
 
-class BillingAccount(models.Model):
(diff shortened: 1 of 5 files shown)


corehq/apps/accounting/models.py
```diff
@@ -320,7 +320,7 @@ def get_default(cls):
 
 DEFAULT_ACCOUNT_FORMAT = 'Account for Project %s'
 
-class BillingAccount(models.Model):
+class BillingAccount(ValidateModelMixin, models.Model):
```
dimagimon: E302 expected 2 blank lines, found 1

nickpell:
@millerdev

PR #13848, opened 2016-11-07
don't process queues and periodic tasks during data migration

snopoke:
I think this should be sufficient for the data migration. All other celery tasks / background jobs are only executed when data is changed and since we're making the domain 'read only' there shouldn't be any data changes.
@gcapalbo

### DIFF
--- a/corehq/apps/data_interfaces/tasks.py
+++ b/corehq/apps/data_interfaces/tasks.py
@@ -11,6 +11,7 @@ from django.template.loader import render_to_string
 from django.utils.translation import ugettext as _
 
 from corehq.apps.data_interfaces.utils import add_cases_to_case_group, archive_forms_old, archive_or_restore_forms
+from corehq.toggles import DATA_MIGRATION
 from .interfaces import FormManagementMode, BulkFormManagementInterface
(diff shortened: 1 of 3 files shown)


snopoke:
I've been through the rest of the periodic tasks and didn't see any that would need to be 'paused'

gcapalbo:
do we need to do this for the pillow retry enqueue process too?

snopoke:
Not sure how that would work for registration SMSs. We'd have to ignore all
of those tasks while any migration is in progress (regardless of domain)
which I don't think would work for the long term since the migration flag
will stay ON for the domain on the old cluster even after the migration is
complete.

For ICDS do we expect any SMS registrations?

We don't need to worry about the pillow retry process. I checked all the
pillows and they're basically just updating ES.

gcapalbo:
we don't expect any sms registrations right now for icds which is why it won't hurt to let those through. but if we block them then we're blocking everyone else's sms registrations which would be bad.

snopoke:
👍 Exactly

gcapalbo:
(i just realized this is good to merge as is, i thought for some reason the toggle check would fail on being passed a None domain but looks like it doesn't)

PR #13851, opened 2016-11-07
rename importer to case_importer_v1

dannyroberts:
and add case_importer_v2 stub

@orangejenny cc @NoahCarnahan (arch)

I'm going to be rewriting case importer, and I wanted to reorganize a bit. The current app is called `importer` which isn't a good name, and I didn't want to just add a `case_importer` app, since it'd be really not obvious that `case_importer` is the logical successor to `importer`. I went with the naming scheme

- `case_importer_v1` <-- current case importer
- `case_importer_v2` <-- future case importer

(In this PR `case_importer_v2` is empty.)

I chatted with @proteusvacuum and he suggested alternatives, such as having the apps live in `case_importer/v1` and `case_importer/v2` (add a level of nesting) or (when I pointed out that our tooling requires app names to be unique) `case_importer/case_importer_v1` and `case_importer/case_importer_v2`. The argument is that this avoids adding another app to this long list, and that the two apps logically go together so it'd make sense to nest them together.

My own arguments for the convention I used in this PR is that (these are all up for debate, but in my opinion) (1) it's the least deviation from our current practice, and thus (2) the most obvious/expected/discoverable for someone stumbling upon it, and (3) least importantly it's shorter than `case_importer/case_importer_v1`. It also clearly shows that the forking point is at the app level and they have no shared code. (You could say the same about the nested version, but if all you see is the line `from corehq.apps.case_importer.case_importer_v1.models import foo`, it's less immediately obvious that they share no code.)

Let me know what you think. If happy with this, I'll do a little QA and we can merge it in soon before I start working on the actual feature.


corehq/apps/case_importer_v1/tests.py
```diff
@@ -125,7 +125,7 @@ def testImportNone(self):
 
     @run_with_all_backends
     def testImporterErrors(self):
-        with mock.patch('corehq.apps.importer.tasks.importer_util.get_spreadsheet', side_effect=ImporterError()):
+        with mock.patch('corehq.apps.case_importer_v1.tasks.importer_util.get_spreadsheet', side_effect=ImporterError()):
```
dimagimon: E501 line too long (121 > 115 characters)

proteusvacuum:
If we're going to make this a convention when rewriting portions of code, could this be a simple first task for the new architecture review process?

czue:
> If we're going to make this a convention when rewriting portions of code, could this be a simple first task for the new architecture review process?

:+1: I had the same reaction

dannyroberts:
Sounds good to me. I'll reach out to the team to initiate a review

dannyroberts:
Ran a test upload locally and that still works. Good to merge from my side!

PR #13852, opened 2016-11-07
Fix admin rpeorts when # domians > 1024

emord:
@nickpell @esoergel 

Would be nice to get this into deploy today. Whenever there are > 1024 domains `ES_MAX_CLAUSE_COUNT` then domains will be `None`

### DIFF
--- a/corehq/apps/hqadmin/reporting/reports.py
+++ b/corehq/apps/hqadmin/reporting/reports.py
@@ -814,7 +814,10 @@ def get_submitted_users(domains):
 
 def get_case_owner_filters(domains):
     mobile_user_ids = list(get_user_ids(True, domains))
-    group_ids = GroupES().domain(domains).get_ids()
+    group_query = GroupES()
+    if domains:
+        group_query = group_query.domain(domains)
+    group_ids = group_query.get_ids()
     return filters.term('owner_id', mobile_user_ids + group_ids)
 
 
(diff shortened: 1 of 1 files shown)


esoergel (approved):
Looks good.  One potential edge case here is `domains == []`.  I'm not sure if/when that could happen, but I could imagine something like:
- select filters which return no domains
- reports end up showing _all_ domains instead of none.

(Non-blocking 'cause I'm not even sure that's possible)

emord:
@esoergel if `domains == []` then `domains_for_es = None` because `if domains` would fail right? not sure what edge case you're talking about?

esoergel:
I was thinking more of the first one:
```python
group_query = GroupES()
if domains:
    group_query = group_query.domain(domains)
group_ids = group_query.get_ids()
```
if `domains == []`, I would expect no groups to be returned, but instead, all groups would be returned.  I'm not familiar enough with the context to know whether that'd be a problem, but figured it was worth flagging just in case.

emord:
Oh I see. If `domains` is not provided then that means that it should be all domains. That should only happen when the # of domains deteremined by the facets in admin reports is > 1024

esoergel:
Yeah, I was thinking of the edge case where domains IS provided, but it's an empty list, as opposed to when it's set to `None`.  Just thinking if there's a semantic distinction between `[]` and `None` from the perspective of the code that calls this.

PR #13853, opened 2016-11-07
Update submodule with UTF-8 

wpride:
Adds fix for http://manage.dimagi.com/default.asp?240409


czue (approved):
pretty sure there's no way this could impact tests but holding off just in case.

PR #13854, opened 2016-11-07
Add 'template="case"' attr to remote-request query element

philomates:
HQ changes for https://github.com/dimagi/commcare-core/pull/440
Can probably be merged whenever because I've verified that it doesn't break case claim on 2.31 

I did not verify these locally since my HQ env isn't up-to-date / I don't know how

@proteusvacuum / @kaapstorm


proteusvacuum:
👍

proteusvacuum:
(for future reference, you can run the test for this with ```./manage.py test ./corehq/apps/app_manager/tests/test_suite_remote_request.py```)

PR #13857, opened 2016-11-07
pass created_by to fix registration

nickpell:
http://manage.dimagi.com/default.asp?241002

Already hotfix-deployed


nickpell:
@millerdev

nickpell:
✅

PR #13859, opened 2016-11-07
suppress errors when there is no case type

nickpell:
http://manage.dimagi.com/default.asp?240972

@millerdev @benrudolph


czue (changes requested):
I assume the assertion was there for a reason, so I think to start it might be better to suppress only when there are no `properties`. if that doesn't fix it I imagine we should setup a `soft_assert` for no case type so we can get to the bottom of what's going on.

PR #13860, opened 2016-11-07
Set trail=False by default for celery tasks

gcapalbo:
One of the times the celery_taskmeta table grew out of control it was because of tasks with lots of children and it was storing information about all the parent-child task relationships. This makes it so that we don't store that info by default, so hopefully will help alleviate the issue.

@snopoke


corehq/apps/users/tasks.py
```diff
@@ -38,7 +39,7 @@ def bulk_upload_async(domain, user_specs, group_specs, location_specs):
     }
 
 
-@task(rate_limit=2, queue='background_queue', ignore_result=True)  # limit this to two bulk saves a second so cloudant has time to reindex
+@hqtask(rate_limit=2, queue='background_queue', ignore_result=True)  # limit this to two bulk saves a second so cloudant has time to reindex
```
dimagimon: E501 line too long (140 > 115 characters)

snopoke:
Can you not use the [CELERY_ANNOTATIONS](http://docs.celeryproject.org/en/2.5-archived/configuration.html#celery-annotations) setting?

gcapalbo:
@snopoke i didn't know that existed. i suppose one downside to using that is that it will ignore whatever value for `trail` is passed to the task itself at definition time, so in order to override it for a task we need to update the settings value. but since for now we want it to be False everywhere, i agree that's the cleaner approach. closing this pr in favor of https://github.com/dimagi/commcare-hq/pull/13874

PR #13867, opened 2016-11-08
Je/upgrade twofactor

emord:
@nickpell 

My changes to the library: https://github.com/Bouke/django-two-factor-auth/pull/178 My branch is based off of an unreleased 1.5.0

### DIFF
--- a/corehq/apps/hqwebapp/two_factor_gateways.py
+++ b/corehq/apps/hqwebapp/two_factor_gateways.py
@@ -29,7 +29,7 @@ class Gateway(object):
     def send_sms(self, device, token):
         message = _('Your authentication token is %s') % token
         self.client.sms.messages.create(
-            to=device.number,
+            to=device.number.as_international,
             from_=self.from_number,
             body=message)
 
@@ -40,7 +40,7 @@ class Gateway(object):
         url = reverse('two_factor:twilio_call_app', kwargs={'token': token})
         url = '%s?%s' % (url, urlencode({'locale': locale}))
         uri = 'https://%s%s' % (Site.objects.get_current().domain, url)
-        self.client.calls.create(to=device.number, from_=self.from_number,
+        self.client.calls.create(to=device.number.as_international, from_=self.from_number,
                                  url=uri, method='GET', if_machine='Hangup', timeout=15)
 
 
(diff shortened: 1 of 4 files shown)


settings.py
```diff
@@ -160,6 +160,7 @@
     'auditcare.middleware.AuditMiddleware',
     'no_exceptions.middleware.NoExceptionsMiddleware',
     'corehq.apps.locations.middleware.LocationAccessMiddleware',
+    'two_factor.middleware.threadlocals.ThreadLocals',
```
emord: I think that this is technically only needed for the twilio gateway, but just went ahead and added in here.

http://django-two-factor-auth.readthedocs.io/en/stable/configuration.html#twilio-gateway

nickpell (approved):
🎉

nickpell:
fyi @calellowitz

emord:
Successfully tested this on staging

corehq/apps/hqwebapp/two_factor_gateways.py
```diff
@@ -29,7 +29,7 @@ def __init__(self):
     def send_sms(self, device, token):
         message = _('Your authentication token is %s') % token
         self.client.sms.messages.create(
-            to=device.number,
+            to=device.number.as_e164,
```
emord: same format as https://github.com/Bouke/django-two-factor-auth/blob/master/two_factor/gateways/twilio/gateway.py#L56

corehq/apps/settings/views.py
```diff
@@ -232,7 +232,7 @@ def dispatch(self, request, *args, **kwargs):
         if not request.couch_user.is_web_user():
             raise Http404
 
-        return super(BaseMyAccountView, self).dispatch(request, *args, **kwargs)
+        return super(MyProjectsList, self).dispatch(request, *args, **kwargs)
```
emord: A lot of these super calls were using the incorrect class

settings.py
```diff
@@ -437,7 +437,7 @@
 
 # after login, django redirects to this URL
 # rather than the default 'accounts/profile'
-LOGIN_REDIRECT_URL = '/'
+LOGIN_REDIRECT_URL = 'homepage'
```
emord: @nickpell I'm wondering if this has something to do with recent url changes. I always got an error locally on redirecting to `/`, but I think `homepage` is the same thing. Does that seem like a recent django change that could have happened?

nickpell: seems possible - bf88d89b touched this

nickpell (approved):

PR #13868, opened 2016-11-08
Add flag to disable CodeMirror (and fix FontAwesome)

esoergel:
@emord

### DIFF
--- a/corehq/apps/hqadmin/templates/hqadmin/raw_couch.html
+++ b/corehq/apps/hqadmin/templates/hqadmin/raw_couch.html
@@ -5,12 +5,10 @@
         <title>Doc {{ doc_id }}</title>
 
         <link type="text/css" rel="stylesheet" media="all" href="{% static 'bootstrap/dist/css/bootstrap.min.css' %}"/>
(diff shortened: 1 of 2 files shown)


emord (approved):
:tight: Code mirror took 1.5 GB of memory when I loaded the icds app

esoergel:
I believe it - &raw=true is a lifesaver for viewing large restores, so it's only sensible to add it here too.

czue:
this seems fine, though I would have expected `raw=true` to just return the JSON of the doc

emord:
agree with cory, but this seems like an improvement for now at least

esoergel:
I thought about that, but a raw json response shows up unformatted in browsers unless you have an extension.  I suppose I could make the page just a large `<pre />` block.

czue:
I think it's a reasonable assumption to assume someone who is using this would have an extension installed or can install one. not a fan of the pre tag since that's not too functionally different from what we have now. the useful thing about the contents actually being the raw contents is being able to save to a file, interact with whatever browser tools you have for JSON, be usable in APIs/scripts, etc..

PR #13870, opened 2016-11-08
suppress errors when there is no case type and properties

biyeun:
A followup to https://github.com/dimagi/commcare-hq/pull/13859
re: http://manage.dimagi.com/default.asp?240972#1243454

@czue does this address your note?

### DIFF
--- a/corehq/apps/case_importer_v1/tasks.py
+++ b/corehq/apps/case_importer_v1/tasks.py
@@ -90,12 +90,13 @@ def do_import(spreadsheet, config, domain, task=None, chunksize=CASEBLOCK_CHUNKS
                         case_type,
                         properties,
                     )
-                _soft_assert = soft_assert(notify_admins=True)
-                _soft_assert(
-                    len(properties) == 0 and not case_type,
-                    'error adding inferred export properties in domain '
-                    '({}): {}'.format(domain, ", ".join(properties))
-                )
+                else:
+                    _soft_assert = soft_assert(notify_admins=True)
+                    _soft_assert(
+                        len(properties) == 0,
+                        'error adding inferred export properties in domain '
+                        '({}): {}'.format(domain, ", ".join(properties))
+                    )
         return err
 
     for i in range(row_count):
--- a/corehq/apps/case_importer_v1/tasks.py
+++ b/corehq/apps/case_importer_v1/tasks.py
@@ -90,7 +90,7 @@ def do_import(spreadsheet, config, domain, task=None, chunksize=CASEBLOCK_CHUNKS
                         case_type,
                         properties,
                     )
-                else:
+                elif len(properties):
                     _soft_assert = soft_assert(notify_admins=True)
                     soft_assert(
                         False,
--- a/corehq/apps/case_importer_v1/tasks.py
+++ b/corehq/apps/case_importer_v1/tasks.py
@@ -13,6 +13,7 @@ from corehq.form_processor.interfaces.dbaccessors import CaseAccessors
 from soil import DownloadBase
 from dimagi.utils.prime_views import prime_views
 from couchdbkit.exceptions import ResourceNotFound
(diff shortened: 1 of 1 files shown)


czue (approved):
yep this looks good. only thought is maybe add a comment saying that we can remove the soft assert if it doesn't ever fire with a date a few months out?

@benrudolph should probably take a look at this when he gets back from vacation but this seems good for now.

czue:
kicked build

corehq/apps/importer/tasks.py
```diff
@@ -82,12 +83,20 @@ def _submit_caseblocks(domain, case_type, caseblocks):
...
-                )
+                if case_type and len(properties):
+                    add_inferred_export_properties.delay(
+                        'CaseImporter',
+                        domain,
+                        case_type,
+                        properties,
+                    )
+                elif len(properties):
+                    _soft_assert = soft_assert(notify_admins=True)
```
czue: actually this is real, the line below should be `_soft_assert`

corehq/apps/importer/tasks.py
```diff
@@ -82,12 +83,20 @@ def _submit_caseblocks(domain, case_type, caseblocks):
...
+                    soft_assert(
+                        False,
```
czue: it would be a little cleaner if you actually replaced `False` with `len(properties) == 0` which is the assertion we are hoping to check, and then we can get rid of the `elif` and use `soft_assert` more like an assertion

czue (changes requested):
sorry just realized @dimagimon was pointing out a real error

biyeun:
@czue updated

czue:
build is failing
```
FAIL: testBasicChunking (corehq.apps.case_importer_v1.tests.ImporterTest)
  File "/mnt/commcare-hq-ro/corehq/apps/case_importer_v1/tests.py", line 318, in testBasicChunking
    res = do_import(file, config, self.domain, chunksize=2)
  File "/mnt/commcare-hq-ro/corehq/apps/case_importer_v1/tasks.py", line 278, in do_import
    _submit_caseblocks(domain, config.case_type, caseblocks)
  File "/mnt/commcare-hq-ro/corehq/apps/case_importer_v1/tasks.py", line 97, in _submit_caseblocks
    '({}): {}'.format(domain, ", ".join(properties))
  File "/mnt/commcare-hq-ro/corehq/util/soft_assert/core.py", line 33, in __call__
    return self._call(assertion, msg, obj)
  File "/mnt/commcare-hq-ro/corehq/util/soft_assert/core.py", line 47, in _call
    raise AssertionError(msg)
AssertionError: error adding inferred export properties in domain (importer-test): location, age, sex
```

corehq/apps/case_importer_v1/tasks.py
```diff
@@ -82,11 +83,18 @@ def _submit_caseblocks(domain, case_type, caseblocks):
...
+                if case_type and len(properties):
+                    add_inferred_export_properties.delay(
+                        'CaseImporter',
+                        domain,
+                        case_type,
+                        properties,
+                    )
+                _soft_assert = soft_assert(notify_admins=True)
+                _soft_assert(
+                    len(properties) == 0 and not case_type,
```
czue: definitely not blocking but I was thinking this might be more readable if it was
```python
if case_type and len(properties):
  add_inferred_export_properties.delay(...)
else:
  _soft_assert(len(properties) == 0, ...)
```
feel free to update if you agree or ignore. i think what you have is functionally equivalent.

czue (approved):

biyeun:
@czue ✅ 

cc @nickpell / @mkangia

czue:
thanks for fighting through that one

benrudolph:
sorry about that, thanks. i didn't expect there to be a situation with no case type / properties. makes me a bit curious about the no case type thing. this looks good

PR #13871, opened 2016-11-08
Remove django-crispy-forms from uninstall reqs

emord:
@nickpell I think this is from a while ago when we moved django-crispy-forms to a submodule, but since you moved it back it's uninstalling and reinstalling on every deploy.

Also I think we could just have this be an empty file now? All of our environments are kept up to date once a week, although that will change with icds, so none of them should ever have these packages. @dimagi/team-commcare-hq anyone have any objections to removing the content of this file?

buddy @esoergel


dannyroberts:
@emord I think it won't cause problems to replace this with an empty file right now, but I wonder what that gets us. Is the idea that it could save time on every deploy, etc. not have to run an uninstall that just does nothing? (Taking out `django-crispy-forms` when we actually do have it installed totally makes sense.) Anyway, I think the bar for clearing this file's pretty low, since like you said it's already been deployed to all our environments and nothing in here is particularly recent, so I'd be pretty easily swayed.

emord:
@dannyroberts I think it doesn't really "get" us anything other than why have something in code that we don't need? It's also been my second time in this file in the past couple of week https://github.com/dimagi/commcare-hq/pull/13612

Anyways I think if we have to add anything to this file from now on we'll have to keep it until we're allowed to update icds again.

dannyroberts:
Makes sense. I'm fine with that then!

millerdev (approved):

PR #13874, opened 2016-11-08
Set trail=False for all celery tasks

gcapalbo:
@snopoke


snopoke (approved):
On the last PR I noticed that some tasks had it set to True. Using this method can you still override it per task? (Just read your comment on #13860. Not possible to override per task).

gcapalbo:
@snopoke yeah there were two tasks that were being used with chain() so i had just set partial=True for them to be safe on that pr. but then i read up on it and did some testing and saw that, at least for that use of it, partial can be False without issue.

PR #13878, opened 2016-11-09
Base64-encode url used in sms install link

gcapalbo:
@snopoke

### DIFF
--- a/corehq/apps/sms/models.py
+++ b/corehq/apps/sms/models.py
@@ -1,4 +1,5 @@
 #!/usr/bin/env python
+import base64
 import jsonfield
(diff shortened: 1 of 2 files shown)


snopoke:
Is this to get around unicode issues?

gcapalbo:
@snopoke no it's a usability issue - we send two sms, one which has an install link used by mobile and one which has a link to sign up. users were getting confused between the two, so this obscures the install link so that only one sms has a clickable link for sign up

snopoke (approved):

gcapalbo:
:white_check_mark:

PR #13880, opened 2016-11-09
Force include locations

proteusvacuum:
Adds the ability to "force sync" specific levels of the location hierarchy without expanding all the locations beneath that level.

https://docs.google.com/document/d/1w2GH-q73QLNMj6I31GVmddxO_qUsvoAASYSATuliS7Q/edit#heading=h.6e4cr4i2mbem
https://trello.com/c/MsU9OGPo/27-tb-location-fixture-force-sync-option
@esoergel 
code buddy: @nickpell


czue:
nice

corehq/apps/locations/models.py
```diff
@@ -114,6 +114,17 @@ class LocationType(models.Model):
     )  # levels below this location type that we start expanding from
     _expand_from_root = models.BooleanField(default=False, db_column='expand_from_root')
     expand_to = models.ForeignKey('self', null=True, related_name='+', on_delete=models.CASCADE)  # levels above this type that are synced
+    _include_without_expanding = models.ForeignKey(
+        'self',
+        null=True,
+        related_name='+',
+        db_column='include_without_expanding',
+        on_delete=models.CASCADE
+    )  # include all leves of this type and their ancestors
```
esoergel: MPTT stores a `level` property on all locations - would it be easier to store an integer here and sync all locations down to level `n`?

corehq/apps/locations/models.py
```diff
@@ -114,6 +114,17 @@ class LocationType(models.Model):
     )  # levels below this location type that we start expanding from
     _expand_from_root = models.BooleanField(default=False, db_column='expand_from_root')
     expand_to = models.ForeignKey('self', null=True, related_name='+', on_delete=models.CASCADE)  # levels above this type that are synced
+    _include_without_expanding = models.ForeignKey(
+        'self',
+        null=True,
+        related_name='+',
+        db_column='include_without_expanding',
+        on_delete=models.CASCADE
```
esoergel: Why cascade?

corehq/apps/locations/views.py
```diff
@@ -212,6 +212,9 @@ def _get_loc_types(self):
             'expand_from': loc_type.expand_from.pk if loc_type.expand_from else None,
             'expand_from_root': loc_type.expand_from_root,
             'expand_to': loc_type.expand_to.pk if loc_type.expand_to else None,
+            'include_without_expanding': (loc_type.include_without_expanding.pk
+                                          if loc_type.include_without_expanding else None),
```
esoergel: nit: I think this is equivalent to `loc_type.include_without_expanding_id`

corehq/apps/locations/models.py
proteusvacuum: good question. Replicated it from above, but I think that might be wrong (although it isn't behaving as I expected this  would with `CASCADE`, where deleting this deletes the referenced model, so I'm a bit confused why I added that in the first place).

corehq/apps/locations/models.py
esoergel: That's a good point.  I agree that it makes more sense to be consistent with `expand_from` and `expand_to`.

corehq/apps/locations/views.py
esoergel: Yeah, it's how django does foreign key references.  When you define a foreign key field (say, `SQLLocation.parent`), Django actually stores the related object's `id` (or `pk` - they're synonyms) as `<field_name>_id`.

* `location.parent.id` looks up the parent and returns the `pk` of the doc, just like how `location.parent.name` would work
* `location.parent_id` *is* the reference to the parent id, so it doesn't need to look up the parent document in the db - it's stored on the `SQLLocation` model.

Aside:
A while back, I wanted to make a `SQLLocation.parent_id` method to mirror the one on `Location`, but django has that attribute set already, so it breaks.  Unfortunately, `SQLLocation.id != SQLLocation.location_id`, so we can't just use that, and we *do* need to do the lookup every time (prefetch_related would help though).  For `LocationType`, though, there's only one id, so it's simpler.

proteusvacuum: Oh, that's cool. I didn't know about that, and actually inadvertently overwrote that functionality for `expand_from` and `expand_to`.

corehq/apps/locations/views.py
```diff
@@ -209,11 +209,11 @@ def _get_loc_types(self):
...
             'code': loc_type.code,
-            'expand_from': loc_type.expand_from.pk if loc_type.expand_from else None,
+            'expand_from': loc_type.expand_from if loc_type.expand_from else None,
             'expand_from_root': loc_type.expand_from_root,
-            'expand_to': loc_type.expand_to.pk if loc_type.expand_to else None,
-            'include_without_expanding': (loc_type.include_without_expanding.pk
-                                          if loc_type.include_without_expanding else None),
+            'expand_to': loc_type.expand_to_id if loc_type.expand_to_id else None,
+            'include_without_expanding': (loc_type.include_without_expanding_id
+                                          if loc_type.include_without_expanding_id else None),
```
esoergel: If it's not set, it'll be `None`, so this line can be just:
```python
'include_without_expanding': loc_type.include_without_expanding_id
```

esoergel (approved):
:boot: :construction_worker_woman:
otherwise g2g

proteusvacuum:
✅

PR #13881, opened 2016-11-09
permit created_by to be blank during validation

nickpell:
fixes payments

@millerdev @proteusvacuum


proteusvacuum:
Looks fine to me... how come that field is being set as blank?

proteusvacuum (approved):

nickpell:
I haven't tracked it down yet ... but ~12K rows are blank so I don't want to error for now

PR #13882, opened 2016-11-09
Update phone number case properties for 99DOTS

proteusvacuum:
@nickpell 
We keep changing the name of these case properties. Hopefully they'll be a bit more stable from now on.


czue (approved):
good idea to make it more future-proof

gcapalbo:
@proteusvacuum i don't have much context here but do we ever expect these cases to be contacts that we want to send SMS to? if so the case property messaging uses for phone number is `contact_phone_number`

proteusvacuum:
@gcapalbo thanks, I didn't know that. Have checked with the app team to see if we need to change this.

PR #13885, opened 2016-11-09
Drop 'Show ' from app summary

orangejenny:
This button bar wraps on screens that are small but not *that* small (i.e., my laptop):
screen shot 2016-11-09 at 5 44 01 pm

Dropping "show" makes it less likely to wrap, and I don't think it takes anything away from understanding/meaning:
screen shot 2016-11-09 at 5 44 18 pm

@dannyroberts


czue:
makes sense to me. build fail not related

PR #13886, opened 2016-11-10
fix change feed / pillowtop docs

czue:
no test coverage


czue:
self merging docs only

PR #13888, opened 2016-11-10
Refactor UCR ES

emord:
less code, more tests, more features

The big change here is extracting the common code in report data sources to a common base class for both sql and es. Along the way I realized that all the code differences didn't necessarily need to be different, so ES now supports the new calculated condtions. I think it may also support mobile ucr, but to be honest haven't looked into it.

I was also able to add more tests and there are now very few db backed tests which aren't on both es and sql (just ones with total rows)

@czue you'll probably be the most familiar/interested

buddy @esoergel

### DIFF
--- a/corehq/apps/es/aggregations.py
+++ b/corehq/apps/es/aggregations.py
@@ -253,6 +253,16 @@ class SumAggregation(Aggregation):
         }
 
 
+class MinAggregation(SumAggregation):
+    """
+    Bucket aggregation that returns the minumum value of a field
+
+    :param name: aggregation name
+    :param field: name of the field to min
+    """
+    type = "min"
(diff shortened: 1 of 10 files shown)


czue:
> less code, more tests, more features

the trifecta!

czue (approved):
Nice this looks great. Clever idea to use multiple inheritance for this.

The idea of such a powerful class being called a Mixin felt a little odd to me though I didn't come up with anything I liked better than that.

emord:
> The idea of such a powerful class being called a Mixin felt a little odd to me though I didn't come up with anything I liked better than that.

Yeah I don't really like it either, but I'm open to other ideas

PR #13892, opened 2016-11-11
show export type in export config page

czue:
@orangejenny http://manage.dimagi.com/default.asp?240997

### DIFF
--- a/corehq/apps/export/models/new.py
+++ b/corehq/apps/export/models/new.py
@@ -517,7 +517,7 @@ class ExportInstance(BlobMixin, Document):
 
     @property
     def is_safe(self):
-        """For compatability with old exports"""
+        """For compatibility with old exports"""
         return self.is_deidentified
 
     @property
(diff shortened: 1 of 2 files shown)


corehq/apps/export/templates/export/customize_export_new.html
```diff
@@ -55,6 +55,18 @@
         <fieldset>
             <div class="form-group">
                 <label
+                    for="export-type"
+                    class="col-sm-3 col-md-2 control-label">
+                    {% if export_instance.type == 'case' %}{% trans "Case Type" %}{% else %}{% trans "Form" %}{% endif %}
```
czue: considered abstracting both the name and value into the classes, but it felt like more trouble than it was worth and not obviously cleaner

corehq/apps/export/models/new.py
```diff
@@ -517,7 +517,7 @@ class Meta:
 
     @property
     def is_safe(self):
-        """For compatability with old exports"""
+        """For compatibility with old exports"""
```
benrudolph: 😅

benrudolph:
nice i've also been frustrated by this

PR #13894, opened 2016-11-11
1.10 updates to urls.py

nickpell:


nickpell:
@emord cc: @kaapstorm

PR #13897, opened 2016-11-11
extra case detail variables

proteusvacuum:
Adds a textbox to allow adding custom detail variables in case lists and case details.
@nickpell 
FYI @ctsims 
screenshot from 2016-11-11 15-32-46


corehq/apps/app_manager/models.py
```diff
@@ -1865,6 +1865,9 @@ class Detail(IndexedSchema, CaseListLookupMixin):
     persist_case_context = BooleanProperty()
     persistent_case_context_xml = StringProperty(default='case_name')
 
+    # Custom variables to add into the <variables /> node
+    custom_variables= StringProperty()
```
dimagimon: E225 missing whitespace around operator

PR #13899, opened 2016-11-11
UCR laboratory

emord:
@czue @snopoke I mentioned this in the scale meeting this morning.

This adds a new backend called laboratory that duplicates data to SQL and ES. Whenever a report is generated (`get_ajax` didn't do anything with exports yet) the report returns results from the SQL backend and then queues a task to compare results between ES and SQL.

:blowfish: / :blowfish: 

buddy @esoergel


corehq/apps/userreports/pillow.py
```diff
@@ -50,20 +50,25 @@ def bootstrap(self, configs=None):
...
 
     def rebuild_tables_if_necessary(self):
+        sql_supported_backends = [UCR_SQL_BACKEND, UCR_LABORATORY_BACKEND]
+        es_supported_backends = [UCR_ES_BACKEND, UCR_LABORATORY_BACKEND]
         self._rebuild_sql_tables(
-            filter(lambda a: get_backend_id(a.config) == UCR_SQL_BACKEND, self.table_adapters))
-        self._rebuild_es_tables(filter(lambda a: get_backend_id(a.config) == UCR_ES_BACKEND, self.table_adapters))
+            filter(lambda a: get_backend_id(a.config) in sql_supported_backends, self.table_adapters))
+        self._rebuild_es_tables(
+                filter(lambda a: get_backend_id(a.config) in es_supported_backends, self.table_adapters))
```
dimagimon: E126 continuation line over-indented for hanging indent

corehq/apps/userreports/laboratory/adapter.py
```diff
@@ -0,0 +1,41 @@
...
+        self.es_adapter.rebuild_table()
+        self.sql_adapter.rebuild_table()
+
+    def drop_table(self):
+        self.es_adapter.drop_table()
+        self.sql_adapter.drop_table()
+
+    def rebuild_table_if_necessary(self):
+        self.es_adapter.rebuild_table_if_necessary()
+        # sql doesn't have this
```
emord: This is the result of a little hackiness right now. Added a todo in the rebuild code to move that to an adapter

corehq/apps/userreports/tasks.py
```diff
@@ -97,3 +103,47 @@ def _iteratively_build_table(config, last_id=None, resume_helper=None):
...
+        return json_response
+
+    spec = get_document_or_not_found(ReportConfiguration, domain, report_config_id)
+    experiment_context = {
+        "domain": domain,
+        "report_config_id": report_config_id,
+        "filter_values": filter_values,
+    }
+    experiment = UCRExperiment(name="UCR DB Experiment", context=experiment_context)
+    with experiment.control() as c:
```
emord: Running the sql results again. I figured this would make it less noisy in case the db changes in between when they look at the UCR and when the task is run

corehq/apps/userreports/reports/view.py
```diff
@@ -398,6 +399,11 @@ def get_ajax(self, params):
         }
         if total_row is not None:
             json_response["total_row"] = total_row
+        if data_source.data_source.config.backend_id == UCR_LABORATORY_BACKEND:
+            compare_ucr_dbs.delay(
+                self.domain, self.report_config_id, self.filter_values,
+                sort_column, sort_order, params
+            )
```
snopoke: Not sure how many of these requests come in but I don't think we should do this for every request.

corehq/apps/userreports/tasks.py
```diff
@@ -97,3 +103,47 @@ def _iteratively_build_table(config, last_id=None, resume_helper=None):
...
+                [(data_source.top_level_columns[int(sort_column)].column_id, sort_order.upper())]
+            )
+
+        datatables_params = DatatablesParams.from_request_dict(params)
+        page = list(data_source.get_data(start=datatables_params.start, limit=datatables_params.count))
+        total_records = data_source.get_total_records()
+        json_response = {
+            'aaData': page,
+            "iTotalRecords": total_records,
+            "iTotalDisplayRecords": total_records,
```
snopoke: Why do we include `total_records` twice?

snopoke (changes requested):
Looks good apart from a few minor questions.

corehq/apps/userreports/reports/view.py
emord: Based on a quick grep of the nginx logs it's around 1k requests/day. I implemented it as a backend so we could possibly slowly scale up these, since it would only happen on data sources that we explicitly set. What do you have in mind for deciding when to do these?

corehq/apps/userreports/tasks.py
emord: Ah. that's some weirdness with the datatables integration, but obviously not needed here. I'll remove the second one.

corehq/apps/userreports/columns.py
```diff
@@ -95,8 +95,8 @@ def _get_expanded_column(data_source_config, report_column, values, lang):
     from corehq.apps.userreports.sql.columns import expand_column as sql_expand_column
     from corehq.apps.userreports.es.columns import expand_column as es_expand_column
     backend_id = get_backend_id(data_source_config)
-    if backend_id == UCR_SQL_BACKEND:
-        fn = sql_expand_column
-    elif backend_id == UCR_ES_BACKEND:
+    if backend_id == UCR_ES_BACKEND:
         fn = es_expand_column
+    else:
+        fn = sql_expand_column
```
esoergel: Should there be a case for `UCR_LABORATORY_BACKEND` here?

snopoke:
I was just thinking something like this:
https://github.com/dimagi/commcare-hq/blob/3646b8825486621bd3b307f4735a442ac3218625/corehq/toggles.py#L142

But as you point out this is only run on selected data sources so probably
not necessary.

emord:
agreed that a deterministic toggle would be better, but this needs to be defined as a backend right now because we should be able to expect both backends to be there. I was thinking of having a management command to migrate data sources from SQL to laboratory as tables/indexes need to be built for every report we want to run this on.

Thoughts on using hte management command approach to randomly assign data sources to laboratory? (that would be a later PR)

snopoke:
I think fine to use a mgmt command but I'm not sure about randomly selecting at this stage. I think to start we should pick a few manually that we know of so that we can monitor more closely.

czue:
took a quick pass and looks good to me. will let @snopoke merge whenever he's happy.

we have a bunch of data sources we could use for this in the TB project if that would be useful.

science! :atom_symbol:

snopoke (approved):

emord:
>  I'm not sure about randomly selecting at this stage

Agree with this, I meant for further down the road. Another blocker is that recently we ran into disk space warning on our ES cluster, so we may want to look into adding another node or space before doing these in batches

snopoke:
ES0 sitting at 55% and growing by 0.17% per day (averaged over last 16 days)
ES1 at 66% and growing by 0.2% per day (averaged over last 16 days)

That gives us about 90 days at current rate (to get to 85%). Probably fine to start testing this on a few data sources now before we need to add a new node. I also don't think we need to migrate everything to ES - it's main use case is for really large tables e.g. for ICDS.

PR #13901, opened 2016-11-14
update checkpoints after 5 mins

snopoke:
On envs that don't have a lot of changes the checkpoints can take a very long time to get updated. This still relies on a change coming through to trigger the update but will still reduce the time between updates.
@gcapalbo


gcapalbo:
build failure looks related to change processing, not sure if it's related to this pr

czue (approved):
i kicked the build but i think it was real (and looking at your last commit it looks like maybe you had intentionally added it to get more info?). 

agree with Giovanni's comment but looks good to me.

gcapalbo:
(looks good to me, but build still failing)

snopoke:
:white_check_mark:

PR #13905, opened 2016-11-14
remove old locations only when available when unsetting locations

mkangia:
http://manage.dimagi.com/default.asp?241177

Breaks when submitted without any locations. Could have handled this at view level but good to have at backend.

buddy @biyeun 
@sravfeyn

### DIFF
--- a/corehq/apps/users/models.py
+++ b/corehq/apps/users/models.py
@@ -1722,14 +1722,14 @@ class CommCareUser(CouchUser, SingleMembershipMixin, CommCareMobileContactMixin)
         """
         from corehq.apps.fixtures.models import UserFixtureType
         from corehq.apps.locations.models import SQLLocation
-
         old_primary_location_id = self.location_id
-        self.assigned_location_ids.remove(old_primary_location_id)
-        self.get_domain_membership(self.domain).assigned_location_ids.remove(old_primary_location_id)
+        if old_primary_location_id:
+            self.assigned_location_ids.remove(old_primary_location_id)
+            self.get_domain_membership(self.domain).assigned_location_ids.remove(old_primary_location_id)
 
         if self.assigned_location_ids:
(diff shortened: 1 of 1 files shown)


corehq/apps/users/models.py
```diff
@@ -1722,14 +1722,14 @@ def unset_location(self, fall_back_to_next=False):
         """
         from corehq.apps.fixtures.models import UserFixtureType
         from corehq.apps.locations.models import SQLLocation
-
         old_primary_location_id = self.location_id
-        self.assigned_location_ids.remove(old_primary_location_id)
-        self.get_domain_membership(self.domain).assigned_location_ids.remove(old_primary_location_id)
+        if old_primary_location_id:
+            self.assigned_location_ids.remove(old_primary_location_id)
+            self.get_domain_membership(self.domain).assigned_location_ids.remove(old_primary_location_id)
```
sravfeyn: Shouldn't this also go under above if block?

mkangia: @sravfeyn  It is... its github messing you up.. 
https://github.com/dimagi/commcare-hq/blob/deb041314b4762d7ccfcfc55c2833d7728cf8108/corehq/apps/users/models.py#L1726

corehq/apps/users/models.py
```diff
@@ -1722,14 +1722,14 @@ def unset_location(self, fall_back_to_next=False):
         """
         from corehq.apps.fixtures.models import UserFixtureType
         from corehq.apps.locations.models import SQLLocation
-
         old_primary_location_id = self.location_id
-        self.assigned_location_ids.remove(old_primary_location_id)
-        self.get_domain_membership(self.domain).assigned_location_ids.remove(old_primary_location_id)
+        if old_primary_location_id:
```
sravfeyn: Ha, 👍

PR #13907, opened 2016-11-14
retry failed save

nickpell:
http://manage.dimagi.com/default.asp?241492

Following the example here: https://github.com/dimagi/commcare-hq/blob/master/corehq/ex-submodules/couchforms/models.py#L214

@proteusvacuum

### DIFF
--- a/corehq/apps/domain/models.py
+++ b/corehq/apps/domain/models.py
@@ -1,6 +1,10 @@
 from datetime import datetime
 from itertools import imap
+import time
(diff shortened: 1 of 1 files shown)


nickpell:
✅

czue:
maybe also add a comment? this feels unintuitive and like code that we will be permanently afraid to delete even though it's kind of hacky (i think i wrote that couchforms version but i would still like to get rid of it if possible)

PR #13908, opened 2016-11-14
location migration

emord:
@nickpell @esoergel 

Assuming it was forgotten about in https://github.com/dimagi/commcare-hq/pull/13452


nickpell (approved):

corehq/apps/locations/migrations/0006_auto_20161114_1858.py
```diff
@@ -0,0 +1,22 @@
...
+
+    dependencies = [
+        ('locations', '0005_locationtype_include_without_expanding'),
+    ]
+
+    operations = [
+        migrations.AlterModelManagers(
+            name='sqllocation',
+            managers=[
+                ('_tree_manager', django.db.models.manager.Manager()),
```
esoergel: Is this autogenerated?  This isn't the model manager we use - I think it should be `corehq.apps.locations.models.LocationManager`.  Then again, I have no idea how this is related to a migration, so maybe a base-class is sufficient?

esoergel (changes requested):

corehq/apps/locations/migrations/0006_auto_20161114_1858.py
emord: Yeah. this was autogenerated. I'll attempt to look into that. It's not clear why it wouldn't have picked that up

emord: Thanks for flagging this.

I think this is actually a bug that was released in 1.9 and it has been fixed in 1.10 https://code.djangoproject.com/ticket/26643 I'm not able to test it though as hq doesn't run on 1.10 yet

I don't think it will cause issues either way. It just changes the manager that is used during migrations

PR #13909, opened 2016-11-14
case search migration

emord:
I think that this should have been included in this commit https://github.com/dimagi/commcare-hq/commit/fdcc595d

@proteusvacuum 

buddy @esoergel


proteusvacuum:
oops. yeah, that seems right.

esoergel (approved):

PR #13911, opened 2016-11-14
Distinguish btwn raw view and no CodeMirror

esoergel:
From comments on https://github.com/dimagi/commcare-hq/pull/13868/files
@emord @czue

I think there IS a use-case for a totally raw json response, so this
adds that.  Additionally, it's useful to disable CodeMirror for really
large documents, where CodeMirror can be quite slow.  For that use-case,
I'd like to keep around the metadata, db statuses, and links, as
available on the normal page.

### DIFF
--- a/corehq/apps/hqadmin/views.py
+++ b/corehq/apps/hqadmin/views.py
@@ -873,13 +873,20 @@ def raw_couch(request):
 def raw_doc(request):
     doc_id = request.GET.get("id")
     db_name = request.GET.get("db_name", None)
-    raw = request.GET.get("raw", False)
     if db_name and "__" in db_name:
         db_name = db_name.split("__")[-1]
     context = _lookup_id_in_database(doc_id, db_name) if doc_id else {}
     other_couch_dbs = sorted(filter(None, couch_config.all_dbs_by_slug.keys()))
     context['all_databases'] = ['commcarehq'] + other_couch_dbs + _SQL_DBS.keys()
-    context['raw'] = raw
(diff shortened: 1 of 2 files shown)


emord (approved):

benrudolph:
yessss i've wanted this for so long

corehq/apps/hqadmin/views.py
```diff
@@ -873,13 +873,20 @@ def raw_couch(request):
...
-    raw = request.GET.get("raw", False)
     if db_name and "__" in db_name:
         db_name = db_name.split("__")[-1]
     context = _lookup_id_in_database(doc_id, db_name) if doc_id else {}
     other_couch_dbs = sorted(filter(None, couch_config.all_dbs_by_slug.keys()))
     context['all_databases'] = ['commcarehq'] + other_couch_dbs + _SQL_DBS.keys()
-    context['raw'] = raw
+    context['use_code_mirror'] = request.GET.get('code_mirror', 'true').lower() == 'true'
+
+    if request.GET.get("raw", False):
```
czue: you could move this up a bit to avoid the work of defining redundant stuff in the context, but don't think it matters much

czue:
thanks for the follow up. agree it's way better to have both.

corehq/apps/hqadmin/views.py
esoergel: fair point: https://github.com/dimagi/commcare-hq/pull/13919

PR #13915, opened 2016-11-15
add function to mapping

czue:
http://manage.dimagi.com/default.asp?241388

introduced in https://github.com/dimagi/commcare-hq/pull/13890

I didn't run this code but am pretty confident it's right


czue:
@sravfeyn / anyone

PR #13916, opened 2016-11-15
Add datadog notification for _send_form_to_hubspot and track_periodi…

mkangia:
…c_data

http://manage.dimagi.com/default.asp?240873

### DIFF
--- a/corehq/apps/analytics/tasks.py
+++ b/corehq/apps/analytics/tasks.py
@@ -408,11 +408,10 @@ def track_periodic_data():
     submit_json = json.dumps(submit)
 
     submit_data_to_hub_and_kiss(submit_json)
-    notify_datadog(
-        { DATADOG_MOBILE_USER_COUNT: number_of_users,
-          DATADOG_EXCEEDING_MAX_FORMS_COUNT: number_of_users_exceeding_max_forms
-        }
-    )
+    notify_datadog({
+        DATADOG_MOBILE_USER_COUNT: number_of_users,
+        DATADOG_EXCEEDING_MAX_FORMS_COUNT: number_of_users_exceeding_max_forms
+    })
 
 
 def submit_data_to_hub_and_kiss(submit_json):
--- a/corehq/apps/analytics/tasks.py
+++ b/corehq/apps/analytics/tasks.py
@@ -22,11 +22,11 @@ from corehq.toggles import deterministic_random
 from corehq.util.decorators import analytics_task
 from corehq.util.soft_assert import soft_assert
 from corehq.util.datadog.utils import (
-    increment_metric,
-    notify_datadog,
-    DATADOG_USER_MAX_FORMS_LIMIT,
-    DATADOG_EXCEEDING_MAX_FORMS_COUNT,
-    DATADOG_MOBILE_USER_COUNT
+    count_by_response_code,
+    update_datadog_metrics,
+    DATADOG_DOMAINS_EXCEEDING_FORMS_GAUGE,
+    DATADOG_WEB_USERS_GAUGE,
+    DATADOG_HUBSPOT_SENT_FORM_METRIC
 )
 
 from dimagi.utils.logging import notify_exception
@@ -50,6 +50,7 @@ HUBSPOT_EXISTING_USER_INVITE_FORM = "7533717e-3095-4072-85ff-96b139bcb147"
 HUBSPOT_CLICKED_SIGNUP_FORM = "06b39b74-62b3-4387-b323-fe256dc92720"
 HUBSPOT_CLICKED_PREVIEW_FORM_ID = "43124a42-972b-479e-a01a-6b92a484f7bc"
 HUBSPOT_COOKIE = 'hubspotutk'
+HUBSPOT_THRESHOLD = 300
 
 
 def _raise_for_urllib3_response(response):
@@ -151,7 +152,6 @@ def _get_client_ip(meta):
     return ip
 
 
-@increment_metric('sent_form_to_hubspot')
 def _send_form_to_hubspot(form_id, webuser, cookies, meta, extra_fields=None, email=False):
     """
     This sends hubspot the user's first and last names and tracks everything they did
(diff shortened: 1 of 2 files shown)


corehq/apps/analytics/tasks.py
```diff
@@ -392,6 +408,11 @@ def track_periodic_data():
     submit_json = json.dumps(submit)
 
     submit_data_to_hub_and_kiss(submit_json)
+    notify_datadog(
+        { DATADOG_MOBILE_USER_COUNT: number_of_users,
```
dimagimon: E201 whitespace after '{'

corehq/apps/analytics/tasks.py
```diff
@@ -392,6 +408,11 @@ def track_periodic_data():
...
+          DATADOG_EXCEEDING_MAX_FORMS_COUNT: number_of_users_exceeding_max_forms
+        }
```
dimagimon: E124 closing bracket does not match visual indentation

corehq/util/datadog/utils.py
```diff
@@ -87,3 +107,8 @@ def get_url_group(url):
         return parts[3] if len(parts) >= 4 else default
 
     return default
+
+
+def notify_datadog(metrics= {}):
```
dimagimon: E251 unexpected spaces around keyword / parameter equals

corehq/apps/analytics/tasks.py
```diff
@@ -144,6 +151,7 @@ def _get_client_ip(meta):
     return ip
 
 
+@increment_metric('sent_form_to_hubspot')
```
snopoke: Instead of wrapping the whole function you could do something like this:
```
@count_by_response_code('commcare.sent_form_to_hubspot')
def _post_to_url(url, data):
    return requests.post(url, data)
```

Doing it this way would capture the status code as well.

snopoke: You should namespace the metrics with `commcare`

corehq/apps/analytics/tasks.py
```diff
@@ -361,6 +375,8 @@ def track_periodic_data():
                 max_workers = domains_to_mobile_users[domain]
 
         project_spaces_created = ", ".join(get_domains_created_by_user(email))
+        if max_forms > DATADOG_USER_MAX_FORMS_LIMIT:
+            number_of_users_exceeding_max_forms += 1
```
snopoke: This isn't a count of users but of domains so should be called something like `domains_with_forms_gt_threshold`

corehq/util/datadog/utils.py
```diff
@@ -13,6 +13,26 @@
 datadog_metric_logger = logging.getLogger('datadog-metrics')
 
 WILDCARD = '*'
+DATADOG_USER_MAX_FORMS_LIMIT = 300
+DATADOG_MOBILE_USER_COUNT = 'commcare.mobile_user_count'
```
snopoke: I'd rename this to `commcare.hubspot.web_users_processed`

corehq/util/datadog/utils.py
```diff
@@ -13,6 +13,26 @@
...
+DATADOG_EXCEEDING_MAX_FORMS_COUNT = 'commcare.users_gt_{}_forms'.format(DATADOG_USER_MAX_FORMS_LIMIT)
```
snopoke: rename to `commcare.hubspot.domains_with_forms_gt_threshold`.

If the threshold changes we don't want to change the metric name.

corehq/util/datadog/utils.py
```diff
@@ -13,6 +13,26 @@
 datadog_metric_logger = logging.getLogger('datadog-metrics')
 
 WILDCARD = '*'
+DATADOG_USER_MAX_FORMS_LIMIT = 300
```
snopoke: I don't think this should be a constant here since this is a hubspot threshold: https://github.com/dimagi/commcare-hq/pull/13916/files#diff-f91ad376200614f22e56c169ab86b9a7R398

corehq/util/datadog/utils.py
snopoke: I'd rename this so give more indication of the purpose: `update_datadog_guage_metric`

corehq/util/datadog/utils.py
```diff
@@ -87,3 +107,8 @@ def get_url_group(url):
         return parts[3] if len(parts) >= 4 else default
 
     return default
+
+
+def notify_datadog(metrics= {}):
+    for metric, value in metrics.items:
```
snopoke: `items()`

snopoke (commented):
Looks good, mostly minor comments.

corehq/util/datadog/utils.py
mkangia: i am looking forward to make this a more generic method and not bound to just `gauge`. Though would definitely update this to `update_datadog_metrics`.

corehq/apps/analytics/tasks.py
mkangia: @snopoke So we are using `commcare.hubspot.domains_with_forms_gt_threshold: domains_with_forms_gt_threshold` but its expected to be reveal users with any domain exceeding 300 forms. Does that sounds alright?

corehq/util/datadog/utils.py
snopoke: yes

corehq/util/datadog/utils.py
```diff
@@ -87,3 +90,8 @@ def get_url_group(url):
         return parts[3] if len(parts) >= 4 else default
 
     return default
+
+
+def update_datadog_metrics(metrics={}):
```
snopoke: `metrics` shouldn't have a default value (and in general you shouldn't use mutable types as default values: see http://docs.python-guide.org/en/latest/writing/gotchas/#mutable-default-arguments)

snopoke (changes requested):
Looks good barring one minor change

snopoke:
Probably worth monitoring this one too: https://github.com/dimagi/commcare-hq/blob/7f4887bc3b7cae1e18c4ca91ca67270dc4a7600e/corehq/apps/analytics/tasks.py#L112-L112

Metric name could be `commcare.hubspot.track_data_post`

corehq/util/datadog/utils.py
mkangia: oh! that bit me again. ruby habit.. getting over it..

snopoke (approved):

snopoke:
:+1: on :white_check_mark:

PR #13921, opened 2016-11-15
add comment clarifying hack

nickpell:
re: https://github.com/dimagi/commcare-hq/pull/13907#issuecomment-260555869

@czue cc: @proteusvacuum


proteusvacuum:
I don't think this is that informative (neither is the ticket)... should the comment be something more in line with https://github.com/dimagi/commcare-hq/blob/master/corehq/ex-submodules/couchforms/models.py#L214 ? or is there another reason?

nickpell:
@proteusvacuum I was going for some kind of permanent reference in case the code changed.  I added a link to the commit in github that introduced the fix at https://github.com/dimagi/commcare-hq/blob/master/corehq/ex-submodules/couchforms/models.py#L214

czue:
thanks for follow up

PR #13923, opened 2016-11-15
Kill map-based demo reports

orangejenny:
@snopoke suggested these might be killable, and the only people with the flag on are Andrea, Rushika, and Biyeun, so they seem pretty safe to kill.

@dannyroberts


snopoke:
:+1: @dimagi/product should confirm

snopoke:
Is it possible to remove any of the base classes too or are they still in use elsewhere?

orangejenny:
@snopoke Good point. Added a commit to remove `GenericCaseListMap`; `GenericMapReport` is used by other reports.

czue:
oh man, good call. 

i think we can also likely delete 'tests/maps_demo/mountains.csv' and `tests/maps_demo/india.geojson`. 

this framework and these demos were actually super cool, too bad we never figured out a way to productize them properly (and i agree better to just kill until we actually have a need/path to do that)

orangejenny:
I am enjoying watching the red count increase as y'all make suggestions.

snopoke:
We also have map reports in UCR so not a complete loss.

orangejenny:
tests passing

PR #13926, opened 2016-11-15
Use default url tag

emord:
@nickpell needed for django 1.9

```RemovedInDjango19Warning: Loading the `url` tag from the `future` library is deprecated and will be removed in Django 1.9. Use the default `url` tag instead.```

buddies @dannyroberts @proteusvacuum


proteusvacuum:
The `future` is now! 👍

nickpell (approved):

PR #13928, opened 2016-11-15
Create excel utility that works for xls or xlsx

dannyroberts:
@emord @proteusvacuum 

This came out requirements for case importer and the following problems

- I need to be able to accept the previous format (which was in xls) as well as a new format (which will be in xlsx)
- Current importer has an annoying date problem that wasn't solved very well; I wanted to have a clean, well-tested interface behind which I could deal with any typing quirks, and present the same thing to the user whether it was xls or xlsx
- both `openpyxl` (for xlsx) and `xlrd` have quirky and sprawling interfaces, largely because they're very comprehensive of excel is feature-packed. Since we use very little of the complete feature-set, I wanted very simple intermediate objects I could use instead that are easier to write code and tests against.

Let me know what you think! I went with the v1/v2 naming again, though it's not 100% the same situation as before. `spreadsheets_v1` has a previous attempt at exposing a spreadsheet reader that's written at a different level of abstraction and that I ended up feeling solved the wrong problem.


corehq/util/spreadsheets_v2/tests/test_spreadsheets.py
```diff
@@ -0,0 +1,64 @@
+from datetime import datetime
+from itertools import izip_longest
+import os
+from django.test import SimpleTestCase
+import openpyxl
```
dimagimon: F401 'openpyxl' imported but unused

emord:
Looks good to me. Is it the intention that this new library will become the only excel library HQ should use?

proteusvacuum:
seems fine to me too

dannyroberts:
> Is it the intention that this new library will become the only excel library HQ should use?

I think practically some of the interfaces are too tangled to change to use this one, but `case_importer_v1` for example could probably switch to using this with reasonably low effort. This utility is also read-only, so isn't useful for creating excel downloads.

dannyroberts:
OK, this is good to go from my end. The v2 code isn't used yet, and v1 name change isn't causing problems (I tested a fixture upload, which uses `spreadsheets_v1`).

dannyroberts:
I also pushed another commit (https://github.com/dimagi/commcare-hq/pull/13928/commits/3f7c8e4cc3b8ddfdab67ed3669251d3cb221785e) since I think you last saw

dannyroberts:
Wee!

PR #13929, opened 2016-11-15
organized bower dependencies

emord:
Tried to group plugins together, so angular, jquery, knockout backbone are all together.

@proteusvacuum @dannyroberts


benrudolph:
nice

PR #13931, opened 2016-11-16
make ExcelFile interface more similar to openpyxl.Worksheet

dannyroberts:
Dipping my toes into case importer code; trying to see if some of the more egregious problems can be fixed in-place before the rewrite. If they can, then that'll be a really good start for backwards compatibility, etc. and people will get the easy wins upfront.

This PR massages case importer's internal `ExcelFile` class to be a little more similar to the `openpyxl.Worksheet` and the interface I'm using in https://github.com/dimagi/commcare-hq/pull/13928.


emord:
Build fail was unrelated. looking good to me once it passes

dannyroberts:
✅  and good to go. Tested basic upload on staging without issue!

PR #13936, opened 2016-11-16
Ensure that appId is included

benrudolph:
@wpride fixes the refresh issue. appid was null so it wasn't actually deleting any dbs
http://manage.dimagi.com/default.asp?241145
cc: @mkangia


ctsims:
@benrudolph Is this what likely was resulting in the experience Matt L and I were seeing where the app wouldn't update?

benrudolph:
It's what was causing app preview to lose non oqps. But, I tracked down the
bug you are referring to and that should be merged and fixed

PR #13937, opened 2016-11-16
get rid of removed in 1.10 BaseCommand.option_list

nickpell:
@emord cc: @kaapstorm


emord (approved):

snopoke:
is `options_list` still respected at all in 1.10? The documented approach is to override the `def add_arguments(self, parser):` method.

nickpell:
@snopoke I'm not sure but I'll keep that in mind

PR #13938, opened 2016-11-16
Fix Bulk Translations Tests

biyeun:
The conclusion was that the strings used in the test were removed from CommCare purposefully in the update to 2.31.0 and that they were likely selected at random to appear in the test. new strings were selected and a note has been added.

@dannyroberts @mkangia @nickpell


nickpell (approved):

biyeun:
@nickpell ✅

PR #13939, opened 2016-11-16
hide unused case importer options

dannyroberts:
@sheelio and I suspect that no one uses either of these options intentionally and are harmful and confusing to everyone else, and as a half-measure before deleting the features entirely (which will be glorious), we're hiding them to see if anyone complains.

The two options being removed are

- Option to upload without headers (it them makes you map column numbers to case properties):
  screen shot 2016-11-16 at 3 14 56 pm

- Option to have one row (case id, property name, property value) per property being uploaded instead of one row per case.
  screen shot 2016-11-16 at 3 15 42 pm

They're basically both just ways to make the process of uploading even :poop:ier

### DIFF
--- a/corehq/apps/case_importer_v1/templates/case_importer_v1/excel_config.html
+++ b/corehq/apps/case_importer_v1/templates/case_importer_v1/excel_config.html
@@ -129,7 +129,7 @@
             </div>
         </fieldset>
 
-        <fieldset>
+        <fieldset class="hide">
             <legend>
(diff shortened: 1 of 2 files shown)


dannyroberts:
@proteusvacuum @emord

dannyroberts:
Got some context from Kieran, the key/value map feature was originally to support a workflow for khayelitsha, which has stopped submitting data. @amsagoff signed off on this plan, and I'll email field-dev@

dannyroberts:
Cancelled build since no tests are relevant

PR #13940, opened 2016-11-16
calculations.first_domain_for_user failing

orangejenny:
Followup for https://github.com/dimagi/commcare-hq/pull/13890

@czue / @dannyroberts

PR #13941, opened 2016-11-16
Fix first_domain_for_user

benrudolph:
@orangejenny the old string domain instead of the object. this is again where it'd be nice to have a convention for domain object. just going with what i see most of the time

cc: @mkangia


orangejenny:
hey, thanks for looking into this. just PRed https://github.com/dimagi/commcare-hq/pull/13940

benrudolph:
👍

PR #13942, opened 2016-11-16
[App Manager V2 Split 004] COPY over additional V1 templates to V2

biyeun:
- only v2 templates modified
- no `.py` files touched
- made sure that older copies were updated to match any changes in master since I last did any work in app maanger

@orangejenny @benrudolph


benrudolph (approved):
Seems fine to me. hard to review closely, but trust your local testing. 👍 on ✅

biyeun:
@benrudolph ✅

PR #13944, opened 2016-11-16
make case_importer_v1 infer types

dannyroberts:
Get rid of "Data Types" column and the need for users to specify property types. The data's already in the excel file, so let's just use that.

Before release:
- [ ] We'll have to update the docs [here](https://confluence.dimagi.com/display/commcarepublic/Mapping+Case+Properties+and+Completing+the+Import) as well

### DIFF
--- a/corehq/apps/case_importer_v1/util.py
+++ b/corehq/apps/case_importer_v1/util.py
@@ -14,8 +14,6 @@ from corehq.apps.case_importer_v1.exceptions import (
     ImporterFileNotFound,
     ImporterRefError,
     InvalidCustomFieldNameException,
-    InvalidDateException,
-    InvalidIntegerException,
     ImporterError)
 from corehq.apps.users.cases import get_wrapped_owner
 from corehq.apps.users.models import CouchUser
@@ -39,7 +37,6 @@ class ImporterConfig(namedtuple('ImporterConfig', [
     'excel_fields',
     'case_fields',
     'custom_fields',
-    'type_fields',
     'search_column',
     'key_column',
     'value_column',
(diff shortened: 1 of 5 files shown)


dannyroberts:
FWIW I tested this on staging with a row that included a number and a date, and that worked fine.

dannyroberts:
@sheelio @proteusvacuum @emord If we're ready to go would be worth announcing since it's user-facing cc @amsagoff

czue:
are there documentation updates that need to change too?

corehq/apps/case_importer_v1/util.py
```diff
@@ -139,10 +135,16 @@ def _row_values(self, sheet, index):
         return [self._fmt_value(cell) for cell in sheet.row(index)]
 
     def _fmt_value(self, cell):
-        # Explicitly format integers, since xlrd treats all numbers as decimals (adds ".0")
-        if cell.ctype == xlrd.XL_CELL_NUMBER and int(cell.value) == cell.value:
-            return int(cell.value)
-        return cell.value
+        if cell.ctype == xlrd.XL_CELL_NUMBER:
+            if int(cell.value) == cell.value:
```
proteusvacuum: doesn't this also cast `1.0` to an int? also, do we do anything differently with `int`s and `floats` in cases?

dannyroberts: That's both currently the case and desired. If you type "1" into a cell in excel, this library will give you '1.0' as its value. In order to have the string "1" appear in the case property—almost always what's desired and unlikely to present a problem even if you did want a float—we cast to int if that doesn't change the numerical value (i.e. doesn't drop precision after the `.`).

proteusvacuum: I see, this is just a formatting thing and doesn't actually change the way the data behaves.

proteusvacuum (approved):

dannyroberts:
@czue thanks for bringing up docs, I updated the description to include that.

dannyroberts:
This is earlier than I intended this to be merged, but I think it's fine. I've tested it and I can update the docs tomorrow before 3pm deploy.

proteusvacuum:
Oh sorry, didn't see the hold off on merge flag, maybe was a bit over zealous

dannyroberts:
No worries, I should have added a tag if I was more concerned. I just updated the docs, so we should be good to go!

PR #13896, opened 2016-11-11
Shadow module form selection: handle child modules

orangejenny:
http://manage.dimagi.com/default.asp?240834

I got bogged down in the suite.xml changes, particularly how form exclusion works with menu mode, for quite some time, so I wrote tests for all of the variants of menu mode in an app with a parent module, a child module, and a shadow module which has the parent as its source.

@dannyroberts / @kaapstorm

PR #13946, opened 2016-11-17
Limit include-able forms to the source module and its children

orangejenny:
In https://github.com/dimagi/commcare-hq/pull/13896, I misunderstood exactly which modules were available in `shadow-module-settings.js` and ended up displaying all of the app's forms as options to exclude from the shadow module. The list of forms should include come from only the source module and any children of the source module.

@czue / @dannyroberts

### DIFF
--- a/corehq/apps/app_manager/static/app_manager/js/shadow-module-settings.js
+++ b/corehq/apps/app_manager/static/app_manager/js/shadow-module-settings.js
@@ -12,20 +12,27 @@ hqDefine('app_manager/js/shadow-module-settings.js', function () {
             self.selectedModule = ko.pureComputed(function () {
                 return _.findWhere(self.modules(), {uniqueId: self.selectedModuleId()});
             });
-            self.allForms = ko.pureComputed(function() {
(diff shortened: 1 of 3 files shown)


corehq/apps/app_manager/static/app_manager/js/shadow-module-settings.js
```diff
@@ -12,20 +12,27 @@ hqDefine('app_manager/js/shadow-module-settings.js', function () {
             self.selectedModule = ko.pureComputed(function () {
                 return _.findWhere(self.modules(), {uniqueId: self.selectedModuleId()});
             });
-            self.allForms = ko.pureComputed(function() {
-                return _.flatten(_.map(self.modules(), function(m) { return m.forms(); }));
+            // Find all forms that could potentially be included in the current module:
+            // Forms belonging to the source module and to the source module's children
```
czue: thanks for the comment. stared at the code for a while trying to figure out what was going on.

czue (approved):

PR #13951, opened 2016-11-17
Test domain calculated properties

orangejenny:
Would have caught the issues fixed by https://github.com/dimagi/commcare-hq/pull/13915 and https://github.com/dimagi/commcare-hq/pull/13940

@czue / @dannyroberts

### DIFF
--- a/corehq/apps/domain/calculations.py
+++ b/corehq/apps/domain/calculations.py
@@ -313,7 +313,7 @@ def dom_calc(calc_tag, dom, extra_arg=''):
     return ans
 
 
-def _all_domain_stats():
+def all_domain_stats():
     webuser_counts = defaultdict(lambda: 0)
     commcare_counts = defaultdict(lambda: 0)
 
@@ -332,7 +332,7 @@ def _all_domain_stats():
     }
 
 
-def _calced_props(dom, id, all_stats):
+def calced_props(dom, id, all_stats):
     return {
         "_id": id,
         "cp_n_web_users": int(all_stats["web_users"].get(dom, 0)),
--- a/corehq/apps/domain/calculations.py
+++ b/corehq/apps/domain/calculations.py
@@ -332,6 +332,51 @@ def _all_domain_stats():
     }
 
 
+def _calced_props(dom, id, all_stats):
+    return {
+        "_id": id,
+        "cp_n_web_users": int(all_stats["web_users"].get(dom, 0)),
+        "cp_n_active_cc_users": int(CALC_FNS["mobile_users"](dom)),
+        "cp_n_cc_users": int(all_stats["commcare_users"].get(dom, 0)),
+        "cp_n_active_cases": int(CALC_FNS["cases_in_last"](dom, 120)),
+        "cp_n_users_submitted_form": total_distinct_users([dom]),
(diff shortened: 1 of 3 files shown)


corehq/apps/domain/tests/test_domain_calculated_properties.py
```diff
@@ -0,0 +1,20 @@
+from __future__ import print_function, unicode_literals
+
+from django.test import TestCase
+
+from corehq.apps.domain.models import Domain
+from corehq.apps.domain.calculations import _all_domain_stats, _calced_props
+
+class DomainCalculatedPropertiesTest(TestCase):
```
dimagimon: E302 expected 2 blank lines, found 1

corehq/apps/domain/calculations.py
```diff
@@ -332,6 +332,51 @@ def _all_domain_stats():
     }
 
 
+def _calced_props(dom, id, all_stats):
```
orangejenny: Underscore prefix seems weird, since this is imported elsewhere, but it's patterned after `_all_domain_stats`.

czue: agree it's weird. we could also change both to non-underscores

czue (approved):
this is great, thanks for adding

corehq/apps/reports/tasks.py
```diff
@@ -169,55 +170,14 @@ def rebuild_export_async(config, schema):
...
-                "cp_n_sms_out_30_d": int(CALC_FNS["sms_out_in_last"](dom, 30)),
-                "cp_n_sms_out_60_d": int(CALC_FNS["sms_out_in_last"](dom, 60)),
-                "cp_n_sms_out_90_d": int(CALC_FNS["sms_out_in_last"](dom, 90)),
-                "cp_n_j2me_30_d": int(CALC_FNS["j2me_forms_in_last"](dom, 30)),
-                "cp_n_j2me_60_d": int(CALC_FNS["j2me_forms_in_last"](dom, 60)),
-                "cp_n_j2me_90_d": int(CALC_FNS["j2me_forms_in_last"](dom, 90)),
-                "cp_j2me_90_d_bool": int(CALC_FNS["j2me_forms_in_last_bool"](dom, 90)),
-                "cp_300th_form": CALC_FNS["300th_form_submission"](dom)
-            }
+            calced_props = calced_props(dom, r["_id"], all_stats)
```
dimagimon: F823 local variable 'calced_props' (defined in enclosing scope on line 34) referenced before assignment

orangejenny:
Note to self: need to fake FormES, SMSES, etc.

https://github.com/dimagi/commcare-hq/blob/master/corehq/apps/es/fake/README.md

czue:
the other (easier?) option would be to just create the indices in the `setUp` function, doing something like [this](https://github.com/dimagi/commcare-hq/blob/master/corehq/apps/data_analytics/tests/test_esaccessors.py#L20).

orangejenny:
@czue thank you, didn't know about that, will check it out

orangejenny:
@czue Fixed tests (that was much easier than I'd expected it to be, so thanks), good to go once tests pass on travis.

czue (approved):
thanks!

PR #13952, opened 2016-11-17
Migrate new export toggle to old one

benrudolph:
@czue i'm going to commit this now, then run on all environments, then merge code that'll switch the new export toggle to this old export toggle. since the number of old exports is static, (unless i manually migrate a domain) we shouldn't have to worry about timing. killing build

cc: @mkangia


snopoke (approved):

czue:
sounds like a plan

PR #13957, opened 2016-11-17
Data Dictionary

emord:
Beginning of backend work for a data dictionary.

:blowfish: or :office: 

buddies @proteusvacuum @dannyroberts

### DIFF
--- a/corehq/apps/data_dictionary/models.py
+++ b/corehq/apps/data_dictionary/models.py
@@ -1,0 +1,38 @@
+from django.db import models
+
+PROPERTY_TYPE_CHOICES = (
+    ('Date', 'Date'),
+    ('Plain', 'Plain'),
+    ('Number', 'Number'),
+    ('Select', 'Select'),
+    ('Integer', 'Integer'),
+    ('', 'No Type Currently Selected')
+)
+
+
+class CaseType(models.Model):
+    domain = models.CharField(max_length=255)
+    name = models.CharField(max_length=255)
+
+    class Meta:
+        unique_together = ('domain', 'name')
+
+
+class CaseProperty(models.Model):
+    case_type = models.ForeignKey(
+        CaseType,
+        on_delete=models.CASCADE,
+        related_name='properties',
+        related_query_name='property'
+    )
+    name = models.CharField(max_length=255)
+    description = models.TextField()
+    deprecated = models.BooleanField(default=False)
+    type = models.CharField(
+        choices=PROPERTY_TYPE_CHOICES,
+        max_length=20,
+        default=''
+    )
+
+    class Meta:
+        unique_together = ('case_type', 'name')
--- a/corehq/apps/data_dictionary/models.py
+++ b/corehq/apps/data_dictionary/models.py
@@ -1,18 +1,19 @@
 from django.db import models
 
 PROPERTY_TYPE_CHOICES = (
-    ('Date', 'Date'),
-    ('Plain', 'Plain'),
-    ('Number', 'Number'),
-    ('Select', 'Select'),
-    ('Integer', 'Integer'),
+    ('date', 'Date'),
+    ('plain', 'Plain'),
+    ('number', 'Number'),
+    ('select', 'Select'),
+    ('integer', 'Integer'),
     ('', 'No Type Currently Selected')
 )
 
 
 class CaseType(models.Model):
-    domain = models.CharField(max_length=255)
-    name = models.CharField(max_length=255)
+    domain = models.CharField(max_length=255, default=None)
+    name = models.CharField(max_length=255, default=None)
+    description = models.TextField(default='')
 
     class Meta:
         unique_together = ('domain', 'name')
@@ -25,8 +26,8 @@ class CaseProperty(models.Model):
         related_name='properties',
         related_query_name='property'
     )
-    name = models.CharField(max_length=255)
-    description = models.TextField()
+    name = models.CharField(max_length=255, default=None)
+    description = models.TextField(default='')
     deprecated = models.BooleanField(default=False)
     type = models.CharField(
         choices=PROPERTY_TYPE_CHOICES,
--- a/corehq/apps/data_dictionary/static/data_dictionary/js/data_dictionary.js
+++ b/corehq/apps/data_dictionary/static/data_dictionary/js/data_dictionary.js
@@ -48,7 +48,7 @@
         }, this);
 
         this.goToCaseType = function (caseType) {
-            self.activeCaseType(caseType);
+            self.activeCaseType(caseType.name());
         };
     };
 
--- a/corehq/apps/data_dictionary/static/data_dictionary/js/data_dictionary.js
+++ b/corehq/apps/data_dictionary/static/data_dictionary/js/data_dictionary.js
@@ -25,10 +25,10 @@
         self.init = function () {
             $.getJSON(dataUrl)
             .done(function (data) {
-                _.each(data.properties, function (property) {
-                    var caseType = new CaseType(property.type);
-                    caseType.init(property.properties);
-                    self.caseTypes.push(caseType);
+                _.each(data.case_types, function (caseType) {
+                    var caseTypeObj = new CaseType(caseType.name);
+                    caseTypeObj.init(caseType.properties);
+                    self.caseTypes.push(caseTypeObj);
                 });
                 self.activeCaseType(self.caseTypes()[0].name());
             });
--- a/corehq/apps/data_dictionary/static/data_dictionary/js/data_dictionary.js
+++ b/corehq/apps/data_dictionary/static/data_dictionary/js/data_dictionary.js
@@ -1,0 +1,61 @@
+(function ($, _) {
+    var CaseProperty = function (data) {
+        this.name = ko.observable(data.name);
+        this.type = ko.observable(data.type);
+        this.description = ko.observable(data.description);
+    };
+
+    var CaseType = function (name) {
+        var self = this;
+        self.name = ko.observable(name);
+        self.properties = ko.observableArray();
+
+        self.init = function (properties) {
+            _.each(properties, function (property) {
+                self.properties.push(new CaseProperty(property));
+            });
+        };
+    };
+
+    var DataDictionaryModel = function (dataUrl) {
+        var self = this;
+        self.caseTypes = ko.observableArray();
+        self.activeCaseType = ko.observable();
+
+        self.init = function () {
+            $.getJSON(dataUrl)
+            .done(function (data) {
+                _.each(data.properties, function (property) {
+                    var caseType = new CaseType(property.type);
+                    caseType.init(property.properties);
+                    self.caseTypes.push(caseType);
+                });
+                self.activeCaseType(self.caseTypes()[0].name());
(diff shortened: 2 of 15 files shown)


corehq/apps/data_dictionary/util.py
```diff
@@ -0,0 +1,54 @@
...
+    current_properties = {}
+    if CaseType.objects.filter(domain=domain).count() != 0:
+        props = CaseProperty.objects.filter(case_type__domain=domain).prefetch_related('case_type')
+        for prop in props:
+            case_type = prop.case_type.name
+            if case_type not in current_properties:
+                current_properties[case_type] = {prop.name}
+            else:
+                current_properties[case_type].add(prop.name)
+    current_case_types = current_properties.keys()
```
dimagimon: F841 local variable 'current_case_types' is assigned to but never used

corehq/apps/data_dictionary/util.py
```diff
@@ -0,0 +1,54 @@
...
+
+    properties = _get_all_case_properties(domain)
+    new_case_properties = []
+    new_case_types = []
```
dimagimon: F841 local variable 'new_case_types' is assigned to but never used

dannyroberts:
@emord can you link me to info about this feature?

emord:
@dannyroberts whoops. meant to link https://docs.google.com/document/d/1ZkcVcEouijoOic9St2cVHFeckbwjD3ftYaXiPXnbtEM/edit#heading=h.y0mj508sv0wx

corehq/apps/data_dictionary/models.py
```diff
@@ -0,0 +1,38 @@
+from django.db import models
+
+PROPERTY_TYPE_CHOICES = (
+    ('Date', 'Date'),
```
snopoke: generally prefer lowercase for the option values

corehq/apps/data_dictionary/models.py
```diff
@@ -0,0 +1,38 @@
...
+    ('Plain', 'Plain'),
+    ('Number', 'Number'),
+    ('Select', 'Select'),
+    ('Integer', 'Integer'),
+    ('', 'No Type Currently Selected')
+)
+
+
+class CaseType(models.Model):
+    domain = models.CharField(max_length=255)
```
snopoke: Should add `default=None` to force integrity errors if no domain is supplied otherwise it will save with domain=''. Same for other non-nullable char fields

corehq/apps/data_dictionary/models.py
```diff
@@ -0,0 +1,38 @@
...
+    name = models.CharField(max_length=255)
+
```
snopoke: Could be good to add a description text property

corehq/apps/data_dictionary/static/data_dictionary/js/data_dictionary.js
```diff
@@ -0,0 +1,61 @@
...
+
+    var DataDictionaryModel = function (dataUrl) {
+        var self = this;
+        self.caseTypes = ko.observableArray();
+        self.activeCaseType = ko.observable();
+
+        self.init = function () {
+            $.getJSON(dataUrl)
+            .done(function (data) {
+                _.each(data.properties, function (property) {
```
snopoke: a bit weird that `data.properties` is actually case types and `property` is actually a case type

corehq/apps/data_dictionary/tests/test_util.py
```diff
@@ -0,0 +1,60 @@
+from django.test import TestCase
+from mock import patch
+
+from corehq.apps.data_dictionary.models import CaseType, CaseProperty
+from corehq.apps.data_dictionary.util import generate_data_dictionary
+
+
+class GenerateDictionaryTest(TestCase):
+    domain = 'data-dictionary'
```
snopoke: I've started making test domains UUID's which makes it harder to get into a bad state, especially when using REUSE_DB.

corehq/apps/data_dictionary/tests/test_util.py
```diff
@@ -0,0 +1,60 @@
...
+
+    @patch('corehq.apps.data_dictionary.util._get_all_case_properties')
+    def test_two_types(self, mock):
+        mock.return_value = {'type': ['property'], 'type2': ['property']}
+        generate_data_dictionary(self.domain)
+
+        self.assertEqual(CaseType.objects.filter(domain=self.domain).count(), 2)
+        self.assertEqual(CaseProperty.objects.filter(case_type__domain=self.domain).count(), 2)
+
+    @patch('corehq.apps.data_dictionary.util._get_all_case_properties')
```
snopoke: If you're going to path this for every test you can do a class level pach

corehq/apps/data_dictionary/util.py
```diff
@@ -0,0 +1,52 @@
...
+    properties = _get_all_case_properties(domain)
+    new_case_properties = []
+    for case_type, properties in properties.items():
+        if not case_type:
+            continue
+
+        case_type_obj, _ = CaseType.objects.get_or_create(domain=domain, name=case_type)
+
+        for prop in properties:
+            if (not current_properties or
```
snopoke: don't think you need this check since you're initializing `current_properties` as an empty dict so `case_type not in current_properties` will work fine

corehq/apps/data_dictionary/util.py
```diff
@@ -0,0 +1,52 @@
...
+            else:
+                current_properties[case_type].add(prop.name)
+
+    properties = _get_all_case_properties(domain)
+    new_case_properties = []
+    for case_type, properties in properties.items():
+        if not case_type:
+            continue
+
+        case_type_obj, _ = CaseType.objects.get_or_create(domain=domain, name=case_type)
```
snopoke: could keep a dict of case types fetched above and only do this call if we don't already have the case type (which is only if the case type doesn't exist or has no properties)

corehq/apps/data_dictionary/util.py
```diff
@@ -0,0 +1,52 @@
...
+
+    CaseProperty.objects.bulk_create(new_case_properties)
+
+
+def _get_all_case_properties(domain):
+    case_type_to_properties = {}
+
+    for case_type in get_case_types_from_apps(domain):
+        properties = set()
+        schema = CaseExportDataSchema.generate_schema_from_builds(domain, None, case_type)
```
snopoke: Interesting that we're using the export schema here to generate the dictionary - I would have expected that the data dictionary would be the authoritative source of case info.

corehq/apps/data_dictionary/views.py
```diff
@@ -0,0 +1,74 @@
...
+from corehq.apps.data_dictionary import util
+from corehq.apps.data_dictionary.models import CaseType, CaseProperty
+from corehq.tabs.tabclasses import ApplicationsTab
+
+
+@login_and_domain_required
+@toggles.DATA_DICTIONARY.required_decorator()
+def generate_data_dictionary(request, domain):
+    util.generate_data_dictionary(domain)
+    return JsonResponse({"status": "success"})
```
snopoke: Why does this need to be exposed as a view?

snopoke (commented):
Great to see some this happening! My main thought is that I was thinking that this data model would be updated in real time when the user makes changes to the app (e.g. save a form with new case management) instead of on request for the entire app. 

Interested to hear your thoughts on why you took this approach.

corehq/apps/data_dictionary/util.py
emord: I think that case exports is the only place in HQ that contains all case properties, I figured it was quickest way to bootstrap the dictionary

corehq/apps/data_dictionary/views.py
emord: It provides a quick way to test regenerating the dictionary if you think something is missing. I think this will be taken out in the final form of this

emord:
> My main thought is that I was thinking that this data model would be updated in real time when the user makes changes to the app (e.g. save a form with new case management) instead of on request for the entire app.

That is definitely the plan. I was putting this smaller PR out there because it's a good first step and way easier to get feedback on the db model I chose

corehq/apps/data_dictionary/util.py
benrudolph: just an fyi on this, when you pass in `None` it only gets all case properties from latest build (not the latest save)

corehq/apps/data_dictionary/util.py
```diff
@@ -0,0 +1,52 @@
...
+        for group_schema in schema.group_schemas:
+            for item in group_schema.items:
+                if item.tag:
+                    name = item.tag
```
benrudolph: why use `tag` as the name? won't this be something generic like `tag`

corehq/apps/data_dictionary/util.py
emord: https://github.com/dimagi/commcare-hq/pull/13957/commits/9e2c4af6d2a405365c4cf0426c7e280af80c56dc

It still runs into the issue where you have n db queries for case types that don't exist, but not worried about that at the moment.

corehq/apps/data_dictionary/tests/test_util.py
emord: Even though they all have different return values?

emord:
Added some commits to add properties on case importing and app changes

corehq/apps/data_dictionary/tests/test_util.py
emord: Thanks. I didn't know you could that classes could be decorated. https://github.com/dimagi/commcare-hq/pull/13957/commits/0003e557125f05949f3bd9d4c42384b744348c34

emord:
not going to add any more functionality to this PR as it's already > 500 lines. It's mostly self contained to the new app. I'll add any more functionality to follow ups

corehq/apps/data_dictionary/templates/data_dictionary/base.html
```diff
@@ -0,0 +1,51 @@
...
+        <li data-bind="css: { active: $data.name() == $root.activeCaseType() }">
+            <a href="#" data-bind="text: $data.name(), click: $root.goToCaseType"></a>
+        </li>
+    </ul>
+{% endblock %}
+
+{% block page_content %}
+    <table id="data-dictionary-table" class="table">
+        <thead>
+            <th>Case Property</th>
```
proteusvacuum: `trans`?

corehq/apps/data_dictionary/util.py
```diff
@@ -0,0 +1,78 @@
...
+            properties[case_type].add(prop.name)
+
+    return case_types, properties
+
+
+def add_properties_to_data_dictionary(domain, case_type, properties):
+    case_type_obj, created = CaseType.objects.get_or_create(domain=domain, name=case_type)
+    if not created:
+        old_properties = CaseProperty.objects.filter(case_type=case_type_obj)
+        old_properties = {p.name for p in old_properties}
```
proteusvacuum: I think you could do a `values_list('name', flat=True)` here to just pull out the names

proteusvacuum (approved):

proteusvacuum:
Nothing blocking, so merging

PR #13960, opened 2016-11-17
Try multiple langs before giving up

benrudolph:
@biyeun / @mkangia  http://manage.dimagi.com/default.asp?240793 in this ticket the user had modules that were named in french only and since their user preference was english it would show Unknown Module. this PR ensures that if we can't find the name for the user's language we try the default app language. every time i go into this code i cringe... 😖


biyeun (approved):
Looks good to me. waiting on ✅

PR #11088, opened 2016-04-04
Inline editing

orangejenny:
Opening for feedback from @dimagi/js-team and @dimagi/product  Won't be ready to merge until
- [x] Knockout gets upgraded to 3.3 (currently on staging: https://github.com/dimagi/commcare-hq/pull/11003)
- [x] Widget supports multiple languages, so that app/module/form names can use it
- [x] App/module/form name and comment get moved to the top of those pages, above the accordions, as described in spec (https://docs.google.com/document/d/1BkHdxoTAd66wr23eFHJb_b5ZT1ZzDzjK8MO5E6VyfOs/edit)
- [x] switch app/form/module names & comments to ajax save (?)
- [x] UAT (http://manage.dimagi.com/default.asp?223336)

Read only view:
screen shot 2016-04-04 at 4 06 31 pm

On hover:
screen shot 2016-04-04 at 4 06 36 pm

On click:
screen shot 2016-04-04 at 4 06 45 pm

When text is blank (there's also an option for placeholder text):
screen shot 2016-04-04 at 4 06 55 pm

Read-only on deploy page:
screen shot 2016-04-04 at 4 07 17 pm

Editing on deploy page:
screen shot 2016-04-04 at 4 07 29 pm

Works with help icons and language badges:
screen shot 2016-04-09 at 10 35 07 am
(I don't love the help icon alignment here, but putting it to the left of the pencil messes up the interaction, since both the help icon and the widget as a whole respond to clicks. Will keep thinking about it, and I'm open to other suggestions.)

PR #13963, opened 2016-11-17
Remove app name & comments from app settings configuration

orangejenny:
http://manage.dimagi.com/default.asp?241908

When moving the app name and comment out of the main section and into the header in https://github.com/dimagi/commcare-hq/pull/11088, I kept them in settings.yaml and created a "none" widget so they wouldn't show. But leaving them as settings means that when someone updates name/comment and then saves the whole settings page, saving "all" settings ends up overwriting the new name/comment.

Verified that name and comment still save and that this change doesn't affect the app's build files.

@czue


czue:
cool, the app manager diff tests caught this!
```

    self.partial_v2_dir
AssertionError: 
```

czue (approved):
sneaky bug - looks good on :white_check_mark:

orangejenny:
yeah, very glad to have the app manager diff tests, @biyeun thank you for putting those together!

PR #13964, opened 2016-11-17
Ensure displayOptions stick

benrudolph:
@biyeun @wpride surprisingly tricky bug. we were passing displayOptions by reference into form entry and then wrapping it into a ko model. the formplayer code wasn't aware of this and it caused some conditions to be evaluated incorrectly. this does a deep clone of displayOptions so we don't have this issue anymore. also fixes minor argument mistake

cc: @mkangia


biyeun (approved):
looks good. waiting on ✅

wpride:
👍 nice catch

PR #13965, opened 2016-11-17
[App Manager v2 #5] Copies the rest of the app manager v1 templates into v2

biyeun:
This copies the rest of the app manager v1 templates into v2.

This PR does not touch any python code (except for one small change for the app settings yaml), it is only copying the templates over and creating the diffs.
The only templates modified are v2 and are not visible to the user.

@benrudolph @orangejenny


benrudolph:
👍

PR #13967, opened 2016-11-18
make case importer accept xls AND xlsx

dannyroberts:
@proteusvacuum @emord

This one's a bit bigger, but the individual commits are self-contained (🐡)


corehq/util/test_utils.py
```diff
@@ -504,3 +504,20 @@ def update_case(domain, case_id, case_properties, user_id=None):
     post_case_blocks(
         [CaseBlock(**kwargs).as_xml()], domain=domain
     )
+
+
+def make_make_path(current_directory):
```
proteusvacuum: lol

corehq/util/spreadsheets_v2/adapters/generic.py
```diff
@@ -0,0 +1,17 @@
...
+@contextmanager
+def open_any_workbook(filename):
+    if filename.endswith('.xls'):
+        with open_xls_workbook(filename) as workbook:
+            yield workbook
+    elif filename.endswith('.xlsx'):
+        with open_xlsx_workbook(filename) as workbook:
+            yield workbook
+    else:
+        raise SpreadsheetFileExtError('File {} does not end in .xls or .xlsx'
```
proteusvacuum: should this be translated?

corehq/util/spreadsheets_v2/adapters/xlsx.py
```diff
@@ -4,16 +4,36 @@
...
         try:
             openpyxl_workbook = openpyxl.load_workbook(f, read_only=True, data_only=True)
-        except (BadZipfile, InvalidFileException) as e:
-            raise SpreadsheetFileError(e.message)
+        except InvalidFileException as e:
+            raise SpreadsheetFileInvalidError(e.message)
+        except BadZipfile as e:
+            f.seek(0)
+            if f.read(8) == XLSX_ENCRYPTED_MARKER:
+                raise SpreadsheetFileEncrypted(u'Workbook is encrypted')
```
proteusvacuum: should this be translated?

corehq/apps/case_importer_v1/util.py
```diff
@@ -104,131 +104,40 @@ def open_workbook(filename=None, **kwargs):
...
     def get_header_columns(self):
-        if self.worksheet.max_row > 0:
-            return self.iter_rows().next()
+        if self.max_row > 0:
+            if self._column_headers:
+                return self.iter_rows().next()
+            else:
+                columns = []
+                for i in range(self.max_row):
+                    columns.append("Column {!d}".format(i))
```
proteusvacuum: I don't think `!d` is a [valid conversion](https://docs.python.org/2.7/library/string.html#formatstrings) . Since `i` is an integer anyway, I think you can just do `{}`
Additional nit, you could use an array comprehension for this.

corehq/apps/case_importer_v1/util.py
```diff
@@ -126,10 +126,7 @@ def get_header_columns(self):
             if self._column_headers:
                 return self.iter_rows().next()
             else:
-                columns = []
-                for i in range(self.max_row):
-                    columns.append("Column {!d}".format(i))
-                return columns
+                return ["Column {}".format(i) for i in range(self.max_row)]
```
proteusvacuum: oh, I see you changed it.

proteusvacuum (approved):
Nothing blocking, looking good!

corehq/apps/case_importer_v1/util.py
dannyroberts: yup, sorry for the fake-out

corehq/util/spreadsheets_v2/adapters/generic.py
dannyroberts: Nope, I don't think so. These errors are all caught in the client code (i.e. case importer) and converted to error messages to be shown to the user. I'm generally opposed to the practice of showing an exception message to users as a sort of blanket ban.

This particular Exception never gets explicitly caught in case importer, but the xls/xlsx is explicitly checked for before this gets called, and if that ever changes this would 500 and still not be shown to the user.

proteusvacuum: 👍 Thanks for the explanation.

emord (approved):

dannyroberts:
@emord thanks for reviewing those last few commits. I think I'm going to freeze this unless I find any bugs and start working on another branch for future stuff

dannyroberts:
@proteusvacuum @emord this is good to go, after some QA (done by me) on staging. Tested create and update on 26 cases with external id, using an xlsx file, and also did the same update with xls just as a sanity check. (The way I wrote and tested it, I'm fairly confident that xlsx and xls present a 100% identical interface to the client code, with the possible exception of edge-case bugs I haven't run into yet like the excel "0 is 1899-12-30" bug we just haven't stumbled upon yet, which we would have already been not covering before this change.)

Talked to sheel and we're happy deploying this today!

PR #13971, opened 2016-11-18
properly serve json as json

czue:
http://manage.dimagi.com/default.asp?241989

this was fun to stumble on.

@proteusvacuum @orangejenny


proteusvacuum:
nice!

PR #13974, opened 2016-11-18
Add New Expressions for Case History

sheelio:
This adds some additional UCR extension expressions for accessing the case history.  Its for something that Neal and I are testing ( not for use by most projects).

### DIFF
--- a/custom/ucr_ext/expressions.py
+++ b/custom/ucr_ext/expressions.py
@@ -536,7 +536,7 @@ def get_last_case_property_update(spec, context):
                 }
             }
         },
-        'value_expression':{
+        'value_expression': {
             'type': 'property_path',
             'property_path': [
                 'updated_unknown_properties',
--- a/custom/ucr_ext/expressions.py
+++ b/custom/ucr_ext/expressions.py
@@ -17,6 +17,8 @@ CUSTOM_UCR_EXPRESSIONS = [
     ('ext_parent_id', 'custom.ucr_ext.expressions.parent_id'),
     ('ext_open_in_month', 'custom.ucr_ext.expressions.open_in_month'),
     ('ext_get_case_forms_by_date', 'custom.ucr_ext.expressions.get_case_forms_by_date'),
+    ('ext_get_case_history_by_date', 'custom.ucr_ext.expressions.get_case_history_by_date'),
+    ('ext_get_last_case_property_update', 'custom.ucr_ext.expressions.get_last_case_property_update'),
 ]
 
 
@@ -50,6 +52,21 @@ class GetCaseFormsByDateSpec(JsonObject):
     start_date = DefaultProperty(required=False)
     end_date = DefaultProperty(required=False)
     form_filter = DefaultProperty(required=False)
+
+
+class GetCaseHistoryByDateSpec(JsonObject):
+    type = TypeProperty('ext_get_case_history_by_date')
+    start_date = DefaultProperty(required=False)
+    end_date = DefaultProperty(required=False)
+    filter = DefaultProperty(required=False)
+
+
+class GetLastCasePropertyUpdateSpec(JsonObject):
+    type = TypeProperty('ext_get_last_case_property_update')
+    case_property = StringProperty(required=True)
+    start_date = DefaultProperty(required=False)
+    end_date = DefaultProperty(required=False)
+    filter = DefaultProperty(required=False)
 
 
 def diff_calendar_months(spec, context):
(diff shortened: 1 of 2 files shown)


custom/ucr_ext/tests.py
```diff
@@ -379,3 +379,295 @@ def test_form_filter(self):
...
+        })
+        self.assertEqual(
+            2,
+            expression(
+                {"some_item": "item_value"},
+                context=EvaluationContext(self.case, 0),
+            )
+        )
+
+class TestGetLastCasePropertyUpdateTest(TestCase):
```
dimagimon: E302 expected 2 blank lines, found 1

czue (approved):
I believe that it should, but have you verified that this will work with both the couch and sql backends? basically just need to confirm the action JSON format is shareable/consistent across the two.

custom/ucr_ext/tests.py
```diff
@@ -379,3 +379,296 @@ def test_form_filter(self):
...
+                },
+                {
+                    'action_type': 'update',
+                    'date': datetime(2015, 3, 4),
+                    'updated_unknown_properties': {
+                        'foo': 'c'
+                    },
+                },
+            ]
+        }
```
snopoke: The JSON representation of SQL case actions is different and doesn't include any of the properties that changed. The SQL one only has `date`,`server_date`, `xform_id`, `sync_log_id`.

sheelio:
@snopoke @czue 

Is it intended that the SQL case docs don't have the full case history?  If so, is there a way to make a separate database call to get that case history (similar to the get_case_forms expression that exists). 

For these monthly indicator tables that Neal and I are testing, its a lot easier to work with the case history vs. the form data.  

(I could also just use these expressions with just couch-backed domains until we've validated the idea and then figure out how to deal with SQL-backed domains if we choose to move forward).

snopoke:
@sheelio we made the decision not to store the full case action in the DB due to the addition storage requirements and because at the time it wasn't being used anywhere. We can still get that data but it's not very efficient - it requires loading all the forms and pulling out the case changes.

If this is experimental then I'd recommend going ahead as you suggest, testing on project that are still on the Couch backend.

czue:
@sheelio related doc expressions using the xform_id from the actions should still work. it would be more painful to reference the properties but would definitely be doable on SQL

custom/ucr_ext/expressions.py
```diff
@@ -404,3 +421,127 @@ def get_case_forms_by_date(spec, context):
...
+                'datatype': 'array',
+                'type': 'ext_root_property_name',
+                'property_name': 'actions'
+            }
+        }
+    return ExpressionFactory.from_spec(spec, context)
+
+
+def get_last_case_property_update(spec, context):
+    GetLastCasePropertyUpdateSpec.wrap(spec)
```
czue: maybe you can add an assertion that the domain is on a couch backend here since this won't work in SQL domains.

czue: and I think it would be possible to write a `get_last_case_property_update_sql` expression - at which point you could remove the assertion. but as @snopoke mentioned it would not be very performant.

sheelio: Cool - happy to add that assertion.  Can you point me to how to check that?

czue: you should be able to use [`should_use_sql_backend`](https://github.com/dimagi/commcare-hq/blob/master/corehq/form_processor/utils/general.py#L29)

sheelio:
@czue 
Refactored the expression so it just extracted the case block from the case's forms.  Its probably not performant but should be ok for me to test with.

custom/ucr_ext/expressions.py
```diff
@@ -54,9 +55,51 @@ class GetCaseFormsByDateSpec(JsonObject):
...
+
+        case_history = []
+        for f in forms:
+            case_history.append(_get_case_block(case_id, f))
+
+        context.set_cache_value(cache_key, case_history)
+        return case_history
+
+
+def _get_case_block(case_id, obj):
```
czue: i think you might want to leverage [`_extract_case_blocks`](https://github.com/dimagi/commcare-hq/blob/master/corehq/ex-submodules/casexml/apps/case/xform.py#L346) for this, which would hopefully be more maintainable and better tested (might involve some small refactors to get it in the format you want)

czue (changes requested):
cool you were able to do this so easily. one comment about code sharing but otherwise seems good. let me know if that exercise doesn't seem like it's going to be easy/right.

czue (approved):
nice

PR #13975, opened 2016-11-18
allow advanced shadow modules

orangejenny:
http://manage.dimagi.com/default.asp?240834

I don't think there's a specific reason that shadow modules can't be advanced modules, it was just the most conservative design decision to make at the time. It's possible that there's some situation with advanced modules that won't work well with shadow modules...but I'm not concerned about this PR causing significant increased support, because when/if shadow modules misbehave, there's always the workaround "don't use shadow modules, make a full copy of the module you want to shadow."

@czue / @kaapstorm

### DIFF
--- a/corehq/apps/app_manager/views/modules.py
+++ b/corehq/apps/app_manager/views/modules.py
@@ -207,7 +207,7 @@ def _get_shadow_module_view_context(app, module, lang=None):
         }
 
     return {
-        'modules': [get_mod_dict(m) for m in app.modules if m.module_type == 'basic'],
+        'modules': [get_mod_dict(m) for m in app.modules if m.module_type != 'shadow'],
         'excluded_form_ids': module.excluded_form_ids,
     }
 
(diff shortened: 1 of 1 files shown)


corehq/apps/app_manager/views/modules.py
```diff
@@ -207,7 +207,7 @@ def get_mod_dict(mod):
         }
 
     return {
-        'modules': [get_mod_dict(m) for m in app.modules if m.module_type == 'basic'],
+        'modules': [get_mod_dict(m) for m in app.modules if m.module_type != 'shadow'],
```
czue: are there any concerns about other module types? (careplan, report)

czue (approved):
looks good. feel free to self merge if my question is not an issue.

corehq/apps/app_manager/views/modules.py
orangejenny: good point, shadow modules don't work with reporting modules. not sure about care plan, but don't think it's worth the time/effort to investigate. fixed.

proteusvacuum:
Oh wow, cool! I think this might simplify a lot of the eNikshay case list filters :)

orangejenny:
caveat emptor!

(but as explained, i think this will work fine)

PR #6944, opened 2015-06-03
don't remove deleted (from the database) documents from UCR tables

czue:
this is mainly being done for performance reasons. the rationale is that
there is no way for users to delete forms or cases from the database
so this only happens in exceptional circumstances. in those exceptional
circumstances it also seems reasonable to make users rebuild their
ucr tables.

@TylerSheffels @snopoke @nickpell seem reasonable?

PR #13976, opened 2016-11-18
Add SQL-backed document store for locations UCR

esoergel:
@sravfeyn
@czue
Is this really all it takes to make UCR not need couch for locations?  I tested it locally and it works as expected on a forced rebuild, but I'm not sure if that gets docs in a different way than the steady-state process.  Marking as don't merge 'till I can get a conclusive answer on that.

### DIFF
--- a/corehq/apps/userreports/expressions/specs.py
+++ b/corehq/apps/userreports/expressions/specs.py
@@ -187,17 +187,22 @@ class RelatedDocExpressionSpec(JsonObject):
 
         self._doc_id_expression = doc_id_expression
         self._value_expression = value_expression
-        # used in caching
-        self._vary_on = json.dumps(self.value_expression, sort_keys=True)
 
     def __call__(self, item, context=None):
         doc_id = self._doc_id_expression(item, context)
         if doc_id:
             return self.get_value(doc_id, context)
 
-    @quickcache(['self._vary_on', 'doc_id'])
-    def get_value(self, doc_id, context):
+    def _vary_on(self, doc_id, context):
+        # For caching. Gets called with the same params as get_value.
+        return [
+            context.root_doc['domain'],
+            doc_id,
+            json.dumps(self.value_expression, sort_keys=True)
+        ]
 
+    @quickcache(_vary_on)
+    def get_value(self, doc_id, context):
         try:
             assert context.root_doc['domain']
             document_store = get_document_store(context.root_doc['domain'], self.related_doc_type)
--- a/corehq/apps/userreports/expressions/specs.py
+++ b/corehq/apps/userreports/expressions/specs.py
@@ -180,7 +180,9 @@ class RelatedDocExpressionSpec(JsonObject):
     value_expression = DictProperty(required=True)
 
     def configure(self, doc_id_expression, value_expression):
-        if get_db_by_doc_type(self.related_doc_type) is None:
+        non_couch_doc_types = ('Location')
+        if (self.related_doc_type not in non_couch_doc_types
+                and get_db_by_doc_type(self.related_doc_type) is None):
             raise BadSpecError(u'Cannot determine database for document type {}!'.format(self.related_doc_type))
 
         self._doc_id_expression = doc_id_expression
--- a/corehq/apps/userreports/expressions/specs.py
+++ b/corehq/apps/userreports/expressions/specs.py
@@ -1,5 +1,6 @@
 import json
 from simpleeval import InvalidExpression
+from corehq.apps.locations.document_store import LOCATION_DOC_TYPE
 from corehq.apps.userreports.document_stores import get_document_store
 from corehq.apps.userreports.exceptions import BadSpecError
 from corehq.form_processor.interfaces.dbaccessors import CaseAccessors
@@ -180,7 +181,7 @@ class RelatedDocExpressionSpec(JsonObject):
     value_expression = DictProperty(required=True)
 
     def configure(self, doc_id_expression, value_expression):
-        non_couch_doc_types = ('Location')
+        non_couch_doc_types = (LOCATION_DOC_TYPE,)
         if (self.related_doc_type not in non_couch_doc_types
                 and get_db_by_doc_type(self.related_doc_type) is None):
             raise BadSpecError(u'Cannot determine database for document type {}!'.format(self.related_doc_type))
--- a/corehq/apps/userreports/tests/test_location_data_source.py
+++ b/corehq/apps/userreports/tests/test_location_data_source.py
@@ -1,0 +1,105 @@
+import uuid
+from django.test import TestCase
+from pillowtop.feed.interface import Change
+
+from corehq.apps.domain.shortcuts import create_domain
+from corehq.apps.locations.document_store import get_location_change_meta
+from corehq.apps.locations.models import SQLLocation, LocationType
+from corehq.apps.locations.tests.util import delete_all_locations
+
+from corehq.apps.userreports.app_manager import _clean_table_name
+from corehq.apps.userreports.models import DataSourceConfiguration
+from corehq.apps.userreports.pillow import get_kafka_ucr_pillow
+from corehq.apps.userreports.tasks import rebuild_indicators
+from corehq.apps.userreports.util import get_indicator_adapter
+
+
+class TestLocationDataSource(TestCase):
+    domain = "delos_corp"
+
+    def setUp(self):
+        delete_all_locations()
+        self.domain_obj = create_domain(self.domain)
+
+        self.region = LocationType.objects.create(domain=self.domain, name="region")
+        self.town = LocationType.objects.create(domain=self.domain, name="town", parent_type=self.region)
+
+        self.data_source_config = DataSourceConfiguration(
+            domain=self.domain,
+            display_name='Locations in Westworld',
+            referenced_doc_type='Location',
+            table_id=_clean_table_name(self.domain, str(uuid.uuid4().hex)),
+            configured_filter={},
+            configured_indicators=[{
+                "type": "expression",
+                "expression": {
+                    "type": "property_name",
+                    "property_name": "name"
+                },
+                "column_id": "location_name",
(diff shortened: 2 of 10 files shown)


czue:
> Is this really all it takes to make UCR not need couch for locations?

i think almost! i believe this is ~all that's needed to support `Location`s in `RelatedDoc` expressions which should cover 95% of the use cases. Not sure if you're 100% confident in the json parity, but if you are I think we can merge this as soon as tests are passing.

Unfortunately there is also the issue of data sources built on top of locations - a feature flag that I think is only used by ICDS. In order to port that over I think we'll need to come up with a solution to making sure updates get into the data source. That will be a bit bigger of an undertaking - basically you'll need to publish a location change event to kafka whenever locations are saved in a way that can be picked up and processed by the UCR pillow. There is some documentation [here](http://commcare-hq.readthedocs.io/change_feeds.html#from-sql) that might be a good starter - I can provide more specifics if that's not enough.

build fail looks related
```
ERROR: test_location_parent_bad_domain (corehq.apps.locations.tests.test_location_expressions.TestLocationParentIdExpression)
  File "/mnt/commcare-hq-ro/corehq/apps/locations/tests/test_location_expressions.py", line 121, in test_location_parent_bad_domain
    self.expression({'location_id': self.child._id}, EvaluationContext({"domain": 'bad-domain'}))
  File "/mnt/commcare-hq-ro/corehq/apps/userreports/expressions/specs.py", line 194, in __call__
    return self.get_value(doc_id, context)
  File "/mnt/commcare-hq-ro/corehq/util/quickcache.py", line 312, in inner
  File "/mnt/commcare-hq-ro/corehq/util/quickcache.py", line 100, in __call__
    content = self.fn(*args, **kwargs)
  File "/mnt/commcare-hq-ro/corehq/apps/userreports/expressions/specs.py", line 202, in get_value
    doc = document_store.get_document(doc_id)
  File "/mnt/commcare-hq-ro/corehq/apps/userreports/document_stores.py", line 18, in get_document
    return self.queryset.get(location_id=doc_id).to_json()
  File "/vendor/lib/python2.7/site-packages/django/db/models/query.py", line 328, in get
    num = len(clone)
  File "/vendor/lib/python2.7/site-packages/django/db/models/query.py", line 144, in __len__
    self._fetch_all()
  File "/vendor/lib/python2.7/site-packages/django/db/models/query.py", line 965, in _fetch_all
    self._result_cache = list(self.iterator())
  File "/vendor/lib/python2.7/site-packages/django/db/models/query.py", line 238, in iterator
    results = compiler.execute_sql()
  File "/vendor/lib/python2.7/site-packages/django/db/models/sql/compiler.py", line 838, in execute_sql
    cursor = self.connection.cursor()
  File "/vendor/lib/python2.7/site-packages/django/db/backends/base/base.py", line 164, in cursor
    cursor = self.make_cursor(self._cursor())
  File "/vendor/lib/python2.7/site-packages/django/db/backends/base/base.py", line 135, in _cursor
    self.ensure_connection()
  File "/vendor/lib/python2.7/site-packages/django/db/backends/base/base.py", line 130, in ensure_connection
    self.connect()
  File "/vendor/lib/python2.7/site-packages/django/db/utils.py", line 98, in __exit__
  File "/vendor/lib/python2.7/site-packages/django/db/backends/base/base.py", line 130, in ensure_connection
    self.connect()
  File "/vendor/lib/python2.7/site-packages/django/db/backends/base/base.py", line 119, in connect
    self.connection = self.get_new_connection(conn_params)
  File "/vendor/lib/python2.7/site-packages/django/db/backends/postgresql_psycopg2/base.py", line 176, in get_new_connection
    connection = Database.connect(**conn_params)
  File "/vendor/lib/python2.7/site-packages/psycopg2/__init__.py", line 164, in connect
    conn = _connect(dsn, connection_factory=connection_factory, async=async)
OperationalError: FATAL:  database "test_commcarehq" does not exist
```

esoergel:
@czue I figured there'd be a missing piece from this unless it just did a full rebuild every time.  The changes publisher vs the document store makes sense after seeing the docs now, I was thinking of those as part of the same entity.  I'll take a crack at the change publishers listed in the docs and let you know if I run into any issues.

Regarding the json being the same, I did write a basic test comparing the two.  I think I'll try to get just [`to_json`](https://github.com/dimagi/commcare-hq/pull/13993/files) deployed, then compare it in a django shell session with a larger, real data set to make sure thare aren't any unanticipated differences.  The main one is the `last_modified` field:

sql: `'2016-10-18T00:58:33.362638'`
couch: `'2016-10-18T00:58:33.403285Z'`

They're slightly different times (because they're both automatic fields that happen on the respective saves).  The couch one has a slightly different format (it has UTC specified), but we don't actually store tz info, so I'm not sure that's correct.  I spoke with Sheel, and he says we don't need that field for ICDS reports anyways.

czue:
cool, sounds good, thanks. in UCR we just throw that date at postgres and hope it deals with it anyways. might be good to double-check the timezone behavior but my guess is it's fine (and the small difference definitely seems unimportant)

esoergel:
@czue this is on staging now.  I'm verifying that it works by making changes to locations and viewing them on this [location-based UCR](https://staging.commcarehq.org/a/ethan-test/configurable_reports/data_sources/preview/373ef1586c4fbfbdfc4ee039c08d2124/) I made for the purpose (feel free to play around on that domain if you like).

After that, additions and updates started showing up just fine, but deletions don't appear to be removing those rows from the data source, as I'd expect.  I do see the change count increment on the system info page, and I've confirmed locally that the signal is being fired.  This does appear to be [expected](https://github.com/dimagi/commcare-hq/blob/master/corehq/apps/userreports/pillow.py#L114-L118) (as discussed in https://github.com/dimagi/commcare-hq/pull/6944), so I think we're okay there.

I had to restart the pillow process although I'm not quite clear on why (I was getting `LeaderNotAvailableError: TopicMetadata(topic='location', error=5, partitions=[])`, but then stopped after a simple restart).  Any idea what that could've been about?

czue:
I think that would fail the first time if you hadn't run `./manage.py create_kafka_topics`. I wonder if that command should be run during deploys to prevent this?

esoergel:
This should be g2g after https://github.com/dimagi/commcare-hq/pull/14096 is merged.
I tested out the `to_json()` stuff on prod and found a bunch of crap data from past migrations and old formats - stuff that shouldn't have showed up in there anyways.  There weren't unanticipated differences in "real" fields aside from `last_modified`, which we already discussed.  I did see a couple locations where `lineage` didn't match, and my guess is that's due to bad data that needs cleaning up.  SQL is the source of truth for that, so I trust it over couch anyways.

esoergel:
ready! @czue @NoahCarnahan

corehq/apps/userreports/expressions/specs.py
```diff
@@ -180,22 +180,29 @@ class RelatedDocExpressionSpec(JsonObject):
     value_expression = DictProperty(required=True)
 
     def configure(self, doc_id_expression, value_expression):
-        if get_db_by_doc_type(self.related_doc_type) is None:
+        non_couch_doc_types = ('Location')
```
czue: nit: should this be using `LOCATION_DOC_TYPE`?

corehq/ex-submodules/pillowtop/dao/interface.py
```diff
@@ -20,7 +20,7 @@ def save_document(self, doc_id, document):
     def delete_document(self, doc_id):
         pass
 
-    def iter_document_ids(self):
+    def iter_document_ids(self, last_id=None):
```
czue: :+1: good catch

czue (commented):
this looks good. my only major thought is that it'd be nice to have an integration test to end-to-end a location into UCR. did you look into that and decide it was too much trouble to implement?

esoergel:
@czue Just added some tests.

czue (commented):
not blocking but i think this doesn't actually test that changes are properly published, does it? it could be nice to test just saving a location and then running the pillow off of whatever is in the kafka feed as opposed to manually processing things (though I think both are valuable). i don't think that would be too hard, but let me know if you disagree.

corehq/apps/userreports/tests/test_location_data_source.py
```diff
@@ -0,0 +1,99 @@
+import uuid
+from django.test import TestCase
+from kafka.common import KafkaUnavailableError
+from pillowtop.feed.interface import Change
```
dimagimon: F401 'Change' imported but unused

corehq/apps/userreports/tests/test_location_data_source.py
```diff
@@ -0,0 +1,99 @@
...
+
+from corehq.apps.domain.shortcuts import create_domain
+from corehq.apps.locations.document_store import get_location_change_meta
```
dimagimon: F401 'get_location_change_meta' imported but unused

corehq/apps/userreports/tests/test_location_data_source.py
```diff
@@ -0,0 +1,99 @@
...
+        sweetwater = self._make_loc("Sweetwater", self.town)
+        las_mudas = self._make_loc("Las Mudas", self.town)
+
+        rebuild_indicators(self.data_source_config._id)
+
+        self.assertDataSourceAccurate(["Westworld", "Sweetwater", "Las Mudas"])
+
+        # Insert new location
+        since = self.pillow.get_change_feed().get_current_offsets()
+        blood_arroyo = self._make_loc("Blood Arroyo", self.town)
```
dimagimon: F841 local variable 'blood_arroyo' is assigned to but never used

esoergel:
@czue good call, that was actually a lot easier than I expected - I thought I had to do a bunch of set-up stuff for that to work.

czue:
:athletic_shoe: :hammer:

czue (approved):
thanks for following up on everything and glad it worked out! :tada:  on :white_check_mark:

PR #13977, opened 2016-11-18
Populate case list/detail select2 with current value

orangejenny:
http://manage.dimagi.com/default.asp?241998

Looks like this is the only one of the recent jquery autocomplete => select2 conversions that doesn't use the `typeahead` binding (which takes care of this at https://github.com/dimagi/commcare-hq/blob/master/corehq/apps/style/static/style/ko/knockout_bindings.ko.js#L589-L594) and also stores a value that might not be in the given list of options.

@czue


czue (approved):
nice thanks. would have taken me a while to find where to make this change.

PR #13981, opened 2016-11-20
The experiment is complete

wpride:
cross-request: https://github.com/dimagi/touchforms/pull/275


benrudolph:
test failure real

emord:
I think you can also remove https://github.com/dimagi/commcare-hq/blob/master/settings.py#L116-L117

PR #13982, opened 2016-11-20
Remove formplayer backend

wpride:
cross-request: https://github.com/dimagi/touchforms/pull/276

### DIFF
--- a/corehq/apps/cloudcare/static/cloudcare/js/form_entry/webformsession.js
+++ b/corehq/apps/cloudcare/static/cloudcare/js/form_entry/webformsession.js
@@ -62,7 +62,6 @@ function WebFormSession(params) {
     self.formContext = params.formContext;
     self.domain = params.domain;
     self.username = params.username;
-    self.formplayerEnabled = params.formplayerEnabled;
     self.debuggerEnabled = params.debuggerEnabled;
     self.post_url = params.post_url;
     self.displayOptions = params.displayOptions;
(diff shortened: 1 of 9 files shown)


czue:
whoa does this mean we're fully converting everyone over?

wpride:
@czue confirming that this won't move everyone onto Nimbus. This was an intermediary flag that allowed us to route calls from the old CloudCare frontend to the Touchforms server to Formplayer instead so that we could validate Formplayer was behaving correctly. Only a few test domains on here.

czue:
got it thanks

corehq/apps/cloudcare/static/cloudcare/js/form_entry/webformsession.js
```diff
@@ -151,36 +150,18 @@ WebFormSession.prototype.serverRequest = function (requestParams, callback, bloc
     this.numPendingRequests++;
     this.onLoading();
 
-    if (self.formplayerEnabled){
```
benrudolph: this we need

benrudolph: https://github.com/dimagi/commcare-hq/blob/br/resize-debugger/corehq/apps/cloudcare/static/cloudcare/js/formplayer/app.js#L133

benrudolph (changes requested):

PR #13984, opened 2016-11-21
count number of domains

snopoke:
count number of "domains with forms > 300" instead of number
of users who belong to "a domain with forms > 300"

@mkangia after looking at the data realized that this was wrong

### DIFF
--- a/corehq/apps/analytics/tasks.py
+++ b/corehq/apps/analytics/tasks.py
@@ -359,6 +359,10 @@ def track_periodic_data():
     number_of_users = 0
     number_of_domains_with_forms_gt_threshold = 0
 
+    for num_forms in domains_to_forms.values():
+        if num_forms > HUBSPOT_THRESHOLD:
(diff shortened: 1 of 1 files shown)


corehq/apps/analytics/tasks.py
```diff
@@ -359,6 +359,10 @@ def track_periodic_data():
     number_of_users = 0
     number_of_domains_with_forms_gt_threshold = 0
 
+    for num_forms in domains_to_forms.values():
+        if num_forms > HUBSPOT_THRESHOLD:
+            number_of_domains_with_forms_gt_threshold += 1
+
```
mkangia: Just thinking if using filter + len would make things quicker.
```
number_of_domains_with_forms_gt_threshold = len(filter(lambda x: x> HUBSPOT_THRESHOLD, domain_to_forms.values()))
```

mkangia:
@snopoke 👍

snopoke:
Won't change the speed

PR #13985, opened 2016-11-21
Return message when user is unauthorized

kkrampa:
@czue 

http://manage.dimagi.com/default.asp?241982

I can't find better way to return message from Tastypie


czue (approved):
this looks fine but maybe we should also add a comment explaining why this is necessary?

PR #13991, opened 2016-11-21
ICDS SMS outgoing backend

emord:
@gcapalbo modeled this after the HTTP backend

https://sms.gov.in/Documents/NIC%20SMS%20SERVICE%20INTEGRATION.pdf
https://sms.gov.in/Documents/NIC%20HTTP%20API.pdf

buddies @proteusvacuum @dannyroberts

### DIFF
--- a/corehq/messaging/smsbackends/icds/models.py
+++ b/corehq/messaging/smsbackends/icds/models.py
@@ -106,7 +106,7 @@ class SQLICDSBackend(SQLSMSBackend):
             text = msg.text.encode("iso-8859-1")
             msg_type = "PM"
         except UnicodeEncodeError:
-            text = msg.text.encode("utf-8")
+            text = msg.text.encode("utf_16_be").encode('hex').upper()
             msg_type = "UC"
         params = {
             "username": config.username,
--- a/corehq/messaging/smsbackends/icds/models.py
+++ b/corehq/messaging/smsbackends/icds/models.py
@@ -1,4 +1,3 @@
-import sys
 from urllib import urlencode
 from urllib2 import urlopen
 
@@ -6,7 +5,6 @@ from crispy_forms import layout as crispy
 from django.conf import settings
 from django.utils.translation import ugettext_lazy as _
 
-from corehq.apps.sms.mixin import BackendProcessingException
 from corehq.apps.sms.models import SQLSMSBackend
 from corehq.apps.sms.forms import BackendForm
 from corehq.apps.sms.util import strip_plus
@@ -119,12 +117,8 @@ class SQLICDSBackend(SQLSMSBackend):
         }
         url_params = urlencode(params)
         url = 'https://smsgw.sms.gov.in/failsafe/HttpLink?%s' % url_params
-        try:
-            response = urlopen("%s?%s" % (url, url_params),
-                               timeout=settings.SMS_GATEWAY_TIMEOUT).read()
-        except Exception as e:
-            msg = "Error sending message from backend: '{}'\n\n{}".format(self.pk, str(e))
-            raise BackendProcessingException(msg), None, sys.exc_info()[2]
+        response = urlopen("%s?%s" % (url, url_params),
+                           timeout=settings.SMS_GATEWAY_TIMEOUT).read()
 
         response_code = self.get_response_code(response)
         if response_code != '000':
--- a/corehq/messaging/smsbackends/icds/models.py
+++ b/corehq/messaging/smsbackends/icds/models.py
@@ -85,10 +85,13 @@ class SQLICDSBackend(SQLSMSBackend):
 
     def get_response_code(self, response):
         api_code_string = "~code=API"
-        begin_response = response.find(api_code_string) + len(api_code_string)
-        assert begin_response != -1
+        begin_response = response.find(api_code_string)
+        if begin_response == -1:
+            return None
+        begin_response += len(api_code_string)
         end_response = response[begin_response:].find('&')
-        assert end_response != -1
+        if end_response == -1:
+            return None
         return response[begin_response:end_response]
 
     def handle_error(self, response_code):
--- a/corehq/messaging/smsbackends/icds/models.py
+++ b/corehq/messaging/smsbackends/icds/models.py
@@ -5,9 +5,35 @@ from crispy_forms import layout as crispy
 from django.conf import settings
 from django.utils.translation import ugettext_lazy as _
 
+from corehq.apps.sms.mixin import BackendProcessingException
 from corehq.apps.sms.models import SQLSMSBackend
 from corehq.apps.sms.forms import BackendForm
 from dimagi.utils.django.fields import TrimmedCharField
+
+ERROR_CODES = {
+    "-2": "Invalid credentials",
+    "-3": "Empty mobile number",
+    "-4": "Empty message",
+    "-5": "HTTPS disabled",
+    "-6": "HTTP disabled",
+    "-13": "Internal Error",
+    "-201": "Email Delivery Disabled",
+    "-401": "Invalid Scheduled Time",
+    "-404": "Invalid MsgType",
+    "-406": "Invalid Port",
+    "-407": "Invalid Expiry minutes",
+    "-408": "Invalid Customer Reference Id",
+    "-409": "Invalid Bill Reference Id",
+    "-410": "Invalid Destination Address",
+    "-432": "Invalid Bill Reference Id Length",
+    "-433": "Invalid Customer Reference Id Length",
+}
+
+RETRY_ERROR_CODES = {"-13"}
+
+
+class ICDSException(Exception):
+    pass
 
 
 class ICDSBackendForm(BackendForm):
@@ -57,6 +83,21 @@ class SQLICDSBackend(SQLSMSBackend):
     def get_form_class(cls):
         return ICDSBackendForm
 
+    def get_response_code(self, response):
+        api_code_string = "~code=API"
+        begin_response = response.find(api_code_string) + len(api_code_string)
+        assert begin_response != -1
+        end_response = response[begin_response:].find('&')
+        assert end_response != -1
+        return respone[begin_response:end_response]
+
+    def handle_error(self, response_code):
+        exception_message = "Error with ICDS backend. Http respone code: %s, %s" % (
+            response_code, ERROR_CODES[response_code]
+        )
+        if response_code in RETRY_ERROR_CODES:
+            raise ICDSException(exception_message)
+
     def send(self, msg, orig_phone_number=None, *args, **kwargs):
         config = self.config
         phone_number = msg.phone_number
@@ -76,5 +117,13 @@ class SQLICDSBackend(SQLSMSBackend):
         }
         url_params = urlencode(params)
         url = 'https://smsgw.sms.gov.in/failsafe/HttpLink?%s' % url_params
-        response = urlopen("%s?%s" % (url, url_params),
-                           timeout=settings.SMS_GATEWAY_TIMEOUT).read()
+        try:
+            response = urlopen("%s?%s" % (url, url_params),
+                               timeout=settings.SMS_GATEWAY_TIMEOUT).read()
+        except Exception as e:
+            msg = "Error sending message from backend: '{}'\n\n{}".format(self.pk, str(e))
+            raise BackendProcessingException(msg), None, sys.exc_info()[2]
+
+        response_code = get_response_code(response)
+        if response_code != '000':
+            self.handle_error(response_code)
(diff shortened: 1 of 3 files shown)


corehq/messaging/smsbackends/icds/models.py
```diff
@@ -0,0 +1,129 @@
...
+    def get_form_class(cls):
+        return ICDSBackendForm
+
+    def get_response_code(self, response):
+        api_code_string = "~code=API"
+        begin_response = response.find(api_code_string) + len(api_code_string)
+        assert begin_response != -1
+        end_response = response[begin_response:].find('&')
+        assert end_response != -1
+        return respone[begin_response:end_response]
```
dimagimon: F821 undefined name 'respone'

corehq/messaging/smsbackends/icds/models.py
```diff
@@ -0,0 +1,129 @@
...
+            "msgType": msg_type
+        }
+        url_params = urlencode(params)
+        url = 'https://smsgw.sms.gov.in/failsafe/HttpLink?%s' % url_params
+        try:
+            response = urlopen("%s?%s" % (url, url_params),
+                               timeout=settings.SMS_GATEWAY_TIMEOUT).read()
+        except Exception as e:
+            msg = "Error sending message from backend: '{}'\n\n{}".format(self.pk, str(e))
+            raise BackendProcessingException(msg), None, sys.exc_info()[2]
```
dimagimon: F821 undefined name 'sys'

corehq/messaging/smsbackends/icds/models.py
```diff
@@ -0,0 +1,129 @@
...
+
+        response_code = get_response_code(response)
```
dimagimon: F821 undefined name 'get_response_code'

corehq/messaging/smsbackends/icds/models.py
```diff
@@ -0,0 +1,129 @@
...
+            raise ICDSException(exception_message)
+
+    def send(self, msg, orig_phone_number=None, *args, **kwargs):
+        config = self.config
+        phone_number = msg.phone_number
+        try:
+            text = msg.text.encode("iso-8859-1")
+            msg_type = "PM"
+        except UnicodeEncodeError:
+            text = msg.text.encode("utf-8")
```
emord: I think this may not work. The docs are not very specific on what this should be, but it looks like it may be UTF-16 with encoded, but not in a way that is obvious to construct in python. The code examples only use UTF-8 and the docs only give a string and say "unicode". I think we'll need to test this out manually

corehq/messaging/smsbackends/icds/models.py
```diff
@@ -0,0 +1,129 @@
...
+    "-406": "Invalid Port",
+    "-407": "Invalid Expiry minutes",
+    "-408": "Invalid Customer Reference Id",
+    "-409": "Invalid Bill Reference Id",
+    "-410": "Invalid Destination Address",
+    "-432": "Invalid Bill Reference Id Length",
+    "-433": "Invalid Customer Reference Id Length",
+}
+
+RETRY_ERROR_CODES = {"-13"}
```
emord: I think that this is the only code that we can retry on and expect a different response. everything else looks like user error except -5 and -6

settings.py
```diff
@@ -1415,6 +1415,7 @@
     'corehq.messaging.smsbackends.apposit.models.SQLAppositBackend',
     'corehq.messaging.smsbackends.grapevine.models.SQLGrapevineBackend',
     'corehq.messaging.smsbackends.http.models.SQLHttpBackend',
+    'corehq.messaging.smsbackends.icds.models.SQLICDSBackend',
```
emord: Currently there's no way for this to only show up on icds. it comes after the localsettings import. Would it be worthwhile to write code to enable environment specific backends?

proteusvacuum (approved):
Will let @gcapalbo comment on the SMS specific parts, but this looks good.

corehq/messaging/smsbackends/icds/models.py
gcapalbo: based on the examples in the docs it looks like this should be:

`text = msg.text.encode('utf_16_be').encode('hex').upper()`

corehq/messaging/smsbackends/icds/models.py
gcapalbo: sounds good. yeah i think we should retry -13 and unexpected or non-existent error codes too

corehq/messaging/smsbackends/icds/models.py
```diff
@@ -0,0 +1,130 @@
...
+    def handle_error(self, response_code):
+        exception_message = "Error with ICDS backend. Http respone code: %s, %s" % (
+            response_code, ERROR_CODES[response_code]
+        )
+        if response_code in RETRY_ERROR_CODES:
+            raise ICDSException(exception_message)
+
+    def send(self, msg, orig_phone_number=None, *args, **kwargs):
+        config = self.config
+        phone_number = msg.phone_number
```
gcapalbo: based on the examples in the docs it looks like we should also be calling [strip_plus](https://github.com/dimagi/commcare-hq/blob/master/corehq/apps/sms/util.py#L50) on the phone number

corehq/messaging/smsbackends/icds/models.py
```diff
@@ -0,0 +1,130 @@
...
+            "message": text,
+            "signature": config.sender_id,
+            "msgType": msg_type
+        }
+        url_params = urlencode(params)
+        url = 'https://smsgw.sms.gov.in/failsafe/HttpLink?%s' % url_params
+        try:
+            response = urlopen("%s?%s" % (url, url_params),
+                               timeout=settings.SMS_GATEWAY_TIMEOUT).read()
+        except Exception as e:
```
gcapalbo: we can actually let this exception raise and not worry about the catch/reraise. the reason we catch it and reraise in the http backend is because the http backend is shared among some other backends, so we put the pk in the message to make it easier to debug.

but since we're only going to have one instance of this backend, we'll always be able to tell from the stack trace which backend it is.

corehq/messaging/smsbackends/icds/models.py
```diff
@@ -0,0 +1,130 @@
...
+        return "ICDS"
+
+    @classmethod
+    def get_form_class(cls):
+        return ICDSBackendForm
+
+    def get_response_code(self, response):
+        api_code_string = "~code=API"
+        begin_response = response.find(api_code_string) + len(api_code_string)
+        assert begin_response != -1
```
gcapalbo: i think we should return None if we can't find the response code, and then retry when the response code is None

but note that we need to check if `response.find(api_code_string)` is -1 before we add `len(api_code_string)`

corehq/messaging/smsbackends/icds/models.py
```diff
@@ -0,0 +1,130 @@
...
+        end_response = response[begin_response:].find('&')
+        assert end_response != -1
+        return response[begin_response:end_response]
+
+    def handle_error(self, response_code):
+        exception_message = "Error with ICDS backend. Http respone code: %s, %s" % (
+            response_code, ERROR_CODES[response_code]
+        )
+        if response_code in RETRY_ERROR_CODES:
```
gcapalbo: i think it would be good to raise an exception also when response code is not in ERROR_CODES (in which case we'd need a different message)

corehq/messaging/smsbackends/icds/models.py
```diff
@@ -0,0 +1,130 @@
...
+from django.conf import settings
+from django.utils.translation import ugettext_lazy as _
+
+from corehq.apps.sms.mixin import BackendProcessingException
+from corehq.apps.sms.models import SQLSMSBackend
+from corehq.apps.sms.forms import BackendForm
+from dimagi.utils.django.fields import TrimmedCharField
+
+ERROR_CODES = {
+    "-2": "Invalid credentials",
```
dannyroberts: Curious about the number-in-a-string format here. Seems like the type of thing that could be easy to mess up (like trying `"-1" == -1` by mistake), but it looks like you've got a system.

corehq/messaging/smsbackends/icds/models.py
```diff
@@ -0,0 +1,130 @@
...
+        except UnicodeEncodeError:
+            text = msg.text.encode("utf-8")
+            msg_type = "UC"
+        params = {
+            "username": config.username,
+            "pin": config.pin,
+            "mnumber": phone_number,
+            "message": text,
+            "signature": config.sender_id,
+            "msgType": msg_type
```
gcapalbo: let's also add `"splitAlgm": "concat"`. this should let us use concatenated sms (sending messages longer than 160 ascii characters, or longer than 70 unicode characters), but we'll want to test it out

corehq/messaging/smsbackends/icds/models.py
emord: https://github.com/dimagi/commcare-hq/pull/13991/commits/c61bcc85596b5efc90cf8ebc6b15c3a65da1e2c1

corehq/messaging/smsbackends/icds/models.py
emord: https://github.com/dimagi/commcare-hq/pull/13991/commits/615fa458f1e3975782cdd8cdc75bc3c0b36fd849

corehq/messaging/smsbackends/icds/models.py
emord: https://github.com/dimagi/commcare-hq/pull/13991/commits/365879f853dbc54598fc220ac43d8c8a774f2625

corehq/messaging/smsbackends/icds/models.py
emord: https://github.com/dimagi/commcare-hq/pull/13991/commits/ffa208f20ad3f005d7343ff617b9ef09fa0492dd

corehq/messaging/smsbackends/icds/models.py
```diff
@@ -0,0 +1,130 @@
...
+        end_response = response[begin_response:].find('&')
+        assert end_response != -1
+        return response[begin_response:end_response]
+
+    def handle_error(self, response_code):
+        exception_message = "Error with ICDS backend. Http respone code: %s, %s" % (
+            response_code, ERROR_CODES[response_code]
+        )
+        if response_code in RETRY_ERROR_CODES:
+            raise ICDSException(exception_message)
```
emord: https://github.com/dimagi/commcare-hq/pull/13991/commits/64589998043f60222998de6424938b73172d61ac

Is there something in the framework whenever we retry an SMS more than x times? Something like on th 5th retry it stops trying and emails the 500 emails

gcapalbo: it will retry based on these settings: https://github.com/dimagi/commcare-hq/blob/master/settings.py#L637-L642

so right now we have it set to do 3 attempts total on 5 minute intervals

corehq/messaging/smsbackends/icds/models.py
```diff
@@ -0,0 +1,132 @@
...
+            "pin": config.pin,
+            "mnumber": phone_number,
+            "message": text,
+            "signature": config.sender_id,
+            "msgType": msg_type,
+            "splitAlgm": "concat",
+        }
+        url_params = urlencode(params)
+        url = 'https://smsgw.sms.gov.in/failsafe/HttpLink?%s' % url_params
+        response = urlopen("%s?%s" % (url, url_params),
```
gcapalbo: i just realized this - we're adding the url_params twice to the url

corehq/messaging/smsbackends/icds/models.py
```diff
@@ -0,0 +1,132 @@
...
+    def get_form_class(cls):
+        return ICDSBackendForm
+
+    def get_response_code(self, response):
+        api_code_string = "~code=API"
+        begin_response = response.find(api_code_string)
+        if begin_response == -1:
+            return None
+        begin_response += len(api_code_string)
+        end_response = response[begin_response:].find('&')
```
gcapalbo: we'll need to actually check for `response.find('&', begin_response)` otherwise we're not getting the right index

corehq/messaging/smsbackends/icds/models.py
```diff
@@ -0,0 +1,132 @@
...
+        if end_response == -1:
+            return None
+        return response[begin_response:end_response]
```
gcapalbo: we should also trim whitespace from the final result since it looks like in some of the examples there can be whitespace around it

gcapalbo:
@emord just a few more things otherwise this should be good to go

gcapalbo:
looks good, just waiting on build

PR #13993, opened 2016-11-21
Add SQLLocation.to_json()

esoergel:
This isn't used anywhere yet, but I want to verify parity first.


corehq/apps/locations/models.py
```diff
@@ -403,6 +403,25 @@ def save(self, *args, **kwargs):
         if sync_to_couch:
             self._migration_do_sync()
 
+    def to_json(self):
+        return {
```
benrudolph: 👍

czue:
this might be a perfect opportunity to use some Science! :atom_symbol:

PR #13995, opened 2016-11-21
Release refresh case management

proteusvacuum:
https://github.com/dimagi/commcare-hq/pull/13592
https://github.com/dimagi/commcare-hq/pull/13592#issuecomment-261964050
@dannyroberts @emord

### DIFF
--- a/corehq/apps/app_manager/tests/data/v2_diffs/partials/case_config.html.diff.txt
+++ b/corehq/apps/app_manager/tests/data/v2_diffs/partials/case_config.html.diff.txt
@@ -9,7 +9,7 @@
  
  <script type="text/html" id="remove-subcase-modal-template">
      <div class="modal-dialog">
-@@ -63,6 +63,21 @@
+@@ -63,76 +63,97 @@
      </div>
  </script>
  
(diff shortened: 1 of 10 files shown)


emord:
looks like an app manager diff failure

proteusvacuum:
Cool, very clear failure message, thanks @biyeun!

emord (approved):

corehq/apps/app_manager/tests/data/v2_diffs/partials/case_config.html.diff.txt
```diff
@@ -31,9 +31,7 @@
  
  <div id="case-config-ko">
      <div data-bind="saveButton: saveButton"></div>
-@@ -70,71 +85,77 @@
-         <div data-bind="template: 'case-config:refresh-form-questions'"></div>
-     {% endif %}
```
dannyroberts: Looks like an instance in which this test was a "false positive"—i.e. made the build fail, but where the code itself was fine and required a change only in the test. Nothing actionable, just noting here, since if that happens to often it dilutes the signal and could become error-prone. My guess is it's a good-enough solution and worth being super-simple to implement.

PR #13999, opened 2016-11-22
don't include suppressed subscriptions in api

nickpell:
http://manage.dimagi.com/default.asp?241937

@kaapstorm


kaapstorm:
I always think it's cool when using less code resolves things.

PR #14000, opened 2016-11-22
force set case props if not set

snopoke:
Seems reasonable to always have these properties set on the case even if there hasn't been a form that explicitly sets them.

@czue @proteusvacuum 
buddy @biyeun

### DIFF
--- a/corehq/form_processor/backends/sql/update_strategy.py
+++ b/corehq/form_processor/backends/sql/update_strategy.py
@@ -72,6 +72,14 @@ class SqlCaseUpdateStrategy(UpdateStrategy):
         if self.case.modified_on is None or modified_on > self.case.modified_on:
             self.case.modified_on = modified_on
 
+        # edge case if form updates case before it's been created (or creation form archived)
+        if not self.case.opened_on:
+            self.case.opened_on = case_update.guess_modified_on()
+        if not self.case.opened_by:
+            self.case.opened_by = case_update.user_id
+        if not self.case.owner_id:
+            self.case.owner_id = case_update.user_id
+
     def _apply_action(self, case_update, action, xform):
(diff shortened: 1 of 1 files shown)


czue (approved):
interesting, I'm curious how you encountered this.

I can think of a few ways this could come up:

1. archiving the create form and rebuilding
2. misbehaving api users / integrations 
3. broken apps / app builder bugs
4. mobile bugs

do we feel like it's the right behavior in all of these scenarios to set these properties? am guessing it's better just thought i'd ask explicitly.

snopoke:
I came across it while trying to migrate my domain to SQL. Agree with those scenarios + one more which use "mobile user submits form to archived case".

In all these cases I think this is the right behaviour (or at least better than having them blank).

czue:
yeah sounds right.

corehq/form_processor/backends/sql/update_strategy.py
```diff
@@ -72,6 +72,14 @@ def _apply_case_update(self, case_update, xformdoc):
         if self.case.modified_on is None or modified_on > self.case.modified_on:
             self.case.modified_on = modified_on
 
+        # edge case if form updates case before it's been created (or creation form archived)
+        if not self.case.opened_on:
+            self.case.opened_on = case_update.guess_modified_on()
```
proteusvacuum: You could probably reuse the `modified_on` variable from above. Not sure if this does whatever work `guess_modified_on` does twice or if that is cached in some way.

snopoke: https://github.com/dimagi/commcare-hq/pull/14035/commits/f08eaf5d1b2b2b6d3cde4f2570dfbdd51474ceb1

PR #14001, opened 2016-11-22
ignore unrecognized kwargs in HQPasswordResetForm.save

nickpell:
http://manage.dimagi.com/default.asp?241713

@kaapstorm


nickpell:
✅

PR #14003, opened 2016-11-22
Explicitly specify archive filename

benrudolph:
@czue @NoahCarnahan this'll ensure that when you unzip an export folder that has only one table it'll be unzipped into a folder with the export name. this prevents the `Forms (2134).csv` effect. this was something that had always annoyed me
screen shot 2016-11-22 at 12 33 13 pm

fyi: @jessica-long

### DIFF
--- a/corehq/apps/export/export.py
+++ b/corehq/apps/export/export.py
@@ -56,6 +56,12 @@ class _Writer(object):
 
         # Create and open a temp file
         assert self._path is None
+
+        if len(export_instances) == 1:
+            name = export_instances[0].name
+        else:
+            name = ''
+
         fd, self._path = tempfile.mkstemp()
         with os.fdopen(fd, 'wb') as file:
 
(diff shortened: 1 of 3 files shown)


NoahCarnahan:
Test failure looks real, otherwise looks good

czue:
still failing
```
===================================================================
ERROR: test_zipped_export_writer_unicode (couchexport.tests.test_writers.ZippedExportWriterTests)
  File "corehq/ex-submodules/couchexport/tests/test_writers.py", line 32, in test_zipped_export_writer_unicode
    self.writer._write_final_result()
  File "corehq/ex-submodules/couchexport/writers.py", line 295, in _write_final_result
    archive.write(path, self._get_archive_filename(name))
  File "corehq/ex-submodules/couchexport/writers.py", line 300, in _get_archive_filename
    return os.path.join(self.archive_basepath, u'{}{}'.format(name, self.table_file_extension))
UnicodeDecodeError: 'ascii' codec can't decode byte 0xe3 in position 0: ordinal not in range(128)

````

benrudolph:
✅

PR #14004, opened 2016-11-22
Allow App Preview to perform updates, attempt to reload session

wpride:
cross-request https://github.com/dimagi/formplayer/pull/193


corehq/apps/cloudcare/static/cloudcare/js/formplayer/app.js
```diff
@@ -393,16 +399,19 @@ FormplayerFrontend.on('refreshApplication', function(appId) {
     if (!appId) {
         throw new Error('Attempt to refresh application for null appId');
     }
+    var sessionId = FormplayerFrontend.request('getCurrentSessionId');
```
benrudolph: what happens when someone is on the main screen of app preview and hits update and sessionId is null?

corehq/apps/cloudcare/static/cloudcare/js/formplayer/app.js
```diff
@@ -412,7 +421,12 @@ FormplayerFrontend.on('refreshApplication', function(appId) {
         tfLoadingComplete(true);
     }).done(function() {
         tfLoadingComplete();
-        FormplayerFrontend.trigger('navigateHome');
+        if (typeof resp.responseJSON.tree !== 'undefined') {
```
benrudolph: this is kind of an awkward check. can just use: http://underscorejs.org/#isUndefined

corehq/apps/cloudcare/static/cloudcare/js/formplayer/app.js
wpride: In that case the sessionId will be null on Formplayer too and we'll just do a normal update with no attempt at replaying anything

benrudolph (approved):
sweet 🍬

PR #14005, opened 2016-11-22
Fix gmaps key in preview app and move to cloudcare app

benrudolph:
@wpride @biyeun this fixes geo questions in app preview. i also refactored to remove the preview app to be in the cloudcare app. realized the preview app literally had nothing inside of it. best read commit-by-commit

cc: @mkangia


biyeun:
@benrudolph conflicts

PR #14006, opened 2016-11-22
Je/data dictionary followup

emord:
@proteusvacuum from your feedback in #13957

### DIFF
--- a/corehq/apps/data_dictionary/templates/data_dictionary/base.html
+++ b/corehq/apps/data_dictionary/templates/data_dictionary/base.html
@@ -34,18 +34,22 @@
 {% endblock %}
 
 {% block page_content %}
-    <table id="data-dictionary-table" class="table">
-        <thead>
-            <th>{% trans "Case Property" %}</th>
-            <th>{% trans "Type" %}</th>
-            <th>{% trans "Description" %}</th>
-        </thead>
-        <tbody data-bind="foreach: activeCaseTypeData">
-            <tr>
-                <td data-bind="text: name"></td>
-                <td data-bind="text: type"></td>
-                <td data-bind="text: description"></td>
-            </tr>
-        </tbody>
-    </table>
+    <div class="row">
+        <div class="col-md-10">
+            <table id="data-dictionary-table" class="table">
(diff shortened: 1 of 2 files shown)


corehq/apps/data_dictionary/templates/data_dictionary/base.html
```diff
@@ -34,18 +34,22 @@ <h2 class="text-hq-nav-header">{% trans "Data Dictionary" %}</h2>
...
-                <td data-bind="text: description"></td>
-            </tr>
-        </tbody>
-    </table>
+    <div class="row">
+        <div class="col-md-10">
+            <table id="data-dictionary-table" class="table">
+                <thead>
+                    <th>{% trans "Case Property" %}</th>
+                    <th>{% trans "Type" %}</th>
```
proteusvacuum: Is this table sortable? I think it might be easier to reason with if `type` is first on the list (the same way the [downloadable case summary](https://india.commcarehq.org/a/enikshay-test/apps/view/e6c88dc7cf8c742cbf397eff07d9a245/summary/case/download/?lang=en) has it. Also, assuming this is "Case Type" and not "Data Type". 
But I'm guessing this UI will probably go through some UX testing before getting fully released

proteusvacuum (approved):

corehq/apps/data_dictionary/templates/data_dictionary/base.html
emord: This table isn't sortable right now, and the UX is not complete. I'm just including this as a quick way to view it while I was developing the backend. 

It is "data type" https://github.com/dimagi/commcare-hq/pull/13957/files#diff-ff11fdee49a311b27e26488e60a55f85R3 so case property should be on the left as it's the "unique key" and type is more information about the property. This table only has one case type per page

proteusvacuum: Cool. That makes sense then. I'd maybe update that to say "Data Type" whenever you build this out since "Type" is an overloaded word when we are thinking about cases. But that doesn't have to happen now.

PR #14007, opened 2016-11-22
Case importer and spreadsheets naming cleanup

dannyroberts:
Not obviously planning a v2 rewrite now that I've been working within the current case importer app. The spreadsheet utils also solve different problems so I've named them accordingly instead of v1/v2 now that that's clearer to me.

~~This also make `workbook_json` use `workbook_reading` and thus get better exceptions, typing, and xls compatibility for free.~~ (Aborted for this PR)


dannyroberts:
Removed the more questionable part of the PR and tested fixture uploads and case uploads. This is just changing names, so anything that would have broken would have broken very obviously.

dannyroberts:
@NoahCarnahan followup on naming cc: @czue buddy

PR #14009, opened 2016-11-22
remove parent_ prefix

benrudolph:
@proteusvacuum / anyone
cc: @mkangia


benrudolph:
ah wait it's supposed to be `parent_id`

proteusvacuum:
Yeah, not sure if @dannyroberts has any plans of changing that in the importer work he is doing? It seems bad to have random reserved words in the importer that could legitimately be case property names.

PR #14010, opened 2016-11-23
Mk 241159 prepopulate locations for edit

mkangia:
http://manage.dimagi.com/default.asp?241159

Pre-populate the locations in the form at render to avoid ajax request causing delay in render.

@benrudolph @esoergel

### DIFF
--- a/corehq/apps/locations/views.py
+++ b/corehq/apps/locations/views.py
@@ -53,7 +53,7 @@ from .permissions import (
 from .models import Location, LocationType, SQLLocation, filter_for_archived
 from .forms import LocationForm, UsersAtLocationForm
 from .tree_utils import assert_no_cycles
-from .util import load_locs_json, location_hierarchy_config, dump_locations
+from .util import load_locs_json, location_hierarchy_config, dump_locations, loc_to_payload
 
 
 logger = logging.getLogger(__name__)
@@ -834,6 +834,16 @@ def location_export(request, domain):
     return response
 
 
+def get_locations_paylod_from_ids(ids, domain, base_queryset):
+    from corehq.apps.locations.util import get_locations_from_ids
+    locations = get_locations_from_ids(ids, domain, base_queryset=base_queryset)
+    return [loc_to_payload(loc) for loc in locations]
+
+
+def base_queryset_for_domain_and_user(domain, user):
+    return SQLLocation.objects.accessible_to_user(domain, user)
+
+
 @locations_access_required
 @location_safe
 def child_locations_for_select2(request, domain):
@@ -842,10 +852,7 @@ def child_locations_for_select2(request, domain):
     query = request.GET.get('name', '').lower()
     user = request.couch_user
 
-    base_queryset = SQLLocation.objects.accessible_to_user(domain, user)
-
-    def loc_to_payload(loc):
-        return {'id': loc.location_id, 'name': loc.display_name}
+    base_queryset = base_queryset_for_domain_and_user(domain, user)
 
     if id:
         try:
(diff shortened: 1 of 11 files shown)


mkangia:
@esoergel i am not very confident if this can cause issues at any other place we are rendering locations. Anything i must test specifically apart from result from tests?

corehq/apps/locations/views.py
```diff
@@ -834,6 +834,16 @@ def location_export(request, domain):
...
 
 
+def get_locations_paylod_from_ids(ids, domain, base_queryset):
+    from corehq.apps.locations.util import get_locations_from_ids
+    locations = get_locations_from_ids(ids, domain, base_queryset=base_queryset)
+    return [loc_to_payload(loc) for loc in locations]
+
+
+def base_queryset_for_domain_and_user(domain, user):
+    return SQLLocation.objects.accessible_to_user(domain, user)
```
esoergel: It just occurred to me that we should make `accessible_to_user` always filter results by domain.  For non-restricted users, this currently doesn't add a domain filter, and I could see things like this confusingly including results from other domains, if that filter isn't explicitly added elsewhere.

corehq/apps/users/forms.py
```diff
@@ -794,13 +799,27 @@ class CommtrackUserForm(forms.Form):
...
     def __init__(self, *args, **kwargs):
+        from corehq.apps.locations.views import base_queryset_for_domain_and_user, get_locations_paylod_from_ids
+
         self.domain = None
+        populate_with = None
         if 'domain' in kwargs:
             self.domain = kwargs['domain']
             del kwargs['domain']
+
+        if 'user' in kwargs:
```
esoergel: Should this just be a required argument?  That might simplify some things, and I can't think of an instance offhand where we wouldn't have `user` available when instantiating the form.

esoergel:
@mkangia I'd make sure to try out any pages where this form is used, which looks like only a couple.  That's load the page, make sure it prepopulates as expected, save, reload.

corehq/apps/users/forms.py
mkangia: That does make sense. Since this form is used for both web and mobile user only and both need the locations pre-populated.

mkangia: Post discussion with Ethan. No changes needed.

corehq/apps/locations/views.py
esoergel: (I could imagine someone seeing this method without context and assuming that it restricts by domain)

mkangia: @esoergel I think for admin user we would still need to keep `self.all()` since they don't have domain membership and have access to all domain. Will update using that.

esoergel: I don't think we need to worry about admin users, because if code is calling `accessible_to_user(domain, user)`, then they are specifying a domain anyways.  The only time I can think of when we might be intentionally dealing with a cross-domain queryset is in migrations and very advanced type things which presumably wouldn't be using `accessible_to_user` anyways.

One other point - you wrote:
> to be
```python
    return SQLLocation.active_objects.filter(domain=domain)
```

But it should actually be:
```python
    return self.filter(domain=domain)
```
This way it will chain properly with other filters that had been applied.

mkangia: Oh yes.. thats right. Admin would get access to locations in the domain only.
Ya! i thought i would give `self` a shot once i actually code this. Any filter would generally be after this so i didn't see chaining as a challenge then. But ya self!

mkangia:
Closing this and opening a new PR https://github.com/dimagi/commcare-hq/pull/14159 since Sravan took care of preloading locations in https://github.com/dimagi/commcare-hq/pull/14143

PR #14012, opened 2016-11-23
Make undeleting users easier

kaapstorm:
Context: [FB 240847](http://manage.dimagi.com/default.asp?240847)

deleted_users

We can undelete mobile workers from the UI if we know their ID. This PR includes a management command to find the IDs of deleted users, and a page to list them.

@nickpell @sravfeyn 

(Feel free to push back against "OCD" commit 4ad95a6 if it's going to cause merge conflicts for Django upgrades.)


corehq/apps/users/views/mobile/users.py
```diff
@@ -789,6 +792,26 @@ def create_mobile_worker(self, in_data):
         }
 
 
+@location_safe
```
czue: I think this should probably be behind a superuser flag (or maybe even a support toggle). because of the complexity and inconsistency of how restoring deleted users actually works it seems unwise to expose this to everyone (even if it is a magic url)

czue (commented):
nice!

nickpell:
👍

PR #14018, opened 2016-11-23
ICDS: Remove Case History Expressions

sheelio:
This removes the references to the case history from the ICDS data sources (since case history does not work on SQL domains).

### DIFF
--- a/custom/icds_reports/ucr/data_sources/child_health_cases_monthly_tableau.json
+++ b/custom/icds_reports/ucr/data_sources/child_health_cases_monthly_tableau.json
@@ -1106,14 +1106,11 @@
                 "type": "nested",
                 "argument_expression": {
                   "type": "named",
-                  "name": "get_latest_cf_update"
+                  "name": "get_latest_cf_repeat"
                 },
                 "value_expression": {
-                  "type": "property_path",
-                  "property_path": [
-                    "updated_unknown_properties",
-                    "comp_feeding"
-                  ]
+                  "type": "property_name",
+                  "property_name": "comp_feeding"
                 }
               }
             },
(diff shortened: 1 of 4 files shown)


custom/icds_reports/ucr/data_sources/child_health_cases_monthly_tableau.json
```diff
@@ -1200,19 +1191,16 @@
             {
               "type": "boolean_expression",
               "operator": "eq",
-              "property_value": "yes",
+              "property_value": "1",
```
czue: why did this change?

custom/icds_reports/ucr/tests/test_child_health_monthly.py
```diff
@@ -193,6 +196,50 @@ def _submit_thr_rations_form(
 
         submit_form_locally(ElementTree.tostring(form), self.domain, **{})
 
+    def _submit_delivery_form(
+            self, form_date, case_id, nutrition_status=None, case_id_2=None):
+
+        form = ElementTree.Element('data')
```
czue: wow this looks fun. well done - i would have just been lazy and used a template.

custom/icds_reports/ucr/tests/test_child_health_monthly.py
```diff
@@ -1124,3 +1237,73 @@ def test_ebf_no_ebf_reasons(self):
...
+        case_id_2 = uuid.uuid4().hex
+        self._create_case(
+            case_id=case_id,
+            dob=date(2015, 5, 12),
+            date_opened=datetime(2015, 5, 14),
+            date_modified=datetime(2016, 3, 12),
+        )
+        self._create_case(
+            case_id=case_id_2,
+            dob=date(2015, 05, 12),
```
czue: github syntax highlighting caught this - `05` is an ocatal 5 in python, which happily is the same as a decimal 5, though probably better to just have `5`

czue (approved):
didn't end up looking at the calculations closely, let me know if you wanted me to

czue:
sorry, I should have caught the couch/sql thing earlier. glad it wasn't too painful to switch.

custom/icds_reports/ucr/data_sources/child_health_cases_monthly_tableau.json
sheelio: Realized the app was using 1s and 0s instead of yes for this particular question.

PR #14019, opened 2016-11-24
"Restore As" permissions

benrudolph:
@czue this implements permissions for the restore as feature (and removes the toggle it was previously behind). this does:

- Allow mobile users to be able to restore as other mobile users
- Allow mobile workers to conditionally be able to view the Login As button in web apps

Being able to pull the list of users in a location safe way was already implemented, but it requires the Edit Mobile Workers permission which is unfortunate.

In order to restore as another mobile worker when you're a mobile worker you'll need to have the `DATA_CLEANUP` permission and you need to have the `edit_data` permission and have that user in your location or its descendants. Wasn't sure if you imagined having a totally separate permission for using restore as

Still todo:

- [x] Better error handling when permission is denied in web apps
- [x] Reconcile that Edit Mobile Workers permission (don't think it should be required for getting a list of mobile workers)

 🐟 or 🏠 

cc some location folk to make sure i'm doing things correctly with locations: @esoergel @proteusvacuum

### DIFF
--- a/corehq/apps/hqwebapp/templatetags/hq_shared_tags.py
+++ b/corehq/apps/hqwebapp/templatetags/hq_shared_tags.py
@@ -231,7 +231,7 @@ def can_use_restore_as(request):
         return False
 
     if (request.couch_user.is_commcare_user() and
-            not request.couch_user.can_edit_data()):
+            not request.couch_user.can_edit_commcare_users()):
         return False
 
     return True
--- a/corehq/apps/hqwebapp/templatetags/hq_shared_tags.py
+++ b/corehq/apps/hqwebapp/templatetags/hq_shared_tags.py
@@ -15,8 +15,10 @@ from django.core.urlresolvers import reverse
 from django.utils.html import format_html
 from django.utils.safestring import mark_safe
 from dimagi.utils.decorators.memoized import memoized
+from django_prbac.utils import has_privilege
 
 from dimagi.utils.make_uuid import random_hex
+from corehq import privileges
 from corehq.apps.domain.models import Domain
 from corehq.util.quickcache import quickcache
 from corehq.util.soft_assert import soft_assert
@@ -221,6 +223,18 @@ def toggle_enabled(request, toggle_or_toggle_name):
 def is_new_cloudcare(request):
     from corehq import toggles
     return _toggle_enabled(toggles, request, toggles.USE_FORMPLAYER_FRONTEND)
+
+
+@register.filter
+def can_use_restore_as(request):
+    if not has_privilege(request, privileges.DATA_CLEANUP):
+        return False
+
+    if (request.couch_user.is_commcare_user() and
+            not request.couch_user.has_permission(request.domain, 'edit_data')):
+        return False
+
+    return True
 
 
 @register.simple_tag
--- a/corehq/apps/hqwebapp/templatetags/hq_shared_tags.py
+++ b/corehq/apps/hqwebapp/templatetags/hq_shared_tags.py
@@ -227,6 +227,9 @@ def is_new_cloudcare(request):
 
 @register.filter
 def can_use_restore_as(request):
+    if not hasattr(request, 'couch_user'):
+        return False
+
     return (
         request.couch_user.can_edit_commcare_users() and
         has_privilege(request, privileges.DATA_CLEANUP)
--- a/corehq/apps/hqwebapp/templatetags/hq_shared_tags.py
+++ b/corehq/apps/hqwebapp/templatetags/hq_shared_tags.py
@@ -231,7 +231,7 @@ def can_use_restore_as(request):
         return False
 
     if (request.couch_user.is_commcare_user() and
-            not request.couch_user.has_permission(request.domain, 'edit_data')):
+            not request.couch_user.can_edit_data()):
         return False
 
     return True
--- a/corehq/apps/hqwebapp/templatetags/hq_shared_tags.py
+++ b/corehq/apps/hqwebapp/templatetags/hq_shared_tags.py
@@ -227,14 +227,10 @@ def is_new_cloudcare(request):
 
 @register.filter
 def can_use_restore_as(request):
-    if not has_privilege(request, privileges.DATA_CLEANUP):
-        return False
-
-    if (request.couch_user.is_commcare_user() and
-            not request.couch_user.can_edit_commcare_users()):
-        return False
-
-    return True
+    return (
+        request.couch_user.can_edit_commcare_users() and
+        has_privilege(request, privileges.DATA_CLEANUP)
+    )
 
 
 @register.simple_tag
--- a/corehq/apps/locations/permissions.py
+++ b/corehq/apps/locations/permissions.py
@@ -327,7 +327,7 @@ def user_can_access_other_user(domain, user, other_user):
         .accessible_to_user(domain, user)
         .values_list('id', flat=True))
 
-    return other_user.sql_locations.filter(id__in=accessible_location_ids).count() > 0
+    return other_user.get_sql_locations(domain).filter(id__in=accessible_location_ids).exists()
 
 
 def can_edit_location(view_fn):
--- a/corehq/apps/locations/permissions.py
+++ b/corehq/apps/locations/permissions.py
@@ -323,11 +323,10 @@ def user_can_access_other_user(domain, user, other_user):
     if user.has_permission(domain, 'access_all_locations'):
         return True
 
-    accessible_location_ids = (SQLLocation.objects
-        .accessible_to_user(domain, user)
-        .values_list('id', flat=True))
-
-    return other_user.get_sql_locations(domain).filter(id__in=accessible_location_ids).exists()
+    return (other_user
+            .get_sql_locations(domain)
+            .accessible_to_user(domain, user)
+            .exists())
 
 
 def can_edit_location(view_fn):
--- a/corehq/apps/locations/permissions.py
+++ b/corehq/apps/locations/permissions.py
@@ -319,6 +319,19 @@ def user_can_access_any_location_id(domain, user, location_ids):
             .exists())
 
 
+def user_can_access_other_user(domain, user, other_user):
+    if user.has_permission(domain, 'access_all_locations'):
+        return True
+
+    accessible_location_ids = (SQLLocation.objects .accessible_to_user(domain, user) .values_list('id', flat=True))
+
+    other_user_locations = other_user.sql_locations
+    return any(map(
+        lambda location_id: location_id in accessible_location_ids,
+        other_user_locations.values_list('id', flat=True) if other_user_locations else []
+    ))
+
+
 def can_edit_location(view_fn):
     """
     Decorator controlling a user's access to a specific location.
--- a/corehq/apps/locations/permissions.py
+++ b/corehq/apps/locations/permissions.py
@@ -327,12 +327,7 @@ def user_can_access_other_user(domain, user, other_user):
         .accessible_to_user(domain, user)
         .values_list('id', flat=True))
 
-    other_user_locations = other_user.sql_locations
-
-    if other_user_locations:
-        return other_user_locations.filter(id__in=accessible_location_ids).count() > 0
-
-    return False
+    return other_user.sql_locations.filter(id__in=accessible_location_ids).count() > 0
 
 
 def can_edit_location(view_fn):
--- a/corehq/apps/locations/permissions.py
+++ b/corehq/apps/locations/permissions.py
@@ -323,13 +323,16 @@ def user_can_access_other_user(domain, user, other_user):
     if user.has_permission(domain, 'access_all_locations'):
         return True
 
-    accessible_location_ids = (SQLLocation.objects .accessible_to_user(domain, user) .values_list('id', flat=True))
+    accessible_location_ids = (SQLLocation.objects
+        .accessible_to_user(domain, user)
+        .values_list('id', flat=True))
 
     other_user_locations = other_user.sql_locations
-    return any(map(
-        lambda location_id: location_id in accessible_location_ids,
-        other_user_locations.values_list('id', flat=True) if other_user_locations else []
-    ))
+
+    if other_user_locations:
+        return other_user_locations.filter(id__in=accessible_location_ids).count() > 0
+
+    return False
 
 
 def can_edit_location(view_fn):
--- a/corehq/apps/ota/tests/test_utils.py
+++ b/corehq/apps/ota/tests/test_utils.py
@@ -24,6 +24,7 @@ class RestorePermissionsTest(LocationHierarchyTestCase):
 
     @classmethod
     def setUpClass(cls):
+        delete_all_users()
         super(RestorePermissionsTest, cls).setUpClass()
         delete_all_users()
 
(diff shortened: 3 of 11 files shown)


corehq/apps/locations/permissions.py
```diff
@@ -319,6 +319,19 @@ def user_can_access_any_location_id(domain, user, location_ids):
             .exists())
 
 
+def user_can_access_other_user(domain, user, other_user):
+    if user.has_permission(domain, 'access_all_locations'):
+        return True
+
+    accessible_location_ids = (SQLLocation.objects .accessible_to_user(domain, user) .values_list('id', flat=True))
+
+    other_user_locations = other_user.sql_locations
```
benrudolph: this function is kind of funky in that when you do have locations, it returns a queryset. when you don't have locations it returns an empty array which seems like it could cause errors pretty easily if you do something like `user.sql_locations.values_list`

corehq/apps/ota/views.py
```diff
@@ -37,6 +38,7 @@
 from .utils import demo_user_restore_response, get_restore_user, is_permitted_to_restore, handle_401_response
 
 
+@location_safe
```
benrudolph: this can break fairly easily because it's order dependent (has to be the first decorator). haven't looked into a way to fix, but thought i'd flag

proteusvacuum: do we need to `@location_safe` the other views in this file too? I think case claim won't work without them (this was on my backlog to do, so no worries if you don't want to add it here, I'll do it in a followup)

corehq/apps/ota/tests/test_utils.py
```diff
@@ -1,21 +1,31 @@
...
+class RestorePermissionsTest(LocationHierarchyTestCase):
     domain = 'goats'
     other_domain = 'sheep'
+    location_type_names = ['country', 'state']
+    location_structure = [
+        ('usa', [
+            ('ma', []),
+        ]),
+        ('canada', [
+            ('montreal', []),
```
proteusvacuum: ahem... Canada doesn't have "states". And Montreal would not be one of them even if it did.
:trollface:

corehq/apps/locations/permissions.py
```diff
@@ -319,6 +319,19 @@ def user_can_access_any_location_id(domain, user, location_ids):
...
+    return any(map(
+        lambda location_id: location_id in accessible_location_ids,
+        other_user_locations.values_list('id', flat=True) if other_user_locations else []
```
proteusvacuum: I think you might be able to do
`other_user_locations.filter(id__in=accessible_location_ids).count() > 0` to not have to loop through each one

corehq/apps/locations/permissions.py
proteusvacuum: that seems bad. @sravfeyn was there a reason you returned an empty list instead of an empty queryset? I think we should be returning `SQLLocation.objects.none()` there instead.

corehq/apps/ota/utils.py
```diff
@@ -99,36 +145,17 @@ def is_permitted_to_restore(domain, couch_user, as_user, has_data_cleanup_privel
...
-            if user_domain != domain:
-                # In this case we may be dealing with a WebUser
-                user = WebUser.get_by_username(as_user)
-                if user and user.is_member_of(domain):
-                    message = None
-                else:
-                    message = u"{} was not in the domain {}".format(username, domain)
-
-    return message is None, message
+    try:
```
proteusvacuum: nice cleanup. 
nit: I find it more readable to have helper methods defined below where they are first referenced.

corehq/apps/ota/utils.py
```diff
@@ -87,7 +90,50 @@ def demo_restore_date_created(commcare_user):
...
+        ))
+
+
+def _ensure_valid_restore_as_user(domain, couch_user, as_user):
+    try:
+        username = as_user.split('@')[0]
+        user_domain = as_user.split('@')[1]
+    except IndexError:
+        raise RestorePermissionDenied(
+            u"Invalid restore user {}. Format is <user>@<domain>".format(as_user)
```
proteusvacuum: I think these messages get shown to the user at some point (at least the view throws a `401` with this as the message). Not sure if you think it worthwhile to add translation tags to them?

corehq/apps/locations/permissions.py
```diff
@@ -319,6 +319,19 @@ def user_can_access_any_location_id(domain, user, location_ids):
             .exists())
 
 
+def user_can_access_other_user(domain, user, other_user):
+    if user.has_permission(domain, 'access_all_locations'):
+        return True
+
+    accessible_location_ids = (SQLLocation.objects .accessible_to_user(domain, user) .values_list('id', flat=True))
```
proteusvacuum: nit: there's a space before the `.`'s, and you don't need the `()`

proteusvacuum (commented):
Looks good to me, made a few general comments, none of which are blocking. The locations stuff looks fine to me, but will let @esoergel confirm. 
Excited to have this!!

corehq/apps/locations/permissions.py
benrudolph: ah oops i didn't mean to commit that. it was originally on multiples lines. i just hit `J` in vim 3 times and then copy and paste to ipdb to debug. then i'm supposed to format it back...

corehq/apps/ota/tests/test_utils.py
benrudolph: haha 😳 it was after the fact when i realized i wanted to test the negative case

corehq/apps/ota/utils.py
benrudolph: yeah i think worth it, though it's likely that no one will be able to do anything with these messages except us

corehq/apps/ota/utils.py
benrudolph: 👍

corehq/apps/ota/views.py
benrudolph: ah interesting, didn't think about that, but yes that would make sense. is there additional work to making case claim location safe or can i just add the decorator? (if more work i might punt to a follow up PR)

corehq/apps/locations/permissions.py
sravfeyn: There is no reason, I overlooked this subtlety PR https://github.com/dimagi/commcare-hq/pull/14020/files

corehq/apps/ota/tests/test_utils.py
czue: isn't canada just the 51st state? so montreal is more like a county

corehq/apps/ota/utils.py
```diff
@@ -87,7 +90,50 @@ def demo_restore_date_created(commcare_user):
...
+    else:
+        if user_domain != domain:
+            # In this case we may be dealing with a WebUser
+            user = WebUser.get_by_username(as_user)
+            if not user or not user.is_member_of(domain):
+                raise RestorePermissionDenied(u"{} was not in the domain {}".format(username, domain))
+
+
+def _ensure_accessible_location(domain, couch_user, as_user):
+    if couch_user.is_commcare_user():
```
czue: not required for this project, but in some future world would we not also want to scope this for web users?

czue (commented):
nice, this looks great. a few questions / comments:

- it appears this also releases the feature from out behind a flag. was that intentional? i thought that might have been a longer conversation
- if we do plan to release we should figure out where in the pricing plan this goes. my guess is that it's a feature that only big projects want so we should put in, e.g. advanced (though we can make community(?) for web users to help with app testing)
- i'll need to check with the project team on permissions. if we give them `can_edit_data` they also get the exports tab and importer and stuff, right? that feels funny and guessing we will need to split it out.

benrudolph:
@czue 

> it appears this also releases the feature from out behind a flag. was that intentional? i thought that might have been a longer conversation

it was intentional, @amsagoff had wanted to take it out of feature flag world. i think the big question that amelia and i were debating was whether or not to hide it behind a pricing plan for app preview. however that's not as critical to do in this PR since app preview itself is still behind a flag. 

> if we do plan to release we should figure out where in the pricing plan this goes. my guess is that it's a feature that only big projects want so we should put in, e.g. advanced (though we can make community(?) for web users to help with app testing)

since cloudcare is behind a flag, this is default behind a pricing flag except for app preview. (and i don't think it's determined how to price that feature)

> i'll need to check with the project team on permissions. if we give them can_edit_data they also get the exports tab and importer and stuff, right? that feels funny and guessing we will need to split it out.

agree it's weird. happy to split out, but thought it might be yagni. the only reason i thought it made some sense is that a savvy mobile worker that has restore as permission can technically view most of the data that you could get from exports anyways and they could also manipulate the data by submitting forms.

corehq/apps/ota/utils.py
benrudolph: i think so, but i didn't think locations was really supported for web users. @esoergel is that true? i thought locations were only configurable for mobile workers

czue:
cool, all makes sense.

> agree it's weird. happy to split out, but thought it might be yagni. the only reason i thought it made some sense is that a savvy mobile worker that has restore as permission can technically view most of the data that you could get from exports anyways and they could also manipulate the data by submitting forms.

Yeah permissions wise it does make sense, I think the issue is going to be that people don't want to see the "data" tab. Also "edit data" comes with some weird stuff like lookup tables I think....

czue:
build fail looks real
```
ERROR: test suite for <class 'corehq.apps.ota.tests.test_utils.RestorePermissionsTest'>
  File "/vendor/lib/python2.7/site-packages/nose/suite.py", line 209, in run
    self.setUp()
  File "/vendor/lib/python2.7/site-packages/nose/suite.py", line 292, in setUp
    self.setupContext(ancestor)
  File "/vendor/lib/python2.7/site-packages/nose/suite.py", line 315, in setupContext
    try_run(context, names)
  File "/vendor/lib/python2.7/site-packages/nose/util.py", line 471, in try_run
    return func()
  File "/mnt/commcare-hq-ro/corehq/apps/ota/tests/test_utils.py", line 71, in setUpClass
    cls.restrict_user_to_location(cls.commcare_user)
AttributeError: type object 'RestorePermissionsTest' has no attribute 'restrict_user_to_location'
```

czue:
@benrudolph what do you think about including with "edit mobile workers" instead? it would meet the needs of the project and i can talk myself into thinking it makes sense (just in terms of - if you can become these people you can probably edit their info too...)

benrudolph:
that's fine with me. that was what i had originally had it under actually

corehq/apps/ota/views.py
proteusvacuum: I think we have to think about this a bit. I'll add it in a later PR.

corehq/apps/locations/permissions.py
```diff
@@ -323,13 +323,16 @@ def user_can_access_other_user(domain, user, other_user):
...
+        .values_list('id', flat=True))
 
     other_user_locations = other_user.sql_locations
-    return any(map(
-        lambda location_id: location_id in accessible_location_ids,
-        other_user_locations.values_list('id', flat=True) if other_user_locations else []
-    ))
+
+    if other_user_locations:
+        return other_user_locations.filter(id__in=accessible_location_ids).count() > 0
```
proteusvacuum: sorry... I think this might not work any more if https://github.com/dimagi/commcare-hq/pull/14020 gets merged, although I think usecases like this is a good reason to keep that as a queryset.

benrudolph: now it will 😄 but i think i can make it even simpler now that i don't have to ensure we're dealing with a query set over an array

benrudolph:
ok i think all feedback has been addressed. the one open item is when to show the Login As button for app preview. i'm going to punt that down the road since i don't think it's critical for this PR

czue:
Nice! No blockers on my side, but I didn't review code closely enough to feel like I can approve.

corehq/apps/ota/views.py
esoergel: I *think* it should work as long as the other decorators use `@wraps`.  It sets an attr on the function itself, and some decorators return a different, anonymous function.

corehq/apps/locations/permissions.py
```diff
@@ -319,6 +319,17 @@ def user_can_access_any_location_id(domain, user, location_ids):
...
 
+def user_can_access_other_user(domain, user, other_user):
+    if user.has_permission(domain, 'access_all_locations'):
+        return True
+
+    accessible_location_ids = (SQLLocation.objects
+        .accessible_to_user(domain, user)
+        .values_list('id', flat=True))
+
+    return other_user.sql_locations.filter(id__in=accessible_location_ids).count() > 0
```
proteusvacuum: Is there ever a time we'd want to use `sql_locations`? If not, I think we should make that property private (out of scope of this PR, but is obviously confusing).

benrudolph: ok, will do. i built most of this with the assumption that these queries will only work against commcare users. what is the status of webusers and locations? should that be first class support?

proteusvacuum (approved):

corehq/apps/locations/permissions.py
esoergel: You can also chain these filters together to avoid making two db calls, and I believe `.exists()` is more performant than `.count() > 0`:
```python
return (other_user.get_sql_locations(domain)
        .accessible_to_user(domain, user)
        .exists())
```
There's also a function in this file called `user_can_access_any_location_id` which does very much the same thing (it's used only in `can_edit_form_location`).  It may be prudent to combine the two (this abstraction seems more appropriate, I can't think of another use-case for `access_any_location_id` beyond `access_other_user`).

esoergel: > what is the status of webusers and locations? should that be first class support?

Yup.  The distinction between web and mobile users is pretty blurry, I'd support @proteusvacuum 's idea of dropping the `sql_locations` method and striving for API compatability.  Web users can edit and archive form submissions, and mobile workers can access some pages on HQ.  Additionally, we usually use `CouchUser.get` methods, which can return either, so it's hard to tell which one you may get.

benrudolph: now that i've realized locations and web users are a real thing, i'm going to update this code :)

corehq/apps/locations/permissions.py
```diff
@@ -327,7 +327,7 @@ def user_can_access_other_user(domain, user, other_user):
         .accessible_to_user(domain, user)
         .values_list('id', flat=True))
 
-    return other_user.sql_locations.filter(id__in=accessible_location_ids).count() > 0
+    return other_user.get_sql_locations(domain).filter(id__in=accessible_location_ids).exists()
```
benrudolph: 👍 will update

benrudolph: https://github.com/dimagi/commcare-hq/pull/14019/commits/899f17b29c5f1495f81ef3f0cc19ebd124dff114

corehq/apps/locations/permissions.py
```diff
@@ -323,11 +323,10 @@ def user_can_access_other_user(domain, user, other_user):
...
 
-    accessible_location_ids = (SQLLocation.objects
-        .accessible_to_user(domain, user)
-        .values_list('id', flat=True))
-
-    return other_user.get_sql_locations(domain).filter(id__in=accessible_location_ids).exists()
+    return (other_user
+            .get_sql_locations(domain)
+            .accessible_to_user(domain, user)
+            .exists())
```
esoergel: :+1: thankya!

corehq/apps/hqwebapp/templatetags/hq_shared_tags.py
```diff
@@ -227,14 +227,10 @@ def is_new_cloudcare(request):
 
 @register.filter
 def can_use_restore_as(request):
```
esoergel: :sunglasses:

esoergel (approved):

benrudolph:
✅ (added one more commit after approval)

snopoke (approved):

czue:
@benrudolph confirming you'll figure out something with @amsagoff on pricing stuff before this gets deployed?

benrudolph:
yeah i'll put it behind the pro plan

PR #14020, opened 2016-11-24
don't use queryset, return object list

sravfeyn:
@proteusvacuum

### DIFF
--- a/corehq/apps/users/models.py
+++ b/corehq/apps/users/models.py
@@ -1638,9 +1638,7 @@ class CommCareUser(CouchUser, SingleMembershipMixin, CommCareMobileContactMixin)
     @property
     def sql_locations(self):
         from corehq.apps.locations.models import SQLLocation
-        if self.assigned_location_ids:
-            return list(SQLLocation.objects.filter(location_id__in=self.assigned_location_ids))
-        return []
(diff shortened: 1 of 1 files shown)


czue (approved):
curious why we don't want it chainable, but makes sense to do this given that we don't.

confirming you double checked no one was _already_ using it as a queryset? feel free to self merge after you confirm.

proteusvacuum (approved):
Thanks for the quick followup! Also curious why we don't want this to be chainable? Having it chainable would allow for filtering this list down further directly in the database, and I don't think would change the way this method is called.

proteusvacuum:
see this commit for an argument in favour of having this method chainable: 
https://github.com/dimagi/commcare-hq/pull/14019/commits/9cc49a86d380f2432d10be013892137aa030e37f
(up to you though as to what you want to do)

sravfeyn:
Hey, I went back n forth between two ways without much arguments in favour for either of those things. I changed it to chainable.

benrudolph:
nice i'm a fan!

czue:
nice

esoergel:
Nice find, @proteusvacuum , and thanks for the update, @sravfeyn !  I agree with the arguments here and am a fan of having it chainable.

PR #14023, opened 2016-11-24
make the calendar fixture configurable per project

czue:
https://trello.com/c/ap6rEqVx/54-adherence-calendar-configurable-fixture-start-end-dates

this ended up being a bit bigger than I expected. should be pretty easy to read by commit though.

this:

- moves the calendar from custom.enikshay to a place in corehq (since it is already being used in other projects)
- adds a model to configure settings per domain
- adds tests
- adds a UI
- tweaks the project settings sidebar a bit

remaining todos (out of scope for this PR)

- [ ] Change the instance ID

@proteusvacuum / @orangejenny buddy


corehq/apps/calendar_fixture/tests.py
```diff
@@ -0,0 +1,57 @@
+import uuid
+from datetime import date, datetime, timedelta
```
dimagimon: F401 'date' imported but unused

PR #14025, opened 2016-11-25
make tableau url a setting

sravfeyn:
@calellowitz @czue

PR #14026, opened 2016-11-25
ssl free tableau auth ticket request

sravfeyn:
This is required for Tableau on nic domain to work till we get SSL. In terms of real security it doesn't matter if we turn off SSL, since no creds pass in the auth request. The auth works based on django server's IP being whitelisted on tableau machine.

@czue @calellowitz 

We could potentially wait for SSL to arrive and not merge this. But I want to test on live server as well to make sure everything is good (I have tested locally)


czue:
looks like this has https://github.com/dimagi/commcare-hq/pull/14025 inside it, which is also why tests are failing.

logic seems fine to me, but I'll let @calellowitz confirm. maybe add a todo to remove it as well if you are planning on having this merged?

sravfeyn:
I made a mess of the commit history. 
Here's PR with cleaner commit history https://github.com/dimagi/commcare-hq/pull/14027 Closing this out

PR #14027, opened 2016-11-25
ignore SSL temporarily

sravfeyn:
Same as https://github.com/dimagi/commcare-hq/pull/14026 with cleaner history


czue (approved):
good by me but would prefer @calellowitz sign off too

calellowitz:
This seems good by me. It will be nice to be able to test if the while we wait on the cert. all those requests are internal anyway so even if they had creds it would be fine. Will wait for the build anyway.

calellowitz (approved):

sravfeyn:
Closing this as I was able to test this locally. Once we get SSL we won't need this workaround.

PR #14030, opened 2016-11-26
Make add_inferred_export_properties serial

benrudolph:
@mkangia / anyone. seeing a lot of ResourceConflict errors coming from this task. this should fix


czue:
nice was noticing these as well

PR #14033, opened 2016-11-28
remove newlines from headers

sravfeyn:
http://manage.dimagi.com/default.asp?242422#1252717 @NoahCarnahan

### DIFF
--- a/corehq/ex-submodules/couchexport/shortcuts.py
+++ b/corehq/ex-submodules/couchexport/shortcuts.py
@@ -63,20 +63,12 @@ def export_response(file, format, filename, checkpoint=None):
         response = StreamingHttpResponse(FileWrapper(file), content_type=format.mimetype)
 
     if format.download:
-        try:
-            filename = unidecode(filename)
-        except Exception:
-            logging.exception("Error with filename: %r" % filename)
-            filename = "data"
-        finally:
-            header_value = 'attachment; filename="{filename}.{format.extension}"'.format(
-                filename=filename,
-                format=format
-            )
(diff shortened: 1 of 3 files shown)


corehq/ex-submodules/couchexport/shortcuts.py
```diff
@@ -69,10 +69,14 @@ def export_response(file, format, filename, checkpoint=None):
...
         finally:
-            response['Content-Disposition'] = 'attachment; filename="{filename}.{format.extension}"'.format(
+            header_value = 'attachment; filename="{filename}.{format.extension}"'.format(
                 filename=filename,
                 format=format
             )
+            # HTTP headers shouldn't contain newlines
+            header_value.replace("\n", " ")
+            header_value.replace("\r", " ")
+            response['Content-Disposition'] = header_value
```
czue: I wonder if we can have a more centralized/consistent way to do this? If you search the codebase for `attachment; filename=` you can find all sorts of different levels of cleanup done to workaround various bugs of this flavor.

I think the most comprehensive version is probably in the main exports code [here](https://github.com/dimagi/commcare-hq/blob/master/corehq/apps/reports/views.py#L574). wonder if we can extract that to a utility and use it here (and ideally, more widely).

sravfeyn: Ha, thanks for the comment. I will try to implement that in all places.

czue: (feel free to just do the easy/testable ones for now - think that would be a good start)

kaapstorm (approved):
:+1:

PR #14036, opened 2016-11-28
Fixed fixture view

kkrampa:
@czue 

I believe that this bug causes problems here: http://manage.dimagi.com/default.asp?237347

If you look at the original code there is iteration over an array using ```in``` operator which causes field_attribute to be an index of the array and in the result of that ``` field_for_attribute.field_value``` always returns null. It seems to me like an obvious bug.

I guess that reindexing the whole view is an expensive operation. I made this PR to raise the problem. (Probably using ```in``` to iterate over array is bad idea at all)

cc @sravfeyn Git history says you are the author of this code.

### DIFF
--- a/corehq/apps/fixtures/tests/test_fixture_data.py
+++ b/corehq/apps/fixtures/tests/test_fixture_data.py
@@ -164,3 +164,9 @@ class FixtureDataTest(TestCase):
                 self.domain, self.tag, 'state_name')
             delhi_id = fixtures['Delhi_state']['district_id']
             self.assertEqual(delhi_id, 'Delhi_id')
+
+    def test_get_item_by_field_value(self):
+        self.assertEqual(
+            FixtureDataItem.by_field_value(self.domain, self.data_type, 'state_name', 'Delhi_state').one().get_id,
+            self.data_item.get_id
+        )
--- a/corehq/apps/fixtures/_design/views/data_items_by_field_value/map.js
+++ b/corehq/apps/fixtures/_design/views/data_items_by_field_value/map.js
@@ -4,7 +4,8 @@ function (doc) {
             if (doc.fields.hasOwnProperty(key)) {
                 if (doc.fields[key] && doc.fields[key].doc_type === 'FieldList'){
                     for (var field_for_attribute in doc.fields[key].field_list){
-                        emit([doc.domain, doc.data_type_id, key, field_for_attribute.field_value]);
+                        var field = doc.fields[key].field_list[field_for_attribute];
+                        emit([doc.domain, doc.data_type_id, key, field.field_value]);
                     }
                 }
                 else {


czue:
huh, i agree this looks like it was obviously wrong. @sravfeyn thoughts?

can we write a test for this case? it's mind-boggling to me that this has been completely broken this whole time and we haven't encountered it (though I admittedly don't know what the implications of this would be)

corehq/apps/fixtures/tests/test_fixture_data.py
```diff
@@ -164,3 +164,6 @@ def test_get_indexed_items(self):
                 self.domain, self.tag, 'state_name')
             delhi_id = fixtures['Delhi_state']['district_id']
             self.assertEqual(delhi_id, 'Delhi_id')
+
+    def test_get_item_by_field_value(self):
+        self.assertEqual(FixtureDataItem.by_field_value(self.domain, self.data_type, 'state_name', 'Delhi_state').one().get_id, self.data_item.get_id)
```
dimagimon: E501 line too long (150 > 115 characters)

kkrampa:
@czue 
Added

sravfeyn:
That looks like a disaster on my part. The change here look good. I am doing a quick review of code/features that use this view and see if anything will break.

sravfeyn:
(The key point to understand following is that fixture fields can be simple `{'city_name': Delhi}` or advanced 
```
{"field_list": [
                {"field_value": "Delhi_IN_ENG", "properties": {"lang": "eng"}},
                {"field_value": "Delhi_IN_HIN", "properties": {"lang": "hin"}},
]}
```
)
The view is mainly used in fixture drillable filter `AsyncDrillableFilter` (which is used in bunch of custom report filters). The reason this hasn't been a problem is because the projects of these custom reports don't use dict for fixtures fields. In fact, the filtering is not supported in that advanced case. `AsyncDrillableFilter` will raise an exception of `FixtureVersionError` if it's used in projects that have this advanced setup. So, it didn't matter whether this view worked for the advanced case or not. It's good to fix the view anyway. But the filtering support is not yet there for advanced use case where this advanced part of view is really useful.

`mc-inscale` doesn't use advanced fixture setup, so I doubt if this PR will actually fix the problem. @kkrampa Did you test if the bug gets fixed with this.

kkrampa:
@sravfeyn 
It's hard to test this on my local machine since I don't have this domain replicated so there is a chance that this PR doesn't help at all in mc-inscale. 
I only imported fixtures from that domain and this change definitely fixed a problem for that imported fixtures.

sravfeyn:
Cool, I am merging this in, since this has to be fixed anyway

czue:
@sravfeyn just flagging this requires a couch view reindex. would you mind kicking that off now? not sure how long this will take though i think it should just reindex the fixtures db so hopefully pretty fast. if it looks like that's not the case we might want to revert and merge/reindex over the weekend

sravfeyn:
Yeah, I am just looking at the reindex and was seeing if there is a way to do selective reindex. I will offline with simon on how to do this smoothly

sravfeyn:
I've started a preindex at 6pm IST using `setup_release` (via screen). Will check back in another two hours and I will revert this PR if it doesn't get finished.

czue:
Cool. was there a reason not to just do a normal preindex?

sravfeyn:
The preindex is complete on prod. You mean, preindex along with deploy? I chatted with simon and thought this was good option in case we had to rollback.

czue:
sorry this isn't important and you can ignore. in my head you can easily
just run a normal preindex, and you can revert and preindex/deploy again if
you want (it shouldn't do any extra work the second time). but it's
entirely possible that I haven't thought that through.

PR #14037, opened 2016-11-28
Gracefully handle deleted locations

kaapstorm:
Context: [FB 242342](http://manage.dimagi.com/default.asp?242342)

When locations are deleted from the UI, HQ cleans up mobile workers that are assigned to them, but not when deleting locations using a bulk upload.

This PR is to allow mobile workers that refer to deleted locations to be edited and saved without throwing a 500.

@sravfeyn cc @nickpell

### DIFF
--- a/corehq/apps/locations/templates/locations/manage/partials/autocomplete_select_widget.html
+++ b/corehq/apps/locations/templates/locations/manage/partials/autocomplete_select_widget.html
@@ -36,6 +36,12 @@ $("#{{ id }}").select2({
             $.get("{{ query_url }}", payload).done(function(data) {
                 callback(data);
                 $(element).trigger('select-ready');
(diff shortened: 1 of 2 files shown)


mkangia:
@kaapstorm Did we add something to take care of this? Remove users in case of bulk deletions and also fix exiting cases? 
The message is suppressed but the user won't see any locations on the mobile worked page until the deleted location is removed from the assigned locations.
Something similar to what i had to do at http://manage.dimagi.com/default.asp?242136#bugevent_1254428

kaapstorm:
@mkangia I agree. I think we need to do both, and this PR doesn't go far enough. I've added it to my to-do list and I'll log a ticket for it.

esoergel:
@kaapstorm can you ping me on the ticket, please?

PR #14038, opened 2016-11-28
exclude archived products from inventory report

czue:
http://manage.dimagi.com/default.asp?242407

@esoergel / cc @orangejenny


czue:
in full transparency I should mention that I didn't run this code but it seems obviously correct

esoergel (approved):

PR #14041, opened 2016-11-28
Max height for tablet mode

benrudolph:
@biyeun / anyone, this adds max height to tablet mode so we don't overflow the screen. now we can have chubby tablets 😄 
screen shot 2016-11-28 at 11 06 27 am

fixes this: http://manage.dimagi.com/default.asp?242464
have you ever used the `calc` function? it seems ridiculously powerful. i don't understand how it's able to work (http://lesscss.org/usage/)


biyeun:
@benrudolph there's some caution about using calc as it really slows down rendering of the dom. also not sure how cross-browser friendly it is

biyeun:
ah, this is the calc that's part of less? that's different. the one that is built into css is the problem.

benrudolph:
aha i think this is the css one. looks like it has decent support if i put the proper prefixes: http://css3clickchart.com/#calc. unclear about the performance, this is by far the simplest and cleanest solution so i'm tempted to just use it. there were no noticable performance differences

PR #14042, opened 2016-11-28
Locations in webuser restore

benrudolph:
@proteusvacuum @esoergel this implements locations for web user restore. it also makes `sql_locations` private. i was going to do the same for `sql_location`, but that looked way way more painful to pinpoint all the usages. want to test on a phone before merging just in case, but works fine on app preview which should be the same thing. also i removed the public api calls to `set_location` and `add_to_assigned_locations`. they were only used in tests, i don't think we should make it feel supported. the restore user is essentially a read only thing that shouldn't be modifying any of the underlying data. here's what my web user's payload looks like:
```xml
<fixture id="commtrack:locations" user_id="8d7a5c6e0e1af478aca5a39bebe5317c">
   <countrys>
      <country id="f655a5755f32e3b3032c5b48960dc220">
         <name>USA</name>
         <site_code>usa</site_code>
         <external_id />
         <latitude />
         <longitude />
         <location_type>Country</location_type>
         <supply_point_id>dce2a8c54d934352bf0b5be0927f7875</supply_point_id>
         <location_data />
         <states>
            <state id="f655a5755f32e3b3032c5b48960dbf7d">
               <name>MA</name>
               <site_code>ma</site_code>
               <external_id />
               <latitude />
               <longitude />
               <location_type>State</location_type>
               <supply_point_id>6911d49a271648fda38034a54588e534</supply_point_id>
               <location_data />
            </state>
         </states>
      </country>
   </countrys>
</fixture>
```

cc: @mkangia

### DIFF
--- a/corehq/apps/locations/tests/test_location_fixtures.py
+++ b/corehq/apps/locations/tests/test_location_fixtures.py
@@ -84,7 +84,7 @@ class LocationFixturesTest(LocationHierarchyPerTest, FixtureHasLocationsMixin):
         self.assertXmlEqual(empty_fixture, fixture)
 
     def test_simple_location_fixture(self, uses_locations):
-        self.user.set_location(self.locations['Suffolk'].couch_location)
+        self.user._couch_user.set_location(self.locations['Suffolk'].couch_location)
 
         self._assert_fixture_has_locations(
             'simple_fixture',
@@ -92,8 +92,8 @@ class LocationFixturesTest(LocationHierarchyPerTest, FixtureHasLocationsMixin):
         )
 
     def test_multiple_locations(self, uses_locations):
-        self.user.add_to_assigned_locations(self.locations['Suffolk'].couch_location)
-        self.user.add_to_assigned_locations(self.locations['New York City'].couch_location)
+        self.user._couch_user.add_to_assigned_locations(self.locations['Suffolk'].couch_location)
+        self.user._couch_user.add_to_assigned_locations(self.locations['New York City'].couch_location)
 
         self._assert_fixture_has_locations(
             'multiple_locations',
@@ -137,7 +137,7 @@ class LocationFixturesTest(LocationHierarchyPerTest, FixtureHasLocationsMixin):
             Mass
             - Suffolk
         """
-        self.user.set_location(self.locations['Suffolk'].couch_location)
+        self.user._couch_user.set_location(self.locations['Suffolk'].couch_location)
         location_type = self.locations['Suffolk'].location_type
         location_type.expand_to = location_type
         location_type.save()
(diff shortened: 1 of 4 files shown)


corehq/apps/users/models.py
```diff
@@ -1636,7 +1636,7 @@ def sql_location(self):
         return None
 
     @property
-    def sql_locations(self):
+    def _sql_locations(self):
```
esoergel: Can we just drop this and use `get_sql_locations` even internally?

esoergel (commented):
This looks great, and I agree with your logic on the removed methods.  As a sanity check, you could look into making some of the existing location fixture tests run also with a web user in addition to a mobile user.  If there are weird edge-cases we haven't thought of, that'd be a good way to find them.

corehq/apps/users/models.py
benrudolph: sounds good to me

benrudolph:
yeah good point, i wanted to add some web user tests as well

corehq/apps/locations/tests/test_location_fixtures.py
```diff
@@ -216,6 +216,59 @@ def test_include_without_expanding(self, uses_locations):
...
+        self.user._couch_user.set_location(self.domain, self.locations['Suffolk'].couch_location)
+
+        self._assert_fixture_has_locations(
+            'simple_fixture',
+            ['Massachusetts', 'Suffolk', 'Boston', 'Revere']
+        )
+
+    def test_multiple_locations(self, uses_locations):
+        self.user._couch_user.add_to_assigned_locations(self.domain, self.locations['Suffolk'].couch_location)
+        self.user._couch_user.add_to_assigned_locations(self.domain, self.locations['New York City'].couch_location)
```
dimagimon: E501 line too long (116 > 115 characters)

proteusvacuum:
👢 👷 , but there's also conflicts

benrudolph:
Ok I'm happy with this. I got it all set up on a phone and then remembered you can't login as a web user on a mobile device 😥 but i did log in with a mobile user that had locations so it didn't break that. it works in app preview and passes tests

corehq/ex-submodules/casexml/apps/phone/models.py
```diff
@@ -80,12 +80,10 @@ def locations(self):
     @property
     def sql_location(self):
```
esoergel: [Not blocking] It may be nice to give `OTARestoreUser` the same API as `CouchUser`, that is: `user.get_sql_location(domain)` and `user.get_sql_locations(domain)`.

esoergel (approved):

esoergel:
Cool stuff!

PR #14047, opened 2016-11-28
Scope breadcrumb styles

benrudolph:
@wpride / anyone. so old cloudcare doesn't look absolutely ridiculous


wpride:
Old Cloudcare, I will have such revenges on you... What they are, yet I know not: but they shall be The terrors of the earth

benrudolph:
Old Cloudcare, O hound of Crete, think'st thou my spouse to get? No to the spital you go, And from the powd'ring tub of infamy Fetch forth lazar kite of Cressid's kind, Doll Tearsheet she by name, and her espouse

PR #14052, opened 2016-11-29
Custom instance injection

proteusvacuum:
Allows users to add custom instance declarations to the suite file.
http://manage.dimagi.com/default.asp?242370#1252888
☣️ 
FYI @ctsims 
code buddy: @emord, @dannyroberts

### DIFF
--- a/corehq/apps/app_manager/views/forms.py
+++ b/corehq/apps/app_manager/views/forms.py
@@ -335,9 +335,17 @@ def _edit_form_attr(request, domain, app_id, unique_form_id, attr):
         instances = json.loads(request.POST.get('custom_instances'))
         try:         # validate that custom instances can be added into the XML
             for instance in instances:
-                etree.fromstring("<instance id='{}' src='{}' />".format(instance.get('instanceId'), instance.get('instancePath')))
+                etree.fromstring(
+                    "<instance id='{}' src='{}' />".format(
+                        instance.get('instanceId'),
+                        instance.get('instancePath')
+                    )
+                )
         except etree.XMLSyntaxError as error:
-            return json_response({'message': _("There was an issue with your custom instances: {}").format(error.message)}, status_code=400)
+            return json_response(
+                {'message': _("There was an issue with your custom instances: {}").format(error.message)},
+                status_code=400
+            )
 
         form.custom_instances = [
             CustomInstance(
--- a/corehq/apps/app_manager/views/forms.py
+++ b/corehq/apps/app_manager/views/forms.py
@@ -77,6 +77,7 @@ from corehq.apps.app_manager.models import (
     ModuleNotFoundException,
     load_case_reserved_words,
     WORKFLOW_FORM,
+    CustomInstance,
 )
 from corehq.apps.app_manager.decorators import no_conflict_require_POST, \
     require_can_edit_apps, require_deploy_apps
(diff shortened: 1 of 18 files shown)


corehq/apps/app_manager/views/forms.py
```diff
@@ -330,6 +331,21 @@ def should_edit(attribute):
             ]
         ) for link in form_links]
 
+    if should_edit('custom_instances'):
+        instances = json.loads(request.POST.get('custom_instances'))
+        try:         # validate that custom instances can be added into the XML
+            for instance in instances:
+                etree.fromstring("<instance id='{}' src='{}' />".format(instance.get('instanceId'), instance.get('instancePath')))
+        except etree.XMLSyntaxError as error:
+            return json_response({'message': _("There was an issue with your custom instances: {}").format(error.message)}, status_code=400)
```
dimagimon: E501 line too long (140 > 115 characters)

czue:
build failing on app builder tests
```
    self.partial_v2_dir
AssertionError: 
APP MANAGER V2 Diff Failure
An edit made to a V1 App Manager file "form_tab_settings.html" does not match the stored diff
of its V2 counterpart.
to rebuild the broken diffs.
These files are located in
/mnt/commcare-hq-ro/corehq/apps/app_manager/tests/../templates/app_manager/v1/partials/form_tab_settings.html
/mnt/commcare-hq-ro/corehq/apps/app_manager/tests/../templates/app_manager/v2/partials/form_tab_settings.html
**************************************
```

benrudolph:
the complex feature master strikes again! :crossed_swords:

snopoke:
Can you motivate why this is necessary and we can't infer which instances are required by interrogating the forms etc?

proteusvacuum:
@snopoke: @ctsims and I discussed this, here's what he said.
> There are three reasons that are relevant to me
> 1) HQ isn't actually able to automate inferring an instance ID from a connector id, so we have to maintain a table 
> 2) HQ would need to add hooks to dozens of places to spot that we are trying to connect instances in there, and we add more with some frequency
> 3) This is already an incredibly advanced use case and I don't think we are adding a ton of value to the world by making what is already a complex process 1% easier v. doing a ton of magic
> 
> [10:58]  
> Mostly this is one of those places where we're just revealing a more and more low-level capacity to just tell the phone to do stuff
> 
> [10:58]  
> and in my experience that process isn't improved by trying to do that "Better"
> 
> [10:58]  
> since people who are doing that stuff arleday know what they want

The current usecase is to use a fixture in detail screen variable declarations that are added through a custom block of XML. In theory we could add the fixture instance automatically pretty easily, but sounds like clayton has some other ideas for places he wanted to use this.

emord (approved):
Looks good to me, but will let @snopoke sign off on your comment

snopoke:
Seems OK for this use case I'm just concerned we're going to get into a place where we're using this back door too much and making apps harder to build / maintain. There is already an item on the GS Efficiency list for improving error messages etc around missing instances.

@dimagi/product this is one of those cases were it would be nice to be able to restrict who can add domains to this toggle.

benrudolph:
might be a dumb question, but why don't we just include every instance in every form? is that a huge performance drain or can mobile decide which ones it needs?

proteusvacuum:
@benrudolph not a dumb question. Theoretically it is totally possible to go through and pull out the references to all the instances everywhere. In this case, it is a bit annoying since the instance is within a blob of custom XML, and we were using a hack to get the instance included where we wanted it. 

I think it is a tradeoff between being "magic" and being "explicit". Since this is a super advanced use case, I think opting for "explicit" has some merit here.

benrudolph:
gotcha, makes sense. thanks for the explanation

PR #14054, opened 2016-11-29
unicode

sravfeyn:
I reminded myself I should put `u` and I forgot before I pushed and committed http://manage.dimagi.com/default.asp?242650

@kaapstorm


kaapstorm:
:+1: Last commit shouldn't affect build.

PR #14040, opened 2016-11-28
soft_assert on bad path to item in export

kaapstorm:
Context: [FB 242106](http://manage.dimagi.com/default.asp?242106)

@czue @benrudolph not sure whether this is the best approach. cc @sravfeyn @nickpell

PR #14055, opened 2016-11-29
Catch ValueError

kaapstorm:
A simpler response to [FB 242106](http://manage.dimagi.com/default.asp?242106); just catch the ValueError if the path is empty in a custom form export.

Supersedes PR #14040. 

@benrudolph @sravfeyn @nickpell


benrudolph:
thanks!

PR #12282, opened 2016-06-29
Created new percent active graph and display active, inactive, and total number of users 

rickypchen:
Addressed changes proposed in google doc: https://docs.google.com/document/d/1DrXuXIXVCDfvEaDRoBX6klct7FP9EaNRcyl3Bc6qeTc/edit#

Highlights:
1. categories changed to high performing and low performing
2. new active users chart (% of active users) and tooltip displaying # of active/inactive/total users 

pinging @benrudolph for code review

PR #14056, opened 2016-11-29
project health fixes

czue:
Fixes http://manage.dimagi.com/default.asp?242542, which I think has been broken since https://github.com/dimagi/commcare-hq/pull/12282/

Was a pretty hard-to-track-down bug and I think points to an issue with how this report generates its data. Applied a bit of a band aid to help catch issues like this preemptively.

read by commit. @orangejenny / anyone


czue:
@NoahCarnahan

PR #14059, opened 2016-11-29
require cloudcare privilege for login as

benrudolph:
@czue @amsagoff puts login as behind cloudcare priv 💰


amsagoff:
Thanks! I still need to discuss pricing with Ryan, so there's a chance this
may not be the final plan, but it seems likely and certainly no harm in
releasing under Pro now.

czue (approved):

PR #12076, opened 2016-06-15
port the rest of aggregated_data path of aggregate inventory report to ES

czue:
probably easiest :fish: 

@snopoke cc @gcapalbo

PR #14062, opened 2016-11-29
Fix bottom level aggregation

emord:
@czue I think this was your original intention, but correct me if I'm wrong https://github.com/dimagi/commcare-hq/pull/12076 and https://github.com/dimagi/commcare-hq/pull/12102

Right now it's creating something like

```
agg1: {
   nested_agg2: {
      nested_agg3: {}
   },
   bottom_agg: {}
}
```

but I think you want `bottom_agg` inside of `nested_agg3` right? 

This hasn't come up before because it's only used in two places and each only has one aggregation term

buddy @benrudolph

### DIFF
--- a/corehq/apps/es/aggregations.py
+++ b/corehq/apps/es/aggregations.py
@@ -556,21 +556,25 @@ class NestedTermAggregationsHelper(object):
         self.terms = terms
         self.bottom_level_aggregation = bottom_level_aggregation
 
-    def get_data(self):
+    @property
+    def query(self):
         previous_term = None
+
+        if self.bottom_level_aggregation is not None:
+            if not isinstance(self.bottom_level_aggregation, SumAggregation):
+                raise ValueError('currently only SumAggregations and its children are supported')
+            term = self.bottom_level_aggregation
+            previous_term = term
+
         for name, field in reversed(self.terms):
             term = TermsAggregation(name, field)
             if previous_term is not None:
                 term = term.aggregation(previous_term)
             previous_term = term
 
-        if self.bottom_level_aggregation is not None:
-            if not isinstance(self.bottom_level_aggregation, SumAggregation):
-                raise ValueError('currently only SumAggregations are supported')
-            term = term.aggregation(self.bottom_level_aggregation)
+        return self.base_query.aggregation(term)
 
-        query = self.base_query.aggregation(term)
-
+    def get_data(self):
         def _add_terms(aggregation_bucket, term, remaining_terms, current_counts, current_key=None):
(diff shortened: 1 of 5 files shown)


corehq/apps/es/aggregations.py
```diff
@@ -547,49 +547,53 @@ class NestedTermAggregationsHelper(object):
...
+        self.inner_most_aggregation = inner_most_aggregation
 
-    def get_data(self):
+    @property
+    def query(self):
         previous_term = None
+
+        if self.inner_most_aggregation is not None:
+            if not isinstance(self.inner_most_aggregation, SumAggregation):
+                raise ValueError('currently only SumAggregations and its children are supported')
```
emord: There are some aggregations like `MinAggregation` that inherit from `SumAggregation` now. They should still work in this helper although I haven't explictly tested that.

czue:
Man, reading elastic queries makes my head hurt... 

I believe you that your thing is right, but just wanted to check:

> This hasn't come up before because it's only used in two places and each only has one aggregation term

Isn't it using multiple terms [here](https://github.com/dimagi/commcare-hq/blob/master/corehq/apps/data_analytics/esaccessors.py#L14)? Nevertheless, that code is tested [here](https://github.com/dimagi/commcare-hq/blob/master/corehq/apps/data_analytics/tests/test_esaccessors.py#L48) and this didn't seem to break the build, so maybe they are equivalent?

czue:
sorry I figured this out, `bottom_level_aggregation` is only used in two places with 1 term. makes sense.

I know I should have done this originally when I wrote it, but would it be possible to add a test for not just the query structure but the results? Thinking we just need a to exercise `get_aggregated_ledger_values`. Then we could be confident this does the right thing before and after. Let me know if that sounds like too much work though.

emord:
Added some tests. 

FWIW https://github.com/dimagi/commcare-hq/blob/master/corehq/apps/reports/tests/test_data_sources.py#L20 would have tested it, if it had run in the past 3 years (https://github.com/dimagi/commcare-hq/commit/2ee61d1a0199ff4a9674798a10fcd5b0efb552bf) ... Doubt it's a simple fix to get it running again (something about adding a location to a fake domain?). Should it just be removed?

czue:
> FWIW https://github.com/dimagi/commcare-hq/blob/master/corehq/apps/reports/tests/test_data_sources.py#L20 would have tested it, if it had run in the past 3 years (2ee61d1) ... Doubt it's a simple fix to get it running again (something about adding a location to a fake domain?). Should it just be removed?

yikes - yeah seems like we should delete it as that comment was clearly aspirational. i can do that tomorrow if you don't get to it

PR #14066, opened 2016-11-30
fix restore perms

czue:
http://manage.dimagi.com/default.asp?242732

@benrudolph not sure if this is the most elegant fix. 

cc @NoahCarnahan

### DIFF
--- a/corehq/apps/ota/utils.py
+++ b/corehq/apps/ota/utils.py
@@ -105,14 +105,24 @@ def is_permitted_to_restore(domain, couch_user, as_user, has_data_cleanup_privil
     try:
         _ensure_valid_domain(domain, couch_user)
         if as_user is not None:
-            _ensure_cleanup_permission(domain, couch_user, as_user, has_data_cleanup_privilege)
-            _ensure_valid_restore_as_user(domain, couch_user, as_user)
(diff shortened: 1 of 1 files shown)


NoahCarnahan (approved):
👍

czue:
i'm just gonna deploy this. seems to be causing pain for a few different people

benrudolph:
Ah shoot. Forgot we did this in old cloudcare. Thanks, I'll follow up with
a test case for this scenario later today

czue:
nice, yeah that'd be good. feel free to also test [this case](https://github.com/dimagi/commcare-hq/pull/14075), or find a smarter way to address these if that makes more sense (i didn't go down the path of figuring out why `as_user` was being set to the same as the calling user)

benrudolph:
Ok, will do. It's an oddity about cloudcare sqlite restore. We always send
the as user whether we need it or not and don't suffix mobile workers with
commcarehq.org which adds a fair bit of complexity that I want to get rid of

czue:
> and don't suffix mobile workers with commcarehq.org which adds a fair bit of complexity that I want to get rid of

haha, yeah just been wrestling with this and was much more annoying than I expected

PR #14067, opened 2016-11-30
Return error, not 500, on AdminRestoreView error

kaapstorm:
Context: [FB 242755](http://manage.dimagi.com/default.asp?242755)

AdminRestoreView.get_context_data wasn't handling 500 responses.

@snopoke cc @sravfeyn @wspride buddy @biyeun

### DIFF
--- a/corehq/apps/hqadmin/views.py
+++ b/corehq/apps/hqadmin/views.py
@@ -497,19 +497,21 @@ class AdminRestoreView(TemplateView):
             xml_payload = etree.fromstring(string_payload)
             restore_id_element = xml_payload.find('{{{0}}}Sync/{{{0}}}restore_id'.format(SYNC_XMLNS))
             num_cases = len(xml_payload.findall('{http://commcarehq.org/case/transaction/v2}case'))
-            formatted_payload = etree.tostring(xml_payload, pretty_print=True)
         else:
-            # get_restore_response return HttpResponse 401 (permissions), 404 (user not found) or 412
             if response.status_code in (401, 404):
-                message = response.content
+                # corehq.apps.ota.views.get_restore_response couldn't find user or user didn't have perms
+                xml_payload = E.error(response.content)
+            elif response.status_code == 412:
+                # RestoreConfig.get_response returned HttpResponse 412. Response content is already XML
+                xml_payload = response.content
             else:
                 message = _('Unexpected restore response {}: {}. '
                             'If you believe this is a bug please report an issue.').format(response.status_code,
                                                                                            response.content)
-            error = E.error(message)
-            formatted_payload = etree.tostring(error)
+                xml_payload = E.error(message)
             restore_id_element = None
             num_cases = 0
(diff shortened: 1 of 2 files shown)


czue:
I think this should be closed in favor of https://github.com/dimagi/commcare-hq/pull/14066

czue:
just thinking that we probably do still want to get 500s when this page fails

czue:
actually offlined with @kaapstorm - i think this is probably a good change but we might want to make sure we get notified on errors we don't expect. not sure if worth digging into that now or leaving as-is since this doesn't come up too frequently.

corehq/apps/hqadmin/views.py
```diff
@@ -490,11 +492,18 @@ def get_context_data(self, **kwargs):
...
-        num_cases = len(xml_payload.findall('{http://commcarehq.org/case/transaction/v2}case'))
-        formatted_payload = etree.tostring(xml_payload, pretty_print=True)
+        if isinstance(response, StreamingHttpResponse):
+            string_payload = ''.join(response.streaming_content)
+            xml_payload = etree.fromstring(string_payload)
+            restore_id_element = xml_payload.find('{{{0}}}Sync/{{{0}}}restore_id'.format(SYNC_XMLNS))
+            num_cases = len(xml_payload.findall('{http://commcarehq.org/case/transaction/v2}case'))
+            formatted_payload = etree.tostring(xml_payload, pretty_print=True)
+        else:
+            # response is HttpResponse 500
```
czue: we don't actually know this to be true right - just that it's an `HttpResponse`? could add a check for the status code or just update the comment to say something like "response is HttpResponse which is unexpected - tell the user something might have gone wrong".

Then I think we should also add a `messages.error` message along the lines of: "unexpected restore response - if you believe this is a bug please report an issue" so it's clear something has gone wrong to the user and they can do something about it.

kaapstorm:
Reopening with better error message.

Example 401 response in a browser:
restore_response

corehq/apps/hqadmin/views.py
```diff
@@ -490,11 +492,24 @@ def get_context_data(self, **kwargs):
...
-        formatted_payload = etree.tostring(xml_payload, pretty_print=True)
+        if isinstance(response, StreamingHttpResponse):
+            string_payload = ''.join(response.streaming_content)
+            xml_payload = etree.fromstring(string_payload)
+            restore_id_element = xml_payload.find('{{{0}}}Sync/{{{0}}}restore_id'.format(SYNC_XMLNS))
+            num_cases = len(xml_payload.findall('{http://commcarehq.org/case/transaction/v2}case'))
+            formatted_payload = etree.tostring(xml_payload, pretty_print=True)
+        else:
+            # get_restore_response return HttpResponse 401 (permissions), 404 (user not found) or 412
+            if response.status_code in (401, 404):
```
czue: should `412` be in this list too based on above comment? that makes sense to me as those errors are also useful.

czue (approved):
looks good, one possible non-blocking followup

PR #14071, opened 2016-11-30
management command to rebuild ucr in a domain

sravfeyn:
@gcapalbo 
cc @snopoke

### DIFF
--- a/corehq/apps/userreports/management/commands/rebuild_tables_by_domain.py
+++ b/corehq/apps/userreports/management/commands/rebuild_tables_by_domain.py
@@ -19,4 +19,4 @@ class Command(BaseCommand):
         print "Rebuilding {} tables".format(len(tables))
 
         for table in tables:
-            tasks.rebuild_indicators(table.config_id)
+            tasks.rebuild_indicators(table._id)
--- a/corehq/apps/userreports/management/commands/rebuild_tables_by_domain.py
+++ b/corehq/apps/userreports/management/commands/rebuild_tables_by_domain.py
@@ -1,0 +1,22 @@
+from django.core.management.base import BaseCommand, CommandError
+from corehq.apps.userreports import tasks
+from corehq.apps.userreports.models import DataSourceConfiguration, StaticDataSourceConfiguration
+
+
+class Command(BaseCommand):
+    help = "Rebuild all user configurable reporting tables in domain"
+    args = 'domain'
+    label = ""
+
+    def handle(self, *args, **options):
+        if len(args) < 1:
+            raise CommandError('Usage is rebuild_tables_by_domain %s' % self.args)
+
+        domain = args[0]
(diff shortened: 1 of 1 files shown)


corehq/apps/userreports/management/commands/rebuild_tables_by_domain.py
```diff
@@ -0,0 +1,22 @@
...
+    help = "Rebuild all user configurable reporting tables in domain"
+    args = 'domain'
+    label = ""
+
+    def handle(self, *args, **options):
+        if len(args) < 1:
+            raise CommandError('Usage is rebuild_tables_by_domain %s' % self.args)
+
+        domain = args[0]
+        tables = StaticDataSourceConfiguration.by_domain(domain)
```
sravfeyn: The `by_domain` method doesn't support `include_docs`. Should I add support for that and use it here to set it to `False`.

snopoke: why would you need `include_docs`? It already returns the docs.

corehq/apps/userreports/management/commands/rebuild_tables_by_domain.py
```diff
@@ -0,0 +1,22 @@
...
+        tables.extend(DataSourceConfiguration.by_domain(domain))
+
+        print "Rebuilding {} tables".format(len(tables))
+
+        for table in tables:
+            tasks.rebuild_indicators(table.config_id)
```
snopoke: I think this should be `table._id`

snopoke (changes requested):

corehq/apps/userreports/management/commands/rebuild_tables_by_domain.py
sravfeyn: I meant so that I can exclude, by setting `include_docs=False`. The default is that it includes.

snopoke: don't think it's really necessary

snopoke (approved):

corehq/apps/userreports/management/commands/rebuild_tables_by_domain.py
```diff
@@ -0,0 +1,22 @@
...
+            raise CommandError('Usage is rebuild_tables_by_domain %s' % self.args)
+
+        domain = args[0]
+        tables = StaticDataSourceConfiguration.by_domain(domain)
+        tables.extend(DataSourceConfiguration.by_domain(domain))
+
+        print "Rebuilding {} tables".format(len(tables))
+
+        for table in tables:
+            tasks.rebuild_indicators(table._id)
```
gcapalbo: not blocking but should this spawn an asynchronous task rather than process all of the tables synchronously? (i'm not saying necessarily it should, just something to think about if these tasks take a long time)

PR #14077, opened 2016-11-30
django 1.10

nickpell:
opening for tests

### DIFF
--- a/requirements/requirements.txt
+++ b/requirements/requirements.txt
@@ -7,7 +7,7 @@ jsonobject-couchdbkit==0.8.0.1
 celery==3.1.18
 # required by django-digest
 decorator==4.0.4
-django==1.9.12
+django==1.10.5
 django-braces==1.8.1
 django-celery==3.2.1
 django-crispy-forms==1.6.0
--- a/requirements/requirements.txt
+++ b/requirements/requirements.txt
@@ -7,7 +7,7 @@ jsonobject-couchdbkit==0.8.0.1
 celery==3.1.18
 # required by django-digest
 decorator==4.0.4
-django==1.10.5
+django==1.10.6
 django-braces==1.8.1
 django-celery==3.2.1
 django-crispy-forms==1.6.0


.travis.yml
```diff
@@ -22,7 +22,7 @@ install:
 before_script:
   - docker version
   - docker-compose version
-script: "scripts/docker test --noinput --stop --verbosity=2 --divide-depth=1"
+script: "scripts/docker test --noinput --verbosity=2 --divide-depth=1"
```
millerdev: Curious why `--stop` was removed?

nickpell: just for testing this PR so I can find >1 bug per iteration on travis

nickpell: will remove before this PR gets merged

snopoke:
@nickpell just wondering if it's a good idea to upgrade to a non-LTS release? The next LTS release (1.11) is due in April 2017.

nickpell:
My two cents: It's easier to manage upgrading 1.8 -> 1.9 -> 1.10 -> 1.11 than 1.8 -> 1.11.  The 1.9 upgrade is ready to go whenever we pull the trigger.  As long as we get onto 1.10 by April 2017 we'll receive security and major bug fixes until December 2017.  I figure that should be enough time to get onto 1.11.

@snopoke

millerdev:
FWIW, I'm with @nickpell on this. It's usually a lot easier to upgrade pervasive libraries like Django more frequently rather than waiting between each LTS release. Upgrading sooner means recommended new code patterns can be used sooner, and theoretically less code needs to be visited during each upgrade effort as a result.

snopoke:
I generally agree with those arguments and am happy for us to proceed, just wanted to make sure we were being intentional about it.

corehq/ex-submodules/casexml/apps/case/tests/util.py
```diff
@@ -207,3 +207,7 @@ def delete_all_xforms():
 @unit_testing_only
 def delete_all_sync_logs():
     FormProcessorTestUtils.delete_all_sync_logs()
+
+@unit_testing_only
```
dimagimon: E302 expected 2 blank lines, found 1

nickpell:
Fixes all known incompatibilities and all the bugs found on staging in the past month.  Planning to release Wed morning EST.

snopoke:
Sorry to put a damper on this but we need to hold off on merging this until we can sort out this issue:
* https://stackoverflow.com/questions/40851475/pass-django-csrf-token-to-angular-with-csrf-cookie-httponly

For NIC security audit we have to change CSRF token to HTTP-ONLY which means that it isn't accessible to JS. The workaround for that is this: http://stackoverflow.com/a/30656040/632517 but it doesn't work in Django 1.10 because it resets the CSRF token on every request.

snopoke:
Other issues described here: https://security.stackexchange.com/a/22936/141942:

> it costs you in terms of usability: with only one token valid at once, the user will not be able to navigate the webapp normally. For example if they hit the 'back' button and submit the form with new values, the submission will fail, and likely greet them with some hostile error message. If they try to open a resource in a second tab, they'll find the session randomly breaks in one or both tabs.

millerdev:
@snopoke on reviewing [the CSRF docs for Django 1.10](https://docs.djangoproject.com/en/1.10/ref/csrf/#how-it-works) as well as [the issue](https://code.djangoproject.com/ticket/20869) and [the commit that made the value sent to the client change on every request](https://github.com/django/django/commit/5112e65ef2df1dbb95ff83026b6a962fb2688661), it looks like the server value of the CSRF token is constant for the duration of the session, but the value sent to the client is scrambled with a random salt to protect it from BREACH attacks. This means it is protected against BREACH _and_ it has none of the usability issues of changing the token on every request. My conclusion from this is that we should upgrade to Django 1.10 as soon as possible since it has a superior CSRF protection implementation than older versions of Django.

PR #14078, opened 2016-11-30
fix get_case_sharing_groups in restore

benrudolph:
@esoergel @proteusvacuum i _think_ this was just broken for a couple months since it seems  like this code hasn't changed in a while, but it's recently been bugging out on me locally. i think there's also an interrupt ticket for it: http://manage.dimagi.com/default.asp?242287

cc: @emord

### DIFF
--- a/corehq/apps/users/models.py
+++ b/corehq/apps/users/models.py
@@ -1557,7 +1557,7 @@ class CommCareUser(CouchUser, SingleMembershipMixin, CommCareMobileContactMixin)
         # get faked location group objects
         groups = []
         for sql_location in self.sql_locations:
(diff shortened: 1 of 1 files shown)


emord (approved):

corehq/apps/users/models.py
```diff
@@ -1557,7 +1557,7 @@ def get_case_sharing_groups(self):
         # get faked location group objects
         groups = []
         for sql_location in self.sql_locations:
-            groups.extend(self.sql_location.get_case_sharing_groups(self._id))
+            groups.extend(sql_location.get_case_sharing_groups(self._id))
```
esoergel: Yikes, thanks for the fix!

PR #14079, opened 2016-11-30
Test user as and begin conforming to a standard

benrudolph:
@czue and fyi @wpride this tests all the different types of restore protocols we support `username@domain` and `username@domain.commcarehq.org`. i want to phase out `username@domain` cause it's complicating code that could be way simpler

PRs for touchforms and formplayer here:
https://github.com/dimagi/formplayer/pull/204
https://github.com/dimagi/touchforms/pull/279 - was able to submit and restore with this change, will let me know if you forsee an issue with this

cc: @emord

### DIFF
--- a/corehq/apps/ota/tests/test_utils.py
+++ b/corehq/apps/ota/tests/test_utils.py
@@ -24,7 +24,6 @@ class RestorePermissionsTest(LocationHierarchyTestCase):
 
     @classmethod
     def setUpClass(cls):
-        delete_all_users()
         super(RestorePermissionsTest, cls).setUpClass()
 
         cls.other_project = Domain(name=cls.other_domain)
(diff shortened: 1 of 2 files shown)


corehq/apps/ota/tests/test_utils.py
```diff
@@ -24,6 +24,7 @@ class RestorePermissionsTest(LocationHierarchyTestCase):
 
     @classmethod
     def setUpClass(cls):
+        delete_all_users()
```
emord: Assuming this was just for when you were testing locally? Looks like this is already in teardown

corehq/apps/ota/tests/test_utils.py
```diff
@@ -232,6 +323,7 @@ class GetRestoreUserTest(TestCase):
...
```
emord: same here

benrudolph: ah yes, it's nice to have for reuse db but seems frivolous

benrudolph: it won't if it errors in the setup call, which has randomly happened a lot

emord (approved):

czue (approved):
looks good, thanks!

PR #14083, opened 2016-11-30
Actually bulk delete locations.

esoergel:
@noahcarnahan @czue
Looks like this hasn't been deleting this locations properly.  I looked at the couchdbkit code for bulk deletion, and it passes through the list passed to it multiple times. Because `couch_locations` uses the output of `iter_docs`, it is a generator, and after the first pass, the generator is exhausted, so the deletion is never committed.

### DIFF
--- a/corehq/apps/locations/models.py
+++ b/corehq/apps/locations/models.py
@@ -526,7 +526,7 @@ class SQLLocation(SyncSQLToCouchMixin, MPTTModel):
         Delete a location and its dependants.
         This also unassigns users assigned to the location.
         """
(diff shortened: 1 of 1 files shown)


benrudolph:
ouch

czue (approved):
man it'd be nice if generators failed hard upon subsequent iterations.

build fail doesn't look related, i'm looking into it.

NoahCarnahan:
I've just created a "SafeGenerator" which should help detect these sorts of issues in the future: https://github.com/dimagi/dimagi-utils/pull/386

PR #14085, opened 2016-12-01
Opening for tests, please ignore

NoahCarnahan:
Not sure if there is a better workflow for this...
I want to test a change to a submodule, but I'm worried that if I open a PR in the submodule repo and then have it merged to master, then someone will assume that submodule should be updated in this repo, but really it still needs this repo's tests to run.

I guess the real issue is that I should get the full build tests running locally.


snopoke:
@NoahCarnahan can this be closed?

PR #14086, opened 2016-12-01
Fixed bugs in case finding report

kkrampa:
@czue

### DIFF
--- a/custom/enikshay/reports/filters.py
+++ b/custom/enikshay/reports/filters.py
@@ -25,12 +25,26 @@ class EnikshayLocationFilter(BaseMultipleOptionFilter):
 
     @property
     def selected(self):
-        selected = super(EnikshayLocationFilter, self).selected
+        location_ids = self.request.GET.getlist(self.slug)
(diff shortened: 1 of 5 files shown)


czue (approved):
code looks fine, but can you explain how this works (maybe with some comments in the code)? I think you added a hack to return all descendant location IDs in the `value` which makes sense, but i don't totally understand what's going on in `selected`. also, would be nice if this could be generalized more, but we can think about that separately

custom/enikshay/reports/filters.py
```diff
@@ -25,12 +25,29 @@ def options(self):
 
     @property
     def selected(self):
```
kkrampa: When a user selects a non-leaf location and he/she hits apply button descendants of selected location will also appear in a select box. That's why I overrode that method to not return descendants. Does it make sense to you?

czue: yep, it does, thanks

kkrampa:
@czue :white_check_mark: 
Can we move this forward?

czue:
sorry was just waiting on build.

PR #14087, opened 2016-12-01
add device id to user model when found during key request

czue:
save device id to user when provided by the phone in a keys request

related to http://manage.dimagi.com/default.asp?242011

@esoergel @NoahCarnahan


snopoke (approved):

proteusvacuum:
This was causing a bunch of `ERROR: BadValueError: u'2016-07-24T06:43:46.440887Z' is not a datetime object` errors

proteusvacuum:
hopefully that error was triggered for every User that ran this, otherwise we may have gotten some User models in a bad state. 
Would be good to add some tests for this change.

PR #14088, opened 2016-12-01
Drop MODULE_FILTER preview

kaapstorm:
Context: http://manage.dimagi.com/default.asp?242690

Hold off pending confirmation from @krsdimagi that this should show up for all users always.

@sravfeyn cc @biyeun


krsdimagi:
@kaapstorm yes, this is for all users.  Thanks!

czue (approved):
lgtm. not merging due to label though i think that it can just be removed.

orangejenny:
this has been a long time coming, hooray!

PR #14089, opened 2016-12-01
generic domain migration flags

snopoke:
Realized that the Couch -> SQL migration needs it's own flag since some Couch domains were created after the timezone work was done and hence have 'already been migrated'

This reuses the code from the timezone migration status model and makes it generic by adding a slug.

All existing data is migrated to the new model.
@nickpell @millerdev


corehq/apps/domain_migration_flags/api.py
```diff
@@ -0,0 +1,66 @@
...
+@skippable_quickcache(['domain'], skip_arg='strict')
+def any_migrations_in_progress(domain, strict=False):
+    return DomainMigrationProgress.objects.filter(
+        domain=domain, migration_status=MigrationStatus.IN_PROGRESS
+    ).exists()
+
+
+def clear_caches(domain, slug):
+    any_migrations_in_progress(domain, strict=True)
+    get_migration_status(domain, slug, strict=True)
```
millerdev: Looks like [this re-primes the cache](https://github.com/dimagi/commcare-hq/blob/b1ffe8295c9ce93c6c988f4bda72dfe41e8f38b3/corehq/util/quickcache.py#L205); it does not clear it (the name `clear_caches` is misleading).

snopoke: renamed

millerdev (approved):

snopoke:
will look at build failure tomorrow

PR #14090, opened 2016-12-01
Fix ICDS SQL functions

kaapstorm:
Context: http://manage.dimagi.com/default.asp?243102

@sheelio @sravfeyn cc @biyeun

### DIFF
--- a/custom/icds_reports/migrations/0002_fix_functions.py
+++ b/custom/icds_reports/migrations/0002_fix_functions.py
@@ -1,8 +1,8 @@
 # -*- coding: utf-8 -*-
 from __future__ import unicode_literals
-
 from django.db import migrations
 from corehq.sql_db.operations import RawSQLMigration
+
 
 migrator = RawSQLMigration(('custom', 'icds_reports', 'migrations', 'sql_templates'))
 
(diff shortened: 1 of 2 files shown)


kaapstorm:
@czue @snopoke when I run `./manage.py migrate`, it says it runs this migration, but SQL logs show that none of the SQL actually gets executed. I then checked whether its dependent migration, [`0001_initial.py`][1], ever ran `create_tables_and_views.sql`, and it seems none of those tables or views were created either. 

(If it's useful to know, I've been testing by deleting their corresponding entries in the `django_migrations` table to get Django to migrate them again. I've been checking whether the scripts succeeded by watching the SQL logs and looking for the tables in PostgreSQL with `\d` and for the functions with `select proname from pg_proc;`)

I can't see anything wrong with these migrate scripts. Can you spot what's wrong?

cc @sheelio 

  [1]: https://github.com/dimagi/commcare-hq/blob/36ae2a99d873ec3200c66b82f993dc62652febb9/custom/icds_reports/migrations/0001_initial.py#L16-L16

sheelio:
@kaapstorm 

Which database are you checking?  These functions are on the icds-ucr database.

snopoke:
The django DB routing is setup so that it will only run these migrations on the UCR database: https://github.com/dimagi/commcare-hq/blob/a0cca91a681709fadd06e8b796d32a428285ed3e/corehq/sql_db/routers.py#L30-L30

So to run them you have to have `ICDS_UCR_DATABASE_ALIAS` set in your settings file and the DB must exist. Then you can run:

```
$ ./manage.py migrate --database <UCR_DB_ALIAS>
```

If the UCR DB alias is `default` then you don't need to specify the `--database` param.

kaapstorm:
Thanks @sheelio @snopoke for the explanations.

PR #14095, opened 2016-12-01
Allow restore as self for web user

benrudolph:
Need to be able to restore as yourself for web users when they preview a form and aren't on new cloudcare

cc: @emord


emord (approved):

czue:
man you'd think this check would be easier.

PR #14100, opened 2016-12-02
Add device ID to user model

kaapstorm:
Fixes PR #14087

@czue @esoergel @NoahCarnahan cc @proteusvacuum @sravfeyn @biyeun

### DIFF
--- a/corehq/apps/mobile_auth/views.py
+++ b/corehq/apps/mobile_auth/views.py
@@ -66,7 +66,7 @@ def fetch_key_records(request, domain):
     payload = FetchKeyRecords(domain, user_id, last_issued).get_payload()
     device_id = request.GET.get('device_id')
     if device_id and isinstance(request.couch_user, CommCareUser):
-        request.couch_user.update_device_id_last_used()
+        request.couch_user.update_device_id_last_used(device_id)
         request.couch_user.save()
     return HttpResponse(payload)
 
--- a/corehq/apps/mobile_auth/views.py
+++ b/corehq/apps/mobile_auth/views.py
@@ -64,7 +64,16 @@ def fetch_key_records(request, domain):
         last_issued = string_to_datetime(last_issued).replace(tzinfo=None)
     user_id = request.couch_user.user_id
     payload = FetchKeyRecords(domain, user_id, last_issued).get_payload()
+    device_id = request.GET.get('device_id')
+    if device_id:
+        _add_device_id_to_user_if_necessary(request.couch_user, device_id)
     return HttpResponse(payload)
+
+
+def _add_device_id_to_user_if_necessary(couch_user, device_id):
+    if isinstance(couch_user, CommCareUser) and device_id and device_id not in couch_user.device_ids:
+        couch_user.device_ids.append(device_id)
+        couch_user.save()
 
 
 @login_or_digest_or_basic_or_apikey()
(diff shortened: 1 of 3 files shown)


corehq/apps/users/models.py
```diff
@@ -806,13 +814,28 @@ def wrap(cls, data, should_save=False):
...
+    @classmethod
+    def migrate_device_ids(cls, data):
+        if (
+            data.get('device_ids') and
+            isinstance(data['device_ids'][0], basestring)
+        ):
+            # device_ids property was only set once, on creation, with a single string, so we only need to worry
+            # about one item in the list
+            data['device_ids'] = [{
+                'device_id': data['device_ids'][0],
```
proteusvacuum: how come you aren't using `DeviceIdLastUsed` here? (alternatively, how come you are using it below if this one was making tests fail)

kaapstorm: What I didn't understand until I tested it is that the data that is passed to wrap is loaded from JSON. So the error when trying to instantiate a DeviceIdLastUsed was hiding the error I got when trying to pass that DeviceIdLastUsed instance to `super().wrap()` (in line 819).

The one below, in the `create()` method works because it's building a Python object, and `commcare_user.created_on` has already been set to a datetime instance by `CouchUser.create()`.

corehq/apps/mobile_auth/views.py
```diff
@@ -64,9 +64,27 @@ def fetch_key_records(request, domain):
         last_issued = string_to_datetime(last_issued).replace(tzinfo=None)
     user_id = request.couch_user.user_id
     payload = FetchKeyRecords(domain, user_id, last_issued).get_payload()
+    device_id = request.GET.get('device_id')
+    if device_id:
+        _touch_user_device_id_last_used(request.couch_user, device_id)
```
proteusvacuum: would be nice to have a test for this too. Looks like that method will only work if the migration has been run on that specific user. Just making sure that this will always be the case.

corehq/apps/users/models.py
proteusvacuum: I see. Thanks for the explanation, that makes sense.

corehq/apps/mobile_auth/views.py
czue: it should be since the migration is called automatically in the `wrap` function. though thinking about this a bit more I'm inclined to not deal with the migration and make a new property. will update this.

corehq/apps/mobile_auth/views.py
```diff
@@ -4,7 +4,7 @@
 from corehq.apps.domain.decorators import login_or_digest_or_basic_or_apikey, domain_admin_required
 from corehq.apps.mobile_auth.utils import new_key_record, get_mobile_auth_payload, bump_expiry
 from corehq.apps.mobile_auth.models import MobileAuthKeyRecord
-from corehq.apps.users.models import CommCareUser
+from corehq.apps.users.models import CommCareUser, DeviceIdLastUsed
```
dimagimon: F401 'DeviceIdLastUsed' imported but unused

corehq/apps/users/tests/test_user_model.py
```diff
@@ -1,7 +1,8 @@
-from django.test import TestCase
+from datetime import datetime, timedelta
+from django.test import TestCase, SimpleTestCase
 
 from corehq.apps.domain.shortcuts import create_domain
-from corehq.apps.users.models import CommCareUser
+from corehq.apps.users.models import CommCareUser, DeviceIdLastUsed
```
dimagimon: F401 'DeviceIdLastUsed' imported but unused

czue:
updated but don't merge until i have a chance to sanity test

czue:
ok this should be good to go. @benrudolph @proteusvacuum

benrudolph:
(note if someone merges this before me please use squash on merge)

benrudolph (approved):

benrudolph:
(killed second to last build)

PR #14103, opened 2016-12-02
Move more location stuff to SQL

esoergel:
(opening to run full test suite)
This is rebased on top of https://github.com/dimagi/commcare-hq/pull/13976


esoergel:
@gcapalbo would you mind taking a look at this?  I started this branch a while ago with a bulk of changes and have been cherry-picking stuff out to merge separately.  This is now a relatively small amount of changes that I'm considering merging in in one go.

corehq/apps/locations/models.py
```diff
@@ -432,6 +432,9 @@ def sync_couch_location(self):
     def save(self, *args, **kwargs):
         from corehq.apps.commtrack.models import sync_supply_point
         from .document_store import publish_location_saved
+        set_site_code_if_needed(self)
+        if self.location_id:
+            self.supply_point_id = sync_supply_point(self)
```
gcapalbo: is this call to sync_supply_point needed? we call it  a few lines below too

gcapalbo:
just the one comment otherwise looks good. it does seem like there are a lot of one for one replacements of couch location with sql location - is that always a safe thing to do, or do we have good test coverage on the affected areas?

esoergel:
> there are a lot of one for one replacements of couch location with sql location - is that always a safe thing to do, or do we have good test coverage on the affected areas?

That's generally a safe thing to do.  The main API differences have all been resolved, and so far everything has worked pretty cleanly.  Part of the reason I split this PR into smaller chunks was so that we could do an incremental rollout in case something does come up.  For each of these changes, I read where the location instance is used in the code and see if there's anything that might raise problems.  Most of the time it's just stuff like `location.name` and `location.external_id`, where we can be pretty confident that it doesn't matter whether it's a `SQLLocation` or a `Location`

corehq/apps/locations/models.py
esoergel: good catch - I think this is a holdover from a bad merge.  Updated.

gcapalbo (approved):
looks like there's merge conflicts

esoergel:
the conflict was trivial so I canceled the merge build.

PR #14111, opened 2016-12-05
Catch when the user's location has been deleted

kaapstorm:
Context: http://manage.dimagi.com/default.asp?243286

Opening for review. I'll add more commits to address the other items in the ticket, and tests.

@esoergel @mkangia cc @biyeun @sravfeyn

### DIFF
--- a/corehq/apps/users/bulkupload.py
+++ b/corehq/apps/users/bulkupload.py
@@ -708,6 +708,10 @@ def parse_users(group_memoizer, domain, user_data_model, location_cache):
             user_data_model.get_model_and_uncategorized(user.user_data)
         )
         role = user.get_role(domain)
+        try:
+            location_code = location_cache.get(user.location_id)
+        except SQLLocation.DoesNotExist:
+            location_code = None
         return {
             'data': model_data,
             'uncategorized_data': uncategorized_data,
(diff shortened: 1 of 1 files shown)


esoergel:
Thanks for the follow-up, Norman.  I'd also be happy to merge these fixes in piecemeal if you don't think they all need to come in together.

kaapstorm:
Good call @esoergel. They don't need to come in together, and piecemeal is a much better idea. Dropping "hold off".

mkangia:
@kaapstorm To understand this better, we are ignoring a case where a location that does not exist is assigned in bulk upload by assigning no location instead?
If thats the case i see this can be misleading since a user would end up having no location even when expected to have some location after a successful import. Would be better to skip the import and notify user via email if this is done async.

kaapstorm:
Hey @mkangia , this change only applies to downloading mobile workers. e.g. A user wants to download a list of mobile workers but at least one of them is currently assigned to a location that no longer exists. As the codebase stands at the moment, HQ will 500 when it tries to get that location. This change will set the mobile worker's location to None in the downloaded document, but won't modify the mobile worker's record.

mkangia:
@kaapstorm i thought this was upload since the file name was 'users/bulkupload`. Sorry!

PR #14112, opened 2016-12-05
[GS Efficiency] seed form with blank xml when creating new form

snopoke:
@millerdev @nickpell 

depends on https://github.com/dimagi/Vellum/pull/823

### DIFF
--- a/corehq/apps/app_manager/templates/app_manager/blank_form.xml
+++ b/corehq/apps/app_manager/templates/app_manager/blank_form.xml
@@ -8,7 +8,7 @@
 				</data>
 			</instance>
 			<itext>
-				<translation lang="en" default="" />
+				<translation lang="{{ lang }}" default="" />
 			</itext>
 		</model>
 	</h:head>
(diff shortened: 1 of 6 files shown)


czue:
yes!

snopoke:
On staging.

corehq/apps/app_manager/templates/app_manager/blank_form.xml
```diff
@@ -0,0 +1,17 @@
...
+	<h:head>
+		<h:title>{{ name }}</h:title>
+		<model>
+			<instance>
+				<data xmlns:jrm="http://dev.commcarehq.org/jr/xforms" xmlns="http://openrosa.org/formdesigner/{{ xmlns }}" uiVersion="1" version="1" name="{{ name }}">
+				</data>
+			</instance>
+			<itext>
+				<translation lang="{{ lang }}" default="" />
+			</itext>
```
millerdev: I think this `<itext>` element could be omitted. Looks like vellum sets the default language based on the list of languages passed to it on initialization, so this doesn't seem to be providing any benefit.

millerdev (approved):
Does this mean HQ can throw an error if the form source attachment is missing?

snopoke:
@millerdev No, HQ won't throw an error if the form source is missing but it will prevent you from creating a build or viewing the case management UI.

corehq/apps/app_manager/templates/app_manager/blank_form.xml
snopoke: 4289d45

benrudolph:
wow this has been a long time coming! nice

snopoke:
@sheelio this is on staging if you want to try it out

millerdev:
> No, HQ won't throw an error if the form source is missing...

Sorry, I didn't phrase my question very clearly. In the past a new form (with empty source) and a form missing it's source attachment (due to error like Riak not having that key) resulted in the same outcome in HQ: it served up an empty form. The error was silently ignored. What I meant to ask was now that the form source is set on form creation it seems that HQ could now throw an error if the source attachment is missing.

snopoke:
@millerdev I've updated the code so that it will error if the attachment is missing. Not sure why we couldn't do this before: fadd2f2

sheelio:
@snopoke 

Staging seems to be down, but I'm guessing this can get released without my review.

millerdev (approved):

PR #14114, opened 2016-12-05
Add a "User IMEIs" column to mobile workers download

kaapstorm:
Adds a comma-separated list of IMEI numbers in the mobile workers download.

e.g.
user_imeis_column

@sheelio @czue cc @biyeun @sravfeyn


czue (approved):
looks good to me. might want @sheelio to confirm whether it will be a problem if this starts showing up for all domains

sheelio:
@czue @kaapstorm 

Shouldn't be an issue as long as the column is not required and uploading it doesn't affect the existing IMEI data

czue:
fyi, conflicts now. also note for reviewer - we may want to use a squash merge for this since some chance we need to deploy the patch to NIC

kaapstorm:
Resolved

PR #14116, opened 2016-12-05
Use Nimbus API for form validation

snopoke:
@nickpell @millerdev 

cc @benrudolph @wspride


corehq/apps/app_manager/xform.py
```diff
@@ -543,7 +545,11 @@ def validate_xform(source):
         source = source.encode("utf-8")
     # normalize and strip comments
     source = ET.tostring(parse_xml(source))
-    validation_results = formtranslate.api.validate(source)
+    try:
+        validation_results = nimbus_api.validate_form(source)
+    except NimbusAPIException:
+        return
```
snopoke: In the case of an error I thought the best thing to do is just ignore it since there is nothing the user can do about it. This does mean that could create a build with an invalid form. The alternative is to give a message saying that we couldn't validate their form but this also prevents them from doing any case management and some other things.

millerdev: Is it possible to give a strictly informational warning message (that does not prevent those other actions)? It seems bad to have "could not validate" look the same as "your form is valid."

.gitmodules
```diff
@@ -10,9 +10,6 @@
 [submodule "submodules/commcare-translations"]
 	path = submodules/commcare-translations
 	url = git://github.com/dimagi/commcare-translations.git
-[submodule "submodules/formtranslate-src"]
-	path = submodules/formtranslate-src
-	url = git://github.com/dimagi/formtranslate.git
```
benrudolph: 😂

corehq/apps/nimbus_api/tests/test_form_validation.py
```diff
@@ -0,0 +1,50 @@
...
+from django.test import SimpleTestCase
+
+from django.conf import settings
+
+from corehq.apps.nimbus_api import const
+from corehq.apps.nimbus_api.exceptions import NimbusRequestException
+from corehq.apps.nimbus_api.form_validation import validate_form
+
+
+@requests_mock.Mocker()
```
benrudolph: cool

benrudolph (approved):
only thing i might suggest is that we roll this out behind a toggle to make sure that formplayer doesn't fall over

snopoke:
updated to show errors when unable to do validation (but still allow you to continue app building - but not create new builds)

snopoke:
Realized that tests will fail since they don't have access to Nimbus. @millerdev options for this are another docker image or just pull the latest build off travis and run it before running the tests. Any opinions on which is better?

I kinda feel that docker is the better long term option since it enables us to use it in dev and elsewhere as well.

millerdev:
Docker sees like the way to go. That's even what [Yelp recommends](https://engineeringblog.yelp.com/2015/03/using-services-to-break-down-monoliths.html#testing) (AotW).

snopoke:
That article actually recommends mocking out the service (which is what I've already done) unless you're doing integration testing.

millerdev:
@snopoke sorry, your question about docker vs pull latest build gave me the impression that mocking wasn't an option. The part of the article I was referring to was "In response to this issue we’re now using Docker containers to spin up private test instances of services."

A mock is fine if you don't actually need to test the service functionality and don't expect the API of the thing you're testing to change or you have some way of validating that the interface hasn't changed.

snopoke:
@millerdev I think when I worded the question it was in the context of the code which was already using mocking. I do think it would be useful to have a Nimbus docker image around though.

:white_check_mark:

benrudolph (approved):
👍 from me will let @millerdev take a look

benrudolph:
thinking about this again and it might make sense to rename this to formplayer_api. we already have way too many names for formplayer/cloudcare/web apps/touchforms/nimbus. nimbus is not yet used anywhere in the code base

snopoke:
Fair point. On a similar topic I did specifically put this separate from
touchforms since I expect that the whole touchforms repo will be going at
some point in the not too distant future.

PR #14118, opened 2016-12-05
revert back to installing the whole app

benrudolph:
@wpride this'll revert us back to installing the whole app until changes to update app are made

cc: @emord


benrudolph:
i left the same post params because i'm expecting we'll change this back at some point, do you think i should just remove those until we're ready to switch back to update?

wpride:
@benrudolph I think keep them in, hopefully will get back on the update train soon

wpride (approved):

benrudolph:
sounds good

PR #13362, opened 2016-09-26
Location export

mkangia:
Also merge submodule https://github.com/dimagi/commcarehq-prelogin/pull/106 along

PR #14119, opened 2016-12-05
Return location queryset even if it's None.

esoergel:
@benrudolph @mkangia
Followup from https://github.com/dimagi/commcare-hq/pull/13362/files/5248880bdb3040f45c8f8e2c27cd146ae01eeddd#r90933211


benrudolph (approved):
👍 thanks @proteusvacuum just wanted to flag that this should get in before deploy

mkangia:
Thank You @esoergel

PR #14120, opened 2016-12-05
Small app preview changes

benrudolph:
@wpride @emord adds some minor bug fixes to app preview styling and adds anlytics


corehq/apps/app_manager/static/app_manager/js/preview_app.js
```diff
@@ -57,13 +57,15 @@ hqDefine('app_manager/js/preview_app.js', function() {
         $appPreview.addClass('preview-tablet-mode');
         $(module.SELECTORS.OFFSET_FOR_PREVIEW).addClass('offset-for-tablet');
         _private.triggerPreviewEvent('tablet-view');
+        window.analytics.workflow('[app-preview] User turned on tablet mode');
```
emord: Do you need an `if (window.analytics)` here?

benrudolph: hmm i don't think so, `window.analytics` is included in the base template: https://github.com/dimagi/commcare-hq/blob/master/corehq/apps/style/templates/style/base.html#L194. is there a situation where it doesn't get loaded?

PR #14123, opened 2016-12-05
don't fail hard on not having cloudcare_enabled

benrudolph:
just for you @wpride so those remote apps don't give you trouble cc: @emord


wpride:
@benrudolph already deleted my poor remote app :(

PR #14125, opened 2016-12-06
Resolve resource conflict error for add_inferred_export_properties

benrudolph:
@emord pesky little bugger. hoping this gets rid of the error emails. my hunch is people are bulk uploading a bunch of users and a lot of tasks get queued up at once. still not quite sure why serial task isn't working. i did check to ensure that the proper redis keys were being generated on task creation. in any case this should at least decrease the load on the server


emord:
Seems better. Do you think it would be worth it to add a soft assert for ones that error out that shows the parameters?

benrudolph:
it already errors out and shows the parameters, and they're all the same parameters which was why i was thinking of caching. is that what you meant?

emord:
Ah. yeah. didn't realize it was already emailing params

PR #14128, opened 2016-12-06
Revert "Revert "[GS Efficiency] check for missing instance declarations in xforms""

snopoke:
Reverts dimagi/commcare-hq#14102

### DIFF
--- a/corehq/apps/app_manager/xform.py
+++ b/corehq/apps/app_manager/xform.py
@@ -1274,8 +1274,12 @@ class XForm(WrappedNode):
         If the id already exists, DOES NOT overwrite.
 
         """
-        conflicting = self.model_node.find('{f}instance[@id="%s"]' % id)
-        if not conflicting.exists():
+        instance_xpath = 'instance[@id="%s"]' % id
+        conflicting = (
+            self.model_node.find('{f}%s' % instance_xpath).exists() or
+            self.model_node.find(instance_xpath).exists()
+        )
+        if not conflicting:
             # insert right after the main <instance> block
             first_instance = self.model_node.find('{f}instance')
             first_instance.addnext(_make_elem('instance', {'id': id, 'src': src}))
--- a/corehq/apps/app_manager/xform.py
+++ b/corehq/apps/app_manager/xform.py
@@ -2133,7 +2133,7 @@ def infer_vellum_type(control, bind):
 
 def find_missing_instances(form_source):
     from corehq.apps.app_manager.suite_xml.post_process.instances import get_all_instances_referenced_in_xpaths
-    instance_declaration_re = r"""<instance.*id=\"([\w\-:]+)\""""
+    instance_declaration_re = r"""<instance[^>]*id=\"([\w\-:]+)\""""
     instance_declarations = set(re.findall(instance_declaration_re, form_source))
 
     missing_instances = set()
--- a/corehq/apps/app_manager/xform.py
+++ b/corehq/apps/app_manager/xform.py
@@ -2138,12 +2138,12 @@ def infer_vellum_type(control, bind):
     return result['name']
 
 
-def find_missing_instances(form_source):
+def find_missing_instances(wrapped_xform):
     from corehq.apps.app_manager.suite_xml.post_process.instances import get_all_instances_referenced_in_xpaths
-    instance_declarations = XForm(form_source).get_instance_ids()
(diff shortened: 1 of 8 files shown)


snopoke:
@nickpell @millerdev this is basically the same code except that it does the validation after the XForm has been rendered

corehq/apps/app_manager/xform.py
```diff
@@ -2133,7 +2133,7 @@ def infer_vellum_type(control, bind):
 
 def find_missing_instances(form_source):
     from corehq.apps.app_manager.suite_xml.post_process.instances import get_all_instances_referenced_in_xpaths
-    instance_declaration_re = r"""<instance.*id=\"([\w\-:]+)\""""
+    instance_declaration_re = r"""<instance[^>]*id=\"([\w\-:]+)\""""
```
millerdev: I think this will miss an instance formatted like `<instance id='casedb' src='...'>` (single quotes). Is that a concern here?

Also, less important, this will match tags like `<instances id="xyz"` and `<instanceid="xyz"`.

A better regexp might be `<instance\s[^>]*id=(['"])([\w\-:]+)\1` – note that this one has different group indices, so you may need to account for that elsewhere in the code. Tested on the following:
```xml
<instance src="..." id="abc">
<instance src="..." id='abc'>
<instanceid="abc">
<instances
  src="..." id="abc">
<instance
  src="..." id="abc">
```

millerdev: Also, using a regexp on the entire xform seems less than ideal (although maybe not a huge concern). This will pick up things that look like instance definitions in parts of the xml where instance definitions should not be found. Is the XML parsed prior to rendering? If yes, wouldn't it be better to parse instances using a real XML parser that knows about things like document structure?

snopoke: I was trying to avoid parsing the form AGAIN but I guess xml parsing should be fast. Updated.

snopoke:
:white_check_mark:

corehq/apps/app_manager/xform.py
```diff
@@ -2129,3 +2136,36 @@ def infer_vellum_type(control, bind):
...
+    if missing_instances:
+        instance_ids = "','".join(missing_instances)
+        message_parts.append(_("Known instances: '{}'").format(instance_ids))
+    if missing_unknown_instances:
+        instance_ids = "','".join(missing_unknown_instances)
+        message_parts.append(_("Unknown instances: '{}'").format(instance_ids))
+
+    if message_parts:
+        raise XFormValidationError(
+            'The form is missing some instance declarations: h{}'.format(', '.join(message_parts)),
```
millerdev: typo: `h` before `{}`

corehq/apps/app_manager/xform.py
```diff
@@ -646,9 +648,15 @@ def all_media_references(self, lang):
         return images + video + audio + inline_video
 
     def get_instance_ids(self):
+        def _get_instances():
+            return itertools.chain(
+                self.model_node.findall('{f}instance'),
+                self.model_node.findall('instance')
```
snopoke: not sure why this is necessary but it definitely is - could be a bug with the parser?

snopoke:
:white_check_mark:

corehq/apps/app_manager/xform.py
```diff
@@ -2129,3 +2148,36 @@ def infer_vellum_type(control, bind):
...
         return None
     return result['name']
+
+
+def find_missing_instances(wrapped_xform):
+    from corehq.apps.app_manager.suite_xml.post_process.instances import get_all_instances_referenced_in_xpaths
+    instance_declarations = wrapped_xform.get_instance_ids()
+    missing_instances = set()
+    missing_unknown_instance = set()
+    instances, unknown_instance_ids = get_all_instances_referenced_in_xpaths('', [wrapped_xform.render()])
```
millerdev: Assuming rendering is relatively low cost, but wanted to point out that it could be avoided since [it's rendering the form anyway](https://github.com/dimagi/commcare-hq/blob/07d0faccb079371fae2eb8c645275f5149c82048/corehq/apps/app_manager/models.py#L995) it might make sense to add an `xform_xml` argument to `find_missing_instances` and `check_for_missing_instances`?

snopoke: Yea, I considered doing that but decided it was ok to render it again

PR #14131, opened 2016-12-06
properly handle editing/deprecating forms that had errors originally

czue:
http://manage.dimagi.com/default.asp?243382

@benrudolph @wpride this should fix Sarah's bug with edits

### DIFF
--- a/corehq/form_processor/backends/sql/processor.py
+++ b/corehq/form_processor/backends/sql/processor.py
@@ -4,6 +4,7 @@ import uuid
 
 from PIL import Image
 from django.db import transaction
+from casexml.apps.case.exceptions import CommCareCaseError
 from casexml.apps.case.xform import get_case_updates
 from corehq.form_processor.backends.sql.dbaccessors import (
     FormAccessorSQL, CaseAccessorSQL, LedgerAccessorSQL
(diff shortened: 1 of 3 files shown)


corehq/form_processor/backends/sql/processor.py
```diff
@@ -171,7 +172,12 @@ def get_cases_from_forms(case_db, xforms):
             for xform in xforms:
                 if xform.is_deprecated:
                     deprecated_form = xform
-                affected_cases.update(case_update.id for case_update in get_case_updates(xform))
+                if not (xform.is_deprecated and xform.problem):
+                    # don't process deprecatd forms which have errors.
+                    # see http://manage.dimagi.com/default.asp?243382 for context.
+                    # note that we have to use .problem instead of .is_error because applying
+                    # the state=DEPRECATED overrides state=ERROR
```
czue: @snopoke I'm not sure if you have any more clever ideas to check if a form being deprecated was previously in an error state. I didn't dig too deeply into sql form processing to try and figure that out.

corehq/form_processor/backends/sql/processor.py
```diff
@@ -4,6 +4,7 @@
 
 from PIL import Image
 from django.db import transaction
+from casexml.apps.case.exceptions import CommCareCaseError
```
dimagimon: F401 'CommCareCaseError' imported but unused

wpride:
@czue failure looks related

 test_case_management (couchforms.tests.test_edits.EditFormTest) ... FAIL

wpride:
:hero:

corehq/form_processor/backends/sql/processor.py
snopoke: We could take the same approach we do with deletions and make the state additive. Not sure what all the implications of that are though.

czue: ok, i think I might punt that for now. actually had to fall back to `.problem` on the couch side too because `is_error` just checks the `doc_type != 'XFormInstance'` which is also too broad

czue:
:white_check_mark:

benrudolph:
:shipit:

PR #14132, opened 2016-12-06
Empty displayText is valid if there is no Display Text column

kaapstorm:
Context: http://manage.dimagi.com/default.asp?240971

default_filters

Items in the report builder's Default Filters list were considered invalid if their displayText property was empty, but the Default Filters list doesn't have a Display Text column. This change accommodates that.

@NoahCarnahan @sravfeyn cc @biyeun


NoahCarnahan:
Code looks good, thanks Norman. Test failure is real though.

PR #14134, opened 2016-12-06
Case import history prefactor

dannyroberts:
isolate `case_importer` `soil` dependency in `CaseUpload` object

### DIFF
--- a/corehq/apps/case_importer/tracking/case_upload_tracker.py
+++ b/corehq/apps/case_importer/tracking/case_upload_tracker.py
@@ -1,0 +1,49 @@
+from corehq.apps.case_importer.util import open_spreadsheet_download_ref, get_spreadsheet
+from dimagi.utils.decorators.memoized import memoized
+from soil import DownloadBase
+from soil.progress import get_task_status
+from soil.util import expose_cached_download
+
+
+class CaseUpload(object):
+    _expiry = 1*60*60
+
+    def __init__(self, upload_id):
+        self.upload_id = upload_id
+
+    @classmethod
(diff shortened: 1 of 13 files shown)


corehq/apps/case_importer/tracking/case_upload_tracker.py
```diff
@@ -0,0 +1,49 @@
+from corehq.apps.case_importer.util import open_spreadsheet_download_ref, get_spreadsheet
+from dimagi.utils.decorators.memoized import memoized
+from soil import DownloadBase
+from soil.progress import get_task_status
+from soil.util import expose_cached_download
+
+
+class CaseUpload(object):
+    _expiry = 1*60*60
```
dimagimon: E226 missing whitespace around arithmetic operator

dannyroberts:
@czue

czue (approved):
Nice cleanup! 

Not a huge deal but in the future it'd be nice if you can break up the commits a bit more. Each one seemed like it was moving a number of things around or combining functionality in a way that could have been split out more which would have been a lot easier to read. As it was I had to figure out which piece of each commit tied to which piece within that same commit. Still not confident I fully followed everything but trusting your your process and QA.

czue:
please go ahead and merge once you confirm you feel you have QA'd sufficiently

dannyroberts:
Yeah, sorry for the commits; as I was doing it, the change I wanted to make kept begging for some other change somewhere else, and in the end they felt too tangled to break up in a sensical way, but point taken.

Tested that case upload, fixture upload, and case export all still work (which all use `soil`) so I'm happy with this.

PR #14135, opened 2016-12-06
case activity report

emord:
http://manage.dimagi.com/default.asp?241962

@benrudolph Not a great solution here. It uses ES in a way it's not really intended to be used in a couple of ways:

1) sorting by a field that isn't in the doc (username)
2) aggregations aren't really meant to return empty results, but in this report that is what is expected. It's possible to change that https://www.elastic.co/guide/en/elasticsearch/reference/1.7/search-aggregations-bucket-terms-aggregation.html#_minimum_document_count but that also means that filters don't apply to those empty buckets, so it would return users that aren't in the domain

:fish:

### DIFF
--- a/corehq/apps/reports/standard/monitoring.py
+++ b/corehq/apps/reports/standard/monitoring.py
@@ -372,8 +372,11 @@ class CaseActivityReport(WorkerMonitoringCaseReportTableBase):
         def _fill_out_buckets(buckets, rows):
             returned_user_ids = {b.key for b in buckets}
             not_returned_user_ids = set(self.paginated_user_ids) - returned_user_ids
+            extra_rows = []
             for user_id in not_returned_user_ids:
-                rows.append(self.Row(self, self.users_by_id[user_id], {}))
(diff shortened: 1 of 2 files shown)


corehq/apps/reports/standard/monitoring.py
```diff
@@ -468,7 +468,7 @@ def es_queryset(self, user_ids, size=None):
         top_level_aggregation = self.add_landmark_aggregations(top_level_aggregation, self.end_date)
 
         if size:
-            top_level_aggregation.size(size)
+            top_level_aggregation = top_level_aggregation.size(size)
```
benrudolph: hehe

corehq/apps/reports/standard/monitoring.py
```diff
@@ -369,6 +369,13 @@ def _format_row(self, row):
 
     @property
     def rows(self):
+        def _fill_out_buckets(buckets, rows):
```
benrudolph: nice

benrudolph (commented):
nice looks like a lot of bugs were quashed in this!

PR #14138, opened 2016-12-06
Fix select2 widths changing in firefox

orangejenny:
http://manage.dimagi.com/default.asp?243133

Fixed in select2 version 4. Added card to js team trello to upgrade the pages still using select2 version 3.

@proteusvacuum


proteusvacuum:
YES. This has bugged me for ages. Thanks!

proteusvacuum (approved):

PR #14139, opened 2016-12-06
Move sync settings to General section

orangejenny:
http://manage.dimagi.com/default.asp?243119

@kaapstorm / @sravfeyn


sravfeyn:
Thanks Jenny!

czue:
nice. have we looked for any doc updates that need to happen?

orangejenny:
@czue Good point. Just did a confluence search and found, curiously, all the references to these settings either specify "General Settings" or don't mention a section.

czue:
nice, thanks for checking!

PR #14140, opened 2016-12-07
have case importer check for blank row before blank external id

dannyroberts:
otherwise it will show a warning on blank rows when keying by external id

minor thing I stumbled upon that inappropriately shows this error on a blank row (if you delete the last row in a spreadsheet, that row shows up with blank cells instead of not showing up at all, so this could actually be quite annoying):
screen shot 2016-12-06 at 8 03 14 pm


dannyroberts:
@czue cc: @sheelio

czue:
:+1:

PR #14145, opened 2016-12-07
Changes in mobile reports

kkrampa:
@czue


czue (approved):
looks good - I just want to make sure I can update the app before I merge and deploy these so nothing breaks. will merge when i've done that.

PR #14149, opened 2016-12-07
Decrease async restore timeout

proteusvacuum:
Decrease async restore retry-after parameter to potentially reduce sync times for people who have this turned on.
FYI @amstone326 
@mkangia

### DIFF
--- a/corehq/ex-submodules/casexml/apps/phone/const.py
+++ b/corehq/ex-submodules/casexml/apps/phone/const.py
@@ -11,7 +11,7 @@ INITIAL_SYNC_CACHE_THRESHOLD = 60  # 1 minute
 INITIAL_ASYNC_TIMEOUT_THRESHOLD = 10
 # The Retry-After header parameter. Ask the phone to retry in this many seconds
 # to see if the task is done.
-ASYNC_RETRY_AFTER = 30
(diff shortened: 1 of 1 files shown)


amstone326:
@proteusvacuum How does changing the retry-after threshold potentially speed up sync times? Wouldn't it just speed up how quickly they receive a retry response?

benrudolph:
if you're sync breaks the inital threshold by 1 second, then you have to wait an additional 30 seconds or is just over the 30 seconds mark (took an additional 31 seconds to compute, but have to wait the rest of the 29 seconds)

proteusvacuum:
Yeah, what @benrudolph said. 

We've been seeing a bunch of restores that take ~15 seconds which means they take ~5 seconds longer than the initial wait period we've set to immediately return a response. The user then actually has to wait 40 seconds to get a response (10 seconds initial wait + 30 seconds "retry time"). 
This will all go away when we get around to automatically updating the `Retry-After` header with the restore, but doing that is a bit far down the line. 

PS, the "15 seconds" is totally anecdotal and is from me running a test with my test user and watching the celery task complete... so not scientific in any way.

amstone326:
Ah cool, thanks for the clarification

PR #14151, opened 2016-12-07
Add Value to valid filter values

emord:
http://manage.dimagi.com/default.asp?241474#1257847

@NoahCarnahan @kaapstorm 

This should probably be a constant, but I wasn't exactly sure how to structure it. Could be `.keys()` for https://github.com/dimagi/commcare-hq/blob/5295ddff4ddf03c5287a4235ed091d666516e233/corehq/apps/userreports/reports/builder/forms.py#L993 There's also a similar structure here https://github.com/dimagi/commcare-hq/blob/5295ddff4ddf03c5287a4235ed091d666516e233/corehq/apps/userreports/reports/builder/forms.py#L895 I'm not sure about the relationship between these structures so not sure the best way to name the constants or if these are the correct values

@benrudolph

### DIFF
--- a/corehq/apps/userreports/reports/builder/forms.py
+++ b/corehq/apps/userreports/reports/builder/forms.py
@@ -52,6 +52,14 @@ from corehq.apps.userreports.sql import get_column_name
 from corehq.apps.userreports.ui.fields import JsonField
 from dimagi.utils.decorators.memoized import memoized
 
+# This dict maps filter types from the report builder frontend to UCR filter types
+REPORT_BUILDER_FILTER_TYPE_MAP = {
+    'Choice': 'dynamic_choice_list',
+    'Date': 'date',
+    'Numeric': 'numeric',
+    'Value': 'pre',
+}
+
 
 class FilterField(JsonField):
     """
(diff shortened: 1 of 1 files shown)


benrudolph (approved):

NoahCarnahan:
Thanks Emord. Just pushed a commit that moves it to a constant.

emord:
Thanks @NoahCarnahan Should there be an equivalent `"pre": "Value"` https://github.com/dimagi/commcare-hq/blob/5295ddff4ddf03c5287a4235ed091d666516e233/corehq/apps/userreports/reports/builder/forms.py#L895-L901 ? That wasn't a part of the bug, but I wasn't clear if that should be the inverse of this dict

NoahCarnahan:
Looks like "pre" filters won't make it to the block that uses that map because of this conditional:
https://github.com/dimagi/commcare-hq/blob/5295ddff4ddf03c5287a4235ed091d666516e233/corehq/apps/userreports/reports/builder/forms.py#L884
So looks ok if that map doesn't need to include it. Also might be worth noting that the maps are slightly different unfortunately because its not a 1-to-1 relationship between the report builder types and the ucr types. Thanks for checking though!

emord:
D'oh. should have looked at more context. Thanks!

PR #14153, opened 2016-12-07
Remove Android / Cloudcare from Address format in Case Details

proteusvacuum:
I think this distinction doesn't matter any more.

http://manage.dimagi.com/default.asp?243558

@mkangia


orangejenny:
The distinction still exists - can't imagine we ever implemented addresses on Java phones - but yeah, Java phones are rare enough that I think this is fine to remove.

PR #14155, opened 2016-12-07
Case import history

dannyroberts:
Currently on staging:

screen shot 2016-12-07 at 2 32 27 pm

This displays below **Select a file to upload** section on case importer landing page currently, but could be convinced to place somewhere else. Currently supported features:

- Shows 10 most recent uploads (including uploads in progress)
- Shows progress of any active upload in the project space (including the one you just uploaded, or multiple uploads running at the same time)
- As long as any of the uploads shown are in progress, it will show you the progress live. (It currently stops polling once all displayed uploads are complete, so you'd have to refresh the page to see new ones.)
- Shows full success message and warnings that you'd see on the polling page
- Shows case type, name of the user who uploaded it, and time (in your chosen timezone)

Working with @sheelio on where to go with this next, but wanted to open this up for review and comments.

### DIFF
--- a/corehq/apps/case_importer/templates/case_importer/import_cases.html
+++ b/corehq/apps/case_importer/templates/case_importer/import_cases.html
@@ -102,13 +102,14 @@
         <!--ko if: state() == states.SUCCESS && case_uploads()-->
         <table class="table table-striped">
             <tr>
-                <th class="col-md-4">{% trans "Status" %}</th>
-                <th class="col-md-4">{% trans "Uploaded by" %}</th>
-                <th class="col-md-4">{% trans "Time" %}</th>
+                <th class="col-md-3">{% trans "Status" %}</th>
+                <th class="col-md-3">{% trans "Case Type" %}</th>
+                <th class="col-md-3">{% trans "Uploaded by" %}</th>
+                <th class="col-md-3">{% trans "Time" %}</th>
             </tr>
             <!--ko foreach: case_uploads-->
             <tr>
-                <td class="col-md-4">
+                <td class="col-md-3">
                     <!--ko if: task_status.state == $root.states.SUCCESS-->
                     <span class="label label-success">{% trans "Success" %}</span>
                     <!--/ko-->
@@ -129,10 +130,13 @@
                     </div>
                     <!--/ko-->
                 </td>
-                <td class="col-md-4">
+                <td class="col-md-3">
+                    <!--ko text: case_type--><!--/ko-->
+                </td>
+                <td class="col-md-3">
                     <!--ko text: user--><!--/ko-->
                 </td>
-                <td class="col-md-4" data-bind="text: created"></td>
+                <td class="col-md-3" data-bind="text: created"></td>
             </tr>
             <!--/ko-->
         </table>
--- a/corehq/apps/case_importer/templates/case_importer/import_cases.html
+++ b/corehq/apps/case_importer/templates/case_importer/import_cases.html
@@ -105,7 +105,7 @@
                 <th class="col-md-2">{% trans "Status" %}</th>
                 <th class="col-md-1">{% trans "Case Type" %}</th>
                 <th class="col-md-1">{% trans "Uploaded by" %}</th>
-                <th class="col-md-6">{% trans "Success Message" %}</th>
+                <th class="col-md-6">{% trans "Details" %}</th>
                 <th class="col-md-2">{% trans "Time" %}</th>
             </tr>
             <!--ko foreach: case_uploads-->
--- a/corehq/apps/case_importer/templates/case_importer/import_cases.html
+++ b/corehq/apps/case_importer/templates/case_importer/import_cases.html
@@ -102,16 +102,22 @@
         <!--ko if: state() == states.SUCCESS && case_uploads()-->
         <table class="table table-striped">
             <tr>
-                <th class="col-md-3">{% trans "Status" %}</th>
-                <th class="col-md-3">{% trans "Case Type" %}</th>
-                <th class="col-md-3">{% trans "Uploaded by" %}</th>
-                <th class="col-md-3">{% trans "Time" %}</th>
+                <th class="col-md-2">{% trans "Status" %}</th>
+                <th class="col-md-1">{% trans "Case Type" %}</th>
+                <th class="col-md-1">{% trans "Uploaded by" %}</th>
+                <th class="col-md-6">{% trans "Success Message" %}</th>
+                <th class="col-md-2">{% trans "Time" %}</th>
             </tr>
             <!--ko foreach: case_uploads-->
             <tr>
-                <td class="col-md-3">
+                <td class="col-md-2">
                     <!--ko if: task_status.state == $root.states.SUCCESS-->
-                    <span class="label label-success">{% trans "Success" %}</span>
+                        <!--ko if: _.isEmpty(task_status.result.errors)-->
+                        <span class="label label-success">{% trans "Success" %}</span>
+                        <!--/ko-->
+                        <!--ko if: !_.isEmpty(task_status.result.errors)-->
+                        <span class="label label-warning">{% trans "Success with warnings" %}</span>
+                        <!--/ko-->
                     <!--/ko-->
                     <!--ko if: task_status.state == $root.states.FAILED-->
                     <span class="label label-danger">{% trans "Failed" %}</span>
@@ -130,13 +136,16 @@
                     </div>
                     <!--/ko-->
                 </td>
-                <td class="col-md-3">
+                <td class="col-md-1">
                     <!--ko text: case_type--><!--/ko-->
                 </td>
-                <td class="col-md-3">
+                <td class="col-md-1">
                     <!--ko text: user--><!--/ko-->
                 </td>
-                <td class="col-md-3" data-bind="text: created"></td>
+                <td class="col-md-6" data-bind="with: task_status">
+                    {% include 'case_importer/partials/ko_import_status.html' %}
+                </td>
+                <td class="col-md-2" data-bind="text: created"></td>
             </tr>
             <!--/ko-->
         </table>
--- a/corehq/apps/case_importer/templates/case_importer/import_cases.html
+++ b/corehq/apps/case_importer/templates/case_importer/import_cases.html
@@ -99,7 +99,7 @@
 
     <h2>Recent Uploads</h2>
     <div id="recent-uploads" class="ko-template">
-        <!--ko if: state() == states.SUCCESS && case_uploads()-->
+        <!--ko if: state() == states.SUCCESS && !_.isEmpty(case_uploads())-->
         <table class="table table-striped">
             <tr>
                 <th class="col-md-2">{% trans "Status" %}</th>
@@ -150,7 +150,7 @@
             <!--/ko-->
         </table>
         <!--/ko-->
-        <!--ko if: state() === states.SUCCESS && !case_uploads()-->
+        <!--ko if: state() === states.SUCCESS && _.isEmpty(case_uploads())-->
         <div class="alert alert-info">
             You have no recent uploads.
         </div>
--- a/corehq/apps/case_importer/templates/case_importer/import_cases.html
+++ b/corehq/apps/case_importer/templates/case_importer/import_cases.html
@@ -100,15 +100,15 @@
     <h2>Recent Uploads</h2>
     <div id="recent-uploads" class="ko-template">
         <!--ko if: state() == states.SUCCESS && case_uploads()-->
-        <table class="table">
+        <table class="table table-striped">
             <tr>
-                <th>{% trans "Time" %}</th>
-                <th>{% trans "Status" %}</th>
+                <th class="col-md-4">{% trans "Status" %}</th>
+                <th class="col-md-4">{% trans "Uploaded by" %}</th>
+                <th class="col-md-4">{% trans "Time" %}</th>
             </tr>
             <!--ko foreach: case_uploads-->
             <tr>
-                <td data-bind="text: created"></td>
-                <td>
+                <td class="col-md-4">
                     <!--ko if: task_status.state == $root.states.SUCCESS-->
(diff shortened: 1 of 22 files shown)


dannyroberts:
@czue buddy

benrudolph:
oh wow this is sweet 😮

czue:
yeah this is awesome!

corehq/apps/users/dbaccessors/couch_users.py
```diff
@@ -14,3 +14,11 @@ def get_user_id_by_username(username):
     if row:
         return row["id"]
     return None
+
+
+def get_display_name_for_user_id(domain, user_id, default=None):
```
snopoke: caching?

corehq/apps/case_importer/tracking/models.py
```diff
@@ -0,0 +1,41 @@
...
+    couch_user_id = models.CharField(max_length=256)
+    case_type = models.CharField(max_length=256)
+
+    @property
+    @memoized
+    def task(self):
+        return get_task(self.task_id)
+
+
+class CaseUploadJSON(jsonobject.JsonObject):
```
snopoke: Interested to hear why you chose this approach to producing JSON instead of having a `to_json` method on the `CaseUploadRecord` that either just creates a dict or uses a serializer to create the dict.

snopoke:
Looks great! I think the main goal for this is to be able to run the same import multiple times without the need to re-configure it. In that case we'd need to store the import config somewhere. I guess we might also want to be able to permanently save a config for reuse (instead of just the last 10). 

The workflow that comes to mind is that after running an import you get the option to make it an import template (button on the history). Then when doing a new import you could select a template instead of going through all the config.

@sheelio can correct me if I'm off base.

corehq/apps/case_importer/tracking/models.py
dannyroberts: I had two objectives in writing this, though I'm not too tied to it and it's definitely more verbose than the alternatives.

1. have an explicitly-defined read-only model for `get_case_uploads` to return, rather than a queryset which lets the caller do all sorts of things to the underlying SQL tables.
2. have a well-defined schema for the JSON object to be passed on for use in javascript

(2) ended up not working out since I later realized I need to do some further transformations before passing it on for use in javascript, but 1 is still a pro. It's definitely adds some redundancy, so I may scrap it anyway for the simpler option, or keep some aspect of it but simplify it somehow.

corehq/apps/users/dbaccessors/couch_users.py
dannyroberts: Good eye. It's not obvious from this code, but `CouchUser.get_by_user_id` is already cached, I believe with enough sophistication that it's properly cleared when a user is saved, so caching this function is at best redundant and at worst would lead to the minor stale cache issues.

dannyroberts:
@snopoke thanks for the review. As to your last comment, that's one direction this feature could go in, but there are a number of other offshoots we're considering that are all enabled by having some sort of record of previous uploads. That includes

- getting access to the original file used for an upload which can help a lot with debugging, by removing uncertainty as to what was done in a particular upload
- maintaining links to the cases touched by an upload and/or the forms used to touch them
- which further opens the door for a feature to revert an upload (by reverting all the forms involved)

as well as just the ability to see what uploads are already in progress so you don't get confused and double-upload something, or two users don't step on each others' toes by doing similar uploads at the same time, etc.

Each of those things requires adding more columns to what's stored for each upload, as does storing the config, but they all start with this basic feature of having a record of past uploads, so I'm just starting with this and then working with @sheelio to prioritize the other add-ons.

The config thing is something we've considered, but it also does have some open questions, like how to make sure the new excel file you choose is compatible with the config, etc. that aren't problems when you're building the config off the excel file like is currently done. That made it enough of an uncertain feature that I'm pushing it down the road a bit and will probably work on some lower-hanging fruit first.

dannyroberts:
@snopoke I forgot to say re: last 10, though you may have picked this up. The case upload records are persisted (in postgresql) forever at this point, and I just made the front end arbitrarily ask for the last 10 and punted any design decisions related to the UI for getting more, etc. So yeah, for now just showing the last 10, but assuming that longer term we'll want to change that and that it won't be too hard to change.

dannyroberts:
@snopoke / any reviewer: I have a number of branches building on this one, and I'm interested in hearing your opinions about rollout/review plan. Some options

- QA and merge this branch, releasing this first-pass to our users, and then PR the other branches to be reviewed individually and merged and deployed as they can be
- combine those other branches into this one and deploy them as a single change-set, rather than giving users a few new features every day

I _think_ the learning curve for users for this change isn't bad at all, so I'd weight being deliberate about the changes we expose lower than I would for app builder, but still could be a consideration. I'm fine with either way, though the long-running branch may get more and more annoying for reviewers to keep track of.

dannyroberts:
OK, I tested case importer on staging, as well as fixture upload and user upload (to validate my touching soil internals didn't break anything). Open question above about rollout, but from a QA perspective this is g2g

corehq/apps/case_importer/templates/case_importer/import_cases.html
```diff
@@ -2,6 +2,53 @@
 {% load report_tags %}
 {% load i18n %}
 
+{% block js-inline %}
+<script id="caseUploadsResource" type="text/plain">{% url 'case_importer_uploads' domain %}</script>
```
snopoke: This is interesting, don't think I've ever seen this pattern before. Why is it better than just assigning the URL to a var directly?

corehq/apps/case_importer/templates/case_importer/import_cases.html
```diff
@@ -48,4 +95,71 @@
...
+                             role="progressbar" aria-valuenow="60" aria-valuemin="0" aria-valuemax="100"
+                             data-bind="attr: {style: 'width: ' + task_status.progress.percent + '%;'}">
+                            <span><!--ko text: task_status.progress.percent --><!--/ko-->%
+                                {% trans "Complete" %}</span>
+                          </div>
+                    </div>
+                    <!--/ko-->
+                </td>
+                <td class="col-md-1">
+                    <!--ko text: case_type--><!--/ko-->
```
snopoke: why use a comment binding there instead of a text binding on the `td` or on a `span`?

corehq/apps/case_importer/templates/case_importer/partials/ko_import_status.html
```diff
@@ -0,0 +1,64 @@
...
+                {% blocktrans with match_count='result.match_count' %}
+                    <strong data-bind="text: {{ match_count }}"></strong> rows were matched and
+                    updated.
+                {% endblocktrans %}
+            </p>
+        <!--/ko-->
+
+        <!--ko if: result.created_count > 0-->
+            <p>
+                {% blocktrans with created_count='result.created_count' %}
```
snopoke: If these are available in the template (`result.created_count`) why use knockout to determine whether or not to show them? Especially since they don't seem to get updated by any JS if the value changes.

corehq/apps/case_importer/templates/case_importer/partials/ko_import_status.html
```diff
@@ -0,0 +1,64 @@
...
+                    <strong data-bind="text: {{ created_count }}"></strong> rows did
```
snopoke: Also confused by this, why use Knockout data-bind with a static value instead of just putting the value in?

corehq/apps/case_importer/templates/case_importer/partials/ko_import_status.html
```diff
@@ -0,0 +1,64 @@
...
+                    in your system with the same external ID.
+                {% endblocktrans %}
+            </p>
+        <!--/ko-->
+    </div>
+    <!--ko if: !_.isEmpty(result.errors)-->
+    <div class="alert alert-warning">
+        <ul class="list-unstyled">
+        <!--ko foreach: _.pairs(result.errors).map(_.partial(_.object, ['error','columns']))-->
+            <!--ko foreach: _.pairs(columns).map(_.partial(_.object, ['column', 'value']))-->
```
snopoke: can this not be pushed into the JS rather than having in the ko comment? It would make it easier to test at a later stage if necessary.

corehq/apps/case_importer/tracking/models.py
```diff
@@ -0,0 +1,41 @@
+from django.db import models
+from dimagi.ext import jsonobject
+from dimagi.utils.decorators.memoized import memoized
+from soil.util import get_task
+
+
+class CaseUploadRecord(models.Model):
+    domain = models.CharField(max_length=256)
```
snopoke: should add an index to this.

corehq/apps/case_importer/tracking/views.py
```diff
@@ -0,0 +1,39 @@
...
+    try:
+        limit = int(request.GET.get('limit'))
+    except (TypeError, ValueError):
+        limit = 10
+
+    def to_user_json(case_upload):
+        tz = get_timezone_for_request()
+        task_status = get_task_status(get_task(case_upload.task_id))
+        return {
+            'created': ServerTime(case_upload.created).user_time(tz).ui_string(),
```
snopoke: nice

snopoke (commented):
Nothing blocking and fine with pushing this out now. Will let you merge in case any of my questions prompt changes.

czue:
looks like @snopoke is doing a better review than I would so will let him review/merge. let me know if anything needed though

corehq/apps/case_importer/templates/case_importer/import_cases.html
dannyroberts: Yeah good question. If I put `"{% url ''case_importer_uploads' domain %}"` inline, then it's escaping for an html context but putting in a string-in-javascript context, so there are a number of things that can go wrong, even though it's rare. Now that I think of it though, I believe `{% url ''case_importer_uploads' domain as url %}{{ url|JSON }}` is the simplest way to do this correctly, and I think that's a better convention, so I'll switch to that. Thanks for asking

corehq/apps/case_importer/templates/case_importer/import_cases.html
```diff
@@ -2,6 +2,53 @@
 {% load report_tags %}
 {% load i18n %}
 
+{% block js-inline %}
+<script id="caseUploadsResource" type="text/plain">{% url 'case_importer_uploads' domain %}</script>
+<script>
+(function () {
+    var caseUploadResource = $('#caseUploadsResource').text();
```
dannyroberts: Probably definition should be in a separate file and binding should be on the page. That's the closest thing we've come to to a convention for this, so you're right I ought to adhere to it.

corehq/apps/case_importer/templates/case_importer/import_cases.html
dannyroberts: Yeah no particular reason. Sounds like you think we should prefer data-bind on an element over comment bindings, it's certainly more common in our code. I can change this.

corehq/apps/case_importer/templates/case_importer/partials/ko_import_status.html
dannyroberts: Sorry this is confusing. This line sets the django template variable `created_count` to the string `'result.created_count'`, which is then used below. The point was just remove as much logic as possible from the thing inside blocktrans, but it's still pretty awkward. Not sure if there's a much better way to do this.

corehq/apps/case_importer/templates/case_importer/partials/ko_import_status.html
dannyroberts: Yeah the django template evaluates this to

```html
<strong data-bind="text: result.created_count"></strong>
```

so it's still a dynamic binding. The idea is that if I change the structure of `result` somehow and want to edit where we're getting that value from, I can change it outside of the blocktrans block, so that the string doesn't have to be translated again. Agree it's pretty awkward and open to better suggestions.

corehq/apps/case_importer/templates/case_importer/partials/ko_import_status.html
snopoke: And what's bad about just having this inside the blocktrans:
```
<strong data-bind="text: result.created_count"></strong>
```

corehq/apps/case_importer/templates/case_importer/partials/ko_import_status.html
dannyroberts: yeah that's a good point. I can do this.

corehq/apps/case_importer/templates/case_importer/partials/ko_import_status.html
snopoke: Yea, I see why you do that, seems OK I guess

corehq/apps/case_importer/tracking/models.py
dannyroberts: yeah maybe even on (domain, created)

dannyroberts:
@snopoke thanks for all the great comments. I think what I'll do is merge this now, PR the other changes I've done on top of this, and then incorporate all your feedback on a branch on top of that, just because that'll be a lot simpler for me than making the changes here and then merging it with the changes I've made on top of this.

dannyroberts:
@snopoke addressed your PR feedback in these commits: https://github.com/dimagi/commcare-hq/compare/case-import-history-phase2...case-import-history-phase-1-feedback, which are on top of this PR: https://github.com/dimagi/commcare-hq/pull/14178. You should be able to comment on the commits even though it's not in PR yet.

I added a little tooling around urls that's pretty generic and useful in other places; I was really not liking the redundancy of passing around urls, but am open to it being not the right fix.

snopoke:
@dannyroberts I think that's a pretty good approach but since you're proposing a new way of doing url's in JS I think it's worth sending to a wider audience (architecture review / JS team)

dannyroberts:
@snopoke cool, I just took it to slack#future-of-js-in-hq and it got a positive reception; I'll also ping them on the PR when I open it

PR #14156, opened 2016-12-07
Report an issue spec

benrudolph:
@emord this implements the rest of the bug report spec: http://manage.dimagi.com/default.asp?240263
screen shot 2016-12-07 at 2 53 07 pm
screen shot 2016-12-07 at 2 53 03 pm
Essentially adds one field to the domain and let's you update that field if you have the proper perms in the bug report

🐟 or 🏠

### DIFF
--- a/corehq/apps/hqwebapp/tests/test_views.py
+++ b/corehq/apps/hqwebapp/tests/test_views.py
@@ -20,26 +20,28 @@ class TestEmailAuthenticationFormAutocomplete(BaseAutocompleteTest):
 class TestBugReport(TestCase):
     domain = 'test-bug-report'
 
-    def setUp(self):
+    @classmethod
+    def setUpClass(cls):
         delete_all_users()
-        self.project = create_domain(self.domain)
-        self.web_user = WebUser.create(
-            self.domain,
+        cls.project = create_domain(cls.domain)
+        cls.web_user = WebUser.create(
+            cls.domain,
             'bug-dude',
             password='***',
         )
-        self.web_user.is_superuser = True
-        self.web_user.save()
-        self.commcare_user = CommCareUser.create(
-            self.domain,
+        cls.web_user.is_superuser = True
+        cls.web_user.save()
+        cls.commcare_user = CommCareUser.create(
+            cls.domain,
             'bug-kid',
             password='***',
         )
-        self.url = reverse("bug_report")
+        cls.url = reverse("bug_report")
 
-    def tearDown(self):
+    @classmethod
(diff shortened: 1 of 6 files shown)


emord (approved):

corehq/apps/style/templates/style/includes/modal_report_issue.html
```diff
@@ -72,9 +72,23 @@ <h4 class="modal-title">{% trans "Report an Issue with CommCare HQ" %}</h4>
                         </div>
                         {% endif %}
 
+                        {% if request.couch_user.is_domain_admin %}
```
sravfeyn: [From the [spec](https://docs.google.com/document/d/12r6fkzpy5vlVj3_i0qwcBzu8VeizJDIAPYVyDpQ4uzw/edit#)] If the project-description already exists, this field should be hidden. It doesn't look like this block is handling that?

corehq/apps/hqwebapp/tests/test_views.py
```diff
@@ -9,3 +15,99 @@ def test_autocomplete_enabled(self):
 
     def test_autocomplete_disabled(self):
         self.verify(False, reverse("login"), "auth-username")
+
+
+class TestBugReport(TestCase):
```
sravfeyn: 💯

corehq/apps/style/templates/style/includes/modal_report_issue.html
benrudolph: hmm i do see that now, when i first read it i thought it was ensuring that they did not fill have to fill it out twice. i feel like that's a confusing UX to have a box appear for project description once, and then never again. there's no reason anyone would know that this corresponds to the box in the project settings if they ever wanted to change it

benrudolph: it's easy enough implement, but i'm going to offline with @orangejenny and see what she thinks about the UX

benrudolph: ok offlined with robin, we are good to go to leave it the way it's currently implemented

emord:
```
FAIL: test_project_description_web_user (corehq.apps.hqwebapp.tests.test_views.TestBugReport)
  File "/mnt/commcare-hq-ro/corehq/apps/hqwebapp/tests/test_views.py", line 86, in test_project_description_web_user
    self.assertEqual(domain_object.project_description, 'Great NGO, Just Great')
AssertionError: None != 'Great NGO, Just Great'
```

benrudolph:
✅

PR #14157, opened 2016-12-07
ICDS: Remove Primary Key from Case ID

sheelio:
This is causing issues with some ICDS reports that are display raw data from the tables. 

@emord @benrudolph to review and merge. 

@sravfeyn We'll also need this deployed to NIC.

### DIFF
--- a/custom/icds_reports/ucr/data_sources/ccs_record_cases.json
+++ b/custom/icds_reports/ucr/data_sources/ccs_record_cases.json
@@ -45,7 +45,7 @@
         },
         "transform": {},
         "is_nullable": false,
-        "is_primary_key": true,
+        "is_primary_key": false,
         "column_id": "case_id",
         "type": "expression"
       },
--- a/custom/icds_reports/ucr/data_sources/child_health_cases.json
+++ b/custom/icds_reports/ucr/data_sources/child_health_cases.json
@@ -112,7 +112,7 @@
         },
         "transform": {},
         "is_nullable": false,
-        "is_primary_key": true,
+        "is_primary_key": false,
         "column_id": "case_id",
         "type": "expression"
       },


emord:
I think `is_primary_key` should be defaulted to `false`, would it be better to just remove all instances of that?

sheelio:
I think HQ will add it back in automatically (these were regular data sources before being converted to static).

Can remove if you think cleaner.

czue:
@sheelio what are NIC reporting timelines? it's quite a process to deploy things and just want to confirm whether these are urgent or can wait until that process is smoother.

sheelio:
Not urgent, we can bundle into a single deploy later.   

This issue will affect the phone and web-based MPR reports until deployed, but I don't think that will be used for a little bit of time.

emord:
This is good to go from my side. 

@calellowitz @benrudolph @czue how do you guys want to handle bundling up deploys? Assuming we have to do several patches in a row, should these PRs be tagged with something like "icds to be patched"?

benrudolph:
i'm in favor of the tagging system. this does seem unmaintainable though

calellowitz:
@czue @emord @sravfeyn 
Is there a reason not to just do a normal fab deploy when the patch server comes online? It gets rid of the need to worry about tagging branches and manually patching. I understand if there are political restrictions around that though

benrudolph:
i think it just depends on how urgent this is to get into the codebase. if it's not time sensitive, then i'm all for waiting for the patch server

czue:
agreed we should wait unless @sheelio or india team explicitly needs something on a particular date.

czue:
i think we can decide at some point soonish whether we need a more maintainable solution to this problem based on how things progress at NIC

sheelio:
Closing this since its already fixed in a different PR.

PR #14158, opened 2016-12-07
Je/split out columns

emord:
http://manage.dimagi.com/default.asp?242310

@czue @proteusvacuum 

This splits out hte columns. I went ahead and removed the labels as well, so now it's just text.

cc @benrudolph

### DIFF
--- a/corehq/apps/reports/standard/deployments.py
+++ b/corehq/apps/reports/standard/deployments.py
@@ -36,7 +36,7 @@ from corehq.apps.reports.datatables import DataTablesHeader, DataTablesColumn, D
 from corehq.apps.reports.filters.select import SelectApplicationFilter
 from corehq.apps.reports.generic import GenericTabularReport
 from corehq.apps.reports.standard import ProjectReportParametersMixin, ProjectReport
-from corehq.apps.reports.util import format_datatables_data
+from corehq.apps.reports.util import format_datatables_data, numcell
 
 
 class DeploymentsReport(GenericTabularReport, ProjectReport, ProjectReportParametersMixin):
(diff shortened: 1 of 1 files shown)


corehq/apps/reports/standard/deployments.py
```diff
@@ -98,9 +70,16 @@ def headers(self):
...
+            DataTablesColumn(_("Application"),
+                             help_text=_("Displays application of last submitted form")),
+            DataTablesColumn(_("Application Version"),
+                             help_text=_("""Displays application version of the last submitted form;
+                                         The currently deployed version may be different."""),
+                             sort_type=DTSortType.NUMERIC),
+            DataTablesColumn(_("CommCare Version"),
+                             help_text=_("""Displays CommCare version the user last submitted with;
+                                         The currently deployed version may be different."""),
+                             sort_type=DTSortType.NUMERIC),
```
proteusvacuum: looks like the helper function returns a string?

proteusvacuum (commented):
Nice cleanup. One question.

proteusvacuum (approved):

corehq/apps/reports/standard/deployments.py
```diff
@@ -98,9 +70,16 @@ def headers(self):
...
             DataTablesColumn(_("Last Sync"),
                              sort_type=DTSortType.NUMERIC),
-            DataTablesColumn(_("Application (Deployed Version)"),
-                help_text=_("""Displays application version of the last submitted form;
-                            The currently deployed version may be different."""))
+            DataTablesColumn(_("Application"),
+                             help_text=_("Displays application of last submitted form")),
+            DataTablesColumn(_("Application Version"),
+                             help_text=_("""Displays application version of the last submitted form;
+                                         The currently deployed version may be different."""),
```
czue: does this not add a big blob of whitespace to the string?

czue (approved):
hawt. :fire: 

one question but otherwise :+1:

corehq/apps/reports/standard/deployments.py
emord: Maybe? but it doesn't matter once it gets down to HTML since whitespace isn't significant there. I just copied this way of doing it from the old Application Deployed Version column

image

czue: yeah agree is no biggie. it does create some oddities in our translation files like this:
```
#: corehq/apps/reports/standard/deployments.py:102
msgid ""
"Displays application version of the last submitted form;\n"
"                            The currently deployed version may be different."
msgstr ""
```
so is something i tend to avoid, but agree is not the most important thing in the world.

PR #14143, opened 2016-12-07
Fix delay issue

sravfeyn:
Fixes the initial location delay issue when loading 'Edit Mobile Worker' page. [FB](http://manage.dimagi.com/default.asp?240625#1257616). Also does small refactor.

PR #14159, opened 2016-12-07
Mk 241159 notify user for error

mkangia:
This PR started with pre-populating locations on user edit page instead of loading it asyncly since that had a huge delay. Sravan took care of that aspect [here](https://github.com/dimagi/commcare-hq/pull/14143). So i just retained some precaution to enforce user to assign a primary location if a location is assigned.

This does couple of things
1. for a user that has access to all locations, still filter for just the current domain.
2. give a warning to user if not selecting primary location after assigning a location and avoid a request that finally returns an error

### DIFF
--- a/corehq/apps/locations/models.py
+++ b/corehq/apps/locations/models.py
@@ -282,7 +282,7 @@ class LocationQueriesMixin(object):
 
     def accessible_to_user(self, domain, user):
         if user.has_permission(domain, 'access_all_locations'):
-            return self.all()
+            return self.filter(domain=domain)
 
         assigned_location_ids = user.get_location_ids(domain)
         if not assigned_location_ids:
--- a/corehq/apps/locations/templates/locations/manage/partials/drilldown_location_widget.html
+++ b/corehq/apps/locations/templates/locations/manage/partials/drilldown_location_widget.html
@@ -34,6 +34,14 @@ $(function() {
             $('#{{ css_id }}').val(null);
         }
     }).trigger('change');
+
+    $('#{{ submit_form_id }}').on('submit', function() {
+         // Notify user in case the source list has options but none selected
+         if(!!$('#{{ source_css_id }}').val() && !$('#{{ css_id }}').val()) {
+             alert("{{ empty_primary_location_message|safe }}");
+             return false;
+         }
+    })
 });
 
 </script>
--- a/corehq/apps/locations/templates/locations/manage/partials/drilldown_location_widget.html
+++ b/corehq/apps/locations/templates/locations/manage/partials/drilldown_location_widget.html
@@ -15,6 +15,10 @@ $(function() {
             data: {'results': $(e).select2('data')},
         }
         $('#{{ css_id }}').select2(options);
+        option_ids = _.map(options.data.results, function(option) { return option.id })
+        if(!option_ids.includes($('#{{ css_id }}').val())) {
+            $('#{{ css_id }}').val(null);
+        }
     };
 
     // This custom event is fired in autocomplete_select_widget.html
--- a/corehq/apps/users/forms.py
+++ b/corehq/apps/users/forms.py
@@ -772,7 +772,7 @@ class PrimaryLocationWidget(forms.Widget):
     Options for this field are dynamically set in JS depending on what options are selected
     for 'assigned_locations'. This works in conjunction with SupplyPointSelectWidget.
     """
-    def __init__(self, css_id, source_css_id, attrs=None):
+    def __init__(self, css_id, source_css_id, submit_form_id, empty_primary_location_message, attrs=None):
         """
         args:
             css_id: css_id of primary_location field
@@ -781,11 +781,15 @@ class PrimaryLocationWidget(forms.Widget):
         super(PrimaryLocationWidget, self).__init__(attrs)
         self.css_id = css_id
         self.source_css_id = source_css_id
+        self.submit_form_id = submit_form_id
+        self.empty_primary_location_message = empty_primary_location_message
 
     def render(self, name, value, attrs=None):
         return get_template('locations/manage/partials/drilldown_location_widget.html').render(Context({
             'css_id': self.css_id,
             'source_css_id': self.source_css_id,
+            'submit_form_id': self.submit_form_id,
+            'empty_primary_location_message': self.empty_primary_location_message,
             'name': name,
             'value': value or ''
         }))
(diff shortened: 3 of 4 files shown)


corehq/apps/users/forms.py
```diff
@@ -813,12 +817,25 @@ def __init__(self, *args, **kwargs):
...
+
         self.fields['assigned_locations'].widget = SupplyPointSelectWidget(
             self.domain, multiselect=True, id='id_assigned_locations'
         )
         self.fields['primary_location'].widget = PrimaryLocationWidget(
             css_id='id_primary_location',
-            source_css_id='id_assigned_locations'
+            source_css_id='id_assigned_locations',
+            submit_form_id=self.helper.form_id,
+            empty_primary_location_message = _("Primary location can't be empty if user has any locations set")
```
dimagimon: E251 unexpected spaces around keyword / parameter equals

mkangia:
Hey, reaching out for review. This is good to go now.
@esoergel @sravfeyn 
buddy : @calellowitz @nickpell

corehq/apps/locations/models.py
```diff
@@ -282,7 +282,7 @@ def couch_locations(self, wrapped=True):
 
     def accessible_to_user(self, domain, user):
         if user.has_permission(domain, 'access_all_locations'):
-            return self.all()
+            return self.filter(domain=domain)
```
czue: yikes good catch. is this affecting any behavior on prod right now?

corehq/apps/locations/templates/locations/manage/partials/drilldown_location_widget.html
```diff
@@ -15,6 +15,10 @@
             data: {'results': $(e).select2('data')},
         }
         $('#{{ css_id }}').select2(options);
+        option_ids = _.map(options.data.results, function(option) { return option.id })
+        if(!option_ids.includes($('#{{ css_id }}').val())) {
+            $('#{{ css_id }}').val(null);
+        }
```
esoergel: Nice!  This definitely seems like the correct behavior.

corehq/apps/locations/templates/locations/manage/partials/drilldown_location_widget.html
```diff
@@ -34,6 +34,14 @@
...
         }
     }).trigger('change');
+
+    $('#{{ submit_form_id }}').on('submit', function() {
+         // Notify user in case the source list has options but none selected
+         if(!!$('#{{ source_css_id }}').val() && !$('#{{ css_id }}').val()) {
+             alert("{{ empty_primary_location_message|safe }}");
+             return false;
+         }
+    })
```
esoergel: I think we discourage using `alert`, is there another way to cue the user?  This could mark the widget with an `.error` class so it's red and show the help text underneath.  Alternatively, it could `.disable` the submit button and warn of the error.

corehq/apps/users/forms.py
```diff
@@ -781,11 +781,15 @@ def __init__(self, css_id, source_css_id, attrs=None):
         super(PrimaryLocationWidget, self).__init__(attrs)
         self.css_id = css_id
         self.source_css_id = source_css_id
+        self.submit_form_id = submit_form_id
+        self.empty_primary_location_message = empty_primary_location_message
```
esoergel: Nit: Does this need to be parameterized?  What's the advantage over just defining it right here?

sravfeyn:
Trying to understand the functionality of second point. The [PR ](https://github.com/dimagi/commcare-hq/pull/14143) I did sets primary-location automatically, if even one location is assigned so that we won't have to check and warn if user has not set primary-location (on js side). On server side the validation for empty primary-location is done and user gets a warning that way as well. What is the issue of second point?

corehq/apps/locations/templates/locations/manage/partials/drilldown_location_widget.html
mkangia: Noted.

mkangia:
@sravfeyn when i worked on this, the default selection for primary location that you added was not there yet. Hence the step taken.
On production, I see that the user can still remove the primary location assigned and submit. We can choose to either let user go through another request to know that its necessary or make sure that the user is not able to submit without it.

corehq/apps/locations/templates/locations/manage/partials/drilldown_location_widget.html
mkangia: Even though the options for primary location was cleared by 'select2' its value was still retained. Which was later handled at server though. Hence made this change.

sravfeyn:
Given that we do automatic-setting of primary location now and that we are going to hide the primary-location field into a default collapsed field to avoid user noticing it, a user removing primary-location is going to be very low frequency and since there is a server side warning even for that, it's probably not worth to add the warning at the cost of more complexity to the primary-location-widget (it's already coupled with css elements from various pages in a non-obvious way, I wrote it ;-) ). 

I could be wrong on assumptions though. Let's differ to @sheelio on the UX?

You could decouple the 1 and 2 changes to get #1 merged without waiting for sheel on #2?

mkangia:
@sravfeyn Thanks Sravan. `hide the primary-location field into a default collapsed field to avoid user noticing it`. That takes care of everything. I am going to go ahead and remove all the checks then and just keep the necessary changes which should include the first point. Just checking if https://github.com/dimagi/commcare-hq/pull/14159/files#diff-365eab9e49926ea6eed7f768d4705a27R18 is needed now?

corehq/apps/locations/templates/locations/manage/partials/drilldown_location_widget.html
sravfeyn: Sorry for duplication again, I think I fixed it in [here](https://github.com/dimagi/commcare-hq/pull/14163/files#diff-365eab9e49926ea6eed7f768d4705a27R38). If the value for primary location is not in the available options, then it is set as [first option](https://github.com/dimagi/commcare-hq/pull/14163/files#diff-365eab9e49926ea6eed7f768d4705a27R38) from available option. Otherwise, it is set to [null](https://github.com/dimagi/commcare-hq/pull/14163/files#diff-365eab9e49926ea6eed7f768d4705a27R32)

mkangia:
thanks @sravfeyn. I think i will close this PR. Made a new one [here](https://github.com/dimagi/commcare-hq/pull/14255) for the first issue.
@esoergel

PR #14160, opened 2016-12-07
Use proper nimbus URL

benrudolph:
@snopoke sorry should have caught this, but we use the fully qualified URL and CORS locally while on prod we just hit `/formplayer` and route to the correct server by using nginx. it was a neat trick to get the formplayer_url to work both locally and in prod from the client side, however it doesn't play as nicely on the server :/

cc: @emord


snopoke:
:+1:

PR #14161, opened 2016-12-07
remove python 2.6 from docs

nickpell:
@millerdev


nickpell:
self-merging doc changes

PR #14166, opened 2016-12-08
Revert "Case import history prefactor"

emord:
Reverts dimagi/commcare-hq#14134

Reverting as I think this is causing errors in the bulk user upload

PR #14170, opened 2016-12-08
Revert "Revert "Case import history prefactor""

dannyroberts:
Reverts dimagi/commcare-hq#14166

@emord cc: @czue


dannyroberts:
I also checked other code usages for a similar error, but the rest of them just pass it directly to the view where None and missing are essentially treated the same way.

emord (approved):

dannyroberts:
thanks for catching this @emord. Sloppy implicit/_ad hoc_ dict schemas are always the trickiest to refactor without something subtle like this happening. I'll keep that in mind next time I do something like this.

czue:
sneaky bugger. :beetle:

PR #14173, opened 2016-12-08
Don't auto select new columns when editing an export

emord:
http://manage.dimagi.com/default.asp?243436#1258842

Feels a little hacky, but needed to feed it through that it was editing an export instead of creating one

@benrudolph


benrudolph (approved):
yep, was how i was going to do. hate adding this kind of complexity 😢

PR #14174, opened 2016-12-08
add xmlns and d3 data

benrudolph:
@esoergel i believe something is completely busted with our `PieGraph` constructor. it's really hard for me to test all the places it's used because many of them are custom reports, so i decided to just to hack the `chart.data` as i saw other instances do. also the xmlns thing is a total hack, but again this report is on its way out and having a hard to justifying spending lots of time fixing it

http://manage.dimagi.com/default.asp?240149

cc: @emord

### DIFF
--- a/corehq/apps/reports/commtrack/data_sources.py
+++ b/corehq/apps/reports/commtrack/data_sources.py
@@ -13,7 +13,7 @@ from dimagi.utils.couch.loosechange import map_reduce
 from corehq.apps.reports.api import ReportDataSource
 from datetime import datetime, timedelta
 from dateutil import parser
-from casexml.apps.stock.const import SECTION_TYPE_STOCK
+from casexml.apps.stock.const import SECTION_TYPE_STOCK, COMMTRACK_REPORT_XMLNS
 from casexml.apps.stock.models import StockReport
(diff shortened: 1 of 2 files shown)


emord (approved):
Seems like a reasonable hack if this is on it's way out

corehq/apps/reports/commtrack/data_sources.py
```diff
@@ -461,6 +461,7 @@ def get_data(self):
             )
 
             form_xmlnses = [form['xmlns'] for form in self.all_relevant_forms.values()]
+            form_xmlnses.append(COMMTRACK_REPORT_XMLNS)
```
esoergel: This somewhat changes the meaning of the report.  If someone DOES want to use the filter to show reporting rates from a specific form, AND the project has SMS stock updates, then those would obscure the data.  In my opinion, that's enough of an edge case that this solution wins the tradeoff, considering that currently there's no way to see consider data submitted by SMS.

esoergel (approved):

corehq/apps/reports/commtrack/data_sources.py
benrudolph: yeah fair point, but also agree probably an edge case and not worth the effort to do this fix properly

PR #14175, opened 2016-12-08
Icds migration

emord:
Added migration for @sheelio 

http://manage.dimagi.com/default.asp?243659#1258866

@benrudolph

### DIFF
--- a/custom/icds_reports/migrations/0003_auto_20161208_2204.py
+++ b/custom/icds_reports/migrations/0003_auto_20161208_2204.py
@@ -1,0 +1,19 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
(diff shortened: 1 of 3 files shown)


benrudolph (approved):

czue:
:athletic_shoe: :hammer_and_wrench:

snopoke:
We should test these of softlayer first, also I think @sravfeyn had some other similar changes that could be bundled in to the same patch.

sravfeyn:
_This_ was the migration I was looking for. I think Sheel might have already tested these out on softlayer @sheelio ?

sheelio:
I've tested it locally against a copy of the DB on softlayer.

sravfeyn:
(Patched and Migrated)

PR #14177, opened 2016-12-08
Copy to clipboard button for raw xml

benrudolph:
@proteusvacuum http://manage.dimagi.com/default.asp?243589

cc: @emord 
copy xml


proteusvacuum:
😍 
Those buttons are really prominent for something that I imagine is used almost never, but as long as I can double click in that area and it does what I expect, this is awesome.

corehq/apps/reports/templatetags/xform_tags.py
```diff
@@ -37,7 +37,7 @@ def render_form_xml(form):
     if isinstance(xml, unicode):
         xml.encode('utf-8', errors='replace')
     formatted_xml = indent_xml(xml) if xml else ''
-    return '<pre class="fancy-code prettyprint linenums"><code class="no-border language-xml">%s</code></pre>' \
```
proteusvacuum: I think this is the only instance of this anywhere in the codebase... which means I think we can remove all the `{% include "imports/fancy-code.html" %}` everywhere, and also the dependency on `jQuery.selectText`. Hooray!

emord (approved):
Maybe one day we can use codemirror :)

corehq/apps/reports/templatetags/xform_tags.py
benrudolph: nice, ok i can add that in

proteusvacuum (approved):
☠️

czue:
:high_heel: :construction_worker_man:

corehq/apps/hqwebapp/templates/imports/fancy-code.html
```diff
@@ -1,26 +0,0 @@
-{% load hq_shared_tags %}
-{% include "imports/bootstrap.google-prettify.html" %}
```
czue: I think this file can also be deleted now

czue (approved):
love this. :skull:

PR #14178, opened 2016-12-09
Case import history: phase 2

dannyroberts:
- first two commits: make case importer return to landing page when initiated instead of special progress polling page
- the rest: expose original upload file in case import history (allow domain admins to download all of them, allow everyone else to download their own; we can modify the permissions in the future to be more permissive, but wanted something simple that definitely doesn't give anyone access to case data they shouldn't have)

screen shot 2016-12-09 at 10 31 14 am


corehq/apps/case_importer/tracking/filestorage.py
```diff
@@ -0,0 +1,87 @@
...
+        :meta_model is a django model used to store meta info
+            must contain columns identifier, filename, length
+        """
+
+        self._bucket = bucket
+        self._db = get_blob_db()
+        self._meta_model = meta_model
+
+    def write_file(self, f, filename):
+        blob_info = self._db.put(f, bucket=self._bucket)
```
millerdev: > I was thinking of (bucket, identifier) as being roughly equivalent to couchdb's (database, _id)

(bucket, identifier) is more like (database + _id, attachment_name), although not exactly.

Yes, that sounds like a fine solution. Maybe keep the bucket prefix on the shorter side if there will be many of these since the bucket prefix is used in the key, which adds storage (and RAM for some configurations) overhead for the blob db.

dannyroberts: I suppose 'case_importer/upload_files' is too long?

dannyroberts:
@millerdev check out my last PR. I went with adding 128 bits of random padding rather than using the upload id, just because it made the interface really unintuitive ("pass me some id, and then I'll pass you back a different id, and then when you want to get your thing back give me both ids") and it wasn't really adding anything.

czue:
@dannyroberts just ping when you're ready for review. one question I had was whether we should be cognizant of the storage overhead of this change and also of a possible site vulnerability? what happens if i just start spamming 1GB importer uploads?

corehq/apps/case_importer/tracking/views.py
```diff
@@ -32,8 +37,34 @@ def to_user_json(case_upload):
...
+    if not user_may_view_file_upload(domain, request.couch_user, case_upload):
+        return HttpResponseForbidden()
+
+    def streaming_content():
+        tempfile = case_upload.get_tempfile()
+        with open(tempfile, 'rb') as f:
+            for chunk in f:
+                yield chunk
+
+    response = StreamingHttpResponse(streaming_content())
```
snopoke: Instead of having Django serve this content we can get Nginx to do it either directly from the NFS drive or from Riak (though I don't think we've configured Nginx for Riak)

corehq/apps/case_importer/urls.py
```diff
@@ -1,18 +1,18 @@
...
 )
 
 urlpatterns = [
     url(r'^excel/config/$', excel_config, name='excel_config'),
     url(r'^excel/fields/$', excel_fields, name='excel_fields'),
     url(r'^excel/commit/$', excel_commit, name='excel_commit'),
-    url(r'^importer_ajax/(?P<download_id>[0-9a-fA-Z]{25,32})/$',
-        importer_job_poll, name='importer_job_poll'),
     url(r'^history/uploads/$', case_uploads, name='case_importer_uploads'),
+    # allow '---' as a dummy value to be able to reverse url in javascript w/o upload_id
```
snopoke: or you could have a more relaxed regex `[\w-]+`. Any reason to have such a strict regex?

corehq/apps/case_importer/tracking/filestorage.py
```diff
@@ -0,0 +1,94 @@
...
+        padding, blob_identifier = identifier.split('/')
+        content = self._db.get(blob_identifier, bucket=self._get_bucket(padding)).read()
+        return make_temp_file(content, suffix)
+
+    @memoized
+    def get_filename(self, identifier):
+        return self._meta_model.objects.values_list('filename', flat=True).get(identifier=identifier)
+
+
+class TransientFileStore(object):
```
snopoke: Why is this better than what it's replacing (the soil interfaces)?

corehq/apps/case_importer/tracking/case_upload_tracker.py
```diff
@@ -13,48 +15,51 @@ def __init__(self, upload_id):
...
+        return get_spreadsheet(self.get_tempfile(), named_columns)
 
     def trigger_upload(self, domain, config):
         from corehq.apps.case_importer.tasks import bulk_import_async
         task = bulk_import_async.delay(config, domain, self.upload_id)
+        original_filename = transient_file_store.get_filename(self.upload_id)
+        with open(self.get_tempfile()) as f:
+            identifier = persistent_file_store.write_file(f, original_filename)
+
+        case_upload_file_meta = CaseUploadFileMeta.objects.get(identifier=identifier)
```
snopoke: Considering this was just created by `persistent_file_store.write_file` it seems odd to have to immediately fetch it. Why not have it returned by `write_file`?

snopoke (commented):
Generally OK. Just confused why you're moving away from using Soil for caching the file contents.

corehq/apps/case_importer/tracking/filestorage.py
```diff
@@ -17,21 +18,27 @@ def __init__(self, bucket, meta_model):
...
+    def _get_bucket(self, padding):
+        return '{}/{}'.format(self._bucket_prefix, padding)
+
     def write_file(self, f, filename):
-        blob_info = self._db.put(f, bucket=self._bucket)
-        self._meta_model(identifier=blob_info.identifier, filename=filename,
+        padding = random_url_id(16)
+        blob_info = self._db.put(f, bucket=self._get_bucket(padding))
+        identifier = '{}/{}'.format(padding, blob_info.identifier)
+        self._meta_model(identifier=identifier, filename=filename,
```
dannyroberts: Sorry this is just indirection. `self._meta_model` is set to `CaseUploadFileMeta`.

corehq/apps/case_importer/urls.py
dannyroberts: No you're right, I'll change to `[\w-]+`; was just going off of what was there before.

corehq/apps/case_importer/tracking/filestorage.py
```diff
@@ -0,0 +1,94 @@
...
+    def write_file(self, f, filename):
+        padding = random_url_id(16)
+        blob_info = self._db.put(f, bucket=self._get_bucket(padding))
+        identifier = '{}/{}'.format(padding, blob_info.identifier)
+        self._meta_model(identifier=identifier, filename=filename,
+                         length=blob_info.length).save()
+        return identifier
+
+    @memoized
+    def get_tempfile(self, identifier):
```
dannyroberts: Yeah I like that suggestion. This was modeled after the maybe even-more-poorly-named `DownloadBase.get_filename()`, which does essentially the same thing. I think maybe most correct would be `get_tempfile_ref_for_contents`; I'll go with that even though it might be a mouth full.

snopoke:
Yea, I agree with that, it would be great to separate those two use cases.
I think if you're proposing a this as a replacement for the file storage /
download expose part of soil then it should be in a separate app and not
part of case importer.

dannyroberts:
> @dannyroberts just ping when you're ready for review. one question I had was whether we should be cognizant of the storage overhead of this change and also of a possible site vulnerability? what happens if i just start spamming 1GB importer uploads?

that's a good question. I'll answer the specific example but also curious if you meant something else by site vulnerability like an exploit that takes advantage of us storing any file they give us (in this case, it's only stored once it's been opened as an excel file, if that's any help), or if you just mean a sort of DoS by quickly filling our storage space like the example you mentioned.

If we want to guard against that, I can think of a couple options. One is to not allow file uploads over a certain size, but I can see that actually hurting a project if the size is set too low; another option is to only allot some maximum amount of storage for a single domain, and start deleting old upload files once it's exceeded. Or I suppose, both. Are there any other places in the code where this could be a problem that we feel we've addressed it well? The only thing that comes to mind right now is form attachments—do we feel like we've dealt with the similar threat there well?

snopoke:
I don't think we've addressed (or even considered) this anywhere else. When
we started using Riak I though we might go with a bucket / domain scheme
which would allow us to use Riak's built in storage tracking etc. but we
aren't using that approach so don't think that's an option. It does seem
reasonable to limit this in some way (per file and probably per domain).

In terms of the per file limit, uploading huge Excel files also causes
issues with celery and makes the task more likely to fail so forcing them
to break up huge files into smaller ones could also reduce the number of
attempts they have to make to get the import through.

corehq/apps/case_importer/tracking/filestorage.py
```diff
@@ -0,0 +1,94 @@
...
+from corehq.apps.case_importer.tracking.models import CaseUploadFileMeta
+from corehq.blobs import get_blob_db
+from corehq.blobs.util import random_url_id
+from corehq.util.files import file_extention_from_filename
+from dimagi.utils.decorators.memoized import memoized
+
+BUCKET_PREFIX = 'case_importer'
+
+
+class PersistentFileStore(object):
```
dannyroberts: Thanks yeah, I agree that could help a lot.

corehq/apps/case_importer/tracking/filestorage.py
```diff
@@ -0,0 +1,94 @@
...
+    def __init__(self, bucket, meta_model):
```
dannyroberts: As I think you could tell, I wrote this from the perspective of what general storage interface I wish we had, and I imagined that different apps that used it would want to store the meta-data in different tables.

> Yea, I agree with that, it would be great to separate those two use cases.
I think if you're proposing a this as a replacement for the file storage /
download expose part of soil then it should be in a separate app and not
part of case importer.

I can certainly do that, if you think this is good. If I do that, it'd make it more important for me to answer the question of whether different apps using this should have their meta data in different tables. The plus side is that makes it pretty easy to see things like the combined size of all case imports, that's maybe not that important and you can get that with a join assuming the tables aren't so huge that that becomes infeasible, which we probably don't have to worry about for this feature. Is it fine practice for lots of different models to contain foreign keys to the same central table? (So `CaseUploadRecord` would have a foreign key to `PersistentFileStorageMeta`, as would lots of other tables?)

If you don't think it's worth me generalizing this, then I can punt that question and make this more explicitly. If you do think it's worth it, maybe this can undergo architecture review.

snopoke:
If we're planning on applying the 'upload history' pattern to lookup
tables, locations, users etc. then I think it's worth generalizing though
if the timing of those is still very far out then perhaps not worth it
until we get there. We it wouldn't be too hard to migrate the data to more
generic model when the time comes.

czue:
>Are there any other places in the code where this could be a problem that we feel we've addressed it well? The only thing that comes to mind right now is form attachments—do we feel like we've dealt with the similar threat there well?

Good general question and point. I think I'm more concerned about users accidentally harming us than us being perfectly protected from DOS attacks (which we are definitely not - form submissions and attachments being a good example). Like uploading the same big file 20 times is actually a real-world use case I could imagine someone doing because they were having issues or confused or had a weird workflow or whatever. So yeah, was wondering about either file size restrictions or something like auto-deleting these after a certain amount of time would be good. Also just quite aware of our Riak storage being a bottleneck right now and so want to be careful about introducing new workflows that could heavily impact it.

corehq/apps/case_importer/tracking/filestorage.py
dannyroberts: @millerdev and I just had a good talk about blobdb interface and how I can very simply improve upon it to not have to have these weird '{my-128-bit}/{blob-db-short-id}' keys.

The outcome is that I'll change blobdb.put to accept an `identifier`, and if it's not passed in, to default to the current behavior of auto-generating short ids. I'll also use the currently unused `basename` param, being careful to make sure it's not passed by any callers. Then @millerdev will have it on his plate to (at some time in the future) make `identifier` a required field so that the default isn't using short ids that result in a collision if your bucket is just a static namespace, and changing everywhere that uses that functionality to generate their own short ids.

dannyroberts:
@snopoke yeah that's a good point, and would definitely be useful to feel good about how the utility is used in this one case before applying it to a bunch of places. For this PR, let's keep it inside `case_importer`, but I'll have it on my mind how this fits in as a general tool, and I talked with @millerdev just now about how that might work.

corehq/apps/case_importer/tracking/case_upload_tracker.py
```diff
@@ -1,9 +1,11 @@
-from corehq.apps.case_importer.tracking.models import CaseUploadRecord
+from corehq.apps.case_importer.exceptions import ImporterRefError
+from corehq.apps.case_importer.tracking.filestorage import transient_file_store, \
+    persistent_file_store
+from corehq.apps.case_importer.tracking.models import CaseUploadRecord, \
```
dimagimon: F401 'CaseUploadFileMeta' imported but unused

corehq/apps/case_importer/tracking/filestorage.py
dannyroberts: (done)

corehq/apps/case_importer/tracking/case_upload_tracker.py
dannyroberts: Hey @snopoke after taking a closer look I saw the reason in your words and went with the approach of having write_file return a file meta object. For transient storage, it just returns it as a namedtuple; for persistent storage, it returns the saved `CaseUploadFileMeta` object.

dannyroberts:
@snopoke @millerdev I think I've now covered all the PR review on this. Let me know if you have anymore feedback!

corehq/apps/case_importer/templates/case_importer/import_cases.html
```diff
@@ -4,11 +4,23 @@
 
 {% block js-inline %}
 <script id="caseUploadsResource" type="text/plain">{% url 'case_importer_uploads' domain %}</script>
+<script id="caseUploadFileDownloadResource" type="text/plain">{% url 'case_importer_upload_file_download' domain '---' %}</script>
+<script id="caseUploadPermissionDeniedTitle" type="text/plain">{% trans 'You may not download this file yourself' %}</script>
+<script id="caseUploadPermissionDeniedContent" type="text/plain">{% trans 'Please ask an administrator or the user who uploaded the file for their help' %}</script>
```
snopoke: I still don't really understand why this is better than this:
```
var caseUploadPermissionDeniedTitle = '{% trans "some text" %}';
```

corehq/apps/case_importer/templates/case_importer/import_cases.html
```diff
@@ -142,10 +155,30 @@
...
                     {% include 'case_importer/partials/ko_import_status.html' %}
                 </td>
-                <td class="col-md-2" data-bind="text: created"></td>
+                <td class="col-md-2">
+                    <!--ko if: upload_file_name-->
+                    <!--ko if: !upload_file_download_allowed-->
+                        <a style="cursor: pointer" data-bind="
+                            text: upload_file_name,
+                            popover: {
+                                title: $root.caseUploadPermissionDeniedTitle,
```
snopoke: couldn't this be `django.gettext`?

corehq/apps/case_importer/tracking/views.py
```diff
@@ -32,8 +37,28 @@ def to_user_json(case_upload):
...
+def case_upload_file(request, domain, upload_id):
+    try:
+        case_upload = CaseUploadRecord.objects.get(upload_id=upload_id, domain=domain)
+    except CaseUploadRecord.DoesNotExist:
+        return HttpResponseNotFound()
+
+    if not user_may_view_file_upload(domain, request.couch_user, case_upload):
+        return HttpResponseForbidden()
+
+    response = StreamingHttpResponse(open(case_upload.get_tempfile_ref_for_upload_ref(), 'rb'))
```
snopoke: I still think it would be better to let nginx handle this than to tie up resources on a web worker.

corehq/apps/case_importer/tracking/filestorage.py
```diff
@@ -0,0 +1,109 @@
...
+        :meta_model is a django model used to store meta info
+            must contain columns identifier, filename, length
+        """
+
+        self._bucket = bucket
+        self._db = get_blob_db()
+        self._meta_model = meta_model
+
+    def write_file(self, f, filename):
+        identifier = random_url_id(16)
```
snopoke: any reason not to just use a UUID here?

snopoke (approved):
A few minor comments but otherwise GTG from me. Still has "Don't merge" label.

corehq/apps/case_importer/tracking/views.py
snopoke: Yes that's right. You have to write the data to the shared drive first. Here's an example: https://github.com/dimagi/commcare-hq/blob/859c1e5821ffeb63afcad4e59fcc93c6a2d54f2d/corehq/apps/hqmedia/tasks.py#L119-L119 (see `settings.SHARED_DRIVE_CONF.transfer_enabled`).

Fine to merge without that though.

Note to self: We should configure nginx so that it can serve data straight out of Riak.

corehq/apps/case_importer/tracking/filestorage.py
dannyroberts: I believe the reason is that this function packs the same number of bits into fewer characters, and riak cares somewhat about key length, since uuids are usually given in hex, and this uses base64. I'm mostly just using what blobdb itself uses.

corehq/apps/case_importer/templates/case_importer/import_cases.html
dannyroberts: Yeah this is changed in https://github.com/dimagi/commcare-hq/compare/case-import-history-phase2...case-import-history-phase-1-feedback, which I haven't PR'd yet. Sorry that's confusing.

corehq/apps/case_importer/tracking/filestorage.py
snopoke: :+1: that's a good reason

corehq/apps/case_importer/templates/case_importer/import_cases.html
snopoke: I thought I'd seen that somewhere.

corehq/apps/case_importer/templates/case_importer/import_cases.html
dannyroberts: I changed this in https://github.com/dimagi/commcare-hq/compare/case-import-history-phase2...case-import-history-phase-1-feedback as well, but not to that.

FWIW the reason I didn't go with

```
var caseUploadPermissionDeniedTitle = '{% trans "some text" %}';
```

is because it's not safe to dump arbitrary text into a string-in-javascript(-in-html) context, both just pedantically, and also in the specific case of translations, an `'` is actually a legitimate concern.

```
var iLoveThis = '{% trans "I love this!" %}';
```

would be rendered in French as

```
var iLoveThis = 'J'adore ça!';
```

which is a javascript syntax error.

dannyroberts:
As for "Don't merge" label, I'm still just doing some final testing on staging—which is actually important since I don't have riak running locally, so bugs that only show up on the s3db backend I wouldn't catch locally.

millerdev (approved):

dannyroberts:
ok still works great on staging, so I'm removing the don't merge label. If no one beats me to it, I'll merge when the final build passes.

dannyroberts:
Weee! Thanks for all the help reviewing, Simon and Daniel!

czue:
@dannyroberts just to check did my concerns about storage space get responded to? i didn't see any follow up

dannyroberts:
@czue sorry I got caught up in the other feedback and didn't realize I hadn't finished that conversation until now. Thanks for pinging me on it.

It's worth noting that we currently have nginx set up to reject (with a `413 Request Entity Too Large`) any requests whose bodies are larger than 500m, so technically the scenario where users repeatedly upload a 1GB file is actually mitigated, but of course if they're repeatedly uploading a 499MB file there's no protection there, and it only takes about twice as long of them doing that to screw us so it's still worth addressing.

One question worth asking is: why is that limit set so high? In fact it used to be set to 100m until around May, when we increased it because of a large multimedia upload that was failing: https://github.com/dimagi/commcarehq-ansible/pull/411. I'm sure at some point someone will try a 1GB upload of multimedia—will we just up the nginx max body size to 1.5GB then? Maybe better would be to decide to hold our current limit or even a lower one, and make sure that areas we know people try very large uploads handle it at least somewhat gracefully by telling the user how to split up their upload.

Narrowing the focus down to case importer and these new file attachments, let me POC a management command to delete old attachments follow a simple set of rules that's unlikely to annoy a user while catching most of what no one's likely to care about—something like "delete all attachments above 2.5GB per domain oldest-first; delete all attachments older than 3 months; and always keep the last 4 uploads per domain, even if in violation of the other two". On top of this, do you think it's worth imposing a lower upload size than then nginx 500m?

dannyroberts:
(And in case it matters, I should have said the "large multimedia upload that was failing" was for ICDS)

czue:
thanks for follow up and no immediate answers to the nginx quandries but agreed with your problem statement and high-level feeling that 500M seems high. feels out of scope for this back and forth to me but maybe good to pursue? not sure.

and just to be clear I wasn't saying that i thought we needed to do something, i was just feeling like we should go through the work of convincing ourselves that we don't.

Like, it would be nice to be able to say something like: "we can happily devote _____ GB to importer files because of ____, and we think that lasts _____ time because of ____, but even if we're wrong it will be fine because ______". 

also possible that the last blank can be filled in with datadog riak disk alerts, in which case we might be able to just punt this for now and address if things start to look like they are causing us real cost at some arbitrary future date. though if there's something we can do now to be able to measure the contribution of this change to that metric that might be good to get in earlier

dannyroberts:
OK, so far I'm experimenting with a datadog gauge for tracking this: https://github.com/dimagi/commcare-hq/pull/14234.

One plan is we could decide like you said "We can happily devote ____ to case upload files", set a low threshold like 10GB that's not controversial, and add an alert that's triggered when that threshold is hit, at which point we can evaluate whether the next step is to raise the limit, introduce a system for auto-deleting old files, or something else.

dannyroberts:
https://app.datadoghq.com/monitors#1400710?group=all&live=4h

czue:
nice! that plan sounds great to me. 

that page isn't showing anything yet. is that because it's not deployed yet or no one has done an import?

dannyroberts:
Because it's not deployed yet. I manually pushed a couple data points from the shell so i could set up the monitor, but it's not getting hourly data yet

PR #14180, opened 2016-12-09
Remove edit_apps restriction for deployment reports

benrudolph:
@czue http://manage.dimagi.com/default.asp?243597#1259307

cc: @emord


emord (approved):

czue:
nice thanks. I found the justification but I think this is better behavior for where we're taking supply so think it is no longer valid.

> Justification: if you're editing apps, you're a system administrator, and only system administrators will want to be looking at those reports a lot.

http://manage.dimagi.com/default.asp?112918

benrudolph:
nice digging, makes sense

PR #14182, opened 2016-12-09
Fix location restriction

benrudolph:
@esoergel @czue this allows mobile works to see the cloudcare tab: http://manage.dimagi.com/default.asp?243701

it also does some work on viewing tabs for location restrictions. there's one bit i'm not sure the best way to fix, will mark in PR

cc: @emord

### DIFF
--- a/corehq/tabs/uitab.py
+++ b/corehq/tabs/uitab.py
@@ -126,9 +126,14 @@ class UITab(object):
         if not self.show_by_default and not self.is_active_tab:
             return False
 
-        if not self.can_access_all_locations and not self.filtered_dropdown_items:
-            # location-safe filtering makes this whole tab inaccessible
-            return False
+        if not self.can_access_all_locations:
+            if self.dropdown_items and not self.filtered_dropdown_items:
+                # location-safe filtering makes this whole tab inaccessible
+                return False
+
+            # Just a button tab, determine if it's location safe
+            if not self.dropdown_items and not url_is_location_safe(self.url):
+                return False
 
         try:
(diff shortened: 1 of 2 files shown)


corehq/tabs/uitab.py
```diff
@@ -126,9 +126,14 @@ def should_show(self):
...
-        if not self.can_access_all_locations and not self.filtered_dropdown_items:
-            # location-safe filtering makes this whole tab inaccessible
-            return False
+        if not self.can_access_all_locations:
+            if self.dropdown_items and not self.filtered_dropdown_items:
+                # location-safe filtering makes this whole tab inaccessible
+                return False
+
+            # Just a button tab, determine if it's location safe
+            if not self.dropdown_items and not url_is_location_safe(self.url):
```
emord: I think the dashboard should be visible to mobile workers https://github.com/dimagi/commcare-hq/pull/13268

benrudolph: Yes, but only if they have access to reports. I tested on a completely unprivileged mobile worker which gets redirected to the LocationList page and then renders a 403: https://github.com/dimagi/commcare-hq/blob/master/corehq/apps/dashboard/views.py#L49

esoergel: The Orgs initiative is planning to tackle this at some point soon: https://trello.com/c/mIQiaisw/25-tb-ucr-making-dashboard-location-safe

esoergel (approved):

czue:
thanks

PR #11569, opened 2016-05-11
Mobile ucr translations

nickpell:
https://trello.com/c/YwYi8mSC/115-mobile-ucr-id-mapping-for-variables

Allows for specifying translation for values in mobile UCR rows using a transform in the report configuration.  I chose to store the translations here (instead of in the app) because we might want to do translations in web reports in the future and it would be good to have one source of truth.  Another reason was that if we decided to move the translations, it would be easier to go from report -> app than app -> report since each report app config has one report and not vice versa.

@NoahCarnahan @czue @orangejenny

PR #14183, opened 2016-12-09
Make the TranslationTransform translate also in web UI

esoergel:
@nickpell @czue

**UPDATE**
This PR implements the TranslationTransform for web, and adds support for the translation format we use elsewhere, that is, a lang->string dict or a simple string:
```python
In [1]: localize('1', 'en')
Out[1]: '1'

In [2]: localize({'en': 'one', 'es': 'dos'}, 'en')
Out[2]: 'one'
```
This maintains support for the old format in mobile reports, but documentation only reflects the new format (converting old stuff is left as an exercise for the reader).  Mobile reports will also now accept the new format, including flat mappings.

There's a `mobile_or_web` flag (defaulted to `'web'`) which is necessary for this to work on mobile, for ease of implementation.

Old description:

> I was looking to see if this existed and Nick pointed me at this transform.  The existing implementation is for mobile UCR, and the transform there happens on the device, not server-side.  Still, it looked pretty easy to implement for web, so I did this as a proof-of-concept.  It works well, and was pretty easy to write.

> In order to ready this to go live, we'd need to:

> * Reconcile the spec (I think that means migrating existing reports).  Nick's implementation uses a list of lists to store translation, and I used the translation semantics used for column headers (Nick mentioned offline that he'd been thinking about switching to that syntax).
> * Add documentation (easy)
> * Figure out how to make this work with both mobile UCR and web-based UCR.  The problem here is that `get_transform_function` should return a noop for mobile UCR, and return a proper translator for web UCR.  Is this something we've run into before?  I'm not sure how to cleanly fork the code like that.

> This was just a POC, so I don't want to go too far on this if it's not something we'd like to do anyways.  I'm leaning towards having two entirely separate transforms, one for mobile, one for the web.  This is pretty inelegant, but makes sense from an implementation perspective.

> How would you two recommend proceeding?

> [here's](https://github.com/dimagi/commcare-hq/blob/master/custom/icds_reports/ucr/reports/mpr_2ci_child_birth_list.json#L108) an existing mobile report using this transform, and [here's](https://github.com/dimagi/commcare-hq/pull/11569/files) the PR which introduced it (shows how it's implemented in the `suite.xml`)

### DIFF
--- a/corehq/apps/userreports/management/commands/migrate_translation_transforms.py
+++ b/corehq/apps/userreports/management/commands/migrate_translation_transforms.py
@@ -1,0 +1,58 @@
+from optparse import make_option
+from django.core.management.base import BaseCommand
+from corehq.apps.userreports.models import ReportConfiguration
+from corehq.dbaccessors.couchapps.all_docs import get_doc_ids_by_class
+from corehq.util.couch import iter_update, DocUpdate
+from corehq.util.log import with_progress_bar
+
+
+def reformat_translations(old_translations):
+    if not isinstance(old_translations, dict):
+        return old_translations
+    new_translations = {}
+    for k, translations in old_translations.items():
+        if isinstance(translations, basestring):
+            new_translations[k] = translations
+        else:
+            new_translations[k] = dict(translations)
+    return new_translations
+
+
+class Command(BaseCommand):
+    help = ("Migrate existing translation transforms to flag as mobile-only "
+            "and update mapping format")
+    option_list = (
+        make_option(
+            '--dry-run',
+            action='store_true',
+            dest='dry_run',
+            default=False,
+            help=("Don't actually perform the update, just list the reports "
+                  "that would be affected"),
+        ),
+    )
+
+    def handle(self, dry_run=False, *args, **options):
+        self.dry_run = dry_run
+        self.reports_using_transform = set()
+        report_ids = get_doc_ids_by_class(ReportConfiguration)
+        res = iter_update(
+            db=ReportConfiguration.get_db(),
+            fn=self.migrate_report,
+            ids=with_progress_bar(report_ids),
+            verbose=True,
+        )
+        print "Found {} reports using the transform:".format(len(self.reports_using_transform))
+        print "\n".join(self.reports_using_transform)
+        print "Updated the following reports:"
+        print "\n".join(res.updated_ids)
+
+    def migrate_report(self, report_config):
+        rc = ReportConfiguration.wrap(report_config)
+        for column in rc.report_columns:
+            if column.transform and column.transform['type'] == 'translation':
+                column.transform['mobile_or_web'] = 'mobile'
+                column.transform['translations'] = reformat_translations(column.transform['translations'])
+                self.reports_using_transform.add(rc._id)
+        if not self.dry_run:
+            return DocUpdate(rc.to_json())
--- a/corehq/apps/userreports/tests/test_transforms.py
+++ b/corehq/apps/userreports/tests/test_transforms.py
@@ -1,5 +1,7 @@
 from datetime import datetime, timedelta
 from django.test import SimpleTestCase
+from mock import patch
+
 from corehq.apps.userreports.exceptions import BadSpecError
 from corehq.apps.userreports.transforms.factory import TransformFactory
 from corehq.apps.userreports.transforms.specs import CustomTransform
@@ -83,3 +85,61 @@ class DaysElapsedTransformTest(SimpleTestCase):
         })
         date = (datetime.utcnow() - timedelta(days=5)).strftime('%Y-%m-%d')
         self.assertEqual(transform.transform(date), 5)
+
+
+
+class TranslationTransform(SimpleTestCase):
+
+    def test_missing_translation(self):
+        transform = TransformFactory.get_transform({
+            "type": "translation",
+            "translations": {},
+        }).get_transform_function()
+        self.assertEqual(transform('foo'), 'foo')
+
+    def test_basic_translation(self):
+        transform = TransformFactory.get_transform({
+            "type": "translation",
+            "translations": {
+                "#0000FF": "Blue"
+            },
+        }).get_transform_function()
+        self.assertEqual(transform('#0000FF'), 'Blue')
+        self.assertEqual(transform('#123456'), '#123456')
+
+    def test_default_language_translation(self):
+        transform = TransformFactory.get_transform({
+            "type": "translation",
+            "translations": {
+                "#0000FF": {
+                    "en": "Blue",
+                    "es": "Azul",
+                },
+                "#800080": {
+                    "en": "Purple",
+                    "es": "Morado",
+                }
+            },
+        }).get_transform_function()
+        self.assertEqual(transform('#0000FF'), 'Blue')
+        self.assertEqual(transform('#800080'), 'Purple')
+        self.assertEqual(transform('#123456'), '#123456')
+
+    @patch('corehq.apps.userreports.transforms.specs.get_language', lambda: "es")
+    def test_spanish_language_translation(self):
+        transform = TransformFactory.get_transform({
+            "type": "translation",
+            "translations": {
+                "#0000FF": {
+                    "en": "Blue",
+                    "es": "Azul",
+                },
+                "#800080": {
+                    "en": "Purple",
+                    "es": "Morado",
+                }
+            },
+        }).get_transform_function()
+        self.assertEqual(transform('#0000FF'), 'Azul')
+        self.assertEqual(transform('#800080'), 'Morado')
+        self.assertEqual(transform('#123456'), '#123456')
--- a/corehq/apps/userreports/tests/test_transforms.py
+++ b/corehq/apps/userreports/tests/test_transforms.py
@@ -143,3 +143,28 @@ class TranslationTransform(SimpleTestCase):
         self.assertEqual(transform('#0000FF'), 'Azul')
         self.assertEqual(transform('#800080'), 'Morado')
         self.assertEqual(transform('#123456'), '#123456')
+
+    def test_dont_translate_for_mobile(self):
+        transform = TransformFactory.get_transform({
+            "type": "translation",
+            "mobile_or_web": "mobile",
+            "translations": {
+                "#0000FF": "Blue",
+                "#800080": [["en", "Purple"], ["es", "Morado"]],  # legacy, mobile-only format
+            },
+        }).get_transform_function()
+        self.assertEqual(transform('#0000FF'), '#0000FF')
+        self.assertEqual(transform('#800080'), '#800080')
+        self.assertEqual(transform('#123456'), '#123456')
+
+    def test_bad_option(self):
+        with self.assertRaises(BadSpecError):
+            TransformFactory.get_transform({
+                "type": "translation",
+                "mobile_or_web": "neither!",
+                "translations": {
+                    "0": "zero",
+                    "1": {"en": "one", "es": "uno"},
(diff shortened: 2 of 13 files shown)


corehq/apps/userreports/tests/test_transforms.py
```diff
@@ -83,3 +85,61 @@ def test_string_transform(self):
         })
         date = (datetime.utcnow() - timedelta(days=5)).strftime('%Y-%m-%d')
         self.assertEqual(transform.transform(date), 5)
+
+
+
+class TranslationTransform(SimpleTestCase):
```
dimagimon: E303 too many blank lines (3)

czue:
I think this looks good/right to me.

not sure I totally understand the complexity of this issue:

> Figure out how to make this work with both mobile UCR and web-based UCR. The problem here is that `get_transform_function` should return a noop for mobile UCR, and return a proper translator for web UCR. Is this something we've run into before? I'm not sure how to cleanly fork the code like that.

would it not be possible to just do different logic depending on the context it is being called in (web vs mobile)?

not sure if the rest of this is helpful, but here's my quick thoughts on what I think we should strive for. i think to provide better feedback on more of the "how it should works" part I'd have to re-familiarize myself with the transform code and how it fits in with this. let me know if you want to do that (might be worth a quick 30 minute design discussion?).

What I think would be good if it were feasible:

- users can define their own translations
- you only have to define the translations in a single place and they work for both web and mobile
- you can change text on mobile if desired
- we reuse as much code as possible
- translation format is relatively consistent with what we do in other places (i.e. apps)
- we have a migration/upgrade path for legacy apps that doesn't break any existing content

esoergel:
> would it not be possible to just do different logic depending on the context it is being called in (web vs mobile)?

That's what I was imagining would be the hard part - figuring out the context.  Currently there's a function with no arguments which returns a transformer function.  Figuring out the context I assume would require passing some flag all the way through the call stack, which could be messy.

I just spent a good little while digging in to transforms to see if there was an easy way to access that bit of state.  It's of course possible, but it does look complex.  Additionally, while digging, I discovered that transforms don't take effect for es-backed reports (at least ReportConfig transforms).  The transform logic is hooked in to [here](https://github.com/dimagi/commcare-hq/blob/master/corehq/apps/reports/sqlreport.py#L255-L256) for SQL, but [the analagous method for ES](https://github.com/dimagi/commcare-hq/blob/master/corehq/apps/reports/api.py#L31-L31) doesn't implement it.

Do you have something more straightforward in mind?  I'd definitely be interested in talking it through if you've got the time.

> we have a migration/upgrade path for legacy apps that doesn't break any existing content
This shouldn't be hard.  Luckily it's easy enough to detect what format you're using - string, dict, or list.  We can just inspect the type of your values and act accordingly.

---------
In case it's useful, here's the stack trace of the `get_transform_fn` invocation:
```python
  File "/home/ethan/commcare-hq/corehq/apps/userreports/tests/test_report_data.py", line 201, in test_translations
    report_data = report_data_source.get_data()
  File "/home/ethan/commcare-hq/corehq/apps/userreports/reports/data_source.py", line 93, in get_data
    return self.data_source.get_data(start, limit)
  File "/home/ethan/commcare-hq/submodules/dimagi-utils-src/dimagi/utils/decorators/memoized.py", line 10, in _inner
    return m(*args, **kwargs)
  File "/home/ethan/commcare-hq/submodules/dimagi-utils-src/dimagi/utils/decorators/memoized.py", line 127, in __call__
  File "/home/ethan/virtualenvs/hq/local/lib/python2.7/site-packages/django/utils/decorators.py", line 34, in _wrapper
    return bound_func(*args, **kwargs)
  File "/home/ethan/commcare-hq/corehq/apps/userreports/decorators.py", line 28, in _inner
    return func(*args, **kwargs)
  File "/home/ethan/virtualenvs/hq/local/lib/python2.7/site-packages/django/utils/decorators.py", line 30, in bound_func
    return func.__get__(self, type(self))(*args2, **kwargs2)
  File "/home/ethan/commcare-hq/corehq/apps/userreports/sql/data_source.py", line 77, in get_data
    ret = super(ConfigurableReportSqlDataSource, self).get_data(start=start, limit=limit)
  File "/home/ethan/commcare-hq/corehq/apps/reports/sqlreport.py", line 255, in get_data
    formatter = DataFormatter(DictDataFormat(self.columns, no_value=None))
  File "/home/ethan/commcare-hq/corehq/apps/userreports/sql/data_source.py", line 66, in columns
    db_columns = [c for c in self.inner_columns if not isinstance(c, CalculatedColumn)]
  File "/home/ethan/commcare-hq/corehq/apps/userreports/mixins.py", line 64, in inner_columns
    for inner_col in col.get_column_config(self.config, self.lang).columns
  File "/home/ethan/commcare-hq/corehq/apps/userreports/reports/specs.py", line 176, in get_column_config
    format_fn=self.get_format_fn(),
  File "/home/ethan/commcare-hq/corehq/apps/userreports/reports/specs.py", line 119, in get_format_fn
    return TransformFactory.get_transform(self.transform).get_transform_function()
  File "/home/ethan/commcare-hq/corehq/apps/userreports/transforms/specs.py", line 115, in get_transform_function
    import traceback; traceback.print_stack()
```

czue:
thanks for the explanation - i didn't have that level of detail in mind and makes sense that is hard/annoying. Also, my guess is that though it probably wouldn't be too hard to pass the context all the way through that would be a pretty tech-debt-inducing change that doesn't seem great to add to the already hard-to-reason-with report generation code.

Do you know if there is any good reason why we do the transforms all the way down there instead of way up when the data is about to be returned back to the user? It almost seems like it would be cleaner if the data source always returned the raw data and then formatting was applied as a post-processing step - maybe somewhere [around here](https://github.com/dimagi/commcare-hq/blob/master/corehq/apps/userreports/sql/data_source.py#L79) (would need to make it backend agnostic). Then it's just the `ConfigurableReportDataSourceMixin.get_data` function that needs to know context, which seems totally fine.

czue:
one possible short term solution if that sounds like too much work would be to just add a "mobile/web" flag to the TranslationTransform spec? that means we can leverage the same code/config, though does not support the "same report for mobile and web" use case

corehq/apps/userreports/transforms/specs.py
```diff
@@ -1,6 +1,8 @@
 from decimal import Decimal
-from dimagi.ext.jsonobject import DictProperty, JsonObject, StringProperty
+from django.utils.translation import get_language
+from dimagi.ext.jsonobject import DictProperty, JsonObject, StringProperty, BooleanProperty
```
dimagimon: F401 'BooleanProperty' imported but unused

esoergel:
@nickpell @czue this should be ready for another pass.
@dannyroberts code buddy
Probably easiest :fish: :heavy_multiplication_x: :tropical_fish:

esoergel:
@czue regarding the backend issue, I wrote a failing test to make it easier to isolate.  Is there a UCR backlog or trello where I can put this?
https://github.com/dimagi/commcare-hq/commit/d4cd4e956b52cebce8e15b2edd73eafd2b1928a0

corehq/apps/userreports/transforms/specs.py
```diff
@@ -109,8 +109,11 @@ class TranslationTransform(Transform):
     """
     type = TypeProperty('translation')
     translations = DictProperty()
+    mobile_or_web = StringProperty(default="web", choices=["mobile", "web"])
```
czue: might be worth a comment explaining why this is necessary

czue: also, if we are making the default "web" (which makes sense to me) don't we need a data migration since all current users of this feature are using it for mobile?

czue (commented):
this looks great - thanks for docs and tests. one possible blocking question on the migration of existing reports

czue:
> @czue regarding the backend issue, I wrote a failing test (on a different branch) to make it easier to isolate. Is there a UCR backlog or trello where I can put this? d4cd4e9

I'm guessing @emord can just put this in whatever he's using to track UCR ES work.

corehq/apps/userreports/transforms/specs.py
esoergel: (offlined, I'll write a migration)

emord:
Yeah. It's been on my backlog to have those methods called in ES, but I didn't realize it meant transforms didn't work. Thanks for chekcing that out!

corehq/apps/userreports/management/commands/migrate_translation_transforms.py
```diff
@@ -0,0 +1,58 @@
...
+            "and update mapping format")
+    option_list = (
+        make_option(
+            '--dry-run',
+            action='store_true',
+            dest='dry_run',
+            default=False,
+            help=("Don't actually perform the update, just list the reports "
+                  "that would be affected"),
+    ),
```
dimagimon: E122 continuation line missing indentation or outdented

custom/icds_reports/ucr/reports/mpr_2bi_preg_delivery_death_list.json
```diff
@@ -109,61 +109,26 @@
         "format": "default",
         "transform": {
           "type": "translation",
+          "mobile_or_web": "mobile",
```
esoergel: @sheelio FYI, I'm reformatting the ICDS translation transforms to be in line with the new spec.  This shouldn't have any functional differences.

esoergel:
I just added a migration to set that flag on existing reports and switch the translations format.  I just did a check on prod and staging, and there's actually only one report which uses this transform - the one that I wrote which prompted this whole thing.  This PR updates the 7 static reports which use the transform.

czue:
> and there's actually only one report which uses this transform - the one that I wrote which prompted this whole thing.

oh man. sorry.

czue (approved):
```
ERROR: test suite for <corehq.tests.nose.HqdbContext object at 0x7fb162d378d0>
  File "/vendor/lib/python2.7/site-packages/nose/suite.py", line 209, in run
    self.setUp()
  File "/vendor/lib/python2.7/site-packages/nose/suite.py", line 296, in setUp
    self.setupContext(context)
  File "/vendor/lib/python2.7/site-packages/nose/suite.py", line 315, in setupContext
    try_run(context, names)
  File "/vendor/lib/python2.7/site-packages/nose/util.py", line 471, in try_run
    return func()
  File "/mnt/commcare-hq-ro/corehq/tests/nose.py", line 224, in setup
    super(HqdbContext, self).setup()
  File "/vendor/lib/python2.7/site-packages/django_nose/plugin.py", line 428, in setup
    self.old_names = self.runner.setup_databases()
  File "/vendor/lib/python2.7/site-packages/django/test/runner.py", line 166, in setup_databases
    **kwargs
  File "/vendor/lib/python2.7/site-packages/django/test/runner.py", line 370, in setup_databases
    serialize=connection.settings_dict.get("TEST", {}).get("SERIALIZE", True),
  File "/vendor/lib/python2.7/site-packages/django/db/backends/base/creation.py", line 368, in create_test_db
    test_flush=not keepdb,
  File "/vendor/lib/python2.7/site-packages/django/core/management/__init__.py", line 120, in call_command
    return command.execute(*args, **defaults)
  File "/vendor/lib/python2.7/site-packages/django/core/management/base.py", line 445, in execute
    output = self.handle(*args, **options)
  File "/vendor/lib/python2.7/site-packages/django/core/management/commands/migrate.py", line 105, in handle
    "'python manage.py makemigrations --merge'" % name_str
CommandError: Conflicting migrations detected (0001_initial, 0002_migrate_translation_transforms in userreports).
To fix them run 'python manage.py makemigrations --merge'
```

PR #14185, opened 2016-12-09
Mk 243262 migration for new location fixture

mkangia:
[Spec](https://docs.google.com/document/d/1KjGPZL4oHyFUeAajnyNYII5Un276_ofc8WiTbWkoaYM/edit#)

🏛 

command to run 
`./manage.py migrate_new_location_fixture`
for dry run use `./manage.py migrate_new_location_fixture --check`


mkangia:
@esoergel would add more checks for finding out applications using hierarchical fixture. But everything else should be in place. Updated Spec as well.

corehq/apps/locations/management/commands/migrate_new_location_fixture.py
```diff
@@ -0,0 +1,40 @@
...
+
+    def handle(self, *args, **options):
+        domains_having_locations = (
+            SQLLocation.objects.order_by('domain').distinct('domain')
+            .values_list('domain', flat=True)
+        )
+        domains_with_hierarchical_fixture = find_applications_with_hierarchical_fixture(
+            domains_having_locations
+        )
+        toggle = Toggle.get(HIERARCHICAL_LOCATION_FIXTURE.slug)
```
proteusvacuum: I think here you want to do `HIERARHICAL_LOCATION_FIXTURE.set(domain, enabled=True, namespace=NAMESPACE_DOMAIN)`, because that also updates the toggle cache

corehq/apps/locations/management/commands/migrate_new_location_fixture.py
```diff
@@ -0,0 +1,40 @@
...
+            toggle.add(domain, True, NAMESPACE_DOMAIN)
+
+
+def find_applications_with_hierarchical_fixture(domains):
+    search_string = 'commtrack:locations'
+    domain_with_application = {}
+    for domain in domains:
+        domain_obj = Domain.get_by_name(domain)
+        for application in domain_obj.applications():
+            raw_doc = json.dumps(application.get_db().get(application.id))
```
proteusvacuum: does this also find references in form xml?

corehq/tabs/tabclasses.py
```diff
@@ -1340,6 +1340,7 @@ def _get_feature_flag_items(domain):
             'title': _('Calendar Fixture'),
             'url': reverse(CalendarFixtureConfigView.urlname, args=[domain])
         })
+    # Check change for this
     if toggles.FLAT_LOCATION_FIXTURE.enabled(domain):
```
mkangia: oops! that was simple.

corehq/apps/locations/management/commands/migrate_new_location_fixture.py
mkangia: Not yet Farid. Waiting for some feedback for more areas to search. Included application doc only for now.

corehq/apps/locations/management/commands/migrate_new_location_fixture.py
mkangia: need to correct that.. i intended to use `toggle.set` only.

czue: you can also just append the (namespaced) domain to the list of enabled items and save it at the end if we don't want to save this document once-per domain. you can just look at what `set_toggle` does behind the scenes

corehq/apps/locations/management/commands/migrate_new_location_fixture.py
czue: I think the rendered complete set of source files might be the right thing to search in

corehq/apps/locations/management/commands/migrate_new_location_fixture.py
```diff
@@ -0,0 +1,40 @@
...
+        for domain in domains_with_hierarchical_fixture:
+            toggle.set(domain, True, NAMESPACE_DOMAIN)
+
+
+def find_applications_with_hierarchical_fixture(domains):
+    search_string = 'commtrack:locations'
+    domain_with_application = {}
+    for domain in domains:
+        domain_obj = Domain.get_by_name(domain)
+        for application in domain_obj.applications():
```
czue: is there a chance we think we need to search app builds too? maybe just released ones? failure mode I'm thinking of is if someone has removed references in the current app version but still has users using earlier app versions.

corehq/apps/locations/management/commands/migrate_new_location_fixture.py
```diff
@@ -0,0 +1,40 @@
...
+            raw_doc = json.dumps(application.get_db().get(application.id))
+            if search_string in raw_doc:
+                search_string[domain] = application.id
```
czue: I think this should be `domain_with_application` not `search_string`

czue (changes requested):
I think we need to be very careful with rolling this out. The failure mode here is that we unknowingly break a huge amount of people who are trying to use their applications. As-is if this PR is merged it will immediately do that.

I think we need to come up with a clear migration pathway that we can feel very confident is not going to break anyone, and I suspect we need to do that in multiple stages. I also think we will want a comprehensive test suite for the logic that checks for location references in applications so we can be sure to trust it.

It seems like it would be much easier to start by just making this the default for new domains, and converting everyone who is not using locations (no org hierarchy defined). Then focusing on figuring out what to do with the hard cases.

corehq/apps/locations/management/commands/migrate_new_location_fixture.py
mkangia: Ideally would do that. I am not completely sure but seems like the cache expiry would get tricky there since that needs to be done for each domain.

corehq/apps/locations/management/commands/migrate_new_location_fixture.py
mkangia: From the limited knowledge i have, in that case would this stand for all domains that migrate from old fixture format in an earlier version to flat fixture in new versions with users still using old version of app? May be because phone restore would still give new flat fixture format and app would be using old references?

czue: yeah exactly. so for example for TB we have eliminated hierarchical from our apps, but we have not turned off sync yet because one person is still using the old app

esoergel:
It's a bit confusing to understand how the `FixtureConfiguration` and the toggles fit together, what is the intended interaction?

I just had an idea - what if you instead worked to drop both toggles and use only the `FixtureConfiguration` to determine what to sync?  This would mean creating a `FixtureConfiguration` for all domains currently using locations (we can be conservative and set `hierarchical=True` at first), then going forward, if a domain doesn't have a `FixtureConfiguration`, we sync only the flat fixture, otherwise we do what the `FixtureConfiguration` says.  That way it's stored clearly in one place, and there's a migration path for existing domains.  (This is just me spitballing, so feel free to ignore if it doesn't sound easier)

czue:
@esoergel I like that idea a lot. The only issue is that for any app that is using both we do need to be explicit about which one we resolve in app manager. A couple thoughts:

1. we could add an option to the config and auto-set it safely during the migration
2. we could add as an option inside the app somewhere
3. we could try to detect it on build by inspecting your app contents (might be hard to implement and a performance hit - and we would have to figure out what to do if you referenced none or both of them in your app and likely need to fall back to 1 or 2 to resolve that...)

at a minimum I think making it explicit in the config object as opposed to mucking around with the feature flags will be much easier to reason with.

corehq/apps/locations/management/commands/migrate_new_location_fixture.py
```diff
@@ -0,0 +1,59 @@
...
+        domain_with_flat_fixture_enabled = find_domains_with_flat_fixture_enabled()
+
+        # 3. Find domains to stay on legacy fixture
+        domains_to_stay_on_hierarchical_fixture = (set(domains_having_locations) -
+                                                   set(domain_with_flat_fixture_enabled))
+
+        # 4. Update domains that need to stay on hierarchical with enabled legacy toggle to be able to access conf
+        # and update their location configuration to use hierarchical fixture for now
+        toggle = Toggle.get(HIERARCHICAL_LOCATION_FIXTURE.slug)
+        for domain in domains_to_stay_on_hierarchical_fixture :
```
dimagimon: E203 whitespace before ':'

mkangia:
@esoergel @czue have updated spec(Revised Changes V2) according to what was discussed last and also the PR accordingly. I have not tested it as of now. Once confirmed that the approach looks fine, will spend time testing it.
I have not looked in the idea above about dropping both toggles since that needs some more brainstorming.

czue:
@mkangia I think it's worth making the chosen format in apps configurable as a part of this work and dropping both toggles. I don't think it should be hard - just use the same logic you are using in the toggles but insert the configuration object in place of it, and set it the same way you are setting the toggles.

czue:
so to be clear I think the steps would be:

1. add a field "default format in apps" to `FixtureConfiguration` with allowed values ["flat", "hiearchical"] and default value "flat"
2. replace all toggle logic in apps to instead run off of this field
3. remove any toggle behavior from fixture generation code and rely only on the config object
4. update migration to set this field to "hieararchical" where you are currently setting the toggle.

@esoergel does that sound right?

mkangia:
@czue if we drop both toggles, it would make the configuration option open for all users and would allow domains on flat fixture to move back to hierarchical configurations whereas we need the opposite to happen?

czue:
good point - we should also hide the config UI if the app type is flat and only flat is turned on (which should also be the default for all new domains)

mkangia:
@czue that sounds good. Just so i understand this correctly,
You are suggesting to keep the 2 fields `sync_flat_fixture` and `sync_hierarchical_fixture` as well along with `default format in apps`?

czue:
> You are suggesting to keep the 2 fields `sync_flat_fixture` and `sync_hierarchical_fixture` as well along with `default format in apps`? 

yep exactly

esoergel:
This all sounds great to me, I'm on the same page.

esoergel:
There are some build failures, my guess is it's mostly just existing tests that need to be updated to reflect the new fixture.

corehq/ex-submodules/casexml/apps/phone/tests/test_ota_fixtures.py
```diff
@@ -13,6 +13,7 @@
 from corehq.apps.users.dbaccessors.all_commcare_users import delete_all_users
 from casexml.apps.case.tests.util import check_xml_line_by_line
 from corehq.form_processor.tests.utils import run_with_all_backends
+from corehq.util.test_utils import flag_enabled
```
dimagimon: F401 'flag_enabled' imported but unused

mkangia:
@esoergel I had a quick sync up with @czue and we concluded to stay on the approach that PR is following. I am having a tough time resolving the build failures. Needs a better knowledge of fixtures to fix them. Moreover, some build failures might highlight genuine concerns. Can you help with test cases? i have fixed  some of them but got stuck at some tricky ones now.

mkangia:
Thank You @esoergel. :1st_place_medal:

esoergel (commented):
I just did another review and it looks good to me.  Was there anything else you wanted to do before wrapping this up?  (The "don't merge" flag is still on).  @czue did you have any remaining followup?

mkangia:
thanks @esoergel One last thing would be to ensure the places where `commtrack:location` exists in the code. I would assume that this notation should be used only for migration and other wise the new notation should be used everywhere.
As of now its mostly in test cases and just one place in code at https://github.com/dimagi/commcare-hq/blob/mk-243262-new-location-fixture/corehq/apps/app_manager/detail_screen.py#L614 which would be good check. Relying on your better understanding of fixtures. Let me know if any code/test-case i should specifically check still using the old notation.

esoergel:
That looks like location xpath validation in the app builder, which I don't think I've ever seen in action.  @czue it looks like you wrote that back in 2014, any idea if that still works or where/how I'd be able to try it out?  [This](https://github.com/dimagi/commcare-hq/blob/3ccacef392e3c087ad4eab7c3a85179882d6538e/corehq/apps/app_manager/models.py#L2115) makes me think it's a case list thing, but I'm not sure how to get the option to appear.

czue:
@esoergel @mkangia that is what powers [this functionality](https://confluence.dimagi.com/display/commtrack/Referencing+Locations+in+Applications). I think we can just get rid of that functionality if you're on the new fixture format. we probably need to support it for legacy apps though.

czue (approved):

czue:
I haven't done a thorough review since last time, but I'm on board with the plan and signed off as long as you both are happy. Would be good to address / understand how this affects the behavior of that case list reference code before merging though.

mkangia:
Thank You @esoergel @czue.
Will go ahead and try to do the same check for case list reference as well then.

@proteusvacuum Just checking if you did you come across this case list reference when working/moving to flat fixture locations.

czue:
> @proteusvacuum Just checking if you did you come across this case list reference when working/moving to flat fixture locations.

there is no way it is currently working for flat fixture so i think we can conclude no one is using it.

esoergel:
@mkangia I just tried this out and saw that it does indeed still exist and work.  The feature is pretty hidden though.  Go to a module on an app that uses cases (on a domain with locations).  Then add a new property, and type in something like `location:state:district/name` (where `state` and `district` are the codes of two location types on that domain).  When I do that, I get the following error message on save:

> Case List has invalid location reference state:county/name. Details: Reference type county cannot be a child of primary type state.

When I flip the order to `location:district:state/name`, it saves successfully.  If you try it with names that aren't real location types, it will give a different error.

Anyways, I think all we need to do for this migration is turn off that validation if the domain uses the flat fixture.  If I type gibberish in there it pretty much just lets it slide.

czue:
thanks @esoergel. sorry i should have mentioned that @mkangia and i offlined about this.

> Anyways, I think all we need to do for this migration is turn off that validation if the domain uses the flat fixture.

I think we should actually just disallow it to use the "location:" syntax at all.

corehq/apps/app_manager/models.py
```diff
@@ -2096,11 +2096,17 @@ def rename_lang(self, old_lang, new_lang):
...
+
         hierarchy = None
         for column in columns:
             if column.field_type == FIELD_TYPE_LOCATION:
-                hierarchy = hierarchy or parent_child(self.get_app().domain)
+                domain = self.get_app().domain
+                project = Domain.get_by_name(domain)
                 try:
+                    if not should_sync_hierarchical_fixture(project):
+                        raise LocationXpathValidationError(_('Location references not supported in Case List.'))
```
mkangia: @czue This would take care of discontinuing the feature to use location references in Case List. Any suggestions for better error message? Any alternative that could be suggested?

czue: Hmm. Maybe "That format is no longer supported. To reference the location hierarchy you need to use the "Custom Calculations in Case List	" feature preview. For more information see: https://confluence.dimagi.com/pages/viewpage.action?pageId=38276915". cc @sheelio

sheelio: I think that message+link is fine.

esoergel:
Build failure was unrelated, just kicked it.

corehq/apps/locations/tests/test_location_fixtures.py
```diff
@@ -82,6 +85,8 @@ def setUp(self):
         delete_all_users()
         self.user = create_restore_user(self.domain, 'user', '123')
 
+
+    @flag_enabled('HIERARCHICAL_LOCATION_FIXTURE')
```
dimagimon: E303 too many blank lines (2)

esoergel (approved):

corehq/apps/locations/management/commands/migrate_new_location_fixture.py
```diff
@@ -0,0 +1,65 @@
...
+        for domain in domain_with_flat_fixture_enabled:
+            # For domains with flat fixture enabled and have configuration defined to use hierarchical fixture
+            # Notify them to make their state clear
+            location_conf_for_domain = LocationFixtureConfiguration.for_domain(domain)
+            if not location_conf_for_domain._state.adding and location_conf_for_domain.sync_hierarchical_fixture:
+                # Can use the following to update them
+                # location_conf_for_domain.sync_flat_fixture = True
+                # location_conf_for_domain.sync_hierarchical_fixture = False
+                # location_conf_for_domain.save()
+                print "Domain that needs attention since its not in a definite state: %s" % domain
```
esoergel: Don't these domains also need the `HIERARCHICAL_LOCATION_FIXTURE` toggle enabled?  I agree that it makes sense to look at them individually, but I imagine in every case we'll want to be able to access the FixtureConfiguration UI and let them keep getting the hierarchical fixture.

mkangia: @esoergel @czue I would expect no domains to be in this area. These would be domains with flat fixture enabled but with configuration set to use hierarchical which contradicts. So this is just a sanity check.
We can always add them to the flag if needed but like i said there should not be any ideally.

corehq/apps/locations/management/commands/migrate_new_location_fixture.py
```diff
@@ -33,10 +46,26 @@ def handle(self, *args, **options):
...
+                HIERARCHICAL_LOCATION_FIXTURE.set(domain, True, NAMESPACE_DOMAIN)
+            else:
+                print "HIERARCHICAL_LOCATION_FIXTURE toggle would be set for domain: %s" % domain
+
+            if not dry_run:
+                print "Enabling legacy fixture for domain: %s" % domain
+                location_configuration = LocationFixtureConfiguration.for_domain(domain)
+                print "Current Configuration, Persisted: %s, Use Flat fixture: %s, Use Hierarchical fixture %s" % (
+                    not location_configuration._state.adding, location_configuration.sync_hierarchical_fixture,
+                    location_configuration.sync_flat_fixture)
```
esoergel: Won't the `location_configuration` always be missing for these domains?  They're 
```python
domains_to_stay_on_hierarchical_fixture = (set(domains_having_locations) -
                                           set(domain_with_flat_fixture_enabled))
```
So if the flat fixture toggle isn't enabled, they shouldn't have a fixture config, right?

mkangia: Yup! Thats right. Just logging for confirmation.

mkangia:
@esoergel lets do this today!

PR #14186, opened 2016-12-09
Call function in domain stats

benrudolph:
@orangejenny i believe introduced here https://github.com/dimagi/commcare-hq/commit/17248ad3#diff-e8c29e4d83c4b54de5ea540784b53448L197

cc: @emord


emord (approved):

orangejenny:
Ugh. Thank you for finding & fixing so quickly.

czue:
thanks. I think we can also fix the tests to catch this.

czue:
https://github.com/dimagi/commcare-hq/pull/14188

PR #14188, opened 2016-12-10
update test to check serializability of output

czue:
@benrudolph @orangejenny

https://github.com/dimagi/commcare-hq/pull/14186#issuecomment-266188786


benrudolph (approved):
👍 sweet

PR #14190, opened 2016-12-10
Remove lenient protocol

benrudolph:
@czue @wpride removes the ability to use `as=<user>@<domain>`. Requires full username. tested old cloudcare, new cloudcare, and edit forms. i think this simplifies many things

cc: @emord

### DIFF
--- a/corehq/apps/ota/utils.py
+++ b/corehq/apps/ota/utils.py
@@ -6,7 +6,7 @@ from corehq.apps.domain.models import Domain
 from corehq.apps.users.util import format_username
 from dimagi.utils.logging import notify_exception
 
-from corehq.apps.users.models import CommCareUser, WebUser, CouchUser
+from corehq.apps.users.models import CouchUser
 
 from dimagi.utils.web import json_response
 from corehq.apps.domain.auth import get_username_and_password_from_request, determine_authtype_from_request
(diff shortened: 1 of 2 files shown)


corehq/apps/ota/utils.py
```diff
@@ -6,7 +6,7 @@
 from corehq.apps.users.util import format_username
 from dimagi.utils.logging import notify_exception
 
-from corehq.apps.users.models import CommCareUser, WebUser
+from corehq.apps.users.models import CommCareUser, WebUser, CouchUser
```
dimagimon: F401 'CommCareUser' imported but unused
F401 'WebUser' imported but unused

czue (approved):
nice this is way simpler

czue:
for some reason feeling a little gunshy about merging (just feels like the type of thing that might have unexpected side effects and we don't have great integration tests). if you're confident in your process I think go for it though.

benrudolph:
also a bit tentative myself. this has already bitten me loads of times, each time though i've added a test case that had bitten me. i'll do another round of QA on staging and then report back

emord (approved):

benrudolph:
Ok tested the possible end points that i know of locally and on staging and haven't seen any errors. pulling the trigger. will monitor it after deploy

PR #14191, opened 2016-12-11
Fix new report builder drag-and-drop issue

NoahCarnahan:
(note this is a PR into `nh/new_rb`, not `master`)
@kaapstorm I found the source of the drag-and-drop issue, although I still can't quite explain it. The drag and drop works as expected if we remove the `rateLimit` property from the observable array. Don't know why that makes it work, but I think the rate limit is unnecessary. Did you add rateLimit for a particular reason?

The second commit just uses an existing drag and drop binding (that resides in `style/ko/knockout_bindings.ko.js`) instead of copying the one from `detail-screen-config.js`.

FYI buddies @esoergel @czue

### DIFF
--- a/corehq/apps/userreports/static/userreports/js/report_config.js
+++ b/corehq/apps/userreports/static/userreports/js/report_config.js
@@ -82,7 +82,6 @@ var reportBuilder = function () {
             self.refreshPreview(newValue);
             self.saveButton.fire('change');
         });
-        self.selectedColumns.extend({ rateLimit: 50 });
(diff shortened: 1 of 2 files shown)


czue (approved):
looks good to me but will let @kaapstorm confirm this seems right

kaapstorm:
This is great. Thank you @NoahCarnahan . 

I can't remember where `rateLimit` came from, or what it was for, so it's good that it's gone.

PR #14193, opened 2016-12-12
Fix missing rows in report builder preview

NoahCarnahan:
(this is *not* a PR into master)

@kaapstorm looks like the case/form list report previews were always showing just one row. This is due to the not-obvious behavior of UCR that requires an aggregation by `["doc_id"]`, not `[]` to see one row per form/case.
If you were seeing missing rows in other circumstances as well, let me know and I'll see if I can track down the cause.


kaapstorm:
:+1: Nice! No, just the list report previews.

PR #14195, opened 2016-12-12
Project description tweaks

benrudolph:
@emord this fixes up a few things with the report an issue spec. hold off on merge while i get sign off from robin
screen shot 2016-12-12 at 8 34 24 am

### DIFF
--- a/corehq/apps/style/templates/style/includes/modal_report_issue.html
+++ b/corehq/apps/style/templates/style/includes/modal_report_issue.html
@@ -115,6 +115,11 @@
                             <div class="col-sm-9">
                                 <input type="file" id="bug-report-file" name="report_issue"/>
                                 <p class="help-block">
+                                    {% blocktrans %}
+                                    Please upload any files relevant to your issue.
+                                    {% endblocktrans %}
+                                </p>
+                                <p class="help-block">
                                     <i class="fa fa-warning"></i>
                                     {% blocktrans %}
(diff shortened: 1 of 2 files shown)


corehq/apps/style/templates/style/includes/modal_report_issue.html
```diff
@@ -81,7 +81,11 @@ <h4 class="modal-title">{% trans "Report an Issue with CommCare HQ" %}</h4>
                                           class="form-control"
                                           id="bug-report-project-description"
                                           rows="3"
-                                          placeholder="{% trans "Please provide a short description of your project." %}">{{ request.project.project_description }}</textarea>
+                                          placeholder="{% trans 'Please provide a short description of your organization and the work it is doing.' %}">{% spaceless %}
+                                          {% if request.project.project_description %}
```
emord: What's the `spaceless` for? django docs say that it's for removing space in between tags, but I don't think there should be any tags in the project description right?

benrudolph: I believe it also removes newlines within in a tag; I was having issues getting the placeholder text to appear when there were multiline statements and I didn't want to put it all on one line. I can double check and verify

benrudolph: (double checked and spaceless strips out all newlines and is necessary for placeholder to work correct)

PR #14196, opened 2016-12-12
Make search and claim endpoints location safe

proteusvacuum:
Adds location safe decorator to search and claim endpoints. 

There is already messaging when you turn this feature on that you will have access to all the cases in that domain, so I figured we're pretty safe for now.

### DIFF
--- a/corehq/apps/ota/views.py
+++ b/corehq/apps/ota/views.py
@@ -54,6 +54,7 @@ def restore(request, domain, app_id=None):
     return response
 
 
+@location_safe
 @json_error
 @login_or_digest_or_basic_or_apikey()
 @check_domain_migration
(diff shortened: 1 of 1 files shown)


esoergel (approved):
Sounds reasonable to me.  Have @sheelio and/or @dimagi/product greenlighted this?

(This change makes case search and case claim accessible to location-restricted users.  Those features allow the user to break out of the sandbox of their location and access data from other parts of the project, but that's kinda the point.)

czue:
Since @sheelio and @amsagoff are both offline I'm going to say yes. This makes sense as a logical extension to the location restriction feature to me and I think very few projects are using this currently.

proteusvacuum:
Yeah, there are only 2 projects with case claim currently. The first one are using their own location restrictions, and the other one wants to have access to all cases regardless of location, so I think this is the right behaviour currently.

benrudolph:
niice fully functional web apps

esoergel:
:+1:

PR #14200, opened 2016-12-12
Stronger email checks for analytics

benrudolph:
@snopoke @emord went back and forth on the best way to do this, but finally settled on this. http://manage.dimagi.com/default.asp?242172 for more context. simon, i didn't use `notify_error` cause i thought it'd be too noisy and not that actionable. let me know if you think i should still use that

### DIFF
--- a/corehq/apps/analytics/tasks.py
+++ b/corehq/apps/analytics/tasks.py
@@ -18,6 +18,7 @@ import logging
 
 from django.conf import settings
 from django.core.urlresolvers import reverse
+from email_validator import validate_email, EmailNotValidError
 from corehq.toggles import deterministic_random
 from corehq.util.decorators import analytics_task
 from corehq.util.soft_assert import soft_assert
@@ -368,7 +369,7 @@ def track_periodic_data():
     submit = []
     for user in users_to_domains:
         email = user.get('email')
-        if not email:
+        if not _email_is_valid(email):
             continue
 
         number_of_users += 1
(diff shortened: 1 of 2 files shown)


corehq/apps/analytics/tasks.py
```diff
@@ -368,7 +369,8 @@ def track_periodic_data():
     submit = []
     for user in users_to_domains:
         email = user.get('email')
-        if not email:
+        if not email or not validate_email(email, verify=True):
```
benrudolph: One thing to note about `validate_email` is that it returns `None` if it's unsure. so this is being extra stringent. We can change this so it only skips when we now it's not valid

benrudolph: there's also this email lib that checks the domain name: https://github.com/JoshData/python-email-validator. overall it seems like a better library. i might go with that

emord:
Curious why you think this will be too noisy? I assume that having an invalid email after registering would not be that common,

benrudolph:
we save the user account even if the email account hasn't been validated yet

corehq/apps/analytics/tasks.py
snopoke: We may also want to make this configurable for the sake of NIC (which may not have good internet access to do the validation).

emord: This isn't being run on nic though since it's submitting to hubspot right?

benrudolph: i think it still gets run (unless there is something excluding it from being picked up as a task), but you're right i doubt anything works so it's probably not a big deal

requirements/requirements.txt
```diff
@@ -95,3 +95,5 @@ github3.py==1.0.0a4
 django-bulk-update==1.1.10
 csiphash==0.0.4
 json-delta==2.0
+email_validator==1.0.1
```
nickpell: @benrudolph Any reason you went with this over the built in django validator? https://docs.djangoproject.com/en/1.8/ref/validators/#validate-email

benrudolph: cause it checks if the email is deliverable (important for hubspot): https://github.com/JoshData/python-email-validator#overview. it's easy to have a fake email that passes the regex but isn't an actual account

nickpell: ooooh very cool

PR #14201, opened 2016-12-12
Remove paranoid selector

millerdev:
Makes `UserCaseXPath` consistent with user case xpath expressions generated by vellum. I don't think this will have any adverse side effects, but I'd appreciate some review on that assertion. For example, [`xpath_references_case`](https://github.com/dimagi/commcare-hq/blob/master/corehq/apps/app_manager/util.py#L98) uses the user case xpath in a `str.replace()` operation, which I think should still work. I think that's the only place where an expression generated by `UserCaseXPath` is used to match other expressions.

@czue @kaapstorm @snopoke @nickpell

### DIFF
--- a/corehq/apps/app_manager/tests/test_suite_form_filter_errors.py
+++ b/corehq/apps/app_manager/tests/test_suite_form_filter_errors.py
@@ -42,7 +42,7 @@ class FormFilterErrorTests(SimpleTestCase, TestXmlMixin):
             <text>
               <locale id="modules.m0"/>
             </text>
-            <command id="m0-f0" relevant="instance('casedb')/casedb/case[@case_type='commcare-user'][hq_user_id=instance('commcaresession')/session/context/userid][1]/due_date &lt;= today()"/>
+            <command id="m0-f0" relevant="instance('casedb')/casedb/case[@case_type='commcare-user'][hq_user_id=instance('commcaresession')/session/context/userid]/due_date &lt;= today()"/>
           </menu>
         </partial>
         """
--- a/corehq/apps/app_manager/tests/data/case_list_form/case-list-form-advanced-autoload.xml
+++ b/corehq/apps/app_manager/tests/data/case_list_form/case-list-form-advanced-autoload.xml
@@ -89,7 +89,7 @@
       <stack>
         <push>
           <command value="'m0-f0'"/>
-          <datum id="case_id_usercase" value="instance('casedb')/casedb/case[@case_type='commcare-user'][hq_user_id=instance('commcaresession')/session/context/userid][1]/@case_id"/>
+          <datum id="case_id_usercase" value="instance('casedb')/casedb/case[@case_type='commcare-user'][hq_user_id=instance('commcaresession')/session/context/userid]/@case_id"/>
           <datum id="case_id_new_dugong_0" value="uuid()"/>
           <datum id="return_to" value="'m1'"/>
         </push>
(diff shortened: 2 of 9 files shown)


nickpell:
test failure:

```
FAIL: test_advanced_suite_auto_select_usercase (corehq.apps.app_manager.tests.test_advanced_suite.AdvancedSuiteTest)
  File "/mnt/commcare-hq-ro/corehq/apps/app_manager/tests/test_advanced_suite.py", line 110, in test_advanced_suite_auto_select_usercase
    './entry[2]')
  File "/mnt/commcare-hq-ro/corehq/apps/app_manager/tests/util.py", line 26, in assertXmlPartialEqual
    self.assertXmlEqual(expected, actual, normalize=False)
  File "/mnt/commcare-hq-ro/corehq/apps/app_manager/tests/util.py", line 32, in assertXmlEqual
    _check_shared(expected, actual, LXMLOutputChecker(), "xml")
  File "/mnt/commcare-hq-ro/corehq/apps/app_manager/tests/util.py", line 126, in _check_shared
    raise AssertionError(message)
AssertionError: XML mismatch
--- want.xml
+++ got.xml
@@ -9,10 +9,10 @@
     <instance id="casedb" src="jr://instance/casedb"/>
     <instance id="commcaresession" src="jr://instance/session"/>
     <session>
-      <datum function="instance('casedb')/casedb/case[@case_type='commcare-user'][hq_user_id=instance('commcaresession')/session/context/userid][1]/@case_id" id="case_id_case_clinic"/>
+      <datum function="instance('casedb')/casedb/case[@case_type='commcare-user'][hq_user_id=instance('commcaresession')/session/context/userid]/@case_id" id="case_id_case_clinic"/>
     </session>
     <assertions>
-      <assert test="count(instance('casedb')/casedb/case[@case_type='commcare-user'][hq_user_id=instance('commcaresession')/session/context/userid][1]) = 1">
+      <assert test="count(instance('casedb')/casedb/case[@case_type='commcare-user'][hq_user_id=instance('commcaresession')/session/context/userid]) = 1">
         <text>
           <locale id="case_autoload.usercase.case_missing"/>
         </text>
```

czue:
> I don't think this will have any adverse side effects, but I'd appreciate some review on that assertion.

would it be a helpful exercise to write up the possible adverse side effects this could have and then reason why they won't be an issue? just not sure i have enough context to definitively say this and would find that write up helpful if we want to convince ourselves this is safe

snopoke:
One option to make this safer is to add an assertion to the entries where this get's used which checks that there is only one user case (in the same way we do for checking that the user only belongs to one group for case sharing.

kaapstorm:
I think dropping `[1]` is the right move, because if a user were to have two user cases and that fact was hidden, then an app could slowly break things more and more, and cleaning it up would be a real pain. (This is not a hypothetical scenario. :upside_down_face: (Don't create duplicate user cases with case import.)) I like @snopoke's idea of an assertion because it allows the scenario to be dealt with in a managed way.

millerdev:
Possible side effects:
- Anywhere where a user case xpath generated by `UserCaseXPath` is used in a string comparison will now possibly have a different outcome, depending on the comparison. The only instance I could find of this was [`xpath_references_case`](https://github.com/dimagi/commcare-hq/blob/master/corehq/apps/app_manager/util.py#L98), and I think the outcome will not change there.
- If there are multiple records for a given user case record then this will now cause all of them to be selected instead of just the first one. This scenario is broken anyway, so failing early is better.

@snopoke isn't it true that most (if not all) of the places where the xpath expression generated by this code are evaluated will be on a phone? Are you proposing to add the assertion in mobile code? If not, can you elaborate on where we'd add such an assertion?

snopoke:
Yes, adding an assertion in the suite file:
https://github.com/dimagi/commcare-core/wiki/Suite20#entry

millerdev:
@snopoke is it possible [there's already an assertion on that](https://github.com/dimagi/commcare-hq/blob/dm/rm-paranoid-selector/corehq/apps/app_manager/suite_xml/sections/entries.py#L449)? I'm not very familiar with suite file generation, and it looks moderately complicated, so any help you can provide about how to add this assertion would be greatly appreciated.

corehq/apps/app_manager/tests/test_suite_form_filter_errors.py
```diff
@@ -42,7 +42,7 @@ def test_no_error_when_user_case(self):
             <text>
               <locale id="modules.m0"/>
             </text>
-            <command id="m0-f0" relevant="instance('casedb')/casedb/case[@case_type='commcare-user'][hq_user_id=instance('commcaresession')/session/context/userid][1]/due_date &lt;= today()"/>
+            <command id="m0-f0" relevant="instance('casedb')/casedb/case[@case_type='commcare-user'][hq_user_id=instance('commcaresession')/session/context/userid]/due_date &lt;= today()"/>
```
dimagimon: E501 line too long (189 > 115 characters)

millerdev:
All tests passed, but travis failed on coveralls due to job exceeded maximum time limit. Kicked.

millerdev:
✅

kaapstorm (approved):

PR #14203, opened 2016-12-12
BlobDB interface changes/cleanup

dannyroberts:
@millerdev is this pretty much what you were thinking when we talked this morning? @snopoke fyi. I think maybe most important thing is if you have a sense of whether I found all uses of `AbstractBlobDB.put`, because if there are any that PyCharm's Find Usages missed that pass in basename, then they'll fail. (And `put` is kind of a hard thing to just grep for, unfortunately!)

### DIFF
--- a/corehq/blobs/atomic.py
+++ b/corehq/blobs/atomic.py
@@ -22,7 +22,7 @@ class AtomicBlobs(object):
         self.puts = None
         self.deletes = None
 
-    def put(self, content, basename="", bucket=DEFAULT_BUCKET):
+    def put(self, content, bucket=DEFAULT_BUCKET):
         if self.puts is None:
             raise InvalidContext("AtomicBlobs context is not active")
         info = self.db.put(content, bucket=bucket)
--- a/corehq/blobs/atomic.py
+++ b/corehq/blobs/atomic.py
@@ -25,7 +25,7 @@ class AtomicBlobs(object):
     def put(self, content, basename="", bucket=DEFAULT_BUCKET):
         if self.puts is None:
             raise InvalidContext("AtomicBlobs context is not active")
-        info = self.db.put(content, basename, bucket)
+        info = self.db.put(content, bucket=bucket)
         self.puts.append((info, bucket))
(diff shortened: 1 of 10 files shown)


dannyroberts:
@czue buddy

czue (approved):
code looks good but i'd feel more comfortable if @millerdev or @snopoke also did a pass since I don't have much context

corehq/blobs/atomic.py
```diff
@@ -25,7 +25,7 @@ def __init__(self, db):
     def put(self, content, basename="", bucket=DEFAULT_BUCKET):
```
millerdev: I think `basename` should be removed here as well.

corehq/blobs/interface.py
```diff
@@ -18,7 +18,7 @@ class AbstractBlobDB(object):
     __metaclass__ = ABCMeta
 
     @abstractmethod
-    def put(self, content, basename="", bucket=DEFAULT_BUCKET):
+    def put(self, content, bucket=DEFAULT_BUCKET, identifier=None):
```
millerdev: The docstring for this method needs to be updated.

corehq/blobs/atomic.py
dannyroberts: Good call

corehq/blobs/interface.py
dannyroberts: Good catch, will fix

millerdev (approved):

dannyroberts:
@millerdev what's your feeling here; is there anything we need to test manually that wouldn't be caught by the travis build?

millerdev:
Don't think so. I think g2g when you're happy with it. FWIW, I checked out this branch locally and did a quick search for `.put(` and reviewed all occurrences, most of which are blob db puts. Nothing jumped out at me as needing attention other than the things you already addressed.

dannyroberts:
Awesome, I'm happy with it then so I'll go ahead and merge it!

PR #14204, opened 2016-12-12
Delete models after test

emord:
These kept failing when running app manager tests back to back

@benrudolph


benrudolph (approved):

emord:
Tests passed. one job exceeded time submitting to coveralls so not going to restart

PR #14206, opened 2016-12-12
Add locations to filter in Submit History

mkangia:
This adds capability to filter for locations and their users in SubmitHistory and CaseList. The locations were already rendered but are not filtered by.

Also added locations filtering for Reassign Cases and Manage Forms, moved out to PR #14382 

Review better at once 🏛 
http://manage.dimagi.com/default.asp?243862


corehq/apps/reports/filters/users.py
```diff
@@ -176,7 +176,7 @@ class ExpandedMobileWorkerFilter(BaseMultipleOptionFilter):
     placeholder = ugettext_lazy(
         "Specify groups and users to include in the report")
     is_cacheable = False
-    options_url = 'emwf_options'
```
esoergel: I think this means `Locations` will start showing up in the dropdown for all reports, right?

mkangia: Yup! This does not belong here. Reverted in following commits

dimagimon:
There are 5 errors:

* corehq/apps/reports/filters/users.py, line 350 - E122 continuation line missing indentation or outdented
* corehq/apps/reports/filters/users.py, line 351 - E122 continuation line missing indentation or outdented
* corehq/apps/reports/filters/users.py, line 352 - E122 continuation line missing indentation or outdented
* corehq/apps/reports/filters/users.py, line 353 - E122 continuation line missing indentation or outdented
* corehq/apps/reports/filters/users.py, line 354 - E122 continuation line missing indentation or outdented

dimagimon:
There are 5 errors:

* corehq/apps/reports/filters/users.py, line 350 - E122 continuation line missing indentation or outdented
* corehq/apps/reports/filters/users.py, line 351 - E122 continuation line missing indentation or outdented
* corehq/apps/reports/filters/users.py, line 352 - E122 continuation line missing indentation or outdented
* corehq/apps/reports/filters/users.py, line 353 - E122 continuation line missing indentation or outdented
* corehq/apps/reports/filters/users.py, line 354 - E122 continuation line missing indentation or outdented

mkangia:
@esoergel Its great that Submit History and Case List could be made location compatible with just minor changes. Good to review.

corehq/apps/locations/permissions.py
```diff
@@ -99,7 +99,7 @@
 
 LOCATION_SAFE_TASTYPIE_RESOURCES = set()
 
-LOCATION_SAFE_HQ_REPORTS = set()
+LOCATION_SAFE_HQ_REPORTS = set(['case_list'])
```
esoergel: I just tried it out locally, you can just add `@location_safe` to the class right at the top: https://github.com/dimagi/commcare-hq/blob/master/corehq/apps/reports/standard/cases/basic.py#L185

corehq/apps/reports/filters/users.py
```diff
@@ -342,9 +344,17 @@ def pull_users_and_groups(cls, domain, mobile_user_and_group_slugs,
...
                 simplified=True
             )
+        users_at_locations = []
+        if location_ids:
+            users_at_locations = util.get_all_users_by_domain(
+                domain=domain,
+                user_ids=user_ids_at_locations_and_descendants(location_ids),
+                simplified=True,
+                CommCareUser=CommCareUser,
+            )
```
esoergel: Oh nice!  This makes a lot of sense, as it will always be empty for reports that don't use locations

corehq/apps/reports/standard/inspect.py
```diff
@@ -92,16 +94,27 @@ def _form_filter(form):
...
-                .filter(time_filter(gte=self.datespan.startdate,
-                                    lt=self.datespan.enddate_adjusted))
-                .filter(self._get_users_filter(mobile_user_and_group_slugs)))
+                 .domain(self.domain)
+                 .filter(time_filter(gte=self.datespan.startdate,
+                                     lt=self.datespan.enddate_adjusted))
+                 .filter(self._get_users_filter(mobile_user_and_group_slugs)))
+
+        if not self.request.can_access_all_locations:
+            query = query.filter(self.scope_filter())
```
esoergel: :+1:

corehq/apps/reports/filters/users.py
mkangia: Also in case of reports with locations, this returned all users in case no location was selected which was weird. Will check that out later.

corehq/apps/locations/permissions.py
mkangia: Ah! Thank you Mr. Hunt. I think i was using location safe over some other class.
This works.

corehq/apps/reports/standard/cases/basic.py
```diff
@@ -114,42 +134,53 @@ def case_owners(self):
...
+            user_types = EMWF.selected_user_types(mobile_user_and_group_slugs)
+            special_owner_ids = self.get_special_owner_ids(
+                admin=HQUserType.ADMIN in user_types,
+                unknown=HQUserType.UNKNOWN in user_types,
+                demo=HQUserType.DEMO_USER in user_types,
+                commtrack=HQUserType.COMMTRACK in user_types,
+            )
+
+
+            # Get group ids for each group that was specified
```
dimagimon: E303 too many blank lines (2)

corehq/apps/reports/standard/cases/basic.py
```diff
@@ -114,42 +134,53 @@ def case_owners(self):
...
+            selected_reporting_group_ids = EMWF.selected_reporting_group_ids(mobile_user_and_group_slugs)
+            selected_sharing_group_ids = EMWF.selected_sharing_group_ids(mobile_user_and_group_slugs)
+
+            report_group_q = HQESQuery(index="groups").domain(self.domain)\
+                                               .doc_type("Group")\
```
dimagimon: E127 continuation line over-indented for visual indent

corehq/apps/reports/standard/cases/basic.py
```diff
@@ -134,45 +134,53 @@ def case_owners(self):
...
 
         # Get user ids for each user that was specifically selected
         selected_user_ids = EMWF.selected_user_ids(mobile_user_and_group_slugs)
 
-        # Get group ids for each group that was specified
-        selected_reporting_group_ids = EMWF.selected_reporting_group_ids(mobile_user_and_group_slugs)
-        selected_sharing_group_ids = EMWF.selected_sharing_group_ids(mobile_user_and_group_slugs)
+        # Get ids for each sharing group that contains a user from selected_reporting_group_users OR
+        # a user that was specifically selected
+        if selected_reporting_group_users and selected_user_ids:
```
esoergel: This should be `or`

esoergel (changes requested):

corehq/apps/reports/standard/cases/basic.py
```diff
@@ -70,7 +85,10 @@ def _build_query(self):
             )
             query = query.NOT(case_es.owner(ids_to_exclude))
         else:  # Only show explicit matches
-            query = query.owner(self.case_owners)
+            selected_case_owners = self.case_owners
+            if selected_case_owners:
+                query = query.owner(selected_case_owners)
+            query = query.OR(self.scope_filter())
```
mkangia: oh yes.. should not be applied to restricted users. will fix that.

corehq/apps/reports/standard/cases/basic.py
mkangia: good catch! Thanks.

mkangia:
@esoergel extracted out changes for Case Reassignment and Manage forms sections in #14382 since those could block rolling out changes for SubmitHistory and CaseList.

corehq/apps/reports/standard/inspect.py
```diff
@@ -140,6 +155,7 @@ def total_records(self):
         return int(self.es_query_result.total)
 
 
+@location_safe
```
esoergel: This all looks good to me!

corehq/apps/reports/standard/cases/basic.py
```diff
@@ -24,6 +26,129 @@
 from .data_sources import CaseInfo, CaseDisplay
 
 
+class CaseExportScope(object):
```
esoergel: This is an awful lot of copy-pasted code - this looks like it would be confusing to update without context.  I'm guessing this is pulled out so that the descendants of `CaseListMixin` don't inherit this behavior automatically.  If that's the case, I don't think that's something we need to worry about.

The behavior only changes for restricted users, right?  In that case it won't matter for regular users whether this code is here or not.  Restricted users won't be able to access descendants of `CaseListMixin` anyways unless those reports are marked as location-safe, in which case they'd want this behavior.

Would it be possible to just include all this in the `CaseListMixin`?

mkangia: This does change in logic and is not an exact over write. This extract is also used later for any place that needs to add this scope for example in Case Reassignment. There are a lot of classes that inherit from CaseListMixin. I am not sure about adding changes to all those classes by updating CaseListMixin, hence extracted this out. This extract can be used for any places we make locations restricted.

esoergel: What are the changes in logic?  The problem with doing it like this is that it's really difficult to tell how the behavior changes between for instance, the Case List Report and the Reassign Cases Report (since they're supposed to be the same).  Is there another way to fork the logic on `can_access_all_locations`?  It's safe to assume that you can change any behavior when `can_access_all_locations == False`, as long as the behavior is consistent when it's `True`.

mkangia: Right.. that sounds so right.. 
The changes are mostly for restricted user access like not allowing filtering by user types, groups etc or adding a scope filter for restricted user. For a user that can access all locations, i think all remains the same plus the locations integration is added.

corehq/apps/reports/filters/users.py
benrudolph: just noting that we should probably respect the `limit_user_ids` variable here. in the application status report, it limits the user ids to those who have submitted to the application selected. this would ignore that and include all user ids

mkangia: Thank you @benrudolph.
Made corresponding changes.

corehq/apps/reports/standard/cases/basic.py
```diff
@@ -111,45 +130,75 @@ def get_special_owner_ids(self, admin, unknown, demo, commtrack):
     @property
     @memoized
     def case_owners(self):
+        """
```
mkangia: @esoergel @sheelio Added this documentation to bring to notice what case owners does and would do. Mostly is from the existing implementation but just want to flag couple of things.

> for selected reporting group ids, returns user_ids belonging to these groups
       also finds the sharing groups which has any user from the above reporting group

The second half does not sound right since that would involve all submission by a sharing group which the user just belongs to but has not submitted the cases, hence returning more cases

> selected user ids
       also finds the sharing groups which has any user from the above reporting group

Same as above.

> ids and descendants ids of selected locations and assigned locations of users selected

Similar issue with finding locations of selected users returning a larger set of cases.

mkangia:
@esoergel All changes moved to CaseListMixin instead of extracting CaseExportScope. Thanks for pointing this out and forcing this down.

corehq/apps/reports/standard/cases/basic.py
esoergel: I'll let @sheelio confirm, but I believe that's expected.  If you select a user or group of users (or location, which is in many ways a group of users), this report should show all cases that those users see on their phones.

corehq/apps/reports/standard/cases/basic.py
```diff
@@ -68,14 +70,31 @@ def _build_query(self):
             )
             query = query.NOT(case_es.owner(ids_to_exclude))
         else:  # Only show explicit matches
-            query = query.owner(self.case_owners)
+            selected_case_owners = self.case_owners
+            if selected_case_owners:
+                query = query.owner(selected_case_owners)
```
esoergel: When you deselect everything, the current behavior is to not return any results.  This changes it to return all results, unfiltered (except by scope, where applicable).  I don't think it matters a lot which one we do @sheelio do you have any thoughts?

One possible point of confusion is that if you select something like "Unknown Users", and there are no unknown users on the project, `case_owners` will return `[]`, and then the report will confusingly render with all cases.  I could see users thinking that means that their data is messed up.

corehq/apps/reports/standard/cases/basic.py
```diff
@@ -111,45 +130,75 @@ def get_special_owner_ids(self, admin, unknown, demo, commtrack):
...
 
         # Get user ids for each user that was specifically selected
         selected_user_ids = EMWF.selected_user_ids(mobile_user_and_group_slugs)
 
-        # Get group ids for each group that was specified
-        selected_reporting_group_ids = EMWF.selected_reporting_group_ids(mobile_user_and_group_slugs)
-        selected_sharing_group_ids = EMWF.selected_sharing_group_ids(mobile_user_and_group_slugs)
+        # Get ids for each sharing group that contains a user from selected_reporting_group_users OR
+        # a user that was specifically selected
+        if selected_reporting_group_users or selected_user_ids:
```
esoergel: Good call.  This'll save us from unnecessary queries in many instances.

esoergel (commented):
Thank you for following up on my feedback, sorry for making this drag for so long.  I had just a one more point, which should be quick to address

corehq/apps/reports/standard/cases/basic.py
mkangia: Oh thats not good. Thanks for catching this. Will revert to old behaviour of not returning any result.

sheelio:
Agreed that we should not return any results if the chosen filter is empty.

esoergel (approved):

esoergel:
Great, this should be good to go, right?  There's a merge conflict, then you can merge if you think it's ready.

benrudolph:
the conflict is most likely with the PR i've made here: https://github.com/dimagi/commcare-hq/pull/14397

may want to go with the changes i made since i've made `limit_user_ids` work with groups as well as locations (up to you on how you want to merge though)

mkangia:
Thank You @benrudolph. Kept your changes and added mine over them.

esoergel:
I approve of the conflict resolution and changes.  The build passed and it's mergeable.  I'm holding off on that because the "Don't Merge" flag is still on.  @mkangia feel free to merge if it's ready.

mkangia:
Sorry for not merging this yet @esoergel . Moving this to QA along with Manage forms and Reassign Cases.

corehq/apps/reports/standard/cases/basic.py
```diff
@@ -160,13 +223,16 @@ def case_owners(self):
                                                  selected_user_ids))
                                  .get_ids())
 
+        selected_location_ids = EMWF.selected_location_ids(mobile_user_and_group_slugs)
```
dimagimon: F841 local variable 'selected_location_ids' is assigned to but never used

mkangia:
@esoergel Tested this. Good to go now. 
Already merged in the changes for Reassign Cases and Manage Forms in this.

Would be nice to just give one final look, quite some merge conflicts resolved.

corehq/apps/groups/models.py
```diff
@@ -222,6 +222,13 @@ def by_user(cls, user_or_user_id, wrap=True, include_names=False):
             return [r['id'] for r in results]
 
     @classmethod
+    def get_case_sharing_accessible_locations(cls, domain, user):
+        return [
+            location.case_sharing_group_object() for location in
+            SQLLocation.objects.accessible_to_user(domain, user).filter(location_type__shares_cases=True)
+        ]
```
esoergel: You could also call this method below in `get_case_sharing_groups` to avoid duplication

esoergel: Oh, never mind, you don't have access to the user in that method.  Probably too much of a pain to change.

corehq/apps/data_interfaces/tasks.py
```diff
@@ -43,39 +43,14 @@ def bulk_archive_forms(domain, couch_user, uploaded_data):
 
 
 @task
-def bulk_form_management_async(archive_or_restore, domain, couch_user, form_ids_or_query_string):
-    # form_ids_or_query_string - can either be list of formids or a BulkFormMangementInterface query url
```
esoergel: I'm really happy to move this out of here.

corehq/apps/data_interfaces/views.py
```diff
@@ -509,11 +522,44 @@ def post(self, request, *args, **kwargs):
...
         )
 
+    def inaccessible_forms_accessed(self, xform_ids, domain, couch_user):
+        xforms = FormAccessors(domain).get_forms(xform_ids)
+        xforms_user_ids = set([xform.user_id for xform in xforms])
+        accessible_user_ids = set(user_ids_at_accessible_locations(domain, couch_user))
+        return xforms_user_ids - accessible_user_ids
+
+    def get_xform_ids(self, request):
+        form_ids_or_query_string = self.get_form_ids_or_query_string(request)
```
esoergel: I just looked up the definition for this:
```python
    def get_form_ids_or_query_string(self, request):
        if 'select_all' in self.request.POST:
            # Altough evaluating form_ids and sending to task would be cleaner,
            # heavier calls should be in in an async task instead
            import urllib
            form_query_string = urllib.unquote(self.request.POST.get('select_all'))
            return form_query_string
        else:
            return self.request.POST.getlist('xform_ids')
```
It forks depending on parameters, then returns one value, forcing you to check the type of the return again here.  It might be more straightforward to just merge `get_form_ids_or_query_string` in to this method and only do one if/else block (so you can drop the `if type()==list` and `isinstance(basestring)`)

esoergel (approved):
I had one or two stylistic comments, but nothing blocking.

czue:
:tada:

sheelio:
Woo hoo!

mkangia:
Exciting!!

PR #14207, opened 2016-12-12
use getScript for clipboard.js

emord:
@benrudolph http://manage.dimagi.com/default.asp?243789#1260203


benrudolph:
@emord added one more commit that removes jquery-ui from touchforms inline (jquery ui used to be used for the datetime picker for cloudcare, but we have since replaced it: https://github.com/dimagi/commcare-hq/pull/13197/files#diff-6fc361ae37a1c17971509c0fe05158f5L26). that should fix the tooltip conflict on the case details page

PR #14211, opened 2016-12-13
Different environments for web apps and preview

benrudolph:
@biyeun @wpride this ensures that display settings only get saved to the environment they're in. namely OQPS is only set for app preview not both web apps and app preview

cc: @emord


emord:
What is an environment?

benrudolph:
it's like the difference between app preview and web apps. it's the same code, but one is run in phone and the other is run in the full screen

biyeun:
hell ya!

PR #14212, opened 2016-12-13
Bring back migrate_app_to_cmitfb.py

orangejenny:
Will need to fix app in http://manage.dimagi.com/default.asp?243879

No changes to script, just reverting.

@emord


emord:
You could also do a `fab production setup_release --set code_branch=some_branch` if you don't want to go through the revert PR workflow

benrudolph:
confirmed no tests

PR #14213, opened 2016-12-13
More noticeable loading bar

benrudolph:
@proteusvacuum @biyeun offlined with B about this a bit. we decided for now to just make the loading bar thicker and a different color. it's now mean and green 💚 
screen shot 2016-12-13 at 12 10 59 pm
http://manage.dimagi.com/default.asp?243629

cc: @emord

### DIFF
--- a/corehq/apps/style/static/cloudcare/css/webforms.css
+++ b/corehq/apps/style/static/cloudcare/css/webforms.css
@@ -15,7 +15,8 @@
 
 /* Override default nprogress bar color to match CommCareHQ */
 #nprogress .bar {
-  background: #004ebc;
+  background: #47b700;
+  height: 4px;
 }
 
 #nprogress .peg {


proteusvacuum:
cool, thanks! 
for things that take a long time (e.g. loading an app) where the screen goes grey or where your action is blocked, is it worth just putting a loading modal up like you have for restores? I don't think this little bar at the top of the screen is a common UI paradigm on most sites, so is easy to miss.

benrudolph:
(i'll let @biyeun weigh in, she had some justifications for keeping it to the top. do agree that the gray screen is regrettable)

biyeun:
@benrudolph @proteusvacuum what i have seen are the three horizontal dots blinking...now that I think about it...

benrudolph:
we came up with an idea, but i think it'll take a bit longer to implement. how do you guys feel about merging this for now and then following up with a more comprehensive loading icon?

PR #14218, opened 2016-12-13
better data dictionary generation

emord:
Improvements on generating the data dictionary. Now includes the current app's case properties and marks a data dictionary as "fully generated" if it's been generated from the export. That will be eventually removed once all domains are converted, but some non-"fully generated" domains currently have case properties from the inferred property task

buddy @proteusvacuum :fish:

### DIFF
--- a/corehq/apps/data_dictionary/tests/test_util.py
+++ b/corehq/apps/data_dictionary/tests/test_util.py
@@ -76,3 +76,11 @@ class GenerateDictionaryTest(TestCase):
 
         self.assertEqual(CaseType.objects.filter(domain=self.domain).count(), 1)
         self.assertEqual(CaseProperty.objects.filter(case_type__domain=self.domain).count(), 1)
+
+    def test_parent_property(self, mock):
+        mock.return_value = {'type': ['property', 'parent/property']}
+        with self.assertNumQueries(4):
+            generate_data_dictionary(self.domain)
+
+        self.assertEqual(CaseType.objects.filter(domain=self.domain).count(), 1)
+        self.assertEqual(CaseProperty.objects.filter(case_type__domain=self.domain).count(), 1)
--- a/corehq/apps/data_dictionary/tests/test_util.py
+++ b/corehq/apps/data_dictionary/tests/test_util.py
@@ -24,7 +24,7 @@ class GenerateDictionaryTest(TestCase):
 
     def test_empty_type(self, mock):
         mock.return_value = {'': ['prop']}
-        with self.assertNumQueries(1):
+        with self.assertNumQueries(2):
             generate_data_dictionary(self.domain)
 
         self.assertEqual(CaseType.objects.filter(domain=self.domain).count(), 0)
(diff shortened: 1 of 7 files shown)


corehq/apps/data_dictionary/util.py
```diff
@@ -1,13 +1,17 @@
+from corehq import toggles
 from corehq.apps.app_manager.dbaccessors import get_case_types_from_apps
 from corehq.apps.app_manager.util import all_case_properties_by_domain
 from corehq.apps.data_dictionary.models import CaseProperty, CaseType
 from corehq.apps.export.models.new import CaseExportDataSchema
 
 
 def generate_data_dictionary(domain):
+    if toggles.OLD_EXPORTS.enabled(domain):
```
emord: @benrudolph I added this check, but from our last conversation it may not actually be necessary right?

corehq/apps/data_dictionary/util.py
```diff
@@ -62,16 +55,32 @@ def _get_current_case_types_and_properties(domain):
...
-        old_properties = set()
-    new_properties = []
-
-    for case_property in properties:
-        if case_property not in old_properties:
-            old_properties.add(case_property)
-            new_properties.append(CaseProperty(name=case_property, case_type=case_type_obj))
-
-    CaseProperty.objects.bulk_create(new_properties)
+    if not properties:
```
proteusvacuum: This seems like a strange way of writing this, since the caller isn't expecting this function to return anything. It might be clearer if you do
```py
if properties:
    _create_properties_for_case_types(...)
```

corehq/apps/data_dictionary/util.py
```diff
@@ -62,16 +55,32 @@ def _get_current_case_types_and_properties(domain):
...
+        if not case_type:
+            continue
+
+        if case_type in current_case_types:
+            case_type_obj = current_case_types[case_type]
+        else:
+            case_type_obj = CaseType.objects.create(domain=domain, name=case_type)
+
+        for prop in props:
+            if '/' in prop:
```
proteusvacuum: is this here because parent / host properties are referenced with `/` (e.g. `index/parent/case_prop`)? If so, might be good to either write a comment, or do something like `property_references_indexed_case = '/' in prop`

proteusvacuum (approved):
A couple of very nit-picky comments. But looks good!

corehq/apps/data_dictionary/tests/test_util.py
```diff
@@ -71,7 +71,15 @@ def test_already_existing_property(self, mock):
         self.assertEqual(CaseType.objects.filter(domain=self.domain).count(), 1)
         self.assertEqual(CaseProperty.objects.filter(case_type__domain=self.domain).count(), 1)
 
-        with self.assertNumQueries(2):
+        with self.assertNumQueries(3):
```
emord: It's really useful for code that is prone to n+1 queries. definitely caught myself accidentally adding sql queries that the tests caught

proteusvacuum (approved):

corehq/apps/data_dictionary/util.py
```diff
@@ -62,16 +59,32 @@ def _get_current_case_types_and_properties(domain):
...
+            continue
+
+        try:
+            case_type_obj = current_case_types[case_type]
+        except KeyError:
+            case_type_obj = CaseType.objects.create(domain=domain, name=case_type)
+
+        for prop in props:
+            # don't add any properites to parent cases
+            if '/' in prop:
```
emord: They get skipped for now. The same thing happens right now if a case property is only referenced through it's child. 

It's possible that we can figure out the hierarchy with the ParentCasePropertyBuilder but we would also need to add the property to each case type that could be a parent of that case, which isn't a great solution

PR #14219, opened 2016-12-13
fix format

emord:
@esoergel how do you mark a translation as wrong. I think that this one was translated incorrectly.

Search https://raw.githubusercontent.com/dimagi/commcare-hq/je/fix-format/locale/es/LC_MESSAGES/django.po for "corehq/apps/accounting/subscription_changes.py:637" The translator left out `number_of_reports`

buddy @proteusvacuum


esoergel:
You can check the translation on transifex: https://www.transifex.com/dimagi/commcare-hq/translate/#es/djangopo/86607457?q=number_of_reports
There you can modify or delete the translation.  In an instance like this, I'd probably talk to whoever's leading that translation effort (Michelle for Spanish, or (I think) Mohini or MikeO for French) so they can speak with the translator.  I already reached out to Michelle for this one.

PR #14221, opened 2016-12-13
Add user properties to data sources schema

millerdev:
- Rename *User Case Management* to *User Properties*. Tried to change it everywhere, but the case properties template was problematic. The labels have changed to be more generic in the [v2 template](https://github.com/dimagi/commcare-hq/blob/9cc8741c6d88558212f23eb7c07e7fc60ea36585/corehq/apps/app_manager/templates/app_manager/v2/partials/case_config_shared.html#L126), which looks great. [v1](https://github.com/dimagi/commcare-hq/blob/9cc8741c6d88558212f23eb7c07e7fc60ea36585/corehq/apps/app_manager/templates/app_manager/v1/partials/case_config_shared.html#L126) still says "case property" and that does not appear to be easy to change since the template is used in multiple contexts. It didn't look easy to pass parameters into the template, and the parameter passing looked like it was going to be messy because of translations and pluralization. I'm hoping v2 templates on the horizon are good enough to meet the requirements of this spec.
- Add user properties to data sources schema when user case is enabled.

Spec: https://docs.google.com/document/d/1kPVwZketyWl6aRo1ZNCM3EngkrTsgmFbnBXAv-FThKU/edit

@benrudolph @emord


emord:
I think that this will have loading User Properties in both the old case management UI and the form. Will there be more work to have a migration (possibly lazy to start with)?

How does this interact with the Case Summary? Right now vellum sends what references it's using to HQ by removing the front of the hashtag https://github.com/dimagi/Vellum/blob/master/src/logic.js#L465-L491 As it is I think this makes it look like user properties are actually case properties

corehq/apps/app_manager/util.py
```diff
@@ -420,7 +420,11 @@ def _add_ancestors(ctype, generation):
             "name": "{} ({})".format(generation_names[i], " or ".join(ctypes)) if i > 0 else base_case_type,
             "key": "@case_type",
             "structure": {p: {} for type in [map[t] for t in ctypes] for p in type},
-            "related": {"parent": generation_names[i + 1]} if i < len(generations) - 1 else None,
+            "related": {"parent": {
+                "hashtag": "#case/" + generation_names[i + 1],
+                "subset": generation_names[i + 1],
+                "key": "@case_id",
+            }} if i < len(generations) - 1 else None,
```
benrudolph: 👍

benrudolph (approved):

benrudolph:
lgtm, also curious how it interacts with old case mgmt

```
    self.partial_v2_dir
AssertionError: 
```

millerdev:
> I think that this will have loading User Properties in both the old case management UI and the form. Will there be more work to have a migration (possibly lazy to start with)?

That's a good question, and I didn't see a clear answer in the spec, although now that I'm looking closer I notice that the screenshot in the spec does not show the *Load the following case properties...* table on the *User Properties* tab. @dimagi/product, can you clarify please?

- should the *Load the following case properties...* table be removed from the *User Properties* tab?
- do we want to do a migration for existing apps similar to what we did for the easy references rollout?

> How does this interact with the Case Summary? Right now vellum sends what references it's using to HQ by removing the front of the hashtag...

It's true, vellum does slice off the front of the hashtag, which I had noted in a TODO comment just above that. We should change it. Is it going to be a problem for the case summary if vellum returns the full hashtag there?

czue:
(just flagging that most of @dimagi/product is offline so if this is time-sensitive we should try to get consensus from some combination of maybe @krsdimagi @ctsims @orangejenny and the rest of the initiative team)

millerdev:
@czue Yeah, I think fine if we don't get answers until January. I'm offline after this week too. Was thinking I'd get a UAT request started on this to test user property easy references in vellum, and knowing that we'll need to get these questions answered before final rollout.

corehq/apps/app_manager/util.py
```diff
@@ -409,46 +409,83 @@ def _add_ancestors(ctype, generation):
...
+        "id": generation_names[i],
+        "name": "{} ({})".format(generation_names[i], " or ".join(ctypes)) if i > 0 else base_case_type,
+        "key": "@case_type",
+        "structure": {p: {} for type in [map[t] for t in ctypes] for p in type},
+        "related": {"parent": {
+            "hashtag": "#case/" + generation_names[i + 1],
+            "subset": generation_names[i + 1],
+            "key": "@case_id",
+        }} if i < len(generations) - 1 else None,
+    } for i, ctypes in enumerate(generations)]
```
snopoke: This seems a bit confusing. If a case has multiple parent types which one get's selected? We don't even support that (at least not with the 'parent' relationship). Also it seems like the properties from ALL parent case types get lumped together.

corehq/apps/app_manager/util.py
```diff
@@ -409,46 +409,83 @@ def _add_ancestors(ctype, generation):
     else:
         generations = []
 
+    subsets = [{
+        "id": generation_names[i],
+        "name": "{} ({})".format(generation_names[i], " or ".join(ctypes)) if i > 0 else base_case_type,
+        "key": "@case_type",
```
snopoke: What's the significance of this. We don't use this in the parent case xpath expression.

ctsims:
Hey @millerdev and @czue .

To maintain consistency with the other hash/bubble/loading setup I'm happy to make the call that we should follow the same path we did there

- Remove the screen
- Migrate loads directly into the default value field

@emord is the code  you used to do that original migration still capable of being used to migrate the user case properties as well, or did that depend on there not being bubbles in the output XForm already and can't be used a second time?

corehq/apps/app_manager/util.py
```diff
@@ -409,46 +409,83 @@ def _add_ancestors(ctype, generation):
...
+            "merge": True,
+            "structure": {
+                "userid": {
+                    "reference": {
+                        "hashtag": "#user",
+                        "source": "casedb",
+                        "subset": USERCASE_TYPE,
+                        "subset_key": "@case_type",
+                        "subset_filter": True,
+                        "key": "hq_user_id",
```
snopoke: how does it know that it must filter with the user's ID from the session?

corehq/apps/app_manager/views/formdesigner.py
```diff
@@ -169,49 +170,8 @@ def get_form_data_schema(request, domain, form_unique_id):
...
-            ...
-        ]
-    }
-    ```
-    A structure may contain nested structure elements. A nested element
-    may contain one of "structure" (a concrete structure definition) or
-    "reference" (a link to some other structure definition). Any
-    structure item may have a human readable "name".
+    definition is a dictionary. For details on the content of the dictionary,
+    see https://github.com/dimagi/Vellum/blob/master/src/datasources.js
```
snopoke: I think I've made this comment before but I still don't understand how this works after reading these docs.

emord:
> is the code you used to do that original migration still capable of being used to migrate the user case properties as well, 

I think that usercase should just be a setvalue correct? The code I/ @orangejenny  wrote is https://github.com/dimagi/commcare-hq/blob/97dee89298830a884cc1cbcc04d09692206b8be2/corehq/apps/app_manager/management/commands/migrate_app_to_cmitfb.py It will need to be changed as part of it is specific to the data structure we used for case preloads in old case management, but it should be pretty similar. 

There's probably a way you can filter apps for the ones that use user case management which will hopefully make the migration much less painful

>  Is it going to be a problem for the case summary if vellum returns the full hashtag there?

@millerdev I believe that in the case summary it would show up as #user/property under the case that the form modifies. The way we store case references doesn't try to do anything particularly intelligently and I left it as is for vellum's case managment. As an example parent/grandparent references show up as parent/property or grandparent/property, which is similar to how they were before vellum case management

corehq/apps/app_manager/util.py
```diff
@@ -409,46 +409,83 @@ def _add_ancestors(ctype, generation):
...
+        "key": "@case_type",
+        "structure": {p: {} for type in [map[t] for t in ctypes] for p in type},
+        "related": {"parent": {
+            "hashtag": "#case/" + generation_names[i + 1],
+            "subset": generation_names[i + 1],
+            "key": "@case_id",
+        }} if i < len(generations) - 1 else None,
+    } for i, ctypes in enumerate(generations)]
+
+    if is_usercase_in_use(app.domain):
```
emord: If we want this rebranding to happen sooner rather than later, I think this is good to merge if you add a feature flag to enable user case in the form builder. That way QA can look at this on prod and this PR won't grow and grow.

Would also be nice to be able to get rid of the legacy structure out of vellum

corehq/apps/app_manager/util.py
emord: This will show both parents if we can't determine which is which. It's confusing, but was the best we came up with with previous iteratoin of case management. I think it looks like `parent (mother, household)` in the form builder tree

corehq/apps/app_manager/util.py
millerdev: It's supposed to be the property that the `id` references. However, I see that's not true anymore because of the way `parent` and `grandparent` case references are now constructed. It could theoretically be removed since it's not currently used by Vellum for these (`#case`) subsets, although it's not hurting anything either.

snopoke:
Except that it's confusing to have it there.

corehq/apps/app_manager/util.py
millerdev: The path of this `reference` is `instance('commcaresession')/session/context/userid`, and `hq_user_id` is the name of the referenced element. So this will end up creating a predicate like

`[hq_user_id = instance('commcaresession')/session/context/userid]`

The path is constructed by concatenating the instance for this source (`"id": "commcaresession"`, so `instance('commcaresession')`) with it's path (`/session`) plus the slash-delimited names of parent elements (`context`) plus the reference's element name (`userid`). Each element in the hierarchy points to a dict with keys describing it such as `structure` (sub elements), `reference` (this element references something), etc.

The thing that is referenced is the `CASE_TYPE` (`commcare-user`) subset in the `casedb` source, and since `subset_filter` is specified, an additional predicate will be added to the reference:

`[@case_type = 'commcare-user']`

corehq/apps/app_manager/views/formdesigner.py
millerdev: It's on my todo list to improve this documentation. Since vellum is currently the only consumer I'm planning to do it there.

millerdev:
>>> How does this interact with the Case Summary? Right now vellum sends what references it's using to HQ by removing the front of the hashtag...
>>
>> Is it going to be a problem for the case summary if vellum returns the full hashtag there?
>
> @millerdev I believe that in the case summary it would show up as #user/property under the case that the form modifies. The way we store case references doesn't try to do anything particularly intelligently and I left it as is for vellum's case managment. As an example parent/grandparent references show up as parent/property or grandparent/property, which is similar to how they were before vellum case management

I think this means "Yes, sending the full hashtag will break the HQ case summary." Additionally, it sounds like the case summary in HQ currently does not provide a way for vellum to send anything other than standard (not user) case references, and those are currently sent in a format that is ambiguous once the user case is in the mix. I am envisioning sending all hashtags in their full, unambiguous form: `#case/prop`, `#case/parent/prop`, `#user/prop`, etc. It sounds like the HQ case summary needs to be changed to support that. Does that all sound correct? If not, let's go to voice.

millerdev:
I think I addressed all the outstanding issues:

- Added feature flag.
- Added support for hashtags in case summary (form case references).
- Removed *Load the following case properties...* table be removed from the *User Properties* tab when the feature flag is enabled.
- Updated CMitFB management command to support user case with `--usercase` flag.
- Fixed superfluous key in schema.
- Fixed tests.

corehq/apps/app_manager/tests/test_util.py
```diff
@@ -89,6 +90,7 @@ def _add_scheduler_to_form(self, form, module, form_abreviation):
         phase.add_form(form)
 
 
+@flag_enabled('USER_PROPERTY_EASY_REFS')
```
emord: Can you add a test that makes sure user cases don't get sent down if this isn't set

corehq/apps/app_manager/management/commands/migrate_app_to_cmitfb.py
```diff
@@ -21,9 +21,15 @@ class Command(BaseCommand):
         Pass either a domain name (to migrate all apps in the domain) or an
         individual app id. Will skip any apps that have already been migrated.
     '''
+    option_list = (
+        make_option('--usercase',
```
emord: I think this style of argument is deprecated in django(not sure what version) you'll want to use `add_arguments` https://docs.djangoproject.com/en/1.10/howto/custom-management-commands/

corehq/apps/app_manager/management/commands/migrate_app_to_cmitfb.py
```diff
@@ -39,23 +46,54 @@ def handle(self, *args, **options):
...
+        refs = {path: [hashtag + case_property]
+                for path, case_property in kwargs["preloads"].iteritems()}
+        if form.case_references:
+            form.case_references["load"].update(refs)
+        else:
+            form.case_references = {"load": refs}
+    save_xform(app, form, ET.tostring(xform.xml))
+
+
+def should_migrate_usercase(app, migrate_usercase):
```
emord: The feature flag is domain level, but this migration can occur on one app at a time. I think that means that if you run this command on an app whose domain does not have the feature flag then their user case properties will show up in the form builder, but they can also add more in the case management screen. This will likely only affect QA at the start, but I could also seeing it happen to internal projects or FM projects.

For the original migration we added a property on the app to describe if it was migrated or not, because we did a staged rollout. That's probably not necessary here and you could force migrating entire domains at once, but I don't think that this will work well as is

corehq/apps/app_manager/management/commands/migrate_app_to_cmitfb.py
```diff
@@ -1,61 +1,116 @@
 import logging
 from lxml import etree as ET
+from optparse import make_option
```
dimagimon: F401 'make_option' imported but unused

corehq/apps/app_manager/management/commands/migrate_app_to_cmitfb.py
millerdev: 759eadfc55162abe20669d0f46934acd6dde0a54

corehq/apps/app_manager/management/commands/migrate_app_to_cmitfb.py
millerdev: 553148e7a74e4a9352c1130dded29ba423f576ab

emord: @orangejenny should look at this management command as well

millerdev:
This can be merged since the changes are behind a feature flag.

millerdev:
Just kidding... Realized that there are a bunch of copy changes in templates that are not behind the flag, and I'm not going to have time to change that today. Better QA this on staging.

corehq/apps/app_manager/models.py
```diff
@@ -5628,9 +5641,14 @@ def app_uses_usercase(app):
         if app_uses_usercase(self) and not domain_has_privilege(self.domain, privileges.USER_CASE):
             errors.append({
                 'type': 'subscription',
-                'message': _('Your application is using User Case functionality. You can remove User Case '
+                'message': _('Your application is using User Properties. You can remove User Properties '
+                             'functionality by opening the User Properties tab in a form that uses it, and '
+                             'clicking "Remove User Properties".')
+                           if toggles.USER_PROPERTY_EASY_REFS.enabled(self.domain) else
```
dimagimon: E131 continuation line unaligned for hanging indent

corehq/apps/app_manager/models.py
```diff
@@ -5628,9 +5641,14 @@ def app_uses_usercase(app):
...
+                           # old message, to be removed with USER_PROPERTY_EASY_REFS toggle
```
dimagimon: E131 continuation line unaligned for hanging indent

millerdev:
Added a really ugly commit to hide all (template) changes behind the feature flag, so this PR can now be merged and QA, which is not expected to happen for a while, can be done on prod.

I purposely used half-width indentation in some places to make the diffs easy to review. This should also make the removal of feature flag and related cruft easy and have minimal impact on template diffs.

corehq/apps/app_manager/models.py
```diff
@@ -4814,7 +4814,7 @@ def validate_app(self):
             })
         except UserCaseXPathValidationError as ucve:
             errors.append({
-                'type': 'invalid user case xpath reference',
+                'type': 'invalid user property xpath reference',
```
millerdev: This was a bug. Should have been changed before.

corehq/apps/app_manager/management/commands/migrate_app_to_cmitfb.py
```diff
@@ -7,55 +7,109 @@
...
+                domains.append(ident)
+            logger.info('migrating {} apps in domain {}'.format(len(ids), ident))
 
         for app_id in app_ids:
             logger.info('migrating app {}'.format(app_id))
             self.migrate_app(app_id)
 
+        if self.migrate_usercase:
+            for domain in domains:
+                USER_PROPERTY_EASY_REFS.set(domain, True, NAMESPACE_DOMAIN)
```
emord: I think that there will be some period of time where the user can't access their user properties here, if you don't set this feature flag whenever you empty out `usercase_preload`. Might not be a big deal until you try to convert a lot of domains though

emord (commented):
One comment about the management command. Likely not blocking, but I'll let you decide if it's worth changing

millerdev:
All tests passed. ✅ The failure was due to a timeout during coveralls.

corehq/apps/app_manager/management/commands/migrate_app_to_cmitfb.py
millerdev: Do you mean while the management command is running? Like after it converts the user properties for all the apps in the domain, but before this line is executed? I assumed the process would be fast enough that it wouldn't be a big deal, but maybe I'm wrong. Also, possibly more concerning is if there was an error migrating an app (that is not the first app to be migrated in the domain). In that case some apps would be migrated and others would not be, and the flag would not be set because the command errored out. I could see that being a problem, but I'm not sure how to work around it. Maybe wait to save the apps until they've all been migrated? -- although even then we run into problems because we don't have a real transactional database.

emord: Yeah. there's not a great solution since its not a transactional DB. 

One solution could be if there is some error, then `usercase_preload` would still be present. We could add an alert to the application page if `usercase_preload` is not empty and the flag is set for the domain. I believe @orangejenny did something similar for apps that were not able to be migrated for the other case management work

millerdev: I don't think this will matter at all during QA and UAT and we can work out a better solution before we start migrating lots of domains. What do you think of merging as-is and work on this problem after we've gotten past QA and UAT (or while those are in progress)?

emord: Yeah. Just merged it. As a note QA sometimes have weird apps, so it's probably worth it to make a note in the QA ticket to create a new domain.

millerdev: Noted, thanks.

PR #14224, opened 2016-12-13
check for makemessages failures and warn

esoergel:
@dannyroberts
I think this has been failing for about a month, but the script still completed, so we just went with it

### DIFF
--- a/scripts/update-translations.sh
+++ b/scripts/update-translations.sh
@@ -33,9 +33,19 @@ echo "Pulling translations from transifex"
 tx pull -f
 
 echo "Gathering all translation strings.  Note that this will probably take a while"
-./manage.py makemessages --all
+./manage.py makemessages --all --ignore 'corehq/apps/app_manager/tests/data/v2_diffs*' --ignore 'node_modules'
+
+if [[ $1 -ne "0" ]]; then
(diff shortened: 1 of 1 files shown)


scripts/update-translations.sh
```diff
@@ -33,9 +33,19 @@ echo "Pulling translations from transifex"
 tx pull -f
 
 echo "Gathering all translation strings.  Note that this will probably take a while"
-./manage.py makemessages --all
+./manage.py makemessages --all --ignore 'corehq/apps/app_manager/tests/data/v2_diffs*' --ignore 'node_modules'
```
esoergel: These diffs have bad stuff in them, figure it's best just to skip.

dannyroberts: Those are definitely both good to skip, we wouldn't want to pull translation-looking things from there even if it didn't happen to crash, so this is an improvement all around

dannyroberts (commented):
Cool, looks like a nasty problem and a great solution!

scripts/update-translations.sh
benrudolph: 👍

benrudolph (approved):

PR #14225, opened 2016-12-14
don't allow case processing with invalid index references

czue:
related to http://manage.dimagi.com/default.asp?243839

need @ctsims to confirm this is appropriately mirroring mobile behavior before merging.

cc @proteusvacuum @emord @NoahCarnahan buddy

### DIFF
--- a/corehq/ex-submodules/casexml/apps/case/exceptions.py
+++ b/corehq/ex-submodules/casexml/apps/case/exceptions.py
@@ -21,6 +21,13 @@ class IllegalCaseId(CommCareCaseError):
     pass
 
 
+class InvalidCaseIndex(CommCareCaseError):
+    """
+    Raise when someone tries to set an index relationship on a case to an invalid case.
+    """
+    pass
+
+
 class UsesReferrals(CommCareCaseError):
     pass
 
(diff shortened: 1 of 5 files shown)


NoahCarnahan:
Looks good to me

snopoke (commented):
LGTM

ctsims:
@proteusvacuum @czue I just ran some tests and reviewed the spec, and it might be a bit more complex on our side.

We don't prevent you from making a case with a missing index, we simply purge the case when we next purge, because there is a parent dependency that doesn't exist, which must be invalid. It's actually kind of tough for us to **check** for a case index at the time of indexing for performance reasons (our databases are much slower than yours) and for document ordering reasons (IE: when HQ sends down cases during a restore I'm a bit worried that keeping them in document/graph order might be annoying/difficult). 

So the current outcome on the mobile side is that we won't reject the forms outright, and they will make it to the server. Will they **block** on the server side (IE: HQ will reject them) or will they be accepted and an error logged? If the latter, that seems good for now.

czue:
> Will they block on the server side (IE: HQ will reject them) or will they be accepted and an error logged? 

the latter - they will just go into the form error quarantine - same as if you submitted an empty case ID or something like that.

(based on this feedback taking off label as this seems like overall better/more consistent behavior)

ctsims:
@czue Sounds good / correct to me.

PR #14227, opened 2016-12-14
Add last_modified to case list properties

wpride:
I have no idea if this is actually this straightforward but worked for me and seems correct to have this property @orangejenny @emord possibly?

### DIFF
--- a/corehq/apps/app_manager/views/modules.py
+++ b/corehq/apps/app_manager/views/modules.py
@@ -252,7 +252,7 @@ def _get_fixture_columns_by_type(domain):
 
 
 def _setup_case_property_builder(app):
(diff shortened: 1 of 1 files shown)


wpride:
screen shot 2016-12-14 at 8 21 28 am

czue (approved):
nice little quick PI! 

looks good to me though will let someone else confirm if this isn't right.

orangejenny:
looks good to me

emord:
Unrelated, but why doe `#owner_name` have a # ?

czue:
guessing, but i think because it's magic that resolves to something fancy and we need a way to differentiate that from a property named `owner_name`

PR #14228, opened 2016-12-14
noisy report alert

emord:
http://manage.dimagi.com/default.asp?243866#1261207

@czue seemed like having an alert for reports getting dropped is useful, but this should give us 10 emails instead of 1000

@proteusvacuum

### DIFF
--- a/corehq/apps/reports/scheduled.py
+++ b/corehq/apps/reports/scheduled.py
@@ -5,7 +5,7 @@ from corehq.apps.reports.models import ReportNotification
 
 def get_scheduled_reports(period, as_of=None):
     as_of = as_of or datetime.utcnow()
-    assert period in ('daily', 'weekly', 'monthly')
+    assert period in ('daily', 'weekly', 'monthly'), period
 
     def _keys(period, as_of):
         minute = guess_reporting_minute(as_of)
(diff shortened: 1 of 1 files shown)


corehq/apps/reports/scheduled.py
```diff
@@ -1,11 +1,15 @@
...
+from corehq.util.soft_assert import soft_assert
+
+
+_soft_assert = soft_assert(exponential_backoff=True)
 
 
 def get_scheduled_reports(period, as_of=None):
     as_of = as_of or datetime.utcnow()
-    assert period in ('daily', 'weekly', 'monthly')
+    assert period in ('daily', 'weekly', 'monthly'), period
```
emord: @millerdev pr followup 2 years after the fact :) https://github.com/dimagi/commcare-hq/pull/4285/files#r17305879

czue: hahaha my bad. thanks for the :mag_right:

czue (approved):
thanks

corehq/apps/reports/scheduled.py
emord: Ha. cory linked the commit for some context in the interrupt ticket. figured I'd just follow up. Don't think it's ever been triggered

PR #14229, opened 2016-12-14
Redirect to app list if appId is null on session navigation

wpride:
Should resolve issue with requests to formplayer being null


corehq/apps/cloudcare/static/cloudcare/js/formplayer/router.js
```diff
@@ -37,7 +37,12 @@ FormplayerFrontend.module("SessionNavigate", function (SessionNavigate, Formplay
             var urlObject = Util.CloudcareUrl.fromJson(
                 Util.encodedUrlToObject(sessionObject || Backbone.history.getFragment())
             );
-            FormplayerFrontend.Menus.Controller.selectMenu(urlObject);
+            if (_.isUndefined(urlObject.appId)) {
```
benrudolph: this one would actually be better as `!appId` since `appId = ''` will attempt to select a menu

benrudolph (approved):

PR #14230, opened 2016-12-14
Serialize named tuple as dict

benrudolph:
@dannyroberts the soil change broke progress bars on exports because it changed the datatype of `progress`. A namedtuple when serialized returns an array while on the frontend we expect it to be a dictionary. i'm assuming this'll probably break the case import progress bar. going to take a look at that quickly (should be easy to fix)
http://manage.dimagi.com/default.asp?243918

cc: @millerdev


benrudolph:
offlined with danny and this should be good to go as is. it's currently on staging as well

dannyroberts (approved):
Thanks @benrudolph!

PR #13989, opened 2016-11-21
Warn on unmigrated apps

orangejenny:
screen shot 2016-11-21 at 1 08 46 pm

fyi @amsagoff 

@emord / @czue 

Hold off on merge until swiss server is migrated (currently waiting on client response).

PR #14231, opened 2016-12-14
Label questions for data schema

benrudolph:
@czue @NoahCarnahan this adds label questions to the export schema. One thing to note about this is that i've implemented this with labels being a different `ExportItem` (`LabelItem`). This means that if you reuse your label id, you will have two columns in your export. i figured you didn't want to mix label data, which is all `"OK"`, with real question data. if you think that's confusing though let me know. it would look like this:

screen shot 2016-12-14 at 12 39 05 pm

it's similar to how if you change a multiple choice question to a text question but keep the same id. it will show up as two columns on the export page.

cc: @millerdev

### DIFF
--- a/corehq/apps/app_manager/xform.py
+++ b/corehq/apps/app_manager/xform.py
@@ -900,6 +900,11 @@ class XForm(WrappedNode):
 
         if the xform is bad, it will raise an XFormException
 
+        :param include_triggers: When set to True will return label questions as well as regular questions
+        :param include_groups: When set will return repeats and group questions
+        :param include_translations: When set to True will return all the translations for the question
+        :param form: When included adds a hashtagValue to the dict returned
+
         """
 
         if not self.exists():
--- a/corehq/apps/export/const.py
+++ b/corehq/apps/export/const.py
@@ -14,7 +14,7 @@ from corehq.apps.export.transforms import (
 
 # When fixing a bug that requires existing schemas to be rebuilt,
 # bump the version number.
-DATA_SCHEMA_VERSION = 6
+DATA_SCHEMA_VERSION = 7
 
 DEID_ID_TRANSFORM = "deid_id"
 DEID_DATE_TRANSFORM = "deid_date"
--- a/corehq/apps/export/models/__init__.py
+++ b/corehq/apps/export/models/__init__.py
@@ -25,6 +25,7 @@ from .new import (
     InferredExportGroupSchema,
     ExportGroupSchema,
     TableConfiguration,
+    LabelItem,
     ScalarItem,
     GeopointItem,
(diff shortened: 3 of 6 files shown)


corehq/apps/app_manager/xform.py
```diff
@@ -900,6 +900,11 @@ def get_questions(self, langs, include_triggers=False,
 
         if the xform is bad, it will raise an XFormException
 
+        :param include_triggers: When set to True will return label questions as well as regular questions
+        :param include_groups: When set will return repeats and group questions
+        :param include_translations: When set to True will return all the translations for the question
+        :param form: When included adds a hashtagValue to the dict returned
```
benrudolph: @orangejenny i was looking through this function and saw that you can pass in an entire form object which just gets used to set a boolean: https://github.com/dimagi/commcare-hq/blob/br/label-questions!/corehq/apps/app_manager/xform.py#L919. I'm curious if we can just drop this argument completely and always include the hashtag value. i'd imagine it'd actually be fairly useful for user facing things

millerdev (approved):

corehq/apps/app_manager/xform.py
orangejenny: Yeah, that would be a good change. I stopped using the `form` in https://github.com/dimagi/commcare-hq/pull/13989 but didn't realize it.

corehq/apps/export/const.py
```diff
@@ -14,7 +14,7 @@
 
 # When fixing a bug that requires existing schemas to be rebuilt,
 # bump the version number.
-DATA_SCHEMA_VERSION = 6
+DATA_SCHEMA_VERSION = 7
```
czue: remind me - do these rebuilds happen just in time and automatically?

czue (approved):
lgtm

NoahCarnahan:
Thanks Ben. Was lack of this export item type preventing migration of existing exports that export labels?

NoahCarnahan:
nvm, I see the slack thread now.

corehq/apps/export/const.py
benrudolph: yes to both

benrudolph:
👍 dig the new profile pic :camera_flash:

corehq/apps/export/const.py
benrudolph: interesting thought, i'm about to make another schema change so i'll put it into that

corehq/apps/app_manager/xform.py
benrudolph: https://github.com/dimagi/commcare-hq/pull/14251/commits/c375fb45efaecc75815f0882c8d60e602095dc7b

PR #14232, opened 2016-12-14
Management command to check invalid case references

emord:
@czue http://manage.dimagi.com/default.asp?243839#1261236

This is a first pass. I dumped the entire index in the csv instead of just hte referenced id so that it would be easier to see the type and relationship. Looks like there's about 95 invalid refs for enikshay

buddy @proteusvacuum


czue (approved):
awesome, thanks. good idea to use ES - I was worried we'd have to load all the cases into memory.

PR #14233, opened 2016-12-14
jQuery 3 retry

orangejenny:
Re-releasing jQuery 3, but leaving new cloudcare on jQuery 2 because [Marionette 2 and jQuery 3 seem to be incompatible](https://github.com/marionettejs/backbone.marionette/issues/3018).

I also left old cloudcare on jQuery 2, because it has some incompatibilities, which are likely fixable, but if we're almost done with old cloudcare, doesn't seem worth the trouble to migrate.

It's possible that issues will arise in the future if HQ's base template calls code that's jq2-incompatible, but don't think it's that likely, since the base template doesn't do much. Hoping that new cloudcare will eventually upgrade marionette.

@benrudolph


corehq/apps/cloudcare/views.py
```diff
@@ -303,6 +310,7 @@ class FormplayerPreviewSingleApp(View):
     urlname = 'formplayer_single_app'
 
     @use_datatables
+    @use_legacy_jquery
     @use_jquery_ui
     @method_decorator(require_cloudcare_access)
     @method_decorator(requires_privilege_for_commcare_user(privileges.CLOUDCARE))
```
benrudolph: i think we also need it for `PreviewAppView`

benrudolph (changes requested):
one minor thing for app preview

corehq/apps/cloudcare/views.py
orangejenny: confused - doesn't line 363 handle `PreviewAppView`?

benrudolph: oops sorry missed that

benrudolph (approved):

orangejenny:
qa passed, this is good to go

orangejenny:
Since approved, I just added a commit and then later reverted it. Self-merging.

PR #14234, opened 2016-12-14
Case importer datadog

dannyroberts:
- Create `datadog_gauge_task` for easily running a function
- use it to track total file size of case upload files over time

### DIFF
--- a/corehq/apps/case_importer/tasks.py
+++ b/corehq/apps/case_importer/tasks.py
@@ -1,7 +1,11 @@
+from celery.schedules import crontab
 from celery.task import task
 from corehq.apps.case_importer.exceptions import ImporterError
+from corehq.apps.case_importer.tracking.analytics import \
+    get_case_upload_files_total_bytes
 from corehq.apps.case_importer.tracking.case_upload_tracker import CaseUpload
 from corehq.apps.case_importer.util import get_importer_error_message
+from corehq.util.datadog.gauges import datadog_gauge_task
 from dimagi.utils.couch.database import is_bigcouch
 from casexml.apps.case.mock import CaseBlock, CaseBlockError
 from corehq.apps.hqcase.utils import submit_case_blocks
@@ -279,3 +283,10 @@ def do_import(spreadsheet, config, domain, task=None, chunksize=CASEBLOCK_CHUNKS
         'errors': errors.as_dict(),
         'num_chunks': num_chunks,
     }
+
+
+total_bytes = datadog_gauge_task(
+    'case_importer.files.total_bytes',
+    get_case_upload_files_total_bytes,
+    run_every=crontab(minute=0)
+)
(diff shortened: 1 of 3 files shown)


dannyroberts:
I should see one of these for staging on datadog shortly after 5pm (in 40 minutes).

dannyroberts:
Ah, I was reminded that staging isn't hooked up to datadog. I tested this in a production shell with copy-pasting, and I've separately tested that the periodic_task part of it works, so this is good to merge by me.

dannyroberts:
(And no need to wait for tests)

benrudolph:
sweet 🔢 🐶

dannyroberts:
Weee! Still open to comments on the utility I added if there are any suggestions

czue:
i like this a lot - looks like it could be a really useful tool in the toolkit

corehq/apps/case_importer/tasks.py
```diff
@@ -279,3 +283,10 @@ def _submit_caseblocks(domain, case_type, caseblocks):
         'errors': errors.as_dict(),
         'num_chunks': num_chunks,
     }
+
+
+total_bytes = datadog_gauge_task(
+    'case_importer.files.total_bytes',
```
snopoke: have been trying to keep all metrics coming from commcare prefixed with `commcare` to make it easier to search in datadog.

corehq/util/datadog/gauges.py
```diff
@@ -0,0 +1,36 @@
+from functools import wraps
+from celery.task import periodic_task
+from corehq.util.datadog import statsd
+
+
+def datadog_gauge_task(name, fn, run_every):
+    """"
```
snopoke: nit: extra `"`

corehq/util/datadog/gauges.py
```diff
@@ -0,0 +1,36 @@
+from functools import wraps
+from celery.task import periodic_task
+from corehq.util.datadog import statsd
+
+
+def datadog_gauge_task(name, fn, run_every):
```
snopoke: this is nice

corehq/apps/case_importer/tasks.py
dannyroberts: Thanks, this is the kind of thing I knew I must have missed. Barely any data's come in so I'll just change this now. Out of curiosity

- is there any reason to restrict the length of these? I assume `'commcare.case_importer.files.total_bytes'` is fine, but if there's reason for it to be shorter, or there's some better convention to follow, I can use something else.
- There's no way to change the name of a metric already in datadog, right? I'm imagining a case where we'd already submitted a few months of data to this, and then decided we wanted it prefixed with `commcare`, but didn't want to lose continuity.

corehq/util/datadog/gauges.py
dannyroberts: 🐦 👁

corehq/apps/case_importer/tasks.py
snopoke: No reason to make them short that I know of. The way I think of these is that the dot prefixes are used to group related metrics.

dannyroberts: Cool, that's how I was thinking of it too.

PR #14235, opened 2016-12-14
use done instead of success on jqXhr

dannyroberts:
for compatibility with jQuery 3


dannyroberts:
no need to wait for the build, since this isn't tested. I checked it's compatible with both the current jquery version and the new one.

PR #14238, opened 2016-12-14
Case import history fix stale tasks

dannyroberts:
I've noticed that in production case uploads more than a couple days old show up as "Waiting to start" in the case upload history, because the associated celery task (`TaskModel` in the `DatabaseBackend`) has been deleted. This PR makes case importer persist task data to its own model once the task is no longer in flux (i.e. once it's succeeded or failed).

When to actually save it was a bit tricky, but I decided to have two entry points just to make sure it happened:

- on task complete, a second task is triggered that just records the info from the first task
- when a user views recent uploads, it also saves task state to any that are complete and not saved yet

The first one might not happen if the task is forcibly killed in such a way that the `finally` of the task try/except doesn't get executed; the second might not happen if you close the window before the tasks complete. Together they cover nearly all use-cases. Finally, for tasks that slip through the cracks, if any, it'll correctly display that the information is missing.

Performance-wise, it's done in such a way that it never saves unless there's something to save, and fetching work that would otherwise be duplicated is cached. I'd be surprised if it adds any significant overhead. If it does, a more async approach would be to have a periodic task run the procedure in bulk every 10 minutes. I like that less because the performance characteristics/extra work done are less predictable to me.


dannyroberts:
(just merging master's dimagi-utils ref, so no need to wait for build)

@benrudolph I think @esoergel is offline; mind taking a look at this?

corehq/apps/case_importer/tracking/models.py
```diff
@@ -23,6 +27,29 @@ class Meta(object):
...
+
+        :return: True if the caller must save model to persist the effect
+
+        """
+        if self.task_status_json is None:
+            # intentionally routing through method to prime local cache
+            task_status_json = self.get_task_status_json()
+            if task_status_json.is_finished():
+                self.task_status_json = task_status_json
+                return True
```
benrudolph: just curious, why not just `save` the model here instead of return `True` and then saving?

benrudolph (approved):
looks good to me

corehq/apps/case_importer/tracking/models.py
dannyroberts: I'm not sure what the best practice is, but I try to not have models save themselves since that steals the ability of the caller to optimize saves. Even if in this case there's no optimization done, I still like being able to tell from the calling code when something was actually saved. When I'm reading code that interacts with models this is constantly going through my head:

```
my_model.do_something()  # is it saved yet??
my_model.save()  # oh, whew, ok here's where it's saved
```
so I like to see that save there. Might just be personal preference; I don't recall ever having formalized this as a best practice even in conversation.

benrudolph: 👍 that's fair

czue: i agree with that (previously unstated) principle. not saving also allows you to chain calls without hitting the DB multiple times. could be a good little guideline for the architecture handbook if we're all aligned.

dannyroberts: https://github.com/dimagi/architecture-guide/pull/1

PR #14239, opened 2016-12-15
New Multiple Locations support in Bulk user upload

sravfeyn:
https://trello.com/c/PI1FHwOi/37-multiple-location-assignment-bulk-upload

@snopoke 
cc @esoergel 

Ready for review. I will need to resolve couple of spec related questions with Sheel before having this merged

### DIFF
--- a/corehq/apps/users/bulkupload.py
+++ b/corehq/apps/users/bulkupload.py
@@ -382,9 +382,7 @@ def create_or_update_users_and_groups(domain, user_specs, group_specs, task=None
             username = row.get('username')
             location_codes = row.get('location_code') or []
             if location_codes and not isinstance(location_codes, list):
-                raise UserUploadError(_("location_code must be listed as multiple columns of "
-                                        "'location_code 1, location_code 2 ...' for each of users's location. "
-                                        "If user has only one location, it should be listed as 'location_code 1'"))
+                location_codes = [location_codes]
             # ignore empty
             location_codes = [code for code in location_codes if code]
             role = row.get('role', '')
--- a/corehq/apps/users/bulkupload.py
+++ b/corehq/apps/users/bulkupload.py
@@ -577,27 +577,6 @@ class GroupNameError(Exception):
         )
 
 
-def get_location_rows(domain):
-    users = CommCareUser.by_domain(domain)
-
-    mappings = []
-    for user in users:
-        # this method is only called when exporting the
-        # locaiton tab (so on domains with multiple
-        # locations per user), so we are relying on
-        # user.locations being a thing that is real
-        # and working
-        locations = user.locations
-        for location in locations:
-            mappings.append([
-                user.raw_username,
-                location.site_code,
-                location.name
-            ])
-
-    return mappings
-
-
 def build_data_headers(keys, header_prefix='data'):
     return json_to_headers(
         {header_prefix: {key: None for key in keys}}
@@ -766,16 +745,6 @@ def dump_users_and_groups(response, domain):
         ('users', user_rows),
         ('groups', group_rows),
     ]
-
-    domain_obj = Domain.get_by_name(domain)
-    # This is only for domains using the multiple locations feature flag
-    if domain_obj.commtrack_enabled and domain_obj.supports_multiple_locations_per_user:
-        headers.append(
-            ('locations', [['username', 'location_code', 'location name (optional)']])
-        )
-        rows.append(
-            ('locations', get_location_rows(domain))
-        )
 
     writer.open(
         header_table=headers,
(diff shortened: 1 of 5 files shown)


czue:
nice! derek will be pleased

snopoke (commented):
Didn't review very closely.

corehq/apps/users/bulkupload.py
```diff
@@ -650,8 +654,6 @@ def _set_progress(progress):
             ) % (_error_message, e.errors))
             ret['errors'].append(_error_message)
 
-    if Domain.get_by_name(domain).supports_multiple_locations_per_user:
-        create_or_update_locations(domain, location_specs, log=ret)
```
esoergel: Yikes, I didn't even know we did this.  That function should be able to be removed now.

esoergel (approved):

corehq/apps/users/bulkupload.py
```diff
@@ -476,7 +380,13 @@ def _set_progress(progress):
             uncategorized_data = row.get('uncategorized_data')
             user_id = row.get('user_id')
             username = row.get('username')
-            location_code = row.get('location_code', '')
+            location_codes = row.get('location_code') or []
+            if location_codes and not isinstance(location_codes, list):
```
sravfeyn: Yeah, this is what I want to do, wanted to get confirmation from Sheel

corehq/apps/users/models.py
```diff
@@ -1777,6 +1777,7 @@ def unset_location_by_id(self, location_id, fall_back_to_next=False):
     def reset_locations(self, location_ids):
         """
         Reset user's assigned_locations to given location_ids and update user data.
+            This should be called after updating primary location via set_location/unset_location
```
sravfeyn: It's hard to enforce it, as it's hard to check from the data whether primary location is updated or not. I was considering to just combining resetting locations and setting primary location into just this one method, but there are multiple ways of checking/enforcing consistency on whether a primary location should be set or not. Instead I thought it's better to leave that logic to callers.

corehq/apps/users/bulkupload.py
sravfeyn: Updated

sravfeyn:
@sheelio 
I have completed most of coding for new [bulk location assignment](https://trello.com/c/PI1FHwOi/37-multiple-location-assignment-bulk-upload). Couple of spec related questions, before I can release this
1. I have removed extra sheet for locations assignment. Same for download
2. location_code column can be just `location_code` one column or multiple columns `location_code 1`, `location_code 2`...
3. Locations won't be created from just bulk user upload. 
4. Is there any restriction on which webusers can assign which locations to which mobile workers? (currently no restrictions are considered)

czue:
not sure if this is time sensitive, but jumping in with quick thoughts on a few of these since @sheelio is offline. let me know if we should be trying to get consensus on this before January

2. seems fine.
3. seems definitely right/desired behavior.
4. this seems tricky and possibly out of scope for current effort if you're using location restrictions. my sense is that the most "correct" behavior would be that you should only be able to modify any user who is assigned to at least one location you can see, and you should only be able to add/remove locations that you have access to. in practice that would be super confusing for a web user so not sure how to figure that out.

czue:
how did we land on the one-column-per-location/group format, anyways? i find it pretty odd (mostly coming from a relational db background), but maybe it's standard practice? (maybe @sheelio the only one who knows that)

sravfeyn:
Re. #4, it's pretty easy to add restrictions in bulk user restrictions. But the spec itself needs some iteration (for e.g., should a user have at least one location assigned manually to be able to use bulk? Can webuser de-assign a location that he/she doesn't have access to). Right now, I can add that a webuser can assign/reassign only those locations that the user has access to. Or we can proceed by adding require-project-admin or no restrictions on webuser at al.

czue:
i think the decision as to what to do right now depends on the time-sensitivity of getting this in. turning off this feature if you have location restrictions enabled seems like the easy short term solution if we need to block on more product capacity to figure out the details though i don't know that impacts current projects.

sravfeyn:
I think Derek is the one who needed this. I will check with him to find the urgency

sravfeyn:
For now I have added logic to disable the feature if location-restritions is enabled. So I am removing the hold-off label

esoergel:
Flagging that the `UploadCommCareUsers` view is not marked as location-safe, so it's not accessible to restricted users anyways.  I did see that you blocked location assignment to the feature flag that was the old means of restricting web-users by location, which is deprecated, and we're working to remove.  I don't think it matters too much how we handle that case, so it's good to be conservative, and I definitely don't think we should be adding any new features there.

At some point (soon?) we'll want to make bulk-user upload location-safe, but that would also require restricting the users you can modify and requiring you to assign locations for new users.  I agree with your assessment that there are some corner cases we should iron out in spec, and that that is out of scope for this PR (no sense in blocking this useful feature for people who can use it now!).

dlanedimagi:
Yes, Derek is very excited about this! and happy to answer any user/pain point/field implementation questions.

PR #14240, opened 2016-12-15
automatically add missing instances if possible

snopoke:
Not sure why I didn't do this in the first place.

@sravfeyn

cc @millerdev


millerdev:
Do we know why some forms have missing instances? Is it because the form builder is not adding them even though it's adding xpath expressions that reference them? If that's the case then there may be a bug in the form builder.

snopoke:
No I think it's mostly people manually adding references to instances in calculates e.g. fixtures / groups etc.

millerdev (approved):

millerdev (approved):

PR #14243, opened 2016-12-15
Add model id to resource not found

benrudolph:
@millerdev am i right in assuming that `self._id` will always be available? it seems like that would be the case, but perhaps there's an edge case where you try and fetch an attachment before the model has been saved? adding this because it's hard to track down the id of a model that throws a resourcenotfound: http://manage.dimagi.com/default.asp?244019

### DIFF
--- a/corehq/blobs/mixin.py
+++ b/corehq/blobs/mixin.py
@@ -130,8 +130,12 @@ class BlobMixin(Document):
                 raise NotFound
             blob = db.get(meta.id, self._blobdb_bucket())
         except NotFound:
-            raise ResourceNotFound(u"{model} attachment: {name!r}".format(
-                                   model=type(self).__name__, name=name))
+            raise ResourceNotFound(
+                u"{model} {model_id} attachment: {name!r}".format(
+                    model=type(self).__name__,
+                    name=name,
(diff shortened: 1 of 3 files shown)


millerdev:
Should be fine to use `self._id` there.

Test failed:
```
FAIL: test_process_change_with_deleted_document (corehq.blobs.tests.test_pillow.TestBlobDeletionProcessor)
  File "/mnt/commcare-hq-ro/corehq/blobs/tests/test_pillow.py", line 21, in test_process_change_with_deleted_document
    self.obj.fetch_attachment("name")
  File "/usr/local/lib/python2.7/contextlib.py", line 35, in __exit__
    self.gen.throw(type, value, traceback)
  File "/vendor/lib/python2.7/site-packages/testil.py", line 244, in raises
    eq_(str(err), msg)
  File "/vendor/lib/python2.7/site-packages/testil.py", line 54, in eq
    raise AssertionError(err)
AssertionError: not equal
"FakeCouchDocument 96e019833dc64d4b8ec52333245f5eb3 attachment: 'name'"
"FakeCouchDocument attachment: 'name'"
```

millerdev (approved):

PR #14249, opened 2016-12-15
rebalance travis tests

emord:
Had the second test fail on me multiple times today. Looked through the last 10 or so builds and the 3rd test is consistently 3 to 6 minutes faster than the first two.

@millerdev I think this is how we rebalance correct? Is there a science to this? I just gave it my best guess

buddy @proteusvacuum

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,9 +7,9 @@ env:
   matrix:
     - TEST=javascript
     # need node installs for app_manager.XpathValidatorTest
-    - TEST=python NOSE_DIVIDED_WE_RUN=05 JS_SETUP=yes
-    - TEST=python NOSE_DIVIDED_WE_RUN=6a JS_SETUP=yes
-    - TEST=python NOSE_DIVIDED_WE_RUN=bf JS_SETUP=yes
+    - TEST=python NOSE_DIVIDED_WE_RUN=03 JS_SETUP=yes
+    - TEST=python NOSE_DIVIDED_WE_RUN=69 JS_SETUP=yes
+    - TEST=python NOSE_DIVIDED_WE_RUN=ba JS_SETUP=yes
     - TEST=python-sharded
   global:
     # TRAVIS_HQ_USERNAME
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,9 +7,9 @@ env:
   matrix:
     - TEST=javascript
     # need node installs for app_manager.XpathValidatorTest
-    - TEST=python NOSE_DIVIDED_WE_RUN=03 JS_SETUP=yes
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -7,9 +7,9 @@ env:
   matrix:
     - TEST=javascript
     # need node installs for app_manager.XpathValidatorTest
-    - TEST=python NOSE_DIVIDED_WE_RUN=05 JS_SETUP=yes
-    - TEST=python NOSE_DIVIDED_WE_RUN=6a JS_SETUP=yes
-    - TEST=python NOSE_DIVIDED_WE_RUN=bf JS_SETUP=yes
+    - TEST=python NOSE_DIVIDED_WE_RUN=03 JS_SETUP=yes
+    - TEST=python NOSE_DIVIDED_WE_RUN=69 JS_SETUP=yes
+    - TEST=python NOSE_DIVIDED_WE_RUN=ba JS_SETUP=yes
```
millerdev: these should be `03` `4a` `bf` - each string represents a range, and they should be contiguous or it will skip some tests. As is, you'd be skipping `46` (that is, any test whose "name" MD5 starts with `4`, `5`, or `6`) and probably `af`. I think you might get an error for `ba` since `b` > `a` so not a valid range.

As for how to pick the values, I'd probably go with `04` `5a` `bf` and see if that works. If the first is still taking to long, then go to `03` `4a` `bf`. I'm assuming because of the change you made here, the first runner is hitting timeout, the second has time to spare, and the third is about right?

millerdev: Oh wait, according to your PR description you might want `04` `57` `8f`, although that's shifting a lot to the third runner.

proteusvacuum: wow, I never knew what those values meant. Thanks for the explanation!

millerdev: It also might make sense to add a 4th runner if 3 can no longer handle all the tests without timeout.

millerdev: it's documented in the [`--divided-we-run` plugin option](https://github.com/dimagi/commcare-hq/blob/b4b36a9bcae86523f54dd656c55e503a3a6ca193/corehq/tests/noseplugins/dividedwerun.py#L26-L32)

emord: Hmm. yeah that does seem like a lot of tests on 3, but I suppose it matters more about the speed than the number. Thanks for the explanation. For some reason I thought they were hex values 

> As is, you'd be skipping 46 (that is, any test whose "name" MD5 starts with 4, 5, or 6) and probably af.

The goal was to make the tests time out less :)

millerdev (approved):

czue:
beautiful
image

PR #14250, opened 2016-12-15
Add '_archive_' as WEIRD_OWNER_ID

proteusvacuum:
Not sure if this is necessary or not for ownership cleanliness.
We are using the magic string `_archive_` to reassign cases in eNikshay so that they go into some limbo archived state. 
@czue would be great to get your input... 
@emord code buddy


czue (approved):
this looks good. basically anything in this list should be something we don't expect to ever have to generate a case footprint for (which applies to `_archive_`)

czue:
respecting label, but i think :+1: to merge

emord (approved):

PR #14251, opened 2016-12-15
Repeat child cases in exports

benrudolph:
@czue @NoahCarnahan more functionality so we can convert more exports. This ensures that we include the ability to export case properties from repeat groups when you create cases from a repeat. Found this when trying to migrate the `nw-veges` domain

cc: @millerdev 

best read commit by commit 🐟 

this is what it looks like:

screen shot 2016-12-15 at 5 53 27 pm
screen shot 2016-12-15 at 5 53 42 pm

### DIFF
--- a/corehq/apps/export/models/new.py
+++ b/corehq/apps/export/models/new.py
@@ -1311,7 +1311,7 @@ class FormExportDataSchema(ExportDataSchema):
                 #
                 # /data/repeat/other_group/question
                 #
-                # We want to currate a path that looks like:
+                # We want to create a path that looks like:
                 #
                 # /data/repeat/case/update/other_group/question
                 path_suffix = case_path[len(subcase_action.repeat_context):]
--- a/corehq/apps/export/models/new.py
+++ b/corehq/apps/export/models/new.py
@@ -57,6 +57,8 @@ from corehq.apps.export.const import (
     MISSING_VALUE,
     EMPTY_VALUE,
     KNOWN_CASE_PROPERTIES,
+    CASE_ATTRIBUTES,
+    CASE_CREATE_ELEMENTS,
     UNKNOWN_INFERRED_FROM,
 )
 from corehq.apps.export.exceptions import BadExportConfiguration
@@ -1233,11 +1235,16 @@ class FormExportDataSchema(ExportDataSchema):
 
     @staticmethod
     def _process_app_build(current_schema, app, form_xmlns):
-        xform = app.get_form_by_xmlns(form_xmlns, log_missing=False)
-        if not xform:
+        form = app.get_form_by_xmlns(form_xmlns, log_missing=False)
+        if not form:
             return current_schema
-        case_updates = xform.get_case_updates(xform.get_module().case_type)
-        xform = xform.wrapped_xform()
+
+        case_updates = form.get_case_updates(form.get_module().case_type)
+        xform = form.wrapped_xform()
+        repeats_with_subcases = {
+            subcase_action for subcase_action in form.actions.subcases
+            if subcase_action.repeat_context
+        }
         xform_schema = FormExportDataSchema._generate_schema_from_xform(
             xform,
             case_updates,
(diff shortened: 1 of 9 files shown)


benrudolph:
(will fix tests)

czue (approved):
looks good though admittedly i reviewed rather quickly. Does `OpenSubCaseAction` also represent when you update (but don't open) subcases? or do we not care because we don't allow you to do that in standard app builder (i think that might be the case)?

corehq/apps/export/models/new.py
```diff
@@ -1209,6 +1212,10 @@ def type(self):
         return FORM_EXPORT
 
     @classmethod
+    def schema_version(cls):
+        return FORM_DATA_SCHEMA_VERSION
```
millerdev: nitpick: would be simpler to use a class attribute here:
```py
class Whatever:
    schema_version = FORM_DATA_SCHEMA_VERSION
```

corehq/apps/export/models/new.py
```diff
@@ -1233,27 +1240,103 @@ def get_latest_export_schema(domain, app_id, form_xmlns):
...
+        :param repeats_with_subcases: A list of OpenSubCaseAction classes that have a
+            repeat_context.
+        :param langs: An array of application languages
+        :param app_id: The app_id of the corresponding app
+        :param app_version: The build number of the app
+        :returns: An instance of a FormExportDataSchema
+        """
+
+        repeats = FormExportDataSchema._get_repeat_paths(xform, langs)
+        schema = FormExportDataSchema()
```
millerdev: minor: if this was a `classmethod` instead of a `staticmethod` you could `cls._get_repeat_paths(...)` and `cls()` here. This could be applied to a bunch of other static methods around here as well.

corehq/apps/export/models/new.py
```diff
@@ -1233,27 +1240,103 @@ def get_latest_export_schema(domain, app_id, form_xmlns):
...
+                _add_to_group_schema(group_schema, path, u'case.{}'.format(case_attribute))
+
+            # Add case updates
+            for case_property, case_path in subcase_action.case_properties.iteritems():
+                # This removes the repeat part of the path. For example, if inside
+                # a repeat group that has the following path:
+                #
+                # /data/repeat/other_group/question
+                #
+                # We want to currate a path that looks like:
```
millerdev: fancy word "currate" :)

millerdev (approved):

corehq/apps/export/models/new.py
benrudolph: 👍

corehq/apps/export/models/new.py
benrudolph: makes sense to do

corehq/apps/export/models/new.py
benrudolph: fancy word, but spell wrong 😞 "curate"

benrudolph: https://github.com/dimagi/commcare-hq/pull/14251/commits/254b6ba58e4a21d9819bd30eb0cc218de0e77f9a

corehq/apps/export/models/new.py
benrudolph: https://github.com/dimagi/commcare-hq/pull/14251/commits/21f1a33f923dae8aca5762c26e355bc927bdcf10

corehq/apps/export/models/new.py
benrudolph: i tried to do this, but it wouldn't place nicely with jsonobject. it kept wanting to wrap it

benrudolph:
> Does OpenSubCaseAction also represent when you update (but don't open) subcases? or do we not care because we don't allow you to do that in standard app builder (i think that might be the case)?

I didn't know this was possible. I'm leaning towards punting for now until I find some real word use cases of it

corehq/apps/export/models/new.py
millerdev: Fair enough. jsonobject is annoying that way.

millerdev (approved):

czue:
> I didn't know this was possible. I'm leaning towards punting for now until I find some real word use cases of it

TB does it :)

but yeah, if the interplay between safe to case and new exports right now is undefined then happy to leave it as is

NoahCarnahan:
Looks great thanks Ben

PR #14253, opened 2016-12-16
Setup two-way sync for keywords

gcapalbo:
First step in moving keywords from couch to postgres - setting up a two-way sync between the couch and sql models.

@kaapstorm 
@emord

### DIFF
--- a/corehq/apps/sms/models.py
+++ b/corehq/apps/sms/models.py
@@ -2320,8 +2320,8 @@ class Keyword(SyncSQLToCouchMixin, models.Model):
     # (i.e., it has a KeywordAction with action equal to ACTION_STRUCTURED_SMS)
     delimiter = models.CharField(max_length=126, null=True)
 
-    # If a SQLXFormsSession is open for a contact when they invoke this Keyword,
-    # override_open_sessions tells what to do with it. If True, the SQLXFormsSession
+    # If a SQLXFormsSession (i.e., an sms survey) is open for a contact when they invoke this
+    # Keyword, override_open_sessions tells what to do with it. If True, the SQLXFormsSession
     # will be closed and this Keyword will be invoked. If False, this Keyword will be
     # skipped and the form session handler will count the text as the next
     # answer in the open survey.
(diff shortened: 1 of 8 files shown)


corehq/apps/sms/tests/migration.py
```diff
@@ -8,13 +12,16 @@ class BaseMigrationTestCase(TestCase):
 
     def setUp(self):
         self.domain = 'test-sms-sql-migration'
-        self.deleteAllLogs()
+        self.deleteAllObjects()
```
emord: nit: It's always weird to me that we use this pattern. deleting objects in the setup seems like there are issues in other tests teardown that should be fixed?

emord (approved):
Looks good, just a few nitpicky comments. Great comments on the model, made it much easier to understand what you were migrating.

corehq/apps/sms/models.py
```diff
@@ -2295,4 +2296,111 @@ def has_migration_completed(cls, name):
...
+
+    # The recipient of this action is a user group (Group) with id given by
+    # recipient_id.
+    RECIPIENT_USER_GROUP = "USER_GROUP"
+
+    # The Keyword that this KeywordAction belongs to
+    keyword = models.ForeignKey('Keyword', on_delete=models.CASCADE)
+
+    # One of the ACTION_* constants representing the action to take
+    action = models.CharField(max_length=126)
```
gcapalbo: I've stopped using `choices` because django doesn't actually enforce that one of those values are selected before you save the model, and it doesn't add a constraint on the db side either in the migration. 

I think `choices` is only used on the UI side if you're using a `ModelForm` to create/edit the object. It does provide self-documentation of the choices though, but the other problem with it is every time you add a new choice, django prompts you that there are changes which require a migration, so we'd have to follow up each change to the choices list with a migration.

The 126 is because for a lot of these I'm not really sure what the max length should be, so I just go with the max possible length while having postgres still consider all strings entered to be "short strings". (see [here](https://www.postgresql.org/docs/9.3/static/datatype-character.html))

corehq/apps/sms/models.py
```diff
@@ -2295,4 +2296,111 @@ def has_migration_completed(cls, name):
...
+    # One of the RECIPIENT_* constants representing the recipient of this action
+    recipient = models.CharField(max_length=126)
+
+    # Represents the id of the recipient, when necessary
+    recipient_id = models.CharField(max_length=126, null=True)
+
+    # Only used for action == ACTION_SMS
+    message_content = models.TextField(null=True)
+
+    # Only used for action in [ACTION_SMS_SURVEY, ACTION_STRUCTURED_SMS]
```
gcapalbo: yeah fair point, i'll add some validation before save

corehq/apps/sms/tests/migration.py
gcapalbo: yeah i had added that a long time ago for another migration and i don't remember why - it isn't needed. i'll take it out

corehq/apps/sms/models.py
emord: Cool. Those all seem reasonable. Maybe we should have a constant for `PSQL_SHORT_STRING` or something?

gcapalbo:
@emord updated

corehq/apps/sms/models.py
gcapalbo: yeah good idea, i've done this in a bunch of places so i'll add it in another pr

corehq/apps/sms/models.py
```diff
@@ -2433,5 +2436,17 @@ class KeywordAction(models.Model):
     # This can be None in which case there is no separator (i.e., "report a100 b200")
     named_args_separator = models.CharField(max_length=126, null=True)
 
+    def save(self, *args, **kwargs):
+        if self.recipient == self.RECIPIENT_USER_GROUP and not self.recipient_id:
+            raise self.InvalidModelStateException("Expected a value for recipient_id")
+
+        if self.action == self.ACTION_SMS and not self.message_content:
```
emord: Is it worthwhile to check `if self.message_content and self.action != self.ACTION_SMS`? Similar for the other validations

gcapalbo: i didn't think so, it would just be ignored for the rest of the actions so it's not a big deal

emord (approved):

gcapalbo:
(approved and tests passed, merging)

PR #14255, opened 2016-12-16
Filter by domain even for user with access to all location

mkangia:
Originating from https://github.com/dimagi/commcare-hq/pull/14159. Extracted just the changes for issue 1 in that PR here.
@esoergel

### DIFF
--- a/corehq/apps/locations/models.py
+++ b/corehq/apps/locations/models.py
@@ -282,7 +282,7 @@ class LocationQueriesMixin(object):
 
     def accessible_to_user(self, domain, user):
         if user.has_permission(domain, 'access_all_locations'):
-            return self.all()
+            return self.filter(domain=domain)
 
         assigned_location_ids = user.get_location_ids(domain)
         if not assigned_location_ids:


corehq/apps/locations/models.py
```diff
@@ -282,7 +282,7 @@ def couch_locations(self, wrapped=True):
 
     def accessible_to_user(self, domain, user):
         if user.has_permission(domain, 'access_all_locations'):
-            return self.all()
+            return self.filter(domain=domain)
```
mkangia: @esoergel just thinking if the same should be done at https://github.com/dimagi/commcare-hq/pull/14255/files#diff-a2c9183455842303ef415bfcd036fdc3R290 for uniformity.

esoergel: `user.get_location_ids(domain)` should only return ids of locations in that domain (otherwise there's some other big problem somewhere), but it couldn't hurt as an additional safety measure.

mkangia: Agreed! `get_location_ids` shadows unwanted locations. Just that it would make things uniform...
for all access locations in domain and for restricted access the same set & assigned locations

PR #14259, opened 2016-12-16
ES reliability

emord:
@benrudolph @czue @proteusvacuum 

Mentioned this in the  meeting today. It's currently only informational, but I imagine a world where this could resave cases/applications/whatever else.

Ran it on softlayer and it completed more quickly than I expected. enikshay-test is missing quite  a few forms/cases from ES which is concerning.

### DIFF
--- a/corehq/apps/domain/calculations.py
+++ b/corehq/apps/domain/calculations.py
@@ -23,6 +23,7 @@ from corehq.apps.domain.models import Domain
 from corehq.apps.reminders.models import CaseReminderHandler
 from corehq.apps.users.models import CouchUser
 from corehq.elastic import es_query, ADD_TO_ES_FILTER
+from corehq.form_processor.interfaces.dbaccessors import CaseAccessors, FormAccessors
 from dimagi.utils.parsing import json_format_datetime
 
 
@@ -74,7 +75,9 @@ def active_mobile_users(domain, *args):
     return num_users if 'inactive' not in args else len(user_ids) - num_users
 
 
-def cases(domain, *args):
+def cases(domain, primary_db=False, *args):
+    if primary_db:
+        return CaseAccessors(domain).get_number_of_cases_in_domain()
     return get_number_of_cases_in_domain(domain)
 
 
@@ -114,7 +117,9 @@ def inactive_cases_in_last(domain, days):
     return data['hits']['total'] if data.get('hits') else 0
 
 
-def forms(domain, *args):
+def forms(domain, primary_db=False, *args):
+    if primary_db:
+        return FormAccessors(domain).get_number_of_forms_in_domain()
     return get_number_of_forms_in_domain(domain)
 
 
--- a/corehq/apps/es/cases.py
+++ b/corehq/apps/es/cases.py
@@ -21,6 +21,9 @@ from . import aggregations, filters
 
 class CaseES(HQESQuery):
     index = 'cases'
+    default_filters = {
+        'is_commcare_case': filters.doc_type("CommCareCase"),
+    }
 
     @property
     def builtin_filters(self):
--- a/corehq/apps/hqadmin/management/commands/check_docs_in_es.py
+++ b/corehq/apps/hqadmin/management/commands/check_docs_in_es.py
@@ -1,0 +1,57 @@
+import csv
+from django.core.management import BaseCommand
+
+from corehq.apps.app_manager.dbaccessors import get_app_ids_in_domain
+from corehq.apps.domain.calculations import cases, forms
+from corehq.apps.domain.models import Domain
+from corehq.apps.es.apps import AppES
+from corehq.apps.es.domains import DomainES
+from corehq.apps.es.groups import GroupES
+from corehq.apps.es.users import UserES
+from corehq.apps.groups.dbaccessors import get_group_ids_by_domain
+from corehq.apps.users.dbaccessors.all_commcare_users import get_all_user_ids_by_domain
+
+
+class Command(BaseCommand):
+    help = "Compares the number of documents in ES to primary DB to detect inconsistency"
+
+    def add_arguments(self, parser):
+        parser.add_argument('--filename', dest='filename', default='es_reliability.csv')
+
+    def handle(self, *args, **options):
+        num_es_domains = DomainES().count()
+        num_couch_domains = Domain.get_db().view("domain/domains").all()[0]['value']
+
+        if num_es_domains != num_couch_domains:
+            print str(num_es_domains) + " != " + str(num_couch_domains)
+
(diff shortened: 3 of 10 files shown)


corehq/apps/domain/calculations.py
```diff
@@ -74,7 +75,9 @@ def active_mobile_users(domain, *args):
     return num_users if 'inactive' not in args else len(user_ids) - num_users
 
 
-def cases(domain, *args):
+def cases(domain, primary_db=False, *args):
```
emord: These two methods aren't used right now (they were earlier in the PR). keeping htem here becasue I think they will fit in well for proactive alerting (to hopefully come soon)

corehq/form_processor/interfaces/dbaccessors.py
```diff
@@ -130,6 +130,9 @@ def form_exists(self, form_id):
     def get_all_form_ids_in_domain(self, doc_type='XFormInstance'):
         return self.db_accessor.get_form_ids_in_domain_by_type(self.domain, doc_type)
 
+    def get_number_of_forms_in_domain(self):
```
emord: This will preferably use count in the future

benrudolph:
sweet, this is cool. my one thing i'm still not sure about is that we actually don't expect cases/forms/etc to be exactly equal in couch and es at a given point in time. we expect that t - 60mins (or some interval) that es and couch will be the same (where t is now). if we were to run this on prod, there'd be a large number of false positives. pretty much any domain that is in the process of submitting a form. i know it makes the problem way harder, but otherwise i think it'd be too noisy to do on a daily basis. still great for debugging a domain though

emord:
@benrudolph yeah. I think that the number could have false positives, but its easy to filter out small differences in excel or something later, and look to see how many domains have large differences. enikshay on india had ~100 in sql, but not es

corehq/apps/hqadmin/management/commands/check_docs_in_es.py
```diff
@@ -0,0 +1,136 @@
...
+
+    def domain_info(self, domain):
+        DomainStats = namedtuple('DomainStats', [
+            'num_cases_es', 'num_forms_es', 'num_apps_es', 'num_users_es', 'num_groups_es',
+            'num_cases_primary', 'num_forms_primary', 'num_apps_primary', 'num_users_primary', 'num_groups_primary',
+            'app_ids_primary', 'user_ids_primary', 'group_ids_primary', 'case_ids_primary', 'form_ids_primary'
+        ])
+
+        case_ids_primary = set(CaseAccessors(domain).get_case_ids_in_domain())
+        form_ids_primary = set(FormAccessors(domain).get_all_form_ids_in_domain())
```
benrudolph: not sure how to get around this but this may OOM. maybe if it's an iterator you could incrementally write to a file

emord: Yeah. I'm running into some issues with big domains https://github.com/dimagi/commcare-hq/pull/14259/commits/68e461bf9c8ffe3ec00c53d97b072ffbcc9d6d62 :/ I don't think it's OOM, but it's not a clear error

czue:
took a quick look and this seems great. agree we'll need to tweak it when we try to use it as an active monitoring thing.

corehq/apps/es/cases.py
```diff
@@ -21,6 +21,9 @@
 
 class CaseES(HQESQuery):
     index = 'cases'
+    default_filters = {
+        'is_commcare_case': filters.doc_type("CommCareCase"),
```
emord: I'm not sure if this is the "right" thing to do. Currently anything that uses `CaseES` is returning all cases including ones that have been deleted. That doesn't seem like what we should be doing, and it's inconsistent with what couch/sql do by default.

corehq/form_processor/backends/sql/dbaccessors.py
```diff
@@ -311,6 +311,10 @@ def get_forms_by_type(domain, type_, limit, recent_first=False):
         return forms[:limit]
 
     @staticmethod
+    def get_number_of_forms_in_domain(domain, doc_type='XFormInstance'):
+        return len(FormAccessorSQL.get_form_ids_in_domain_by_type(domain, doc_type))
```
emord: These don't use `count`, but that's not really necessary as there's not any huge sql domains at the moment

corehq/form_processor/interfaces/dbaccessors.py
emord: Couch domains now use count/reduce

emord (commented):
dumped more context in http://manage.dimagi.com/default.asp?244248 for the next interrupt team

corehq/apps/es/cases.py
czue: the pillow should remove soft deletions. see [code](https://github.com/dimagi/commcare-hq/blob/master/corehq/ex-submodules/pillowtop/processors/elastic.py#L32) and [test](https://github.com/dimagi/commcare-hq/blob/master/testapps/test_pillowtop/tests/test_case_pillow.py#L47)

are there lots of `CommCareCase-Deleted` in the index?

czue: see my other comment, but i would have thought deleted cases would be removed from the index

czue:
build is failing
```
FAIL: test_case_pillow (testapps.test_pillowtop.tests.test_case_pillow.CasePillowTest)
  File "/mnt/commcare-hq-ro/corehq/util/test_utils.py", line 264, in inner
  File "/mnt/commcare-hq-ro/corehq/util/test_utils.py", line 237, in __call__
  File "/mnt/commcare-hq-ro/corehq/util/test_utils.py", line 252, in call_with_settings
    fn(*args, **kwargs)
  File "/mnt/commcare-hq-ro/corehq/util/test_utils.py", line 233, in fn_with_pre_and_post
    self.fn(*args, **kwargs)
  File "/mnt/commcare-hq-ro/testapps/test_pillowtop/tests/test_case_pillow.py", line 40, in test_case_pillow
    self.assertEqual(1, results.total)
AssertionError: 1 != 0
```

corehq/form_processor/backends/sql/dbaccessors.py
czue: my guess is we might want to address this before we add any automatic running of this

corehq/apps/es/cases.py
emord: Hmm. It does appear like that's the case, but I ran into a couple of deleted cases when I was running this command. I don't have a great idea on why that is though. Possibly another bug or inconsistency in pillow top?

emord:
superseded by #14654

PR #14260, opened 2016-12-16
Better error message in RestorePermissionDenied

proteusvacuum:


corehq/apps/ota/utils.py
```diff
@@ -109,7 +109,11 @@ def is_permitted_to_restore(domain, couch_user, as_user, has_data_cleanup_privil
         if as_user is not None and not _restoring_as_yourself(couch_user, as_user):
             as_user_obj = CouchUser.get_by_username(as_user)
             if not as_user_obj:
-                raise RestorePermissionDenied(_(u'Invalid restore as user {}').format(as_user))
+                if '.commcarehq.org' not in as_user:
```
benrudolph: this is not always true, you can restore as a web user so this error message wouldn't necessarily make sense

benrudolph: that being said, i'm not that opposed to it given that most restore as calls are for mobile users

czue:
build failed
```
FAIL: test_web_user_as_user_bad_domain (corehq.apps.ota.tests.test_utils.RestorePermissionsTest)
  File "/mnt/commcare-hq-ro/corehq/apps/ota/tests/test_utils.py", line 154, in test_web_user_as_user_bad_domain
    self.assertRegexpMatches(message, 'Invalid restore as user')
AssertionError: Regexp didn't match: 'Invalid restore as user' not found in u'User not found. Usernames should be of the form <user>@<domain>.commcarehq.org'
```

PR #14266, opened 2016-12-16
Allow releases table to scroll horizontally

orangejenny:
http://manage.dimagi.com/default.asp?243237

Tables won't shrink beyond what their content allows, so allow scrolling on the table's container.

@biyeun / @proteusvacuum


proteusvacuum (approved):
Yay!! 
I think a large part of the issue is also that half the screen is taken up by the (non collapsible) sidebar, but that might be a harder thing to change.

orangejenny:
Yeah, in general full screen mode would be nice within app manager if you're on mobile...but that is indeed a slightly bigger design question.

czue:
yes!

PR #13293, opened 2016-09-20
Nikshay migration

nickpell:
For review

@proteusvacuum @czue 
- [x] add tests

PR #14267, opened 2016-12-16
more case properties

nickpell:
small updates
@czue @proteusvacuum 
cc: @calellowitz @mkangia

### DIFF
--- a/custom/enikshay/nikshay_datamigration/factory.py
+++ b/custom/enikshay/nikshay_datamigration/factory.py
@@ -105,7 +105,7 @@ class EnikshayCaseFactory(object):
                 'case_type': 'occurrence',
                 'update': {
                     'hiv_status': outcome.HIVStatus,
-                    'nikshay_id': outcome.PatientId.PregId,
+                    'nikshay_id': self.patient_detail.PregId,
 
                     'migration_created_case': True,
                 },
--- a/custom/enikshay/nikshay_datamigration/factory.py
+++ b/custom/enikshay/nikshay_datamigration/factory.py
@@ -69,7 +69,7 @@ class EnikshayCaseFactory(object):
                     'current_address': self.patient_detail.paddress,
                     'current_address_district_choice': self.patient_detail.Dtocode,
                     'current_address_state_choice': self.patient_detail.scode,
-                    'dob_known': False,
+                    'dob_known': 'no',
                     'first_name': self.patient_detail.first_name,
                     'last_name': self.patient_detail.last_name,
                     'middle_name': self.patient_detail.middle_name,
--- a/custom/enikshay/nikshay_datamigration/factory.py
+++ b/custom/enikshay/nikshay_datamigration/factory.py
@@ -107,7 +107,9 @@ class EnikshayCaseFactory(object):
                 'case_type': 'occurrence',
                 'update': {
                     'hiv_status': outcome.HIVStatus,
+                    'name': 'Occurrence #1',
                     'nikshay_id': self.patient_detail.PregId,
+                    'occurrence_episode_count': 1,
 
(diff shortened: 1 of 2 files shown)


custom/enikshay/nikshay_datamigration/factory.py
```diff
@@ -63,11 +63,13 @@ def person(self):
                 # 'owner_id': self._location.location_id,
                 'update': {
                     'aadhaar_number': self.patient_detail.paadharno,
+                    'age': self.patient_detail.page,
                     'age_entered': self.patient_detail.page,
                     'contact_phone_number': validate_number(self.patient_detail.pmob),
                     'current_address': self.patient_detail.paddress,
                     'current_address_district_choice': self.patient_detail.Dtocode,
                     'current_address_state_choice': self.patient_detail.scode,
+                    'dob_known': False,
```
proteusvacuum: I think this is set to "no" in the app

proteusvacuum: (would be good to check the other boolean values for what they are set to)

proteusvacuum: Also, I think when this is false, we still calculate a DOB in the app, so would be good to do that here too. (for [eg check this case](https://india.commcarehq.org/a/enikshay-test/reports/case_data/45953ec6-304a-4c30-984a-05e6840dfde9/).)

proteusvacuum (changes requested):

custom/enikshay/nikshay_datamigration/factory.py
nickpell: ah good catch

nickpell: > I think this is set to "no" in the app

fixed

> (would be good to check the other boolean values for what they are set to)

The other booleans are all set by the migration (not the app).  I'll follow up on https://github.com/dimagi/commcare-hq/pull/13293/files/d28dab894c613df238ab727085f226a57d453434#r92884819 in a separate PR

> Also, I think when this is false, we still calculate a DOB in the app, so would be good to do that here too. (for eg check this case.)

Question: since ```dob``` never gets loaded in the app, is it needed?  I suppose it might be used in reports or elsewhere.

proteusvacuum: > The other booleans are all set by the migration (not the app)

I see. I think it would still be good to map `True` and `False` to `('true', 'false') ` (or `('yes', 'no')`) because those are the conventions we've been using in the app. 

> Question: since dob never gets loaded in the app, is it needed? I suppose it might be used in reports or elsewhere.

Looks to me like `dob` is used in [some places](https://india.commcarehq.org/a/enikshay-test/apps/view/e6c88dc7cf8c742cbf397eff07d9a245/modules-7/forms-0/source/). Take a look at how the "age" hidden value is calculated there, and how "dob" is fetched. (I'm not saying this is good...)

I'd be wary of going off the form summary for figuring out where case properties are used as we have a lot of casedb lookups that are happening in the app. I've been unzipping and grepping through the app build files to find references to things if I'm not sure.

proteusvacuum (approved):

custom/enikshay/nikshay_datamigration/factory.py
nickpell: > I see. I think it would still be good to map True and False to ('true', 'false') (or ('yes', 'no')) because those are the conventions we've been using in the app.

https://github.com/dimagi/commcare-hq/pull/14296

nickpell: https://github.com/dimagi/commcare-hq/pull/14338

PR #14268, opened 2016-12-16
update Outcome model

nickpell:
Instead of trying to migrate ```CharField```s to ```DateTimeFields```, ```IntegerField```s etc, I'm just dropping the ```Outcome``` table and adding it back with the right columns.  Then I can re-import the table.

@czue @proteusvacuum

### DIFF
--- a/custom/enikshay/nikshay_datamigration/models.py
+++ b/custom/enikshay/nikshay_datamigration/models.py
@@ -153,20 +153,53 @@ class PatientDetail(models.Model):
 
 class Outcome(models.Model):
     PatientId = models.OneToOneField(PatientDetail, primary_key=True)
-    Outcome = models.CharField(max_length=255)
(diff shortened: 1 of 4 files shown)


custom/enikshay/nikshay_datamigration/models.py
```diff
@@ -153,20 +153,53 @@ def _list_of_dot_names(self):
...
+            ('1', '1'),
+            ('2', '2'),
+            ('3', '3'),
+            ('4', '4'),
+            ('5', '5'),
+            ('6', '6'),
+            ('7', '7'),
+        )
+    )
+    OutcomeDate = models.CharField(max_length=255, null=True)  # somethings DD/MM/YYYY, sometimes DD-MM-YYYY
```
czue: of course.

czue (approved):
makes sense

czue:
(not looking closely at the new model, but this seems like a good approach)

PR #14270, opened 2016-12-16
fix typo

nickpell:
This might affect the behavior of your loggers, but I'm assuming this was meant to be set to ```"propagate"``` @benrudolph @emord @snopoke 

@proteusvacuum thanks for noticing


benrudolph:
lol... and this is how spelling mistakes propogate

PR #14271, opened 2016-12-17
ccinternal --> internal

czue:
from https://github.com/dimagi/commcare-hq/pull/14262

searched repo for ccinternal and found these

@benrudolph


benrudolph:
thanks!

PR #14277, opened 2016-12-19
Change start/stop syntax

gcapalbo:
Changes the syntax to start/stop reminders from "start" and "stop" to "reminder on" and "reminder off" for EWS.


gcapalbo:
code buddies @orangejenny @biyeun

custom/ewsghana/handlers/reminder.py
```diff
@@ -0,0 +1,25 @@
...
+class ReminderOnOffHandler(KeywordHandler):
+
+    def help(self):
+        if user_needs_reminders(self.user):
+            self.respond(DEACTIVATE_REMINDERS)
+        else:
+            self.respond(REACTIVATE_REMINDERS)
+
+    def handle(self):
+        action = self.args[0].lower()
```
orangejenny: Could be extra-defensive here and show help if no argument is provided.

orangejenny (approved):

custom/ewsghana/handlers/reminder.py
gcapalbo: @orangejenny yep, the way it works behind the scenes is that for each of these `KeywordHandler`s, if no args are present, it automatically invokes `.help()` instead of `.handle()`

orangejenny: Ah, I missed that. Nice!

gcapalbo:
(approved and tests passed, merging)

PR #14278, opened 2016-12-19
Escape warning on deploy page

orangejenny:
http://manage.dimagi.com/default.asp?244192

@biyeun / @gcapalbo

### DIFF
--- a/corehq/apps/app_manager/templates/app_manager/v1/app_view.html
+++ b/corehq/apps/app_manager/templates/app_manager/v1/app_view.html
@@ -43,7 +43,7 @@
                 permissions: {
                     cloudcare: {{ is_cloudcare_allowed|BOOL }}
                 },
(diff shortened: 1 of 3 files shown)


gcapalbo:
@orangejenny was the issue happening because the french translation had characters that needed to be escaped?

if so, was just curious if applying escapejs in this way worked to resolve it. i would have thought you'd have to do something similar to [this](http://stackoverflow.com/questions/22469235/complicated-quote-escaping-in-django-template-translations)

orangejenny:
Aww, seeing Tyler on SO makes me a little nostalgic. The combined trans & escapejs is a pattern we use a handful of other places, but I did test the code and it fixed the problem (which, indeed, was an apostrophe in the French translation).

gcapalbo:
ok, thanks for confirming

gcapalbo (approved):

PR #14261, opened 2016-12-16
Remove deprecated case importer features

dannyroberts:

PR #14254, opened 2016-12-16
Case importer field mapping improvements

dannyroberts:
- Adds a notion of special fields, statically defined (like `owner_name`, etc.)
- Shows description of special fields in the UI if you use them
- Admonishes you if you use a non "discoverable" property (currently just `parent_ref`)
- Adds a "Did you mean?" feature for the case field, which suggests any case property or discoverable special field
- Switches interactive part of page to knockout
- Gets rid of jumpiness in case field input box
- Small, but make 'name' show in the menu and be recognized as a known property
- Autofill by default (without you having to press autofill); change name to "Set to default".

screen shot 2016-12-15 at 10 36 50 pm

EDIT Dec 19, 2016: The final version of this feature also

- makes 'external_id' a special field in the UI (so it comes up with a description)
- makes the "did you mean" functionality case-insensitive (so it'll still suggest "owner_name" for "OWNER NAME")

PR #14280, opened 2016-12-19
remove unused is_date_field value from html

dannyroberts:
Small cleanup step that didn't fit into https://github.com/dimagi/commcare-hq/pull/14261 because it's on top of https://github.com/dimagi/commcare-hq/pull/14254


czue (approved):
tests aren't running - maybe because the base branched changed midway through PR?

esoergel (approved):

esoergel:
(Merging 'cause I assume tests won't hit this - it's all html template JS, the js tests were in a module)

PR #14281, opened 2016-12-19
Reduce bulk payload size for ES

emord:
This is my theory for why a lot of apps are missing from ES. Apps tend to be the largest docs we have and attempting to save 100 MB of apps tends to timeout on our ES. We do retry the request, but it will often fail again, and the reindex will exit. I noticed this while doing a manual index of apps.

EDIT: There were > 7k apps missing from ES

I found this email from Oct 24 (when the last app reindex was):

```
  File "/home/cchq/www/production/releases/2016-10-24_19.20/corehq/apps/hqcase/management/commands/ptop_preindex.py", line 130, in handle
    job.get()
  File "/home/cchq/www/production/releases/2016-10-24_19.20/python_env/local/lib/python2.7/site-packages/gevent/greenlet.py", line 251, in get
    raise self._exception
BulkProcessingFailed: Processing batch failed
```

It doesn't really give info, but I think it's probably due to this.

Because 10 MB retries have (from my experience) 100% success rate, the estimated time remaining went from 2.5 hours to 1.25 hours

@czue guessing you have the most context on pillowtop. Does this seem like a likely explanation to you?

@orangejenny buddy @proteusvacuum

### DIFF
--- a/corehq/ex-submodules/pillowtop/reindexer/reindexer.py
+++ b/corehq/ex-submodules/pillowtop/reindexer/reindexer.py
@@ -14,6 +14,7 @@ from pillowtop.utils import prepare_bulk_payloads, build_bulk_payload, ErrorColl
 
 MAX_TRIES = 3
 RETRY_TIME_DELAY_FACTOR = 15
+MAX_PAYLOAD_SIZE = 10 ** 7  # ~10 MB
 
 
 class Reindexer(six.with_metaclass(ABCMeta)):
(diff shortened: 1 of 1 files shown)


orangejenny:
Code looks reasonable to me, will give @czue a chance to comment.

The fact that we didn't notice the failure during the last reindex seems like a possible process failure. Should this failure make a louder noise, or should we have different/clearer expectations around watching error emails when a reindex happens?

czue:
> The fact that we didn't notice the failure during the last reindex seems like a possible process failure. Should this failure make a louder noise, or should we have different/clearer expectations around watching error emails when a reindex happens?

:+1: to this

benrudolph:
nice find. probably out of scope for this, but it would be cool if our reindexer could detect this and retry with a smaller payload

PR #14284, opened 2016-12-19
Case import history link forms

dannyroberts:
Keep track of the ids of the forms and cases involved in an import. Expose this information at secret URLs.

There's no way currently to correlate rows in excel files to cases created or updated, especially when importing new cases (where there's no id yet, so no id column), and that can be very frustrating. This PR adds the bookkeeping behind the scenes that would let you see what cases were touched (and what forms were used to touch them), but just exposes it as plain text at a hidden url. A simple next step would be to add links from the case import history under a feature flag, so that people know about and want to use it in its low-polish form can.

@czue @snopoke There's definitely a space/usage question here. My gut since it's just storing ids, it's not a huge deal to store this in postgresql. I imagine storage associated with this feature would grow at about the same rate as the case upload attachments in riak (since there's one row per case in the attachments) but with a much lower constant factor. Since just like those attachments, there's a lot of pruning we can do of old data, my thinking is that we can revisit when we revisit the case upload attachments in riak, and add pruning to both using the same rules when we do that.


dannyroberts:
@esoergel buddy

czue (commented):
code looks good.

I don't totally understand the need for `CaseUploadCaseRecord` at all. Can we not just grab the form and parse the case IDs just in time using the same code we would to extract them for case processing? I realize that would be a bit slower, but I can't imagine it's the type of operation we need to make fast. I think that also alleviates the storage concerns.

dannyroberts:
That's totally fair. The only thing that came to mind is preserving the order, but case importer is the one building the form, so we control the order of the case blocks. If whatever utility pulls out case ids from a form doesn't preserve order for some reason, all we'd have to do is fix that utility, and we'd be able to insure that the order of the case ids is the same order as the rows in your upload. I think I just talked myself into your suggestion. I'll try it and see if anything unexpected comes up; seems like it should work the way you described.

dannyroberts:
@czue thanks for the suggestion. The way I was doing it wasn't even preserving order, and I just hadn't tested that. After adding the test, it was pretty easy to switch to extracting case ids from the form, and the test verifies that they end up in original order.

czue (approved):
looks good!

czue:
strange build failure, kicked but not sure if it was real (if so seems a weird dependency)

esoergel:
Still getting a weird test failure:
```
ERROR: test_checkpoint_with_multiple_topics (corehq.apps.change_feed.tests.test_kafka_change_feed.KafkaCheckpointTest)
  File "/mnt/commcare-hq-ro/corehq/util/decorators.py", line 24, in decorated
    return fn(*args, **kwds)
  File "/mnt/commcare-hq-ro/corehq/apps/change_feed/tests/test_kafka_change_feed.py", line 83, in test_checkpoint_with_multiple_topics
    checkpoint=checkpoint, checkpoint_frequency=1, change_feed=feed
  File "/mnt/commcare-hq-ro/corehq/apps/change_feed/consumer/feed.py", line 149, in __init__
    checkpoint_doc = self.checkpoint.get_or_create_wrapped()
  File "corehq/ex-submodules/pillowtop/checkpoints/manager.py", line 42, in get_or_create_wrapped
    checkpoint = get_or_create_checkpoint(self.checkpoint_id)
  File "corehq/ex-submodules/pillowtop/checkpoints/manager.py", line 22, in get_or_create_checkpoint
    'timestamp': datetime.utcnow()
  File "/vendor/lib/python2.7/site-packages/django/db/models/manager.py", line 127, in manager_method
    return getattr(self.get_queryset(), name)(*args, **kwargs)
  File "/vendor/lib/python2.7/site-packages/django/db/models/query.py", line 405, in get_or_create
    return self.get(**lookup), False
  File "/vendor/lib/python2.7/site-packages/django/db/models/query.py", line 328, in get
    num = len(clone)
  File "/vendor/lib/python2.7/site-packages/django/db/models/query.py", line 144, in __len__
    self._fetch_all()
  File "/vendor/lib/python2.7/site-packages/django/db/models/query.py", line 965, in _fetch_all
    self._result_cache = list(self.iterator())
  File "/vendor/lib/python2.7/site-packages/django/db/models/query.py", line 238, in iterator
    results = compiler.execute_sql()
  File "/vendor/lib/python2.7/site-packages/django/db/models/sql/compiler.py", line 838, in execute_sql
    cursor = self.connection.cursor()
  File "/vendor/lib/python2.7/site-packages/django/db/backends/base/base.py", line 164, in cursor
    cursor = self.make_cursor(self._cursor())
  File "/vendor/lib/python2.7/site-packages/django/db/backends/base/base.py", line 137, in _cursor
    return self.create_cursor()
  File "/vendor/lib/python2.7/site-packages/django/db/utils.py", line 98, in __exit__
  File "/vendor/lib/python2.7/site-packages/django/db/backends/base/base.py", line 137, in _cursor
    return self.create_cursor()
  File "/vendor/lib/python2.7/site-packages/django/db/backends/postgresql_psycopg2/base.py", line 212, in create_cursor
    cursor = self.connection.cursor()
InterfaceError: connection already closed
```
Only thing I could guess is that it might be related to the `transaction.atomic`, although I still don't see how.

dannyroberts:
✅ It was that a `tearDownClass` wasn't calling super, which manages the connection.

czue:
> It was that a `tearDownClass` wasn't calling super, which manages the connection.

thanks for leaving this here. just ran into a similar thing and found the answer right in my email!

czue:
Hi @dannyroberts :wave: . I just came back to this PR to write a thank you for documenting this and saw that I already thanked you 16 months ago. The levels of consistency I have demonstrated here (causing the same error for myself *again*, looking in the same place for a solution *again*, and coming back to the PR to say thinks *again*) have left me feeling strange.

While I'm here....

Hi @czue from 2019 :wave: . You're here again to say thanks for documenting this random error with forgetting to call `teardownClass` in a `TestCase`, right? Well, no need, you've already done it twice! Hope the future is treating you well. If there are self-driving cars yet you should totally buy one! :robot:

PR #14285, opened 2016-12-19
update MediaUploader

orangejenny:


orangejenny:
@biyeun

orangejenny:
tests pass => self-merging

PR #14287, opened 2016-12-19
don't use $.load

orangejenny:
jQuery 3 issue. Throws a js error but doesn't appear to cause problems (I can still interact with live preview).

@biyeun / @gcapalbo


biyeun (approved):

orangejenny:
no test coverage, merging

PR #11381, opened 2016-04-22
clone domain additions

snopoke:
based on Matt's feedback https://docs.google.com/document/d/1mm2wI3Ax0xxvXnB_ZuY5quKquiDuLR9-KDBT3gMFBnA/edit

@czue

PR #14290, opened 2016-12-20
Remove keywords and reminders from clone domain

gcapalbo:
Looks like this was added in https://github.com/dimagi/commcare-hq/pull/11381

Copying over reminders and keywords to a new domain is a bit more involved than just copying the data as you have to also resolve any new references to form unique ids and resolve any new references to contact ids like group or user ids. I think there might not have been an issue in the original use of this for the migration to the ICDS SQL domain because it didn't have any reminders or keywords that had those kinds of references.

@snopoke @czue I'm proposing removing copying of reminders and keywords from this command for now. Can we leave it out or do we still need this?

fyi @emord

### DIFF
--- a/corehq/apps/hqadmin/management/commands/clone_domain.py
+++ b/corehq/apps/hqadmin/management/commands/clone_domain.py
@@ -21,8 +21,6 @@ types = [
     'apps',
     'user_fields',
     'user_roles',
-    'reminders',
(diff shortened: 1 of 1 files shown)


czue (approved):
sure I think this is fine. if we need this functionality again we can think more deliberately about bringing it back.

PR #14298, opened 2016-12-20
perf optimization

nickpell:
Hey, let's not write O(N^2) code.

Expecting to get the most benefit by calling ```get_nikshay_ids_to_preexisting_nikshay_person_cases``` once

@proteusvacuum @czue

### DIFF
--- a/custom/enikshay/nikshay_datamigration/factory.py
+++ b/custom/enikshay/nikshay_datamigration/factory.py
@@ -244,7 +244,7 @@ class EnikshayCaseFactory(object):
         return '-'.join(self.patient_detail.PregId.split('-')[:4])
 
 
-def nikshay_code_to_location(domain):
+def get_nikshay_codes_to_location(domain):
     return {
         location.metadata.get('nikshay_code'): location
         for location in SQLLocation.objects.filter(domain=domain)
@@ -252,7 +252,7 @@ def nikshay_code_to_location(domain):
     }
 
 
-def nikshay_id_to_preexisting_nikshay_person_cases(domain):
+def get_nikshay_ids_to_preexisting_nikshay_person_cases(domain):
     case_accessor = CaseAccessors(domain)
     return {
         person_case.dynamic_case_properties()['nikshay_id']: person_case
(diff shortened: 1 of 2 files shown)


custom/enikshay/nikshay_datamigration/factory.py
```diff
@@ -248,17 +240,27 @@ def _followups(self):
...
+        location.metadata.get('nikshay_code'): location
+        for location in SQLLocation.objects.filter(domain=domain)
+        if 'nikshay_code' in location.metadata
+    }
+
+
+def get_nikshay_ids_to_preexisting_nikshay_person_cases(domain):
+    case_accessor = CaseAccessors(domain)
+    return {
+        person_case.dynamic_case_properties()['nikshay_id']: person_case
```
czue: still think we should also do the external id thing here

czue:
hahaha this is ridiculous. i suspect this will make the migration way way faster.

custom/enikshay/nikshay_datamigration/management/commands/create_enikshay_cases.py
```diff
@@ -26,10 +27,17 @@ def handle(self, domain, **options):
         num_succeeded = 0
         num_failed = 0
         logger.info('Starting migration of %d patient cases on domain %s.' % (total, domain))
+        nikshay_codes_to_location = get_nikshay_codes_to_location(domain)
+        nikshay_ids_to_preexisting_nikshay_person_cases = get_nikshay_ids_to_preexisting_nikshay_person_cases(
+            domain
+        )
```
czue: you never have the same nikshay ID twice in the same file right? otherwise we'd have to update or invalidate this when we add people.

czue (approved):

custom/enikshay/nikshay_datamigration/management/commands/create_enikshay_cases.py
proteusvacuum: This is ensured when the data is first added into the postgres tables

custom/enikshay/nikshay_datamigration/factory.py
nickpell: https://github.com/dimagi/commcare-hq/pull/14307

PR #14299, opened 2016-12-21
don't pass args that don't exist

nickpell:
```copy_ucr_data``` does not take any args

@mkangia @calellowitz


czue:
funny, wonder how this ever worked.

PR #14309, opened 2016-12-21
Show full path by default in UCR location filter

esoergel:
You can still disable it if desired.
@czue @noahcarnahan
Cory, I think we discussed this before and there was an argument that the full path display may be too long for some applications.  Is this an acceptable compromise.

### DIFF
--- a/corehq/apps/userreports/README.md
+++ b/corehq/apps/userreports/README.md
@@ -1181,7 +1181,10 @@ user                 | Select a user
 owner                | Select a possible case owner owner (user, group, or location)
 
 
-Location choice providers also support an "include_descendants" property to include descendant locations in the results, which defaults to `false`.
+Location choice providers also support two additional configuration options:
+
+* "include_descendants" - Include descendant locations in the results. Defaults to `false`.
+* "show_full_path" - display the full path to the location in the filter.  Defaults to `true`.
 
 Example assuming "village" is a location ID, which is converted to names using the location `choice_provider`:
 ```json
@@ -1193,7 +1196,8 @@ Example assuming "village" is a location ID, which is converted to names using t
   "datatype": "string",
   "choice_provider": {
       "type": "location",
-      "include_descendants": false
+      "include_descendants": false,
+      "show_full_path": true
   }
 }
 ```
@@ -1326,13 +1330,13 @@ Expanded columns have a type of `"expanded"`. Expanded columns will be "expanded
 
 If you have a data source like this:
 ```
-+---------+----------+-------------+
++---------|----------|-------------+
 | Patient | district | test_result |
-+---------+----------+-------------+
++---------|----------|-------------+
 | Joe     | North    | positive    |
 | Bob     | North    | positive    |
 | Fred    | South    | negative    |
-+---------+----------+-------------+
++---------|----------|-------------+
 ```
 and a report configuration like this:
 ```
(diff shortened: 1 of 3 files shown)


corehq/apps/userreports/README.md
```diff
@@ -1326,13 +1330,13 @@ Expanded columns have a type of `"expanded"`. Expanded columns will be "expanded
 
 If you have a data source like this:
 ```
-+---------+----------+-------------+
++---------|----------|-------------+
```
esoergel: Emacs insists on changing this automatically every time I save the file.

czue: ha. and here i thought you were a vim guy.

czue (commented):
this changes the behavior of existing reports right? maybe we should make the default false?

czue:
build fail is real
```
FAIL: test_query_no_search_all (corehq.apps.userreports.tests.test_choice_provider.LocationChoiceProviderTest)
  File "/mnt/commcare-hq-ro/corehq/apps/userreports/tests/test_choice_provider.py", line 110, in test_query_no_search_all
    self._test_query(self.choice_query_context('', limit=20, page=0))
  File "/mnt/commcare-hq-ro/corehq/apps/userreports/tests/test_choice_provider.py", line 101, in _test_query
    self.static_choice_provider.query(query_context))
AssertionError: Lists differ: [Choice(value=u'9dd6f9d75bc362... != [Choice(value=u'9dd6f9d75bc362...
First differing element 0:
Choice(value=u'9dd6f9d75bc36216b196f5bdf9d024f7', display=u'Massachusetts/Suffolke/Bostone')
Choice(value=u'9dd6f9d75bc36216b196f5bdf9d05964', display=u'Massachusetts')
```

corehq/apps/userreports/README.md
esoergel: `evil-mode` is where it's at!

esoergel:
Tests should be fixed now (I cherry-picked this from a bigger branch and forgot to test in isolation)

> this changes the behavior of existing reports right? maybe we should make the default false?

It changes the way locations are displayed in the filter.  Where people used to see:

> Boston [city]
Cambridge [city]
Massachusetts [state]
Middlesex [county]
Somerville [city]
Suffolk [county]

They'll now see:
> Massachusetts/Suffolk/Boston
Massachusetts/Middlesex/Cambridge
Massachusetts
Massachusetts/Middlesex
Massachusetts/Middlesex/Somerville
Massachusetts/Suffolk

Note that the sort order is a little wacky either way, which I'm attempting to address elsewhere.  This shouldn't affect the data in the report, it's just a (in my opinion) clearer display in the filter, matching the way this is displayed in the Case List and other location-aware reports.
My thinking was that since it's mostly a cosmetic change, it would be okay to upgrade the behavior of existing reports too.  Do you agree with that or should I not affect existing reports?

czue:
hmm, where can I see this filter in action today? I want to see how it looks on the TB domain which will be affected. our hierarchy has 7 levels and that format starts to get pretty unweildy.

esoergel:
@czue the case list report is your best bet.  You can either scroll down through the options or type to autocomplete.  I just looked at the TB domain and it appears that the location naming scheme includes the path anyways, so this may be a bit redundant for them.

Right now the UCR filter uses `filter_path_by_user_input`, but then doesn't display the name, so it will filter by parents, but that's hidden from view (if you typed "Mass", my above examples wouldn't change).

Another thing we could do for usability would be to match only against name and site code if path display is disabled.

czue:
sorry just trying to be paranoid - the other thing i think would be an issue is if this impacts mobile UCR filter select screens, which I think ICDS is using. do you know if it does? am guessing these would be way too long to show on mobile.

generally am a bit worried about changing defaults out from under people and not fully thinking through consequences so still think that the safer path would be to keep the existing behavior the default or write a migration that compensates for it, but if you think you can reach high confidence this won't have issues then we can go with it.

esoergel:
Those are valid concerns.  I do believe that this would make a better default, but I imagine people generally just copy/paste from the README anyways, so that's a sort-of default.  The safety of defaulting to False wins against the unsurety of affecting existing reports.

PR #14318, opened 2016-12-22
assign owner_id

nickpell:
@proteusvacuum @czue

### DIFF
--- a/custom/enikshay/nikshay_datamigration/factory.py
+++ b/custom/enikshay/nikshay_datamigration/factory.py
@@ -127,6 +127,9 @@ class EnikshayCaseFactory(object):
 
         if self._location:
             kwargs['attrs']['owner_id'] = self._location.location_id
+        else:
+            kwargs['attrs']['update']['migration_error'] = 'location_not_found'
+            kwargs['attrs']['update']['migration_error_details'] = self._nikshay_code
 
         if self.patient_detail.paadharno is not None:
             kwargs['attrs']['update']['aadhaar_number'] = self.patient_detail.paadharno
--- a/custom/enikshay/nikshay_datamigration/factory.py
+++ b/custom/enikshay/nikshay_datamigration/factory.py
@@ -97,7 +97,6 @@ class EnikshayCaseFactory(object):
             'attrs': {
                 'case_type': PERSON_CASE_TYPE,
                 'external_id': self.nikshay_id,
-                'owner_id': self._location.location_id,
                 'update': {
                     'age': self.patient_detail.page,
                     'age_entered': self.patient_detail.page,
@@ -125,6 +124,9 @@ class EnikshayCaseFactory(object):
                 },
             },
         }
+
+        if self._location:
+            kwargs['attrs']['owner_id'] = self._location.location_id
 
         if self.patient_detail.paadharno is not None:
             kwargs['attrs']['update']['aadhaar_number'] = self.patient_detail.paadharno
@@ -273,7 +275,7 @@ class EnikshayCaseFactory(object):
 
     @property
     def _location(self):
-        return self.nikshay_codes_to_location[self._nikshay_code]
+        return self.nikshay_codes_to_location.get(self._nikshay_code)
 
     @property
     def _nikshay_code(self):
--- a/custom/enikshay/nikshay_datamigration/factory.py
+++ b/custom/enikshay/nikshay_datamigration/factory.py
@@ -126,6 +126,7 @@ class EnikshayCaseFactory(object):
         }
 
         if self._location:
+            assert self._location.location_type.name == 'phi'
             kwargs['attrs']['owner_id'] = self._location.location_id
         else:
             kwargs['attrs']['owner_id'] = '_archive_'
--- a/custom/enikshay/nikshay_datamigration/factory.py
+++ b/custom/enikshay/nikshay_datamigration/factory.py
@@ -152,6 +152,7 @@ class EnikshayCaseFactory(object):
         kwargs = {
             'attrs': {
                 'case_type': OCCURRENCE_CASE_TYPE,
+                'owner_id': '-',
                 'update': {
                     'current_episode_type': 'confirmed_tb',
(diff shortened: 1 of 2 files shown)


custom/enikshay/nikshay_datamigration/tests/test_create_enikshay_cases.py
```diff
@@ -263,6 +263,18 @@ def test_case_update(self):
...
 
+    @run_with_all_backends
+    def test_location_not_found(self):
+        self.loc.delete()
+        call_command('create_enikshay_cases', self.domain.name)
+
+        person_case_ids = self.case_accessor.get_case_ids_in_domain(type='person')
+        self.assertEqual(1, len(person_case_ids))
+        person_case = self.case_accessor.get_case(person_case_ids[0])
+        self.assertEqual(person_case.owner_id, '')
```
nickpell: is blank ```owner_id``` the right thing to do when there's no location? @proteusvacuum

proteusvacuum: Hmm. I think it might be better to set it to `'_archive_'` and set `person.archive_reason = 'migration_location_not_found'`

custom/enikshay/nikshay_datamigration/factory.py
```diff
@@ -129,6 +128,12 @@ def get_person_case_structure(self):
             },
         }
 
+        if self._location:
```
proteusvacuum: I think we should check that this location has location_type "phi", otherwise it should fail.

proteusvacuum (changes requested):
Coming along! 
I think you also need to set all the owner ids of "extension" cases (`episode`, `occurrence`, `test`) to `'-'`

custom/enikshay/nikshay_datamigration/factory.py
nickpell: does ```'_archive_'``` have any special meaning in HQ?

proteusvacuum: Kind of. It is a "[weird owner id](https://github.com/dimagi/commcare-hq/blob/db17403a17c815ed663943e90afca3d5032d2c07/corehq/apps/users/util.py#L19-L26)" so we don't do stuff in the restore with those IDs. 
We are using it elsewhere in the eNikshay app to signify "a closed case that we might want to open in the app after searching for it with case claim at some point".

nickpell: added an assert

custom/enikshay/nikshay_datamigration/tests/test_create_enikshay_cases.py
```diff
@@ -263,6 +263,19 @@ def test_case_update(self):
...
+    @run_with_all_backends
+    def test_location_not_found(self):
+        self.loc.delete()
+        call_command('create_enikshay_cases', self.domain.name)
+
+        person_case_ids = self.case_accessor.get_case_ids_in_domain(type='person')
+        self.assertEqual(1, len(person_case_ids))
+        person_case = self.case_accessor.get_case(person_case_ids[0])
+        self.assertEqual(person_case.owner_id, '_archive_')
+        self.assertEqual(person_case.dynamic_case_properties()['migration_archive_reason'], 'migration_location_not_found')
```
dimagimon: E501 line too long (123 > 115 characters)

custom/enikshay/nikshay_datamigration/tests/test_create_enikshay_cases.py
nickpell: added ```'_archive_'``` and set an archive reason (used the prefix ```migration_``` to be consistent with other case properties added by the migration)

czue:
fyi conflicts

custom/enikshay/nikshay_datamigration/factory.py
```diff
@@ -129,6 +128,15 @@ def get_person_case_structure(self):
...
+            assert self._location.location_type.name == 'phi'
+            kwargs['attrs']['owner_id'] = self._location.location_id
+        else:
+            kwargs['attrs']['owner_id'] = '_archive_'
+            kwargs['attrs']['update']['migration_archive_reason'] = 'migration_location_not_found'
```
proteusvacuum: this should still be `archive_reason` since that is the property we are going to be using to figure out why the cases were archived

custom/enikshay/nikshay_datamigration/factory.py
proteusvacuum: sorry, when, I said "fail", I meant that we should still assign to the `_archive_` owner id, set the `archive_reason`, and `migration_error` to `migration_not_phi_location` or something.

custom/enikshay/nikshay_datamigration/factory.py
```diff
@@ -129,6 +128,15 @@ def get_person_case_structure(self):
             },
         }
 
+        if self._location:
+            assert self._location.location_type.name == 'phi'
```
proteusvacuum: I think you want to use `location_type.code` in this check

custom/enikshay/nikshay_datamigration/factory.py
```diff
@@ -129,6 +128,15 @@ def get_person_case_structure(self):
...
+            kwargs['attrs']['owner_id'] = self._location.location_id
+        else:
+            kwargs['attrs']['owner_id'] = '_archive_'
```
proteusvacuum: I think it might be better if you use the constant defined in `casexml.apps.case.const`

nickpell:
fixed conflicts

proteusvacuum (changes requested):
Couple of comments, sorry I wasn't clear in the first place

nickpell:
@proteusvacuum updated

proteusvacuum (approved):
nit: There's also the "UNOWNED_EXTENSION_OWNER_ID" constant that you could use instead of the hardcoded `-`, but that's not important. 
G2M on ✅

PR #14319, opened 2016-12-22
use '' for blank phone number

nickpell:
@proteusvacuum @czue

### DIFF
--- a/custom/enikshay/nikshay_datamigration/factory.py
+++ b/custom/enikshay/nikshay_datamigration/factory.py
@@ -17,9 +17,9 @@ EPISODE_CASE_TYPE = 'episode'
 TEST_CASE_TYPE = 'test'
 
 
-def validate_number(string_value):
+def validate_phone_number(string_value):
     if string_value is None or string_value.strip() == '':
-        return None
+        return ''
     else:
         return int(string_value)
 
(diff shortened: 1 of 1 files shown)


proteusvacuum (approved):

custom/enikshay/nikshay_datamigration/factory.py
```diff
@@ -17,9 +17,9 @@
 TEST_CASE_TYPE = 'test'
 
 
-def validate_number(string_value):
+def validate_phone_number(string_value):
     if string_value is None or string_value.strip() == '':
-        return None
+        return ''
     else:
         return int(string_value)
```
proteusvacuum: good point about `int` @nickpell I think we can just keep as a string? 

This is actually for an import script, and the data is pretty messy. 

@nickpell do you think it would be worthwhile to do something like [this](https://github.com/dimagi/commcare-hq/blob/mk-enikshay-nikshay-api/custom/enikshay/integrations/ninetyninedots/repeater_generators.py#L133-L138)? to format the phone number properly or do you think that would lead to too many failures?

Right now, the numbers we collect explicitly do not have a country code attached, but I'll raise this with the requirements team and see if that is something we need. I didn't realize that was a requirement to use the messaging framework.

gcapalbo: @proteusvacuum ok, one option i see is sometimes people will choose to store the original/locally formatted number as one case property (since sometimes people prefer to see that in commcare if it comes up), and then store the internationally formatted number as contact_phone_number for messaging

gcapalbo: (also might want to update the phonenumbers ref if you end up using it to format the numbers, looks like it hasn't been updated in a year or so)

nickpell: It looks like all of the numbers in the data have 10 digits (leading digit non-zero).  Would it be sufficient to assert that the returned string has 10 characters?  Noting that if we have to integrate with messaging proper formatting would be required.

gcapalbo: @nickpell if all the numbers are exactly 10 digits without a leading zero (and no other characters), then assuming these are all Indian numbers, all we should have to do is add '91' to the beginning of each in order to put it in international format

nickpell: https://github.com/dimagi/commcare-hq/pull/14333

PR #14322, opened 2016-12-22
Optimize exports

benrudolph:
@czue @NoahCarnahan Did some more profiling and found that a slow down was in the schema merge, which was not where I had originally thought. I took just two schemas from icds and merged them together and it took 149 seconds to complete the merge 100 times . This PR lowers that number to to 2.89 seconds (50x speedup). 

Before:
```python
In [2]: timeit.timeit('ExportDataSchema._merge_schemas(d, d)', setup='from corehq.apps.export.models import *; d = FormExportDataSchema.get("d8518dec00db650d19671d96081143f0")', number=100)
Out[2]: 149.8663728237152
```

After:
```python
In [6]: timeit.timeit('ExportDataSchema._merge_schemas(d, d)', setup='from corehq.apps.export.models import *; d = FormExportDataSchema.get("d8518dec00db650d19671d96081143f0")', number=100)
Out[6]: 2.897454023361206
```

The optimization does two things, broken down by commit:
1. It calls the `keyfn` way less. When originally written I pretty much thought that it was negligible so i didn't think about it. On profiling the icds schema, it was getting called 42,238 times. this commit just uses a better data structure to drop the calls to 470.
2. This completely drops the `copyfn`. When I wrote this, in theory I was hoping to not do any mutation to objects because it "feels" better. Turns out in practice it was highly inefficient to export to json and then reimport into a new model. This does mean that the schemas are modified when they are passed in and should not be used/saved afterward (which is consistent with how it works now)
cc: @millerdev

### DIFF
--- a/corehq/apps/export/models/new.py
+++ b/corehq/apps/export/models/new.py
@@ -160,7 +160,7 @@ class ExportItem(DocumentSchema):
 
     @classmethod
     def merge(cls, one, two):
-        item = cls(one.to_json())
+        item = one
         item.last_occurrences = _merge_dicts(one.last_occurrences, two.last_occurrences, max)
         item.inferred = one.inferred or two.inferred
         item.inferred_from |= two.inferred_from
@@ -932,7 +932,6 @@ class MultipleChoiceItem(ExportItem):
                     value=option1.value,
                     last_occurrences=_merge_dicts(option1.last_occurrences, option2.last_occurrences, max)
                 ),
-            copyfn=lambda option: Option(option.to_json())
         )
 
         item.options = options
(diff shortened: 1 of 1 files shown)


benrudolph:
Was able to build a schema for the ccs_record in 6 minutes with this new code. Not great but better than 30 minutes...

NoahCarnahan (approved):
Nice find. looks good to me.

czue (approved):
wow, nice.

> When I wrote this, in theory I was hoping to not do any mutation to objects because it "feels" better. Turns out in practice it was highly inefficient to export to json and then reimport into a new model. This does mean that the schemas are modified when they are passed in and should not be used/saved afterward (which is consistent with how it works now)

I'm not too familiar with the surrounding code, but if you wanted to enforce this you might be able to use a different class to initially wrap the objects up front that prevents saving? E.g. something [like this](https://github.com/dimagi/commcare-hq/blob/master/corehq/apps/groups/models.py#L290).

czue:
will merge now and let you decide whether to follow up

benrudolph:
are you saying to unwrap and wrap the schemas in a different class at the start of the merge call? i think this would work since the schemas being passed in would no longer be mutated. for now i think i'm going to leave as is to stay away from wrapping

czue:
my comment probably didn't make sense. what i was picturing is something like:

```python
schemas = ExportSchema.get_everything()
merged = merge_schemas(schemas)
do_something_with_merged(merged)
```

and i was saying that if it looked something like that you might be able to do something like this
```python
schemas = ReadonlyExportSchema.get_everything()  # instantiate read-only schemas so they can't be saved since they will be mutated during merge.
merged = merge_schemas(schemas)
do_something_with_merged(merged)
```
but I probably just don't know what's going on here well enough for that to make sense.

benrudolph:
Ah i see now. thanks for the clarification, that's definitely possible. i'm going to put it on my list of things to do, but i don't think it's super high priority wise (let me know if you disagree)

czue:
agree

PR #14325, opened 2016-12-22
Do not include shared_properties

benrudolph:
@czue @NoahCarnahan this optimization shaves off a few seconds for me locally on a 10 build icds app (went down from 18 seconds to 15 seconds). What this does is essentially disallows getting shared case properties. when set to `True` it will call a function called `get_all_shared_apps_in_domain` which is `memoized` but still a hefty call to make for project with lots of apps. since we already go through all apps in the domain, this won't change the functionality much. *however* if you export an app that has shared properties in another app and that app doesn't have any submissions, then the shared case properties won't show up. we can mitigate this by including the current version of every app in the analysis (currently it only includes the current version of the app of the case you selected). put this one in a separate PR cause it's more wishy washy and i'm not sure it's worth it

### DIFF
--- a/corehq/apps/export/models/new.py
+++ b/corehq/apps/export/models/new.py
@@ -11,6 +11,7 @@ from django.core.urlresolvers import reverse
 
 from corehq.apps.export.esaccessors import get_ledger_section_entry_combinations
 from dimagi.utils.decorators.memoized import memoized
+from dimagi.utils.decorators.profile import profile
 from couchdbkit import SchemaListProperty, SchemaProperty, BooleanProperty, DictProperty
 
 from corehq import feature_previews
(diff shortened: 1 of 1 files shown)


czue:
this PR description is hurting my brain.

are you saying the only edge case is if you export a case type that is shared across two apps A and B. and B does not have any submissions at all. then the additional case properties that are defined in B will not show up when you initiate the export from A?

on the one hand that seems like a very obscure issue and seems worth it if performance gain is significant. on the other i'm sure it will cost someone on our support time time at some point.

czue (changes requested):

benrudolph:
blehh, sometimes i feel like building exports off the application is the best thing and other times i feel like it's the worst thing...

you've understood it correctly, but i don't think it'll actually give us that much of a performance boost. i'm gonna go ahead and close for now. i think delayed schemas are the only true way to get load times down under a few seconds

corehq/apps/export/models/new.py
```diff
@@ -1043,6 +1044,7 @@ class Meta:
         app_label = 'export'
 
     @classmethod
+    @profile('generate_schema_from_builds')
```
benrudolph: d'oh

PR #14333, opened 2016-12-23
validate that phone number has length 10

nickpell:
@proteusvacuum @czue 

Holding off on adding the country code until we know from the requirements team how to proceed.


nickpell:
✅

custom/enikshay/nikshay_datamigration/factory.py
```diff
@@ -18,10 +18,12 @@
 
 
 def validate_phone_number(string_value):
-    if string_value is None or string_value.strip() == '':
+    if string_value is None or string_value.strip() in ['', '0']:
         return ''
     else:
-        return int(string_value)
+        phone_number = str(int(string_value))
+        assert len(phone_number) == 10
```
proteusvacuum: I think phone numbers in India can also be 8 numbers...

nickpell: fwiw all of the phone numbers in the data had 10 digits

gcapalbo: just wanted to reiterate that in order to work with messaging, numbers have to be in international format. if all the numbers are exactly 10 digits, then all we should have to do is add '91' to the front of them to put them in international format for India.

how else is this phone number used? if you need to keep a version of it in national format (without the '91'), you'd have to put it in another case property since `contact_phone_number` should always be in international format

PR #14335, opened 2016-12-23
Fix ucr date comparisons

NoahCarnahan:
~Opening for tests~
http://manage.dimagi.com/default.asp?244364

### DIFF
--- a/corehq/apps/userreports/tests/test_report_filters.py
+++ b/corehq/apps/userreports/tests/test_report_filters.py
@@ -107,8 +107,8 @@ class DateFilterTestCase(SimpleTestCase):
             return filter.create_filter_value(reports_core_value).to_sql_values()
 
         val = get_query_value(compare_as_string=False)
-        self.assertEqual(type(val['my_slug_startdate']), datetime)
-        self.assertEqual(type(val['my_slug_enddate']), datetime)
+        self.assertEqual(type(val['my_slug_startdate']), date)
+        self.assertEqual(type(val['my_slug_enddate']), date)
 
         val = get_query_value(compare_as_string=True)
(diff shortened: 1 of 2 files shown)


corehq/apps/userreports/tests/test_report_filters.py
```diff
@@ -107,14 +116,155 @@ def get_query_value(compare_as_string):
...
+        self.assertEqual(1, self.docs_returned(view.export_table))
+
+    def test_string_date_filter(self):
+        # Confirm that "compare_as_string" date filters include rows that match the start and/ord end date
+        view = self._create_view({
+            "date_as_string_filter": "2017-01-01 to 2017-01-01",
+            "date_as_string_filter-start": "2017-01-01",
+            "date_as_string_filter-end": "2017-01-01",
+        })
+        self.assertEqual(1, self.docs_returned(view.export_table))
```
NoahCarnahan: This test would have failed before this PR

NoahCarnahan:
This PR fixes a bug with the report builder.
The report builder saves date questions in string type indicators. If a user wants to filter the report on a date indicator, a date filter with the "compare_as_string" option is created. However, the date filter object converts the dates selected in the UI to datetimes, and then converts them back to strings. This process would result in the filter excluding values that were the same as the start date because it was appending a time to the lower bound value:

`"2016-12-23 00:00:00" > "2016-12-23"`
 
 This change ensures that the starting lower value in the filter is in the form "2016-12-23", not "2016-12-23 00:00:00".

cc @biyeun

biyeun (approved):

NoahCarnahan:
whoops, meant to remove the tag. Self merging since it has been approved.

PR #14338, opened 2016-12-23
populate dob

nickpell:
Drops the tables so I can reimport the data, casting ```page``` as an integer; populates ```dob``` according to the method used in the app.

@proteusvacuum @czue 

After this gets merged to master I'll run the migration in a private release and then repopulate the tables.  A full deploy isn't necessary since I assume the migration from then on would be run from branches containing this PR.


proteusvacuum:
conflix

nickpell:
rebased

proteusvacuum (approved):

PR #14339, opened 2016-12-23
Escape fewer things from filenames

esoergel:
@dannyroberts @NoahCarnahan

Prompted by http://manage.dimagi.com/default.asp?244453
Looks like the url escaping was done in response to http://manage.dimagi.com/default.asp?228419, but that is more restrictive than is strictly necessary.

This PR tests a couple things we've had issues with in the past, and verifies that it gives reasonable output.

### DIFF
--- a/corehq/apps/export/tasks.py
+++ b/corehq/apps/export/tasks.py
@@ -1,4 +1,3 @@
-import re
 import logging
 from celery.task import task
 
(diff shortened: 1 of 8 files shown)


czue:
nice tests!

can we try to combine / reconcile this with https://github.com/dimagi/commcare-hq/blob/master/corehq/apps/reports/util.py#L475? seems like we should be trying to centralize this logic in a single place which we're not doing a good job of now.

benrudolph (approved):
Nice this has always annoyed me (missed cory's comment before approving, agree it'd be nice to have only one of these functions)

esoergel:
@czue good call, I didn't know about that function.  I moved it into `corehq.util.files`, which seemed like a better place for it, and added a bunch of test cases.

czue (approved):
nice this looks great.

PR #14342, opened 2016-12-24
update visibility of TB reports to show to all users

czue:
@NoahCarnahan / anyone


czue:
self merging this since it's only custom reports and is blocking a user manual

NoahCarnahan:
👍

PR #14345, opened 2016-12-24
Add ability to use test phi to eNikshay migration

proteusvacuum:
@czue @nickpell 
Added the ability to directly assign to a test PHI location id so it immediately assigns to our test user.

### DIFF
--- a/custom/enikshay/nikshay_datamigration/factory.py
+++ b/custom/enikshay/nikshay_datamigration/factory.py
@@ -1,3 +1,4 @@
+from collections import namedtuple
 from datetime import date, datetime
 
 from dimagi.utils.decorators.memoized import memoized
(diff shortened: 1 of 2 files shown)


nickpell (approved):

custom/enikshay/nikshay_datamigration/factory.py
```diff
@@ -292,18 +297,27 @@ def _followups(self):
 
     @property
     def state(self):
+        if self.test_phi is not None:
+            return MockLocation('FAKESTATE', 'fake_state_id', MockLocationType('state', 'state'))
```
czue: is there a chance we should use real (as in existing `Location` objects) but fake (as in test or QA locations) data for this instead of mocks? It seems like that would allow us to exercise things that may depend on the hierarchy and make it easier to sync phones.

proteusvacuum: Good idea. @nickpell do you think this is still useful for testing? otherwise I'll leave it as is.

nickpell: so far I've gotten by with just setting a few nikshay codes manually to test Locations, but if that becomes a blocker I'll take this suggestion

PR #14347, opened 2016-12-26
add UI for overwriting a module's case list/details using another module

czue:
faced with the prospect of redoing a giant case list / details screen i decided it'd be a better time spend to just build this feature.

This adds the section at the bottom to the case list and detail screens of all advanced modules, and basic modules using a feature flag.
image

@dimagi/product @orangejenny I think it'd be really good to productize the basic module version under the GS efficiency umbrella so am keen to hear what you think that takes versus this UI

@proteusvacuum thought you might enjoy this too.

### DIFF
--- a/corehq/apps/app_manager/templates/app_manager/v1/partials/case_list_module_overwrite.html
+++ b/corehq/apps/app_manager/templates/app_manager/v1/partials/case_list_module_overwrite.html
@@ -1,23 +1,29 @@
 {% load i18n %}
 {% load hq_shared_tags %}
 {% load xforms_extras %}
+{% load app_manager_extras %}
 {% if module.module_type != 'basic' or request|toggle_enabled:'COPY_CASE_CONFIGS' %}
 <form class="form-inline" method='POST' action='{% url "overwrite_module_case_list" domain app.id module.id %}'>
     {% csrf_token %}
     <fieldset>
         <legend>{% trans "Use Configuration from Existing Module" %}</legend>
-        <p class="help-block">{% trans "This will <strong>overwrite</strong> the current case list/details screen with the configuration from the selected module." %}</p>
-        <label for="other_module_id">{% trans "Module" %}: </label>
-        <select name='source_module_id' class='form-control'>{% for other_mod in app.get_modules %}
-            {% if other_mod.id !=  module.id and other_mod.module_type != 'report' and other_mod.case_type == module.case_type %}
-                <option value={{ other_mod.id }}>{{ other_mod.name|html_trans:langs }}</option>
-            {% endif %}
-        {% endfor %}</select>
-        <input name="detail_type" type="hidden" data-bind="value: columnKey" />
-        <button class='btn btn-danger' type="submit">
-            <i class="fa fa-copy"></i>
-            {% trans "Overwrite Case List" %}
-        </button>
+            <p class="help-block">{% trans "This will <strong>overwrite</strong> the current case list/details screen with the configuration from the selected module." %}</p>
+    {% with app|get_available_modules_for_case_list_configuration:module as available_modules %}
+        {% if available_modules %}
+            <label for="other_module_id">{% trans "Module" %}: </label>
+            <select name='source_module_id' class='form-control'>{% for other_mod in app.get_modules %}
+                {% if other_mod.id !=  module.id and other_mod.module_type != 'report' and other_mod.case_type == module.case_type %}
+                    <option value={{ other_mod.id }}>{{ other_mod.name|html_trans:langs }}</option>
+                {% endif %}
+            {% endfor %}</select>
+            <input name="detail_type" type="hidden" data-bind="value: columnKey" />
+            <button class='btn btn-danger' type="submit">
+                <i class="fa fa-copy"></i>
+                {% trans "Overwrite Case List" %}
+        {% else %}
+            <p><em>{% trans "Didn't find any other modules with the same case type."%}</em></p>
+        {% endif %}
+    {% endwith %}
     </fieldset>
 </form>
 {% endif %}
--- a/corehq/apps/app_manager/templates/app_manager/v1/partials/case_list_module_overwrite.html
+++ b/corehq/apps/app_manager/templates/app_manager/v1/partials/case_list_module_overwrite.html
@@ -1,0 +1,20 @@
+{% load i18n %}
+{% load xforms_extras %}
+<form class="form-inline" method='POST' action='{% url "overwrite_module_case_list" domain app.id module.id %}'>
+    {% csrf_token %}
+    <fieldset>
+        <legend>{% trans "Use Configuration from Existing Module" %}</legend>
+        <p class="help-block">{% trans "This will <strong>overwrite</strong> the current case list/details screen with the configuration from the selected module." %}</p>
+        <label for="other_module_id">{% trans "Module" %}: </label>
+        <select name='source_module_id' class='form-control'>{% for other_mod in app.get_modules %}
+            {% if other_mod.id !=  module.id and other_mod.module_type != 'report' %}
+                <option value={{ other_mod.id }}>{{ other_mod.name|html_trans:langs }}</option>
+            {% endif %}
+        {% endfor %}</select>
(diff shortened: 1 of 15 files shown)


corehq/apps/app_manager/templatetags/app_manager_extras.py
```diff
@@ -0,0 +1,14 @@
...
+
+register = template.Library()
+
+
+@register.filter
+def get_available_modules_for_case_list_configuration(app, module):
+    return [
+        m for m in app.get_modules()
+        if m.id != module.id
+           and m.module_type not in ('reports', 'careplan')
```
dimagimon: E131 continuation line unaligned for hanging indent

corehq/apps/app_manager/templatetags/app_manager_extras.py
```diff
@@ -0,0 +1,14 @@
...
+           and m.case_type == module.case_type
```
dimagimon: E131 continuation line unaligned for hanging indent

corehq/apps/app_manager/templates/app_manager/v2/partials/case_detail.html
```diff
@@ -17,6 +17,7 @@
 </legend>
 <div data-bind="with: longScreen">
     {% include 'app_manager/v2/partials/case_list_properties.html' %}
+    {% include 'app_manager/v1/partials/case_list_module_overwrite.html' %}
```
biyeun: you should make a v2 copy of this template

corehq/apps/app_manager/templates/app_manager/v2/partials/case_list.html
```diff
@@ -77,6 +77,7 @@
         </div>
     </div>
     {% include 'app_manager/v2/partials/case_list_properties.html' %}
+    {% include 'app_manager/v1/partials/case_list_module_overwrite.html' %}
```
biyeun: v2 partial

corehq/apps/app_manager/templates/app_manager/v2/partials/case_detail.html
czue: ok cool i can do that.

corehq/apps/app_manager/templates/app_manager/v2/partials/case_list.html
biyeun: fyi, the diff script makes sure the v2 template stays up to date

corehq/apps/app_manager/tests/data/v2_diffs/partials/case_detail.html.diff.txt
```diff
@@ -15,6 +15,6 @@
  <div data-bind="with: longScreen">
 -    {% include 'app_manager/v1/partials/case_list_properties.html' %}
 +    {% include 'app_manager/v2/partials/case_list_properties.html' %}
+     {% include 'app_manager/v1/partials/case_list_module_overwrite.html' %}
```
biyeun: v2

corehq/apps/app_manager/tests/data/v2_diffs/partials/case_list.html.diff.txt
```diff
@@ -27,10 +27,10 @@
      </div>
 -    {% include 'app_manager/v1/partials/case_list_properties.html' %}
 +    {% include 'app_manager/v2/partials/case_list_properties.html' %}
+     {% include 'app_manager/v1/partials/case_list_module_overwrite.html' %}
```
biyeun: v2

corehq/apps/app_manager/templates/app_manager/v1/partials/case_list_module_overwrite.html
```diff
@@ -0,0 +1,29 @@
+{% load i18n %}
+{% load hq_shared_tags %}
+{% load xforms_extras %}
+{% load app_manager_extras %}
+{% if module.module_type != 'basic' or request|toggle_enabled:'COPY_CASE_CONFIGS' %}
+<form class="form-inline" method='POST' action='{% url "overwrite_module_case_list" domain app.id module.id %}'>
+    {% csrf_token %}
+    <fieldset>
```
biyeun: one of the reasons for a v2 template is that this styling is updated to a panel for the settings pages. so you're not just copying in vain either.

biyeun:
@czue looks good, can you add a v2 version of the new template file? I can make the updates to the style for it in my pr once this is merged.

czue:
thanks! updated with v2 versions.

orangejenny:
> @dimagi/product @orangejenny I think it'd be really good to productize the basic module version under the GS efficiency umbrella so am keen to hear what you think that takes versus this UI

My initial thought is that this seems pretty close to being a general "copy module" feature, which I think would be a bit easier to grasp and less error-prone (somebody's going to overwrite the case list but not set the case type correctly...). Allowing copying a module also fits in with how we allow copying other first order objects (apps, forms) and [has been asked for before](https://dimagi.uservoice.com/forums/194738-app-builder/suggestions/3870007-copy-modules-between-domains).

An alternative might be to keep a similar UI, but only make it visible when the case list/detail config is empty and you have some other module that uses cases. That'd at least keep this from being the 5th, or 12th, or whatever additional case detail section.

Can you suggest how to best push this forward via gs efficiency? I agree this would be a good addition to the product, and I'd be happy to chat about it or help with a spec.

czue:
> My initial thought is that this seems pretty close to being a general "copy module" feature, which I think would be a bit easier to grasp and less error-prone. Allowing copying a module also fits in with how we allow copying other first order objects (apps, forms) and has been asked for before.

Agreed. I thought about implementing this as copy module, and that might be the right answer for our basic users. For more complicated apps I think this feature is much more flexible since you can already copy forms if you like, but it also allows you to work with two existing modules that just should be "kept in sync" as well as allows you to take the configuration as a starting point to tweak it, without having to realize up front that you need to start from a copy. So I think even in a world where copy module exists we'd want to expose this to some set of users somewhere. Does the new app builder have an advanced area for modules? (also this has also [been asked for](https://dimagi.uservoice.com/forums/194747-user-experience-design/suggestions/3332532-copy-case-detail-config-from-one-module-to-another))

> somebody's going to overwrite the case list but not set the case type correctly...

Not sure how closely you looked at the PR, but this already doesn't let you select a source module unless the case types are the same. Not sure if there's something else you're worried about?

> An alternative might be to keep a similar UI, but only make it visible when the case list/detail config is empty and you have some other module that uses cases. That'd at least keep this from being the 5th, or 12th, or whatever additional case detail section.

I don't think we should only hide it when empty, as per the use case above. I'd rather find a non-invasive home for it. We can hide it when there's no matching modules which would be an easy change.

> Can you suggest how to best push this forward via gs efficiency? I agree this would be a good addition to the product, and I'd be happy to chat about it or help with a spec.

If there's a planned "advanced" area for modules/case list screens in the new app manager UI i'd be in favor of putting it there. If there's not, what about just making one in a default-hidden accordion and putting it there?

orangejenny:
> If there's a planned "advanced" area for modules/case list screens in the new app manager UI i'd be in favor of putting it there. If there's not, what about just making one in a default-hidden accordion and putting it there?

The new form view page has an "Actions" tab, which is where copy lives. You could add an actions tab to module view. There isn't currently an advanced area or an accordion.

czue:
> The new form view page has an "Actions" tab, which is where copy lives. You could add an actions tab to module view. There isn't currently an advanced area or an accordion.

that sounds good to me. should i wait until new app manager is released or should I do that now? haven't been tracking progress closely.

orangejenny:
The redesign isn't really relevant. This

> The new form view page has an "Actions" tab, which is where copy lives. You could add an actions tab to module view. There isn't currently an advanced area or an accordion.

is also true in current app manager, except that the tab is called "Advanced" in the old version and "Actions" in the new.

proteusvacuum:
Cool! When we decouple modules and case lists this will no longer be necessary 😉

snopoke:
> Cool! When we decouple modules and case lists this will no longer be necessary :wink:

Been wanting that for SO long!

PR #14348, opened 2016-12-26
Hard Separate Old Cloudcare styling from Formplayer/preview

biyeun:
also fixes the required question styling bug from http://manage.dimagi.com/default.asp?243260

@benrudolph / @orangejenny cc @czue


benrudolph (approved):
sweet so now everything named cloudcare is only used in old cloudcare right?

biyeun:
@benrudolph yes, that's the idea!

benrudolph:
err probably should've asked this before merging, (i don't actually think it'll be an issue) but did you test edit forms?

PR #14349, opened 2016-12-26
Delayed schemas

benrudolph:
@czue @NoahCarnahan this is a WIP for delayed schemas, but wanted to get some feedback on this first before plowing ahead to see if there're any better ideas out there. or maybe we decide this is too complex and we try to optimize more. broken up by commit, but the important one to look at is: https://github.com/dimagi/commcare-hq/commit/3682b983fe23c6c5b87fdbcfea7413023b3a9197

the general strategy is this:

1. First generate the schema for the current applications
2. Attempt to generate the schema for the historical applications (previous builds, open to better naming)
3. If it finishes within in X seconds, just continue as usual
4. If it doesn't finish, kill the task, mark the schema as "needs more processing," kick off another task at the end to build the delayed schema again and remerge/save

it sounds fairly simple, but in my head there are a fair number of edge cases that could happen and leave the export schema in a weird state. we lose the semi-atomicity we had before where the schema is built and if it was saved we know it was complete and successful. this simplified many things. lots of error handling needs to be added to this PR (hence WIP), but is a general strategy.

some things for myself to consider:

- what happens when multiple processes attempt to generate a delayed schema?
- what happens when the delayed schema errors, and `last_app_versions` is set to the most recent versions and `_get_app_build_ids_to_process` returns an empty array?
- is there an easy way to tell which apps have been processed by the schema besides using `last_app_versions`?


corehq/apps/export/models/new.py
```diff
@@ -1063,29 +1066,35 @@ def generate_schema_from_builds(cls, domain, app_id, identifier, force_rebuild=F
...
-        if app_id:
-            app_build_ids.append(app_id)
-        else:
-            app_build_ids.extend(cls._get_current_app_ids_for_domain(domain))
 
-        for app_doc in iter_docs(Application.get_db(), app_build_ids, chunksize=10):
-            if (not app_doc.get('has_submissions', False) and
-                    app_doc.get('copy_of')):
-                continue
+        # These build ids represent the current state of the app and show what items are currently
```
dimagimon: E303 too many blank lines (2)

corehq/apps/export/tasks.py
```diff
@@ -96,6 +97,27 @@ def _cached_add_inferred_export_properties(sender, domain, case_type, properties
     inferred_schema.save()
 
 
-@task
+@task(soft_time_limit=SCHEMA_GENERATION_TIMEOUT)
```
benrudolph: my assumption here is that `soft_time_limit` will start ticking from when we kick off the task vs. when the task actually starts. need to confirm

corehq/apps/export/models/new.py
```diff
@@ -1066,32 +1066,35 @@ def generate_schema_from_builds(cls, domain, app_id, identifier, force_rebuild=F
...
-        try:
-            current_schema = task.get(timeout=SCHEMA_GENERATION_TIMEOUT)
-        except TimeoutError:
-            # In this case, generating the schema has taken too long. Wait until later to
-            # generate the rest of the schema.
-            # TODO: Figure out how to properly handle this case
-            pass
+
+        # Process all historical applications if we are not currently processing them
+        if not current_schema.processing_delayed_schemas:
```
benrudolph: note to self: could we ever end up in a state where `processing_delayed_schemas` is `True`, but we actually do need to process the historical applications?

NoahCarnahan:
This seems like a good approach.
I was just thinking though, have we considered building the schemas in a non-lazy approach though? Perhaps we should just kick off a rebuild schema task every time a build is created?

benrudolph:
Ah interesting, I always wondered if that might be worth it, but then thought it'd be a lot of extra computation power. The minor thing with that though is that we should rather kick off a build every time there's a form submitted to a new build (rather than a new build created). We could do that in the pillow though. worth discussing at our next meeting

PR #14350, opened 2016-12-26
just curious

benrudolph:


emord:
I think you have to change https://github.com/dimagi/commcare-hq/blob/master/Dockerfile#L1

benrudolph:
oops yep thanks

czue:
haha nice try.

based on the pip install paths it seems to still be using 2.7, though not totally sure why. e.g.
```
Requirement already satisfied (use --upgrade to upgrade): django-two-factor-auth==1.4.0 from git+git://github.com/emord/django-two-factor-auth.git@cea4533f6048fe4de7431a182a44490c84d0486c in /vendor/lib/python2.7/site-packages (from -r requirements/requirements.txt (line 80))
```

benrudolph:
that would've been incredible. we're probably not close to this. yesterday i had wanted to play with a python3 only lib: https://docs.python.org/3/library/asyncio.html

emord:
Yeah there's no way we're close to this. 

Not sure if you figured out why it's doing it but I think it's because it pulls the docker image that is built from master. There's probably a way to get it to create two versions from every master, one with 3, one with 2, but likely not worth it now.

For fun you can run `python -3 manage.py ...` to see warnings that will be errors in python 3

benrudolph:
😞 agree not worth it. mostly do it for 💩s and 😆 s

PR #14351, opened 2016-12-27
catch cases when dotpType is '0'

nickpell:
Catches an edge case for 224 patients.

@czue @proteusvacuum

### DIFF
--- a/custom/enikshay/nikshay_datamigration/models.py
+++ b/custom/enikshay/nikshay_datamigration/models.py
@@ -170,6 +170,7 @@ class PatientDetail(models.Model):
     @property
     def treatment_supporter_designation(self):
         return {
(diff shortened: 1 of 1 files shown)


custom/enikshay/nikshay_datamigration/models.py
```diff
@@ -170,6 +170,7 @@ def occupation(self):
     @property
     def treatment_supporter_designation(self):
         return {
+            '0': 'undetermined_by_migration',
```
czue: is this not in the enums we have? buggy data?

nickpell: it's only 224 out of ~400K patients, so I'm guessing buggy data.

PR #14358, opened 2016-12-28
Export directory names

esoergel:
@czue @sravfeyn @benrudolph 
http://manage.dimagi.com/default.asp?243332
This is a two-fold fix:

Commit # 1 changes the default name of new form exports from something like:
`Surveys > Survey Category 1 (Ex. Household) > Survey 1: 2016-12-24`
to something like:
`Surveys - Survey Category 2 (Ex. Village) - Survey 2 (2016-12-27)`
This way what you see on the "edit export" page is a valid name, and will remain unchanged in the download.

Commit # 2 strips any illegal characters from these names before creating the download.

This is a similar issue to https://github.com/dimagi/commcare-hq/pull/14339

### DIFF
--- a/corehq/apps/export/export.py
+++ b/corehq/apps/export/export.py
@@ -11,6 +11,7 @@ from soil import DownloadBase
 from couchexport.export import FormattedRow, get_writer
 from couchexport.files import Temp
 from couchexport.models import Format
+from corehq.util.files import safe_filename
 from corehq.apps.export.esaccessors import (
     get_form_export_base_query,
     get_case_export_base_query,
@@ -61,6 +62,7 @@ class _Writer(object):
             name = export_instances[0].name
         else:
             name = ''
+        name = safe_filename(name)
 
         fd, self._path = tempfile.mkstemp()
         with os.fdopen(fd, 'wb') as file:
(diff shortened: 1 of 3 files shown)


benrudolph:
it seems odd to me that this would be the root cause since exports have been named like this for ages. i do like the name change regardless though. hope it fixes it!
```
ERROR: Ensure that we can export stock properties in a form export
  File "/mnt/commcare-hq-ro/corehq/apps/export/tests/test_get_export_file.py", line 257, in test_form_stock_columns
    with writer.open([export_instance]):
  File "/usr/local/lib/python2.7/contextlib.py", line 17, in __enter__
    return self.gen.next()
  File "/mnt/commcare-hq-ro/corehq/apps/export/export.py", line 65, in open
    name = safe_filename(name)
  File "/mnt/commcare-hq-ro/corehq/util/files.py", line 22, in safe_filename
    filename = filename if isinstance(filename, unicode) else filename.decode('utf8')
AttributeError: 'NoneType' object has no attribute 'decode'
```

benrudolph (commented):
👍 on ✔️

benrudolph:
offlined with ethan and have convince myself that this is indeed the root cause. good hunting!

PR #14360, opened 2016-12-28
skip ownership cleanliness checks during case processing

czue:
@nickpell @benrudolph what do you think about this plan? I recognize it's not the... cleanest? but I think should get the job done.


benrudolph:
👍

PR #14363, opened 2016-12-28
allow named expressions to reference other named expressions

czue:
as long as no loops. this has been on my list for ages and got frustrated by it again today. didn't realize how easy it would be.

@orangejenny / anyone.


czue:
@sheelio @lwyszomi @kkrampa fyi in case this helps with report building and simplification

PR #14364, opened 2016-12-28
Allow non-superusers to import apps.

esoergel:
@orangejenny @nickpell
http://manage.dimagi.com/default.asp?244593
I think this was done in error.  As-is, non superusers can access the page, but they see a blank form.  If they edit the html to remove the "hide" class, they can submit the form sucessfully.  Server-side validation checks for the "edit apps" permission.  This page is documented on the public wiki, so I assume it's intended to be usable by non superusers.

### DIFF
--- a/corehq/apps/app_manager/templates/app_manager/v1/import_app.html
+++ b/corehq/apps/app_manager/templates/app_manager/v1/import_app.html
@@ -39,9 +39,8 @@
         {% csrf_token %}
 {% if app %}
         <p>Import application <strong>{{ app.name }}</strong> from domain <strong>{{ app.domain }}</strong>?</p>
-{% else %}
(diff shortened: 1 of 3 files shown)


nickpell (approved):

orangejenny:
Looks like there's a v2 template to update

```
APP MANAGER V2 Diff Failure
An edit made to a V1 App Manager file "import_app.html" does not match the stored diff
of its V2 counterpart.
to rebuild the broken diffs.
These files are located in
/mnt/commcare-hq-ro/corehq/apps/app_manager/tests/../templates/app_manager/v1/import_app.html
/mnt/commcare-hq-ro/corehq/apps/app_manager/tests/../templates/app_manager/v2/import_app.html
**************************************
```

czue:
nice

PR #14366, opened 2016-12-28
Handle advanced forms that open cases in repeats

benrudolph:
anyone / http://manage.dimagi.com/default.asp?244564


corehq/apps/export/models/new.py
```diff
@@ -1240,8 +1240,13 @@ def _process_app_build(cls, current_schema, app, form_xmlns):
...
         case_updates = form.get_case_updates(form.get_module().case_type)
         xform = form.wrapped_xform()
+        if isinstance(form.actions, AdvancedFormActions):
+            open_case_actions = form.actions.open_cases
+        else:
+            open_case_actions = form.actions.subcases
+
         repeats_with_subcases = {
-            subcase_action for subcase_action in form.actions.subcases
+            subcase_action for subcase_action in open_case_actions
```
czue: nit: should loop variable be renamed to `open_case_action`?

czue: noting that `AdvancedForm` has a `get_open_subcase_actions` function, so depending on the intent you might want to use that instead (otherwise renaming the variable would be more clear)

corehq/apps/export/tests/test_export_data_schema.py
```diff
@@ -539,9 +539,12 @@ def setUpClass(cls):
         cls.first_build.copy_of = cls.current_app.get_id
         cls.first_build.version = 3
 
+        cls.advanced_app = Application.wrap(cls.get_json('advanced_application'))
```
czue: don't think is necessary to change but if easy I find it more easy to reason with tests when the app is created by hand. E.g. using `AppFactory` or just manually adding stuff. E.g. something [like this](https://github.com/dimagi/commcare-hq/blob/master/corehq/apps/app_manager/tests/test_models.py#L23).

czue (approved):
no blockers but will give you a chance to respond

corehq/apps/export/models/new.py
benrudolph: ah that might actually be what i'm looking for 👍

corehq/apps/export/tests/test_export_data_schema.py
benrudolph: 👍 originally was doing it because i also needed to define the form xml. agree it's better to define the app programmatically for a variety of reasons. will update

corehq/apps/export/models/new.py
benrudolph: i actually don't think `get_open_subcase_actions` is what i'm looking for since in an advanced forms you can seemingly open a case with a repeat group in another module without a case_index. as i'm typing that, it sounds weird, but that's what i was seeing in my app. totally possible i'm misunderstanding something. i'm not that well versed in the abilities of advanced forms/modules

czue:
thanks for follow up

PR #14368, opened 2016-12-28
use right enum values for hiv_status

nickpell:
@czue @proteusvacuum cc: @esoergel

### DIFF
--- a/custom/enikshay/nikshay_datamigration/models.py
+++ b/custom/enikshay/nikshay_datamigration/models.py
@@ -263,6 +263,15 @@ class Outcome(models.Model):
     loginDate = models.DateTimeField()
     OutcomeDate1 = models.CharField(max_length=255)  # datetimes and NULL
 
+    @property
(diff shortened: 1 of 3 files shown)


esoergel (approved):

custom/enikshay/nikshay_datamigration/models.py
```diff
@@ -263,6 +263,15 @@ class Outcome(models.Model):
     loginDate = models.DateTimeField()
     OutcomeDate1 = models.CharField(max_length=255)  # datetimes and NULL
 
+    @property
+    def hiv_status(self):
+        return {
+            None: None,
+            'NULL': None,
```
czue: should these be empty strings? or it doesn't matter?

nickpell: it can be anything falsy since I'm only assigning ```hiv_status``` if it is truthy

PR #14369, opened 2016-12-29
process patients in sorted order

nickpell:
to make it easier to test on a specific subset of patients

@czue @proteusvacuum cc: @esoergel


czue:
:+1:

PR #14370, opened 2016-12-29
fix named expressions destructive serialization

czue:
http://manage.dimagi.com/default.asp?244625

should not have just worked around the test in https://github.com/dimagi/commcare-hq/pull/14363/

:hankey: :disappointed: 

@orangejenny  / anyone


czue:
@esoergel @benrudolph @nickpell and whoever else is online, important that this get merged before next deploy otherwise it might cause people to lose their named expressions. have already hotfixed india and I think prod hasn't been deployed in a long time.

PR #14371, opened 2016-12-29
trigger analytics for app preview only on click events

biyeun:
@orangejenny / @benrudolph 
cc @czue 

- fixes analytics to only trigger on click events
@dimagi/product


benrudolph (approved):
annoying, but makes sense 👍

PR #14375, opened 2016-12-30
use gettext in case detail config

orangejenny:
Noticed this file is light on translations.

@czue


czue:
our west africa office thanks you. :senegal:

benrudolph:
i always wonder whether they love or hate us when we put a giant pile of strings in transifex :writing_hand:

PR #14380, opened 2016-12-30
More usage of gettext

orangejenny:
Friday afternoon refactor: grep for user-facing strings that aren't using `gettext`. Doesn't catch them all, but makes decent headway.

@czue / @biyeun

### DIFF
--- a/corehq/apps/reports/static/reports/javascripts/language_field.js
+++ b/corehq/apps/reports/static/reports/javascripts/language_field.js
@@ -1,7 +1,5 @@
 $(function() {
     $('.languageselect').select2({
-        // TODO: Translate this string.
-        // It's hard because this is static.
-        placeholder: "Select a language..."
+        placeholder: gettext("Select a language..."),
     });
 });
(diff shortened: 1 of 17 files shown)


corehq/apps/reports/static/reports/javascripts/language_field.js
```diff
@@ -1,7 +1,5 @@
 $(function() {
     $('.languageselect').select2({
-        // TODO: Translate this string.
-        // It's hard because this is static.
```
benrudolph: :shipit:

benrudolph (commented):
👍 on ✔️ . ❤️ friday afternoon refactors

czue (approved):
this is great! 

build fail seems related. 

```
Warning: ReferenceError: Can't find variable: gettext
    at undefined (http://localhost:8000/static/app_manager/js/form_workflow.js:53)
    at hqDefine (http://localhost:8000/static/hqwebapp/js/hqModules.js:45)
    at undefined (http://localhost:8000/static/app_manager/js/form_workflow.js:189) Use --force to continue.
Aborted due to warnings.
```

PR #14381, opened 2016-12-30
populate treatment_initiated property

nickpell:
@czue @proteusvacuum


czue:
huh, we are just always assuming initiated?

nickpell:
yeah, I think it's correct to assume that

PR #14382, opened 2017-01-01
Mk 243862 reassign cases manage forms

mkangia:
Branching out from https://github.com/dimagi/commcare-hq/pull/14206 to make location safe:
1. Reassign Cases
2. Manage Forms


corehq/apps/data_interfaces/interfaces.py
```diff
@@ -85,12 +87,22 @@ def rows(self):
...
+        else:
+            active_users = self.get_all_users_by_domain(user_filter=tuple(HQUserType.all()), simplified=True)
+            context.update(groups=[dict(ownerid=group.get_id, name=group.name, type="group")
+                                   for group in self.all_case_sharing_groups],
+                           )
         context.update(
             users=[dict(ownerid=user.user_id, name=user.username_in_report, type="user")
                    for user in active_users],
-            groups=[dict(ownerid=group.get_id, name=group.name, type="group")
-                    for group in self.all_case_sharing_groups],
```
esoergel: lol, this old dict format is a throwback, huh?

corehq/apps/data_interfaces/interfaces.py
```diff
@@ -85,12 +87,22 @@ def rows(self):
...
         context = super(CaseReassignmentInterface, self).report_context
-        active_users = self.get_all_users_by_domain(user_filter=tuple(HQUserType.all()), simplified=True)
+        if not self.request.can_access_all_locations:
+            accessible_location_ids = (SQLLocation.active_objects.accessible_location_ids(
+                self.request.domain,
+                self.request.couch_user)
+            )
+            accessible_user_ids = user_ids_at_locations(accessible_location_ids)
+            active_users = self.get_all_users_by_domain(user_filter=tuple(HQUserType.all()), simplified=True,
+                                                        user_ids=tuple(accessible_user_ids))
```
esoergel: Non-blocking, but I'm concerned about the performance of this call.  I just looked at the definition of `get_all_users_by_domain`, and it does `CommCareUser.get` for every id.  That means potentially hundreds or thousands of database hits.  It would be faster to do this in bulk.  Something like this should do it:

```python
user_query = UserES().domain(domain).location(accessible_location_ids)
active_users = get_simplified_users(user_query)
```

Or some other way of calling `_report_user_dict`.  (I just ran in to something similar recently and added `get_simplified_users` to make it easy to go from ES queries to report users)

corehq/apps/data_interfaces/interfaces.py
```diff
@@ -178,6 +178,7 @@ def selected(self):
         return FormManagementMode(self.request.GET.get(self.slug)).mode_name
 
 
+@location_safe
```
esoergel: :+1: 
Looks like the queries happen in `SubmitHistoryMixin`, which should be location safe already

corehq/apps/data_interfaces/tasks.py
```diff
@@ -53,6 +54,7 @@ def get_ids_from_url(query_string, domain, couch_user):
         _request.user = couch_user.get_django_user()
         _request.domain = domain
         _request.couch_user.current_domain = domain
+        _request.can_access_all_locations = can_access_all_locations
```
esoergel: You can also use `couch_user.has_permission(domain, 'access_all_locations')` if you wanted to avoid passing in another parameter.

corehq/apps/data_interfaces/views.py
```diff
@@ -498,7 +500,8 @@ def post(self, request, *args, **kwargs):
             mode,
             self.domain,
             self.request.couch_user,
-            form_ids_or_query_string
+            form_ids_or_query_string,
+            self.request.can_access_all_locations
         )
```
esoergel: This should probably check permissions on the forms being updated before calling the task.  How does that happen currently?

corehq/apps/data_interfaces/tasks.py
mkangia: aha! nice. thanks

corehq/apps/data_interfaces/interfaces.py
mkangia: Thats right. Thank You @esoergel.
Using the first suggestion. I think `get_all_users_by_domain` just handles the case where a user is not found and gives a temp user but i don't think that would be an issue here.

mkangia: Moreover, is there a way to improve `get_all_users_by_domain` as well?

corehq/apps/data_interfaces/interfaces.py
```diff
@@ -83,12 +88,22 @@ def rows(self):
     @property
     def report_context(self):
         context = super(CaseReassignmentInterface, self).report_context
-        active_users = self.get_all_users_by_domain(user_filter=tuple(HQUserType.all()), simplified=True)
+        if not self.request.can_access_all_locations:
+            accessible_location_ids = (SQLLocation.active_objects.accessible_location_ids(
+                self.request.domain,
+                self.request.couch_user)
+            )
+            accessible_user_ids = user_ids_at_locations(accessible_location_ids)
```
dimagimon: F841 local variable 'accessible_user_ids' is assigned to but never used

corehq/apps/data_interfaces/views.py
mkangia: @esoergel i am not sure if something like this already exist and have not added any check as of now.
If need to add one what should be check to ensure all cases are covered. Thinking about:
1. form user_id should be the user's id

corehq/apps/data_interfaces/interfaces.py
mkangia: yup! this was just to allow page access to restricted user.

corehq/apps/data_interfaces/interfaces.py
```diff
@@ -7,7 +7,10 @@
 
 from corehq.apps.app_manager.const import USERCASE_TYPE
 from corehq.apps.es import cases as case_es
+from corehq.apps.es.users import user_ids_at_locations, UserES
```
dimagimon: F401 'user_ids_at_locations' imported but unused

corehq/apps/data_interfaces/interfaces.py
mkangia: @esoergel whats that?

corehq/apps/data_interfaces/interfaces.py
esoergel: > Moreover, is there a way to improve get_all_users_by_domain as well?
Haha, probably!  There may be some easy optimizations, I'm not sure.  That function also does way too much, in my opinion, I'd be interested to see if we could replace most usages with much simpler utilities (like `CommCareUser.by_domain`).

corehq/apps/data_interfaces/interfaces.py
esoergel: Oh, nothing, just commenting that this uses a `dict` constructor instead of a `dict` literal:
```python
dict(ownerid=group.get_id, name=group.name, type="group")
# vs
{'ownerid': group.get_id, 'name': group.name, 'type': "group"}
```
(Totally fine to leave as-is, it's just unusual code I don't see often)

esoergel (approved):
cool, looks good to me

corehq/apps/data_interfaces/views.py
esoergel: Yeah, I think the form user ID should come from an accessible user.

esoergel:
Oh whoops, except for this issue: https://github.com/dimagi/commcare-hq/pull/14382#discussion_r94460791

corehq/apps/data_interfaces/interfaces.py
mkangia: Aha!

mkangia:
Thank You @esoergel . Added a check at view level to ensure accessible forms accessed. Also moved the logic to fetch form ids from archiving to view. Changes [here](https://github.com/dimagi/commcare-hq/pull/14382/files#diff-9f0c2d0ac3b32ff875ff48f73f7b3e10R500)

corehq/apps/data_interfaces/views.py
```diff
@@ -511,6 +519,38 @@ def post(self, request, *args, **kwargs):
             )
         )
 
+    def inaccessible_forms_accessed(self, xform_ids, domain, couch_user):
```
esoergel: This should have something like this at the top:
```python
if couch_user.has_permission(self.domain, 'access_all_locations'):
    return True
```
so we don't needlessly run this logic (and that query) on non-restricted users.

mkangia: oops! added that earlier but got lost when moving code. thanks ethan.

esoergel:
:+1:

mkangia:
@ethan Post your last review, i just added ability for a restricted user to assign case to accessible locations as well.
Tested all and this looks good. Merging this in base branch.

PR #14390, opened 2017-01-02
Refactor formplayer breadcrumbs

benrudolph:
@biyeun @wpride this refactors some of the HTML in formplayer to have the formplayer templates inside the menu navigation. this allows us to use the menu breadcrumbs while rendering form entry, which means we have breadcrumbs that actually work. i realized this was also necessary for form entry error handling. we were hiding the menu navigation, which contained the cloudcare notifcations div, so no errors were being shown during form entry. broken up by commit. hold on merge because i want to ensure with B that style for one question per screen are OK

screen shot 2017-01-02 at 4 35 18 pm

needs this guy: https://github.com/dimagi/formplayer/pull/232

cc: @millerdev

### DIFF
--- a/corehq/apps/cloudcare/templates/form_entry/templates.html
+++ b/corehq/apps/cloudcare/templates/form_entry/templates.html
@@ -56,8 +56,8 @@
 </script>
 
 <script type="text/html" id="form-fullform-ko-template">
-    <div class="breadcrumb-form-container"
-         data-bind="css: { 'breadcrumb-form-single-question': showInFormNavigation }">
+    <div class="webforms-nav-container"
+         data-bind="css: { 'webforms-nav-single-question': showInFormNavigation }">
       <div class="webforms-nav"
            data-bind="template: { name: 'form-navigation-ko-template' }"></div>
     </div>
--- a/corehq/apps/cloudcare/templates/form_entry/templates.html
+++ b/corehq/apps/cloudcare/templates/form_entry/templates.html
@@ -93,7 +93,6 @@
 </script>
 
 <script type="text/html" id="instance-viewer-ko-template">
-        <hr/>
         <div id="instance-xml-home" class="debugger" data-bind="
             css: {
                 'debugger-minimized': isMinimized,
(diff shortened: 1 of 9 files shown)


corehq/apps/style/static/preview_app/less/preview_app/form.less
```diff
@@ -40,12 +40,7 @@
 }
 
 .form-single-question {
-  padding-top: 66px;
-  min-height: 443px;
-}
-
-.preview-tablet-mode .form-single-question {
-  min-height: 642px;
```
benrudolph: @biyeun it's kind of hard to tell what these numbers are supposed to be representing. i think the styles look fine now, but would be good if you took a look or could point me to why those are needed

wpride:
:tight:

corehq/apps/style/static/preview_app/less/preview_app/form.less
biyeun: it's so that the background color goes all the way to the bottom. I don't think this should be removed.

benrudolph: ah ok, i thought that's what it was doing, but if it's too large it makes the screen for OQPS scrollable

corehq/apps/style/static/preview_app/less/preview_app/form.less
```diff
@@ -1,7 +1,8 @@
 .form-container {
-  padding-top: 35px;
   background-color: white;
-  min-height: 403px;
+  min-height: -webkit-calc(~"100vh - " (@webforms-nav-height + @webforms-breadcrumb-height));
+  min-height: -moz-calc(~"100vh - " (@webforms-nav-height + @webforms-breadcrumb-height));
+  min-height: calc(~"100vh - " (@webforms-nav-height + @webforms-breadcrumb-height));
```
benrudolph: @biyeun added these calcs which give you this:

screen shot 2017-01-03 at 3 48 23 pm

corehq/apps/cloudcare/templates/form_entry/templates.html
```diff
@@ -93,7 +93,6 @@ <h1 class="title" data-bind="text: title, visible: !showInFormNavigation()"></h1
 </script>
 
 <script type="text/html" id="instance-viewer-ko-template">
-        <hr/>
```
benrudolph: took me ages to figure out why the phone was scroll 1px... damn you `hr`!

PR #14305, opened 2016-12-21
Use Keyword instead of SurveyKeyword

gcapalbo:
Updates all messaging references to keywords to use the ORM model rather than the Couch model.

Opening for review only, don't merge yet - I need to run the sync in all environments first and do a little bit of testing.

@emord

PR #14391, opened 2017-01-02
Stop two-way sync of keywords

gcapalbo:
Stops the couch <-> postgres sync for keywords as everything is using the postgres model now.

Once https://github.com/dimagi/commcare-hq/pull/14305 gets merged, I'll merge master to reduce the diff here. No need to review until that happens.

This is ready for release but I'd like to let https://github.com/dimagi/commcare-hq/pull/14305 be in prod for a few days before we do so.

@emord


gcapalbo:
https://github.com/dimagi/commcare-hq/pull/14305 has been released and seems to be going well; this is ready to go from my end

gcapalbo:
code buddies @benrudolph @millerdev

millerdev (approved):
Don't know much about this code, but removing so many lines looks just great!

PR #14393, opened 2017-01-02
remove iso_string_to_datetime soft assert

dannyroberts:
it has been firing regularly from the beginning
and has never been actionable

due to old date time formats used in old docs


czue (approved):
kicked build

PR #14394, opened 2017-01-02
Use EMW Filter in Worker Activity Report

esoergel:
http://manage.dimagi.com/default.asp?244563
Behind a toggle.  Needed somewhat urgently for ICDS.  I'll be throwing this up on staging too.
@czue @sravfeyn

### DIFF
--- a/corehq/apps/reports/filters/users.py
+++ b/corehq/apps/reports/filters/users.py
@@ -278,6 +278,7 @@ class ExpandedMobileWorkerFilter(BaseMultipleOptionFilter):
         user_ids = cls.selected_user_ids(mobile_user_and_group_slugs)
         user_types = cls.selected_user_types(mobile_user_and_group_slugs)
         group_ids = cls.selected_group_ids(mobile_user_and_group_slugs)
+        location_ids = cls.selected_location_ids(mobile_user_and_group_slugs)
 
         user_type_filters = []
         if HQUserType.ADMIN in user_types:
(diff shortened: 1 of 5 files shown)


corehq/apps/reports/util.py
```diff
@@ -235,6 +235,16 @@ def parts():
             is_active=user.get('is_active', None)
         )
 
+def get_simplified_users(user_es_query):
```
dimagimon: E302 expected 2 blank lines, found 1

czue:
note for reviewer in case you get to this before me: please use squash merge so it's easier to hotfix

corehq/apps/reports/filters/users.py
```diff
@@ -295,9 +296,13 @@ def user_es_query(cls, domain, mobile_user_and_group_slugs):
             return q.OR(*user_type_filters)
         else:
             # return matching user types and exact matches
+            location_ids = list(SQLLocation.active_objects
+                                .get_locations_and_children(location_ids)
+                                .location_ids())
             id_filter = filters.OR(
                 filters.term("_id", user_ids),
                 filters.term("__group_ids", group_ids),
+                user_es.location(location_ids),
```
czue: does this change the behavior of this filter for all reports that use it? and if so, is that desired?

czue (approved):
looks good to me - just one clarifying question

corehq/apps/reports/filters/users.py
czue: offlined with @benrudolph and we convinced ourselves that this is better behavior.

benrudolph:
also noting this should fix this as well: http://manage.dimagi.com/default.asp?244877#1267846

sravfeyn:
Should this be tagged with `To be patched on NIC` or is there a change in patch process?

benrudolph:
already patched

sravfeyn:
Sweet!

corehq/apps/reports/filters/users.py
esoergel: :+1:

PR #14395, opened 2017-01-02
Use a different 403 message for missing location assignment

esoergel:
@sheelio to confirm text.
@sravfeyn @mkangia @proteusvacuum

There are three situations where we 403:

1. A restricted user attempts to access a non-whitelisted feature
2. A restricted user does not have an assigned location
3. A restricted user attempts to access a whitelisted page, but for an unallowed location

Currently all three scenarios receive the same message:

> This project has restricted data access rules. Please contact your project administrator to be assigned access to data in this project. More information is available <a href="{link}">here</a>.

This PR changes it to display the following message for situation 2:

> Your user role allows you to access data based on your assigned location in the organization hierarchy. You do not currently have an assigned location, and will be unable to access CommCareHQ until that is corrected. Please contact your project administrator to be assigned to a location.

I think we could also do with better messages for the other scenarios, happy to hear suggestions.

### DIFF
--- a/corehq/apps/locations/middleware.py
+++ b/corehq/apps/locations/middleware.py
@@ -1,4 +1,13 @@
+from django.utils.translation import ugettext_lazy
+from corehq.apps.hqwebapp.views import no_permissions
 from .permissions import is_location_safe, location_restricted_response
+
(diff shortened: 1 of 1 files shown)


benrudolph (approved):
:100: that error message makes a lot of sense to me

PR #14396, opened 2017-01-03
prevent modifying non-custom software plans

nickpell:
Solves the underlying issue in http://manage.dimagi.com/default.asp?244858 @czue 

After the migration gets run, I'll check for any existing subscriptions on the default plans if they're on the custom software plan versions and flag with Ops any that look wrong.


nickpell:
This will fix self-service subscriptions; fixing the FM interface for contracted subscriptions will require one more PR to put custom report builder plans into the code

czue:
nice, thanks

PR #14397, opened 2017-01-03
Locations for pull_users_and_groups

benrudolph:
@czue @esoergel this makes locations work with reports that use `pull_users_and_groups`. should fix the submit history report http://manage.dimagi.com/default.asp?244877

cc: @millerdev

### DIFF
--- a/corehq/apps/reports/filters/users.py
+++ b/corehq/apps/reports/filters/users.py
@@ -7,7 +7,7 @@ from corehq.apps.domain.models import Domain
 from corehq.apps.groups.models import Group
 from corehq.apps.reports.util import namedtupledict
 from corehq.apps.users.models import CommCareUser
-from corehq.apps.locations.dbaccessors import get_user_ids_by_location
+from corehq.apps.es.users import user_ids_at_locations_and_descendants
 from corehq.util import remove_dups, flatten_list
 from dimagi.utils.decorators.memoized import memoized
 from corehq.apps.commtrack.models import SQLLocation
@@ -323,8 +323,7 @@ class ExpandedMobileWorkerFilter(BaseMultipleOptionFilter):
         users = []
 
         if location_ids:
-            for location_id in location_ids:
-                user_ids |= set(get_user_ids_by_location(domain, location_id))
+            user_ids |= set(user_ids_at_locations_and_descendants(location_ids))
 
         if limit_user_ids:
             user_ids = set(limit_user_ids).intersection(user_ids)
--- a/corehq/apps/reports/filters/users.py
+++ b/corehq/apps/reports/filters/users.py
@@ -7,6 +7,7 @@ from corehq.apps.domain.models import Domain
 from corehq.apps.groups.models import Group
 from corehq.apps.reports.util import namedtupledict
 from corehq.apps.users.models import CommCareUser
+from corehq.apps.locations.dbaccessors import get_user_ids_by_location
 from corehq.util import remove_dups, flatten_list
 from dimagi.utils.decorators.memoized import memoized
 from corehq.apps.commtrack.models import SQLLocation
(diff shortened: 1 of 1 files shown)


corehq/apps/reports/filters/users.py
```diff
@@ -315,13 +316,18 @@ def user_es_query(cls, domain, mobile_user_and_group_slugs):
...
-        user_ids = cls.selected_user_ids(mobile_user_and_group_slugs)
+        user_ids = set(cls.selected_user_ids(mobile_user_and_group_slugs))
         user_types = cls.selected_user_types(mobile_user_and_group_slugs)
         group_ids = cls.selected_group_ids(mobile_user_and_group_slugs)
+        location_ids = cls.selected_location_ids(mobile_user_and_group_slugs)
         users = []
 
         if limit_user_ids:
-            user_ids = set(limit_user_ids).intersection(set(user_ids))
+            user_ids = set(limit_user_ids).intersection(user_ids)
```
czue: not totally sure how `limit_user_ids` is meant to work, but if the purpose is to limit the results should it not come after the location expansion?

benrudolph: mm i wasn't sure either, it doesn't seem to limit the user ids for groups though either

benrudolph: bug or feature?

benrudolph: looking at where it's used, i think it should limit both locations and groups. will update

benrudolph: https://github.com/dimagi/commcare-hq/pull/14397/commits/37b3c21d53ba02b2014fcd523c8f214e19198ccd -- this function could use some tests. going to leave for now so we can patch up icds, but will followup with a test

millerdev (approved):

corehq/apps/reports/filters/users.py
```diff
@@ -315,13 +316,18 @@ def user_es_query(cls, domain, mobile_user_and_group_slugs):
...
-        user_ids = cls.selected_user_ids(mobile_user_and_group_slugs)
+        user_ids = set(cls.selected_user_ids(mobile_user_and_group_slugs))
         user_types = cls.selected_user_types(mobile_user_and_group_slugs)
         group_ids = cls.selected_group_ids(mobile_user_and_group_slugs)
+        location_ids = cls.selected_location_ids(mobile_user_and_group_slugs)
         users = []
 
+        if location_ids:
+            for location_id in location_ids:
+                user_ids |= set(get_user_ids_by_location(domain, location_id))
```
esoergel: This will return only users assigned directly to a location, not to its children.  I think you want a bulk lookup probably in ES.

corehq/apps/reports/filters/users.py
```diff
@@ -343,10 +349,14 @@ def pull_users_and_groups(cls, domain, mobile_user_and_group_slugs,
...
         user_dict = {}
         for group in groups:
-            user_dict["%s|%s" % (group.name, group._id)] = util.get_all_users_by_domain(
+            users_in_group = util.get_all_users_by_domain(
                 group=group,
                 simplified=True
             )
+            if limit_user_ids:
+                users_in_group = filter(lambda user: user['user_id'] in limit_user_ids, users_in_group)
+            user_dict["%s|%s" % (group.name, group._id)] = users_in_group
```
esoergel: The return value of this method is a little strange.  I just checked usages, and it looks like all but one use only `combined_users`, the remaining one uses `combined_users` and `admin_and_demo_users`.  It's probably out of scope to refactor this, but it's worth noting that the other properties don't appear to be used anywhere (I'd of course look closer before counting on that, though).

esoergel (changes requested):
One blocking comment.  @mkangia actually also just implemented this same functionality here:
https://github.com/dimagi/commcare-hq/pull/14206/files#diff-07031c94878b608626b437e634d152ddR317
You could pull from his changes to get the set of users assigned to location descendants too.

corehq/apps/reports/filters/users.py
benrudolph: ah right, always get tripped up by that function name. maybe a convention that suggests that this isn't the method you want to use most of the time (my assumption is that most of the time you want to get all user_ids for that location and its children, but that could be wrong).

corehq/apps/reports/filters/users.py
benrudolph: yeah wanted to stay away from doing any larger refactors in the PR, but agree this function could use some love

corehq/apps/reports/filters/users.py
benrudolph: https://github.com/dimagi/commcare-hq/pull/14397/commits/73bf2f465097931106030d6741937a06f728f50d

esoergel: In reporting that's probably true.  For management you often want users assigned to a particular location.  What kind of convention were you thinking of?

esoergel (approved):

corehq/apps/reports/filters/users.py
benrudolph: right, that makes sense. not sure, but something that was explicit like the sister call:
```
user_ids_at_locations_and_descendants
user_ids_at_locations_and_no_descendants
```
that's kind of verbose and i'm not sure how much i love it, but it would definitely make it clear to the caller that you are not getting the entire tree

PR #14399, opened 2017-01-03
add metric for submissions timings

dannyroberts:


dannyroberts:
@snopoke not sure what kind of testing we want/need to do with this. Probably want to test that submissions still work, but won't be able to test that the datadog metric is useful until we deploy to production.

I believe that adding 'backend' as a tag is actually a perfect fit; we can then pretty easily then graph the overall or just sql or just couch separately. Adding number of cases as a tag is likely not to be that helpful. The only thing I'm thinking is if datadog lets graph calculations that include the value of a tag (i.e. graph `metric / (1 + #cases)`), but I don't think it does.

snopoke:
Nice. You can certainly do those kind of calculations with separate metrics but I don't think you can do it with a tag value.

Nothing wrong with your implementation but you can actually use the statsd client to do the timing directly:
```
with statsd.timed('user.query.time'):
        pass
```

https://github.com/DataDog/datadogpy/blob/master/datadog/dogstatsd/base.py#L183

dannyroberts:
OK I decided to move 'cases' out of the tag and instead submit a second metric, 'normalized_timings' as well, which divides by (1 + cases) to get conceptually "time per thing (form or case) saved".

I'd seen `statsd.timed` in the docs, but think chose against it because
- I wanted to filter out failures by only submitting on a 201
- I wanted what I'm submitting to depend on the result of the operation being timed (`len(cases)`)
- I wanted to use a gauge instead of a `timing`/histogram

I think gauge is the right thing (that's what we use in couch.timings for example); not sure if histogram works but I think it's more for being able to say "the timings from the last day fall into this distribution: ..."

Fixed the build (glad to see things are tested at this level) and I think I'm happy with these metrics.

dannyroberts:
Going to put this on staging and submit a form. The only other concern is that sending these two metrics will itself noticeably reduce submission performance (or overload statsd). We can take a look at https://app.datadoghq.com/metric/explorer?live=true&page=0&is_auto=false&from_ts=1483303972243&to_ts=1483476772243&tile_size=m&exp_metric=nginx.timings&exp_scope=url%3A%2Fa%2F_%2Freceiver%2Fsecure%2F_%2F&exp_agg=avg&exp_row_type=metric around the deploy to see if there's a clear jump.

dannyroberts:
Confirming that a submission works on staging with this deployed

dannyroberts:
@snopoke, happy with this as a first pass? Any concerns with performance beyond what would be caught by eyeballing [nginx timings of the submit url](https://app.datadoghq.com/metric/explorer?live=true&page=0&is_auto=false&from_ts=1483303972243&to_ts=1483476772243&tile_size=m&exp_metric=nginx.timings&exp_scope=url%3A%2Fa%2F_%2Freceiver%2Fsecure%2F_%2F&exp_agg=avg&exp_row_type=metric)?

czue:
nice i'm excited for this

snopoke (approved):

snopoke:
Looks good to me. There shouldn't be any performance issues since sending data to statsd is a datagram socket send to a local port.

dannyroberts:
Looks like we're getting good data from this: https://app.datadoghq.com/dash/238629/submissions-timings-couch-vs-sql?live=true&page=0&is_auto=false&from_ts=1484668065378&to_ts=1485272865378&tile_size=m. Good to see SQL submissions are consistently faster (especially since I think they're still a minority, so there should be less load).

czue:
super cool!

snopoke:
Very cool @dannyroberts. Just added this for further comparison:
https://github.com/dimagi/commcare-hq/pull/14635

PR #14400, opened 2017-01-03
Clean up app manager js tests

orangejenny:
Noticed we still have a b2/b3 split for app manager's javascript tests, don't think it needs to exist anymore.

@biyeun / @benrudolph 
code buddy @czue


benrudolph:
nice thanks

PR #14405, opened 2017-01-03
use advanced_plan_v0 for managed and hosted plans

nickpell:
needed alongside https://github.com/dimagi/commcare-hq/pull/14396

kinda confused how the managed/hosted plans ever got set in the first place.  In any case, this ensures that the migration is consistent with what is currently on our production environments.


proteusvacuum (approved):

biyeun:
@nickpell informed me that accounting tests passed locally

PR #14409, opened 2017-01-03
Fail gracefully if a user is missing a location.

esoergel:
http://manage.dimagi.com/default.asp?240767
@sravfeyn
Sometimes users are assigned to locations that no longer exist. This shouldn't happen, but it does, and at least this way they'll be able to edit their locations and it'll fix itself.

### DIFF
--- a/corehq/apps/users/forms.py
+++ b/corehq/apps/users/forms.py
@@ -750,17 +750,14 @@ class SupplyPointSelectWidget(forms.Widget):
 
     def render(self, name, value, attrs=None):
         location_ids = value.split(',') if value else []
-        from corehq.apps.locations.util import get_locations_from_ids
-        try:
-            locations = get_locations_from_ids(location_ids, self.domain)
-        except SQLLocation.DoesNotExist:
-            locations = []
+        locations = list(SQLLocation.active_objects
+                         .filter(domain=self.domain, location_id__in=location_ids))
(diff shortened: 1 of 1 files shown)


corehq/apps/users/forms.py
```diff
@@ -750,17 +750,14 @@ def __init__(self, domain, attrs=None, id='supply-point', multiselect=False, que
...
     def render(self, name, value, attrs=None):
         location_ids = value.split(',') if value else []
-        from corehq.apps.locations.util import get_locations_from_ids
-        try:
-            locations = get_locations_from_ids(location_ids, self.domain)
-        except SQLLocation.DoesNotExist:
-            locations = []
+        locations = list(SQLLocation.active_objects
+                         .filter(domain=self.domain, location_id__in=location_ids))
         initial_data = [{'id': loc.location_id, 'name': loc.display_name} for loc in locations]
```
czue: nit picky but since you're only pulling these two fields you could use `values_list` to reduce the amount of data you need to pull from the DB

esoergel: `display_name` is a dynamic property of the model.  It adds the type name to display `Cambridge [city]`.  I thought briefly about assembling that from `'location_id', 'name', 'location_type__name'`, but that sounds like a maintenance headache, and I expect this to be < 10 locations at a time anyways.

PR #14414, opened 2017-01-03
Catch any failure when updating translations

esoergel:
@dannyroberts @biyeun

### DIFF
--- a/scripts/update-translations.sh
+++ b/scripts/update-translations.sh
@@ -32,25 +32,38 @@ fi
 echo "Pulling translations from transifex"
 tx pull -f
 
+if [[ $? -ne "0" ]]; then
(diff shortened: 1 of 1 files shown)


dannyroberts:
Looks good to me, although setting flags that make it fail harder seems like a quicker and more future-proof solution, e.g. http://redsymbol.net/articles/unofficial-bash-strict-mode/. The settings that article suggests (that have relevance to our current problem) you can get by sticking

```
#!/bin/bash
set -euo pipefail
```

at the beginning of the file. `set -e` makes any subprocess failure abort the process; `set -u` makes referencing unset variables abort the process; `set -o pipefail` a whole line fail if the command before the pipe fails. If it were me I'd probably use that over trying to catch and propagate potential failures individually.

PR #14419, opened 2017-01-04
add lazy migration for app Forms with case_references = []

dannyroberts:
http://manage.dimagi.com/default.asp?239236


dannyroberts:
@orangejenny @emord. Ran into this again on staging. If the tests pass (which I expect them to), I think this is all that's needed to solve this problem for good.

dannyroberts:
✅

orangejenny:
This is great, thanks for taking initiative.

PR #14421, opened 2017-01-04
remove test cases from migration

nickpell:
@proteusvacuum @czue


czue (approved):
nice, i think this also means the migration creates way fewer cases, yeah? 

assume reqs team is signed off on throwing these away?

nickpell:
> nice, i think this also means the migration creates way fewer cases, yeah?

oh yeah!

> assume reqs team is signed off on throwing these away?

it's been discussed via email

PR #14423, opened 2017-01-04
Rename stylesheets dir and drop some unused files

orangejenny:
@czue


dimagimon:
Could not review pull request. It may be too large, or contain no reviewable changes.

corehq/apps/hqwebapp/templates/home.html
```diff
@@ -10,8 +10,8 @@
 {% endblock %}
 {% block head %}{{ block.super }}
     <link rel="stylesheet" href="{% static 'hqwebapp/css/mainlogin.css' %}"/>
-    <link rel="stylesheet" href="{% static 'hqwebapp/stylesheets/blueprint/screen.css' %}" media="screen, projection">
```
orangejenny: I suspect this file isn't useful anymore - is there a use case for having prelogin disabled? Guessing we aren't keeping it up to date [("over 10 countries worldwide"!).](https://github.com/dimagi/commcare-hq/blob/master/corehq/apps/hqwebapp/templates/home.html#L34)

corehq/apps/hqwebapp/static/hqwebapp/stylesheets/modules.css
```diff
@@ -1,11 +0,0 @@
-/* UNICEF RapidSMS UI
```
czue: classic.

corehq/apps/hqwebapp/templates/home.html
czue: yeah I think we can get rid of it. there was a time when prelogin was a static site you had to build separately but now that it's shipped with HQ i think we can delete this.

czue (approved):
this is great thanks for digging these up!

PR #14426, opened 2017-01-04
Rename img to images

orangejenny:
Long, but all the changes are mechanical.

Merge with https://github.com/dimagi/commcarehq-prelogin/pull/110

@czue / @biyeun


corehq/apps/dump_reload/tests/test_couch_dump_load.py
```diff
@@ -128,7 +128,7 @@ def _make_doc(doc_type, domain):
 
     def test_multimedia(self):
         from corehq.apps.hqmedia.models import CommCareAudio, CommCareImage, CommCareVideo
-        image_path = os.path.join('corehq', 'apps', 'hqwebapp', 'static', 'hqwebapp', 'img', 'commcare-hq-logo.png')
+        image_path = os.path.join('corehq', 'apps', 'hqwebapp', 'static', 'hqwebapp', 'images', 'commcare-hq-logo.png')
```
dimagimon: E501 line too long (119 > 115 characters)

czue (approved):
looks good - one question though. it seems like a lot of these images moved but didn't have any references in code. how does that work? are they being included in some magic relative path somewhere or can they be deleted?

orangejenny:
I think most of them can be deleted - I hesitate to delete images but don't think I can actually come up with a good reason, and most of the non-referenced ones look like old logos or pre-B3 icons. Will do in a separate PR.

orangejenny:
Actually, might as well do it in this PR. @biyeun any objection to deleting images?

benrudolph:
knowing this is happening makes me feel better and healthier 🌻

orangejenny:
fixed conflicts

PR #14431, opened 2017-01-04
Ability to include_without_expanding to the same level as 'expand_from'

proteusvacuum:
With the "include without expanding" feature, we should be able to select the same location type or the location type that is at the same level as the "include_without_expanding" location type.
This allows us to include all "districts" in eNikshay and doing an "expand_from = district", which will expand only the locations beneath the user's assigned district. 

Hard to explain in words, but I think the test case I wrote should make sense. 

@esoergel 
CC @nickpell

### DIFF
--- a/corehq/apps/locations/static/locations/js/location_types.js
+++ b/corehq/apps/locations/static/locations/js/location_types.js
@@ -300,8 +300,11 @@ hqDefine('locations/js/location_types.js', function(){
             if (self.expand_from() !== ROOT_LOCATION_ID){
                 if (typeof(self.expand_from()) === 'undefined' || self.expand_from() === null){
                     options = self.parents().reverse();
+                    options.push(self);
                 } else {
-                    options = self.view.loc_types_by_id()[self.expand_from()].parents();
+                    var expand_from_type = self.view.loc_types_by_id()[self.expand_from()];
+                    options = expand_from_type.parents();
+                    options.push(expand_from_type);
                     options = options.reverse();
                 }
                 return options;
--- a/corehq/apps/locations/tests/test_location_fixtures.py
+++ b/corehq/apps/locations/tests/test_location_fixtures.py
@@ -76,6 +76,7 @@ class LocationFixturesTest(LocationHierarchyPerTest, FixtureHasLocationsMixin):
 
     def setUp(self):
         super(LocationFixturesTest, self).setUp()
+        delete_all_users()
         self.user = create_restore_user(self.domain, 'user', '123')
 
     def test_no_user_locations_returns_empty(self, uses_locations):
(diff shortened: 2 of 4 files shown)


proteusvacuum:
oops, will update test tomorrow.

czue (approved):

corehq/apps/locations/tests/test_location_fixtures.py
```diff
@@ -214,6 +215,21 @@ def test_include_without_expanding(self, uses_locations):
             ['Massachusetts', 'Suffolk', 'Boston', 'Revere', 'New York']
         )
 
+    def test_include_without_expanding_same_level(self, uses_locations):
+        # I want a list of all the counties, but only the cities in my county
+        self.user._couch_user.set_location(self.locations['Boston'].couch_location)
+        location_type = self.locations['Boston'].location_type
+
+        # Get all the counties
+        location_type.include_without_expanding = self.locations['Middlesex'].location_type
```
esoergel: You can also do
```python
location_type = self.location_types['county']
```
(this is fine too, just wasn't sure if you knew that was available)

corehq/apps/locations/tests/test_location_fixtures.py
```diff
@@ -214,6 +215,21 @@ def test_include_without_expanding(self, uses_locations):
...
+        # Expand downwards from my county
+        location_type.expand_from = self.locations['Middlesex'].location_type
+        location_type.save()
+        self._assert_fixture_has_locations(
+            'include_without_expanding_same_level',
+            ['Massachusetts', 'New York', 'Middlesex', 'Suffolk', 'New York City', 'Boston', 'Revere']
+        )  # (New York City is of type "county")
```
esoergel: This totally threw me off

corehq/apps/locations/static/locations/js/location_types.js
```diff
@@ -300,8 +300,11 @@ hqDefine('locations/js/location_types.js', function(){
             if (self.expand_from() !== ROOT_LOCATION_ID){
                 if (typeof(self.expand_from()) === 'undefined' || self.expand_from() === null){
                     options = self.parents().reverse();
+                    options.push(self);
                 } else {
-                    options = self.view.loc_types_by_id()[self.expand_from()].parents();
+                    var expand_from_type = self.view.loc_types_by_id()[self.expand_from()];
+                    options = expand_from_type.parents();
+                    options.push(expand_from_type);
```
esoergel: I think you can extend your same logic beyond the level of "expand_from_type".  Say you've got the setup in these tests:
```python
    location_type_names = ['state', 'county', 'city']
    location_structure = [
        ('Massachusetts', [
            ('Middlesex', [
                ('Cambridge', []),
                ('Somerville', []),
            ]),
            ('Suffolk', [
                ('Boston', []),
                ('Revere', []),
            ])
        ]),
        ('New York', [
            ('New York City', [
                ('Manhattan', []),
                ('Brooklyn', []),
                ('Queens', []),
            ]),
        ]),
    ]
```
You want a user in Cambridge to get all the cities in Massachusetts, and all counties project-wide.  In that case, `expand_from` would be `'state'`, and `include_without_expanding` would be `county`.  The way it's done here, you can't select an `include_without_expanding` that's below your `expand_from`.

esoergel (approved):

corehq/apps/locations/static/locations/js/location_types.js
proteusvacuum: Ah! that's true. I don't remember why we put this limitation on there in the first place.

PR #14434, opened 2017-01-05
Truncate mobile worker's commcarehq.org

benrudolph:
@czue this'll truncate the user's `<domain>.commcarehq.org` portion http://manage.dimagi.com/default.asp?245080

cc: @millerdev @gcapalbo

### DIFF
--- a/corehq/apps/cloudcare/static/cloudcare/js/formplayer/users/models.js
+++ b/corehq/apps/cloudcare/static/cloudcare/js/formplayer/users/models.js
@@ -11,6 +11,17 @@ FormplayerFrontend.module("Users.Models", function(Models, FormplayerFrontend, B
             }.bind(this));
         },
 
+        isMobileWorker: function() {
+            return this.username.endsWith('commcarehq.org');
+        },
+
+        getDisplayUsername: function() {
+            if (this.isMobileWorker()) {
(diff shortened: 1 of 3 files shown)


czue (approved):
:+1: thanks. even comes with tests!

corehq/apps/cloudcare/static/cloudcare/js/formplayer/users/models.js
```diff
@@ -11,6 +11,17 @@ FormplayerFrontend.module("Users.Models", function(Models, FormplayerFrontend, B
             }.bind(this));
         },
 
+        isMobileWorker: function() {
+            return this.username.endsWith('commcarehq.org');
```
gcapalbo: not a blocker, just wondering if you think it's worth grabbing this from `settings.HQ_ACCOUNT_ROOT` as is done here https://github.com/dimagi/commcare-hq/blob/24b3679279ad19093b8aca28b765eff2c5c5a8c4/corehq/apps/users/util.py#L29

benrudolph: i would, but unfortunately we don't have a great way of getting django variables into javascript files at the moment and i don't think it's worth the complexity of passing that arg around all the way to the model

PR #14435, opened 2017-01-05
don't use breadcrumbs for incomplete forms

benrudolph:
@biyeun looks like you made this header breadcrumbs. was that intentional? https://github.com/dimagi/commcare-hq/commit/dd0ed819

changed it back because i thought it looked weird
before:
screen shot 2017-01-05 at 12 54 19 pm
after:
screen shot 2017-01-05 at 12 55 24 pm

cc: @millerdev @gcapalbo

### DIFF
--- a/corehq/apps/cloudcare/static/cloudcare/js/formplayer/constants.js
+++ b/corehq/apps/cloudcare/static/cloudcare/js/formplayer/constants.js
@@ -2,4 +2,8 @@
 
 FormplayerFrontend.Constants = {
     ALLOWED_SAVED_OPTIONS: ['oneQuestionPerScreen'],
+
+    // These should match corehq/apps/cloudcare/const.py
+    WEB_APPS_ENVIRONMENT: 'web-apps',
+    PREVIEW_APP_ENVIRONMENT: 'preview-app',
 };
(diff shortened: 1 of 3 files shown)


gcapalbo (approved):

biyeun (changes requested):
@benrudolph yeah this was definitely intentional for app preview. Since you edited this in the template, it seems like that will not remain for app preview. I suggest adding the header in but not removing the breadcrumbs and hiding the breadcrumbs for webapps.

Please make sure that things look just as good on preview when doing stuff on webapps side, as this is now technically released! It sucks that our templates are not decoupled better.

biyeun:
@benrudolph I'm noticing issues with breadcrumbs all over app preview now :(

biyeun:
might be due to some of the earlier changes

benrudolph:
hey B, I'm not sure which breadcrumb errors you're talking about, could you send me a ticket? I do always test on app preview and I haven't noticed any oddities, but it's definitely possible I missed some cases, sorry about that.

A few reasons I wasn't too concerned about this change is that: 
- Incomplete Forms is not a real workflow for app preview, though is for Web Apps
- The "login as" doesn't have a breadcrumbs section (though that may just be something we need to fix)
- The breadcrumbs that do get displayed don't actually work, which to me is slightly confusing

Anyways all that being said, I agree that we should at least show the title of "Incomplete Forms" at the top. Will update this PR with that

benrudolph:
this splits out incomplete forms for web apps and preview (https://github.com/dimagi/commcare-hq/pull/14435/commits/e0c192379a0ac037c29803bae362c001fca67cd1).

screen shot 2017-01-09 at 11 25 53 am

biyeun (approved):

PR #14441, opened 2017-01-05
Make separate call for case details

wpride:
@benrudolph 

Make separate request for case details on case selection

ticket: http://manage.dimagi.com/default.asp?244524

cross-request: https://github.com/dimagi/formplayer/pull/236

cc @czue


czue:
boss

benrudolph (approved):
nice, not too invasive 👍

PR #14442, opened 2017-01-06
Only process current apps

benrudolph:
@NoahCarnahan @czue This implements a strategy Cory and I discussed the other day. Essentially it it only processes the most current applications and ignores the old ones. The old builds only give information about deleted questions, and we decided it probably wasn't worth the performance hit. It gives the ability to process the older builds if requested. I'm not set on the UX / wording yet, so I put behind a feature flag. Some scenarios I manually tested:

- You already have an export instance, the schema is deleted and then rebuilt with only latest builds. The deleted questions you have selected, stay selected
- You have an export with the most recent builds, you then process the old builds, deleted questions do indeed show up.

Here's what the current UX looks like:
At the beginning:
screen shot 2017-01-06 at 11 50 52 am
While loading:
screen shot 2017-01-06 at 11 51 46 am
After loading:
screen shot 2017-01-06 at 11 51 08 am

Probably can be reviewed all at once or commit by commit
cc: @millerdev @gcapalbo

### DIFF
--- a/corehq/apps/export/models/new.py
+++ b/corehq/apps/export/models/new.py
@@ -1046,7 +1046,8 @@ class ExportDataSchema(Document):
         app_label = 'export'
 
     @classmethod
-    def generate_schema_from_builds(cls, domain, app_id, identifier, force_rebuild=False):
+    def generate_schema_from_builds(cls, domain, app_id, identifier, force_rebuild=False,
+            only_process_current_builds=False):
         """Builds a schema from Application builds for a given identifier
 
         :param domain: The domain that the export belongs to
@@ -1065,11 +1066,13 @@ class ExportDataSchema(Document):
         else:
             current_schema = cls()
 
-        app_build_ids = cls._get_app_build_ids_to_process(
-            domain,
-            app_id,
-            current_schema.last_app_versions,
-        )
+        app_build_ids = []
+        if not only_process_current_builds:
+            app_build_ids = cls._get_app_build_ids_to_process(
+                domain,
+                app_id,
+                current_schema.last_app_versions,
+            )
         app_build_ids.extend(cls._get_current_app_ids_for_domain(domain, app_id))
 
         for app_doc in iter_docs(Application.get_db(), app_build_ids, chunksize=10):
--- a/corehq/apps/export/models/new.py
+++ b/corehq/apps/export/models/new.py
@@ -1070,10 +1070,7 @@ class ExportDataSchema(Document):
             app_id,
             current_schema.last_app_versions,
         )
-        if app_id:
-            app_build_ids.append(app_id)
-        else:
-            app_build_ids.extend(cls._get_current_app_ids_for_domain(domain))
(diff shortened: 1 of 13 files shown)


corehq/apps/export/static/export/spec/ExportInstance.spec.js
```diff
@@ -5,6 +5,9 @@ describe('ExportInstance model', function() {
     var constants = hqImport('export/js/const.js');
     var viewModels = hqImport('export/js/models.js');
     var basicFormExport, savedFormExport;
+    hqImport('hqwebapp/js/urllib.js').registerUrl(
+        "build_schema", "/a/---/data/export/build_full_schema/"
+    )
```
dimagimon: Missing semicolon. (semi)

corehq/apps/export/static/export/spec/ExportInstance.spec.js
```diff
@@ -72,6 +75,93 @@ describe('ExportInstance model', function() {
...
+                requests.push(xhr);
+            };
+        });
+
+        afterEach(function() {
+            xhr.restore();
+            clock.restore();
+        });
+
+        it('successfully check for pending build', function() {
```
benrudolph: this is more or less an integration test of the full cycle

corehq/apps/export/models/new.py
```diff
@@ -1087,7 +1093,14 @@ def generate_schema_from_builds(cls, domain, app_id, identifier, force_rebuild=F
                 identifier,
             )
 
-            current_schema.record_update(app.copy_of or app._id, app.version)
+            # Only record the version of builds on the schema. We don't care about
+            # whether or not the schema has seen the current build because that always
+            # gets processed.
+            if app.copy_of:
+                current_schema.record_update(app.copy_of, app.version)
```
benrudolph: this is kind of sneaky. we no longer record the app.version of the current app, since it's then easier to figure out how many builds we've gone through. i think this is ok because we always process the most current app

benrudolph (commented):
⚡️

NoahCarnahan (approved):
Looks good to me.

One idea I had about an alternate UX:
I doubt many people outside of Dimagi will know what "Process all applications for deleted questions" means. Perhaps this would be an improvement?:
- Split the "Show Deleted Questions and Advanced Properties" button into two separate buttons: "Show Deleted Questions" and "Show Advanced Properties".
- When the user clicks "Show Deleted Questions" there is a modal window that explains that the old builds need to be processed and that that will take some time, etc.
- After the user dismisses the modal we show the progress bar as you have (but maybe up near the "Show Deleted Questions" button.

czue:
+1 to noah's idea of splitting out deleted and advanced if it's not too complicated/hard

czue:
took a quick look and looks good!

benrudolph:
sweet, i think that UX makes sense to me, thanks

benrudolph:
Merging for now so i can split up the second portion

PR #14445, opened 2017-01-06
add owner to related cases table

czue:
this is really helpful to understand what's going on for referral type of workflows where you have cases in the hierarchy with different owners. i realize it's kind of a useless column if the app doesn't use case sharing or referrals at all, but i think the benefits are worth it. cc @dimagi/product in case anyone wants to push back.

image


benrudolph:
yeah **baby!** looks good to me

PR #14448, opened 2017-01-06
mobile ucr UI cleanup

czue:
@orangejenny I feel like you might appreciate this since I think i remember you complaining about it at some point. has been driving me nuts when I use / dev on it.

this provides some more readable stuff in mobile UCR as well as DRY's up a few things. 

Main filter choice

From
image

To
image

Auto filter choices
From
image

To
image

Date ranges
From
image

To
image


czue:
:white_check_mark:

PR #14449, opened 2017-01-06
Revert "add owner to related cases table"

proteusvacuum:
Reverts dimagi/commcare-hq#14445
Was causing the case_data table to 500:
http://manage.dimagi.com/default.asp?245175

@biyeun


czue:
dang, sorry - will fix

PR #14451, opened 2017-01-07
Drop couch lineage, only store relation in SQL

esoergel:
@czue @sravfeyn
UCR is on SQL now, so I don't think we need this anymore.
http://manage.dimagi.com/default.asp?245138

### DIFF
--- a/corehq/apps/locations/models.py
+++ b/corehq/apps/locations/models.py
@@ -403,7 +403,7 @@ class SQLLocation(SyncSQLToCouchMixin, MPTTModel):
 
     @classmethod
     def _migration_get_fields(cls):
-        return ["domain", "name", "lineage", "site_code", "external_id",
+        return ["domain", "name", "site_code", "external_id",
                 "metadata", "is_archived"]
 
     @classmethod
@@ -761,14 +761,11 @@ class Location(SyncCouchToSQLMixin, CachedCouchDocumentMixin, Document):
     latitude = FloatProperty()
     longitude = FloatProperty()
 
-    # a list of doc ids, referring to the parent location, then the
-    # grand-parent, and so on up to the root location in the hierarchy
-    lineage = StringListProperty()
-
     @classmethod
     def wrap(cls, data):
         last_modified = data.get('last_modified')
         data.pop('location_type', None)  # Only store location type in SQL
+        data.pop('lineage', None)  # Don't try to store lineage
         # if it's missing a Z because of the Aug. 2014 migration
         # that added this in iso_format() without Z, then add a Z
         # (See also Group class)
@@ -778,18 +775,19 @@ class Location(SyncCouchToSQLMixin, CachedCouchDocumentMixin, Document):
         return super(Location, cls).wrap(data)
 
     def __init__(self, *args, **kwargs):
-        from corehq.apps.locations.util import get_lineage_from_location, get_lineage_from_location_id
         if 'parent' in kwargs:
+            # if parent is in the kwargs, this was set from a constructor
+            # if parent isn't in kwargs, this was probably pulled from the db
+            # and we should look to SQL as source of truth
             parent = kwargs['parent']
             if parent:
                 if isinstance(parent, Document):
-                    lineage = get_lineage_from_location(parent)
+                    self._sql_parent = parent.sql_location
                 else:
                     # 'parent' is a doc id
-                    lineage = get_lineage_from_location_id(parent)
+                    self._sql_parent = SQLLocation.objects.get(location_id=parent)
             else:
-                lineage = []
-            kwargs['lineage'] = lineage
+                self._sql_parent = None
             del kwargs['parent']
 
         location_type = kwargs.pop('location_type', None)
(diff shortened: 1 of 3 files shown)


corehq/apps/locations/util.py
```diff
@@ -321,11 +321,3 @@ def get_locations_from_ids(location_ids, domain, base_queryset=None):
...
         raise SQLLocation.DoesNotExist('One or more of the locations was not found.')
     return locations
-
-
-def get_lineage_from_location_id(location_id):
-    return get_lineage_from_location(Location.get(location_id))
-
-
-def get_lineage_from_location(location):
-    return list(reversed(location.path))
```
esoergel: I'm not sure exactly how the original problem manifested, but this could've been a source - `path` includes `self`, but `lineage` isn't supposed to.  Additionally, `path` was defined as `lineage` + `self._id`, so I could see those working together to repeatedly add on ids.

corehq/apps/locations/models.py
```diff
@@ -778,18 +775,19 @@ def wrap(cls, data):
         return super(Location, cls).wrap(data)
 
     def __init__(self, *args, **kwargs):
-        from corehq.apps.locations.util import get_lineage_from_location, get_lineage_from_location_id
         if 'parent' in kwargs:
+            # if parent is in the kwargs, this was set from a constructor
+            # if parent isn't in kwargs, this was probably pulled from the db
+            # and we should look to SQL as source of truth
```
czue: thanks for this comment. super confusing behavior.

corehq/apps/locations/models.py
```diff
@@ -957,14 +953,18 @@ def parent_id(self):
 
     @property
     def parent(self):
-        parent_id = self.parent_location_id
-        return Location.get(parent_id) if parent_id else None
+        if hasattr(self, '_sql_parent'):
+            return self._sql_parent.couch_location if self._sql_parent else None
```
czue: is this branch just for memoization? or is there logic that has `_sql_parent` being temporarily different from `sql_location.parent`, e.g. when you instantiate something but haven't saved it yet.

corehq/apps/locations/tests/test_reupholster.py
```diff
@@ -29,28 +28,28 @@ def test_path(self):
...
-        new_district.save()
-        self.assertEqual(new_state._id, new_district.sql_location.parent.location_id)
+        original_parent = self.all_locs[1].sql_location
+        new_parent = make_loc('New York', type='state').sql_location
+        district = make_loc('NYC', type='block', parent=original_parent.couch_location).sql_location
+        self.assertEqual(original_parent.site_code, district.couch_location.parent.site_code)
+
+        district.parent = new_parent
+        district.save()
+        self.assertEqual(new_parent.site_code, district.couch_location.parent.site_code)
```
czue: are there other tests it would be worth adding here? e.g. a lineage check?

corehq/apps/locations/tests/test_reupholster.py
```diff
@@ -29,28 +28,28 @@ def test_path(self):
...
-        new_district.lineage = []
-        new_district.save()
-        self.assertEqual(None, new_district.sql_location.parent)
+        original_parent = self.all_locs[1].sql_location
+        district = make_loc('NYC', type='block', parent=original_parent.couch_location).sql_location
+        self.assertEqual(original_parent.site_code, district.couch_location.parent.site_code)
+
+        district.parent = None
+        district.save()
+        self.assertEqual(None, district.couch_location.parent)
```
czue: same as above.

corehq/apps/locations/util.py
czue: huh.

czue (approved):
looks good though is the type of change that i'm a little nervous about not having great test coverage for. curious whether you think there is more we should do to prep/test? 

One comfort is that it seems like sql hierarchies are good/consistent, so as long as that stays true then any bugs introduced would presumably be quickly addressable. did you search for write locations to `lineage` to make sure nothing tries to do that on a couch location?

corehq/apps/locations/models.py
esoergel: Yeah, for the latter case - if you instantiate it with parent, but haven't saved yet.

corehq/apps/locations/tests/test_reupholster.py
esoergel: Yeah, I suppose it couldn't hurt.  These changes are a little scary, so I'm happy to test whatever makes sense.

esoergel:
I did do a search for lineage, there are fairly few references left.  The only one that's still writing to it is `copy_domain`.  I've got another branch where I was working on rewriting that to use SQL (and test it), so I figured it made the most sense to just leave that broken briefly, since I don't think it's frequently used.

I'll add lineage checks to those test cases you highlighted, but beyond that I'm not sure what else to test.  This code alters the way it's synced when you create a couch location and save to SQL, and it switches to SQL as the source of truth for relationships.  There are a bunch of tests testing syncing to SQL, which apply here.  Let me know if you have suggestions for potential edge cases that might not be covered.

czue:
sounds good to me. i didn't have any other thoughts just figured it was worth asking the question. good to merge now or when you add those test cases

PR #14452, opened 2017-01-07
Switch copy form and xforms sections on form view actions tab

orangejenny:
Came up during hallway testing. Copying is more common than dealing with the xform, so it should go on top.

@czue / @biyeun


czue:
makes sense

PR #14454, opened 2017-01-07
Make dot interpolation support quoted strings

orangejenny:
http://manage.dimagi.com/default.asp?243540

code buddy @czue 
app manager @snopoke / @dannyroberts

### DIFF
--- a/corehq/apps/app_manager/xpath.py
+++ b/corehq/apps/app_manager/xpath.py
@@ -29,35 +29,39 @@ CASE_REFERENCE_VALIDATION_ERROR = ugettext_lazy(
 
 def dot_interpolate(string, replacement):
     """
-    Replaces non-decimal dots in `string` with `replacement`
+    Replaces non-decimal, non-quoted dots in `string` with `replacement`
     """
-    i = 0
-    quote = ""
+    quote = ""  # " or ' if inside a string, else blank
     new = ""
+    i = 0
     while i < len(string):
+        # Ignore backslash-escaped characters
         if string[i] == "\\":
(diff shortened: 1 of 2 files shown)


corehq/apps/app_manager/xpath.py
```diff
@@ -29,10 +31,40 @@
...
+            if string[i] in "\"'":
+                # We're entering a quoted string: just record the quote type
+                quote = string[i]
+                new += string[i]
+            else:
+                # Non-quote character, not inside a quoted string
+                if string[i] == ".":
+                    # Replace dot with replacement, unless this looks like a decimal number
+                    if (i == 0 or re.match(r'\D', string[i - 1])):
+                        if (i == len(string)  - 1 or re.match(r'\D', string[i + 1])):
```
dimagimon: E221 multiple spaces before operator

czue (approved):
looks good and thorough. the implementation looks like it could get a little unwieldy to maintain (especially if this gets any more complicated). the tests are great and will help a lot. just wondering if we might be able to do something simpler like first extract all the quoted bits using something [like this](http://stackoverflow.com/a/19675957/8207) and then only run the interpolation on the rest, then rejoin everything back together? Maybe that's not any better but feels like it could be a little easier to reason with if it wasn't too hard.

czue:
will let you self-merge or take a crack at my comment if it feels worth pursuing.

orangejenny:
Agree that this is a finicky piece of code, but didn't find regex solutions any better. They get messy when dealing with quotes within quotes and then when rejoining the string. Going to leave this as-is.

PR #14455, opened 2017-01-08
reveal No mobile workers message only when no mobile workers…

mkangia:
… in system

http://manage.dimagi.com/default.asp?245147

Update to show "There are currently no Mobile Workers in this project." only when there are actually no users and retain the message "No Mobile Workers were found that matched your query." when search returns no result.

buddy @dannyroberts

### DIFF
--- a/corehq/apps/users/views/mobile/users.py
+++ b/corehq/apps/users/views/mobile/users.py
@@ -673,7 +673,7 @@ class MobileWorkerListView(JSONResponseMixin, BaseUserSettingsView):
         users_query = self._user_query(query, page - 1, limit)
 
         # run with a blank query to fetch total records with same scope as in search
-        total_records = self._user_query('', 0, 0).run().total
+        total_records = self._user_query('', 0, 0).count()
         if in_data.get('showDeactivatedUsers', False):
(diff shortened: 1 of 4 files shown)


corehq/apps/users/views/mobile/users.py
```diff
@@ -671,6 +671,9 @@ def get_pagination_data(self, in_data):
 
         # backend pages start at 0
         users_query = self._user_query(query, page - 1, limit)
+
+        # run with a blank query to fetch total records with same scope as in search
+        total_records = self._user_query('', 0, 0).run().total
```
benrudolph: nit: should use `count` here instead of `run().total`. it also ensures that `size` is `0` so it doesn't fetch any results

benrudolph (commented):
looks good, one comment

corehq/apps/users/views/mobile/users.py
mkangia: Thank You @benrudolph. Did set offset and limit to 0 to ensure no results. But this definitely sounds the better way.

benrudolph: ah right, didn't realize that was what those params were doing 😄

PR #14456, opened 2017-01-08
Fix ICDS Aggregation Scripts

sheelio:
@sravfeyn 

This fixes a bug in the aggregation scripts.  Is it possible to deploy to NIC and re-run the aggregation scripts?


czue:
squashed. https://github.com/dimagi/commcare-hq/commit/fd3380c1818ac7ec06a799d408a96ec392b55e75 is what needs patching

sravfeyn:
@sheelio, I have deployed, updated migrations and re-ran the task. Everything went ok and it looks like Tableau dashboard is updated as well. I am not sure if all the issues you were aiming to fix are fixed though.

mkangia:
Nice!!

PR #14457, opened 2017-01-08
allow user to preview form only when permission to edit_data

mkangia:
http://manage.dimagi.com/default.asp?244649

buddy @dannyroberts 
@benrudolph


czue:
app manager diff tests are failing.

corehq/apps/app_manager/views/forms.py
```diff
@@ -576,6 +576,7 @@ def get_form_view_context_and_template(request, domain, form, langs, messages=me
             {'instanceId': instance.instance_id, 'instancePath': instance.instance_path}
             for instance in form.custom_instances
         ],
+        'can_preview_form': request.couch_user.has_permission(domain, 'edit_data')
```
benrudolph: any reason not to combine `can_preview_form` and `allow_cloudcare`? i believe `allow_cloudcare` is only used in this template and they seem to have the same intent

benrudolph (changes requested):
looks good, one question

corehq/apps/app_manager/views/forms.py
mkangia: @benrudolph just to keep the intent clear. `allow_cloudcare` seems picking from the form instance type whereas `can_preview_form` would be user permission.  Don't mind combining them if that looks better.

benrudolph: sounds good, i'm fine with that

benrudolph (approved):

corehq/apps/app_manager/templates/app_manager/v1/form_view_base.html
```diff
@@ -323,7 +323,7 @@
             {% trans "Edit" %}
         </button>
         {% endif %}
-        {% if allow_cloudcare %}
+        {% if can_preview_form and allow_cloudcare %}
```
snopoke: I think this should also get added to the `v2/form_view_base.html` template. Then you can run `manage.py build_app_manager_v2_diffs`

snopoke (changes requested):

corehq/apps/app_manager/templates/app_manager/v2/form_view_base.html
```diff
@@ -211,7 +211,7 @@
 
         gaTrackLink($('.breadcrumb .view-form'), 'App Builder', 'View Form', 'Breadcrumb');
 
-        {% if allow_cloudcare %}
+        {% if can_preview_form and allow_cloudcare %}
```
snopoke: This isn't the same place as was edited in the V1 file. I think you should add this edit to the V1 file and then also hide this section in the V2 file: https://github.com/dimagi/commcare-hq/blob/a5c29f2cacafcc9bd2651223d4067fa537d6c7eb/corehq/apps/app_manager/templates/app_manager/v2/form_view_base.html#L276-L281

There are also other places where you can launch the App Preview from. Search for "Preview App" and "preview_app.js".

@biyeun @orangejenny FYI that this will also affect app manager V2. Since V2 isn't released yet let us know if you'd rather handle this in a different way.

snopoke (changes requested):

corehq/apps/app_manager/templates/app_manager/v2/form_view_base.html
orangejenny: If Manish tests locally in v2 (turn on `Prototype for case management onboarding (App Manager V2)`) and verified it looks all right, that's good due diligence.

mkangia: @orangejenny The change is to hide the "Preview Form" link in case user does not have "Edit Data" permission. I tested on V2 and it seems that there is no link for preview form in v2 but it does set the url for link [here](https://github.com/dimagi/commcare-hq/blob/master/corehq/apps/app_manager/templates/app_manager/v2/form_view_base.html#L248) so the change in v2 seems to be making no difference.

mkangia: Also what Simon said makes sense. Also stated by @benrudolph [here](http://manage.dimagi.com/default.asp?244649#bugevent_1268745). So added changes for Preview App as well and made the corresponding change in [commit](https://github.com/dimagi/commcare-hq/pull/14457/commits/4876d3fd5b5c1a5421b09d86f737c4401a8ad4b6)

orangejenny:
v2 changes look good now

orangejenny (approved):

mkangia:
@snopoke offlined with @orangejenny and made quite some more changes. Should be good now. Thank you @orangejenny

mkangia:
@snopoke Changes done and build fixed. Please review.

snopoke (approved):

snopoke:
Just confirming that tested this locally before I merge (both v1 & v2)

mkangia:
@snopoke tested locally that "Show App Preview" and "Preview Form" for v1 and "Preview App" for v2 are displayed only when user has permission to "Edit data".

PR #14460, opened 2017-01-09
Release debugger

benrudolph:
@biyeun @wpride (also @snopoke since you're in charge of the product portion of this) This finishes off implementing the debugger release: https://docs.google.com/document/d/1PWznhCNj65zKUPo7PGCDwyVyYSXUJQwbjiKi8vJKZvM/edit#. I left the UI mostly the same because the proposed UI doesn't make as much sense for app preview, and i didn't think it was a show stopper. all analytics have already been added. can be read 🐟 or 🏠 

this PR does the following:

- release debugger from behind the flag (flag still in use for old cloudcare, maybe we should get rid of entirely?)
- puts the debugger under edit_data permissions as the spec says
- fixes form height scrolling issues. here's what it looks like now for OQPS and regular
screen shot 2017-01-09 at 11 56 25 am
screen shot 2017-01-09 at 11 55 51 am
- fixes the setting that allows you to turn off the auto sync option. this may be something we want to be able to set globally, but for now i decided to just leave the functionality the same as it was before

- [x] Only  request on open of debugger and only do auto update when debugger is open
- [x] Do pass on docs (https://confluence.dimagi.com/display/commcarepublic/Using+Web+Apps and https://confluence.dimagi.com/display/commcarepublic/Beginner+Tutorial+Part+4+-+Testing+in+Preview)

cc: @millerdev @gcapalbo

### DIFF
--- a/corehq/apps/cloudcare/static/cloudcare/js/form_entry/fullform-ui.js
+++ b/corehq/apps/cloudcare/static/cloudcare/js/form_entry/fullform-ui.js
@@ -446,6 +446,8 @@ Formplayer.ViewModels.CloudCareDebugger = function() {
     self.isMinimized = ko.observable(true);
     self.instanceXml = ko.observable('');
     self.formattedQuestionsHtml = ko.observable('');
+    // Whether or not to auto update after every question answer
+    self.autoUpdate = ko.observable(true);
     self.toggleState = function() {
         self.isMinimized(!self.isMinimized());
         // Wait to set the content heigh until after the CSS animation has completed.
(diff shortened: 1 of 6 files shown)


benrudolph:
Gave a go at some documentation: https://confluence.dimagi.com/display/ccinternal/CloudCare+Data+Preview

It's fairly simple, mainly cause I think it's fairly simple tool. Now that this is not being released for app preview, I think it's usage will pretty much only be Dimagi users. In that case, I'm not too worried about the docs. Most users who can `edit_data` are web users who are building apps, and as app preview becomes more common, fewer people will be using form preview. Still going to coordinate with Clark on refining the docs and figuring out where it should live and be linked to

czue:
woohoo!

PR #14463, opened 2017-01-09
Use nimbus API to validate forms

benrudolph:
@snopoke how do you feel about bumping this? i think formplayer can handle the load


snopoke:
:+1: the only reason we made this random was because of formplayer performance so if you're happy bumping then that's fine with me.

benrudolph:
yep, these requests seem to be very fast: https://app.datadoghq.com/dash/231591/formplayer-performance?live=true&page=0&is_auto=false&from_ts=1483968342796&to_ts=1483971942796&tile_size=m

snopoke:
sweet

PR #14465, opened 2017-01-09
warn if overriding case list reg workflow

snopoke:
http://manage.dimagi.com/default.asp?243646

case list reg form warning

@kaapstorm @amstone326


amstone326:
Nice! Thanks for following up with this!

millerdev (approved):

PR #14469, opened 2017-01-09
fixing view for exports modal

calellowitz:
@benrudolph 
A fix for the export modal. Let me know if this isnt what you were intending.

Unfortunately this also requires a main db reindex so we should hold off on merge until friday.


benrudolph:
👍 with this you were able to find the linked applications in both the form and case modal? that code is fairly old so don't have all the context on it

calellowitz:
@benrudolph yep

PR #14476, opened 2017-01-10
Don't cache RelatedDocType get_value

benrudolph:
@czue http://manage.dimagi.com/default.asp?245341 this fixes the caching bug you were seeing. huge shout out to @snopoke for walking down that rabbit hole with me!

cc: @millerdev @gcapalbo

### DIFF
--- a/corehq/apps/userreports/tests/test_expressions.py
+++ b/corehq/apps/userreports/tests/test_expressions.py
@@ -743,13 +743,13 @@ class RelatedDocExpressionTest(SimpleTestCase):
         self.test_simple_lookup()
 
         my_doc = self.database.get('my-id')
+
+        context = EvaluationContext(my_doc, 0)
+        self.assertEqual('foo', self.expression(my_doc, context))
+
+        my_doc = self.database.get('my-id')
         self.database.mock_docs.clear()
-
-        self.assertEqual({}, self.database.mock_docs)
-        self.assertEqual('foo', self.expression(my_doc, EvaluationContext(my_doc, 0)))
-
-        same_expression = ExpressionFactory.from_spec(self.spec)
-        self.assertEqual('foo', same_expression(my_doc, EvaluationContext(my_doc, 0)))
+        self.assertEqual('foo', self.expression(my_doc, context))
 
 
 class RelatedDocExpressionDbTest(TestCase):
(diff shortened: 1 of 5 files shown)


snopoke (approved):
So not related to SQL case backend at all. Surprised we haven't seen it before.

czue (approved):
you guys :moyai: !

looks good thanks. so basically quickcache was caching the output of `get_value` and that propagated across updates but now we cache the doc in the session so it's correct per change?

benrudolph:
yep you got it

czue:
build fail real (probably can either delete the test or update to test the doc-level caching)
```
FAIL: test_caching (corehq.apps.userreports.tests.test_expressions.RelatedDocExpressionTest)
  File "/mnt/commcare-hq-ro/corehq/apps/userreports/tests/test_expressions.py", line 749, in test_caching
    self.assertEqual('foo', self.expression(my_doc, EvaluationContext(my_doc, 0)))
AssertionError: 'foo' != None
``

**Comment by benrudolph** [comment]
shoot, looking

File: corehq/apps/userreports/tests/test_expressions.py [hunk 1]
```diff
@@ -743,13 +743,13 @@ def test_caching(self):
...
-        self.assertEqual({}, self.database.mock_docs)
-        self.assertEqual('foo', self.expression(my_doc, EvaluationContext(my_doc, 0)))
+        context = EvaluationContext(my_doc, 0)
+        self.assertEqual('foo', self.expression(my_doc, context))
 
-        same_expression = ExpressionFactory.from_spec(self.spec)
-        self.assertEqual('foo', same_expression(my_doc, EvaluationContext(my_doc, 0)))
+        my_doc = self.database.get('my-id')
+        self.database.mock_docs.clear()
+        self.assertEqual('foo', self.expression(my_doc, context))
```
czue: does this actually test caching at all anymore? was hard for me to wrap my head around how, if it does. not blocking though and want to get this in so merging.

benrudolph: yes, if you remove the caching portion the context tries to fetch the document from the database and returns an empty result. if this line were to change to `self.assertEqual('foo', self.expression(my_doc, EvaluationContext(my_doc, 0)))` it'd fail

czue: oh nice, i see. thanks for explaining.

PR #14477, opened 2017-01-10
always return value from to_python

snopoke:
http://manage.dimagi.com/default.asp?245087

@NoahCarnahan @biyeun


NoahCarnahan (approved):
Looks good to me! Glad you found this, because I bet it would have taken me a long time to locate this...

PR #14478, opened 2017-01-10
ensure exported tables are kept in order

snopoke:
http://manage.dimagi.com/default.asp?244229

Followup to https://github.com/dimagi/commcare-hq/pull/14331

@NoahCarnahan


NoahCarnahan (approved):
Ooof. Can't believe I missed that :( Thanks for the follow up.

PR #14483, opened 2017-01-10
getting json doc instead of wrapped class

calellowitz:
@benrudolph @snopoke CC: @gcapalbo 
http://manage.dimagi.com/default.asp?245333#1271904

looks like cloudcare preview was totally broken. this should fix it


benrudolph (approved):
nice thanks, fwiw this url is not actually anywhere in the UI. this was just for people who know what url to type in

PR #14485, opened 2017-01-10
redirecting loc restricted mobile workers to cloudcare

calellowitz:
@sravfeyn @snopoke @esoergel 
http://manage.dimagi.com/default.asp?245077#1271974

@benrudolph is there a reason formplayer is marked as location safe and old cloudcare isnt. I think this fails for domains on old cloudcare. I am not sure thats a big deal, and its not a regression but it doesnt seem ideal.


esoergel (approved):

benrudolph (approved):
👍 there's no reason except that we aren't supporting old cloudcare anymore. so if something doesn't work i try and get them to move to new cloudcare rather than fix some functionality on old cloudcare (unless it's a regression).

benrudolph:
:kick_scooter: 🚧

PR #14488, opened 2017-01-11
use name as serializable expression

czue:
http://manage.dimagi.com/default.asp?245531

ugh, i almost regret doing this change. this part of the code is very hard to reason with and i can't figure out how to repro the issues locally


czue:
:white_check_mark: @orangejenny / @proteusvacuum / anyone

PR #14489, opened 2017-01-11
full throttle nimbus api

benrudolph:
@snopoke still not even seeing a dent in the perf graphs after bumping to 0.4. sorry for blocking on that originally, i think this is good to go for 100%. the benefits of monitoring!

cc: @emord


snopoke:
oops, premature merge. @benrudolph do you think it's reasonable to mock the API call for this test?

benrudolph:
makes sense to me. we have a test for it in formplayer so no need to test it twice

PR #14491, opened 2017-01-11
Small debugger button

benrudolph:
@biyeun @orangejenny this makes the debugger button smaller and less intrusive for app preview. what do you think? i also think this is way less intrusive when you use one question per screen. there's always loads of whitespace at the bottom.

veolmahtic

cc: @emord


emord:
Would it make sense to have that icon on the bottom right? In your gif it looks like it would be less likely to interfere with question labels on the right than left

benrudolph:
👍

benrudolph:
moved to the right side, but got rid of the hack to make sure the bar wasn't the full with of the screen. i think it looks fine on old cloudcare:

screen shot 2017-01-11 at 6 51 36 pm

this is preview:

screen shot 2017-01-11 at 6 52 13 pm

biyeun:
good move to put it on the bottom right! NICE suggestion @emord :+1:

biyeun (approved):

emord (approved):

PR #14497, opened 2017-01-11
UCLA custom handler

emord:
https://docs.google.com/document/d/1Hc7Cd53AcQJ5o1WybukAHdaUnbEtwDgMbXB0waXBUpc/edit

@gcapalbo buddy: @benrudolph

### DIFF
--- a/custom/ucla/api.py
+++ b/custom/ucla/api.py
@@ -32,4 +32,12 @@ def ucla_message_bank_content(reminder, handler, recipient):
         notify_exception(None, message=message)
         return None
 
+    case_props = recipient.dynamic_case_properties()
+    try:
+        assert 'risk_profile' in case_props
+    except AssertionError:
+        message = "case does not include risk_profile"
+        notify_exception(None, message=message)
+        return None
+
     return "custom message"
--- a/custom/ucla/api.py
+++ b/custom/ucla/api.py
@@ -1,2 +1,18 @@
+from corehq.apps.fixtures.models import FixtureDataType
+from dimagi.utils.logging import notify_exception
+
+
 def ucla_message_bank_content(reminder, handler, recipient):
-    return None
+    message_bank = filter(
+        lambda f: f.tag == 'message_bank',
+        FixtureDataType.by_domain(reminder.domain)
+    )
+
+    if not message_bank:
+        message = "Lookup Table message_bank not found in {}".format(reminder.domain)
+        notify_exception(None, message=message)
+        return None
+
+    message_bank = message_bank[0]
+
+    return "custom message"
--- a/custom/ucla/api.py
+++ b/custom/ucla/api.py
@@ -1,10 +1,19 @@
 from corehq.apps.fixtures.models import FixtureDataType, FixtureDataItem
+from corehq.apps.users.models import WebUser
 from corehq.form_processor.utils import is_commcarecase
-from dimagi.utils.logging import notify_exception
+from corehq.util.soft_assert import soft_assert
 
 
 RISK_PROFILE_FIELD = 'risk_profile'
 REQUIRED_FIXTURE_FIELDS = [RISK_PROFILE_FIELD, 'sequence', 'message']
+
+
+def notify_dimagi_project_admins(domain, message):
+    users = list(WebUser.get_dimagi_emails_by_domain(domain))
+    # if there isn't a dimagi email in the project, notify admins
+    notify_admins = len(users) == 0
+    _assert = soft_assert(users, notify_admins=notify_admins, send_to_ops=False)
+    _assert(False, message)
 
 
 def ucla_message_bank_content(reminder, handler, recipient):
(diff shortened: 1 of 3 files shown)


custom/ucla/tests.py
```diff
@@ -0,0 +1,142 @@
+from collections import namedtuple
+import uuid
+from django.test import TestCase
+
+from corehq.apps.fixtures.models import (
+    FixtureDataType, FixtureTypeField, FixtureDataItem, FieldList, FixtureItemField
+)
+from corehq.form_processor.interfaces.dbaccessors import CaseAccessors
```
dimagimon: F401 'CaseAccessors' imported but unused

custom/ucla/api.py
```diff
@@ -0,0 +1,62 @@
...
+        lambda m: m.fields_without_attributes['sequence'] == current_message_seq_num,
+        custom_messages
+    )
+
+    if len(custom_messages) != 1:
+        if not custom_messages:
+            message = "No message for risk {}, seq {} in domain {}"
+        else:
+            message = "Multiple messages for risk {}, seq {} in domain {}"
+        message = message.format(risk_profile, current_message_seq_num, domain)
```
benrudolph: nit: if risk profile can be unicode this could through an error

benrudolph (approved):
👍 one question about the `notify_exception`. i don't have too much context, but is that a kind of error that we can do something about? from this is looks like it should almost be a message sent back to the caller of the api

custom/ucla/tests.py
```diff
@@ -0,0 +1,141 @@
...
+
+class UCLACustomHandler(TestCase):
+    domain = uuid.uuid4().hex
+    case_type = 'ucla-reminder'
+    fixture_name = 'message_bank'
+
+    def _reminder(self):
+        return Reminder(
+            domain=self.domain,
+            schedule_iteration_num=0,
```
gcapalbo: we should make this be 1 by default (the iteration num starts counting at 1)

custom/ucla/tests.py
```diff
@@ -0,0 +1,141 @@
...
+            current_event_sequence_num=0,
+        )
+
+    def _handler(self, num_events=0):
```
gcapalbo: i think it would be good for all the tests to use num_events=1 for this

custom/ucla/tests.py
```diff
@@ -0,0 +1,141 @@
...
+    def test_no_relevant_message_invalid_risk(self):
+        self._setup_fixture_type()
+        with create_test_case(self.domain, self.case_type, 'test-case', {'risk_profile': 'risk2'}) as case:
+            self.assertIsNone(ucla_message_bank_content(self._reminder(), self._handler(), case))
+
+    @run_with_all_backends
+    def test_no_relevant_message_invalid_seq_num(self):
+        self._setup_fixture_type()
+        with create_test_case(self.domain, self.case_type, 'test-case', {'risk_profile': 'risk1'}) as case:
+            self.assertIsNone(ucla_message_bank_content(self._reminder(), self._handler(num_events=1), case))
```
gcapalbo: we could use a reminder with schedule_iteration_num=2 for this test, keeping num_events=1 on the handler

custom/ucla/api.py
gcapalbo: i think to be safe we could just make all the notify messages be unicode by prepending `u` to them

gcapalbo:
looks great @emord, just a few comments

@benrudolph regarding the notify_exception, yeah maybe that's not the best fit. these errors can happen at reminder runtime, so i was thinking there should be some way to alert us of config issues in the project so we can fix them. maybe a better thing to do would be to send an email to all "dimagi.com" web users in the project rather than notify_exception.

custom/ucla/tests.py
```diff
@@ -0,0 +1,157 @@
...
+    def test_no_relevant_message_invalid_risk(self):
+        self._setup_fixture_type()
+        with create_test_case(self.domain_name, self.case_type, 'test-case', {'risk_profile': 'risk2'}) as case:
+            self.assertRaises(AssertionError, ucla_message_bank_content, self._reminder(), self._handler(), case)
+
+    @run_with_all_backends
+    def test_no_relevant_message_invalid_seq_num(self):
+        self._setup_fixture_type()
+        with create_test_case(self.domain_name, self.case_type, 'test-case', {'risk_profile': 'risk1'}) as case:
+            self.assertRaises(AssertionError, ucla_message_bank_content, self._reminder(iteration_num=2), self._handler(), case)
```
dimagimon: E501 line too long (128 > 115 characters)

emord:
Updated to include all the feedback. I went with soft assert to email all of the dimagi people on the project and if there aren't any, then fall back to notifying admins

gcapalbo (approved):

gcapalbo:
looks great, just waiting on tests

PR #14500, opened 2017-01-11
Make dashboard location-safe

esoergel:
@czue @calellowitz IIRC you two were just doing some work regarding this behavior.
This PR makes the dashboard location-safe, along with the "default" views for the reports, data, and users tiles.

It does preserve the behavior Cal implemented, which is redirecting mobile workers to Cloudcare instead of the dashboard, but only if they're location restricted.  I don't see a reason why we shouldn't just do this for ALL mobile workers, actually, but that felt like too impactful a change to make without more explicit thought.

@NoahCarnahan @sravfeyn

### DIFF
--- a/corehq/apps/dashboard/models.py
+++ b/corehq/apps/dashboard/models.py
@@ -1,5 +1,6 @@
 from corehq.apps.export.views import ExportsPermissionsMixin
-from django.core.urlresolvers import reverse
+from django.core.urlresolvers import reverse, resolve, Resolver404
+from corehq.tabs.uitab import url_is_location_safe
 from corehq.apps.app_manager.dbaccessors import get_brief_apps_in_domain
 from corehq.apps.reports.models import ReportConfig, FormExportSchema, CaseExportSchema
 from dimagi.utils.decorators.memoized import memoized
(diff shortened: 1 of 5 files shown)


corehq/apps/dashboard/models.py
```diff
@@ -35,6 +36,15 @@ def is_visible(self):
...
         """
+        if not self.request.can_access_all_locations:
+            url = self.tile_config.get_url(self.request)
+            try:
+                match = resolve(url)
+            except Resolver404:
+                pass
+            else:
+                if 'domain' in match.kwargs and not url_is_location_safe(url):
+                    return False
```
esoergel: Note that this only affects "domain" views, so non-domain things like the help site and the exchange will appear if they'd normally appear.  "Location safety" only applies to domain things, so I figured it made sense to preserve the existing logic otherwise.

czue:
> It does preserve the behavior Cal implemented, which is redirecting mobile workers to Cloudcare instead of the dashboard, but only if they're location restricted. I don't see a reason why we shouldn't just do this for ALL mobile workers, actually, but that felt like too impactful a change to make without more explicit thought.

I think we should do this for all mobile workers. It has always been the default behavior historically.

czue:
didn't review code, but approve of the concept!

NoahCarnahan (approved):
Looks good to me!

NoahCarnahan:
Not merging so you can make the change for all mobile workers, but feel free to merge if you want to do that later or in a separate PR.

sheelio:
I think for ICDS, we specifically didn't want to re-direct mobile workers to CloudCare if they had access to other parts of the website (this is because we're using mobile workers for users who just view a few reports).   It is easier to create / manage mobile workers in bulk than do that with web users.

PR #14501, opened 2017-01-12
allow specifying case type for autoload case actions

snopoke:
http://manage.dimagi.com/default.asp?244499

Allowing the user to specify the case type has two benefits:
1. we can use it when creating case indexes
2. we can use it to autosuggest case properties from the appropriate case type which makes doing the case management much easier

@calellowitz @proteusvacuum

### DIFF
--- a/corehq/apps/app_manager/static/app_manager/js/case-config-ui-advanced.js
+++ b/corehq/apps/app_manager/static/app_manager/js/case-config-ui-advanced.js
@@ -197,6 +197,10 @@ hqDefine('app_manager/js/case-config-ui-advanced.js', function () {
                         self.applyAccordion('open', 0);
                     }
                 });
+
+                $('.hq-help-template').each(function () {
+                    COMMCAREHQ.transformHelpTemplate($(this), true);
+                });
             });
         };
     };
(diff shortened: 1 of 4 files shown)


corehq/apps/app_manager/static/app_manager/js/case-config-ui-advanced.js
```diff
@@ -696,6 +708,9 @@ hqDefine('app_manager/js/case-config-ui-advanced.js', function () {
                             return 'Lookup table tag required';
                         }
                     }
+                    if (!self.case_type()) {
+                        return 'Expected case type required';
```
proteusvacuum: Will this mean we have to set these expected case types on all case management pages that currently have autoloads currently when we save (even if the change is to a different load action)?

proteusvacuum (approved):
One non-blocking question, but looks great.

corehq/apps/app_manager/static/app_manager/js/case-config-ui-advanced.js
snopoke: This validation doesn't actually prevent the user from saving (strangely) so it shouldn't have any effect in the example you mention.

czue:
this is excellent - thanks!

corehq/apps/app_manager/static/app_manager/js/case-config-ui-advanced.js
proteusvacuum: hah, weird. Thanks!

PR #14504, opened 2017-01-12
Redirect all mobile workers to cloudcare

esoergel:
@NoahCarnahan @czue
from https://github.com/dimagi/commcare-hq/pull/14500

### DIFF
--- a/corehq/apps/dashboard/views.py
+++ b/corehq/apps/dashboard/views.py
@@ -36,17 +36,6 @@ def default_dashboard_url(request, domain):
     if domain in settings.CUSTOM_DASHBOARD_PAGE_URL_NAMES:
         return reverse(settings.CUSTOM_DASHBOARD_PAGE_URL_NAMES[domain], args=[domain])
 
-    # Send mobile users without a role to CloudCare
-    if (
-        couch_user and couch_user.is_commcare_user()
-        and couch_user.get_role(domain) is None
-    ):
-        if toggles.USE_FORMPLAYER_FRONTEND.enabled(domain):
-            formplayer_view = FormplayerMain.urlname
-        else:
-            formplayer_view = "corehq.apps.cloudcare.views.default"
-        return reverse(formplayer_view, args=[domain])
-
     if couch_user and user_has_custom_top_menu(domain, couch_user):
         return reverse('saved_reports', args=[domain])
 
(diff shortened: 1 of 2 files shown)


czue:
sounds like we might need to wait on this: https://github.com/dimagi/commcare-hq/pull/14500#issuecomment-272214709

esoergel:
@sheelio I assume that means the ICDS users have a role assigned then, right?  Do you think it'd be acceptable to redirect all mobile workers without a role to cloudcare?  My understanding is that if they don't have a role, they won't be able to access anything else.

sheelio:
Yup that sounds great if you can do that.

czue:
shoot, i just realized ICDS and TB want opposite behavior (for TB our users are primarily cloudcare users who need edit data permissions to login as someone else, but we want cloudcare to be the primary landing page). I think this means we need to make this a setting/configurable. yuck.

czue:
@esoergel you can still make the change you suggested, but we might need to undo what @calellowitz did for http://manage.dimagi.com/default.asp?245077? Cal did you change the landing page or just fix the dashboard?

esoergel:
I did a little more investigating, there are two functions in play here - `redirect_to_default` and `default_dashboard_url`.  The former seems like the more sensible place for this logic to be, I think I mistakenly put the redirect in `default_dashboard_url` when I first instituted location safety.

The existing behavior is to redirect to cloudcare IFF it's a web user who `can_view_some_reports`, otherwise direct to the default dashboard view.  That seems backwards, I wonder if there was supposed to be a `not` there.  The change I'm proposing reverses that, and switches the check from `can_view_some_reports` to "has role".  That is, direct mobile users with a role to the dashboard, and mobile users without a role to cloudcare.

That seems like the desireable behavior to me, then in the future we can add some configuration option to send some users with roles to cloudcare.

@dimagi/product what do you think?

czue:
> That is, direct mobile users with a role to the dashboard, and mobile users without a role to cloudcare.

This definitely sounds right to me and agree the behavior you observed sounds like a mistake.

@sheelio and I were chatting about this and have two small changes we're going to propose/roll out that are enterprise product aligned and solve the ICDS/TB conflict.

1. add a "default landing page" option to roles. This might currently have options of dashboard, web apps, and reports, and will default to dashboard for web users and web apps for mobile workers.
2. remove the "dashboard" tab entirely from all mobile worker views. can go into more reasoning if there is push back but basically the dashboard is really useful for orienting you into HQ, but the use case for mobile workers using HQ is much more commonly someone who has been trained to do specific stuff and doesn't need that orientation.

czue:
I think this can be closed in favor of https://github.com/dimagi/commcare-hq/pull/14530

PR #14506, opened 2017-01-12
Introduce is_two_way flag on PhoneNumber model

gcapalbo:
(opening for tests)


corehq/apps/reminders/tests/test_send_reminders.py
```diff
@@ -318,7 +318,10 @@ def setUp(self):
         self.handler.save()
 
         self.user = CommCareUser.create(self.domain, 'chw.bob3', 'abc', phone_number='14445551234')
-        self.user.save_verified_number(self.domain, '14445551234', True)
+        entry = user.get_or_create_phone_entry('14445551234')
```
dimagimon: F821 undefined name 'user'

corehq/apps/reminders/tests/test_util.py
```diff
@@ -9,24 +9,30 @@ class ReminderUtilTest(TestCase):
...
+    def test_get_two_way_number_for_recipient(self):
+        self.assertIsNone(get_two_way_number_for_recipient(self.user))
 
         self.user.phone_numbers = ['123', '456', '789']
-        self.user.save_verified_number('test', '456', True)
-        self.assertEqual(get_verified_number_for_recipient(self.user).phone_number, '456')
-
-        self.user.save_verified_number('test', '789', True)
-        self.assertEqual(get_verified_number_for_recipient(self.user).phone_number, '456')
+        entry = user.get_or_create_phone_entry('456')
```
dimagimon: F821 undefined name 'user'

corehq/apps/reminders/tests/test_util.py
```diff
@@ -9,24 +9,30 @@ class ReminderUtilTest(TestCase):
...
+        entry.set_two_way()
+        entry.set_verified()
+        entry.save()
+        self.assertEqual(get_two_way_number_for_recipient(self.user).phone_number, '456')
+
+        entry = user.get_or_create_phone_entry('789')
```
dimagimon: F821 undefined name 'user'

dimagimon:
There are 7 errors:

* corehq/apps/sms/tests/test_phone_numbers.py, line 385 - F841 local variable 'number2' is assigned to but never used
* corehq/apps/sms/tests/test_phone_numbers.py, line 430 - F841 local variable 'number2' is assigned to but never used
* corehq/apps/sms/tests/test_phone_numbers.py, line 757 - F841 local variable 'number3' is assigned to but never used
* corehq/apps/sms/tests/test_phone_numbers.py, line 772 - F841 local variable 'number1' is assigned to but never used
* corehq/apps/sms/tests/test_phone_numbers.py, line 773 - F841 local variable 'number2' is assigned to but never used
* corehq/apps/sms/tests/test_phone_numbers.py, line 774 - F841 local variable 'number3' is assigned to but never used
* corehq/apps/sms/tests/test_phone_numbers.py, line 775 - F841 local variable 'number4' is assigned to but never used

dimagimon:
There are 7 errors:

* corehq/apps/sms/tests/test_phone_numbers.py, line 385 - F841 local variable 'number2' is assigned to but never used
* corehq/apps/sms/tests/test_phone_numbers.py, line 430 - F841 local variable 'number2' is assigned to but never used
* corehq/apps/sms/tests/test_phone_numbers.py, line 757 - F841 local variable 'number3' is assigned to but never used
* corehq/apps/sms/tests/test_phone_numbers.py, line 772 - F841 local variable 'number1' is assigned to but never used
* corehq/apps/sms/tests/test_phone_numbers.py, line 773 - F841 local variable 'number2' is assigned to but never used
* corehq/apps/sms/tests/test_phone_numbers.py, line 774 - F841 local variable 'number3' is assigned to but never used
* corehq/apps/sms/tests/test_phone_numbers.py, line 775 - F841 local variable 'number4' is assigned to but never used

corehq/apps/sms/models.py
```diff
@@ -582,52 +592,61 @@ def by_couch_id(cls, couch_id):
...
 
     @classmethod
-    def _by_suffix(cls, phone_number):
-        # Decided not to cache this method since in order to clear the cache
-        # we'd have to clear using all suffixes of a number (which would involve
-        # up to ~10 cache clear calls on each save). Since this method is used so
-        # infrequently, it's better to not cache vs. clear so many keys on each
-        # save. Once all of our IVR gateways provide reliable caller id info,
-        # we can also remove this method.
+    def get_reserved_number(cls, phone_number):
```
emord: I think you could make this one call with `cls.objects.filter(phone_number__contains='70').filter(Q(is_two_way=True) | Q(verified=False, pending_verification=True)).first()` May help with the upcoming phone number report "already in use" status

dimagimon:
There are 37 errors:

* corehq/apps/sms/tests/test_phone_numbers.py, line 658 - E251 unexpected spaces around keyword / parameter equals
* corehq/apps/sms/tests/test_phone_numbers.py, line 666 - E123 closing bracket does not match indentation of opening bracket's line
* corehq/apps/sms/tests/test_phone_numbers.py, line 668 - E251 unexpected spaces around keyword / parameter equals
* corehq/apps/sms/tests/test_phone_numbers.py, line 676 - E123 closing bracket does not match indentation of opening bracket's line
* corehq/apps/sms/tests/test_phone_numbers.py, line 686 - E122 continuation line missing indentation or outdented
* corehq/apps/sms/tests/test_phone_numbers.py, line 689 - E251 unexpected spaces around keyword / parameter equals
* corehq/apps/sms/tests/test_phone_numbers.py, line 690 - E251 unexpected spaces around keyword / parameter equals
* corehq/apps/sms/tests/test_phone_numbers.py, line 692 - E251 unexpected spaces around keyword / parameter equals
* corehq/apps/sms/tests/test_phone_numbers.py, line 693 - E251 unexpected spaces around keyword / parameter equals
* corehq/apps/sms/tests/test_phone_numbers.py, line 700 - E251 unexpected spaces around keyword / parameter equals
* corehq/apps/sms/tests/test_phone_numbers.py, line 705 - E251 unexpected spaces around keyword / parameter equals
* corehq/apps/sms/tests/test_phone_numbers.py, line 713 - E123 closing bracket does not match indentation of opening bracket's line
* corehq/apps/sms/tests/test_phone_numbers.py, line 715 - E251 unexpected spaces around keyword / parameter equals
* corehq/apps/sms/tests/test_phone_numbers.py, line 718 - E251 unexpected spaces around keyword / parameter equals
* corehq/apps/sms/tests/test_phone_numbers.py, line 723 - E251 unexpected spaces around keyword / parameter equals
* corehq/apps/sms/tests/test_phone_numbers.py, line 725 - E251 unexpected spaces around keyword / parameter equals
* corehq/apps/sms/tests/test_phone_numbers.py, line 730 - E251 unexpected spaces around keyword / parameter equals
* corehq/apps/sms/tests/test_phone_numbers.py, line 733 - E251 unexpected spaces around keyword / parameter equals
* corehq/apps/sms/tests/test_phone_numbers.py, line 741 - E251 unexpected spaces around keyword / parameter equals
* corehq/apps/sms/tests/test_phone_numbers.py, line 746 - E251 unexpected spaces around keyword / parameter equals
* corehq/apps/sms/tests/test_phone_numbers.py, line 756 - E122 continuation line missing indentation or outdented
* corehq/apps/sms/tests/test_phone_numbers.py, line 759 - E251 unexpected spaces around keyword / parameter equals
* corehq/apps/sms/tests/test_phone_numbers.py, line 760 - E251 unexpected spaces around keyword / parameter equals
* corehq/apps/sms/tests/test_phone_numbers.py, line 762 - E251 unexpected spaces around keyword / parameter equals
* corehq/apps/sms/tests/test_phone_numbers.py, line 763 - E251 unexpected spaces around keyword / parameter equals
* corehq/apps/sms/tests/test_phone_numbers.py, line 770 - E251 unexpected spaces around keyword / parameter equals
* corehq/apps/sms/tests/test_phone_numbers.py, line 776 - E251 unexpected spaces around keyword / parameter equals
* corehq/apps/sms/tests/test_phone_numbers.py, line 784 - E251 unexpected spaces around keyword / parameter equals
* corehq/apps/sms/tests/test_phone_numbers.py, line 785 - E251 unexpected spaces around keyword / parameter equals
* corehq/apps/sms/tests/test_phone_numbers.py, line 787 - E251 unexpected spaces around keyword / parameter equals
* corehq/apps/sms/tests/test_phone_numbers.py, line 794 - E251 unexpected spaces around keyword / parameter equals
* corehq/apps/sms/tests/test_phone_numbers.py, line 797 - E251 unexpected spaces around keyword / parameter equals
* corehq/apps/sms/tests/test_phone_numbers.py, line 810 - E251 unexpected spaces around keyword / parameter equals
* corehq/apps/sms/tests/test_phone_numbers.py, line 813 - E251 unexpected spaces around keyword / parameter equals
* corehq/apps/sms/tests/test_phone_numbers.py, line 817 - E251 unexpected spaces around keyword / parameter equals
* corehq/apps/sms/tests/test_phone_numbers.py, line 823 - E251 unexpected spaces around keyword / parameter equals
* corehq/apps/sms/tests/test_phone_numbers.py, line 827 - E251 unexpected spaces around keyword / parameter equals

corehq/apps/sms/models.py
gcapalbo: my thinking here was that both of those calls are cached so it'd be better/faster to lookup to the cache rather than the db

gcapalbo:
@emord this is ready for release now

corehq/apps/sms/mixin.py
```diff
@@ -132,64 +128,42 @@ def validate_number_format(cls, phone_number):
...
+    def _get_or_create_phone_entry(self, phone_number):
+        phone_number = apply_leniency(phone_number)
+        self.validate_number_format(phone_number)
 
-        return  The PhoneNumber
-        raises  InvalidFormatException if the phone number format is invalid
-        raises  PhoneNumberInUseException if the phone number is already in use by another contact
-        """
-        from corehq.apps.sms.models import PhoneNumber
+        entries = self.get_phone_entries()
```
emord: Is the reasoning for not using `get_or_create` is this is cached?

corehq/apps/sms/views.py
```diff
@@ -707,7 +707,7 @@ def get_contact_info(domain):
     case_ids = []
     mobile_worker_ids = []
     data = []
-    for p in PhoneNumber.by_domain(domain):
+    for p in PhoneNumber.by_domain(domain).filter(is_two_way=True):
```
emord: Why was this condition added?

corehq/apps/sms/migrations/0024_add_is_two_way.py
```diff
@@ -0,0 +1,45 @@
...
+
+from corehq.sql_db.operations import HqRunPython
+from django.db import migrations, models
+import datetime
+
+
+def set_pending_verification_numbers(apps, schema_editor):
+    PhoneNumber = apps.get_model('sms', 'PhoneNumber')
+    db_alias = schema_editor.connection.alias
+    PhoneNumber.objects.using(db_alias).filter(verified=False).update(is_two_way=False, pending_verification=True)
```
emord: Should `verified=True` imply `is_two_way=True` and should it be added to this migration?

emord (commented):
Looks good, but want to verify my understanding of this code.

Before this PR a verified number implies is_two_way. How did the code handle an unverified number that should be two way? 

After this PR a number can be flagged is_two_way and pending_verification or verified or neither. 

It's pretty hard for me to grasp how often the DB lookups are called and when it makes sense to cache vs make separate more filtered call. Similar with the `CriticalSection`s

There are also some methods that return objects (`by_owner_id`), and others that return queryset (`by_domain`), but it isn't really clear unless you look at implementation details. I assume there are good reasons for that (caching/locking) but is there a standard for how to tell what to expect?

gcapalbo:
On Thu, Jan 19, 2017 at 12:16 PM, Jonathan Emord <notifications@github.com>
wrote:

> *@emord* commented on this pull request.
>
> Looks good, but want to verify my understanding of this code.
>
> Before this PR a verified number implies is_two_way. How did the code
> handle an unverified number that should be two way?
>
Yeah so the idea here is to get away from using the verified flag to denote
a two-way number since we may be doing away with that in the future. We're
also starting to store store one-way numbers in the db with this pr so that
we can report on them, and also do automatic promotion eventually (once a
two-way number gets freed up we can promote another duplicate one-way
number to two-way, for example).

Previously, one-way numbers weren't stored in the db, so all entries were
either two way if verified, or pending verification if not verified.

> After this PR a number can be flagged is_two_way and pending_verification
> or verified or neither.
>
Functionally, nothing has changed yet. So all we expect to see is:

two-way numbers: is_two_way True, verified True, pending_verification False

pending verification numbers: is_two_way False, verified False,
pending_verification True

one-way numbers: all three flags False

We may be making some changes to the verification process in a future state
which would let us get rid of the verification and pending verification
flags and would simplify this, but this refactor is the first step to take
before getting there.

> It's pretty hard for me to grasp how often the DB lookups are called and
> when it makes sense to cache vs make separate more filtered call.
>
I tried to cache the things we do often - for example looking up a two-way
number to see who owns it (which happens each time we receive an inbound
sms), and looking up an entire contact's phone entries (which happens each
time we update a case or user to sync their phone entries).

> Similar with the CriticalSections
>
 The CriticalSections are used to lock out code that can execute
simultaneously to prevent race conditions. So everywhere the same lock key
is used in a CriticalSection statement, we don't want those blocks of code
executing at the same time otherwise the data could end up in an
inconsistent state.  For example, we have to do this when promoting a
number to two-way so that if two threads try to do it at the same time on
different one-way entries belonging to different contacts but with the same
phone number, only one of them will get marked as such.

> There are also some methods that return objects (by_owner_id), and others
> that return queryset (by_domain), but it isn't really clear unless you
> look at implementation details. I assume there are good reasons for that
> (caching/locking) but is there a standard for how to tell what to expect?
>
I didn't want to cache a queryset so by_owner_id returns a list, but i
think everything else returning a list of things returns querysets which i
tend to like since they allow .count() and .filter()

corehq/apps/sms/migrations/0024_add_is_two_way.py
gcapalbo: yup, it does imply that right now, and that is done [here](https://github.com/dimagi/commcare-hq/pull/14506/files#diff-246f581da2e92dbc370d3d52c2277312R35)

basically the migration sets everything to two-way and then demotes the pending verification ones

corehq/apps/sms/mixin.py
gcapalbo: yeah exactly, we call this in a few places in the framework, and we almost always will get back a cached result so this helps us avoid hitting the db

corehq/apps/sms/views.py
gcapalbo: this was actually wrong before - here we're producing a list of contacts that we can use the chat functionality with, which we should only get from the set of two way numbers

emord:
Thanks for the context. I didn't realize there were 0 one-way numbers in  the database.

PR #14507, opened 2017-01-12
UCR ES transforms

emord:
@esoergel This is what you reported in your PR a month or so ago.

Also caught this coming up in the laboratory experiments for enikshay

buddy @benrudolph

### DIFF
--- a/corehq/apps/userreports/es/data_source.py
+++ b/corehq/apps/userreports/es/data_source.py
@@ -1,3 +1,4 @@
+from collections import OrderedDict
 import copy
 
 from django.utils.decorators import method_decorator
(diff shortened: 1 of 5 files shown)


emord:
Relevant comment/bug report: https://github.com/dimagi/commcare-hq/pull/14183#issuecomment-266596018

corehq/apps/userreports/es/data_source.py
```diff
@@ -123,16 +128,24 @@ def _parse_buckets(buckets, remaining_agg_columns, current_bucket_name, past_buc
...
 
-        ret = []
+        ret = OrderedDict()
         start = start or 0
         end = start + (limit or len(hits))
         relevant_hits = hits[start:end]
 
         for row in relevant_hits:
-            r = {}
+            r_ = {}
```
esoergel: out of curiosity, what's with the trailing underscore?

esoergel (approved):
That parsing code is wicked complicated, but this all looks good to me.

corehq/apps/userreports/es/data_source.py
esoergel: (offlined, it was so `pdb` works)

emord:
> That parsing code is wicked complicated, but this all looks good to me.

I bet sqlalchemy does some pretty complicated stuff too :)

It's likely that a lot of this could be pushed into the reporting code and as an `ESReport` class, like the `SqlReport` class. Then the UCR could inherit from that and it could be used in othe reports

PR #14508, opened 2017-01-12
replacing db calls with dict lookups

calellowitz:
@esoergel cc: @czue @snopoke @sravfeyn 
While investigating http://manage.dimagi.com/default.asp?245323 , we noticed this small optimization that replaces db calls (13k lookups in the case of the icds domain) with dict lookups when pulling the users list for most reports. I am not sure this will be enough of a speedup to make that report load for all users, but should definitely be a step in the right direction.
🎩 to ethan for the initial spot.

### DIFF
--- a/corehq/apps/reports/util.py
+++ b/corehq/apps/reports/util.py
@@ -138,7 +138,7 @@ def get_all_users_by_domain(domain=None, group=None, user_ids=None,
 
         if user_filter[HQUserType.REGISTERED].show:
             # now add all the registered users who never submitted anything
-            users.extend(user for id, user in registered_users_by_id.items()if id not in submitted_user_ids)
+            users.extend(user for id, user in registered_users_by_id.items() if id not in submitted_user_ids)
 
     if simplified:
         return [_report_user_dict(user) for user in users]
--- a/corehq/apps/reports/util.py
+++ b/corehq/apps/reports/util.py
@@ -136,9 +136,8 @@ def get_all_users_by_domain(domain=None, group=None, user_ids=None,
 
         if user_filter[HQUserType.REGISTERED].show:
             # now add all the registered users who never submitted anything
-            for user_id in registered_user_ids:
+            for user_id, user in registered_user_ids.items():
                 if user_id not in submitted_user_ids:
-                    user = CommCareUser.get_by_user_id(user_id)
                     users.append(user)
 
     if simplified:
--- a/corehq/apps/reports/util.py
+++ b/corehq/apps/reports/util.py
@@ -119,7 +119,9 @@ def get_all_users_by_domain(domain=None, group=None, user_ids=None,
         submitted_user_ids = set(get_all_user_ids_submitted(domain))
         registered_users_by_id = dict([(user.user_id, user) for user in CommCareUser.by_domain(domain)])
         if include_inactive:
-            registered_users_by_id.update(dict([(u.user_id, u) for u in CommCareUser.by_domain(domain, is_active=False)]))
+            registered_users_by_id.update(dict(
+                [(u.user_id, u) for u in CommCareUser.by_domain(domain, is_active=False)]
+            ))
         for user_id in submitted_user_ids:
             if user_id in registered_users_by_id and user_filter[HQUserType.REGISTERED].show:
                 user = registered_users_by_id[user_id]
--- a/corehq/apps/reports/util.py
+++ b/corehq/apps/reports/util.py
@@ -117,14 +117,14 @@ def get_all_users_by_domain(domain=None, group=None, user_ids=None,
             user_filter = HQUserType.all()
         users = []
         submitted_user_ids = get_all_user_ids_submitted(domain)
-        registered_user_ids = dict([(user.user_id, user) for user in CommCareUser.by_domain(domain)])
+        registered_users_by_id = dict([(user.user_id, user) for user in CommCareUser.by_domain(domain)])
         if include_inactive:
-            registered_user_ids.update(dict([(u.user_id, u) for u in CommCareUser.by_domain(domain, is_active=False)]))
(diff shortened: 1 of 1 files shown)


esoergel (approved):
Thanks for the fix!

corehq/apps/reports/util.py
```diff
@@ -136,9 +136,8 @@ def _create_temp_user(user_id):
 
         if user_filter[HQUserType.REGISTERED].show:
             # now add all the registered users who never submitted anything
-            for user_id in registered_user_ids:
+            for user_id, user in registered_user_ids.items():
```
orangejenny: Nitpick: `registered_users` would be better than `registered_user_ids`

esoergel: I'd go with `registered_users_by_id` to indicate it's a dict.  @millerdev and I have a date to refactor this whole function, it's one of those things that started small, then grew and grew until it does like 3 or 4 unrelated things, is really unperformant, and is hard to reason about.

czue (approved):
no way, we already had it in the dict? that is amazing.

agree with renaming the variable.

corehq/apps/reports/util.py
snopoke: seems like you could just make this a list comprehension:
```
users.extend(user for user in registered_users_by_id.values() if user.user_id not in submitted_user_ids)
```

snopoke (approved):

corehq/apps/reports/util.py
```diff
@@ -117,14 +117,14 @@ def _create_temp_user(user_id):
             user_filter = HQUserType.all()
         users = []
         submitted_user_ids = get_all_user_ids_submitted(domain)
-        registered_user_ids = dict([(user.user_id, user) for user in CommCareUser.by_domain(domain)])
+        registered_users_by_id = dict([(user.user_id, user) for user in CommCareUser.by_domain(domain)])
         if include_inactive:
-            registered_user_ids.update(dict([(u.user_id, u) for u in CommCareUser.by_domain(domain, is_active=False)]))
+            registered_users_by_id.update(dict([(u.user_id, u) for u in CommCareUser.by_domain(domain, is_active=False)]))
```
dimagimon: E501 line too long (122 > 115 characters)

corehq/apps/reports/util.py
```diff
@@ -116,15 +116,15 @@ def _create_temp_user(user_id):
         if not user_filter:
             user_filter = HQUserType.all()
         users = []
-        submitted_user_ids = get_all_user_ids_submitted(domain)
-        registered_user_ids = dict([(user.user_id, user) for user in CommCareUser.by_domain(domain)])
+        submitted_user_ids = set(get_all_user_ids_submitted(domain))
+        registered_users_by_id = dict([(user.user_id, user) for user in CommCareUser.by_domain(domain)])
         if include_inactive:
-            registered_user_ids.update(dict([(u.user_id, u) for u in CommCareUser.by_domain(domain, is_active=False)]))
+            registered_users_by_id.update(dict([(u.user_id, u) for u in CommCareUser.by_domain(domain, is_active=False)]))
```
dimagimon: E501 line too long (122 > 115 characters)

calellowitz:
updated with name change and list comprehension @orangejenny @czue @snopoke

corehq/apps/reports/util.py
```diff
@@ -116,15 +116,15 @@ def _create_temp_user(user_id):
         if not user_filter:
             user_filter = HQUserType.all()
         users = []
-        submitted_user_ids = get_all_user_ids_submitted(domain)
-        registered_user_ids = dict([(user.user_id, user) for user in CommCareUser.by_domain(domain)])
+        submitted_user_ids = set(get_all_user_ids_submitted(domain))
+        registered_users_by_id = dict([(user.user_id, user) for user in CommCareUser.by_domain(domain)])
```
calellowitz: this is where the dict is initially created that already had the mobile workers

corehq/apps/reports/util.py
```diff
@@ -136,10 +138,7 @@ def _create_temp_user(user_id):
 
         if user_filter[HQUserType.REGISTERED].show:
             # now add all the registered users who never submitted anything
-            for user_id in registered_user_ids:
-                if user_id not in submitted_user_ids:
-                    user = CommCareUser.get_by_user_id(user_id)
-                    users.append(user)
+            users.extend(user for id, user in registered_users_by_id.items()if id not in submitted_user_ids)
```
emord: nit: it looks weird to not have a space before `if`

corehq/apps/reports/util.py
esoergel: :+1: on the `set()`

corehq/apps/reports/util.py
```diff
@@ -116,15 +116,17 @@ def _create_temp_user(user_id):
...
         users = []
-        submitted_user_ids = get_all_user_ids_submitted(domain)
-        registered_user_ids = dict([(user.user_id, user) for user in CommCareUser.by_domain(domain)])
+        submitted_user_ids = set(get_all_user_ids_submitted(domain))
+        registered_users_by_id = dict([(user.user_id, user) for user in CommCareUser.by_domain(domain)])
         if include_inactive:
-            registered_user_ids.update(dict([(u.user_id, u) for u in CommCareUser.by_domain(domain, is_active=False)]))
+            registered_users_by_id.update(dict(
+                [(u.user_id, u) for u in CommCareUser.by_domain(domain, is_active=False)]
+            ))
```
esoergel: nit (fine to punt): 
```python
# this assembles a list, then converts that to a dict:
dict([(u._id, u) for u in users])
# this uses a generator so you never build that list:
dict((u._id, u) for u in users)
# this is my preference overall:
{u._id: u for u in users}
```
My hunch is that the performance differences are marginal compared to the db calls here, but it's still good to think about.

PR #14511, opened 2017-01-13
update clone domain to copy custom settings

czue:
@proteusvacuum this is why they didn't carry over.

cc @snopoke @orangejenny buddy

### DIFF
--- a/corehq/apps/hqadmin/management/commands/clone_domain.py
+++ b/corehq/apps/hqadmin/management/commands/clone_domain.py
@@ -3,6 +3,7 @@ from optparse import make_option
 from django.core.management.base import BaseCommand, CommandError
 
 from corehq.apps.app_manager.models import Application
+from corehq.apps.locations.models import LocationFixtureConfiguration
 from corehq.apps.userreports.dbaccessors import get_report_configs_for_domain, get_datasources_for_domain
 from corehq.blobs.mixin import BlobMixin
 from corehq.apps.userreports.models import (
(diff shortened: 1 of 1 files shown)


corehq/apps/hqadmin/management/commands/clone_domain.py
```diff
@@ -317,15 +325,19 @@ def save_couch_copy(self, doc, new_domain=None):
         self.log_copy(doc.doc_type, old_id, new_id)
         return old_id, new_id
 
-    def save_sql_copy(self, model, new_domain):
-        old_id = model.id
+    def save_sql_copy(self, model, new_domain, domain_is_pk=False):
+        old_id = model.domain if domain_is_pk else model.id
```
snopoke: you could just change this to `model.pk` which will work for all models (except ones with composite primary keys)

corehq/apps/hqadmin/management/commands/clone_domain.py
```diff
@@ -317,15 +325,19 @@ def save_couch_copy(self, doc, new_domain=None):
...
+
         if self.no_commmit:
-            model.id = 'new-{}'.format(old_id)
+            if not domain_is_pk:
+                model.id = 'new-{}'.format(old_id)
         else:
             model.save()
-        new_id = model.id
+
+        new_id = model.domain if domain_is_pk else model.id
```
snopoke: same as above

corehq/apps/hqadmin/management/commands/clone_domain.py
czue: good call, updated

snopoke (approved):

proteusvacuum:
cool! another thing I noticed that didn't carry over are the Repeaters. Not sure if there is a list of things that still need to get added to this.

czue:
Yeah we did it opportunistically for icds so not everything is supported.
Sorry about that. Saw the ticket so can handle through there

PR #14515, opened 2017-01-13
Better UX for processing deleted apps

benrudolph:
@NoahCarnahan @czue Not fully done, but mostly there. Want to add some automated tests and add one convenience property to the export instance. (it does have the progress bar i just wasn't quick enough to catch it in a screen shot)

screen shot 2017-01-13 at 12 44 23 pm
screen shot 2017-01-13 at 12 44 16 pm

- [x] Automated tests
- [x] Add schema_id to export instance
- [x] Change Settings back to Privacy Settings

cc: @emord

### DIFF
--- a/corehq/apps/app_manager/dbaccessors.py
+++ b/corehq/apps/app_manager/dbaccessors.py
@@ -249,7 +249,7 @@ def get_built_app_ids_with_submissions_for_app_ids_and_versions(domain, app_ids_
         results.extend(
             get_built_app_ids_with_submissions_for_app_id(domain, app_id, app_ids_and_versions.get(app_id))
         )
-    return [result['id'] for result in results]
+    return results
 
 
 def get_latest_app_ids_and_versions(domain, app_id=None):
--- a/corehq/apps/app_manager/dbaccessors.py
+++ b/corehq/apps/app_manager/dbaccessors.py
@@ -235,6 +235,23 @@ def get_built_app_ids_with_submissions_for_app_id(domain, app_id, version=None):
     return [result['id'] for result in results]
 
 
+def get_built_app_ids_with_submissions_for_app_ids_and_versions(domain, app_ids_and_versions=None):
+    """
+    Returns all the built app_ids for a domain that has submissions.
+    If version is specified returns all apps after that version.
+    :domain:
+    :app_ids_and_versions: A dictionary mapping an app_id to build version
+    """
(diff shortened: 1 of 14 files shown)


czue:
nice, UI looks great

NoahCarnahan (approved):
Looks solid

benrudolph:
Ok feeling better about this. Still behind a flag but some changes affect current functionality. flagging some of the important bits

corehq/apps/app_manager/dbaccessors.py
```diff
@@ -235,6 +235,23 @@ def get_built_app_ids_with_submissions_for_app_id(domain, app_id, version=None):
     return [result['id'] for result in results]
 
 
+def get_built_app_ids_with_submissions_for_app_ids_and_versions(domain, app_ids_and_versions=None):
```
benrudolph: I had to modify the way we get built app ids for cases. Before it would return some apps that didn't have submissions and wouldn't be processed. This now ensures we only fetch apps with submissions. This is important for the new Show Deleted because number of processed apps will always be greater than 0 in some cases since we don't process ones without submissions. And thus the modal will always pop up.

Another thing about this dbaccessor is that it makes view calls in a loop. Rather than write another view for just this function, I utilized the already established view. It goes through each app id in the domain and gets all build ids for that app. I don't expect this loop to ever be extremely long, but if you have lots of apps in your domain it can make a lot of view calls.

NoahCarnahan (approved):
Looks good, just saw one little thing. Also, looks like test failure is related.

corehq/apps/export/templates/export/dialogs/process_deleted_questions.html
```diff
@@ -0,0 +1,45 @@
...
+        <div class="progress " data-bind="visible: showBuildSchemaProgressBar">
+          <div class="progress-bar" role="progressbar" data-bind="
+              style: {
+                  width: buildSchemaProgress() + '%'
+              }
+          "></div>
+        </div>
+      </div>
+      <div class="modal-footer">
+        <button type="button" class="btn btn-default" data-dismiss="modal">Close</button>
```
benrudolph: https://github.com/dimagi/commcare-hq/pull/14515/commits/9afc990b52f63f921f3c9da1ac1ee261e9aadd3c

benrudolph:
✅

PR #14516, opened 2017-01-13
If Formplayer returns an error on menu navigation, remove that step

wpride:
Fixes http://manage.dimagi.com/default.asp?245505

### DIFF
--- a/corehq/apps/cloudcare/static/cloudcare/js/formplayer/app.js
+++ b/corehq/apps/cloudcare/static/cloudcare/js/formplayer/app.js
@@ -100,14 +100,11 @@ $(document).on("ajaxStart", function () {
     tfLoadingComplete();
 });
 
-FormplayerFrontend.on('showError', function (errorMessage, isHTML, removeStep) {
+FormplayerFrontend.on('showError', function (errorMessage, isHTML) {
     if (isHTML) {
         showHTMLError(errorMessage, $("#cloudcare-notifications"));
     } else {
         showError(errorMessage, $("#cloudcare-notifications"));
-    }
-    if (removeStep) {
-        FormplayerFrontend.trigger('navigation:back');
     }
 });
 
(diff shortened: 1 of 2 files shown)


corehq/apps/cloudcare/static/cloudcare/js/formplayer/app.js
```diff
@@ -100,12 +100,15 @@ $(document).on("ajaxStart", function () {
...
 });
 
-FormplayerFrontend.on('showError', function (errorMessage, isHTML) {
+FormplayerFrontend.on('showError', function (errorMessage, isHTML, removeStep) {
     if (isHTML) {
         showHTMLError(errorMessage, $("#cloudcare-notifications"));
     } else {
         showError(errorMessage, $("#cloudcare-notifications"));
     }
+    if (removeStep) {
```
benrudolph: it feels weird to have the `showError` trigger a back movement. how do you feel about just having the normal `showError` and then afterward call `navigation:back`

corehq/apps/cloudcare/static/cloudcare/js/formplayer/menus/api.js
```diff
@@ -26,7 +26,8 @@ FormplayerFrontend.module("Menus", function (Menus, FormplayerFrontend, Backbone
                         FormplayerFrontend.trigger(
                             'showError',
                             response.exception,
-                            response.type === 'html'
+                            response.type === 'html',
+                            true
                         );
```
benrudolph: the nav back would just be here

wpride: I shied away from this because it felt like breaking some abstraction barrier to change state within the API class

benrudolph: aren't we still breaking the abstraction it's just two trigger calls instead of one?

benrudolph: i do see what you mean though, i'm on the fence as to whether this is better

wpride:
@benrudolph yeah I changed my mind yours is better

benrudolph (approved):

PR #14517, opened 2017-01-13
POC hide reports sidebar on mobile

snopoke:
http://manage.dimagi.com/default.asp?245529

### Menu hidden
reports_menu_hidden

### Menu shown
reports_menu_visible

### DIFF
--- a/corehq/apps/reports/templates/reports/standard/base_template.html
+++ b/corehq/apps/reports/templates/reports/standard/base_template.html
@@ -119,15 +119,15 @@
 {% endblock %}
 
 {% block page_navigation %}
-    <button type="button" class="pull-right btn btn-default visible-xs" data-toggle="offcanvas" aria-expanded="false" aria-controls="navbar">
-        <i class="fa fa-navicon"></i>
+    <button type="button" class="pull-right btn btn-invisible visible-xs" data-toggle="offcanvas" aria-expanded="false" aria-controls="navbar">
+        <i class="fa fa-ellipsis-v"></i>
     </button>
 {{ block.super }}
 {% endblock %}
 
 {% block pre_page_content %}{{ block.super }}
-    <button id="sidebar-menu" type="button" class="pull-left btn btn-default visible-xs" data-toggle="offcanvas" aria-expanded="false" aria-controls="navbar">
-        <i class="fa fa-navicon"></i>
+    <button id="sidebar-menu" type="button" class="pull-left btn btn-invisible visible-xs" data-toggle="offcanvas" aria-expanded="false" aria-controls="navbar">
+        <i class="fa fa-ellipsis-v"></i>
     </button>
 {% endblock %}
 
--- a/corehq/apps/reports/templates/reports/standard/base_template.html
+++ b/corehq/apps/reports/templates/reports/standard/base_template.html
@@ -101,7 +101,7 @@
 {% block title %}{{ report.title|default:"Project Reports"|safe }}{% endblock %}
 
 {% block page_breadcrumbs %}
-    <a href="#" class="btn btn-link btn-invisible visible-xs btn-report-menu" data-toggle="offcanvas"><i class="fa fa-bars"></i> Reports Menu</a>
+    <a href="#" class="btn btn-link btn-invisible visible-xs btn-report-menu" data-toggle="offcanvas"><i class="fa fa-bars"></i> {% trans "Reports Menu" %}</a>
     <ol id="hq-breadcrumbs" class="breadcrumb breadcrumb-hq-section">
         <li>
             <a href="{{ report.default_url }}"><strong>{% trans report.section_name|default:"Reports" %}</strong></a>
--- a/corehq/apps/reports/templates/reports/standard/base_template.html
+++ b/corehq/apps/reports/templates/reports/standard/base_template.html
@@ -3,6 +3,12 @@
 {% load hq_shared_tags %}
 {% load i18n %}
 
+{% block page_sidebar_classes %}sidebar-offcanvas col-xs-4 col-sm-3 col-lg-2{% endblock %}
+
+{% block page_row_classes %}row-offcanvas row-offcanvas-left{% endblock %}
+
+{% block page_content_classes %}col-xs-12 col-sm-9 col-lg-10{% endblock %}
+
 {% block js %}{{ block.super }}
     {% block reports-js %}
     {% compress js %}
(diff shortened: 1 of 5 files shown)


czue (approved):
nice this looks great! thanks for taking care of it!

will let @biyeun review and merge but seems great to me.

biyeun (changes requested):
Can we have the button in line with the header text for the filters? Alternatively, can we remove the button styling on this and just have the icon? The whitespacing is not great here. If there's any confusion I can take a pass by Monday with the styling.

snopoke:
I took a quick crack at it but didn't get far. Removing the button styling doesn't look great. If you have time on Monday that would be great otherwise I'll fiddle with it next week.

czue:
@biyeun do you think the style change needs to block merge? just wondering because i think this is already a significant improvement and would be great to demo this workflow for some TB partners early next week

czue:
@biyeun can you please weigh in on my last question?

biyeun:
@czue, sorry have been partially offline yesterday and over the weekend and didn't realize this was so pressing. 

@snopoke I think the best thing you can do for the design here is to remove the border and the rounded edges from this button and make the background transparent. Give it a separate style like `btn-corner`. 

@orangejenny not sure if you had any additional thoughts here. 

One of the biggest issues with this is that you have two borders touching each other creating very weird whitespace artifacts. It should be trivial to make this change.

orangejenny:
@biyeun I defer to you, but yeah, the borders touching is the thing that looks off to me. Removing the button border when the left-hand menu is shown, and moving the button into the "Report Filters" legend when the menu isn't being shown (either as a button or just a plain icon), would fix that.

orangejenny:
Poked at this in the debugger locally. Just getting rid of the background and border on the icon (new `invisible-btn` class?) makes it a bit better. @biyeun do you think that's all right?

screen shot 2017-01-17 at 11 32 43 am
screen shot 2017-01-17 at 11 32 53 am

biyeun:
@czue To be honest, this particular widget **is going to need a lot more design thought than just a casual glance at this PR**---and this is on me to do this thinking. I'm bogged down getting some stuff thru before Amelia heads out and I go on vacation, but I really want to make it clear that **this widget still needs work**. Since there is pressure to merge this in right away before putting in more design thought, I really want to caution against this widget suddenly being used everywhere. I'm afraid this is going to happen simply due to merging this in. 

It seems strange that this didn't go through an initial design brainstorming / review for the interaction and visual design and was just implemented through the interrupt stream and none of the design people were notified. Any new widget like this that isn't a one-off and client specific should be going through the design team, not slipped in through interrupt. Seems like we could have thought of something better than a drawer slider that would make sense for all the pages' navs not just report builder. @orangejenny not sure if you had more thoughts here.

----

Another thought for @snopoke for this particular PR:
The bar icon doesn't make sense because we use it for additional menus / drop down menus elsewhere on HQ. I would use [this icon](http://fontawesome.io/icon/ellipsis-v/) instead because it isn't really associated with anything else on HQ.

czue:
hey @biyeun all of that sounds right. understand this is not the ideal process you wanted to go through, and also hope you can also understand that we need to be moving quickly on these things.

given that we don't have time for a design review right now since this is not a product priority, i think the options are merge-as-is, or do extra work to complicate the implementation to put behind a flag so that only TB sees this. if we don't think this change is harmful (which i haven't heard anyone say) i would advocate for letting it through now. 

> I really want to caution against this widget suddenly being used everywhere

this seems totally valid.

> I'm afraid this is going to happen simply due to merging this in.

however i think we need to come up with a better plan than this to enforce it.

snopoke:
Thanks for the feedback. I completely agree that a general solution to this should go through design review etc. I consider this a stop gap until that happens. I also agree that this should not be applied to any other part of HQ (in fact I'd argue that it isn't useful in most of HQ at all).

I did a quick analytics analysis and found that over the 60 days:

* Mobile + tablet views = 2.5% of total page views
* Mobile + tablet views of reports = 0.38 % of total page views

This indicates to me that the large majority of our users won't even see this change.

I'll update the styling as suggested tomorrow.

biyeun:
@orangejenny the change is still not ideal, but it's better than without. I still advocate for the icon change

snopoke:
Updated:

reports_sidebar_hidden
reports_sidebar_visible

orangejenny:
Code looks fine, thanks for updating both the border/color and the icon. Commented in the ticket on release plan.

biyeun:
Responded in the original ticket FYI @orangejenny @czue @snopoke

snopoke:
@biyeun @czue @orangejenny updated per suggestions in the ticket:
* menu button now in breadcrumbsm
* menu is full width when visible

Other points:
* report content is still visible below menu
* not transition for showing menu (I tried a few thing but didn't get anything working smoothly)

czue:
@biyeun @orangejenny can you take a look at this and see if it is ok to get in now?

corehq/apps/reports/templates/reports/standard/base_template.html
```diff
@@ -18,6 +24,13 @@
 {% endblock %}
 
 {% block js-inline %}{{ block.super }}
+    <script>
+        $(document).ready(function () {
```
orangejenny: Should be `$(function() {` for jQuery 3. I think this style will still work, but it's deprecated: https://jquery.com/upgrade-guide/3.0/#deprecated-document-ready-handlers-other-than-jquery-function

corehq/apps/reports/templates/reports/standard/base_template.html
```diff
@@ -67,12 +80,28 @@
...
+    {% endcompress %}
+  {% endif %}
+
     {% block reports-css %}{% endblock %}
 {% endblock %}
 
 {% block title %}{{ report.title|default:"Project Reports"|safe }}{% endblock %}
 
 {% block page_breadcrumbs %}
+    <a href="#" class="btn btn-link btn-invisible visible-xs btn-report-menu" data-toggle="offcanvas"><i class="fa fa-bars"></i> Reports Menu</a>
```
orangejenny: `trans`

orangejenny:
Looks reasonable to me, just very minor comments, will wait on merging in case you want to respond. Transition would be cool but agree it isn't necessary, and it can be added later without changing the basic interaction/design.

Thank you for being patient with the back & forth on this.

snopoke:
Updated. Thanks for the great feedback!

orangejenny (approved):

PR #14518, opened 2017-01-13
Replace dimagi email addresses with example addresses in localsettings example

NoahCarnahan:
@esoergel

### DIFF
--- a/localsettings.example.py
+++ b/localsettings.example.py
@@ -87,7 +87,7 @@ ENABLE_PRELOGIN_SITE = False
 
 ####### # Email setup ########
 # email settings: these ones are the custom hq ones
-EMAIL_LOGIN = "notifications@yourdomain.com"
+EMAIL_LOGIN = "notifications@example.com"
 EMAIL_PASSWORD = "******"
 EMAIL_SMTP_HOST = "smtp.gmail.com"
 EMAIL_SMTP_PORT = 587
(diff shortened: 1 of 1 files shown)


localsettings.example.py
```diff
@@ -87,22 +87,22 @@
...
-EXCHANGE_NOTIFICATION_RECIPIENTS = ['commcarehq-dev+exchange@dimagi.com']
+ADMINS = (('HQ Dev Team', 'commcarehq-dev+www-notifications@yourdomain.com'),)
+BUG_REPORT_RECIPIENTS = ['commcarehq-support@yourdomain.com']
+NEW_DOMAIN_RECIPIENTS = ['commcarehq-dev+newdomain@yourdomain.com']
+EXCHANGE_NOTIFICATION_RECIPIENTS = ['commcarehq-dev+exchange@yourdomain.com']
 
-SERVER_EMAIL = 'commcarehq-noreply@dimagi.com' #the physical server emailing - differentiate if needed
-DEFAULT_FROM_EMAIL = 'commcarehq-noreply@dimagi.com'
-SUPPORT_EMAIL = "commcarehq-support@dimagi.com"
+SERVER_EMAIL = 'commcarehq-noreply@yourdomain.com' #the physical server emailing - differentiate if needed
```
dimagimon: E261 at least two spaces before inline comment
E262 inline comment should start with '# '

czue (approved):

esoergel:
Note that whatever we put here will in all likelihood receive emails from some instance or another eventually.  If we put `None` or something (or comment out the variables), will the email backend fail in a way that the server admins would notice?  Alternatively, `example.com` is expressly meant for this [purpose](http://www.iana.org/domains/reserved)

NoahCarnahan:
sweet, didn't know about example.com. That sounds like the right approach.

NoahCarnahan:
Self merging since this has been approved

PR #14519, opened 2017-01-13
initial_page_data template tag

orangejenny:
I'd like to get much more of our JavaScript into actual JavaScript files, out of html templates. One of the blockers is making data accessible to js files. This PR adds a template tag that stores given data as json in hidden divs, and a library to access that data from js files. Not ideal that it's a global namespace, but I don't think it's worse than what we're currently doing, I'd guess that most of our current inline js isn't scoped in any way.

@dimagi/js-team

### DIFF
--- a/corehq/apps/fixtures/static/fixtures/js/view-table.js
+++ b/corehq/apps/fixtures/static/fixtures/js/view-table.js
@@ -1,0 +1,17 @@
+(function () {
+    var initialPageData = $.parseJSON($("#initial-page-json").text());
+    if (initialPageData.renderReportTables) {
+        var reportTables = new HQReportDataTables(initialPageData.reportTablesOptions);
+        if (typeof standardHQReport !== 'undefined') {
+            standardHQReport.handleTabularReportCookies(reportTables);
+        }
+        reportTables.render();
+    }
+
+    $(function() {
+        $('.header-popover').popover({
+            trigger: 'hover',
+            placement: 'bottom'
+        });
+    });
+})();
--- a/corehq/apps/fixtures/static/fixtures/js/view-table.js
+++ b/corehq/apps/fixtures/static/fixtures/js/view-table.js
@@ -1,7 +1,24 @@
 (function () {
-    var initialPageData = $.parseJSON($("#initial-page-json").text());
+    var initialPageData = {};
+    _.each($(".initial-page-data"), function(div) {
+        var data = $(div).data();
+        initialPageData[data.name] = data.value;
+    });
     if (initialPageData.renderReportTables) {
-        var reportTables = new HQReportDataTables(initialPageData.reportTablesOptions);
+        var reportTables = new HQReportDataTables(_.pick(initialPageData, [
+            'slug',
+            'defaultRows',
+            'startAtRowNum',
+            'showAllRowsOption',
+            'autoWidth',
+            'aoColumns',
+            'customSort',
+            'ajaxSource',
+            'ajaxParams',
+            'fixColumns',
+            'fixColsNumLeft',
+            'fixColsWidth',
+        ]));
         if (typeof standardHQReport !== 'undefined') {
             standardHQReport.handleTabularReportCookies(reportTables);
         }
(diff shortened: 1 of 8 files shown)


corehq/apps/fixtures/static/fixtures/js/view-table.js
```diff
@@ -0,0 +1,37 @@
...
+            'fixColumns',
+            'fixColsNumLeft',
+            'fixColsWidth',
+        ], function(name) {
+            var value = data(name);
+            if (value !== undefined) {
+                options[name] = value;
+            }
+        });
+        var reportTables = new HQReportDataTables(options);
```
dimagimon: 'HQReportDataTables' is not defined. (no-undef)

corehq/apps/fixtures/static/fixtures/js/view-table.js
```diff
@@ -0,0 +1,37 @@
...
+        if (typeof standardHQReport !== 'undefined') {
+            standardHQReport.handleTabularReportCookies(reportTables);
```
dimagimon: 'standardHQReport' is not defined. (no-undef)

corehq/apps/fixtures/static/fixtures/js/view-table.js
```diff
@@ -0,0 +1,37 @@
...
+        }
+        reportTables.render();
+    }
+
+    $(function() {
+        $('.header-popover').popover({
+            trigger: 'hover',
+            placement: 'bottom'
```
dimagimon: Missing trailing comma. (comma-dangle)

millerdev (approved):

biyeun:
:+1:

czue:
i'm not on @dimagi/js-team so apologies if this has all been discussed, but i'd be interested to hear a bit more about the rationale behind this approach - as opposed to simply doing variable declarations directly in the template and then passing them along to the constructors that live in js files. is the goal to have actually 0 inline JS, and if so what are some of the reasons behind that goal?

some downsides i could imagine from this approach:

- more magic/complexity making things harder to reason about and understand
- larger pages
- possibly less performant

czue:
(did this already go through architecture review on @dimagi/js-team ? if not i'm tempted to call for one before moving forwards)

orangejenny:
The `initial_page_data` template tag + `initial_page_data.js` library is the same pattern as `register_url`/`urllib.js`, which has been discussed by the js team, yes.

Can you expand on your concerns? The operations & markup added by the library & template tag are pretty trivial. I'm also not following how this approach is magical - accessing the js vars requires explicitly importing `initial_page_data.js`, nothing invisible is happening.

czue:
regarding the magic I am mostly thinking that reading:

```javascript
var myVar = {{ my_var|JSON }};
doSomeStuff(myVar);
```
is quite a bit easier to reason with than

```
{% initial_page_data 'myVar' my_var %}
// in some other file
var myVar = data('ajaxSource');
```
agree we can train ourselves on it, but I think it makes grepping and finding the source of where those `initial_page_data` calls are being made quite a bit more complex, and introduces a non-trivial amount of cross-file dependencies that are larger than just the urls.

still curious about my original question about our goals and the rationale behind them.

orangejenny:
We don't have a decision on whether the goal is to eradicate inline js or just drastically reduce it. The answer to that will likely end up depending on where we go with a module system and what the setup for a single page looks like.

I'm not a fan of the 
```javascript
var myVar = {{ my_var|JSON }};
doSomeStuff(myVar);
```
style because it means that js execution is happening in both django templates and in js files (given the number of places where js files execute code in a domready handler). Using the template/HTML as a static data store, and then having all execution in js files, reduces the number of places you need to look and reason about when trying to figure out where and when some behavior is getting fired.

czue:
cool... without getting too bogged down in the details i'm still feeling tempted to do a review unless you feel that @dimagi/js-team has already specifically signed off on this in which case am fine moving forwards as-is. you certainly are more up to speed on js best practices than I am, though am not personally conviinced yet.

orangejenny:
I'll send an email to see if anyone has additional thoughts.

orangejenny:
Tweaked a little bit based on arch review: slightly reduced the amount of markup used, and moved the generation of data tables options out of the template, into python. I think this is good to go.

millerdev (approved):

czue (approved):
nice thanks for bearing with me. i like this a lot.

PR #14522, opened 2017-01-13
UCR ES  total row

emord:
Enables total rows in ES

buddy @benrudolph

### DIFF
--- a/corehq/apps/userreports/tests/test_report_aggregation.py
+++ b/corehq/apps/userreports/tests/test_report_aggregation.py
@@ -2,7 +2,7 @@ from django.conf import settings
 from django.http import HttpRequest
 from django.test import TestCase
 
-from corehq.apps.userreports.const import UCR_BACKENDS, UCR_SQL_BACKEND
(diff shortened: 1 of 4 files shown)


corehq/apps/userreports/tests/test_report_aggregation.py
```diff
@@ -533,7 +583,7 @@ def test_total_row_with_expanded_column(self):
                     ],
                     [3, 1, 0, 3],
                     [6, 0, 2, 2],
-                    [9, 1, 2, 5],
```
emord: Whenever you have a `min` column SQL reports will just sum up all of the mins, which seems pretty useless to me https://github.com/dimagi/sql-agg/blob/c1b70052fcbcfda219ac5532fa0ffcb87d84594c/sqlagg/base.py#L86-L88

Noted it as a bug, but it's been around for a while so doubt it's high priority. ES handles this properly

snopoke: I guess it depends on what you expect from a totals column. Not obvious to me that there is a single correct answer for a 'min' column.

PR #14525, opened 2017-01-13
Je/ucr es percent col

emord:
based off of https://github.com/dimagi/commcare-hq/pull/14522. https://github.com/dimagi/commcare-hq/commit/c6efa17b436f1d2ccd3d350c4a2103272e221cef is the only commit of interest

@benrudolph

### DIFF
--- a/corehq/apps/userreports/es/data_source.py
+++ b/corehq/apps/userreports/es/data_source.py
@@ -12,6 +12,7 @@ from corehq.apps.es.es_query import HQESQuery
 from corehq.apps.reports.api import ReportDataSource
 from corehq.apps.reports.sqlreport import DataFormatter, DictDataFormat
 from corehq.apps.userreports.decorators import catch_and_raise_exceptions
+from corehq.apps.userreports.es.columns import safe_es_column
 from corehq.apps.userreports.exceptions import UserReportsError
 from corehq.apps.userreports.mixins import ConfigurableReportDataSourceMixin
 from corehq.apps.userreports.reports.sorting import ASCENDING, DESCENDING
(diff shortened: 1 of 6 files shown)


emord:
@czue any interest in expanding the test outside of enikshay's UCRs? After this, the only things that I expect don't work are date based columns and advanced sorting and the `LocationColumn`

czue:
> @czue any interest in expanding the test outside of enikshay's UCRs? After this, the only things that I expect don't work are date based columns and advanced sorting and the LocationColumn

yeah definitely. let me know if you want some help making that happen. imagine you could probably ping the ucr skype or email lists for volunteers

corehq/apps/userreports/es/data_source.py
```diff
@@ -219,29 +220,33 @@ def _get_total_aggregated_results(self):
 
     @method_decorator(catch_and_raise_exceptions)
     def get_total_row(self):
-        total_results = self._get_total_aggregated_results()
+        def _clean_total_row(aggregations, aggregation_name):
+            agg = getattr(aggregations, aggregation_name)
+            if hasattr(agg, 'value'):
+                return agg.value
+            elif hasattr(agg, 'doc_count'):
+                return agg.doc_count
```
benrudolph: just curious what is the separation for the two attributes? does ES return different attr names based on the agg?

benrudolph (approved):
nice

corehq/apps/userreports/es/data_source.py
emord: Aggregations like sum or min use 'value' where as filtered aggregations (used in expanded columns) use doc_count. Ideally, this should use a method on the column's config to interpret the results, but that'll take a small refactor. Will add it to the list of things to clean up

PR #14527, opened 2017-01-14
always hide the dashboard tab for mobile workers, regardless of role

czue:
second half of https://github.com/dimagi/commcare-hq/pull/14504#issuecomment-272603395

@esoergel @sheelio cc @orangejenny


snopoke:
is this still relevant in light of https://github.com/dimagi/commcare-hq/pull/14530

czue:
yeah this affects the tab in the top bar so is still needed

PR #14530, opened 2017-01-15
support defining a default landing page per role

czue:
@esoergel @sheelio this implements the first half of https://github.com/dimagi/commcare-hq/pull/14504#issuecomment-272603395

"add a "default landing page" option to roles. This might currently have options of dashboard, web apps, and reports, and will default to dashboard for web users and web apps for mobile workers."

can be read by commit. ui looks like this: 

image

currently options are "use default", "dashboard", "webapps" and "reports".

cc @orangejenny

### DIFF
--- a/corehq/apps/hqwebapp/views.py
+++ b/corehq/apps/hqwebapp/views.py
@@ -37,6 +37,7 @@ from couchdbkit import ResourceNotFound
 from two_factor.views import LoginView
 from two_factor.forms import AuthenticationTokenForm, BackupTokenForm
 from corehq.apps.domain.dbaccessors import get_doc_count_in_domain_by_class
+from corehq.apps.users.landing_pages import get_redirect_url
 
 from corehq.form_processor.utils.general import should_use_sql_backend
 from dimagi.utils.couch.cache.cache_core import get_redis_default_cache
@@ -151,12 +152,11 @@ def redirect_to_default(req, domain=None):
                 couch_user = req.couch_user
                 from corehq.apps.users.models import DomainMembershipError
                 try:
-                    if (couch_user.is_commcare_user() and
-                            couch_user.can_view_some_reports(domain)):
-                        if toggles.USE_FORMPLAYER_FRONTEND.enabled(domain):
-                            url = reverse(FormplayerMain.urlname, args=[domain])
-                        else:
-                            url = reverse("cloudcare_main", args=[domain, ""])
+                    role = couch_user.get_role(domain)
+                    if role and role.default_landing_page:
+                        url = get_redirect_url(role.default_landing_page, domain)
+                    elif couch_user.is_commcare_user():
+                        url = reverse("cloudcare_default", args=[domain])
                     else:
                         from corehq.apps.dashboard.views import dashboard_default
                         return dashboard_default(req, domain)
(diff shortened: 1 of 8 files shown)


benrudolph:
this is pretty neat

corehq/apps/hqwebapp/views.py
```diff
@@ -146,22 +145,29 @@ def redirect_to_default(req, domain=None):
...
+                    else:
+                        # web users without roles are redirected to the dashboard default
+                        # view since some domains allow web users to request access if they
+                        # don't have it
+                        return dashboard_default(req, domain)
+                else:
+                    if role and role.default_landing_page:
+                        url = get_redirect_url(role.default_landing_page, domain)
+                    elif couch_user.is_commcare_user():
+                        url = reverse("cloudcare_default", args=[domain])
```
snopoke: should this still have the toggle logic to redirect to formplayer if it's enabled?

corehq/apps/hqwebapp/views.py
```diff
@@ -146,22 +145,29 @@ def redirect_to_default(req, domain=None):
...
+                    if couch_user.is_commcare_user():
+                        raise Http404()
+                    else:
+                        # web users without roles are redirected to the dashboard default
+                        # view since some domains allow web users to request access if they
+                        # don't have it
+                        return dashboard_default(req, domain)
+                else:
+                    if role and role.default_landing_page:
+                        url = get_redirect_url(role.default_landing_page, domain)
```
snopoke: Could catch the potential (though unlikely) `ValueError`.

snopoke:
Nice, this looks good. Will @jjwack need to make any changes once this gets merged to keep the workflow for "Succeed"? (http://manage.dimagi.com/default.asp?245517)

corehq/apps/hqwebapp/views.py
czue: hmm, yeah do you think that's better than letting it fail hard since something has gotten into the DB in a bad state?

corehq/apps/hqwebapp/views.py
czue: i tested on a formplayer and non-formplayer domain and it worked out of the box. I was actually confused by the logic that was there and why i didn't need it. I'll try to get to the bottom of that.

czue:
oh nice catch on succeed. yeah I think @jjwack will just need to set the default landing page on the role as "reports" and then it should work, if I read that correctly?

corehq/apps/hqwebapp/views.py
snopoke: Should definitely notify but rather just redirect to the dashboard instead of 500. Since it's unlikely probably fine to leave as is.

corehq/apps/hqwebapp/views.py
benrudolph: what most likely happened is that there are two different urls for formplayer and non-formplayer. however a while ago the non-formplayer route used to support both formplayer and non-formplayer. i separated the view out but left the non-formplayer view the same (to not break anything). that view should be totally deleted soon. we'll want to add the logic back in to differentiate between the two views

benrudolph: kind of confusing sorry...

czue: ok restored that logic and added a test

snopoke (approved):

czue:
merging cause i want to build on top of this.

PR #14533, opened 2017-01-16
UCR: Add Get Case Sharing Groups Expression

sheelio:
@snopoke @calellowitz 

This adds a UCR expression for accessing case sharing groups based on a user ID.  I need it for the Pradan project's UCRs.

### DIFF
--- a/corehq/apps/userreports/expressions/specs.py
+++ b/corehq/apps/userreports/expressions/specs.py
@@ -345,7 +345,7 @@ class CaseSharingGroupsExpressionSpec(JsonObject):
 
     def _get_case_sharing_groups(self, user_id, context):
         domain = context.root_doc['domain']
-        cache_key = (self.__class__.__name__, user_id)
+        cache_key = (self.__class__.__name__, domain, user_id)
         if context.get_cache_value(cache_key) is not None:
             return context.get_cache_value(cache_key)
 
--- a/corehq/apps/userreports/expressions/specs.py
+++ b/corehq/apps/userreports/expressions/specs.py
@@ -354,6 +354,7 @@ class CaseSharingGroupsExpressionSpec(JsonObject):
             return []
 
         case_sharing_groups = user.get_case_sharing_groups()
+        case_sharing_groups = [g.to_json() for g in case_sharing_groups]
         context.set_cache_value(cache_key, case_sharing_groups)
         return case_sharing_groups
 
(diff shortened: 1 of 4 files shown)


corehq/apps/userreports/tests/test_expressions.py
```diff
@@ -1064,6 +1065,55 @@ def test_single_extension(self):
...
+        super(TestGetCaseSharingGroupsExpression, self).tearDown()
+
+    @run_with_all_backends
+    def test_no_subcases(self):
+        user = CommCareUser.create(domain=self.domain, username='test_no_group', password='123')
+        case_sharing_groups = self.expression({'user_id': user._id}, self.context)
+        self.assertEqual(len(case_sharing_groups), 0)
+
+    @run_with_all_backends
+    def test_single_child(self):
```
snopoke: `test_single_group`

corehq/apps/userreports/tests/test_expressions.py
```diff
@@ -1064,6 +1065,55 @@ def test_single_extension(self):
...
+
+    def tearDown(self):
+        for group in Group.by_domain(self.domain):
+            group.delete()
+        for user in CommCareUser.all():
+            user.delete()
+        super(TestGetCaseSharingGroupsExpression, self).tearDown()
+
+    @run_with_all_backends
+    def test_no_subcases(self):
```
snopoke: `test_no_groups`

corehq/apps/userreports/expressions/specs.py
```diff
@@ -327,6 +328,36 @@ def _get_subcases(self, case_id, context):
...
+        user_id = self._user_id_expression(item, context)
+        if not user_id:
+            return []
+
+        assert context.root_doc['domain']
+        return self._get_case_sharing_groups(user_id, context)
+
+    def _get_case_sharing_groups(self, user_id, context):
+        domain = context.root_doc['domain']
+        cache_key = (self.__class__.__name__, user_id)
```
snopoke: not really necessary but could include `domain` in the key as a safety for future in case the caching mechanism changes or something

snopoke (commented):
Looks good, could do with a test that ensures the domain checking works.

sheelio:
Thanks will add that test

snopoke (approved):

snopoke:
One other thing to do is to update the docs

sheelio:
@snopoke 

Change the expression so that it would return JSON docs instead of Group objects and updated the docs.

sheelio:
@snopoke 
Fixed the test failures so this should be good to go.

PR #14534, opened 2017-01-16
clean up / fix up role management

czue:
The main feature here is a new option on roles "Allow non-admins to assign this role to other users" which will let someone with permission to edit web users or mobile workers be able to assign a subset of roles (the ones which have this property set).

Along the way I uncovered and attempted to address a number of quirks with our role security model. In particular the main change is that only admin users can create other admin users, and everyone else who has "edit web user" permissions defaults to not being able to set roles unless those roles have been explicitly granted the new permission.

is broken out by commit though might actually be easier to read all at once, not sure.

### DIFF
--- a/corehq/apps/users/views/__init__.py
+++ b/corehq/apps/users/views/__init__.py
@@ -186,10 +186,7 @@ class BaseEditUserView(BaseUserSettingsView):
     @property
     @memoized
     def editable_role_choices(self):
-        roles = UserRole.by_domain(self.domain)
-        if not self.request.couch_user.is_domain_admin(self.domain):
-            roles = filter(lambda role: role.is_non_admin_editable, roles)
-        return [UserRole.role_to_choice(role) for role in roles]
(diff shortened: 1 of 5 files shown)


czue:
@dimagi/product @sheelio fyi, cc @orangejenny

corehq/apps/users/views/__init__.py
```diff
@@ -301,13 +305,6 @@ def user_role_choices(self):
             return [UserRole.role_to_choice(AdminUserRole(domain=self.domain))] + editable_choices
 
     @property
-    @memoized
-    def form_user_update(self):
-        form = super(EditWebUserView, self).form_user_update
-        form.load_roles(current_role=self.existing_role, role_choices=self.user_role_choices)
-        return form
```
esoergel: Was this just not used anywhere?

esoergel: oh, nm, I see it was overriding the other usage.

esoergel (approved):

czue:
@sheelio reviewed this from a product standpoint and said it was good to go

PR #14535, opened 2017-01-16
enikshay adherence update task

sravfeyn:
(mainly pseudo-code for review)


custom/enikshay/case_utils.py
```diff
@@ -111,6 +111,10 @@ def get_open_episode_case_from_person(domain, person_case_id):
 def get_adherence_cases_between_dates(domain, person_case_id, start_date, end_date):
     case_accessor = CaseAccessors(domain)
     episode = get_open_episode_case_from_person(domain, person_case_id)
+    return get_adherence_cases_between_dates_from_episode(domain, episode, start_date, end_date)
+
+
+def get_adherence_cases_between_dates_from_episode(domain, episode, start_date, end_date)
```
dimagimon: E901 SyntaxError: invalid syntax

custom/enikshay/tasks.py
```diff
@@ -0,0 +1,88 @@
...
+            adherence_date = adherence_case.adherence_date
+            if adherence_date < PURGE_DATE:
+                episode.aggregated_score_date_calculated = adherence_date
+            else:
+                episode.aggregated_score_date_calculated = PURGE_DATE
+
+            # calculate 'aggregated_score_count_taken'
+            adherence_cases = get_adherence_cases_between_dates_from_episode(
+                DOMAIN
+                episode,
```
dimagimon: E901 SyntaxError: invalid syntax

custom/enikshay/tasks.py
```diff
@@ -0,0 +1,88 @@
...
+                    case.adherence_value in DOSE_TAKEN_INDICATORS
+                    for case in cases
+                ])
+            total_taken_count = is_dose_taken_by_date.values().count(True)
+            episode.aggregated_score_count_taken = total_taken_count
+
+            # calculate 'expected' score
+            adherence_schedule_id = episode.adherence_schedule_id or DAILY_SCHEDULE_ID
+            doses_per_week = DOESES_BY_SCHEDULE_BY_ID[adherence_schedule_id]
+            episode.expected = ((aggregated_score_date_calculated - adherence_schedule_date_start) / 7) * doses_per_week
```
dimagimon: E501 line too long (120 > 115 characters)

czue:
@sravfeyn let me know if you want this reviewed

sravfeyn:
Yes, please! I am looking for adherence logic review to make sure that it is spec-correct. I will try to restructure the code, to make it easy for tests and will write some tests. So I am not looking for review on that yet.

@proteusvacuum @ctsims

custom/enikshay/tasks.py
```diff
@@ -0,0 +1,109 @@
...
+    """
+    inde
+    """
+    by_date = defaultdict(list)
+    for case in adherence_cases:
+        adherance_date = parse_date(case.dynamic_case_properties().get('adherence_date'))
+        by_date[adherance_date].append(case)
+    return by_date
+
+def get_adherence_schedule_date_start(episode_case):
```
dimagimon: E302 expected 2 blank lines, found 1

custom/enikshay/tasks.py
```diff
@@ -0,0 +1,109 @@
...
+                ])
+            total_taken_count = is_dose_taken_by_date.values().count(True)
+            episode.aggregated_score_count_taken = total_taken_count
+
+            # calculate 'expected' score
+            dose_data = get_doses_data()
+            adherence_schedule_id = episode.adherence_schedule_id or DAILY_SCHEDULE_ID
+            # Todo: what if 'adherence_schedule_id' doesn't exist in fixtures
+            doses_per_week = dose_data.get(adherence_schedule_id)
+            episode.expected = ((aggregated_score_date_calculated - adherence_schedule_date_start) / 7) * doses_per_week
```
dimagimon: E501 line too long (120 > 115 characters)

sravfeyn:
@ctsims @czue @proteusvacuum Bumping this to get [spec signoff](https://docs.google.com/document/d/1FjSdLYOYUCRBuW3aSxvu3Z5kvcN6JKbpDFDToCgead8/edit?pli=1#heading=h.ldp1ikmaxcg)

custom/enikshay/tasks.py
```diff
@@ -0,0 +1,109 @@
...
+
+def update_adherence_properties():
+    # edge cases around datetime
+    PURGE_DATE = datetime.datetime.today() - datetime.timedelta(days=60)
+
+    for episode in get_open_episode_cases():
+        adherence_schedule_date_start = get_adherence_schedule_date_start(episode)
+        # Todo: What if adherence_schedule_date_start is None
+        if adherence_schedule_date_start > PURGE_DATE:
+            episode.aggregate_date = adherence_schedule_date_start - 1
```
proteusvacuum: I think you will have to submit a form against the episode to update it (but you probably already knew that)... and I think it might be good to chunk those submissions into batches.

custom/enikshay/tasks.py
```diff
@@ -0,0 +1,109 @@
...
+def get_open_episode_cases():
+    """
+    get list of all open 'episode' type cases
+    """
+    case_accessor = CaseAccessors(DOMAIN)
+    case_ids = get_open_case_ids_in_domain_by_type(CASE_TYPE_PERSON)
+    return case_accessor.iter_cases(case_ids)
+
+
+def get_latest_adherence_case_for_episode(episode):
```
proteusvacuum: 👍

custom/enikshay/tasks.py
proteusvacuum: I *think* this should be `aggregated_score_date_calculated`. @ctsims to confirm

custom/enikshay/tasks.py
```diff
@@ -0,0 +1,109 @@
...
+
+from .case_utils import get_adherence_cases_between_dates_from_episode
+from corehq.apps.fixtures.models import FixtureDataItem
+
+
+DOMAIN = 'enikshay'
+DOSE_TAKEN_INDICATORS = [
+    'directly_observed_dose',
+    'unobserved_dose',
+    'self_administered_dos',
```
proteusvacuum: `self_administered_dose`

custom/enikshay/tasks.py
```diff
@@ -0,0 +1,109 @@
...
+    # edge cases around datetime
+    PURGE_DATE = datetime.datetime.today() - datetime.timedelta(days=60)
+
+    for episode in get_open_episode_cases():
+        adherence_schedule_date_start = get_adherence_schedule_date_start(episode)
+        # Todo: What if adherence_schedule_date_start is None
+        if adherence_schedule_date_start > PURGE_DATE:
+            episode.aggregate_date = adherence_schedule_date_start - 1
+            episode.expected = 0
+            episode.taken = 0
```
proteusvacuum: I think this should be `aggregated_score_count_taken`

custom/enikshay/tasks.py
```diff
@@ -0,0 +1,109 @@
...
+    'unobserved_dose',
+    'self_administered_dos',
+]
+DAILY_SCHEDULE_ID = 'schedule_daily'
+CASE_TYPE_PERSON = 'person'
+
+
+def get_doses_data():
+    # return 'doses_per_week' by 'schedule_id' from the Fixture data
+    fixtures = FixtureDataItem.get_indexed_items(DOMAIN, DAILY_SCHEDULE_ID, 'schedule_id')
```
proteusvacuum: The lookup table is called `adherence_schedules`, and the field is called `id`: https://india.commcarehq.org/a/enikshay/fixtures/view_lookup_tables/?table_id=1ba69af232d61c7479fc6fdecb94cc7e

custom/enikshay/tasks.py
proteusvacuum: I think you should call this property `expected_doses_taken` or something more descriptive (also elsewhere)

proteusvacuum (commented):
Looks like a good start sravs. There was only a few things that I clarified.

custom/enikshay/tasks.py
```diff
@@ -0,0 +1,109 @@
...
+    return dict((k, int(fixture['doses_per_week'])) for k, fixture in fixtures.items())
+
+
+def get_open_episode_cases():
```
sravfeyn: Sorry, I meant to do `get_open_case_ids_in_domain_by_type(CASE_TYPE_EPISODE)`. Not by `CASE_TYPE_PERSON`.

custom/enikshay/tasks.py
sravfeyn: Yeah, something like `submit_case_blocks`. Thanks

sravfeyn: 👍

custom/enikshay/tasks.py
sravfeyn: 👍

custom/enikshay/tasks.py
sravfeyn: Okay, I have got it from the spec (assumed the property was so named in the app). Should we update the spec with that title for record? I can update the code

custom/enikshay/tasks.py
```diff
@@ -0,0 +1,109 @@
...
+        by_date[adherance_date].append(case)
+    return by_date
+
+def get_adherence_schedule_date_start(episode_case):
+    return parse_datetime(case.dynamic_case_properties().get('adherence_schedule_date_start'))
+
+
+def update_adherence_properties():
+    # edge cases around datetime
+    PURGE_DATE = datetime.datetime.today() - datetime.timedelta(days=60)
```
sravfeyn: What kind of verification? By `purge_date`, I understood as date 60 days ago. Please can you list any more spec details as well?

custom/enikshay/tasks.py
```diff
@@ -0,0 +1,109 @@
...
+    for episode in get_open_episode_cases():
+        adherence_schedule_date_start = get_adherence_schedule_date_start(episode)
+        # Todo: What if adherence_schedule_date_start is None
+        if adherence_schedule_date_start > PURGE_DATE:
+            episode.aggregate_date = adherence_schedule_date_start - 1
+            episode.expected = 0
+            episode.taken = 0
+        else:
+            adherence_case = get_latest_adherence_case_for_episode(episode)
+            # ToDo: what if adherence_case doesn't exist?
```
sravfeyn: @proteusvacuum, I have added some 'edge-case' questions titled 'Todo', can you respond on what should happen in such cases.

custom/enikshay/tasks.py
```diff
@@ -0,0 +1,109 @@
...
+
+
+DOMAIN = 'enikshay'
+DOSE_TAKEN_INDICATORS = [
+    'directly_observed_dose',
+    'unobserved_dose',
+    'self_administered_dos',
+]
+DAILY_SCHEDULE_ID = 'schedule_daily'
+CASE_TYPE_PERSON = 'person'
```
proteusvacuum: FYI, all the case type constants are defined in `case_utils.py`

custom/enikshay/tasks.py
```diff
@@ -0,0 +1,109 @@
+import datetime
+from collections import defaultdict
+from django.utils.dateparse import parse_datetime, parse_date
+
+from .case_utils import get_adherence_cases_between_dates_from_episode
+from corehq.apps.fixtures.models import FixtureDataItem
+
+
+DOMAIN = 'enikshay'
```
proteusvacuum: This seems pretty brittle, and would be good to be able to use on our `enikshay-test` domain too. Maybe make this a feature flag thing?

custom/enikshay/tasks.py
```diff
@@ -0,0 +1,109 @@
...
+    indexed_cases = case_accessor.get_reverse_indexed_cases([episode.case_id])
+    latest_date = 0
+    latest_case = None
+    for case in indexed_cases:
+        adherence_date = parse_datetime(case.dynamic_case_properties().get('adherence_date'))
+        if (not case.closed and
+           case.type == CASE_TYPE_ADHERENCE and
+           adherence_date > latest_date):
+            latest_date = adherence_date
+            latest_case = adherence_date
```
proteusvacuum: I think this should be `latest_case = case`

custom/enikshay/tasks.py
proteusvacuum: Verify with the @ctsims that 60 days is the appropriate time window to do the purge.

custom/enikshay/tasks.py
proteusvacuum: I think the `aggregated_score_count_taken` will be 0, and the `aggregated_score_date_calculated` will be the `PURGE_DATE`

custom/enikshay/tasks.py
```diff
@@ -0,0 +1,109 @@
...
+                    case.adherence_value in DOSE_TAKEN_INDICATORS
+                    for case in cases
+                ])
+            total_taken_count = is_dose_taken_by_date.values().count(True)
+            episode.aggregated_score_count_taken = total_taken_count
+
+            # calculate 'expected' score
+            dose_data = get_doses_data()
+            adherence_schedule_id = episode.adherence_schedule_id or DAILY_SCHEDULE_ID
+            # Todo: what if 'adherence_schedule_id' doesn't exist in fixtures
```
proteusvacuum: Probably just set `expected_doses_taken` to 0 and raise a soft_assert.

custom/enikshay/tasks.py
proteusvacuum: I think this should actually be `aggregated_score_count_expected`. That's what is used in the app.
@ctsims

custom/enikshay/tasks.py
```diff
@@ -0,0 +1,109 @@
...
+    return parse_datetime(case.dynamic_case_properties().get('adherence_schedule_date_start'))
+
+
+def update_adherence_properties():
+    # edge cases around datetime
+    PURGE_DATE = datetime.datetime.today() - datetime.timedelta(days=60)
+
+    for episode in get_open_episode_cases():
+        adherence_schedule_date_start = get_adherence_schedule_date_start(episode)
+        # Todo: What if adherence_schedule_date_start is None
```
proteusvacuum: I don't think these calculations should be run in that case because I think that would indicated that an adherence schedule hasn't been selected. @ctsims to confirm.

custom/enikshay/tasks.py
sravfeyn: Yeah. I want to look at flexibility of this. Just wanted to get spec sign off for now

proteusvacuum: cool. I'm signed off. Let me know if you have any more questions

custom/enikshay/tasks.py
sravfeyn: 👍

custom/enikshay/tasks.py
dimagimon: E302 expected 2 blank lines, found 1

custom/enikshay/tasks.py
```diff
@@ -0,0 +1,109 @@
...
+            adherence_date = parse_datetime(adherence_case.dynamic_case_properties().get('adherence_date'))
+            if adherence_date < PURGE_DATE:
+                episode.aggregated_score_date_calculated = adherence_date
+            else:
+                episode.aggregated_score_date_calculated = PURGE_DATE
+
+            # calculate 'aggregated_score_count_taken'
+            adherence_cases = get_adherence_cases_between_dates_from_episode(
+                DOMAIN
+                episode,
```
dimagimon: E901 SyntaxError: invalid syntax

custom/enikshay/tasks.py
dimagimon: E501 line too long (120 > 115 characters)

sravfeyn:
@proteusvacuum @czue This is ready for final review. I am removing `don't merge` label

custom/enikshay/tests/test_adherence_updates.py
```diff
@@ -0,0 +1,292 @@
...
+        self.case_updater.purge_date = pytz.UTC.localize(input[0])
+        # setup episode and adherence cases
+        adherence_schedule_date_start, adherence_schedule_id = input[1]
+        adherence_cases = input[2]
+        episode = self._create_episode_case(adherence_schedule_date_start, adherence_schedule_id)
+        self._create_adherence_cases(adherence_cases)
+
+        return EpisodeUpdate(episode, self.case_updater)
+
+    def test_adherence_schedule_date_start_late(self):
```
sravfeyn: @ctsims 
Hey Clayton, these are some test-cases for adherence update. Can you review these test cases and make sure they are upto spec?

custom/enikshay/tasks.py
```diff
@@ -0,0 +1,180 @@
...
+                    "No fixture item found with schedule_id {}".format(adherence_schedule_id)
+                )
+            return update
+
+    def case_block(self):
+        update = self.update_json()
+        if update:
+            return CaseBlock(**{
+                'case_id': self.episode.case_id
+                'create': False,
```
dimagimon: E901 SyntaxError: invalid syntax

custom/enikshay/tasks.py
```diff
@@ -0,0 +1,180 @@
...
+    queue=getattr(settings, 'CELERY_PERIODIC_QUEUE', 'celery')
+)
+def enikshay_adherence_task():
+    domains = toggles.UATBC_ADHERENCE_TASK.get_enabled_domains()
+    for domain in domains:
+        updater = PeriodicCaseUpdater(domain)
+        updater.run()
+
+
+class PeriodicCaseUpdater(object):
```
czue: could be nice to add a few more doc strings / comments to this code. is a bit hard to wrap your head around what it's doing without having a good amount of familiarity with the project needs

custom/enikshay/tasks.py
```diff
@@ -0,0 +1,180 @@
...
+        return open case of type 'adherence' reversed-indexed to episode and
+            with latest 'adherence_date' property
+        """
+        case_accessor = CaseAccessors(self.case_updater.domain)
+        indexed_cases = case_accessor.get_reverse_indexed_cases([self.episode.case_id])
+        # sometime far back
+        latest_date = pytz.UTC.localize(datetime.datetime(1990, 1, 1))
+        latest_case = None
+        for case in indexed_cases:
+            adherence_date = parse_datetime(case.dynamic_case_properties().get('adherence_date'))
```
czue: does this crash if `adherence_date` is missing or malformed? if so, might want to account for that since my current read is that the code as-is would crash and nothing in the entire domain would get updated.

custom/enikshay/tasks.py
```diff
@@ -0,0 +1,180 @@
...
+
+
+class PeriodicCaseUpdater(object):
+
+    def __init__(self, domain):
+        self.domain = domain
+        self.purge_date = pytz.UTC.localize(datetime.datetime.today() - datetime.timedelta(days=60))
+
+    def run(self):
+        for episode in self._get_open_episode_cases():
```
czue: what do you think about putting this whole loop in a try/catch/notify block so individual errors don't cause the entire domain's data to fail?

custom/enikshay/tasks.py
```diff
@@ -0,0 +1,180 @@
...
+    def get_property(self, property):
+        return self.episode.dynamic_case_properties().get(property)
+
+    def get_latest_adherence_case_for_episode(self):
+        """
+        return open case of type 'adherence' reversed-indexed to episode and
+            with latest 'adherence_date' property
+        """
+        case_accessor = CaseAccessors(self.case_updater.domain)
+        indexed_cases = case_accessor.get_reverse_indexed_cases([self.episode.case_id])
```
czue: it's pretty tempting to try and push this logic into elasticsearch. as-is it seems like this will have to load every single adherence case for every single episode in the entire domain every time it runs right? I think that is going to be pretty untenable long term. could we possibly use the case search index for this?

custom/enikshay/tasks.py
```diff
@@ -0,0 +1,180 @@
...
+    for domain in domains:
+        updater = PeriodicCaseUpdater(domain)
+        updater.run()
+
+
+class PeriodicCaseUpdater(object):
+
+    def __init__(self, domain):
+        self.domain = domain
+        self.purge_date = pytz.UTC.localize(datetime.datetime.today() - datetime.timedelta(days=60))
```
czue: the intention of this is to define the window in which we don't need to worry about running this task based on your schedule start date right? i'm trying to wrap my head around whether this is ok. is there a specific part of the spec you can refer me to? it feels wrong to me that we don't do anything to reconcile your first 60 days of data.

custom/enikshay/tasks.py
```diff
@@ -0,0 +1,180 @@
...
+        if not adherence_schedule_date_start:
+            # adherence schedule hasn't been selected, so no update necessary
+            return {}
+
+        if adherence_schedule_date_start > self.case_updater.purge_date:
+            return {
+                'aggregated_score_date_calculated': adherence_schedule_date_start - datetime.timedelta(days=1),
+                'expected_doses_taken': 0,
+                'aggregated_score_count_taken': 0
+            }
```
czue: an api like this feels like a pretty good place to use a real class or a `namedtuple`

custom/enikshay/tasks.py
czue: i also think linking to the spec doc somewhere in this file would be helpful

custom/enikshay/tasks.py
```diff
@@ -0,0 +1,180 @@
...
+                update["aggregated_score_date_calculated"] = self.case_updater.purge_date
+                update["aggregated_score_count_taken"] = 0
+            elif adherence_date < self.case_updater.purge_date:
+                update["aggregated_score_date_calculated"] = adherence_date
+            else:
+                update["aggregated_score_date_calculated"] = self.case_updater.purge_date
+
+            # calculate 'aggregated_score_count_taken'
+            if adherence_case:
+                adherence_cases = get_adherence_cases_between_dates_from_episode(
```
czue: if we are going to load all the adherence cases from the episode to get the latest one, we should at least reuse that list here to not have to go back to the DB. Again, moving to elasticsearch would probably help a lot for performance of this.

custom/enikshay/tests/test_adherence_updates.py
```diff
@@ -0,0 +1,291 @@
...
+        cls.setupFixtureData()
+        FormProcessorTestUtils.delete_all_cases()
+
+    def setUp(self):
+        super(TestAdherenceUpdater, self).setUp()
+        delete_all_users()
+        self.factory = CaseFactory(domain=self.domain)
+        self.user = CommCareUser.create(
+            self.domain,
+            "jon-snow@user",
```
czue: nice to see you sticking with enikshay testing conventions. :wolf:

custom/enikshay/tests/test_adherence_updates.py
```diff
@@ -0,0 +1,291 @@
...
+    @classmethod
+    def setUpClass(cls):
+        super(TestAdherenceUpdater, cls).setUpClass()
+        cls.domain = 'adherence-enikshay-test'
+        cls.setupFixtureData()
+        FormProcessorTestUtils.delete_all_cases()
+
+    def setUp(self):
+        super(TestAdherenceUpdater, self).setUp()
+        delete_all_users()
```
czue: is this definitely necessary?

custom/enikshay/tests/test_adherence_updates.py
```diff
@@ -0,0 +1,291 @@
...
+                        ]
+                    )
+                },
+                item_attributes={},
+            )
+            data_item.save()
+            cls.data_items.append(data_item)
+
+    def tearDown(self):
+        delete_all_users()
```
czue: nit: I think this could be `self.user.delete()` which might be slightly less work

custom/enikshay/tests/test_adherence_updates.py
```diff
@@ -0,0 +1,291 @@
...
+        super(TestAdherenceUpdater, self).tearDown()
+
+    @classmethod
+    def tearDownClass(cls):
+        for domain in Domain.get_all():
+            domain.delete()
```
czue: i think this is a particularly slow function because it calls all sorts of other stuff. is there a reason this is necessary?

custom/enikshay/tests/test_adherence_updates.py
czue: these look like great tests. @ctsims can you please check the logic and confirm it looks right?

custom/enikshay/tests/utils.py
```diff
@@ -27,6 +27,115 @@
 from corehq.apps.users.models import CommCareUser
 
 
+def get_person_case_structure(case_id, user_id, extra_update={}):
```
czue: nit: [mutable default args](http://docs.python-guide.org/en/latest/writing/gotchas/#mutable-default-arguments)

custom/enikshay/tests/utils.py
```diff
@@ -27,6 +27,115 @@
...
+        'aadhaar_number': "499118665246",
+        MERM_ID: "123456789",
+        'dob': "1987-08-15",
+        'age': '20',
+        'sex': 'male',
+        'current_address': 'Mr. Everest',
+        'secondary_contact_name_address': 'Mrs. Everestie',
+        'previous_tb_treatment': 'yes',
+        'nikshay_registered': "false",
+    }
```
czue: didn't review these properties for correctness. let me know if you want me to

czue: oh cool, see you were just refactoring this.

czue (commented):
this is looking great sravs. thanks for putting it all together and for the great tests.

wanted to start a conversation about a few possible scalability concerns I have. let me know if it's better to offline some of this. also would like @ctsims to sign off on the tests. also had a bunch of small stuff.

custom/enikshay/tasks.py
sravfeyn: It's mentioned here
https://docs.google.com/a/dimagi.com/document/d/1FjSdLYOYUCRBuW3aSxvu3Z5kvcN6JKbpDFDToCgead8/edit?pli=1&disco=AAAAAv94gJg

I don't honestly understand the project context of purge date, I have tried to follow the spec and write good test cases to validate

custom/enikshay/tasks.py
sravfeyn: Great point, will add it.

custom/enikshay/tasks.py
sravfeyn: I will chat offline with you re. this

custom/enikshay/tasks.py
sravfeyn: I haven't accounted for exceptions, will add try-catches wherever needed

custom/enikshay/tests/test_adherence_updates.py
sravfeyn: I was going to remove, had to put it there for some cleanup.

custom/enikshay/tests/test_adherence_updates.py
sravfeyn: Was supposed to be temporary cleanup. Removed it

custom/enikshay/tests/utils.py
sravfeyn: I am not updating `extra_update`, so it should be okay, no?

custom/enikshay/tasks.py
czue: (although I think i realized later that these are case properties that get passed to `update` so maybe not worth it)

custom/enikshay/tests/utils.py
czue: it should be, though is more future proof to make the default `None` and then have `extra_update = extra_update or {}` at the top of the function. just in case later someone else updates this function to modify it some how

custom/enikshay/tests/utils.py
sravfeyn: yeah

custom/enikshay/tests/utils.py
sravfeyn: Cool, updated

custom/enikshay/tests/test_adherence_updates.py
```diff
@@ -0,0 +1,363 @@
...
+    def test_no_adherence_cases(self):
+        self.assert_update(
+            (
+                datetime(2016, 1, 20),
+                (datetime(2016, 1, 10), 'schedule1'),
+                []
+            ),
+            {   # change
+                'aggregated_score_date_calculated': date(2016, 1, 20),
+                'expected_doses_taken': (10.0 / 7) * int(self.fixture_data['schedule1']),
```
ctsims: I think this should be 0 here. Since there's no adherence data there's no expected doses. 

The "10" in this case should ~always match (adherence_latest_date_recorded - adherence_schedule_date_start)

custom/enikshay/tests/test_adherence_updates.py
```diff
@@ -0,0 +1,363 @@
...
+
+    def test_no_adherence_cases(self):
+        self.assert_update(
+            (
+                datetime(2016, 1, 20),
+                (datetime(2016, 1, 10), 'schedule1'),
+                []
+            ),
+            {   # change
+                'aggregated_score_date_calculated': date(2016, 1, 20),
```
ctsims: I think this should match 'adherence_latest_date_recorded' from below ( no data, so it's 1 day before).

adherence_latest_date_recorded and aggregated_score_date_calculated will be the same if the purge date is after aggregated_score_date_calculated

custom/enikshay/tests/test_adherence_updates.py
```diff
@@ -0,0 +1,363 @@
...
+                datetime(2016, 1, 20),
+                (datetime(2016, 1, 10), 'schedule1'),
+                [(datetime(2016, 1, 15), DOSE_TAKEN_INDICATORS[0])]
+            ),
+            {
+                'aggregated_score_date_calculated': date(2016, 1, 15),
+                'expected_doses_taken': (5.0 / 7) * int(self.fixture_data['schedule1']),
+                'aggregated_score_count_taken': 1,
+                'adherence_latest_date_recorded': date(2016, 1, 15),
+                'adherence_total_doses_taken': 1
```
ctsims: This looks right to me

custom/enikshay/tests/test_adherence_updates.py
```diff
@@ -0,0 +1,363 @@
...
+            }
+        )
+
+    def test_adherence_date_greater_than_purge_date(self):
```
ctsims: Great catch on this test case! Looks right to me, and is an important one to identify

custom/enikshay/tests/test_adherence_updates.py
```diff
@@ -0,0 +1,363 @@
...
+        self.assert_update(
+            (
+                datetime(2016, 1, 20),
+                (datetime(2016, 1, 10), 'schedule1'),
+                [
+                    # same day, different time. Set hours different so that case-id becomes different
+                    (datetime(2016, 1, 11, 1), DOSE_TAKEN_INDICATORS[0]),
+                    (datetime(2016, 1, 11, 3), 'unknown'),
+                    (datetime(2016, 1, 12), DOSE_TAKEN_INDICATORS[0]),
+                    (datetime(2016, 1, 14), 'unknown'),
```
ctsims: reminder that we want to use whatever we set up as the _missing_data thing instead of unknown, just for consistency with what will be coming in from the device.

custom/enikshay/tests/test_adherence_updates.py
```diff
@@ -0,0 +1,363 @@
...
+                    (datetime(2016, 1, 21), DOSE_TAKEN_INDICATORS[3])  # dose missed
+                ]
+            ),
+            {
+                'aggregated_score_date_calculated': date(2016, 1, 20),
+                'expected_doses_taken': (10.0 / 7) * int(self.fixture_data['schedule1']),
+                # ask clayton
+                'aggregated_score_count_taken': 2,
+                'adherence_latest_date_recorded': date(2016, 1, 21),
+                'adherence_total_doses_taken': 3
```
ctsims: This should be 2, shouldn't it? 2 taken doses, 2 missed doses.

custom/enikshay/tests/test_adherence_updates.py
```diff
@@ -0,0 +1,363 @@
...
+                    (datetime(2016, 1, 12), DOSE_TAKEN_INDICATORS[0]),
+                    (datetime(2016, 1, 14), 'unknown'),
+                    (datetime(2016, 1, 21), DOSE_TAKEN_INDICATORS[3])  # dose missed
+                ]
+            ),
+            {
+                'aggregated_score_date_calculated': date(2016, 1, 20),
+                'expected_doses_taken': (10.0 / 7) * int(self.fixture_data['schedule1']),
+                # ask clayton
+                'aggregated_score_count_taken': 2,
```
ctsims: this is right

custom/enikshay/tests/test_adherence_updates.py
```diff
@@ -0,0 +1,363 @@
...
+            {
+                'aggregated_score_date_calculated': date(2016, 1, 11),
+                'expected_doses_taken': (1.0 / 7) * int(self.fixture_data['schedule1']),
+                'aggregated_score_count_taken': 1,
+                'adherence_latest_date_recorded': date(2016, 1, 11),
+                'adherence_total_doses_taken': 1
+            }
+        )
+
+    def test_two_doses_on_same_day_one_missed(self):
```
ctsims: One other potential test case could be two doses on the same day, but different indicators (IE: one directly observed, one self reported), on the same day, which should also be recorded as 1 dose (not two)

custom/enikshay/tests/test_adherence_updates.py
```diff
@@ -0,0 +1,470 @@
...
+            ),
+            {
+                # 1 day before adherence_schedule_date_start
+                'aggregated_score_date_calculated': date(2016, 1, 9),
+                # set to zero
+                'expected_doses_taken': 0,
+                'aggregated_score_count_taken': 0,
+                'adherence_total_doses_taken': 0
+                # 1 day before should be adherence_schedule_date_start
+                'adherence_latest_date_recorded': date(2016, 1, 9),
```
dimagimon: E901 SyntaxError: invalid syntax

custom/enikshay/tests/test_adherence_updates.py
sravfeyn: Makes, sense. Updated

custom/enikshay/tests/test_adherence_updates.py
sravfeyn: Updated

custom/enikshay/tests/test_adherence_updates.py
sravfeyn: Updated

custom/enikshay/tests/test_adherence_updates.py
sravfeyn: Yeah, updated

custom/enikshay/tests/test_adherence_updates.py
sravfeyn: Added this and few more test cases with various permutations of missed, unknown, less and greater dates.

corehq/apps/reports/tests/test_esaccessors.py
```diff
@@ -1029,6 +1029,33 @@ def _send_case_to_es(self,
         self.es.indices.refresh(CASE_INDEX_INFO.index)
         return case
 
+    def test_reverse_index(self):
+        child_case_id = uuid.uuid4().hex
+        parent_case_id = uuid.uuid4().hex
+        from casexml.apps.case.sharedmodels import CommCareCaseIndex
+        from casexml.apps.case.const import CASE_INDEX_CHILD, CASE_INDEX_EXTENSION
```
dimagimon: F401 'CASE_INDEX_CHILD' imported but unused

sravfeyn:
Hey @czue, I have implemented all of your feedback
- Switch to ES
- Remove redundant case lookups
- docstrings
- other assorted comments.

Can you review last few comments and merge if everything is ok?

I've also incorporated spec changes by Clayton, he is now signed off on the test cases (@ctsims).

custom/enikshay/tasks.py
```diff
@@ -0,0 +1,267 @@
...
+            except Exception, e:
+                logger.error("Error calculating adherence values for episode case_id({}): {}".format(
+                    episode.case_id,
+                    e
+                ))
+
+    def _get_open_episode_cases(self):
+        # return all open 'episode' cases
+        case_accessor = CaseAccessors(self.domain)
+        case_ids = case_accessor.get_open_case_ids_in_domain_by_type(CASE_TYPE_EPISODE)
```
sravfeyn: Should I switch this also to ES? This is just one time call per task run. 

I wasn't able to find ES util for this, it's pretty easy to write, I will write one if it's needed

custom/enikshay/tasks.py
```diff
@@ -0,0 +1,267 @@
...
+        """
+        # sometime far back
+        latest_date = pytz.UTC.localize(datetime.datetime(1990, 1, 1))
+        latest_case = None
+        for case in self.get_valid_adherence_cases():
+            adherence_date = parse_datetime(case.dynamic_case_properties().get('adherence_date'))
+            if adherence_date > latest_date:
+                latest_date = adherence_date
+                latest_case = case
+        return latest_case
```
czue: sorry i think i should have been clearer in my previous comment. so i think the main reason to push this to ES would be so that you can have the database return just the latest case and not have to grab every single one and do this check in python. unfortunately, since `adherence_date` is a dynamic property i don't think it's indexed which was why I was wondering if we could use the case search index or possibly make our own thing.

custom/enikshay/tasks.py
```diff
@@ -0,0 +1,267 @@
...
+            two adherence_cases on one day at different time, it will be counted as one
+        """
+        by_date = defaultdict(list)
+        for case in adherence_cases:
+            adherence_date = parse_datetime(case.dynamic_case_properties().get('adherence_date')).date()
+            adherence_value = case.dynamic_case_properties().get('adherence_value')
+            if adherence_value in DOSE_TAKEN_INDICATORS:
+                by_date[adherence_date].append(case)
+
+        return len(by_date.keys())
```
czue: likewise i think maybe this aggregation could be pushed into ES (with the same caveats/complexity mentioned above).

czue (commented):
thanks for the updates Sravan! I'm still a bit worried about the processing times here if we're doing aggregations in memory in python. Since I'm not going to be around to push this conversation through to a final decision I wonder if it makes sense to loop in @NoahCarnahan or @snopoke and possibly even architecture team to try and figure out whether we think the current approach is sustainable? it is feeling risky to me

ctsims:
@sravfeyn In terms of what we move to ES: I think it would make a lot of sense for us to think through what we need to be indexing for this (and other) behaviors to be performant on eNikshay. 

Unfortunately I think this may be really tough to due to the complexities that have come up (IE: Needing to ignore "missing data" doses and needing to do "implicit de-aggregation" at a per-day basis) 

Would it be helpful for future work for us to articulate what, like, the "clean" data set is for adherence data that we'd end up running operations on or is it easier to figure out what's aggregatable task-by-task?

corehq/pillows/mappings/case_mapping.py
```diff
@@ -31,7 +31,7 @@
                                         'referenced_id': {'type': 'string'},
                                         'referenced_type': {'type': 'string'}
                                     },
-                                    'type': 'object'},
+                                    'type': 'nested'},
```
sravfeyn: Yeah, I will add reindex label

custom/enikshay/tasks.py
sravfeyn: Ha, now I see the issue you are pointing out. There is potential custom ES queries that can be written here. I will dig into this a bit more and come up with a proposal

@NoahCarnahan @snopoke

sravfeyn:
Clayton, definitely sounds like the final adherence data and future adherence calculations requirements will have an impact on how we solve ES/performance issues here. I will chat with you offline tomorrow re this

custom/enikshay/tasks.py
proteusvacuum: haven't followed this too closely, but the thing that @czue was mentioning is the `CaseSearchES`, which indexes each dynamic property already (it is turned on for this domain, so should be safe to use). You can see some examples [here](https://github.com/dimagi/commcare-hq/blob/8667a4159ad39d3307f92a3812d74006d610fd1b/corehq/apps/es/case_search.py#L38-L50)

proteusvacuum: (you may have looked at that already though)

sravfeyn: Yeah, this turned out to require more than `CaseSearchES`, currently looking into this issue

custom/enikshay/tests/test_adherence_updates.py
```diff
@@ -0,0 +1,498 @@
...
+from django.test import TestCase
+
+from corehq.apps.fixtures.models import FixtureDataType, FixtureTypeField, \
+    FixtureDataItem, FieldList, FixtureItemField
+from corehq.form_processor.tests.utils import FormProcessorTestUtils
+
+from casexml.apps.case.mock import CaseFactory
+from corehq.apps.users.models import CommCareUser
+from corehq.apps.userreports.tasks import rebuild_indicators
+from corehq.form_processor.tests.utils import FormProcessorTestUtils
```
dimagimon: F811 redefinition of unused 'FormProcessorTestUtils' from line 7

sravfeyn:
@NoahCarnahan @proteusvacuum hey guys, can you review this, specifically the latest commits that push lot of python calculations to ES. I have one more spec update coming tomorrow, but wanted to have your review so that I can respond tomorrow. 

@emord buddy

corehq/apps/reports/tests/test_esaccessors.py
```diff
@@ -1029,6 +1029,33 @@ def _send_case_to_es(self,
...
+
+        case = self.es.indices.refresh(CASE_INDEX_INFO.index)
+
+        result = (CaseES()
+             .domain(self.domain)
+             .reverse_indexed(parent_case_id)
+             .run())
+
+        self.assertEqual(result.total, 1)
+        print result.hits
```
proteusvacuum: @sravfeyn this print statement is still here

proteusvacuum:
@sravfeyn I'm not sure I have enough context on this any more to do a proper review -- do you want someone to review alongside the spec? Or just to review moving to ES?

corehq/pillows/mappings/case_mapping.py
emord: This won't be reindexed because the `CASE_INDEX` did not change. @benrudolph this should get included with your PR

benrudolph: 👍 good catch

NoahCarnahan (commented):
Looks good Sraven. I'm curious if it's possible to use the existing cases or case search index for the ES queries? Didn't look too closely but I maybe they not store the indices in a searchable way?

custom/enikshay/ucr/data_sources/adherence.json
```diff
@@ -0,0 +1,101 @@
+{
+    "domains": [
```
emord: Can you add the `server_environment` here

custom/enikshay/ucr/data_sources/adherence.json
```diff
@@ -0,0 +1,101 @@
...
+        "enikshay-test-3",
+        "enikshay-nikshay-migration-test",
+        "enikshay-domain-copy-test",
+        "enikshay-aks-audit",
+        "np-migration-3"
+    ],
+    "config": {
+        "referenced_doc_type": "CommCareCase",
+        "engine_id": "ucr",
+        "backend_id": "ES",
```
emord: @sravfeyn what are the benefits you see from using ES here? It looks like you're just doing some basic filtering for now.

custom/enikshay/ucr/data_sources/adherence.json
```diff
@@ -0,0 +1,101 @@
...
+        "enikshay-test-2",
+        "enikshay-test-3",
+        "enikshay-nikshay-migration-test",
+        "enikshay-domain-copy-test",
+        "enikshay-aks-audit",
+        "np-migration-3"
+    ],
+    "config": {
+        "referenced_doc_type": "CommCareCase",
+        "engine_id": "ucr",
```
emord: You don't need `engine_id` for ES data sources

corehq/apps/reports/analytics/esaccessors.py
```diff
@@ -62,6 +63,17 @@ def get_active_case_counts_by_owner(domain, datespan, case_types=None, owner_ids
     return _get_case_case_counts_by_owner(domain, datespan, case_types, False, owner_ids)
 
 
+def get_reverse_indexed_cases_es(domain, case_id):
```
emord: If this is not used I think it should be possible to remove the changes to the index and associated tests from this pr and move it to another pr if it will be needed later

sravfeyn: Ha, makes sense, I will remove it if I don't need it. Still going back n forth on this

custom/enikshay/ucr/data_sources/adherence.json
sravfeyn: Added

custom/enikshay/tests/utils.py
```diff
@@ -212,7 +269,16 @@ def lab_referral(self):
 
     def create_adherence_cases(self, adherence_dates, adherence_source='99DOTS'):
         return self.factory.create_or_update_cases([
-            self._get_adherence_case_structure(adherence_date, adherence_source, "unobserved_dose")
+            get_adherence_case_structure(
+                adherence_date.strftime('%Y-%m-%d')
+                self.episode_id,
```
dimagimon: E901 SyntaxError: invalid syntax

custom/enikshay/ucr/data_sources/adherence.json
```diff
@@ -0,0 +1,158 @@
...
+        "enikshay",
+        "enikshay-test-2",
+        "enikshay-test-3",
+        "enikshay-nikshay-migration-test",
+        "enikshay-domain-copy-test",
+        "enikshay-aks-audit",
+        "np-migration-3"
+    ],
+    "server_environment": [
+      "softlayer",
```
emord: Trailing commas aren't valid in JSON. i'm not sure if python handles this correctly

corehq/apps/reports/analytics/esaccessors.py
```diff
@@ -1,6 +1,7 @@
 from collections import defaultdict, namedtuple
 from datetime import datetime
 
+from casexml.apps.case.models import CommCareCase
```
dimagimon: F401 'CommCareCase' imported but unused

sravfeyn:
Offlined with Noah re performance issues, here is some of [the reasoning](https://docs.google.com/document/d/1_ARZHUrH0G-3wHGyiHA44sO8s2sDgi0B7ss4rqguVQY/edit) for current approach.

I will add datadog metrics to profile the task in a separate PR

custom/enikshay/tasks.py
```diff
@@ -0,0 +1,295 @@
...
+        else:
+            dose_taken_by_date = self._cache_dose_taken_by_date
+
+        if bool(lte) != bool(lte):
+            raise Exception("Both of lte and gte should be specified or niether of them")
+
+        if not lte:
+            return dose_taken_by_date.values().count(True)
+        else:
+            # any efficient way to do this - numpy, python bisect?
```
NoahCarnahan: This doesn't look like it will be a bottleneck to me. Filtering this list is just `O(n)`.

custom/enikshay/tasks.py
```diff
@@ -0,0 +1,295 @@
...
+                    lte=adherence_schedule_date_start,
+                    gte=update["aggregated_score_date_calculated"]
+                )
+
+                # calculate 'expected_doses_taken' score
+                dose_data = self.case_updater.get_doses_data()
+                adherence_schedule_id = self.get_property('adherence_schedule_id') or DAILY_SCHEDULE_ID
+                doses_per_week = dose_data.get(adherence_schedule_id)
+                if doses_per_week:
+                    update['expected_doses_taken'] = ((
```
NoahCarnahan: This will be a float, just want to confirm that's intended?

custom/enikshay/tasks.py
```diff
@@ -0,0 +1,295 @@
...
+    """
+    def __init__(self, episode_case, case_updater):
+        """
+        Args:
+            episode_case: An 'episode' case object
+            case_repeater: EpisodeAdherenceUpdater object
+        """
+        self.episode = episode_case
+        self.case_updater = case_updater
+        self._cache_dose_taken_by_date = False
```
NoahCarnahan: I don't think the value of this property ever changes.

custom/enikshay/data_store.py
```diff
@@ -0,0 +1,35 @@
...
+        )
+
+    @memoized
+    def dose_known_adherences(self, episode_id):
+        return self.es.filter(self._base_filters(episode_id)).run().hits
+
+    def latest_adherence_date(self, episode_id):
+        result = self.dose_known_adherences(episode_id)
+        if len(result) > 0:
+            latest = sorted(result, key=lambda x: x['adherence_date'])[-1]
```
NoahCarnahan: You're sorting in python because `dose_known_adherences` is memoized, right? Might just be worth adding a comment to this function to explain that that is why you aren't sorting in ES. I could imagine someone coming along later and thinking that they were speeding things up by moving the sort into ES.

custom/enikshay/ucr/data_sources/adherence.json
```diff
@@ -0,0 +1,158 @@
...
+    ],
+    "config": {
+        "referenced_doc_type": "CommCareCase",
+        "backend_id": "ES",
+        "description": "",
+        "base_item_expression": {
+        },
+        "table_id": "adherence",
+        "display_name": "adherence",
+        "configured_filter": {
```
NoahCarnahan: I just realized that you could exclude the cases that don't have DOSE_KNOWN_INDICATORS from the data source itself. However, if you ever wanted to change this list, we would need to rebuild the data source.

NoahCarnahan (changes requested):
Looking good Sravan, just a couple small things.

custom/enikshay/tasks.py
sravfeyn: Good catch! Fixed

custom/enikshay/tasks.py
```diff
@@ -0,0 +1,295 @@
...
+            two adherence_cases on one day at different time, it will be counted as one
+        """
+
+        if not self._cache_dose_taken_by_date and self._cache_dose_taken_by_date != {}:
+            # we can't memoize this as the arg is list.
+            dose_taken_by_date = self.calculate_doses_taken_by_day(adherence_cases)
+        else:
+            dose_taken_by_date = self._cache_dose_taken_by_date
+
+        if bool(lte) != bool(lte):
```
sravfeyn: Oh yeah, fixed.

custom/enikshay/tasks.py
sravfeyn: 👍

custom/enikshay/tasks.py
sravfeyn: Hey @ctsims - I asked you this before. We decided on making this a floor right?

custom/enikshay/tasks.py
```diff
@@ -0,0 +1,296 @@
...
+            total count of adherence_cases excluding duplicates on a given day. If there are
+            two adherence_cases on one day at different time, it will be counted as one
+        """
+
+        if not self._cache_dose_taken_by_date and self._cache_dose_taken_by_date != {}:
+            # memoize this, `calculate_doses_taken_by_day` can't be memoized using decorator as it uses list arg
+            dose_taken_by_date = self.calculate_doses_taken_by_day(adherence_cases)
+            self._cache_dose_taken_by_date = dose_taken_by_date
+        else:
+            dose_taken_by_date = self._cache_dose_taken_by_date
```
NoahCarnahan: Hey, I just realized that you can probably do away with this memoization dict (which sort of obfuscates what's going on) by just passing `dose_taken_by_date` to `count_doses_taken()` instead of `adherence_cases`. I think this change would be nice to have, but not blocking the PR since I know this is on a tight timeline.

e.g.
```python
adherence_cases = self.get_valid_adherence_cases()
dose_taken_by_date = self.calculate_doses_taken_by_day(adherence_cases)
update["adherence_total_doses_taken"] = self.count_doses_taken(dose_taken_by_date)
# calculate 'aggregated_score_count_taken'
update["aggregated_score_count_taken"] = self.count_doses_taken(
	dose_taken_by_date,
	lte=adherence_schedule_date_start,
	gte=update["aggregated_score_date_calculated"]
)
```

NoahCarnahan:
(test failure looks real)

custom/enikshay/tasks.py
sravfeyn: Yeah, I was going back n forth between these two. I will actually change it to this. This is cleaner

sravfeyn:
Yeah, I wasn't able to reproduce this test failure on local. I am gonna look in a bit.

sravfeyn:
@NoahCarnahan ✅

custom/enikshay/tasks.py
sravfeyn: Offlined, made it to floor

NoahCarnahan (approved):

PR #14537, opened 2017-01-16
fix jquery upgrade bug in form display

snopoke:
http://manage.dimagi.com/default.asp?245714

`#` can't be used as a jquery selector


millerdev (approved):

millerdev:
Can't really tell how jquery is involved here, but taking your word that this fixes a crash.

snopoke:
@millerdev bootstrap uses the `href` as a jquery selector to find the tab that it's supposed to show.

PR #14528, opened 2017-01-15
fix link to data access restrictions

czue:
@orangejenny / anyone

PR #14543, opened 2017-01-16
ucr view permissions

calellowitz:
@sravfeyn @snopoke cc: @NoahCarnahan 
This models UCR view permissions off the ones currently used for other reports and fixes a broken link I found along the way

### DIFF
--- a/corehq/apps/locations/permissions.py
+++ b/corehq/apps/locations/permissions.py
@@ -95,7 +95,7 @@ LOCATION_ACCESS_DENIED = mark_safe(ugettext_lazy(
     "This project has restricted data access rules. Please contact your "
     "project administrator to be assigned access to data in this project. "
     'More information is available <a href="{link}">here</a>.'
-).format(link="https://wiki.commcarehq.org/display/commcarepublic/Data+Access+Restrictions"))
+).format(link="https://wiki.commcarehq.org/display/commcarepublic/Data+Access+and+User+Editing+Restrictions"))
 
 LOCATION_SAFE_TASTYPIE_RESOURCES = set()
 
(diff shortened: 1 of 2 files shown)


corehq/apps/userreports/reports/view.py
```diff
@@ -274,7 +274,11 @@ def post(self, request, *args, **kwargs):
             raise Http403()
 
     def has_permissions(self, domain, user):
-        return True
+        if domain is None:
+            return False
+        if not user.is_active:
+            return False
+        return user.can_view_report(domain, 'corehq.reports.DynamicReport{}'.format(self.report_config_id))
```
calellowitz: ucrs when listed in the permissions object have the form `corehq.reports.DynamicReport{id}` which comes from the way the report classes are generated (and the fact that we store access to a report by class)

NoahCarnahan (approved):
👍

corehq/apps/locations/permissions.py
```diff
@@ -95,7 +95,7 @@
     "This project has restricted data access rules. Please contact your "
     "project administrator to be assigned access to data in this project. "
     'More information is available <a href="{link}">here</a>.'
-).format(link="https://wiki.commcarehq.org/display/commcarepublic/Data+Access+Restrictions"))
+).format(link="https://wiki.commcarehq.org/display/commcarepublic/Data+Access+and+User+Editing+Restrictions"))
```
czue: already addresed in https://github.com/dimagi/commcare-hq/pull/14528/, though seems to merge cleanly.

corehq/apps/userreports/reports/view.py
czue: not blocking but maybe a helper function with a good doc string might help with future maintenance of this?

czue (approved):
:boot:

corehq/apps/userreports/reports/view.py
snopoke: @calellowitz would be good to follow up on this

PR #14547, opened 2017-01-17
Handle multiple default product plans

kaapstorm:
This fixes accounting migration [0047_ensure_default_product_plans.py](https://github.com/dimagi/commcare-hq/blob/2992da2646d61181c46cb36eba068be6eb37fc53/corehq/apps/accounting/migrations/0047_ensure_default_product_plans.py#L21-L21) when multiple default product plans are returned.

@nickpell could you confirm this logic is valid? It makes sense to me that we don't want multiple defaults, but I want to check with you that the default plan set by this change will be correct.

@calellowitz this PR should fix the Swiss deploy.

buddies @dannyroberts @mkangia

### DIFF
--- a/corehq/apps/accounting/bootstrap/utils.py
+++ b/corehq/apps/accounting/bootstrap/utils.py
@@ -17,6 +17,16 @@ FEATURE_TYPES = [
 
 
 def ensure_plans(config, dry_run, verbose, apps):
+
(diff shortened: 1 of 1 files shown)


nickpell:
I prefer https://github.com/dimagi/commcare-hq/pull/14559 because it's clear what code can be deleted after migrations are squashed.  Squashing in the medium-term seems like a good idea to reduce the complexity of ```ensure_plans```

PR #14550, opened 2017-01-17
Fix exception on rendering exceed message

mkangia:
Resolves: http://manage.dimagi.com/default.asp?245875

### DIFF
--- a/corehq/apps/export/views.py
+++ b/corehq/apps/export/views.py
@@ -1727,8 +1727,8 @@ class GenericDownloadNewExportMixin(object):
             count += get_export_size(instance, filters)
         if count > MAX_EXPORTABLE_ROWS:
             raise ExportAsyncException(
-                _("This export contains " + count + " rows. Please change the " +
-                "filters to be less than " + MAX_EXPORTABLE_ROWS + "rows.")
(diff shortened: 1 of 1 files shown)


corehq/apps/export/views.py
```diff
@@ -1727,8 +1727,8 @@ def _check_export_size(self, export_instances, filters):
             count += get_export_size(instance, filters)
         if count > MAX_EXPORTABLE_ROWS:
             raise ExportAsyncException(
-                _("This export contains " + count + " rows. Please change the " +
-                "filters to be less than " + MAX_EXPORTABLE_ROWS + "rows.")
+                _("This export contains {count} rows. Please change the filters to be less than {max} "
+                  "rows.".format(count=count, max=MAX_EXPORTABLE_ROWS))
```
czue: looks good but you should actually put the format call after the translation, since you want to inject the variables into the translated string.

czue (changes requested):
small change needed but then g2g

benrudolph:
https://github.com/dimagi/commcare-hq/pull/14551 same change here

mkangia:
Thank you @benrudolph. Closing this.

PR #14553, opened 2017-01-17
Allow drilldown reports to show up in report access control list

gcapalbo:
http://manage.dimagi.com/default.asp?245800

@proteusvacuum


gcapalbo:
also pinging @esoergel @snopoke @biyeun for fyi as they're listed as area owners for reporting

PR #14554, opened 2017-01-17
[WIP] Add RCH integration to pull data from RCH API and list out

mkangia:


dimagimon:
There are 5 errors:

* custom/rch/forms.py, line 6 - F401 'Fieldset' imported but unused
* custom/rch/models.py, line 139 - E226 missing whitespace around arithmetic operator
* custom/rch/models.py, line 187 - E226 missing whitespace around arithmetic operator
* custom/rch/utils.py, line 5 - E203 whitespace before ':'
* custom/rch/views.py, line 6 - F401 'AreaMapping' imported but unused

custom/rch/forms.py
```diff
@@ -0,0 +1,109 @@
...
+from django.utils.translation import ugettext_lazy
+
+from crispy_forms import layout as crispy
+from crispy_forms.layout import Layout, Fieldset, ButtonHolder, Submit
+
+from custom.rch.models import AreaMapping
+
+
+def get_state_names_choices():
+    states = AreaMapping.objects.order_by().values_list('stname', flat=True).distinct()
```
snopoke: what does `order_by()` do?

custom/rch/forms.py
```diff
@@ -0,0 +1,109 @@
...
+        options.add((state, state))
+    return tuple(options)
+
+
+def get_district_names_choices():
+    districts = AreaMapping.objects.order_by().values_list('dtname', flat=True).distinct()
+    options = set()
+    for district in districts:
+        options.add((district, district))
+    return tuple(options)
```
snopoke: you could do all of this in a helper function and just pass in the field you want to query

custom/rch/models.py
```diff
@@ -0,0 +1,212 @@
...
+    MDDS_DistrictID = models.PositiveSmallIntegerField(null=True)
+    PHC_Name = models.CharField(null=True, max_length=255)
+    PHC_ID = models.IntegerField(null=True)
+    Subcentre_Name = models.CharField(null=True, max_length=255)
+    Village_ID = models.IntegerField(null=True)
+    Mobile_no = models.BigIntegerField(null=True)
+    Religion = models.CharField(null=True, max_length=255)
+    State_Name = models.CharField(null=True, max_length=255)
+    StateID = models.PositiveSmallIntegerField(null=True)
+    Updated_On = models.DateTimeField(null=True)
```
snopoke: Is there a reason to make these camel case? Normally DB columns should be lower case (makes querying easier since you don't have to quote column names)

custom/rch/models.py
```diff
@@ -0,0 +1,212 @@
...
+class AreaMapping(models.Model):
+    stcode = models.IntegerField(null=False)
+    stname = models.CharField(max_length=255, null=False)
+    dtcode = models.IntegerField(null=False)
+    dtname = models.CharField(max_length=255, null=False)
+    pjcode = models.IntegerField(null=False)
+    pjname = models.CharField(max_length=255, null=False)
+    awcid = models.BigIntegerField(null=False)
+    awcname = models.CharField(max_length=255, null=False)
+    villcode = models.IntegerField(null=False)
```
snopoke: why `villcode` and not `village_code`? The second option is clearer and consistent with `village_name`

custom/rch/models.py
```diff
@@ -0,0 +1,212 @@
...
+    Infant2_Baby_Cried = models.CharField(null=True, max_length=255)
+    Infant2_Gender = models.CharField(null=True, max_length=255)
+    Subcentre_ID = models.IntegerField(null=True)
+    Village_Name = models.CharField(null=True, max_length=255)
+
+    def save(self, *args, **kwargs):
+        try:
+            if self.Delivery_Time:
+                datetime.datetime.strptime(self.Delivery_Time, '%H:%M:%S')
+        except ValueError:
```
snopoke: Why are these necessary? Doesn't the field do it's own validation? Custom validation should be done in the `clean` method: https://docs.djangoproject.com/en/1.8/ref/models/instances/#validating-objects

custom/rch/models.py
```diff
@@ -0,0 +1,212 @@
...
+
+        try:
+            if self.Discharge_Time:
+                datetime.datetime.strptime(self.Discharge_Time, '%H:%M:%S')
+        except ValueError:
+            self.Discharge_Time = None
+
+        if self.Maternal_Death:
+            self.Maternal_Death = 0
+            if self.Maternal_Death == 'Y' or self.Maternal_Death == 'y':
```
snopoke: This conversion should happen before the value gets assigned to the field.

custom/rch/models.py
```diff
@@ -0,0 +1,212 @@
...
+    @classmethod
+    def update_beneficiaries(cls):
+        date_str = str(datetime.date.fromordinal(datetime.date.today().toordinal()-1))
+        for state_id in STATE_DISTRICT_MAPPING:
+            for district_id in STATE_DISTRICT_MAPPING[state_id]:
+                records = fetch_mother_beneficiaries_records(date_str, date_str, state_id, district_id)
+                for record in records:
+                    rch_mother = cls()
+                    for prop in record:
+                        rch_mother.__setattr__(prop.keys()[0], prop.values()[0])
```
snopoke: rather use `setattr(rch_mother, key, value)`

custom/rch/models.py
```diff
@@ -0,0 +1,212 @@
...
+    def update_beneficiaries(cls):
+        date_str = str(datetime.date.fromordinal(datetime.date.today().toordinal()-1))
+        for state_id in STATE_DISTRICT_MAPPING:
+            for district_id in STATE_DISTRICT_MAPPING[state_id]:
+                records = fetch_child_beneficiaries_records(date_str, date_str, state_id, district_id)
+                for record in records:
+                    rch_child = cls()
+                    for prop in record:
+                        rch_child.__setattr__(prop.keys()[0], prop.values()[0])
+                    rch_child.save()
```
snopoke: seems very similar to method on `RCHMother`, could extract a helper function?

custom/rch/views.py
```diff
@@ -0,0 +1,88 @@
...
+        context['mother_beneficiaries'] = mother_beneficiaries.all()
+        context['child_beneficiaries'] = child_beneficiaries.all()
+
+        return context
+
+    def post(self, request, *args, **kwargs):
+        context = self.get_context_data()
+        mother_beneficiaries = RCHMother.objects
+        child_beneficiaries = RCHChild.objects
+        state = request.POST['state']
```
snopoke: should use `POST.get` to avoid key errors if the parameter is missing

requirements/requirements.txt
```diff
@@ -97,3 +97,4 @@ csiphash==0.0.4
 json-delta==2.0
 email_validator==1.0.1
 dnspython==1.13.0
+git+https://github.com/mvantellingen/python-zeep.git@46653fac2a92a99bc93d20c772cd7d797c301105
```
snopoke: any reason not to use the version on pypi: https://pypi.python.org/pypi/zeep?

settings.py
```diff
@@ -370,6 +371,10 @@
     'custom.enikshay.nikshay_datamigration',
 )
 
+RCH_APPS = (
+    'custom.rch',
+)
```
snopoke: you don't need this if you've already added it to `HQ_APPS`

custom/rch/models.py
```diff
@@ -0,0 +1,212 @@
...
+    Name_Child = models.CharField(null=True, max_length=255)
+    Name_Father = models.CharField(null=True, max_length=255)
+    Name_Mother = models.CharField(null=True, max_length=255)
+    Received_AllVaccines = models.CharField(null=True, max_length=255)
+    SubCentre_ID = models.SmallIntegerField(null=True)
+    VILLAGE_Name = models.CharField(null=True, max_length=255)
+    Weight = models.FloatField(null=True)
+
+    @classmethod
+    def update_beneficiaries(cls):
```
snopoke: Where do these get called?

custom/rch/views.py
```diff
@@ -0,0 +1,88 @@
+from django.utils.translation import ugettext_noop
+from django.views.generic import TemplateView
+from django.core import serializers
+import json
+from custom.rch.forms import BeneficiariesFilterForm
+from custom.rch.models import RCHMother, AreaMapping, RCHChild
+
+
+class BeneficariesList(TemplateView):
```
snopoke: Are these views just for debugging purposes?

snopoke (changes requested):

custom/rch/forms.py
mkangia: I thought this was needed here. Cant remember why. Removing it.

custom/rch/models.py
mkangia: The field names were kept same as the property names received intentionally so that https://github.com/dimagi/commcare-hq/pull/14554#discussion-diff-96423237R146 works. Ideally we should keep this camel case but then we would need a one to one mapping for these with the properties received from API. It would cost us some computation.

custom/rch/models.py
mkangia: This is to take care if Delivery_Time is not in correct format. There was once the API returned "%H:%M" (AS-IS) in this field. So this takes care of it so that save call does not fail. I don't think we want the save to fail hence not validating it but just ignoring it if not in required format. Also read that clean method is not called on save(). Should it be called explicitly then?

custom/rch/models.py
mkangia: These need to get bind to task decorators to be run everyday only between 6-8PM IST. Need to figure that out.

custom/rch/models.py
mkangia: I feel models would be right place and not helper function since these are actually adding records. Moved to parent class if that sounds right.

custom/rch/models.py
mkangia: Its just to keep it same as the fields received from the API as shared [above](https://github.com/dimagi/commcare-hq/pull/14554#discussion-diff-96421945R32)

custom/rch/views.py
mkangia: Yup! just views to list down Beneficiaries to show what data has been received from API.

requirements/requirements.txt
mkangia: This took quite a while, to figure out the version that would work.  This commit version is just above one stable version and this particular [commit](https://github.com/mvantellingen/python-zeep/commit/46653fac2a92a99bc93d20c772cd7d797c301105) is a fix for empty tags. The recent version failed due to some parsing errors.

custom/rch/forms.py
```diff
@@ -0,0 +1,77 @@
+from crispy_forms.helper import FormHelper
+from django import forms
+from django.utils.translation import ugettext_lazy
+
+from crispy_forms import layout as crispy
+from crispy_forms.layout import Layout, Fieldset, ButtonHolder, Submit
```
dimagimon: F401 'Fieldset' imported but unused

custom/rch/utils.py
```diff
@@ -0,0 +1,38 @@
+from zeep.client import Client
+
+
+def etree_to_dict(t):
+    return {t.tag : map(etree_to_dict, t.iterchildren()) or t.text}
```
dimagimon: E203 whitespace before ':'

custom/rch/views.py
```diff
@@ -0,0 +1,97 @@
+import json
+
+from django.utils.translation import ugettext_noop
+from django.views.generic import TemplateView
+from django.core import serializers
+from django.utils.decorators import method_decorator
+
+from corehq.apps.domain.decorators import require_superuser
+from custom.rch.forms import BeneficiariesFilterForm
+from custom.rch.models import RCHMother, AreaMapping, RCHChild
```
dimagimon: F401 'AreaMapping' imported but unused

custom/rch/utils.py
```diff
@@ -0,0 +1,38 @@
+from zeep.client import Client
+
+
+def etree_to_dict(t):
+    return {t.tag: map(etree_to_dict, t.iterchildren()) or t.text}
+
+
+PROJECTID = ''  # From some secure place
+ID = ''  # From some secure place
+PASSWORD = ''  # From some secure place
```
mkangia: @snopoke How should this be done? Adding secure ids and password.

custom/rch/models.py
```diff
@@ -0,0 +1,207 @@
...
+    Name_Child = models.CharField(null=True, max_length=255)
+    Name_Father = models.CharField(null=True, max_length=255)
+    Name_Mother = models.CharField(null=True, max_length=255)
+    Received_AllVaccines = models.CharField(null=True, max_length=255)
+    SubCentre_ID = models.SmallIntegerField(null=True)
+    VILLAGE_Name = models.CharField(null=True, max_length=255)
+    Weight = models.FloatField(null=True)
+
+    @classmethod
+    def update_beneficiaries(cls, beneficiary_type=CHILD_DATA_TYPE):
```
snopoke: probably don't need the `beneficiary_type` kwarg since I doubt you'll ever call it with something different.

custom/rch/models.py
snopoke: :+1:

snopoke: :+1:

custom/rch/models.py
```diff
@@ -0,0 +1,207 @@
...
+    @classmethod
+    def update_beneficiaries(cls, beneficiary_type):
+        date_str = str(datetime.date.fromordinal(datetime.date.today().toordinal() - 1))
+        for state_id in STATE_DISTRICT_MAPPING:
+            for district_id in STATE_DISTRICT_MAPPING[state_id]:
+                records = fetch_beneficiaries_records(date_str, date_str, state_id, beneficiary_type, district_id)
+                for record in records:
+                    rch_beneficiary = cls()
+                    for prop in record:
+                        setattr(rch_beneficiary, prop.keys()[0], prop.values()[0])
```
snopoke: To make this a bit safer you could get all the fields on the model and make sure the props are a valid field. You could also do validation with it if you want:
```python
valid_fields = {field.name: field for field in cls._meta.fields}
field = prop.keys()[0]
if field not in valid_fields:
    # skip or do something
if isinstance(valid_fields[field], models.TimeField):
    # check that it is correct format
```

custom/rch/models.py
snopoke: I think it would be better to put this in a separate function which gets called in `update_beneficiaries` prior to calling save.

custom/rch/emails.py
```diff
@@ -0,0 +1,26 @@
+from django.conf import settings
+from corehq.apps.hqwebapp.tasks import send_mail_async
+
+
+def notify_parsing_error(for_date, state_id, beneficiary_type, district_id):
+    subject = "[RCHServiceParsingError] [{for_date}] Issue with parsing RCH response".format(for_date=for_date)
+    content = ("Unexpected response received for date:{for_date}, state_id:{s_id}, beneficiary_type:{bt}, "
+               "district_id: {d_id}".format(
+                for_date=for_date, s_id=state_id, bt=beneficiary_type, d_id=district_id))
```
dimagimon: E121 continuation line under-indented for hanging indent

custom/rch/emails.py
```diff
@@ -0,0 +1,26 @@
...
+    send_mail_async.delay(subject, content, settings.DEFAULT_FROM_EMAIL, ['mkangia@dimagi.com'])
+
+
+def notify_insecure_access_response(for_date, state_id, beneficiary_type, district_id):
+    subject = "[RCHServiceInsecureResponse] [{for_date}] RCH Insecure access response received".format(
+        for_date=for_date)
+    content = ("Insecure access response received for date:{for_date}, state_id:{s_id}, beneficiary_type:{bt}, "
+               "district_id: {d_id}. Probably no updates in RCH for the requested date. Must check this if "
+               "consecutively happening.".format(
+                for_date=for_date, s_id=state_id, bt=beneficiary_type, d_id=district_id))
```
dimagimon: E121 continuation line under-indented for hanging indent

custom/rch/emails.py
```diff
@@ -0,0 +1,26 @@
+from django.conf import settings
+from corehq.apps.hqwebapp.tasks import send_mail_async
+
+
+def notify_parsing_error(for_date, state_id, beneficiary_type, district_id):
+    subject = "[RCHServiceParsingError] [{for_date}] Issue with parsing RCH response".format(for_date=for_date)
+    content = ("Unexpected response received for date:{for_date}, state_id:{s_id}, beneficiary_type:{bt}, "
+               "district_id: {d_id}".format(
+                    for_date=for_date, s_id=state_id, bt=beneficiary_type, d_id=district_id))
```
dimagimon: E126 continuation line over-indented for hanging indent

custom/rch/emails.py
```diff
@@ -0,0 +1,26 @@
...
+    send_mail_async.delay(subject, content, settings.DEFAULT_FROM_EMAIL, ['mkangia@dimagi.com'])
+
+
+def notify_insecure_access_response(for_date, state_id, beneficiary_type, district_id):
+    subject = "[RCHServiceInsecureResponse] [{for_date}] RCH Insecure access response received".format(
+        for_date=for_date)
+    content = ("Insecure access response received for date:{for_date}, state_id:{s_id}, beneficiary_type:{bt}, "
+               "district_id: {d_id}. Probably no updates in RCH for the requested date. Must check this if "
+               "consecutively happening.".format(
+                    for_date=for_date, s_id=state_id, bt=beneficiary_type, d_id=district_id))
```
dimagimon: E126 continuation line over-indented for hanging indent

custom/rch/emails.py
```diff
@@ -0,0 +1,26 @@
...
+               "district_id: {d_id}. Probably no updates in RCH for the requested date. Must check this if "
+               "consecutively happening.".format(for_date=for_date, s_id=state_id, bt=beneficiary_type,
+                                                 d_id=district_id))
+    send_mail_async.delay(subject, content, settings.DEFAULT_FROM_EMAIL, ['mkangia@dimagi.com'])
+
+
+def notify_error_in_service(message, for_date):
+    subject = "[RCHServiceDown] [{for_date}] Unable to access RCH Service".format(for_date=for_date)
+    content = "Error in RCH Service: {message}".format(message=message)
+    send_mail_async.delay(subject, content, settings.DEFAULT_FROM_EMAIL, ['mkangia@dimagi.com'])
```
snopoke: Can you not just use `notify_exception` for these?
```python
notify_exception(None, 'RCH Insecure access response received', {'date': for_date, 'state_id': state_id, 'beneficiary_type': bt, 'district_id': d_id})
```

mkangia: Added a separate notification and only to me as of now because:
1. Not useful to all devs since its specific to RCH
2. Should not get missed in exception emails since this should be looked at asap considering we can fetch details only for last 7 days.

custom/rch/views.py
```diff
@@ -0,0 +1,93 @@
...
+from django.views.generic import TemplateView
+from django.core import serializers
+from django.utils.decorators import method_decorator
+
+from corehq.apps.domain.decorators import require_superuser
+from custom.rch.forms import BeneficiariesFilterForm
+from custom.rch.models import RCHMother, RCHChild
+
+
+class BeneficariesList(TemplateView):
```
mkangia: There would be quite some customisation to this. So would be better to keep it outside admin. Plus might be used by some non-admins as well, later.

custom/rch/models.py
mkangia: Agreed. Added this to match the signature in parent class which is called later. Should that be a good practice?

snopoke:
@mkangia let me know when I should review this

mkangia:
@snopoke running late on a couple of changes. Also the deploy for this has been delayed due to permission issues for deploy. Getting that aligned as well.

custom/rch/models.py
```diff
@@ -0,0 +1,240 @@
...
+
+    class Meta:
+        abstract = True
+
+    @classmethod
+    def update_beneficiaries(cls, beneficiary_type, days_before=1):
+        date_str = str(datetime.date.fromordinal(datetime.date.today().toordinal() - days_before))
+        for state_id in STATE_DISTRICT_MAPPING:
+            for district_id in STATE_DISTRICT_MAPPING[state_id]:
+                records = fetch_beneficiaries_records(date_str, date_str, state_id, beneficiary_type, district_id)
```
snopoke: are `from_date` and `to_date` meant to be the same?

custom/rch/views.py
```diff
@@ -0,0 +1,108 @@
...
+        elif beneficiaries_in == 'cas':
+            self.get_cas_records()
+
+        context['mother_beneficiaries_total'] = self.mother_beneficiaries.count()
+        context['child_beneficiaries_total'] = self.child_beneficiaries.count()
+        context['mother_beneficiaries'] = self.mother_beneficiaries.order_by()[:20]
+        context['child_beneficiaries'] = self.child_beneficiaries.order_by()[:20]
+
+
+        return context
```
dimagimon: E303 too many blank lines (2)

custom/rch/models.py
mkangia: yup. Could pass just one day as well. Will update

custom/rch/models.py
```diff
@@ -0,0 +1,146 @@
...
+    details_doc_id = models.CharField(max_length=255, null=True)
+    district_id = models.PositiveSmallIntegerField(null=True)
+    state_id = models.PositiveSmallIntegerField(null=True)
+    village_id = models.IntegerField(null=True)
+    village_name = models.CharField(null=True, max_length=255)
+    name = models.CharField(null=True, max_length=255)
+    aadhar_num = models.BigIntegerField(null=True)
+    dob = models.DateTimeField(null=True)
+    rch_id = models.BigIntegerField(null=True)
+    doc_type = models.CharField(max_length=1, choices=[('0', 'mother'), ('1','child')], null=False)
```
dimagimon: E231 missing whitespace after ','

mkangia:
@snopoke 
The process of taking data from RCH application needs to be reconsidered since we can't pull data from RCH like we have been assuming till now. :(
This came from our meeting at UIDAI for using Aadhar number as the common key for beneficiary matching. This might be needed later. Will revive it then.

i am going to close this pull request without merging it.

snopoke:
OK, I'm keen to chat about this more to understand what's changed.

PR #14555, opened 2017-01-17
delay application of select2

snopoke:
http://manage.dimagi.com/default.asp?244947

Was able to reproduce locally and found that if I ran `$('#report_filter_group').select2();` in the console after the filters had re-loaded then the group options were correctly rendered.

Adding the delay fixed it for me locally, not sure if this is the best solution @dimagi/js-team?

### DIFF
--- a/corehq/apps/reports/templates/reports/filters/multi_option.html
+++ b/corehq/apps/reports/templates/reports/filters/multi_option.html
@@ -26,54 +26,53 @@
             select_params: {{ select.options|JSON }},
             current_selection: ko.observableArray({{ select.selected|JSON }}),
         });
-    });
-    {% if not endpoint %}
-        _.delay(function() {
+
+        {% if not endpoint %}
             $('#{{ css_id }}').select2();
-        });
-    {% else %}
-        {% comment %}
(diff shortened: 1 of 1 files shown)


corehq/apps/reports/templates/reports/filters/multi_option.html
```diff
@@ -28,7 +28,9 @@
         });
     });
     {% if not endpoint %}
-        $('#{{ css_id }}').select2();
+        _.delay(function() {
+            $('#{{ css_id }}').select2();
```
millerdev: When does this bit of JS get executed? Is it before or after DOM ready or unknown?

orangejenny:
I think moving the `select2` call into the document ready handler directly above it will also fix this problem (tried it out locally) and makes more sense when you're reading the code.

snopoke:
@orangejenny thanks, I hadn't noticed that. Updated to move the `select2` calls inside the document ready handler.

millerdev (approved):

esoergel (approved):
Much easier to read like this: https://github.com/dimagi/commcare-hq/pull/14555/files/?w=1
Killing build.

PR #14557, opened 2017-01-17
re add phone number a/b test

calellowitz:
@NoahCarnahan @biyeun cc: @sravfeyn 
This just reverts the pr that took out the phone number a/b test since sales wants it back in now.

### DIFF
--- a/corehq/apps/analytics/ab_tests.py
+++ b/corehq/apps/analytics/ab_tests.py
@@ -50,3 +50,13 @@ LIVE_PREVIEW = ABTestConfig(
     'live_preview_dec2016',
     (LIVE_PREVIEW_ENABLED, LIVE_PREVIEW_DISABLED)
 )
+
+NEW_USER_NUMBER_OPTION_SHOW_NUM = 'show_number'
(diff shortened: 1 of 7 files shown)


corehq/apps/analytics/ab_tests.py
```diff
@@ -50,3 +50,13 @@ def context(self):
...
     (LIVE_PREVIEW_ENABLED, LIVE_PREVIEW_DISABLED)
 )
+
+NEW_USER_NUMBER_OPTION_SHOW_NUM = 'show_number'
+NEW_USER_NUMBER_OPTION_HIDE_NUM = 'hide_number'
+
+
+NEW_USER_NUMBER = ABTestConfig(
+    'New User Phone Number',
+    'new_phone_aug2016',
```
NoahCarnahan: I wonder if this id should be different? Not sure if product/sales wants this to appear as the same test or a new one...

NoahCarnahan (approved):
Looks good to me, just one comment (that might not require any action)

sravfeyn:
LGTM pending noah's comment

calellowitz:
@NoahCarnahan I think that slug is only used in the cookie (so shouldnt matter), but just in case i updated it

PR #14561, opened 2017-01-17
Always load role choices, even on POST

esoergel:
This was causing it to fail locally when assigning a role.  On POST, the choices list was never populated, so the selection of course wasn't in the list of options.
@calellowitz @czue

### DIFF
--- a/corehq/apps/users/views/__init__.py
+++ b/corehq/apps/users/views/__init__.py
@@ -207,14 +207,16 @@ class BaseEditUserView(BaseUserSettingsView):
             raise NotImplementedError("You must specify a form to update the user!")
 
         if self.request.method == "POST" and self.request.POST['form_type'] == "update-user":
-            return self.user_update_form_class(data=self.request.POST)
+            form = self.user_update_form_class(data=self.request.POST)
+        else:
+            form = self.user_update_form_class()
+            form.initialize_form(domain=self.request.domain, existing_user=self.editable_user)
 
-        form = self.user_update_form_class()
-        form.initialize_form(domain=self.request.domain, existing_user=self.editable_user)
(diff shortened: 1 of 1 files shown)


corehq/apps/users/views/__init__.py
```diff
@@ -207,14 +207,16 @@ def form_user_update(self):
...
+        else:
+            form = self.user_update_form_class()
+            form.initialize_form(domain=self.request.domain, existing_user=self.editable_user)
 
-        form = self.user_update_form_class()
-        form.initialize_form(domain=self.request.domain, existing_user=self.editable_user)
         if self.can_change_user_roles:
             form.load_roles(current_role=self.existing_role, role_choices=self.user_role_choices)
         else:
             del form.fields['role']
```
esoergel: This used to be applied in a subclass, so it was applied regardless of where the function returned
https://github.com/dimagi/commcare-hq/pull/14534/files#diff-b88c82ab45221007abc47af5a1c62207L281
https://github.com/dimagi/commcare-hq/pull/14534/files#diff-df796c59286ba51267f9d91ce003127aL269

esoergel:
@NoahCarnahan buddy ping

calellowitz (approved):

calellowitz:
merging now because the master build appears broken and we want to get this pr deployed

czue:
nice, thanks for the fix. for my own education what was the user-facing bug here? just want to know what i missed in my testing.

esoergel:
@czue I'd load the mobile worker page, see the options in the role dropdown, select one, click save, and the save would fail with a message about "`user-role:1231231j231` is not one of the available choices" (I think that's the default message for a django `forms.ChoiceField` validation)

PR #14562, opened 2017-01-17
making predictably random toggles return False in testing

calellowitz:
@esoergel cc: @sravfeyn @snopoke

### DIFF
--- a/corehq/toggles.py
+++ b/corehq/toggles.py
@@ -3,6 +3,8 @@ from functools import wraps
 import hashlib
 from django.http import Http404
 import math
+
+from django.conf import settings
 from corehq.util.quickcache import quickcache
 from toggle.shortcuts import toggle_enabled, set_toggle
 
@@ -137,7 +139,9 @@ class PredictablyRandomToggle(StaticToggle):
         return '{}:{}:{}'.format(self.namespaces, self.slug, item)
 
     def enabled(self, item, **kwargs):
-        if item in self.always_disabled:
+        if settings.UNIT_TESTING:
+            return False
+        elif item in self.always_disabled:
             return False
         return (
             (item and deterministic_random(self._get_identifier(item)) < self.randomness)


esoergel (approved):

benrudolph:
i'm not convinced this is a good strategy. a test failure for when a toggle turns on could be useful to know and prevent unwanted errors when the toggle is turned up to full blast. what were the reasons for going with this over fixing the tests?

snopoke:
In particular, if you were trying to fix anything related for form validation (which is currently using a random toggle). The old method is going to be removed at some stage.

calellowitz:
This was a response to a broken master build, where changing the sensitivity of a toggle (in this case cloudcare) had broken a set of tests designed to test both paths, because the dummy test domain switched from hashing to the flag being off to the flag being on. It felt wrong that our tests would change with toggle sensitivity, and that it was impossible to test the both paths, as these tests did, without searching for a domain name that hashed correctly, (which would have been fine for cloudcare but if it were a toggle on 80% or all new domains or something else like that would have been much more difficult).
In this setup, you can explicitly turn the toggle on, which tests that interact with the toggle should do anyway, but in the state where both paths exist in the wild, as they do almost always with a random toggle, the ability to test both at will seemed better than playing the hash guessing game with domain names, and allows us to write tests that are stable.

calellowitz:
For reference the broken tests were https://github.com/dimagi/commcare-hq/blob/master/corehq/apps/hqwebapp/tests/test_default_landing_page.py

benrudolph:
Good to know, I do agree that it is wrong that our tests change with different toggle sensitivity, but I would rather them break at 20% than at 100%. The way we can properly write tests to test both the flag on and off is to use mocking (`flag_enabled` decorator is useful for this). In this case it wasn't a big deal because the tests didn't catch a real bug, but if they had and the build had broken at 70% i would've known that a) some tests need to be re written and b) this will probably be broken for more users if those tests caught a real bug. In this scenario those tests wouldn't break until I was done rolling out the feature.

> In this setup, you can explicitly turn the toggle on

How does this setup more easily let you turn a toggle on? Isn't this saying it's always disabled unless you do something like mocking?

esoergel:
It would definitely be nice to catch potential bugs in tests, but I'm not convinced the way this was working was the way to go about it.  It's surprising to have a feature flag enabled without explicitly enabling it, since the usual default is off.  It turned on arbitrarily for a percentage of domain names, so some tests had the flag enabled and some didn't, with no way to tell which was which.  Note also that this is consistent, so the same N% of tests would have the flag each test run.

Ideally we'd run the full test suite twice, once with and once without the flag, but doubling our run times for each random toggle is clearly a non-starter.  We could write something to turn each flag on for the full run 50% of the time, so we at least get full test coverage pretty quickly.  However, any degree of randomness in tests can lead to cryptic failures that aren't consistently reproducible, so I'd prefer to avoid it whenever possible.

I don't think there's a perfect solution to this problem, unfortunately.

benrudolph:
The way we are current going about it is definitely not ideal (and there's a strong chance I've just gotten used to that) and I agree in a perfect world we would run each test twice with the flag toggled. However this solution I think just delays the inevitable. These test failures aren't crytic/random in the way that accounting tests used to be 😖 (see https://github.com/dimagi/commcare-hq/pull/8725). These tests will predictably fail. That test would've failed on my PR, and I'd argue that would have been useful information for me. Now the same test is going to fail when I take out the toggle. I'd almost argue it makes more sense to return True since the intent of every predictably random toggle is to go to 100%. All that said I don't feel strongly enough to push for this to be reverted; I see the argument in both ways

PR #14563, opened 2017-01-17
Flag enikshay custom reports as location-safe

esoergel:
@czue
http://manage.dimagi.com/default.asp?245889
Following up there, but the short of it is that this enables access to those reports by restricted users, but doesn't do any actual checking to prevent inappropriate data access.  If these reports are built with that in mind already, we should be able to just merge this in, otherwise they'll need more work.

### DIFF
--- a/custom/enikshay/reports/case_finding.py
+++ b/custom/enikshay/reports/case_finding.py
@@ -1,5 +1,6 @@
 # coding=utf-8
 
+from corehq.apps.locations.permissions import location_safe
 from corehq.apps.reports.datatables import DataTablesHeader, DataTablesColumn
 from custom.enikshay.reports.const import AGE_RANGES, PATIENT_TYPES
 from custom.enikshay.reports.generic import EnikshayReport, EnikshayMultiReport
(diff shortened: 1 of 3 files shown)


czue (approved):
thanks! @kkrampa can you confirm whether the reports are already properly handling location restrictions?

kkrampa:
I reused UCR's filters code so restrictions should be the same as in UCR.

czue:
ok awesome, then I think this is g2g. build failed on something that doesn't look related but is weird. kicked.
```
FAIL: test_mobile_worker_landing_page('webapps_role', 'cloudcare_main', ['']) (corehq.apps.hqwebapp.tests.test_default_landing_page.TestDefaultLandingPages)
  File "/mnt/commcare-hq-ro/corehq/util/test_utils.py", line 331, in test
    return test_func(self, *args)
  File "/mnt/commcare-hq-ro/corehq/apps/hqwebapp/tests/test_default_landing_page.py", line 124, in test_mobile_worker_landing_page
    response.request['PATH_INFO'])
AssertionError: u'/a/temerant/cloudcare/apps/' != '/a/temerant/cloudcare/apps/v2/'

```

PR #14567, opened 2017-01-18
don't wipe domain membership

snopoke:
http://manage.dimagi.com/default.asp?245969

@benrudolph @czue @calellowitz


czue (approved):
looks good. some cobwebs in this code. :spider_web:

benrudolph:
nice find

PR #14569, opened 2017-01-18
bump to 100

benrudolph:
@wpride @czue the number of requests are holding constant... so i don't think think this will matter much. the real test will be app preview

cc: @emord

### DIFF
--- a/corehq/toggles.py
+++ b/corehq/toggles.py
@@ -731,7 +731,7 @@ USE_FORMPLAYER_FRONTEND = PredictablyRandomToggle(
     'Use New CloudCare',
     TAG_PRODUCT_PATH,
     [NAMESPACE_DOMAIN],
-    randomness=0.7,
+    randomness=1.0,
     always_disabled=[
         'hsph-betterbirth',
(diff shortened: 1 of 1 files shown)


corehq/toggles.py
```diff
@@ -731,7 +731,7 @@ def _commtrackify(domain_name, toggle_is_enabled):
     'Use New CloudCare',
     TAG_PRODUCT_PATH,
     [NAMESPACE_DOMAIN],
-    randomness=0.7,
+    randomness=1.0,
```
czue: nit but you could just make this a regular toggle now.

czue (approved):

corehq/toggles.py
benrudolph: ah, i'll do that. my thinking was it'd be easy to turn down, but i guess it still takes a revert and deploy. would be sweet if we could turn the randomness up in the ui...

czue:
PID.

benrudolph:
actually going to merge for now because why not and then i'll then i'll follow up

PR #14571, opened 2017-01-18
Fix case tiles

wpride:
Looks like this got dropped here https://github.com/dimagi/commcare-hq/commit/2a2d7308054822372b60e73686efe18198e6c528

Will (hopefully) fix http://manage.dimagi.com/default.asp?245909

cc @benrudolph


benrudolph:
oops sorry. that was broken for a long time haha

benrudolph (approved):

wpride:
Indeed, I think no one knew or cared that this existed until now

PR #14572, opened 2017-01-18
management command to restore domain membership

snopoke:
@czue @benrudolph @calellowitz

### DIFF
--- a/corehq/apps/users/management/commands/restore_domain_membership.py
+++ b/corehq/apps/users/management/commands/restore_domain_membership.py
@@ -1,0 +1,87 @@
+from couchdbkit import ResourceNotFound
+from django.core.management.base import BaseCommand, CommandError
+
+from corehq.apps.change_feed.consumer.feed import KafkaChangeFeed
+from corehq.apps.change_feed.topics import COMMCARE_USER, get_multi_topic_offset, \
+    get_multi_topic_first_available_offsets
+from corehq.apps.users.models import DomainMembership, CommCareUser
+
+defaults = DomainMembership().to_json()
+del defaults['domain']
+extra_keys = ['first_name', 'last_name', 'user_data']
+
+
+class Command(BaseCommand):
+    args = "since_kafka_offset"
+    help = "See http://manage.dimagi.com/default.asp?245969 for context"
+
+    def add_arguments(self, parser):
+        parser.add_argument('--print-kafka-offsets', action='store_true',
+                            help='Just print the kafka offsets')
+
+    def handle(self, *args, **options):
+        if options['print_kafka_offsets']:
+            end = get_multi_topic_offset([COMMCARE_USER])[COMMCARE_USER]
+            start = get_multi_topic_first_available_offsets([COMMCARE_USER])[COMMCARE_USER]
+            print "\nKakfa topic offset range: {} - {}".format(start, end)
+            return
+
+        if len(args) == 0:
+            raise CommandError("Usage: python manage.py resync_location_user_data %s" % self.args)
+
+        since = args[0]
+        change_feed = KafkaChangeFeed(topics=[COMMCARE_USER], group_id='user-repair')
+        for change in change_feed.iter_changes(since=since, forever=False):
+            if change.deleted:
+                continue
+
+            user = change.get_document()
+            if user_looks_ok(user):
+                continue
+
+            restore_domain_membership(user)
+
+
+def user_looks_ok(user):
+    if 'commcare_project' not in user.user_data:
+        return True
+
+    domain_membership = user.domain_membership
+    if any(key in domain_membership for key in extra_keys):
+        # membership has not been reset
+        return True
+
+    for key, default_value in defaults.items():
+        if domain_membership[key] != default_value:
+            # domain membership has been updated
+            return True
+
+    return False
+
+
+def restore_domain_membership(user):
+    doc_id = user._id
+    db = CommCareUser.get_db()
+    revisions = get_doc_revisions(db, doc_id)
+    for rev in revisions[1:]:
+        doc = get_doc_rev(db, doc_id, rev)
+        prev_user = CommCareUser.wrap(doc)
+        if user_looks_ok(prev_user):
+            prev_domain_membership = prev_user.domain_membership
+            user.domain_membership = prev_domain_membership
+            user.save()
+            return
+
+
+def get_doc_revisions(db, doc_id):
+    res = db.get(doc_id, revs=True)
+    start = res['_revisions']['start']
+    ids = res['_revisions']['ids']
+    return ["{}-{}".format(start-i, rev) for i, rev in enumerate(ids)]
+
+
+def get_doc_rev(db, doc_id, rev):
+    try:
+        return db.get(doc_id, rev=rev)
+    except ResourceNotFound:
+        return None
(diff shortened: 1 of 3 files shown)


corehq/apps/users/management/commands/restore_domain_membership.py
```diff
@@ -0,0 +1,87 @@
...
+        if options['print_kafka_offsets']:
+            end = get_multi_topic_offset([COMMCARE_USER])[COMMCARE_USER]
+            start = get_multi_topic_first_available_offsets([COMMCARE_USER])[COMMCARE_USER]
+            print "\nKakfa topic offset range: {} - {}".format(start, end)
+            return
+
+        if len(args) == 0:
+            raise CommandError("Usage: python manage.py resync_location_user_data %s" % self.args)
+
+        since = args[0]
```
snopoke: not sure how to determine the `since` value since we don't store the date in the change (should fix that)

corehq/apps/users/management/commands/restore_domain_membership.py
```diff
@@ -0,0 +1,87 @@
...
+            user.domain_membership = prev_domain_membership
+            user.save()
+            return
+
+
+def get_doc_revisions(db, doc_id):
+    res = db.get(doc_id, revs=True)
+    start = res['_revisions']['start']
+    ids = res['_revisions']['ids']
+    return ["{}-{}".format(start-i, rev) for i, rev in enumerate(ids)]
```
dimagimon: E226 missing whitespace around arithmetic operator

corehq/apps/users/management/commands/restore_domain_membership.py
```diff
@@ -0,0 +1,87 @@
...
+                continue
+
+            user = change.get_document()
+            if user_looks_ok(user):
+                continue
+
+            restore_domain_membership(user)
+
+
+def user_looks_ok(user):
```
snopoke: This will fail for users created by the API or by bulk import (unless they have a role or location assigned) meaning that we could end up checking a lot of users we don't have to.

snopoke (commented):
not 100% sold on the approach... open to other ideas

corehq/apps/users/management/commands/restore_domain_membership.py
snopoke: binary search for first change where `'commcare_project' in user.user_data`

corehq/apps/users/management/commands/restore_domain_membership.py
```diff
@@ -0,0 +1,227 @@
...
+    def search(self):
+        if self.match_at_index(self.min):
+            return self.min
+        if not self.match_at_index(self.max):
+            return
+
+        min_index = self.min + 1  # since we've already checked min
+        max_index = self.max - 1  # since we've already checked max
+        while max_index >= min_index:
+            midpoint = (min_index + max_index)//2
```
dimagimon: E226 missing whitespace around arithmetic operator

corehq/apps/users/management/commands/restore_domain_membership.py
```diff
@@ -0,0 +1,227 @@
...
+            if self._match_at_index(midpoint):
+                max_index = midpoint-1  # search down
```
dimagimon: E226 missing whitespace around arithmetic operator

corehq/apps/users/management/commands/restore_domain_membership.py
```diff
@@ -0,0 +1,227 @@
...
+            else:
+                min_index = midpoint+1  # search up
```
dimagimon: E226 missing whitespace around arithmetic operator

corehq/apps/users/management/commands/restore_domain_membership.py
```diff
@@ -0,0 +1,227 @@
...
+            return
+
+    print 'Unable to fix user: {} ({})'.format(user.domain, doc_id)
+
+
+def get_doc_revisions(db, doc_id):
+    res = db.get(doc_id, revs=True)
+    start = res['_revisions']['start']
+    ids = res['_revisions']['ids']
+    return ["{}-{}".format(start-i, rev) for i, rev in enumerate(ids)]
```
dimagimon: E226 missing whitespace around arithmetic operator

corehq/apps/users/management/commands/restore_domain_membership.py
dimagimon: E226 missing whitespace around arithmetic operator

corehq/apps/users/management/commands/restore_domain_membership.py
dimagimon: E226 missing whitespace around arithmetic operator

corehq/apps/users/management/commands/restore_domain_membership.py
dimagimon: E226 missing whitespace around arithmetic operator

snopoke:
OK, think this is GTG

corehq/apps/users/management/commands/restore_domain_membership.py
```diff
@@ -0,0 +1,227 @@
...
+        min_index = self.min + 1  # since we've already checked min
+        max_index = self.max - 1  # since we've already checked max
+        while max_index >= min_index:
+            midpoint = (min_index + max_index) // 2
+            if self._match_at_index(midpoint):
+                max_index = midpoint - 1  # search down
+            else:
+                min_index = midpoint + 1  # search up
+
+        return self._min_matching_index
```
benrudolph: nice bin search

benrudolph (approved):
i tried thinking of all the cases where we could accidentally revert a user that we weren't supposed to. the only thing i could think of is if someone purposefully changes the membership back to the default settings. is that handled in this? that said, not sure if it's worth worrying about

snopoke:
@benrudolph yes, that is a possibility:
* user created via bulk import or the API
* user has role / location set
* user has role / location unset

Unfortunately there is no way to differentiate between that case and the bug. I can add one other check which will verify that the location fields in the user match those in the domain membership which will leave only the role being set / unset as an edge case.

@czue any thoughts on whether we should go ahead with this?

czue:
this looks great and really thorough. have we run `--check` yet? definitely seems worth running to me. I think the end result of running this is definitely going to result in much better state than not running it even if we do revert some data.

Maybe we could work with flos to email all domain admins with the list of affected users after we run? That way if we do revert something that we shouldn't have at least the project admins are aware.

benrudolph:
cool, agree that we should still run this. i'm signed off on merge when you're ready

benrudolph:
tests passed locally, mergin

corehq/apps/users/management/commands/restore_domain_membership.py
```diff
@@ -0,0 +1,236 @@
...
+            prev_domain_membership = prev_user.domain_membership
+            user.domain_membership = prev_domain_membership
+            user.save()
+            return
+
+    print 'Unable to fix user: {} ({})'.format(user.domain, doc_id)
+
+
+def get_doc_revisions(db, doc_id):
+    res = db.get(doc_id, revs=True)
```
nickpell: @snopoke Is this management command still in use?  We're trying to move couchdbkit off of restkit to the [cloudant API](https://python-cloudant.readthedocs.io/en/latest/document.html#cloudant.document.Document), but the cloudant API doesn't support the ```revs``` kwarg.

snopoke: no, it was a one time command - fine to remove

PR #14573, opened 2017-01-18
updates to jan 2017 pricing plans

nickpell:
Does two things:

1) Uses the right product prices
2) Uses the number of included SMS that previously was set on production but not updated in the code.

A followup step is to check any subscriptions that were created since Jan 3 (when included SMS was reset to 500) and decide with Ops whether to update these subscriptions or grandfather them in.


nickpell:
@calellowitz

calellowitz (approved):

PR #14577, opened 2017-01-18
switch from phonenumberslite to phonenumbers

nickpell:
@emord staging deploy went through


czue:
what's motivating this? this package has a super high memory overhead which was why we originally switched. see [this link](https://github.com/daviddrysdale/python-phonenumbers#memory-usage) for more information

nickpell:
We're trying to get [requirements](https://github.com/dimagi/commcare-hq/blob/master/requirements/requirements.txt#L77) to use the main branch of ```django-two-factor-auth``` instead of our forked version.  Upgrading our requirements seems like a simpler path than getting https://github.com/Bouke/django-two-factor-auth/pull/178 accepted.

fwiw it sounds from that link like the memory overhead is loaded on-demand, and in HQ we aren't using any of those functions that are listed.  Maybe though there is extra overhead by default that isn't described in the link?

emord:
Did you check with an upgrade to two factor 1.5? We should make sure this would actually work first

emord:
@czue As mentioned in that ticket, pip uses wheel by default now and when installing wheel packages there actually isn't a difference between `phonenumbers` and `phonenumberslite` https://pypi.python.org/pypi/phonenumberslite (the `whl` package is 10x size as `tar.gz`)

czue:
@emord not sure what ticket you're referring to, but sounds like my concerns aren't relevant either way.

emord:
Err. Sorry it's not a ticket, it was my PR that was linked by bubbles earlier

nickpell:
closing in favor of https://github.com/dimagi/commcare-hq/pull/15128

PR #14578, opened 2017-01-19
don't verify nimbus SSL

snopoke:
http://manage.dimagi.com/default.asp?246037

This is to get around the issue on NIC where we have two certs for
the nginx site and requests is using the wrong one (the first
in the list).

@benrudolph @calellowitz

### DIFF
--- a/corehq/apps/nimbus_api/form_validation.py
+++ b/corehq/apps/nimbus_api/form_validation.py
@@ -42,7 +42,8 @@ def validate_form(form_xml):
         response = requests.post(
             get_nimbus_url() + const.ENDPOINT_VALIDATE_FORM,
             data=form_xml,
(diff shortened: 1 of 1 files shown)


benrudolph (approved):

czue:
thanks for super quick turnaround!

millerdev:
Would it make sense to only skip verification if some flag that says it's OK to skip verification is set (and only set that flag on NIC)? In general it's best practice to always verify SSL.

millerdev:
Curious if people care about the comment I made above, or if we think it's not an issue to always be skipping SSL verification here?

snopoke:
I was thinking that it wasn't enough of an issue for a short term workaround. Since we now have working SSL for ICDS we can revert this: #14716

PR #14581, opened 2017-01-19
Migrate inferred schemas

benrudolph:
@NoahCarnahan @czue this enables inferred schemas for forms and uses them in the migrate command. can be read 🐟 

*Caveat in this PR*: This changes the InferredSchema to CaseInferredSchema and FormInferredSchema. If we merge this, we drop all InferredSchemas because I exclude them in the view. I don't think that's a huge deal since I don't think many people depend on InferredSchemas yet because we haven't migrated apps that needed to add things to the InferredSchema. However case imports and user case properties are added to the InferredSchema. The user case properties will be re-added as soon as a save on the user happens, so I'm not worried about that. We do have the potential to drop items that were imported in the case upload and not used in the application. If they are used in the export, they will remain in any export. I'm kind of OK dropping those, but that might be me being lazy and not thinking many people rely on them...

Last thing to do before this wretched migration is over is to revamp the feature flag that lets you add custom export properties to add them to the inferred schema.

cc: @emord 

Note: Reindex is on the meta database


corehq/couchapps/inferred_schemas_by_case_type_or_xmlns/views/view/map.js
```diff
@@ -0,0 +1,7 @@
+function(doc) {
```
dimagimon: Parsing error: Unexpected token (

corehq/apps/export/tests/test_export_conversion.py
```diff
@@ -608,9 +683,10 @@ def test_skippable_export_columns(self, _, __):
         index, column = table.get_column(
             [PathNode(name='initial_processing_complete')], None, None
         )
-        self.assertIsInstance(column, UserDefinedExportColumn)
+        self.assertIsInstance(column, ExportColumn)
```
dimagimon: F821 undefined name 'ExportColumn'

corehq/apps/export/tests/test_export_conversion.py
```diff
@@ -608,9 +683,10 @@ def test_skippable_export_columns(self, _, __):
         index, column = table.get_column(
             [PathNode(name='initial_processing_complete')], None, None
         )
-        self.assertIsInstance(column, UserDefinedExportColumn)
```
dimagimon: E303 too many blank lines (3)

NoahCarnahan (approved):
Looks good to me.

With regards to losing case properties that only exist in case uploads: Do you have a ballpark number of how many domains this would effect? It sounds not great to drop this to me, but then again it won't matter if they do another upload or build out their apps to include those properties by the time they want to do an export.

Also just to make sure I'm understanding this right: Is the goal of adding `FormInferredSchema`s to support migrating remote apps?

benrudolph:
> With regards to losing case properties that only exist in case uploads: Do you have a ballpark number of how many domains this would effect? It sounds not great to drop this to me, but then again it won't matter if they do another upload or build out their apps to include those properties by the time they want to do an export.

Fair will try and look into it and see if there's an easy way to tell

> Also just to make sure I'm understanding this right: Is the goal of adding FormInferredSchemas to support migrating remote apps?

That and support migrating other domains that are too old to have had build_ids attached to their form submissions

corehq/apps/export/utils.py
```diff
@@ -58,8 +59,10 @@ def convert_saved_export_to_export_instance(
         ExportMigrationMeta,
         ConversionMeta,
         TableConfiguration,
-        InferredSchema,
+        CaseInferredSchema,
+        FormInferredSchema,
     )
+    from corehq.apps.export.system_properties import ROW_NUMBER_COLUMN
```
dimagimon: F401 'ROW_NUMBER_COLUMN' imported but unused

corehq/apps/export/utils.py
```diff
@@ -507,6 +523,7 @@ def _convert_stock_column(new_table, old_column):
 
 
 def _get_normal_column(new_table, column_path, transform):
+    from .models import PathNode
```
dimagimon: F401 'PathNode' imported but unused

benrudolph:
✅ i think worth discussing in scale team meeting before merging this

benrudolph:
after discussing with the scale team, we've decided to just merge this as is

snopoke:
I assume you'll merge this when ready

benrudolph:
👍 going to wait til formplayer settles down so don't have to wait on a preindex

benrudolph:
merging and kicking off preindex

PR #14587, opened 2017-01-19
Make sure locs in URL are accessible

esoergel:
Reports using the location choice provider have the options restricted to only those locations which are accessible to the user.  However, a user could modify a URL to include, say, a parent location, and get data they shouldn't have access to.  More realistically, if someone on the project shared a URL, there would be no verification that the restricted user loading it has access to those locations.  This PR adds in that check.
@proteusvacuum We offlined about this one.
@NoahCarnahan code buddy


proteusvacuum (approved):
Nice find! Thanks for fixing.

PR #14588, opened 2017-01-19
excluding swiss domain from nimbus form validation

calellowitz:
@sravfeyn 
They dont have a running formplayer so this was stopping builds


nickpell:
how will this affect https://www.commcarehq.org/a/icrc-almanach/dashboard/apps/?

calellowitz:
@nickpell It means that form validation will use old cloudcare instead of nimbus, which is important because there is not a functioning nimbus server on swiss. That was preventing them from doing builds.

nickpell:
(that link is in prod but we offlined and it's ok)

PR #14590, opened 2017-01-19
Phone number report

emord:
@gcapalbo I was hoping to get more tests in before I had to leave for the day, but I didn't have time. Can work on those more tomorrow, but wanted to get it up for review for deploy tomorrow.

https://docs.google.com/document/d/1uBDnmIRhqburT_zD6bzjVCWoGSr355g4ckyfeoFKcT8/edit?pli=1#

buddy @benrudolph

### DIFF
--- a/corehq/apps/reports/standard/sms.py
+++ b/corehq/apps/reports/standard/sms.py
@@ -1243,21 +1243,23 @@ class PhoneNumberReport(BaseCommConnectLogReport):
 
         return query.order_by('phone_number', 'couch_id')
 
+    @memoized
     def _get_users_without_phone_numbers(self):
         query = (
             PhoneNumber.objects.filter(domain=self.domain).filter(owner_doc_type__in=['CommCareUser', 'WebUser'])
         )
-        users_with_phone_numbers = {x[0]: x for x in query.values_list('owner_id', 'owner_doc_type').distinct()}
 
         if self.selected_group:
             users_by_id = {
                 id: {'_id': id, 'doc_type': 'CommCareUser'}
                 for id in self.user_ids_in_selected_group
             }
+            query.filter(owner_id__in=users_by_id.keys())
         else:
             users_by_id = {u.get_id: u for u in CouchUser.by_domain(self.domain)}
 
-        user_ids = set(users_by_id.keys()) - set(users_with_phone_numbers.keys())
+        user_ids_with_phone_numbers = {x[0] for x in query.values_list('owner_id').distinct()}
+        user_ids = set(users_by_id.keys()) - user_ids_with_phone_numbers
         user_types_with_id = {(id, users_by_id[id]['doc_type']) for id in user_ids}
 
         FakePhoneNumber = namedtuple('FakePhoneNumber', ['owner_id', 'owner_doc_type'])
--- a/corehq/apps/reports/standard/sms.py
+++ b/corehq/apps/reports/standard/sms.py
@@ -1178,6 +1178,10 @@ class PhoneNumberReport(BaseCommConnectLogReport):
     def _show_cases(self):
         return self.contact_type == 'cases'
 
+    @classmethod
+    def show_in_navigation(cls, domain=None, project=None, user=None):
+        return domain and toggles.PHONE_NUMBERS_REPORT.enabled(domain)
+
     def _fmt_owner(self, owner_doc_type, owner_id, owner_cache, link_user=True):
         doc_info = self.get_recipient_info(owner_doc_type, owner_id, owner_cache)
         table_cell = self._fmt_contact_link(owner_id, doc_info)
--- a/corehq/apps/reports/standard/sms.py
+++ b/corehq/apps/reports/standard/sms.py
@@ -1097,3 +1097,10 @@ class SMSOptOutReport(ProjectReport, ProjectReportParametersMixin, GenericTabula
     @property
     def export_rows(self):
         return self._get_rows(paginate=False)
+
+
+class PhoneNumberReport(BaseCommConnectLogReport):
+    name = ugettext_noop("Phone Number Report")
+    slug = 'phone_number_report'
+    ajax_pagination = True
+    fields = []
--- a/corehq/apps/reports/standard/sms.py
+++ b/corehq/apps/reports/standard/sms.py
@@ -1184,8 +1184,11 @@ class PhoneNumberReport(BaseCommConnectLogReport):
             return "Verified"
         elif number.pending_verification:
             return "Verification Pending"
-        elif not (number.is_two_way or number.pending_verification) and PhoneNumber.get_reserved_number(number):
-            return "Already in use"
+        elif (
+                not (number.is_two_way or number.pending_verification) and
+                PhoneNumber.get_reserved_number(number.phone_number)
+             ):
+            return "Already In Use"
         return "Not Verified"
 
     def _fmt_row(self, number, owner_cache):
--- a/corehq/apps/reports/standard/sms.py
+++ b/corehq/apps/reports/standard/sms.py
@@ -1262,7 +1262,7 @@ class PhoneNumberReport(BaseCommConnectLogReport):
         else:
             users_by_id = {u['id']: u for u in get_user_id_and_doc_type_by_domain(self.domain)}
 
-        user_ids_with_phone_numbers = {x[0] for x in query.values_list('owner_id').distinct()}
+        user_ids_with_phone_numbers = set(query.values_list('owner_id', flat=True).distinct())
         user_ids = set(users_by_id.keys()) - user_ids_with_phone_numbers
         user_types_with_id = {(id, users_by_id[id]['doc_type']) for id in user_ids}
 
--- a/corehq/apps/reports/standard/sms.py
+++ b/corehq/apps/reports/standard/sms.py
@@ -1228,14 +1228,11 @@ class PhoneNumberReport(BaseCommConnectLogReport):
                 query = query.filter(phone_number=self.phone_number_filter)
         elif self.filter_type == 'contact':
             if self._show_cases:
-                query = query.filter(owner_doc_type='CommCareCase', phone_number__isnull=False)
+                query = query.filter(owner_doc_type='CommCareCase')
             else:
                 query = query.filter(owner_doc_type__in=['CommCareUser', 'WebUser'])
                 if self.selected_group:
                     query = query.filter(owner_id__in=self.user_ids_in_selected_group)
-
-                if self.has_phone_number == 'has_phone_number':
-                    query = query.filter(phone_number__isnull=False)
 
             if self.verification_status == 'not_verified':
                 query = query.filter(pending_verification=False, verified=False)
--- a/corehq/apps/reports/standard/sms.py
+++ b/corehq/apps/reports/standard/sms.py
@@ -1205,7 +1205,7 @@ class PhoneNumberReport(BaseCommConnectLogReport):
             ]
 
         return [
-            self._fmt_owner(number.owner_doc_type, number.owner_id, owner_cache),
+            self._fmt_owner(number.owner_doc_type, number.owner_id, owner_cache, link_user),
             '---',
             '---',
             '---',
--- a/corehq/apps/reports/standard/sms.py
+++ b/corehq/apps/reports/standard/sms.py
@@ -138,6 +138,7 @@ def _sms_count(user, startdate, enddate):
 
 
 class BaseCommConnectLogReport(ProjectReport, ProjectReportParametersMixin, GenericTabularReport, DatespanMixin):
+    contact_index_in_result = 1
 
     def _fmt(self, val):
         if val is None:
@@ -244,8 +245,8 @@ class BaseCommConnectLogReport(ProjectReport, ProjectReportParametersMixin, Gene
         table[0].insert(0, _("Contact Id"))
         table[0].insert(0, _("Contact Type"))
         for row in table[1:]:
-            contact_info = row[1].split("|||")
-            row[1] = contact_info[0]
+            contact_info = row[self.contact_index_in_result].split("|||")
+            row[self.contact_index_in_result] = contact_info[0]
             row.insert(0, contact_info[2])
             row.insert(0, contact_info[1])
         return result
@@ -1110,9 +1111,11 @@ class PhoneNumberReport(BaseCommConnectLogReport):
     name = ugettext_noop("Phone Number Report")
     slug = 'phone_number_report'
     ajax_pagination = True
+    exportable = True
     fields = [
         PhoneNumberReportFilter
     ]
+    contact_index_in_result = 0
 
     @property
     def headers(self):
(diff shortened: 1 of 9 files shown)


corehq/apps/sms/static/sms/js/phone_number_filter.js
```diff
@@ -0,0 +1,42 @@
+function SMSPhoneNumberFilterViewModel(initial_value, groups) {
```
dimagimon: 'SMSPhoneNumberFilterViewModel' is defined but never used (no-unused-vars)

corehq/apps/sms/filters.py
```diff
@@ -1,7 +1,10 @@
+from django.core.urlresolvers import reverse
```
dimagimon: F401 'reverse' imported but unused

corehq/apps/sms/filters.py
```diff
@@ -1,7 +1,10 @@
...
 from django.utils.translation import ugettext_noop, ugettext_lazy
 from corehq import toggles
-from corehq.apps.reports.filters.base import BaseMultipleOptionFilter, BaseSingleOptionFilter
+from corehq.apps.es.groups import GroupES
+from corehq.apps.reports.filters.base import BaseMultipleOptionFilter, BaseSingleOptionFilter, BaseReportFilter
 from corehq.apps.reports.filters.search import SearchFilter
+from corehq.apps.locations.util import load_locs_json, location_hierarchy_config
```
dimagimon: F401 'load_locs_json' imported but unused
F401 'location_hierarchy_config' imported but unused

corehq/apps/reports/standard/sms.py
```diff
@@ -1097,3 +1103,183 @@ def shared_pagination_GET_params(self):
...
+                for id in self.user_ids_in_selected_group
+            }
+        else:
+            users_by_id = {u.get_id: u for u in CouchUser.by_domain(self.domain)}
+
+        user_ids = set(users_by_id.keys()) - set(users_with_phone_numbers.keys())
+        user_types_with_id = {(id, users_by_id[id]['doc_type']) for id in user_ids}
+
+        FakePhoneNumber = namedtuple('FakePhoneNumber', ['owner_id', 'owner_doc_type'])
+        return sorted([FakePhoneNumber(id, type) for id, type in user_types_with_id])
```
emord: This means that all users without a number will be sorted by id. There wasn't anything in the spec about it, but it would be pretty difficult to fix this for the case of getting users in a group that don't have phone numbers

corehq/apps/reports/standard/sms.py
```diff
@@ -1097,3 +1103,183 @@ def shared_pagination_GET_params(self):
...
+        )
+        users_with_phone_numbers = {x[0]: x for x in query.values_list('owner_id', 'owner_doc_type').distinct()}
+
+        if self.selected_group:
+            users_by_id = {
+                id: {'_id': id, 'doc_type': 'CommCareUser'}
+                for id in self.user_ids_in_selected_group
+            }
+        else:
+            users_by_id = {u.get_id: u for u in CouchUser.by_domain(self.domain)}
```
emord: I think that this could be fed into the `owner_cache` in `get_rows` but that's an optimization I'm not sure we need

corehq/apps/reports/standard/sms.py
```diff
@@ -1103,4 +1109,177 @@ class PhoneNumberReport(BaseCommConnectLogReport):
...
+                query = query.filter(owner_doc_type='CommCareCase', phone_number__isnull=False)
+            else:
+                query = query.filter(owner_doc_type__in=['CommCareUser', 'WebUser'])
+                if self.selected_group:
+                    query = query.filter(owner_id__in=self.user_ids_in_selected_group)
+
+                if self.has_phone_number == 'has_phone_number':
+                    query = query.filter(phone_number__isnull=False)
+
+            if self.verification_status == 'not_verified':
```
benrudolph: nit: some of these strings could be constants

benrudolph (approved):
code looks fine to me

corehq/apps/reports/standard/sms.py
```diff
@@ -1103,4 +1109,177 @@ class PhoneNumberReport(BaseCommConnectLogReport):
...
+        doc_info = self.get_recipient_info(owner_doc_type, owner_id, owner_cache)
+        table_cell = self._fmt_contact_link(owner_id, doc_info)['html']
+        return table_cell
+
+    def _fmt_status(self, number):
+        if number.verified:
+            return "Verified"
+        elif number.pending_verification:
+            return "Verification Pending"
+        elif not (number.is_two_way or number.pending_verification) and PhoneNumber.get_reserved_number(number):
```
gcapalbo: should be `get_reserved_number(number.phone_number)` i think right?

corehq/apps/reports/standard/sms.py
```diff
@@ -1103,4 +1109,177 @@ class PhoneNumberReport(BaseCommConnectLogReport):
...
+
+        for number in data:
+            yield self._fmt_row(number, owner_cache)
+
+    def _get_queryset(self):
+        query = PhoneNumber.objects.filter(domain=self.domain)
+
+        if self.filter_type == 'phone_number':
+            if self.phone_number_filter:
+                query = query.filter(phone_number__contains=self.phone_number_filter)
```
gcapalbo: this might not perform well if the phone number table gets huge. i'd be ok with making this an exact match for now

corehq/apps/reports/standard/sms.py
```diff
@@ -1103,4 +1109,177 @@ class PhoneNumberReport(BaseCommConnectLogReport):
...
+        elif self.filter_type == 'contact':
+            if self._show_cases:
+                query = query.filter(owner_doc_type='CommCareCase', phone_number__isnull=False)
```
gcapalbo: we can remove the not null filter on `phone_number`, that should never happen. it was created as a nullable column when we migrated from couch and i just haven't updated it

corehq/apps/reports/standard/sms.py
```diff
@@ -1103,4 +1109,177 @@ class PhoneNumberReport(BaseCommConnectLogReport):
...
+            else:
+                query = query.filter(owner_doc_type__in=['CommCareUser', 'WebUser'])
+                if self.selected_group:
+                    query = query.filter(owner_id__in=self.user_ids_in_selected_group)
+
+                if self.has_phone_number == 'has_phone_number':
+                    query = query.filter(phone_number__isnull=False)
```
gcapalbo: can also remove this filter for the same reason

corehq/apps/reports/standard/sms.py
```diff
@@ -1103,4 +1109,177 @@ class PhoneNumberReport(BaseCommConnectLogReport):
...
+            elif self.verification_status == 'verified':
+                query = query.filter(verified=True)
+
+        return query.order_by('phone_number', 'couch_id')
+
+    def _get_users_without_phone_numbers(self):
+        query = (
+            PhoneNumber.objects.filter(domain=self.domain).filter(owner_doc_type__in=['CommCareUser', 'WebUser'])
+        )
+        users_with_phone_numbers = {x[0]: x for x in query.values_list('owner_id', 'owner_doc_type').distinct()}
```
gcapalbo: i think we only need to select owner_id here right?

corehq/apps/reports/standard/sms.py
gcapalbo: yeah i don't think that's necessary. but it would be good to only grab id and doc_type from the view instead of grabbing the whole doc. it should be possible since doc_type is exposed in the key, so if we query for include_docs=False we can grab it from the emitted key

corehq/apps/reports/standard/sms.py
gcapalbo: i think we can also simplify this a bit when we're filtering on a group by filtering on owner_id in the group's user ids

corehq/apps/reports/standard/sms.py
gcapalbo: yup, we want to sort this result by id, that's what was in the spec too

corehq/apps/reports/standard/sms.py
```diff
@@ -1103,4 +1109,177 @@ class PhoneNumberReport(BaseCommConnectLogReport):
...
+        ]
+
+    @property
+    def rows(self):
+        return self._get_rows()
+
+    @property
+    def total_records(self):
+        if self._show_users_without_phone_numbers:
+            return len(self._get_users_without_phone_numbers())
```
gcapalbo: i think we can memoize `_get_users_without_phone_numbers()` so it doesn't get called twice

gcapalbo:
looks good @emord, let me know once updated and i can test a bit locally as well

gcapalbo:
@emord i just realized i think you'll have to do some tweaking to get the excel export working. i think you'll have to override `export_rows` to return `self._get_rows(paginate=False)`. you may need to also make some tweaks to how `export_table` is calculated since it's assuming the contact is the second column in the report in the `BaseCommConnectLogReport`. i think it's fine to also just override these properties for this report if it's too hard/unnatural to reuse what's in the `BaseCommConnectLogReport`

corehq/apps/reports/standard/sms.py
```diff
@@ -1097,3 +1105,191 @@ def shared_pagination_GET_params(self):
...
+
+    def _fmt_status(self, number):
+        if number.verified:
+            return "Verified"
+        elif number.pending_verification:
+            return "Verification Pending"
+        elif (
+                not (number.is_two_way or number.pending_verification) and
+                PhoneNumber.get_reserved_number(number.phone_number)
+             ):
```
dimagimon: E126 continuation line over-indented for hanging indent

corehq/apps/users/dbaccessors/__init__.py
```diff
@@ -1,3 +1,3 @@
 from .all_commcare_users import (get_all_commcare_users_by_domain,
                                  get_user_docs_by_username)
-from .couch_users import get_user_id_by_username
+from .couch_users import get_user_id_by_username, get_user_id_and_doc_type_by_domain
```
dimagimon: F401 'get_user_id_and_doc_type_by_domain' imported but unused
F401 'get_user_id_by_username' imported but unused

corehq/apps/reports/standard/sms.py
emord: https://github.com/dimagi/commcare-hq/pull/14590/commits/ef635f858ffe5d683d793d39c321dd46a4d88f63

corehq/apps/reports/standard/sms.py
emord: https://github.com/dimagi/commcare-hq/pull/14590/commits/380f13d5423aaf2a9e72caf9ce7fb62d5de70f80 

Pulled the `__contains` from https://github.com/dimagi/commcare-hq/blob/master/corehq/apps/reports/standard/sms.py#L765 Might be worth thinking about that as well?

corehq/apps/reports/standard/sms.py
emord: https://github.com/dimagi/commcare-hq/pull/14590/commits/d92d6a2b6b46548e1f1775673533d72c3e181dc3

corehq/apps/reports/standard/sms.py
emord: https://github.com/dimagi/commcare-hq/pull/14590/commits/9b3d9671f9ba8f0eb359966108380aaee9577d38

corehq/apps/reports/standard/sms.py
emord: https://github.com/dimagi/commcare-hq/pull/14590/commits/9b3d9671f9ba8f0eb359966108380aaee9577d38

corehq/apps/reports/standard/sms.py
emord: Ah. I didn't think about using the key. https://github.com/dimagi/commcare-hq/pull/14590/commits/388c3f1a7904cbec163555cab923912e3269ac32

emord:
For the exports the easiest thing to do was to pull out the index into a property of the report https://github.com/dimagi/commcare-hq/pull/14590/commits/2d5fbd63abce7c75b5489b1df99455fe4b8d60b4

corehq/apps/reports/standard/sms.py
gcapalbo: yeah i think you're right. we're going to have to do some cleanup with that report anyway, will add that to the list

corehq/apps/reports/standard/sms.py
```diff
@@ -1243,21 +1243,23 @@ def _get_queryset(self):
...
             users_by_id = {
                 id: {'_id': id, 'doc_type': 'CommCareUser'}
                 for id in self.user_ids_in_selected_group
             }
+            query.filter(owner_id__in=users_by_id.keys())
         else:
             users_by_id = {u.get_id: u for u in CouchUser.by_domain(self.domain)}
 
-        user_ids = set(users_by_id.keys()) - set(users_with_phone_numbers.keys())
+        user_ids_with_phone_numbers = {x[0] for x in query.values_list('owner_id').distinct()}
```
gcapalbo: can also just do `query.values_list('owner_id', flat=True).distinct()`

corehq/apps/reports/standard/sms.py
```diff
@@ -1097,3 +1105,191 @@ def shared_pagination_GET_params(self):
...
+                id: {'_id': id, 'doc_type': 'CommCareUser'}
+                for id in self.user_ids_in_selected_group
+            }
+            query.filter(owner_id__in=users_by_id.keys())
+        else:
+            users_by_id = {u['id']: u for u in get_user_id_and_doc_type_by_domain(self.domain)}
+
+        user_ids_with_phone_numbers = {x[0] for x in query.values_list('owner_id').distinct()}
+        user_ids = set(users_by_id.keys()) - user_ids_with_phone_numbers
+        user_types_with_id = {(id, users_by_id[id]['doc_type']) for id in user_ids}
```
gcapalbo: i think this can just be a list, right? might perform a bit better as a list since it wouldn't have to keep track of set-related info

corehq/apps/reports/standard/sms.py
```diff
@@ -1097,3 +1105,191 @@ def shared_pagination_GET_params(self):
...
+
+        FakePhoneNumber = namedtuple('FakePhoneNumber', ['owner_id', 'owner_doc_type'])
+        return sorted([FakePhoneNumber(id, type) for id, type in user_types_with_id])
```
gcapalbo: does this actually sort by id? might want to add key to `sorted` to make it explicit

emord: Moved that up to the list function. Agreed that it's not obvious from the namedtuple usage, but tuples do sort by comparing the first element, then second, and so on

gcapalbo:
i just tested this locally @emord and it looks great, i just made one small fix and also put it behind a toggle until we sync all the numbers in all domains. 

going to merge now so this gets in with today's deploy; the previous build passed and i ran the report tests locally for the last two commits

thanks for all your work on this!

czue:
excited to see this

PR #14595, opened 2017-01-20
remove get_form_ids_or_querystring to avoid double check on request a…

mkangia:
…nd then on return type

Follow up from https://github.com/dimagi/commcare-hq/pull/14206/files#diff-9f0c2d0ac3b32ff875ff48f73f7b3e10R532
@esoergel

### DIFF
--- a/corehq/apps/data_interfaces/views.py
+++ b/corehq/apps/data_interfaces/views.py
@@ -488,16 +488,6 @@ class XFormManagementView(DataInterfaceSection):
     urlname = 'xform_management'
     page_title = ugettext_noop('Form Management')
 
-    def get_form_ids_or_query_string(self, request):
-        if 'select_all' in self.request.POST:
-            # Altough evaluating form_ids and sending to task would be cleaner,
-            # heavier calls should be in in an async task instead
-            import urllib
-            form_query_string = urllib.unquote(self.request.POST.get('select_all'))
-            return form_query_string
-        else:
-            return self.request.POST.getlist('xform_ids')
-
     def post(self, request, *args, **kwargs):
         form_ids = self.get_xform_ids(request)
         if not self.request.can_access_all_locations:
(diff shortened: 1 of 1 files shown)


corehq/apps/data_interfaces/views.py
```diff
@@ -531,10 +521,11 @@ def inaccessible_forms_accessed(self, xform_ids, domain, couch_user):
         return xforms_user_ids - accessible_user_ids
 
     def get_xform_ids(self, request):
-        form_ids_or_query_string = self.get_form_ids_or_query_string(request)
-        if type(form_ids_or_query_string) == list:
-            xform_ids = form_ids_or_query_string
-        elif isinstance(form_ids_or_query_string, basestring):
+        if 'select_all' in self.request.POST:
+            # Altough evaluating form_ids and sending to task would be cleaner,
+            # heavier calls should be in an async task instead
```
esoergel: hmm, I just took a closer look at what this existing code (and this comment) were doing - the comment refers to behavior that was just changed in the last PR (the "heavier call" no longer happens in the async task).

Also, it's out of scope to change, but that request mocking business seems like a needlessly complex way of performing an elasticsearch query.

esoergel:
Thanks for the change!

corehq/apps/data_interfaces/views.py
mkangia: Thanks @esoergel. Happy to make any improvements in this. What do you suggest?

esoergel: I was just thinking of dropping the comment since it's no longer accurate.

esoergel: The bigger fix would be to abstract out the ES query that accepts a querystring and returns form ids, then just call it from here and from the view where it lives now.  That's a bit out of scope, and there may be complications which led the original author of this to choose this request mocking business instead, so I wouldn't worry about it too much.

PR #14598, opened 2017-01-20
Add Task to Sync Phone Numbers

gcapalbo:
@emord this will create all the missing phone entries (i.e., the one-way entries) to populate for the report


emord (approved):
Is this something that will need to be kicked off manually, or will there be a follow up to make it run in the background for all domains?

gcapalbo:
yeah there's going to be a followup to run this in the background for all projects. for now i just wanted a way to run it for one project that we're doing this for

PR #14575, opened 2017-01-18
debug code for bootstrap-tab-history-custom.js

orangejenny:
@proteusvacuum I started seeing the blank screen issue you've been getting, but only intermittently on india, and I haven't had much luck working in the chrome debugger to figure out what's going on. Adding some temporary debug code to narrow down where the problem is.

PR #14599, opened 2017-01-20
More debugging in bootstrap-tab-history

orangejenny:
@proteusvacuum 

Revert https://github.com/dimagi/commcare-hq/pull/14575 since it didn't teach us anything, and add some logging around the events. My new theory is that the order of window load and domready isn't deterministic.

### DIFF
--- a/corehq/apps/hqwebapp/static/hqwebapp/js/lib/bootstrap-tab-history-custom.js
+++ b/corehq/apps/hqwebapp/static/hqwebapp/js/lib/bootstrap-tab-history-custom.js
@@ -1,5 +1,7 @@
 // copied and modified from https://gist.github.com/dsully/1938283
+console.log("Loading bootstrap-tab-history-custom.js");
 $(function () {
+    console.log("Document ready handler in bootstrap-tab-history-custom.js");
 
     function loadPage(url) {
         window.location.href = url;
@@ -25,7 +27,6 @@ $(function () {
                 State.data.tab = '';
             }
         }
-        console.log("State.data.tab = " + State.data.tab);
 
         var link;
         if (State.data.tab) {
@@ -52,15 +53,13 @@ $(function () {
                 }, null, State.url);
             }
         }
-        if (link.length) {
-            console.log(link.get(0).outerHTML);
-        } else {
-            console.log("No link found");
-        }
         link.parent().removeClass('active');    // force tab to load
         link.tab('show');
     };
-    $(window).on('load', statechange);
+    $(window).on('load', function() {
+        console.log("Window ready handler in bootstrap-tab-history-custom.js");
+        statechange();
+    });
     History.Adapter.bind(window, 'statechange', statechange);
     History.Adapter.bind(window, 'statechange', function () {
         var State = History.getState();


orangejenny:
would like this to go into today's deploy, please @calellowitz

corehq/apps/hqwebapp/static/hqwebapp/js/lib/bootstrap-tab-history-custom.js
```diff
@@ -1,5 +1,7 @@
 // copied and modified from https://gist.github.com/dsully/1938283
+console.log("Loading bootstrap-tab-history-custom.js");
```
dimagimon: Unexpected console statement. (no-console)

corehq/apps/hqwebapp/static/hqwebapp/js/lib/bootstrap-tab-history-custom.js
```diff
@@ -1,5 +1,7 @@
...
 $(function () {
+    console.log("Document ready handler in bootstrap-tab-history-custom.js");
```
dimagimon: Unexpected console statement. (no-console)

corehq/apps/hqwebapp/static/hqwebapp/js/lib/bootstrap-tab-history-custom.js
```diff
@@ -52,15 +53,13 @@ $(function () {
...
-            console.log(link.get(0).outerHTML);
-        } else {
-            console.log("No link found");
-        }
         link.parent().removeClass('active');    // force tab to load
         link.tab('show');
     };
-    $(window).on('load', statechange);
+    $(window).on('load', function() {
+        console.log("Window ready handler in bootstrap-tab-history-custom.js");
```
dimagimon: Unexpected console statement. (no-console)

benrudolph:
app preview is also struggling from this... i can't find a solution. i read a few threads that i thought was relevant didn't seem to help. window load and dom ready is no longer deterministic in jquery3

benrudolph:
for example, if a load event happens in a ready event, in jquery 3, the load event can fire before the ready event and then the load event never gets called. this is a change in functionality with the upgrade

proteusvacuum:
No tests, so merging

orangejenny:
@benrudolph Did you find jQuery docs on that specifically, or is it something they didn't announce but people have observed? My guesses about it being event-related are based on them having made changes like https://jquery.com/upgrade-guide/3.0/#breaking-change-document-ready-handlers-are-now-asynchronous

For this specific bug, I'm hoping that if it is event-related and the logs support that, it'll be fixable by doing something like calling `statechange` another time, maybe when the js is parsed or maybe on domready instead of window load.

benrudolph:
that is the exact change that i'm dealing with: https://jquery.com/upgrade-guide/3.0/#breaking-change-document-ready-handlers-are-now-asynchronous. there was also a long issue on it on github: https://github.com/jquery/jquery/issues/3194. the reason this is failing is definitely because of that breaking change

PR #14602, opened 2017-01-20
Fix tabs on app settings in app manager v2

orangejenny:
Something mixed up the tabs on app settings (`Data and Sharing` doesn't go under `Languages`), I think a bad merge in linked apps:

screen shot 2017-01-20 at 4 43 52 pm

After (normal app):
screen shot 2017-01-20 at 4 43 24 pm

After (linked app):
screen shot 2017-01-20 at 4 43 17 pm

@calellowitz


calellowitz:
This is probably just my unfamiliarity with formplayer v2 but where are the deploy page and delete app options for the linked app.

orangejenny:
The deploy page is okay (the app title in the top left links to it). But you're right that the delete action is now missing...will fix.

orangejenny:
Fixed. This made an oddly large difference in the diff file.

screen shot 2017-01-20 at 5 38 59 pm

screen shot 2017-01-20 at 5 39 04 pm

calellowitz:
not sure why the diff file is so large but looks good to me

PR #14603, opened 2017-01-20
admin phone report

emord:
@gcapalbo 

buddy @benrudolph

### DIFF
--- a/corehq/apps/hqadmin/reports.py
+++ b/corehq/apps/hqadmin/reports.py
@@ -16,6 +16,9 @@ from corehq.apps.app_manager.models import Application
 from corehq.apps.reports.dispatcher import AdminReportDispatcher
 from corehq.apps.reports.generic import ElasticTabularReport, GenericTabularReport
 from corehq.apps.reports.standard.domains import DomainStatsReport, es_domain_query
+from corehq.apps.reports.standard.sms import PhoneNumberReport
+from corehq.apps.sms.filters import RequiredPhoneNumberFilter
+from corehq.apps.sms.models import PhoneNumber
 from django.utils.translation import ugettext as _, ugettext_noop, ugettext_lazy
 from corehq.elastic import es_query, parse_args_for_es, fill_mapping_with_facets
 from corehq.apps.app_manager.commcare_settings import get_custom_commcare_settings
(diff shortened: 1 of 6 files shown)


corehq/apps/hqadmin/reports.py
```diff
@@ -1287,3 +1290,46 @@ def get_data(domains):
...
+
+
+class AdminPhoneNumberReport(PhoneNumberReport):
+    name = ugettext_lazy("Admin Phone Number Report")
+    slug = 'phone_number_report'
+    fields = [
+        RequiredPhoneNumberFilter,
+    ]
+
+    dispatcher = AdminReportDispatcher
```
emord: Didn't work with inheriting `PhoneNumberReport` and `AdminReport` and multi-inheritance gets fairly confusing anyways. 

These are the only 3 attributes we would gain from inheriting `AdminReport`

corehq/apps/hqadmin/reports.py
```diff
@@ -1287,3 +1290,46 @@ def get_data(domains):
...
+        value = RequiredPhoneNumberFilter.get_value(self.request, domain=None)
+        if isinstance(value, basestring):
+            return value.strip()
+
+        return None
+
+    def _get_rows(self, paginate=True, link_user=True):
+        owner_cache = {}
+        if self.phone_number_filter:
+            data = PhoneNumber.objects.filter(phone_number=self.phone_number_filter)
```
gcapalbo: we should call `apply_leniency` on the phone number before using it in the filter

gcapalbo: i also noticed this filter mentions you can enter a partial number but we're only matching on exact number. since this is an admin report and it won't be used too frequently i think it's ok to open this up and match using `__contains`

gcapalbo:
looks good overall @emord

corehq/apps/hqadmin/reports.py
emord: Cool. Removed __contains because we didn't want it in the other report. Added it back here: https://github.com/dimagi/commcare-hq/pull/14603/commits/2653ac1f5ba79d64d950c50ccee650353dfe8e4b

gcapalbo (approved):

PR #14604, opened 2017-01-20
Make sure locs in URL are accessible

esoergel:
From https://github.com/dimagi/commcare-hq/pull/14592/
@czue @NoahCarnahan
After causing a failure I'm a little hesitant to just merge this as-is, do either of you know of a staging app where I can test this out?


czue:
> do either of you know of a staging app where I can test this out?

I don't, unfortunately. I think you just need to make a report in report builder and then setup a mobile UCR module with a location auto-filter though which might not be too bad. might be just as easy to write a test though.

esoergel:
Took me a bit to figure out how to get it set up, but I managed to get a mobile UCR with a location filter working (I'd been meaning to explore that functionality for a bit anyways).  It all worked fine after I sorted out the configuration issues.

PR #14607, opened 2017-01-22
ensure that boolean fields are assigned boolean values

nickpell:
for django 1.10 compatibility

### DIFF
--- a/corehq/apps/users/models.py
+++ b/corehq/apps/users/models.py
@@ -1088,13 +1088,19 @@ class CouchUser(Document, DjangoUserMixin, IsMemberOfMixin, UnicodeMixIn, EulaMi
             django_user = User(username=self.username)
         for attr in DjangoUserMixin.ATTRS:
             attr_val = getattr(self, attr)
-            if not attr_val and attr != 'last_login':
+            if attr in [
+                'is_active',
+                'is_staff',
+                'is_superuser',
+            ]:
+                attr_val = bool(attr_val)
+            elif not attr_val and attr != 'last_login':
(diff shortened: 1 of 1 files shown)


nickpell:
@gcapalbo @proteusvacuum

corehq/apps/users/models.py
```diff
@@ -1088,13 +1088,19 @@ def sync_to_django_user(self):
             django_user = User(username=self.username)
         for attr in DjangoUserMixin.ATTRS:
             attr_val = getattr(self, attr)
-            if not attr_val and attr != 'last_login':
+            if attr in [
+                'is_active',
+                'is_staff',
+                'is_superuser',
+            ]:
+                attr_val = bool(attr_val)
```
gcapalbo: is this just to set a default when `attr_val` is None? if so it might be better to just explicitly default it to False when it's None.

nickpell: in the tests it was getting passed ```''```.  had something to do with the couch-sql integration for users.  In django <1.10 it worked, but in 1.10 you have to assign boolean values to these fields.

gcapalbo: i see. to be safe it might be best to say:

`attr_val = attr_val if attr_val is True else False`

it just seems a bit dangerous to set `is_superuser` to the outcome of `bool(attr_val)`, in case `attr_val` gets set to something like `'False'`

nickpell: Good point - updated

gcapalbo (approved):

PR #14608, opened 2017-01-22
update django-nose

nickpell:
for django 1.10


nickpell:
@millerdev

PR #14609, opened 2017-01-22
update django REST framework

nickpell:
for django 1.10 compatibility

http://www.django-rest-framework.org/topics/release-notes/


nickpell:
@emord

PR #14611, opened 2017-01-22
replace deprecated get_fields_with_model

nickpell:


nickpell:
@gcapalbo @proteusvacuum

PR #14612, opened 2017-01-22
update django-statici18n

nickpell:
https://github.com/zyegfryed/django-statici18n/blob/master/docs/changelog.rst


nickpell:
@emord tested on staging

PR #14613, opened 2017-01-22
remove nonexisting kwarg

nickpell:
passing wrong kwarg to ```create``` fails hard in django 1.10


nickpell:
@gcapalbo @proteusvacuum

PR #14615, opened 2017-01-22
update management commands for django 1.10

nickpell:
only affects management commands that have test coverage.  still plenty more to go


nickpell:
closing in favor of more granular PRs

PR #14616, opened 2017-01-22
update request context processor import

nickpell:
needed for django 1.10


nickpell:
@emord tested on staging

PR #14618, opened 2017-01-23
Get system column without transform

benrudolph:
@emord there were assertions on this method when it would try to add `user_id` because that path maps to two system properties. this ensures we take the one that doesn't have a transform


snopoke:
http://manage.dimagi.com/default.asp?246207

PR #14620, opened 2017-01-23
Added date filter to TB notification register

kkrampa:
@czue


czue (approved):
@calellowitz this is what should get deployed to india.

PR #14622, opened 2017-01-23
break out of loop

snopoke:
http://manage.dimagi.com/default.asp?246172

@czue is this what you meant?


czue (approved):
wow yeah, definitely. so this basically happened to work (probably incorrectly) as long as the loop was successfully executed on the first pass. 

surprised it took this long to run into it!

millerdev:
Would it be hard to add a test for the bug?

PR #14623, opened 2017-01-23
allow TaskStatusResultError to be non-int

dannyroberts:
because some errors are using it this way

http://manage.dimagi.com/default.asp?246211 @snopoke


snopoke:
Thank @dannyroberts

PR #14624, opened 2017-01-23
Move case attachments to blob db

millerdev:
Probably best viewed by commit 🐠 

@snopoke @dannyroberts


corehq/blobs/migrate.py
```diff
@@ -499,6 +500,12 @@ def migrate(self, filename=None, reset=False, max_retry=2, chunk_size=100):
         xform.SubmissionErrorLog,
         ("HQSubmission", xform.XFormInstance),
     ], CouchAttachmentMigrator),
+    Migrator("cases", [
+        cases.CommCareCase,
+        ('CommCareCase-deleted', cases.CommCareCase),
+        ('CommCareCase-Deleted', cases.CommCareCase),
+        ('CommCareCase-Deleted-Deleted', cases.CommCareCase),
```
millerdev: I'm pretty sure these are the only classes that need to be migrated. There is a [`SupplyPointCase`](https://github.com/dimagi/commcare-hq/blob/c334bd9ed7f1c8dd7a8ac8329237349a68d4b3a2/corehq/apps/commtrack/models.py#L285) subclass of `CommCareCase`, but it looks like it's never saved with that doctype (at least there were none saved in couch on prod when I looked last week). Also there are a bunch of subclasses in the `custom` code dir, but it looked like they were all reporting helpers and never saved (and there were none saved in prod couch).

corehq/apps/app_manager/models.py
```diff
@@ -4114,7 +4114,7 @@ def wrap(cls, data):
         if "_attachments" in data:
             data = data.copy()
             attachments = data.pop("_attachments").copy()
-            if cls.migrating_blobs_from_couch:
+            if cls._migrating_blobs_from_couch:
```
millerdev: Previously this had always evaluated to `True` because jsonobject replaced the class attribute boolean value with a `jsonobject.properties.BooleanProperty`. It now works as expected, which is why `_attachments` had to be removed from some test JSON docs.

corehq/ex-submodules/pillowtop/tests/test_elasticsearch.py
```diff
@@ -198,8 +198,8 @@ def _bad_es_getter():
             from elasticsearch import Elasticsearch
             return Elasticsearch(
                 [{
-                    'host': 'localhost',
-                    'port': '9000',  # bad port
+                    'host': settings.ELASTICSEARCH_HOST,
+                    'port': settings.ELASTICSEARCH_PORT - 2,  # bad port
```
millerdev: This test was failing for me because I have a service running on `localhost:9000`, and the ES service was running in boot2docker, which has a different IP address than localhost.

corehq/blobs/mixin.py
```diff
@@ -432,13 +443,25 @@ def deferred_put_attachment(self, content, name=None, content_type=None,
...
                 # list deferred blobs to avoid modification during iteration
                 for name, info in list(self._deferred_blobs.iteritems()):
-                    self.put_attachment(name=name, **info)
-                assert not self._deferred_blobs, self._deferred_blobs
+                    if info is not None:
+                        self.put_attachment(name=name, **info)
+                    else:
+                        delete_names.append(name)
+            for name in delete_names:
+                self.delete_attachment(name)
```
snopoke: Why does this have to happen outside of `atomic_blobs` context?

corehq/apps/app_manager/models.py
snopoke: I've been bitten by this before.

snopoke:
:boot: :building_construction:

snopoke (approved):

corehq/blobs/mixin.py
millerdev: The code in the `atomic_blobs` context is executed [just before](https://github.com/dimagi/commcare-hq/blob/c334bd9ed7f1c8dd7a8ac8329237349a68d4b3a2/corehq/blobs/mixin.py#L206) the [doc is saved](https://github.com/dimagi/commcare-hq/blob/c334bd9ed7f1c8dd7a8ac8329237349a68d4b3a2/corehq/blobs/mixin.py#L207). This is outside the context so they are deleted after successful save. If we deleted them inside the context then they would be deleted even if the save failed.

It is possible that an exception could be raised just outside the context (after the doc is saved) before all deletes have gone through, which would result in orphaned blobs. I think that's not very likely, and better to err on the side of keeping data we don't need than deleting data we might need.

millerdev:
One other thing I forgot to ask about this: With the form attachments migration there were ES indexes to update, but it looks like case attachments are not indexed in ES. Does that sound right, or should this PR be updating ES indexes too?

snopoke:
The only case index that has attachments indexed is `reportcase` but I don't think it needs any changes for this: https://github.com/dimagi/commcare-hq/blob/624a9fc206d39d79761f8ab1cd2cc05e32733a4e/corehq/pillows/mappings/reportcase_mapping.py#L66-L66

snopoke:
This is GTG from me, holding off on merge in case you want to wait for @dannyroberts to do a pass.

millerdev:
Thanks @snopoke. I'm going to wait on @dannyroberts review before merging.

corehq/apps/app_manager/models.py
dannyroberts: Very true. When I designed `jsonobject` I was both blessed with having couchdbkit's interface already designed for me to match exactly, and cursed with the blemishes of that design. woh, woh

corehq/ex-submodules/pillowtop/tests/test_elasticsearch.py
dannyroberts: I'm not sure I get the `- 2` part

dannyroberts (approved):

corehq/ex-submodules/pillowtop/tests/test_elasticsearch.py
millerdev: Random different port than the configured one. This is not a great feature of the way this test is written because for any other random port there could be another service running that would cause an incorrect failure.

PR #14627, opened 2017-01-23
Attempt to fix intermittent app manager blank screen

orangejenny:
Blank screen is happening when window is loaded before the load event is attached to it. [jQuery docs](https://api.jquery.com/ready/) imply that `ready` maps to `DomContentLoaded`, which is the equivalent of `document.readyState === "interactive"`, whereas it looks like the window load event maps to `document.readyState === "complete"`...so, if the window has loaded already, the handler is never going to run, so call `statechange`.

https://developer.mozilla.org/en-US/docs/Web/API/Document/readyState#readystatechange_as_an_alternative_to_DOMContentLoaded_event

@proteusvacuum / @benrudolph


proteusvacuum:
no tests, so merging.

benrudolph:
i would've thought the solution here is to use `document.onreadystatechange` and have an `if` clause for `"interactive"` and one for `"complete"`

benrudolph:
oh i see what you're doing

PR #14631, opened 2017-01-23
quick sms export backend code

emord:
@benrudolph @gcapalbo 

I started looking at this today, but I'll be on interrupt, so probably won't get back to it until after home week. I think this will be pretty easy. It could probably use a bit of refactoring of the exports framework though, so this might be good for Ben to review and think about.

Can be tested like this
```
from corehq.apps.export.models.new import SMSExportInstance, SMSExportDataSchema
from corehq.apps.export.export import get_export_file
instance = SMSExportInstance.generate_instance_from_schema(SMSExportDataSchema.get_latest_export_schema('demo', None, None))
f = get_export_file([instance], [])
f.file.path # where the export is saved locally
```


corehq/apps/export/models/new.py
```diff
@@ -792,6 +800,76 @@ def _new_from_schema(cls, schema):
         )
 
 
+class SMSExportInstance(ExportInstance):
+    type = SMS_EXPORT
+    identifier = SMS_EXPORT
+
+    @classmethod
+    def _new_from_schema(cls, schema):
+        return cls(
```
emord: This should be moved outside this file and likely just be a json structure

corehq/apps/export/models/new.py
```diff
@@ -792,6 +800,76 @@ def _new_from_schema(cls, schema):
...
+                            label="Message",
+                            item=ExportItem(
+                                path=[PathNode(name='text')]
+                            ),
+                            selected=True,
+                        ),
+                        ExportColumn(
+                            label="Type",
+                            item=ExportItem(
+                                path=[PathNode(name='workflow')]
```
emord: This will need a new transform to implement this https://github.com/dimagi/commcare-hq/blob/je/sms-export/corehq/apps/reports/standard/sms.py#L289-L305

corehq/apps/export/models/new.py
```diff
@@ -1607,6 +1698,36 @@ def _add_to_group_schema(group_schema, path_start, prop, app_id, app_version):
         return schema
 
 
+class SMSExportDataSchema(ExportDataSchema):
```
emord: It's possible that this could be removed with some refactoring. All this provides at the moment is the domain.

corehq/apps/export/models/new.py
```diff
@@ -1607,6 +1698,36 @@ def _add_to_group_schema(group_schema, path_start, prop, app_id, app_version):
...
+    @classmethod
+    def schema_version(cls):
+        return SMS_DATA_SCHEMA_VERSION
+
+    @classmethod
+    def _get_inferred_schema(cls, domain, identifier):
+        return None
+
+    @classmethod
+    def _get_current_app_ids_for_domain(cls, domain, app_id):
```
emord: I didn't like that these needed to be implemented here. They are called in the base classes generate schema class.

@benrudolph I was thinking that this framework could be pretty useful in the future if we want to move other reports to the export framework (like worker activity or whatever). What do you think about splitting up the base classes into a `ExportDataSchema` and `AppDependentExportDataSchema`? I wasn't sure how non app dependent exports should fit into this framework

corehq/apps/export/models/new.py
benrudolph: this is interesting. yeah this does feel a lot like "static" generation vs. "dynamic" generation. a json structure somewhere would make sense to me

corehq/apps/export/models/new.py
benrudolph: 👍 transforms can be fairly easily added

corehq/apps/export/models/new.py
benrudolph: I don't think you actually have to implement this. instead of overriding the `_new_from_schema` private method, i would override the `generate_schema_from_builds` method that would just return the schema. then you shouldn't need to touch any of `ExportInstance` functions or overwrite any of these methods. some of these constructs don't make as much sense now that we're not using apps. so the naming of `generate_schema_from_builds` is kind of bad.

i wouldn't be too opposed to splitting out `ExportDataSchema` to `AppDependentExportDataSchema`, however i still think we'd need a split between forms and cases, so i'm not sure how much that is buying us. 

what do you think? overriding just the generate_schema function i think would save us writing a decent amount of code

corehq/apps/export/models/new.py
```diff
@@ -792,6 +800,76 @@ def _new_from_schema(cls, schema):
         )
 
 
+class SMSExportInstance(ExportInstance):
+    type = SMS_EXPORT
+    identifier = SMS_EXPORT
```
benrudolph: not sure if this is correct here. `identifier` is used to specify the `xmlns` or `case_type` in an export. i think it may be Ok to have this be `None` if we don't need to further drilldown in the export

benrudolph:
nice, this is exciting. i don't doubt this'll probably take some thinking about because we definitely only thought about this in terms of forms and cases (one of the downsides of the previous framework was that it was so generalized it was hard to know what was going on and we only used it for forms and cases). it doesn't look like it'll be too bad though

corehq/apps/export/transforms.py
```diff
@@ -19,6 +19,37 @@ def owner_id_to_display(owner_id, doc):
 def case_id_to_case_name(case_id, doc):
     return _cached_case_id_to_case_name(case_id)
 
+
+def workflow_transform(workflow, doc):
+    from corehq.apps.sms.models import (
```
dimagimon: F401 'WORKFLOW_FORWARD' imported but unused

corehq/apps/export/models/new.py
```diff
@@ -792,6 +800,76 @@ def _new_from_schema(cls, schema):
...
+    def _new_from_schema(cls, schema):
+        return cls(
+            domain=schema.domain,
+            tables=[
+                TableConfiguration(
+                    label="Messages",
+                    columns=[
+                        RowNumberColumn(),
+                        ExportColumn(
+                            label="Contact ID",
```
gcapalbo: the current export also has a column for Contact Type, which takes the value of one of:

Mobile Worker
Case
Web User
Unknown

corehq/apps/export/transforms.py
gcapalbo: we'll need to include WORKFLOW_FORWARD in the relevant_workflows below, looks like it's actually also missing in the existing export process

corehq/apps/export/models/new.py
```diff
@@ -792,6 +800,76 @@ def _new_from_schema(cls, schema):
...
+                            item=ExportItem(
+                                path=[PathNode(name='date')]
+                            ),
+                            selected=True,
+                        ),
+                        ExportColumn(
+                            label="Username",
+                            item=ExportItem(
+                                path=[PathNode(name='couch_recipient')],
+                                transform=USERNAME_TRANSFORM,
```
gcapalbo: does this transform handle cases too? this id could point to a case, mobile worker, web user, or be None. couch_recipient_doc_type gives the doc type that this id points to

corehq/apps/export/models/new.py
```diff
@@ -1730,6 +1823,25 @@ def _add_to_group_schema(group_schema, path_start, prop, app_id, app_version):
...
+            only_process_current_builds=False, task=None):
+        return cls(domain=domain)
+
+    @classmethod
+    def schema_version(cls):
+        return SMS_DATA_SCHEMA_VERSION
+
+    @staticmethod
+    def get_latest_export_schema(domain, app_id, case_type):
+        return SMSExportDataSchema(domain=domain)
```
benrudolph: since you don't actually return the saved schema, this will create a new schema on every request to an export

corehq/apps/export/models/new.py
```diff
@@ -915,6 +924,77 @@ def get_filters(self):
...
 
+class SMSExportInstance(ExportInstance):
+    type = SMS_EXPORT
+    identifier = None
+
+    @classmethod
+    def _new_from_schema(cls, schema):
+        return cls(
+            domain=schema.domain,
+            tables=[
```
benrudolph: nit: the other instance handle system properties by defining them here: https://github.com/dimagi/commcare-hq/blob/master/corehq/apps/export/system_properties.py and then they get inserted in this function: https://github.com/dimagi/commcare-hq/blob/master/corehq/apps/export/models/new.py#L710.

corehq/apps/export/models/new.py
```diff
@@ -915,6 +924,77 @@ def get_filters(self):
...
+                        ExportColumn(
+                            label="Type",
+                            item=ExportItem(
+                                path=[PathNode(name='workflow')],
+                                transform=WORKFLOW_TRANSFORM,
+                            ),
+                            selected=True,
+                        ),
+                    ],
+                    selected=True
```
benrudolph: this tableconfiguration doesn't seem to have a `path` attribute. that should be required (didn't actually enforce at the db level, but should have). if it is set the `MAIN_TABLE` it will automatically be selected

corehq/apps/export/models/new.py
benrudolph: another reason to define it there is so that when you add system properties, the export will automatically add those regardless of whether or not you have a saved version (handled in the `__insert_system_properties` function)

gcapalbo:
@emord i just realized, would it be hard to also include the `MESSAGE_LOG_METADATA` toggle in this export? if it's enabled for the user requesting the export we would need to also include the `couch_id` of each `SMS` in the export

emord:
Going to close this out and open a new PR and rebase on master

corehq/apps/export/transforms.py
emord: https://github.com/dimagi/commcare-hq/pull/14823/commits/b730c519e142ee4eb98a3d536c7806fae739e35a

corehq/apps/export/models/new.py
emord: https://github.com/dimagi/commcare-hq/pull/14823/files#diff-9087195bf99fb8ba2fda775b0556aceaR944

corehq/apps/export/models/new.py
emord: changed it to the owner transform https://github.com/dimagi/commcare-hq/pull/14823/files#diff-9087195bf99fb8ba2fda775b0556aceaR947

corehq/apps/export/models/new.py
emord: That was intentional. Since these are static theres no reason to save it to the DB and it means there's no extra db calls

corehq/apps/export/models/new.py
emord: https://github.com/dimagi/commcare-hq/pull/14823/files#diff-9087195bf99fb8ba2fda775b0556aceaR777

corehq/apps/export/models/new.py
emord: It wasn't automatically selected when I tested it without `selected=True` https://github.com/dimagi/commcare-hq/pull/14823/files#diff-9087195bf99fb8ba2fda775b0556aceaR955

emord:
@gcapalbo added the metadata in hte new pr

PR #14633, opened 2017-01-24
Restored case finding mobile v1

kkrampa:
@czue

### DIFF
--- a/custom/enikshay/ucr/reports/case_finding_mobile_v2.json
+++ b/custom/enikshay/ucr/reports/case_finding_mobile_v2.json
@@ -76,9 +76,33 @@
                 "calculate_total": false
             },
             {
+                "display": "Suspects with positive test results",
+                "type": "field",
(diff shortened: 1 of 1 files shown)


custom/enikshay/ucr/reports/case_finding_mobile_v2.json
```diff
@@ -0,0 +1,416 @@
+{
```
kkrampa: Intentionally left and not added to settings

czue: why?

kkrampa: I guess they want to use this in further version of app but I can remove this if you want.

czue: no it's fine, was just curious.

czue (approved):

PR #14137, opened 2016-12-06
App Preview a/b test

biyeun:
Adds ab test for preview
adds analytics relevant to preview AB test
to activate ab test we have to put the appropriate date we want to set as the onboarding date in localsettings, so there will be a separate ansible deploy once this goes through

@benrudolph

PR #14639, opened 2017-01-24
Live preview A/B test not setting property

orangejenny:
Followup for https://github.com/dimagi/commcare-hq/pull/14137/commits/ccefab721b30864c9bc2443538d5cc147dbb4e6e ...`_kmq.push` and `analytics.workflow` take slightly different arguments. This is probably why we haven't been seeing data on the test come through Kiss.

The event name isn't relevant here, just made one up.

@benrudolph / @biyeun 
cc @mkangia


orangejenny:
tests passing

PR #14641, opened 2017-01-24
Add get_reporting_groups ucr expression

NoahCarnahan:
@nickpell. Code should be pretty self-explanatory.


nickpell (approved):

czue:
cool! what's the use case for this?

NoahCarnahan:
Mike O was asking if it existed and since `get_case_sharing_groups` already existed, I figured this would only take a few minutes to add.

PR #14642, opened 2017-01-24
Revert "Move case attachments to blob db"

emord:
Reverts dimagi/commcare-hq#14624

@millerdev


emord:
Reverted after offlining with @millerdev due to an issue where sql domains with case attachemtns could not submit

PR #14643, opened 2017-01-25
Create "/lookup_tables/" versions of all "/fixtures/" urls.

NoahCarnahan:
A few days ago Sarah asked for this on field/dev. Seemed like a pretty easy thing to fix so I thought I'd give it a try.

This PR preserves the existing "/fixtures/" urls, but also adds a "/lookup_tables/" url for each existing "fixtures" url. Also, the previous url names all refer to the new "/lookup_tables/" urls, while the old ones are put within a namespace to avoid collisions.

I decided to take this approach instead of using redirects because it seemed slightly performant, but   perhaps it would be considered better practice to use redirects instead.

@nickpell code buddy
@snopoke might have some opinions about weather this is the right approach or not

### DIFF
--- a/urls.py
+++ b/urls.py
@@ -65,7 +65,11 @@ domain_specific = [
     url(r'^case/', include('corehq.apps.hqcase.urls')),
     url(r'^case/', include('corehq.apps.case_search.urls')),
     url(r'^cloudcare/', include('corehq.apps.cloudcare.urls')),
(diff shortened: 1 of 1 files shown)


snopoke:
I think if the end goal is to remove the `fixtures` URLs (which I assume it is) then redirects are a better option otherwise we'll continue to have people bookmarking the old URLs.

PR #14646, opened 2017-01-25
Fix loading race condition

benrudolph:
@orangejenny finally got to the bottom of this damn bug. essentially the new behavior in jquery (or web browsers) is that the readyState for an iframe can change to `complete` even if the javascript in the `ready` event in the iframe hasn't finished loading.

this combats that by moving the JS out of the ready event and to the bottom of the page. then the `complete` event is accurate and we're free to send messages to the preview app window.

cc: @emord


orangejenny:
This looks like it was fun to figure out...

PR #14647, opened 2017-01-25
Remove USE_FORMPLAYER_FRONTEND toggle

benrudolph:
@czue this was followup from the PR where i bumped it to 100%. we need to add these domains when this gets merged. will coordinate with interrupt. all the domains with a check box need to be added to the toggle

- [x] 'hsph-betterbirth'		
- [x] 'figo-ppiud-srilanka'	
- [x] 'broadreach-sa'	
- [x] 'goal-global'	
- [ ] 'ipm-senegal'		
- [x] 'madla-malaria'	
- [x] 'myrice'
- [x] 'pact'		
- [x] 'icrc-almanach' <- for swiss

(this'll probably break a bunch of tests)
cc: @emord

### DIFF
--- a/corehq/apps/app_manager/templates/app_manager/v1/form_view_base.html
+++ b/corehq/apps/app_manager/templates/app_manager/v1/form_view_base.html
@@ -46,7 +46,7 @@
         }
     </style>
 
-    {% if request|toggle_enabled:"USE_FORMPLAYER_FRONTEND"%}
+    {% if not request|toggle_enabled:"USE_OLD_CLOUDCARE"%}
         <script src="{% static 'cloudcare/js/formplayer/utils/util.js' %}"></script>
     {% endif %}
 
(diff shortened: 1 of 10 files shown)


emord:
I think if you want to avoid the battles with travis of fix 1 test, restart test, fix next test that breaks you can take `stop` out of https://github.com/dimagi/commcare-hq/blob/master/.travis.yml#L34

benrudolph:
👍 good call

czue:
huh, this seems a bit more annoying than I expected. don't feel obligated to do this on my account, though does feel like a better end state to be in, i think. 100% toggle also seems fine though if this doesn't feel worth it.

benrudolph:
✅

emord:
@benrudolph added these domains to prod. will add to swiss once I deploy

benrudolph:
thanks j$

PR #14649, opened 2017-01-25
Release app preview to everyone else

benrudolph:
@orangejenny @wpride this rolls out app preview to 20% of domains. if it is an onboarding domain, it will _never_ show app preview. i think this is worth the sacrifice since i don't expect a ton of domains to fall into this category. if it becomes an issue, we can address it. would also be good to know how long we have to be in this state

cc: @emord

### DIFF
--- a/corehq/apps/app_manager/views/formdesigner.py
+++ b/corehq/apps/app_manager/views/formdesigner.py
@@ -27,6 +27,7 @@ from corehq.apps.app_manager.util import (
     app_callout_templates,
     is_usercase_in_use,
 )
+from corehq.apps.cloudcare.utils import should_show_preview_app
 from corehq.apps.fixtures.fixturegenerators import item_lists_by_domain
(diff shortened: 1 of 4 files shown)


benrudolph:
changed my mind, by passed the onboarding check if you are a dimagi user. will this mess with the ab test data? my guess is there's already messy data since the current status is anyone with the toggle enabled can use app preview on onboarding domains

orangejenny:
I think we're okay from the test perspective - you're right that the data is already, ideally we'd filter dimagi users out of the test altogether (not sure if that's possible or if we're just going to live with the messiness).

Should this also update https://github.com/dimagi/commcare-hq/blob/master/corehq/apps/app_manager/views/releases.py#L238 ?

benrudolph:
i don't think the other spot matters for our purposes very much. we're only tracking dimagi users and those should have the toggle explicitly enabled

PR #14651, opened 2017-01-25
adding url param to mark atypical users

calellowitz:
@gcapalbo This is from the third request in https://docs.google.com/document/d/1IiK3kj-wlHaOdokTP2RQ9f1ehTXiJkf0VCpthc5wvRc/edit#heading=h.edog63lkny5l

The idea is that users that sign up for hq via the job application or similar processes will be given urls to the signup page that include the internal parameter. we can then mark them in hubspot as atypical users and exclude them from our workflows

### DIFF
--- a/corehq/apps/registration/static/registration/js/new_user.ko.js
+++ b/corehq/apps/registration/static/registration/js/new_user.ko.js
@@ -186,6 +186,7 @@ hqDefine('registration/js/new_user.ko.js', function () {
                 eula_confirmed: self.eulaConfirmed(),
                 phone_number: _private.getPhoneNumberFn() || self.phoneNumber(),
                 xform: defaults.xform,
+                atypical_user: defaults.atypical_user
             };
         };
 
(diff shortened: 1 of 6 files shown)


corehq/apps/users/models.py
```diff
@@ -2012,6 +2012,7 @@ class WebUser(CouchUser, MultiMembershipMixin, CommCareMobileContactMixin):
     login_attempts = IntegerProperty(default=0)
     attempt_date = DateProperty()
     fcm_device_token = StringProperty()
+    atypical_user = BooleanProperty(default=False)
```
gcapalbo: might be worth a small comment describing what an `atypical_user` is

gcapalbo (approved):

corehq/apps/registration/static/registration/js/new_user.ko.js
```diff
@@ -186,6 +186,7 @@ hqDefine('registration/js/new_user.ko.js', function () {
                 eula_confirmed: self.eulaConfirmed(),
                 phone_number: _private.getPhoneNumberFn() || self.phoneNumber(),
                 xform: defaults.xform,
+                atypical_user: defaults.atypical_user
```
dimagimon: Missing trailing comma. (comma-dangle)

PR #14654, opened 2017-01-25
tools for auditing ES data

snopoke:
This is part of the ES reliability work Phase 1:
https://docs.google.com/document/d/1kULVsYdqB4Qo_FhHcTj66-xIYHzYWGHoSL5Euej_u6w/edit#heading=h.xdllplxs4tpe

I decided not to build off https://github.com/dimagi/commcare-hq/pull/14259 but rather re-purpose code that's aready in the repo which was used for the dump / load validation.

Differences from #14259:
* doesn't cover all ES indices yet (since this is still investigative)
* does the breakdowns by doc type


corehq/util/markup.py
```diff
@@ -39,3 +44,79 @@ def inner(text):
 
 shell_red = _shell_color_template('31')
 shell_green = _shell_color_template('32')
+
+
+class SimpleTableWriter(object):
```
benrudolph: ❤️ this

benrudolph (approved):
👍 from me, still seems like we'll get false positives when checking the `Only in primary` but think that's fine for now

snopoke:
@benrudolph not concerned about false positives yet but that will definitely be an issue for any automated tasks. I did see your comment on https://github.com/dimagi/commcare-hq/pull/14259 which seems like a good approach to try out.

czue:
very cool! 

would we ever make a web UI for this?

snopoke:
We could make a web UI. I think we can consider this down the line.

PR #14655, opened 2017-01-25
Redirect report builder reports to the paywall on community plans

emord:
http://manage.dimagi.com/default.asp?246184

Before if a user had a bookmark to the report, they could still access the report page. This redirects them to paywall home, but should still allow them to get to any custom built UCRs

@kaapstorm 

buddies @benrudolph @snopoke

### DIFF
--- a/corehq/apps/userreports/reports/view.py
+++ b/corehq/apps/userreports/reports/view.py
@@ -21,7 +21,7 @@ from dimagi.utils.modules import to_function
 from django.conf import settings
 from django.contrib import messages
 from django.core.urlresolvers import reverse
-from django.http import HttpResponse, Http404, HttpResponseBadRequest
+from django.http import HttpResponse, Http404, HttpResponseBadRequest, HttpResponseRedirect
 from django.utils.translation import ugettext as _, ugettext_noop
 from braces.views import JSONResponseMixin
 from corehq.apps.locations.permissions import conditionally_location_safe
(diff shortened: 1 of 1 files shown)


benrudolph (approved):
👢 :building_construction:

benrudolph:
actually looks like it was purposely killed

kaapstorm:
:ambulance: :construction_worker_man:

corehq/apps/userreports/reports/view.py
```diff
@@ -124,8 +125,12 @@ def domain(self):
...
     @conditionally_location_safe(has_location_filter)
     def dispatch(self, request, *args, **kwargs):
-        original = super(ConfigurableReport, self).dispatch(request, *args, **kwargs)
-        return original
+        if self.spec.report_meta.created_by_builder and has_report_builder_access(request):
+            original = super(ConfigurableReport, self).dispatch(request, *args, **kwargs)
+            return original
+        else:
+            from corehq.apps.userreports.views import paywall_home
+            return HttpResponseRedirect(paywall_home(self.domain))
```
NoahCarnahan: I think this is causing non-report builder UCRs to all be redirected to the report builder paywall. This view is used for both hand built UCRs and report builder UCRs, so the hand built ones fail `self.spec.report_meta.created_by_builder`, which result in the user getting sent to the paywall.
@kaapstorm I believe this is the issue that you are trying to solve in https://github.com/dimagi/commcare-hq/pull/14678

PR #14656, opened 2017-01-25
Fix botched readme edit

NoahCarnahan:
@nickpell


NoahCarnahan:
self merging because docs

PR #14659, opened 2017-01-25
Kill settings base template

orangejenny:
Noticed that everything in `settings/base_template.html` already exists in `settings/base.html`.

@mkangia


benrudolph (approved):
huh thought i killed this a long time ago. nice that it's gone for good now!

PR #14662, opened 2017-01-26
Send invitations to CommCare in domain's language

kaapstorm:
[FB 246203](http://manage.dimagi.com/default.asp?246203)

@emord cc buddies @proteusvacuum @czue

### DIFF
--- a/corehq/apps/domain/utils.py
+++ b/corehq/apps/domain/utils.py
@@ -110,4 +110,4 @@ def guess_domain_language(domain_name):
     """
     domain = Domain.get_by_name(domain_name)
     counter = Counter([app.default_language for app in domain.applications()])
-    return counter.most_common(1)[0][0]
+    return counter.most_common(1)[0][0] if counter else 'en'
--- a/corehq/apps/domain/utils.py
+++ b/corehq/apps/domain/utils.py
@@ -1,4 +1,4 @@
-from collections import defaultdict
+from collections import Counter
 import os
 import re
 
@@ -109,12 +109,5 @@ def guess_domain_language(domain_name):
     the language code of the most common default language across apps.
     """
     domain = Domain.get_by_name(domain_name)
-    language_count = defaultdict(lambda: 0)
-    lang = 'en'
-    last_max = 0
-    for app in domain.applications():
-        language_count[app.default_language] += 1
-        if language_count[app.default_language] > last_max:
-            lang = app.default_language
-            last_max = language_count[app.default_language]
-    return lang
+    counter = Counter([app.default_language for app in domain.applications()])
+    return counter.most_common(1)[0][0]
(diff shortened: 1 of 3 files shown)


corehq/apps/domain/utils.py
```diff
@@ -100,3 +101,20 @@ def get_domain_url_slug(hr_name, max_length=25, separator='-'):
             text += separator + word
 
     return text[:max_length]
+
+
+def guess_domain_language(domain_name):
+    """
+    A domain does not have a default language, but its apps do. Return
+    the language code of the most common default language across apps.
```
czue: I wonder if domains should have a default language....

corehq/apps/domain/utils.py
```diff
@@ -100,3 +101,20 @@ def get_domain_url_slug(hr_name, max_length=25, separator='-'):
...
+    """
+    domain = Domain.get_by_name(domain_name)
+    language_count = defaultdict(lambda: 0)
+    lang = 'en'
+    last_max = 0
+    for app in domain.applications():
+        language_count[app.default_language] += 1
+        if language_count[app.default_language] > last_max:
+            lang = app.default_language
+            last_max = language_count[app.default_language]
```
czue: (not blocking) I think you could do something clever with the [`Counter`](https://docs.python.org/2/library/collections.html#collections.Counter) class to simplify this into a one-liner if you were feeling inspired.

czue: e.g. `Counter([app.lang for app in apps]).most_common(1)[0][0]` (not tested)

czue (approved):
looks good!

corehq/apps/domain/utils.py
czue: (just musing, think the product implications of doing this would be out of scope of interrupt)

corehq/apps/domain/utils.py
kaapstorm: Nice! I was not aware of `Counter`!

corehq/apps/domain/utils.py
```diff
@@ -109,12 +109,5 @@ def guess_domain_language(domain_name):
...
-    lang = 'en'
-    last_max = 0
-    for app in domain.applications():
-        language_count[app.default_language] += 1
-        if language_count[app.default_language] > last_max:
-            lang = app.default_language
-            last_max = language_count[app.default_language]
-    return lang
+    counter = Counter([app.default_language for app in domain.applications()])
+    return counter.most_common(1)[0][0]
```
czue: will this crash if there are no applications?

czue: (could be a good test case to add)

kaapstorm: doh. ...

proteusvacuum (approved):

PR #14663, opened 2017-01-26
Don't 500 when a dev wants to change their subscription locally

kaapstorm:
Follows the same pattern as [`utils.get_customer_cards()`](https://github.com/dimagi/commcare-hq/blob/ef6a15aaa82b7ceee6042a60119931ebf552c669/corehq/apps/accounting/utils.py#L204-L204)

@nickpell cc buddies @proteusvacuum @czue

### DIFF
--- a/corehq/apps/accounting/models.py
+++ b/corehq/apps/accounting/models.py
@@ -2854,7 +2854,14 @@ class StripePaymentMethod(PaymentMethod):
 
     @property
     def all_cards(self):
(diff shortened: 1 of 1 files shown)


czue (approved):
makes sense to me. probably not worth it now, but if we find ourselves using this pattern a lot we could consider making a decorator kind of like [`json_error`](https://github.com/dimagi/commcare-hq/blob/master/corehq/util/view_utils.py#L45)

nickpell:
👍

PR #14666, opened 2017-01-26
Check ReportBuilderView permission to show button

kaapstorm:
[FB 245748](http://manage.dimagi.com/default.asp?245748)

Checks the same permission as the [`ReportBuilderView.dispatch()` decorator](https://github.com/dimagi/commcare-hq/blob/2e62f859a5c3678a2a4340e1539257525b315c27/corehq/apps/userreports/views.py#L245-L245) to determine whether to show the "Edit Report" button.

@emord cc @czue @proteusvacuum

### DIFF
--- a/corehq/apps/userreports/util.py
+++ b/corehq/apps/userreports/util.py
@@ -4,6 +4,7 @@ import hashlib
 from django.conf import settings
 from django.utils.translation import ugettext_lazy as _
 
+from corehq.apps.users.models import Permissions
 from corehq.util.soft_assert import soft_assert
 from corehq import privileges, toggles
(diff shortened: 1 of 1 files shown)


emord (approved):

czue:
does this actually work with location restrictions? i would have thought you would have to mimic something in `LocationAccessMiddleware` or possibly just check the `access_all_locations` permission.

kaapstorm:
You're right @czue. The user that this change hides the button from does actually have permission to edit the report. -- At least, their role says "View Report", but the Edit Report page doesn't kick them out with a 403 -- which may also be a problem, but it's a different problem. :confused: I'll close this.

PR #14671, opened 2017-01-26
Support hand-off workflow

esoergel:
Adds a bunch of questions to the internal project info page and allows GS staff to trigger a hand-off to support email.
https://docs.google.com/document/d/1uXOUxnTmi70-0Ei9V0yrrVQL3Suxgh-Nr3TGPDRIg2s/edit
@millerdev 

Definitely easiest :fishing_pole_and_fish: :x: :shark:

### DIFF
--- a/corehq/apps/domain/forms.py
+++ b/corehq/apps/domain/forms.py
@@ -822,82 +822,82 @@ class PrivacySecurityForm(forms.Form):
 
 
 class DomainInternalForm(forms.Form, SubAreaMixin):
-    sf_contract_id = CharField(label=ugettext_noop("Salesforce Contract ID"), required=False)
-    sf_account_id = CharField(label=ugettext_noop("Salesforce Account ID"), required=False)
-    initiative = forms.MultipleChoiceField(label=ugettext_noop("Initiative"),
+    sf_contract_id = CharField(label="Salesforce Contract ID", required=False)
+    sf_account_id = CharField(label="Salesforce Account ID", required=False)
+    initiative = forms.MultipleChoiceField(label="Initiative",
                                            widget=forms.CheckboxSelectMultiple(),
                                            choices=tuple_of_copies(DATA_DICT["initiatives"], blank=False),
                                            required=False)
     workshop_region = CharField(
-        label=ugettext_noop("Workshop Region"),
+        label="Workshop Region",
         required=False,
-        help_text=ugettext_noop("e.g. US, LAC, SA, Sub-Saharan Africa, Southeast Asia, etc."))
+        help_text="e.g. US, LAC, SA, Sub-Saharan Africa, Southeast Asia, etc.")
     self_started = ChoiceField(
-        label=ugettext_noop("Self Started?"),
+        label="Self Started?",
         choices=tf_choices('Yes', 'No'),
         required=False,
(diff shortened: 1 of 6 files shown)


corehq/apps/domain/templates/domain/email/support_handoff.html
```diff
@@ -0,0 +1,28 @@
+<p>
+    Hi There! My name is Robin, and I run the support team at Dimagi. I work with {{ contact_name }} and look forward to helping you use CommCare successfully.
```
millerdev: Do we really want to hard-code Robin's name in here?

corehq/apps/domain/views.py
```diff
@@ -2527,6 +2536,23 @@ def page_context(self):
             'areas': dict([(a["name"], a["sub_areas"]) for a in settings.INTERNAL_DATA["area"]]),
         }
 
+    def send_handoff_email(self):
+        partner_contact = self.internal_settings_form.cleaned_data['partner_contact']
+        dimagi_contact = self.internal_settings_form.cleaned_data['dimagi_contact']
+        recipients = filter(None, [partner_contact, dimagi_contact])
```
millerdev: Why is the filter necessary here? It looks like they are both required fields (see form `clean()` method).

millerdev (commented):
A test node failed due to timeout. Kicked.

millerdev (approved):

corehq/apps/domain/views.py
esoergel: Good point.  When I first wrote it, they weren't required.  Will update.

corehq/apps/domain/templates/domain/email/support_handoff.html
esoergel: Yeah, it's a little weird.  Honestly, I find the whole automated email thing a little weird, but they wanted a personal touch, I think.  Have you got an alternative suggestion?

corehq/apps/domain/forms.py
```diff
@@ -822,82 +822,82 @@ def save(self, domain):
 
 
 class DomainInternalForm(forms.Form, SubAreaMixin):
-    sf_contract_id = CharField(label=ugettext_noop("Salesforce Contract ID"), required=False)
-    sf_account_id = CharField(label=ugettext_noop("Salesforce Account ID"), required=False)
-    initiative = forms.MultipleChoiceField(label=ugettext_noop("Initiative"),
```
esoergel: I was adding a bunch of strings and really didn't want to translate those, but it felt weird to leave these strings translated.  This is a form that's only accessible by Dimagi employees and is meant to be filled out in English.  Translations are done either by Dimagi staff or by a paid external translator who charges by the word.  Any change to a string requires a retranslation, and the translators don't have a way of knowing the urgency of translating a string.In short, a human has to do work for every string that's marked for translations, and for internal pages, we don't get any benefit from that work.

corehq/apps/domain/templates/domain/email/support_handoff.html
millerdev: Short of adding a setting to HQ to set the support contact's name, not really anything great. The other option is to make the language less personal, but it sounds like people didn't want that.

PR #14672, opened 2017-01-26
Externalize js: Management Commands

orangejenny:
Low-hanging fruit. Review commit by commit.

@mkangia


orangejenny:
tests pass

snopoke:
Is this page useful? Could we just kill it?

orangejenny:
@snopoke Yeah, good point. Analytics say it's accessed occasionally, but I suspect those aren't real usages and are people semi-randomly clicking because it's an item in the top menu dropdown. Checked with A team and didn't find anyone who uses it. Will kill.

snopoke:
👍☠

PR #14673, opened 2017-01-27
Combobox for form entry

benrudolph:
@wpride had a little fun implementing the combobox for web entry :sweat_smile: i had initially tried to do this with atwho.js, but it just isn't the right tool for what we're trying to do. the jquery typeahead library has some great capabilities. you can read the docs on combobox/behavior here: https://confluence.dimagi.com/display/commcarepublic/Advanced+CommCare+Android+Formatting#AdvancedCommCareAndroidFormatting-SingleSelect"ComboBox"

screen shot 2017-01-27 at 9 49 08 am
screen shot 2017-01-27 at 9 48 52 am
screen shot 2017-01-27 at 9 49 03 am

cc: @emord @snopoke

design: @orangejenny

### DIFF
--- a/corehq/apps/cloudcare/static/cloudcare/js/form_entry/entrycontrols_full.js
+++ b/corehq/apps/cloudcare/static/cloudcare/js/form_entry/entrycontrols_full.js
@@ -321,6 +321,78 @@ DropdownEntry.prototype.onPreProcess = function(newValue) {
     }
 };
 
+function ComboboxEntry(question, options) {
+    var self = this;
+    EntrySingleAnswer.call(this, question, options);
+
+    self.matchType = options.matchType;
+    self.templateType = 'combobox';
+    self.helpText = function() { return 'Combobox'; };
+    self.options = _.map(question.choices(), function(choice, idx) {
+        return { display: choice, id: idx + 1 };
+    });
+
+    self.afterRender = function() {
+        var selector = '#' + self.entryId;
+        if (self.answer()) {
+            // The - 1 is to offset for the one-indexed select questions
+            $(selector).val(question.choices()[self.answer() - 1]);
+        }
+        $.typeahead({
+            input: selector,
+            minLength: 1,
+            hint: true,
+            cancelButton: false,
+            filter: function(item) {
+                return ComboboxEntry.filter(this.rawQuery, item, self.matchType);
+            },
+            source: {
+                data: self.options,
+            },
+            callback: {
+                onResult: function(node, query) {
+                    self.rawAnswer(query);
+                },
+            },
+        });
+    };
+}
+
+ComboboxEntry.filter = function(query, d, matchType) {
+    if (matchType === Formplayer.Const.COMBOBOX_MULTIWORD) {
+        // Multiword filter, matches the start of any word in the choice
+        var words = d.display.split(' ');
+        return _.any(words, function(word) { return word.startsWith(query); });
+    } else if (matchType === Formplayer.Const.COMBOBOX_FUZZY) {
+        // Fuzzy filter, matches if any of the query is in choice
+        return d.display.includes(query);
+    } else {
+        // Standard filter, matches only start of word
+        return d.display.startsWith(query);
+    }
+};
+
+ComboboxEntry.prototype = Object.create(EntrySingleAnswer.prototype);
+ComboboxEntry.prototype.constructor = EntrySingleAnswer;
+ComboboxEntry.prototype.onPreProcess = function(newValue) {
+    var value;
+    if (newValue === Formplayer.Const.NO_ANSWER || newValue === '') {
+        this.answer(Formplayer.Const.NO_ANSWER);
+        this.question.error(null);
+        return;
+    }
+
+    value = _.find(this.options, function(d) {
+        return d.display === newValue;
+    });
+    if (value) {
+        this.answer(value.id);
+        this.question.error(null);
+    } else {
+        this.question.error(gettext('Not a valid choice'));
+    }
+};
+
 $.datetimepicker.setDateFormatter({
     parseDate: function (date, format) {
         var d = moment(date, format);
@@ -545,6 +617,7 @@ function getEntry(question) {
     var options = {};
     var isNumeric = false;
     var isMinimal = false;
+    var isCombobox = false;
 
     var displayOptions = _getDisplayOptions(question);
     var isPhoneMode = ko.utils.unwrapObservable(displayOptions.phoneMode);
@@ -574,9 +647,24 @@ function getEntry(question) {
     case Formplayer.Const.SELECT:
         if (question.style) {
             isMinimal = ko.utils.unwrapObservable(question.style.raw) === Formplayer.Const.MINIMAL;
+            isCombobox = ko.utils.unwrapObservable(question.style.raw).startsWith(Formplayer.Const.COMBOBOX);
         }
         if (isMinimal) {
             entry = new DropdownEntry(question, {});
+        } else if (isCombobox) {
+
+            entry = new ComboboxEntry(question, {
+                /*
+                 * The appearance attribute is either:
+                 *
+                 * combobox
+                 * combobox multiword
+                 * combobox fuzzy
+                 *
+                 * The second word designates the matching type
+                 */
+                matchType: question.style.raw().split(' ')[1],
+            });
         } else {
             entry = new SingleSelectEntry(question, {});
         }
--- a/corehq/apps/cloudcare/static/cloudcare/js/form_entry/entrycontrols_full.js
+++ b/corehq/apps/cloudcare/static/cloudcare/js/form_entry/entrycontrols_full.js
@@ -336,14 +336,21 @@ function ComboboxEntry(question, options) {
     self.matchType = options.matchType;
     self.lengthLimit = Infinity;
     self.templateType = 'str';
-    self.helpText = function() { return 'Combobox'; };
+
     self.options = ko.computed(function() {
         return _.map(question.choices(), function(choice, idx) {
             return { name: choice, id: idx + 1 };
         });
     });
+    self.options.subscribe(function () {
+        self.renderAtwho();
+        if (!self.isValid(self.rawAnswer())) {
+            self.question.error(gettext('Not a valid choice'));
+        }
+    });
+    self.helpText = function() { return 'Combobox'; };
 
-    self.afterRender = function() {
+    self.renderAtwho = function() {
         var $input = $('#' + self.entryId);
         $input.atwho({
             at: '',
@@ -362,6 +369,19 @@ function ComboboxEntry(question, options) {
                 },
             },
         });
+    };
+    self.isValid = function(value) {
+        if (!value) {
+            return true;
+        }
+        return _.include(
+            _.map(self.options(), function(option) { return option.name }),
+            value
+        );
+    };
+
+    self.afterRender = function() {
+        self.renderAtwho();
     };
 }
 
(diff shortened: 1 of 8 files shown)


corehq/apps/cloudcare/static/cloudcare/js/form_entry/entrycontrols_full.js
```diff
@@ -321,6 +321,78 @@ DropdownEntry.prototype.onPreProcess = function(newValue) {
     }
 };
 
+function ComboboxEntry(question, options) {
+    var self = this;
+    EntrySingleAnswer.call(this, question, options);
```
dimagimon: 'EntrySingleAnswer' is not defined. (no-undef)

corehq/apps/cloudcare/static/cloudcare/js/form_entry/entrycontrols_full.js
```diff
@@ -321,6 +321,78 @@ DropdownEntry.prototype.onPreProcess = function(newValue) {
...
+    } else if (matchType === Formplayer.Const.COMBOBOX_FUZZY) {
+        // Fuzzy filter, matches if any of the query is in choice
+        return d.display.includes(query);
+    } else {
+        // Standard filter, matches only start of word
+        return d.display.startsWith(query);
+    }
+};
+
+ComboboxEntry.prototype = Object.create(EntrySingleAnswer.prototype);
```
dimagimon: 'EntrySingleAnswer' is not defined. (no-undef)

corehq/apps/cloudcare/static/cloudcare/js/form_entry/entrycontrols_full.js
```diff
@@ -321,6 +321,78 @@ DropdownEntry.prototype.onPreProcess = function(newValue) {
...
+ComboboxEntry.prototype.constructor = EntrySingleAnswer;
```
dimagimon: 'EntrySingleAnswer' is not defined. (no-undef)

corehq/apps/cloudcare/static/cloudcare/js/form_entry/entrycontrols_full.js
```diff
@@ -321,6 +321,86 @@ DropdownEntry.prototype.onPreProcess = function(newValue) {
...
+/**
+ * The ComboboxEntry is an advanced android formatting entry. It is enabled
+ * when the user specifies combobox in the appearance attributes for a
+ * single select question.
+ *
+ * Docs: https://confluence.dimagi.com/display/commcarepublic/Advanced+CommCare+Android+Formatting#AdvancedCommCareAndroidFormatting-SingleSelect"ComboBox"
+ */
+function ComboboxEntry(question, options) {
+    var self = this;
+    EntrySingleAnswer.call(this, question, options);
```
dimagimon: 'EntrySingleAnswer' is not defined. (no-undef)

wpride:
nice! cc @amstone326

wpride (approved):

wpride:
@benrudolph test failure looks real

orangejenny:
Design looks fine to me, though I'm curious what was needed that's in jquery typeahead but not in atwho/select2.

benrudolph:
@orangejenny atwhojs is really only good for progressively adding autocomplete fields. like slack emojs. it's hard to restrict it to _just_ one slack emoji and nothing else, which is what i'm attempting to do here. and the space that it adds after you enter an autocomplete field is killer

orangejenny:
Vellum uses atwho as its "autocomplete" (it allowed us to get rid of jquery ui without having to introduce select2 or yet another library) and handles both of those issues. To play with it, add a lookup table and click on the `Lookup Table Data` node; the `Value Field` and `Display Text Field` use it in this manner.

The config: https://github.com/dimagi/Vellum/blob/master/src/atwho.js#L54-L74

amstone326:
Hey @benrudolph @wpride I apologize for not making this clearer initially, but I had intended for you to be able to reuse the `ComoboxFilterRule` classes that are in core when implementing this on formplayer, so that the filter behaviors between this and mobile would be automatically aligned without having to reimplement any filter code. Does it seem like a big lift to integrate that in here? There's 1 implementation of a filter rule for each type of combobox, and the class exposes a `choiceShouldBeShown(String choice, CharSequence textEntered)` method that returns whether or not a specific choice should be included in the visible answer options based on the text entered by the user.

benrudolph:
@orangejenny interesting... that does seem to do what i want. i can try and switch it back to that, thanks (though depending on aliza's comment it may not make sense)

@amstone326 we can do that, however it would require an ajax request every time someone types a character. in this version we don't make a request to the server until they've entered a valid choice. i'm kind of torn as to whether being DRY here is better since it requires a fairly big lift given the filter logic isn't crazy complicated

amstone326:
@benrudolph Ah ok, did not realize that. It sounds like that's not the best idea then, but as a compromise I think it would be worthwhile to try to make the separate implementations equivalent, which I think they aren't quite right now. The "standard" one looks right to me, but for the others:

-The "fuzzy" implementation on mobile uses a [util method we have](https://github.com/dimagi/commcare-core/blob/master/src/main/java/org/commcare/cases/util/StringUtils.java#L69-L137) for fuzzy-matching strings, which means that options that are close-but-not-quite matches will be shown. Would it make sense to just straight up copy over the logic of that code to formplayer and use it here? I'm particularly worried about this one having parity because this is the version that UATBC has elected to use, and they were specifically concerned about having the fuzzy-matching behavior because they said they will be using the combo box with answer choices that are commonly misspelled (complex location names).

-For the "multiword" implementation, I could be misreading the code but I think what you implemented is sort of the inverse of what I have. On mobile, what it does is split the _query_ up into words and then shows any answer choice which contains all of those words. Whereas it looks like you're splitting the _answer choice_ up into words and then checking if the query _starts with_ any of those words. Let me know if I'm misreading that one though.

corehq/apps/cloudcare/static/cloudcare/js/form_entry/entrycontrols_full.js
```diff
@@ -321,6 +321,86 @@ DropdownEntry.prototype.onPreProcess = function(newValue) {
...
+    } else if (matchType === Formplayer.Const.COMBOBOX_FUZZY) {
+        // Fuzzy filter, matches if any of the query is in choice
+        return d.display.includes(query);
+    } else {
+        // Standard filter, matches only start of word
+        return d.display.startsWith(query);
+    }
+};
+
+ComboboxEntry.prototype = Object.create(EntrySingleAnswer.prototype);
```
dimagimon: 'EntrySingleAnswer' is not defined. (no-undef)

corehq/apps/cloudcare/static/cloudcare/js/form_entry/entrycontrols_full.js
```diff
@@ -321,6 +321,86 @@ DropdownEntry.prototype.onPreProcess = function(newValue) {
...
+ComboboxEntry.prototype.constructor = EntrySingleAnswer;
```
dimagimon: 'EntrySingleAnswer' is not defined. (no-undef)

corehq/apps/cloudcare/static/cloudcare/js/form_entry/entrycontrols_full.js
```diff
@@ -321,6 +321,86 @@ DropdownEntry.prototype.onPreProcess = function(newValue) {
...
+        this.answer(value.id);
+        this.question.error(null);
+    } else {
+        this.question.error(gettext('Not a valid choice'));
+    }
+};
+
 $.datetimepicker.setDateFormatter({
     parseDate: function (date, format) {
         var d = moment(date, format);
```
dimagimon: 'moment' is not defined. (no-undef)

benrudolph:
@orangejenny -- https://github.com/dimagi/commcare-hq/pull/14673/commits/8eeed4cc10afcad6ddde00af29706cbcb1b1cf98 switched back to atwho. for some reason i constantly miss new settings on the docs page even though i think i've read them all. thanks for pointing me in the right direction

@amstone326 -- https://github.com/dimagi/commcare-hq/pull/14673/commits/099ef9e4812bbecdbd748ff217aad0763b33419d ok took another go at the filters. i think i'm doing what you've said now. i decided not to always return `false` if the query was less than 3 for the fuzzy match though because i thought that would then mean you could never match 3 letter choices? unless you guys have more logic that matches when it's an exact match

corehq/apps/cloudcare/static/cloudcare/js/form_entry/entrycontrols_full.js
```diff
@@ -321,6 +321,112 @@ DropdownEntry.prototype.onPreProcess = function(newValue) {
...
+    self.renderAtwho = function() {
+        var $input = $('#' + self.entryId);
+        $input.atwho({
+            at: '',
+            data: self.options(),
+            maxLen: Infinity,
+            tabSelectsMatch: false,
+            suffix: '',
+            callbacks: {
+                filter: function(query, data, searchKey) {
```
dimagimon: 'searchKey' is defined but never used (no-unused-vars)

corehq/apps/cloudcare/static/cloudcare/js/form_entry/entrycontrols_full.js
```diff
@@ -321,6 +321,112 @@ DropdownEntry.prototype.onPreProcess = function(newValue) {
...
+                    return _.filter(data, function(item) {
+                        return ComboboxEntry.filter(query, item, self.matchType);
+                    });
+                },
+                matcher: function(flag, subtext, should_startWithSpace) {
```
dimagimon: 'should_startWithSpace' is defined but never used (no-unused-vars)

corehq/apps/cloudcare/static/cloudcare/js/form_entry/entrycontrols_full.js
```diff
@@ -321,6 +321,112 @@ DropdownEntry.prototype.onPreProcess = function(newValue) {
...
+                },
+            },
+        });
+    };
+    self.isValid = function(value) {
+        if (!value) {
+            return true;
+        }
+        return _.include(
+            _.map(self.options(), function(option) { return option.name }),
```
dimagimon: Missing semicolon. (semi)

corehq/apps/cloudcare/static/cloudcare/js/form_entry/entrycontrols_full.js
```diff
@@ -321,6 +321,112 @@ DropdownEntry.prototype.onPreProcess = function(newValue) {
...
+    } else if (matchType === Formplayer.Const.COMBOBOX_FUZZY) {
+        // Fuzzy filter, matches if query is "close" to answer
+        return window.Levenshtein.get(d.name.toLowerCase(), query.toLowerCase()) <= 2;
+    } else {
+        // Standard filter, matches only start of word
+        return d.name.startsWith(query);
+    }
+};
+
+ComboboxEntry.prototype = Object.create(EntrySingleAnswer.prototype);
```
dimagimon: 'EntrySingleAnswer' is not defined. (no-undef)

orangejenny:
Sweet, thank you for reworking it with atwho, admit I tend to be draconian about adding js libraries.

amstone326:
@benrudolph Re: your last comment, yea I'm using this https://github.com/dimagi/commcare-core/blob/master/src/main/java/org/javarosa/core/model/FuzzyMatchFilterRule.java#L56-L60 to check for either a fuzzy match OR exact substring, so I think it makes sense to implement Levenshtein the standard way and then check for either as well. (Side note: it looks like you're just importing a pre-existing javascript implementation of Levenshtein in this PR, so how did you change the behavior anyway? My impression was that Levenshtein distance is a standard formula).

Other than that, looks good to me 👍 Thanks for changing

benrudolph:
@amstone326 cool, that's easy to add.

> Side note: it looks like you're just importing a pre-existing javascript implementation of Levenshtein in this PR, so how did you change the behavior anyway? My impression was that Levenshtein distance is a standard formula

I don't understand this part. I am using a pre existing library since it's just a standard algo as you mentioned. I didn't change the functionality of it at all

amstone326:
@benrudolph Sorry, realized I misunderstood your last comment, can ignore that question.

benrudolph:
good to merge

PR #14676, opened 2017-01-27
less aggresive case type rename

snopoke:
Only change the case type of an advanced form action if:
* action is a load action and reference the edited module
* action is an open action in a form belonging to the edited module
  or there are no modules with the old case type in the app


czue (approved):
thanks! kind of surprised this hasn't bitten anyone before.

benrudolph (approved):

PR #14677, opened 2017-01-27
Chunk es query

benrudolph:
@esoergel @NoahCarnahan curious what you guys think of this strategy. It's fairly hacky, but this will chunk your ES query by a term. This way if you're icds and have 100,000 user_ids, we don't insert all of them in the actual query. This doesn't address the problem of loading 100,000 user_ids into memory, but should help with ReadTimeouts when querying ES.

cc: @snopoke @emord

### DIFF
--- a/corehq/apps/es/utils.py
+++ b/corehq/apps/es/utils.py
@@ -52,6 +52,8 @@ def chunk_query(query, term, chunk_size=1000):
         an array of ESQuerys with the chunked term.
 
     '''
+    from corehq.apps.userreports.expressions.getters import recursive_lookup
+
     query = deepcopy(query)
 
     path_to_term = _chunk_query([], None, query.es_query, term)
@@ -61,7 +63,7 @@ def chunk_query(query, term, chunk_size=1000):
     if path_to_term is None:
         return [query]
 
-    term_list = _dict_lookup(path_to_term, query.es_query)
+    term_list = recursive_lookup(query.es_query, path_to_term)
 
     # Cannot chunk term that is not a list
     if not isinstance(term_list, list):
@@ -72,21 +74,11 @@ def chunk_query(query, term, chunk_size=1000):
         chunked_query = deepcopy(query)
 
         # Set query to the chunk
-        _dict_lookup(path_to_term[:-1], chunked_query.es_query)[term] = terms
+        recursive_lookup(chunked_query.es_query, path_to_term[:-1])[term] = terms
 
         chunked_queries.append(chunked_query)
 
     return chunked_queries
-
-
-def _dict_lookup(path, dict_):
-    partial = dict_
-    for part in path:
-        try:
-            partial = partial[part]
-        except IndexError:
-            return None
-    return partial
 
 
 def _chunk_query(path_to_term, key, es_query, term):
--- a/corehq/apps/es/utils.py
+++ b/corehq/apps/es/utils.py
@@ -1,3 +1,6 @@
+from copy import deepcopy
+
+
 def values_list(hits, *fields, **kwargs):
     """modeled after django's QuerySet.values_list"""
     flat = kwargs.pop('flat', False)
@@ -32,3 +35,61 @@ def flatten_field_dict(results, fields_property='fields'):
             new_val = val[0]
         field_dict[key] = new_val
     return field_dict
+
+
+def chunk_query(query, term, chunk_size=1000):
+    query = deepcopy(query)
+
+    path_to_term = _chunk_query([], None, query.es_query, term)
+    chunked_queries = []
+
+    # No path to term could be found, return just the query
+    if path_to_term is None:
+        return [query]
+
+    term_list = _dict_lookup(path_to_term, query.es_query)
+
+    # Cannot chunk term that is not a list
+    if not isinstance(term_list, list):
+        return [query]
+
+    for index in xrange(0, len(term_list), chunk_size):
+        terms = term_list[index:index + chunk_size]
+        chunked_query = deepcopy(query)
+
+        # Set query to the chunk
+        _dict_lookup(path_to_term[:-1], chunked_query.es_query)[term] = terms
+
+        chunked_queries.append(chunked_query)
+
+    return chunked_queries
+
+
+def _dict_lookup(path, dict_):
+    partial = dict_
+    for part in path:
+        try:
+            partial = partial[part]
+        except IndexError:
+            return None
+    return partial
+
+
+def _chunk_query(path_to_term, key, es_query, term):
+    if key is not None:
+        path_to_term.append(key)
+    if key == term:
+        return path_to_term
+
+    if isinstance(es_query, dict):
+        for key, partial_query in es_query.iteritems():
+            result = _chunk_query(list(path_to_term), key, partial_query, term)
+            if result:
+                return result
+    elif isinstance(es_query, list):
+        for index, partial_query in enumerate(es_query):
+            result = _chunk_query(list(path_to_term), index, partial_query, term)
+            if result:
+                return result
+
+    return None
--- a/corehq/apps/es/utils.py
+++ b/corehq/apps/es/utils.py
@@ -38,6 +38,20 @@ def flatten_field_dict(results, fields_property='fields'):
 
 
 def chunk_query(query, term, chunk_size=1000):
+    '''
+    This takes a ESQuery and a term and then chunks the query into identical queries
+    except having chunked the term. For example, if you have a query that has 100,000
+    queries and the chunk_size is 1000, it will return 100 queries.
+
+    :param query: An ESQuery
+    :param term: A term that you wish to be chunked, such as 'form.meta.userID'
+    :param chunk_size: The chunk size
+
+    :return: If it cannot find anything to chunk or the chunked term is not an array,
+        it will just return the same query wrapped in an array. Otherwise, it will return
+        an array of ESQuerys with the chunked term.
+
+    '''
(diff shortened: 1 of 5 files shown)


dimagimon:
There are 23 errors:

* corehq/apps/es/tests/test_utils.py, line 9 - E121 continuation line under-indented for hanging indent
* corehq/apps/es/tests/test_utils.py, line 10 - E121 continuation line under-indented for hanging indent
* corehq/apps/es/tests/test_utils.py, line 11 - E121 continuation line under-indented for hanging indent
* corehq/apps/es/tests/test_utils.py, line 12 - E121 continuation line under-indented for hanging indent
* corehq/apps/es/tests/test_utils.py, line 17 - E121 continuation line under-indented for hanging indent
* corehq/apps/es/tests/test_utils.py, line 18 - E121 continuation line under-indented for hanging indent
* corehq/apps/es/tests/test_utils.py, line 19 - E121 continuation line under-indented for hanging indent
* corehq/apps/es/tests/test_utils.py, line 20 - E121 continuation line under-indented for hanging indent
* corehq/apps/es/tests/test_utils.py, line 21 - E121 continuation line under-indented for hanging indent
* corehq/apps/es/tests/test_utils.py, line 22 - E121 continuation line under-indented for hanging indent
* corehq/apps/es/tests/test_utils.py, line 26 - E121 continuation line under-indented for hanging indent
* corehq/apps/es/tests/test_utils.py, line 27 - E121 continuation line under-indented for hanging indent
* corehq/apps/es/tests/test_utils.py, line 31 - E121 continuation line under-indented for hanging indent
* corehq/apps/es/tests/test_utils.py, line 32 - E121 continuation line under-indented for hanging indent
* corehq/apps/es/tests/test_utils.py, line 36 - E121 continuation line under-indented for hanging indent
* corehq/apps/es/tests/test_utils.py, line 37 - E121 continuation line under-indented for hanging indent
* corehq/apps/es/tests/test_utils.py, line 38 - E121 continuation line under-indented for hanging indent
* corehq/apps/es/tests/test_utils.py, line 39 - E121 continuation line under-indented for hanging indent
* corehq/apps/es/tests/test_utils.py, line 46 - E121 continuation line under-indented for hanging indent
* corehq/apps/es/tests/test_utils.py, line 47 - E121 continuation line under-indented for hanging indent
* corehq/apps/es/tests/test_utils.py, line 48 - E121 continuation line under-indented for hanging indent
* corehq/apps/es/tests/test_utils.py, line 56 - E126 continuation line over-indented for hanging indent
* corehq/apps/es/tests/test_utils.py, line 57 - E122 continuation line missing indentation or outdented

snopoke (approved):

esoergel:
I haven't done a thorough review yet, but did you look at [our implementation](https://github.com/dimagi/commcare-hq/blob/master/corehq/apps/es/es_query.py#L220-L229) of the [scan/scroll api](https://www.elastic.co/guide/en/elasticsearch/guide/1.x/scan-scroll.html)?  The ES docs make it sound like that's the preferable way to paginate queries which return a large number of hits:
> The scan search type and the scroll API are used together to retrieve large numbers of documents from Elasticsearch efficiently, without paying the penalty of deep pagination.

I'm not too familiar with that API. @NoahCarnahan do you know if there are limitations that would prevent us from using that here?

emord:
@esoergel This is using the scroll api https://github.com/dimagi/commcare-hq/pull/14677/files#diff-3dca43ab3e970c335479aa007c944de6R161

What this is doing is chunking the actual filters, so if you have a report that queries 2000 users, it first queries forms/cases for the first 1000 and then queries the 2nd 1000. I think it's preventing a timeout on ES for when ICDS has 10000 users and it queires for all of those at the same time

esoergel:
Oh, I see you're using it in conjunction with `scroll` - can you elaborate on what that's about?  Is it so ES doesn't hold the whole thing in memory or something?

esoergel:
Okay, J$ and I just offlined about this and I think I understand the problem a little better.  This does seem like a reasonable solution.

corehq/apps/export/export.py
```diff
@@ -136,22 +138,35 @@ def get_export_file(export_instances, filters, progress_tracker=None):
...
+        user_id_term = 'user_id'
+    elif export_instance.type == FORM_EXPORT:
+        user_id_term = 'form.meta.userID'
+    else:
+        user_id_term = None
+
     # size here limits each scroll request, not the total number of results
-    return query.size(1000).scroll()
+    for chunked_query in chunk_query(query, user_id_term, chunk_size=USER_ID_CHUNK_SIZE):
+        yield chunked_query.size(1000).scroll()
```
esoergel: There's a fair bit of technical ES internals here, which might be nice to hide.  Have you considered wrapping `chunk_query` and `scroll` in a function call that handles all the chunking, results flattening, and returns a simple generator of docs?  The only disadvantage I can think of to that is if you DO want the results in chunks as well (if say, you had subsequent things you wanted batched).

benrudolph: that's a good idea, i was hesitant at first since i thought there'd be a lot of magic crammed into a function (chunking a term and scrolling) and they seemed like different things. thinking about it more i'm warming up to it

corehq/apps/es/utils.py
```diff
@@ -32,3 +35,75 @@ def flatten_field_dict(results, fields_property='fields'):
...
+
+        # Set query to the chunk
+        _dict_lookup(path_to_term[:-1], chunked_query.es_query)[term] = terms
+
+        chunked_queries.append(chunked_query)
+
+    return chunked_queries
+
+
+def _dict_lookup(path, dict_):
```
benrudolph: 👍

corehq/apps/export/export.py
```diff
@@ -136,22 +138,35 @@ def get_export_file(export_instances, filters, progress_tracker=None):
...
             # TODO: Don't get the docs multiple times if you don't have to
-            docs = _get_export_documents(export_instance, filters)
-            _write_export_instance(writer, export_instance, docs, progress_tracker)
+            for docs in _iter_export_documents(export_instance, filters):
+                _write_export_instance(writer, export_instance, docs, progress_tracker)
 
     return ExportFile(writer.path, writer.format)
 
 
-def _get_export_documents(export_instance, filters):
```
esoergel: Test failure looks real:
```
ERROR: Failure: ImportError (cannot import name _get_export_documents)
  File "/vendor/lib/python2.7/site-packages/nose/loader.py", line 418, in loadTestsFromName
    addr.filename, addr.module)
  File "/vendor/lib/python2.7/site-packages/nose/importer.py", line 47, in importFromPath
    return self.importFromDir(dir_path, fqname)
  File "/vendor/lib/python2.7/site-packages/nose/importer.py", line 94, in importFromDir
    mod = load_module(part_fqname, fh, filename, desc)
  File "/mnt/commcare-hq-ro/corehq/apps/export/tests/test_export_filters.py", line 7, in <module>
    from corehq.apps.export.export import (
ImportError: cannot import name _get_export_documents
```

corehq/apps/es/utils.py
```diff
@@ -32,3 +35,61 @@ def flatten_field_dict(results, fields_property='fields'):
...
+
+    # No path to term could be found, return just the query
+    if path_to_term is None:
+        return [query]
+
+    term_list = _dict_lookup(path_to_term, query.es_query)
+
+    # Cannot chunk term that is not a list
+    if not isinstance(term_list, list):
+        return [query]
```
NoahCarnahan: Should we make sure the chunked term is always part of an "or" statement? Would it ever make sense to chunk anything else?

NoahCarnahan (approved):
lgtm

corehq/apps/es/utils.py
benrudolph: good question, i think it could be chunked if there wasn't an OR filter (would only be one filter), but we certainly don't want to do it to an AND filter. don't think it'd be too hard to check that

benrudolph: actually that ends up being kind of tricky because there are situations where the path to the term could have an `AND` clause in it. see the example in the test: https://github.com/dimagi/commcare-hq/pull/14677/files#diff-154c27feeb92507c0618c4e130dbb73dR17

corehq/apps/export/export.py
benrudolph: in the end i decided against this. i don't want `chunk_query` to be something that people use. it does require quite a bit of knowledge about the actual ES query in order for it to work and has a lot of caveats. in the end i'd rather see this function removed for a better solution (at the database level) than have its usage spread

NoahCarnahan:
Tests are passing now. Shall I merge or did you want to discuss/think about this more?

corehq/apps/es/utils.py
esoergel: You could also take the reverse approach.  Currently this takes an existing query with a large terms filter and chunks apart the filter values to produce multiple queries.  You could instead assume that this will always be used for a top-level terms filter, then take an existing query without that large terms filter and produce multiple duplicate queries each with a chunk of the terms.  Something like:
```python
def chunk_query(query, field, values):
    for chunk in chunked(values):
        yield query.filter(filters.term(field, chunk))
```

benrudolph: offlined with ethan, decided on keeping the current implementation

benrudolph:
Actually i'm going to hold off on this for a bit. Just tested an icds export with a location that has loads of user ids and it went through (albeit slow). hesitant to add this code if it's not needed

PR #14638, opened 2017-01-24
assign correct report builder plans to contracted subscriptions

nickpell:
long-awaited fix to http://manage.dimagi.com/default.asp?244858

This makes sure that we assign the right software plan version with 5 reports included, rather than just the most recent one created by Ops.

recommend 🐡 

Follow-up will require fixing subscriptions affected by this bug.

@proteusvacuum @gcapalbo

PR #14678, opened 2017-01-27
Enterprise Plan gets report builder

kaapstorm:
I haven't been able to reproduced locally (and I'm not sure why) so I don't know whether this PR resolves the problem with [Enterprise Plan customers](https://www.commcarehq.org/a/airstanzania/settings/project/subscription/) not being able to view [report builder-built reports](https://www.commcarehq.org/a/airstanzania/reports/builder/subscribe/pricing/). 

This PR doesn't address the issue with Pro Plan customers [with report builder](https://www.commcarehq.org/a/tz-lad/settings/project/subscription/), who also can't view [their reports](https://www.commcarehq.org/a/tz-lad/reports/builder/subscribe/activating_subscription/). Is that because they have more than 5, @nickpell?

I assume this is related to PR #14638 

@nickpell cc @NoahCarnahan @proteusvacuum

### DIFF
--- a/corehq/apps/accounting/bootstrap/config/cchq_software_plan_bootstrap.py
+++ b/corehq/apps/accounting/bootstrap/config/cchq_software_plan_bootstrap.py
@@ -47,7 +47,7 @@ BOOTSTRAP_CONFIG = {
             FeatureType.SMS: dict(monthly_limit=0),
         }
     },
-    (SoftwarePlanEdition.ENTERPRISE, False, False): {
+    (SoftwarePlanEdition.ENTERPRISE, False, True): {
         'role': 'enterprise_plan_v0',
         'product_rate': dict(monthly_fee=Decimal('0.00')),
         'feature_rates': {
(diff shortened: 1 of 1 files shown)


NoahCarnahan:
Hey Norman, I think this might not actually be the fix. I think this might be the problematic PR though: https://github.com/dimagi/commcare-hq/pull/14655

I'm not sure about the tz-lad report, but the airstanzania one is not a report builder report. Therefore report builder access is unnecessary. Following up on 14655 so I can comment on the code itself...

kaapstorm:
Urgh. OK, let's revert #14655.

emord:
I think this is the fix for a separate but similar issue. Right now enterprise domains are being redirected when clicking "create new report". @nickpell I think this also needs a migration correct?

emord:
After offlining with @nickpell I believe that this PR is not needed. There's some legacy subscription workaround that needs to use a feature flag to use report builder.

PR #14681, opened 2017-01-27
Fix encoding issue with case detail mapping ids

orangejenny:
http://manage.dimagi.com/default.asp?246489


orangejenny:
@mkangia / @millerdev

benrudolph (approved):

kaapstorm (approved):

PR #14683, opened 2017-01-27
Better UCR Rebuilding

emord:
I moved over a report from mz-lad and it had some errors because the dynamic mapping didn't work. This was easier to fix that I thought it would be 

buddies @benrudolph @snopoke


czue (approved):
cool stuff. kicked build

PR #14688, opened 2017-01-29
[don't merge] playing with dropping custom apps tests

czue:
not even sure if this works anymore, but curious how it affects timings if it does.


czue:
this doesn't actually work. probably need to figure out how to remove them from nose.

PR #14689, opened 2017-01-29
migrated case property updates

nickpell:
Identified by Kriti and I in https://docs.google.com/spreadsheets/d/1qLTDb2K0x_8n_kUDy6t3-xRDiLw8uSMOJD90vmjnheE/edit#gid=0

@proteusvacuum @czue cc: @NoahCarnahan


custom/enikshay/nikshay_datamigration/models.py
```diff
@@ -286,8 +286,8 @@ class Outcome(models.Model):
     @property
     def hiv_status(self):
         return {
-            None: None,
-            'NULL': None,
+            None: 'unknown',
```
czue: are there lots of these in the data? it's going to cause everyone to show up on the "HIV not updated" task lists, which is maybe fine/good.

czue (approved):
nice this looks great. glad we caught the treatment outcome thing before running this!

custom/enikshay/nikshay_datamigration/models.py
proteusvacuum: When this is set to unknown or 'reactive', we need to make a drtb-hiv-referral case to the drtb-hiv location under the dto location.

nickpell: @proteusvacuum is this needed for 1.0?  Wondering if I can run the migration today and add the drtb-hiv-referral cases later or if that will break the 1.0 app.

proteusvacuum: I think we want this for 1.0

nickpell: https://github.com/dimagi/commcare-hq/pull/14770

nickpell: > are there lots of these in the data?

~16K

PR #14691, opened 2017-01-30
excluding mikolo from nimbus form validation

calellowitz:
@orangejenny 
Formplayer was crashing and burning (out of memory issues) so this unblocks ismaila's app building.


calellowitz:
@benrudolph

PR #14696, opened 2017-01-31
copy module config tweaks

czue:
follow up from https://github.com/dimagi/commcare-hq/pull/14347#issuecomment-269341432 and some minor bug fixes.

@orangejenny cc @proteusvacuum @kaapstorm

### DIFF
--- a/corehq/apps/app_manager/templates/app_manager/v2/module_view.html
+++ b/corehq/apps/app_manager/templates/app_manager/v2/module_view.html
@@ -269,6 +269,9 @@
                 </li>
               {% endif %}
             {% endfor %}
+            {% if details %}
+                <li><a href="#module-actions" data-toggle="tab">{% trans "Actions" %}</a></li>
+            {% endif %}
           {% endif %}
         {% endblock %}{# endblock tab-headers #}
       </ul>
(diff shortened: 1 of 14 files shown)


czue:
:white_check_mark:

corehq/apps/app_manager/templates/app_manager/v2/module_view.html
```diff
@@ -292,6 +295,11 @@
               </div>
             {% endif %}
           {% endfor %}
+          {% if details %}
+            <div class="tab-pane" id="module-actions">
+                {% include 'app_manager/v1/partials/module_actions.html' %}
```
orangejenny: This might as well include the v2 version of module_actions.html instead of v1. Not that it matters now since they're identical, but will be confusing if that changes.

orangejenny:
One minor comment, feel free to self-merge after that.

corehq/apps/app_manager/templates/app_manager/v2/module_view.html
czue: good catch. https://github.com/dimagi/commcare-hq/pull/14696/commits/4775c0b0fd15641c21870c9209fdd760211b796e

PR #14699, opened 2017-01-31
Reverse toggle

benrudolph:
@orangejenny thanks, definitely think we should do this

### DIFF
--- a/corehq/toggles.py
+++ b/corehq/toggles.py
@@ -1017,16 +1017,11 @@ LINKED_APPS = StaticToggle(
     [NAMESPACE_DOMAIN]
 )
 
-NIMBUS_FORM_VALIDATION = PredictablyRandomToggle(
-    'nimbus_form_validation',
-    'Use Nimbus to validate XForms',
+FORMTRANSLATE_FORM_VALIDATION = StaticToggle(
+    'formtranslate_form_validation',
+    'Use formtranslate to validate XForms',
     TAG_PRODUCT_PATH,
-    [NAMESPACE_DOMAIN],
-    randomness=1.0,
-    always_disabled=[
-        'icrc-almanach',
-        'mikolo'
-    ]
+    [NAMESPACE_DOMAIN]
 )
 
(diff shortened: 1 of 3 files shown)


orangejenny:
thanks for fixing test :)

benrudolph:
✅

corehq/toggles.py
```diff
@@ -1017,16 +1017,11 @@ def _commtrackify(domain_name, toggle_is_enabled):
...
-    'Use Nimbus to validate XForms',
+FORMTRANSLATE_FORM_VALIDATION = StaticToggle(
+    'formtranslate_form_validation',
+    'Use formtranslate to validate XForms',
     TAG_PRODUCT_PATH,
-    [NAMESPACE_DOMAIN],
-    randomness=1.0,
-    always_disabled=[
-        'icrc-almanach',
-        'mikolo'
```
czue: does this need to be coordinated with deployer to get added?

czue (approved):

corehq/toggles.py
benrudolph: more important that the swiss one gets added when swiss is deployed. we added mikolo due to memory issues in web apps. i think that has now been fixed. thanks for reminding

benrudolph:
@emord @kaapstorm if swiss gets deployed during your interrupt, can you add `icrc-almanach` to this toggle?

corehq/toggles.py
emord: If there's more of these, maybe it would make sense to implement always_enabled arument?

benrudolph: yeah, i think the functionality that would be nice to have is something like "initial" where we seed the flag with a few domains, but we're still able to turn them off without a code change. `always_disabled` is useful when it's a random toggle

benrudolph: yeah, i think the functionality that would be nice to have is something like "initial" where we seed the flag with a few domains, but we're still able to turn them off without a code change. `always_disabled` is useful when it's a random toggle

benrudolph:
handling the swiss deploy myself

PR #14700, opened 2017-01-31
Handle HTML entities in inline edit widget

orangejenny:
http://manage.dimagi.com/default.asp?246264

@mkangia / @benrudolph


benrudolph (approved):
clever

PR #14705, opened 2017-01-31
set last_update when updating checkpoint

snopoke:
Since we introduced the max delay for updating checkpoints we've basically been updating them on every change. Oops.


czue (approved):
classic

czue:
might need to merge master for the build to finish.

PR #14710, opened 2017-02-01
support save to case in app summary, exports, data dictionary, etc.

czue:
depends on https://github.com/dimagi/Vellum/pull/850

probably easiest read by commit.


czue:
@emord @millerdev #dsahackathon cc @proteusvacuum @kaapstorm

corehq/apps/app_manager/models.py
```diff
@@ -818,6 +818,7 @@ class FormBase(DocumentSchema):
     form_links = SchemaListProperty(FormLink)
     schedule_form_id = StringProperty()
     custom_instances = SchemaListProperty(CustomInstance)
+    case_references_data = DictProperty()
```
czue: not sure if there are any implications of this moving up to the base class that I should worry about / test. it seemed to be fine everywhere i tried it.

corehq/apps/app_manager/models.py
```diff
@@ -1115,6 +1136,45 @@ def get_module(self):
     def get_case_type(self):
         return self._parent.case_type
 
+    def _add_save_to_case_questions(self, form_questions, app_case_meta):
+        def _make_save_to_case_question(path):
+            from corehq.apps.reports.formdetails.readable import FormQuestionResponse
+            # todo: this is a hack - just make an approximate save-to-case looking question
```
czue: eventually we should support save to case questions as a first class thing that `get_questions` knows how to parse and work with, but this seemed unneccessary to get this working. if we want those to ever be supported in form exports or form summaries we would need to do that work and then update this to use the real question.

corehq/apps/app_manager/models.py
czue: Also I really want to define a schema for this instead of using a DictProperty. I don't think there are any reasons not to do that but I didn't get around to it.

corehq/apps/app_manager/models.py
```diff
@@ -2722,14 +2783,6 @@ def get_case_updates(self, case_type):
 
         return updates.union(scheduler_updates)
 
-    @property
-    def case_references(self):
```
czue: @millerdev I assume there was no major reason you added these other than just trying to be defensive, yeah? [context](https://github.com/dimagi/commcare-hq/commit/62aaf78675f8d96605016e9acf52ff7d060f7c50)

corehq/apps/export/tests/test_export_data_schema.py
```diff
@@ -657,6 +657,49 @@ def test_build_with_advanced_app(self):
...
+                }
+            }
+        }
+        cls.current_app = factory.app
+        cls.current_app.save()
+
+    @classmethod
+    def tearDownClass(cls):
+        cls.current_app.delete()
+        CaseExportDataSchema.get_latest_export_schema(cls.domain, cls.current_app._id, cls.case_type).delete()
```
benrudolph: nit: there's a function to just delete all export schemas, but this does the trick too

corehq/apps/export/tests/test_export_data_schema.py
```diff
@@ -657,6 +657,49 @@ def test_build_with_advanced_app(self):
...
+
+    def test(self):
```
benrudolph: nit: a more descriptive test name

benrudolph (approved):
👏 happy you got this working for exports!

corehq/apps/export/tests/test_export_data_schema.py
czue: this is presumably less work, though, no?

corehq/apps/export/tests/test_export_data_schema.py
czue: https://github.com/dimagi/commcare-hq/pull/14710/commits/877be947df5e203d6cb5d97afff05edf127bced9

czue:
the only reason there's a "don't merge" label on this is just to confirm with @millerdev or @emord that there isn't something I'm missing that should be considered / QA'd that I haven't already tested. otherwise this is g2g from my perspective. will need to be merged and deployed at the same time as https://github.com/dimagi/Vellum/pull/850

corehq/apps/app_manager/models.py
```diff
@@ -5761,7 +5815,7 @@ def get_profile_setting(self, s_type, s_id):
     def has_careplan_module(self):
         return any((module for module in self.modules if isinstance(module, CareplanModule)))
 
-    @quickcache(['self.version'])
+    @quickcache(['self._id', 'self.version'])
```
czue: heh, this is a funny bug.

corehq/apps/app_manager/models.py
czue: also had to change a test because of this: https://github.com/dimagi/commcare-hq/pull/14710/commits/4ef9ca348b124dd428669d0fbec1786904f9eb7a

corehq/apps/app_manager/models.py
```diff
@@ -1103,6 +1112,18 @@ def case_list_modules(self):
     def is_case_list_form(self):
         return bool(self.case_list_modules)
 
+    def get_save_to_case_updates(self, case_type):
+        """
+        Get a flat list of case property names from save to case questions
+        """
+        updates = set()
+        if self.case_references_data and 'save' in self.case_references_data:
```
snopoke: could just be `if 'save' in self.case_references:`

czue:
took off label since I realized this is actually totally safe to merge as long as it comes before https://github.com/dimagi/Vellum/pull/850. still want @millerdev to weigh in on some comments though

corehq/apps/app_manager/models.py
snopoke: interesting, so this was using a single cache key for all apps HQ with the same version... that's baaaad

corehq/apps/app_manager/tests/test_case_meta.py
```diff
@@ -1,12 +1,15 @@
...
 from mock import patch
 from nose.tools import nottest
 
 from corehq.apps.app_manager.models import Application, Module, OpenCaseAction, ParentSelect, OpenSubCaseAction, \
     AdvancedModule, LoadUpdateAction, AdvancedOpenCaseAction, CaseIndex
+from corehq.apps.app_manager.tests.util import TestXmlMixin
 
 
-class CaseMetaTest(SimpleTestCase):
+class CaseMetaTest(SimpleTestCase, TestXmlMixin):
```
snopoke: :+1: tests!

snopoke (commented):
Looking good

corehq/apps/app_manager/models.py
czue: decided to take a crack at this in most recent changes. it's not perfect because the current structure of each question being a key doesn't lend itself well to being typed, but I still think it'll be better to work with.

czue:
I've done a fair amount of testing on this including writing a bunch of tests, searching for every possible usage of the property and switching between this and master using easy references and verifying everything is fine and backwards compatible, so I'm feeling pretty good about getting this in even though it's a bigish change. definitely open to dissenting opinions on that though.

corehq/apps/app_manager/models.py
```diff
@@ -792,6 +792,43 @@ def short_comment(self):
...
+    path = StringProperty()
+    properties = ListProperty(unicode)
+    create = BooleanProperty(default=False)
+    close = BooleanProperty(default=False)
+
+
+class CaseReferences(DocumentSchema):
+    _allow_dynamic_properties = False
+    load = DictProperty()
+    save = DictProperty()
```
emord: This change seems a little scary. We don't really have integration tests between HQ and Vellum so if there was a bug in vellum that caused this to not have the correct format, it wouldn't save and the user would be kind of screwed. With that said that is tested in Vellum's tests, so it shouldn't happen, but could be an extra point of failure.

emord (approved):
This change looks good to me. I think vellum's could be deployed first since the current HQ code only looks at `load` and doesn't check `save`

czue:
> I think vellum's could be deployed first since the current HQ code only looks at `load` and doesn't check `save`

Currently if the vellum change gets deployed HQ will choke on the extra "save" data [here](https://github.com/dimagi/commcare-hq/blob/master/corehq/apps/app_manager/views/forms.py#L758)

corehq/apps/app_manager/models.py
czue: yeah, agreed. this has already been the case historically because of [this validation](https://github.com/dimagi/commcare-hq/blob/master/corehq/apps/app_manager/views/forms.py#L758). I think [this set of tests](https://github.com/dimagi/commcare-hq/pull/14710/files#diff-c1311f89ab8e0af93474ab88b4019498R110) is the closest proxy to an integration test we have, but there definitely isn't an easy way to check if vellum starts changing things up.

snopoke: I think having this validated is great. If there were a bug in Vellum that produced the wrong format then we'd end up needing to do a data migration to fix it which seems like a more difficult situation to fix.

snopoke: could these not be `SchemaDictProperty`?

corehq/apps/app_manager/models.py
```diff
@@ -792,6 +792,43 @@ def short_comment(self):
         return self.comment if len(self.comment) <= 500 else self.comment[:497] + '...'
 
 
+class CaseLoadReference(DocumentSchema):
+    _allow_dynamic_properties = False
+    path = StringProperty()
+    properties = ListProperty(unicode)
```
snopoke: Just a note for a future world where all case management happens in the form builder this should also include `case_type`

snopoke (approved):

corehq/apps/app_manager/models.py
czue: > could these not be `SchemaDictProperty`?

I think they could, will try and update now.

dimagimon:
There are 6 errors:

* corehq/apps/app_manager/models.py, line 40 - F401 'copy' imported but unused
* corehq/apps/app_manager/tests/test_case_meta.py, line 92 - E121 continuation line under-indented for hanging indent
* corehq/apps/app_manager/tests/test_case_refs_from_fb.py, line 123 - E126 continuation line over-indented for hanging indent
* corehq/apps/app_manager/tests/test_case_refs_from_fb.py, line 133 - E121 continuation line under-indented for hanging indent
* corehq/apps/app_manager/tests/test_case_refs_from_fb.py, line 139 - E121 continuation line under-indented for hanging indent
* corehq/apps/app_manager/tests/test_case_refs_from_fb.py, line 141 - E121 continuation line under-indented for hanging indent

dimagimon:
There are 6 errors:

* corehq/apps/app_manager/models.py, line 40 - F401 'copy' imported but unused
* corehq/apps/app_manager/tests/test_case_meta.py, line 92 - E121 continuation line under-indented for hanging indent
* corehq/apps/app_manager/tests/test_case_refs_from_fb.py, line 123 - E126 continuation line over-indented for hanging indent
* corehq/apps/app_manager/tests/test_case_refs_from_fb.py, line 133 - E121 continuation line under-indented for hanging indent
* corehq/apps/app_manager/tests/test_case_refs_from_fb.py, line 139 - E121 continuation line under-indented for hanging indent
* corehq/apps/app_manager/tests/test_case_refs_from_fb.py, line 141 - E121 continuation line under-indented for hanging indent

corehq/apps/app_manager/models.py
czue: updated in https://github.com/dimagi/commcare-hq/pull/14710/commits/4b5657f21849bf6a31fdeaa04ddf377c019539df and added some doc strings to better explain the intent of these different classes/methods

dimagimon:
There are 5 errors:

* corehq/apps/app_manager/tests/test_case_meta.py, line 92 - E121 continuation line under-indented for hanging indent
* corehq/apps/app_manager/tests/test_case_refs_from_fb.py, line 123 - E126 continuation line over-indented for hanging indent
* corehq/apps/app_manager/tests/test_case_refs_from_fb.py, line 133 - E121 continuation line under-indented for hanging indent
* corehq/apps/app_manager/tests/test_case_refs_from_fb.py, line 139 - E121 continuation line under-indented for hanging indent
* corehq/apps/app_manager/tests/test_case_refs_from_fb.py, line 141 - E121 continuation line under-indented for hanging indent

corehq/apps/app_manager/models.py
```diff
@@ -792,6 +792,82 @@ def short_comment(self):
...
+    def get_load_references(self):
+        """
+        Returns a generator of `CaseLoadReference` objects containing all the load references.
+        """
+        for path, properties in self.load.items():
+            yield CaseLoadReference(path=path, properties=list(properties))
+
+    def get_save_references(self):
+        """
+        Returns a generator of `CaseSaveReferenceWithPath` objects containing all the load references.
```
millerdev: should say "containing all the _save_ references"

corehq/apps/app_manager/views/forms.py
```diff
@@ -755,6 +754,10 @@ def is_valid(value):
         }
     else:
         refs = json.loads(data.get('case_references', '{}'))
-    if set(refs) - {"load"} or not all(is_valid(v) for v in refs["load"].values()):
+
+    try:
+        references = CaseReferences.wrap(copy.deepcopy(refs))
```
millerdev: looks like `copy.deepcopy` can be removed

millerdev (approved):

corehq/apps/app_manager/views/forms.py
czue: good call, updated and added a test to convince myself. https://github.com/dimagi/commcare-hq/pull/14710/commits/237e1f89e82ce3a79cb4ece4d7da3a6f1e70eee5

corehq/apps/app_manager/models.py
czue: https://github.com/dimagi/commcare-hq/pull/14710/commits/613189a4bbc0e6e753e2e180454a47a3114e2c36

corehq/apps/app_manager/views/forms.py
```diff
@@ -1,3 +1,4 @@
+import copy
```
dimagimon: F401 'copy' imported but unused

corehq/apps/app_manager/views/forms.py
czue: actually i thought this comment was somewhere else, but removed it here too. https://github.com/dimagi/commcare-hq/pull/14710/commits/ca63db1e42a26cb806768da474174457b860b6e4

millerdev:
Kicked travis node, which appeared to have stalled after 20 min. Not sure what the issue was. The last few lines in the log were not helpful:
```
test_probono_properties (corehq.apps.analytics.tests.test_subscription_properties.TestSubscriptionProperties) ... ok
test_properties (corehq.apps.analytics.tests.test_subscription_properties.TestSubscriptionProperties) ... ok
test_basic (corehq.apps.api.tests.EWSLocationResourceTest) ... ok
corehq.apps.api.tests.ILSLocationResourceTest ... SKIP: divided-we-run: 4 not in range 03
```

millerdev (approved):

PR #14713, opened 2017-02-01
move case locking to encapsulate save

snopoke:
http://manage.dimagi.com/default.asp?246758
Was looking at these errors and wondering how they happen when I noticed that the per case locks are released before we do the case save.

Hopefully keeping the lock until after the save will get rid of these. This is particularly important for SQL domains that don't have a rev check.


snopoke:
Best viewed with [w=1](https://github.com/dimagi/commcare-hq/pull/14713/files?w=1)

czue (approved):
looks good. tests are timing out with what seem like anomalously long deltas. maybe somehow some locking getting triggered by test cases using the same case ids?

emord:
Looks like build was cancelled. may have to merge master?

corehq/form_processor/submission_post.py
```diff
@@ -139,36 +142,47 @@ def run(self):
...
-                    self.save_processed_models(xforms, case_stock_result)
-                    cases = case_stock_result.case_models
+            from casexml.apps.case.xform import get_and_check_xform_domain
+            domain = get_and_check_xform_domain(xforms[0])
+            if self.case_db:
+                assert self.case_db.domain == domain
+                case_db_cache = self.case_db
+                case_db_cache.cached_xforms.extend(xforms)
+            else:
+                case_db_cache = self.interface.casedb_cache(domain=domain, lock=True, deleted_ok=True, xforms=xforms)
```
dimagimon: E501 line too long (117 > 115 characters)

snopoke:
finally got to the bottom of this one. Required making the case db cache object re-entrant for when we close extension cases with another form triggered during form processing.

czue:
wow that's sneaky.

PR #14714, opened 2017-02-02
Set project desc to "not applicable" if N/A

kaapstorm:
[FB 246696](http://manage.dimagi.com/default.asp?246696): `domain` is not available/applicable to bugs reported from admin pages.

@emord cc @czue @proteusvacuum @nickpell


czue (approved):
:+1:

PR #14719, opened 2017-02-02
SQL ucr ledgers

emord:
@proteusvacuum This test was rough. Ended up stealing a lot of the setup from https://github.com/dimagi/commcare-hq/blob/6620b6e658d5e9206770938c22be08df789e16f1/corehq/apps/commtrack/tests/test_stock_report.py#L32

buddies @snopoke @benrudolph


corehq/apps/userreports/tests/test_indicators.py
```diff
@@ -526,56 +541,96 @@ def test_ledger_balances_indicator(self, get_values_by_product):
...
-            product = cls._make_product(product_code)
-            cls._make_stock_state(product, section, value)
+        super(TestGetValuesByProduct, cls).setUpClass()
+        cls.data = [
+            {"product": "coke", "section": "soh", "balance": 32},
+            {"product": "coke", "section": "consumption", "balance": 63},
+            {"product": "surge", "section": "soh", "balance": 85},
+            {"product": "fanta", "section": "soh", "balance": 11},
+        ]
+        product_ids = set(map(lambda p: p['product'], cls.data))
```
proteusvacuum: nit: I think you could use a set comprehension here
```py
{p['product'] for p in cls.data}
```

corehq/apps/userreports/indicators/__init__.py
```diff
@@ -104,32 +104,35 @@ def get_values(self, item, context=None):
...
-            section_id=ledger_section,
-            case_id=case_id,
-            sql_product__code__in=product_codes,
-        ).values_list('sql_product__code', 'stock_on_hand')
-        return defaultdict(lambda: 0, values_by_product)
+        ret = defaultdict(lambda: 0)
+        ledgers = LedgerAccessors(domain).get_ledger_values_for_case(case_id)
+        ledgers = filter(lambda l: l.section_id == ledger_section, ledgers)
+        ledgers = filter(lambda l: l.entry_id in product_codes, ledgers)
+        for ledger in ledgers:
```
proteusvacuum: I think these three lines will loop through each of the items in `ledgers` three times. You could instead:
```py
return {
    l.entry_id: l.stock_on_hand if l.stock_on_hand else 0 
    for l in ledgers 
    if l.section_id == ledger_section and l.entry_id in product_codes
}
```

proteusvacuum (approved):
Looks good! Few small comments, but non-blocking.

corehq/apps/userreports/indicators/__init__.py
```diff
@@ -104,32 +104,34 @@ def get_values(self, item, context=None):
...
+
+        return ret
 
     def get_columns(self):
         return map(self._make_column, self.product_codes)
 
     def get_values(self, item, context=None):
         case_id = self.case_id_expression(item)
-        values = self._get_values_by_product(self.ledger_section, case_id, self.product_codes)
+        values = self._get_values_by_product(self.ledger_section, case_id, self.product_codes, self.domain)
```
czue: other expressions like `RelatedDoc` use `context.root_doc['domain']` for this. I do like the idea of making the domain an explicit part of the `EvaluationContext`, though. one easy compromise could be to just make it a property that returns `context.root_doc['domain']` for now?

czue: or alternatively you could update the other references to follow this pattern. not sure whether there's any significant upside/downside of doing this in `FactoryContext` versus `EvaluationContext`. just thinking it would be nice if it were consistent across things.

corehq/apps/userreports/indicators/__init__.py
```diff
@@ -104,32 +104,34 @@ def get_values(self, item, context=None):
 
 class LedgerBalancesIndicator(ConfigurableIndicator):
```
czue: this is not at all related to your PR but just occurred to me while reviewing so writing it somewhere. it feels like what we should do is expose a `get_case_ledgers` expression that then can play nice with all the map/reduce stuff and would be more versatile for anyone doing anything off of ledgers. i haven't totally thought through whether that could be a replacement for this functionality (might be tricky with expanding columns), but seems like it would simplify things if we could use it.

(feel free to ignore this comment)

corehq/apps/userreports/indicators/__init__.py
emord: Nice. I fixed it using the evaluation context and removed all of the related code to changing the factory context. Originally I thought that the context passed in here would be the factory context. May be a good spot for a rename but I don't think I have enough context to do that

proteusvacuum: > enough context

😆

PR #14721, opened 2017-02-02
Externalize JavaScript: Exchange

orangejenny:
Tested on staging. Review commits individually.

@mkangia


snopoke (approved):
Slowly but surely!

PR #14722, opened 2017-02-02
Externalize js: import app

orangejenny:
@mkangia

### DIFF
--- a/corehq/apps/app_manager/templates/app_manager/v2/import_app.html
+++ b/corehq/apps/app_manager/templates/app_manager/v2/import_app.html
@@ -1,33 +1,15 @@
 {% extends "app_manager/v2/apps_base.html" %}
 {% load i18n %}
 {% load hq_shared_tags %}
+{% load compress %}
 
-{% block js-inline %}{{ block.super }}
-    <script src="{% static 'hqwebapp/js/lib/compression.js' %}"></script>
-    <script>
-        $(function () {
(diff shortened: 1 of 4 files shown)


corehq/apps/app_manager/templates/app_manager/v2/import_app.html
```diff
@@ -1,40 +1,24 @@
...
+    {% endcompress %}
 {% endblock %}
 
 {% block breadcrumbs %}
     <li>{% trans 'Import Application' %}</li>
 {% endblock %}
 
 {% block page_content %}
+    {% registerurl 'import_app' domain %}
+    {% initial_page_data 'export_json' app.export_json %}
```
snopoke: I know you didn't change this but just wondering if this is even necessary to have here. Can't we just do the copy all server side?

snopoke (approved):

corehq/apps/app_manager/templates/app_manager/v2/import_app.html
orangejenny: Possibly...I'm not really sure what the use case is where `app` has a value. It looks like there's sometimes an app or a different domain provided via GET (looking at https://github.com/dimagi/commcare-hq/blob/master/corehq/apps/app_manager/views/apps.py#L406-L407) but I haven't found code that triggers that workflow. Basically stopped investigating for fear that I was getting into a rabbit hole.

PR #14730, opened 2017-02-03
make corehq/apps/userreports compatible with python 3

nickpell:
@emord @millerdev

### DIFF
--- a/corehq/apps/userreports/models.py
+++ b/corehq/apps/userreports/models.py
@@ -1,3 +1,4 @@
+from __future__ import absolute_import
 from collections import namedtuple
 from copy import copy, deepcopy
 import json
--- a/corehq/apps/userreports/models.py
+++ b/corehq/apps/userreports/models.py
@@ -2,6 +2,7 @@ from collections import namedtuple
 from copy import copy, deepcopy
 import json
 from datetime import datetime
+from six.moves import filter as ifilter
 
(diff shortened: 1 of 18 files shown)


nickpell:
✅

corehq/apps/userreports/models.py
```diff
@@ -395,7 +397,7 @@ def map_col(column):
                     'XFormInstance': _('Forms'),
                     'CommCareCase': _('Cases')
                 }.get(self.config.referenced_doc_type, "Layer"),
-                'columns': filter(None, [map_col(col) for col in self.columns])
+                'columns': ifilter(None, [map_col(col) for col in self.columns])
```
emord: I think you want `list(filter(None, ...))` here. I'm not if we guarantee that this could handle an iterator

millerdev: `futurize` would convert this to a list comprehension. I think that should be our standard approach.

emord: May wan to keep the filter if `map_col` is expensive though

millerdev: or (preferably) do this
```py
[x for x in (map_col(col) for col in self.columns) if x]
```

nickpell: > ```[x for x in (map_col(col) for col in self.columns) if x]```

nice, I like this!

nickpell:
✅

PR #14737, opened 2017-02-03
fixing updates to latest saved version

calellowitz:
@gcapalbo @esoergel 
https://docs.google.com/document/d/1nHuMqwWagM9oph5mMqB8E6xbVVtZ_SwULDE4vkIHtwY/edit


calellowitz:
more context http://manage.dimagi.com/default.asp?237199

calellowitz:
cc: @amstone326

corehq/apps/app_manager/tasks.py
```diff
@@ -1,9 +1,23 @@
...
 
 
 @task(queue='background_queue', ignore_result=True)
 def create_user_cases(domain_name):
     from corehq.apps.callcenter.utils import sync_usercase
     for user in CommCareUser.by_domain(domain_name):
         sync_usercase(user)
+
+
+@serial_task('{app._id}', max_retries=0)
```
gcapalbo: looks like the default timeout on the lock is 5 minutes - do we anticipate making the build to ever take longer than 5 minutes?

gcapalbo (approved):
looks ok to me - is there anyone else with more context that you'd like to take a pass?

corehq/apps/app_manager/tasks.py
```diff
@@ -1,9 +1,23 @@
...
+def make_async_build(app):
+    latest_build = app.get_latest_app(released_only=False)
+    if latest_build.version == app.version:
+        return
+    errors = app.validate_app()
+    if not errors:
```
gcapalbo: will it be transparent to the user that there are errors with the build preventing them from updating to the latest version?

calellowitz: a failed build wont prevent an update, but will be invisible. not sure how to communicate that to the mobile, but the failed build will likely break the app anyway

corehq/apps/app_manager/tasks.py
calellowitz: that is definitely possible. do you think it makes more sense to use a normal task and then just wrap the logic in a critical section?

esoergel: That's true, 5 minutes may be a little low for some domains.  @proteusvacuum do you have any idea how long a reasonable build timeout would be for eNikshay?

Note that there's not a huge issue with an obviously long timeout, as that only triggers when the task hangs indefinitely.

corehq/apps/app_manager/views/download.py
```diff
@@ -73,7 +75,8 @@ def download_media_suite(request, domain, app_id):
 
     """
     if not request.app.copy_of:
-        request.app.set_media_versions(None)
+        previous_version = request.app.get_latest_app(released_only=False)
```
esoergel: It's not super obvious what `request.app.get_latest_app(released_only=False)` does (it wraps `get_app`, which is itself hella complex).  Is there a more direct getter you can use in these instances?

esoergel (approved):

corehq/apps/app_manager/tasks.py
gcapalbo: @calellowitz i think it's good to still use the serial_task, just need to pass a different value for `timeout` to the decorator

corehq/apps/app_manager/tasks.py
gcapalbo: yeah, i was just thinking more along the lines of someone not realizing that the update they just made on hq won't make it to mobile. if it's too big of a lift, probably no need to do it in this pr, but i think we would want to have some way to communicate to the user that the app won't build on the latest version when they're making updates in hq with errors

corehq/apps/app_manager/views/download.py
```diff
@@ -311,13 +314,17 @@ def download_profile(request, domain, app_id):
     See ApplicationBase.create_profile
 
     """
+    if not request.app.copy_of:
+        make_async_build.delay(request.app)
     return HttpResponse(
         request.app.create_profile()
     )
```
gcapalbo: i might be missing some context, but does this task need to finish before we return the download response, or do we expect the returned data to sometimes be a build behind?

corehq/apps/app_manager/tasks.py
```diff
@@ -1,9 +1,23 @@
...
 
 
 @task(queue='background_queue', ignore_result=True)
 def create_user_cases(domain_name):
     from corehq.apps.callcenter.utils import sync_usercase
     for user in CommCareUser.by_domain(domain_name):
         sync_usercase(user)
+
+
+@serial_task('{app._id}', max_retries=0, timeout=10*60)
```
dimagimon: E226 missing whitespace around arithmetic operator

corehq/apps/app_manager/views/download.py
calellowitz: thats the one used throughout `app_manager` i can look for another though

corehq/apps/app_manager/views/download.py
calellowitz: we expect the download to finish before the build. the build is so that the next time, the version comparison is against the last build

corehq/apps/app_manager/tasks.py
calellowitz: increased timeout to 10 minutes which should be long enough

gcapalbo:
:+1: looks good. if 10 minutes still feels tight, note that there's little downside to setting it to a longer timeout as Ethan said, like 1 hour or 10 hours

gcapalbo (approved):

corehq/apps/app_manager/tasks.py
```diff
@@ -1,9 +1,23 @@
...
+def make_async_build(app):
+    latest_build = app.get_latest_app(released_only=False)
+    if latest_build.version == app.version:
+        return
+    errors = app.validate_app()
+    if not errors:
+        copy = app.make_build(
+            previous_version=latest_build
+        )
+        copy.save(increment_version=False)
```
snopoke: I think these builds should have a comment associated with them to indicate that they were auto-created otherwise it could be confusing for users to suddenly see a whole bunch of builds that they didn't trigger.

corehq/apps/app_manager/views/download.py
esoergel: Hmm, you're probably right that it's safer to be consistent if there are other things already doing this.

corehq/apps/app_manager/tasks.py
```diff
@@ -1,9 +1,26 @@
...
 
 
 @task(queue='background_queue', ignore_result=True)
 def create_user_cases(domain_name):
     from corehq.apps.callcenter.utils import sync_usercase
     for user in CommCareUser.by_domain(domain_name):
         sync_usercase(user)
+
+
+@serial_task('{app._id}', max_retries=0, timeout=60*60)
```
dimagimon: E226 missing whitespace around arithmetic operator

corehq/apps/app_manager/tasks.py
calellowitz: what kind of ui are you imagining? since this is triggered by the phone, the logical place to alert the user would be on the phone but im not sure we have that capability. It is tough to alert on hq since they could be on any page (or not on hq at all) when this build is triggered, and will not even be logged in as the same user (mobile user on the phone, web user on hq). If the build fails it will likely cause problems on mobile anyway, that should become apparent.
In general, i like the idea of alerting but i'm not really sure how or what it would like in this case.

calellowitz:
Bumped timeout and added a build comment, let me know if there is anything else

corehq/apps/app_manager/tasks.py
gcapalbo: I was thinking something like a banner at the top of the page when viewing the app that says something like "There are errors preventing the latest version of this app from building. Phones will not receive the latest  version until these errors are corrected: <list of errors>"

I don't think we would want to notify the error on mobile since we should be alerting the person building the app that their change won't be deployed.

I think some build errors already display at the top of the page automatically whether you try to build or not, but am not sure if some errors will only be present during an actual build attempt.

gcapalbo:
:+1: i think that's everything

PR #14744, opened 2017-02-03
Update toggles.py

krsdimagi:
[ci skip]

@esoergel @robin824 adding help links to the feature flags page.

### DIFF
--- a/corehq/toggles.py
+++ b/corehq/toggles.py
@@ -522,7 +522,7 @@ VELLUM_PRINTING = StaticToggle(
     "Enables the Print Android App Callout",
     TAG_PRODUCT_PATH,
     help_link='https://confluence.dimagi.com/display/ccinternal/Printing+from+a+form+in+CommCare+Android',
-    namespace=[NAMESPACE_DOMAIN]
+    namespaces=[NAMESPACE_DOMAIN]
 )
 
 VELLUM_DATA_IN_SETVALUE = StaticToggle(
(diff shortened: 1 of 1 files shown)


corehq/toggles.py
```diff
@@ -298,15 +299,16 @@ def toggle_values_by_name(username=None, domain=None):
     'detail-list-tab-nodesets',
     'Associate a nodeset with a case detail tab',
     TAG_PRODUCT_PATH,
-    [NAMESPACE_DOMAIN],
     help_link='https://confluence.dimagi.com/display/internal/Case+Detail+Nodesets',
+    namespaces=[NAMESPACE_DOMAIN]
```
esoergel: @krsdimagi both versions of this are actually valid.  In python, you can use both "positional" arguments and "keyword" (aka "named") arguments.  The interplay between the two is a little strange, if you're interested in learning more, those are some terms to google.

corehq/toggles.py
```diff
@@ -516,7 +521,8 @@ def toggle_values_by_name(username=None, domain=None):
     'printing',
     "Enables the Print Android App Callout",
     TAG_PRODUCT_PATH,
-    [NAMESPACE_DOMAIN]
+    help_link='https://confluence.dimagi.com/display/ccinternal/Printing+from+a+form+in+CommCare+Android',
+    namespace=[NAMESPACE_DOMAIN]
```
esoergel: This should be `namespaces`

esoergel (changes requested):

corehq/toggles.py
krsdimagi: Thanks!  Will take a look.

esoergel (approved):

PR #14745, opened 2017-02-03
Add endpoint to retrieve a list of apps for installation on phone

proteusvacuum:
Send a list of apps accessible to a particular user to the phone for use in installation.

We'll be using this so that people can download any app that is accessible to them immediately after downloading CommCare from the play store.

@amstone326 
#hackathon

cc: @kaapstorm @czue

### DIFF
--- a/corehq/apps/app_manager/tests/data/list_apps/list_apps.xml
+++ b/corehq/apps/app_manager/tests/data/list_apps/list_apps.xml
@@ -1,0 +1,6 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<apps>
+  <app profile="http://onering.com/a/rohan/apps/download/None/profile.xml?latest=true" domain="rohan" name="Untitled Application" environment="production" media-profile="http://onering.com/a/rohan/apps/download/None/media_profile.xml?latest=true"/>
+  <app profile="http://onering.com/a/angmar/apps/download/None/profile.xml?latest=true" domain="angmar" name="Untitled Application" environment="production" media-profile="http://onering.com/a/angmar/apps/download/None/media_profile.xml?latest=true"/>
+  <app profile="http://onering.com/a/mordor/apps/download/None/profile.xml?latest=true" domain="mordor" name="Untitled Application" environment="production" media-profile="http://onering.com/a/mordor/apps/download/None/media_profile.xml?latest=true"/>
+</apps>
--- a/corehq/apps/app_manager/tests/data/list_apps/list_apps.xml
+++ b/corehq/apps/app_manager/tests/data/list_apps/list_apps.xml
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <apps>
-  <app profile="http://onering.com/a/rohan/apps/download/None/profile.xml?latest=true" domain="rohan" name="Untitled Application" environment="production" media-profile="http://onering.com/a/rohan/apps/download/None/media_profile.xml?latest=true"/>
-  <app profile="http://onering.com/a/angmar/apps/download/None/profile.xml?latest=true" domain="angmar" name="Untitled Application" environment="production" media-profile="http://onering.com/a/angmar/apps/download/None/media_profile.xml?latest=true"/>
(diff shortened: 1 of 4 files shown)


czue:
I like where this is going.

corehq/apps/app_manager/views/phone.py
```diff
@@ -0,0 +1,70 @@
...
+    profile = StringField('@profile')
+
+
+class AppList(XmlObject):
+    ROOT_NAME = 'apps'
+    apps = NodeListField('app', App)
+
+
+@check_lockout
+@basicauth()
```
czue: what's the benefit of restricting this to basic auth? as opposed to using `login_or_digest_or_basic_or_apikey` or similar

corehq/apps/app_manager/views/phone.py
```diff
@@ -0,0 +1,70 @@
...
+            get_all_latest_builds_for_user(couch_user)
+        ).serializeDocument(),
+        content_type='application/xml',
+    )
+
+
+def get_all_latest_builds_for_user(user):
+    app_ids = [
+        (domain, app_id)
+        for domain in user.domains
```
czue: i think this would fail hard if you ever hit the API as a mobile worker, yeah? was that intended? If so, might be better to make it a more explicit restriction, though i could also imagine a world where it's good to support it

corehq/apps/app_manager/views/phone.py
```diff
@@ -0,0 +1,70 @@
...
+    NodeListField,
+)
+
+
+class App(XmlObject):
+    ROOT_NAME = 'app'
+
+    domain = StringField('@domain')
+    name = StringField('@name')
+    environment = StringField('@environment')
```
czue: what's the intended usage of this field?

czue (commented):
excited to see the mobile UX of this. had one possible change that might be worth looking at

corehq/apps/app_manager/tests/data/list_apps/list_apps.xml
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<apps>
+  <app profile="http://onering.com/a/rohan/apps/download/None/profile.xml?latest=true" domain="rohan" name="Untitled Application" environment="production" media-profile="http://onering.com/a/rohan/apps/download/None/media_profile.xml?latest=true"/>
+  <app profile="http://onering.com/a/angmar/apps/download/None/profile.xml?latest=true" domain="angmar" name="Untitled Application" environment="production" media-profile="http://onering.com/a/angmar/apps/download/None/media_profile.xml?latest=true"/>
+  <app profile="http://onering.com/a/mordor/apps/download/None/profile.xml?latest=true" domain="mordor" name="Untitled Application" environment="production" media-profile="http://onering.com/a/mordor/apps/download/None/media_profile.xml?latest=true"/>
```
snopoke: are the `None`s in these URLs intentional?

corehq/apps/app_manager/views/phone.py
proteusvacuum: Mobile polls both prod and india for your available apps, so this was intended to show where the app was coming from.

corehq/apps/app_manager/views/phone.py
proteusvacuum: No benefit of restricting... mostly just laziness.
The `login_or_digest_or_basic_or_apikey` and others of the same type require a domain, and we don't have one. 
Also, mobile uses basic auth for this endpoint, so I figured this was enough.

corehq/apps/app_manager/views/phone.py
proteusvacuum: I don't think so. The `SingleMembershipMixin` which `CouchUser` inherits from has a `domains` property which just returns `[self.domain]`
The intention is to support both mobile and web workers.

corehq/apps/app_manager/tests/data/list_apps/list_apps.xml
proteusvacuum: https://github.com/dimagi/commcare-hq/pull/14745/commits/14877305eeac2f1c04a491f759580684c87a9a2a

proteusvacuum:
✅ new buddy: @emord

PR #14747, opened 2017-02-03
Update toggles.py

robin824:
Updating more Feature flag docs @esoergel @krsdimagi


corehq/toggles.py
```diff
@@ -320,6 +320,9 @@ def toggle_values_by_name(username=None, domain=None):
     'Multimedia Case Properties',
     TAG_PRODUCT_PATH,
     [NAMESPACE_DOMAIN, NAMESPACE_USER]
+    "Turning on this feature flag allows you to save multimedia/image capture photos as case properties in the App builder."
```
proteusvacuum: I think you need `description=(` here

czue: also a comma at the end of the previous line

PR #14748, opened 2017-02-03
App manager.translations

dannyroberts:
depends on https://github.com/dimagi/commcare-translations/pull/37

From the Feb 3 hackathon. Aliza pulled out past translation files from the git history and labeled them by commcare version (2.25, 2.26, etc.) in commcare-translations. This PR makes the app builder pull its default translation strings from the translation file corresponding to the commcare version selected for the app.

This fixes a long standing class of bugs that result from updating a single master translation file whenever we release a new commcare version, which would effect even apps that weren't on the latest version. Since translations get added and deleted, that meant that at the very least, the bulk ui translations excel file would contain irrelevant strings and be missing relevant ones, which causes confusion and support burden.

This PR finally brings ui translations into a state that not just mostly-works, but is actually cleanly versioned and logically consistent. As a nice touch, we also include the commcare version in the meta data of the excel file download, so that when you change your app version but then upload an old bulk translations file, it can tell the difference between lines you've edited and ones that are just there because they used to be in the old version, making for a clean and seamless upgrade experience.


amstone326:
@dannyroberts Can you ping me whenever this gets merged so that I can merge the corresponding PR in our deploy scripts (https://github.com/dimagi/mobile-deploy/pull/12)?

dannyroberts:
@amstone326 I didn't want to merge this while I was super offline, but just did another pass at reassuring myself with some basic tests on staging, and I'm happy to have it merged now.

@nickpell buddy, can you take a look at this PR and the related https://github.com/dimagi/commcare-translations/pull/37?

amstone326:
@dannyroberts Re-reading this PR description just filled me with joy 😃 

Going to merge https://github.com/dimagi/mobile-deploy/pull/12 right now, which will be in time for the 2.34 release! Woohoo

dannyroberts:
Wee! Thanks again @amstone326 for getting me to finally do this!

PR #14756, opened 2017-02-06
update vellum

czue:
vellum:e8a839484c09a0429c0a75a350464e4ad1038321

@millerdev @emord @orangejenny I'm not sure where the bower changes are coming from. is this right or did I mess something up?

ran `./scripts/vellum-to-hq master --vellum-dir=/home/czue/src/Vellum`

and it only thought i was changing these:

```
Vellum Changes Since Last HQ Deploy
e8a8394 Merge pull request #850 from dimagi/cz/save-to-case
7580536 Merge pull request #849 from dimagi/cz/update-readme
18a45a1 Merge pull request #848 from dimagi/cz/update-readme
```


emord:
I think you just have a really old vellum install. I think you can use `bower prune` to get rid of those. `jstree-actions` is something that's introduced in @orangejenny onboarding branch so we have it installed even though it's not included

PR #14440, opened 2017-01-05
Daily Saved Export / Dashboard Feed redesign

NoahCarnahan:
Apologies in advance for this monster... 😳

This PR reorganizes and adds functionality to the exports page. The changes are detailed in "phase 2" of [this document](https://docs.google.com/document/d/16svaBmyNiFGYEyaOMHKLUoJrLG9M7DjtwaNvrdusoH0/edit).

A summary of the major changes:
- move daily saved exports and excel dashboards to their own pages at the same level as "forms" and "cases"
- rename "excel dashboards" to "dashboard feeds"
- Allow users to save and edit filters for daily saved exports (previously impossible to do it for the *saved* export). This UI is in a modal on the export list page.

I tried to keep the commits of manageable size and a clean history, but there ends up being a bit of churn from e52b36b and on. I think all of the commits after that point are reconciling the changes from this ["locations export" PR](https://github.com/dimagi/commcare-hq/pull/13362) that happened simultaneously. Also given the shear number of commits it might be easier to review this all at once, not sure.

cc:
code buddy @biyeun 
initiative buddy @kaapstorm 
location exports expert @mkangia

PR #14757, opened 2017-02-06
Pass a string, not an Exception instance

kaapstorm:
Context: [FB 246864](http://manage.dimagi.com/default.asp?246864)

This change doesn't resolve the underlying exception, but does avoid HQ from trying to encode an Exception instance as JSON. 

I'm opening this PR to discuss why the KeyError is being raised in the first place. I can't see why ["emw" is hardcoded here](https://github.com/dimagi/commcare-hq/blob/d8a8c3101c74d63d61351ad6472a325b6bdf4ea3/corehq/apps/export/views.py#L2250-L2250) (where the KeyError is being raised). It seems to refer to [`ExpandedMobileWorkerFilter.slug`](https://github.com/dimagi/commcare-hq/blob/0f511188e207f92ffebb64cd6229791257c6ea06/corehq/apps/reports/filters/users.py#L195-L195), but it's not obvious to me why that particular slug is used here. How do we know it's not going to be the slug of a subclass?

@mkangia @NoahCarnahan @benrudolph

### DIFF
--- a/corehq/apps/export/views.py
+++ b/corehq/apps/export/views.py
@@ -2247,7 +2247,7 @@ class DownloadNewFormExportView(GenericDownloadNewExportMixin, DownloadFormExpor
         return form_filters
 
     def get_multimedia_task_kwargs(self, in_data, filter_form, export_object, download_id):
-        filter_slug = in_data['form_data']['emw']
+        filter_slug = in_data['form_data'][LocationRestrictedMobileWorkerFilter.slug]
         mobile_user_and_group_slugs = self._get_mobile_user_and_group_slugs(filter_slug)
         return filter_form.get_multimedia_task_kwargs(export_object, download_id, mobile_user_and_group_slugs)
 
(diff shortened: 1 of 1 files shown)


benrudolph:
hmm strange, it seems likely that `emw` was removed from the in_data and that change was not reflected on the server for multimedia uploads: https://github.com/dimagi/commcare-hq/commit/83ad01d8. @mkangia does that sound correct?

NoahCarnahan:
I agree it looks like it shouldn't be hard coded there, but don't know off the top of my head how the correct slug should be determined. I think Manish will have some insight, but I'm happy to look more closely at it if not.

mkangia:
@kaapstorm Did a similar fix in Noah's changes [here](https://github.com/dimagi/commcare-hq/pull/14440/commits/83ad01d8d2f7142b99751522718e9932939b66f2). "emw" was used earlier and was updated. I think can do the same change `LocationRestrictedMobileWorkerFilter.slug` which should be the new param used now. Thanks for addressing this.

kaapstorm:
Thanks for the help guys. Commit 9a5dbff should resolve the error.

@emord cc @czue @proteusvacuum

czue (approved):

PR #14726, opened 2017-02-03
Add select options to case summary view

proteusvacuum:
For multiple choice questions, you can now see the possible values that will save to the case. Hooray! 

screenshot from 2017-02-03 10-55-14

@czue @kaapstorm 
#INChackathon

PR #14760, opened 2017-02-06
Respect showLabels option in case summary options

proteusvacuum:
https://github.com/dimagi/commcare-hq/pull/14726#pullrequestreview-20142767

@czue @kaapstorm


czue (approved):
nice!

PR #14761, opened 2017-02-06
allow superusers to use restore as

benrudolph:
@snopoke @emord

fyi: @robin824


emord (approved):

czue (approved):
out of scope for this PR, but I was surprised to see that superuser status isn't baked into our permissions checks at a lower level. is there ever a scenario where `user.can_do_x()`, `user.has_permission(x)`, or `user.is_admin(domain)` should ever be `False` for superusers?

benrudolph:
i was surprised too until i actually tried it out. i can't imagine there ever being a situation where we'd want it to be false

PR #14762, opened 2017-02-06
Get app_manager ready for Python 3

millerdev:
Used `python-future`'s `futurize` script to do most of this, and then hand-edited (in separate commits) changes that were unnecessary or wrong. Probably best to review commits individually.

Things to discuss:
- Do we want to standardize on use of `future` vs `six`? I had previously thought `future` depended on `six`, but on a subsequent reading of the docs I see it doesn't. See the [`future` FAQ on this topic](http://python-future.org/faq.html#what-is-the-relationship-between-future-and-six). I personally think `future` is better because it "[allows] standard Py3 code to run with almost no modification on both Py3 and Py2."
- Are we going to try to make all code written from now on Python 3 compatible? I think it's a good idea, and if others agree we should make @dimagimon enforce it.
- String types are going to require special care. `future` uses `from builtins import str` to make `str` act like `unicode` in Python 2. This can cause confusion if you're not expecting it. Also, I think there may be a bug where `futurize` forgets to add that import in some cases, but that's a separate issue.

@emord @nickpell

### DIFF
--- a/corehq/apps/app_manager/fields.py
+++ b/corehq/apps/app_manager/fields.py
@@ -1,3 +1,5 @@
+from builtins import map
+from builtins import object
 import logging
 from django.core.urlresolvers import reverse
 from django.http import Http404
@@ -83,7 +85,7 @@ class ApplicationDataSourceUIHelper(object):
     def bootstrap(self, domain):
         self.all_sources = get_app_sources(domain)
         self.application_field.choices = sorted(
-            [(app_id, source['name']) for app_id, source in self.all_sources.items()],
+            [(app_id, source['name']) for app_id, source in list(self.all_sources.items())],
             key=lambda id_name_tuple: (id_name_tuple[1] or '').lower()
         )
         self.source_field.choices = []
@@ -96,12 +98,12 @@ class ApplicationDataSourceUIHelper(object):
         if self.enable_cases:
             _add_choices(
                 self.source_field,
-                [(ct['value'], ct['text']) for app in self.all_sources.values() for ct in app['case']]
+                [(ct['value'], ct['text']) for app in list(self.all_sources.values()) for ct in app['case']]
             )
         if self.enable_forms:
             _add_choices(
                 self.source_field,
-                [(ct['value'], ct['text']) for app in self.all_sources.values() for ct in app['form']]
+                [(ct['value'], ct['text']) for app in list(self.all_sources.values()) for ct in app['form']]
             )
 
         # NOTE: This corresponds to a view-model that must be initialized in your template.
@@ -313,28 +315,22 @@ class ApplicationDataRMIHelper(object):
     @property
     @memoized
     def _no_app_forms(self):
-        return filter(lambda f: not f.get('has_app', False), self._all_forms)
+        return [f for f in self._all_forms if not f.get('has_app', False)]
 
     @property
     @memoized
     def _remote_app_forms(self):
-        return filter(lambda f: f.get('has_app', False) and f.get('show_xmlns', False), self._all_forms)
+        return [f for f in self._all_forms if f.get('has_app', False) and f.get('show_xmlns', False)]
 
     @property
     @memoized
     def _deleted_app_forms(self):
-        return filter(
-            lambda f: f.get('has_app', False) and f.get('app_deleted') and not f.get('show_xmlns', False),
-            self._all_forms
-        )
+        return [f for f in self._all_forms if f.get('has_app', False) and f.get('app_deleted') and not f.get('show_xmlns', False)]
 
     @property
     @memoized
     def _available_app_forms(self):
-        return filter(
-            lambda f: f.get('has_app', False) and not f.get('app_deleted') and not f.get('show_xmlns', False),
-            self._all_forms
-        )
+        return [f for f in self._all_forms if f.get('has_app', False) and not f.get('app_deleted') and not f.get('show_xmlns', False)]
 
     @property
     @memoized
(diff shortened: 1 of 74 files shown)


dimagimon:
There are 14 errors:

* corehq/apps/app_manager/fields.py, line 328 - E501 line too long (130 > 115 characters)
* corehq/apps/app_manager/fields.py, line 333 - E501 line too long (134 > 115 characters)
* corehq/apps/app_manager/remote_app.py, line 4 - E401 multiple imports on one line
* corehq/apps/app_manager/remote_app.py, line 5 - F811 redefinition of unused 'urllib' from line 4
* corehq/apps/app_manager/suite_xml/features/scheduler.py, line 32 - E127 continuation line over-indented for visual indent
* corehq/apps/app_manager/suite_xml/generator.py, line 4 - E401 multiple imports on one line
* corehq/apps/app_manager/templatetags/url_extras.py, line 5 - E401 multiple imports on one line
* corehq/apps/app_manager/templatetags/url_extras.py, line 37 - E501 line too long (120 > 115 characters)
* corehq/apps/app_manager/tests/test_dbaccessors.py, line 163 - F821 undefined name 'app_build_verions'
* corehq/apps/app_manager/tests/test_dbaccessors.py, line 164 - F821 undefined name 'app_build_verions'
* corehq/apps/app_manager/tests/test_dbaccessors.py, line 170 - F821 undefined name 'app_build_verions'
* corehq/apps/app_manager/tests/test_dbaccessors.py, line 171 - F821 undefined name 'app_build_verions'
* corehq/apps/app_manager/views/apps.py, line 199 - E127 continuation line over-indented for visual indent
* corehq/apps/app_manager/views/apps.py, line 201 - E127 continuation line over-indented for visual indent

corehq/apps/app_manager/models.py
```diff
@@ -307,13 +317,13 @@ def get_action_properties(self, action):
         if 'external_id' in action_properties and action.external_id:
             yield 'external_id', action.external_id
         if 'update' in action_properties:
-            for name, path in action.update.items():
+            for name, path in list(action.update.items()):
```
nickpell: ```list``` not needed

corehq/apps/app_manager/models.py
```diff
@@ -307,13 +317,13 @@ def get_action_properties(self, action):
...
                 yield name, path
         if 'case_properties' in action_properties:
-            for name, path in action.case_properties.items():
+            for name, path in list(action.case_properties.items()):
```
nickpell: ```list``` not needed

corehq/apps/app_manager/models.py
```diff
@@ -307,13 +317,13 @@ def get_action_properties(self, action):
...
                 yield name, path
         if 'preload' in action_properties:
-            for path, name in action.preload.items():
+            for path, name in list(action.preload.items()):
```
nickpell: ```list``` not needed

corehq/apps/app_manager/models.py
```diff
@@ -411,7 +421,7 @@ class AdvancedAction(IndexedSchema):
     __eq__ = DocumentSchema.__eq__
 
     def get_paths(self):
-        for path in self.case_properties.values():
+        for path in list(self.case_properties.values()):
```
nickpell: ```list``` not needed

corehq/apps/app_manager/models.py
```diff
@@ -511,7 +521,7 @@ def get_paths(self):
         for path in super(LoadUpdateAction, self).get_paths():
             yield path
 
-        for path in self.preload.keys():
+        for path in list(self.preload.keys()):
```
nickpell: ```list``` not needed

corehq/apps/app_manager/models.py
millerdev: It's possible that it's not needed, but it's also possible that it is. The generator would fail if the thing that's iterating over this generator function mutates `action.update` during iteration.  Same below.

nickpell (approved):

corehq/apps/app_manager/models.py
nickpell: interesting, hadn't thought about that

emord (commented):
> Do we want to standardize on use of future vs six? I had previously thought future depended on six, but on a subsequent reading of the docs I see it doesn't. 

I'm not excited that this overrides standard python types. I like the explicitness of `six.text_type`

> Are we going to try to make all code written from now on Python 3 compatible? I think it's a good idea, and if others agree we should make @dimagimon enforce it.

I'm not sure we're at that point yet, but I'm pro this idea once we are able to verify it.

> String types are going to require special care. future uses from builtins import str to make str act like unicode in Python 2. This can cause confusion if you're not expecting it.

Agreed with it causing  confusion. Do you see an issue with using `six.text_type?

millerdev:
> I'm not excited that this overrides standard python types.

I'm curious to explore this further. Other than needing to remember `from builtins import ...` I don't see a big issue with the overridden types. (And I think remembering to import from `builtins` could be mitigated by @dimagimon automation.) The advantage I see with `future` over `six` is that it means less work long term because it allows writing code the way it should be for Python 3.

> Agreed with it causing confusion. Do you see an issue with using `six.text_type`?

I don't like the aesthetics of it, but other than that I have no real objections. I suppose it will be easy enough to replace `six.text_type` with `str` when we're finally switched over to Python 3.

corehq/apps/app_manager/fields.py
```diff
@@ -377,12 +372,12 @@ def _get_applications_by_type(self, as_dict=True):
...
         )
-        _app_fmt = lambda c: (c[0], map(lambda f: RMIDataChoice(
+        _app_fmt = lambda c: (c[0], [RMIDataChoice(
             f['app']['id'] if f.get('has_app', False) else self.UNKNOWN_SOURCE,
             f['app']['name'] if f.get('has_app', False) else _("Unknown Application"),
             f
-        ), c[1]))
-        apps_by_type = map(_app_fmt, apps_by_type)
+        ) for f in c[1]])
+        apps_by_type = list(map(_app_fmt, apps_by_type))
```
millerdev: ha, I didn't look at the context carefully enough. I thought `apps_by_type` was being referenced in multiple places (it is), but didn't realize it was being reassigned. `dict(map(...))` should be fine.

corehq/apps/app_manager/fields.py
```diff
@@ -396,8 +391,8 @@ def _get_applications_by_type(self, as_dict=True):
 
     @staticmethod
     def _map_chosen_by_choice_as_dict(chosen_by_choice):
-        for k, v in chosen_by_choice.items():
-            chosen_by_choice[k] = map(lambda f: f._asdict(), v)
+        for k, v in list(chosen_by_choice.items()):
```
millerdev: `chosen_by_choice` is being mutated inside the loop, so I thought it would break in Python 3 without `list(...)`. However, it does appear to work, probably because it's assigning to the same key:
```
Python 3.5.1 (default, Mar  2 2016, 03:38:02)
[GCC 4.2.1 Compatible Apple LLVM 7.0.2 (clang-700.1.81)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> x = {1: 2, 3: 4, 5: 6}
>>> for k, v in x.items():
...     x[k] = v ** 2
...
>>> x
{1: 4, 3: 16, 5: 36}
```
Indeed [the documentation](https://docs.python.org/3/library/stdtypes.html#dictionary-view-objects) implies that it's safe since we are not adding or deleting entries: "Iterating views while adding or deleting entries in the dictionary may raise a `RuntimeError` or fail to iterate over all entries."

millerdev:
Closing for now. I'm leaning toward six/modernize rather than futurize, so I may redo this at some point, or if we come up with new reasons to use futurize then maybe this will be resurrected.

PR #14766, opened 2017-02-06
allow specifying a timing threshold

snopoke:


emord (approved):
The output of this would be one line each for the setup, test and tear down, correct? So it would only output something if one of those steps was > threshold? (not actionable, just want to make sure I understand it)

snopoke:
@emord that's correct

corehq/tests/noseplugins/timing.py
```diff
@@ -32,6 +32,9 @@ def options(self, parser, env):
                           default=env.get('NOSE_PRETTY_TIMING'),
                           help='Print timing info in a format that is better '
                                'for reviewing in text mode (not CSV).')
+        parser.add_option('--threshold', type=int,
+                          default=env.get('NOSE_TIMING_THRESHOLD'),
+                          help='Only print timing info for events above this threshold.')
```
millerdev: Assuming the threshold units are seconds. Might be nice to document that here.

millerdev (approved):

czue:
does travis print these out for us?

snopoke:
Not at present but I do plan on adding that in.

benrudolph (approved):

PR #14768, opened 2017-02-06
Add a Coalesce UCR Expression

sheelio:
@czue 

This is super useful for a bunch of UCRs I've written (but can be replicated using other expressions so ignore these changes if you don't think useful).

### DIFF
--- a/corehq/apps/userreports/expressions/specs.py
+++ b/corehq/apps/userreports/expressions/specs.py
@@ -414,4 +414,7 @@ class CoalesceExpressionSpec(JsonObject):
     def __call__(self, item, context=None):
         expression_value = self._expression(item, context)
         default_value = self._default_expression(item, context)
-        return expression_value or default_value
+        if expression_value is None or expression_value == '':
+            return default_value
+        else:
+            return expression_value
--- a/corehq/apps/userreports/expressions/specs.py
+++ b/corehq/apps/userreports/expressions/specs.py
@@ -400,3 +400,18 @@ class SplitStringExpressionSpec(JsonObject):
             return string_value.split(self.delimiter)[index_value]
         except IndexError:
             return None
+
+
+class CoalesceExpressionSpec(JsonObject):
+    type = TypeProperty('coalesce')
+    expression = DictProperty(required=True)
+    default_expression = DictProperty(required=True)
+
+    def configure(self, expression, default_expression):
+        self._expression = expression
+        self._default_expression = default_expression
+
+    def __call__(self, item, context=None):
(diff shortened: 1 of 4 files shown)


corehq/apps/userreports/expressions/specs.py
```diff
@@ -400,3 +400,18 @@ def __call__(self, item, context=None):
...
+    default_expression = DictProperty(required=True)
+
+    def configure(self, expression, default_expression):
+        self._expression = expression
+        self._default_expression = default_expression
+
+    def __call__(self, item, context=None):
+        expression_value = self._expression(item, context)
+        default_value = self._default_expression(item, context)
+        return expression_value or default_value
```
czue: is the intention that you want this to use `default_value` on falsey things like `0`? i could see that behavior being a bit surprising.

czue (commented):
code looks good.

this feels a bit like a more limited version of [`switch`](https://github.com/dimagi/commcare-hq/blob/master/corehq/apps/userreports/README.md#switch-expression). do you use switch or what is the main benefit of having this instead of that? it doesn't even feel like much more code, with the possible exception of handling `null` and `""` in the same case.

corehq/apps/userreports/expressions/specs.py
sheelio: That's actually not intentional - I can update the code.

sheelio:
Have used Switch - didn't realize I could use it to get a coalesce-esque behaviour.  Can you specific "" and null as cases in the switch?

czue:
> Can you specify "" and null as cases in the switch?

not sure if I've done it specifically but I would expect this to work

sheelio:
Tested the switch option and JSON doesn't let you specify "null" or "None" as a key (in the "cases" dictionary of the switch expression).

Think it might makes sense to keep this expression instead (and I'll update the code so it doesn't coalesce out falsey things like 0)

czue (approved):
sounds good. feel free to merge after you've done that.

corehq/apps/userreports/tests/test_expressions.py
```diff
@@ -1267,3 +1267,57 @@ def test_split_string_delimiter(self):
...
+            'expression': 0,
+            'default_expression': 'default',
+        }))
+
+    def testBlankDefaultValue2(self):
+        self.assertEqual('foo', self.expression({
+            'expression': 'foo',
+        }))
+
+    def testBlankDefaultValue2(self):
```
dimagimon: F811 redefinition of unused 'testBlankDefaultValue2' from line 1317

PR #14770, opened 2017-02-06
create drtb-hiv-referral case

nickpell:
@proteusvacuum @czue

### DIFF
--- a/custom/enikshay/nikshay_datamigration/factory.py
+++ b/custom/enikshay/nikshay_datamigration/factory.py
@@ -15,6 +15,7 @@ from custom.enikshay.nikshay_datamigration.models import Outcome
 PERSON_CASE_TYPE = 'person'
 OCCURRENCE_CASE_TYPE = 'occurrence'
 EPISODE_CASE_TYPE = 'episode'
+DRTB_HIV_REFERRAL_CASE_TYPE = 'drtb-hiv-referral'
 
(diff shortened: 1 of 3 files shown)


czue (approved):
lgtm. will also let @proteusvacuum take a quick look since he knows the data models better

custom/enikshay/nikshay_datamigration/factory.py
```diff
@@ -97,7 +98,11 @@ def get_case_structures_to_create(self):
         person_structure = self.get_person_case_structure()
         ocurrence_structure = self.get_occurrence_case_structure(person_structure)
         episode_structure = self.get_episode_case_structure(ocurrence_structure)
-        return [episode_structure]
+        if not self._outcome or self._outcome.hiv_status == 'unknown':
```
proteusvacuum: this should also trigger when `hiv_status = reactive`

proteusvacuum: if there is no outcome, do you set the hiv_status of the episode to `unknown`?

nickpell: >this should also trigger when hiv_status = reactive

updated

nickpell: > if there is no outcome, do you set the hiv_status of the episode to unknown?

yes

proteusvacuum (approved):

PR #14771, opened 2017-02-06
Add hand-off info to support email

esoergel:
https://docs.google.com/document/d/1uXOUxnTmi70-0Ei9V0yrrVQL3Suxgh-Nr3TGPDRIg2s/edit
I also did a little refactoring.  This can also be used from a non-domain page, so I pulled the domain-specific stuff out.  Commit-by-commit should be straightforward.
@millerdev

### DIFF
--- a/corehq/apps/hqwebapp/views.py
+++ b/corehq/apps/hqwebapp/views.py
@@ -556,13 +556,16 @@ def bug_report(req):
             u"Feature Flags: {feature_flags}\n"
             u"Feature Previews: {feature_previews}\n"
             u"Is scale backend: {scale_backend}\n"
+            u"Has Support Hand-off Info: {has_handoff_info}\n"
+            u"Internal Project Information: {internal_info_link}\n"
             u"Project description: {project_description}\n"
         ).format(
             software_plan=software_plan,
-            feature_flags=toggles.toggles_dict(username=report['username'],
-                                               domain=domain).keys(),
+            feature_flags=toggles.toggles_dict(username=report['username'], domain=domain).keys(),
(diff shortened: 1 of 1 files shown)


corehq/apps/hqwebapp/views.py
```diff
@@ -529,32 +513,63 @@ def bug_report(req):
...
-        subscriber__domain=report['domain'],
-    )
-
-    if len(matching_subscriptions) >= 1:
-        report['software_plan'] = matching_subscriptions[0].plan_version
+    if report['domain']:
+        domain = report['domain']
+    elif len(couch_user.domains) == 1:
+        # This isn't a domain page, but the user has only one domain, so let's use that
+        domain = couch_user.domains[0]
```
esoergel: This wasn't a requested change, just seemed reasonable.

benrudolph: there's also the `current_domain` property on the couch_user which may be useful

benrudolph: https://github.com/dimagi/commcare-hq/blob/master/corehq/apps/users/middleware.py#L36

benrudolph (approved):

benrudolph:
cool!

millerdev (approved):

corehq/apps/hqwebapp/views.py
esoergel: Cool, I hadn't seen that one.  I don't think it'll be available on non-domain pages, though, since it's underneath a `if 'domain' in view_kwargs:` check.

benrudolph: 👍 right on

PR #14772, opened 2017-02-06
update models

nickpell:
1.  makes two fields nullable that didn't have nulls in the first dump
2.  deletes two fields that are not formatted properly in the last dump and aren't used anyway

@proteusvacuum @czue

### DIFF
--- a/custom/enikshay/nikshay_datamigration/models.py
+++ b/custom/enikshay/nikshay_datamigration/models.py
@@ -27,7 +27,7 @@ class PatientDetail(models.Model):
     page = models.IntegerField()
     poccupation = models.CharField(max_length=255)
     paadharno = models.BigIntegerField(null=True)
-    paddress = models.CharField(max_length=255)
+    paddress = models.CharField(max_length=255, null=True)
     pmob = models.CharField(max_length=255, null=True)  # validate numerical in factory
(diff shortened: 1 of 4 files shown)


proteusvacuum (approved):

custom/enikshay/nikshay_datamigration/models.py
```diff
@@ -285,8 +285,8 @@ class Outcome(models.Model):
     )
     InitiatedDate = models.CharField(max_length=255, null=True)  # dates, None, and NULL
     userName = models.CharField(max_length=255)
-    loginDate = models.DateTimeField()
-    OutcomeDate1 = models.CharField(max_length=255)  # datetimes and NULL
+    # loginDate = models.DateTimeField()
+    # OutcomeDate1 = models.CharField(max_length=255)  # datetimes and NULL
```
czue: why did you do this this way as opposed to adding a migration to remove them? just feels like manually changing migration files might get us into trouble.

nickpell: I don't follow.  https://github.com/dimagi/commcare-hq/pull/14772/files#diff-9b2457f77a1639e5734852448332815b should remove them, right?

czue: sorry ignore i was confused.

PR #14774, opened 2017-02-07
Tweaks to export migration

benrudolph:
A couple of tweaks I noticed while migrating staging

@NoahCarnahan as I was doing this I was wondering if there was any extra work we had to do to migrate old exports to new exports with your new changes to daily saved exports? My guess is no, but wanted to confirm

### DIFF
--- a/corehq/apps/export/static/export/js/models.js
+++ b/corehq/apps/export/static/export/js/models.js
@@ -486,6 +486,7 @@ hqDefine('export/js/models.js', function () {
         });
 
         self.showAdvanced = ko.observable(false);
+        self.showDeleted = ko.observable(false);
         self.customPathString.subscribe(self.onCustomPathChange.bind(self));
     };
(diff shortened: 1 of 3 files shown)


snopoke (approved):

corehq/apps/export/static/export/js/models.js
```diff
@@ -486,6 +486,7 @@ hqDefine('export/js/models.js', function () {
         });
 
         self.showAdvanced = ko.observable(false);
+        self.showDeleted = ko.observable(false);
```
snopoke: was this just missing before?

benrudolph: yeah, this model is used very infrequently. it's used when we enabled the feature flag to allow users to define their own columns/tables. When you create your own table, this is the model that gets created

emord (approved):

NoahCarnahan:
Thanks for checking. There shouldn't be any migrations needed for my changes (used default values on the objects). But let me know if anything comes up that looks suspicious.

PR #14776, opened 2017-02-07
50% Do not process old builds for export generation

benrudolph:
@snopoke @emord


czue (approved):
:chart_with_upwards_trend:

snopoke (approved):
Need to wait for build?

benrudolph:
actually not anymore, there used to be some chance that this would break tests, but we've since made the randomness always return false

PR #14777, opened 2017-02-07
Ensure that atwho works in iframe

benrudolph:
Ensures that the correct offsets get set when atwho gets loaded in an iframe


snopoke (approved):

emord (approved):

benrudolph:
JS tests passed, merging

PR #14780, opened 2017-02-07
cache report_class_name

benrudolph:
@snopoke during my dive in the report dispatcher, i noticed it took 4 seconds to call `get_report`. that function gets called at least 2, sometimes 3 times per report load. i switched around the function return value so that we can cache it. here's the prof data:

```
loading profile stats for get_report_class_name-20170207T120127706584
         4248460 function calls (3478542 primitive calls) in 4.230 seconds

   Ordered by: cumulative time, call count
   List reduced from 531 to 200 due to restriction <200>

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
        1    0.000    0.000    4.230    4.230 /Users/benrudolph/Dimagi/commcare-hq/corehq/apps/reports/dispatcher.py:125(get_report_class_name)
        1    0.000    0.000    4.230    4.230 /Users/benrudolph/Dimagi/commcare-hq/corehq/apps/reports/dispatcher.py:117(get_report)
        1    0.000    0.000    4.230    4.230 /Users/benrudolph/Dimagi/commcare-hq/corehq/apps/reports/dispatcher.py:112(get_reports_dict)
        1    0.000    0.000    4.230    4.230 /Users/benrudolph/Dimagi/commcare-hq/corehq/apps/reports/dispatcher.py:79(get_reports)
        1    0.000    0.000    4.227    4.227 /Users/benrudolph/Dimagi/commcare-hq/corehq/apps/reports/dispatcher.py:87(process)
        1    0.000    0.000    4.227    4.227 /Users/benrudolph/Dimagi/commcare-hq/corehq/reports.py:63(REPORTS)
        2    0.000    0.000    4.218    2.109 /Users/benrudolph/Dimagi/commcare-hq/corehq/reports.py:182(_safely_get_report_configs)
        2    0.034    0.017    4.209    2.105 /Users/benrudolph/Dimagi/commcare-hq/corehq/apps/userreports/models.py:605(by_domain)
      840    0.007    0.000    4.170    0.005 /Users/benrudolph/Dimagi/commcare-hq/corehq/apps/userreports/models.py:599(all)
      838    0.008    0.000    3.691    0.004 /Users/benrudolph/Dimagi/commcare-hq/corehq/apps/userreports/models.py:671(_get_report_config)
```
cc: @emord

### DIFF
--- a/corehq/apps/reports/dispatcher.py
+++ b/corehq/apps/reports/dispatcher.py
@@ -5,6 +5,7 @@ from django.utils.translation import ugettext
 from django.views.generic.base import View
 
 from dimagi.utils.decorators.datespan import datespan_in_request
+from dimagi.utils.modules import to_function
 
 from django_prbac.exceptions import PermissionDenied
 from django_prbac.utils import has_privilege
@@ -17,6 +18,7 @@ from corehq.apps.domain.models import Domain
 from corehq.apps.domain.utils import get_domain_module_map
 from corehq.apps.hqwebapp.templatetags.hq_shared_tags import toggle_enabled
 from corehq.apps.reports.exceptions import BadRequestError
+from corehq.util.quickcache import quickcache
 
 
 datespan_default = datespan_in_request(
@@ -119,6 +121,11 @@ class ReportDispatcher(View):
         """
         return self.get_reports_dict(domain).get(report_slug, None)
 
+    @quickcache(['domain', 'report_slug'], timeout=60)
+    def get_report_class_name(self, domain, report_slug):
+        report_cls = self.get_report(domain, report_slug)
+        return report_cls.__module__ + '.' + report_cls.__name__ if report_cls else ''
+
     def _redirect_slug(self, slug):
         return self.slug_aliases.get(slug) is not None
 
@@ -145,8 +152,8 @@ class ReportDispatcher(View):
 
         report_kwargs = kwargs.copy()
 
-        cls = self.get_report(domain, report_slug)
-        class_name = cls.__module__ + '.' + cls.__name__ if cls else ''
+        class_name = self.get_report_class_name(domain, report_slug)
+        cls = to_function(class_name)
 
         permissions_check = permissions_check or self.permissions_check
         if (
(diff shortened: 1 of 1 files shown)


benrudolph:
Unless I'm confused, this is why none of our reports _ever_ load sub 4 seconds: https://app.datadoghq.com/screen/47018/report-metrics

corehq/apps/reports/dispatcher.py
```diff
@@ -145,8 +152,8 @@ def dispatch(self, request, domain=None, report_slug=None, render_as=None,
 
         report_kwargs = kwargs.copy()
 
-        cls = self.get_report(domain, report_slug)
-        class_name = cls.__module__ + '.' + cls.__name__ if cls else ''
+        class_name = self.get_report_class_name(domain, report_slug)
+        cls = to_function(class_name) if class_name else None
```
snopoke: if it's already a class why do we need to use `to_function`?

snopoke (commented):
Wow, nice find.

snopoke:
Could we also cache [get_reports](https://github.com/dimagi/commcare-hq/blob/4f6972bda6be7f63bbe1636d6be4e468f0931390/corehq/apps/reports/dispatcher.py#L76-L76)?

benrudolph:
it's trickier because we can't cache the report class, but i'll look into  a work around

corehq/apps/reports/dispatcher.py
benrudolph: `class_name` is actually string. i did that so it was easier to cache

benrudolph:
(offlined, just going to leave as is)

snopoke (approved):

emord (approved):

corehq/apps/reports/dispatcher.py
```diff
@@ -119,6 +121,11 @@ def get_report(self, domain, report_slug, *args):
         """
         return self.get_reports_dict(domain).get(report_slug, None)
 
+    @quickcache(['domain', 'report_slug'], timeout=60)
```
czue: any reason not to make this timeout even longer? if i'm reading the code right the only way this can change is during a deploy, no?

czue (commented):
nice find!

corehq/apps/reports/dispatcher.py
benrudolph: that's correct, the only reason is to not lose my position in the travis queue :trollface: (i can just use ci skip)

PR #14781, opened 2017-02-07
Use Elasticsearch.index() instead of .update()

kaapstorm:
Context: [FB 246601](http://manage.dimagi.com/default.asp?246601)

Elasticsearch.update() will merge documents, but Elasticsearch.index() will replace them. Replacing allows us to remove case properties.

@snopoke cc @czue @proteusvacuum


benrudolph:
lol...

does this mean we have to do an ES reindex? probably not worth the trouble unless there are widespread complaints

snopoke:
Yea, don't think a reindex is necessary.

snopoke (approved):

czue (approved):
nice find! also thanks for including a test

PR #14783, opened 2017-02-07
share hierarchical fixtures for domains that were on flat fixture and…

mkangia:
… have NOT chosen to remove hierarchical fixture from restores

This should fix issues with domains on flat fixture needing hierarchical location fixture to be synced as well, mirroring the behaviour that was there earlier.


mkangia:
Closing this since Ethan & Sheel figured a better way out.

PR #14787, opened 2017-02-07
Kill live preview A/B test

orangejenny:
@biyeun / @benrudolph

### DIFF
--- a/corehq/apps/cloudcare/utils.py
+++ b/corehq/apps/cloudcare/utils.py
@@ -3,8 +3,7 @@ from corehq.apps.analytics import ab_tests
 
 
 def should_show_preview_app(request, domain_obj, username):
-    live_preview_ab = ab_tests.ABTest(ab_tests.LIVE_PREVIEW, request)
-    if domain_obj.is_onboarding_domain and live_preview_ab.version == ab_tests.LIVE_PREVIEW_ENABLED:
+    if domain_obj.is_onboarding_domain:
         return True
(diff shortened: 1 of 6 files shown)


biyeun (approved):

corehq/apps/cloudcare/utils.py
```diff
@@ -3,8 +3,7 @@
 
 
 def should_show_preview_app(request, domain_obj, username):
-    live_preview_ab = ab_tests.ABTest(ab_tests.LIVE_PREVIEW, request)
-    if domain_obj.is_onboarding_domain and live_preview_ab.version == ab_tests.LIVE_PREVIEW_ENABLED:
+    if domain_obj.is_onboarding_domain:
         return True
     elif domain_obj.is_onboarding_domain and not request.couch_user.is_dimagi:
```
benrudolph: i think we can remove this whole if block and just defer to the toggle. right now this is saying all onboarding domains that aren't dimagi should not show live preview

benrudolph (changes requested):

corehq/apps/cloudcare/utils.py
benrudolph: acutally i misread this. it wouldn't be as bad as i suggested, but still can just drop the whole thing

orangejenny: fixed

benrudolph (approved):

PR #14789, opened 2017-02-07
add option to purge existing custom data fields from users when saving

czue:
was going to implement this as a management command but then seemed easy enough to do in the UI.

image

@dimagi/product / @sheelio do you see any issues with adding this? i can obviously flag it if needed but seems helpful to me.

I think @esoergel has the most context on this feature? cc @proteusvacuum @kaapstorm buddies. review how you like. :fish: / :classical_building:

### DIFF
--- a/corehq/apps/custom_data_fields/edit_model.py
+++ b/corehq/apps/custom_data_fields/edit_model.py
@@ -165,6 +165,8 @@ class CustomDataModelMixin(object):
     def post(self, request, *args, **kwargs):
         if self.form.is_valid():
             self.save_custom_fields()
+            if self.show_purge_existing and self.form.cleaned_data['purge_existing']:
+                self.update_existing_models()
             msg = _(u"{} fields saved successfully").format(
                 unicode(self.entity_string)
             )
@@ -172,3 +174,9 @@ class CustomDataModelMixin(object):
             return self.get(request, success=True, *args, **kwargs)
         else:
             return self.get(request, *args, **kwargs)
+
+    def update_existing_models(self):
+        """
+        Subclasses with show_purge_exiting set to True should override this to update existing models
+        """
+        raise NotImplementedError()
--- a/corehq/apps/custom_data_fields/edit_model.py
+++ b/corehq/apps/custom_data_fields/edit_model.py
@@ -97,6 +97,7 @@ class CustomDataModelMixin(object):
     urlname = None
     template_name = "custom_data_fields/custom_data_fields.html"
     field_type = None
+    show_purge_existing = False
     entity_string = None  # User, Group, Location, Product...
 
     @use_jquery_ui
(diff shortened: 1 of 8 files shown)


corehq/apps/custom_data_fields/templates/custom_data_fields/custom_data_fields.html
```diff
@@ -183,10 +183,27 @@
...
+        <fieldset>
+            <div class="col-sm-10 col-sm-offset-2">
+                <div class="checkbox">
+                    <label>
+                        <input name="remove-existing" type="checkbox" data-bind="checked: purge_existing">{% trans "Remove unused fields from existing data."%}
+                        <span class="hq-help-template"
+                              data-title="{% trans "Remove Unused Fields" %}"
+                              data-content="{% blocktrans with entity_string=view.entity_string|lower %}
+                                  If checked, this will remove any extra information from any existing {{ entity_string }}
+                                  that has data fields that are no longer used.
```
czue: i struggled a bit with this text and being too lazy to add a pluralized version of `{{ entity_string }}` to the template. open to any improvements.

sheelio:
@czue  Seems ok.   Would update the text a bit:
"Remove any other fields from existing users.   This could break existing applications if they rely on these other fields"

biyeun:
Looks good 👍

corehq/apps/custom_data_fields/edit_model.py
```diff
@@ -19,6 +19,7 @@ class CustomDataFieldsForm(forms.Form):
     The main form for editing a custom data definition
     """
     data_fields = forms.CharField(widget=forms.HiddenInput)
+    purge_existing = forms.BooleanField(widget=forms.HiddenInput, required=False, initial=False)
```
esoergel: Oh man this code is all way too abstract.  I apologize on behalf of Ethan of 2 1/2 years ago.

corehq/apps/users/tasks.py
```diff
@@ -172,3 +171,9 @@ def reset_demo_user_restore_task(couch_user, domain):
 
     DownloadBase.set_progress(reset_demo_user_restore_task, 100, 100)
     return {'messages': results}
+
+
+@task
+def remove_unused_custom_fields_from_users_task(domain):
+    from corehq.apps.users.custom_data import remove_unused_custom_fields_from_users
+    remove_unused_custom_fields_from_users(domain)
```
esoergel: :+1:

esoergel (approved):
This looks good to me.  It would be nice if we could display specifically what fields (if any) would be deleted, but I imagine there'd be performance concerns - not sure if you can get that without pulling all users and doing it in python.

kaapstorm (approved):

czue:
@sheelio updated text. just waiting for a build to run.... :snail:

corehq/apps/custom_data_fields/edit_model.py
czue: hahaha it did take me a long time to figure out how some pieces worked, though the ability to be generically dropped in in multiple places is nice. i did at first try to reconcile the logic that handles the removal in the edit entity form with this, but that was too coupled to the view / request to be able to do it.

PR #14791, opened 2017-02-07
Remove old multiple locations flag

esoergel:
Let's let tests run on this and maybe give it a few more days just in case.


czue:
does IPM not need this anymore?

esoergel:
@czue I don't believe so, I've been hassling people to drop the flag for a little while and someone from that domain (Mike, I think?) removed it: https://www.commcarehq.org/hq/flags/edit/multiple_locations/

I do want to keep this PR open for a little while longer just in case.  The last domain switched the flag off about a week ago, IIRC.

czue:
awesome

esoergel:
Cool, this should be g2g by now - haven't heard of any problems warranting a revert.  @czue @NoahCarnahan

czue:
good :skull:

PR #14793, opened 2017-02-07
fix settings text

biyeun:
a commit made yesterday was causing errors on the yaml parsing
@orangejenny


orangejenny:
thanks for catching & fixing

orangejenny (approved):

PR #14794, opened 2017-02-07
Always install ipython on prod environments

esoergel:
If I ever want to run a django shell, I want ipython available.  I don't
think there's a real downside to this, but interested to hear if anyone
disagrees.

### DIFF
--- a/requirements/dev-requirements.txt
+++ b/requirements/dev-requirements.txt
@@ -15,7 +15,7 @@ fixture
 sniffer
 psutil  # for memory profiling
 django-extensions
-ipython==4.2.1
+ipython==5.2.2
 ipdb==0.9.4
 wheel==0.29.0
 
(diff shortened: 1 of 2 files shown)


requirements/dev-requirements.txt
```diff
@@ -15,7 +15,7 @@ fixture
 sniffer
 psutil  # for memory profiling
 django-extensions
-ipython==4.2.1
+ipython==5.2.2
```
benrudolph: this doesn't work with ipdb do you think we can move this version back down?

esoergel: Are you sure?  I just tried it out and it works for me.  Could be a mac vs linux thing.  If that's the case, then yeah, we'll probably have to go with 4.x

esoergel: FYI, I did find some issues with ipdb and ipython5, but it looks like they've been resolved: https://github.com/gotcha/ipdb/pull/108

benrudolph: yes! nice that was the error i was getting. auto complete also wasn't working unless i upgraded ipdb

PR #14795, opened 2017-02-07
Fixes issue with static widths on select2s in case management + live preview

biyeun:
response to this ticket
http://manage.dimagi.com/default.asp?245920

codebuddy @calellowitz 
cc @orangejenny

### DIFF
--- a/corehq/apps/style/static/style/less/_hq/forms.less
+++ b/corehq/apps/style/static/style/less/_hq/forms.less
@@ -112,13 +112,7 @@ legend .subtext {
 }
 
 #case-config-ko, #usercase-config-ko {
-    @select2Width: 210px;
-
-    .case-config-select2s(@select2Width);
-
-    .wide-select2s {
-        .case-config-select2s(@select2Width * 1.5);
-    }
+    .case-config-select2s(100%);
 }
 
 // hack to fix issues with placeholder not showing up fully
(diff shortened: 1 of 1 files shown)


calellowitz (approved):

orangejenny (changes requested):
Doesn't seem to fix issue (will attach screenshot)

orangejenny:
screen shot 2017-02-07 at 4 09 39 pm

biyeun:
huh that's odd.

biyeun:
@orangejenny fixed. Finally had some time to sit with this. slightly non-trivial trying to figure out the core issue....

orangejenny:
Hey, the `100%` width breaks the `text-overflow: ellipsis` on the case property name select, causing overlap pretty similar to the original issue:
screen shot 2017-03-07 at 11 02 16 am

The styling in this area is a pain :-/  Wonder if there's a different approach we could take, maybe wrapping the question select2 under the case property select2? Or possibly punting on this altogether since @millerdev is pretty close to getting user properties in the form builder released, which will get rid of the loading area altogether?

biyeun:
@orangejenny IMO this is still better and more manageable than having it go completely off the table cell and distort the layout. At least now it's contained. Believe me I wrestled with this thing.

orangejenny:
Ha, I believe you, I've had fun times working with this code, too.

Given that I think user properties in form builder will make this a non-issue soon, doesn't seem worth spending more of your time on it.

orangejenny (approved):

PR #14799, opened 2017-02-07
Coerce integer custom location data to strings

proteusvacuum:
Yipes.
When doing a bulk upload with integers as custom data, it was breaking restores with `ERROR: TypeError: cannot serialize 9 (type int)`, because `ElementTree` [doesn't do it for you](http://stackoverflow.com/questions/19534288/can-xml-etree-elementtree-write-integer-values-for-a-given-element).

@esoergel @nickpell @emord

### DIFF
--- a/corehq/apps/locations/bulk_management.py
+++ b/corehq/apps/locations/bulk_management.py
@@ -143,7 +143,7 @@ class LocationStub(object):
         self.do_delete = do_delete
         self.external_id = str(external_id) if isinstance(external_id, int) else external_id
         self.index = index
-        self.custom_data = custom_data or {}
+        self.custom_data = {key: unicode(value) for key, value in custom_data.items()} if custom_data else {}
         self.uncategorized_data = uncategorized_data or {}
         if not self.location_id and not self.site_code:
             raise LocationExcelSheetError(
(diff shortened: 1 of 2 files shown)


czue:
would it make sense to just coerce all custom data properties to strings during restore if they aren't alread?

proteusvacuum:
Yeah, we weren't sure of the best place to put this. I figured that catching it early in the import process was the safest, so that if we are ever using these values anywhere else in the code they are properly saved as strings, and will be treated as such by everything else.

Doing it at restore point would have been better if this bug was widespread. We did a quick pass of the server logs and didn't find any other instances of this error popping up for other domains. We also manually changed all the eNikshay values to strings so that we didn't have to do a new upload. 

Do you think it's still worth catching it in the restore? I don't think that would be hard to add.

czue:
that's logical. doing it in restore doesn't seem like it hurts and would be consistent with what we do in lookup tables, but also seems fine not to worry about it if we don't expect it to ever crop up.

esoergel:
@czue I had the same thought, but offlined with @proteusvacuum and he pointed out that we get an error email if/when this happens, so if it does come up again, we'll for sure hear about it.

proteusvacuum:
@esoergel: https://github.com/dimagi/commcare-hq/pull/14822

PR #14803, opened 2017-02-08
fix operator

nickpell:
@NoahCarnahan


nickpell:
self-merging since tests are broken

NoahCarnahan:
👍

PR #14804, opened 2017-02-08
remove erroring tests

nickpell:
these fail on [travis](https://api.travis-ci.org/jobs/199330608/log.txt?deansi=true) and in the docker setup (but not ./manage.py test ...).  temporarily removing while I figure out what's going on, then adding back and figuring out if there are any implications for already migrated data


nickpell:
self-merging because tests are broken

czue:
(canceled build)

PR #14809, opened 2017-02-08
Edit forms route

wpride:
This works but I want to clean it up before merging. But before doing that want to ask - how do you think we should roll this out? All at once or under random toggle? QA? 

@benrudolph


corehq/apps/cloudcare/views.py
```diff
@@ -401,6 +401,7 @@ def form_context(request, domain, app_id, module_id, form_id):
 
     root_context = {
         'form_url': form_url,
+        'formplayer_url' : settings.FORMPLAYER_URL,
```
dimagimon: E203 whitespace before ':'

benrudolph:
i'm personally a fan of a random toggle and it seems pretty easy to add in, what do you think?

corehq/apps/cloudcare/views.py
benrudolph: this will be wrong for regular touchforms

corehq/apps/cloudcare/touchforms_api.py
```diff
@@ -61,7 +61,7 @@ def get_full_context(self, root_extras=None, session_extras=None):
         # always tell touchforms to include footprinted cases in its case db
         session_data["additional_filters"] = {"footprint": True}
         session_data.update(session_extras)
-        xform_url = reverse("xform_player_proxy")
+        xform_url = root_extras.get('formplayer_url')
```
benrudolph: this may break for sms

wpride: @gcapalbo does SMS use this route?

corehq/apps/cloudcare/views.py
dimagimon: E203 whitespace before ':'

corehq/apps/cloudcare/touchforms_api.py
```diff
@@ -61,7 +61,10 @@ def get_full_context(self, root_extras=None, session_extras=None):
         # always tell touchforms to include footprinted cases in its case db
         session_data["additional_filters"] = {"footprint": True}
         session_data.update(session_extras)
-        xform_url = reverse("xform_player_proxy")
+        if toggles.EDIT_FORMPLAYER.enabled(session_data.get('domain')):
```
wpride: @gcapalbo and if so is there a way we can differentiate that its coming from SMS instead of from edit forms or old cloudcare?

corehq/apps/cloudcare/touchforms_api.py
gcapalbo: i checked and doesn't look like the method where this change happened ever gets called in the sms or ivr form workflows. but i ran the test suites for both to be sure and they passed, so we're good

wpride:
@gcapalbo Awesome, thank you!

corehq/apps/cloudcare/touchforms_api.py
gcapalbo: (answered above, can leave as is)

gcapalbo: although in case you do ever need to differentiate i believe you can use the device id param that gets passed in

benrudolph (approved):

PR #14806, opened 2017-02-08
Rename Cloud[Cc]are to Web Apps 

kaapstorm:
([FB 247083](http://manage.dimagi.com/default.asp?247083))

Not sure about [this reference to CloudCare][1]. Do our existing customers call it CloudCare or Web Apps?

@sravfeyn @jjwack cc buddy @benrudolph 

  [1]: https://github.com/dimagi/commcare-hq/blob/411868d46b7459e8b0d3fcc7951b192db78c08fc/corehq/apps/accounting/subscription_changes.py#L373-L373

PR #14810, opened 2017-02-08
No more old CloudCare

kaapstorm:
cf. [FB 247083](http://manage.dimagi.com/default.asp?247083), #14806 

This, because @benrudolph says, "only 7 domains are still using old cloudcare"

PRing separately. 

@sravfeyn @dimagi/product


benrudolph (approved):

biyeun:
HOORAY

PR #14817, opened 2017-02-08
Update Location Feature Flags

sheelio:
@esoergel 

(Might be worth testing locally if easy)


esoergel:
Confirmed locally that there are no syntax errors.

PR #14818, opened 2017-02-08
Kill eula modal

orangejenny:
Pretty sure this is no longer used. The registrations flows (signup and web user accept invitation) use a different eula modal (https://github.com/dimagi/commcare-hq/blob/master/corehq/apps/registration/templates/registration/partials/eula_modal.html), and the eula in the footer is now a real link, not a modal (https://github.com/dimagi/commcare-hq/blob/master/corehq/apps/style/templates/style/base.html#L347).

@biyeun / @gcapalbo


gcapalbo (approved):
looks ok to me. do you think it's worth trying to put some kind of soft assert in here to make sure it's not used before removing it, or do you feel confident enough that it's not worth adding a soft assert?

benrudolph:
doesn't this get used for cloudcare?

orangejenny:
Dug some more, this is still used (it's a catch-all that blocks access to ~all of HQ for users who have somehow not agreed to the EULA). Might look at renaming or commenting to make this clearer (my confusion started because we have two `eula_modal.html` files).

PR #14821, opened 2017-02-08
release copy case list config

czue:
this is a PR into https://github.com/dimagi/commcare-hq/pull/14820

@orangejenny / @dimagi/product - thought i'd chance it and see if we thought there was more work needed to release this.

current UI

image

If no available source modules
image

### DIFF
--- a/corehq/apps/app_manager/templates/app_manager/v2/module_view.html
+++ b/corehq/apps/app_manager/templates/app_manager/v2/module_view.html
@@ -74,9 +74,9 @@
                 </li>
               {% endif %}
             {% endfor %}
-            {% if details %}{% if  module.module_type != 'basic' or request|toggle_enabled:'COPY_CASE_CONFIGS' %}
+            {% if details %}
                 <li><a href="#module-actions" data-toggle="tab">{% trans "Actions" %}</a></li>
-            {% endif %}{% endif %}
+            {% endif %}
           {% endif %}
         {% endblock %}{# endblock tab-headers #}
       </ul>
@@ -100,11 +100,11 @@
               </div>
             {% endif %}
           {% endfor %}
-          {% if details %}{% if  module.module_type != 'basic' or request|toggle_enabled:'COPY_CASE_CONFIGS' %}
+          {% if details %}
             <div class="tab-pane" id="module-actions">
                 {% include 'app_manager/v2/partials/module_actions.html' %}
             </div>
-          {% endif %}{% endif %}
+          {% endif %}
         {% endif %}
       {% endblock %}
     </div>
--- a/corehq/apps/app_manager/templates/app_manager/v2/module_view.html
+++ b/corehq/apps/app_manager/templates/app_manager/v2/module_view.html
@@ -75,7 +75,7 @@
                 </li>
               {% endif %}
             {% endfor %}
-            {% if details and app|get_available_modules_for_case_list_configuration:module %}
+            {% if details and app.advanced_app_builder and app|get_available_modules_for_case_list_configuration:module %}
                 <li><a href="#module-actions" data-toggle="tab">{% trans "Actions" %}</a></li>
             {% endif %}
           {% endif %}
@@ -101,7 +101,7 @@
               </div>
             {% endif %}
           {% endfor %}
-          {% if details %}
+          {% if details and app.advanced_app_builder%}
             {% with app|get_available_modules_for_case_list_configuration:module as available_modules %}
             {% if available_modules %}
             <div class="tab-pane" id="module-actions">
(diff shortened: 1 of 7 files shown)


orangejenny:
I'm open to releasing this if it's likely to be useful to self starts, defer to your judgment on that.

What would you think of hiding the Actions tab altogether if there aren't two+ modules with the same case type? It's one less tab to distract people, and I think more with what we usually do (e.g., non-case modules in v2 aren't going to have case management/list/detail tabs at all).

corehq/apps/app_manager/templates/app_manager/v2/module_view.html
```diff
@@ -271,9 +271,9 @@
                 </li>
               {% endif %}
             {% endfor %}
-            {% if details %}{% if  module.module_type != 'basic' or request|toggle_enabled:'COPY_CASE_CONFIGS' %}
+            {% if details %}
```
orangejenny: Let's also put this behind the `advanced_app_builder` setting in v2, which is how we're planning to hide semi-advanced functionality. Similar code at https://github.com/dimagi/commcare-hq/blob/master/corehq/apps/app_manager/templates/app_manager/v2/partials/case_config.html#L76

czue:
> I'm open to releasing this if it's likely to be useful to self starts, defer to your judgment on that.

I'm not actually sure. What I want to start is for GS to discover it and say whether it's useful so we can decide. I could make it superuser only, but then I worry about the problem of FMs having different UIs from partners which has come up as a source of confusion historically. But am starting to think that this is the best option for now (superuser + flag).

> What would you think of hiding the Actions tab altogether if there aren't two+ modules with the same case type? It's one less tab to distract people, and I think more with what we usually do (e.g., non-case modules in v2 aren't going to have case management/list/detail tabs at all).

I debated this and decided not to because of an old joel article that stuck with me that is probably no longer the latest in UX thinking. Some discussion on the matter [here](http://ux.stackexchange.com/questions/12756/dont-hide-or-disable-menu-items) and [here](http://ux.stackexchange.com/questions/34271/greyed-out-vs-invisible). 

Am guessing what's most important is being consistent, so if HQ is moving towards a world where we hide functionality that isn't possible to use then seems like that's what we should do. 

Is there an equivalent of a UX "style" guide? It could be useful to record some of these things so the team can be aligned and refer to them.

corehq/apps/app_manager/templates/app_manager/v2/module_view.html
czue: will do

orangejenny:
Worthwhile reading, thanks for the links. I like the Spolsky desription - "Users see the disabled menu item that they want to click on, and are left entirely without a clue of what they are supposed to do to get the menu item to work." For this, there isn't something users can/should do to make the action work, it just isn't relevant to their app, which makes me inclined to leave off the tab so people don't have the (mild) frustration of seeing a feature they can't use. Reg from case list is a contrasting example, where people can understand why it's useful, but their app has to be configured in non-obvious way for it to work properly, so we disable the action and explain how you can go about enabling it.

I think you can make a reasonable argument in either direction in this case. There's some benefit to knowing this functionality exists even if your current app doesn't need it, but HQ has so many features that I'm not sure that knowing about one more is useful. If you leave the tab, it might help to also leave the "Overwrite..." header so the unavailable functionality isn't a mystery.

Feel free to self-merge once the advanced_app_builder check is added to v2.

> Is there an equivalent of a UX "style" guide? It could be useful to record some of these things so the team can be aligned and refer to them.

Agree this is useful, and we're planning to update the HQ style guide [(this one)](https://www.commcarehq.org/styleguide/) more frequently with design best practices. I don't think this particular issue has a super clear-cut answer, but maybe there's a general principle here, something like "things to consider when adding semi-advanced functionality", that's worth writing down.

czue:
> For this, there isn't something users can/should do to make the action work, it just isn't relevant to their app, which makes me inclined to leave off the tab so people don't have the (mild) frustration of seeing a feature they can't use. 

This is a great point and convinces me that it should be removed entirely.

Thanks for the other thoughts and responses. I'm genuinely unsure whether this should be released or not - I wish there was an even more advanced place to stick it but not sure what that is (sounds like maybe app manager v2 will solve this to some degree). Will ruminate on it some more and propose something when I make the other updates. Appreciate the input!

czue:
@orangejenny ok I updated this to hide the tab if nothing can be done and with the advanced flag for v2. I think that's everything but let me know if I missed sometihng.

orangejenny:
looks good to me, thanks for updates!

orangejenny (approved):

gcapalbo (approved):
looks good. is this ready to release? feel free to self merge if so - wasn't sure from the comments

PR #13100, opened 2016-09-04
Tranform dates in exports to request timezone

benrudolph:
@czue @dannyroberts been meaning to do this for a while. all dates will now be in iso format instead of the other datetime format with the Z at the end (except if they want to transform to excel, then it'll be in excel format)

https://dimagi.uservoice.com/forums/194740-data/suggestions/9904056-have-data-export-timezone-match-project-domain-tim

PR #14823, opened 2017-02-08
Je/sms export final

emord:
Replacing #14631 

@benrudolph @gcapalbo 

This export is static, so it doesn't allow the user to create new exports and there's not a list view. Also all domains will use "new" export

cc @proteusvacuum


corehq/apps/export/tests/test_sms_export.py
```diff
@@ -0,0 +1,202 @@
+import datetime
+import json
+import uuid
+from django.test.testcases import SimpleTestCase
+from mock import patch
+
+from couchexport.models import Format
+
+from corehq.apps.export.export import (
```
dimagimon: F401 'ExportFile' imported but unused
F401 '_Writer' imported but unused
F401 '_get_writer' imported but unused
F401 '_write_export_instance' imported but unused

corehq/apps/export/views.py
```diff
@@ -2283,6 +2286,57 @@ def get_filters(self, filter_form_data, mobile_user_and_group_slugs):
...
+    def _get_filter_form(self, filter_form_data):
+        filter_form = self.filter_form_class(
+            self.domain_object, self.timezone, filter_form_data,
+        )
+        if not filter_form.is_valid():
+            raise ExportFormValidationException()
+        return filter_form
+
+    def _get_export(self, domain, export_id):
+        user = self.request.couch_user
```
dimagimon: F841 local variable 'user' is assigned to but never used

corehq/apps/export/models/new.py
```diff
@@ -920,6 +933,53 @@ def get_filters(self):
...
 
 
+class SMSExportInstance(ExportInstance):
+    type = SMS_EXPORT
+    identifier = None
+    name = "Messages"
+
+    @classmethod
+    def _new_from_schema(cls, schema):
+        columns = [
```
emord: Could also make these nametuple but this felt readable to me since it's only used here

corehq/apps/export/views.py
```diff
@@ -2283,6 +2286,56 @@ def get_filters(self, filter_form_data, mobile_user_and_group_slugs):
         return form_filters
 
 
+class DownloadNewSmsExportView(GenericDownloadNewExportMixin, BaseDownloadExportView):
+    urlname = 'new_export_download_sms'
+    page_title = ugettext_noop("Export SMS")
+    form_or_case = None  # todo: remove this property from exports
```
emord: I think this should get removed from exports totally and replaced with something like `export_type`, which already exists. Want to dig into this, but was going to separate that out into a different PR

benrudolph: 👍 didn't even know that existed

corehq/apps/export/models/new.py
benrudolph: why not define all of these in the system_properties.py file?

corehq/apps/export/models/new.py
```diff
@@ -920,6 +933,53 @@ def get_filters(self):
...
+            main_table.columns.append(ExportColumn(
+                label="Message Log ID",
+                item=ExportItem(
+                    path=[PathNode(name='_id')]
+                ),
+                selected=True,
+            ))
+
+        instance = cls(domain=schema.domain, tables=[main_table])
+        cls._insert_system_properties(instance.domain, schema.type, instance.tables[0])
```
benrudolph: this should already get called on `generate_instance_from_schema`

corehq/apps/export/views.py
```diff
@@ -2283,6 +2286,56 @@ def get_filters(self, filter_form_data, mobile_user_and_group_slugs):
         return form_filters
 
 
+class DownloadNewSmsExportView(GenericDownloadNewExportMixin, BaseDownloadExportView):
```
benrudolph: i would've expected to have seen more views like the edit view and create view. are those not for this PR?

benrudolph (commented):
looking good, couple of questions

corehq/apps/export/models/new.py
emord: Fair I'll update that. originally didn't think of them as "system" properties, but since it's a static thing, i guess they all are

corehq/apps/export/views.py
emord: Going to wait on a refactor if this still exists after old exports are no longer supported

corehq/apps/export/views.py
emord: Since this is a "static" report I didn't make those. When you click "SMS Exports" you're automatically taken to the download page because there's nothing to edit or create

corehq/apps/export/models/new.py
emord: Looks like that doesn't happen because I'm creating the schema without any `group_schemas` since all properties are system properties. Do you think it's worth moving those properties over to a `group_schema` in `SMSExportDataSchema`?

corehq/apps/export/models/new.py
```diff
@@ -920,6 +933,53 @@ def get_filters(self):
...
+    identifier = None
+    name = "Messages"
+
+    @classmethod
+    def _new_from_schema(cls, schema):
+        columns = [
+            ("Contact Type", "couch_recipient_doc_type", DOC_TYPE_TRANSFORM),
+            ("Contact ID", "couch_recipient", None),
+            ("Timestamp", "date", None),
+            ("Owner Name", "couch_recipient", OWNER_ID_TRANSFORM),
```
gcapalbo: i don't think the owner_id transform will work here since it doesn't handle cases right? the couch_recipient could point to a case, mobile worker, web user, or be blank

corehq/apps/export/esaccessors.py
```diff
@@ -26,6 +27,12 @@ def get_case_export_base_query(domain, case_type):
             .sort("opened_on"))
 
 
+def get_sms_export_base_query(domain):
+    return (SMSES()
+            .domain(domain)
+            .sort("date"))
```
gcapalbo: i forgot to mention, we should also exclude any messages that are outbound and had errors. we can do that by excluding those messages with `direction` equal to [OUTGOING](https://github.com/dimagi/commcare-hq/blob/master/corehq/apps/sms/models.py) and `processed` False

corehq/apps/export/models/new.py
```diff
@@ -920,6 +933,53 @@ def get_filters(self):
...
+                    selected=True
+                )
+                for col in columns
+            ],
+        )
+        if schema.include_metadata:
+            main_table.columns.append(ExportColumn(
+                label="Message Log ID",
+                item=ExportItem(
+                    path=[PathNode(name='_id')]
```
gcapalbo: does this name reference work? i know we call it `couch_id` on the SMS (postgres) model and that should get copied to to ES too

gcapalbo:
@emord once you're done should we put this up on staging for UAT/QA before release?

corehq/apps/export/models/new.py
emord: `_id` will be the id of the document in ES. I thought those would be the same because they are the same for the couch documents that get moved over.

It looks like `SMS` is fully on SQL now? In that case I think this should be changed to `id` since we would want the postgres id? We could (maybe should) change the indexer to make elasticsearch and postgres the same

corehq/apps/export/esaccessors.py
emord: `processed` isn't included in the sms mapping so this will require a reindex https://github.com/dimagi/commcare-hq/blob/8667a4159ad39d3307f92a3812d74006d610fd1b/corehq/pillows/mappings/sms_mapping.py#L36

corehq/apps/export/models/new.py
gcapalbo: yeah SMS models are only in postgres now.  i'd prefer to keep couch_id as the uuid for an SMS though because we may need to shard it some day and then we couldn't use the integer id anymore

i don't know if _id in the ES doc will be the right thing, it might be if it gets populated by [this](https://github.com/dimagi/commcare-hq/blob/2e0d3a06d394753b188b3fbcdc97b3aaffbaf901/corehq/pillows/sms.py#L47)

corehq/apps/export/esaccessors.py
gcapalbo: hmm, ok. there's not too many of them anyway, maybe it's easier to just exclude them when building the export?

corehq/pillows/mappings/sms_mapping.py
```diff
@@ -1,65 +1,52 @@
...
-                    "type": "custom",
-                    "tokenizer": "whitespace",
-                    "filter": ["lowercase"]
-                },
-            }
-        }
+SMS_INDEX = es_index("smslogs_2017-02-09")
+SMS_MAPPING = {
+    '_meta': {
+       'comment': 'Autogenerated [corehq.apps.sms.models.SMSLog] mapping from ptop_generate_mapping 09/23/2013',
```
dimagimon: E121 continuation line under-indented for hanging indent

corehq/apps/export/tests/test_sms_export.py
```diff
@@ -0,0 +1,238 @@
...
+                            u'Direction',
+                            u'Message',
+                            u'Type',
+                        ],
+                        u'rows': rows_value,
+                    }
+                }
+            )
+
+    @patch('corehq.apps.export.transforms._cached_case_id_to_case_name')
```
dimagimon: F811 redefinition of unused 'test_export_doc_type_transform' from line 126

corehq/apps/export/models/new.py
emord: Offlined. _id is the correct thing to use

corehq/apps/export/models/new.py
emord: Yeah. you're right. I misread the owner id transform. Created a new one https://github.com/dimagi/commcare-hq/pull/14823/commits/6c8b135a70731bfc2c7e650da7767b42cf9b4469#diff-0111f3dfa65bf39568e054504f4430afR23

emord:
This is on staging now

corehq/apps/export/models/new.py
benrudolph: ah gotcha, understood. i don't think we're in the greatest situation, essentially we're bypassing logic meant to add all of this stuff in smart ways, but since this is actually a static export, it doesn't actually matter that much. begs the question of whether we should model static exports slightly differently. in any case i think this is a fine workaround. thanks for looking into it

corehq/apps/export/views.py
benrudolph: right makes sense thanks

corehq/apps/export/models/new.py
emord: Yeah. I agree that we should look into making a distiniction between "static" exports and form and case ones. I'm hesitant to start that refactor with base classes before we make another static export though.

corehq/apps/es/tests/test_sms.py
```diff
@@ -0,0 +1,35 @@
...
+        json_output = {
+            "query": {
+                "filtered": {
+                    "filter": {
+                        "and": [
+                            {"term": {"domain.exact": "demo"}},
+                            {
+                                "not": {
+                                    "and": (
+                                        {"term": {"direction": "o"}},
```
gcapalbo: i noticed this in the filter too, that direction is lowercase o instead of capital O. is that because ES is doing case-insensitive matching?

emord: Right. By default it is case insensitive https://github.com/dimagi/commcare-hq/blob/master/corehq/pillows/base.py#L9

gcapalbo: :+1: thanks

gcapalbo:
looks great @emord, i just ran it on staging. would it be hard to convert the timestamps to the timezone of the project rather than have them be utc?

benrudolph:
there was an effort to do that here: https://github.com/dimagi/commcare-hq/pull/13100. got bogged down in product stuff

benrudolph (approved):

gcapalbo:
this is ready to go from my side - @dimagi/product / @sheelio just let us know once you have a chance to check it out on staging for uat

sheelio:
Did a pass. Looks generally good - minor feedback below. 
- After clicking "Download" and the file is downloaded, I still see the "Filters Inactive While Download Is In Progress" overlay.  I think form or case exports remove that bubble once done. 
- The data file is inside a folder inside a zip (guessing we can drop the folder if easy)
- Checking this doesn't integrate with the saved / daily export stuff (which is fine)

emord:
@sheelio 

Just tried a form export and the overlay is still present once the download is done

I think the data file should be in a zip because text is highly compressible so there will likely be good gains for a csv file

It doesn't integrate with saved/daily export, but I could look into that if we determine it's important

sheelio:
@emord 

Okay, all sounds good.  I was mainly pushing to get rid of the extra directory that you get inside the CSV (but its really minor so okay to just merge as is).

emord:
kicked off the preindex on prod. Should be ready for merging on monday

gcapalbo:
:+1: great thanks @emord

emord:
preindexing finished. this should be good to merge.

FYI @NoahCarnahan I didn't run this on india/swiss, but those should be quick to index

PR #14830, opened 2017-02-08
update ptop_reindexer_v2

nickpell:


corehq/apps/hqcase/management/commands/ptop_reindexer_v2.py
```diff
@@ -89,14 +98,9 @@ def confirm():
         reindexer = REINDEX_FNS[index]()
         reindexer_options = {
             key: value for key, value in options.items()
-            if value is not None and key in [option.dest for option in self.option_list]
+            if value is not None
```
snopoke: Won't this pass other options through as well like 'verbosity' or 'traceback'?

nickpell: good point.  I was testing this in django 1.10 earlier.

nickpell:
✅

PR #14832, opened 2017-02-08
Je/ilsgateway test

emord:
Opening for tests. The first commit was where the real gains were. There's some issues here where tests make changes to the users, locations etc and don't clean it up. so it's pretty hard to transfer some of them over.

buddy @proteusvacuum


custom/ilsgateway/tests/test_handlers.py
```diff
@@ -0,0 +1,1313 @@
...
+from corehq.apps.reminders.util import get_two_way_number_for_recipient
+from corehq.apps.sms.api import incoming
+from corehq.apps.users.models import CommCareUser
+from corehq.util.translation import localize
+from corehq.util.test_utils import flag_enabled
+from custom.ilsgateway.models import (
+    DeliveryGroups, SupplyPointStatus, SupplyPointStatusTypes, SupplyPointStatusValues
+)
+from custom.ilsgateway.tanzania.reminders import ARRIVED_HELP, ARRIVED_DEFAULT, ARRIVED_KNOWN, EMG_ERROR, EMG_HELP
+from custom.ilsgateway.tanzania.reminders import (
```
dimagimon: F401 'SOH_BAD_FORMAT' imported but unused
F401 'STOCKOUT_CONFIRM' imported but unused
F401 'STOCKOUT_HELP' imported but unused
F401 'SUBMITTED_INVALID_QUANTITY' imported but unused
F401 'SUPERVISION_CONFIRM_NO' imported but unused
F401 'SUPERVISION_CONFIRM_YES' imported but unused
F401 'YES_HELP' imported but unused

emord:
@dimagi/test-team

custom/ilsgateway/tests/test_handlers.py
dimagimon: F401 'YES_HELP' imported but unused

custom/ilsgateway/tests/handlers/arrived.py
```diff
@@ -1,38 +0,0 @@
-from custom.ilsgateway.tanzania.reminders import ARRIVED_HELP, ARRIVED_DEFAULT, ARRIVED_KNOWN
-from custom.ilsgateway.tests.handlers.utils import ILSTestScript
-
-
-class ILSArrivedTest(ILSTestScript):
```
snopoke: are all of these duplicates or just no longer needed?

benrudolph: they're all consolidated in `test_handlers` i believe. github is just not showing the diff

emord: Yeah that's correct. The benefit here is from running the setup once instead of 20x. Although if https://github.com/dimagi/commcare-hq/pull/14840 fixes prepare_domain then it's possible that this won't have as large of an effect as noted above

emord:
:white_check_mark: looking at PRs around it, it only looks like it shaved off a min or so from total run time :/ Hard to tell if it didn't do much, or if it would be faster on average and was a chance slow build

snopoke:
@emord were you comparing the 'Total' travis time or the 'Elapsed' time. 

One reason why we may not see such a big speedup is that the tests might be split over the different builds (due to how to split the test suite up) and so we'd still incur the cost of multiple setups.

millerdev:
> tests might be split over the different builds (due to how to split the test suite up) and so we'd still incur the cost of multiple setups.

I think this could be avoided if a module- or package-level setup/teardown were used. I'm pretty sure I made the `divided-we-run` plugin keep all tests in a module or package setup/teardown context together. The disadvantage of doing that is it can create a huge block of stuff that gets allocated to one test node, boosting its runtime relative to other nodes.

Aside: I can't remember if or why I chose to split up tests with `setUpClass`/`tearDownClass`. Possibly because of the disadvantage mentioned above.

PR #14836, opened 2017-02-09
site_code to lower-case in bulk-location-upload

sravfeyn:
http://manage.dimagi.com/default.asp?246874 @nickpell @esoergel


kaapstorm (approved):

esoergel:
Awesome, thanks for the super quick turnaround, sravs!

PR #14838, opened 2017-02-09
test tweaks

snopoke:
@dimagi/test-team 

```
Before: Ran 20 tests in 53.440s
After: Ran 34 tests in 17.672s
```

Extra test count is because tests marked with `run_with_all_backeds` are run twice but only counted once.

The change was to split test classes based on the backed so that the setup & tear down could be done once instead of once per test.

### DIFF
--- a/corehq/apps/cloudcare/tests/test_api.py
+++ b/corehq/apps/cloudcare/tests/test_api.py
@@ -30,7 +30,7 @@ class CaseAPITestMixin(object):
 
     @classmethod
     def do_setup(cls):
-        create_domain(cls.domain)
+        cls.project = create_domain(cls.domain)
         cls.password = "****"
 
         def create_user(username):
@@ -64,6 +64,7 @@ class CaseAPITestMixin(object):
     def do_teardown(cls):
         for user in cls.users:
             user.delete()
+        cls.project.delete()
         clear_toggle_cache(toggles.CLOUDCARE_CACHE.slug, TEST_DOMAIN, toggles.NAMESPACE_DOMAIN)
         FormProcessorTestUtils.delete_all_cases_forms_ledgers(TEST_DOMAIN)
 
@@ -244,7 +245,7 @@ class CaseAPIMiscTests(TestCase):
     def setUpClass(cls):
         super(CaseAPIMiscTests, cls).setUpClass()
         cls.domain = uuid.uuid4().hex
-        create_domain(cls.domain)
+        cls.project = create_domain(cls.domain)
         cls.username = uuid.uuid4().hex
         cls.password = "****"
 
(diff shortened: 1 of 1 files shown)


corehq/apps/cloudcare/tests/test_api.py
```diff
@@ -1,81 +1,74 @@
...
+from toggle.shortcuts import update_toggle_cache, clear_toggle_cache
 
 TEST_DOMAIN = "test-cloudcare-domain"
 
 
-class CaseAPITest(TestCase):
-    """
-    Tests some of the Case API functions
-    """
+class CaseAPIBaseTest(object):
```
emord: Could possibly change this to `CaseAPIBaseTestMixin` to make it clear it doesn't generate a test case?

corehq/apps/cloudcare/tests/test_api.py
```diff
@@ -1,81 +1,74 @@
...
-        super(CaseAPITest, cls).setUpClass()
+    def do_setup(cls):
         create_domain(cls.domain)
         cls.password = "****"
 
         def create_user(username):
+            username = format_username(username, cls.domain)
+            ccu = CommCareUser.get_by_username(username)
+            if ccu:
+                ccu.delete()
```
emord: Nit: This pattern always frustrates me. I understand how it's useful when developing tests, but why not use `addCleanup(ccu.delete)` ? If a test doesn't clean up properly it makes it really difficult to add tests to the app and reason about which other tests are leaving artifacts around.

Also it looks like this is getting cleaned up in the teardown so this could just be removed

corehq/apps/cloudcare/tests/test_api.py
```diff
@@ -236,6 +206,66 @@ def testFiltersWithFootprint(self):
...
+        cls.do_teardown()
+        super(CaseListApiSQLTests, cls).tearDownClass()
+
+
+class CaseAPIMiscTests(TestCase):
+    @classmethod
+    def setUpClass(cls):
+        super(CaseAPIMiscTests, cls).setUpClass()
+        cls.domain = uuid.uuid4().hex
+        create_domain(cls.domain)
```
emord: Does this domain get cleaned up?

corehq/apps/cloudcare/tests/test_api.py
snopoke: Sorry, wasn't meant to leave that in there. Was just an artifact of the test failing in setup and using reuse db. I'll clean it up

corehq/apps/cloudcare/tests/test_api.py
snopoke: doesn't look like it. Thanks, will add that.

corehq/apps/cloudcare/tests/test_api.py
emord: Ha. No worries. I think I was just frustrated from trying to reason through some similar issues in ilsgateway tests

snopoke:
@emord updated. Previous test passed and ran test locally after changes so I've killed the build.

PR #14844, opened 2017-02-09
rely on transaction rollback for cleanup

nickpell:
testing if this works because of [transactions](https://docs.djangoproject.com/en/1.10/topics/testing/tools/#testcase)

@emord

### DIFF
--- a/corehq/apps/hqadmin/tests/test_prbac.py
+++ b/corehq/apps/hqadmin/tests/test_prbac.py
@@ -28,22 +28,9 @@ class TestCchqPrbacBootstrap(TestCase):
 
     @classmethod
     def setUpClass(cls):
+        super(TestCchqPrbacBootstrap, cls).setUpClass()
         Grant.objects.all().delete()
         Role.objects.all().delete()
-
-    @classmethod
-    def tearDownClass(cls):
-        # re-bootstrap for other tests
-        call_command('cchq_prbac_bootstrap', testing=True)
-
-        DefaultProductPlan.objects.all().delete()
-        SoftwarePlanVersion.objects.all().delete()
-        SoftwarePlan.objects.all().delete()
-        SoftwareProductRate.objects.all().delete()
-        SoftwareProduct.objects.all().delete()
-        FeatureRate.objects.all().delete()
-        Feature.objects.all().delete()
-        call_command('cchq_software_plan_bootstrap', testing=True)
 
     def test_dry_run(self):
         """


corehq/apps/hqadmin/tests/test_prbac.py
```diff
@@ -28,23 +28,10 @@ class TestCchqPrbacBootstrap(TestCase):
 
     @classmethod
     def setUpClass(cls):
+        super(TestCchqPrbacBootstrap, cls).setUpClass()
         Grant.objects.all().delete()
```
emord: Would this work if you got rid of these teardowns in the setup?

emord (commented):
Did you test this locally by running this test and inspecting the database afterwards? I think that just because this test passed doesn't mean that it doesn't leave around some artifacts that could make a future test fail

corehq/apps/hqadmin/tests/test_prbac.py
nickpell: That should break tests.  When I make these edits tests still pass:
```
diff --git a/corehq/apps/hqadmin/tests/test_prbac.py b/corehq/apps/hqadmin/tests/test_prbac.py
index 61a4d95..296b8f7 100644
--- a/corehq/apps/hqadmin/tests/test_prbac.py
+++ b/corehq/apps/hqadmin/tests/test_prbac.py
@@ -32,6 +32,14 @@ class TestCchqPrbacBootstrap(TestCase):
         Grant.objects.all().delete()
         Role.objects.all().delete()
 
+    @classmethod
+    def tearDownClass(cls):
+        assert Role.objects.count() == 0
+        assert Grant.objects.count() == 0
+        super(TestCchqPrbacBootstrap, cls).tearDownClass()
+        assert Role.objects.count() != 0
+        assert Grant.objects.count() != 0
+
     def test_dry_run(self):
         """
         When --dry-run is passed, no models should be created
```

which looks to be, before the superclass ```tearDownClass``` is called, 
```
Grant.objects.all().delete()
Role.objects.all().delete()
```
still has an effect, and after ```tearDownClass``` is executed, the transaction opened in ```setUpClass``` is rolled back, so the rows reappear in the DB.  Which is consistent with my interpretation of transactions in setup(class)/teardown(class)

emord: Ah. ok so those are needed for some random unclean shutdowns elsewhere. thanks for looking into that

PR #14846, opened 2017-02-09
App manager: stay on same tab when refreshing

orangejenny:
http://manage.dimagi.com/default.asp?236308

This has been asked for a few times. Discovered today that we already have this behavior in accounting. Added it to app manager. Putting in a [QA request](http://manage.dimagi.com/default.asp?247338) so this gets a second set of eyes on it - when app manager nav breaks in prod it's pretty unfortunate.

Review commits individually, as I made some changes to stay-on-tab.js, didn't just move it.

code buddy @gcapalbo 
accounting @nickpell 
app manager @snopoke


snopoke:
Nice, this will be great. Last time I looked at this I though the functionality was in `bootstrap-tab-history-custom.js`. Is there an overlap between these two approaches or are they just completely different?

corehq/apps/app_manager/tests/data/v2_diffs/templates/apps_base.html.diff.txt
```diff
@@ -74,21 +74,19 @@
    {% endif %}
  
  {% endblock %}
-@@ -64,9 +81,11 @@
```
nickpell: What do these mean?

nickpell (approved):

corehq/apps/app_manager/tests/data/v2_diffs/templates/apps_base.html.diff.txt
orangejenny: it's coming from the blank line I deleted in `v2/apps_base.html`

orangejenny:
@snopoke `bootstrap-tab-history-custom.js` still controls the app manager left-hand nav via url hashes, and it already handles refreshing correctly. This is only for the inner tabs on a page (like module view's settings / case mgmt / case list / case detail).

proteusvacuum:
YEEESSSSSSSSSSSSSSSSS

czue:
yes! this is great

gcapalbo (approved):

orangejenny:
QA passed, this is good to go.

PR #14848, opened 2017-02-09
Add username and password from repeater to Nikshay API 

proteusvacuum:
@mkangia @emord

### DIFF
--- a/custom/enikshay/integrations/nikshay/repeater_generator.py
+++ b/custom/enikshay/integrations/nikshay/repeater_generator.py
@@ -38,59 +38,24 @@ class NikshayRegisterPatientPayloadGenerator(BasePayloadGenerator):
 
     def get_payload(self, repeat_record, episode_case):
         """
-        mandatory_fields for Nikshay Registration API as of now
-        scode
-        dcode
-        pname
-        pgender
-        paddress
-        pmob
-        ptbyr
-        pregdate
-        cname
-        caddress
-        cmob
-        dcpulmunory
-        dotname
-        dotdesignation
-        dotmob
-        dotpType
-        dotcenter
-        regBy # Used date_of_diagnosis instead of date_of_registration
-        IP_From # Pending
-        Local_ID
-        tcode # Pending
-        page
-        poccupation
-        dotphi
-        Ptype
-        pcategory # Pending
-        Source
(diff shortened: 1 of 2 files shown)


proteusvacuum:
@NoahCarnahan @mkangia 
This will allow us to just enter the username / password that Nikshay gives us when we create the Repeater. Much easier than adding it in localsettings, etc.

custom/enikshay/integrations/nikshay/tests/test_repeaters.py
```diff
@@ -159,6 +160,21 @@ def test_payload_properties(self):
     def _assert_case_property_equal(self, case, case_property, expected_value):
         self.assertEqual(case.dynamic_case_properties().get(case_property), expected_value)
 
+    def test_username_password(self):
+        episode_case = self._create_nikshay_enabled_case()
+        username = "arwen"
+        password = "adhafang"
```
emord: Did you mean `Hadhafang`?

emord (approved):

custom/enikshay/integrations/nikshay/tests/test_repeaters.py
proteusvacuum: damn, my references are really getting checked these days.

emord: This is the real reason we formed the test team

custom/enikshay/integrations/nikshay/repeater_generator.py
```diff
@@ -38,59 +38,24 @@ def content_type(self):
...
-        # "IP_Address": "mk-ip-address",
-        # "IP_From": "127.0.0.1",
         person_case = get_person_case_from_episode(episode_case.domain, episode_case.get_id)
-
         episode_case_properties = episode_case.dynamic_case_properties()
         person_case_properties = person_case.dynamic_case_properties()
+
+        try:
+            username = repeat_record.repeater.username
+        except AttributeError:
```
mkangia: @proteusvacuum incorporated these changes in another PR. Curious, In what scenario is this expected to raise an exception? `repeater` would always be there and `username` is also always present as a blank string in case not mentioned explicitly.

proteusvacuum: I did this so I didn't have to add a username to all the tests that already existed... so you are right that this is somewhat misleading 🐙

PR #14851, opened 2017-02-09
Externalize js: Signup

orangejenny:
Review commits individually.

The one mildly interesting thing about this one is it makes `static` available as a filter, so that I can use it more easily with inital_page_data: `{% initial_page_data 'path' 'this/is/my/path'|static %}`

@gcapalbo


orangejenny:
tests passing

PR #14852, opened 2017-02-10
New Reminders Models

gcapalbo:


corehq/messaging/scheduling/models/abstract.py
```diff
@@ -0,0 +1,287 @@
...
+        if isinstance(timezone, tzinfo):
+            return timezone
+
+        if isinstance(timezone, basestring):
+            try:
+                return coerce_timezone_value(timezone)
+            except ValidationError:
+                pass
+
+        return pytz.UTC
```
dimagimon: F821 undefined name 'pytz'

corehq/messaging/scheduling/models/abstract.py
```diff
@@ -0,0 +1,287 @@
...
+        elif self.recipient_type == 'Location':
+            location = self.recipient
+            if self.memoized_schedule.include_descendant_locations:
+                location_ids = location.get_descendants(include_self=True).filter(is_archived=False).location_ids()
+            else:
+                location_ids = [location.location_id]
+
+            user_ids = set()
+            for location_id in location_ids:
+                for user in get_all_users_by_location(self.domain, location_id):
```
dimagimon: F821 undefined name 'get_all_users_by_location'

corehq/messaging/scheduling/models/abstract.py
```diff
@@ -0,0 +1,289 @@
...
+
+        if isinstance(value, TimedSchedule):
+            self.timed_schedule_id = value.pk
+        elif isinstance(value, AlertSchedule):
+            self.alert_schedule_id = value.pk
+        else:
+            raise self.UnknownScheduleType()
+
+
+class ScheduleInstance(UUIDGeneratorMixin, ScheduleForeignKeyMixin):
```
gcapalbo: good point, looking at the docs, there's a [cleaner way](https://docs.djangoproject.com/en/1.10/ref/models/fields/#uuidfield) to autogenerate the uuid - will update

corehq/messaging/scheduling/models/abstract.py
```diff
@@ -0,0 +1,155 @@
+from corehq.apps.users.models import CommCareUser, WebUser
```
dimagimon: F401 'CommCareUser' imported but unused
F401 'WebUser' imported but unused

corehq/messaging/scheduling/scheduling_partitioned/dbaccessors.py
```diff
@@ -0,0 +1,47 @@
+from corehq.sql_db.util import (
+    get_object_from_partitioned_database,
+    save_object_to_partitioned_database,
+    run_query_across_partitioned_databases,
+)
+from datetime import datetime
```
dimagimon: F401 'datetime' imported but unused

corehq/messaging/scheduling/scheduling_partitioned/dbaccessors.py
```diff
@@ -0,0 +1,47 @@
...
+    save_object_to_partitioned_database(instance, str(instance.pk))
+
+
+def delete_schedule_instance(instance):
+    from corehq.messaging.scheduling.scheduling_partitioned.models import ScheduleInstance
+
+    if not isinstance(instance, ScheduleInstance):
+        raise ValueError("Expected an instance of ScheduleInstance")
+
+    delete_object_from_partitioned_database(instance, str(instance.pk))
```
dimagimon: F821 undefined name 'delete_object_from_partitioned_database'

corehq/messaging/scheduling/scheduling_partitioned/dbaccessors.py
dimagimon: F401 'datetime' imported but unused

corehq/messaging/scheduling/scheduling_partitioned/dbaccessors.py
```diff
@@ -0,0 +1,55 @@
...
+    get_object_from_partitioned_database,
+    save_object_to_partitioned_database,
+    delete_object_from_partitioned_database,
+    run_query_across_partitioned_databases,
+)
+from django.db.models import Q
+from uuid import UUID
+
+
+def _validate_schedule_instance(obj)
```
dimagimon: E901 SyntaxError: invalid syntax

corehq/messaging/scheduling/scheduling_partitioned/dbaccessors.py
```diff
@@ -0,0 +1,55 @@
...
+    from corehq.messaging.scheduling.scheduling_partitioned.models import ScheduleInstance
```
dimagimon: E113 unexpected indentation

corehq/messaging/scheduling/scheduling_partitioned/dbaccessors.py
```diff
@@ -0,0 +1,55 @@
...
+
+    if not isinstance(obj, ScheduleInstance):
+        raise ValueError("Expected an instance of ScheduleInstance")
+
+
+def _validate_uuid(value)
+    if not isinstance(value, UUID):
```
dimagimon: E113 unexpected indentation

corehq/messaging/scheduling/migrations/0001_initial.py
```diff
@@ -0,0 +1,148 @@
...
+                'abstract': False,
+            },
+        ),
+        migrations.AddField(
+            model_name='timedevent',
+            name='schedule',
+            field=models.ForeignKey(on_delete=django.db.models.deletion.CASCADE, to='scheduling.TimedSchedule'),
+        ),
+        migrations.AddField(
+            model_name='timedevent',
```
emord: Kind of confused by this migration. I would have expected it to be in the `CreateModel`

corehq/messaging/scheduling/models/content.py
```diff
@@ -0,0 +1,45 @@
...
+    def send(self, recipient):
+        print '*******************************'
+        print 'To:', recipient
+        print 'Subject: ', self.subject
+        print 'Message: ', self.message
+        print '*******************************'
+
+
+class SMSSurveyContent(Content):
+    form_unique_id = models.CharField(max_length=126)
```
emord: I think this would be fine, but flagging that this should be 255 based on https://github.com/dimagi/commcare-hq/blob/master/corehq/form_processor/models.py#L185

(i just put it on my backlog to make constants for these things and try to standardize these lengths across HQ)

corehq/messaging/scheduling/scheduling_partitioned/dbaccessors.py
```diff
@@ -0,0 +1,55 @@
...
+def delete_schedule_instance(instance):
+    _validate_schedule_instance(instance)
+    _validate_uuid(instance.schedule_instance_id)
+    delete_object_from_partitioned_database(instance, instance.schedule_instance_id)
+
+
+def get_active_schedule_instance_ids(start_timestamp, end_timestamp):
+    from corehq.messaging.scheduling.scheduling_partitioned.models import ScheduleInstance
+
+    q_expression = Q(
```
emord: could be `active_events_in_time_period` ?

corehq/messaging/scheduling/migrations/0001_initial.py
gcapalbo: yeah, this is just a byproduct of using makemigrations. i don't think django tries to figure out which models have dependencies on each other when determining the order in which to create them, so if a model is created before a foreign key model is, the foreign key gets added later rather than with the initial create

corehq/messaging/scheduling/models/content.py
gcapalbo: actually this points to this: https://github.com/dimagi/commcare-hq/blob/bceffe6caa457bf41c67e9524a41ccb0a8ad0cd6/corehq/apps/app_manager/models.py#L881

corehq/messaging/scheduling/scheduling_partitioned/dbaccessors.py
gcapalbo: good point, gave it a better name here https://github.com/dimagi/commcare-hq/pull/14852/commits/b340f7b78d83c14ba787436c9db0036126fdf146

corehq/messaging/scheduling/scheduling_partitioned/dbaccessors.py
```diff
@@ -0,0 +1,70 @@
...
+    for schedule_instance_id in run_query_across_partitioned_databases(
+        ScheduleInstance,
+        active_filter,
+        values=['schedule_instance_id']
+    ):
+        yield schedule_instance_id
+
+
+def get_schedule_instances_for_schedule(schedule):
+    from corehq.messaging.scheduling.models import AlertSchedule, TimedSchedule, ScheduleForeignKeyMixin
```
dimagimon: F401 'ScheduleForeignKeyMixin' imported but unused

corehq/messaging/scheduling/models/abstract.py
```diff
@@ -0,0 +1,122 @@
+import uuid
+from datetime import datetime
+from dimagi.utils.decorators.memoized import memoized
+from django.db import models
+from corehq.messaging.scheduling.exceptions import (
```
dimagimon: F401 'NoAvailableSchedule' imported but unused
F401 'UnknownScheduleType' imported but unused

corehq/messaging/scheduling/scheduling_partitioned/dbaccessors.py
```diff
@@ -0,0 +1,111 @@
...
+
+def _validate_uuid(value):
+    _validate_class(value, UUID)
+
+
+def get_alert_schedule_instance(schedule_instance_id):
+    from corehq.messaging.scheduling.scheduling_partitioned.models import AlertScheduleInstance
+
+    _validate_uuid(schedule_instance_id)
+    return get_object_from_partitioned_database(AlertScheduleInstance, schedule_instance_id, 'schedule_instance_id')
```
dimagimon: E501 line too long (116 > 115 characters)

corehq/messaging/scheduling/scheduling_partitioned/dbaccessors.py
```diff
@@ -0,0 +1,111 @@
...
+
+
+def get_timed_schedule_instance(schedule_instance_id):
+    from corehq.messaging.scheduling.scheduling_partitioned.models import TimedScheduleInstance
+
+    _validate_uuid(schedule_instance_id)
+    return get_object_from_partitioned_database(TimedScheduleInstance, schedule_instance_id, 'schedule_instance_id')
```
dimagimon: E501 line too long (116 > 115 characters)

corehq/messaging/scheduling/tasks.py
```diff
@@ -0,0 +1,91 @@
+from celery.task import task
+from corehq.messaging.scheduling.models import AlertSchedule, TimedSchedule
```
dimagimon: F401 'AlertSchedule' imported but unused
F401 'TimedSchedule' imported but unused

corehq/messaging/scheduling/scheduling_partitioned/tests/test_dbaccessors.py
```diff
@@ -0,0 +1,282 @@
+import re
```
dimagimon: F401 're' imported but unused

corehq/messaging/scheduling/scheduling_partitioned/tests/test_model_creation.py
```diff
@@ -0,0 +1,57 @@
...
+@partitioned
+class PartitionedModelsTest(TestCase):
+
+    def assertModelExists(self, model_class, db):
+        try:
+            with transaction.atomic(using=db):
+                # We have to let Django rollback this nested transaction if an Exception
+                # is raised otherwise we won't be able to run any more queries.
+                model_class.objects.using(db).count()
+        except ProgrammingError as e:
```
dimagimon: F841 local variable 'e' is assigned to but never used

corehq/messaging/scheduling/scheduling_partitioned/tests/test_dbaccessors_non_partitioned.py
```diff
@@ -0,0 +1,245 @@
...
+    AlertSchedule,
+    TimedSchedule,
+)
+from corehq.messaging.scheduling.scheduling_partitioned.models import (
+    AlertScheduleInstance,
+    TimedScheduleInstance,
+)
+from corehq.util.exceptions import AccessRestricted
+from datetime import datetime, date
+from django.conf import settings
```
dimagimon: F401 'settings' imported but unused

corehq/messaging/scheduling/scheduling_partitioned/tests/test_dbaccessors_non_partitioned.py
```diff
@@ -0,0 +1,245 @@
...
+
+        self.timed_instance2 = self.make_timed_schedule_instance(self.uuid5, schedule_id=self.schedule_id2)
+        save_timed_schedule_instance(self.timed_instance2)
+
+        self.timed_instance3 = self.make_timed_schedule_instance(self.uuid6, schedule_id=self.schedule_id2,
+            active=False)
+        save_timed_schedule_instance(self.timed_instance3)
+
+    def tearDown(self):
+         AlertScheduleInstance.objects.using(self.db).filter(domain=self.domain).delete()
```
dimagimon: E111 indentation is not a multiple of four

corehq/messaging/scheduling/scheduling_partitioned/tests/test_dbaccessors_non_partitioned.py
```diff
@@ -0,0 +1,245 @@
...
+         TimedScheduleInstance.objects.using(self.db).filter(domain=self.domain).delete()
```
dimagimon: E111 indentation is not a multiple of four

corehq/messaging/scheduling/scheduling_partitioned/tests/test_model_creation_non_partitioned.py
```diff
@@ -0,0 +1,33 @@
+import re
```
dimagimon: F401 're' imported but unused

corehq/messaging/scheduling/scheduling_partitioned/tests/test_dbaccessors_non_partitioned.py
```diff
@@ -0,0 +1,245 @@
...
+            active=active,
+            timed_schedule_id=schedule_id or uuid.uuid4(),
+            start_date=date(2017, 3, 1),
+        )
+
+
+class TestSchedulingNonPartitionedDBAccessorsGetAndSave(BaseSchedulingNontPartitionedDBAccessorsTest):
+
+    def tearDown(self):
+        AlertScheduleInstance.objects.using(self.db).filter(domain=self.domain).delete()
```
snopoke: doesn't seem necessary to specify `using` since `self.db` here is always `default`

corehq/messaging/scheduling/models/timed_schedule.py
```diff
@@ -0,0 +1,197 @@
...
+    def set_first_event_due_timestamp(self, instance, start_date=None):
+        """
+        If start_date is None, we set it automatically ensuring that
+        self.next_event_due does not get set in the past for the first
+        event.
+        """
+        if start_date:
+            instance.start_date = start_date
+        else:
+            instance.start_date = ServerTime(datetime.utcnow()).user_time(instance.timezone).done().date()
```
snopoke: What's the purpose of this conversion? I would have assumed that the dates would all be stored in UTC.

corehq/messaging/scheduling/models/timed_schedule.py
```diff
@@ -0,0 +1,197 @@
...
+        return schedule
+
+    def set_simple_monthly_schedule(self, time, days, content, total_iterations=REPEAT_INDEFINITELY):
+        with transaction.atomic():
+            self.schedule_length = self.MONTHLY
+            self.total_iterations = total_iterations
+            self.save()
+
+            for event in self.timedevent_set.all():
+                event.content.delete()
```
snopoke: is this necessary - won't the delete of the event cascade to the content? If so couldn't you just do `self.timedevent_set.all().delete()`

corehq/messaging/scheduling/models/timed_schedule.py
```diff
@@ -0,0 +1,197 @@
...
+                event = TimedEvent(
+                    schedule=self,
+                    order=order,
+                    day=day,
+                    time=time
+                )
+
+                if order > 1:
+                    content.pk = None
+                    content.save()
```
snopoke: So this duplicates the content for each event. Would it be bad to have them all pointing at the same content object?

corehq/messaging/scheduling/scheduling_partitioned/tests/test_dbaccessors_non_partitioned.py
```diff
@@ -0,0 +1,245 @@
...
+    AlertScheduleInstance,
+    TimedScheduleInstance,
+)
+from corehq.util.exceptions import AccessRestricted
+from datetime import datetime, date
+from django.conf import settings
+from django.test import TestCase
+
+
+@only_run_with_non_partitioned_database
```
snopoke: Why is this decorator needed?

corehq/messaging/scheduling/scheduling_partitioned/tests/test_model_creation_partitioned.py
```diff
@@ -0,0 +1,57 @@
...
+class PartitionedModelsTest(TestCase):
+
+    def assertModelExists(self, model_class, db):
+        try:
+            with transaction.atomic(using=db):
+                # We have to let Django rollback this nested transaction if an Exception
+                # is raised otherwise we won't be able to run any more queries.
+                model_class.objects.using(db).count()
+        except ProgrammingError:
+            self.assertTrue(False)
```
snopoke: Could also be `self.fail()` which seems clearer

snopoke (commented):
This is looking good. I didn't review the logic much, mostly just the sharding. 

Was a but confused by the sharding test decorators.

corehq/messaging/scheduling/scheduling_partitioned/tests/test_dbaccessors_non_partitioned.py
gcapalbo: this isn't needed when running with travis since there are different jobs for partitioned vs non partitioned, but it makes a difference when running locally. without it, it would try to run these tests even if i'm running with partitioned settings

corehq/messaging/scheduling/scheduling_partitioned/tests/test_dbaccessors_non_partitioned.py
gcapalbo: using is required because it's a subclass of `PartitionedModel` (introduced in this PR) which requires using() to be called on every queryset as a precaution

corehq/messaging/scheduling/models/timed_schedule.py
gcapalbo: imagine the scenario where it's midnight GMT but you're scheduling a reminder at 10pm eastern for "the next available time" (i.e., start_date is None). datetime.utcnow() would say that's it march 18th, but in eastern time it's still march 17th. so the point of this is that in that we'd use march 17th as the start date to correctly set the next available time.

corehq/messaging/scheduling/models/timed_schedule.py
gcapalbo: it's the other way around - the event has a pointer to the content, so deleting the content will cascade the delete to the event. in that case, i suppose i could spare the delete of the event.

for clarity i may just remove event.delete() and replace it with a bulk delete - it would be just one query that wouldn't be needed but would make things less prone to fail in case something ever changes.

corehq/messaging/scheduling/models/timed_schedule.py
gcapalbo: i had thought about this a bit, and while it makes no difference to the framework, it seemed right to me to keep a 1:1 relationship between event and content, otherwise it would be easy in some future state for someone to think they're only updating the content for one event when instead they're updating multiple

corehq/messaging/scheduling/scheduling_partitioned/tests/test_model_creation_partitioned.py
gcapalbo: agree, will update

corehq/messaging/scheduling/models/abstract.py
```diff
@@ -0,0 +1,124 @@
+import uuid
+from datetime import datetime
```
dimagimon: F401 'datetime' imported but unused

corehq/messaging/scheduling/models/alert_schedule.py
```diff
@@ -0,0 +1,72 @@
+from corehq.messaging.scheduling.models.abstract import Schedule, Event, Broadcast
+from corehq.messaging.scheduling import util
+from datetime import timedelta, datetime
```
dimagimon: F401 'datetime' imported but unused

corehq/messaging/scheduling/tests/test_schedules.py
```diff
@@ -0,0 +1,359 @@
...
+    get_timed_schedule_instances_for_schedule,
+)
+from corehq.messaging.scheduling.models import (
+    TimedSchedule,
+    SMSContent,
+)
+from corehq.messaging.scheduling.tasks import (
+    refresh_timed_schedule_instances,
+)
+from corehq.messaging.scheduling.scheduling_partitioned.models import (
```
dimagimon: F401 'TimedScheduleInstance' imported but unused

corehq/messaging/scheduling/tests/test_schedules.py
```diff
@@ -0,0 +1,359 @@
...
+        self.assertNumInstancesForSchedule(1)
+        [instance] = get_timed_schedule_instances_for_schedule(self.schedule)
+        self.assertTimedScheduleInstance(instance, 0, 1, datetime(2017, 3, 16, 16, 0), True, date(2017, 3, 16),
+            self.user2)
+
+        self.assertEqual(send_patch.call_count, 0)
+
+
+
+@partitioned
```
dimagimon: E303 too many blank lines (3)

corehq/form_processor/backends/sql/dbaccessors.py
```diff
@@ -103,6 +123,10 @@ def hash_doc_ids_python(doc_ids):
     def hash_doc_id_python(doc_id):
         if isinstance(doc_id, unicode):
             doc_id = doc_id.encode('utf-8')
+        elif isinstance(doc_id, UUID):
+            # Hash the 16-byte string
+            doc_id = doc_id.bytes
```
gcapalbo: it seemed like the most consistent thing to do is to just hash the actual bytestring for uuids rather than try and do it to a string representation of it. the same approach is used in `hash_doc_uuid_sql` (@snopoke fyi)

corehq/sql_db/models.py
```diff
@@ -0,0 +1,73 @@
...
+    A Model Manager which requires that .using() is called before any
+    other methods like .filter() or .get().
+    """
+
+    def get_queryset(self):
+        queryset = super(RequireDBManager, self).get_queryset()
+        return RequireDBQuerySet(queryset)
+
+
+class PartitionedModel(models.Model):
```
gcapalbo: this purpose of this is to just make it so that we don't accidentally run queries on the wrong db - it essentially forces you to specify a database when calling save(), delete(), or any orm queries against the objects Manager (@snopoke  fyi)

corehq/sql_db/routers.py
```diff
@@ -35,7 +36,7 @@ def allow_migrate(db, app_label):
 
     if app_label == PROXY_APP:
         return db == partition_config.get_proxy_db()
-    elif app_label == FORM_PROCESSOR_APP:
+    elif app_label in (FORM_PROCESSOR_APP, SCHEDULING_PARTITIONED_APP):
```
gcapalbo: the routing here is only used by the scheduling models for migrations and not for running queries, since the sharding is done in python. i actually created a separate app, `scheduling_partitioned`, apart from the `scheduling` app, to hold all scheduling models that can be partitioned. it's going to create the partitioned models in the proxy db (in case we ever want to use pl/proxy even though it's not using it right now) as well as reuse all the form processing dbs for the actual partitioned models that hold data.

@snopoke fyi

corehq/sql_db/util.py
```diff
@@ -0,0 +1,102 @@
...
+    for db_name in db_names:
+        qs = model_class.objects.using(db_name).filter(q_expression)
+        if values:
+            if len(values) == 1:
+                qs = qs.values_list(*values, flat=True)
+            else:
+                qs = qs.values_list(*values)
+
+        for result in qs:
+            yield result
```
gcapalbo: this whole file has some generic utils for handling the sharding in python. it works on individual objects (get, save, delete) and also lets you run a query across all partitioned databases and consolidate the result

@snopoke fyi

gcapalbo:
ok @snopoke @emord this is good to go from my side. thanks for all the feedback - i've addressed everything but let me know if there's any other followup from my comments.

also @snopoke i marked some places of interest related to the partitioning that you may want to have a look at in case you haven't already.

what this releases so far is all the scheduling models and code for scheduling and handling the events for the different kinds of schedules.

corehq/util/mixin.py
```diff
@@ -37,3 +41,15 @@ class ValidateModelMixin(object):
     def save(self, *args, **kwargs):
         self.full_clean()
         super(ValidateModelMixin, self).save(*args, **kwargs)
+
+
+class DisabledDbMixin(object):
```
gcapalbo: i had moved this over to corehq.util.mixin from corehq.form_processor.models because i started off using it, but then i opted to instead take the approach used in the new `PartitionedModel`. i suppose i'll just leave this here though, since it's still a generic class

@snopoke fyi

corehq/messaging/scheduling/util.py
```diff
@@ -0,0 +1,8 @@
+from datetime import datetime
+
+
+def utcnow():
+    """
+    Defined here to do patching in tests.
```
emord: I think you could also use something like https://github.com/dimagi/commcare-hq/blob/master/corehq/apps/userreports/tests/test_pillow.py#L84

emord (approved):
Nothing blcoking from me

corehq/messaging/scheduling/models/timed_schedule.py
```diff
@@ -0,0 +1,201 @@
...
+            month_offset = (instance.schedule_iteration_num - 1) % 12
+
+            year = instance.start_date.year + year_offset
+            month = instance.start_date.month + month_offset
+
+            days_in_month = calendar.monthrange(year, month)[1]
+            if current_event.day > 0:
+                if current_event.day > days_in_month:
+                    # If the day refers to a date that is not possible to schedule
+                    # (for example, February 30th), just move to the next month.
```
gcapalbo: so actually the behavior here is that if someone starts a monthly schedule on the 30th, it would skip the event for February.

if someone wanted to schedule on the last day of the month, they could enter a -1 for the day. we don't have a way to do "last day of the month" in our ui spec since it's not a requirement yet, but i imagine some day that would be a nice thing to add and the backend is ready for it now.

corehq/messaging/scheduling/util.py
gcapalbo: yeah, datetime.utcnow() was used in a lot of different files though so it seemed easier for patching to centralize it this way

gcapalbo:
@snopoke let me know if there was anything else you wanted to review on this before we merge - if you need another day or so that's fine, i was just hoping to merge this some time this week

corehq/form_processor/tests/utils.py
```diff
@@ -172,8 +172,36 @@ def _delete_all(db, viewname, **view_kwargs):
...
+    """
+    Only runs the test with the non-partitioned database settings.
+    """
+    if settings.USE_PARTITIONED_DATABASE:
+        return nottest(cls)
+
+    return cls
+
+
+def only_run_with_partitioned_database(cls):
```
snopoke: What's the need for these? Is it basically the same as https://github.com/dimagi/commcare-hq/blob/d50fdf17133c68881b3600d025641a7402d524ae/corehq/form_processor/tests/test_sharding.py#L20-L20

corehq/messaging/scheduling/models/timed_schedule.py
snopoke: That cascade thing always confuses me.

corehq/messaging/scheduling/scheduling_partitioned/tests/test_dbaccessors_partitioned.py
```diff
@@ -0,0 +1,302 @@
...
+    def test_get_timed_schedule_instance(self):
+        self.test_save_timed_schedule_instance()
+        instance = get_timed_schedule_instance(self.p2_uuid)
+        self.assertTrue(isinstance(instance, TimedScheduleInstance))
+        self.assertEqual(instance.schedule_instance_id, self.p2_uuid)
+
+        with self.assertRaises(TimedScheduleInstance.DoesNotExist):
+            get_timed_schedule_instance(uuid.uuid4())
+
+    def test_access_restricted(self):
```
snopoke: Is it necessary to repeat this test? I didn't look closely but it looks very similar to the one in `TestSchedulingNonPartitionedDBAccessorsGetAndSave`

corehq/sql_db/util.py
```diff
@@ -0,0 +1,102 @@
...
+    """
+    if settings.USE_PARTITIONED_DATABASE:
+        db_name = ShardAccessor.get_database_for_doc(partition_value)
+    else:
+        db_name = 'default'
+
+    obj.delete(using=db_name)
+
+
+def run_query_across_partitioned_databases(model_class, q_expression, values=None):
```
snopoke: Will this ever need to support paging?

corehq/messaging/scheduling/scheduling_partitioned/models.py
```diff
@@ -0,0 +1,203 @@
...
+
+    @schedule.setter
+    def schedule(self, value):
+        if not isinstance(value, AlertSchedule):
+            raise ValueError("Expected an instance of AlertSchedule")
+
+        self.alert_schedule_id = value.schedule_id
+
+
+class TimedScheduleInstance(ScheduleInstance):
```
snopoke: Just to confirm, these are the only 2 sharded models (`TimedScheduleInstance` and `AlertScheduleInstance`) and they don't have any foreign keys to any other models so we don't have to worry about cross DB sharding issues.

snopoke (approved):

corehq/form_processor/tests/utils.py
gcapalbo: yeah so there are certain tests that i only want to run with a sharded db, and certain ones i only want to run on the monolith db. if i just mark a test with the sql_backend=True attribute, docker picks it up for testing in both kinds of dbs, so in order to only run it for only one type of db we need these extra decorators.

yeah i suppose i could have just used `@skipUnless(settings.USE_PARTITIONED_DATABASE)` as the tests you referenced do, but this moves that logic into a single decorator and makes it a little cleaner. i don't feel too strongly about it though, i can remove this if you prefer i just use `skipUnless`

corehq/messaging/scheduling/scheduling_partitioned/tests/test_dbaccessors_partitioned.py
gcapalbo: yeah so they're both tests which test that the routing for the migrations is working properly, but one tests the migrations in a monolith db and the other in a sharded db setup

gcapalbo: sorry i was referring to the wrong tests, what i said is true about the test_model_creation_partitioned and test_model_creation_non_partitioned tests.

for these dbaccessors tests, yes they're also a bit different. the ones that run in the sharded db setup explicitly make sure that each record gets to the right db and nowhere else.

corehq/sql_db/util.py
gcapalbo: good question, so far none of our uses of this need paging, and i think that might get a bit tricky since to do paging we need to also sort across dbs. i think it's doable but maybe should go in its own dbaccessor function

corehq/messaging/scheduling/scheduling_partitioned/models.py
gcapalbo: yeah these are the only two that we are sharding so far. they do have foreign keys to objects in the main db (each schedule instance model has a pointer to its schedule definition), but those are not handled with a django ForeignKey to avoid issues with querying / joining. those foreign keys are defined here and are only traversable through a python property:

https://github.com/dimagi/commcare-hq/blob/4997ea281f2b9c2f0b86c5e278b8a030a7e10a4b/corehq/messaging/scheduling/scheduling_partitioned/models.py#L160

https://github.com/dimagi/commcare-hq/blob/4997ea281f2b9c2f0b86c5e278b8a030a7e10a4b/corehq/messaging/scheduling/scheduling_partitioned/models.py#L178

corehq/form_processor/tests/utils.py
gcapalbo: (i'm going to merge this so it gets in with today's deploy everywhere - let me know if there's anything you'd prefer i do about these test decorators and i can handle in a separate pr)

corehq/messaging/scheduling/scheduling_partitioned/tests/test_dbaccessors_partitioned.py
snopoke: I'm referring specifically to the `test_access_restricted` rests which are identical between the two files and since they just access restriction I don't see the need for them to be repeated for sharded and non-sharded envs.

gcapalbo: i see, https://github.com/dimagi/commcare-hq/pull/15489

PR #14855, opened 2017-02-10
Externalize JavaScript: more low-hanging fruit

orangejenny:
@gcapalbo


czue:
looks like this is just html and js changes - can we kill python builds?

orangejenny:
Probably could have, yeah. I think app manager is the only place where we have python testing html/js changes.

corehq/apps/domain/static/domain/js/internal_settings.js
```diff
@@ -0,0 +1,59 @@
...
+        update_subareas();
+    });
+    $('[name="initiative"]').change(function() {
+        update_workshop_region();
+    });
+
+});
+
+$(function() {
+    $('#id_deployment_date').datepicker({
```
gcapalbo: it looks like a lot of the external js files in this pr now reference element ids from the page, or other details of elements on the specific page. just curious why we're moving some of the page-specific information out of the template? it seems like it would be hard for another template to reuse this external file unless it was structured in the exact same way

orangejenny: Yeah, the js files added in this PR aren't meant to be reusable. I'm splitting them out for the sake of organization, ability to compress & cache, and eventually to work better with a module system.

orangejenny: FWIW, I think eventually this specific code should/might be replaced by code that looks for a class instead of an  id and has the options encoded as data attributes, so it could be used more generically.

gcapalbo: ok, thanks for the explanation

PR #14856, opened 2017-02-10
default to True

sravfeyn:
http://manage.dimagi.com/default.asp?247071#edit_0_247071 @czue


sravfeyn:
It should be okay to not migrate roles that have the value to `False` to `True`, right?

czue (changes requested):
Not sure about the migration - when do those objects actually persist? Either way I think this is a good change.

However, I think we need to make sure that `AdminRole` doesn't have this property set. Not sure if that's as easy as updating the `__init__` function in the subclass or you need to do something fancier.

PR #14857, opened 2017-02-10
Tweak combobox

benrudolph:
Default to showing more items http://manage.dimagi.com/default.asp?246510. Think i got it right this time

cc: @kaapstorm

### DIFF
--- a/corehq/apps/cloudcare/static/cloudcare/js/form_entry/entrycontrols_full.js
+++ b/corehq/apps/cloudcare/static/cloudcare/js/form_entry/entrycontrols_full.js
@@ -373,6 +373,9 @@ function ComboboxEntry(question, options) {
                 matcher: function() {
                     return $input.val();
                 },
+                sorter: function(query, data) {
+                    return data;
+                },
             },
         });
     };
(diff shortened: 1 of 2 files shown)


corehq/apps/cloudcare/static/cloudcare/js/form_entry/entrycontrols_full.js
```diff
@@ -373,6 +373,9 @@ function ComboboxEntry(question, options) {
                 matcher: function() {
                     return $input.val();
                 },
+                sorter: function(query, data) {
+                    return data;
+                },
```
benrudolph: this needs to be changed. the default sorter would eliminate items that didn't directly match the query... which is dumb considering there's a function to filter the items already.

kaapstorm: Yeah, surely a sorter isn't supposed eliminate anything; it just moves the items the user is most likely to want up to the top of the already-filtered items, right?

kaapstorm (approved):

corehq/apps/cloudcare/static/cloudcare/js/form_entry/entrycontrols_full.js
benrudolph: yeah that's what i had thought, but that wasn't the behavior i was seeing unfortunately

PR #14858, opened 2017-02-10
Revert "remove erroring tests"

czue:
Reverts dimagi/commcare-hq#14804

@nickpell I don't like the idea of these being permanently not run on travis. were you planning to look into why they broke and get them running again?


czue:
actually these were failing locally for me so i just fixed them.

custom/enikshay/nikshay_datamigration/factory.py
```diff
@@ -306,7 +321,11 @@ def get_drtb_hiv_referral_case_structure(self, episode_structure):
                 related_type=EPISODE_CASE_TYPE,
             )],
         }
-
+        if self.existing_drtb_hiv_case:
```
czue: @nickpell we need to do this for any case being created by migration. was a bit surprised you hadn't already accounted for this given the similarities to other case types

czue: would be good to add a test case for this too if not too hard

nickpell: https://github.com/dimagi/commcare-hq/pull/14979

PR #14860, opened 2017-02-10
Test refactoring sync_mode

benrudolph:
Taking a stab at getting casexml tests down. This one refactors sync_mode just a bit by moving things that can be done in SetUpClass to there instead of setUp.

Before:
```

Ran 68 tests in 108.208s
```
After:
```

Ran 68 tests in 54.738s
```
@dimagi/test-team

### DIFF
--- a/corehq/ex-submodules/casexml/apps/phone/tests/test_sync_mode.py
+++ b/corehq/ex-submodules/casexml/apps/phone/tests/test_sync_mode.py
@@ -1882,20 +1882,22 @@ class SteadyStateExtensionSyncTest(SyncBaseTest):
     Test that doing multiple clean syncs with extensions does what we think it will
     """
 
-    def setUp(self):
-        super(SteadyStateExtensionSyncTest, self).setUp()
-        self.other_user = create_restore_user(
-            self.project.name,
+    @classmethod
+    def setUpClass(cls):
+        super(SteadyStateExtensionSyncTest, cls).setUpClass()
+        cls.other_user = create_restore_user(
+            cls.project.name,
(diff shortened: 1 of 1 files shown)


snopoke (approved):
:1st_place_medal:

benrudolph:
Realized it's especially brutal to do extra work in the `setUp` and `tearDown` call because not only does this get called once per test, but for those decorated with `run_with_all_backends` it gets called twice (i believe)

benrudolph:
For all of casexml:

Before:
```
Ran 390 tests in 353.083s
```
After:
```
Ran 390 tests in 147.810s
```

snopoke:
Yea, that's the same as I found with this one https://github.com/dimagi/commcare-hq/pull/14838

kaapstorm (approved):

emord:
Would it be possible to have `run_with_all_backends` decorate the class instead of individual tests? Or does that not work becasue run_with_all_backends logs two setup calls per test, instead of splitting them into two tests?

corehq/ex-submodules/casexml/apps/phone/tests/test_sync_mode.py
```diff
@@ -42,21 +42,26 @@ class SyncBaseTest(TestCase):
     """
     Shared functionality among tests
     """
+    @classmethod
+    def setUpClass(cls):
+        super(SyncBaseTest, cls).setUpClass()
+        delete_all_users()
```
emord: Does it still work if you remove this line?

emord (approved):

PR #14861, opened 2017-02-10
update logging to prevent duplicate error reports

snopoke:
anyone

Tested this locally and prior to this change 2 emails get sent for every 500. After the change only one (the correct one).

See https://docs.djangoproject.com/en/1.10/releases/1.9/#changes-to-the-default-logging-configuration


snopoke:
killed the build

PR #14863, opened 2017-02-10
Use format_html so Django 1.9 doesn't escape XML

kaapstorm:
Context: [FB 247336](http://manage.dimagi.com/default.asp?247336)

Django 1.9 escaping strings returned by template tags again. 

@sravfeyn cc @czue @benrudolph

### DIFF
--- a/corehq/apps/reports/templatetags/xform_tags.py
+++ b/corehq/apps/reports/templatetags/xform_tags.py
@@ -5,7 +5,7 @@ from django.template import RequestContext
 from django.template.loader import render_to_string
 from django.core.urlresolvers import reverse
 from django import template
-from django.utils.html import escape
+from django.utils.html import format_html
 from couchdbkit.exceptions import ResourceNotFound
 from corehq import privileges
 from corehq.apps.cloudcare import CLOUDCARE_DEVICE_ID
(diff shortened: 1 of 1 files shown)


kaapstorm:
raw_xml

corehq/apps/reports/templatetags/xform_tags.py
```diff
@@ -37,8 +37,8 @@ def render_form_xml(form):
     if isinstance(xml, unicode):
         xml.encode('utf-8', errors='replace')
     formatted_xml = indent_xml(xml) if xml else ''
-    return '<pre class="prettyprint linenums"><code class="no-border language-xml">%s</code></pre>' \
-           % escape(formatted_xml)
+    return format_html('<pre class="prettyprint linenums"><code class="no-border language-xml">{}</code></pre>',
+                       formatted_xml)
```
kaapstorm: I'm not seeing line numbers. Should I be? 

And to render the XML I dropped the `escape()` call. I didn't expect that either.

czue (approved):
nice thanks. this is definitely an improvement. I just checked an older version and it looks like it's been this way for a while so seems fine. my guess is we had something like codemirror setup but it broke and no one noticed? could be nice to get working again, but doesn't feel in scope for this ticket.

benrudolph:
i actually unintentionally removed it when i removed the double click thing that it used to do. it use the google pretty code thing

PR #14864, opened 2017-02-10
Remove do not process old builds toggle

benrudolph:
@czue @NoahCarnahan full throttle. haven't had any issues thus far

cc: @kaapstorm


czue (approved):
nice

PR #14866, opened 2017-02-10
run both sharded and js tests on same runner

benrudolph:
@dimagi/test-team maybe a bad idea, but we can certainly fit both these test suites on one runner. it's not as clean though

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,12 +5,11 @@ branches:
     - master
 env:
   matrix:
-    - TEST=javascript
     # need node installs for app_manager.XpathValidatorTest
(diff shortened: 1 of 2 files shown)


snopoke:
Makes sense to me

emord (approved):

millerdev (approved):

benrudolph:
TEST=python-sharded-and-javascript JS_SETUP=yes
 23 min 36 sec

👷 🔫

snopoke:
Nice, although this doesn't reduce the total build time it does mean less concurrent VM's required per PR which should improve our throughput.

benrudolph:
1 less VM and 1 less pull down of docker, so i think it shaves 5 or so minutes

PR #14869, opened 2017-02-10
case block ordering

snopoke:
The processing order for case blocks for the same case in a single form is currently the order they appear in the form. This means that an update block could get processed before a create block.

First commit is a test prefactor


czue:
build fail looks real
```
FAIL: test_default_to_phone_date_on_same_day (casexml.apps.case.tests.test_rebuild.TestActionSortKey)
  File "corehq/ex-submodules/casexml/apps/case/tests/test_rebuild.py", line 536, in test_default_to_phone_date_on_same_day
    self._test(expected_actions)
  File "corehq/ex-submodules/casexml/apps/case/tests/test_rebuild.py", line 574, in _test
    self.assertEqual(action, sorted_actions[index])
AssertionError: None: update - 2001-01-01 00:00:01 (2001-01-01 00:00:00) != None: update - 2001-01-01 00:00:00 (2001-01-01 00:00:01)
```

snopoke:
:white_check_mark:

czue (approved):
:+1:

PR #14873, opened 2017-02-10
99DOTS Repeater Changes

proteusvacuum:
I spoke to Nakull this morning and we came up with some changes that were needed for the 99DOTS repeaters. 
Biggest change is that we weren't triggering a 99DOTS update whenever the episode or person case's properties that are tracked on 99D were changing.
🐡 
@NoahCarnahan @mkangia 
cc: @emord


NoahCarnahan (approved):
Looks good 👍

PR #14874, opened 2017-02-10
Externalize JavaScript some more

orangejenny:
Was originally going to convert all of locations, got sidetracked and ended up just doing a couple more random easy pages.

@gcapalbo / @millerdev

### DIFF
--- a/corehq/apps/locations/static/locations/js/location_types.js
+++ b/corehq/apps/locations/static/locations/js/location_types.js
@@ -321,6 +321,41 @@ hqDefine('locations/js/location_types.js', function(){
             };
         };
     }
+
+    $(function() {
+        var initial_page_data = hqImport('hqwebapp/js/initial_page_data.js').get,
+            loc_types = initial_page_data('location_types'),
+            commtrack_enabled = initial_page_data('commtrack_enabled'),
+            model = new LocationSettingsViewModel(loc_types, commtrack_enabled);
+
+        var $settings = $('#settings');
+
+        var warnBeforeUnload = function() {
+            return gettext("You have unsaved changes.");
+        };
+
+        $settings.submit(function() {
+            var valid = model.presubmit();
+            if (valid) {
+                // Don't warn if they're leaving the page due to form submission
+                window.onbeforeunload = undefined;
+            }
+            return valid;
+        });
+
+        $settings.koApplyBindings(model);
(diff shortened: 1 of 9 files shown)


gcapalbo (approved):

corehq/apps/locations/static/locations/js/location_types.js
```diff
@@ -321,6 +321,41 @@ hqDefine('locations/js/location_types.js', function(){
             };
         };
     }
+
+    $(function() {
```
millerdev: Something to think about, but probably not immediately actionable. Importing this module now has a side effect (setup on document ready handler). It looks like it didn't have any side effects before, it just exposed an object. Is this additional import/runtime complexity something we should be concerned about? Should try to define some type of distinction between modules that have import side effects and modules that just expose data/functions?

millerdev (commented):
> Was originally going to convert all of locations, got sidetracked and ended up just doing a couple more random easy pages.

👍 for release early/small

corehq/apps/locations/static/locations/js/location_types.js
orangejenny: Definitely think we should be concerned about it.

In my head, I've been bucketing js files into "libraries" and "entry points" where the former don't have side effects and might be used by several pages, while the latter consist mostly of side effects (applying knockout bindings or setting up event handlers) and are typically used by only one page, or by a few closely-related pages. I'm not certain at this point that's the correct or only distinction to be making, might make sense to have a third "widget" bucket for files that define a widget or knockout model and then also initialize any instances of it.

When a js file is included on only one html template, I've been giving it the same name as that template, but otherwise I haven't been using any particular conventions. I think having  definitive conventions - and doing the necessary file renaming and code re-organization - will be part of the move to a module system, and will fit in with figuring out each page's entry point (I've been thinking we'll do a single entry point per page using something like [RequireJS's data-main attribute](http://requirejs.org/docs/api.html#data-main)) and also which scripts get included on which pages.

Happy to talk more, now or later, especially if you're thinking we should be making these decisions sooner rather than later.

millerdev: OK, great! It sounds like you're keeping the right stuff in mind as you're doing this. No need (for me) to discuss more now.

PR #14877, opened 2017-02-12
Conditionally run both backends

benrudolph:
@dimagi/test-team This makes it possible to _not_ run both backends for tests marked as `run_with_all_backends`. We run our `casexml` tests 4 times, twice during the normal run when tests are marked as `run_with_all_backends` and then again for the python sharded tests. It's redundant to run the Couch backend for testing the python sharding, and it's particularly painful because these are some of our slowest tests. I also think this'll just be generally useful, since there have been many times where i've been debugging something in a test not related to the backend and having the test run twice in a row is annoying. Big PR, but mainly it's find and replace. Best read 🐟 

cc: @kaapstorm


benrudolph:
(this isn't as easy as i'd hoped, we have a lot of dependencies on the fact that couch tests are run first)

PR #14878, opened 2017-02-12
use setupclass in accounting

emord:
@dimagi/test-team 

There were a lot of tests creating a new domain for every test and x users as well.

cc @proteusvacuum


czue (approved):
nice

PR #14879, opened 2017-02-13
squash accounting migrations

nickpell:


benrudolph:
nice!

NoahCarnahan:
woah. This squashes every single one?

PR #14881, opened 2017-02-13
ICDS Branding

sravfeyn:
https://docs.google.com/document/d/1QCUHA7PLvZPspuY23Pn98HIrD4zv_KuakFoJwi1dGiE/edit#

@nickpell @snopoke


corehq/apps/export/templates/export/partial/export_list_controller.html
```diff
@@ -16,6 +16,8 @@
     {% endif %}
     <p>
         {% block lead_text %}
+        {% blocktrans with hq_name=django_settings.COMMCARE_HQ_NAME %}
+            Exports are a way to download data from {{ hq_name }}
```
snopoke: missing `endblocktrans`

sravfeyn:
Corresponding ansible change https://github.com/dimagi/commcarehq-ansible/commit/96cd91e1498b06dafcb581c7a4fae9ebe3459b00

corehq/util/context_processors.py
```diff
@@ -87,7 +87,7 @@ def websockets_override(request):
         return {}
 
 
-def enterprise_mode(request):
+def django_settings(request):
     return {
-        'enterprise_mode': settings.ENTERPRISE_MODE
+        'django_settings': settings
```
snopoke: I don't think it's a good idea to expose ALL settings here.

snopoke (changes requested):
Looking good so far

corehq/util/context_processors.py
sravfeyn: Why is this bad?

snopoke: It just makes it easier to expose secrets to the world. A typo that got missed in code review could put our entire settings dict into a template. Best to be explicit about what templates should have access to.

sravfeyn:
@snopoke Updated. Except for 1,2,3 and 5 in the spec, everything else is implemented.

PR #14883, opened 2017-02-13
Remove blue box for group in form entry

benrudolph:
http://manage.dimagi.com/default.asp?247456 @czue @biyeun removes blue box when group has no children
screen shot 2017-02-13 at 12 03 10 pm
screen shot 2017-02-13 at 12 02 57 pm

cc: @kaapstorm


czue:
sweet, thanks!

PR #14888, opened 2017-02-13
add failing test for instance removal bug

czue:


corehq/apps/app_manager/suite_xml/post_process/instances.py
```diff
@@ -110,18 +110,20 @@ def __call__(self, fn):
...
 }
 
 
-@register_factory(*INSTANCE_BY_ID.keys())
+@register_factory(*INSTANCE_KWARGS_BY_ID.keys())
 def preset_instances(domain, instance_name):
-    return INSTANCE_BY_ID.get(instance_name, None)
+    kwargs = INSTANCE_KWARGS_BY_ID.get(instance_name, None)
+    if kwargs:
+        return Instance(**kwargs)
```
czue: whoa this is crazy. sorry, probably just being slow, but where was the instance getting mutated / how does this fix it?

corehq/apps/app_manager/tests/data/suite/search_command_detail.xml
```diff
@@ -5,7 +5,7 @@
         <locale id="m0.case_short.title"/>
       </text>
     </title>
-    <field sort="default">
+    <field>
```
NoahCarnahan: @czue So this is a little odd...
Your addition of `self.app.build_spec = BuildSpec(version='tests', build_number=1)` causes a different test, `test_case_search_action` to fail, unless this change is also made. I haven't looked into why the existence of the `BuildSpec` would make a difference though. Does this change look acceptable? My reading of the spec is that the `sort` attribute of `field`s is obsolete anyways: https://github.com/dimagi/commcare-core/wiki/Suite20

czue: huh, that's strange. I think that's fine thanks. (i added that because otherwise `wrap` [tries to hit couch](https://github.com/dimagi/commcare-hq/blob/master/corehq/apps/app_manager/models.py#L4669) which makes it fail as a `SimpleTestCase`... not sure if there was an easier way to address that)

czue:
I can't approve this because I opened it, but I approve. Still curious how this fixes it though.

corehq/apps/app_manager/suite_xml/post_process/instances.py
NoahCarnahan: Yeah so the instance wasn't getting mutated, but it was being passed to two different `RemoteRequest` object constructors, which store `Instance`s in `NodeListFields`s: https://github.com/dimagi/commcare-hq/blob/b26f60148b8fc2c535f53dd1ccc0929e01a96932/corehq/apps/app_manager/suite_xml/xml_models.py#L478

I haven't dug too deeply into the eulxml.xmlmap source, but I'm assuming that the `NodeListField` expects the nodes you pass it to not exist elsewhere in the tree, but I don't know for sure. I can try to confirm this explanation if you think that's worth spending time on.

czue: > I haven't dug too deeply into the eulxml.xmlmap source, but I'm assuming that the NodeListField expects the nodes you pass it to not exist elsewhere in the tree

Ah, this makes sense (and also makes sense why you were worried about the uniqueness).

But I still don't quite understand why this wouldn't come up before? I see 37 occurrences of `<instance id="casedb"` in the enikshay suite.

>  I can try to confirm this explanation if you think that's worth spending time on

no that seems fine, though might try and dig in and satisfy my own intellectual curiosity at some point

PR #14960, opened 2017-02-13
Abt custom report: Adjust kenya targets configuration

NoahCarnahan:
@esoergel 
Killing build because no test coverage


NoahCarnahan:
self merging because this needs to get in before deploy

PR #14963, opened 2017-02-13
Allow custom ordering of reports in sidebar

NoahCarnahan:
@czue I was going to write a spec for Kamil to implement this, but I had to do a bit of digging to figure out how best to approach it, and at that point it seemed easier to just write the code than to write a spec for it.

This PR adds a new model, `ReportsSidebarOrdering`, which can be created in the django admin interface. Here is an example config:
```json
[
   ["Adherence", [
      "DynamicReport7613ac1402e2c41db782526e9c43e040",
      "DynamicReport1233ac1402e2c41db782526e9c43e040"
   ]],
   ["Test Results", [
      "DynamicReport4563ac1402e2c41db782526e9c43e040",
      "DynamicReportmy-static-ucr-id"
   ]]
]
```

This will create two new sections about the "Reports" section of the sidebar containing the specified reports. The specified reports will not appear in "Reports"

cc @esoergel


corehq/apps/reports/tests/test_sidebar.py
```diff
@@ -0,0 +1,46 @@
...
+            ]),
+        ]
+
+        mock_ordering = ReportsSidebarOrdering(
+            domain="wut",
+            config=[
+                ["new section", ["qux", "foo", "dne"]]
+            ],
+        )
+        def mock_get(*args, **kwargs):
```
dimagimon: E301 expected 1 blank line, found 0

corehq/tabs/tabclasses.py
```diff
@@ -59,9 +64,42 @@ def sidebar_items(self):
...
         custom_reports = CustomProjectReportDispatcher.navigation_sections(
             request=self._request, domain=self.domain)
-        sidebar_items = tools + report_builder_nav + custom_reports + project_reports
+        sidebar_items = tools + report_builder_nav + self._regroup_sidebar_items(custom_reports + project_reports)
         return self._filter_sidebar_items(sidebar_items)
 
+    def _regroup_sidebar_items(self, sidebar_items):
+        try:
+            ordering = ReportsSidebarOrdering.objects.get(domain=self.domain)
+        except ObjectDoesNotExist:
```
esoergel: Oh weird, I didn't know this even existed, I've always used `ReportsSidebarOrdering.DoesNotExist`

corehq/tabs/tabclasses.py
```diff
@@ -59,9 +64,42 @@ def sidebar_items(self):
             request=self._request, domain=self.domain)
         custom_reports = CustomProjectReportDispatcher.navigation_sections(
             request=self._request, domain=self.domain)
-        sidebar_items = tools + report_builder_nav + custom_reports + project_reports
+        sidebar_items = tools + report_builder_nav + self._regroup_sidebar_items(custom_reports + project_reports)
         return self._filter_sidebar_items(sidebar_items)
 
+    def _regroup_sidebar_items(self, sidebar_items):
```
esoergel: This'd also work as a utility function outside of the class.

corehq/apps/reports/models.py
```diff
@@ -1097,3 +1099,21 @@ def remove_custom_export(cls, domain, export_id):
...
+    )
+    config = JSONField(
+        default=list,
+        help_text=(
+            "An array of arrays. Each array represents a heading in the sidebar navigation. "
+            "The first item in each array is a string, which will be the title of the heading. The second item in "
+            "the array is another array, each item of which is the name of a report class. Each of these reports "
+            "will be listed under the given heading in the sidebar nav."
+        )
+    )
```
esoergel: Is there an easy way to add any basic validation to this?  (It's obviously a super internal thing, so maybe that's not necessary, but it'd be nice)

corehq/tabs/tabclasses.py
```diff
@@ -59,9 +64,42 @@ def sidebar_items(self):
...
+
+        reports_to_move = {}  # report class name to SidebarPosition mapping
+        for heading, reports in ordering.config:
+            for i, report_class_name in enumerate(reports):
+                reports_to_move[report_class_name] = SidebarPosition(heading, i)
+
+        # A mapping of headings to lists of 2-tuples. The tuples are index (under the heading) and the item itself.
+        new_sections = {x[0]: [] for x in ordering.config}
+        # A mapping of headings to lists of sidebar items.
+        old_sections = {x[0]: [] for x in sidebar_items}
```
esoergel: you could unpack these tuples into something like this
```python
old_sections = {heading: [] for heading, _ in sidebar_items} 
```

corehq/tabs/tabclasses.py
```diff
@@ -59,9 +64,42 @@ def sidebar_items(self):
...
+                    new_sections[new_position.heading].append((new_position.index, item))
+                else:
+                    old_sections[heading].append(item)
+
+        # Flatten the itermediary data structures
+        flat_new_sections = [
+            (heading, [item for index, item in sorted(new_sections[heading])])
+            for heading, _ in ordering.config
+        ]
+        flat_old_sections = [(heading, old_sections[heading]) for heading, _ in sidebar_items]
```
esoergel: Shouldn't this use `old_sections` instead of `sidebar_items`?  I think this will make reports in `new_sections` show up twice.

esoergel: From the test above it looks like that doesn't happen, but I don't understand how that could be.

esoergel (commented):
I had a few questions, but nothing major.  Let me know if you wanna discuss in person.

esoergel:
(That's not a real test failure)

corehq/apps/reports/models.py
```diff
@@ -1097,3 +1099,21 @@ def remove_custom_export(cls, domain, export_id):
...
+        blank=False,
+        unique=True
+    )
+    config = JSONField(
+        default=list,
+        help_text=(
+            "An array of arrays. Each array represents a heading in the sidebar navigation. "
+            "The first item in each array is a string, which will be the title of the heading. The second item in "
+            "the array is another array, each item of which is the name of a report class. Each of these reports "
+            "will be listed under the given heading in the sidebar nav."
```
czue: this might be a case where providing an example would be clearer than describing it in words.

corehq/apps/reports/models.py
czue: agreed this would be good if easy. could potentially do it with a jsonobject schema?

corehq/tabs/tabclasses.py
```diff
@@ -59,9 +64,42 @@ def sidebar_items(self):
...
 
+    def _regroup_sidebar_items(self, sidebar_items):
+        try:
+            ordering = ReportsSidebarOrdering.objects.get(domain=self.domain)
+        except ObjectDoesNotExist:
+            return sidebar_items
+
+        reports_to_move = {}  # report class name to SidebarPosition mapping
+        for heading, reports in ordering.config:
+            for i, report_class_name in enumerate(reports):
```
czue: I suspect this would just cause every report page in the domain to 500 if someone configures the ordering wrong right? seems like a bad failure mode. so i think should either add error handling here or add the validation going into the DB above

corehq/tabs/tabclasses.py
czue: I think `ReportsSidebarOrdering.DoesNotExist` is more appropriate here and `ObjectDoesNotExist` is meant to be used if there might be more than one model type that could raise it. though don't think the distinction is too important here since it's so clear.

corehq/tabs/tabclasses.py
```diff
@@ -59,9 +64,42 @@ def sidebar_items(self):
...
+        sidebar_items = tools + report_builder_nav + self._regroup_sidebar_items(custom_reports + project_reports)
         return self._filter_sidebar_items(sidebar_items)
 
+    def _regroup_sidebar_items(self, sidebar_items):
+        try:
+            ordering = ReportsSidebarOrdering.objects.get(domain=self.domain)
+        except ObjectDoesNotExist:
+            return sidebar_items
+
+        reports_to_move = {}  # report class name to SidebarPosition mapping
```
czue: it might be a little clearer to extract everything here and below to a utility function and replace this line with:
```python
return _reorder_sidebar_items(reports, ordering)
```
not sure if that would also simplify the testing interface at all, but seems like it might.

czue (changes requested):
Looks good! from my perspective just needs some validation/error handling somewhere and then should be g2g. had a few other minor things you can respond to or ignore.

corehq/apps/reports/models.py
NoahCarnahan: I can define a jsonobject subclass then try to wrap the field's content during the model's clean method. Is that what you had in mind? Or do we have some sort of json field class that allows specifying a jsonobject class as a schema?

corehq/apps/reports/models.py
NoahCarnahan: I had an example here to begin with, but the admin app renders the help text in `<p>` tags so I can't nice formatting for the json without mucking with the admin app a little, which didn't seem worth it. What do you think of putting an example in a comment here instead of adding it to the help_text?

corehq/tabs/tabclasses.py
NoahCarnahan: Since its just getting the headings  from `sidebar_items`, `old_sections.keys()` and `x[0] for x in sidebar_items` are equivalent. Although probably more readable to just use `old_sections` throughout. I'll update

NoahCarnahan: Ah actually I take that back. Need to iterate through `sidebar_items` to get the order of the keys correctly.

corehq/apps/reports/models.py
czue: ah, that's annoying. makes sense and sounds good.

corehq/apps/reports/models.py
czue: yeah was thinking something exactly like that. You could also add a property called `wrapped_config` which would return the jsonobject version of the config which might be nicer to work with in the calling code as well.

unfortunately, we don't have the equivalent of `JsonField(schema=MyDocument)` though that would be pretty nice for a future PID...

corehq/apps/reports/models.py
esoergel: Will it render html?  I wonder if you could just throw a `<pre/>` tag around the example.  You might need to use `mark_safe` somewhere, though.

corehq/apps/reports/models.py
NoahCarnahan: Decided to just go with a "manual" check of the structure here: https://github.com/dimagi/commcare-hq/pull/14963/commits/9c1e981cc82e5432f5266831cadb2f180abd8a4e.
Since my config structure is nested lists instead of nested objects, I think jsonobject won't actually be that useful for validating. For example I don't think there is a way in jsonobject to specify that a top level JsonObject should be a list, or that the first item of a list should have certain properties while the second should some other properties.

I suppose I could change the config structure to be object based instead of list based, but given that this isn't going to be a widely used feature, I'm inclined to think it's not worth restructuring. (Plus it would be more tedious to write the config).

czue: that makes sense and sounds good. yeah to make it play nice with JsonObject I think you would want to rewrite the structure as:
```json
[
  {
    "header": "Adherence"
   "reports": [
       "r1", "r2" 
    ]
}
]
```
and then validate the items, but i can see that being a little more unwieldy. though has the benefit of being easier to extend. but yeah, seems fine for now to do it this way.

czue (approved):

corehq/apps/reports/tests/test_sidebar.py
```diff
@@ -0,0 +1,37 @@
+from django.test import SimpleTestCase
+from mock import patch, MagicMock
```
dimagimon: F401 'MagicMock' imported but unused
F401 'patch' imported but unused

corehq/apps/reports/tests/test_sidebar.py
```diff
@@ -0,0 +1,37 @@
...
+
+from corehq.apps.reports.models import ReportsSidebarOrdering
```
dimagimon: F401 'ReportsSidebarOrdering' imported but unused

corehq/apps/reports/tests/test_sidebar.py
```diff
@@ -0,0 +1,37 @@
...
+from corehq.tabs.tabclasses import ProjectReportsTab
```
dimagimon: F401 'ProjectReportsTab' imported but unused

corehq/tabs/tabclasses.py
```diff
@@ -71,33 +68,7 @@ def _regroup_sidebar_items(self, sidebar_items):
...
-                    old_sections[heading].append(item)
-
-        # Flatten the intermediary data structures
-        flat_new_sections = [
-            (heading, [item for index, item in sorted(new_sections[heading])])
-            for heading, _ in ordering.config
-        ]
-        flat_old_sections = [(heading, old_sections[heading]) for heading, _ in sidebar_items]
-        return flat_new_sections + flat_old_sections
+        return regroup_sidebar_items(ordering.config, sidebar_items)
```
esoergel: :+1:

esoergel (approved):

PR #14964, opened 2017-02-13
Dynamically generate sql backend test cases

benrudolph:
@dimagi/test-team in light of my failure to not run couch tests on travis by modifying the `run_with_all_backends`, I've developed a new strategy. It might be too sneaky, but curious to hear your thoughts. Essentially this decorator will create a new class by subclassing the decorated class and then override its settings. Then we can tag that new class as `sql_backend`. Once we've tagged all tests that are sql_backend, we can change the travis config to just run `--attr=sql_backend`

cc: @kaapstorm

### DIFF
--- a/corehq/ex-submodules/casexml/apps/phone/tests/test_sync_mode.py
+++ b/corehq/ex-submodules/casexml/apps/phone/tests/test_sync_mode.py
@@ -16,8 +16,7 @@ from corehq.apps.groups.models import Group
 from corehq.apps.users.dbaccessors.all_commcare_users import delete_all_users
 from corehq.apps.receiverwrapper.util import submit_form_locally
 from corehq.form_processor.interfaces.dbaccessors import CaseAccessors
-from corehq.form_processor.tests.utils import FormProcessorTestUtils, \
-    run_with_all_backends
+from corehq.form_processor.tests.utils import FormProcessorTestUtils, generate_sql_backend_test_case
 from corehq.toggles import LOOSE_SYNC_TOKEN_VALIDATION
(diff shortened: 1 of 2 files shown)


kaapstorm:
My 5c: Creating classes can be a little magical, but I like this use case. I don't think it's too obscure.

snopoke:
This is quite sneaky. If I wanted to run just a single SQL test locally I'd have to know that the test class is called `CouchTest_SQL` or else use some other way of singling it out using the attribute and a pattern matcher:
```
--attr=sql_backed -m=*the_test_i_want_to_run
```

I think I might be in favor of a more manual approach which makes it clear what's happening:
```python
class TestCouch(TestCase):
    pass

@attr(sql_backend=True)
@override_settings(TESTS_SHOULD_USE_SQL_BACKEND=True)
class TestSQL(TestCouch):
    pass
```

benrudolph:
That's true, though I don't think it's a loss in functionality of what we currently have. Currently you can't run just the sql test, you have to run both the couch and sql and step through both. This decorator creates the same behavior. The reason i went this route was because it was slightly easier to do the change. I'm not too tied to it though, was a fun experiment

snopoke:
That's fair but I think you do lose the ability to run a single SQL test at
all (even with the couch equivalent). If you ran `manage.py
path.to.test.py:TestClass.test_method`
that would only run the Couch test (previously it would have run both). In
order to run the SQL test at all you either have to run the entire file or
you have to know that you must reference the dynamic class `TestClass_SQL`.

benrudolph:
ah good point

benrudolph:
(closing for now)

PR #14965, opened 2017-02-13
BubbleGraphSeries should inherit from XYGraphSeries

orangejenny:
Bubble graphs weren't displaying the autocomplete option and hints for `is-data`, `point-style`, `secondary-y`.

@gcapalbo / @NoahCarnahan


gcapalbo (approved):

orangejenny:
js-only change, killing build and merging

PR #14972, opened 2017-02-14
cache hq multimedia

benrudolph:
@czue this should fix the caching annoyances for app preview

screen shot 2017-02-14 at 2 33 31 pm

cc: @kaapstorm

### DIFF
--- a/corehq/apps/hqmedia/views.py
+++ b/corehq/apps/hqmedia/views.py
@@ -24,6 +24,7 @@ from corehq.util.files import file_extention_from_filename
 
 from soil import DownloadBase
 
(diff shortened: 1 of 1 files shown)


snopoke (approved):

czue (approved):
nice!

millerdev:
How will this affect the app if the multimedia in the app is changed after a user has viewed (the old version of) the multimedia in the app? Will it load the cached version forever? Is there a way to invalidate the cache?

benrudolph:
Good question, all images uploaded have their own id that the application references. So as soon as you upload a new image it points to a new reference (thus no need to cache bust)

PR #14973, opened 2017-02-14
ICDS SMS Indicators - First Pass

gcapalbo:
Creates a small framework for evaluating and running indicators and implements the first one.

@emord


custom/icds/tasks.py
```diff
@@ -0,0 +1,33 @@
...
+
+
+@task(queue=CELERY_REMINDERS_QUEUE, ignore_result=True)
+def run_indicator(domain, user_id, indicator_class):
+    """
+    Runs the given indicator for the given user and sends the SMS if needed.
+
+    :param domain: The domain the indicator is being run for
+    :param user_id: The id of either an AWW or LS CommCareUser
+    :param indicator_class: a subclass of AWWIndicator or LSIndicator
```
dimagimon: W291 trailing whitespace

corehq/apps/users/models.py
```diff
@@ -1992,9 +1992,12 @@ def __repr__(self):
             self=self
         ))
 
+    def get_usercase(self):
+        return CaseAccessors(self.domain).get_case_by_domain_hq_user_id(self._id, USERCASE_TYPE)
+
     @skippable_quickcache(['self._id'], lambda _: settings.UNIT_TESTING)
     def get_usercase_id(self):
-        case = CaseAccessors(self.domain).get_case_by_domain_hq_user_id(self._id, USERCASE_TYPE)
+        case = self.get_usercase()
```
snopoke: You could update the quickcache to allow skipping with a kwarg:
```python
@skippable_quickcache(['self._id'], lambda user, skip_cache: skip_cache or settings.UNIT_TESTING)
def get_usercase_id(self, skip_cache=False):
    ...
```

gcapalbo: @snopoke not sure i understand - i didn't want to bypass the cache, i just wanted to get the actual user case and not just the user case id

custom/icds/messaging/indicators.py
```diff
@@ -0,0 +1,117 @@
...
+
+        context = {
+            'more_than_one_week': more_than_one_week,
+            'more_than_one_month': more_than_one_month,
+            'awc': self.user.sql_location.name,
+        }
+
+        message = self.render_template(context, language_code=language_code)
+
+        return [message]
```
emord: If you call this on an indicator that should not be sent it would return `[u'']` which I think would send an empty message unless there's something deeper in SMS that prevents that. None of should happen, but would it be better to use `filter(None, ...)` here or in the task? (maybe a soft assert too to be extra paranoid)

gcapalbo: yeah that shouldn't happen as long as `should_send` is called first, but agree it seems confusing. maybe to make this cleaner we should just get rid of `should_send`, and have `get_messages` return empty list when no messages need to be sent

corehq/apps/users/models.py
snopoke: gotcha

custom/icds/messaging/indicators.py
emord: offlined and we may change this later.

PR #14978, opened 2017-02-14
move locations queries

emord:
@esoergel Moved these around after we offlined

buddies @czue @millerdev

### DIFF
--- a/corehq/apps/es/tests/test_esquery.py
+++ b/corehq/apps/es/tests/test_esquery.py
@@ -141,28 +141,9 @@ class TestESQuery(ElasticTestMixin, TestCase):
                 .xmlns('banana')
         self.checkQuery(query, json_output)
 
-    @patch('corehq.apps.locations.models.SQLLocation.objects.get_locations_and_children_ids')
-    def test_users_at_locations_and_descendants(self, locations_patch):
-        location_ids = ['09d1a58cb849e53bb3a456a5957d998a', '09d1a58cb849e53bb3a456a5957d99ba']
-        children_ids = ['19d1a58cb849e53bb3a456a5957d998a', '19d1a58cb849e53bb3a456a5957d99ba']
-        all_ids = location_ids + children_ids
-        locations_patch.return_value = location_ids + children_ids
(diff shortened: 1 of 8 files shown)


corehq/apps/es/tests/test_esquery.py
```diff
@@ -141,28 +141,9 @@ def test_filtered_forms(self):
                 .xmlns('banana')
         self.checkQuery(query, json_output)
 
-    @patch('corehq.apps.locations.models.SQLLocation.objects.get_locations_and_children_ids')
-    def test_users_at_locations_and_descendants(self, locations_patch):
```
emord: Because the sql call was being patched this was basically testing the same thing that `test_users_at_locations` tests

corehq/apps/es/users.py
```diff
@@ -56,19 +56,6 @@ def show_only_inactive(self):
...
-        return self.location(location_ids)
-
-    def users_at_locations(self, location_ids):
-        return self.location(location_ids)
-
-    def users_at_accessible_locations(self, domain_name, user):
-        from corehq.apps.locations.models import SQLLocation
-        accessible_location_ids = SQLLocation.active_objects.accessible_location_ids(domain_name, user)
-        return self.users_at_locations(accessible_location_ids)
-
```
esoergel: Oh awesome, were these just not used anywhere else?

corehq/apps/locations/dbaccessors.py
```diff
@@ -79,7 +79,7 @@ def get_users_location_ids(domain, user_ids):
 
 
 def user_ids_at_locations(location_ids):
-    return UserES().location(location_ids).exclude_source().run().hits
+    return UserES().location(location_ids).get_ids()
```
esoergel: :+1:

esoergel (approved):
Big fan of all these changes, thanks, J:moneybag:!

corehq/apps/es/users.py
emord: They were only used inside of the functions that I moved over. I moved these with them

PR #14980, opened 2017-02-14
Add REUSE_DB=wipe option to clean database

esoergel:
You ever find yourself not properly cleaning up after something and forced with a `REUSE_DB=reset`?  Perhaps you had a bug in your `setUpClass`, so it never tore down some stuff and now you get a CouchUser with username blah already exists".  Fear no more - this deletes all docs in Couch and SQL, letting you keep testing without deleting the test dbs and rebuilding and migrating them.

@dimagi/test-team looking for feedback, then I'll ping the HQ team to notify once we're happy with it.

### DIFF
--- a/corehq/tests/nose.py
+++ b/corehq/tests/nose.py
@@ -15,21 +15,21 @@ import logging
 import os
 import sys
 import threading
-import types
 from fnmatch import fnmatch
-
-from django.apps import apps
 
 from couchdbkit import ResourceNotFound
 from couchdbkit.ext.django import loading
 from django.core.management import call_command
 from mock import patch, Mock
 from nose.plugins import Plugin
-from django.apps import AppConfig
 from django.conf import settings
 from django.db.backends.base.creation import TEST_DATABASE_PREFIX
 from django.db.utils import OperationalError
 from django_nose.plugin import DatabaseContext
+
+from dimagi.utils.couch.database import iter_bulk_delete
+from corehq.util.test_utils import unit_testing_only
+from corehq.preindex import get_preindex_plugin
 
 log = logging.getLogger(__name__)
 
@@ -173,11 +173,13 @@ class HqdbContext(DatabaseContext):
     Other supported `REUSE_DB` values:
 
     - `REUSE_DB=reset` : drop existing, then create and migrate new test
-      databses, but do not teardown after running tests. This is
+      databases, but do not teardown after running tests. This is
       convenient when the existing databases are outdated and need to be
       rebuilt.
     - `REUSE_DB=teardown` : skip database setup; do normal teardown after
       running tests.
+    - `REUSE_DB=wipe` : same as `REUSE_DB=1` except delete all objects from the
+      databases before running tests.  Much faster than `reset`.
     - `REUSE_DB=migrate` : same as `REUSE_DB=1` except migrate databases
       before running tests.
     """
@@ -215,6 +217,8 @@ class HqdbContext(DatabaseContext):
         if self.skip_setup_for_reuse_db and self._databases_ok():
             if self.reuse_db == "migrate":
                 call_command('migrate_multi', interactive=False)
+            if self.reuse_db == "wipe":
+                empty_databases()
             return  # skip remaining setup
 
         if self.reuse_db == "reset":
@@ -296,6 +300,24 @@ def print_imports_until_thread_change():
 
     # Register the import hook. See https://www.python.org/dev/peps/pep-0302/
     sys.meta_path.append(InfoImporter())
+
+
+@unit_testing_only
+def empty_databases():
+    """
+    Best effort at emptying all documents from all databases.
+    Useful when you break a test and it doesn't clean up properly. This took
+    about 5 seconds to run when trying it out.
+    """
+    for db in get_preindex_plugin('couchapps').get_dbs('all_docs'):
+        all_ids = [result['id'] for result in db.view(
+            '_all_docs',
+            include_docs=False,
+            reduce=False,
+        ) if '_design' not in result['id']]  # don't delete design docs
+        iter_bulk_delete(db, all_ids, chunksize=500)
+    call_command('flush', interactive=False)
+
 
 if os.environ.get("HQ_TESTS_PRINT_IMPORTS"):
     print_imports_until_thread_change()
--- a/corehq/tests/nose.py
+++ b/corehq/tests/nose.py
@@ -22,6 +22,7 @@ from couchdbkit.ext.django import loading
 from django.core.management import call_command
 from mock import patch, Mock
 from nose.plugins import Plugin
+from nose.tools import nottest
 from django.conf import settings
 from django.db.backends.base.creation import TEST_DATABASE_PREFIX
 from django.db.utils import OperationalError
@@ -276,6 +277,7 @@ def print_imports_until_thread_change():
     sys.meta_path.append(InfoImporter())
 
 
+@nottest
 @unit_testing_only
 def get_all_test_dbs():
     all_dbs = couch_config.all_dbs_by_db_name.values()
--- a/corehq/tests/nose.py
+++ b/corehq/tests/nose.py
@@ -27,6 +27,7 @@ from django.db.backends.base.creation import TEST_DATABASE_PREFIX
 from django.db.utils import OperationalError
 from django_nose.plugin import DatabaseContext
 
+from corehq.tests.noseplugins.cmdline_params import CmdLineParametersPlugin
 from corehq.util.couchdb_management import couch_config
 from corehq.util.test_utils import unit_testing_only
 
@@ -171,27 +172,22 @@ class HqdbContext(DatabaseContext):
 
     Other supported `REUSE_DB` values:
 
-    - `REUSE_DB=reset` : drop existing, then create and migrate new test
-      databases, but do not teardown after running tests. This is
-      convenient when the existing databases are outdated and need to be
-      rebuilt.
-    - `REUSE_DB=teardown` : skip database setup; do normal teardown after
-      running tests.
-    - `REUSE_DB=flush` : same as `REUSE_DB=1` except delete all objects from the
-      databases before running tests.  Much faster than `reset`.
-    - `REUSE_DB=migrate` : same as `REUSE_DB=1` except migrate databases
-      before running tests.
+    The following flags are also useful in conjunction with REUSE_DB=1:
+        --reset-db
+        --flush-db
+        --migrate-db
+        --teardown-db
     """
 
     def __init__(self, tests, runner):
-        self.reuse_db = reuse_db = os.environ.get("REUSE_DB")
-        self.skip_setup_for_reuse_db = reuse_db and reuse_db != "reset"
-        self.skip_teardown_for_reuse_db = reuse_db and reuse_db != "teardown"
+        reuse_db = CmdLineParametersPlugin.get('db_action') or os.environ.get("REUSE_DB")
+        self.reuse_db = reuse_db
+        self.skip_setup_for_reuse_db = reuse_db and reuse_db != "reset_db"
+        self.skip_teardown_for_reuse_db = reuse_db and reuse_db != "teardown_db"
         super(HqdbContext, self).__init__(tests, runner)
 
     def should_skip_test_setup(self):
-        # FRAGILE look in sys.argv; can't get nose config from here
-        return "--collect-only" in sys.argv
+        return CmdLineParametersPlugin.get('collect_only')
 
     def setup(self):
         if self.should_skip_test_setup():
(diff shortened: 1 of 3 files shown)


benrudolph:
❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ 

wanted this for so long!!

benrudolph (approved):

corehq/tests/nose.py
```diff
@@ -297,5 +301,23 @@ def find_module(self, name, path=None):
...
 
+
+@unit_testing_only
+def empty_databases():
+    """
+    Best effort at emptying all documents from all databases.
+    Useful when you break a test and it doesn't clean up properly. This took
+    about 5 seconds to run when trying it out.
+    """
+    for db in get_preindex_plugin('couchapps').get_dbs('all_docs'):
```
snopoke: Could be clearer to use `CouchConfig().all_dbs_by_slug().values()`

snopoke (commented):
Since this is a once off command that you'd run (as opposed to with every test run) would it not be better to just make a management command that does it?

Also the `flush` management command will delete ALL data include data added during migration which might put the DB in a weird state e.g. no roles etc.

millerdev:
I like @snopoke's idea of a management command, although this is a bit more distance between that and what you're thinking about when tests fail due to bad/stale data. `REUSE_DB` is what you're thinking about when that happens.

Related: could this perhaps encourage or at least allow people to always run tests with `REUSE_DB=wipe` and fall into a (bad) habit of writing tests that do not clean up after themselves?

esoergel:
@snopoke 
> Since this is a once off command that you'd run (as opposed to with every test run) would it not be better to just make a management command that does it?

I was thinking of this as an analog to `REUSE_DB=reset`.  I imagine a typical use case would be:
1. run tests
2. inadvertently break setUpClass
3. run tests again, get a failure
4. run tests again with a wipe, which should succeed

My assumption is that every time you use this, you'll also want to run the test you're working on.

> Also the flush management command will delete ALL data include data added during migration which might put the DB in a weird state e.g. no roles etc.

Yeah, I saw that in the django docs.  It did work for the test I was working on, so I'm inclined to leave as-is until we run into a problem where this makes things worse (probably necessitating a reset).  When you say "no roles", what roles do you mean?  Postgres roles?

Do you have any alternative ideas on how to empty the postgres tables?  One approach I considered was taking a list of SQL models to run `Model.objects.delete()` on, and opportunistically adding to it as needed.  My hope was that if this works as-is, we won't need to bother (since that might not be an obvious fix for people unfamiliar with the implementation).

corehq/tests/nose.py
esoergel: :+1: I just copied this from somewhere, but that does look like a better approach.  Will update.

esoergel:
@millerdev 
> Related: could this perhaps encourage or at least allow people to always run tests with REUSE_DB=wipe and fall into a (bad) habit of writing tests that do not clean up after themselves?

That would be concerning.  There are two issues with that approach, though - `wipe` adds 5-10 seconds to the test run, which is enough that I wouldn't run it every time if I'm doing TDD.  Also, it would (hopefully) fail on travis.

We could also include some output when `wipe` runs like "deleting N docs, make sure to clean up after yourself!".  That way people would at least be aware it's happening and not forget to set it back to `REUSE_DB=1`.

snopoke:
I still think this belongs in a separate command but I concede that it has the same workflow as `reset`. I'd be in favor of even having a prompt which forces the user to confirm the action (I could have done with one of those on `reset` before).

millerdev:
> `wipe` adds 5-10 seconds to the test run

Given how slow many of our tests are, I doubt 5-10s is a big deal to some people. Just sayin'

I'd be in favor of a prompt on `wipe`, and would also be nice to prompt on `reset` too, although I think that already happens for SQL DB if not running with `--noinput`

esoergel:
@snopoke @millerdev I offlined with Daniel about this and he showed me [this](https://github.com/dimagi/commcare-hq/blob/dc25141/corehq/tests/nose.py#L174).  I converted the `REUSE_DB` options to flags, so now it's only a yes or no environment variable.  This way if you want to flush or reset the db, for instance, you have to specify manually each time you run the command.

I'm pretty happy with where this is now, so let me know what you think!

corehq/tests/noseplugins/cmdline_params.py
```diff
@@ -0,0 +1,59 @@
...
+            help="Migrate the test databases before running tests." + reuse_msg,
+        )
+        parser.add_option(
+            '--teardown-db',
+            action="store_true",
+            default=False,
+            help="Skip database setup; do normal teardown after running tests." + reuse_msg,
+        )
+        parser.add_option(
+            "--collect-only",
```
millerdev: Interesting that you're redefining this parameter here. Did you check that this does not interfere with the [collect plugin](http://nose.readthedocs.io/en/latest/plugins/collect.html)? I have no idea what happens if two plugins define the same option. But pretty cool if it works.

corehq/tests/noseplugins/cmdline_params.py
```diff
@@ -0,0 +1,59 @@
...
+            action="store_true",
+            default=False,
+        )
+
+    def configure(self, options, conf):
+        db_action = None
+        for action in ['reset_db', 'flush_db', 'migrate_db', 'teardown_db']:
+            if getattr(options, action):
+                db_action = action
```
millerdev: Looks like the last action found wins, but it would be nice if it threw an error if more than one of these options is present. They are mutually exclusive, and it's likely a user error if there's more than one. That was one of the nice things about using the value of the environment variable: only one could be specified. I suppose you could make it a single option with different values:

`--reusedb=[reset|flush|migrate|teardown]`

or you could do something here to error if more than one is present. What do you think?

corehq/tests/nose.py
```diff
@@ -171,27 +172,22 @@ class HqdbContext(DatabaseContext):
...
+        --flush-db
+        --migrate-db
+        --teardown-db
     """
 
     def __init__(self, tests, runner):
-        self.reuse_db = reuse_db = os.environ.get("REUSE_DB")
-        self.skip_setup_for_reuse_db = reuse_db and reuse_db != "reset"
-        self.skip_teardown_for_reuse_db = reuse_db and reuse_db != "teardown"
+        reuse_db = CmdLineParametersPlugin.get('db_action') or os.environ.get("REUSE_DB")
```
millerdev: Looks like `REUSE_DB=flush_db` would also work. Is that intentional? I think you want `bool(os.environ.get("REUSE_DB"))`

corehq/tests/noseplugins/cmdline_params.py
```diff
@@ -0,0 +1,59 @@
...
+        type(self).parameters['db_action'] = db_action
```
millerdev: what do you think of adding a check for `REUSE_DB`?
```py
 if db_action is not None and not os.environ.get("REUSE_DB"):
    raise ValueError("REUSE_DB=1 is required with --" + db_action.replace("_", "-"))
```

corehq/tests/noseplugins/cmdline_params.py
millerdev: Nit: could `db_action = action[:-3]  # strip _db suffix` and make fewer changes to `nose.py`

millerdev (commented):
Looking good!

corehq/tests/nose.py
esoergel: actually, has this never worked?  env variables are strings, so if it exists at all it'll `bool` to `True`:
```
$ REUSE_DB=0 python -c "import os; print bool(os.environ.get('REUSE_DB'))"
True
```
Thinking I'll change this to:
```python
reuse_db = CmdLineParametersPlugin.get('db_action') or os.environ.get("REUSE_DB") != "0"
```

corehq/tests/noseplugins/cmdline_params.py
esoergel: Daniel and I offlined, this works fine still.

corehq/tests/noseplugins/cmdline_params.py
esoergel: Yeah, I thought about `--reusedb=flush`, it does seem like a more sensible parameter.  My only hesitation is that it's obnoxious to type, but that's not a great argument.

esoergel (commented):
responded to feedback.

corehq/tests/noseplugins/cmdline_params.py
```diff
@@ -3,6 +3,17 @@
...
 
+REUSE_DB_HELP = """
+To be used in conjunction with the environment variable REUSE_DB=1.
+reset: Drop existing test dbs, then create and migrate new ones, but do not
+    teardown after running tests. This is convenient when the existing databases
+    are outdated and need to be rebuilt.
+flush: Flush all objects from the old test databases before running tests.
+    Much faster than `reset`.
+migrate: Migrate the test databases before running tests.
+    teardown: Skip database setup; do normal teardown after running tests.
```
millerdev: indentation error. Also, have you checked the format of this when printed with `--help`? Unfortunately I think argparse will indiscriminately remove all your nice indentation here, which is a shame. Not sure there is any better approach though :(

corehq/tests/noseplugins/cmdline_params.py
```diff
@@ -12,33 +23,11 @@ class CmdLineParametersPlugin(Plugin):
...
-            action="store_true",
-            default=False,
-            help="Migrate the test databases before running tests." + reuse_msg,
-        )
         parser.add_option(
-            '--teardown-db',
-            action="store_true",
-            default=False,
-            help="Skip database setup; do normal teardown after running tests." + reuse_msg,
+            '--reuse_db',
```
millerdev: nit: maybe it's just me, but underscores in command-line options drive me crazy. I think the convention would be `--reuse-db` or `--reusedb`. My preference is the latter because it's a tiny bit shorter/easier to type, but up to you.

millerdev: Also, fwiw, argparse automatically converts dashes to underscores when saving the argument on the options object.

millerdev (commented):
Tests failed:
```
  File "/mnt/commcare-hq-ro/corehq/tests/nose.py", line 180, in __init__
    reuse_db = CmdLineParametersPlugin.get('reuse_db') or string_to_boolean(os.environ.get("REUSE_DB"))
  File "submodules/dimagi-utils-src/dimagi/utils/parsing.py", line 22, in string_to_boolean
    raise ValueError("%s is not a parseable boolean!" % val)
ValueError:  is not a parseable boolean!
```

corehq/tests/noseplugins/cmdline_params.py
esoergel: Yeah, it concats all whitespace to a single space.  I did this just to make it easier to read in the file.  I saw some stack overflow posts on writing your own renderer which preserves whitespace, but I'm not about to do that.

corehq/tests/noseplugins/cmdline_params.py
esoergel: yeah, I meant to do `reusedb`.  I'm familiar with the conversion to underscores although that bothers me more than having underscores in the first place.

millerdev (approved):

corehq/tests/noseplugins/cmdline_params.py
```diff
@@ -0,0 +1,44 @@
...
+REUSE_DB_HELP = """
+To be used in conjunction with the environment variable REUSE_DB=1.
+reset: Drop existing test dbs, then create and migrate new ones, but do not
+    teardown after running tests. This is convenient when the existing databases
+    are outdated and need to be rebuilt.
+flush: Flush all objects from the old test databases before running tests.
+    Much faster than `reset`.
+migrate: Migrate the test databases before running tests.
+teardown: Skip database setup; do normal teardown after running tests.
+"""
```
esoergel: @dimagi/team-commcare-hq 
Hey CommCareHQ team - this is a change to REUSE_DB.  The environment variable is now only a boolean - so `REUSE_DB=1` if you want to reuse the db and `REUSE_DB=0` (or just don't set it) if you don't.

The various options such as `reset` and `migrate` are now specified via a command-line parameter.  This new parameter is `--reusedb=migrate` and has four options as described in the code above or via `--help`.

This PR introduces a new option, called `flush`, which will delete all documents in couch and postgres (feel free to extend if you come across more it should clean).  This is useful for when you're working on a test and made a mistake in a `setUpClass` for instance, leaving behind test data that then gives you errors like "user X already exists!".  Functionally it's similar to `reset`, except it takes around 5 seconds to run, rather than the minute or more a normal db rebuild takes.  Rather than deleting the databases, rebuilding them, and running migrations, this essentially deletes all rows in the databases.

`reset` is still nice to have around as a big hammer if the test db just gets all screwed up, but my hope is that `flush` and `migrate` will replace it in day-to-day use.

benrudolph: for those looking for a quick copy paste for sniffer:
```
REUSE_DB=1 sniffer -x path.to.test -x --reusedb=migrate
```

PR #14982, opened 2017-02-15
Revert "Revert "Display graphs in Web Apps""

benrudolph:
Reverts dimagi/commcare-hq#14981 @wpride ping when g2g


benrudolph:
i think this is no longer necessary

PR #14983, opened 2017-02-15
A wild stab in the dark

kaapstorm:
Opening for sake of discussion. Untested. 

Context: [FB 246992](http://manage.dimagi.com/default.asp?246992)
TL;DR: The first index in the XPath in a case list detail field template is relative: 
```xml
    <xpath function="instance('casedb')/casedb/case[
        @case_id=instance('casedb')/casedb/case[
            @case_id=index/host
        ]/index/host
    ]/case_name"/>
```
but it should be absolute:
```xml
    <xpath function="instance('casedb')/casedb/case[
        @case_id=instance('casedb')/casedb/case[
            @case_id=current()/index/host
        ]/index/host
    ]/case_name"/>
```

@snopoke cc @benrudolph @czue


corehq/apps/app_manager/detail_screen.py
```diff
@@ -568,7 +572,10 @@ def xpath(self):
         property = parts.pop()
         indexes = parts
 
-        use_relative = property != '#owner_name'
+        if property == '#owner_name' or self.is_case_list_column:
```
snopoke: could you not just say `use_relative = bool(indexes) or property != '#owner_name'`

snopoke:
yes, sorry, `not indexes`. I'm pretty sure what Clayton was saying is that
the first index query for ALL indexed details queries should use `current()`

snopoke:
Pretty sure this will break tests

kaapstorm:
:open_mouth: It didn't when I ran app_manager tests locally. So I'm adding a test.

czue:
what's the status of this PR?

kaapstorm:
I think it's good to go. What do you reckon, @snopoke ?

corehq/apps/app_manager/tests/test_case_list_form.py
```diff
@@ -153,6 +163,14 @@ def test_case_list_form_parent_child_basic(self):
         )
 
         factory.form_requires_case(update_person_form, parent_case_type='house')
+        register_person_module.case_details.short.columns.append(
+            DetailColumn(
+                header={'en': 'a'},
+                model='case',
+                field='house/case_name',  # Include a parent case property in the case list
```
snopoke: Although this is technically legal I'd prefer if we stuck to the normal references and made this `parent/case_name`

corehq/apps/app_manager/detail_screen.py
```diff
@@ -568,7 +568,7 @@ def xpath(self):
         property = parts.pop()
         indexes = parts
 
-        use_relative = property != '#owner_name'
+        use_relative = not indexes and property != '#owner_name'
```
snopoke: I think this would be easier to read if it was inverted:
```python
use_absolute = indexes or property == '#owner_name'
```

snopoke (changes requested):
minor changes

corehq/apps/app_manager/detail_screen.py
kaapstorm: Agreed

czue:
thanks for powering through this!

PR #14985, opened 2017-02-15
move task to later

snopoke:
Move it to later so that there isn't an exact overlap with
the 'update_calculated_properties' task. This should
allow us to see which one has the memory leak.


snopoke:
@benrudolph

snopoke:
:white_check_mark: @benrudolph

benrudolph:
nice sorry somehow this i missed this

PR #14987, opened 2017-02-15
Added DHIS2 connection settings

kaapstorm:
It looks like this:
dhis2_conn

([Spec](https://docs.google.com/document/d/1PJrGSedNhRlNxR7fK5HUj30Y3R7_HqfK7qbysWIJEsY/edit#) -- "Test Connection" button not yet implemented.)

@benrudolph @sheelio cc @snopoke


corehq/apps/dhis2/models.py
```diff
@@ -0,0 +1,20 @@
+from dimagi.ext.couchdbkit import Document, StringProperty
+from dimagi.utils.couch.cache import cache_core
+
+
+class Dhis2Connection(Document):
```
snopoke: It would be useful if you could map out the data model for this feature so that we can get a better picture of how the pieces fit together. I think it would also be useful to validate that before diving into the code.

snopoke (commented):
Looks fine but would like to see some more up-front design work.

corehq/apps/dhis2/models.py
```diff
@@ -0,0 +1,20 @@
...
+    domain = StringProperty()
+    server_url = StringProperty()
+    username = StringProperty()
+    password = StringProperty()
+
+    @classmethod
+    def for_domain(cls, domain):
+        res = cache_core.cached_view(
```
benrudolph: nit this should be in a `dbaccessors` file

benrudolph (approved):
deferring to simon's thoughts about design work (i don't have much context on what this is)

corehq/apps/dhis2/models.py
snopoke: and should rather use the `quickcache` decorator

corehq/apps/domain/views.py
```diff
@@ -43,6 +43,9 @@
     enable_case_search,
     disable_case_search,
 )
+from corehq.apps.dhis2.dbaccessors import get_dhis2_connection
+from corehq.apps.dhis2.forms import Dhis2ConnectionForm
+from corehq.apps.dhis2.models import Dhis2Connection
```
dimagimon: F401 'Dhis2Connection' imported but unused

PR #14994, opened 2017-02-15
Allow reg from case list in CLMI-only modules

orangejenny:
http://manage.dimagi.com/default.asp?247629

Tested on staging that the build succeeds and that the app behaves as expected on the phone. When you register a case from a CLMI-only module, you get sent to the home screen afterwards, which I think is fine.

@snopoke 
cc @gcapalbo


snopoke:
Pardon the ignorance but what's "CLMI"?

orangejenny:
Oops, sorry for jargon - "case list menu item", a checkbox in module settings > case mgmt that lets you set up a case list on the phone that doesn't have any forms associated with it.

snopoke:
Got it. I know what that is just couldn't get the acronym.

gcapalbo (approved):

PR #14865, opened 2017-02-10
Remove a runner

benrudolph:

PR #14995, opened 2017-02-15
Je/custom run with all backends

emord:
Just look at first/third commit. second commit was changing windows line endings to unix to match rest of hq and pep8 was yelling at me.

Noticed this when looking at test output from https://github.com/dimagi/commcare-hq/pull/14865 Since this is custom code and we know that they will only be on sql domains, we shouldn't test on couch.

before: `Ran 124 tests in 522.590s`

after: `Ran 124 tests in 214.298s`

@czue @millerdev


benrudolph (approved):
cha-ching. that reasoning makes sense to me

benrudolph:
maybe this'll put us over the edge for 3 runners

custom/ucla/tests.py
```diff
@@ -1,20 +1,20 @@
...
 from corehq.apps.users.models import WebUser
-from corehq.form_processor.tests.utils import run_with_all_backends
 from corehq.util.test_utils import create_test_case
 from custom.ucla.api import ucla_message_bank_content
 
 Reminder = namedtuple('Reminder', ['domain', 'schedule_iteration_num', 'current_event_sequence_num'])
 Handler = namedtuple('Handler', ['events'])
 
 
+@override_settings(TESTS_SHOULD_USE_SQL_BACKEND=True)
```
czue: confirming you checked this domain?

czue (approved):
cool, looks good. just asked a dumb question since it was the only one I wasn't sure myself was on SQL.

custom/ucla/tests.py
emord: Yep. it's for a new domain. (I wrote this code in the past month)

PR #15000, opened 2017-02-16
Fix a couple preview bugs

benrudolph:
@czue this fixes the loading of a combobox from a case and removes the green checkmark. 

@dimagi/ux -- cory and i decided it would be good to remove the green checkmark from app preview until it works across tabs. it will still show the red dot when it's out of date

cc: @kaapstorm


orangejenny:
> cory and i decided it would be good to remove the green checkmark from app preview until it works across tabs. it will still show the red dot when it's out of date

sounds reasonable

biyeun (approved):

PR #15001, opened 2017-02-16
Module case type tooltip

benrudolph:
screen shot 2017-02-16 at 11 19 31 am

Similar to the form icon tooltip this displays what case type the module is without having to open it. I find this very useful in the current app manager. I put it behind the support toggle.

cc: @kaapstorm

### DIFF
--- a/corehq/apps/app_manager/templates/app_manager/v1/managed_app.html
+++ b/corehq/apps/app_manager/templates/app_manager/v1/managed_app.html
@@ -119,7 +119,7 @@
                             {% elif module.module_type == 'shadow' %}class="fa fa-folder-open-o appmanager-icon-type"
                             {% else %}class="fa fa-folder-open"{% endif %}
 
-                            {% if module.case_type and request|toggle_enabled:"SUPPORT" %}
+                            {% if module.case_type and app.get_case_types|length > 1 %}
                             title="Case Type: {{ module.case_type }}"
                             data-toggle="tooltip"
                             data-placement="top"
--- a/corehq/apps/app_manager/templates/app_manager/v1/managed_app.html
+++ b/corehq/apps/app_manager/templates/app_manager/v1/managed_app.html
@@ -110,12 +110,22 @@
                     <li class="edit-module-li {% ifequal module.id selected_module.id %}{% if not form %}
                         active{% endif %}{% endifequal %}"
                         data-index="{{ module.id }}">
-                        <a href="{% url "view_module" domain app.id module.id %}">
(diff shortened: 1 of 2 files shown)


benrudolph:
@czue might be interested

kaapstorm:
Oooo! Nice!

kaapstorm:
Failure is real. You just need to update the template diff.

czue (approved):
good stuff. adding "announce" label to try and spread the gospel

corehq/apps/app_manager/templates/app_manager/v1/managed_app.html
```diff
@@ -110,12 +110,22 @@
...
-                            {% elif module.module_type == 'report' %}<i class="fa fa-bar-chart appmanager-icon-type"></i>
-                            {% elif module.module_type == 'shadow' %}<i class="fa fa-folder-open-o appmanager-icon-type"></i>
-                            {% else %}<i class="fa fa-folder-open"></i>{% endif %}
+                            <i
+                            {% if module.module_type == 'advanced' %}class="fa fa-flask appmanager-icon-type"
+                            {% elif module.module_type == 'report' %}class="fa fa-bar-chart appmanager-icon-type"
+                            {% elif module.module_type == 'shadow' %}class="fa fa-folder-open-o appmanager-icon-type"
+                            {% else %}class="fa fa-folder-open"{% endif %}
+
+                            {% if module.case_type and request|toggle_enabled:"SUPPORT" %}
```
czue: why is this "support" only?

benrudolph: last time i tried this with form icons it was decided to go under support toggle. happy to take out though, though should ping @dimagi/ux and @dimagi/product

benrudolph:
thanks, keep forgetting to add that tag

corehq/apps/app_manager/templates/app_manager/v1/managed_app.html
orangejenny: How about releasing it to everyone, but limiting it to apps that have multiple case types? That'd keep it from being distracting for new users / simple apps.

FWIW, the form icons are going to be released to everyone in app manager v2, we're just holding off for the sake of bundling visual changes together. I'm not worried about that here, though, because this is change you only see on hover.

benrudolph: good call, changes made: https://github.com/dimagi/commcare-hq/pull/15001/commits/99056d13f46a304eb5b8a71981ff83552c80dcb9

PR #15004, opened 2017-02-16
pass object ID's to tasks rather than the full objects

snopoke:
Should have better memory profile than passing full objects. Also just best practice.

### DIFF
--- a/corehq/apps/export/dbaccessors.py
+++ b/corehq/apps/export/dbaccessors.py
@@ -86,16 +86,16 @@ def _get_export_instance(cls, key):
     return [cls.wrap(result['doc']) for result in results]
 
 
-def get_all_daily_saved_export_instances():
+def get_all_daily_saved_export_instance_ids():
     from .models import ExportInstance
     results = ExportInstance.get_db().view(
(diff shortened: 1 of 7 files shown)


mkangia (approved):
Good call. Nice!

corehq/apps/export/dbaccessors.py
```diff
@@ -86,16 +86,16 @@ def _get_export_instance(cls, key):
...
     from .models import ExportInstance
     results = ExportInstance.get_db().view(
         "export_instances_by_is_daily_saved/view",
         startkey=[True],
         endkey=[True, {}],
-        include_docs=True,
+        include_docs=False,
         reduce=False,
     ).all()
-    return [_properly_wrap_export_instance(result['doc']) for result in results]
```
benrudolph: 😖

benrudolph (approved):
looks good, though my intuition would have been that the memory leak would have been from generating the export, not the export instance object

snopoke:
Yea, I agree that this almost certainly isn't the memory leak. It's just
the first thing that struck me when I glanced at it.

PR #15005, opened 2017-02-16
Add missing template

wpride:
Fixes http://manage.dimagi.com/default.asp?247394

@benrudolph @biyeun is there any reason not to unify the imports here with those in the main template?


benrudolph:
yes we should do that...

PR #15011, opened 2017-02-16
Remove poster from reqs

emord:
Ran `ag --py poster` and got no results. I don't think we've used this for a while. (hasn't even been updated in 6 years)

buddies @czue @millerdev


millerdev (approved):

czue (approved):
nice find

PR #15013, opened 2017-02-16
Don't error if export table configuration has empty label

NoahCarnahan:
Addresses http://manage.dimagi.com/default.asp?246861
If the label was falsey, the writer would attempt to use the TableConfiguration for the label, but as the configuration is not a string, it would 500. This ensures that every TableConfiguration has a truthy label when we attempt to write the table.
cc @esoergel 
fyi @kaapstorm (who also worked on this bug)

### DIFF
--- a/corehq/apps/export/tests/test_get_export_file.py
+++ b/corehq/apps/export/tests/test_get_export_file.py
@@ -513,6 +513,43 @@ class WriterTest(SimpleTestCase):
                 }
             )
 
+    def test_empty_table_label(self):
+        export_instance = FormExportInstance(
+            export_format=Format.JSON,
+            domain=DOMAIN,
+            case_type=DEFAULT_CASE_TYPE,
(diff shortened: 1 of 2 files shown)


corehq/apps/export/tests/test_get_export_file.py
```diff
@@ -513,6 +513,43 @@ def test_multiple_write_export_instance_calls(self):
...
+                        item=ScalarItem(
+                            path=[PathNode(name='form'), PathNode(name='q1')],
+                        ),
+                        selected=True
+                    ),
+                ]
+            )]
+        )
+        writer = _get_writer([export_instance])
+        with writer.open([export_instance]):
```
NoahCarnahan: Before this bug fix, the test would raise an exception at this point.

esoergel (approved):

kaapstorm:
Nice find.

PR #15014, opened 2017-02-16
Kill CommTrackTest

esoergel:
This class was essentially a dumping ground for the `setUp` needs of bunch of vaguely related tests.  Much of that wasn't necessary at all, and in most cases could go in `setUpClass`.  I cut down on a bunch of this needless `setUp`.  The `test_xml.py` tests still take around 135s (down from 220s), it wasn't trivial to pull stuff out in to a `setUpClass`, but I'm confident that could be improved further.  All told, this brings the total runtime of these tests from 407s to 225s - a savings of 3:02.
@dimagi/test-team

### DIFF
--- a/corehq/apps/commtrack/tests/test_fixture.py
+++ b/corehq/apps/commtrack/tests/test_fixture.py
@@ -1,20 +1,36 @@
+import datetime
 import random
 import string
 from xml.etree import ElementTree
 
+from django.test import TestCase
+
 from casexml.apps.case.xml import V1
+from casexml.apps.phone.models import SyncLog
 from casexml.apps.phone.tests.utils import generate_restore_payload, create_restore_user
 from corehq.apps.app_manager.tests.util import TestXmlMixin
 from corehq.apps.programs.fixtures import program_fixture_generator
 from corehq.apps.products.fixtures import product_fixture_generator
 from corehq.apps.products.models import Product
 from corehq.apps.programs.models import Program
-from corehq.apps.commtrack.tests.util import CommTrackTest
-from casexml.apps.phone.models import SyncLog
-import datetime
+from corehq.apps.commtrack.tests import util
 
 
-class FixtureTest(CommTrackTest, TestXmlMixin):
+class FixtureTest(TestCase, TestXmlMixin):
+    domain = "fixture-test"
+
+    @classmethod
+    def setUpClass(cls):
+        super(FixtureTest, cls).setUpClass()
+        cls.domain_obj = util.bootstrap_domain(cls.domain)
+        util.bootstrap_location_types(cls.domain)
+        util.bootstrap_products(cls.domain)
+        cls.user = create_restore_user(cls.domain)
+
+    @classmethod
+    def tearDownClass(cls):
+        cls.domain_obj.delete()
+        super(FixtureTest, cls).tearDownClass()
 
     def _random_string(self, length):
         return ''.join(random.choice(string.ascii_lowercase)
@@ -109,7 +125,7 @@ class FixtureTest(CommTrackTest, TestXmlMixin):
         )
 
     def test_product_fixture(self):
-        user = create_restore_user(self.domain.name)
+        user = self.user
         xml = self.generate_product_fixture_xml(user)
         fixture = product_fixture_generator(user, V1)
 
@@ -119,7 +135,7 @@ class FixtureTest(CommTrackTest, TestXmlMixin):
         )
 
     def test_selective_product_sync(self):
-        user = create_restore_user(self.domain.name)
+        user = self.user
 
         expected_xml = self.generate_product_fixture_xml(user)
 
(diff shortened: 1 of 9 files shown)


corehq/apps/commtrack/tests/test_fixture.py
```diff
@@ -1,20 +1,36 @@
...
+        super(FixtureTest, cls).setUpClass()
+        cls.domain_obj = util.bootstrap_domain(cls.domain)
+        util.bootstrap_location_types(cls.domain)
+        util.bootstrap_products(cls.domain)
+        cls.user = create_restore_user(cls.domain)
 
-class FixtureTest(CommTrackTest, TestXmlMixin):
+    @classmethod
+    def tearDownClass(cls):
+        cls.domain_obj.delete()
```
millerdev: do you need to `cls.user.delete()` here?

millerdev (approved):
> brings the total runtime of these tests from 407s to 225s - a savings of 3:02

📉 woohoo!

Actually, approval is contingent on the answer to my one comment on the code.

corehq/apps/commtrack/tests/test_fixture.py
esoergel: I don't think so - `create_restore_user` creates a restore user instance, which isn't a thing that can be saved to the db, and a CommCareUser, which is directly tied to a domain, meaning `domain_obj.delete()` takes care of cleaning it up.

emord: You may need to delete locations too

corehq/apps/commtrack/tests/test_locations.py
```diff
@@ -12,77 +17,74 @@ def _count_root_locations(domain):
...
 
-            self.assertEqual(sql_village.name, test_village.name)
-            self.assertEqual(sql_village.domain, test_village.domain)
-        except SQLLocation.DoesNotExist:
-            self.fail("Synced SQL object does not exist")
+    def tearDown(self):
+        self.user.delete()
+        SQLLocation.objects.all().delete()
+        delete_all_docs_by_doc_type(Location.get_db(), ['Location'])
+        super(LocationsTest, self).tearDown()
```
emord: Do you need to clean up location types as well? Also why not use the `delete_all_locations` helper?

corehq/apps/commtrack/tests/test_products.py
```diff
@@ -1,18 +1,32 @@
...
+        cls.domain_obj = bootstrap_domain(cls.domain)
+
+    @classmethod
+    def tearDownClass(cls):
+        cls.domain_obj.delete()
+        super(ProductsTest, cls).tearDownClass()
 
     def test_archive(self):
-        original_list = Product.by_domain(self.domain.name, wrap=False)
+        bootstrap_products(self.domain)
```
emord: Do the products get cleaned up from this test?

corehq/apps/commtrack/tests/test_programs.py
```diff
@@ -63,3 +62,4 @@ def test_delete(self):
             self.default_program._id,
             SQLProduct.objects.get(product_id=self.products[0]._id).program_id
         )
+        self.domain.delete()
```
emord: It's better to use `self.addCleanup(self.domain.delete)` so that even if this test errors out it will still be deleted

corehq/apps/commtrack/tests/test_programs.py
```diff
@@ -1,22 +1,22 @@
...
 
 
-class ProgramsTest(CommTrackTest):
-
-    def setUp(self):
-        super(ProgramsTest, self).setUp()
+class ProgramsTest(TestCase):
+    def test_programs(self):
+        self.domain = bootstrap_domain(TEST_DOMAIN)
+        bootstrap_products(self.domain.name)
```
emord: Same question about deleting products

emord (commented):
Looks good. just a few comments about tes tcleanup. (there's some more places where I don't think it is cleaned up properly, but nothing out of line of my other comments)

esoergel:
@emord SQLLocations, LocationTypes, SQLProducts, and all couch documents are deleted by `domain.delete()`: https://github.com/dimagi/commcare-hq/blob/5504e4c90db29c287a2db2f726335e8874b0c6b3/corehq/apps/domain/deletion.py#L93-L95

corehq/apps/commtrack/tests/test_locations.py
esoergel: I didn't use `delete_all_locations` because I don't want to delete types (this is copy/pasted from there).  Types are defined in `setUpClass` and aren't modified by the tests.  They're cleaned up by `domain_obj.delete()`

benrudolph (commented):
Only did a quick review, looks like others are reviewing, but this is awesome! 🏃

czue:
nice, you're a hero for taking this on - this has bugged me for ages!

corehq/apps/commtrack/tests/test_programs.py
emord: I think this is the only outstanding comment, but merging anyways as this is a huge improvement.

esoergel: good call, was thinking this was in a `tearDownClass`.  Will update.

PR #15015, opened 2017-02-16
User properties soft assert and moved message

millerdev:
Triggers if form has new-style easy-reference user properties as well as old-style user case preload configuration.

Also added a message to the User Properties tab about where to find them now that they have moved:

screenshot 2017-02-16 15 04 56

@orangejenny @czue @emord


emord:
> Triggers if form has new-style easy-reference user properties as well as old-style user case preload configuration.

How does the app/form get migrated so that hte user properties show up in form builder?

millerdev:
> How does the app/form get migrated so that hte user properties show up in form builder?

management command: `migrate_app_to_cmitfb --usercase`

I'm planning to run that once Kevin has notified domain admins of the migration. The first group will be Dimagi FMs, and assuming that goes smoothly we'll likely migrate everyone else using usercase next. We will do the same for India. Swiss will be migrated once Kevin communicates with them.

emord (approved):
:+1: Thanks I misread this at first

czue:
:+1: nice idea on the screenshot

PR #15018, opened 2017-02-16
ICDS VHND survey indicator

emord:
@gcapalbo this'll be easiest commit by commit. The first part is refactoring and including the performance improvement you mentioned on the previous pr

@czue @millerdev

### DIFF
--- a/corehq/apps/locations/dbaccessors.py
+++ b/corehq/apps/locations/dbaccessors.py
@@ -94,7 +94,7 @@ def user_ids_at_accessible_locations(domain_name, user):
     return user_ids_at_locations(accessible_location_ids)
 
 
-def assigned_location_by_user_id(domain, location_ids):
+def get_user_ids_from_assigned_location_ids(domain, location_ids):
     """
     Returns {user_id: [location_id, location_id, ...], ...}
     """
@@ -116,7 +116,7 @@ def assigned_location_by_user_id(domain, location_ids):
     return ret
 
 
-def primary_location_by_user_id(domain, location_ids):
+def get_user_ids_from_primary_location_ids(domain, location_ids):
     """
     Returns {user_id: primary_location_id, ...}
     """
--- a/corehq/apps/locations/dbaccessors.py
+++ b/corehq/apps/locations/dbaccessors.py
@@ -92,3 +92,23 @@ def user_ids_at_accessible_locations(domain_name, user):
     from corehq.apps.locations.models import SQLLocation
     accessible_location_ids = SQLLocation.active_objects.accessible_location_ids(domain_name, user)
     return user_ids_at_locations(accessible_location_ids)
+
+
+def assigned_location_by_user_id(domain, location_ids):
+    """Get the ids of the locations the users are assigned to"""
+    result = (
+        UserES()
+        .domain(domain)
+        .location(location_ids)
+        .non_null('assigned_location_ids')
+        .fields(['assigned_location_ids', '_id'])
+        .run().hits
+    )
+    ret = {}
+    for r in result:
+        if 'assigned_location_ids' in r:
+            locs = r['assigned_location_ids']
+            if not isinstance(locs, list):
+                locs = [r['assigned_location_ids']]
+            ret[r['_id']] = locs
+    return ret
--- a/corehq/apps/locations/dbaccessors.py
+++ b/corehq/apps/locations/dbaccessors.py
@@ -95,7 +95,9 @@ def user_ids_at_accessible_locations(domain_name, user):
 
 
 def assigned_location_by_user_id(domain, location_ids):
-    """Get the ids of the locations the users are assigned to"""
+    """
+    Returns {user_id: [location_id, location_id, ...], ...}
+    """
     result = (
         UserES()
         .domain(domain)
@@ -111,4 +113,24 @@ def assigned_location_by_user_id(domain, location_ids):
             if not isinstance(locs, list):
                 locs = [r['assigned_location_ids']]
             ret[r['_id']] = locs
+    return ret
+
+
+def primary_location_by_user_id(domain, location_ids):
+    """
+    Returns {user_id: primary_location_id, ...}
+    """
+    result = (
+        UserES()
+        .domain(domain)
+        .primary_location(location_ids)
+        .non_null('location_id')
+        .fields(['location_id', '_id'])
+        .run().hits
+    )
+    ret = {}
+    for r in result:
+        if 'location_id' in r:
+            loc = r['location_id']
+            ret[r['_id']] = loc
     return ret
--- a/custom/icds/messaging/indicators.py
+++ b/custom/icds/messaging/indicators.py
@@ -12,7 +12,7 @@ from django.template import TemplateDoesNotExist
 from django.template.loader import render_to_string
 
 from corehq.apps.locations.dbaccessors import (
-    primary_location_by_user_id,
+    get_user_ids_from_primary_location_ids,
     get_users_location_ids,
 )
 from corehq.apps.locations.models import SQLLocation
@@ -87,7 +87,7 @@ class LSIndicator(SMSIndicator):
     @property
     @memoized
     def locations_by_user_id(self):
-        return primary_location_by_user_id(self.domain, set(self.awc_locations))
+        return get_user_ids_from_primary_location_ids(self.domain, set(self.awc_locations))
 
     @property
     @memoized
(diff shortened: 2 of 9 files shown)


custom/icds/messaging/indicators.py
```diff
@@ -143,19 +182,51 @@ def get_messages(self, language_code=None):
...
+            return string_to_datetime(date).date() if date else None
+
+        now_date = self.now.date()
+        user_ids_with_forms_in_time_frame = set()
+        for user_id, form in self.forms.items():
+            vhnd_date = convert_to_date(form['form']['vhsnd_date_planned'])
+            if (now_date - vhnd_date).days < 37:
+                user_ids_with_forms_in_time_frame.add(user_id)
+
+        user_ids_without_forms = self.aww_user_ids - user_ids_with_forms_in_time_frame
```
dimagimon: F841 local variable 'user_ids_without_forms' is assigned to but never used

custom/icds/tests/test_ls_vhnd_survey.py
```diff
@@ -0,0 +1,142 @@
+from datetime import date, datetime, timedelta
```
dimagimon: F401 'date' imported but unused

custom/icds/tests/test_ls_vhnd_survey.py
```diff
@@ -0,0 +1,142 @@
...
+        user.set_location(location)
+        UserESFake.save_doc(user._doc)
+        return user
+
+    @property
+    def today(self):
+        tz = pytz.timezone('Asia/Kolkata')
+        return datetime.now(tz=tz).date()
+
+    def _make_form(self, user_id, date):
```
dimagimon: F811 redefinition of unused 'date' from line 1

custom/icds/tests/test_ls_vhnd_survey.py
```diff
@@ -0,0 +1,142 @@
...
+        def es_formatted_date(date):
+            return date.strftime('%Y-%m-%d')
+
+        return {
+            u'form': {
+                u'meta': {
+                    u'user_id': user_id
+                },
+                u'vhsnd_date_planned':  date,
```
dimagimon: E241 multiple spaces after ':'

czue (commented):
don't have too much context for this so will let @gcapalbo review again. nice test to code ratio though.

corehq/apps/locations/dbaccessors.py
```diff
@@ -92,3 +92,23 @@ def user_ids_at_accessible_locations(domain_name, user):
     from corehq.apps.locations.models import SQLLocation
     accessible_location_ids = SQLLocation.active_objects.accessible_location_ids(domain_name, user)
     return user_ids_at_locations(accessible_location_ids)
+
+
+def assigned_location_by_user_id(domain, location_ids):
```
gcapalbo: not sure i understand why this is necessary - this looks like it's getting all the locations that the users are assigned to, by searching on primary location id? i think it's safe to only use a user's primary location for these indicators

custom/icds/messaging/indicators.py
```diff
@@ -115,8 +140,6 @@ class LSSubmissionPerformanceIndicator(LSIndicator):
     def __init__(self, domain, user):
         super(LSSubmissionPerformanceIndicator, self).__init__(domain, user)
 
-        child_locations = [l.location_id for l in self.user.sql_location.child_locations()]
-        self.aww_user_ids = user_ids_at_locations(child_locations)
```
gcapalbo: so the way i had imagined this working was that we'd call `get_user_ids_by_location` for each child location and then build a map of user id: location id which we could then use as an index to self.awc_locations to get the location name when we need to.

i realize that's using couch instead of es, and i think for something like this i'd rather rely on the primary data source for this info

custom/icds/tests/test_ls_vhnd_survey.py
```diff
@@ -0,0 +1,142 @@
...
+        self.addCleanup(aww_2.delete)
+        last_subs.return_value = {
+            self.aww.get_id: self._make_form(self.aww.get_id, self.today - timedelta(days=15)),
+            aww_2.get_id: self._make_form(aww_2.get_id, self.today - timedelta(days=37))
+        }
+        indicator = LSVHNDSurveyIndicator(self.domain, self.ls)
+        messages = indicator.get_messages()
+        self.assertEqual(len(messages), 1)
+        message = messages[0]
+        self.assertTrue('AWC2' in message)
```
gcapalbo: could we also add a `self.assertFalse('AWC1' in message)`?

custom/icds/tests/test_ls_vhnd_survey.py
```diff
@@ -0,0 +1,142 @@
...
+
+    def test_no_form_submitted(self, last_subs):
+        last_subs.return_value = {}
+        indicator = LSVHNDSurveyIndicator(self.domain, self.ls)
+        messages = indicator.get_messages()
+        self.assertEqual(len(messages), 1)
+        self.assertTrue('AWC1' in messages[0])
```
gcapalbo: :+1: great tests

corehq/apps/locations/dbaccessors.py
emord: `location` filters on the assigned locations, not primary location: https://github.com/dimagi/commcare-hq/blob/je/ls-indicator-num-2/corehq/apps/es/users.py#L112-L131

I was trying to not rely too much on icds's current structure and it seemed like this would be equivalent. I can write a separate dbaccessor for using primary location id though

custom/icds/messaging/indicators.py
emord: That would require way more queries though (20x I think?)

Curious why you prefer couch vs es? If we have inconsistencies (especially for things that are mostly static like users and locations) I think that would be a much larger issue.

corehq/apps/locations/dbaccessors.py
gcapalbo: hmm, ok. yeah i'm not sure we want to open it up to all assigned locations because they also won't necessarily be in the list of supervisor child locations so we might not have a name for them. i think it's fine to only use primary location

custom/icds/messaging/indicators.py
gcapalbo: just seemed like the kind of thing that we should get from the primary data source rather than a secondary data source. we can probably get it down to one query if we remove user id from the list of emitted values [here](https://github.com/dimagi/commcare-hq/blob/master/corehq/couchapps/users_extra/views/users_by_location_id/map.js). we could probably remove user id there since it's not giving us any extra information, and then we could query that couch view with a keys parameter

corehq/apps/locations/dbaccessors.py
emord: Ah. I see how this was confusing before. The User ES mock was incorrect. I fixed that https://github.com/dimagi/commcare-hq/pull/15018/commits/7ae622b4e116868cd71c74f32a047f4ccdc4d3e3

and changed this to use primary location https://github.com/dimagi/commcare-hq/pull/15018/commits/d85e72200d3fbba333412c14ef23f0d693339a55

custom/icds/tests/test_ls_vhnd_survey.py
emord: good idea https://github.com/dimagi/commcare-hq/pull/15018/commits/dbcd7d809c492b4ca4a2888f05863bc587debb83

custom/icds/messaging/indicators.py
gcapalbo: i see you already added primary location to es, that's fine for now, and fine to table this suggestion

custom/icds/tests/test_ls_vhnd_survey.py
```diff
@@ -0,0 +1,143 @@
...
+        UserESFake.save_doc(user._doc)
+        return user
+
+    @property
+    def today(self):
+        tz = pytz.timezone('Asia/Kolkata')
+        return datetime.now(tz=tz).date()
+
+    def _make_form(self, user_id, date):
+        def es_formatted_date(date):
```
gcapalbo: doesn't look like this ends up getting used?

corehq/apps/locations/dbaccessors.py
```diff
@@ -92,3 +92,45 @@ def user_ids_at_accessible_locations(domain_name, user):
...
+    for r in result:
+        if 'assigned_location_ids' in r:
+            locs = r['assigned_location_ids']
+            if not isinstance(locs, list):
+                locs = [r['assigned_location_ids']]
+            ret[r['_id']] = locs
+    return ret
+
+
+def primary_location_by_user_id(domain, location_ids):
```
gcapalbo: i was a little confused by the name of this function because it seems like it should take a user_id as a parameter. could we rename it to something like `get_user_ids_from_primary_location_ids`? same with the one above it too

gcapalbo (approved):
just a couple of nitpicky comments otherwise looks good

PR #15020, opened 2017-02-17
Include style for edit forms

benrudolph:
@wpride http://manage.dimagi.com/default.asp?247788 this doesn't address the full ticket, but this will handle the issue of unwanted "Sorry this question is required" showing up. I think there is some discrepancy between old edit forms and new edit forms (will offline with you). tested on both old and new backend

screen shot 2017-02-17 at 10 46 45 am

cc: @kaapstorm (also @dimagi/ux not sure if you have input, but this is drastically changing the look of edit forms to match web apps)


wpride (approved):

orangejenny:
> this is drastically changing the look of edit forms to match web apps

This sounds great - consistency is good, and web apps are better-looking than old edit forms.

PR #15022, opened 2017-02-17
record releases on sentry with additional info

snopoke:
This doesn't show up on the UI anywhere yet but is supposed to be in the pipeline


benrudolph (approved):
sweet

mkangia (approved):

PR #15016, opened 2017-02-16
Stick fixture index schema behind a feature flag

esoergel:
(The feature was initially added here: https://github.com/dimagi/commcare-hq/pull/14545/files)
@sheelio @NoahCarnahan
http://manage.dimagi.com/default.asp?247786

PR #15023, opened 2017-02-17
ensure fixture uniformity for all fields

czue:
I think this is the right fix for http://manage.dimagi.com/default.asp?247786

Includes a revert of dimagi/commcare-hq#15016

### DIFF
--- a/corehq/apps/locations/fixtures.py
+++ b/corehq/apps/locations/fixtures.py
@@ -2,6 +2,7 @@ from itertools import groupby
 from collections import defaultdict
 from xml.etree.ElementTree import Element
 from casexml.apps.phone.models import OTARestoreUser
+from corehq.apps.custom_data_fields import CustomDataFieldsDefinition
 from corehq.apps.locations.models import SQLLocation, LocationType, LocationFixtureConfiguration
 from corehq import toggles
 
@@ -78,12 +79,13 @@ class LocationFixtureProvider(object):
         if not should_sync_locations(last_sync, all_locations, restore_user):
             return []
 
-        return self.serializer.get_xml_nodes(self.id, restore_user, all_locations)
+        data_fields = _get_location_data_fields(restore_user.domain)
+        return self.serializer.get_xml_nodes(self.id, restore_user, all_locations, data_fields)
 
 
 class HierarchicalLocationSerializer(object):
 
-    def get_xml_nodes(self, fixture_id, restore_user, all_locations):
+    def get_xml_nodes(self, fixture_id, restore_user, all_locations, data_fields):
         if not should_sync_hierarchical_fixture(restore_user.project):
             return []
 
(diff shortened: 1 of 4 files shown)


czue:
@esoergel @sheelio I think this is the right fix for http://manage.dimagi.com/default.asp?247786. One possible concern is that this will include empty location data elements for all known fields in the domain for all locations in the restore. this could make the sync payload a bit bigger. I'm not too concerned about that but figured i'd flag.

czue:
(fixing tests)

corehq/apps/locations/tests/test_location_fixtures.py
```diff
@@ -258,6 +263,105 @@ def test_index_location_fixtures(self):
...
+            field for field in fixture.find('locations/location[@id="{}"]/location_data'.format(mass.location_id))
+        ]
+        self.assertEqual(3, len(mass_data))
+        self.assertEqual(self.field_slugs | set(['driver_friendliness']), set([f.tag for f in mass_data]))
+        self.assertEqual(
+            'poor',
+            fixture.find(
+                'locations/location[@id="{}"]/location_data/driver_friendliness'.format(mass.location_id)
+            ).text
+        )
```
esoergel: Wait, I'm confused, shouldn't the `driver_friendliness` field be either excluded here or included for every location?  It looks like it'll only be included on `Mass`.  FWIW, I think it's fine to force the project to add fields to the schema if they want them to appear.

esoergel (changes requested):
One blocking comment about non-schema custom data

corehq/apps/locations/fixtures.py
```diff
@@ -306,4 +312,12 @@ def _fill_in_location_element(xml_root, location):
         field_node.text = unicode(val if val is not None else '')
         xml_root.append(field_node)
 
-    xml_root.append(_get_metadata_node(location))
+    xml_root.append(_get_metadata_node(location, data_fields))
+
+
+def _get_location_data_fields(domain):
+    from corehq.apps.locations.views import LocationFieldsView
+    fields_definition = CustomDataFieldsDefinition.get_or_create(domain, LocationFieldsView.field_type)
```
czue: good call, will change.

corehq/apps/locations/tests/test_location_fixtures.py
czue: i was debating this. so you're saying you think we should just not include them? that makes sense to me, was just worried that maybe some apps might break, but that seems low probability.

esoergel: Yeah, I think we should skip that data.  I'd be concerned that these apps might break anyways if the fixture is heterogeneous because of non-schema data.

czue:
@esoergel responded to feedback

esoergel (approved):
Thanks!

czue:
thanks for fixing tests

PR #15025, opened 2017-02-17
delete system forms when deleting cases

snopoke:


corehq/apps/users/tasks.py
```diff
@@ -73,6 +73,16 @@ def tag_forms_as_deleted_rebuild_associated_cases(user_id, domain, form_id_list,
 
 
 @task(queue='background_queue', ignore_result=True, acks_late=True)
+def tag_system_forms_as_deleted(domain, deleted_forms, deleted_cases, deletion_id, deletion_date):
+    form_ids_to_delete = set()
+    for case_id in deleted_cases:
+        xform_ids = CaseAccessors(domain).get_case_xform_ids(case_id)
+        form_ids_to_delete |= set(xform_ids) - deleted_forms
+
+    FormAccessors(domain).soft_delete_forms(list(form_ids_to_delete), deletion_date, deletion_id)
```
snopoke: Hmm, just thinking about this again I think it might be necessary to check if these are all system forms. They could be forms submitted by another user (if the case was re-assigned to this user).

corehq/apps/users/tasks.py
```diff
@@ -73,6 +73,33 @@ def tag_forms_as_deleted_rebuild_associated_cases(user_id, domain, form_id_list,
...
+        if form.domain != domain:
+            return False
+
+        case_ids = get_case_ids_from_form(form)
+        cases_touched_by_form_not_deleted = case_ids - deleted_cases
+        for case in CaseAccessors(domain).iter_cases(cases_touched_by_form_not_deleted):
+            if not case.is_deleted:
+                return False
+
+        # all cases touched by this form are deleted
```
benrudolph: 👍

corehq/apps/users/tests/retire.py
```diff
@@ -197,3 +199,72 @@ def test_multiple_case_blocks_some_deleted(self, rebuild_case):
 
         self.assertEqual(rebuild_case.call_count, len(case_ids) - 1)
         self.assertItemsEqual(rebuild_case.call_args_list, expected_call_args)
+
+    @run_with_all_backends
+    def test_all_case_forms_deleted(self):
+        domain = create_domain(self.domain)
+        self.addCleanup(domain.delete)
```
benrudolph: nit: `create_domain` and `domain.delete` are brutal. would be better to have this in a setupclass if possible (and this will be called twice)

benrudolph (approved):

corehq/apps/users/tests/retire.py
snopoke: ed5bf05

benrudolph (approved):

PR #15026, opened 2017-02-17
Move debugger out of drag scroll dominion

benrudolph:
@biyeun this moves the debugger out of the dominion of the dragscroll so you can select things again. I tried `nochilddrag` with no luck. I still find that library terribly uncooperative

http://manage.dimagi.com/default.asp?247674

cc: @kaapstorm


benrudolph:
✅

kaapstorm:
lgtm, but will defer to @biyeun

PR #15027, opened 2017-02-17
Allow changing of language for app preview

benrudolph:
@wpride fun friday! multiple languages in app preview. super easy to do actually (good job on making formplayer respect the locale parameter)

language_demo

(it may not look like it's change the language but "Juste Leaga" is french for Justice League)

cc: @kaapstorm


wpride:
That is fun!

wpride (approved):

PR #15028, opened 2017-02-17
Session Replaying in Live Preview

wpride:
Let's you re-run your previous session after updating to latest build

This is a remake of https://github.com/dimagi/commcare-hq/pull/14004

@benrudolph @biyeun do you remember what the firefox issue was there?

cross-request: https://github.com/dimagi/formplayer/pull/295


corehq/apps/cloudcare/static/cloudcare/js/formplayer/app.js
```diff
@@ -124,6 +124,10 @@ FormplayerFrontend.on('startForm', function (data) {
     FormplayerFrontend.request("clearMenu");
     FormplayerFrontend.Menus.Util.showBreadcrumbs(data.breadcrumbs);
 
+    var urlObject = Util.currentUrlToObject();
+    urlObject.setSessionId(data.session_id);
+    Util.setUrlToObject(urlObject);
```
benrudolph: I think this caused a bug last time it was introduced... struggling to remember what it was

benrudolph (commented):
My main issue with this is that the history gets erased when you reload your session so it totally negates the back button. this is a fairly confusing user experience (at least it has been for me)

corehq/apps/cloudcare/static/cloudcare/js/formplayer/app.js
biyeun: i think it had something to do with one question per screen. i vaguely remember this too

wpride:
@benrudolph OK think I fixed all the issues except one. When you update and first return to the question, the answer doesn't populate immediately - only once you navigate forward and back does it reappear. Guessing there's some event we just need to publish

benrudolph:
sounds likely, let's offline

PR #15034, opened 2017-02-17
use add_arguments

nickpell:
@emord cc: @sravfeyn

### DIFF
--- a/corehq/apps/style/management/commands/update_manifest.py
+++ b/corehq/apps/style/management/commands/update_manifest.py
@@ -24,6 +24,7 @@ class Command(BaseCommand):
     def add_arguments(self, parser):
         parser.add_argument(
             'params',
+            choices=['save', 'soft'],
             nargs='*',
         )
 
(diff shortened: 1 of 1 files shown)


corehq/apps/style/management/commands/update_manifest.py
```diff
@@ -18,10 +18,15 @@ class ResourceCompressError(Exception):
 
 class Command(BaseCommand):
     help = "Prints the paths of all the static files"
-    args = "save or soft"
 
     root_dir = settings.FILEPATH
 
+    def add_arguments(self, parser):
+        parser.add_argument(
+            'params',
```
emord: Is there a way to define this or that in the parser? Or could a help text be added here to say the only valid things are save or soft?

nickpell: yeah, I can add ```choices```

nickpell: updated

emord (approved):

PR #15036, opened 2017-02-17
Abt custom reports: Update Tanzania targets

NoahCarnahan:
@esoergel


esoergel (approved):

esoergel:
Looks like you're running tests, is that intentional?

NoahCarnahan:
nope, just canceled them.

PR #15040, opened 2017-02-17
Force translated strings to be strings

orangejenny:
http://manage.dimagi.com/default.asp?247798

This causes a build error because the value `3.0` is interpreted as a decimal:
screen shot 2017-02-17 at 4 27 25 pm

Noticed this also affects graphing configuration (setting a series name to `1.0` will break the app), decided fixing it in `trans` was safe and more straightforward than figuring out what other app config might behave this way and fixing each area.

@gcapalbo

### DIFF
--- a/corehq/apps/app_manager/templatetags/xforms_extras.py
+++ b/corehq/apps/app_manager/templatetags/xforms_extras.py
@@ -31,11 +31,11 @@ def trans(name, langs=None, include_lang=True, use_delim=True, prefix=False):
     for lang in langs:
         if lang in name and name[lang]:
             affix = ("" if langs and lang == langs[0] else tag(lang))
-            return affix + str(name[lang]) if prefix else str(name[lang]) + affix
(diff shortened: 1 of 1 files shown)


corehq/apps/app_manager/templatetags/xforms_extras.py
```diff
@@ -31,11 +31,11 @@ def trans(name, langs=None, include_lang=True, use_delim=True, prefix=False):
     for lang in langs:
         if lang in name and name[lang]:
             affix = ("" if langs and lang == langs[0] else tag(lang))
-            return affix + name[lang] if prefix else name[lang] + affix
+            return affix + str(name[lang]) if prefix else str(name[lang]) + affix
```
gcapalbo: just curious if there might ever be unicode values here we need to worry about, in which case str() would cause an error, but unicode() would work

orangejenny: Excellent point. Looks like tests broke for the same reason. Fixed.

gcapalbo (approved):

PR #15041, opened 2017-02-17
Better handle overflow in case history

orangejenny:
http://manage.dimagi.com/default.asp?247778

This will wrap long URLs onto multiple lines. Experimented with having the table cell scroll, or having the whole table scroll, but both of those were a bit ugly and annoying to interact with.

screen shot 2017-02-17 at 5 19 49 pm

@gcapalbo

### DIFF
--- a/corehq/apps/style/static/style/less/_hq/utils.less
+++ b/corehq/apps/style/static/style/less/_hq/utils.less
@@ -16,13 +16,6 @@
   display: none;
 }
 
-.overflow-container {
-    position: relative;
-
-    .overflow-content {
-        position: absolute;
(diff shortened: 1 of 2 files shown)


gcapalbo (approved):
nice

corehq/apps/style/static/style/less/_hq/utils.less
```diff
@@ -15,3 +15,7 @@
 .ko-template {
   display: none;
 }
+
+.break-all-words {
+    word-break: break-all;
```
benrudolph: very cool, didn't know about this css property

PR #15043, opened 2017-02-17
Warn when saving to case property with a slash

orangejenny:
Context in http://manage.dimagi.com/default.asp?247753

Only added to basic modules because it's not very intelligent validation, it's just mean to warn users who don't realize that `/` is special.

screen shot 2017-02-17 at 6 43 34 pm

@gcapalbo / @snopoke / @czue

### DIFF
--- a/corehq/apps/app_manager/views/forms.py
+++ b/corehq/apps/app_manager/views/forms.py
@@ -551,7 +551,7 @@ def get_form_view_context_and_template(request, domain, form, langs, messages=me
                       and not toggles.USER_TESTING_SIMPLIFY.enabled(request.domain))
     valid_index_names = DEFAULT_CASE_INDEX_IDENTIFIERS.values()
     if allow_usercase:
-        valid_index_names.append(USERCASE_PREFIX[0:-1]) # strip trailing slash
+        valid_index_names.append(USERCASE_PREFIX[0:-1])     # strip trailing slash
 
     form_has_schedule = isinstance(form, AdvancedForm) and form.get_module().has_schedule
     context = {
--- a/corehq/apps/app_manager/views/forms.py
+++ b/corehq/apps/app_manager/views/forms.py
@@ -26,6 +26,7 @@ from corehq.apps.app_manager.views.schedules import get_schedule_context
 from corehq.apps.app_manager.views.utils import back_to_main, \
     CASE_TYPE_CONFLICT_MSG, get_langs
 
+from casexml.apps.case.const import DEFAULT_CASE_INDEX_IDENTIFIERS
 from corehq import toggles, privileges
 from corehq.apps.accounting.utils import domain_has_privilege
 from corehq.apps.app_manager.exceptions import (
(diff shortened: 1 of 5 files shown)


czue (approved):
nice looks good - I forgot about the user case as well. had one comment which you can respond to and merge

corehq/apps/app_manager/views/forms.py
```diff
@@ -546,12 +547,19 @@ def get_form_view_context_and_template(request, domain, form, langs, messages=me
         form.get_unique_id()
         app.save()
 
+    allow_usercase = (domain_has_privilege(request.domain, privileges.USER_CASE)
+                      and not toggles.USER_TESTING_SIMPLIFY.enabled(request.domain))
+    valid_index_names = DEFAULT_CASE_INDEX_IDENTIFIERS.values()
+    if allow_usercase and USERCASE_PREFIX[-1] == "/":
```
czue: I'm curious why this check is here? are you trying to future-proof it against someone removing the trailing `/`? if so the check seems like it might go around the truncation of the prefix, not the addition of the prefix to `valid_index_names`

corehq/apps/app_manager/views/forms.py
```diff
@@ -546,12 +547,19 @@ def get_form_view_context_and_template(request, domain, form, langs, messages=me
         form.get_unique_id()
         app.save()
 
+    allow_usercase = (domain_has_privilege(request.domain, privileges.USER_CASE)
```
gcapalbo: i don't have much context but am just curious if you also need to check if usercase is enabled for the project here, or if that doesn't matter

corehq/apps/app_manager/views/forms.py
orangejenny: The usercase prefix used to be `user:`, so I wanted to not add it to the prefixes if that was the case. Although, odds are good it was changed to `user/` to resemble `parent/` etc and isn't likely to change back... @kaapstorm do you think this check is worthwhile, or is `user/` set in stone at this point?

corehq/apps/app_manager/views/forms.py
orangejenny: This code is just being moved, so I think it's okay as-is.

corehq/apps/app_manager/views/forms.py
kaapstorm: I'm sure it's safe to set it in stone, if we hadn't felt that way already.

orangejenny: cool, thanks. updated.

corehq/apps/app_manager/views/forms.py
```diff
@@ -546,12 +547,19 @@ def get_form_view_context_and_template(request, domain, form, langs, messages=me
...
+                      and not toggles.USER_TESTING_SIMPLIFY.enabled(request.domain))
+    valid_index_names = DEFAULT_CASE_INDEX_IDENTIFIERS.values()
+    if allow_usercase:
+        valid_index_names.append(USERCASE_PREFIX[0:-1]) # strip trailing slash
```
dimagimon: E261 at least two spaces before inline comment

kaapstorm (approved):

PR #15046, opened 2017-02-19
last two sms indicators

emord:
@gcapalbo These are the last two indicators. The first several commits are the same as on my other PR. You only need to review the last 4 https://github.com/dimagi/commcare-hq/compare/49796bae7611263ff7e65a23a24a3c3324f9305b~1...bb864811e268263e4f0349df35ee8fd4c2dd1556

These tests aren't as comprehensive as the others since I mocked the report fixture. Otherwise I think I would have had to build up a lot more mocks for the app config, UCRs and form/case data

cc @millerdev @czue


custom/icds/messaging/indicators.py
```diff
@@ -121,6 +126,11 @@ def location_names_from_user_id(self, user_ids):
         return {self.awc_locations[l] for l in loc_ids}
 
 
+class AWWAggregatePerformanceIndicator(AWWIndicator):
+    def get_messages(self, language_code=None):
+        return LSAggregatePerformanceIndicator(self.domain, self.supervisor).get_messages(language_code)
```
gcapalbo: i'm not sure this is right - this will send to this one AWW the aggregated info for all AWCs under the LS right? we'd only want the AWW's info to be included here

gcapalbo:
would be good if @czue could take a pass to check we're not missing anything wrt UCR. (the commits to review start with https://github.com/dimagi/commcare-hq/pull/15046/commits/43a29ee5574e6dbadcea173ca06b1cda4e2bb2e5)

this is the thing i emailed you about running the mobile UCRs to include message content from them in SMS

custom/icds/messaging/indicators.py
```diff
@@ -227,3 +237,71 @@ def convert_to_date(date):
...
+        app = get_app(self.domain, SUPERVISOR_APP_ID)
+        return {
+            report_config.report_id: report_config
+            for module in app.modules if isinstance(module, ReportModule)
+            for report_config in module.report_configs
+            if report_config.report_id in report_ids
+        }
+
+    def get_report_fixture(self, report_id):
+        return ReportFixturesProvider._report_config_to_fixture(
```
czue: nit: might make sense to remove the `_` from the signature now that it's being called from other places

custom/icds/messaging/indicators.py
```diff
@@ -227,3 +237,71 @@ def convert_to_date(date):
...
+    def weighed_fixture(self):
+        return self.get_report_fixture(CHILDREN_WEIGHED_REPORT_ID)
+
+    @property
+    def days_open_fixture(self):
+        return self.get_report_fixture(DAYS_AWC_OPEN_REPORT_ID)
+
+    def get_value_from_fixture(self, fixture, attribute):
+        xpath = './rows/row[@is_total_row="True"]/column[@id="{}"]'.format(attribute)
+        return fixture.findall(xpath)[0].text
```
czue: I'm curious whether it was an intentional design decision to work directly off the fixture XML as opposed to using the UCR data source and working with the results directly? That would be more "direct" though I also see value in preserving the abstraction barrier. Also simplifies the tests quite a bit.

czue (commented):
Reviewed the UCR stuff and looks good to me. Just had one nitpicky comment and a question that is not actionable.

custom/icds/messaging/indicators.py
emord: I briefly thought about that, but using the mobile ucr allowed me to pass in the user and not worry about constructing the filters, which seemed like it would be way easier. It also means that changes in the app will be reflected immediately in the sms we send, which could be a feature or a liability depending on what we want.

millerdev:
didn't review since others seem to have covered it already.

custom/icds/messaging/indicators.py
emord: Thanks. I initially misunderstood the spec. I filtered those here https://github.com/dimagi/commcare-hq/pull/15046/commits/aaafd1e844d9b10ce907d272ec361b2e3b6cc4f0

custom/icds/messaging/indicators.py
```diff
@@ -224,3 +273,72 @@ def convert_to_date(date):
...
+        report_ids = [
+            HOME_VISIT_REPORT_ID,
+            THR_REPORT_ID,
+            CHILDREN_WEIGHED_REPORT_ID,
+            DAYS_AWC_OPEN_REPORT_ID,
+        ]
+        app = get_app(self.domain, SUPERVISOR_APP_ID)
+        return {
+            report_config.report_id: report_config
+            for module in app.modules if isinstance(module, ReportModule)
```
dimagimon: F821 undefined name 'ReportModule'

custom/icds/tests/test_ls_aggregate_performance.py
```diff
@@ -0,0 +1,112 @@
+from datetime import datetime, timedelta
```
dimagimon: F401 'datetime' imported but unused
F401 'timedelta' imported but unused

custom/icds/tests/test_ls_aggregate_performance.py
```diff
@@ -0,0 +1,112 @@
...
+from corehq.apps.domain.shortcuts import create_domain
+from corehq.apps.locations.tests.util import (
+    LocationStructure,
+    LocationTypeStructure,
+    setup_location_types_with_structure,
+    setup_locations_with_structure,
+)
+from corehq.apps.users.models import CommCareUser
+
+from custom.icds.const import (
```
dimagimon: F401 'CHILDREN_WEIGHED_REPORT_ID' imported but unused
F401 'DAYS_AWC_OPEN_REPORT_ID' imported but unused
F401 'HOME_VISIT_REPORT_ID' imported but unused
F401 'THR_REPORT_ID' imported but unused

emord:
@gcapalbo I updated the indicators to match the revised spec

custom/icds/messaging/indicators.py
```diff
@@ -108,6 +127,37 @@ def location_names_from_user_id(self, user_ids):
...
 
+class AWWAggregatePerformanceIndicator(AWWIndicator):
+    template = 'ls_aggregate_performance.txt'
+
+    def get_value_from_fixture(self, fixture, attribute):
+        xpath = './rows/row[@is_total_row="False"]'
+        rows = fixture.findall(xpath)
+        for row in rows:
+            owner_id = row.find('./column[@id="owner_id"]')
+            if owner_id.text == self.user.sql_location.name:
```
gcapalbo: is there a way to match on location id instead of location name?

emord: Not if we want to use mobile ucr to handle the filtering and generation.

the column says `owner_id` but it's actually transformed into a name. ex https://github.com/dimagi/commcare-hq/blob/master/custom/icds_reports/ucr/reports/ls_awc_days_open.json#L42-L56

We could possibly change the report to include two columns, but I don't know how feasible that is

custom/icds/messaging/indicators.py
gcapalbo: would this fail hard if the attribute doesn't exist? i think we'd want it to

custom/icds/messaging/indicators.py
```diff
@@ -108,6 +127,37 @@ def location_names_from_user_id(self, user_ids):
...
+                return row.find('./column[@id="{}"]'.format(attribute)).text
```
gcapalbo: same question about failing hard

custom/icds/messaging/indicators.py
```diff
@@ -224,3 +274,72 @@ def convert_to_date(date):
...
+        thr_count = self.get_value_from_fixture(self.thr_fixture, 'open_count')
+        num_weigh = self.get_value_from_fixture(self.weighed_fixture, 'open_weighed')
+        num_weigh_avail = self.get_value_from_fixture(self.weighed_fixture, 'open_count')
+        num_days_open = self.get_value_from_fixture(self.days_open_fixture, 'awc_opened_count')
+
+        context = {
+            "visits": "{} / {}".format(visit_on_time, visits),
+            "thr_distribution": "{} / {}".format(thr_gte_21, thr_count),
+            "children_weighed": "{} / {}".format(num_weigh, num_weigh_avail),
+            "days_open": "{}".format(int(num_days_open) // len(self.awc_locations)),
```
gcapalbo: would it be possible to use the number of rows in the report for this? they should be the same but i think the number of rows in the report would be the more correct way to calculate the average

gcapalbo: also they wanted this rounded to the nearest whole number so i think it would have to be:

`int(round(1.0 * x / y))`

custom/icds/messaging/indicators.py
gcapalbo: i think we'd also want it to fail hard if the AWC is not found here

custom/icds/messaging/indicators.py
```diff
@@ -108,6 +127,37 @@ def location_names_from_user_id(self, user_ids):
         return {self.awc_locations[l] for l in loc_ids}
 
 
+class AWWAggregatePerformanceIndicator(AWWIndicator):
+    template = 'ls_aggregate_performance.txt'
```
gcapalbo: we'll need a different template for the aww performance report since some of the headings are different (like timely visits vs. home visits)

custom/icds/messaging/indicators.py
gcapalbo: i see, i think if that's going to be difficult it's probably also fine if we just fail hard if we can't find it. right now it would return None if none match

custom/icds/messaging/indicators.py
```diff
@@ -224,3 +286,73 @@ def convert_to_date(date):
...
+        num_weigh = self.get_value_from_fixture(self.weighed_fixture, 'open_weighed')
+        num_weigh_avail = self.get_value_from_fixture(self.weighed_fixture, 'open_count')
+        num_days_open = int(self.get_value_from_fixture(self.days_open_fixture, 'awc_opened_count'))
+        num_awc_locations = len(self.days_open_fixture.findall('./rows/row[@is_total_row="False"]'))
+
+        context = {
+            "visits": "{} / {}".format(visit_on_time, visits),
+            "thr_distribution": "{} / {}".format(thr_gte_21, thr_count),
+            "children_weighed": "{} / {}".format(num_weigh, num_weigh_avail),
+            "days_open": "{}".format(int(round(1.0 * num_days_open / num_awc_locations))),
```
gcapalbo: we should check if `num_awc_locations` is 0 and just display 0 here when that happens

custom/icds/messaging/indicators.py
```diff
@@ -224,3 +286,73 @@ def convert_to_date(date):
...
+
+    @property
+    def weighed_fixture(self):
+        return self.get_report_fixture(CHILDREN_WEIGHED_REPORT_ID)
+
+    @property
+    def days_open_fixture(self):
+        return self.get_report_fixture(DAYS_AWC_OPEN_REPORT_ID)
+
+    def get_value_from_fixture(self, fixture, attribute):
```
gcapalbo: can we add the same IndicatorError exception raising here as you did for the AWW one?

gcapalbo:
looks great @emord just two small comments

gcapalbo (approved):

PR #15048, opened 2017-02-20
Use class to allow text entry in case claim in live preview

wpride:
Fixes http://manage.dimagi.com/default.asp?247698


benrudolph (approved):

benrudolph:
notests merging

PR #15049, opened 2017-02-20
performance sms tweaks

czue:
some feedback from demoing/explaining this to someone. including sorting groups, fixing bugs, and allowing customizing the schedule in advanced mode.

fixes the root cause of http://manage.dimagi.com/default.asp?247833 and also makes it much harder to create invalid configurations.

probably easiest read by commit.

@gcapalbo / @millerdev

### DIFF
--- a/corehq/apps/performance_sms/forms.py
+++ b/corehq/apps/performance_sms/forms.py
@@ -26,7 +26,9 @@ class PerformanceFormMixin(object):
 
     def clean(self):
         cleaned_data = super(PerformanceFormMixin, self).clean()
-        print 'calling clean()'
+        if self.errors:
+            # don't bother processing if there are already errors
+            return cleaned_data
         config = PerformanceConfiguration.wrap(self.config.to_json())
         self._apply_updates_to_config(config, cleaned_data)
         try:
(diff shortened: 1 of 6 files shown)


corehq/apps/performance_sms/forms.py
```diff
@@ -166,6 +166,9 @@ def __init__(self, domain, config, *args, **kwargs):
             *form_layout
         )
 
+    def clean_schedule(self):
+        return ScheduleConfiguration.wrap(self.cleaned_data['schedule'])
```
millerdev: Ignore me if I'm way off, but it seems that the `schedule` field in _advanced mode_ is a different format and has a different meaning than `schedule` in _simple mode_: in _advanced_ it's a dict, while in _simple_ it's a string (I think). This makes it harder to reason about the data in an abstract way and could be a source of subtle bugs if the two ever get confused.

Suggestion: the _simple mode_ field should be named something like `schedule_interval` while the _advanced mode_ field can keep the name `schedule`.

millerdev (approved):
Approval conditional on whether my comment deserves attention or not.

corehq/apps/performance_sms/forms.py
czue: nope, that's right and a good suggestion. https://github.com/dimagi/commcare-hq/pull/15049/commits/ac71e947ccd2a84a4614c620ef685eb09743c123

gcapalbo (approved):

PR #15052, opened 2017-02-20
DHIS2 API

kaapstorm:
Classes for managing DHIS2 API calls, and tests.

(Tests are going to fail without a DHIS2 test server, hence the `[ci skip]` commit message. Also, the server needs to be for testing only, because once DataValues have been added you can't completely tear down test data -- DHIS2 creates DataValueAudits that can only be deleted directly in PostgreSQL, so test organisation units, etc. can never be removed.)

@benrudolph cc @sheelio @snopoke


kaapstorm:
@sheelio do we have a DHIS2 server we can use for testing? These tests create an org unit and add dummy data to it which we won't be able to delete, so maybe not great for a server we want to use for other stuff. Otherwise I can mark these tests to be skipped, and just test locally.

benrudolph (commented):
is there anyway we can make this not dependent on a separate server? maybe by mocking out some of the responses?

kaapstorm:
Good call @benrudolph I'll do that.

kaapstorm:
To do:
 - [ ] Mock DHIS2 responses

kaapstorm:
Closing this to PR off master

PR #15056, opened 2017-02-20
restrict archiving form that creates cases if case has other forms

snopoke:
Prevent archiving a form if it creates any cases which have other forms updating them. To put it another way, you can only archive a form which creates a case if that form is the only form which touches the case.

The reason for this change is to avoid the scenario where a case exists but is missing the form that created it which results in some confusing (and incorrect) data.

### DIFF
--- a/corehq/apps/reports/views.py
+++ b/corehq/apps/reports/views.py
@@ -55,7 +55,7 @@ from casexml.apps.case.cleanup import rebuild_case_from_forms, close_case
 from casexml.apps.case.dbaccessors import get_open_case_ids_in_domain
 from casexml.apps.case.models import CommCareCase
 from casexml.apps.case.templatetags.case_tags import case_inline_display
-from casexml.apps.case.xform import extract_case_blocks
+from casexml.apps.case.xform import extract_case_blocks, get_case_updates
 from casexml.apps.case.xml import V2
 from casexml.apps.stock.models import StockTransaction
 from couchdbkit.exceptions import ResourceNotFound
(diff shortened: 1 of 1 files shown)


benrudolph (approved):
so this eliminates the ability to erase the initial changes to a case that a form makes. the reason we're doing this is that it's somewhat weird to have the second "update" form create the case. this was an edge case during your migrations right? seems fine to me, wonder if we'll get any complaints

czue:
I agree this makes sense and that it's better to ask for forgiveness than permission if it comes up (and the answer likely being "sorry, this was a bug that you could ever do it which we've fixed"). Unless someone makes a very compelling case I can't see us losing that debate.

Probably good to let flos know about the change in case anyone gets confused or reports a bug.

corehq/apps/reports/views.py
```diff
@@ -1859,26 +1868,60 @@ def archive_form(request, domain, instance_id):
...
+
+def _get_cases_with_forms_message(domain, cases_with_other_forms, case_id_from_request):
+    def _get_case_link(case_id, name):
+        if case_id == case_id_from_request:
+            return _(u"%(case_name)s (this case)") % {'case_name': name}
+        else:
+            return u'<a href="{}#!history">{}</a>'.format(reverse('case_details', args=[domain, case_id]), name)
+
+    case_links = ', '.join([
+       _get_case_link(case_id, name)
```
dimagimon: E121 continuation line under-indented for hanging indent

snopoke:
:white_check_mark:

PR #15068, opened 2017-02-21
Grammar and typo

kaapstorm:
... because "log" is a verb, and "in", "into" and "out" are prepositions. 

(A "login" is a noun, and it means "user account" or "session". (Having an English teacher as a mom is something that you can never recover from. :grimacing: ))

@benrudolph


benrudolph (approved):
haha

orangejenny:
❤️

PR #15070, opened 2017-02-21
update admin url patterns

nickpell:
removes warning:

```
RemovedInDjango20Warning: Passing a 3-tuple to django.conf.urls.include() is deprecated. Pass a 2-tuple containing the list of patterns and app_name, and provide the namespace argument to include() instead.
  url(r'^admin/', include(admin.site.urls)),
```

@sravfeyn


nickpell:
✅

PR #15075, opened 2017-02-21
cleaning up accounting tests

nickpell:
Some first steps.  One of the next things I'd like to do is stop creating a new domain every time we create a ```WebUser``` for accounting tests.


benrudolph (approved):
nice 🎸 ⭐️

PR #15077, opened 2017-02-21
Migration progress and blob db migration command updates

millerdev:
- Improve long-running, resumeable migration status output.
- Convert `run_blob_migration` management command args to new format so the command can be run on staging (newer version of Django).

Best to review commits individually 🐠 

@czue @emord @snopoke

### DIFF
--- a/corehq/blobs/management/commands/run_blob_migration.py
+++ b/corehq/blobs/management/commands/run_blob_migration.py
@@ -22,13 +22,24 @@ class Command(BaseCommand):
     Example: ./manage.py run_blob_migration [options] saved_exports
     """
     help = USAGE
-    option_list = (
-        make_option('--log-dir', help="Migration log directory."),
-        make_option('--reset', action="store_true", default=False,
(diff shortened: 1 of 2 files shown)


corehq/blobs/management/commands/run_blob_migration.py
```diff
@@ -22,13 +22,29 @@ class Command(BaseCommand):
...
-        make_option('--chunk-size', type="int", default=100,
-            help="Maximum number of records to read from couch at once."),
-    )
+
+    def add_arguments(self, parser):
+        parser.add_argument(
+            'slug',
+            choices=sorted(MIGRATIONS),
+            help="Migration slug: {}".format(', '.join(sorted(MIGRATIONS))),
+        )
```
millerdev: @nickpell: I cherry-picked 47b4d09a06ed8f0558230ea7b0a0229865bac9c3 and added the slug parameter. Hope that doesn't cause problems for your Django 1.10 branch, but let me know if it does.

PR #15078, opened 2017-02-21
Je/print statements

emord:
@nickpell @millerdev 

ran modernize to fix print statements and exception catching.

### DIFF
--- a/custom/ilsgateway/views.py
+++ b/custom/ilsgateway/views.py
@@ -252,7 +252,7 @@ def end_report_run(request, domain):
         rr.complete = True
         rr.has_error = True
         rr.save()
-    except ReportRun.DoesNotExist, ReportRun.MultipleObjectsReturned:
+    except (ReportRun.DoesNotExist, ReportRun.MultipleObjectsReturned) as e:
         pass
     return HttpResponseRedirect(reverse(ILSConfigView.urlname, kwargs={'domain': domain}))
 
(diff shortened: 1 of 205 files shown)


millerdev (approved):
Guessing you want to get this merged ASAP to avoid conflicts. I did not review every line, but it's pretty automatic and I think not much can go wrong with these particular conversions. Assuming you looked at every line that was changed to verify that it made sense.

emord:
@millerdev yeah I did a quick check on each line by doing `git add -p .` I did catch one error, but everything else looked ok

nickpell:
@emord curious, what was the error you found?

custom/ilsgateway/views.py
```diff
@@ -252,7 +252,7 @@ def end_report_run(request, domain):
         rr.complete = True
         rr.has_error = True
         rr.save()
-    except ReportRun.DoesNotExist, ReportRun.MultipleObjectsReturned:
```
emord: @nickpell this line had the error. Whenever an exception tries to catch two things like this, but doesn't have a variable assocaited with it, it is changed to:

```
except ReportRun.DoesNotExist as xxx_todo_changeme:
    ReportRun.MultipleObjectsReturned = xxx_todo_changeme
```

nickpell: ah ok.  I've since this one before.

PR #15081, opened 2017-02-21
Allow advanced case list filter hacks

orangejenny:
http://manage.dimagi.com/default.asp?247816

@gcapalbo / @snopoke


gcapalbo (approved):

czue:
would it be hard to add a test for this?

orangejenny:
added test

PR #15082, opened 2017-02-21
App manager v2: new forms in case lists should update a case

orangejenny:
@biyeun


orangejenny:
HTML-only change, killing build

benrudolph:
couldn't html changes like these affect the app manager diff test?

orangejenny:
yeah, fair point. i was in haste to get this in before subsequent user tests (and did run `build_app_manager_v2_diffs`), but you're right, shouldn't have canceled these tests.

benrudolph:
ok makes sense 👍 (fwiw i think it's fine to merge if needed to get in for something urgent especially when you ran the test locally, i was more concerned about confusion over when it's ok to merge something without tests running)

PR #15088, opened 2017-02-21
make filters async

emord:
http://manage.dimagi.com/default.asp?247762#1291495

I'm not positive that this will fix it. My hunch is that this report is choking because it tries to load every user and device id in the report at once and making it use async select2 would fix that. I tried to deploy to staging, but kept on failing multiple times, so putting it up for PR anyways

@benrudolph @gcapalbo


corehq/apps/reports/filters/api.py
```diff
@@ -247,3 +249,30 @@ def paginate_options(data_sources, query, start, size):
...
         options.extend(objects)
     return total, options
+
+
+class DeviceLogFilter(LoginAndDomainMixin, JSONResponseMixin, View):
+    field = None
+
+    def get(self, request, domain):
+        q = self.request.GET.get('q', None)
+        field_filter = {self.field + "__startswith": q}
```
emord: Couldn't think of anything more performant that would still look through the device entry log as those aren't in ES. Could use full text search if we had jdjango 1.10 (https://docs.djangoproject.com/en/1.10/ref/contrib/postgres/search/)

corehq/apps/reports/filters/api.py
```diff
@@ -247,3 +249,30 @@ def paginate_options(data_sources, query, start, size):
...
+        values = (
+            DeviceReportEntry.objects
+            .filter(domain=domain)
+            .filter(**field_filter)
+            .distinct(self.field)
+            .values_list(self.field, flat=True)
+        )[:10]
```
benrudolph: django doesn't smartly add `limit 10` right? specifying that could help with the number of objects loaded into memory

corehq/apps/reports/filters/api.py
benrudolph: interesting, we could also use a raw sql query if we need to

corehq/apps/reports/filters/api.py
```diff
@@ -247,3 +249,30 @@ def paginate_options(data_sources, query, start, size):
...
+        return self.render_json_response({
+            'results': [{'id': v, 'text': v} for v in values],
+            'total': len(values)
```
benrudolph: is this supposed to be the total values of the entire query? otherwise this number doesn't make that much sense since you could just derive this value from the `results` field

benrudolph (commented):
cool, i'm going to try and deploy staging today, so hopefully it will be deployed. had a couple of questions, but other than that looks good

corehq/apps/reports/filters/api.py
emord:
```
In [11]: print DeviceReportEntry.objects.filter(domain='dimagi', username__startswith='a').distinct('username').values_list('username', flat=True)[:10].query
SELECT DISTINCT ON ("phonelog_devicereportentry"."username") "phonelog_devicereportentry"."username" FROM "phonelog_devicereportentry" WHERE ("phonelog_devicereportentry"."username"::text LIKE a% AND "phonelog_devicereportentry"."domain" = dimagi) LIMIT 10
```

it does add limit 10 which

benrudolph: nice that's good to know 👍

corehq/apps/reports/filters/api.py
emord: It's not needed. I just copied it from one of the other async filters. Will remove when I fix these tests

corehq/apps/reports/filters/api.py
emord: raw sql makes me shudder especially when its not absolutely necessary

benrudolph: agreed

benrudolph (approved):

PR #15090, opened 2017-02-22
Update uninstall-requirements.txt

emord:
@nickpell did you run into issues here https://github.com/dimagi/commcare-hq/pull/14608/files ? Causes a bit of extra work on tests because it uninstalls and reinstalls nose every time


nickpell (approved):
I had added this because it didn't look like it was updating the requirement (it was a github commit) but should be safe to remove now that travis is updated.

PR #15091, opened 2017-02-22
Try to reduce DocumentConflict errors

benrudolph:
@gcapalbo what do you think about this for reducing conflict errors?

cc: @emord

### DIFF
--- a/corehq/apps/export/utils.py
+++ b/corehq/apps/export/utils.py
@@ -3,13 +3,14 @@ from datetime import datetime
 
 from dimagi.utils.couch.undo import DELETED_SUFFIX
 from dimagi.utils.modules import to_function
+from dimagi.utils.couch import CriticalSection
 from toggle.shortcuts import set_toggle
 
 from corehq.toggles import OLD_EXPORTS, NAMESPACE_DOMAIN, ALLOW_USER_DEFINED_EXPORT_COLUMNS
 from corehq.util.log import with_progress_bar
 from corehq.apps.hqwebapp.templatetags.hq_shared_tags import toggle_js_domain_cachebuster
 from corehq.apps.reports.dbaccessors import (
-    stale_get_exports_json,
+    get_exports_json,
     stale_get_export_count,
 )
 from corehq.apps.reports.models import (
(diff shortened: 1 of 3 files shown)


corehq/apps/reports/dbaccessors.py
```diff
@@ -23,6 +23,13 @@ def _get_exports(domain, include_docs=True, reduce=False, **kwargs):
     )
 
 
+def get_exports_json(domain):
```
benrudolph: literally the same as the function below except without the stale argument. will remove when i'm done migrating

corehq/apps/reports/views.py
```diff
@@ -523,19 +524,21 @@ def _export_default_or_custom_data(request, domain, export_id=None, bulk_export=
 @require_form_export_permission
 @require_GET
 def hq_download_saved_export(req, domain, export_id):
-    saved_export = SavedBasicExport.get(export_id)
-    return _download_saved_export(req, domain, saved_export)
+    with CriticalSection(['saved-export-{}'.format(export_id)]):
+        saved_export = SavedBasicExport.get(export_id)
+        return _download_saved_export(req, domain, saved_export)
```
gcapalbo: was the issue before that `_download_saved_export` was saving the `saved_export` and that's where you were getting the doc update conflicts with?

corehq/apps/export/utils.py
```diff
@@ -596,21 +597,22 @@ def migrate_domain(domain, dryrun=False, force_convert_columns=False):
...
-            except Exception as e:
-                print('Failed parsing {}: {}'.format(old_export['_id'], e))
-                raise e
-            else:
-                metas.append(migration_meta)
+            with CriticalSection(['saved-export-{}'.format(old_export._id)], timeout=120):
+                try:
+                    _, migration_meta = convert_saved_export_to_export_instance(
+                        domain,
+                        SavedExportSchema.wrap(old_export),
```
gcapalbo: does this `old_export` get saved throughout this process? if so i think you still might have an issue since the entire get-and-save operation needs to be under the `CriticalSection`.  as it is now, `get_exports_json` is outside the `CriticalSection` so between the time that is called and this executes, something else could have saved the object. One way to fix would be to call `.get()` to get the latest object again under the `CriticalSection`

benrudolph: yeah i did think that still might be an issue, but thought this would cover most cases since the `last_accessed` thing happen very quickly and often. this would grab the lock and ensure none of those requests messed it up. agree though that it'd be more foolproof to call `.get()`. will change to that

corehq/apps/reports/views.py
benrudolph: that is my current theory yes

corehq/apps/export/utils.py
```diff
@@ -596,21 +597,22 @@ def migrate_domain(domain, dryrun=False, force_convert_columns=False):
...
-            except Exception as e:
-                print('Failed parsing {}: {}'.format(old_export['_id'], e))
-                raise e
-            else:
-                metas.append(migration_meta)
+            with CriticalSection(['saved-export-{}'.format(old_export._id)], timeout=120):
+                try:
+                    _, migration_meta = convert_saved_export_to_export_instance(
+                        domain,
+                        SavedExportSchema.get(old_export._id),
```
gcapalbo: should this be `old_export['_id']` (and in the `CriticalSection` reference too)? i noticed it's that way in the except block

benrudolph: sorry yes that was dumb thanks, realized i tested this on a domain that already had all of its exports migrated so none of them actually ran

benrudolph:
ok should be fixed

PR #15092, opened 2017-02-22
Export migration tweaks

benrudolph:
@emord @gcapalbo first commit improves trackback readability, second commit skips the first migration if we are forcing the conversion. before it would do a dryrun migration to ensure we didn't skip any columns, however this is not necessary when we force the column convert

### DIFF
--- a/corehq/apps/export/utils.py
+++ b/corehq/apps/export/utils.py
@@ -1,6 +1,5 @@
 from __future__ import print_function
 from datetime import datetime
-import sys
 
 from dimagi.utils.couch.undo import DELETED_SUFFIX
 from dimagi.utils.modules import to_function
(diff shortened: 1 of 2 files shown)


corehq/apps/export/utils.py
```diff
@@ -608,7 +609,8 @@ def migrate_domain(domain, dryrun=False, force_convert_columns=False):
                 )
             except Exception as e:
                 print('Failed parsing {}: {}'.format(old_export['_id'], e))
-                raise e
+                exc_type, exc_value, exc_traceback = sys.exc_info()
+                raise exc_type, exc_value, exc_traceback
```
gcapalbo: could just calling `raise` with no args work too to reraise with all the info?

benrudolph: ah i didn't know that i was getting incorrect tracebacks and this fixed it, will test out with no args

benrudolph: thanks that does work. updated: https://github.com/dimagi/commcare-hq/pull/15092/commits/f333c776c7436174899aac6477937860eb58fd71

gcapalbo (approved):

emord (approved):

PR #15097, opened 2017-02-22
Catch exceptions when generating schemas

benrudolph:
@gcapalbo @emord not in love with this solution, but when migrating exports, some app builds are borked and processing them generates an error. I don't think this actually fails to often, but just in case i've added a soft assert. This falls into the category of "nice to be notified but not always an actionable error"


gcapalbo (approved):
looks good. i don't think the build ran - maybe merge master?

PR #15102, opened 2017-02-22
fix management command

nickpell:
@emord cc: @esoergel @biyeun


emord (approved):

esoergel (approved):

esoergel:
Just a thought - do you need tests running on all of these?

emord:
@esoergel for what it's worth this managment command is actually tested https://github.com/dimagi/commcare-hq/blob/80d6885dd9885adac05c7b265d19f7f9d39e4efd/corehq/apps/cleanup/tests/test_fix_forms_and_apps_with_missing_xmlns.py#L167

but agreed most of these probably won't catch anything

nickpell:
This one has test coverage, and since it's fixing a bug (in django 1.10) I'd like to run tests on just this one.  But after this batch there isn't test coverage for the rest of the management commands, so I'll submit those changes in larger groups.

nickpell:
The reason I suspect most of the remaining management commands don't have coverage is that optparse didn't cause them to fail during tests on django 1.10

PR #15105, opened 2017-02-22
update cchq_prbac_bootstrap.py to use add_arguments

nickpell:
@esoergel @biyeun


esoergel (approved):
Cool, is Django finally migrating to argparse or something?

nickpell:
yep, django 1.10 no longer accepts optparse

PR #15108, opened 2017-02-22
use case insensitive search

emord:
@nickpell got frustrated I couldn't search for enterprise and had to use Enterprise


nickpell (approved):
brilliant

PR #15111, opened 2017-02-22
don't pass RequestContext to render

nickpell:
passing ```RequestContext``` instead of ```dict``` to ```render```  becomes problematic in django 1.10

testing on staging

### DIFF
--- a/corehq/apps/hqwebapp/views.py
+++ b/corehq/apps/hqwebapp/views.py
@@ -93,12 +93,14 @@ def server_error(request, template_name='500.html'):
     traceback_key = uuid.uuid4().hex
     cache.cache.set(traceback_key, traceback_text, 60*60)
 
-    return HttpResponseServerError(t.render(RequestContext(request,
-        {'MEDIA_URL': settings.MEDIA_URL,
-         'STATIC_URL': settings.STATIC_URL,
-         'domain': domain,
-         '500traceback': traceback_key,
-        })))
+    return HttpResponseServerError(t.render(
+        RequestContext(request, {
+            'MEDIA_URL': settings.MEDIA_URL,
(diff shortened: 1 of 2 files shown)


nickpell:
no problems observed on staging

nickpell:
@esoergel @biyeun

corehq/apps/hqwebapp/views.py
```diff
@@ -93,23 +93,27 @@ def server_error(request, template_name='500.html'):
...
-         '500traceback': traceback_key,
-        })))
+    return HttpResponseServerError(t.render(
+        RequestContext(request, {
+            'MEDIA_URL': settings.MEDIA_URL,
+            'STATIC_URL': settings.STATIC_URL,
+            'domain': domain,
+            '500traceback': traceback_key,
+        }).dicts[0]
+    ))
```
esoergel: I think the preferable way to do this would be to no longer bother assembling a `RequestContext` object - `render` takes an optional `request` argument which I believe does the same thing
https://docs.djangoproject.com/en/1.10/topics/templates/#django.template.backends.base.Template.render

```python
return HttpResponseServerError(t.render(
    {
        'MEDIA_URL': settings.MEDIA_URL,
        'STATIC_URL': settings.STATIC_URL,
        'domain': domain,
        '500traceback': traceback_key,
    },
    request
))
```

nickpell: updated, going to test on staging

nickpell:
✅

corehq/apps/hqwebapp/views.py
esoergel: Thanks, :champagne: !

esoergel (approved):

PR #15112, opened 2017-02-22
Update prelogin to master

NoahCarnahan:
@calellowitz


orangejenny:
ideally prelogin updates should wait on tests, since hq has tests that are affected by prelogin

PR #15113, opened 2017-02-22
Je/grandfather priv

emord:
http://manage.dimagi.com/default.asp?247697#1292273
@nickpell I think this is the correct workflow.

There is an old pro plan version that does not have access to restrict access to location, but the new version does. "Grandfathering" it in this way should give all old versions access to it

buddies @benrudolph @gcapalbo


corehq/apps/hqadmin/management/commands/cchq_prbac_grandfather_privs.py
```diff
@@ -15,8 +15,19 @@ def handle(self, privs, **kwargs):
         dry_run = kwargs.get('dry_run')
         verbose = kwargs.get('verbose')
         noinput = kwargs.get('noinput')
+        skip_edition = kwargs.get('skip_edition')
+
+        query = SoftwarePlanVersion.objects
+
+        filters = {}
```
dimagimon: F841 local variable 'filters' is assigned to but never used

nickpell (approved):

PR #15116, opened 2017-02-22
last submission returns list

emord:
http://manage.dimagi.com/default.asp?248123

Always returns a list of one since if a user doesn't have a submission then the dict doesn't contain the user id as a key

@gcapalbo @benrudolph


gcapalbo (approved):

benrudolph:
tests failed on downloading some JS libs in the JS tests. this is clearly not related, merging

PR #15118, opened 2017-02-23
Fix options list for parent module selection

orangejenny:
http://manage.dimagi.com/default.asp?248126

@millerdev / @nickpell


czue:
just to check if you save while this bug is deployed does it blow away your parent module config?

orangejenny:
yes

czue:
ok this feels hotfix worthy. I'll deploy now

PR #15121, opened 2017-02-23
allow property_name to accept expressions

czue:
this supports a relatively obscure use case for base item expression doing fancy stuff using a property name expression based on the iterator value.

should hopefully be backwards compatible though.

@calellowitz


czue:
:white_check_mark:

PR #15122, opened 2017-02-23
Initial testing docs

benrudolph:
@dimagi/test-team first go at the docs, please feel free to add commits

### DIFF
--- a/docs/testing.rst
+++ b/docs/testing.rst
@@ -1,0 +1,60 @@
+=======
+Testing best practices
+=======
+
+Using setUpClass
+================
+
+Doing a lot of work in the ``setUp`` call of a test class means that it will be run on every test. This
+quickly adds a lot of run time to the tests. Some things that can be easily moved to ``setUpClass`` are domain
+creation, user creation, or any other static models needed for the test.
+
+Using SimpleTestCase
+====================
+
+The SimpleTestCase runs tests without a database. Many times this can be achieved through the use of the `mock
+library <http://www.voidspace.org.uk/python/mock/>`. A good rule of thumb is to have 80% of your tests be unit
+tests that utilize ``SimpleTestCase``, and then 20% of your tests be integration tests that utilize the
+database and ``TestCase``.
+
+
+Tearing down tests
+==================
+
+It is good practice to ensure that all objects you have created in the test database are deleted when the test
+class finishes running. This often happens in the ``tearDown`` method or the ``tearDownClass`` method.
+However, unneccessary cleanup "just to be safe" can add a large amount of time onto your tests.
+
+Combining test classes
+======================
+
+Sometimes classes share the same base class and inherit the ``setUpClass`` function. Below is an example:
+
+.. code:: python
+
+    class MyBaseTestClass(TestCase):
+
+        @classmethod
+        def setUpClass(cls):
+            ...
+
+
+    class MyTestClass(MyBaseTestClass):
+
+        def test(self):
+            ...
+
+    class MyTestClassTwo(MyBaseTestClass):
+
+        def test(self):
+            ...
+
+
+In the above example the ``setUpClass`` is run twice, once for ``MyTestClass`` and ``MyTestClassTwo``. If
+``setUpClass`` has expensive operations, then it's best for all the tests to be combined under one test class.
+
+Squashing Migrations
+====================
+
+There is overhead to running many migrations at once. Django allows you to squash migrations which will help
+speed up the migrations when running tests.
--- a/docs/testing.rst
+++ b/docs/testing.rst
@@ -13,7 +13,7 @@ Using SimpleTestCase
 ====================
 
 The SimpleTestCase runs tests without a database. Many times this can be achieved through the use of the `mock
-library <http://www.voidspace.org.uk/python/mock/>`. A good rule of thumb is to have 80% of your tests be unit
+library <http://www.voidspace.org.uk/python/mock/>`_. A good rule of thumb is to have 80% of your tests be unit
 tests that utilize ``SimpleTestCase``, and then 20% of your tests be integration tests that utilize the
 database and ``TestCase``.
 
--- a/docs/testing.rst
+++ b/docs/testing.rst
@@ -68,7 +68,7 @@ In the above example the ``setUpClass`` is run twice, once for ``MyTestClass`` a
 However this can lead to giant Test classes. If you find that all the tests in a package or module are sharing
 the same set up, you can write a setup method for the entire package or module. More information on that can be found `here <http://pythontesting.net/framework/nose/nose-fixture-reference/#package>`_.
 
-Tearing down tests
+Test tear down
 ==================
 
 It is important to ensure that all objects you have created in the test database are deleted when the test
--- a/docs/testing.rst
+++ b/docs/testing.rst
@@ -72,8 +72,8 @@ database and ``TestCase``.
 
 CommCareHQ also has some custom in mocking tools.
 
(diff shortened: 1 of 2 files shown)


benrudolph:
hehe killed build

also didn't add anything about travis. we may want a section in there

emord:
Haven't thought about this in depth yet, but maybe we should have a list of common mocks that one can use. I just found out about the `UserESFake` last week and it could have been useful in the past

docs/testing.rst
```diff
@@ -0,0 +1,60 @@
...
+However, unneccessary cleanup "just to be safe" can add a large amount of time onto your tests.
+
+Combining test classes
+======================
+
+Sometimes classes share the same base class and inherit the ``setUpClass`` function. Below is an example:
+
+.. code:: python
+
+    class MyBaseTestClass(TestCase):
```
emord: This might be a bad example. I think that you would want the base class to inherit from object and have the child classes inherit tow classes http://stackoverflow.com/questions/4566910/abstract-test-case-using-python-unittest

benrudolph: ah i think this is supposed to be a bad example, or are you saying that this is a bad "bad example". that said maybe we shouldn't put bad examples in it

benrudolph: trying to make an example out of what happened in ilsgateway

emord: Oh. Gotcha that makes sense. Although these setups would actually run 3 times because  all three inherit from `TestCase`

docs/testing.rst
```diff
@@ -0,0 +1,60 @@
...
+The SimpleTestCase runs tests without a database. Many times this can be achieved through the use of the `mock
+library <http://www.voidspace.org.uk/python/mock/>`_. A good rule of thumb is to have 80% of your tests be unit
+tests that utilize ``SimpleTestCase``, and then 20% of your tests be integration tests that utilize the
+database and ``TestCase``.
+
+
+Tearing down tests
+==================
+
+It is good practice to ensure that all objects you have created in the test database are deleted when the test
```
millerdev: Nit: "It is good practice to ensure..." sounds like it's good but not required. Suggest changing to "It is important to ensure..."

docs/testing.rst
millerdev: I'm pro not putting bad examples in docs without clearly marking them with something like
```
# BAD CODE; DO NOT DO THIS
```
Also might be worth putting an example of the right way to do it.

docs/testing.rst
```diff
@@ -0,0 +1,60 @@
+=======
+Testing best practices
+=======
+
+Using setUpClass
+================
+
+Doing a lot of work in the ``setUp`` call of a test class means that it will be run on every test. This
```
emord: There's also package and module. http://pythontesting.net/framework/nose/nose-fixture-reference/#package

benrudolph:
thanks for the feedback guys. i've updated with all your comments. emord i added a small list of mock libraries HQ has, though i'm not sure if it's exhaustive

docs/testing.rst
```diff
@@ -0,0 +1,98 @@
...
+It is important to ensure that all objects you have created in the test database are deleted when the test
+class finishes running. This often happens in the ``tearDown`` method or the ``tearDownClass`` method.
+However, unneccessary cleanup "just to be safe" can add a large amount of time onto your tests.
+
+Combining test classes
+======================
+
+Sometimes classes share the same base class and inherit the ``setUpClass`` function. Below is an example:
+
+.. code:: python
```
millerdev: These code samples are not showing up in the [rich text view](https://github.com/dimagi/commcare-hq/blob/4bed22abf2f22b582d9af0d4ba63cf1da31e9e08/docs/testing.rst)

docs/testing.rst
```diff
@@ -0,0 +1,98 @@
...
+        def test(self):
+            ...
+
+    class MyTestClassTwo(MyBaseTestClass):
+
+        def test(self):
+            ...
+
+
+In the above example the ``setUpClass`` is run three times, once for ``MyBaseTestClass``, ``MyTestClass`` and ``MyTestClassTwo``. If
```
millerdev: > ``setUpClass`` is run three times

In `BAD EXAMPLE`, this is not true, it's only run twice because `MyBaseTestClass` has no test methods.

Also, you say "In the above example" but both GOOD and BAD examples are above and it's not clear which you're referring to, and the GOOD example is above the BAD example, so that adds to the confusion.

docs/testing.rst
```diff
@@ -0,0 +1,98 @@
...
+``setUpClass`` has expensive operations, then it's best for all the tests to be combined under one test class.
```
millerdev: Nit: this feels like a separate point, and isn't really relevant to the examples given since neither GOOD nor BAD examples are doing what this is saying: combine all tests under one test class. Maybe put it in a separate paragraph?

docs/testing.rst
benrudolph: 👍 will take a look, thought it was working. thanks for checking

docs/testing.rst
benrudolph: 👍 sorry i had originally thought that it was only run twice too, but then changed based on this comment: https://github.com/dimagi/commcare-hq/pull/15122#discussion_r102724606. i should just run it and figure out myself

millerdev: Did that:
```py
from unittest import TestCase

class MyBaseTestClass(TestCase):

    runs = 0

    @classmethod
    def setUpClass(cls):
        MyBaseTestClass.runs += 1
        print("setup", MyBaseTestClass.runs)

    def _my_helper_function(self):
        print("help")


class MyTestClass(MyBaseTestClass):

    def test(self):
        pass

class MyTestClassTwo(MyBaseTestClass):

    def test(self):
        pass
```
output:
```
$ nosetests sampletest.py -vs
('setup', 1)
test (sampletest.MyTestClass) ... ok
('setup', 2)
test (sampletest.MyTestClassTwo) ... ok

Ran 2 tests in 0.001s

OK
```

docs/testing.rst
```diff
@@ -0,0 +1,97 @@
...
+        def test(self):
+            ...
+
+    class MyTestClassTwo(MyBaseTestClass):
+
+        def test(self):
+            ...
+
+
+In the above example the ``setUpClass`` is run twice, once for ``MyTestClass`` and once for ``MyTestClassTwo``. If ``setUpClass`` has expensive operations, then it's best for all the tests to be combined under one test class.
```
millerdev: I hate to keep beating this horse, but the example below does not combine all tests "under one test class" as this sentence describes. It has two test classes with separate `setUpClass` methods.

This code example appears to be demonstrating how to share a helper method with multiple test classes using a mixin, which is a good technique to document. However, it seems like a separate point from merging tests to avoid running expensive setup repeatedly.

benrudolph: Please keep beating the horse haha (sorry I haven't had the time I wanted to sit down and write this cleanly and appreciate this review. i'll ping when i think ready for another review, didn't mean to make you review again)

benrudolph: also i do see what you mean, i had initially thought your comment was pointing to another issue

benrudolph: https://github.com/dimagi/commcare-hq/pull/15122/commits/75c4346ab188651ce3f727dc7edfd9c80327c69d -- ok i think this example is better

millerdev (approved):
Not sure if you're done with this yet, but looks great!

emord (approved):

benrudolph:
for now, will add more incrementally 📈

PR #15124, opened 2017-02-23
don't re-create the ES client on every call

snopoke:
From what I can tell this is how the ES client is supposed to be used (since it has connection pools etc inside it).

### DIFF
--- a/corehq/elastic.py
+++ b/corehq/elastic.py
@@ -26,7 +26,7 @@ def get_es_new():
     Get a handle to the configured elastic search DB.
     Returns an elasticsearch.Elasticsearch instance.
     """
-    if not get_es_new._es_client:
+    if not getattr(get_es_new, '_es_client', None):
         es_hosts = getattr(settings, 'ELASTICSEARCH_HOSTS', None)
         if not es_hosts:
             es_hosts = [settings.ELASTICSEARCH_HOST]
--- a/corehq/elastic.py
+++ b/corehq/elastic.py
@@ -21,7 +21,7 @@ from corehq.pillows.mappings.user_mapping import USER_INDEX_INFO
 from corehq.pillows.mappings.xform_mapping import XFORM_INDEX_INFO
 
 
-def get_es_new(**kwargs):
+def get_es_new():
     """
     Get a handle to the configured elastic search DB.
     Returns an elasticsearch.Elasticsearch instance.
--- a/corehq/elastic.py
+++ b/corehq/elastic.py
@@ -38,7 +38,12 @@ def get_es_new():
             }
             for host in es_hosts
         ]
-        get_es_new._es_client = Elasticsearch(hosts, **kwargs)
+        get_es_new._es_client = Elasticsearch(
+            hosts,
+            sniff_on_start=True,
+            sniff_on_connection_fail=True,
+            sniffer_timeout=60
+        )
     return get_es_new._es_client
 
 
--- a/corehq/elastic.py
+++ b/corehq/elastic.py
@@ -26,18 +26,20 @@ def get_es_new(**kwargs):
     Get a handle to the configured elastic search DB.
     Returns an elasticsearch.Elasticsearch instance.
     """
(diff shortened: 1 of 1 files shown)


corehq/elastic.py
```diff
@@ -26,18 +26,20 @@ def get_es_new(**kwargs):
...
-
-    hosts = [
-        {
-            'host': host,
-            'port': settings.ELASTICSEARCH_PORT,
-        }
-        for host in es_hosts
-    ]
-    return Elasticsearch(hosts, **kwargs)
+    if not get_es_new._es_client:
```
emord: Should we also turn on connection sniffing periodically to make sure that the cluster/connection is still in a good state? http://elasticsearch-py.readthedocs.io/en/master/index.html#sniffing

snopoke: :+1: 1e13edc

emord: I assume that the client automatically handles reconnecting to ES if something goes wrong with the sniffer? I actually don't know too much about it, just did a brief look through it

emord (approved):

czue:
nice

emord:
`AttributeError: 'function' object has no attribute '_es_client'`

esoergel (approved):

emord:
@snopoke sounds like we should monitor the ES queries to see if this had a bad effect. shortened from web team slack:

```
Daniel Miller [10:42] 
elasticsearch queries in my local test setup (docker) are taking 4s each. anyone seen that before or have ideas as to why/how to fix?

[10:50]  
looks like all queries, and this definitely was not happening last week when I was profiling the same test

[10:50]  
 ```   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
       10   40.442    4.044   40.442    4.044 {method 'connect' of '_socket.socket' objects}
```

[10:51]  
socket.connect is slow, so maybe a network issue in my docker setup?

Jonathan Emord [10:52] 
revert https://github.com/dimagi/commcare-hq/pull/15124 locally and test it out

Daniel Miller [10:59] 
yeah, reverting that fixed it
```

@millerdev

**Comment by snopoke** [comment]
@emord @millerdev anything happen on this? Seems odd to me that this would make it slower. I can see how it may affect the initial setup time for the client (since it's doing sniffing) but after that it should actually be faster.

@millerdev did you try disabling sniffing to see if that was the cause?

**Comment by millerdev** [comment]
@snopoke the issue I observed on Friday seemed somehow related to running in a VirtualBox VM. Both @gcapalbo and I observed much slower test times with this PR merged. I run test services in Docker/boot2docker (VirtualBox VM as Docker host) on Mac. Giovanni said he was "running in an ubuntu vm, but not using docker." Other members of the team (I think all using Docker) on Linux could not reproduce the issue.

Here's the test we ran:
```
./manage.py test --with-timing --pretty-timing custom.ewsghana.tests.handlers.receipt:ReceiptsTest
```

Normally it takes about 15s. When I run it with this PR merged it takes >800s. It took ~300s for Giovanni once he updated master to a point after this PR was merged. If I revert the commits in this PR the test time drops back down to ~15s.

I was profiling a different part of the code when I ran into this, so I just used the revert workaround to continue profiling. I'll likely dig deeper into this to figure out a better fix that can be deployed everywhere if it doesn't get fixed by someone else first.

I'll try disabling sniffing to see if that helps first.

snopoke:
Guven that our ES cluster is pretty static I don't think we get a lot of
benefit from sniffing so if it is that (which I highly suspect it is) then
we could just remove it and rely our localsettings to know all the ES nodes.

NoahCarnahan:
@snopoke @millerdev 
I have a related but different issue. I'm running elasticsearch with docker in a virtualbox VM. However I run django locally on the VM's host. When sniffing is on, the runserver process tries to connect to the node es node at 172.17.0.4, which I guess is a private address only accessible within the VM, so I get a timeout and am unable to connect to es. I guess I could fix this by running runserver in the VM as well, but would probably be easier for me to just rely on localsettings, so this is another vote for that approach I guess :)

PR #15066, opened 2017-02-20
Externalize JavaScript: module view

orangejenny:
Haven't dealt with the `nav_menu_media_js.html` includes yet, but moved all the other js into `module-view.js`.

Best reviewed by commit.

@gcapalbo / @snopoke

PR #15129, opened 2017-02-23
Revert externalize js module view

orangejenny:
http://manage.dimagi.com/default.asp?248159

Reverts https://github.com/dimagi/commcare-hq/pull/15066

@NoahCarnahan


orangejenny:
Should go into today's deploy; kill build if needed.

PR #15133, opened 2017-02-24
Edit forms fixes

benrudolph:
@wpride @snopoke some fixes we discussed for edit forms. Also will the first commit is the reason edit forms wasn't working for formplayer. i think someone accidentally hardcoded `oneQuestionPerScreen` to true

### DIFF
--- a/corehq/apps/reports/views.py
+++ b/corehq/apps/reports/views.py
@@ -1771,6 +1771,14 @@ class EditFormInstance(View):
             non_parents = filter(lambda cb: cb.path == [], case_blocks)
             if len(non_parents) == 1:
                 edit_session_data['case_id'] = non_parents[0].caseblock.get(const.CASE_ATTR_ID)
+                case = CaseAccessors(domain).get_case(edit_session_data['case_id'])
+                if case.closed:
+                    return _error(_(
+                        u'Case {} is closed. Please reopen the '
+                        u'case before editing the form'
+                    ).format(case.case_id))
+                elif case.deleted:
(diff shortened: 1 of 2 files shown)


corehq/apps/reports/views.py
```diff
@@ -1771,6 +1771,14 @@ def _error(msg):
...
             if len(non_parents) == 1:
                 edit_session_data['case_id'] = non_parents[0].caseblock.get(const.CASE_ATTR_ID)
+                case = CaseAccessors(domain).get_case(edit_session_data['case_id'])
+                if case.closed:
+                    return _error(_(
+                        u'Case {} is closed. Please reopen the '
+                        u'case before editing the form'
+                    ).format(case.case_id))
+                elif case.deleted:
+                    return _error(_(u'Case {} is deleted. Cannot edit this form.').format(case.case_id))
```
snopoke: Should this not also apply to any of the cases edited by the form?

Also would be nice to use have the error link to the case using the case name instead of just have the case ID.

benrudolph: :+1: to the link. As far as I know this is the only case we load for basic forms. I think there may be more in the form that could be turned up

benrudolph: i think we show a 404 when we link to a deleted case, but i cleaned up the UX for archived case https://github.com/dimagi/commcare-hq/pull/15133/commits/3f3ffacd710a58159b1e460625440e24ffc6a8d9

screen shot 2017-02-27 at 11 45 29 am

PR #15134, opened 2017-02-24
fix archive forms

czue:
http://manage.dimagi.com/default.asp?248250

Introduced in https://github.com/dimagi/commcare-hq/pull/15111/

This is pretty weird. The default instantiation of `Template` actually has a [different api](https://github.com/django/django/blob/stable/1.9.x/django/template/base.py#L199) for this function. I think it only exists to [support this use case](https://github.com/django/django/blob/stable/1.9.x/django/template/base.py#L176), but at first blush I would argue this is a bug in django 1.9.

@nickpell cc @calellowitz


nickpell:
weird, thanks for fixing

PR #15137, opened 2017-02-24
use the right user filter in links to submit history

czue:
http://manage.dimagi.com/default.asp?248253

@esoergel @mkangia guessing this is due to releasing location restrictions? Please confirm it's right to always use this filter now on submit history (seemed so locally, but i didn't try testing, e.g. with a domain that didn't have enterprise).

cc @calellowitz


calellowitz:
will wait on @esoergel or @mkangia but looks good to me

esoergel:
Yeah, this looks right - guessing it's just an oversight.

mkangia:
Thank You @czue

PR #15140, opened 2017-02-24
cache docker images on travis

snopoke:
@dimagi/test-team


emord:
https://docs.travis-ci.com/user/caching/#Things-not-to-cache

snopoke:
Yea, I read that, still found lots of people who do it so think it's worth
a test.

benrudolph:
how do we evaluate this? would we have to make another dummy commit to see if it's faster?

PR #15141, opened 2017-02-24
case bug tests

czue:
picked an arbitrary file and worked on it a bit before realizing the whole thing only takes a few seconds to run.

deleted a bunch of xml files at least.

@dimagi/test-team / @calellowitz


benrudolph:
💀 ❌ Ⓜ️ L

PR #15143, opened 2017-02-24
sync mode tests

czue:
didn't end up making much progress at all on this, but i guess every little bit helps?


benrudolph:
nice! every bit does help. i think multiple people contributed 1 min speed ups which adds up

PR #15144, opened 2017-02-24
don't let raise_conflicting_dates go off against a canceled subscription

nickpell:
http://manage.dimagi.com/default.asp?248270

@esoergel @biyeun

### DIFF
--- a/corehq/apps/accounting/models.py
+++ b/corehq/apps/accounting/models.py
@@ -77,6 +77,8 @@ SMALL_INVOICE_THRESHOLD = 100
 
 UNLIMITED_FEATURE_USAGE = -1
 
+CONSISTENT_DATES_CHECK = Q(date_start__lt=F('date_end')) | Q(date_end__isnull=True)
+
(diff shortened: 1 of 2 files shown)


corehq/apps/accounting/models.py
```diff
@@ -1124,7 +1126,11 @@ def raise_conflicting_dates(self, date_start, date_end):
         assert date_start is not None
         for sub in Subscription.objects.filter(
             subscriber=self.subscriber
-        ).exclude(id=self.id).all():
+        ).exclude(
+            id=self.id,
+        ).exclude(
+            CONSISTENT_DATES_CHECK,
```
esoergel: I think this should be `filter`, not `exclude`

esoergel (changes requested):

corehq/apps/accounting/models.py
nickpell: updated

esoergel (approved):

PR #15152, opened 2017-02-26
Change sinon lib

benrudolph:
Think this is causing travis to fail

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -71,7 +71,7 @@
   "devDependencies": {
     "chai": "3.3.0",
     "mocha": "2.3.3",
-    "sinon": "http://sinonjs.org/releases/sinon-1.17.0.js",
+    "sinon": "http://sinonjs.org/releases/sinon-1.17.7.js",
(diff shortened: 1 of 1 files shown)


bower.json
```diff
@@ -71,7 +71,7 @@
   "devDependencies": {
     "chai": "3.3.0",
     "mocha": "2.3.3",
-    "sinon": "http://sinonjs.org/releases/sinon-1.17.0.js",
+    "sinon": "http://sinonjs.org/releases/sinon-1.17.7.js",
```
emord: Why do these use bower.json versus our package.json https://github.com/dimagi/commcare-hq/blob/master/package.json ? seems like our test dependencies are split up

benrudolph: server vs. browser js libs. sinon gets loaded into the browser where as npm is for server js like cli calls to the js-xpath module

benrudolph: the description is mighty confusing though

emord: Ah. That makes sense. Thanks

PR #15154, opened 2017-02-27
avoid rendering 30 day trial message by default as its better self-re…

mkangia:
…ndered only when needed by the template

This is silently ignored by chrome but causes JS error in say, safari


mkangia:
Seems like this got fixed and is not needed anymore.

PR #15158, opened 2017-02-27
Mk 248129 Notify Daily Saved Exports Change

mkangia:
[Change Request](https://manage.dimagi.com/default.asp?248129)

best to review all at once

Also for any design suggestions @dimagi/ux 
screen shot 2017-03-09 at 6 55 54 pm


mkangia:
@NoahCarnahan Considering you would know the best for whats needed here to ensure this renders only for domains that have access to Daily Saved Exports and Export Dashboard Integration.
For now have added just domain privileges check for both. Thats sufficient right?
Also if you have any feedback for changes.

corehq/apps/export/views.py
```diff
@@ -973,7 +975,20 @@ def dispatch(self, request, *args, **kwargs):
         if not (self.has_edit_permissions or self.has_view_permissions
                 or (self.is_deid and self.has_deid_view_permissions)):
             raise Http404()
-        return super(BaseExportListView, self).dispatch(request, *args, **kwargs)
+
+        self.request = request
+
+        if domain_has_daily_saved_export_access(self.domain) or domain_has_excel_dashboard_access(self.domain):
```
NoahCarnahan: I think you also want to do this check:
https://github.com/dimagi/commcare-hq/blob/2d42bbf3088f7ac1ec7a881df1f7f4d27ca5909c/corehq/apps/export/views.py#L1511

corehq/apps/export/views.py
```diff
@@ -973,7 +975,20 @@ def dispatch(self, request, *args, **kwargs):
...
+            self.set_notify_new_daily_saved_export()
+
+        return super(BaseExportListView, self).dispatch(self.request, *args, **kwargs)
+
+    def set_notify_new_daily_saved_export(self):
+        self.request.notify_new_daily_saved_export = False
+
+        if self.request.session.get('notify_new_daily_saved_export', True):
+            self.request.notify_new_daily_saved_export = True
+            self.request.session['notify_new_daily_saved_export'] = False
```
NoahCarnahan: I don't think it's a great user experience for the modal to show on every new session. I think it should be a one-time thing. I know it's a bit more work, but I think the "read" status of the notification should reside somewhere in permanent storage. You could look at the [`corehq.apps.notifications.models.Notification`](https://github.com/dimagi/commcare-hq/blob/2d42bbf3088f7ac1ec7a881df1f7f4d27ca5909c/corehq/apps/notifications/models.py#L20) model might provide some inspiration.

NoahCarnahan (changes requested):

corehq/apps/export/views.py
```diff
@@ -973,7 +975,22 @@ def dispatch(self, request, *args, **kwargs):
...
-        return super(BaseExportListView, self).dispatch(request, *args, **kwargs)
+
+        self.request = request
+
+        if (use_new_daily_saved_exports_ui(self.domain) and
+                not DailySavedExportNotification.notified(self.request.couch_user.user_id, self.domain) and
+                (
+                    domain_has_daily_saved_export_access(self.domain) or
+                    domain_has_excel_dashboard_access(self.domain)
+                )):
```
dimagimon: E123 closing bracket does not match indentation of opening bracket's line

corehq/apps/export/views.py
```diff
@@ -973,7 +975,22 @@ def dispatch(self, request, *args, **kwargs):
...
-        return super(BaseExportListView, self).dispatch(request, *args, **kwargs)
+
+        self.request = request
+
+        if use_new_daily_saved_exports_ui(self.domain) and \
+            not DailySavedExportNotification.notified(self.request.couch_user.user_id, self.domain) and \
+                (
+                    domain_has_daily_saved_export_access(self.domain) or
+                    domain_has_excel_dashboard_access(self.domain)
+                ):
```
dimagimon: E123 closing bracket does not match indentation of opening bracket's line

NoahCarnahan (approved):
Looks good!

mkangia:
@NoahCarnahan thanks for the review. Just added one more [change](https://github.com/dimagi/commcare-hq/pull/15158/commits/c2fd9c030c46ca193cf750ab7d5d46cb8cbd8f97) to check that user joined before the feature release. I have just added a random date for now. 
Do you know the exact date to be checked against? Also pinging kevin for the same on ticket.

corehq/apps/export/models/new.py
```diff
@@ -2357,6 +2360,24 @@ def notified(cls, user_id, domain):
     def mark_notified(cls, user_id, domain):
         cls.objects.get_or_create(user_id=user_id, domain=domain)
 
+    @classmethod
+    def user_added_before_feature_release(cls, user_added_on):
+        return user_added_on < datetime(2016, 2, 1)
```
NoahCarnahan: I think `datetime(2017, 1, 25)` is when it was released.

NoahCarnahan:
Looks good thanks. Ok to merge? (I see the "hold off on merge" tag is still on the PR)

mkangia:
@NoahCarnahan just keeping some time if any feedback by UX team or kevin. Will merge in a while then but before todays deploy

NoahCarnahan:
👍

orangejenny:
Hi Manish, ux feedback: Could this be an `alert-warning` box instead, in or right above the Exports panel, instead of a modal? A modal seems too in-your-face to me, especially since it looks like users will see this modal every time they visit the page. Modals are more for blocking issues, things the user *has* to deal with. Someone who's looking for their daily saved exports will already be scanning the page. 

One other minor thing: making `Daily Saved Exports` and `Excel Dashboard Integration` text into links would be a nice touch.

mkangia:
@orangejenny the users would see this only once and also only for domains that were moved to new daily saved export. Happy to change this if you think it should be still done.

orangejenny:
Hey, if they'll only see it once, and it's limited to people who actually have daily saved exports (which it looks like it is), then it works for me.

Curious, how are we going to decide when to remove this code?

mkangia:
Thanks @orangejenny. since its just as a notification for very few users hence keeping effort low.

As of now the plan is to remove this post march. By that time we are hoping all would know. Also we are storing when the notification is viewed. So we can always check how many users saw this. Also, I think we have an idea about how many domains were moved to this new export. So we can tally if thats available @NoahCarnahan.

orangejenny:
Cool, thanks for the explanation!

PR #15162, opened 2017-02-27
Save last sent date on repeat records that will not be retried

NoahCarnahan:
This change will make the repeat_record_report more useful for canceled repeat records.
https://manage.dimagi.com/default.asp?248041
cc @kaapstorm


kaapstorm (approved):

czue:
thanks!

PR #15163, opened 2017-02-27
create indexes on ucr sql tables

emord:
@gcapalbo @benrudolph to be used on icds reports


gcapalbo (approved):
it looks like this is creating the ability to create the indexes and the indexes are created elsewhere?

if you haven't already, i think it would be good to inspect the number of records in the table(s) where the indexes are being created to understand how long adding the indexes will take.

emord:
Yeah. This likely won't happen until the DB settles down more, but just adding hte ability to do it in the column definition of a UCR. We'll check when we add these to the DB

czue (approved):

PR #12915, opened 2016-08-22
Kill 1.x apps

dannyroberts:
Remove all but most basic plumbing for 1.x apps

PR #15166, opened 2017-02-28
Don't 500 on remote apps

orangejenny:
https://manage.dimagi.com/default.asp?248127

The wrapping bug I think was a side effect of https://github.com/dimagi/commcare-hq/pull/12915 (fyi @dannyroberts ), and then there were a couple of places I wrote app manager v2 code that didn't account for remote apps.

@millerdev

### DIFF
--- a/corehq/apps/app_manager/views/apps.py
+++ b/corehq/apps/app_manager/views/apps.py
@@ -263,7 +263,7 @@ def get_apps_base_context(request, domain, app):
         'app_subset': {
             'commcare_minor_release': app.commcare_minor_release,
             'doc_type': app.get_doc_type(),
-            'form_counts_by_module': [len(m.forms) for m in app.modules] if app.get_doc_type() != "RemoteApp" else [],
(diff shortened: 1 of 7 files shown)


corehq/apps/app_manager/views/apps.py
```diff
@@ -263,7 +263,7 @@ def get_apps_base_context(request, domain, app):
         'app_subset': {
             'commcare_minor_release': app.commcare_minor_release,
             'doc_type': app.get_doc_type(),
-            'form_counts_by_module': [len(m.forms) for m in app.modules],
+            'form_counts_by_module': [len(m.forms) for m in app.modules] if app.get_doc_type() != "RemoteApp" else [],
```
dimagimon: E501 line too long (118 > 115 characters)

dannyroberts:
wow! It's a real testament to our getting off remote apps that it takes 4 months now between when we (I) break remote apps and when we notice it!

orangejenny:
@dannyroberts Totally agree :)

PR #15171, opened 2017-02-28
Prepare user model for app status

benrudolph:
@dimagi/scale-team working on the user model for the app status report by including the last_submission in the user model. still have some open questions.

if we want to support the filter by application dropdown, we'll need to change the model to support `SchemaListProperty` so that we can have the last submission for each application.

I talked to a few GS folk and most seemed to think that it was less important to be able to filter by application and often used it as a way to filter users. some of the primary use cases were:

1. To be able to tell which users haven't updated their phones (with this addition we should be able to sort by build number)
2. Be able to tell which users are not submitting data. This should be possible with last_submission date
3. Be able to tell which users are not syncing. I punted on being able to sort by last sync time, but wouldn't be too hard to add

🐟 or 🏠

### DIFF
--- a/corehq/apps/users/models.py
+++ b/corehq/apps/users/models.py
@@ -771,9 +771,17 @@ class DeviceIdLastUsed(DocumentSchema):
         return all(getattr(self, p) == getattr(other, p) for p in self.properties())
 
 
+class LastSubmission(DocumentSchema):
+    submission_date = DateTimeProperty()
+    app_id = StringProperty()
+    build_id = StringProperty()
+    build_version = IntegerProperty()
+    commcare_version = StringProperty()
+
+
 class ReportingMetadata(DocumentSchema):
 
-    last_submission_date = DateTimeProperty()
+    last_submission = SchemaProperty(LastSubmission)
 
 
 class CouchUser(Document, DjangoUserMixin, IsMemberOfMixin, UnicodeMixIn, EulaMixin):
--- a/corehq/apps/users/models.py
+++ b/corehq/apps/users/models.py
@@ -774,7 +774,6 @@ class DeviceIdLastUsed(DocumentSchema):
 class ReportingMetadata(DocumentSchema):
 
     last_submission_date = DateTimeProperty()
-    last_sync_date = DateTimeProperty()
 
 
 class CouchUser(Document, DjangoUserMixin, IsMemberOfMixin, UnicodeMixIn, EulaMixin):
--- a/corehq/apps/users/models.py
+++ b/corehq/apps/users/models.py
@@ -782,7 +782,7 @@ class LastSubmission(DocumentSchema):
 
 class ReportingMetadata(DocumentSchema):
 
-    last_submission = SchemaProperty(LastSubmission)
+    last_submissions = SchemaListProperty(LastSubmission)
 
 
 class CouchUser(Document, DjangoUserMixin, IsMemberOfMixin, UnicodeMixIn, EulaMixin):
--- a/corehq/apps/users/models.py
+++ b/corehq/apps/users/models.py
@@ -806,7 +806,7 @@ class CouchUser(Document, DjangoUserMixin, IsMemberOfMixin, UnicodeMixIn, EulaMi
     assigned_location_ids = StringListProperty()
     has_built_app = BooleanProperty(default=False)
 
-    reporting_metadata = SchemaDictProperty(ReportingMetadata)
+    reporting_metadata = SchemaProperty(ReportingMetadata)
 
     _user = None
 
--- a/corehq/apps/users/models.py
+++ b/corehq/apps/users/models.py
@@ -771,6 +771,12 @@ class DeviceIdLastUsed(DocumentSchema):
         return all(getattr(self, p) == getattr(other, p) for p in self.properties())
 
 
+class ReportingMetadata(DocumentSchema):
+
+    last_submission_date = DateTimeProperty()
+    last_sync_date = DateTimeProperty()
+
+
 class CouchUser(Document, DjangoUserMixin, IsMemberOfMixin, UnicodeMixIn, EulaMixin):
     """
     A user (for web and commcare)
@@ -799,6 +805,8 @@ class CouchUser(Document, DjangoUserMixin, IsMemberOfMixin, UnicodeMixIn, EulaMi
     location_id = StringProperty()
     assigned_location_ids = StringListProperty()
     has_built_app = BooleanProperty(default=False)
+
+    reporting_metadata = SchemaDictProperty(ReportingMetadata)
 
     _user = None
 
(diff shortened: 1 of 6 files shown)


corehq/apps/users/models.py
```diff
@@ -771,6 +771,19 @@ def __eq__(self, other):
...
 
+class LastSubmission(DocumentSchema):
+    submission_date = DateTimeProperty()
+    app_id = StringProperty()
+    build_id = StringProperty()
+    build_version = IntegerProperty()
+    commcare_version = StringProperty()
+
+
+class ReportingMetadata(DocumentSchema):
```
benrudolph: this might be over-engineering, we could possible want to just stick `last_submission` on the user model itself

corehq/ex-submodules/pillowtop/processors/form.py
```diff
@@ -28,16 +32,59 @@ def process_change(self, pillow_instance, change):
...
+    except Http404:
+        pass
+
+    if app and not app.has_submissions:
+        app.has_submissions = True
+        app.save()
+
+
+def mark_latest_submission(domain, user_id, app_id, build_id, version, metadata, received_on):
+    user = CouchUser.get_by_user_id(user_id, domain)
```
benrudolph: this is cached

corehq/ex-submodules/pillowtop/processors/form.py
```diff
@@ -28,16 +32,59 @@ def process_change(self, pillow_instance, change):
...
+        return
+
+    try:
+        received_on_datetime = string_to_utc_datetime(received_on)
+    except ValueError:
+        return
+
+    current_last_submission = user.reporting_metadata.last_submission.submission_date
+
+    app_version_info = get_app_version_info(
```
benrudolph: this can do an app lookup, but that is also cached

corehq/apps/users/models.py
```diff
@@ -800,6 +813,8 @@ class CouchUser(Document, DjangoUserMixin, IsMemberOfMixin, UnicodeMixIn, EulaMi
     assigned_location_ids = StringListProperty()
     has_built_app = BooleanProperty(default=False)
 
+    reporting_metadata = SchemaProperty(ReportingMetadata)
```
czue: will review closer later but was thinking about this last night - is there a reason this needs to be on the user model as opposed to a separate thing we can join with in elasticsearch? just thinking that saving all this extra data will make that model heavier (and therefore slower to access when loaded on every page). have also been wanting to clear all the computed stuff out of the domain model at some point...

snopoke: +1 to having it separated from the user model. I also think it should be tracked per device / user combo and combined with [DeviceIdLastUsed](https://github.com/dimagi/commcare-hq/blob/7bd298e2d8870bca8ad32ab77b61c3ffac94fb27/corehq/apps/users/models.py#L766-L766)

benrudolph: i did this way because i knew it'd make the query easier on the report side since i was not sure how much ES supported doing joins. the reason it's semi-important that this be able to be done in one query is so we can grab just the top 10 users and sort by one of those properties (currently that the reason we have to fetch every user). would be interested to see how much these kinds of extra properties affect the load time. my intuition was that adding things like string/integers are fairly lightweight

there looks to be some support for ES joins, but it's not that fully featured: https://www.elastic.co/guide/en/elasticsearch/reference/current/joining-queries.html

emord: > the reason it's semi-important that this be able to be done in one query is so we can grab just the top 10 users and sort by one of those properties (currently that the reason we have to fetch every user)

I think the way I worked around this in case activity was, if the report is sorted on username then I grabbed  the paginated users from UserES and then added that to a user_id filter for cases. If it was sorted on something else, then I would just add a limit to the case query and look up those user ids in UserES. I think you could do the same here by using last_form_submission.

The resulting query code probably won't be as readable as this, so it'll fit in with the rest of reports :P

benrudolph: > if the report is sorted on username

i don't think we can assume that report is sorted on username 😞

emord: Right. You wouldn't assume

https://github.com/dimagi/commcare-hq/blob/master/corehq/apps/reports/standard/monitoring.py#L296

I looked at the sort column and if it was username, i paginated the users then did the case query. Otherwise I grabbed the top x from the cases and then found the usernames

benrudolph:
Might be misunderstanding what you're suggesting but I think if we sort by
username the report loses its usefulness

corehq/apps/users/models.py
emord: Offlined with ben. My strategy doesn't work well here because this report needs afilter on users

corehq/ex-submodules/pillowtop/tests/test_form_submission_metadata_tracker_processor.py
```diff
@@ -0,0 +1,166 @@
+from django.test import TestCase
+from dimagi.utils.parsing import string_to_utc_datetime
+
+from corehq.apps.users.models import CommCareUser, CouchUser, LastSubmission
```
dimagimon: F401 'LastSubmission' imported but unused

corehq/ex-submodules/pillowtop/tests/test_form_submission_metadata_tracker_processor.py
```diff
@@ -0,0 +1,166 @@
...
+            self.user._id,
+            'other-app-id',
+            self.build_id,
+            self.version,
+            self.metadata,
+            submission_date,
+        )
+        user = CouchUser.get_by_user_id(self.user._id, self.domain)
+        self.assertEqual(len(user.reporting_metadata.last_submissions), 2)
+
```
dimagimon: W391 blank line at end of file

benrudolph:
screen shot 2017-03-01 at 10 59 36 am
Ok this looks like this on the user model now. I offlined with Simon and I think we're in a agreement that this is a decent stop gap until we figure out a longer term solution. Currently using a star schema for ES is a route we want to explore, but it will require us to upgrade ES in order to do parent child joins. We could also build up a star schema for SQL, but I still think we should try and push out a few short term fixes for the icds team, and then for the longer term plan work on this new star schema. My next task is going to be looking into upgrading ES to 5.X

snopoke (approved):

snopoke:
still has a don't merge label on it

emord:
> My next task is going to be looking into upgrading ES to 5.X

I think what's stopping us from upgrading to ES 2.x is that there are some queries (I think in custom reports and hq admin reports) that use facets which aren't supported in 2.x. not sure if we have blockers from 2.x to 5.x but most queries use our ORM so it shouldn't be awful

I also think we need to upgrade to 2.x then to 5.x because indicies made in 1.7 can't be read in 5.x so the process will be

upgrade to 2.x
reindex everything
upgrade to 5.x

benrudolph:
fun stuff. thanks for the info, still unsure if we're going to do this in ES or SQL

benrudolph:
approved, merging

PR #15172, opened 2017-02-28
Ensure timestamp is always naive datetime for comparison

gcapalbo:
@emord @benrudolph

### DIFF
--- a/corehq/apps/data_interfaces/models.py
+++ b/corehq/apps/data_interfaces/models.py
@@ -1,3 +1,4 @@
+import pytz
 import re
 from collections import defaultdict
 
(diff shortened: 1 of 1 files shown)


corehq/apps/data_interfaces/models.py
```diff
@@ -217,6 +218,16 @@ def _try_date_conversion(self, date_or_string):
 
         return date_or_string
 
+    def clean_datetime(self, timestamp):
+        if not isinstance(timestamp, datetime):
+            timestamp = datetime.combine(timestamp, time(0, 0))
+
+        if timestamp.tzinfo:
+            # Convert to UTC and make it a naive datetime for comparison to datetime.utcnow()
+            timestamp = timestamp.astimezone(pytz.utc).replace(tzinfo=None)
```
benrudolph: I think you could just use: https://github.com/dimagi/dimagi-utils/blob/master/dimagi/utils/parsing.py#L36

benrudolph (approved):
one comment that might be useful, but it's fairly minor. will let you decide

corehq/apps/data_interfaces/models.py
gcapalbo: hmm, looks like that takes a string though and at this point i've already got a timezone-aware datetime

benrudolph: ah oops missed that. looks good then

PR #15174, opened 2017-02-28
Pass restoreAs into formplayer for edits

wpride:
Needed to get the right user data when editing forms on Formplayer

cross-request: https://github.com/dimagi/formplayer/pull/304


corehq/apps/cloudcare/static/cloudcare/js/touchforms-inline.js
```diff
@@ -65,6 +65,7 @@
                 data.post_url = options.submitUrl;
                 data.domain = options.domain;
                 data.username = options.username;
+                data.restoreAs = options.restoreAs;
```
benrudolph: nice

benrudolph (approved):

PR #15176, opened 2017-02-28
icds reports

emord:
@sheelio Here's some changes and questions to the icds reports

I don't think that the month indexes will be hugely helpful now (although may be a ~2x speedup), but will likely be a huge help once we scale out and have lots of different months of data

These indexes will mostly help with the where clauses. I haven investigated indexes for the group by clauses yet because I haven't been able to get postgres to generate different query plans for group bys

buddies @gcapalbo @benrudolph


custom/icds_reports/migrations/sql_templates/create_functions.sql
```diff
@@ -484,11 +493,17 @@ BEGIN
 	EXECUTE 'CREATE INDEX ' || quote_ident(_tablename || '_indx1') || ' ON ' || quote_ident(_tablename) || '(state_id, district_id, block_id, supervisor_id, awc_id)';
 	EXECUTE 'CREATE INDEX ' || quote_ident(_tablename || '_indx2') || ' ON ' || quote_ident(_tablename) || '(ccs_status)';
 	EXECUTE 'CREATE INDEX ' || quote_ident(_tablename || '_indx3') || ' ON ' || quote_ident(_tablename) || '(trimester)';
+    -- is a month query actually useful since the tables are partitioned by months? (general q for every table)
```
sheelio: I'm actually not sure if the month index is useful - the partitions might make it moot.

custom/icds_reports/migrations/sql_templates/create_functions.sql
```diff
@@ -1166,6 +1190,9 @@ BEGIN
 	all_text = 'All';
 	null_value = NULL;
 
+    -- should there be indexes created on this table?
+    -- it doesn't appear to be filtered in this file. maybe tableau touches it?
+
```
sheelio: This table is used by Tableau (there is a join between this table and any of the monthly tables for Tableau queries).

custom/icds_reports/tasks.py
```diff
@@ -14,12 +17,22 @@ def move_ucr_data_into_aggregation_tables():
         with connections[settings.ICDS_UCR_DATABASE_ALIAS].cursor() as cursor:
 
             path = os.path.join(os.path.dirname(__file__), 'sql_templates', 'update_locations_table.sql')
+            # This seems like it shouldn't be updated often. Should only happen
+            # if there's been a change in the table since yesterday?
```
sheelio: Can we move it to a weekly task?  Not sure if an easy way to check if the table has been updated since a date.

gcapalbo:
just curious if we've looked into the size of the tables we're  creating indexes on just to get an idea for how long the create index statements will run

emord:
@gcapalbo I went to the preview data and about half are quite small < 10k rows and should be quick

The indexes that are based on forms are huge. 3 of them didn't load when I went to preview data. I think it's  because they get hung on count https://github.com/dimagi/commcare-hq/blob/master/corehq/apps/userreports/views.py#L1175-L1180

custom/icds_reports/migrations/0008_20170301.py
```diff
@@ -0,0 +1,18 @@
...
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('icds_reports', '0007_20170215'),
+    ]
+
+    operations = [
+        migrator.get_migration('create_functions.sql'),
+    ]
```
gcapalbo: nit: i typically prefer giving migrations names rather than relying on the default django date/time

emord: Changed it. Was going off of the previous migrations which just a date stamp

emord:
@sheelio The changes to the static data source would trigger a rebuild, when the configuration is loaded (on deploy). It looks like it could be kind of messy to make it not trigger a rebuild, but I can do that if you think it's preferable. We could also split out some of the data source changes into separate PRs so that only some of them are run at one time on a deploy

sheelio:
@emord 

I'm not quite sure how long the indexing will take, but it might be worth turning off the aggregation tasks while the re-indexes happen, and then deploy another change to turn on the aggregation

custom/icds_reports/migrations/sql_templates/create_functions.sql
emord: Thanks I think that the primary key you've set should take care of that

custom/icds_reports/migrations/sql_templates/create_functions.sql
emord: agreed

custom/icds_reports/tasks.py
emord: Not going to worry about this now. we can take a look at the logs now that we have timestamps later

emord:
@sheelio disabled the aggregation task so that when the tables are rebuilt it doesn't remove any data from tableau. I'll create a new PR after this is merged to turn on aggregation again

gcapalbo (approved):
looks good to me. do we need to do any special coordinating with a deploy for this or can this be merged?

emord:
@sheelio can confirm, but I don't think we need to do any special coordinating. This will rebuild on next deploy and the one after we'll turn the aggregation back on

sheelio:
Confirmed.

PR #15178, opened 2017-02-28
use same table for ucr

emord:
@dimagi/test-team I looked at this briefly for the test team day, but it's pretty complicated to do this on most tests. Most tests are currently creating different data sources for each test, but some look like they could be combined as one.

There are really big gains per test if we can decrease the amount of rebuilding we do

before: `Ran 4 tests in 26.804s`
after: `Ran 8 tests in 8.489s`

@gcapalbo @benrudolph

### DIFF
--- a/corehq/apps/userreports/adapter.py
+++ b/corehq/apps/userreports/adapter.py
@@ -21,6 +21,9 @@ class IndicatorAdapter(object):
     def refresh_table(self):
         raise NotImplementedError
 
+    def clear_table(self):
+        raise NotImplementedError
+
     def get_query_object(self):
         raise NotImplementedError
 
--- a/corehq/apps/userreports/adapter.py
+++ b/corehq/apps/userreports/adapter.py
@@ -1,6 +1,7 @@
 from __future__ import absolute_import
 from dimagi.utils.decorators.memoized import memoized
 from dimagi.utils.logging import notify_exception
+from corehq.util.test_utils import unit_testing_only
 
 
 class IndicatorAdapter(object):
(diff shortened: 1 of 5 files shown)


corehq/apps/userreports/adapter.py
```diff
@@ -21,6 +21,9 @@ def drop_table(self):
     def refresh_table(self):
         raise NotImplementedError
 
+    def clear_table(self):
```
emord: For use in the future when we use the same data source, but want to test different data

corehq/apps/userreports/es/adapter.py
```diff
@@ -164,6 +164,9 @@ def rebuild_table_if_necessary(self):
     def refresh_table(self):
         self.es.indices.refresh(index=self.table_name)
 
+    def clear_table(self):
+        self.rebuild_table()
```
emord: ES doesn't actually have a way to delete all data in one index

benrudolph:
wow nice that is some huge improvements. looks like build was cancelled?

emord:
@benrudolph build won't work until aws comes back

benrudolph:
ah yes just getting caught up that. crazy

gcapalbo:
looks good to me. i don't have too much context but should some of those clear_table methods have a unit testing only decorator, or are there valid scenarios where we'd run them live?

emord:
@gcapalbo sounds reasonable https://github.com/dimagi/commcare-hq/pull/15178/commits/b604d81513a8ce2884b05a316688cbadce376f19

gcapalbo (approved):

PR #15181, opened 2017-02-28
add logging to debug https://manage.dimagi.com/default.asp?246697

nickpell:
This was in the Interrupt queue for a month, and we have no idea what's going on.  Adding some debug statements to help understand where it's hanging.

@dannyroberts


dannyroberts (approved):
Looks good to me! Good call on adding debugging statements rather than letting it sit there confusing us

PR #15182, opened 2017-02-28
support max aggregation on strings in ES

emord:
@gcapalbo @benrudolph

### DIFF
--- a/corehq/apps/userreports/reports/specs.py
+++ b/corehq/apps/userreports/reports/specs.py
@@ -186,8 +186,27 @@ class FieldColumn(ReportColumn):
     def get_fields(self, data_source_config=None, lang=None):
         return [self.field]
 
+    def _data_source_col_config(self, data_source_config):
+        return filter(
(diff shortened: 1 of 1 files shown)


gcapalbo (approved):

corehq/apps/userreports/reports/specs.py
```diff
@@ -186,8 +186,27 @@ def get_column_config(self, data_source_config, lang):
     def get_fields(self, data_source_config=None, lang=None):
         return [self.field]
 
+    def _data_source_col_config(self, data_source_config):
+        return filter(
+            lambda c: c['column_id'] == self.field, data_source_config.configured_indicators
+        )[0]
```
benrudolph: fine to fail hard here if this doesn't return any items?

benrudolph (approved):

corehq/apps/userreports/reports/specs.py
emord: Yeah that should be fine because if this fails it's an invalid report configuration. It does seem like brittle code though

PR #11441, opened 2016-04-28
Feature flag for vellum beta release

orangejenny:
@millerdev / @emord 

Will have no effect until vellum-to-hq is run with target directory vellum_beta

PR #15185, opened 2017-02-28
Feature flag for vellum beta

orangejenny:
Basically just re-applies https://github.com/dimagi/commcare-hq/pull/11441

@millerdev 

Won't have any effect until `scripts/vellum-to-hq master --target=vellum_beta --branch=onboarding-redux` is run.


orangejenny:
Also note that the `onboarding-redux` code will be live (for people with the flag on) once this is merged and `vellum-to-hq` is run, even though https://github.com/dimagi/Vellum/pull/847 hasn't been merged yet.

I think this must have been what we did for the case management in the form builder release; the PR actually gets merged at the point we do the full rollout.

PR #15190, opened 2017-03-01
Atwho items indicator

benrudolph:
@czue @dimagi/ux made some adjustments to address the usability of the combobox in web apps.

first commit adds place holder text:

screen shot 2017-03-01 at 10 18 12 am

second commit adds an indicator of how many items are showing (i also dropped the number of items showing down to 5 from 10 because i didn't like the scroll bar that was being created)

screen shot 2017-03-01 at 10 40 14 am

cc: @snopoke


czue:
nice this looks way more discoverable. thanks!

benrudolph:
adding label because i think this will happen to the atwho dropdown in the debugger

benrudolph:
fixed

orangejenny:
I like the item count indicator (I don't like the scrollbar in atwho, either).

benrudolph:
✔️

PR #15194, opened 2017-03-01
Ability to handle multiple GPS questions

benrudolph:
@biyeun @wpride when you try to load two gps questions at the same time live preview (and web apps) go kaboom. this ensures only one instance of it loads

### DIFF
--- a/corehq/apps/cloudcare/static/cloudcare/js/form_entry/entrycontrols_full.js
+++ b/corehq/apps/cloudcare/static/cloudcare/js/form_entry/entrycontrols_full.js
@@ -614,7 +614,7 @@ function GeoPointEntry(question, options) {
         self.rawAnswer([]);
     };
 
-    window.gMapsCallback = function() {
+    self.gMapsCallback = function() {
         self.geocoder = new google.maps.Geocoder();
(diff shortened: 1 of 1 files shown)


corehq/apps/cloudcare/static/cloudcare/js/form_entry/entrycontrols_full.js
```diff
@@ -626,12 +626,19 @@ function GeoPointEntry(question, options) {
...
     self.afterRender = function() {
-        if (typeof google === "undefined") {
-            $.getScript(self.apiKey + '&callback=gMapsCallback');
+        if (typeof google === "undefined" && !window.gMapsRequested) {
+            // First entry to attempt to load google
+            window.gMapsRequested = true;
+            $.getScript(self.apiKey, self.gMapsCallback);
+        } else if (typeof google === "undefined" && window.gMapsRequested) {
+            // Waiting for gmaps to load, recursively call afterRender
+            setTimeout(self.afterRender, 400);
```
benrudolph: this solution is meh, but any other solution i think is fairly non-trivial unless we want to always load the google api script

benrudolph:
cc: @snopoke

PR #15198, opened 2017-03-01
Increase length for location name to 255

esoergel:
255 is finite and arbitrarily large.  When I ran `./manage.py makemigrations`, there were a bunch of other changes that django pulled in, any idea what that's about?
edit: looks like that is a result of these changes: https://github.com/dimagi/commcare-hq/commit/2e2ec6f5
I thought those wouldn't require a migration, as [`blank` is only a form validation parameter](https://docs.djangoproject.com/en/1.10/ref/models/fields/#django.db.models.Field.blank)?
@gcapalbo @millerdev


corehq/apps/locations/migrations/0007_auto_20170301_1607.py
```diff
@@ -0,0 +1,53 @@
...
+            model_name='sqllocation',
+            name='name',
+            field=models.CharField(max_length=255, null=True),
+        ),
+        migrations.AlterField(
+            model_name='sqllocation',
+            name='supply_point_id',
+            field=models.CharField(blank=True, db_index=True, max_length=255, null=True, unique=True),
+        ),
+    ]
```
esoergel: This results in the following SQL:
```SQL
BEGIN;
--
-- Change managers on sqllocation
--
--
-- Alter field external_id on sqllocation
--
--
-- Alter field latitude on sqllocation
--
--
-- Alter field longitude on sqllocation
--
--
-- Alter field metadata on sqllocation
--
--
-- Alter field name on sqllocation
--
ALTER TABLE "locations_sqllocation" ALTER COLUMN "name" TYPE varchar(255) USING "name"::varchar(255);
--
-- Alter field supply_point_id on sqllocation
--

COMMIT;
```
So I guess it's fine

gcapalbo:
it looks like the extra migrations are coming from this change (adding blank=True):
https://github.com/dimagi/commcare-hq/commit/2e2ec6f579f59c15c93ee331145103a22e771b9f

those should all just be a no-op anyway since the column definition isn't actually changing, just the django metadata about it is. worth running locally though to make sure it doesn't change anything.

it might be worth splitting into two migrations, one to take care of the blank=True stuff, and one to increase the length of name, just so it's clear what each one does.

also in general i like to name the migrations instead of using the django default date/time name

gcapalbo:
looks like we were editing at the same time - great, yeah looks like you confirmed the blank=True thing is a no op for postgres. i think django just updates its metadata with the migration

esoergel:
Good suggestions, updated.

gcapalbo (approved):

PR #15206, opened 2017-03-01
Fix web apps crash

dannyroberts:
https://sentry.io/dimagi/commcarehq/issues/223558134/events/4989578483/


benrudolph (approved):
nice, thanks for fixing!

PR #15208, opened 2017-03-02
Metrics for kafka

benrudolph:
@snopoke setup some metrics for kafka, have any additional ideas for monitoring it?

### DIFF
--- a/corehq/ex-submodules/pillowtop/pillow/interface.py
+++ b/corehq/ex-submodules/pillowtop/pillow/interface.py
@@ -125,7 +125,7 @@ class PillowBase(object):
 
     def _record_checkpoint_in_datadog(self):
         for topic, value in self.get_last_checkpoint_sequence().iteritems():
-            datadog_gauge('commcare.change_feed.topics'.format(topic), value, tags=[
+            datadog_gauge('commcare.change_feed.processed_offsets'.format(topic), value, tags=[
                 'pillow_name:{}'.format(self.get_name()),
                 'topic:{}'.format(topic),
             ])
--- a/corehq/ex-submodules/pillowtop/pillow/interface.py
+++ b/corehq/ex-submodules/pillowtop/pillow/interface.py
@@ -132,12 +132,12 @@ class PillowBase(object):
             assert len(topics) == 1
             sequence = {topics[0]: int(sequence)}
 
+        datadog_counter('commcare.change_feed.change_feed.checkpoint', tags=[
+            'pillow_name:{}'.format(self.get_name()),
+        ])
+
         for topic, value in sequence.iteritems():
             datadog_gauge('commcare.change_feed.processed_offsets'.format(topic), value, tags=[
-                'pillow_name:{}'.format(self.get_name()),
-                'topic:{}'.format(topic),
-            ])
-            datadog_counter('commcare.change_feed.topics.counter'.format(topic), tags=[
                 'pillow_name:{}'.format(self.get_name()),
                 'topic:{}'.format(topic),
             ])
--- a/corehq/ex-submodules/pillowtop/pillow/interface.py
+++ b/corehq/ex-submodules/pillowtop/pillow/interface.py
@@ -150,13 +150,14 @@ class PillowBase(object):
         self.__record_change_metric_in_datadog('commcare.change_feed.changes.exceptions', change)
 
     def __record_change_metric_in_datadog(self, metric, change):
-        datadog_counter(metric, tags=[
-            'datasource:{}'.format(change.metadata.data_source_name),
-            'document_type:{}'.format(change.metadata.document_type),
-            'domain:{}'.format(change.metadata.domain),
-            'is_deletion:{}'.format(change.metadata.is_deletion),
-            'pillow_name:{}'.format(self.get_name()),
-        ])
+        if change.metadata is not None:
+            datadog_counter(metric, tags=[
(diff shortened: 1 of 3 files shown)


corehq/ex-submodules/pillowtop/pillow/interface.py
```diff
@@ -118,6 +123,41 @@ def process_change(self, change):
     def fire_change_processed_event(self, change, context):
         pass
 
+    def _record_checkpoint_in_datadog(self):
+        for topic, value in self.get_last_checkpoint_sequence().iteritems():
```
benrudolph: can we always assume that this is a dict?

benrudolph:
(need to update tags, it's supposed to be an array)

corehq/ex-submodules/pillowtop/pillow/interface.py
snopoke: no, it could be an int for single topic checkpoints

corehq/ex-submodules/pillowtop/pillow/interface.py
```diff
@@ -118,6 +123,41 @@ def process_change(self, change):
...
+            datadog_gauge('commcare.change_feed.topics'.format(topic), points=value, tags={
+                'pillow_name': self.get_name(),
+                'topic': topic,
+            })
+            datadog_counter('commcare.change_feed.topics.counter'.format(topic), tags={
+                'pillow_name': self.get_name(),
+                'topic': topic,
+            })
```
snopoke: Is this just a counter for the number of checkpoints?

corehq/ex-submodules/pillowtop/pillow/interface.py
```diff
@@ -118,6 +123,41 @@ def process_change(self, change):
     def fire_change_processed_event(self, change, context):
         pass
 
+    def _record_checkpoint_in_datadog(self):
+        for topic, value in self.get_last_checkpoint_sequence().iteritems():
+            datadog_gauge('commcare.change_feed.topics'.format(topic), points=value, tags={
+                'pillow_name': self.get_name(),
+                'topic': topic,
+            })
```
snopoke: not sure about this name, how about `processed_offsets`?

corehq/ex-submodules/pillowtop/pillow/interface.py
```diff
@@ -118,6 +123,41 @@ def process_change(self, change):
...
+            datadog_counter('commcare.change_feed.topics.counter'.format(topic), tags={
+                'pillow_name': self.get_name(),
+                'topic': topic,
+            })
+
+        for topic, offset in self.get_change_feed().get_current_offsets().iteritems():
+            datadog_gauge('commcare.change_feed.current_offsets'.format(topic), points=offset, tags={
```
snopoke: Also here, `current_offsets` could be confused with what the pillow is currently at. Maybe `max_offsets`?

corehq/util/datadog/gauges.py
```diff
@@ -1,6 +1,7 @@
+from django.conf import settings
```
dimagimon: F401 'settings' imported but unused

corehq/ex-submodules/pillowtop/pillow/interface.py
benrudolph: ok thought so, will check that case

corehq/ex-submodules/pillowtop/pillow/interface.py
benrudolph: yeah so we can see the rate at which we create checkpoints. maybe not that useful

corehq/ex-submodules/pillowtop/pillow/interface.py
benrudolph: 👍

corehq/ex-submodules/pillowtop/pillow/interface.py
snopoke: Could be useful.

benrudolph:
ok i think ready for re-review whenever you get a chance

corehq/ex-submodules/pillowtop/pillow/interface.py
```diff
@@ -118,6 +124,49 @@ def process_change(self, change):
...
+            topics = change_feed.topics
+            assert len(topics) == 1
+            sequence = {topics[0]: int(sequence)}
+
+        for topic, value in sequence.iteritems():
+            datadog_gauge('commcare.change_feed.processed_offsets'.format(topic), value, tags=[
+                'pillow_name:{}'.format(self.get_name()),
+                'topic:{}'.format(topic),
+            ])
+            datadog_counter('commcare.change_feed.topics.counter'.format(topic), tags=[
```
snopoke: Sorry, this name doesn't mean much to me. How about `change_feed.checkpoint`.

Also is it necessary to split this by topic? Could just have it per pillow.

snopoke (approved):
one nit

corehq/ex-submodules/pillowtop/pillow/interface.py
benrudolph: yes sorry overlooked that, will update. good point will make it per pillow

benrudolph: https://github.com/dimagi/commcare-hq/pull/15208/commits/2d1f39ea61be72af59ce91a8e9620c7d4a678414

PR #15211, opened 2017-03-02
Add arg to get_default_translations

benrudolph:
@dannyroberts assuming this is the right thing to do based on the other calls to it https://manage.dimagi.com/default.asp?248740

cc: @snopoke


dannyroberts (approved):
Thanks, sorry for messing that up

PR #15217, opened 2017-03-02
catch admin_reports_stats_data error rather than 500

dannyroberts:
just a silly validation error that was causing a 500


dannyroberts:
@nickpell probably no need to wait for tests, they wouldn't cover this anyway

PR #15220, opened 2017-03-02
update ucla message handlers

emord:
@gcapalbo https://manage.dimagi.com/default.asp?248775

@benrudolph

### DIFF
--- a/settings.py
+++ b/settings.py
@@ -1476,6 +1476,11 @@ ALLOWED_CUSTOM_CONTENT_HANDLERS = {
     "FRI_SMS_SHIFT": "custom.fri.api.shift_custom_content_handler",
     "FRI_SMS_OFF_DAY": "custom.fri.api.off_day_custom_content_handler",
     "UCLA_MESSAGE_BANK": "custom.ucla.api.ucla_message_bank_content",
+    "UCLA_GENERAL_HEALTH": "custom.ucla.api.general_health_message_bank_content",
+    "UCLA_MENTAL_HEALTH": "custom.ucla.api.mental_health_message_bank_content",
+    "UCLA_SEXUAL_HEALTH": "custom.ucla.api.sexual_health_message_bank_content",
+    "UCLA_MED_ADHERENCE": "custom.ucla.api.med_adherence_message_bank_content",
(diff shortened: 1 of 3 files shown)


settings.py
```diff
@@ -1476,6 +1476,11 @@
     "FRI_SMS_SHIFT": "custom.fri.api.shift_custom_content_handler",
     "FRI_SMS_OFF_DAY": "custom.fri.api.off_day_custom_content_handler",
     "UCLA_MESSAGE_BANK": "custom.ucla.api.ucla_message_bank_content",
+    "UCLA_GENERAL_HEALTH": "custom.ucla.api.general_health_message_bank_content",
+    "UCLA_MENTAL_HEALTH": "custom.ucla.api.mental_health_message_bank_content",
+    "UCLA_SEXUAL_HEALTH": "custom.ucla.api.sexual_health_message_bank_content",
+    "UCLA_MED_ADHERENCE": "custom.ucla.api.med_adherence_message_bank_content",
+    "UCLA_SUBSTANCE_ABUSE": "custom.ucla.api.substance_use_message_bank_content",
```
emord: Should this be standardized to substance_abuse or substance_use ?

settings.py
```diff
@@ -1476,6 +1476,11 @@
     "FRI_SMS_SHIFT": "custom.fri.api.shift_custom_content_handler",
     "FRI_SMS_OFF_DAY": "custom.fri.api.off_day_custom_content_handler",
     "UCLA_MESSAGE_BANK": "custom.ucla.api.ucla_message_bank_content",
```
emord: Your ticket didn't mention keeping this, but I assume we need to while they migrate their app?

settings.py
gcapalbo: sorry, substance_use for both

settings.py
gcapalbo: i should have said, we can get rid of it, it's only in the test phases right now so it's not actually being used

gcapalbo:
looks good to me. could we also add the lookup table name to the error messages at the end of the api.py file under `if len(custom_messages) != 1:`

gcapalbo (approved):

PR #15222, opened 2017-03-03
A couple more test metrics

benrudolph:
(opening to ensure this records on travis correctly)

### DIFF
--- a/docker/run.sh
+++ b/docker/run.sh
@@ -48,6 +48,8 @@ function run_tests() {
     fi
     shift
 
+    echo $TRAVIS_BUILD_ID
+
     now=`date +%s`
     setup $TEST
     delta=$((`date +%s` - $now))
--- a/docker/run.sh
+++ b/docker/run.sh
@@ -59,6 +59,7 @@ function run_tests() {
     delta=$((`date +%s` - $now))
 
     send_metrics_to_datadog "tests" $delta
+    send_counter_metric_to_datadog
 }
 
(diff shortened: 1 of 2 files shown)


snopoke (approved):

benrudolph:
@dimagi/test-team ✔️ should give us a few more dashboards. will tackle the docker setup separately

docker/run.sh
```diff
@@ -52,30 +52,44 @@ function run_tests() {
...
     delta=$((`date +%s` - $now))
 
-    send_metrics_to_datadog "tests" $delta
+    send_timing_metric_to_datadog "tests" $delta
+    send_counter_metric_to_datadog
 }
 
-function send_metrics_to_datadog() {
+function send_timing_metric_to_datadog() {
+    send_metric_to_datadog "travis.timings.$1" $2 "gauge"
```
millerdev: I think there's a way to tell datadog that a metric represents a time. At least some of the riak metrics show up as times (with time units that automatically scale based on the magnitude of the metric) when you put them in a graph. See metric `riak.node_put_fsm_time_95` for example. I'm not sure how to do this, but it would be cool if we could add units to these. Here's the datadog riak integration in case that helps: http://docs.datadoghq.com/integrations/riak/

benrudolph: nice you're right. you have to edit the metadata of the metric in the metric summary page: https://app.datadoghq.com/metric/summary. this could very useful

PR #15223, opened 2017-03-03
Ignore deleted locations

kkrampa:
@gcapalbo 
It seems they deleted some location but transactions remain in db.


gcapalbo (approved):
thanks, build failure was unrelated, i just restarted it

PR #15225, opened 2017-03-03
what if self.existing_report is None? related to fogbugz # 248552

javierwilson:
@benrudolph @snopoke

### DIFF
--- a/corehq/apps/userreports/views.py
+++ b/corehq/apps/userreports/views.py
@@ -578,12 +578,18 @@ class ConfigureChartReport(ReportBuilderView):
             report_form = self.report_form
         except Exception as e:
             self.template_name = 'userreports/report_error.html'
-            error_response = {
-                'report_id': self.existing_report.get_id,
-                'is_static': self.existing_report.is_static,
-                'error_message': '',
-                'details': six.text_type(e)
-            }
+            if self.existing_report is not None:
+                error_response = {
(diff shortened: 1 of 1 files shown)


benrudolph (commented):
nice and when this errors nothing in the template expects to have `report_id` or anything?

corehq/apps/userreports/views.py
```diff
@@ -578,12 +578,18 @@ def page_context(self):
             report_form = self.report_form
         except Exception as e:
             self.template_name = 'userreports/report_error.html'
-            error_response = {
-                'report_id': self.existing_report.get_id,
-                'is_static': self.existing_report.is_static,
-                'error_message': '',
-                'details': six.text_type(e)
-            }
+            if self.existing_report is not None:
```
benrudolph: also as just a small nit pick, this could be slightly DRYer:

```py
error_response = {
    'error_message': '',
    'details': six.text_type(e)
}
if self.existing_report is not None:
    error_response.update({
        'report_id': self.existing_report.get_id,
        'is_static': self.existing_report.is_static
    })
```

javierwilson: You're right. I'll update my branch and commit.

javierwilson:
applied @benrudolph 's suggestion. @snopoke

benrudolph (approved):

PR #15227, opened 2017-03-03
Fix due list missing fixture_columns_by_type

orangejenny:
Caused by https://github.com/dimagi/commcare-hq/pull/15161

@millerdev / @czue


czue (approved):
thanks!

PR #15229, opened 2017-03-03
lower export scroll window size

dannyroberts:
https://manage.dimagi.com/default.asp?248384


benrudolph (approved):
wow interesting. pretty crazy we can't iterate through 1000 forms in 5 minutes

dannyroberts:
Yeah it does seem pretty insane; not at all certain that my understanding of the error message pointing to the conclusion that we're overrunning the timeout for the scroll id is correct, or that that implies that the processing of docs in a scroll window is taking longer than 5 minutes, but that was one obvious logical chain and I thought I'd test it out and see if it fixes the problem

PR #15230, opened 2017-03-03
remove sniffing

emord:
@NoahCarnahan @snopoke 

sounds like this is causing more issues than helping? thread starts here: https://github.com/dimagi/commcare-hq/pull/15124#issuecomment-282340245

I've also noticed some 500s `ERROR: ESError: TransportError(N/A, 'Unable to sniff hosts.')`

@benrudolph @gcapalbo

### DIFF
--- a/corehq/elastic.py
+++ b/corehq/elastic.py
@@ -38,12 +38,7 @@ def get_es_new():
             }
             for host in es_hosts
         ]
-        get_es_new._es_client = Elasticsearch(
-            hosts,
-            sniff_on_start=True,
-            sniff_on_connection_fail=True,
-            sniffer_timeout=60
-        )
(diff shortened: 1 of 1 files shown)


NoahCarnahan:
That was speedy :) Do we need to add all the nodes to prod's localsettings if we make this change?

emord:
https://github.com/dimagi/commcarehq-ansible/blob/2f261754ca034ea5a0dd9e09482d667a75a5b864/ansible/roles/commcarehq/templates/localsettings.py.j2#L308-L312

They should be there already. they look through the env file

NoahCarnahan:
ah, great

NoahCarnahan (approved):

PR #15234, opened 2017-03-04
nikshay patient registration repeater to pick non-test submissions

mkangia:
https://manage.dimagi.com/default.asp?248743

1. Send patients registrations to nikshay only when submitted by non-test location
2. update repeaters to fire only when state is pending or failure. This is needed coz the repeat records are getting fired even in cancelled state. Not the best fix though.

Best reviewed 🐟 / 🐟

### DIFF
--- a/corehq/apps/repeaters/tasks.py
+++ b/corehq/apps/repeaters/tasks.py
@@ -11,7 +11,8 @@ from corehq.apps.repeaters.dbaccessors import iterate_repeat_records
 from corehq.apps.repeaters.const import (
     CHECK_REPEATERS_INTERVAL,
     CHECK_REPEATERS_KEY,
-)
+    RECORD_PENDING_STATE,
+    RECORD_FAILURE_STATE)
 
 logging = get_task_logger(__name__)
 
@@ -60,9 +61,10 @@ def process_repeat_record(repeat_record):
         if repeat_record.repeater.doc_type.endswith(DELETED_SUFFIX):
             if not repeat_record.doc_type.endswith(DELETED_SUFFIX):
                 repeat_record.doc_type += DELETED_SUFFIX
+                repeat_record.save()
         else:
-            repeat_record.fire()
-        repeat_record.save()
+            if repeat_record.state == RECORD_PENDING_STATE or repeat_record.state == RECORD_FAILURE_STATE:
+                repeat_record.fire()
     except Exception:
         logging.exception('Failed to process repeat record: {}'.format(repeat_record._id))
 
--- a/corehq/apps/repeaters/tasks.py
+++ b/corehq/apps/repeaters/tasks.py
@@ -62,8 +62,7 @@ def process_repeat_record(repeat_record):
             if not repeat_record.doc_type.endswith(DELETED_SUFFIX):
                 repeat_record.doc_type += DELETED_SUFFIX
                 repeat_record.save()
-        else:
-            if repeat_record.state == RECORD_PENDING_STATE or repeat_record.state == RECORD_FAILURE_STATE:
+        elif repeat_record.state == RECORD_PENDING_STATE or repeat_record.state == RECORD_FAILURE_STATE:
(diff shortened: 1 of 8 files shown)


custom/enikshay/integrations/nikshay/repeaters.py
```diff
@@ -34,8 +38,21 @@ def allowed_to_forward(self, episode_case):
...
+
+
+def test_submission(episode_case):
+    person_case = get_person_case_from_episode(episode_case.domain, episode_case.get_id)
+    try:
+        phi_location = SQLLocation.objects.get(location_id=person_case.owner_id)
+    except SQLLocation.DoesNotExist:
+        raise NikshayLocationNotFound(
+            "Location with id {location_id} not found. This is the owner for person with id: {person_id}"
+                .format(location_id=person_case.owner_id, person_id=person_case.case_id)
```
dimagimon: E131 continuation line unaligned for hanging indent

corehq/apps/repeaters/tasks.py
```diff
@@ -60,9 +61,10 @@ def process_repeat_record(repeat_record):
         if repeat_record.repeater.doc_type.endswith(DELETED_SUFFIX):
             if not repeat_record.doc_type.endswith(DELETED_SUFFIX):
                 repeat_record.doc_type += DELETED_SUFFIX
+                repeat_record.save()
         else:
-            repeat_record.fire()
-        repeat_record.save()
+            if repeat_record.state == RECORD_PENDING_STATE or repeat_record.state == RECORD_FAILURE_STATE:
```
mkangia: @benrudolph notifying you if you have context on this. I saw that you made some changes to this last. More context [here](https://manage.dimagi.com/default.asp?248743#bugevent_1297570)

benrudolph: seems right not to send when it's been cancelled, i think the cancelled state was added after this logic was added, so makes sense that this doesn't account for it.

nit: can make this a `elif` instead of an `else` then an `if`

benrudolph (approved):
looks good. a test for the cancelled state might be useful

corehq/apps/repeaters/tests/test_repeater.py
```diff
@@ -219,6 +217,39 @@ def test_check_repeat_records(self):
...
+
+        # not trigger records succeeded triggered after cancellation
+        with patch('corehq.apps.repeaters.models.simple_post_with_cached_timeout') as mock_fire:
+            check_repeaters()
+            self.assertEqual(mock_fire.call_count, 0)
+            for repeat_record in RepeatRecord.all():
+                self.assertEqual(repeat_record.state, RECORD_SUCCESS_STATE)
+
+
+    @run_with_all_backends
```
dimagimon: E303 too many blank lines (2)

corehq/apps/repeaters/tasks.py
mkangia: Thank You @benrudolph. Adding test and elif logic.

NoahCarnahan:
This looks good to me, thanks Manish. Why do you say this is "not the best fix though"?

NoahCarnahan:
Ah shoot, just realized you had the "hold off on merge" tag on. Hopefully that was just because you were waiting for Ben and me to weigh in. I think getting this fixed released sooner rather than later would definitely be a big help, as the india server is getting prettied overwhelmed by these.

mkangia:
Thanks @NoahCarnahan for the review. Yup! kept the flag for us to be on the same page first.

> Why do you say this is "not the best fix though"?
Well the method `check_repeaters` would still check all repeat records(pending, failed, cancelled and successful) even though it would just run pending and failed which is an extra overhead. We need to either update the couch view `receiverwrapper/repeat_records_by_next_check` or update how we fetch repeat records.

NoahCarnahan:
Ah got it, yeah I think updating the view to exclude canceled records is probably the right move

PR #15235, opened 2017-03-05
Move where metrics are being recorded

benrudolph:
It seems that the checkpoint doesn't get called too often, so i've moved where we record the current_offsets to be updated every time we process a change. Edit: this might've been erroring more than I had thought which might've caused the metrics not to be recorded


benrudolph:
the diff is pretty weird on this... i think it's because i was just moving around the lines and not changing what they did

PR #15242, opened 2017-03-06
Bump timeout on run_case_update_rules_for_domain serial task

gcapalbo:
@emord @benrudolph

### DIFF
--- a/corehq/apps/data_interfaces/tasks.py
+++ b/corehq/apps/data_interfaces/tasks.py
@@ -74,7 +74,7 @@ def run_case_update_rules(now=None):
 
 @serial_task(
     '{domain}',
-    timeout=24 * 60 * 60,
+    timeout=36 * 60 * 60,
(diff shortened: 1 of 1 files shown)


corehq/apps/data_interfaces/tasks.py
```diff
@@ -74,7 +74,7 @@ def run_case_update_rules(now=None):
 
 @serial_task(
     '{domain}',
-    timeout=24 * 60 * 60,
+    timeout=36 * 60 * 60,
```
snopoke: A bit confused by this. This is the lock timeout so are you saying that this task can run for longer than 24 hours?

gcapalbo: it shouldn't, but in case it runs over i'm trying to make sure we won't spawn another one the next day

emord (approved):

PR #15233, opened 2017-03-03
ICDS Tableau Report Changes

sheelio:
@kkrampa 
I've refactored this code a fair bit.  I don't believe the location restrictions are working correctly.

Is it possible to update the drilldown filter to support location restrictions with the following behaviour:
- If user is assigned to a location, the drilldown filter defaults to that location (but still shows the ancestors)
- In addition, if a user is in a role that is location restricted, they should not be able to choose any locations through the filter that they do not have access to.  This should support the user being assigned to multiple locations

I'm guessing this can be done by updating the API that pulls locations (for the filter) to be location restricted supported, and by defaulting the assigned location to user's primary assigned location.

PR #15243, opened 2017-03-06
Moved restricted location filter to common module

kkrampa:
@esoergel Might be useful here #15233


esoergel:
Cool, sounds good to me.

PR #15244, opened 2017-03-06
Mk nikshay audit changes

mkangia:
https://docs.google.com/document/d/12_38B4LFhdzz7XSGDOKxoreBLis6th_O9r5yxTS18ZE/edit#bookmark=id.hoy96tlo3liu

Handle XSS injection vulnerability by safe saving of name and comments for app, form and module.
I need to test this one last time so keeping the "hold off on merge"

cc: review buddy @nickpell 

Was not sure for review so Added quite some reviewers. Please ignore if that feels right.


dimagimon:
There are 16 errors:

* corehq/apps/style/static/style/js/components/inline_edit.js, line 103 - Missing semicolon. (semi)
* corehq/apps/style/static/style/js/components/inline_edit.js, line 105 - Expected indentation of 28 space characters but found 27. (indent)
Missing semicolon. (semi)
* corehq/apps/style/static/style/js/components/inline_edit.js, line 106 - Expected indentation of 28 space characters but found 27. (indent)
* corehq/apps/style/static/style/js/components/inline_edit.js, line 107 - Expected indentation of 28 space characters but found 27. (indent)
* corehq/apps/style/static/style/js/components/inline_edit.js, line 108 - Expected indentation of 28 space characters but found 27. (indent)
* corehq/apps/style/static/style/js/components/inline_edit.js, line 112 - Expected indentation of 28 space characters but found 27. (indent)
'sanitized_value' is already defined (no-redeclare)
Missing semicolon. (semi)
* corehq/apps/style/static/style/js/components/inline_edit.js, line 113 - Expected indentation of 28 space characters but found 27. (indent)
* corehq/apps/style/static/style/js/components/inline_edit.js, line 114 - Expected indentation of 28 space characters but found 27. (indent)
* corehq/apps/style/static/style/js/components/inline_edit.js, line 115 - Expected indentation of 28 space characters but found 27. (indent)
* corehq/apps/style/static/style/js/components/inline_edit.js, line 119 - Expected indentation of 28 space characters but found 27. (indent)
'sanitized_value' is already defined (no-redeclare)
Missing semicolon. (semi)
* corehq/apps/style/static/style/js/components/inline_edit.js, line 120 - Expected indentation of 28 space characters but found 27. (indent)
* corehq/apps/style/static/style/js/components/inline_edit.js, line 121 - Expected indentation of 28 space characters but found 27. (indent)
* corehq/apps/style/static/style/js/components/inline_edit.js, line 122 - Expected indentation of 28 space characters but found 27. (indent)
* corehq/apps/app_manager/views/apps.py, line 24 - F401 'html_trans' imported but unused
* corehq/apps/app_manager/views/forms.py, line 36 - F401 'html_trans' imported but unused
* corehq/apps/app_manager/views/modules.py, line 22 - F401 'html_trans' imported but unused

corehq/apps/style/static/style/js/components/inline_edit.js
```diff
@@ -98,9 +98,23 @@ hqDefine('style/js/components/inline_edit.js', function() {
                     dataType: 'JSON',
                     data: data,
                     success: function (data) {
+                        function sanitizeValue() {
+                            var sanitize_properties = [".variable-app_name", ".variable-module_name", ".variable-form_name"]
```
dimagimon: Missing semicolon. (semi)

corehq/apps/style/static/style/js/components/inline_edit.js
```diff
@@ -98,9 +98,23 @@ hqDefine('style/js/components/inline_edit.js', function() {
...
+                                    .text();
+                                    self.value(sanitized_value);
+                                    self.readOnlyValue = data.update[sanitize_properties[i]];
+                                    break;
+                                }
+                            }
+                        }
                         self.isSaving(false);
                         self.hasError(false);
+                        sanitizeValue()
```
dimagimon: Missing semicolon. (semi)

corehq/apps/style/static/style/js/components/inline_edit_v2.js
```diff
@@ -98,8 +98,21 @@ hqDefine('style/js/components/inline_edit_v2.js', function() {
                     dataType: 'JSON',
                     data: data,
                     success: function (data) {
+                        function sanitizeValue() {
+                            var sanitize_properties = [".variable-app_name", ".variable-module_name", ".variable-form_name"]
```
dimagimon: Missing semicolon. (semi)

corehq/apps/style/static/style/js/components/inline_edit_v2.js
```diff
@@ -98,8 +98,21 @@ hqDefine('style/js/components/inline_edit_v2.js', function() {
...
+                                    .text();
+                                    self.value(sanitized_value);
+                                    self.readOnlyValue = data.update[sanitize_properties[i]];
+                                    break;
+                                }
+                            }
+                        }
                         self.isSaving(false);
                         self.hasError(false);
+                        sanitizeValue()
```
dimagimon: Missing semicolon. (semi)

corehq/apps/style/static/style/js/components/inline_edit.js
```diff
@@ -98,9 +98,23 @@ hqDefine('style/js/components/inline_edit.js', function() {
                     dataType: 'JSON',
                     data: data,
                     success: function (data) {
+                        function sanitizeValue() {
+                            var sanitize_properties = [".variable-app_name", ".variable-module_name", ".variable-form_name"]
+                            for(var i=0; i<sanitize_properties.length; i++) {
+                                if(data.update.hasOwnProperty(sanitize_properties[i])) {
+                                    var sanitized_value = $("<div/>").html(DOMPurify.sanitize(self.value()));
+                                    .text();
```
dimagimon: Parsing error: Unexpected token .

corehq/apps/style/static/style/js/components/inline_edit_v2.js
```diff
@@ -98,8 +98,21 @@ hqDefine('style/js/components/inline_edit_v2.js', function() {
                     dataType: 'JSON',
                     data: data,
                     success: function (data) {
+                        function sanitizeValue() {
+                            var sanitize_properties = [".variable-app_name", ".variable-module_name", ".variable-form_name"]
+                            for(var i=0; i<sanitize_properties.length; i++) {
+                                if(data.update.hasOwnProperty(sanitize_properties[i])) {
+                                    var sanitized_value = $("<div/>").html(DOMPurify.sanitize(self.value()));
+                                    .text();
```
dimagimon: Parsing error: Unexpected token .

corehq/apps/style/static/style/js/components/inline_edit_v2.js
```diff
@@ -98,8 +98,20 @@ hqDefine('style/js/components/inline_edit_v2.js', function() {
...
+                                    var sanitized_value = $("<div/>").html(DOMPurify.sanitize(self.value())).text();
+                                    self.value(sanitized_value);
+                                    self.readOnlyValue = data.update[sanitize_properties[i]];
+                                    break;
+                                }
+                            }
+                        }
                         self.isSaving(false);
                         self.hasError(false);
+                        sanitizeValue();
```
orangejenny: Why does this need to be re-sanitized? Hasn't `data.update` already been sanitized on the server?

orangejenny:
Hi Manish, is the major weakness here the `updateDOM` function?

If the problem is `updateDOM`, it seems like changing https://github.com/dimagi/commcare-hq/blob/master/corehq/apps/style/static/style/js/main.js#L95 to use `text` instead of `html` would be sufficient to keep the name/comment from being interpreted as html.

If the problem isn't just `updateDOM`, does this also need to update other uses of the inline edit widget, such as build comments on the deploy page? If it does, we should talk more about how to prevent future uses of the inline edit widget from introducing new vulnerabilities.

corehq/apps/style/static/style/js/components/inline_edit_v2.js
mkangia: @orangejenny You are right. I sanitised this on the view first and then on the server so didn't realise i could just pick up the server value now. Updated!

mkangia:
@orangejenny I tried changing the `updateDOM` method to use text which did avoid the html interpretation but that would not sanitise the value on the server. The idea behind this is to escape this on server as well.

we are looking to just handle this on these particular input fields as of now for the security audit. We can look for a more robust side wise option if needed later.

corehq/apps/app_manager/templates/app_manager/v1/form_view_base.html
```diff
@@ -293,7 +293,7 @@
         name: 'comment',
         id:'comment-id',
         readOnlyClass: 'app-comment',
-        value: '{{ form.comment|escapejs }}',
+        value: '{{ form.comment|html_name|escapejs }}',
```
millerdev: curious if/why this was required for the audit?

corehq/apps/app_manager/views/forms.py
```diff
@@ -260,14 +260,15 @@ def should_edit(attribute):
 
     if should_edit("name"):
         name = request.POST['name']
-        form.name[lang] = name
+        form.name[lang] = html_name(name)
```
millerdev: I'm a little surprised it's supported to put HTML in a form name. Also don't understand why we want to put `<span class="label label-info">Empty</span>` in the name if it's empty.

snopoke: Agree that using `html_name` isn't the right thing here. I think you just want `html.strip_tags`

snopoke (changes requested):
I think this still needs some more work.

In general I think the approach should be:
* strip HTML tags from user input on submission using `strip_tags`
* use `html_name` when displaying a field in html where we want the `Empty` text to show if the field is blank
* don't use `html_name` for any field that's rendered with JS or where we don't need to show the `Empty` text e.g. app comments.

snopoke (commented):
Tested this locally and I think the main issue is that the user input isn't sanitized on save. Instead of using `strip_tags` I think it should use `html.escape` or `html.conditional_escape`

mkangia:
> Tested this locally and I think the main issue is that the user input isn't sanitized on save

@snopoke i just tried saving name as `&amp;gt;&lt;img src="https://pbs.twimg.com/profile_images/655066410087940096/QSUlrrlm.png"&gt;` and it was saved as `&gt;` in app. You are saying that this didn't work or you tried some other input?

corehq/apps/app_manager/views/forms.py
mkangia: thank you Daniel and Simon.. html.strip_tags works well. Changing it.

snopoke:
@mkangia I used that same input but then sanitized it with `html.escape` before saving it to the app then it wasn't an issue. 

I did also have to change this line here to get the sanitized version (`app.name`) instead of the raw input: https://github.com/dimagi/commcare-hq/blob/09c8a4222055b585bd2d5c9c7fa41ecab8c7661d/corehq/apps/app_manager/views/apps.py#L682-L682

corehq/apps/app_manager/views/forms.py
mkangia: thank you Daniel and Simon. Updated to use html.strip_tags

corehq/apps/app_manager/views/modules.py
```diff
@@ -4,6 +4,7 @@
 import logging
 from lxml import etree
 
+from django.utils import html
 from django.template.loader import render_to_string
```
dimagimon: F401 'feature_previews' imported but unused

orangejenny:
Offlined with Manish, but commenting here in case anyone else has thoughts.

The security issue is only about `updateDOM`, and is only an issue because `updateDOM` uses `.html(...)` to display the new name. So when I said...

> does this also need to update other uses of the inline edit widget, such as build comments on the deploy page?

...that wasn't relevant, this isn't related to the inline edit widget. The changes to `inline_edit.js` aren't needed.

The `main.js` change is all that's needed to fix the actual vulnerability. I don't understand why the server-side changes are being requested at all. There are hundreds of places HQ accepts user input that could theoretically include an injected script, but it doesn't matter unless that input gets put on the page in a way that interprets it as html. Doing server-side stripping for these few places and nothing for the rest of HQ seems arbitrary and confusing.

millerdev:
> I don't understand why the server-side changes are being requested at all.

Are we in a position where we need to argue with the auditors if we don't want to do this tag stripping?

> There are hundreds of places HQ accepts user input that could theoretically include an injected script, but it doesn't matter unless that input gets put on the page in a way that interprets it as html.

Indeed, Django templates escape HTML tags by default unless the content is explicitly marked safe. Most of the `strip_tags` added in this PR seem unnecessary, maybe even undesirable. Possible exception is the one in [xforms_extras.py](https://github.com/dimagi/commcare-hq/pull/15244/files#diff-2a3821fed4cf521dc09530913d2474efR122) because the string being constructed is wrapped in `mark_safe` later on.

mkangia:
Thank you @orangejenny and @millerdev.

> Indeed, Django templates escape HTML tags by default unless the content is explicitly marked safe. Most of the strip_tags added in this PR seem unnecessary, maybe even undesirable. Possible exception is the one in xforms_extras.py because the string being constructed is wrapped in mark_safe later on.

i have removed any view level parsing and just kept changes to strip tags when storing value.
We might need to add that back for form_extras in case we choose to not make any changes on server or is it good to have anyway? Have removed it for now.

> I don't understand why the server-side changes are being requested at all.
Are we in a position where we need to argue with the auditors if we don't want to do this tag stripping?

We can just do the @orangejenny suggested in updateDOM and not do any server side validation if we are confident that it should be enough and address any followup arguments if raised later.

@orangejenny it seems like updateDOM which is called in inline_edit through `postSave` is updating values from the update object thats sent back by server, but thats not reflecting in html coz in inline_edit the viewModel's value is not updated and is still what was entered by user. Is that intentional? The object's value and its representation in html should be the same?

millerdev:
> i have removed any view level parsing and just kept changes to strip tags when storing value.

Sorry, I did not communicate clearly. You did the opposite of what I had in mind. That is, the `strip_tags` removed in e08682a was the only one that I thought was reasonable to keep, and all the others added in this PR should probably be removed.

orangejenny:
> it seems like updateDOM which is called in inline_edit through postSave is updating values from the update object thats sent back by server, but thats not reflecting in html coz in inline_edit the viewModel's value is not updated and is still what was entered by user

That code just wasn't anticipating that the server call would change the value, and it also can't really make any assumptions about what will be in the data that comes back from the server. I think the best way to deal with this would be to add a condition to the `success` function like
```
if (data.value) {
   self.value(data.value);
}
```
and then have the server pass back a `value` along with the `update`.

mkangia:
Hey @millerdev,

> That is, the strip_tags removed in e08682a was the only one that I thought was reasonable to keep, and all the others added in this PR should probably be removed.

Just to confirm, that call was added to strip tags for module and form name when server side changes were not done. I removed that now because it is not needed any more since we are already stripping tags when saving the value. Any other reason for this to still be added?

mkangia:
Thank You @orangejenny.
Checked it more closely now. `updateDOM` is updating the module and form name in the side bar and in top nav bar which shows path to module/form `Application > Module Name`.
So changes in inline_edit take care of the edit field in the form by updating with server value and the changes in updateDOM would now update the text instead of html taking care of XSS.

Seeing now why you said inline_edit changes are not needed. I don't mind removing them, but i think they still add value by making the value in backend same as whats visible in view eventually after striping tags in nav bar and side bar.
Also, having tags stripped in nav bar and side bar using method `html_trans` and edit field without it when rendering results in different values which is confusing in the view which is also taken care of by the changes.
But overall i think this is for the bigger team to decide as a practice.

corehq/apps/app_manager/views/apps.py
```diff
@@ -678,7 +679,8 @@ def should_edit(attribute):
 
     if should_edit("name"):
         clear_app_cache(request, domain)
-        name = hq_settings['name']
+        name = html.strip_tags(hq_settings['name'])
+        app.name = name
```
snopoke: This is wrong, `app.name` is already getting set above on line 677/8. I also don't think you should use `strip_tags`, if we're going to sanitize this at all on the server side (seems like there's some feeling that we shouldn't) then we should use `html.escape` instead.

The same thing goes for the other places that use `html.strip_tags`

snopoke (commented):
reviewing this again I think I'm in agreement with @orangejenny and @millerdev. I don't think the server side changes or the changes to `inline_edit` are necessary or desirable. The only issue is `updateDom`.

millerdev:
> Just to confirm, that call was added to strip tags for module and form name when server side changes were not done. I removed that now because it is not needed any more since we are already stripping tags when saving the value. Any other reason for this to still be added?

I'm going to step back and talk about theory to hopefully help us get on the same page. We should not use `strip_tags` or any other function to attempt to sanitize user input by removing parts of it. Instead, we should store exactly what the user submits. We should always escape it properly when we pass it to a service or save it to a database or display it back to them so it does not allow a code injection attack, either on our server or in a browser.

mkangia:
Removed all changes and just kept the change for updateDOM. 🤕

millerdev (approved):

snopoke (approved):

snopoke:
Thanks for persevering on this @mkangia

PR #15249, opened 2017-03-06
Some visual bug fixes to v2 & Preview

biyeun:
- attempts to address the scrollbar issue on windows through css-based scrollbar hiding magic
- fixes the `import_app` view in V2
- fixes the `panel` style in app preview
- fixes the Complete button style at the end of oqps in live preview so that it's not taller than the nav bar.

buddy @esoergel

### DIFF
--- a/corehq/apps/app_manager/templates/app_manager/v2/apps_base.html
+++ b/corehq/apps/app_manager/templates/app_manager/v2/apps_base.html
@@ -83,11 +83,6 @@ appmanager-main-container{% if formdesigner %} formdesigner-content-wrapper{% en
     <script src="{% static 'hqwebapp/js/lib/jquery.textchange.min.js' %}"></script>
 
     <script src="{% static 'hqwebapp/js/stay-on-tab.js' %}"></script>
-    {% if not is_appmanager_admin %}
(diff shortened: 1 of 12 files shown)


corehq/apps/app_manager/templates/app_manager/v2/apps_base.html
```diff
@@ -83,8 +83,10 @@
     <script src="{% static 'hqwebapp/js/lib/jquery.textchange.min.js' %}"></script>
 
     <script src="{% static 'hqwebapp/js/stay-on-tab.js' %}"></script>
+    {% if not is_appmanager_admin %}
```
orangejenny: How about moving these two scripts to `managed_app.html` instead of using this flag?

orangejenny (changes requested):

corehq/apps/app_manager/templates/app_manager/v2/apps_base.html
biyeun: sounds like a good plan

biyeun:
@orangejenny updates made

orangejenny:
thanks!

orangejenny (approved):

PR #15250, opened 2017-03-06
use add_arguments

nickpell:
@esoergel


esoergel (approved):

esoergel:
@nickpell note that this runs on deploy, so I'd want to make sure you've run it as is done in deploy, otherwise I could see it failing silently.

PR #15258, opened 2017-03-06
App manager v2 copy updates

orangejenny:
@biyeun / @krsdimagi 

How's this? I think that including info about the app properties panel is too much for a first-time intro, and it's hard to write something that's both new-user-friendly and accurate (it isn't only properties about the case, it also contains usercase info and will eventually contain more), so I dropped that part.

cc @millerdev


millerdev (approved):

biyeun:
i like how you worded this from the user's perspective 👍

biyeun (approved):

biyeun:
@orangejenny don't forget to rebuild the diffs

orangejenny:
@biyeun thanks, updated

PR #15260, opened 2017-03-06
don't import from django.core.urlresolvers

nickpell:
get rid of warning:

```
RemovedInDjango20Warning: Importing from django.core.urlresolvers is deprecated in favor of django.urls.
  from django.core.urlresolvers import reverse as _reverse
```

@dannyroberts


dannyroberts (approved):

dannyroberts:
any idea what the build error is about?

nickpell:
I jumped the gun and should wait until we're on django 1.10 before merging this PR

PR #14706, opened 2017-01-31
Infinite repeat mods

kmbrgandhi:
Added a cancel option on the Data Forwarding Records page for pending or failed payloads, as well as an automatic cancellation after 3 retries.  Added a requeue option for cancelled payloads.

PR #15262, opened 2017-03-06
Cancel repeat record

dannyroberts:
Repeat records get fired in a few steps

- non-succeeded records end up in the `receiverwrapper/repeat_records_by_next_check` view
- that view gets polled every 5 minutes by the `check_repeaters` task
- that task queues a `process_repeat_record` task for each record

Towards the end of January, a feature was introduced to cancel repeat records. This was done by setting a (new) `cancelled` property on the record to true and nulling the `next_check`. `receiverwrapper/repeat_records_by_next_check` however did not check this new property, nor did it actually check to filter out records with null `next_check`s, so these cancelled records still made it into the records-to-be-queued view, and thus into the queue. It appears there are currently 250k records with `succeeded` set to false, but with a null `next_check`. Every time `check_repeaters` is called, it has to make its way through these 250k records before queuing any actual records. My hypothesis is that these records remain in the queue indefinitely and thus as the number grows these records have been taking more and more time to get through, to the point that eventually we can barely queue any real records in the 5 minute time slot.

This PR fixes this by removing them from the records-to-be-queued view, `receiverwrapper/repeat_records_by_next_check`.

EDIT: At the time I wrote this PR message, I was conflating two things, namely the existance of the `cancelled` property and the existence of the "Cancel" button. In fact, the `cancelled` property has been around a long time, and is the mechanism for stopping retries after trying on 3 different occasions. So, while this was the fix, I have no reason to believe it was caused by the introduction of the "Cancel" button.

EDIT 2: Actually it seems like my last edit was incorrect and the recent commit https://github.com/dimagi/commcare-hq/pull/14706 did in fact introduce `cancelled` and `cancel()`—but since it didn't actually have the effect of getting them out of the to-be-queued queue, it resulted in the buildup we saw.

### DIFF
--- a/corehq/apps/repeaters/tests/test_dbaccessors.py
+++ b/corehq/apps/repeaters/tests/test_dbaccessors.py
@@ -27,25 +27,25 @@ class TestRepeatRecordDBAccessors(TestCase):
             domain=cls.domain,
             failure_reason='Some python error',
             repeater_id=cls.repeater_id,
-            next_event=before,
+            next_check=before,
         )
         success = RepeatRecord(
             domain=cls.domain,
             succeeded=True,
             repeater_id=cls.repeater_id,
-            next_event=before,
+            next_check=before,
         )
         pending = RepeatRecord(
             domain=cls.domain,
             succeeded=False,
             repeater_id=cls.repeater_id,
-            next_event=before,
+            next_check=before,
         )
         other_id = RepeatRecord(
             domain=cls.domain,
             succeeded=False,
             repeater_id=cls.other_id,
-            next_event=before,
+            next_check=before,
         )
 
         cls.records = [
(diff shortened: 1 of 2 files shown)


emord (approved):

dannyroberts:
```
FAIL: test_iterate_repeat_records (corehq.apps.repeaters.tests.test_dbaccessors.TestRepeatRecordDBAccessors)
  File "/mnt/commcare-hq-ro/corehq/apps/repeaters/tests/test_dbaccessors.py", line 100, in test_iterate_repeat_records
    self.assertEqual(len(records), 3)  # Should grab all but the succeeded one
AssertionError: 0 != 3
```

dannyroberts:
I fixed the failing test, and repeaters tests pass. Given that it seems unlikely that a test in a different app would fail, and that this is bug is actively causing problems, I'm going to merge now and deploy to production and india.

NoahCarnahan:
FYI @mkangia it looks like this takes of the remaining todo from this pr: https://github.com/dimagi/commcare-hq/pull/15234

PR #15264, opened 2017-03-06
fix html syntax in es translation

dannyroberts:
was causing href string to be many lines long due to missing quote

fixes: https://manage.dimagi.com/default.asp?248926


dannyroberts:
(killed the build out of courtesy. not relevent here)

PR #15265, opened 2017-03-07
fix quote escaping in es translations

dannyroberts:
followup to https://github.com/dimagi/commcare-hq/pull/15264


dannyroberts:
@nickpell FYI merging to deploy

PR #15273, opened 2017-03-07
misc accounting test cleanup

nickpell:
@dannyroberts

### DIFF
--- a/corehq/apps/accounting/tests/base_tests.py
+++ b/corehq/apps/accounting/tests/base_tests.py
@@ -3,7 +3,6 @@ from django.test import TestCase
 from django_prbac.models import Role
 
 from corehq.apps.accounting.tests import generator
-from corehq.apps.domain.models import Domain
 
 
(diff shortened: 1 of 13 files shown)


dannyroberts:
not sure what's going on here
```
ERROR: test_migration (corehq.apps.tzmigration.tests.test_timezonemigration.TimeZoneMigrationTest)
  File "/mnt/commcare-hq-ro/corehq/apps/tzmigration/tests/test_timezonemigration.py", line 94, in test_migration
    xform_instance, = get_forms_by_type(self.domain, 'XFormInstance', limit=10)
ValueError: too many values to unpack
```

nickpell:
my theory is some test is not cleaning up a domain it's created, and removing the step to delete all domains from ```tearDownClass``` is causing some later test to fail.  I might try to track down the domain, or just delete all domains at the start of the failing test (I don't think the domain is coming from an accounting test so I don't feel terrible about punting)

nickpell:
✅

PR #15277, opened 2017-03-07
fix ShortFormRepeaters

dannyroberts:
which used to break on SQL forms


dannyroberts:
@nickpell cc: @NoahCarnahan

NoahCarnahan (approved):
Looks good to me. Thanks for the test.

PR #15278, opened 2017-03-07
catch schema error

emord:
https://manage.dimagi.com/default.asp?248971

This'll provide a slightly better message when someone has an error in their app that prevents the case properties from being displayed. Happy to hear any thoughts on better wording

@millerdev @gcapalbo @benrudolph

### DIFF
--- a/corehq/apps/app_manager/views/formdesigner.py
+++ b/corehq/apps/app_manager/views/formdesigner.py
@@ -9,8 +9,13 @@ from django.shortcuts import render
 from django.views.decorators.http import require_GET
 from django.conf import settings
 from django.contrib import messages
+
+from dimagi.utils.logging import notify_exception
+
 from corehq.apps.app_manager.views.apps import get_apps_base_context
 from corehq.apps.app_manager.views.notifications import get_facility_for_form, notify_form_opened
+
+from corehq.apps.app_manager.exceptions import AppManagerException
 
 from corehq.apps.app_manager.views.utils import back_to_main, bail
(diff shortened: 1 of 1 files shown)


corehq/apps/app_manager/views/formdesigner.py
```diff
@@ -203,8 +208,14 @@ def get_form_data_schema(request, domain, form_unique_id):
         data.append(get_session_schema(form))
         if form.requires_case() or is_usercase_in_use(domain):
             data.append(get_casedb_schema(form))
-    except Exception:
-        logger.exception("schema error")
+    except AppManagerException as e:
+        notify_exception(request, message=e.message)
+        return HttpResponseBadRequest(_(
+            "There is an error in the case management of your application. "
+            "Please fix the error to see case properties in this tree"
```
millerdev: Will it be obvious what that error is and how it should be fixed? I do like this much better than what we had before, but I wonder if it's still likely to wind up reported on fogbugz or field-dev?

Also, what do you think of including the text of the exception message in the error, both here and below?

millerdev (approved):
Approving because it's definitely better than what we had before, but see my other suggestion in the comments.

corehq/apps/app_manager/views/formdesigner.py
emord: I think that's a great idea, but it will require a vellum change. It's only set up to have text come back instead of html. This would be fine for this error (which is just text without a link) but there are some errors that come back that would include `<a href='your form'>form 1</a>` which would not be very helpful

PR #15285, opened 2017-03-07
use file object instead of payload

emord:
http://manage.dimagi.com/default.asp?248922#1299064

blob mixin can handle a file or string https://github.com/dimagi/commcare-hq/blob/je/efficient-exports/corehq/blobs/mixin.py#L83

it uses put_attachment to set the payload, but now it's using a string (that's what `self.file.payload` is https://github.com/dimagi/commcare-hq/blob/je/efficient-exports/corehq/ex-submodules/couchexport/files.py#L24)

This should stop loading the entire file in memory which i think could be pretty large for xls 

exports @benrudolph blobdb @millerdev 
buddy @gcapalbo

### DIFF
--- a/corehq/apps/export/export.py
+++ b/corehq/apps/export/export.py
@@ -32,7 +32,7 @@ class ExportFile(object):
         self.format = format
 
     def __enter__(self):
-        return self.file.payload
+        return self.file.file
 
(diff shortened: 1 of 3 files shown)


benrudolph (approved):
nice catch 💪

millerdev (approved):

PR #15286, opened 2017-03-08
exit early in repeat queue

dannyroberts:
if a repeat record should not be fired

There are a lot of things in the queue from before that are cancelled=True or have already gone over their retry limit but for some reason aren't cancelled.


corehq/apps/repeaters/tasks.py
```diff
@@ -60,6 +60,12 @@ def check_repeaters():
 
 @task(queue=settings.CELERY_REPEAT_RECORD_QUEUE)
 def process_repeat_record(repeat_record):
+    if repeat_record.state == RECORD_FAILURE_STATE and repeat_record.overall_tries >= repeat_record.max_possible_tries:
```
dimagimon: E501 line too long (119 > 115 characters)

PR #15288, opened 2017-03-08
do not put inherit block in if statement

benrudolph:
@dannyroberts @biyeun this more closely matches how we do this in `style/base.html` don't know if it's the fix. i was only able to test compress locally (not staging)

### DIFF
--- a/corehq/apps/cloudcare/templates/landing_page/base.html
+++ b/corehq/apps/cloudcare/templates/landing_page/base.html
@@ -3,18 +3,23 @@
 {% load compress %}
 {% load statici18n %}
 
-{% block less_debug %}
+{% block css %}
+
+{% if less_debug %}
 <link type="text/less"
       rel="stylesheet"
       media="all"
-      href="{% static 'cloudcare/less/formplayer-webapp.debug.less' %}" />
-{% endblock %}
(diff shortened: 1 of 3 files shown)


corehq/apps/cloudcare/templates/landing_page/base.html
```diff
@@ -3,18 +3,23 @@
...
 <link type="text/less"
       rel="stylesheet"
       media="all"
-      href="{% static 'cloudcare/less/formplayer-webapp.debug.less' %}" />
-{% endblock %}
+      href="{% static 'cloudcare/less/formplayer-webapp.debug.less' %}"/>
 
-{% block css %}
+{% else %}
+{% compress css %}
```
benrudolph: this page is only behind a feature flag that no one uses, but it was missing a compress tag

dannyroberts:
I guess this is worth a shot? If we find out what's causing it, we probably also want to make whatever thing causes a compress block not to be run fail really loudly.

dannyroberts (approved):

biyeun:
hmmmm.... did this actually fix the issue?

dannyroberts:
looks like it!

benrudolph:
magic...

PR #15289, opened 2017-03-08
Add need_processing metric

benrudolph:
@snopoke @dannyroberts this ends up being annoying to calculate and graph properly in datadog, so just creating another metric. also i realize i might be confusing datadog with this. do you think i need to add the topic name in the metric name? for example, if sms is 10 behind and form is 20 behind, it will record 10, and 20 for the same metric. i'm wondering if that would mess with numbers. i think as long as we disaggregate the number it should be ok

cc: @biyeun

### DIFF
--- a/corehq/ex-submodules/pillowtop/pillow/interface.py
+++ b/corehq/ex-submodules/pillowtop/pillow/interface.py
@@ -6,6 +6,7 @@ from corehq.util.soft_assert import soft_assert
 from corehq.util.datadog.gauges import datadog_counter, datadog_gauge
 from dimagi.utils.logging import notify_exception
 from pillowtop.const import CHECKPOINT_MIN_WAIT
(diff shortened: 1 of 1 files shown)


snopoke:
not sure if this will confuse DD, seems ok to try. Will this work with couch sequences which aren't int's? Might need 'force_to_int'

benrudolph:
👍 will check

benrudolph:
Using force_seq_int in the normalize sequence call: https://github.com/dimagi/commcare-hq/pull/15289/commits/c627fb2a7eeb9e952311e9b43e484ed600be851c

the current offsets is already forced to an int here: https://github.com/dimagi/commcare-hq/blob/br/add-one-metric/corehq/ex-submodules/pillowtop/feed/couch.py#L34

snopoke (approved):

PR #7957, opened 2015-08-20
reduce gunicorn worker count

dannyroberts:

PR #15290, opened 2017-03-08
gunicorn config tweaks

snopoke:
https://github.com/dimagi/commcare-hq-deploy/pull/195


benrudolph:
we purposely changed this a while back: https://github.com/dimagi/commcare-hq/pull/7957. it was a long time ago, so may be irrelevant, but as i commented before we might want to make this a function of the ram available on the machine. we were previously seeing OOM errors on prod

kaapstorm (commented):
fwiw, `psutil.virtual_memory()` will give you total and free RAM. We could probably use something like 
```
workers = max(multiprocessing.cpu_count() * 2, (psutil.virtual_memory().total - BASE_RAM) / WORKER_RAM)
```

snopoke:
Ah, that's irritating. Maybe we should push some of those settings to the command line which will be easier to configure per env.

snopoke:
Updated to make #workers configurable per env

kaapstorm (approved):

benrudolph (approved):
nice

PR #15297, opened 2017-03-08
ignore forms that are errors

emord:
http://manage.dimagi.com/default.asp?249103#1299502

Form was an errored form that caused it to not have an attachment associated with it (https://github.com/dimagi/commcare-hq/blob/master/corehq/form_processor/models.py#L262)

An errored form shouldn't ever have an effect on cases so we should just skip them

@gcapalbo

### DIFF
--- a/corehq/apps/users/tasks.py
+++ b/corehq/apps/users/tasks.py
@@ -58,7 +58,7 @@ def tag_forms_as_deleted_rebuild_associated_cases(user_id, domain, form_id_list,
     cases_to_rebuild = set()
 
     for form in FormAccessors(domain).iter_forms(form_id_list):
-        if form.domain != domain or form.is_error:
+        if form.domain != domain or not form.is_normal:
             continue
 
         # rebuild all cases anyways since we don't know if this has run or not if the task was killed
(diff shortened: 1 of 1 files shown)


corehq/apps/users/tasks.py
```diff
@@ -58,7 +58,7 @@ def tag_forms_as_deleted_rebuild_associated_cases(user_id, domain, form_id_list,
     cases_to_rebuild = set()
 
     for form in FormAccessors(domain).iter_forms(form_id_list):
-        if form.domain != domain:
+        if form.domain != domain or form.is_error:
```
snopoke: I wonder if we should exclude all forms except 'normal' ones i.e. errors, duplicates, archived, deprecated etc.

emord: Yeah. I had thought about that, but I wasn't sure if anything not normal could change the case. seemed safer to just remove errored forms from this

emord: Although if you think it's safe to use `not form.is_normal` i'm happy to change it

gcapalbo (approved):

corehq/apps/users/tasks.py
snopoke: Yea, I that would be fine. None of the other's affect cases.

snopoke (approved):

PR #15300, opened 2017-03-08
show stack trace when downgrading logos errors

nickpell:
this has been an issue a few times, and has been difficult to track down without a stack trace

@mkangia


nickpell:
bump

PR #15301, opened 2017-03-08
slightly better error message

emord:
http://manage.dimagi.com/default.asp?248241#1299526

A better fix would be to search through headers without case, but I think that's a larger refactor

@gcapalbo

### DIFF
--- a/corehq/apps/locations/bulk_management.py
+++ b/corehq/apps/locations/bulk_management.py
@@ -10,7 +10,7 @@ from collections import Counter, defaultdict
 
 from django.core.exceptions import ValidationError
 from django.db import transaction
-from django.utils.translation import ugettext as _
+from django.utils.translation import string_concat, ugettext as _, ugettext_lazy
 
 from dimagi.utils.decorators.memoized import memoized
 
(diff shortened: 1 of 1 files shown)


corehq/apps/locations/bulk_management.py
```diff
@@ -327,13 +327,16 @@ def validate_and_parse_stubs_from_excel(self):
...
+            message = (u"'types' sheet should contain exactly '{expected}' as the sheet headers. "
+                       "'{missing}' are missing")
+            if actual - expected:
+                message = message + " '{extra}' are not recognized"
+            message = message.format(
+                expected=", ".join(expected),
+                missing=", ".join(expected - actual),
+                extra=", ".join(actual - expected),
             )
+            raise LocationExcelSheetError(_(message))
```
gcapalbo: if we wanted to translate, we'd have to do so before formatting the params

emord: Thanks. I updated it to use lazy strings that way makemessages actually picks up these strings to translate

gcapalbo: :+1: thanks

gcapalbo (approved):

PR #15306, opened 2017-03-08
Smartly order export columns

benrudolph:
@NoahCarnahan this was kind of annoying ticket that i've been sitting on https://manage.dimagi.com/default.asp?248514. essentially the user has changed their form so much that the columns appear out of order when they create an export (by default we order by however the schema is ordered). this makes it so we re order the schema to match whatever the current form looks like

cc: @biyeun

### DIFF
--- a/corehq/apps/export/models/new.py
+++ b/corehq/apps/export/models/new.py
@@ -1344,7 +1344,11 @@ class ExportDataSchema(Document):
         if inferred_schema:
             current_schema = cls._merge_schemas(current_schema, inferred_schema)
 
-        current_schema = cls._reorder_schema_from_app(current_schema, app_id, identifier)
+        try:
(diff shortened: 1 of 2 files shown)


corehq/apps/export/models/new.py
```diff
@@ -1341,6 +1342,8 @@ def generate_schema_from_builds(cls, domain, app_id, identifier, force_rebuild=F
         if inferred_schema:
             current_schema = cls._merge_schemas(current_schema, inferred_schema)
 
+        current_schema = cls._reorder_schema_from_app(current_schema, app_id, identifier)
```
benrudolph: we could probably be smarter about saving the fetched app/schema during when we process the apps, but this logic is cleaner and simpler so i opted for it. now since we only process the most recent app, generating an export isn't as slow as it used to be

NoahCarnahan (approved):
Nice. All looks good.

PR #15308, opened 2017-03-09
Split off Locations Permissions

esoergel:
Currently location pages check the `can_edit_commcare_users` permission.  This feature switches location pages to a new permission - `can_edit_locations`.  This also includes a migration which will default all existing UserRoles' location permission to whatever their commcare_users permission is (thus preserving the current behavior).
@millerdev


corehq/apps/users/migrations/0001_add_location_permission.py
```diff
@@ -0,0 +1,21 @@
...
+    update_all_roles()
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+    ]
+
+    operations = [
+        HqRunPython(migrate),
```
emord: Do you think it would be hard to add `edit_locations` to the Permissions schema and a management command in 1 PR, then run the management command in a shell and then have a second PR that adds the permission checking?

esoergel: No, I don't think that'd be especially hard.  The main reason I did it like this was so it'd happen automatically to (for instance) swiss, NIC, local envs, and third party instances.

corehq/apps/locations/permissions.py
```diff
@@ -103,11 +104,16 @@
...
-    Decorator controlling domain-level access to locations.
-    """
-    return login_and_domain_required(
+    """Decorator controlling domain-level access to locations features."""
+    return (login_and_domain_required(
         requires_privilege_raise404(privileges.LOCATIONS)(view_fn)
+    ))
+
+
+def require_can_edit_locations(view_fn):
```
millerdev: looking at the symmetry between this and `locations_access_required`. First thought is is it too late to change `locations_access_required` to `require_can_view_locations`?

It's not immediately clear to me if the word _access_ implies _view_ or _edit_ permissions (although I would guess _view_ given the presence of `require_can_edit_locations`).

corehq/apps/users/migrations/0001_add_location_permission.py
millerdev: When and where will the management command be used? Am I correct in thinking that this migration will have already been run by the time the management command can be run in a production environment?

millerdev (approved):
No blocking concerns, so I think fine to merge unless you want to address comments in this PR.

corehq/apps/users/migrations/0001_add_location_permission.py
emord: Yeah. it seems reasonable to not want to run this on a bunch of different places (especially NIC)

Was just thinking that this migration isn't something that will need to be re-run on every new environment/test and this was one of the things that slowed down accounting so hard. I suppose we could change this to a no-op migration after it's been run everywhere

corehq/apps/locations/permissions.py
esoergel: I agree it's a little confusing.  `locations_access_required` checks whether the domain can use locations (is it available on their plan?).  `require_can_edit_locations` checks whether the user has permission to create, update, and delete locations.  I pulled that name off of `require_can_edit_commcare_users`, `require_can_edit_web_users`, and a few others.

The former isn't really view _or_ edit access - it doesn't refer to individual locations, but to the section of the site concerning locations.

corehq/apps/users/migrations/0001_add_location_permission.py
esoergel: The management command won't be run anywhere, actually.  I initially implemented it that way to be able to run it separately, but then realized it'd need to run as a migration.  Ideally I'd be able to run it from a setup_release before the deploy, then have the migration only run if the command hasn't been run manually already.  This would mean setting a flag somewhere in the command and checking for it in the migration.  That started to feel a little unwieldly, and there are only 93k relevant docs on prod, so I expect it to be quick enough not to matter.

Given that, this should probably all be in the migration at this point, but I also agree with Emord that this should noop after a while.  I think what I'd like to do is merge this as is, then open a PR to make it a noop, and merge that after a couple weeks.  Do y'all have thoughts on the best way to noop this?  Make the migration simply `return None`?  Use `CCHQ_IS_FRESH_INSTALL=1` (but that'd leave this code around forever)?  Can I delete the migration without messing up django's migration bookkeeping?

emord: was thinking like this https://github.com/dimagi/commcare-hq/blob/8667a4159ad39d3307f92a3812d74006d610fd1b/corehq/form_processor/migrations/0037_get_form_by_id_fn.py, but that seems to be mostly for sharded dbs

I would say you could squash it but this doesn't have any other migrations. I wouldn't delete it in case someone makes a migration here later with the same name.

Maybe just an empty operations array? Not too worried about this because I don't think it'll grow quite to what accounting was

esoergel: Oh nice, that sounds perfect.  I'll merge this now then and get that in PR, thanks!

esoergel: https://github.com/dimagi/commcare-hq/pull/15320

PR #15309, opened 2017-03-09
a better pillowtop status page for system info

emord:
We're at a point where having colors based on the percentage of changes gone through means nothing. 2% of the commcarehq feed would be 3.5 million changes behind which is pretty absurd. 

Would like to get rid of the progress bar, but not sure of a better way to have a quick look at the page to check if anything is up

Also adding the # changes behind on a feed after the raw numbers

@gcapalbo @dannyroberts


benrudolph (approved):
nice, been struggling with the progress bar thing in datadog as well

PR #15310, opened 2017-03-09
App preview settings for all

benrudolph:
@dimagi/ux this enables settings for everyone. actually kind of excited about this because it means people can change the language in their application


biyeun (approved):

orangejenny:
nice!

PR #15311, opened 2017-03-09
Hide case detail modal on back

benrudolph:
@biyeun this hides the case detail modal when you hit back http://manage.dimagi.com/default.asp?248051


benrudolph:
✅

biyeun:
🎉

PR #15315, opened 2017-03-09
minor issue: cancel records for deleted repeaters

dannyroberts:
there were only a handful of these in the queue


dannyroberts:
@orangejenny

PR #15316, opened 2017-03-09
put a max height on case list table icons

biyeun:
should hopefully fix this issue

https://manage.dimagi.com/default.asp?248334#1299884

@orangejenny 
buddy @benrudolph


benrudolph (approved):
lol at that bug report

PR #15317, opened 2017-03-09
combine pillows

emord:
@benrudolph @dannyroberts what do you think of this approach? General idea is you can have multiple processors in one pillow. it still requires only one doc type per pillow, but i was able to get intrahealth custom pillow from 7 to 2

This would also need a change in the deploy directory to include the new pillow

buddy @gcapalbo


corehq/ex-submodules/fluff/pillow.py
```diff
@@ -153,11 +152,30 @@ def get_fluff_pillow(indicator_class, delete_filtered=False):
...
+        for cls in indicator_classes
+    ]
+    domains = list(set(d for cls in indicator_classes for d in cls.domains))
+    doc_types = list(set(cls.document_class._doc_type for cls in indicator_classes))
+    assert len(doc_types) == 1
+
+    return FluffPillow(
+        indicator_name=name,
+        kafka_topic=kafka_topic,
+        processor=processors,
```
benrudolph: nit: may want to rename to `processors` (though might be more work than it's worth)

corehq/ex-submodules/pillowtop/pillow/interface.py
```diff
@@ -244,7 +248,8 @@ def get_change_feed(self):
         return self._change_feed
 
     def process_change(self, change):
-        self._processor.process_change(self, change)
+        for processor in self._processors:
```
benrudolph: 🙌

benrudolph (approved):
this is awesome and way more concise than i expected!

emord:
Hmm. something in the custom intrahealth tests must not be cleaning up properly. This still contains a lot of testing cleanup and combines 5 m4change pillows so should be good to merge and i'll look at the intrahealth tests for another PR

benrudolph (approved):
nice, looks good to me

corehq/ex-submodules/fluff/pillow.py
```diff
@@ -153,11 +152,30 @@ def get_fluff_pillow(indicator_class, delete_filtered=False):
...
     )
 
 
+def get_multi_fluff_pillow(indicator_classes, name, kafka_topic, delete_filtered=False):
+    processors = [
+        FluffPillowProcessor(cls, delete_filtered=delete_filtered)
+        for cls in indicator_classes
+    ]
+    domains = list(set(d for cls in indicator_classes for d in cls.domains))
+    doc_types = list(set(cls.document_class._doc_type for cls in indicator_classes))
```
snopoke: nit: could use set literal `list({a for a in 'abc'})`

snopoke:
missed this PR in the chaos - nice work.

PR #15319, opened 2017-03-09
More hqdefine

orangejenny:
Adding the last few missing `hqDefine` calls to the recent externalizing js PRs.

`w=1`

@millerdev / @dannyroberts

### DIFF
--- a/corehq/apps/app_manager/static/app_manager/js/source_files.js
+++ b/corehq/apps/app_manager/static/app_manager/js/source_files.js
@@ -1,37 +1,39 @@
-$(function(){
-    $('.toggle-next').click(function(e){
-        e.preventDefault();
-        $(this).parents('tr').next('tr').toggleClass("hide");
-    });
-
-    var current_version = hqImport('hqwebapp/js/initial_page_data.js').get('current_version'),
-        built_versions = hqImport('hqwebapp/js/initial_page_data.js').get('built_versions'),
-        $form = $("#compare-form"),
-        $input = $form.find("input");
-
-    built_versions = _.sortBy(_.filter(built_versions, function(v) {
-        return v.version != current_version;
-    }), function(v) { return parseInt(v.version); }).reverse();
-    version_map = _.indexBy(built_versions, 'version');
-
-    $input.select2({
-        data: _.map(built_versions, function(v) {
-            return {
-                id: v.version,
-                text: v.version + ": " + (v.comment || "no comment"),
-            };
-        }),
-    });
-
-    $form.find("button").click(function() {
-        var version = $input.val();
-        if (!version) {
-            alert("Please enter a version to compare");
-            return;
-        } else if (!version_map[version]) {
-            alert(version + " is not a valid version");
-            return;
-        }
-        window.location = hqImport('hqwebapp/js/urllib.js').reverse('diff', version_map[version].build_id);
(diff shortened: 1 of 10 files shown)


corehq/apps/app_manager/static/app_manager/js/source_files.js
```diff
@@ -1,37 +1,40 @@
...
-        var version = $input.val();
-        if (!version) {
-            alert("Please enter a version to compare");
-            return;
-        } else if (!version_map[version]) {
-            alert(version + " is not a valid version");
-            return;
-        }
-        window.location = hqImport('hqwebapp/js/urllib.js').reverse('diff', version_map[version].build_id);
+/* globals hqDefine */
```
millerdev: nit: could we add `hqDefine` and `hqRequire` to a _.jshintrc_ file somewhere?

millerdev (approved):
Guess we don't need any `hqRequire` calls anywhere to actually run the code in these modules? I forget how `hqDefine` works...

orangejenny:
There isn't an `hqRequire`, modules get "required" by including the script on the page and are run by the `hqDefine` call.

corehq/apps/app_manager/static/app_manager/js/source_files.js
orangejenny: Going to leave this alone for now. It's annoying, but so much of our js defines and accesses globals willy-nilly, I actually like having this explicit statement.

dannyroberts:
I think @millerdev just mean `hqImport`

dannyroberts:
and to @millerdev's point, I'd also envisioned a convention where `hqDefine` modules don't modify global (DOM and such) state on first run, just expose methods that can be called to do so. If all we're doing is taking a file that contains

```
$(function () {
});
```

and wrapping that in an `hqDefine`, I don't think it really gets us anything—in fact it kind of obfuscates the fact that code that does stuff beyond defining functions is getting run just by importing it.

I'm in favor of one of two options—either (1) leaving files like that be so it's obvious that they're modifiying the dom just by being included or (2) putting them into a named function like `init()` that is returned as part of the module, and letting a controller call it the init function explicitly.

Thoughts?

orangejenny:
Oops, yeah, `hqImport`.

@dannyroberts check out the PR conversation on https://github.com/dimagi/commcare-hq/pull/14874  I've moved quite a bit of DOM-affecting code into `hqDefine` modules for the sake of getting it out of HTML files. I can see an argument for keeping side effects out of the modules themselves...but I'd rather put the controlling code inside the modules, and accept that this means you can't assume that a module doesn't have side effects, rather than decide on a new place for the controller code - inside the same file as the module, but not inside the `hqDefine`? in a separate file that initialize the page? ...Ultimately, I think deciding where controlling code goes will be part of hammering out how we use RequireJS* which I'm deliberately not doing right now. I don't think this state is perfect but do think it's temporary.

* or potentially another module system, though I suspect RequireJS will end up winning

dannyroberts:
Okay, sounds like you've thought about it. I'd been going with the convention that side effects just live outside of modules in the same file, or that they get internalized into an init function called elsewhere. But since I'm barely touching this anymore, I'll let you pick the convention you like!

orangejenny:
Cool. Feedback always welcome, and I expect we'll be having more conversations about this in the future :)

PR #15322, opened 2017-03-10
Put expensive ucrs on a queue to save

emord:
idea i had while about to go to sleep. Not sure that the `is_expensive` condition is the best heuristic but the idea would be to put long calculations (like indicators that need to grab other docs from the db) in a queue to be processed later so that network slowness won't clog up the queue

at the current rate i don't think the main kafka ucr pillow will finish anytime soon (although it may speed up on the weekend)

@dannyroberts


benrudolph (commented):
interesting, this seems like a good idea. some concerns would be if we make any assumptions about the change being processed immediately, for example if the celery task error'd i don't think we would catch it in the pillow retry queue

snopoke:
Another option could be to use gevent to spawn tasks.

benrudolph:
Or have a gevent celery queue 😊 The one hard thing about gevent is that we
lose all the logging and task management that celery gives us

snopoke:
What do you mean?

benrudolph:
if we spawn gevent jobs and the job gets killed or we want to know how many
gevent jobs there are currently spawned, it's hard to know. Celery has all
of that monitoring information built in

snopoke:
Ah right, I thought you were saying if we switched the celery queues to gevent then we'd lose that info.

emord:
> for example if the celery task error'd i don't think we would catch it in the pillow retry queue

This is why I included acks_late and retry on the celery task. http://docs.celeryproject.org/en/latest/faq.html#should-i-use-retry-or-acks-late

I think that this would retry if there's an error

benrudolph:
sweet agree that should work. i do remember there being an issue with `acks_late` a while back. pinging @gcapalbo to see if he remembers if there was an issue or not (sorry if i'm totally misremembering)

snopoke:
is this not going to completely overload celery?

benrudolph:
seems likely, maybe we can put it behind a random toggle?

gcapalbo:
yes, acks_late doesn't really help us much. it turns out it only works when the worker itself crashes, which is a scenario that almost never happens. when we see sigterm and sigsegv errors, that's just the worker child process that gets killed and tasks in those scenarios don't get retried. so basically acks_late doesn't give us much and if we want to truly retry something we have to implement something similar to what the pillow retry process does

emord:
Oops I meant to put retry=True for that task as well. @gcapalbo wouldn't that retry in the event of exceptions?

corehq/apps/userreports/tasks.py
```diff
@@ -171,3 +173,10 @@ def _run_report(backend_to_use):
 
     objects = experiment.run()
     return objects
+
+
+@task(queue=UCR_INDICATOR_CELERY_QUEUE, ignore_result=True, acks_late=True, retry=True)
```
gcapalbo: retry=True is not a valid option to pass a task decorator as far as i know. there is a retry process described [here](http://docs.celeryproject.org/en/3.1/userguide/tasks.html#retrying), but the worker lost errors which happen for sigterm, sigsegv, etc. weren't catchable by it the last time i tried

gcapalbo:
@emord answered the retry question [here](https://github.com/dimagi/commcare-hq/pull/15322/files#r105786414). if it's absolutely critical these tasks never get dropped i think the only way to do that is to implement something like what the pillow retry process does

benrudolph:
would it be possible to try to catch any error and just place it on the pillow retry queue: https://github.com/dimagi/commcare-hq/blob/master/corehq/ex-submodules/pillowtop/pillow/interface.py#L259

gcapalbo:
@benrudolph i think that's what Emord just added: https://github.com/dimagi/commcare-hq/pull/15322/commits/3f9c65c8fdf9eec3bce18f1ba4f4ddcb6683b5e4

i just want to reiterate that this doesn't catch worker lost errors (sigterm, sigsegv, etc.). if it's critical these tasks finish we'd have to implement a separate queueing mechanism like the pillow retry process does, or just put it into the pillow retry process from the beginning if that's possible

benrudolph:
ah sorry missed that. ok i do think it's fairly critical that we don't drop changes. can make for some very hard to debug errors

corehq/apps/userreports/tasks.py
```diff
@@ -1,11 +1,14 @@
 from __future__ import absolute_import
-import datetime
+from datetime import datetime, timedelta
```

emord:
@gcapalbo thanks for the close review.

I needed to do a little refactoring to make this safer, but I think this should not lose any documents.

It now creates a `PillowError` instance and then sends a change over to celery that includes all of the expensive tables that the pillow knows about.

Celery picks that up and goes through each change and if everything goes well, it deletes the PillowError, otherwise it logs it and sets a next attempt. If the worker crashes for any reason the original PillowError should still be there and it will be retried the next time the queue goes through this.

I tested it locally by running:

`./manage.py celery worker --queues=ucr_indicator_queue --events --loglevel=INFO --hostname=local_ucr_indicator_queue -P gevent --concurrency=4`

and ` ./manage.py run_ptop --pillow-name=kafka-ucr-main` and verifying that changes were hitting the celery worker and came into the database

snopoke (commented):
This is a very interesting approach. When you save the `PillowError` you should set `queued` to `True` to prevent it from getting picked up by the normal pillow retry queue.

If this fails here then the normal queue will pick it up and it will re-process all the configs for this doc (which I think is OK).

One other thing that I think is important is to be able to track how many of these are in the 'queue'. We need a way to distinguish them from normal `PillowErrors`. Perhaps a 'state' field? (which might also simplify some of the overly complex retry logic in there currently). Some metrics would also go a long way to maintaining visibility into what's in the queue.

My takeaway from this is that we're converting the `PillowError` queue into a `AsyncProcessing` queue.

emord:
> When you save the PillowError you should set queued to True to prevent it from getting picked up by the normal pillow retry queue.

I don't think it should be saved as True because then if the task picks it up and segfaults or is killed or something, the job would be lost.

> One other thing that I think is important is to be able to track how many of these are in the 'queue'. We need a way to distinguish them from normal PillowErrors. Perhaps a 'state' field? (which might also simplify some of the overly complex retry logic in there currently). Some metrics would also go a long way to maintaining visibility into what's in the queue.

Do you think the pillow_id field would be insufficient to track this? (this would be kafka-ucr-main and kafka-ucr-static)

> My takeaway from this is that we're converting the PillowError queue into a AsyncProcessing queue.

I think this still counts as a PillowError since it's an error that's coming from a pillow. this just changes the fact that pillows can now work asynchronously

corehq/apps/userreports/pillow.py
```diff
@@ -21,11 +21,11 @@
 from pillowtop.pillow.interface import ConstructedPillow
 from pillowtop.processors import PillowProcessor
 from pillowtop.utils import ensure_matched_revisions, ensure_document_exists
+from pillow_retry.models import PillowError
+import six
```
dimagimon: F811 redefinition of unused 'six' from line 7

corehq/apps/userreports/models.py
```diff
@@ -94,6 +94,7 @@ class DataSourceConfiguration(UnicodeMixIn, CachedCouchDocumentMixin, Document):
     meta = SchemaProperty(DataSourceMeta)
     is_deactivated = BooleanProperty(default=False)
     last_modified = DateTimeProperty()
+    asynchronous = BooleanProperty(default=False)
```
snopoke: Is this something that would get set manually? Who would decide whether or not to set this?

snopoke (approved):

corehq/apps/userreports/models.py
emord: I think it'll be a quick way to change a data source that is taking too much time. 

Assuming it'll mostly be static ucrs and likely the eniskhay ones since those have been destroying india's (although much less so since noah's change)

PR #15325, opened 2017-03-10
Add the ability for mobile workers to reset their passwords

kaapstorm:
Context: [FB 240812](http://manage.dimagi.com/default.asp?240812)

Adds a workflow for the mobile worker's domain to be passed throughout the password reset process, so they can use a username like "foo" instead of "foo@domain.commcarehq.org". Offers a mobile-friendly interface extending the same base template as the mobile worker login.

:tropical_fish: 

@snopoke


corehq/apps/domain/forms.py
```diff
@@ -17,8 +17,9 @@
 from django import forms
 from django.conf import settings
 from django.contrib.auth import get_user_model
-from django.contrib.auth.forms import SetPasswordForm
+from django.contrib.auth.forms import SetPasswordForm, PasswordResetForm
 from django.contrib.auth.hashers import UNUSABLE_PASSWORD_PREFIX
+from django.contrib.auth.models import User
```
dimagimon: F401 'User' imported but unused

corehq/apps/domain/forms.py
```diff
@@ -81,6 +81,7 @@
     REPORT_BUILDER_TRIAL,
 )
 from corehq.toggles import HIPAA_COMPLIANCE_CHECKBOX
+from corehq.util.soft_assert import soft_assert
```
dimagimon: F401 'soft_assert' imported but unused

kaapstorm:
Hey @proteusvacuum, Simon's pretty busy with other stuff. Would you have a chance to eyeball this?

corehq/apps/domain/views.py
```diff
@@ -3271,6 +3276,10 @@ def exception_safe_password_reset(request, *args, **kwargs):
     # http://streamhacker.com/2009/09/19/django-ia-auth-password-reset/
     # http://www.rkblog.rk.edu.pl/w/p/password-reset-django-10/
     # http://blog.montylounge.com/2009/jul/12/django-forgot-password/
+    kwargs['password_reset_form'] = partial(HQPasswordResetForm, domain)
+    if domain:
+        kwargs['extra_context']['domain_name'] = domain
+        kwargs['extra_email_context'] = {'domain_name': domain}
     try:
         return password_reset(request, *args, **kwargs)
     except None:
```
proteusvacuum: Weird. What does `except None` do? (I know you didn't write this, just never seen it before).

proteusvacuum:
```
In [8]: try:
   ...:     raise Exception
   ...: except None:
   ...:     print "hello"
   ...:     
---------------------------------------------------------------------------
Exception                                 Traceback (most recent call last)
<ipython-input-8-f373c0667aad> in <module>()
      1 try:
----> 2     raise Exception
      3 except None:
      4     print "hello"
      5 

Exception: 

In [9]: try:
   ...:     raise None
   ...: except None:
   ...:     print "hello"
   ...:     
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-9-005bbeca5615> in <module>()
      1 try:
----> 2     raise None
      3 except None:
      4     print "hello"
      5 

TypeError: exceptions must be old-style classes or derived from BaseException, not NoneType
```

corehq/apps/hqwebapp/urls.py
```diff
@@ -52,6 +56,38 @@
...
         kwargs={'template_name': 'login_and_password/mobile_login.html'}),
+
+    # This url is linked to from web login page
+    url(r'^reset_pwd_email/$', exception_safe_password_reset,
+        {'template_name': 'login_and_password/password_reset_form.html',
+         'email_template_name': 'login_and_password/domain_reset_pwd_email.html',
+         'from_email': settings.DEFAULT_FROM_EMAIL,
+         'extra_context': {'current_page': {'page_name': _('Password Reset')}}},
+        name='domain_reset_pwd_email'),
+    # This url is linked to from mobile login page
```
proteusvacuum: is the mobile login page ever called? It looks like only when this is true:
```py
def is_mobile_url(url):
    # Minor hack
    return ('reports/custom/mobile' in url)
```
and I don't think that is ever true any more?

corehq/apps/hqwebapp/urls.py
```diff
@@ -52,6 +56,38 @@
...
+    url(r'^reset_pwd_email/$', exception_safe_password_reset,
+        {'template_name': 'login_and_password/mobile_password_reset_form.html',
+         'email_template_name': 'login_and_password/domain_mobile_reset_pwd_email.html',
+         'from_email': settings.DEFAULT_FROM_EMAIL,
+         'extra_context': {'current_page': {'page_name': _('Password Reset')}}},
+        name='domain_mobile_reset_pwd_email'),
```
proteusvacuum: ditto to above comment

corehq/apps/hqwebapp/views.py
```diff
@@ -211,6 +212,32 @@ def password_change(req):
...
+def domain_reset_pwd(request, domain, uidb64, token, template_name='login_and_password/reset_pwd.html'):
+    domain_instance = Domain.get_by_name(domain)
+    if not domain_instance:
+        raise Http404
+    request.project = domain_instance
+
+    extra_context = {
+        'domain': domain,
+        'display_name': domain_instance.display_name() or domain,
+        'current_page': {'page_name': _('Welcome to %s!') % domain_instance.display_name()},
```
proteusvacuum: might be nicer for the translators if this was named, otherwise there is no context

corehq/apps/hqwebapp/views.py
```diff
@@ -211,6 +212,32 @@ def password_change(req):
...
+        request, uidb64, token, template_name,
+        set_password_form=HQSetPasswordForm, extra_context=extra_context
+    )
+
+
+def domain_reset_pwd_complete(request, domain,
+                              template_name='login_and_password/password_reset_complete.html',
+                              extra_context=None):
+    extra_context = {'domain_name': domain} if extra_context is None else dict(extra_context, domain_name=domain)
+    return password_reset_complete(request, template_name, extra_context)
```
proteusvacuum: does this method validate that the password reset was actually completed? Otherwise, what happens if I go to this link without first going through the other steps?

corehq/apps/hqwebapp/views.py
```diff
@@ -211,6 +212,32 @@ def password_change(req):
     return render(req, template_name, {"form": password_form})
 
 
+# Corresponds to django.contrib.auth.views.password_reset_confirm()
+# Handles when a user clicks the link in their "reset password" e-mail and prompts for a new password
+def domain_reset_pwd(request, domain, uidb64, token, template_name='login_and_password/reset_pwd.html'):
+    domain_instance = Domain.get_by_name(domain)
+    if not domain_instance:
+        raise Http404
```
proteusvacuum: Might want to add a message? I think I'd get very confused if I clicked a link in my email and it gave me a 404...

proteusvacuum (commented):
couple of questions. looks good though

corehq/apps/domain/templates/login_and_password/domain_mobile_reset_pwd.html
```diff
@@ -0,0 +1,52 @@
...
+                <button tabindex="3" type="submit" class="btn btn-primary">{% trans "Submit" %}</button>
+            </div>
+        </div>
+    </form>
+
+    <script>
+        $(function () {
+            $('#id_new_password1').attr('tabindex', '1').focus();
+            $('#id_new_password2').attr('tabindex', '2');
+            $('input[type="password"]').addClass('form-control');
```
kaapstorm: These fields are defined in `django.contrib.auth.forms.SetPasswordForm`, so these attributes can't be set there. I chose to set them in this template, but are right -- they should be dealt with in the subclass `HQSetPasswordForm` if possible. Turns out it's totally possible, so I've moved these to the `__init__()` method.

corehq/apps/domain/templates/login_and_password/domain_mobile_reset_pwd_email.html
```diff
@@ -0,0 +1,16 @@
+{% load i18n %}{% autoescape off %}
+{% blocktrans %}You're receiving this email because you requested a password reset for your user account at {{ site_name }}{% endblocktrans %}.
+
+{% trans "Please go to the following page and choose a new password:" %}
+{% block reset_link %}
+{{ protocol }}://{{ domain }}{% url "domain_mobile_reset_pwd_confirm" domain=domain_name uidb64=uid token=token %}
+{% endblock %}
+{% trans "Your username, in case you've forgotten:" %} {{ user.username }}
```
kaapstorm: I agree. This template is a copy of the one sent to web users. I'll change the wording here, and fix the web users' template in a separate branch/PR.

corehq/apps/domain/templates/login_and_password/domain_mobile_reset_pwd_email.html
```diff
@@ -0,0 +1,16 @@
...
+
+{% trans "Your domains, in case you've forgotten:" %}{%for x in user.domain_membership.all%}{%if not forloop.first%},{%endif%} {{ x.domain }}{% endfor %}
```
kaapstorm: :+1: Again, I'll change it here, and fix the original template in a separate branch.

corehq/apps/domain/templates/login_and_password/domain_reset_pwd.html
```diff
@@ -0,0 +1,45 @@
...
+                  <label class="col-sm-4 control-label">{{ field.label }}</label>
+                  <div class="col-sm-4">{{ field }}
+                  {% for error in field.errors %}
+                      <span class="help-block">{{ error }}</span>
+                  {% endfor %}
+                  </div>
+                </div>
+                {% endfor %}
+
+                {# hidden submit button to enable [enter] key #}
```
kaapstorm: That does sound about right! I'll check, and if so, fix the original separately.

corehq/apps/domain/templates/login_and_password/domain_mobile_reset_pwd_email.html
```diff
@@ -0,0 +1,16 @@
+{% load i18n %}{% autoescape off %}
+{% blocktrans %}You're receiving this email because you requested a password reset for your user account at {{ site_name }}{% endblocktrans %}.
+
+{% trans "Please go to the following page and choose a new password:" %}
```
kaapstorm: Currently, despite the misleading filename, this template serves as both the HTML and plaintext template, so making this a link would break the plaintext render. If we care about plaintext emails (Do we still care about plaintext emails? Is that still a thing in 2017?) then we should split this out into two templates, with a link in the HTML one, and this URL in the plaintext one. 

(There is just something about plaintext, isn't there? Plaintext is the vinyl of email. Plaintext smells like a new book.)

corehq/apps/domain/templates/login_and_password/domain_mobile_reset_pwd_email.html
kaapstorm: I think I'll need a list comprehension to pull out `x.domain` before I can join it. I'll move it to the view.

corehq/apps/domain/views.py
kaapstorm: lol. This can go to a third branch/PR.

corehq/apps/hqwebapp/views.py
kaapstorm: It'll just show you a link to the login page.

corehq/apps/domain/views.py
kaapstorm: See PR #16899

PR #15331, opened 2017-03-10
Revert "Fixes issue with static widths on select2s in case management + live preview"

orangejenny:
Reverting dimagi/commcare-hq#14795 because it caused https://manage.dimagi.com/default.asp?249148 and stretches out the select2 in conditional case opening:

screen shot 2017-03-10 at 1 44 46 pm

@biyeun / @calellowitz / @dannyroberts


orangejenny:
LESS-only change, no test coverage, killed build

PR #15333, opened 2017-03-10
hotfixing bad property for advanced modules

calellowitz:
@orangejenny


orangejenny:
hotfix, not waiting on tests

PR #15335, opened 2017-03-10
Ce/case config

calellowitz:
@orangejenny 
Not sure this was totally necessary, but it was annoying enough that i took a stab at fixing it. I am able to sucessfully load and modify an advanced module with case management, but didnt try anything more advanced than that.

### DIFF
--- a/corehq/apps/app_manager/static/app_manager/js/case-config-ui-advanced.js
+++ b/corehq/apps/app_manager/static/app_manager/js/case-config-ui-advanced.js
@@ -214,10 +214,10 @@ hqDefine('app_manager/js/case-config-ui-advanced.js', function () {
         };
     };
 
-    var CaseConfigViewModel = function (config, params) {
+    var CaseConfigViewModel = function (caseConfig, params) {
         var self = this;
 
-        self.config = config;
+        self.caseConfig = caseConfig;
 
         self.getCaseTags = function (type, action) {
             var tags = [],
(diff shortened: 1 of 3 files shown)


corehq/apps/app_manager/templates/app_manager/v1/partials/case_config_advanced.html
```diff
@@ -178,7 +178,7 @@ <h3 class="panel-title" data-bind="html: header"></h3>
                 <div class="form-group">
                     <label>{% trans "Case Type" %}</label>
                     <select class="form-control" data-bind="
-                        options: config.caseTypes,
+                        options: caseConfig.caseTypes,
```
orangejenny: Do https://github.com/dimagi/commcare-hq/blob/eb893f0091554b485f8cb26aa205fe6d1e25f2e5/corehq/apps/app_manager/templates/app_manager/v1/partials/case_config_advanced.html#L30 and https://github.com/dimagi/commcare-hq/blob/eb893f0091554b485f8cb26aa205fe6d1e25f2e5/corehq/apps/app_manager/templates/app_manager/v1/partials/case_config_advanced.html#L155 also need to use `caseConfig`?

corehq/apps/app_manager/static/app_manager/js/case-config-ui-advanced.js
```diff
@@ -437,16 +437,16 @@ hqDefine('app_manager/js/case-config-ui-advanced.js', function () {
...
+                self.caseConfig.applyAccordion('open', index);
             }
         };
 
         self.removeFormAction = function (action) {
             if (action.actionType === 'open') {
                 self.open_cases.remove(action);
             } else if (action.actionType === 'load') {
-                var index = self.config.caseConfigViewModel.load_update_cases.indexOf(action),
+                var index = self.caseConfig.caseConfigViewModel.load_update_cases.indexOf(action),
```
orangejenny: What about https://github.com/dimagi/commcare-hq/blob/eb893f0091554b485f8cb26aa205fe6d1e25f2e5/corehq/apps/app_manager/static/app_manager/js/case-config-ui-advanced.js#L454?

orangejenny:
Thumbs up for following up a hotfix with a related refactor. I do think there's value in `case-config-ui.js` and `case-config-ui-advanced.js` using similar naming.

corehq/apps/app_manager/templates/app_manager/v1/partials/case_config_advanced.html
calellowitz: finally got around to looking following up on this. Those two examples should be fine as that `config` is a local variable within the `case-config:condition` template. Im open to changing it for consitency's sake but it should work either way.

calellowitz:
commented on the outstanding questions. fixed the last reference in the followup commit.

PR #15337, opened 2017-03-10
remove unused *args from handle()

nickpell:
Cleaning up because args not added in ```add_arguments()``` won't be recognized by django 1.10.  Will follow up by updating management commands where ```*args``` is used without ```add_arguments()```

@emord cc: @mkangia

### DIFF
--- a/corehq/apps/hqadmin/management/commands/delete_related_cases.py
+++ b/corehq/apps/hqadmin/management/commands/delete_related_cases.py
@@ -15,9 +15,7 @@ class Command(BaseCommand):
         parser.add_argument('case_id', type=unicode)
         parser.add_argument('--filename', dest='filename', default='case-delete-info.csv')
 
-    def handle(self, *args, **options):
-        domain = options['domain']
-        case_id = options['case_id']
+    def handle(self, domain, case_id, **options):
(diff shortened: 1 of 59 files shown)


corehq/apps/hqadmin/management/commands/delete_related_cases.py
```diff
@@ -15,9 +15,7 @@ def add_arguments(self, parser):
         parser.add_argument('case_id', type=unicode)
         parser.add_argument('--filename', dest='filename', default='case-delete-info.csv')
 
-    def handle(self, *args, **options):
-        domain = options['domain']
-        case_id = options['case_id']
+    def handle(self, domain, case_id, **options):
```
mkangia: Is this moving options as arguments? If yes, would it be nice to keep those still in options to avoid changes in implementation?

mkangia (approved):
Just left a suggestion/query.

corehq/apps/hqadmin/management/commands/delete_related_cases.py
nickpell: I prefer to make non-optional arguments required arguments in ```handle``` since it's clearer that these arguments are expected.

PR #15339, opened 2017-03-11
correctly check if task is pending for GroupResult

benrudolph:
@dannyroberts noticed that the progress bar isn't correctly working for the prime restore. this should correctly check if any of the a GroupResult's tasks are pending

cc: @biyeun


snopoke:
Build time out. :boot:

dannyroberts (approved):

PR #14758, opened 2017-02-06
[WIP] Add repeaters and corresponding generators to FollowUp end point

mkangia:

PR #15340, opened 2017-03-12
Mk nikshay hiv integration

mkangia:
Extracted out from #14758 implementing [HIV Test API Integration](https://docs.google.com/document/d/1yUWf3ynHRODyVVmMrhv5fDhaK_ufZSY7y0h9ke5rBxU/edit#heading=h.hxfnqahoeag)

cc: Review buddy @sravfeyn


custom/enikshay/integrations/nikshay/tests/test_repeaters.py
```diff
@@ -295,3 +301,151 @@ def test_handle_failure(self):
...
+        )
+
+        self.person_case = CaseAccessors(self.domain).get_case(self.person_id)
+        payload = (json.loads(
+            NikshayHIVTestPayloadGenerator(None).get_payload(self.repeat_record, self.person_case))
+        )
+
+        self.assertEqual(payload["CPTDeliverDate"], "02/01/2016")
+        self.assertEqual(payload["InitiatedDate"], "03/04/2016")
+        self.assertEqual(payload["ARTCentreDate"], "03/04/2016")
```
dimagimon: W292 no newline at end of file

custom/enikshay/integrations/nikshay/views.py
```diff
@@ -5,3 +5,9 @@ class RegisterNikshayPatientRepeaterView(AddCaseRepeaterView):
     urlname = 'register_nikshay_patient'
     page_title = "Register Nikshay Patients"
     page_name = "Register Nikshay Patients"
+
+
+class NikshayHIVTestRepeaterView(AddCaseRepeaterView):
+     urlname = 'nikshay_patient_hiv_test'
```
dimagimon: E111 indentation is not a multiple of four

custom/enikshay/integrations/nikshay/views.py
```diff
@@ -5,3 +5,9 @@ class RegisterNikshayPatientRepeaterView(AddCaseRepeaterView):
...
+     page_title = "Nikshay Patients HIV Test"
```
dimagimon: E111 indentation is not a multiple of four

custom/enikshay/integrations/nikshay/views.py
```diff
@@ -5,3 +5,9 @@ class RegisterNikshayPatientRepeaterView(AddCaseRepeaterView):
...
+     page_name = "Nikshay Patients HIV Test"
```
dimagimon: E111 indentation is not a multiple of four

custom/enikshay/case_utils.py
```diff
@@ -35,12 +35,11 @@ def get_parent_of_case(domain, case_id, parent_case_type):
```
NoahCarnahan: > There seems to be an issue where the case_type is not correct in indices
This sounds like it could be a fairly serious bug. Have you been able to reliably reproduce this wrong case_type on indicies issue? If so, we might want to send over to interrupt to investigate further.

custom/enikshay/integrations/nikshay/repeater_generator.py
```diff
@@ -24,10 +31,13 @@
     disease_classification,
     dcexpulmonory,
     dcpulmonory,
+    hiv_status,
+    art_initiated,
 )
 from custom.enikshay.case_utils import update_case
 
 ENIKSHAY_ID = 8
+NIKSHAY_NULL_DATE = '1990-01-01'
```
NoahCarnahan: I think this should be `1900`, not `1990`.

custom/enikshay/integrations/nikshay/repeater_generator.py
```diff
@@ -104,6 +114,63 @@ def handle_exception(self, exception, repeat_record):
...
+            "ARTCentreDate": datetime.datetime.strptime(
+                person_case_properties.get('art_initiation_date', NIKSHAY_NULL_DATE), '%Y-%m-%d'
+            ).strftime('%d/%m/%Y'),
+            "InitiatedOnART": art_initiated.get(person_case_properties.get('art_initiated', 'no')),
+            "InitiatedDate": datetime.datetime.strptime(
+                person_case_properties.get('art_initiation_date', NIKSHAY_NULL_DATE), '%Y-%m-%d'
+            ).strftime('%d/%m/%Y'),
+            "Source": ENIKSHAY_ID,
+            "regby": repeat_record.repeater.username,
+            "password": repeat_record.repeater.password,
```
NoahCarnahan: Seems weird to be sending the password in the request body, but assuming this is a quirk of the Nikshay api?

custom/enikshay/integrations/nikshay/repeaters.py
```diff
@@ -34,17 +37,60 @@ def allowed_to_forward(self, episode_case):
...
+        # CPTDeliverDate changes OR
+        # InitiatedDate/Art Initiated date changes
+        allowed_case_types_and_users = self._allowed_case_type(person_case) and self._allowed_user(person_case)
+        if allowed_case_types_and_users:
+            episode_case = get_open_episode_case_from_person(person_case.domain, person_case.get_id)
+            episode_case_properties = episode_case.dynamic_case_properties()
+
+            return (
+                episode_case_properties.get('nikshay_registered', 'false') == 'true' and
+                episode_case_properties.get('nikshay_id') and
```
NoahCarnahan: Should this be `episode_case_properties.get('nikshay_id', False)` as above, in case nikshay id is blank for whatever reason?

NoahCarnahan (commented):
Looks good, just a few things

custom/enikshay/integrations/nikshay/repeater_generator.py
mkangia: Yup! thats how nikshay API accepts password.

NoahCarnahan: 😝

custom/enikshay/integrations/nikshay/repeaters.py
mkangia: Expecting this to return `None`. Wont it return empty string if its empty string? so should not make a difference to add that default False. Nikshay. Also can't see a case where this should be blank string since its only added [here](https://github.com/dimagi/commcare-hq/blob/master/custom/enikshay/integrations/nikshay/repeater_generator.py#L81) and not updated again.

custom/enikshay/integrations/nikshay/repeater_generator.py
mkangia: Actually it should be 1900 only. Nikshay wants a value which is way before so its not valid. This date was suggested by them only.

custom/enikshay/integrations/nikshay/repeaters.py
NoahCarnahan: Oh right, it returns `None`. I was thinking this would raise an exception if the key wasn't in the the dict, but you're right.

custom/enikshay/integrations/nikshay/repeater_generator.py
NoahCarnahan: Ah ok, I had seen 1900 in the spec (not 1990)

NoahCarnahan (approved):

custom/enikshay/case_utils.py
mkangia: Farid and I noticed this bug here..
https://india.commcarehq.org/hq/admin/raw_doc/?id=39d45f04-3005-4ea1-8692-a5bf1e7ff685&db_name= 
references https://india.commcarehq.org/hq/admin/raw_doc/?id=190867c3-38db-4344-849f-c53849b252bd with incorrect case type. I am not sure if this is fixed now. Farid suggested this work around.

custom/enikshay/integrations/nikshay/repeater_generator.py
mkangia: Sorry you are right. Changing this. Thanks Noah

NoahCarnahan: 👍

mkangia: Done!

custom/enikshay/case_utils.py
proteusvacuum: yeah, it's a quirk of advanced modules I think. I think Simon fixed this here: https://github.com/dimagi/commcare-hq/pull/14501, but I'm not sure we have updated it everywhere yet.

custom/enikshay/integrations/nikshay/repeater_generator.py
proteusvacuum: 🔒 🔑 
Hopefully they'll add HTTPS soon too! 🙄

custom/enikshay/integrations/nikshay/repeater_generator.py
```diff
@@ -104,6 +114,63 @@ def handle_exception(self, exception, repeat_record):
...
+            "IP_FROM": "127.0.0.1",
+        }
+
+        return json.dumps(properties_dict)
+
+    def handle_success(self, response, payload_doc, repeat_record):
```
proteusvacuum: I'm assuming you've validated that every `200-300` response is actually a success in Nikshay's eyes? I remember there were some quirks with this earlier.

custom/enikshay/integrations/nikshay/repeaters.py
```diff
@@ -34,17 +37,60 @@ def allowed_to_forward(self, episode_case):
...
+        else:
+            return False
+
+
+class NikshayHIVTestRepeater(CaseRepeater):
+    class Meta(object):
+        app_label = 'repeaters'
 
+    include_app_id_param = False
+    friendly_name = _("Forward eNikshay Patient's HIV Test to Nikshay")
```
proteusvacuum: nit: I've been adding the case type to the name for future wanderers.

custom/enikshay/integrations/nikshay/repeaters.py
```diff
@@ -34,17 +37,60 @@ def allowed_to_forward(self, episode_case):
...
+        if allowed_case_types_and_users:
+            episode_case = get_open_episode_case_from_person(person_case.domain, person_case.get_id)
+            episode_case_properties = episode_case.dynamic_case_properties()
+
+            return (
+                episode_case_properties.get('nikshay_registered', 'false') == 'true' and
+                episode_case_properties.get('nikshay_id') and
+                not test_submission(person_case) and
+                (
+                    related_dates_changed(person_case) or
```
proteusvacuum: there's the [`case_properties_changed`](https://github.com/dimagi/commcare-hq/blob/9e1208aab9876194c0378c987ae194809051c4be/custom/enikshay/integrations/ninetyninedots/repeaters.py#L185-L201) method which will allow you to do this in one line:
```py
... and case_properties_changed(person_case, ['art_initiation_date', 'cpt_initiation_date', 'hiv_status'])
```

custom/enikshay/integrations/nikshay/repeaters.py
```diff
@@ -70,10 +116,45 @@ def episode_pending_registration_changed(case):
...
+
+def create_hiv_test_repeat_records(sender, case, **kwargs):
+    create_repeat_records(NikshayHIVTestRepeater, case)
+
 case_post_save.connect(create_case_repeat_records, CommCareCaseSQL)
+case_post_save.connect(create_hiv_test_repeat_records, CommCareCaseSQL)
 
 # TODO: Remove this when eNikshay gets migrated to SQL
 case_post_save.connect(create_case_repeat_records, CommCareCase)
+case_post_save.connect(create_hiv_test_repeat_records, CommCareCase)
```
proteusvacuum: technically, I don't think we need these any more

custom/enikshay/integrations/nikshay/tests/test_repeaters.py
```diff
@@ -295,3 +301,151 @@ def test_handle_failure(self):
...
+
+        MockRepeater = namedtuple('MockRepeater', 'username password')
+        MockRepeatRecord = namedtuple('MockRepeatRecord', 'repeater')
+        self.repeat_record = MockRepeatRecord(MockRepeater(username="arwen", password="Hadhafang"))
+
+    def create_case_structure(self):
+        return {case.get_id: case for case in filter(None, self.factory.create_or_update_cases(
+            [self.person, self.episode]))}
+
+    def _create_nikshay_registered_case(self):
```
proteusvacuum: this is shadowing a different method with the same name in the superclass. can you rename this one?

custom/enikshay/integrations/nikshay/tests/test_repeaters.py
```diff
@@ -295,3 +301,151 @@ def test_handle_failure(self):
...
+        self.episode_case = self.cases['episode']
+        self._create_nikshay_registered_case()
+
+        MockRepeater = namedtuple('MockRepeater', 'username password')
+        MockRepeatRecord = namedtuple('MockRepeatRecord', 'repeater')
+        self.repeat_record = MockRepeatRecord(MockRepeater(username="arwen", password="Hadhafang"))
+
+    def create_case_structure(self):
+        return {case.get_id: case for case in filter(None, self.factory.create_or_update_cases(
+            [self.person, self.episode]))}
```
proteusvacuum: why did you need to do this?

proteusvacuum (commented):
Couple of comments that you could follow up with in a subsequent PR. Looks good, thanks for doing this!

PR #15341, opened 2017-03-12
Add indexes to awc_id

emord:
https://docs.google.com/document/d/1dp_QXUvE6V9UKsgDrpBNXvVnDxQJDltCUJcPvlqE1SA/edit

@snopoke @benrudolph you guys had an [UNCLEAR] before one of the indexes so I didn't add it here. I think this was just based on the mobile ucrs? it's probable that adding indexes all of the ids (block_id, district_id, supervisor_id etc) would help the web based ones

@sheelio

### DIFF
--- a/corehq/apps/userreports/indicators/factory.py
+++ b/corehq/apps/userreports/indicators/factory.py
@@ -37,6 +37,7 @@ def _build_raw_indicator(spec, context):
         datatype=wrapped.datatype,
         is_nullable=wrapped.is_nullable,
         is_primary_key=wrapped.is_primary_key,
+        create_index=wrapped.create_index,
     )
     return RawIndicator(
         wrapped.display_name,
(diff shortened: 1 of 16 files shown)


benrudolph:
nice thanks for doing this. simon also had some ideas about combined indexes that would help. the `[UNCLEAR]` was because i attempted to first make that index concurrently and i eventually killed it. however when i went to recreate non concurrently, it said it already existed. i should've just deleted and recreated.. anyways, long story short, we should create an index for that column.

benrudolph (approved):

snopoke:
2 things:
* before this get's merged we should rename the indexes we created to match the naming schema here
* would be good to get input from @sheelio since we only looked at reports being used by the AWW App.

sheelio:
I think Jon has added all the ones I would suggest (indexes on supervisor_id as well, and on month for the monthly tables).

snopoke (approved):

emord:
Merging so this gets in before an ICDS deploy @sheelio

PR #15344, opened 2017-03-13
upgrade datadog library

snopoke:
https://github.com/DataDog/datadogpy/blob/master/CHANGELOG.md

### DIFF
--- a/corehq/util/datadog/utils.py
+++ b/corehq/util/datadog/utils.py
@@ -2,11 +2,8 @@ import re
 import logging
 from functools import wraps
 
-from datadog.api.exceptions import DatadogException
-
 from corehq.util.datadog import statsd, COMMON_TAGS, datadog_logger
 from datadog import api
-
 
 from corehq.util.datadog.const import ALERT_INFO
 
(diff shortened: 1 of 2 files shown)


benrudolph (approved):
nice some good stuff in there. the `_TimedContextManagerDecorator` could be very useful, though very similar to your timing context

corehq/util/datadog/utils.py
```diff
@@ -49,8 +46,8 @@ def create_datadog_event(title, text, alert_type=ALERT_INFO, tags=None, aggregat
                 title=title, text=text, tags=tags,
                 alert_type=alert_type, aggregation_key=aggregation_key,
             )
-        except DatadogException:
-            datadog_logger.exception('Error creating Datadog event')
+        except Exception as e:
```
kaapstorm: Just curious, how come you decided against `DatadogException`?

kaapstorm (approved):

snopoke:
They removed the generic base exception so the alternative is to catch 4
exceptions.

PR #15345, opened 2017-03-13
Remove border on legend

benrudolph:
@dimagi/ux we need to remove the border on empty groups: https://manage.dimagi.com/default.asp?249226. however instead of having a bunch of logic to determine when or when not to show the borders, i've decided to just remove it completely (nested empty groups could get annoying). i think it looks clean without. curious if you think this is OK. screen shots below are very contrived

Before
screen shot 2017-03-13 at 12 31 13 pm

After
screen shot 2017-03-13 at 12 27 28 pm

### DIFF
--- a/corehq/apps/style/static/cloudcare/less/formplayer-common/appicon.less
+++ b/corehq/apps/style/static/cloudcare/less/formplayer-common/appicon.less
@@ -89,6 +89,21 @@
   }
 }
 
+.gridicon {
+  background-color: transparent;
+  background-size: contain;
+  background-color: transparent;
+  background-repeat: no-repeat;
+  background-position: 50%;
+  // For icons with their own image, don't make assumptions about the shadowing
+  .box-shadow(0 0 0 0 rgba(0, 0, 0, 0));
+}
+
+.gridicon.gridicon-circle {
+  background-color: @cc-neutral-mid;
+  .box-shadow(0 0 5px 0 rgba(0, 0, 0, 0.45));
+}
+
 .make-appicon-size(170px, 115px, 35px, 16px);
 .make-gridicon-size(70px);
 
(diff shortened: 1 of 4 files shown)


ctsims:
@benrudolph As a heads up when we use this pattern in the app we're going to use internal repeats, which could be lots of elements long, so if we don't update spacing we'll get huge gaps in the form. Can we revisit that before pushing? (I can provide a sample for that demonstrates the gap issue if needed for testing).

benrudolph:
:+1: sure thing, i'll take another pass

ctsims:
hey ben! Thanks so much for quick turnaround.

Did end up making a sample form if helpful:
https://www.commcarehq.org/a/corpora/cloudcare/apps/v2/preview/60a7d2f0ab5c464c23f77bf8ef562bf1/#%7B%22appId%22%3A%2260a7d2f0ab5c464c23f77bf8ef562bf1%22%2C%22previewCommand%22%3A%22m0-f2%22%7D

benrudolph:
lol at that form

corehq/apps/style/static/cloudcare/less/formplayer-common/appicon.less
```diff
@@ -89,6 +89,21 @@
   }
 }
 
+.gridicon {
+  background-color: transparent;
```
orangejenny: nit: this rule is here twice

orangejenny (approved):

corehq/apps/style/static/cloudcare/less/formplayer-common/appicon.less
benrudolph: thanks: https://github.com/dimagi/commcare-hq/pull/15345/commits/0dd0a07664f356988f2d5d41569a2f8336048031

benrudolph:
ok this should be g2g, just one minor change after the approval to review (removing the redundant rule)

ctsims:
thanks, Ben!

PR #15349, opened 2017-03-13
combine intrahealth pillows

emord:
not a fun test to track down, but I think I got it. sqlreport wasn't properly removing the session

@gcapalbo


gcapalbo (approved):
restarted the build, not sure if the failure was related or a blip

corehq/apps/reports/sqlreport.py
```diff
@@ -274,9 +274,11 @@ def _get_data(self, start=None, limit=None):
         session = connection_manager.get_scoped_session(self.engine_id)
         try:
             return qc.resolve(session.connection(), self.filter_values)
-        except:
+        except Exception:
             session.rollback()
             raise
+        finally:
+            session.remove()
```
snopoke: @emord any reason you chose `session.remove()` instead of `session.close()`?

Was going to change it [here](https://github.com/dimagi/commcare-hq/pull/20575) but though you may have a reason for using `remove`

PR #15351, opened 2017-03-13
divide up the main kafka queue

emord:
@snopoke @benrudolph there was talk of splitting up the ucr queues to get more throughput. we could do this for arbitrarily many ucr pillows. Inspired by our nose plugin to divide tests

@gcapalbo


benrudolph (commented):
very cool! i took a quick look at simon's PR that would utilize kafka groups and it does seem like we're fairly far away from that world. i think this would be a good solution for the interim

settings.py
```diff
@@ -1562,11 +1562,6 @@
             'instance': 'corehq.apps.change_feed.pillow.get_domain_db_kafka_pillow',
         },
         {
-            'name': 'kafka-ucr-main',
-            'class': 'corehq.apps.userreports.pillow.ConfigurableReportKafkaPillow',
-            'instance': 'corehq.apps.userreports.pillow.get_kafka_ucr_pillow',
-        },
-        {
```
benrudolph: so we would declare the separate partitions in localsettings then?

emord: yeah. that's what the associated ansible PR would do. would be nice to be able to pass in the partitions here, but that's a little more complicated, especially with how the checkpoints work. 

Also if we do this I (or someone else) would need to go manually set checkpoints for the new pillows if they are changed

corehq/apps/userreports/pillow.py
```diff
@@ -180,16 +189,21 @@ def rebuild_table(self, sql_adapter):
...
             data_source_provider=DynamicDataSourceProvider(),
             auto_repopulate_tables=False,
+            ucr_division=ucr_division
         ),
         pillow_name=pillow_id,
     )
 
 
+get_kafka_ucr_pillow_08 = partial(get_kafka_ucr_pillow, '08')
+get_kafka_ucr_pillow_9f = partial(get_kafka_ucr_pillow, '9f')
```
snopoke: Instead of defining these statically we could add a `params` field to the pillowtops settings and pass that through to the constructor. Then we can just decide in localsettings who to partition for each env.

corehq/apps/userreports/pillow.py
```diff
@@ -1,7 +1,12 @@
 from __future__ import absolute_import
 from collections import defaultdict
-from alembic.autogenerate.api import compare_metadata
 from datetime import datetime, timedelta
+from functools import partial
```
dimagimon: F401 'partial' imported but unused

corehq/apps/userreports/pillow.py
emord: Moved this to params dictionary and updated the ansible config to do this https://github.com/dimagi/commcare-hq/pull/15351/commits/180ac6a79805c140fc8165e375d9ce9d6cecfdd5

localsettings.example.py
```diff
@@ -207,6 +207,16 @@
...
 
+LOCAL_PILLOWTOPS = {
+    'core_ext': [
+        {
+            'name': 'kafka-ucr-main',
+            'class': 'corehq.apps.userreports.pillow.ConfigurableReportKafkaPillow',
+            'instance': 'corehq.apps.userreports.pillow.get_kafka_ucr_pillow',
+        },
+    ]
+}
```
snopoke: Should add this to `localsettings.dev.py` as well (or instead).

snopoke (changes requested):

snopoke:
Realized last night that regardless of what's in settings the pillows that run on a machine is controlled by which supervisor conf files get created by the deploy.

We could just put them all in the settings and update [these files](https://github.com/dimagi/commcare-hq-deploy/tree/master/fab/pillows) for the different ENV's.

emord:
@snopoke The settings is how the system info page gets a list of pillows to display the checkpoint for. if we put all of them in the settings there's going to be a lot of noise on that page (not that there isn't already). i'd rather not contribute more to that

emord:
Err. actually it would only be adding two more and I just removed 10ish, so I'll just add all the ones we're using

snopoke (approved):

emord:
The test that failed was near the end and a test I had broken before in this branch. confirmed it's working locally. Going to merge this so that we can relieve backlog from prod

PR #15353, opened 2017-03-13
_processors -> processors

emord:
Makes _processors public since they're used in a lot of places now

@gcapalbo


corehq/apps/userreports/pillow.py
```diff
@@ -145,7 +145,7 @@ def process_change(self, pillow_instance, change):
 
 
 class ConfigurableReportKafkaPillow(ConstructedPillow):
-    # the only reason this is a class is to avoid exposing _processor
+    # the only reason this is a class is to avoid exposing processors
```
emord: I suppose this class could be changed now, but i'm not sure how deep that's going to go

PR #15354, opened 2017-03-14
don't look for reports, if they are invalid

sravfeyn:
http://manage.dimagi.com/default.asp?249368#1301522 @proteusvacuum

### DIFF
--- a/corehq/apps/app_manager/models.py
+++ b/corehq/apps/app_manager/models.py
@@ -4125,7 +4125,7 @@ class ReportModule(ModuleBase):
                 'type': 'report config ref invalid',
                 'module': self.get_module_info()
             })
-        if not self.reports:
+        elif not self.reports:
             errors.append({
                 'type': 'no reports',
                 'module': self.get_module_info(),


corehq/apps/app_manager/models.py
```diff
@@ -4125,7 +4125,7 @@ def validate_for_build(self):
                 'type': 'report config ref invalid',
                 'module': self.get_module_info()
             })
-        if not self.reports:
+        elif not self.reports:
```
snopoke: not sure what this get's us. If there are no reports then the previous if won't ever execute.

proteusvacuum:
Looks like this was fixed in https://github.com/dimagi/commcare-hq/pull/15377

corehq/apps/app_manager/models.py
sravfeyn: If the condition above is `True` then there is an error in reports config, then calling `self.reports` will raise an Exception. So making this `elif` meant it would be called only if there were no errors.

PR #15357, opened 2017-03-14
Add caching for UCR lookups in SMS indicators

gcapalbo:
Makes it so that we only load the UCRs once per supervisor instead of once per user.
@emord @snopoke

### DIFF
--- a/custom/icds/messaging/indicators.py
+++ b/custom/icds/messaging/indicators.py
@@ -46,7 +46,7 @@ REPORT_IDS = [
 
 @quickcache(['domain'], timeout=4 * 60 * 60, memoize_timeout=4 * 60 * 60)
 def get_report_configs(domain):
-    app = get_app(domain, SUPERVISOR_APP_ID)
+    app = get_app(domain, SUPERVISOR_APP_ID, latest=True)
     return {
         report_config.report_id: report_config
         for module in app.modules if isinstance(module, ReportModule)
--- a/custom/icds/messaging/indicators.py
+++ b/custom/icds/messaging/indicators.py
@@ -24,6 +24,7 @@ from corehq.apps.reports.analytics.esaccessors import (
     get_last_submission_time_for_users,
     get_last_form_submissions_by_user,
 )
+from corehq.util.quickcache import quickcache
 from custom.icds.const import (
     CHILDREN_WEIGHED_REPORT_ID,
     DAYS_AWC_OPEN_REPORT_ID,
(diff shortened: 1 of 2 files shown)


custom/icds/tests/test_ls_aggregate_performance.py
```diff
@@ -37,6 +37,7 @@ def test_report_parsing(self, days_open, weighed, thr, visits):
         thr.return_value = ET.fromstring(self.get_xml('thr_fixture'))
         visits.return_value = ET.fromstring(self.get_xml('visit_fixture'))
         indicator = LSAggregatePerformanceIndicator('domain', 'user')
+        self.addCleanup(indicator.clear_caches)
```
gcapalbo: this doesn't matter right now since the actual reference is mocked out, but just adding this here as a precaution in case this test ever changes

custom/icds/messaging/indicators.py
```diff
@@ -35,6 +36,36 @@
...
+    HOME_VISIT_REPORT_ID,
+    THR_REPORT_ID,
+    CHILDREN_WEIGHED_REPORT_ID,
+    DAYS_AWC_OPEN_REPORT_ID,
+]
+
+
+@quickcache(['domain'], timeout=4 * 60 * 60, memoize_timeout=4 * 60 * 60)
+def get_report_configs(domain):
+    app = get_app(domain, SUPERVISOR_APP_ID)
```
snopoke: Should this not be the latest starred build?

custom/icds/tests/test_ls_aggregate_performance.py
gcapalbo: (removed this, turns out it made the test fail since it wasn't actually creating any users anyway)

custom/icds/messaging/indicators.py
gcapalbo: i'm not sure i know enough about mobile ucr to answer that - pinging @emord since he worked on this initially.

it looks like what we're doing is inspecting the ReportModules of the app and getting the report configs out of it. as long as that all still works correctly with the latest starred build, i think using the latest starred build sounds right to me.

@snopoke @emord does that sound right to you?

emord: Yeah. I agree. I should have used the latest  starred build for this. although I imagine there won't be too much app dev happening on NIC

gcapalbo: updated https://github.com/dimagi/commcare-hq/pull/15357/commits/716556a37ce4bf4cfec76bf0ceb77627432080af

PR #15358, opened 2017-03-14
add more info to case modified assertion

snopoke:
Trying to get to the bottom of why these are happening: https://sentry.io/dimagi/commcarehq/issues/234393735/


benrudolph (approved):
looks good. still seems like it's going to be hard to track down

snopoke:
@benrudolph yea, not sure what other info to put in there. Getting the `server_modified_on` should enable me to see which transaction modified the case.

amstone326:
Cool, thanks @snopoke! I haven't had the bandwidth to do more sleuthing on this in the last few weeks, but it's still on my medium-term to-do list. Would it be possible to add me directly to these soft asserts so I can view them for myself instead of you having to export and share them with me every time? If that involves me getting set up with sentry, I'm fine with doing that, just let me know how.

benrudolph:
ah smart idea

PR #15360, opened 2017-03-14
Handle toggle ResourceNotFound

orangejenny:
https://manage.dimagi.com/default.asp?249303

@dannyroberts

### DIFF
--- a/corehq/apps/toggle_ui/views.py
+++ b/corehq/apps/toggle_ui/views.py
@@ -180,7 +180,6 @@ class ToggleEditView(ToggleBaseView):
 
 def enable_vellum_beta(request):
     slug = "vellum_beta"
-    toggle = None
     try:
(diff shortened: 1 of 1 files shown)


corehq/apps/toggle_ui/views.py
```diff
@@ -180,7 +180,11 @@ def post(self, request, *args, **kwargs):
 
 def enable_vellum_beta(request):
     slug = "vellum_beta"
-    toggle = Toggle.get(slug)
+    toggle = None
```
millerdev: nit: this line is not necessary (no need to declare or initialize variables in python)

millerdev (approved):

corehq/apps/toggle_ui/views.py
orangejenny: javascript habits die hard :)  fixed.

PR #15361, opened 2017-03-14
adding error message for user when multiple location types share code

calellowitz:
@esoergel cc: @NoahCarnahan
more context here
http://manage.dimagi.com/default.asp?249219#1301563
but this stops it from 500ing when you make a new location type with the same name as an old code, and tries to explain to the user what to do about that

### DIFF
--- a/corehq/apps/locations/views.py
+++ b/corehq/apps/locations/views.py
@@ -3,6 +3,7 @@ import logging
 
 from django.contrib import messages
 from django.core.cache import cache
+from django.db import IntegrityError
 from django.http import HttpResponse, HttpResponseRedirect, Http404
 from django.http.response import HttpResponseServerError
 from django.shortcuts import render, get_object_or_404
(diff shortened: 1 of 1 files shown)


NoahCarnahan (approved):
Looks good to me, thanks Cal.

corehq/apps/locations/views.py
```diff
@@ -336,7 +337,14 @@ def mk_loctype(name, parent_type, administrative,
 
         for loc_type in hierarchy:
             # make all locations in order
-            mk_loctype(**loc_type)
+            try:
+                mk_loctype(**loc_type)
+            except IntegrityError:
```
esoergel: I'm not sure about this, as this could conceivably arise by other means.

There's a bit of code in this view that's supposed to detect this:  https://github.com/dimagi/commcare-hq/blob/master/corehq/apps/locations/views.py#L327-L330 The problem is it doesn't work if the codes are inferred.  You could fix this by generating the code earlier rather than just not validating autogenerated codes.

Even better, this could avoid conflicts entirely (for autogenerated codes) by making sure the generated code is unique, for instance, using https://github.com/dimagi/commcare-hq/blob/master/corehq/apps/commtrack/util.py#L255-L255.

PR #15363, opened 2017-03-14
catch more instances where add_arguments is needed

nickpell:
@emord cc: @mkangia

### DIFF
--- a/corehq/apps/app_manager/management/commands/upload_app_forms.py
+++ b/corehq/apps/app_manager/management/commands/upload_app_forms.py
@@ -17,7 +17,7 @@ class Command(BaseCommand):
 
     def add_arguments(self, parser):
         parser.add_arguments(
-            'path_to_dir',
+            'path',
         )
         parser.add_arguments(
             'app_id',
@@ -45,9 +45,7 @@ class Command(BaseCommand):
         )
 
 
-    def handle(self, *args, **options):
-        if len(args) != 2:
-            raise CommandError('Usage: %s\n%s' % (self.args, self.help))
+    def handle(self, path, app_id, **options):
         if options['deploy'] and not options['user']:
             raise CommandError('Deploy argument requires a user')
         elif options['deploy']:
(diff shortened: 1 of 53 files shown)


corehq/apps/app_manager/management/commands/upload_app_forms.py
```diff
@@ -45,9 +45,7 @@ def add_arguments(self, parser):
         )
 
 
-    def handle(self, *args, **options):
-        if len(args) != 2:
-            raise CommandError('Usage: %s\n%s' % (self.args, self.help))
+    def handle(self, path, app_id, **options):
```
dimagimon: E303 too many blank lines (2)

corehq/ex-submodules/couchforms/management/commands/purge_forms_and_cases.py
```diff
@@ -155,6 +159,6 @@ def _print_final_debug_info(xform_ids, filtered_xform_ids, case_ids):
 
 
 def _notify_parsed_args(domain_name, app_id, version_number, test_run):
-    print('Received request for domain : %s with app_id : %s to soft delete data before version number %s' % \
+    print('Received request for domain : %s with app_id : %s to soft delete data before version number %d' % \
```
dimagimon: E502 the backslash is redundant between brackets

emord:
DIdn't look too hard, but the general idea seems good.

However part of me wants to say don't commit this. Instead you could add a comment or some message to each of these saying "Whoa this actually gets run? Here's a link to a commit that will fix it" And then after six months we could look through these and see which were changed and delete the ones that weren't.

(also you've already done this work so feel free to go ahead and merge)

nickpell:
interesting idea.  What about putting at the start of every infrequently used management command
```raise Exception("Delete this command if not used by 2017-10-01")```
or the like?

emord:
Yeah.that seems good too

PR #15367, opened 2017-03-14
Upgrade kafka client

emord:
The logs are littered with https://github.com/dpkp/kafka-python/pull/432/files which were changed to debug in 0.9.5.

@gcapalbo


gcapalbo:
are there any other changes we need to be aware of with the upgrade?

emord:
I don't see anything that regresses https://github.com/dpkp/kafka-python/blob/master/CHANGES.md#095-dec-6-2015

The blocking thing was a change to a parameter, it shouldn't block us

PR #15370, opened 2017-03-14
Show the correct levels for expand-to if this is a forked hierarchy

proteusvacuum:
@esoergel 
@sravfeyn 
Noticed that when you select an `expand_from` level above the current level in a forked hierarchy, you are given the wrong options. This gives you all options visible below the location set as "expand_from".
(The backend was already doing this).


proteusvacuum:
bump

PR #15372, opened 2017-03-15
Case list sort: ability to control where blanks display

orangejenny:
https://manage.dimagi.com/default.asp?247874

Depends on https://github.com/dimagi/commcare-core/pull/536, will be available when mobile 2.35 is released.

@dannyroberts 

fyi @amstone326 

screen shot 2017-03-14 at 8 48 57 pm

### DIFF
--- a/corehq/apps/app_manager/templates/app_manager/v1/partials/case_list.html
+++ b/corehq/apps/app_manager/templates/app_manager/v1/partials/case_list.html
@@ -107,6 +107,7 @@
                         </span>
                     </th>
                     <th>{% trans "Direction" %}</th>
+                    <th>{% trans "Blanks" %}</th>
                     {% if request|toggle_enabled:'SORT_CALCULATION_IN_CASE_LIST' %}
                         <th>{% trans "Sort Calculation" %}{{ request|toggle_tag_info:"SORT_CALCULATION_IN_CASE_LIST" }}</th>
                     {% elif request|toggle_enabled:'SHOW_DEV_TOGGLE_INFO' %}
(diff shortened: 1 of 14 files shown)


amstone326:
@orangejenny What do you think of adding a question mark hover-hint for this? Something like "Should cases for which this sort property is empty be shown at the top of the case list or the bottom?". Thinking it might not be obvious what this property does otherwise.

orangejenny:
@amstone326 good thought, added in https://github.com/dimagi/commcare-hq/pull/15372/commits/9df33939d64da487801485f028f2fcfa7fe83ede

corehq/apps/app_manager/templates/app_manager/v1/partials/case_list.html
```diff
@@ -90,23 +90,34 @@
...
+                                      is useful if for example you want higher
+                                      priority cases to appear closer to the
+                                      top of the list. The case list will sort
+                                      by the first property, then the second,
+                                      etc.
+                                  {% endblocktrans %}" >
+                        </span>
+                    </th>
+                    <th>{% trans "Direction" %}</th>
+                    <!-- ko if: hqImport('app_manager/js/app_manager.js').checkCommcareVersion("2.35") -->
```
dannyroberts: At some point it was best practice to put version checks for features in https://github.com/dimagi/commcare-hq/blob/master/corehq/apps/app_manager/feature_support.py, as a way to document that feature's dependency on that version. Not sure if we abandoned that on purpose, but I still like that convention.

orangejenny: thanks, fixed!

dannyroberts: nice! Glad too that it was a really easy fix like that :)

dannyroberts (approved):

PR #15373, opened 2017-03-15
Use custom expressions to avoid expensive eNikshay UCR calculations

NoahCarnahan:
http://manage.dimagi.com/default.asp?249246
I wrote these custom expressions to avoid [this line](https://github.com/dimagi/commcare-hq/blob/73ee772b70df3b31295923dd55a4b09af0e2583a/corehq/apps/userreports/expressions/specs.py#L302), which is very time consuming if the number of forms is large.

Other optimizations could be done, such as caching the value of `FormProcessorInterface(domain).get_case_forms(case_id)`, but at least right now that call is insignificant compared to all the calls to `form.to_json()`, which require a trip to riak.

Hold off on merge until I have tested this.
@calellowitz @emord

### DIFF
--- a/custom/enikshay/expressions.py
+++ b/custom/enikshay/expressions.py
@@ -28,7 +28,7 @@ class FirstCaseFormWithXmlns(JsonObject):
     def _get_forms(self, case_id, context):
         domain = context.root_doc['domain']
 
-        cache_key = (self.__class__.__name__, case_id)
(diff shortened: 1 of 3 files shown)


NoahCarnahan:
Actually, I'm going to move the expression additions and the static report update into separate PRs so that I can test the expressions with prod data before updating the static report.

PR #15375, opened 2017-03-15
Mk nikshay repeaters update

mkangia:
Turns out we should be sending updates for 
1. Migrated patients
2. patients registered via API
3. patients that provide nikshay_id during registration

So, update to send HIV Test update and Treatment Outcome for patients that have nikshay_id available

### DIFF
--- a/custom/enikshay/integrations/nikshay/repeater_generator.py
+++ b/custom/enikshay/integrations/nikshay/repeater_generator.py
@@ -142,8 +142,12 @@ class NikshayTreatmentOutcomePayload(BaseNikshayPayloadGenerator):
         person_case = get_person_case_from_episode(episode_case.domain, episode_case.get_id)
         episode_case_properties = episode_case.dynamic_case_properties()
         base_properties = self._base_properties(repeat_record, person_case)
+        nikshay_id = (
+            episode_case_properties.get("nikshay_id", None) or
+            person_case.dynamic_case_properties().get('nikshay_id')
+        )
         base_properties.update({
-            "PatientID": episode_case_properties.get("nikshay_id"),
+            "PatientID": nikshay_id,
             "OutcomeDate": episode_case_properties.get(TREATMENT_OUTCOME_DATE),
             "Outcome": treatment_outcome.get(episode_case_properties.get(TREATMENT_OUTCOME)),
             "MO": u"{} {}".format(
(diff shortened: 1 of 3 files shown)


NoahCarnahan (approved):
Looks fine to me, thanks.

nickpell (approved):

mkangia:
@NoahCarnahan @proteusvacuum 
I think it would be better to have nikshay_id at a single location i.e on `person` case. 
We initially added nikshay_id on `episode` case so now its at two different places which confuses with nikshay_id added on `person` case by either data migration done from nikshay or during patient registration. Hence a check was needed at two places in the PR.
If that sounds ok i am gonna update the patient registration generator to add nikshay_id on `person` instead of `episode` thus keeping all three HIVTest, TreatmentOutcome and registration details within a single case.

mkangia:
Thanks for the review @NoahCarnahan . I am going to wait for @proteusvacuum to share his review on this and then merge it post some test cases.

mkangia:
Test cases added!

proteusvacuum:
@mkangia I'm a bit confused about when the Nikshay ID gets stored to the person case. 
Can you show me an example form submission, or where in the migration this is happening? I'm pretty sure the migration sets the Nikshay ID on episode [here](url). Also, the question that updates the Nikshay ID in the app only saves it to the episode (AFAIK).

mkangia:
@proteusvacuum i think i stumbled upon a person case with nikshay_id and then i went through nick's PR for data migration where https://github.com/dimagi/commcare-hq/pull/14307/files#diff-2f3af5f0c009eb077786b485da32afe3R110 made me believe that nikshay_id should be on person for migrated cases.

proteusvacuum:
@mkangia I'm pretty certain the `nikshay_id` is stored to the `episode` case and never the `person`. 
[Here](https://github.com/dimagi/commcare-hq/blob/master/custom/enikshay/nikshay_datamigration/factory.py#L267) is where it gets saved in the migration. We don't save it to the person anywhere (and in the app it gets saved to the episode, too).
That commit you linked is no longer up to date. Please see the latest master version of that file [here](https://github.com/dimagi/commcare-hq/blob/master/custom/enikshay/nikshay_datamigration/tests/test_create_enikshay_cases.py#L148)

mkangia:
Closing this and extracting just the required changes. 
Sorry for the confusion!

mkangia:
Extracted required change #15457

PR #15376, opened 2017-03-15
fix requirement for sending treatment outcome

mkangia:
Fixing the check for sending treatment outcome which should be the opposite of what was set. Bad miss :(


benrudolph (approved):
(approving, don't have much context, but i think this needs to be deployed)

PR #15377, opened 2017-03-15
do not validate when app is a build

benrudolph:
@snopoke @wpride was able to reproduce easily by faking a static ucr. couldn't reproduce with a dynamic one. offending line: https://github.com/dimagi/commcare-hq/blob/br/dont-validate-app/corehq/apps/userreports/models.py#L730

cc: @biyeun


snopoke (approved):

wpride:
💰

PR #15381, opened 2017-03-15
Add custom expressions to avoid expensive eNikshay UCR calculations

NoahCarnahan:
This is one half of this PR: https://github.com/dimagi/commcare-hq/pull/15373

http://manage.dimagi.com/default.asp?249246
I wrote these custom expressions to avoid [this line](https://github.com/dimagi/commcare-hq/blob/73ee772b70df3b31295923dd55a4b09af0e2583a/corehq/apps/userreports/expressions/specs.py#L302), which is very time consuming if the number of forms is large.

Other optimizations could be done, such as caching the value of `FormProcessorInterface(domain).get_case_forms(case_id)`, but at least right now that call is insignificant compared to all the calls to `form.to_json()`, which require a trip to riak.

@calellowitz @emord

### DIFF
--- a/custom/enikshay/expressions.py
+++ b/custom/enikshay/expressions.py
@@ -50,7 +50,7 @@ class FirstCaseFormWithXmlns(JsonObject):
 def first_case_form_with_xmlns_expression(spec, context):
     wrapped = FirstCaseFormWithXmlns.wrap(spec)
     wrapped.configure(
-        ExpressionFactory.from_spec(wrapped.case_id_expression)
+        ExpressionFactory.from_spec(wrapped.case_id_expression, context)
     )
     return wrapped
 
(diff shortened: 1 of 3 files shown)


NoahCarnahan:
FYI no need to review `custom/enikshay/ucr/data_sources/expression_test.json`

custom/enikshay/expressions.py
```diff
@@ -1,10 +1,98 @@
...
+
+        cache_key = (self.__class__.__name__, case_id, self.xmlns, self.reverse)
+        if context.get_cache_value(cache_key) is not None:
+            return context.get_cache_value(cache_key)
+
+        xforms = FormProcessorInterface(domain).get_case_forms(case_id)
+        xforms = sorted(
+            [form for form in xforms if form.xmlns == self.xmlns and form.domain == domain],
+            key=lambda x: x.received_on
+        )
```
snopoke: Did we consider storing the from XMLNS in the `CaseTransaction` table to allow filtering in the DB?

NoahCarnahan: I don't recall if we considered it or not. It would be useful in this situation, but hard to say if it's worth the general table size increase.

PR #15382, opened 2017-03-15
set nikshay id on person case instead of episode case on patient regi…

mkangia:
…station

cc review buddy @nickpell


nickpell (approved):

NoahCarnahan (approved):
Looks ok to me

proteusvacuum (changes requested):
what's the rationale behind this? in the app we store Nikshay id on the episode. If you were to update the Nikshay ID in the app, it would conflict with the one on the person case that comes from Nikshay.

mkangia:
Closing this as this is not needed.

> Nikshay id always is on episode case.

PR #15383, opened 2017-03-15
Use custom expression in eNikshay static ucr datasource

NoahCarnahan:
This is the second half of https://github.com/dimagi/commcare-hq/pull/15381
I've tested that the report will contain the same data with the `expression_test` static datasource, which this PR removes. So no need to review the report config changes closely (or really at all, unless you want to).

@calellowitz @emord 
Would be great to have this in before today's deploy


emord (approved):
:shipit:

PR #15384, opened 2017-03-15
don't listen to unnecessary kafka topics

emord:
From https://github.com/dimagi/commcare-hq/blob/master/corehq/apps/userreports/ui/forms.py#L171 I think ucrs should only listen to locations, forms or cases. No need for it to attempt processing all of the other stuff

@gcapalbo


gcapalbo:
i don't know much about ucr but am wondering if a change like this might cause confusion later on if someone doesn't realize it's not listening to all topics.

also does it rely on any user, group, or app data?

emord:
It does occasionally use users and groups, but that is only when using `related_doc` which is retrieved by id based on some case or form property.

This is the code that filters it based on referenced_doc_type:

https://github.com/dimagi/commcare-hq/blob/master/corehq/apps/userreports/models.py#L119

Perhaps @nickpell @NoahCarnahan can confirm. I think they've both done work on ucrs

NoahCarnahan:
It's safe to exclude to the other types. It looks like there is currently nothing stoping someone from creating a static UCR datasource based on another model type, but it's not an intended workflow. Might be worth just adding an assertion to the `DataSourceConfiguration` model to confirm that `referenced_doc_type` is a case, form, or location.

emord:
Adding an assertion was a good idea. Turns out there's one static ucr that listens to users. 

ran the following on prod and india `{conf.referenced_doc_type for conf in DataSourceConfiguration.all()}` and confirmed that there are only 3 doc types that are referenced

gcapalbo (approved):
:+1: thanks

looks like the build is failing on the `BadSpecError` in one test

corehq/apps/dump_reload/tests/test_couch_dump_load.py
```diff
@@ -100,6 +101,11 @@ def _dump_line_to_doc_class(line):
     def test_docs_with_domain(self):
         # one test for all docs that have a 'domain' property
         doc_types = []
+        special_handlers = {
+            "DataSourceConfiguration": {
+                "referenced_doc_type": lambda x: random.choice(VALID_REFERENCED_DOC_TYPES)
```
emord: Not a huge fan of randomness in tests like this, but did it to keep it in line with the randomness in the original test

emord:
merging. was approved. just merged master and fixed one test

PR #15387, opened 2017-03-16
Set user_id if it's needed and it's blank

gcapalbo:
https://manage.dimagi.com/default.asp?249552
@emord


emord (approved):
kicked the build. looks like we're starting to get some test timeouts on the second test somewhat consistently

PR #15388, opened 2017-03-16
handle case when occurrence case missing/closed for episode

mkangia:


mkangia:
@NoahCarnahan i am going to merge this and deploy.

mkangia:
@proteusvacuum 
This change at least unblocks form submissions but is it a concern that this scenario is possible? that there is an episode case(closed) with a closed occurrence case.
Is it fine to not send a record to Nikshay and simply ignore it if it has a closed occurrence.

proteusvacuum:
If the occurrence is closed, I think the episode should be closed at that point too. Is that not what you are seeing? Sorry, still digging myself out of a swamp here, so I might not have all the context on this. Feel free to take to email if you have more questions.

PR #15391, opened 2017-03-16
if no GET or POST params, redirect back to select screen

sravfeyn:
https://manage.dimagi.com/default.asp?249347 @proteusvacuum


sravfeyn:
@emord (new buddy)

emord:
It looks like this ticket was resolved without this being merged. Is it still needed? Also should there be some error/warning message telling they user why they are redirected?

sravfeyn:
The issue isn't resolved. This[ link ](https://manage.dimagi.com/default.asp?249347)from ticket still is 500ing

PR #15401, opened 2017-03-16
ICDS Reports: Change Definition of Active AWC

sheelio:
Including some additional forms in calculating the definition of "is_active"


sheelio:
@emord 

This is a pretty quick change if you're able to review and merge it.

PR #15402, opened 2017-03-16
eNikshay: Add management command for removing unnecessary repeater submissions

NoahCarnahan:
@proteusvacuum cc @calellowitz 
As previously discussed (but adding here for posterity), a previous repeat bug resulted in many case updates being submitted, all of which updated the same two case properties. Some cases have over 4000 forms modifying them. These forms slow down UCRs in the domain.
This management command archives all by the last system form that updated the two case properties. One slight downside to this approach is that some of those forms would have been intentional, since we would expect a few of those repeater generated forms on the case. However, archiving these old forms should not affect the case's current properties.

### DIFF
--- a/custom/enikshay/management/commands/archive_extra_repeater_forms.py
+++ b/custom/enikshay/management/commands/archive_extra_repeater_forms.py
@@ -1,0 +1,56 @@
+from django.core.management.base import BaseCommand
+
+from casexml.apps.case.xform import get_case_updates
+from casexml.apps.case.xml.parser import CaseUpdateAction
+from corehq.form_processor.interfaces.dbaccessors import CaseAccessors
+from corehq.form_processor.interfaces.processor import FormProcessorInterface
+from corehq.util.log import with_progress_bar
+
+domain = "enikshay"
+
+
+class Command(BaseCommand):
+
+    def add_arguments(self, parser):
+        parser.add_argument('log_file')
+
+        parser.add_argument(
+            '--commit',
+            action='store_true',
+            dest='commit',
+            default=False,
+            help="Don't do a dry run, but actually archive the forms",
+        )
+
+    def handle(self, log_file, **options):
+        to_archive = []
+        episode_case_ids = CaseAccessors(domain).get_case_ids_in_domain("episode")
+        for episode_case_id in with_progress_bar(episode_case_ids[:100]):
+            nikshay_to_archive = []
+            dots_99_to_archvie = []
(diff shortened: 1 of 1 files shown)


custom/enikshay/management/commands/archive_extra_repeater_forms.py
```diff
@@ -0,0 +1,56 @@
...
+        to_archive = []
+        episode_case_ids = CaseAccessors(domain).get_case_ids_in_domain("episode")
+        for episode_case_id in with_progress_bar(episode_case_ids[:100]):
+            nikshay_to_archive = []
+            dots_99_to_archvie = []
+            case_forms = FormProcessorInterface(domain).get_case_forms(episode_case_id)
+            for form in case_forms:
+                if form.user_id in ("system", "", None) and form.metadata.username == "system":
+                    updates = get_case_updates(form)
+                    update_actions = [update.get_update_action() for update in updates if update.id == episode_case_id]
```
dimagimon: E501 line too long (119 > 115 characters)

custom/enikshay/management/commands/archive_extra_repeater_forms.py
```diff
@@ -0,0 +1,56 @@
...
+                    for action in update_actions:
+                        if isinstance(action, CaseUpdateAction):
+                            if set(action.dynamic_properties.keys()) == {"nikshay_registered", "nikshay_error"}:
+                                nikshay_to_archive.append(form)
+                            elif set(action.dynamic_properties.keys()) == {"dots_99_registered", "dots_99_error"}:
+                                dots_99_to_archvie.append(form)
+
+
+            nikshay_to_archive = nikshay_to_archive[:-1]
```
dimagimon: E303 too many blank lines (2)

custom/enikshay/management/commands/archive_extra_repeater_forms.py
```diff
@@ -0,0 +1,58 @@
...
+            action='store_true',
+            dest='commit',
+            default=False,
+            help="Don't do a dry run, but actually archive the forms",
+        )
+
+    def handle(self, log_file, **options):
+        to_archive = []
+        episode_case_ids = CaseAccessors(domain).get_case_ids_in_domain("episode")
+        for episode_case_id in with_progress_bar(episode_case_ids[:100]):
```
proteusvacuum: why `[:100]`?

NoahCarnahan: ah shit, meant to remove that :P (limited the number when I was testing it)

PR #15403, opened 2017-03-16
cleanup some ucr tests

emord:
cleans up the ucr columns test and splits them out into multiple classes

@gcapalbo

### DIFF
--- a/corehq/apps/userreports/tests/test_columns.py
+++ b/corehq/apps/userreports/tests/test_columns.py
@@ -2,13 +2,15 @@ from __future__ import absolute_import
 import uuid
 
 from sqlagg import SumWhen
-from django.test import SimpleTestCase, TestCase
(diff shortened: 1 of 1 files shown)


corehq/apps/userreports/tests/test_columns.py
```diff
@@ -134,7 +141,9 @@ def _new_case(self, properties):
             update=properties,
         ).as_xml()
         post_case_blocks([case_block], {'domain': self.domain})
-        return CommCareCase.get(id)
+        case = CommCareCase.get(id)
```
gcapalbo: looks like this is assuming the couch form processing backend is in use. should this use CaseAccessors?

emord: Yeah. that's probably true. although I don't think it's necessary as it was already here and that's still the default

gcapalbo (approved):

PR #15404, opened 2017-03-16
DHIS2 logging

kaapstorm:
This is the last PR in the DHIS2 series. This one just adds API request logging.

I've built on top of standard Python logging, and used Django generic views to keep things simple and minimal. Currently the templates are **extremely** minimal. One more commit to follow to make them more usable.

:tropical_fish: 

@snopoke cc @benrudolph @sheelio


corehq/apps/dhis2/api.py
```diff
@@ -0,0 +1,120 @@
+import httplib
```
dimagimon: F401 'httplib' imported but unused

corehq/apps/dhis2/handlers.py
```diff
@@ -0,0 +1,61 @@
...
+    >>> logger = logging.getLogger('my_logger')
+    >>> logger.debug({'some_property': 'xyzzy', 'another_property': 6.283185})
+
+    """
+    def __init__(self, model_class, *args, **kwargs):
+        super(DjangoModelHandler, self).__init__(*args, **kwargs)
+        self.model_class = model_class
+
+    def emit(self, record):
+        created = time.localtime(record.created)
```
snopoke: Generally better to store dates at UTC. Also you can use [datetime.utcfromtimestamp](https://docs.python.org/2/library/datetime.html#datetime.datetime.utcfromtimestamp)

corehq/apps/dhis2/api.py
```diff
@@ -0,0 +1,120 @@
...
+
+    @staticmethod
+    def json_or_error(response):
+        """
+        Return HTTP status, JSON
+
+        :raises JsonApiError: if HTTP status is not in the 200 (OK) range
+        """
+        if 200 <= response.status_code < 300:
+            return response.status_code, response.json() if response.content else None
```
snopoke: I don't see a huge advantage to not returning the full response object here. At least I think this should return a namedtuple.

corehq/apps/dhis2/handlers.py
```diff
@@ -0,0 +1,61 @@
...
+        ))
+        self.model_class.objects.create(**kwargs)
+
+
+class JsonApiHandler(DjangoModelHandler):
+    """
+    Used for logging JSON API requests to JsonApiLog instances.
+    """
+    def __init__(self, *args, **kwargs):
+        super(JsonApiHandler, self).__init__(JsonApiLog, *args, **kwargs)
```
snopoke: There doesn't seem to be much point to this subclass

snopoke (commented):
In general I quite like this patterns since it decouples recording the log from the rest of it but I'm not sure it warrants this level of separation here since it's currently just a single usage.

I realise that this could be used in future though it's not clear when if that would happen.

Also, in order to be generalizable I think a few pieces are missing:

* being able to separate the logging for different usages
* being able to use the API without doing any logging

snopoke:
@kaapstorm Seems like a good thing to send past architecture review.

corehq/apps/dhis2/models.py
```diff
@@ -111,156 +115,17 @@ def get_dataset(self):
...
+    domain = models.CharField(max_length=126, db_index=True)  # 126 seems to be a popular length
+    timestamp = models.DateTimeField()
+    request_method = models.CharField(max_length=12)
+    request_url = models.CharField(max_length=255)
+    request_headers = jsonfield.JSONField(blank=True)
+    request_params = jsonfield.JSONField(blank=True)
+    request_body = jsonfield.JSONField(blank=True)
+    request_error = models.TextField(blank=True)
+    response_status = models.IntegerField()
+    response_body = models.TextField(blank=True)
```
kaapstorm: Does Dimagi have a position in the NULL vs empty string debate? i.e. Which (one or more) of the following is true:
- [ ] NULL == "I don't know" 
- [ ] NULL == "I do know and it's not applicable" 
- [ ] NULL == "I do know and it has no value"

I'm happy to adopt whatever convention we've chosen. Your comment makes me think that our convention is either "True, True, True" or "True, True, False". For now I'll assume "True, True, False" and change this accordingly.

If we don't have a convention, I'll take the initial position of "True, I'm open to persuasion, False." 

If the convention is "True, False, False" then there would be no change because all the values are known. If it's "True, True, (True|False)" then all those properties should be nullable (including `response_status` if the request errors before it is sent).

snopoke: `blank` is only useful for validating data from a django form

From a data point of view I always prefer to separate NULL from blank and assume:
* NULL = there is no value
* blank = there is a value and it's an empty string

corehq/apps/dhis2/api.py
```diff
@@ -0,0 +1,134 @@
...
+        """
+        Unpack path and data from args and kwargs in order to log them separately
+        """
+        logger.log(log_level, {
+            'domain': json_api_request.domain_name if json_api_request.domain_name is not None else '[N/A]',
+            'request_method': func.__name__.upper(),
+            'request_url': json_api_request.server_url + path,
+            'request_headers': json.dumps(json_api_request.headers),
+            'request_params': json.dumps(params),
+            'request_body': '' if data is None else json.dumps(data, default=json_serializer),
```
snopoke: could do this: https://github.com/dimagi/commcare-hq/blob/420c9b5b0cb13731933f3e85c9c7a9e4b7478f68/corehq/apps/repeaters/repeater_generators.py#L153-L153

corehq/apps/dhis2/api.py
```diff
@@ -0,0 +1,134 @@
...
+
+    @log_request
+    def put(self, path, data, **kwargs):
+        path = path.lstrip('/')
+        headers = dict(self.headers, **{'Content-type': 'application/json'})
+        json_data = json.dumps(data, default=json_serializer)
+        try:
+            response = requests.put(self.server_url + path, json_data, headers=headers, auth=self.auth, **kwargs)
+        except requests.RequestException as err:
+            raise JsonApiError(str(err))
```
snopoke: This error handling could also go in a decorator: `wrap_errors`?

corehq/apps/dhis2/forms.py
```diff
@@ -13,10 +13,20 @@
 from corehq.apps.style import crispy as hqcrispy
 
 
+LOG_LEVEL_CHOICES = (
+    (logging.CRITICAL, 'Critical'),
+    (logging.ERROR, 'Error'),
+    (logging.WARNING, 'Warning'),
+    (logging.INFO, 'Info'),
+    (logging.DEBUG, 'Debug'),
+)
```
snopoke: Should only include the ones we actually log things at - so at present it would just be `error` and `info`

corehq/apps/dhis2/handlers.py
```diff
@@ -0,0 +1,61 @@
...
+    def model_class(self):
+        """
+        Lazyload model class so that we can configure it in settings.py before we can load app modules
+        """
+        if isinstance(self._model_class, basestring):
+            self._model_class = import_string(self._model_class)
+        return self._model_class
+
+    def emit(self, record):
+        created = time.localtime(record.created)
```
snopoke: Still think you could use `datetime.fromtimestamp` or `datetime.utcfromtimestamp`

corehq/apps/dhis2/models.py
```diff
@@ -111,156 +116,17 @@ def get_dataset(self):
...
+    domain = models.CharField(max_length=126, db_index=True)  # 126 seems to be a popular length
+    timestamp = models.DateTimeField()
+    request_method = models.CharField(max_length=12)
+    request_url = models.CharField(max_length=255)
+    request_headers = jsonfield.JSONField(blank=True)
+    request_params = jsonfield.JSONField(blank=True)
+    request_body = jsonfield.JSONField(blank=True, null=True)  # NULL for GET, but POST can take an empty body
+    request_error = models.TextField(null=True)
+    response_status = models.IntegerField(null=True)
+    response_body = models.TextField(blank=True, null=True)
```
snopoke: I'd add in the log level here as well so that it can be filtered on the UI

corehq/apps/domain/templates/domain/admin/dhis2/log_detail.html
```diff
@@ -0,0 +1,61 @@
...
+                    <th>{% trans "Request headers" %}:</th>
+                    <td><pre>{{ log.request_headers }}</pre></td>
+                </tr>
+                {% if log.request_params %}
+                <tr>
+                    <th>{% trans "Request params" %}:</th>
+                    <td><pre>{{ log.request_params }}</pre></td>
+                </tr>
+                {% endif %}
+                {% if log.request_body %}
```
snopoke: I'd prefer to just leave these blank instead of completely removing them from the UI

corehq/apps/domain/views.py
```diff
@@ -3183,6 +3183,58 @@ def page_context(self):
         }
 
 
+class Dhis2LogListView(BaseAdminProjectSettingsView, ListView):
```
snopoke: Any reason you chose to use this and not `GenericTabularReport` e.g. https://github.com/dimagi/commcare-hq/blob/898e161ef3ef375d573fac58612027a53db1fb74/corehq/apps/domain/views.py#L2268-L2268

corehq/apps/dhis2/dbaccessors.py
```diff
@@ -12,3 +12,16 @@ def get_dhis2_connection(domain_name):
         reduce=False,
     ).first()
     return Dhis2Connection.wrap(result['doc']) if result else None
+
+
+@quickcache(['domain_name'])
```
snopoke: not sure how often the data would change but if it's not often you should increase the timeout here (or just clear the cache when it does change)

corehq/apps/dhis2/api.py
```diff
@@ -0,0 +1,132 @@
...
+        path = path.lstrip('/')
+        return self.send_request(
+            requests.delete, self.server_url + path, headers=self.headers, auth=self.auth, **kwargs
+        )
+
+    def post(self, path, data, **kwargs):
+        path = path.lstrip('/')
+        # Make a copy of self.headers so as not to set content type on requests that don't send content
+        headers = dict(self.headers, **{'Content-type': 'application/json'})
+        json_data = json.dumps(data, default=json_serializer)
```
snopoke: any reason not to use the `DjangoJSONEncoder`?

corehq/apps/dhis2/migrations/0003_jsonapilog_log_level.py
```diff
@@ -0,0 +1,20 @@
...
+        ('dhis2', '0002_auto_20170322_1323'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='jsonapilog',
+            name='log_level',
+            field=models.IntegerField(null=True),
+        ),
+    ]
```
snopoke: can merge this migration with the previous one

corehq/apps/dhis2/api.py
```diff
@@ -0,0 +1,132 @@
...
+            'request_method': method_func.__name__.upper(),
+            'request_url': json_api_request.server_url + path,
+            'request_headers': json.dumps(json_api_request.headers),
+            'request_params': json.dumps(params),
+            'request_body': '' if data is None else json.dumps(data, default=json_serializer),
+
+            'request_error': request_error,
+            'response_status': response_status,
+            'response_body': response_body,
+        })
```
snopoke: Can you articulate why you think this indirection is necessary / good? You could as easily call a function to save the model.

snopoke:
@kaapstorm I think this is close - would be good to push on it so that we can get it out the door

corehq/apps/dhis2/api.py
kaapstorm: By "indirection", do you mean using the logging framework? Sorry I didn't reply to your previous comment, "I'm not sure it warrants this level of separation here since it's currently just a single usage."

I think you make a good point, and I agree with the simplicity of your approach. I chose to use logging instead of just saving a model because as soon as we had two log levels it started to feel like I was reinventing a wheel that came standard with Python. (TBH, I don't think the logging framework is very elegant, but I thought it was better to build with what Python provided instead of in parallel, even if it turns out to be once-off.)

corehq/apps/dhis2/api.py
kaapstorm: Ooo! I didn't know about that. Changed.

corehq/apps/dhis2/api.py
snopoke: That would make sense if we were actually using the log level configuration. What we're doing is choosing to call `log` based on the logic in the request wrapper which bypasses the logging frameworks own log level configuration.

Also the logging framework doesn't really fit well here since you can't adjust it 'per project'.

snopoke (approved):

snopoke:
any reason not to merge this?

corehq/apps/dhis2/models.py
```diff
@@ -1,176 +1,139 @@
-import httplib
-import json
-import logging
-from collections import namedtuple
+from itertools import chain
```
dimagimon: F401 'chain' imported but unused

PR #15408, opened 2017-03-16
App Diff feedback

benrudolph:
@orangejenny this makes the form diff compatible with app manager v2, translates the modal. it also changes a few things i noticed when trying on india (broken down by commit)

cc: @biyeun

### DIFF
--- a/corehq/apps/app_manager/static/app_manager/js/app_diff.js
+++ b/corehq/apps/app_manager/static/app_manager/js/app_diff.js
@@ -47,13 +47,13 @@ hqDefine('app_manager/js/app_diff.js', function () {
     var reverse = hqImport('hqwebapp/js/urllib.js').reverse;
     var sanitize = DOMPurify.sanitize;
 
-    var init = function(selector, appIdOne, appIdTwo, options) {
+    var init = function(selector, options) {
         var $el = $(selector);
 
         if (!$el.length) {
             throw new Error(selector + ' does not resolve to an element');
         }
-        return new AppDiff($el, appIdOne, appIdTwo, options);
+        return new AppDiff($el, options);
     };
 
     /**
@@ -66,10 +66,8 @@ hqDefine('app_manager/js/app_diff.js', function () {
      * appIdOne {String} - An app id
      * appIdTwo {String} - An app id
      */
-    var AppDiff = function($el, appIdOne, appIdTwo, options) {
+    var AppDiff = function($el, options) {
         var self = this;
-        self.appIdOne = appIdOne;
-        self.appIdTwo = appIdTwo;
         self.$el = $el;
         self.controller = new Controller();
         self.options = options || {};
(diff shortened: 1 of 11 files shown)


corehq/apps/app_manager/static/app_manager/js/app_diff.js
```diff
@@ -47,13 +47,13 @@ hqDefine('app_manager/js/app_diff.js', function () {
     var reverse = hqImport('hqwebapp/js/urllib.js').reverse;
     var sanitize = DOMPurify.sanitize;
 
-    var init = function(selector, appIdOne, appIdTwo, options) {
+    var init = function(selector, options) {
         var $el = $(selector);
 
         if (!$el.length) {
-            throw new Error(selector + ' does not resolve to an element');
+            window.console.warn(selector + ' does not resolve to an element');
```
benrudolph: it's pretty bad to break the release page if the element for this doesn't exist, so decided to change it to a warning

corehq/apps/app_manager/static/app_manager/js/app_diff.js
```diff
@@ -82,10 +80,11 @@ hqDefine('app_manager/js/app_diff.js', function () {
          *
          * When called renders a diff into the $el element
          */
-        self.renderDiff = function() {
+        self.renderDiff = function(appIdOne, appIdTwo) {
```
benrudolph: this enables use to reused the cached responses to the server because we don't have to instantiate a new AppDiff object for every diff we create

corehq/apps/app_manager/static/app_manager/js/app_diff.js
```diff
@@ -82,10 +80,11 @@ hqDefine('app_manager/js/app_diff.js', function () {
...
+            self.$el.text(gettext('Crunching app diff...'));
```
benrudolph: ensures we show loading text on reload

corehq/apps/app_manager/static/app_manager/js/releases.js
```diff
@@ -258,24 +258,15 @@ hqDefine('app_manager/js/releases.js', function () {
...
 
-        self.previousReleaseId = function(index) {
-            var savedApp;
-            for (var i = index + 1; i < self.savedApps().length; i++) {
-                savedApp = self.savedApps()[i];
-                if (savedApp.is_released()) {
-                    return savedApp.id();
-                }
+        self.previousBuildId = function(index) {
+            if (self.savedApps()[index + 1]) {
```
benrudolph: this was just dumb. i was comparing the latest starred version instead of the latest builds

corehq/apps/app_manager/templates/app_manager/v2/partials/releases.html
```diff
@@ -12,9 +12,21 @@
        });
     });
     $(function () {
+        hqImport('hqwebapp/js/urllib.js').registerUrl(
+            'form_data',
+            '/a/' + '{{ domain }}' + '/apps/view/---/form_data/'
```
orangejenny: Surprised to see a url built like this instead of using `registerurl` in a template. Although I think releases.js does some funky stuff with urls, is that why this is like this?

corehq/apps/app_manager/static/app_manager/js/app_diff.js
orangejenny: nice

orangejenny:
Looks good, thanks for the quick turnaround!

orangejenny (approved):

corehq/apps/app_manager/templates/app_manager/v2/partials/releases.html
benrudolph: huh i never knew about that. played with it a bit briefly and got some errors, i'll update in a followup. (have another PR for case diff stuff)

PR #15417, opened 2017-03-17
fix expanded audio

emord:
@millerdev


millerdev (approved):
A fine stop-gap measure. As discussed offline, it would be nice to change `expanded-audio` and `inline-video` to just `audio` and `video` (and change Vellum to use those in paths).

PR #15424, opened 2017-03-17
add ucr static partitions

emord:
goes with https://github.com/dimagi/commcare-hq-deploy/pull/208

safe to be merged without it though

@gcapalbo @snopoke


snopoke:
Was thinking about this again and another way we could subdivide is by topic (ie. have different pillows for forms and cases)

emord:
Oh yeah. that could be interesting. similar to what we do with the custom fluff pillows now.

PR #15426, opened 2017-03-17
Block 99DOTS submissions from test locations

proteusvacuum:
@mkangia @NoahCarnahan 
cc @sravfeyn

### DIFF
--- a/custom/enikshay/integrations/nikshay/repeaters.py
+++ b/custom/enikshay/integrations/nikshay/repeaters.py
@@ -121,13 +121,18 @@ class NikshayTreatmentOutcomeRepeater(CaseRepeater):
             return False
 
         episode_case_properties = episode_case.dynamic_case_properties()
-        return (
+        if (
             episode_case_properties.get('nikshay_registered', 'false') == 'true' and
             episode_case_properties.get('nikshay_id', False) and
             case_properties_changed(episode_case, [TREATMENT_OUTCOME]) and
-            episode_case_properties.get(TREATMENT_OUTCOME) in treatment_outcome.keys() and
-            not is_submission_from_test_location(get_person_case_from_episode(episode_case.domain, episode_case))
-        )
+            episode_case_properties.get(TREATMENT_OUTCOME) in treatment_outcome.keys()
+        ):
+            try:
+                person_case = get_person_case_from_episode(episode_case.domain, episode_case)
+            except ENikshayCaseNotFound:
+                return False
+            return not is_submission_from_test_location(person_case)
+        return False
 
 
 def person_hiv_status_changed(case):
--- a/custom/enikshay/integrations/nikshay/repeaters.py
+++ b/custom/enikshay/integrations/nikshay/repeaters.py
@@ -1,7 +1,6 @@
 from django.utils.translation import ugettext_lazy as _
 from django.core.urlresolvers import reverse
 
-from corehq.apps.locations.models import SQLLocation
 from corehq.apps.repeaters.models import CaseRepeater
 from corehq.form_processor.models import CommCareCaseSQL
 from corehq.toggles import NIKSHAY_INTEGRATION
@@ -14,8 +13,9 @@ from custom.enikshay.case_utils import (
     get_person_case_from_episode,
     get_open_episode_case_from_person,
 )
-from custom.enikshay.exceptions import NikshayLocationNotFound, ENikshayCaseNotFound
+from custom.enikshay.exceptions import ENikshayCaseNotFound
 from custom.enikshay.const import TREATMENT_OUTCOME, EPISODE_PENDING_REGISTRATION
+from custom.enikshay.integrations.utils import is_submission_from_test_location
 from custom.enikshay.integrations.ninetyninedots.repeaters import case_properties_changed
 from custom.enikshay.integrations.nikshay.field_mappings import treatment_outcome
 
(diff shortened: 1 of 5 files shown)


mkangia:
@proteusvacuum Can you add the method `is_test_submission` to TreatmentOutcomeRepeater as well? I am guessing this would get deployed earlier than my PR, so good to get it through.

mkangia (approved):

custom/enikshay/integrations/ninetyninedots/repeaters.py
```diff
@@ -110,7 +111,15 @@ def allowed_to_forward(self, case):
...
             return False
 
-        return registered_episode and props_changed
+        return (
+            registered_episode
+            and props_changed
+            and (
+                case.type == CASE_TYPE_PERSON and not is_submission_from_test_location(person_case)
+                or case.type == CASE_TYPE_EPISODE and not is_submission_from_test_location(
+                        get_person_case_from_episode(episode_case.domain, episode_case))
```
dimagimon: E126 continuation line over-indented for hanging indent

custom/enikshay/integrations/nikshay/repeaters.py
```diff
@@ -117,24 +117,17 @@ def get_custom_url(cls, domain):
...
-
-def test_submission(person_case):
-    try:
-        phi_location = SQLLocation.objects.get(location_id=person_case.owner_id)
-    except SQLLocation.DoesNotExist:
-        raise NikshayLocationNotFound(
-            "Location with id {location_id} not found. This is the owner for person with id: {person_id}"
-            .format(location_id=person_case.owner_id, person_id=person_case.case_id)
+            episode_case_properties.get(TREATMENT_OUTCOME) in treatment_outcome.keys() and
+            not is_submission_from_test_location(get_person_case_from_episode(episode_case.domain, episode_case))
```
mkangia: @proteusvacuum Can you please add this in a try block to avoid a case where `get_person_case_from_episode` raises a `NikshayCaseNotFound`. It seems that this won't ever happen but good to add since this exception causes phone submissions to break. Have added similar checks to other calls to such methods as well.

custom/enikshay/integrations/nikshay/repeaters.py
```diff
@@ -117,24 +117,17 @@ def get_custom_url(cls, domain):
 
     def allowed_to_forward(self, episode_case):
         allowed_case_types_and_users = self._allowed_case_type(episode_case) and self._allowed_user(episode_case)
+        if not allowed_case_types_and_users:
+            return False
```
mkangia: 👍

custom/enikshay/integrations/utils.py
```diff
@@ -1,13 +1,14 @@
...
         phi_location = SQLLocation.objects.get(location_id=person_case.owner_id)
     except SQLLocation.DoesNotExist:
-        raise NikshayLocationNotFound(
-            "Location with id {location_id} not found. This is the owner for person with id: {person_id}"
-            .format(location_id=person_case.owner_id, person_id=person_case.case_id)
-        )
+        message = ("Location with id {location_id} not found. This is the owner for person with id: {person_id}"
+                   .format(location_id=person_case.owner_id, person_id=person_case.case_id))
+        notify_exception(None, message="[ENIKSHAY] {}".format(message))
+        return True
```
proteusvacuum: @mkangia what do you think of doing this here instead of failing hard here? I figured since you fall back to "yes" in the line below if the metadata isn't set, this might make sense too (if for some reason the location doesn't exist).

mkangia: I think i was fine with this failing hard here coz:
1. we should not let this proceed if the location a person is assigned to does not exist. Such a scenario is undesirable and would cause problems everywhere we fetch information for a location that owns the case. So good to raise it.

custom/enikshay/integrations/nikshay/repeaters.py
```diff
@@ -117,24 +117,22 @@ def get_custom_url(cls, domain):
...
-            .format(location_id=person_case.owner_id, person_id=person_case.case_id)
-        )
-    return phi_location.metadata.get('is_test', "yes") == "yes"
+        ):
+            try:
+                person_case = get_person_case_from_episode(episode_case.domain, episode_case)
+            except ENikshayCaseNotFound:
+                return False
+            return not is_submission_from_test_location(person_case)
+        return False
```
mkangia: @proteusvacuum i think this is slightly more complicated than needed. We can just move this try block up and then have all conditions at once.
Though the change is definitely better since it avoids fetching an extra case and may be avoids a DB fetch in most of the situation. Just wondering if it written in a better manner. If not, then this its fine.

proteusvacuum: Agreed that this is a little verbose. I think I want to do the check here since it gets run on every case update... so ideally we only do it if all the other checks fail.

proteusvacuum: https://github.com/dimagi/commcare-hq/pull/15426/commits/6309ece717d23762a56c095c985d139442bb56c9

custom/enikshay/integrations/utils.py
proteusvacuum: 👍 changed it back.

proteusvacuum:
✅

PR #15427, opened 2017-03-17
unicode report builder choices

calellowitz:
@NoahCarnahan @millerdev 
to address the first traceback in http://manage.dimagi.com/default.asp?249347#1304078
There was a non ascii character that was failing to be encoded. I have not been able to reproduce this bug however, so its possible my understanding is off. Does this seem safe @millerdev

### DIFF
--- a/corehq/apps/userreports/indicators/factory.py
+++ b/corehq/apps/userreports/indicators/factory.py
@@ -76,10 +76,10 @@ def _build_choice_list_indicator(spec, context):
     base_display_name = wrapped_spec.display_name
 
     def _construct_display(choice):
-        return '{base} ({choice})'.format(base=base_display_name, choice=choice)
+        return u'{base} ({choice})'.format(base=base_display_name, choice=choice)
 
     def _construct_column(choice):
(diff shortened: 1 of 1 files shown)


millerdev (approved):
My guess is that none of the string formatting parameters are bytes, so yes, seems like the right thing to do.

corehq/apps/userreports/indicators/factory.py
```diff
@@ -76,10 +76,10 @@ def _build_choice_list_indicator(spec, context):
     base_display_name = wrapped_spec.display_name
 
     def _construct_display(choice):
-        return '{base} ({choice})'.format(base=base_display_name, choice=choice)
+        return u'{base} ({choice})'.format(base=base_display_name, choice=choice)
 
     def _construct_column(choice):
-        return '{col}_{choice}'.format(col=spec['column_id'], choice=choice)
+        return u'{col}_{choice}'.format(col=spec['column_id'], choice=choice)
```
NoahCarnahan: This one might need to strip/convert unicode characters, as I'm not sure how postgres handles unicode in column names. If you've tested this locally, it's probably okay though.

PR #15430, opened 2017-03-18
Mk repeater edit view

mkangia:
Add Edit views for 
1. Repeaters
2. CaseRepeaters
3. FormRepeaters

Best review 🏛


corehq/apps/domain/urls.py
```diff
@@ -161,6 +167,12 @@
         name=AddCaseRepeaterView.urlname),
     url(r'^forwarding/new/(?P<repeater_type>\w+)/$', AddRepeaterView.as_view(), name=AddRepeaterView.urlname),
     url(r'^forwarding/test/$', test_repeater, name='test_repeater'),
+    url(r'^forwarding/CaseRepeater/edit/(?P<repeater_id>\w+)/$', EditCaseRepeaterView.as_view(),
+        {'repeater_type': 'CaseRepeater'},name=EditCaseRepeaterView.urlname),
```
dimagimon: E231 missing whitespace after ','

corehq/apps/domain/urls.py
```diff
@@ -161,6 +167,12 @@
         name=AddCaseRepeaterView.urlname),
     url(r'^forwarding/new/(?P<repeater_type>\w+)/$', AddRepeaterView.as_view(), name=AddRepeaterView.urlname),
     url(r'^forwarding/test/$', test_repeater, name='test_repeater'),
+    url(r'^forwarding/CaseRepeater/edit/(?P<repeater_id>\w+)/$', EditCaseRepeaterView.as_view(),
```
benrudolph: nit: somewhat unconventional for us to put camel case in urls

corehq/apps/domain/views.py
```diff
@@ -2506,23 +2506,31 @@ def page_context(self):
...
+
     def make_repeater(self):
-        repeater = self.repeater_class(
-            domain=self.domain,
-            url=self.add_repeater_form.cleaned_data['url'],
-            use_basic_auth=self.add_repeater_form.cleaned_data['use_basic_auth'],
-            username=self.add_repeater_form.cleaned_data['username'],
-            password=self.add_repeater_form.cleaned_data['password'],
-            format=self.add_repeater_form.cleaned_data['format']
-        )
```
benrudolph: why change this?

corehq/apps/repeaters/models.py
```diff
@@ -315,6 +318,16 @@ def handle_exception(self, exception, repeat_record):
         generator = self.get_payload_generator(self.format_or_default_format())
         return generator.handle_exception(exception, repeat_record)
 
+    @property
+    def is_form_repeater(self):
```
benrudolph: i think conventionally we define this on the base class as NotImplemented and then on the FormRepeater class we have it `return True`

corehq/apps/repeaters/views.py
```diff
@@ -26,13 +31,69 @@ def dispatch(self, request, *args, **kwargs):
...
-    def make_repeater(self):
-        repeater = super(AddCaseRepeaterView, self).make_repeater()
+    def set_repeater_attr(self, repeater, cleaned_data):
+        repeater = super(AddCaseRepeaterView, self).set_repeater_attr(repeater, cleaned_data)
         repeater.white_listed_case_types = self.add_repeater_form.cleaned_data['white_listed_case_types']
         repeater.black_listed_users = self.add_repeater_form.cleaned_data['black_listed_users']
         return repeater
 
 
+class EditRepeaterView(AddRepeaterView):
```
benrudolph: it seems semantically strange that the EditRepeaterView inherits from the AddRepeaterView. maybe there should just be a mixin for the functions they share?

benrudolph (commented):
a couple of nit picky comments. will give you a chance to respond. overall looks great! very excited to have this

corehq/apps/domain/urls.py
mkangia: Used the same convention views & urls to add repeaters. No strong opinion in favour or against it.

corehq/apps/domain/views.py
mkangia: I broke this method to make this more flexible so that this class can be extended easily. For ex extracted out initialize_repeater so that it can be initialised with an existing repeater when editing and not always a new repeater. Similarly extracted post_save

corehq/apps/repeaters/models.py
mkangia: Don't think we can do that here. Not sure if there is better way for this since all objects are fetched as instance of `Repeater` class and the doc_type would be the Repeater which won't always be FormRepeater or CaseRepeater but a class inherited from either of those.

corehq/apps/repeaters/views.py
mkangia: All methods in `EditRepeaterView` are over writing methods from `AddRepeaterView`. Not sure if extracting all other methods in a mixin would make this better or may be a `BaseRepeaterView` class would be better?

corehq/apps/hqwebapp/templatetags/hq_shared_tags.py
```diff
@@ -652,3 +652,8 @@ def render(self, context):
     nodelist = NodeList([FakeNode()])
 
     return AddToBlockNode(nodelist, 'initial_page_data')
+
+
+@register.filter
+def concat_list(str_elements):
+    return ", ".join(str_elements)
```
dimagimon: W292 no newline at end of file

corehq/apps/repeaters/views.py
benrudolph: personally i think a BaseRepeaterView would make more sense to me

corehq/apps/hqwebapp/templatetags/hq_shared_tags.py
benrudolph: this might be useful: https://docs.djangoproject.com/en/1.10/ref/templates/builtins/#join

corehq/apps/domain/urls.py
benrudolph: gotcha. if it's already being used, that's fine with me

benrudolph (commented):
looking good

corehq/apps/hqwebapp/templatetags/hq_shared_tags.py
mkangia: silly me. thanks

mkangia:
Restarting this [here](https://github.com/dimagi/commcare-hq/pull/18579)

PR #15431, opened 2017-03-19
disable device id logging for demo users

sravfeyn:


corehq/apps/mobile_auth/views.py
```diff
@@ -66,8 +66,11 @@ def fetch_key_records(request, domain):
     payload = FetchKeyRecords(domain, user_id, last_issued).get_payload()
     device_id = request.GET.get('device_id')
     if device_id and isinstance(request.couch_user, CommCareUser):
-        request.couch_user.update_device_id_last_used(device_id)
-        request.couch_user.save()
+        if not request.couch_user.is_demo_user:
```
snopoke: nit, could make this part of the previous `if` condition

corehq/apps/mobile_auth/views.py
```diff
@@ -66,8 +66,11 @@ def fetch_key_records(request, domain):
...
+            # temporary NIC fix; db conflicts happen when multiple logins happen
+            # because of following, so disable it for demo users
```
snopoke: Don't think this should be temporary - it doesn't make sense to keep track of device ID's for demo users.

corehq/apps/mobile_auth/views.py
snopoke: Ah, I see, `is_demo_user` is only an attribute on `CommCareUser`

corehq/apps/mobile_auth/views.py
sravfeyn: Cool, will update it. (not updating right now, assuming that will restart the build again)

snopoke: @sravfeyn fine to merge this once you've updated the comment

PR #15432, opened 2017-03-19
add migration metadata

nickpell:
Adds ```migration_created_case: 'true'``` to drtb-hiv-referral, and records the nikshay ID of the migrated record in ```migration_created_from_record```.  This way it is easily to tell for each case if 1) it was created in a migration and 2) which nikshay record it came from.

I was able to figure out this stuff when deleting records for GU-MSN, but it would have been easier to just have this data stored in the case.

Recommend 🐡 

@mkangia

### DIFF
--- a/custom/enikshay/nikshay_datamigration/factory.py
+++ b/custom/enikshay/nikshay_datamigration/factory.py
@@ -149,6 +149,7 @@ class EnikshayCaseFactory(object):
                     'sex': self.patient_detail.sex,
 
                     'migration_created_case': 'true',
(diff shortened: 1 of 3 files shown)


mkangia (approved):
Did you run this for the data that was already migrated as well?

nickpell:
I don't think I'm going to run this on the already migrated data right now, since I don't have a way to update single properties without possibly overwriting changes that users might have made.  If I update the script to enable that, I'll update the already migrated data.

PR #15438, opened 2017-03-20
don't call "add inferred case props" unless we need to

snopoke:
This should relieve some stress on celery for large user imports


benrudolph (approved):
nice, good call

kaapstorm (approved):

PR #15441, opened 2017-03-20
option to merge es docs

sravfeyn:
http://manage.dimagi.com/default.asp?249683#1303997 @snopoke

### DIFF
--- a/corehq/ex-submodules/pillowtop/tests/test_elasticsearch.py
+++ b/corehq/ex-submodules/pillowtop/tests/test_elasticsearch.py
@@ -158,7 +158,11 @@ class TestSendToElasticsearch(SimpleTestCase):
         doc = {'_id': uuid.uuid4().hex, 'doc_type': 'MyCoolDoc', 'property': 'foo'}
         self._send_to_es_and_check(doc)
 
-    def _send_to_es_and_check(self, doc, update=False, delete=False, esgetter=None):
+    def _send_to_es_and_check(self, doc, update=False, es_merge_update=False,
+                              delete=False, esgetter=None):
+        if update and es_merge_update:
+            old_doc = self.es.get_source(self.index, TEST_INDEX_INFO.type, doc['_id'])
+
         send_to_elasticsearch(
             index=self.index,
             doc_type=TEST_INDEX_INFO.type,
@@ -167,6 +171,7 @@ class TestSendToElasticsearch(SimpleTestCase):
             name='test',
             data=doc,
             update=update,
+            es_merge_update=es_merge_update,
             delete=delete,
             except_on_failure=True,
             retries=1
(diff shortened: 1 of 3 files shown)


sravfeyn:
We need to run `./manage.py ptop_reindexer_v2 domain --in-place` on all envs after this gets merged

corehq/ex-submodules/pillowtop/tests/test_elasticsearch.py
```diff
@@ -195,6 +205,17 @@ def test_replace_doc(self):
...
+    def test_merge_doc(self):
+        doc = {'_id': uuid.uuid4().hex, 'doc_type': 'MyCoolDoc', 'property': 'bar'}
+        self._send_to_es_and_check(doc)
+
+        update = doc.copy()
+        del update['property']
+        update['new_prop'] = 'new_val'
+        self._send_to_es_and_check(update, update=True)
+        # merging should still keep old 'property'
+        self._send_to_es_and_check(update, update=True, es_merge_update=True)
```
snopoke: Don't think this test is actually working. `self._send_to_es_and_check(update, update=True)` will replace the doc with the new one so `self._send_to_es_and_check(update, update=True, es_merge_update=True)` doesn't actually test the merging.

I think you just want to do something like this
```python
self._send_to_es_and_check('_id': doc['_id'], 'new_prop': 'new_val', update=True, es_merge_update=True)
# Doc in ES should contain new prop and old prop - make sure old prop is still there.
updated_doc = self.es.get_source(self.index, TEST_INDEX_INFO.type, doc['_id'])
self.assertIn('property', doc)
self.assertEqual(doc['property'], 'bar')
```

snopoke (changes requested):

corehq/ex-submodules/pillowtop/tests/test_elasticsearch.py
sravfeyn: I have updated[ `_send_to_es_and_check` ](https://github.com/dimagi/commcare-hq/pull/15441/files#diff-56cd81bc88100a7354b093f20e6da7c3R184) to reflect above, doesn't look right?

snopoke: I did see that but I still don't think this test is correct since it's calling `_send_to_es_and_check` with the same doc that it's using to check against.

I guess if you just got rid of the call to `_send_to_es_and_check` on line 215 then it would be OK.

calellowitz: @snopoke updated to reflect last comment, let me know if this is good to go

orangejenny (approved):

calellowitz:
im merging this now based on approval from @orangejenny so we can kick off the reindex and the issue was in a test, but if you want more @snopoke, happy to revisit tomorrow

PR #15445, opened 2017-03-20
Catch bad dates in HIV Test repeater

proteusvacuum:
@mkangia Found it.

To test this remotely all I had to do was:
`Repeater({repeater_id}).register({case})`. This returned a new RepeatRecord which I was able to then call `get_payload` on. This returned a `ValueError` because the dates are set to `''`.

This error was getting swallowed because of this: http://manage.dimagi.com/default.asp?249894

It looks like in the app, when dates aren't relevant in the Edit/Update Treatment Details form they are set to blanks, rather than not getting set (check [this form submission](https://india.commcarehq.org/a/enikshay/reports/form_data/78bf0ed6-ff79-496d-ac42-0e858cf29d22/download/), for [this case](https://india.commcarehq.org/a/enikshay/reports/case_data/88e9f05f-9644-44f9-9466-243456dc52b3/#!history?form_id=78bf0ed6-ff79-496d-ac42-0e858cf29d22)). This could pose a problem with the checks you are making in the `allowed_to_forward` method -- can you verify that this repeater won't get triggered every time one of those forms is submitted? 

FYI @NoahCarnahan

### DIFF
--- a/custom/enikshay/integrations/nikshay/repeater_generator.py
+++ b/custom/enikshay/integrations/nikshay/repeater_generator.py
@@ -181,19 +181,11 @@ class NikshayHIVTestPayloadGenerator(BasePayloadGenerator):
         properties_dict = {
             "PatientID": episode_case_properties.get('nikshay_id'),
             "HIVStatus": hiv_status.get(person_case_properties.get('hiv_status')),
-            "HIVTestDate": datetime.datetime.strptime(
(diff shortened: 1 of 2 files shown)


mkangia:
Thank You for finding this @proteusvacuum. 

> can you verify that this repeater won't get triggered every time one of those forms is submitted?
If you are asking when all is fine, we want this repeater to get triggered on each form submission to check if any relevant dates are changed and send that update.

mkangia (approved):

PR #15446, opened 2017-03-20
adding needed keyword

calellowitz:
@orangejenny


orangejenny (approved):

calellowitz:
@snopoke @sravfeyn 
The updated calculations pass through this function before hitting the one modified in https://github.com/dimagi/commcare-hq/pull/15441
This addresses that

snopoke:
Thanks, surprising that the tests passed - perhaps good to add a test for this function.

orangejenny:
I don't think we have a good way to test the ES part of calculated properties. (We've been bitten by this kind of bug before, and added a test for calculated properties, but it doesn't include the ones that use ES.)

PR #15450, opened 2017-03-21
Don't override celery queue

benrudolph:
@gcapalbo this prevents us from overriding the queue name in the task decorator

cc: @biyeun

### DIFF
--- a/corehq/apps/reports/tasks.py
+++ b/corehq/apps/reports/tasks.py
@@ -59,18 +59,18 @@ from .models import (
     ReportNotification,
     UnsupportedScheduledReportError,
 )
-from .scheduled import get_scheduled_reports
+from .scheduled import get_scheduled_report_ids
 
 
 logging = get_task_logger(__name__)
 EXPIRE_TIME = 60 * 60 * 24
 
 
-def send_delayed_report(report):
+def send_delayed_report(report_id):
     """
     Sends a scheduled report, via  celery background task.
     """
-    send_report.delay(report._id)
+    send_report.delay(report_id)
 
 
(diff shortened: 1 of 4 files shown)


corehq/apps/reports/tasks.py
```diff
@@ -70,17 +70,7 @@ def send_delayed_report(report):
...
-
-def get_report_queue(report):
-    # This is a super-duper hacky, hard coded way to deal with the fact that MVP reports
-    # consistently crush the celery queue for everyone else.
-    # Just send them to their own longrunning background queue
-    if report.domain in get_mvp_domains():
-        return 'background_queue'
-    else:
-        return 'celery'
+    send_report.delay(report._id)
```
benrudolph: actually can probably refactor this so that we don't pass in the entire report

corehq/apps/reports/tasks.py
```diff
@@ -59,28 +59,18 @@
...
-from .scheduled import get_scheduled_reports
+from .scheduled import get_scheduled_report_ids
 
 
 logging = get_task_logger(__name__)
 EXPIRE_TIME = 60 * 60 * 24
 
 
-def send_delayed_report(report):
+def send_delayed_report(report_id):
```
gcapalbo: not blocking, but we can probably just get rid of this function now

gcapalbo (approved):

corehq/apps/reports/tasks.py
benrudolph: thanks you're totally right. for now i'm going to leave because i'm just trying to keep my head above water with the interrupt queue

PR #15452, opened 2017-03-21
Dont send test submissions for TreatmentOutcomeRepeater

proteusvacuum:
@mkangia I missed one (test wasn't testing the right thing...)
@NoahCarnahan 
cc: @sravfeyn


mkangia (approved):

mkangia:
@proteusvacuum one possible concern:
when a patient is updated to be as cured, the occurrence is closed and the person is archived. See this [case](https://india.commcarehq.org/a/enikshay/reports/case_data/ff42d7a2-7a37-4769-955d-b619343473b4/)
What would happen in this case?
will this check return true(which is needed) since person and occurrence would get updated after episode?

proteusvacuum:
@mkangia will let you follow up on whether the repeater is doing the thing you expect (might be worth just writing a test for the scenario you set out).

PR #15455, opened 2017-03-21
Fix repeater report

benrudolph:
@proteusvacuum this should fix the report to include those records http://manage.dimagi.com/default.asp?249894

cc: @biyeun

### DIFF
--- a/corehq/apps/repeaters/tests/test_dbaccessors.py
+++ b/corehq/apps/repeaters/tests/test_dbaccessors.py
@@ -30,6 +30,13 @@ class TestRepeatRecordDBAccessors(TestCase):
             repeater_id=cls.repeater_id,
             next_check=before,
         )
+        failed_hq_error = RepeatRecord(
+            domain=cls.domain,
+            failure_reason='Some python error',
+            repeater_id=cls.repeater_id,
+            next_check=before,
+        )
+        failed_hq_error.doc_type += '-Failed'
         success = RepeatRecord(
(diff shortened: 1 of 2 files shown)


proteusvacuum (approved):
Yay.

mkangia:
Thank You @benrudolph

benrudolph:
you are welcome! thank you for reporting

biyeun:
:fire:

proteusvacuum:
(This db took a while to reindex last time, might want to hold off until the weekend)

mkangia:
@benrudolph can we include some timestamp on that key to have some sense of ordering when fetching them for the listing of repeat records?

benrudolph:
hmm yeah, i can revert this and add a timestamp. it sounds like though we might want to take another pass on this report given the amount of work you guys are doing with repeaters

mkangia:
Thanks @benrudolph. 
From what i know we are not modifying anything about how repeaters work but mostly using them and hence notifying any bugs/challenges we find. I think this report is still a good to have and should not wait.
@proteusvacuum?

benrudolph:
(don't think we should wait on this particular change either, just saying
it may help in the long term to take another look at revamping it if you're
going to be doing a lot more repeater work)

proteusvacuum:
Yeah, there's a bunch of things that we should add to this report (e.g. the payload_doc id). I think fine to leave this as-is, and @mkangia and I can work on figuring it out.

PR #15460, opened 2017-03-21
add fall back value for art initiated

mkangia:


custom/enikshay/integrations/nikshay/repeater_generator.py
```diff
@@ -189,7 +189,8 @@ def get_payload(self, repeat_record, person_case):
             "HIVTestDate": _format_date(person_case_properties, 'hiv_test_date'),
             "CPTDeliverDate": _format_date(person_case_properties, 'cpt_initiation_date'),
             "ARTCentreDate": _format_date(person_case_properties, 'art_initiation_date'),
-            "InitiatedOnART": art_initiated.get(person_case_properties.get('art_initiated', 'no')),
+            "InitiatedOnART": art_initiated.get(
+                person_case_properties.get('art_initiated', 'no'), art_initiated['no']),
```
snopoke: Is this because of bad data in the case?

snopoke (approved):

custom/enikshay/integrations/nikshay/repeater_generator.py
mkangia: @snopoke This is added here for empty string but yes would take care of any bad values as well.

NoahCarnahan (approved):

mkangia:
@snopoke @NoahCarnahan thank you for the review. Added one more small change for IP address that was requested.

custom/enikshay/integrations/nikshay/repeater_generator.py
```diff
@@ -63,13 +64,14 @@ def _get_credentials(self, repeat_record):
 
     def _base_properties(self, repeat_record):
         username, password = self._get_credentials(repeat_record)
+        server_ip = socket.gethostbyname(socket.gethostname())
```
NoahCarnahan: The [docs](https://docs.python.org/2.7/library/socket.html#socket.gethostname) for `socket.gethostname()` mention that "This operation assumes that there is a valid address-to-host mapping for the host, and the assumption does not always hold." Do you know what happens if that assumption does not hold? This might raise an exception, and we might want to catch it, because we might not want to prevent that from forwarding the record (although maybe we do, I don't know how critical the IP info is).

mkangia: i checked this on softlayer proxy machine and it returned an IP address. I think it should then work from celery machine as well where it would get triggered.
If it raises an exception it would be logged in the repeat record and i can ensure that it works as expected for first occurrence.
Its not a critical info but i am guessing we should not hardcode it.

proteusvacuum:
@mkangia  Does this provide the public IP address or the address of the machine that this was run on? Is there a reason for us to send them our internal IPs?

mkangia:
Hey Farid this should send the public IP of the machine this was run on, which should be the celery machine i think. Soni requested that the IP address should be of the machine from where the request is sent and not `127.0.0.1`. Initially it was phone but since the web sends it, it was fine to send web machine IP address.

proteusvacuum:
weird. I feel like sending the public IP of our proxy server might make more sense, but if this is what they want... 
I don't really know what they use this information for, but I can't imagine having the IP addresses of our celery machines would be useful in any way... sigh.

mkangia:
@proteusvacuum its just for identification of source of request from different machines, mainly for tracking. I don't see any obvious use of it though but its a requirement.
This would end up being the celery machine since the tasks are picked up in background. It won't matter if its the proxy machine IP as well but then for that we would need to hardcode it right?

PR #15462, opened 2017-03-22
better async indicators

emord:
Fixes the issues with async indicators. Modeled this after the current repeat record queue

@snopoke buddy @sravfeyn


corehq/apps/userreports/tasks.py
```diff
@@ -1,27 +1,35 @@
 from __future__ import absolute_import
-from datetime import datetime
-import sys
+from datetime import datetime, timedelta
```

corehq/apps/userreports/models.py
```diff
@@ -683,6 +687,48 @@ def _get_report_config(cls, static_config, domain):
...
+                    doc_id=doc_id,
+                    doc_type=doc_type,
+                    domain=domain,
+                    pillow=pillow_id,
+                    indicator_config_ids=table_ids
+                )
+            else:
+                current_table_ids = indicator.indicator_config_ids
+                new_table_ids = list(set(current_table_ids).union(set(table_ids)))
+                indicator.indicator_config_ids = new_table_ids
```
emord: If the indicator is already queued up to be put into celery, then this will add any new UCR data sources to that indicator

corehq/apps/userreports/tasks.py
```diff
@@ -178,25 +186,52 @@ def _run_report(backend_to_use):
...
+    queue=settings.CELERY_PERIODIC_QUEUE,
+)
+def queue_async_indicators():
+    start = datetime.utcnow()
+    cutoff = start + ASYNC_INDICATOR_QUEUE_TIME
+    time_for_crit_section = ASYNC_INDICATOR_QUEUE_TIME.seconds - 10
+
+    with CriticalSection(['queue-async-indicators'], timeout=time_for_crit_section):
+        redis_client = get_redis_client().client.get_client()
+        indicators = AsyncIndicator.objects.all()[:10000]
```
emord: This will limit the queue to 10k indicators at one time. That should make sure it does not overwhelm rabbitmq

corehq/apps/userreports/tasks.py
```diff
@@ -178,25 +186,52 @@ def _run_report(backend_to_use):
...
+        paginator = Paginator(indicators, 1000)
+        for page in paginator.page_range:
+            for indicator in paginator.page(page).object_list:
+                now = datetime.utcnow()
+                if now > cutoff:
```
benrudolph: never loved the way repeaters used this to cut off the task. maybe this is slightly better and more reliable? http://docs.celeryproject.org/en/latest/userguide/configuration.html#std:setting-task_soft_time_limit

corehq/apps/userreports/tests/test_pillow.py
```diff
@@ -251,36 +251,109 @@ def test_process_doc_from_sql_stale(self):
...
-    #                     update={'update-prop-child': i}
-    #                 ).as_xml()
-    #             ], domain=self.domain
-    #         )
-    #     self.pillow.process_changes(since=since, forever=False)
-    #     rows = self.adapter.get_query_object()
-    #     self.assertEqual(rows.count(), 0)
-    #     errors = PillowError.objects.filter(doc_id='child-id', pillow=self.pillow.pillow_id)
-    #     self.assertEqual(errors.count(), 1)
-    #     errors.delete()
```
snopoke: any reason to keep this around?

corehq/apps/userreports/tests/test_pillow.py
```diff
@@ -251,36 +251,109 @@ def test_process_doc_from_sql_stale(self):
...
+            row = rows[0]
+            self.assertEqual(int(row.parent_property), i)
+
+            errors = PillowError.objects.filter(doc_id='child-id', pillow=self.pillow.pillow_id)
+            self.assertEqual(errors.count(), 0)
+            self.assertEqual(indicators.count(), 0)
+
+    @patch('corehq.apps.userreports.tasks._get_config_by_id')
+    def test_async_save_fails(self, config):
+        config.return_value = None
```
snopoke: commed here would be useful to indicate that this is why it's going to fail

corehq/apps/userreports/tasks.py
snopoke: A bit concerned that this could still end up with thousands of tasks in the celery queue if they aren't getting processed fast enough. There's nothing here to stop us from adding more tasks into the queue.

corehq/apps/userreports/tests/test_pillow.py
emord: This was deleted

corehq/apps/userreports/tasks.py
emord: There's a lock for 6 hours on queueing the same indicator and it's only possible to queue 10k indicators at one time because this will only pull the oldest created indicators thanks to https://github.com/dimagi/commcare-hq/pull/15462/files#diff-3af365e79b93f051d416585bcba61b7eR703

Maybe we should bump up the 6 hours? 6 hours assumes about 1 indicator every 2 seconds with one thread. if we have 100 concurrency (same as repeaters on prod) then each indicator could take 200 seconds before we run into issues.

We also have proactive warnings about too many tasks in the celery queue so I don't think it's something that will slide by unnoticed (assuming datadog can be put on nic eventually)

corehq/apps/userreports/models.py
```diff
@@ -683,6 +687,48 @@ def _get_report_config(cls, static_config, domain):
         return ReportConfiguration.wrap(doc)
 
 
+class AsyncIndicator(models.Model):
```
sravfeyn: docstrings (nice to have)

corehq/apps/userreports/tests/test_pillow.py
```diff
@@ -251,36 +251,109 @@ def test_process_doc_from_sql_stale(self):
...
+        self.assertEqual(indicators.count(), 0)
+        self.pillow.process_changes(since=since, forever=False)
+        self.assertEqual(indicators.count(), 1)
+
+        queue_async_indicators()
+        rows = self.adapter.get_query_object()
+        self.assertEqual(rows.count(), 0)
+
+        errors = PillowError.objects.filter(doc_id='child-id', pillow=self.pillow.pillow_id)
+        self.assertEqual(errors.count(), 0)
```
sravfeyn: Since there are multiple assertions in the test, a small explanatory comment wherever applicable is nice to have

corehq/apps/userreports/tasks.py
```diff
@@ -178,25 +186,52 @@ def _run_report(backend_to_use):
...
-                error_id
-            )
-        )
-    else:
-        error.delete()
+
+        redis_client = get_redis_client().client.get_client()
+        queued_lock_key = _get_indicator_queued_lock_key(indicator)
+        lock = redis_client.lock(queued_lock_key)
+        release_lock(lock, degrade_gracefully=True)
```
sravfeyn: For my understanding, how will the lock get releases if there is an exception above?

sravfeyn:
I am good, will let simon sign off

corehq/apps/userreports/models.py
```diff
@@ -683,6 +687,48 @@ def _get_report_config(cls, static_config, domain):
...
+
+    class Meta(object):
+        unique_together = ('doc_id', 'pillow',)
+        ordering = ["date_created"]
+
+    @classmethod
+    def update_indicators(cls, change, pillow, table_ids):
+        doc_id = change.id
+        pillow_id = pillow.pillow_id
+        with CriticalSection([get_async_indicator_modify_lock_key(doc_id, pillow_id)]):
```
emord: I don't think so. Locks are want we would want in the following situation:

add indicator for doc_id 1, pillow static, table_id first_table
indicator begins processing in save_document
save_document begins processing
update this indicator to add table_id second_table (the record that save_document has only grabbed that table)
save_document ends processing
save_document deletes the record without processing second_table

I'm not sure of a good way to solve that without locks because UCRs can be over mulitple sql databases or ES so there wouldn't be a way to roll those back without a large refactor

corehq/apps/userreports/tasks.py
emord: The lock has a timeout https://github.com/dimagi/commcare-hq/pull/15462/files#diff-1f15a4ea1fec5c4ee5fa155d4d80c3d7R208 so that will time out and it will be requeued in celery

emord (commented):
updated with comments and better var names

corehq/apps/userreports/tasks.py
emord: I don't think it's very easy to do that because we use a deprecated decorator https://github.com/celery/celery/blob/3.1/celery/task/base.py#L171-L179

would be good to see how easy it is to upgrade to the recommended way to make tasks. not sure how hard that will be though

corehq/apps/userreports/tasks.py
sravfeyn: 👍

emord:
I think I fixed the test. I didn't have celery propagating errors locally. I think i've responded to all feedback @snopoke @benrudolph @sravfeyn

corehq/apps/userreports/tasks.py
snopoke: +1 to upgrading

corehq/apps/userreports/models.py
```diff
@@ -683,6 +687,53 @@ def _get_report_config(cls, static_config, domain):
...
+                    domain=domain,
+                    pillow=pillow_id,
+                    indicator_config_ids=config_ids
+                )
+            else:
+                current_config_ids = indicator.indicator_config_ids
+                new_config_ids = list(set(current_config_ids).union(set(config_ids)))
+                indicator.indicator_config_ids = new_config_ids
+
+            indicator.save()
```
snopoke: could avoid saving this if there have been no change

snopoke (approved):

corehq/apps/userreports/models.py
emord: avoided: https://github.com/dimagi/commcare-hq/pull/15462/commits/dbdc37b226583fc794127a741f44d8e8a2eef12d

emord:
:white_check_mark: (made a couple of commits since approval)

snopoke:
What's the rollout plan for this?

emord:
@snopoke 

1) manual testing on prod in a test domain. (will automate some submissions to get a few things in the queue at the same time)
2) add some of the more expensive ucrs that block on I/O (related_doc/get_case_forms etc). may start with some of the test enikshay UCRs. I also have some data on expensive UCRs on prod now that would be good to add as well
3) Add this to ICDS once we have some relief from the backlog. Sheel and I have improvements that I'm optimistic will help us with that

snopoke:
👍 would be good to have a Trello card or something to track the status.

emord:
@snopoke https://trello.com/c/kg34AZI5/686-asynchronous-ucr-indicators

snopoke:
Thanks

PR #15463, opened 2017-03-22
Use correct properties in episode data source

kkrampa:
@NoahCarnahan 
http://manage.dimagi.com/default.asp?249815


kkrampa:
@NoahCarnahan Could you restart the build?

PR #15470, opened 2017-03-22
Better ensure doc exists

benrudolph:
@snopoke how about this? no assumptions about whether SQL case is deleted or missing. context: http://manage.dimagi.com/default.asp?249534

cc: @millerdev

### DIFF
--- a/corehq/ex-submodules/pillowtop/feed/interface.py
+++ b/corehq/ex-submodules/pillowtop/feed/interface.py
@@ -56,13 +56,15 @@ class Change(object):
     def set_document(self, document):
         self.document = document
 
-    def get_document(self):
+    def get_document(self, raise_error=False):
         if not self.document and self.document_store and not self._document_checked:
             try:
                 self.document = self.document_store.get_document(self.id)
             except DocumentNotFoundError:
                 self.document = None
                 self._document_checked = True  # set this flag to avoid multiple redundant lookups
+                if raise_error:
+                    raise
         return self.document
 
     def __repr__(self):
--- a/corehq/ex-submodules/pillowtop/feed/interface.py
+++ b/corehq/ex-submodules/pillowtop/feed/interface.py
@@ -45,6 +45,7 @@ class Change(object):
         self.deleted = deleted
         self.metadata = metadata
         self.document_store = document_store
+        self.error_raised = None
         self._document_checked = False
         self._dict = {
             'id': self.id,
@@ -56,15 +57,14 @@ class Change(object):
     def set_document(self, document):
         self.document = document
 
-    def get_document(self, raise_error=False):
+    def get_document(self):
         if not self.document and self.document_store and not self._document_checked:
             try:
                 self.document = self.document_store.get_document(self.id)
-            except DocumentNotFoundError:
+            except DocumentNotFoundError, e:
                 self.document = None
                 self._document_checked = True  # set this flag to avoid multiple redundant lookups
-                if raise_error:
-                    raise
+                self.error_raised = e
         return self.document
 
     def __repr__(self):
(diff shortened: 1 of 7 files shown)


corehq/ex-submodules/pillowtop/utils.py
```diff
@@ -268,11 +268,14 @@ def _convert_rev_to_int(rev):
...
-    doc = change.get_document()
-    if doc is None:
-        pillow_logging.warning("Unable to get document from change: {}".format(change))
-        raise DocumentNotFoundError()  # force a retry
+    try:
+        change.get_document(raise_error=True)
+    except DocumentMissingError:
+        raise
+    except DocumentNotFoundError:
+        pass
```
snopoke: If we don't raise the error here won't it error somewhere else because the doc returned is `None`

benrudolph: i had hoped that the doc_filter_fn would auto filter out docs that are None, but that doesn't look like the case actually. in this scenario though we should just return without error in the processor

benrudolph: actually need to prevent caching in get_document as well

corehq/ex-submodules/pillowtop/utils.py
```diff
@@ -12,7 +12,7 @@
 
 from pillowtop.exceptions import PillowNotFoundError
 from pillowtop.logger import pillow_logging
-from pillowtop.dao.exceptions import DocumentMismatchError, DocumentNotFoundError
+from pillowtop.dao.exceptions import DocumentMismatchError, DocumentNotFoundError, DocumentMissingError
```
dimagimon: F401 'DocumentNotFoundError' imported but unused

benrudolph:
Ok did another pass on this. Now `change.get_document` stores the error raised. This way we don't have to fetch the document twice to get the error. Secondly, I've added a check in all the places where we call it to ensure we handle the case when doc is `None`

corehq/ex-submodules/pillowtop/feed/interface.py
```diff
@@ -60,9 +61,10 @@ def get_document(self):
         if not self.document and self.document_store and not self._document_checked:
             try:
                 self.document = self.document_store.get_document(self.id)
-            except DocumentNotFoundError:
+            except DocumentNotFoundError, e:
                 self.document = None
                 self._document_checked = True  # set this flag to avoid multiple redundant lookups
+                self.error_raised = e
```
snopoke: Will this still be necessary if we swap this code around:
```python
- doc = change.get_document()
  ensure_document_exists(change)
  ensure_matched_revisions(change)
+ doc = change.get_document()
```

benrudolph: no, but it makes the calls order dependent which i think would be fairly confusing if anyone were to ever modify the code. bad to store the error raised?

snopoke: not necessarily, just an unusual pattern

corehq/ex-submodules/pillowtop/feed/interface.py
```diff
@@ -60,9 +61,10 @@ def get_document(self):
         if not self.document and self.document_store and not self._document_checked:
             try:
                 self.document = self.document_store.get_document(self.id)
-            except DocumentNotFoundError:
+            except DocumentNotFoundError, e:
```
millerdev: nit: should say `except DocumentNotFoundError as e`

benrudolph: 👍 i can never remember which one is supposed to be used

emord: if you use flake8 to check code you can install https://docs.openstack.org/developer/hacking/#python-3-x-compatibility. i attempted to add it to dimagimon but never quite figured it out

benrudolph: i do use flake8. nice i'll look into that

PR #15471, opened 2017-03-22
Use relational field in select related

kkrampa:
@snopoke 
https://manage.dimagi.com/default.asp?250071
Is that look good to you?


gcapalbo (approved):
i just noticed the same thing: https://github.com/dimagi/commcare-hq/pull/15473

closing my pr since we made the same change

PR #15473, opened 2017-03-22
Fix select_related reference

gcapalbo:
https://manage.dimagi.com/default.asp?250078
@proteusvacuum


gcapalbo:
already fixed here: https://github.com/dimagi/commcare-hq/pull/15471

PR #15478, opened 2017-03-22
Latitude and longitude in degrees, not meters

wpride:
These measurements are in degrees, not meters


locale/en/LC_MESSAGES/django.po
```diff
@@ -14833,11 +14833,11 @@ msgid "Messages"
 msgstr ""
 
 #: corehq/apps/export/models/new.py:2101
-msgid "{}: latitude (meters)"
+msgid "{}: latitude (degrees)"
```
benrudolph: we don't usually modify these by hand, django generates these

benrudolph (commented):
nice good catch

PR #15480, opened 2017-03-22
eNikshay ID generation

esoergel:
Adding autogeneration of issuer_id to nikshay commcare users on save.


custom/enikshay/models.py
```diff
@@ -0,0 +1,62 @@
+from django.db import models
+from dimagi.ext.couchdbkit import Document, StringProperty, IntegerProperty
+from corehq.util.decorators import serial_task
+
+
+# option 1
+# store a SQL model mapping user_id to pk (nikshay id)
+# The nice thing about this approach is that django and postgres handle all the concurrency issues for us!
+# also, this makes it easy to look up a user by nikshay id
```
nickpell: I'm a fan for this reason

custom/enikshay/models.py
```diff
@@ -0,0 +1,62 @@
...
+# the downside is that it's a db row for every user.
```
nickpell: This doesn't bother me - I still expect this table to be small enough that its size won't matter.

esoergel:
Just built out the rest of this, ready for real review.
@nickpell

esoergel:
@proteusvacuum FYI, you may be interested.  First commit shows the two approaches I considered, ended up going with the first.

nickpell (approved):

custom/enikshay/models.py
```diff
@@ -0,0 +1,11 @@
+from django.db import models
+
+
+class IssuerId(models.Model):
+    """
+    This model is used to ensure unique, incrementing issuer IDs for users,
+    and to look up a user given an issuer ID.
+    obj.pk represents the serial issuer ID, later representations will be added as fields
```
proteusvacuum: is there a reason you didn't name this `issuer_id` to be explicit?

custom/enikshay/user_setup.py
```diff
@@ -168,6 +169,22 @@ def set_available_tests(location, location_form):
         location.metadata['tests_available'] = 'cbnaat'
 
 
+def save_user_callback(sender, couch_user, **kwargs):
+    if toggles.ENIKSHAY.enabled(couch_user.domain):
+        set_issuer_id(couch_user.domain, couch_user)
```
proteusvacuum: do we need this for every user?

proteusvacuum (commented):
Looks good to me. Had a couple of questions.

custom/enikshay/models.py
esoergel: I thought about it, but it seemed surprising to have the object primary key renamed.  I'm guessing this will grow aliases for various representations of the issuer_id.  No strong feelings about it, though.

custom/enikshay/user_setup.py
esoergel: I don't know, actually awaiting clarification on that.  One thing that's not obvious is that `couch_user` here will only ever be a `CommCareUser` (the signal is `commcare_user_post_save`).  I tried naming the parameter `commcare_user`, but Django requires that signal receivers have parameter names that match up with the `providing_args`.

custom/enikshay/models.py
proteusvacuum: sounds fine if we use an alias for `issuer_id`. I don't really like the idea of accessing the `pk` to also mean 'issuer_id'.

proteusvacuum: although I guess the model is `IssuerId`, so I suppose it could make sense.

proteusvacuum (approved):

PR #15481, opened 2017-03-22
fix old exports page in django 1.10

nickpell:
http://manage.dimagi.com/default.asp?250116

@emord

### DIFF
--- a/corehq/apps/style/forms/widgets.py
+++ b/corehq/apps/style/forms/widgets.py
@@ -129,7 +129,7 @@ class _Select2Mixin(object):
         js = ('select2-3.4.5-legacy/select2.js',)
 
     def render(self, name, value, attrs=None, choices=()):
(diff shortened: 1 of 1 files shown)


nickpell:
cc: code buddy @esoergel

emord (approved):

benrudolph:
do we not have to reconcile the `choices` param in some way?

nickpell:
@benrudolph It seems to work fine without ```choices```.  I don't think it was being used in the first place.  Compare [current docs](https://docs.djangoproject.com/en/1.10/ref/forms/widgets/#django.forms.Widget.render) with [past docs](https://docs.djangoproject.com/en/1.9/ref/forms/widgets/#django.forms.Widget.render).

benrudolph:
weird, ok makes sense. thanks

PR #15484, opened 2017-03-22
log time for expensive ucrs

emord:
@snopoke as discussed on #pillows Logs the time for a specific table on a specific doc

@sravfeyn

### DIFF
--- a/corehq/apps/userreports/pillow.py
+++ b/corehq/apps/userreports/pillow.py
@@ -19,12 +19,30 @@ from corehq.sql_db.connections import connection_manager
 from corehq.util.soft_assert import soft_assert
 from fluff.signals import get_migration_context, get_tables_to_rebuild, reformat_alembic_diffs
 from pillowtop.checkpoints.manager import PillowCheckpoint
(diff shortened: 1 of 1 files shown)


sravfeyn:
Nice! I was just wondering if we had this info (given UCRs getting stalled lately)

corehq/apps/userreports/pillow.py
```diff
@@ -27,6 +28,23 @@
...
+        result = method(*args, **kw)
+        te = datetime.now()
+        seconds = (te - ts).total_seconds()
+        if seconds > 0.1:
+            table = args[1]
+            doc = args[2]
+            message = u"UCR data source {} on doc_id {} took {} seconds to process".format(
+                table.config._id, doc['_id'], seconds
+            )
+            pillow_logging.warning(message)
```
sravfeyn: Where can we see pillow warning logs? The logging level for pillowtop is `ERROR`

snopoke: nice catch - @emord you'll have to update the logging config in settings.py

PR #15485, opened 2017-03-23
helpful feature-flag 404

sravfeyn:
@emord


orangejenny:
Nice - for some reason I'd been thinking that you'd need to update a bunch of individual pages to make this work, but this makes perfect sense.

PR #15486, opened 2017-03-23
Only add export name prefix if the sheet name is duplicated

benrudolph:
@NoahCarnahan what do you think about this? now we only add the export name prefix if the sheet name is duplicated. i think this is the best scenario. for people who just bulk export a bunch of "Forms" sheets, it will automatically make them unique sheet names. For people who directly specify their sheet names, we'll leave them be.

https://manage.dimagi.com/default.asp?249738

cc: @millerdev

### DIFF
--- a/corehq/apps/export/export.py
+++ b/corehq/apps/export/export.py
@@ -1,6 +1,7 @@
 import contextlib
 import os
 import tempfile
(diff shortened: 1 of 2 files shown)


NoahCarnahan:
👍

PR #15487, opened 2017-03-23
Fix repeater report and -Failed

benrudolph:
Sorry the other PR got messed up

### DIFF
--- a/corehq/apps/repeaters/dbaccessors.py
+++ b/corehq/apps/repeaters/dbaccessors.py
@@ -33,7 +33,7 @@ def get_repeat_record_count(domain, repeater_id=None, state=None):
         endkey = [domain, repeater_id, {}]
     elif repeater_id and state:
         startkey = [domain, repeater_id, state]
-        endkey = [domain, repeater_id, state]
+        endkey = [domain, repeater_id, state, {}]
     elif not repeater_id and state:
         ids = sorted(_get_repeater_ids_by_domain(domain))
         if not ids:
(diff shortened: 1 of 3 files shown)


proteusvacuum (approved):

corehq/apps/repeaters/dbaccessors.py
```diff
@@ -70,23 +70,23 @@ def get_paged_repeat_records(domain, skip, limit, repeater_id=None, state=None):
...
         'limit': limit,
         'skip': skip,
+        'descending': True,
     }
 
     if repeater_id and not state:
-        kwargs['startkey'] = [domain, repeater_id]
-        kwargs['endkey'] = [domain, repeater_id, {}]
+        kwargs['endkey'] = [domain, repeater_id]
+        kwargs['startkey'] = [domain, repeater_id, {}]
```
millerdev: I'm not understanding why `startkey` and `endkey` got swapped here (with empty dict in `startkey`), but not on line 36 (where `endkey` has the empty dict). I also have never quite understood the purpose of the empty dict, so maybe if I understood that it would make more sense.

millerdev (commented):
I'm not sure if my comment is a blocker or just me misunderstanding. If not, this is approved.

corehq/apps/repeaters/dbaccessors.py
benrudolph: good question, so this has to do with how couch treats `descending=True`. this changes the order of the index in couch:

> View rows are sorted by the key; specifying descending=true will reverse their order. Note that the descending option is applied before any key filtering, so you may need to swap the values of the startkey and endkey options to get the expected results. The sorting itself is described in ViewCollation.

https://wiki.apache.org/couchdb/HTTP_view_API

The reason you don't see me switching the endkey and start key on line 36 is that i don't use descending=True since i'm just getting the count (it's a different function).

As for why i'm now using an empty dict, i think it's best understood if i use an example. say you have an index that looks like this:

```
['domain1', 'repeater-1', 'PENDING', '2017']
['domain1', 'repeater-1', 'PENDING', '2018']
['domain1', 'repeater-2', 'FAIL', '2017']
['domain1', 'repeater-2', 'PENDING', '2017']
```
those are ordered the way couch would order the keys in an index (repeater-2 is string comparison greater than repeater-1). when i say i want `startkey=['domain1', 'repeater-1', 'PENDING']` and `endkey=['domain1', 'repeater-1', 'PENDING',{}]`, it means i want everything in between those two keys. `{}` is a way of specifying the greatest value in a string comparison. since everything is less than `{}` it will essentially ignore the timestamp param, which is what i want. i'm only using the timestamp param to have couch order the results. does that make sense? happy to offline

benrudolph: just found this: https://developer.couchbase.com/documentation/mobile/current/guides/couchbase-lite/native-api/query/index.html#querying-key-ranges. it may be a better explanation than i just gave

millerdev: yes, that's helpful. I figured it was something like that, but found it difficult to search for `{}` in the couch docs on querying. (I didn't look very hard). Thanks for the detailed explanation.

benrudolph:
merging and preindexing

PR #15488, opened 2017-03-23
Disallow appending twice

benrudolph:
https://manage.dimagi.com/default.asp?250139 -- manish i'm guessing you tried to send the failed repeat record again and it failed on the same exact thing causing it to add a `-Failed-Failed` doc type

### DIFF
--- a/corehq/apps/repeaters/const.py
+++ b/corehq/apps/repeaters/const.py
@@ -5,6 +5,7 @@ MAX_RETRY_WAIT = timedelta(days=7)
 MIN_RETRY_WAIT = timedelta(minutes=60)
 CHECK_REPEATERS_INTERVAL = timedelta(minutes=5)
 CHECK_REPEATERS_KEY = 'check-repeaters-key'
+REPEATER_FAILED_SUFFIX = '-Failed'
 
 POST_TIMEOUT = 45  # seconds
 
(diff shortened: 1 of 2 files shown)


proteusvacuum:
lol. Is the reason the `-Failed` suffix is added so that the other dbaccessors don't pull those records out of the db? It seems redundant with the `repeater.failure_reason`, and seems to be adding complexity everywhere (also, there is no corresponding `-Cancelled` doctype). 

Might be out of scope for interrupt, but couldn't we just set the `doc.succeeded` to `false` here instead? I think that already filters out failed records elsewhere else...

benrudolph:
Yeah i'm guessing it was first done so it could be removed from the view without changing the view? i don't know, it's a good chunk of tech debt though. i was also scratching my head when looking at this. i agree it might be very easy to remove this completely (or very subtly break something)

proteusvacuum (approved):

benrudolph:
ok a couple of things

removed useless exception that has been removed from the codebase: https://github.com/dimagi/commcare-hq/pull/15488/commits/edaceab3fcffae9468a57a909463ca399cc64528
and also just used succeeded = False https://github.com/dimagi/commcare-hq/pull/15488/commits/0957b0956c04cc525defef1e553b51200f6802bc

PR #15490, opened 2017-03-23
Mk add password encoder

mkangia:
Changes for UATBC security audit
1. hash passwords when submitted to HQ on various pages [encoder](https://github.com/dimagi/commcare-hq/pull/15490/files#diff-cbd1f30f97ad4f6bc782bb148e3e9945),  [decoder](https://github.com/dimagi/commcare-hq/pull/15490/files#diff-261364967e94356f22e293ec0909b1bdR153)
2. prevent replay attack by maintaining history of login attempts [model](https://github.com/dimagi/commcare-hq/pull/15490/files#diff-53c4b5ba374ddc3b784e8979cdc6598bR32),  [used here](https://github.com/dimagi/commcare-hq/pull/15490/files#diff-261364967e94356f22e293ec0909b1bdR155)
3. restrict user when setting password to one from previously used 3 passwords [model](https://github.com/dimagi/commcare-hq/pull/15490/files#diff-53c4b5ba374ddc3b784e8979cdc6598bR38),  [validator added](https://github.com/dimagi/commcare-hq/pull/15490/files#diff-0f124b50bf8e5b06583471b5fb9196aaR9)
4. add commcare app setting to enable password hashing from phone: This needs to be done by adding property cc-use-pw-obfuscation=yes on app.

Details and Spec, [doc](https://docs.google.com/document/d/12_38B4LFhdzz7XSGDOKxoreBLis6th_O9r5yxTS18ZE/edit#bookmark=id.m1zxn02spgiv)

[QA](https://docs.google.com/document/d/1w_OoVJ14Pf5gwf5DyjG9DgF58lpYFSojYBZ8Mw_4nRI/edit)


dimagimon:
There are 8 errors:

* corehq/apps/domain/static/domain/js/encoder.js, line 1 - 'HexParsr' is defined but never used (no-unused-vars)
* corehq/apps/domain/static/domain/js/encoder.js, line 6 - 'b64EncodeUnicode' is not defined. (no-undef)
* corehq/apps/domain/static/domain/js/encoder.js, line 7 - Expected indentation of 8 space characters but found 6. (indent)
* corehq/apps/domain/static/domain/js/encoder.js, line 12 - Missing semicolon. (semi)
* corehq/apps/domain/static/domain/js/encoder.js, line 14 - 'p_encode' is not defined. (no-undef)
* corehq/apps/domain/static/domain/js/encoder.js, line 16 - Missing semicolon. (semi)
* corehq/apps/domain/static/domain/js/encoder.js, line 19 - 'p_encode' is not defined. (no-undef)
'b64EncodeUnicode' is not defined. (no-undef)
* corehq/apps/domain/static/domain/js/encoder.js, line 20 - Missing semicolon. (semi)

corehq/apps/domain/static/domain/js/encoder.js
```diff
@@ -0,0 +1,21 @@
+function HexParsr(salt1, salt2) {
```
orangejenny: Wrap as an `hqDefine` module rather than defining a global function

corehq/apps/domain/templates/login_and_password/login.html
```diff
@@ -1,10 +1,26 @@
...
 
 {% block title %}{% trans "Log into CommCare HQ" %}{% endblock title %}
 
 {% block page_content %}
     {% block login-content %}
     {% include "login_and_password/partials/login_full.html" %}
     {% endblock %}
+
+    {% if implement_password_hashing %}
+        <script src="{% static 'domain/js/encoder.js' %}"></script>
```
orangejenny: This should go in the `js` block

corehq/apps/domain/templates/login_and_password/login.html
```diff
@@ -1,10 +1,26 @@
...
+            $(document).ready(function(){
+                (function(){
+                  p_field=$("#id_auth-password");
+                  p_field.parents("form")[0].onsubmit = function(){
+                    p_encoder = new HexParsr("{{psalt1}}","{{psalt2}}");
+                    p_field.val(p_encoder.xyz(p_field.val()))
+                  };
+                })()
+            });
+        </script>
```
orangejenny: It should be external. It looks like this could just be moved into `encoder.js`

mkangia: Moving this to `encoder.js` itself along with logic to generate these values.
I added it like this thinking that static js should not contain any dynamic content since they are precompiled and stored. Guess thats not true.

corehq/apps/domain/templates/login_and_password/login.html
mkangia: using compress block.

corehq/apps/hqwebapp/views.py
```diff
@@ -70,6 +70,10 @@
 from corehq.util.datadog.metrics import JSERROR_COUNT
 from corehq.util.datadog.utils import create_datadog_event, log_counter, sanitize_url
 from corehq.util.view_utils import reverse
+from corehq.apps.hqwebapp.utils import (
+    generate_password_salt1,
+    generate_password_salt2,
+)
```
mkangia: Correct. That sounds better. Moved!

corehq/apps/domain/static/domain/js/encoder.js
mkangia: Considering we don't need any of this JS to use anywhere else is it fine to just have it within a function and not use hqDefine?
Plus for hqDefine would need to include hqModules JS on this page as well.

dimagimon:
There are 12 errors:

* corehq/apps/domain/static/domain/js/encoder.js, line 7 - 'b64EncodeUnicode' is not defined. (no-undef)
* corehq/apps/domain/static/domain/js/encoder.js, line 8 - Expected indentation of 12 space characters but found 10. (indent)
* corehq/apps/domain/static/domain/js/encoder.js, line 13 - Missing semicolon. (semi)
* corehq/apps/domain/static/domain/js/encoder.js, line 15 - 'p_encode' is not defined. (no-undef)
* corehq/apps/domain/static/domain/js/encoder.js, line 17 - Missing semicolon. (semi)
* corehq/apps/domain/static/domain/js/encoder.js, line 20 - 'p_encode' is not defined. (no-undef)
'b64EncodeUnicode' is not defined. (no-undef)
* corehq/apps/domain/static/domain/js/encoder.js, line 21 - Missing semicolon. (semi)
* corehq/apps/domain/static/domain/js/encoder.js, line 25 - Expected indentation of 8 space characters but found 6. (indent)
* corehq/apps/domain/static/domain/js/encoder.js, line 26 - Expected indentation of 8 space characters but found 6. (indent)
* corehq/apps/domain/static/domain/js/encoder.js, line 27 - Expected indentation of 10 space characters but found 8. (indent)
'p_encoder' is not defined. (no-undef)
* corehq/apps/domain/static/domain/js/encoder.js, line 28 - Expected indentation of 10 space characters but found 8. (indent)
'p_encoder' is not defined. (no-undef)
* corehq/apps/hqwebapp/utils.py, line 2 - F401 'uuid' imported but unused

dimagimon:
There are 12 errors:

* corehq/apps/domain/static/domain/js/encoder.js, line 7 - 'b64EncodeUnicode' is not defined. (no-undef)
* corehq/apps/domain/static/domain/js/encoder.js, line 8 - Expected indentation of 12 space characters but found 10. (indent)
* corehq/apps/domain/static/domain/js/encoder.js, line 9 - Expected indentation of 14 space characters but found 16. (indent)
* corehq/apps/domain/static/domain/js/encoder.js, line 11 - Missing semicolon. (semi)
* corehq/apps/domain/static/domain/js/encoder.js, line 15 - Missing semicolon. (semi)
* corehq/apps/domain/static/domain/js/encoder.js, line 18 - 'b64EncodeUnicode' is not defined. (no-undef)
* corehq/apps/domain/static/domain/js/encoder.js, line 20 - Missing semicolon. (semi)
* corehq/apps/domain/static/domain/js/encoder.js, line 24 - Expected indentation of 8 space characters but found 6. (indent)
* corehq/apps/domain/static/domain/js/encoder.js, line 25 - Expected indentation of 8 space characters but found 6. (indent)
* corehq/apps/domain/static/domain/js/encoder.js, line 26 - Expected indentation of 10 space characters but found 8. (indent)
'password_encoder' is not defined. (no-undef)
* corehq/apps/domain/static/domain/js/encoder.js, line 27 - Expected indentation of 10 space characters but found 8. (indent)
'password_encoder' is not defined. (no-undef)
* corehq/apps/hqwebapp/utils.py, line 2 - F401 'uuid' imported but unused

dimagimon:
There are 19 errors:

* corehq/apps/domain/static/domain/js/encoder.js, line 7 - 'b64EncodeUnicode' is not defined. (no-undef)
* corehq/apps/domain/static/domain/js/encoder.js, line 8 - Expected indentation of 12 space characters but found 10. (indent)
* corehq/apps/domain/static/domain/js/encoder.js, line 9 - Expected indentation of 14 space characters but found 16. (indent)
* corehq/apps/domain/static/domain/js/encoder.js, line 11 - Missing semicolon. (semi)
* corehq/apps/domain/static/domain/js/encoder.js, line 15 - Missing semicolon. (semi)
* corehq/apps/domain/static/domain/js/encoder.js, line 18 - 'b64EncodeUnicode' is not defined. (no-undef)
* corehq/apps/domain/static/domain/js/encoder.js, line 20 - Missing semicolon. (semi)
* corehq/apps/domain/static/domain/js/encoder.js, line 24 - Expected indentation of 8 space characters but found 6. (indent)
* corehq/apps/domain/static/domain/js/encoder.js, line 25 - Expected indentation of 8 space characters but found 6. (indent)
* corehq/apps/domain/static/domain/js/encoder.js, line 27 - 'password_encoder' is not defined. (no-undef)
Expected indentation of 14 space characters but found 12. (indent)
* corehq/apps/domain/static/domain/js/encoder.js, line 28 - Expected indentation of 14 space characters but found 12. (indent)
'password_encoder' is not defined. (no-undef)
* corehq/apps/domain/static/domain/js/encoder.js, line 31 - Expected indentation of 8 space characters but found 6. (indent)
* corehq/apps/domain/static/domain/js/encoder.js, line 32 - Expected indentation of 8 space characters but found 6. (indent)
* corehq/apps/domain/static/domain/js/encoder.js, line 36 - 'password_encoder' is not defined. (no-undef)
* corehq/apps/domain/static/domain/js/encoder.js, line 37 - 'password_encoder' is not defined. (no-undef)
* corehq/apps/domain/static/domain/js/encoder.js, line 39 - Missing semicolon. (semi)
* corehq/apps/domain/auth.py, line 10 - F401 'settings' imported but unused
* corehq/apps/hqwebapp/forms.py, line 11 - F401 'settings' imported but unused
* corehq/apps/hqwebapp/utils.py, line 2 - F401 'uuid' imported but unused

dimagimon:
There are 19 errors:

* corehq/apps/domain/static/domain/js/encoder.js, line 7 - 'b64EncodeUnicode' is not defined. (no-undef)
* corehq/apps/domain/static/domain/js/encoder.js, line 8 - Expected indentation of 12 space characters but found 10. (indent)
* corehq/apps/domain/static/domain/js/encoder.js, line 9 - Expected indentation of 14 space characters but found 16. (indent)
* corehq/apps/domain/static/domain/js/encoder.js, line 11 - Missing semicolon. (semi)
* corehq/apps/domain/static/domain/js/encoder.js, line 15 - Missing semicolon. (semi)
* corehq/apps/domain/static/domain/js/encoder.js, line 18 - 'b64EncodeUnicode' is not defined. (no-undef)
* corehq/apps/domain/static/domain/js/encoder.js, line 20 - Missing semicolon. (semi)
* corehq/apps/domain/static/domain/js/encoder.js, line 24 - Expected indentation of 8 space characters but found 6. (indent)
* corehq/apps/domain/static/domain/js/encoder.js, line 25 - Expected indentation of 8 space characters but found 6. (indent)
* corehq/apps/domain/static/domain/js/encoder.js, line 27 - 'password_encoder' is not defined. (no-undef)
Expected indentation of 14 space characters but found 12. (indent)
* corehq/apps/domain/static/domain/js/encoder.js, line 28 - Expected indentation of 14 space characters but found 12. (indent)
'password_encoder' is not defined. (no-undef)
* corehq/apps/domain/static/domain/js/encoder.js, line 31 - Expected indentation of 8 space characters but found 6. (indent)
* corehq/apps/domain/static/domain/js/encoder.js, line 32 - Expected indentation of 8 space characters but found 6. (indent)
* corehq/apps/domain/static/domain/js/encoder.js, line 36 - 'password_encoder' is not defined. (no-undef)
* corehq/apps/domain/static/domain/js/encoder.js, line 37 - 'password_encoder' is not defined. (no-undef)
* corehq/apps/domain/static/domain/js/encoder.js, line 39 - Missing semicolon. (semi)
* corehq/apps/domain/auth.py, line 10 - F401 'settings' imported but unused
* corehq/apps/hqwebapp/forms.py, line 11 - F401 'settings' imported but unused
* corehq/apps/hqwebapp/utils.py, line 2 - F401 'uuid' imported but unused

corehq/apps/domain/static/domain/js/encoder.js
```diff
@@ -0,0 +1,42 @@
...
+            password_field.val(password_encoder.encode(password_field.val()));
+          };
+      }
+      var reset_password_fields = $("#id_old_password, #id_new_password1, #id_new_password2");
+      if(reset_password_fields.length) {
+          $(reset_password_fields[0]).parents("form")[0].onsubmit = function() {
+              for(var i=0; i<reset_password_fields.length; i++) {
+                  password_field = $(reset_password_fields[i]);
+                  password_encoder = new HexParsr();
+                  password_field.val(password_encoder.encode(password_field.val()));
```
millerdev: if the submit fails due to connectivity issue or whatever, will a subsequent submission cause this encoded value to be encoded again? since it would double-encode the old password too, I'm pretty sure any subsequent submission would fail, but it doesn't seem like a good user experience.

corehq/apps/domain/templates/login_and_password/login.html
```diff
@@ -1,10 +1,19 @@
...
 
 {% block page_content %}
     {% block login-content %}
     {% include "login_and_password/partials/login_full.html" %}
     {% endblock %}
+
+    {% if implement_password_hashing %}
+        {% compress js %}
+            {{ block.super }}
+            <script src="{% static 'domain/js/encoder.js' %}"></script>
```
millerdev: re: https://github.com/dimagi/commcare-hq/pull/15490#discussion_r107991151

I'm not sure if this is correct yet. Looks like it's in the `page_content` block. I think it should be in the `js` block.

mkangia: Moved.

dimagimon:
There are 19 errors:

* corehq/apps/domain/static/domain/js/encoder.js, line 7 - 'b64EncodeUnicode' is not defined. (no-undef)
* corehq/apps/domain/static/domain/js/encoder.js, line 8 - Expected indentation of 12 space characters but found 10. (indent)
* corehq/apps/domain/static/domain/js/encoder.js, line 9 - Expected indentation of 14 space characters but found 16. (indent)
* corehq/apps/domain/static/domain/js/encoder.js, line 11 - Missing semicolon. (semi)
* corehq/apps/domain/static/domain/js/encoder.js, line 15 - Missing semicolon. (semi)
* corehq/apps/domain/static/domain/js/encoder.js, line 18 - 'b64EncodeUnicode' is not defined. (no-undef)
* corehq/apps/domain/static/domain/js/encoder.js, line 20 - Missing semicolon. (semi)
* corehq/apps/domain/static/domain/js/encoder.js, line 24 - Expected indentation of 8 space characters but found 6. (indent)
* corehq/apps/domain/static/domain/js/encoder.js, line 25 - Expected indentation of 8 space characters but found 6. (indent)
* corehq/apps/domain/static/domain/js/encoder.js, line 27 - 'password_encoder' is not defined. (no-undef)
Expected indentation of 14 space characters but found 12. (indent)
* corehq/apps/domain/static/domain/js/encoder.js, line 28 - Expected indentation of 14 space characters but found 12. (indent)
'password_encoder' is not defined. (no-undef)
* corehq/apps/domain/static/domain/js/encoder.js, line 31 - Expected indentation of 8 space characters but found 6. (indent)
* corehq/apps/domain/static/domain/js/encoder.js, line 32 - Expected indentation of 8 space characters but found 6. (indent)
* corehq/apps/domain/static/domain/js/encoder.js, line 36 - 'password_encoder' is not defined. (no-undef)
* corehq/apps/domain/static/domain/js/encoder.js, line 37 - 'password_encoder' is not defined. (no-undef)
* corehq/apps/domain/static/domain/js/encoder.js, line 39 - Missing semicolon. (semi)
* corehq/apps/domain/auth.py, line 10 - F401 'settings' imported but unused
* corehq/apps/hqwebapp/forms.py, line 11 - F401 'settings' imported but unused
* corehq/apps/hqwebapp/utils.py, line 2 - F401 'uuid' imported but unused

corehq/apps/domain/static/domain/js/encoder.js
mkangia: oh nice! didn't think of that.
I think one way would be to keep track of if the password has been encoded and expire that in case the value is changed.
```
var password_field = $("#id_auth-password");
      var password_encoded = false
      if(password_field.length) {
          password_field.on("change", function(){
            password_encoded = false
          })
          password_field.parents("form")[0].onsubmit = function() {
            if(!password_encoded) {
                password_encoder = new HexParsr();
                password_field.val(password_encoder.encode(password_field.val()));
                password_encoded = true
            }
          };
      }
```
Not sure how often this would come in handy though.

corehq/apps/domain/static/domain/js/encoder.js
orangejenny: hqModules is already included because it's part of `style/base.html`. But yeah, wrapping it in a function is fine. It'll likely be turned into a module eventually, but it doesn't have to happen now.

orangejenny:
JavaScript updates look good, thanks for responding.

corehq/apps/domain/static/domain/js/encoder.js
```diff
@@ -0,0 +1,42 @@
+(function() {
+    function HexParsr() {
+        // private property
+        var _paddingLeft = "sha256$" + Math.random().toString(36).slice(2, 8);
+        var _paddingRight = Math.random().toString(36).slice(2, 8) + "=";
```
millerdev: It is possible that these `Math.random().toString(36)` calls will return a string that is shorter than 8 characters:
```
-> (0.0).toString(36)
<- "0"
```

millerdev (commented):
@mkangia are you planning to clean up the lint issues in this PR?

Also, it would be nice to have some tests that verify that values produced by the JavaScript can be decoded on the Python side. I'm thinking you'd use the JS to generate some encoded passwords (you'd do this once before writing the python test), and then write some python tests that recover the password from those previously encoded values and verify that it gets the correct result. Bonus points if you can figure out a way to run the JS encoder and python decoder in the same test (not sure if we have the test infrastructure to do that right now).

dimagimon:
There are 20 errors:

* corehq/apps/domain/static/domain/js/encoder.js, line 5 - Expected '===' and instead saw '=='. (eqeqeq)
* corehq/apps/domain/static/domain/js/encoder.js, line 12 - 'b64EncodeUnicode' is not defined. (no-undef)
* corehq/apps/domain/static/domain/js/encoder.js, line 13 - Expected indentation of 12 space characters but found 10. (indent)
* corehq/apps/domain/static/domain/js/encoder.js, line 14 - Expected indentation of 14 space characters but found 16. (indent)
* corehq/apps/domain/static/domain/js/encoder.js, line 16 - Missing semicolon. (semi)
* corehq/apps/domain/static/domain/js/encoder.js, line 20 - Missing semicolon. (semi)
* corehq/apps/domain/static/domain/js/encoder.js, line 23 - 'b64EncodeUnicode' is not defined. (no-undef)
* corehq/apps/domain/static/domain/js/encoder.js, line 25 - Missing semicolon. (semi)
* corehq/apps/domain/static/domain/js/encoder.js, line 29 - Expected indentation of 8 space characters but found 6. (indent)
* corehq/apps/domain/static/domain/js/encoder.js, line 30 - Expected indentation of 8 space characters but found 6. (indent)
* corehq/apps/domain/static/domain/js/encoder.js, line 32 - 'password_encoder' is not defined. (no-undef)
Expected indentation of 14 space characters but found 12. (indent)
* corehq/apps/domain/static/domain/js/encoder.js, line 33 - Expected indentation of 14 space characters but found 12. (indent)
'password_encoder' is not defined. (no-undef)
* corehq/apps/domain/static/domain/js/encoder.js, line 36 - Expected indentation of 8 space characters but found 6. (indent)
* corehq/apps/domain/static/domain/js/encoder.js, line 37 - Expected indentation of 8 space characters but found 6. (indent)
* corehq/apps/domain/static/domain/js/encoder.js, line 41 - 'password_encoder' is not defined. (no-undef)
* corehq/apps/domain/static/domain/js/encoder.js, line 42 - 'password_encoder' is not defined. (no-undef)
* corehq/apps/domain/static/domain/js/encoder.js, line 44 - Missing semicolon. (semi)
* corehq/apps/domain/auth.py, line 10 - F401 'settings' imported but unused
* corehq/apps/hqwebapp/forms.py, line 11 - F401 'settings' imported but unused
* corehq/apps/hqwebapp/utils.py, line 2 - F401 'uuid' imported but unused

corehq/apps/domain/static/domain/js/encoder.js
mkangia: Ah! right.. Adding a retry if length not apt.

dimagimon:
There are 5 errors:

* corehq/apps/domain/static/domain/js/encoder.js, line 32 - 'password_encoder' is not defined. (no-undef)
* corehq/apps/domain/static/domain/js/encoder.js, line 33 - 'password_encoder' is not defined. (no-undef)
* corehq/apps/domain/static/domain/js/encoder.js, line 41 - 'password_encoder' is not defined. (no-undef)
* corehq/apps/domain/static/domain/js/encoder.js, line 42 - 'password_encoder' is not defined. (no-undef)
* corehq/apps/domain/static/domain/js/encoder.js, line 44 - Missing semicolon. (semi)

millerdev (approved):
looking good

corehq/apps/hqwebapp/tasks.py
```diff
@@ -78,3 +81,14 @@ def mail_admins_async(self, subject, message, fail_silently=False, connection=No
             }
         )
         self.retry(exc=e)
+
+
+@periodic_task(run_every=crontab(hour=0, minute=0), queue=getattr(settings, 'CELERY_PERIODIC_QUEUE','celery'))
```
dimagimon: E231 missing whitespace after ','

mkangia:
@millerdev the code changes are now updated to include the changes we discussed with Clayton for replay attack for web login only, mainly [here](https://github.com/dimagi/commcare-hq/pull/15490/files#diff-261364967e94356f22e293ec0909b1bdR143)

> Also, it would be nice to have some tests that verify that values produced by the JavaScript can be decoded on the Python side. I'm thinking you'd use the JS to generate some encoded passwords (you'd do this once before writing the python test), and then write some python tests that recover the password from those previously encoded values and verify that it gets the correct result. Bonus points if you can figure out a way to run the JS encoder and python decoder in the same test (not sure if we have the test infrastructure to do that right now).

I gave this a shot but could not achieve any JS integration to generate passwords. So just used the password hashes that were generated from JS/mobile and used for testing in [tests](https://github.com/dimagi/commcare-hq/pull/15490/files#diff-1396f790c705a17024a86fa4ee9dad9c). Considering that JS would generate the same passwords considering JS is not changed, using pre-built hashes should suffice.

corehq/apps/hqwebapp/tests/test_utils.py
```diff
@@ -0,0 +1,50 @@
...
+            ).exists())
+
+    def test_replay_attack(self):
+        password_hash = "sha256$1e2d5bc2hhMjU2JDFlMmQ1Yk1USXpORFUyZjc5MTI3PQ==f79127="
+        username = "james@007.com"
+        HashedPasswordLoginAttempt.objects.create(
+            username=username,
+            password_hash=password_hash
+        )
+        self.assertEqual(decode_password(password_hash, username), '')
```
millerdev: this test should use `decode_password(password_hash, username)` to create the initial `HashedPasswordLoginAttempt` object, and then assert the return value is `''` on a second call. That should show that `memoized` blocks the replay protection.

corehq/apps/hqwebapp/utils.py
```diff
@@ -81,3 +86,70 @@ def aliased_language_name(lang_code):
...
+        # return the password received AS-IS
+        return password
+
+
+# Memoized for multiple decode attempts in the same request:
+# 1. an attempt to decode a password should be done just once in a request for the login attempt
+# check to work correctly.
+# 2. there should be no need to decode a password multiple times in
+# the same request.
+@memoized
```
mkangia: I am not sure if using `memorised` here never expires. From testing i thought it was just over the request and starts afresh for next login web request. I am going to check this further.

corehq/apps/hqwebapp/utils.py
```diff
@@ -81,3 +86,70 @@ def aliased_language_name(lang_code):
...
+            if HashedPasswordLoginAttempt.objects.filter(
+                username=username,
+                password_hash=password,
+                used_at__gte=(datetime.today() - timedelta(HASHED_PASSWORD_EXPIRY))
+            ).exists():
+                return ''
+            else:
+                HashedPasswordLoginAttempt.objects.create(
+                    username=username,
+                    password_hash=password
```
mkangia: Yikes! That thought was in my mind when i was doing it, not sure why i didn't connect it to what i was implementing(may be coz it was encoded with some algorithm). Thanks for the correction. Bad miss!

corehq/apps/hqwebapp/tests/test_utils.py
mkangia: In development environment it actually does not block the replay protection. There is something else happening here thats making it work as expected. Need to check.

corehq/apps/hqwebapp/utils.py
mkangia: I think this seemed to work well on staging coz of multiple web workers since the request for simultaneous login using the same hash went to different instances and it worked in case of local machine due to reload/restart of server. Checked it again on local and it fails like you said. This still needs some caching in place or some other tweak.

corehq/apps/hqwebapp/tests/test_utils.py
mkangia: That is correct @millerdev. It was failing. I misunderstood that to do something with memoization within a single test instance.

corehq/apps/domain/forms.py
```diff
@@ -1331,7 +1331,22 @@ class HQSetPasswordForm(SetPasswordForm):
                                     """))
 
     def clean_new_password1(self):
-        return clean_password(self.cleaned_data.get('new_password1'))
+        from corehq.apps.hqwebapp.utils import decode_password
+        password1 = decode_password(self.cleaned_data.get('new_password1'))
+        return clean_password(password1)
+
+    def clean_new_password2(self):
```
millerdev: Looks like this function is duplicated in about 3 places. Would it makes sense to refactor to remove the duplication?

corehq/apps/hqwebapp/utils.py
```diff
@@ -81,3 +86,91 @@ def aliased_language_name(lang_code):
...
+def verify_password(password, password_salt):
+    hasher = get_password_hasher()
+    return hasher.verify(password, password_salt)
+
+
+# quickcache for multiple decode attempts in the same request:
+# 1. an attempt to decode a password should be done just once in a request for the login attempt
+# check to work correctly.
+# 2. there should be no need to decode a password multiple times in the same request.
+@quickcache(['password_hash'], timeout=0)
```
dannyroberts: I saw the offlined conversation in #web-team

I agree that `quickcache` isn't a good way to make sure something is run only once per request. You could do that by setting a thread-local (which is basically a global variable that is actually local to the request).

```
# in the file's global namespace
_thread_local = threading.local()

# in the function
    if getattr(_thread_local, 'decode_password_already_run', False):
        raise YourException("your message about why this should never happen")
    else:
        _thread_local.decode_password_already_run = True
```

corehq/apps/hqwebapp/utils.py
```diff
@@ -81,3 +86,91 @@ def aliased_language_name(lang_code):
...
+            # completely
+            return ''
+    else:
+        # return the password received AS-IS
+        return password
+
+
+@memoized
+def get_password_hasher():
+    return get_hasher()
```
mkangia: Nice! Will do.

corehq/apps/hqwebapp/utils.py
millerdev: does django create a new thread per request? or does it use a thread pool where a single thread is re-used for many requests? Using a thread-local seems like a bit of a hack because it couples this logic with an implementation detail of the server architecture, which seems fragile.

corehq/apps/hqwebapp/utils.py
```diff
@@ -81,3 +86,91 @@ def aliased_language_name(lang_code):
...
+
+
+def hash_password(password):
+    hasher = get_password_hasher()
+    return hasher.encode(password, hasher.salt())
+
+
+def verify_password(password, password_salt):
+    hasher = get_password_hasher()
+    return hasher.verify(password, password_salt)
```
mkangia: I don't think its stateful. `verify` just compares a password i.e plain text and password salt generated by the hashers algorithm like SHA1 to verify that the password salt was generated for the same password.

corehq/apps/hqwebapp/utils.py
dannyroberts: It's a bit of a hack, but it's been a pretty stable hack for years. We use it in `get_request()` and it's worked very well. (Django makes each request a "thread", but our threads have been monkey patched to be greenlets/gevent threads by `gevent.monkey_patch`)

Agree though that adding the hack here is really not ideal.

mkangia: Thank You @millerdev @dannyroberts 

```
You'll need to do it somewhere where you have access to the request object itself or cache the result on an object with the same lifecycle as the request (for example, the view).
```
i don't think it would be possible to track down all views calling that function. It might also be within django at places.

I looked at the documentation for `quickcache` and seemed like it was the right way. Added that it should vary on password_hash argument passed and by default it expires in 10 seconds. Also added timeout 0 so that its not saved in any other cache.
```
In addition to caching in the default shared cache,
          quickcache caches in memory for 10 seconds
          (conceptually the length of a single request).
```

corehq/apps/hqwebapp/utils.py
```diff
@@ -81,3 +86,91 @@ def aliased_language_name(lang_code):
...
+# 2. there should be no need to decode a password multiple times in the same request.
+@quickcache(['password_hash'], timeout=0)
+def decode_password(password_hash, username=None):
+    def replay_attack():
+        # Replay attack where the same hash used from previous login attempt
+        login_attempts = HashedPasswordLoginAttempt.objects.filter(
+            username=username,
+        )
+        for login_attempt in login_attempts:
+            if verify_password(password_hash, login_attempt.password_hash):
```
dannyroberts: isn't `verify_password` supposed to take a hashed password and a plaintext password and compare them for match? This looks like it passes in two hashed passwords, so won't it just always return false?

corehq/apps/hqwebapp/utils.py
```diff
@@ -81,3 +86,91 @@ def aliased_language_name(lang_code):
...
+# check to work correctly.
+# 2. there should be no need to decode a password multiple times in the same request.
+@quickcache(['password_hash'], timeout=0)
+def decode_password(password_hash, username=None):
+    def replay_attack():
+        # Replay attack where the same hash used from previous login attempt
+        login_attempts = HashedPasswordLoginAttempt.objects.filter(
+            username=username,
+        )
+        for login_attempt in login_attempts:
```
dannyroberts: Might there be more than a couple login attempts over time? At some point this becomes an infeasible way to check

mkangia: Have added a task [here](https://github.com/dimagi/commcare-hq/pull/15490/files#diff-f09b26756bbf9bb3c3289709ae0ec5bbR87) that would remove any attempts older than a set time period(30 days for now).

corehq/apps/hqwebapp/utils.py
mkangia: > isn't verify_password supposed to take a hashed password and a plaintext password and compare them for match?
Thats correct.
We are storing encoded(not salted) passwords in DB which can be decoded for passwords if DB is compromised. The `password_hash` here is the encoded password and `login_attempt.password_hash` is the salt/hashed password.
discussion [here](https://github.com/dimagi/commcare-hq/pull/15490#discussion-diff-112281775R152)

I should improve the naming. Will reconsider the names.

dannyroberts:
Hey @mkangia, I might be the only one who doesn't know here, but can you back up and explain what this PR is doing, or link to a spec, just to provide some context. (Maybe put it in the description so it's prominent for others.) I'm finding myself confused and I think it's because I lack the bigger picture

mkangia:
@dannyroberts most of this is over email but there is a doc for reference we have been using. Updated description. Sorry for not putting that in description earlier.

corehq/apps/domain/static/domain/js/encoder.js
```diff
@@ -0,0 +1,49 @@
...
+                for(var i=0; i<reset_password_fields.length; i++) {
+                    password_field = $(reset_password_fields[i]);
+                    var password_encoder = new HexParsr();
+                    password_field.val(password_encoder.encode(password_field.val()));
+                }
+            };
+        }
+    });
+
+    return HexParsr
```
dimagimon: Missing semicolon. (semi)

corehq/apps/registration/static/registration/js/new_user.ko.js
```diff
@@ -178,10 +178,14 @@ hqDefine('registration/js/new_user.ko.js', function () {
         self.currentStep = ko.observable(0);
 
         var _getDataForSubmission = function () {
+            var password = self.password()
```
dimagimon: Missing semicolon. (semi)

corehq/apps/registration/static/registration/js/new_user.ko.js
```diff
@@ -178,10 +178,14 @@ hqDefine('registration/js/new_user.ko.js', function () {
...
+            if(typeof(hex_parser) != 'undefined') {
```
dimagimon: Expected '!==' and instead saw '!='. (eqeqeq)

corehq/apps/registration/static/registration/js/new_user.ko.js
```diff
@@ -178,10 +178,14 @@ hqDefine('registration/js/new_user.ko.js', function () {
...
+                var password = (new hex_parser()).encode(self.password())
```
dimagimon: 'password' is already defined (no-redeclare)
'hex_parser' is not defined. (no-undef)
Missing semicolon. (semi)

corehq/apps/registration/static/registration/js/new_user.ko.js
```diff
@@ -178,10 +178,14 @@ hqDefine('registration/js/new_user.ko.js', function () {
         self.currentStep = ko.observable(0);
 
         var _getDataForSubmission = function () {
+            var password = self.password();
+            if(typeof(hex_parser) !== 'undefined') {
+                password = (new hex_parser()).encode(self.password());
```
dimagimon: 'hex_parser' is not defined. (no-undef)

mkangia:
@millerdev @dannyroberts Offlined with @snopoke on alternates for `quickcache` and he suggested a way around using `get_request` and maintaining this over the request object.
Its working well. Here is the [commit](https://github.com/dimagi/commcare-hq/pull/15490/commits/c69e47fed4e861239f903f705aaf15d23a870353)

Thank you @snopoke

corehq/apps/domain/forms.py
```diff
@@ -1331,7 +1331,22 @@ class HQSetPasswordForm(SetPasswordForm):
...
+        from corehq.apps.hqwebapp.utils import decode_password
+        password1 = self.cleaned_data.get('new_password1')
+        password2 = decode_password(self.cleaned_data.get('new_password2'))
+        if password1 and password2:
```
mkangia: Currently its breaking in case decode_password returns an empty string. This can only happen if the decoded password is not properly generated. Also replay attack is just for login.
Fixed by checking for empty password and raised ValidationError with `password_empty` message.

corehq/apps/domain/forms.py
```diff
@@ -16,7 +16,7 @@
 from dateutil.relativedelta import relativedelta
 from django import forms
 from django.conf import settings
-from django.contrib.auth import get_user_model
+from django.contrib.auth import get_user_model, password_validation
```
dimagimon: F401 'password_validation' imported but unused

corehq/apps/users/forms.py
```diff
@@ -16,19 +16,20 @@
 from django.template.loader import get_template
 from django.template import Context
 from django_countries.data import COUNTRIES
-
+from django.contrib.auth import password_validation
```
dimagimon: F401 'password_validation' imported but unused

corehq/apps/domain/static/domain/js/encoder.js
```diff
@@ -0,0 +1,53 @@
...
+                return String.fromCharCode('0x' + p1);
+            });
+        };
+
+        this.addPadding = function(secret_password) {
+            return _paddingLeft + secret_password + _paddingRight;
+        };
+
+        this.encode = function(password) {
+            password = password.trim()
```
dimagimon: Missing semicolon. (semi)

corehq/apps/domain/static/domain/js/encoder.js
```diff
@@ -0,0 +1,53 @@
...
+            if(password) {
+                var secret_password = this.addPadding(window.btoa(b64EncodeUnicode(password)));
+                return this.addPadding(window.btoa(secret_password));
+            }
+            return password
```
dimagimon: Missing semicolon. (semi)

corehq/apps/hqwebapp/password_validation.py
```diff
@@ -0,0 +1,34 @@
...
+
+    # hook to password_changed called on all validators when password is changed for a Django User
+    def password_changed(self, password, user):
+        UsedPasswords.objects.create(
+            user=user,
+            password_hash=hash_password(password)
+        )
+
+    def get_help_text(self):
+        return _("Your password can't be same as last {restricted} passwords.").format(
```
mkangia: Wont it be nice to be explicit about the number of passwords?

corehq/apps/hqwebapp/migrations/0002_auto_20170416_2117.py
```diff
@@ -0,0 +1,24 @@
...
+        ('hqwebapp', '0001_initial'),
+    ]
+
+    operations = [
+        migrations.CreateModel(
+            name='HashedPasswordLoginAttempt',
+            fields=[
+                ('id', models.AutoField(auto_created=True, primary_key=True, serialize=False, verbose_name='ID')),
+                ('username', models.CharField(db_index=True, max_length=255)),
+                ('password_hash', models.CharField(db_index=True, max_length=255)),
```
mkangia: This model would be removed once this gets moved to redis

corehq/apps/hqwebapp/migrations/0005_auto_20170426_2139.py
```diff
@@ -0,0 +1,20 @@
...
+        ('hqwebapp', '0004_auto_20170423_1750'),
+    ]
+
+    operations = [
+        migrations.RenameField(
+            model_name='usedpasswords',
+            old_name='password',
+            new_name='password_hash',
+        ),
+    ]
```
mkangia: Would it be safe now to merge them since the 2 models are being now removed? Those tables are already present at least on staging.

corehq/apps/hqwebapp/utils.py
```diff
@@ -81,3 +87,101 @@ def aliased_language_name(lang_code):
             if code == lang_code:
                 return name
         raise KeyError('Unknown language code %s' % lang_code)
+
+
+def extract_password(password):
+    # Passwords set with expected salts length and padding would respect this regex
+    reg_exp = r"^sha256\$([a-z|0-9|A-Z]{6})(.*)([a-z|0-9|A-Z]{6})=$"
```
mkangia: thanks for the correction with the `|` character.

Replacing `(.*)` with `([a-zA-Z0-9=]*)`.  `=` comes in coz the second salt used always has `=` in it at the end. Was there a specific reason for adding `+\`?

corehq/apps/domain/forms.py
```diff
@@ -1322,17 +1323,14 @@ def clean_email(self):
             return self.cleaned_data['email']
 
 
-class HQSetPasswordForm(SetPasswordForm):
+class HQSetPasswordForm(EncodedPasswordChangeForm, SetPasswordForm):
```
snopoke: instead of changing this class would it be possible to create a new class and then only use the new class if enabled?
```
class EncodedPasswordComplianceForm(EncodedPasswordChangeFormMixin, HQSetPasswordForm):
    pass

form = EncodedPasswordComplianceForm if settings.ENABLE_PASSWORD_HASHING else HQSetPasswordForm
```

corehq/apps/hqwebapp/migrations/0002_auto_20170416_2117.py
snopoke: I think just remove the migrations - we can clean up staging by manually dropping the tables.

custom/nic_compliance/utils.py
```diff
@@ -0,0 +1,117 @@
...
+        # Replay attack where the same hash used from previous login attempt
+        login_attempts = get_login_attempts(username)
+        for login_attempt in login_attempts:
+            if verify_password(password_hash, login_attempt):
+                return True
+
+    def record_login_attempt():
+        client = get_redis_client()
+        login_attempts = client.get(login_attempts_redis_key_for_user(username), [])
+        key_name = login_attempts_redis_key_for_user(username)
```
snopoke: could move this up one line and reuse it rather than calling `login_attempts_redis_key_for_user` twice

corehq/apps/users/forms.py
```diff
@@ -413,9 +414,14 @@ def __init__(self, project, user_id, **kwargs):
         )
 
     def clean_new_password1(self):
```
snopoke: This seems to get overridden by `EncodedPasswordChangeForm`

custom/nic_compliance/forms.py
```diff
@@ -0,0 +1,37 @@
...
+    def clean_new_password1(self):
+        from corehq.apps.domain.forms import clean_password
+        from corehq.apps.hqwebapp.utils import decode_password
+        new_password = decode_password(self.cleaned_data.get('new_password1'))
+        if new_password == '':
+            raise ValidationError(
+                _("Password cannot be empty"), code='new_password1_empty',
+            )
+
+        return clean_password(new_password)
```
snopoke: I think this should be under an if statement: ` if self.project.strong_mobile_passwords:`

snopoke (commented):
Looking good. I think my only question now is the approach that we're taking with the django forms and whether or not we should be changing the current forms or overriding them.

corehq/apps/users/forms.py
```diff
@@ -23,12 +23,13 @@
 from corehq.apps.domain.models import Domain
 from corehq.apps.locations.models import SQLLocation
 from corehq.apps.locations.permissions import user_can_access_location_id
+from corehq.apps.settings.forms import EncodedPasswordChangeForm
```
dannyroberts: I think you want to change this to
```python
from custom.nic_compliance.forms import EncodedPasswordChangeForm
```
now, even though this still technically works (since `corehq.apps.settings.forms` also imports this). I'll also comment on the other places that I'd make the same change.

settings.py
```diff
@@ -892,6 +893,8 @@
 SENTRY_QUERY_URL = 'https://sentry.io/{org}/{project}/?query='
 SENTRY_API_KEY = None
 
+ENABLE_PASSWORD_HASHING = False
```
dannyroberts: I know this is annoying and touches a lot of different places, but could we change the name of this setting and related functions to

1. include the phrase `NIC_COMPLIANCE` so it's clear that this is tied to an NIC compliance feature and
2. use a word like 'obfuscation' instead of 'hashing'?

Unless calling it 'hashing' is itself a requirement for NIC compliance, I'd prefer to call it obfuscating because hashing is the concept of taking a very large/infinite input space (such as "all strings") and mapping it to very (comparatively) small output space (such as "all 128-bit integers"); moreover colloquially it implies the properties we associate with a "good" hashing function (such as consistent distribution), and in the context of passwords specifically, with those of a cryptographic hashing function (such as bit scrambling and irreversibility). Since the obfuscation scheme we're using for NIC compliance doesn't have any of those properties, I think simply calling it "obfuscation" would lead to less confusion/surprise.

dannyroberts (commented):
Looks to me like moving most of the code to `custom/nic_compliance` was a really big improvement. A few comments below.

corehq/apps/settings/forms.py
```diff
@@ -54,8 +56,19 @@ def __init__(self, user, *args, **kwargs):
...
+        old_password = decode_password(self.cleaned_data['old_password'])
+        if old_password == '':
+            raise ValidationError(
+                _("Password cannot be empty"), code='password_empty',
+            )
+        if not self.user.check_password(old_password):
+            raise forms.ValidationError(
+                self.error_messages['password_incorrect'],
+                code='password_incorrect',
+            )
```
millerdev: It's surprising to me that we need to raise these two validation errors in this PR since they should be handled at other places in the code.

Could it be changed to something like (a little tricky, but try/finally is necessary if `clean_old_password` is called multiple times per request):
```py
def clean_old_password(self):
    encoded_pass = self.cleaned_data['old_password']
    cleaned_pass = decode_password(encoded_pass)
    self.cleaned_data['old_password'] = cleaned_pass
    try:
        return super(HQPasswordChangeForm, self).clean_old_password()
    finally:
        self.cleaned_data['old_password'] = encoded_pass
```

If only called once per request:
```py
def clean_old_password(self):
    self.cleaned_data['old_password'] = decode_password(
        self.cleaned_data['old_password'])
    return super(HQPasswordChangeForm, self).clean_old_password()
```

corehq/apps/hqwebapp/password_validation.py
```diff
@@ -0,0 +1,34 @@
...
+
+from corehq.apps.hqwebapp.const import RESTRICT_USED_PASSWORDS_NUM
+from corehq.apps.hqwebapp.models import UsedPasswords
+from corehq.apps.hqwebapp.utils import verify_password, hash_password
+
+
+class UsedPasswordValidator(object):
+    def validate(self, password, user):
+        used_passwords = UsedPasswords.objects.filter(
+            user=user,
```
snopoke: The `UserPasswords` model has been removed and we're now storing the data in redis with a timeout.

millerdev: oh, ignore this comment. I started to do a review a while ago and added this comment, but then never submitted it. I forgot to delete it when I did another review yesterday.

mkangia:
Thank you for the master merge/update @dannyroberts

dannyroberts:
👍 I actually did it inside GitHub's new Resolve Conflict tool. It's super cool, hadn't used it before!

mkangia:
@dannyroberts pretty cool.. also clearly highlights the conflict in settings file. 👍

custom/nic_compliance/forms.py
mkangia: this builds up on the implementation for `clean_new_password1` from `HQSetPasswordForm` and  `HQPasswordChangeForm` which does not use that if block. Only `SetUserPasswordForm` had a different implementation so in that class this method is overridden.

corehq/apps/users/forms.py
mkangia: Ya. Thats intentional. 
[details](https://github.com/dimagi/commcare-hq/pull/15490/files/61e87fa26387a986e1fbcc11f2b164bb0662db1f#diff-68f388f9cbc609677f66c12e41cf02b8R17)

corehq/apps/domain/forms.py
```diff
@@ -71,6 +71,7 @@
     AREA_CHOICES, SUB_AREA_CHOICES, BUSINESS_UNITS, TransferDomainRequest)
 from corehq.apps.hqwebapp.tasks import send_mail_async, send_html_email_async
 from corehq.apps.reminders.models import CaseReminderHandler
+from corehq.apps.settings.forms import EncodedPasswordChangeForm
```
mkangia: wow!!! That should not work...but it is.. this is so confusing..

corehq/apps/domain/forms.py
mkangia: Would need do that for various forms then.. one form would not work for all different forms.
Making this as a `mixin` would make it clear?

corehq/apps/settings/forms.py
mkangia: > It's surprising to me that we need to raise these two validation errors in this PR since they should be handled at other places in the code.

`if not self.user.check_password(old_password)` is to check if user entered their current password correctly. This is a password reset form. This would not be used anywhere else.

I think we can effectively just keep the `check_password`. In case of an empty password which would happen only in case someone messes up the hashing and then would get the message "Incorrect password" instead of "empty password"

corehq/apps/users/forms.py
mkangia: Thanks for flagging this @dannyroberts.

settings.py
mkangia: That makes sense @dannyroberts. 

> Unless calling it 'hashing' is itself a requirement for NIC compliance
:1st_place_medal:

mkangia:
@dannyroberts @snopoke @millerdev 
Thank you for your review for this PR. Has been quite a journey.

I have tried to cover all review points and replied to the ones where needed. Apologies if i missed any.

Since testing this is quite a battle and time consuming, i would test it once i get approval on changes.
Request you for one final pass.

corehq/apps/settings/forms.py
snopoke: I think @millerdev point is that the super method already does the password validation so it shouldn't be necessary to copy that in there. Rather you could just call the super method as he's suggested in his code examples.

corehq/apps/users/forms.py
snopoke: Perhaps you misunderstand, if this is getting overridden then it won't get called at all unless the function in `EncodedPasswordChangeForm` calls it explicitly with `super`.

As it stands right now this function will never get called.

custom/nic_compliance/forms.py
snopoke: See comment on `SetUserPasswordForm.clean_new_password1`. I don't think this is working as you describe.

custom/nic_compliance/utils.py
```diff
@@ -0,0 +1,119 @@
...
+        request = get_request()
+        if request:
+            # 1. an attempt to decode a password should be done just once in a request for the login attempt
+            # check to work correctly and not consider it a replay attack in case of multiple calls
+            # 2. also there should be no need to decode a password multiple times in the same request.
+            if not hasattr(request, 'decoded_password'):
+                request.decoded_password = {}
+
+            # return decoded password set on request object for the obfuscated_password
+            if request.decoded_password.get(obfuscated_password):
```
snopoke: Since `_decode_password()` can return an empty string this should be `if obfuscated_password in request.decoded_password`

settings.py
```diff
@@ -897,6 +898,8 @@
 SENTRY_QUERY_URL = 'https://sentry.io/{org}/{project}/?query='
 SENTRY_API_KEY = None
 
+OBFUSCATE_PASSWORD_FOR_NIC_COMPLIANCE = False
+RESTRICT_USED_PASSWORDS_FOR_NIC_COMPLIANCE = False
```
snopoke: Any reason to have these separate? I don't see that we'd ever enable one and not the other.

mkangia: Just that both of these are for different functionalities. Also helped in testing one thing at a time.

custom/nic_compliance/utils.py
mkangia: Thanks for catching this.

corehq/apps/users/forms.py
snopoke: Never mind.

snopoke (approved):

snopoke:
Didn't mean to approve - still think that some changes from other comments are necessary.

corehq/apps/settings/forms.py
millerdev: Yes, that's what I meant. Thanks for clarifying @snopoke

mkangia: Thank you @snopoke @millerdev 
Super method does check on raw password. This overrides to use the decoded password.
I don't think calling super can be done here.

snopoke: Either of the code snippets that @millerdev pastd should work since they put the decoded password into `self.cleaned_data` which is where the super method gets the password from.

mkangia: Cool. Sorry for the back and forth on this simple thing.

Will update to keep just the blank password check here to retain better message to user and call the method using super for password check.

custom/nic_compliance/utils.py
```diff
@@ -0,0 +1,119 @@
...
+
+from corehq.util.view_utils import get_request
+from custom.nic_compliance.const import EXPIRE_LOGIN_ATTEMPTS_IN, REDIS_LOGIN_ATTEMPTS_LIST_PREFIX
+
+PASSWORD_HASHER = get_hasher()
+
+
+def extract_password(obfuscated_password):
+    # Passwords set with expected padding length and format would respect this regex
+    reg_exp = r"^sha256\$([a-z0-9A-Z]{6})([a-zA-Z0-9=]*)([a-z0-9A-Z]{6})=$"
```
millerdev: Should move this to a module-level constant and wrap it in `re.compile(...)`. Then below, call `reg_exp.match(whatever)` instead of `re.match(reg_exp, whatever)`.

custom/nic_compliance/utils.py
```diff
@@ -0,0 +1,119 @@
...
+    else:
+        # return the password received AS-IS
+        return obfuscated_password
+
+
+def hash_password(password):
+    return PASSWORD_HASHER.encode(password, PASSWORD_HASHER.salt())
+
+
+def verify_password(password, password_salt):
```
millerdev: Nitpick: the second argument is not a salt. Consider renaming to `encoded` as in https://github.com/django/django/blob/1.11/django/contrib/auth/hashers.py#L200

millerdev (commented):
Looks like there are a few comments from @snopoke that have not been addressed yet.

corehq/apps/settings/forms.py
```diff
@@ -54,8 +56,15 @@ def __init__(self, user, *args, **kwargs):
...
 
-    def clean_new_password1(self):
-        return clean_password(self.cleaned_data.get('new_password1'))
+    def clean_old_password(self):
+        from corehq.apps.hqwebapp.utils import decode_password
+        old_password = decode_password(self.cleaned_data['old_password'])
+        if old_password == '':
+            raise ValidationError(
+                _("Password cannot be empty"), code='password_empty',
+            )
```
mkangia: Yes that is correct.

This check would only run in case the encoding is incorrect/messed up with, since one can not enter blank password because the front end does not allow. Hence, the user would get a message password empty and not password incorrect which would be a better message for the user that we are getting empty password and not incorrect password.

custom/nic_compliance/forms.py
```diff
@@ -0,0 +1,39 @@
+from django import forms
+from django.utils.translation import ugettext as _
+from django.core.exceptions import ValidationError
+from django.contrib.auth import password_validation
+
+
+class EncodedPasswordChangeForm(object):
```
mkangia: The name change sounds good.

About the doc string, this can be used by any forms using password fields, for ex: this mixin is currently being used by classes inheriting from `PasswordChangeForm` as well.
You saying this specifically for `clean_new_password1` method?

custom/nic_compliance/static/nic_compliance/js/encoder.js
```diff
@@ -0,0 +1,53 @@
...
+                return String.fromCharCode('0x' + p1);
+            });
+        };
+
+        this.addPadding = function(secret_password) {
+            return _paddingLeft + secret_password + _paddingRight;
+        };
+
+        this.encode = function(password) {
+            password = password.trim();
```
mkangia: > Why are we trimming the password?

Since we are obfuscating the password now, even a blank string gets encoded to a not-empty string and submits to server whereas otherwise if obfuscating is not enabled the browser does not allow submit. So just added the trim.

I tried submitting with a password and then the same password along with an empty space in beginning, where both got successfully submitted. So i thought this was safe here.
If it can be dangerous, i can remove it. It would just allow someone to enter blank passwords in case obfuscation is enabled but would eventually get a message that password can not be empty from server instead of proactively from browser.

custom/nic_compliance/tests/test_password_validation.py
```diff
@@ -0,0 +1,56 @@
...
+from corehq.apps.domain.models import Domain
+from corehq.apps.users.dbaccessors.all_commcare_users import delete_all_users
+from corehq.apps.users.models import WebUser
+from custom.nic_compliance.password_validation import UsedPasswordValidator
+from custom.nic_compliance.const import RESTRICT_USED_PASSWORDS_NUM
+
+
+class TestUsedPasswordsRestriction(TestCase):
+    def setUp(self):
+        delete_all_users()
```
mkangia: Will remove it.

custom/nic_compliance/utils.py
mkangia: Nice one! thank you.

custom/nic_compliance/forms.py
```diff
@@ -0,0 +1,39 @@
...
+
+    def clean_new_password2(self):
+        from corehq.apps.hqwebapp.utils import decode_password
+        password2 = decode_password(self.cleaned_data.get('new_password2'))
+        if password2 == '':
+            raise ValidationError(
+                _("Password cannot be empty"), code='new_password2_empty',
+            )
+
+        password1 = self.cleaned_data.get('new_password1')
```
mkangia: Seems like `clean_new_password1` would be called before `clean_new_password2`. 
```
    def _clean_fields(self):
        for name, field in self.fields.items():
```
This is the snippet from django forms and that fields is an ordered list and when adding the fields "all fields are kept in the order defined in the class" since we are not specifying any order

More details [here](https://github.com/django/django/blob/1.10/django/forms/forms.py#L93)

corehq/apps/settings/forms.py
millerdev: > This check would only be true in case the encoding is incorrect/messed up

Can a normal (not trying to do anything sneaky) user trigger this condition? I think we should not be adding code to give nice/better messages to people who are messing with the obfuscation.

custom/nic_compliance/forms.py
millerdev: `PasswordChangeForm` is a subclass of `SetPasswordForm`

custom/nic_compliance/forms.py
millerdev: `self.fields` ordering appears to depend on `creation_counter` and the MRO, the fields are not sorted alphabetically. In other words it depends when and where the methods are defined in the source code, which is fragile (could be changed with seemingly benign refactoring) and not something I think we should depend on here.

custom/nic_compliance/static/nic_compliance/js/encoder.js
millerdev: It doesn't seem like a good practice to trim characters off the password entered by the user. Making the password shorter makes it weaker and easier to guess. I still think we should not do this.

corehq/apps/settings/forms.py
mkangia: Sounds right. Removing it.

custom/nic_compliance/forms.py
mkangia: Cool.

custom/nic_compliance/forms.py
mkangia: Offlined with Simon. This is fine.

mkangia:
@millerdev 
i have made all the changes.. Also tested everything
realised i had not added this for new mobile worker form
added that [now](https://github.com/dimagi/commcare-hq/pull/15490/commits/8b25a7672526ec2e16e0793e6371dba660aa28ec)
Since the form class did not inherit from `SetPasswordForm` i could not use the mixin.. so just had to add `decode_password` within the `clean_password` method itself

Thank You for the patience and review with this PR

millerdev (approved):
I don't see any other blockers. Good to go if @snopoke approves.

snopoke (approved):
Code wise I think this is GTG. @mkangia relying on you to make the call from a QA point of view.

mkangia:
Thank you @millerdev @snopoke .

Yup! i have tested this. Will do just one last round of sanity check and merge this.
Well let this stay deployed for next two days without enabling obfuscation. Will make ansible changes tomorrow.

PR #15491, opened 2017-03-23
Notify an error on long running rule tasks

gcapalbo:
@proteusvacuum

### DIFF
--- a/corehq/apps/data_interfaces/tasks.py
+++ b/corehq/apps/data_interfaces/tasks.py
@@ -17,10 +17,12 @@ from .interfaces import FormManagementMode, BulkFormManagementInterface
 from .dispatcher import EditDataInterfaceDispatcher
 from corehq.util.log import send_HTML_email
 from dimagi.utils.couch import CriticalSection
+from dimagi.utils.logging import notify_error
 
 
 logger = get_task_logger('data_interfaces')
 ONE_HOUR = 60 * 60
+ONE_DAY_IN_SECONDS = 24 * 60 * 60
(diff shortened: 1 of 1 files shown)


corehq/apps/data_interfaces/tasks.py
```diff
@@ -89,6 +92,13 @@ def run_case_update_rules_for_domain(domain, now=None):
 
         for case_ids in case_id_chunks:
             for case in CaseAccessors(domain).iter_cases(case_ids):
+                time_elapsed = datetime.utcnow() - start_run
+                if time_elapsed.seconds > ONE_DAY_IN_SECONDS:
+                    notify_error(
+                        "Halting rule run for domain %s as it's been running for more than a day." % domain
```
proteusvacuum: would it be useful to know which rule is being halted so that it can be looked into further? 
Also, won't this rule just get re-run next time the task gets triggered? Or is that the intended behaviour?

gcapalbo: yeah so these rules get run every day and while i don't anticipate this scenario to happen often, if it ever does happen it means there's a huge backlog of cases to be closed. so what will happen is it will halt today and then start again tomorrow. knowing the domain should be enough to look into this and see if something is actually wrong or if it's just a large backlog that will be processed eventually.

proteusvacuum: makes sense. Thanks for the explanation.

proteusvacuum (approved):

PR #15493, opened 2017-03-23
Add publish timestamp to ChangeMeta

benrudolph:
@snopoke @emord simon and i discussed this briefly, and after reading that article it could be useful for detecting if a specific change is behind. also just useful for debugging

cc: @millerdev

### DIFF
--- a/corehq/apps/change_feed/tests/test_pillow.py
+++ b/corehq/apps/change_feed/tests/test_pillow.py
@@ -1,4 +1,5 @@
 # coding=utf-8
+from datetime import datetime
 from mock import patch
 from django.conf import settings
 from django.test import SimpleTestCase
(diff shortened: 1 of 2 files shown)


emord (approved):
Is there a way to access these changes without changing the pillows to support it? seems like we'll want some datadog/logging here

benrudolph:
we can view the change information on a pillow error: https://github.com/dimagi/commcare-hq/blob/br/add-publish-timestamp/corehq/ex-submodules/pillow_retry/models.py#L43. but yeah you're right, it doesn't take advantage of it anywhere else for now

benrudolph:
(will think about adding further support in logging/datadog a bit later, think it's fine to start tracking this now)

snopoke:
👍

PR #15503, opened 2017-03-23
force all topics to bytes

emord:
This code is already deployed to NIC (hand edited on pillow and web0/web1)

The kafka library hard errors anytime you give it unicode (even if it's only ascii characters)

https://github.com/dimagi/commcarehq-ansible/blob/cd6dcf4008e93ba37a44477538b65732d1528f6a/ansible/roles/commcarehq/templates/localsettings.py.j2#L164 json.loads always returns the dictionary elements as unicode

@snopoke @sravfeyn

### DIFF
--- a/corehq/apps/userreports/pillow.py
+++ b/corehq/apps/userreports/pillow.py
@@ -249,6 +249,7 @@ class ConfigurableReportKafkaPillow(ConstructedPillow):
 def get_kafka_ucr_pillow(pillow_id='kafka-ucr-main', ucr_division=None,
                          include_ucrs=None, exclude_ucrs=None, topics=None):
     topics = topics or KAFKA_TOPICS
+    topics = [str(t) for t in topics]
     return ConfigurableReportKafkaPillow(
         processor=ConfigurableReportPillowProcessor(
             data_source_provider=DynamicDataSourceProvider(),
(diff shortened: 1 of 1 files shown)


corehq/apps/userreports/pillow.py
```diff
@@ -249,6 +249,7 @@ def rebuild_table(self, sql_adapter):
 def get_kafka_ucr_pillow(pillow_id='kafka-ucr-main', ucr_division=None,
                          include_ucrs=None, exclude_ucrs=None, topics=None):
     topics = topics or KAFKA_TOPICS
+    topics = [str(t) for t in topics]
```
millerdev: Will this work on Python 3? (guessing it does) or should this do `bytes(t)`

emord: I'm not sure. I was going to test it, but that requires trying to spin up an hq management command in py3 which isn't going to be easy. Will punt it for now, but will make a note to come back to this when it's more feasible to test. (will likely upgrade the kafka client before py3 anyways)

snopoke: should switch to `kafka.util.kafka_bytestring` which is used internally by the kafka lib

emord: yeah. just saw that in your partition PR https://github.com/dimagi/commcare-hq/pull/15530

PR #15504, opened 2017-03-23
don't append configs

emord:
config != configs

only affects prod because it's the only one that uses ucr_division right now


emord:
merging because tests pass locally and it's causing ucr issues on prod.

PR #15510, opened 2017-03-24
double the throughput on laboratory UCRs

emord:
Calls `self.config.get_all_values(doc)` once instead of twice for the UCRs. That is where all the db hits happen  to generate the rows.

This should be a huge boon to icds since all the UCRs in laboratory now. w=1 may be helpful for some of the review

familiar with UCR @NoahCarnahan @nickpell 

buddy @sravfeyn FYI @sheelio

### DIFF
--- a/corehq/apps/userreports/laboratory/adapter.py
+++ b/corehq/apps/userreports/laboratory/adapter.py
@@ -1,4 +1,3 @@
-from sqlalchemy.exc import IntegrityError
 from corehq.apps.userreports.adapter import IndicatorAdapter
 from corehq.apps.userreports.es.adapter import IndicatorESAdapter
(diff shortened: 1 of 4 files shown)


corehq/apps/userreports/laboratory/adapter.py
```diff
@@ -1,3 +1,4 @@
+from sqlalchemy.exc import IntegrityError
```
dimagimon: F401 'IntegrityError' imported but unused

NoahCarnahan (approved):
Nice find

emord:
@javierwilson @sravfeyn Not sure on the status of the nic deploy, but if this could get in that would be awesome

javierwilson:
sure. we'll deploy soon, and we may even succeed....

snopoke:
This is great, thanks!

PR #15513, opened 2017-03-24
Add filtering by timeEnd and xmlns to get_case_forms

emord:
@NoahCarnahan @sheelio 

You can check out how I structured this in the test in this commit https://github.com/dimagi/commcare-hq/commit/00e58ee02d9e94960c6ce76348951a0eedf71520

@sheelio I think it's important to note that ES isn't our primary DB, which means it is usually up to date, but likely shouldn't be trusted to be 100%. In practice I think this will sometimes give you an off-by-one on your forms where the latest form for the case is not yet in ES. I"m not sure if that's acceptable for your UCRs though

:blowfish: 

buddy @sravfeyn

### DIFF
--- a/corehq/apps/es/fake/es_query_fake.py
+++ b/corehq/apps/es/fake/es_query_fake.py
@@ -1,4 +1,6 @@
+import datetime
 import logging
+import pytz
 import uuid
 from corehq.apps.es.es_query import ESQuerySet
 from corehq.apps.es.utils import values_list
@@ -131,6 +133,36 @@ class ESQueryFake(object):
             return doc[field] in valid_terms
 
         return self._filtered(_term_query)
+
+    def date_range(self, field, gt=None, gte=None, lt=None, lte=None):
+        def format_time(t):
+            if t:
+                t = datetime.datetime(*(t.timetuple()[:6]))
+                return pytz.UTC.localize(t)
+            return t
+
+        gt = format_time(gt)
+        gte = format_time(gte)
+        lt = format_time(lt)
+        lte = format_time(lte)
+
+        def _date_comparison(doc):
+            if gt and doc:
+                if not doc[field] > gt:
+                    return False
+            if gte and doc:
+                if not doc[field] >= gte:
+                    return False
+            if lt and doc:
+                if not doc[field] < lt:
+                    return False
+            if lte and doc:
+                if not doc[field] <= lte:
+                    return False
+
+            return True
+
+        return self._filtered(_date_comparison)
 
     def __getattr__(self, item):
         """
--- a/corehq/apps/userreports/expressions/specs.py
+++ b/corehq/apps/userreports/expressions/specs.py
@@ -304,14 +304,33 @@ class FormsExpressionSpec(JsonObject):
         if context.get_cache_value(cache_key) is not None:
             return context.get_cache_value(cache_key)
 
-        xforms = FormProcessorInterface(domain).get_case_forms(case_id)
+        xforms = self._get_case_forms(case_id, context)
         if self.xmlns:
             xforms = [f for f in xforms if f.xmlns in self.xmlns]
 
-        xforms = [f.to_json() for f in xforms if f.domain == domain]
+        xforms = [self._get_form_json(f, context) for f in xforms if f.domain == domain]
 
         context.set_cache_value(cache_key, xforms)
         return xforms
+
+    def _get_case_forms(self, case_id, context):
+        cache_key = (self.__class__.__name__, 'helper', case_id)
+        if context.get_cache_value(cache_key) is not None:
+            return context.get_cache_value(cache_key)
+
+        domain = context.root_doc['domain']
+        xforms = FormProcessorInterface(domain).get_case_forms(case_id)
+        context.set_cache_value(cache_key, xforms)
+        return xforms
+
+    def _get_form_json(self, form, context):
+        cache_key = (self.__class__.__name__, 'xform', form.id)
+        if context.get_cache_value(cache_key) is not None:
+            return context.get_cache_value(cache_key)
+
+        form_json = form.to_json()
+        context.set_cache_value(cache_key, form_json)
+        return form_json
 
 
 class SubcasesExpressionSpec(JsonObject):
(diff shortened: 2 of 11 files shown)


corehq/apps/userreports/expressions/specs.py
```diff
@@ -314,6 +315,60 @@ def _get_forms(self, case_id, context):
         return xforms
 
 
+class FormsInDateExpressionSpec(JsonObject):
+    type = TypeProperty('get_case_forms_in_date')
```
sravfeyn: Should this be added to [UCR docs](https://github.com/dimagi/commcare-hq/blob/master/corehq/apps/userreports/README.md)?

corehq/apps/es/fake/es_query_fake.py
```diff
@@ -132,6 +134,36 @@ def _term_query(doc):
 
         return self._filtered(_term_query)
 
+    def date_range(self, field, gt=None, gte=None, lt=None, lte=None):
```
sravfeyn: This doesn't accept both `gt` and `lt` at the same time, right? If so, worth adding a comment indicating so.

sravfeyn: Sorry, I take back, it looks like it does.

sravfeyn:
Looks good, are you planning to finish the 'TODO' latter? If not you can self-merge.

corehq/apps/userreports/expressions/specs.py
```diff
@@ -278,6 +279,7 @@ def get_variables(self, item, context):
 class FormsExpressionSpec(JsonObject):
     type = TypeProperty('get_case_forms')
     case_id_expression = DefaultProperty(required=True)
+    xmlns = ListProperty(required=False)
```
NoahCarnahan: This should also get a docs update

corehq/apps/userreports/expressions/specs.py
emord: Actually decided I don't like this being in the main ucr app since it relies on ES. going to put it in the custom icds folder

sheelio:
@emord 

Guessing I should hold off on trying to use this until you move it back to the custom app?

emord:
@sheelio moved it to the custom app

@sravfeyn that todo is for quite a while in the future. I don't anticipate an issue until the cases get ~1k forms or so

NoahCarnahan (approved):

emord:
@sheelio I updated with some more incremental caching. When I ran the tests they were ~10 s faster before/after caching over 70ish tests, so hopefully that will translate into some gains on icds. I think there are still some get_case_forms in the actual data source that need to be moved over though

sheelio:
@emord 

Do you want make to make commits against this PR with the changes to the data sources?

sheelio:
@emord 

After looking at this, I think I want to restructure the "icds_get_case_forms_by_date" and "icds_get_case_history" expressions to take a db_xmlns parameter that we'd pass through to ES.  However, the expressions would still then further filter on XMLNS using standard UCR filtering.  

Do you want me to just push those changes to this PR?

custom/icds_reports/ucr/expressions.py
```diff
@@ -133,7 +139,11 @@ def __call__(self, item, context=None):
 
     def _get_forms(self, case_id, from_date, to_date, context):
         domain = context.root_doc['domain']
-        cache_hash = "{}{}{}".format(from_date.toordinal(), to_date.toordinal(), self.count)
+        cache_hash = "{}".format(self.count)
+        if from_date:
+            cache_hash += "{}".format(from_date.toordinal())
+        if to_date:
+            cache_hash += "{}".format(to_date.toordinal())
```
NoahCarnahan: I think you might need `None` values for from_date and to_date to be represented in the hash. Otherwise, an expression with "from:2017-03-28 to:None" and "from:None to:2017-03-28" will have the same hash.

Speaking of which, would it be preferable to just use a longer tuple that includes each argument as the `cache_key`?

corehq/apps/userreports/expressions/specs.py
```diff
@@ -304,15 +304,34 @@ def _get_forms(self, case_id, context):
...
         if self.xmlns:
             xforms = [f for f in xforms if f.xmlns in self.xmlns]
 
-        xforms = [f.to_json() for f in xforms if f.domain == domain]
+        xforms = [self._get_form_json(f, context) for f in xforms if f.domain == domain]
 
         context.set_cache_value(cache_key, xforms)
         return xforms
 
+    def _get_case_forms(self, case_id, context):
```
NoahCarnahan: Good thinking on this cacheing

NoahCarnahan (changes requested):

emord:
> After looking at this, I think I want to restructure the "icds_get_case_forms_by_date" and "icds_get_case_history" expressions to take a db_xmlns parameter that we'd pass through to ES. However, the expressions would still then further filter on XMLNS using standard UCR filtering.

@sheelio I think icds_get_case_forms_by_date already has that. I passed the same xmlns that was provided to the db. Can you make a pr into this pr with any other changes?

custom/icds_reports/ucr/expressions.py
emord: Good catch. originally i did hashing because xmlns is a list of strings and lists can't be used in keys, so now just make them tuples: https://github.com/dimagi/commcare-hq/pull/15513/commits/c8caa5802f72640b08165ae7189228507bf74ca6

custom/icds_reports/ucr/expressions.py
```diff
@@ -139,17 +138,9 @@ def __call__(self, item, context=None):
...
-        if from_date:
-            cache_hash += "{}".format(from_date.toordinal())
-        if to_date:
-            cache_hash += "{}".format(to_date.toordinal())
-        if self.xmlns:
-            cache_hash += ''.join(self.xmlns)
-
-        cache_hash = hashlib.md5(cache_hash).hexdigest()[:4]
+        cache_key = (self.__class__.__name__, case_id, self.count, from_date,
+                     to_date, tuple(self.xmlns))
```
NoahCarnahan: 👍

NoahCarnahan (approved):

PR #15514, opened 2017-03-24
app status

calellowitz:
@emord @benrudolph @snopoke 
Paginating the app status report. A few notes, this requires a forms reindex to work properly but can be done in place. I can setup a release to do that if this is approved. It will also require a user reindex, which changes the mapping file so needs to be done as a preindex.
There are a few small changes to functionality. Let me know if any of these are blockers and I can work out a solution. You can no longer sort by last sync date. I spoke with Ben and it sounds like that was planned. Sorting also works slightly differently. Null values always appear last, which seems fine to me. This just means that if some users havent submitted a form they will be at the end whether the sorting last submission in ascending or descending order. Finally, forms without an app id field no longer have app names, because we no longer fetch the form, and our last way to figure out what app a submission is from was to look at the xform metadata.


dimagimon:
There are 6 errors:

* corehq/apps/reports/standard/deployments.py, line 59 - E501 line too long (116 > 115 characters)
* corehq/apps/reports/standard/deployments.py, line 60 - E241 multiple spaces after ','
* corehq/apps/reports/standard/deployments.py, line 101 - E303 too many blank lines (2)
* corehq/apps/reports/standard/deployments.py, line 152 - E501 line too long (117 > 115 characters)
* corehq/apps/reports/standard/deployments.py, line 340 - E121 continuation line under-indented for hanging indent
* corehq/pillows/mappings/user_mapping.py, line 93 - E126 continuation line over-indented for hanging indent
E501 line too long (387 > 115 characters)

corehq/pillows/mappings/user_mapping.py
```diff
@@ -85,6 +85,23 @@
                 'last_name': {'type': 'string'},
                 'password': {'type': 'string'},
                 'registering_device_id': {'type': 'string'},
+                'reporting_metadata': {'dynamic': False,
+                                       'properties': {
+                                           'last_submissions': {
+                                               'dynamic': False,
+                                               'properties': {
+                                                    'submission_date': {'format': "yyyy-MM-dd||yyyy-MM-dd'T'HH:mm:ssZZ||yyyy-MM-dd'T'HH:mm:ss.SSSSSS||yyyy-MM-dd'T'HH:mm:ss.SSSSSS'Z'||yyyy-MM-dd'T'HH:mm:ss'Z'||yyyy-MM-dd'T'HH:mm:ssZ||yyyy-MM-dd'T'HH:mm:ssZZ'Z'||yyyy-MM-dd'T'HH:mm:ss.SSSZZ||yyyy-MM-dd'T'HH:mm:ss||yyyy-MM-dd' 'HH:mm:ss||yyyy-MM-dd' 'HH:mm:ss.SSSSSS||mm/dd/yy' 'HH:mm:ss",
```
E501 line too long (387 > 115 characters)

corehq/ex-submodules/pillowtop/processors/form.py
```diff
@@ -34,7 +34,7 @@ def process_change(self, pillow_instance, change):
             # the same effect, an app having has_submissions set to True.
             mark_has_submission(domain, build_id)
 
-        user_id = doc.get('user_id')
+        user_id = doc.get('form', {}).get('meta', {}).get('userID')
```
benrudolph: i think this'll work for couch, but not sql. sql stores the user id in a property whereas couch stores the entire form in the doc so it gets it from the metadata

testapps/test_elasticsearch/tests/test_prod_es_indices.py
```diff
@@ -55,7 +55,7 @@ def test_prod_config(self):
     },
     {
         "alias": "hqusers",
-        "index": "test_hqusers_2017-02-03",
+        "index": "test_hqusers_2017-03-22",
```
benrudolph: why do users have to be reindexed? if the user model changes in the form reindexer then it will automatically be queued to be transformed to the elasticsearch

corehq/apps/reports/standard/deployments.py
```diff
@@ -82,58 +80,57 @@ def selected_app_id(self):
...
             self.domain,
             mobile_user_and_group_slugs,
-            include_inactive=False,
-            limit_user_ids=limit_user_ids,
         )
-        return users_data.combined_users
+        user_query = (user_query
+                      .set_sorting_block(self.get_sorting_block())
+                      .size(self.pagination.count)
+                      .start(self.pagination.start))
```
benrudolph: 👏

benrudolph (commented):
nice, can't wait for this. my impression is that this change would pretty much make this report super fast (comparatively). does seem right?

corehq/pillows/mappings/user_mapping.py
emord: this should be the date string from https://github.com/dimagi/commcare-hq/blob/master/corehq/pillows/core.py#L20
 can you also change the other places in this file that have this string?

corehq/apps/reports/standard/deployments.py
```diff
@@ -82,58 +81,56 @@ def selected_app_id(self):
...
-                    xform_dict['form']['meta'],
-                )
+                app_version_info_from_form = AppVersionInfo(last_sub.get('build_version'),
+                                                            last_sub.get('commcare_version'), None)
 
             if app_name is None and self.selected_app_id:
                 continue
 
-            last_sync_log = SyncLog.last_for_user(user.user_id)
+            last_sync_log = SyncLog.last_for_user(user['_id'])
```
snopoke: Wouldn't be hard to add this to the `reporting_metadata` in the user as well (not in this PR though)

corehq/apps/reports/standard/deployments.py
```diff
@@ -82,58 +81,56 @@ def selected_app_id(self):
...
 
-                if xform_dict.get('app_id'):
+                if last_sub.get('app_id'):
                     try:
-                        app = get_app(self.domain, xform_dict.get('app_id'))
+                        app = get_app(self.domain, last_sub['app_id'])
                     except ResourceNotFound:
                         pass
                     else:
                         app_name = app.name
```
snopoke: If all we need is the app name then a well cached util function would be good. Either way the app lookup should be heavily cached.

snopoke (commented):
Looking good.

testapps/test_elasticsearch/tests/test_prod_es_indices.py
calellowitz: We need to the user index, to index a series of properties that werent in the mapping file. It is my understanding that that requires a reindex. Is this untrue? Either way, a user reindex should be pretty short.

benrudolph: ah sorry of course

corehq/ex-submodules/pillowtop/processors/form.py
calellowitz: i tested it in sql and it seems to work. `doc` should be the xform itself (which is why later on we can grab meta from it anyway), and the other method definitely doesnt work for couch (not sure if it does for sql)

benrudolph: ah this might be why - https://github.com/dimagi/commcare-hq/blob/master/corehq/form_processor/serializers.py#L44 (thought didn't look closely). yeah the other method only works on sql, forgot we didnt' store the user_id directly on the model for couch

dimagimon:
There are 5 errors:

* corehq/apps/reports/standard/deployments.py, line 34 - F401 'get_app_name' imported but unused
* corehq/apps/reports/standard/deployments.py, line 82 - F811 redefinition of unused 'get_app_name' from line 34
* corehq/apps/reports/util.py, line 9 - F401 'ResourceNotFound' imported but unused
* corehq/apps/reports/util.py, line 15 - F401 'get_app' imported but unused
* corehq/pillows/mappings/user_mapping.py, line 84 - E126 continuation line over-indented for hanging indent

calellowitz:
@snopoke @emord 
I think I hit everything but let me know if something is missing. It also looks like the form reindexers havent implemented the `in-place` flag. Is there a specific reason for that, or is that something I can add fairly easily

corehq/pillows/mappings/user_mapping.py
```diff
@@ -80,11 +71,28 @@
...
                                'type': 'date'},
                 'last_name': {'type': 'string'},
                 'password': {'type': 'string'},
                 'registering_device_id': {'type': 'string'},
+                'reporting_metadata': {'dynamic': False,
+                                       'properties': {
+                                           'last_submissions': {
+                                               'dynamic': False,
+                                               'properties': {
+                                                    'submission_date': {'format': DATE_FORMATS_STRING,
```

corehq/apps/reports/standard/deployments.py
```diff
@@ -80,60 +79,62 @@ def headers(self):
     def selected_app_id(self):
         return self.request_params.get(SelectApplicationFilter.slug, None)
 
+    @memoized
```
snopoke: If you use `quickcache` then it will cache over multiple requests.

calellowitz: Do you have an appropriate cache timeout in mind. I am not particularly scared of an outdated cache, because I dont think app names change often but I do not have a sense of what is a reasonable amount of time to keep it.

snopoke: 10mins? Basically so that if you run the report a few times you can get the benefit of the cache. This is particularly important for large apps like ICDS which take a long time to load and 'parse'

calellowitz:
@snopoke I think the only outstanding question is about the form reindex. If we cant do it in place we will need to create a new index right?

emord:
@calellowitz what makes you think in place won't work?

`./manage.py ptop_reindexer_v2 --in-place  sql-form`
`./manage.py ptop_reindexer_v2 --in-place  form`

both seem to work for me locally. With that said ES has been in a bad place recently. Maybe we should take this opportunity to bump the number of shards for forms?

calellowitz:
@emord I think we need to use `./manage.py ptop_reindexer_v2 couch-app-form-submission` and `./manage.py ptop_reindexer_v2 sql-app-form-submission` not `form` and `sql-form`

snopoke:
Since neither of those 2 delete or create ES indexes there is no need for `--in-place`. You should just be able to run them as is.

benrudolph:
> both seem to work for me locally. With that said ES has been in a bad place recently. Maybe we should take this opportunity to bump the number of shards for forms?

What is the thought behind increasing shards? I'm reading more about it, and I'm not sure why increased shards will help with load? Besides distributing space more evenly, my impression is that there is a bit of overhead to having more shards running on one node. It seems like the best thing would be to add ES nodes (if that's an option)

calellowitz:
> Since neither of those 2 delete or create ES indexes there is no need for --in-place. You should just be able to run them as is.

Awesome. Sounds ready to go then

snopoke:
From my perspective the main goal was to allow better allocation of data
(and hence requests). With only 5 shards each shard has 20% of the data and
so if you have 2,2,1 on the 3 machines that's quite an uneven distribution.
If you add in another index you might get forms (2,2,1) cases (2,2,1) which
is even more uneven.

This article seems to suggest 30GB per shard or 1.5-3 shards per node
(presumably per index):
https://qbox.io/blog/optimizing-elasticsearch-how-many-shards-per-index

I haven't read extensively on this topic so will defer to anyone who feels
that have a better grasp on the topic.

emord:
I think their reasoning is based on the memory limit but a ES employee said otherwise in this response: https://discuss.elastic.co/t/maximum-shard-size-in-elasticsearch/35202

smaller shards also might help reallocation happen more efficiently since it'd be moving around less data

benrudolph:
Ok I'm convinced it makes sense to increase the number of shards. I think one thing that is contributing to this mess is the fact that ES is attempting to rebalance the shards. The shard rebalancing has only been a thing recently. With more shards that seems like it'll be less of an issue

calellowitz:
removed hold off on merge tag, looking to get this out as soon as possible

benrudolph:
great, has the reindex been run?

calellowitz:
after offlining with ben, holding off on merge to run the reindexes

benrudolph:
Ok offlined with Cal a bit on this one. I think this a bit more complex than we think. Here's what I think is needed:

1. Split out the code that modifies the tracker pillow. Have that merged so new changes are processed. Then run a reindex for couch and sql. I believe if we run this reindex in a separate release, we will miss new changes processed since this does not change the index / checkpoint. (I believe that is how that works)

2. Once that has been merged and run, we can then deploy this to a private dir and run the user reindex. 

3. Merge this PR. On deploy, it will notice that a new user index has been created and process any changes that have taken place since the checkpoint.

calellowitz:
removed dont merge tag, and fixed conflicts, will merge after tests run

PR #15516, opened 2017-03-24
Paginate exports

benrudolph:
@NoahCarnahan this is the beginnings of paginating exports. This currently just creates a drop in replacement for `_Writer` that automatically pages export files. There are some other tasks that need to be tackled but will be done in different PRs to keep things small.

This is what it ends up looking like, it keeps the current row number too:
screen shot 2017-03-24 at 9 10 20 pm

cc: @millerdev

### DIFF
--- a/corehq/apps/export/export.py
+++ b/corehq/apps/export/export.py
@@ -89,8 +89,10 @@ class _Writer(object):
                         )
                     table_titles[table] = sheet_name
             self.writer.open(headers, file, table_titles=table_titles, archive_basepath=name)
-            yield
-            self.writer.close()
+            try:
+                yield
+            finally:
+                self.writer.close()
 
     def write(self, table, row):
         """
@@ -146,8 +148,10 @@ class _PaginatedWriter(object):
                 table_titles=self._get_paginated_table_titles(),
                 archive_basepath=self.name
             )
-            yield
-            self.writer.close()
+            try:
+                yield
+            finally:
+                self.writer.close()
 
     def _get_name(self, export_instances):
         if len(export_instances) == 1:
--- a/corehq/apps/export/export.py
+++ b/corehq/apps/export/export.py
@@ -5,12 +5,14 @@ import tempfile
 import datetime
 
 from couchdbkit import ResourceConflict
+from collections import Counter
 
 from soil import DownloadBase
 
 from couchexport.export import FormattedRow, get_writer
 from couchexport.files import Temp
 from couchexport.models import Format
+from corehq.toggles import PAGINATED_EXPORTS
 from corehq.util.files import safe_filename
 from corehq.apps.export.esaccessors import (
     get_form_export_base_query,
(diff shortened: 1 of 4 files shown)


corehq/apps/export/export.py
```diff
@@ -109,6 +112,164 @@ def path(self):
...
+        """
+
+        assert self._path is None
+
+        self.name = self._get_name(export_instances)
+        self.headers = self._get_headers(export_instances)
+        self.table_names = self._get_table_names(export_instances)
+
+        # Create and open a temp file
+        fd, self._path = tempfile.mkstemp()
```
benrudolph: i'm going to have to do more research into how this works. on icds, these exports could potentially last for days and have gigabytes of data

corehq/apps/export/export.py
```diff
@@ -109,6 +112,164 @@ def path(self):
...
+        with os.fdopen(fd, 'wb') as file_handle:
+            self.writer.open(
+                self._get_paginated_headers().iteritems(),
+                file_handle,
+                table_titles=self._get_paginated_table_titles(),
+                archive_basepath=self.name
+            )
+            yield
+            self.writer.close()
```
benrudolph: also something i'm not sure about is keeping files open for a very long time. currently it keeps every file open until this call to close. we could add more logic to close an individual page as we finish with it

corehq/apps/export/export.py
```diff
@@ -109,6 +112,164 @@ def path(self):
...
+        # Create and open a temp file
+        fd, self._path = tempfile.mkstemp()
+        with os.fdopen(fd, 'wb') as file_handle:
+            self.writer.open(
+                self._get_paginated_headers().iteritems(),
+                file_handle,
+                table_titles=self._get_paginated_table_titles(),
+                archive_basepath=self.name
+            )
+            yield
```
millerdev: You probably want to wrap this `yield` in `try`/`finally`. Otherwise, if there's an exception raised inside the context, the next line (`self.writer.close()`) will not be called.

millerdev (commented):
Reviewed pretty quickly, don't feel like I have enough context here to provide much useful feedback.

NoahCarnahan (approved):
Looks good Ben.

Am I right in saying that this sort of pagination is intended to deal with the issue of excel files becoming too big, but not do anything about the zip file itself being too large?

benrudolph:
yes that's correct. there are a bunch of issues with downloading gigabytes of data, but one of them is trying to open a file that is a gb in excel. other issues that we need to tackle is not storing the file in memory and then putting in redis. then we need a workflow that allows us to send a link to the user when the export is ready

corehq/apps/export/export.py
benrudolph: good call, will udpate

emord: if you use flake8 to check code you can install https://docs.openstack.org/developer/hacking/#python-3-x-compatibility to catch this. tried adding it to dimagimon but didn't have any luck

PR #15517, opened 2017-03-25
exclude UCR's for domains that don't exist

snopoke:


corehq/apps/userreports/pillow.py
```diff
@@ -49,6 +49,31 @@ def timed(*args, **kw):
...
+        if ucr_start <= table_hash <= ucr_end:
+            filtered_configs.append(config)
+    return filtered_configs
+
+
+def _exclude_missing_domains(configs):
+    from corehq.apps.es import DomainES
+    from corehq.elastic import ESError
+
+    config_domains = {conf.domain for conf in configs}
```
emord: I was going to do this by adding a `server_environment` variable to the static config and filtering on that, but this looks good for now. 

Could you limit this to only being filtered for static reports? This filter isn't useful for dynamic reports and I think this might not work well on prod, because I think there are some issues when you feed a large array to ES. @benrudolph you ran into that in reports right?

benrudolph: yeah we were trying to stuff 10k ids into an ES query. it's been an issue on and off, i think it just increases the time it takes to process the query and sometimes causes it to timeout

snopoke: updated

emord (approved):

PR #15531, opened 2017-03-27
adding retry

calellowitz:
@biyeun
This makes async builds (those triggered by updating the phone to the latest saved version of an app), retry if an older build is ongoing. I picked some relatively arbitrary numbers for retries and delay, but wanted to make sure the total time spent retrying was as long as the timeout for the lock.

cc: @amstone326 FYI

### DIFF
--- a/corehq/apps/app_manager/tasks.py
+++ b/corehq/apps/app_manager/tasks.py
@@ -12,7 +12,7 @@ def create_user_cases(domain_name):
         sync_usercase(user)
 
 
-@serial_task('{app._id}', max_retries=0, timeout=60*60)
+@serial_task('{app._id}', max_retries=30, default_retry_delay=120, timeout=60*60)
 def make_async_build(app):
     latest_build = app.get_latest_app(released_only=False)
     if latest_build.version == app.version:
--- a/corehq/apps/app_manager/tasks.py
+++ b/corehq/apps/app_manager/tasks.py
@@ -12,7 +12,7 @@ def create_user_cases(domain_name):
         sync_usercase(user)
 
 
-@serial_task('{app._id}', max_retries=30, default_retry_delay=120, timeout=60*60)
+@serial_task('{app._id}-{app.version}', max_retries=0, timeout=60*60)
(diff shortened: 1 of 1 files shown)


corehq/apps/app_manager/tasks.py
```diff
@@ -12,7 +12,7 @@ def create_user_cases(domain_name):
         sync_usercase(user)
 
 
-@serial_task('{app._id}', max_retries=0, timeout=60*60)
+@serial_task('{app._id}', max_retries=30, default_retry_delay=120, timeout=60*60)
```
dimagimon: E226 missing whitespace around arithmetic operator

calellowitz:
switched key to include version because it should be safe and will stop the waiting on a lock for retries. including @gcapalbo and @snopoke since they were on the first pr and may have context

corehq/apps/app_manager/tasks.py
```diff
@@ -12,7 +12,7 @@ def create_user_cases(domain_name):
         sync_usercase(user)
 
 
-@serial_task('{app._id}', max_retries=0, timeout=60*60)
+@serial_task('{app._id}-{app.version}', max_retries=0, timeout=60*60)
```
dimagimon: E226 missing whitespace around arithmetic operator

snopoke (approved):

gcapalbo:
@calellowitz can you explain where the retries are happening? it looks like this change just allows concurrent builds of different versions of the same app and celery still won't retry anything (maybe that's what's intended, just want to make sure).

biyeun (approved):

calellowitz:
@gcapalbo Im sorry, I changed the approach in the second commit and did not explain that well. The initial bug was that async builds would be missed, if subsequent builds were kicked off while one was completing due to the lock on app id. Adding version to that key should fix the problem, since we only want one build per version anyway, and now we dont need to worry about waiting for retries.

gcapalbo:
:+1: sounds good, thanks for clarifying

PR #15532, opened 2017-03-27
bump refresh_interval

benrudolph:
@emord @snopoke Updated here, this I believe won't get run unless we do a reindex (https://github.com/dimagi/commcare-hq/blob/br/refresh-interval/corehq/apps/hqcase/management/commands/ptop_preindex.py#L95). However, I've already updated the refresh_interval using the command line. I opted to not change the max_bytes_per_sec in order to isolate changes. We should really be doing some sort of performance benchmarking, but we aren't that close to having that setup. Instead I'm just going to rely on our monitoring and see what effect these changes have. I still think this isn't a bad setting to have for all indices. I don't think we care that much about a slightly longer delay in things showing up in ES, but feel free to push back

cc: @millerdev

### DIFF
--- a/corehq/ex-submodules/pillowtop/es_utils.py
+++ b/corehq/ex-submodules/pillowtop/es_utils.py
@@ -17,7 +17,7 @@ INDEX_REINDEX_SETTINGS = {
 
 INDEX_STANDARD_SETTINGS = {
     "index": {
-        "refresh_interval": "1s",
+        "refresh_interval": "5s",
         "merge.policy.merge_factor": 10,
         "store.throttle.max_bytes_per_sec": "5mb",
         "store.throttle.type": "node",


emord (approved):
> I still think this isn't a bad setting to have for all indices. I don't think we care that much about a slightly longer delay in things showing up in ES, but feel free to push back

I think the only one I'd be wary of is users since our users pages rely on ES pretty heavily (may be the only source of truth). 5s probably isn't a big deal though

benrudolph:
True, that code also fakes the user on the initial creation, so it will still always show whether or not that user was created and allow for you to click on the edit button. it will delay though in the user appearing in the user list.

this is the dialog i'm referring to in case not clear:
screen shot 2017-03-27 at 12 19 08 pm

PR #15533, opened 2017-03-27
warn emord about really expensive ucrs

emord:
@sravfeyn

### DIFF
--- a/corehq/apps/userreports/pillow.py
+++ b/corehq/apps/userreports/pillow.py
@@ -26,6 +26,7 @@ from pillowtop.utils import ensure_matched_revisions, ensure_document_exists
 from pillow_retry.models import PillowError
 
 REBUILD_CHECK_INTERVAL = 60 * 60  # in seconds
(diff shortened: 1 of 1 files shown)


corehq/apps/userreports/pillow.py
```diff
@@ -45,6 +46,7 @@ def timed(*args, **kw):
                 table.config._id, doc['_id'], seconds
             )
             pillow_logging.warning(message)
+            _slow_ucr_assert(seconds < 5, message)
```
emord: planning on reducing this eventually, but just want to get a feel for how many are longer than this

orangejenny (approved):

corehq/apps/userreports/pillow.py
sravfeyn: Do we get softasserts onto Sentry. Might be easy to do these kinds of analysis on Sentry than in email

emord: This doesn't do much other than alert me that a particular ucr is slow. I still need to go in and see why it's slow, so I don't think sentry would have much gain

PR #15535, opened 2017-03-27
POC epoch_second

benrudolph:
@emord @snopoke add this to the list of things that we would get if we could upgrade ES. looking into playing with some of the date formats


snopoke:
:+1:

PR #15536, opened 2017-03-27
[eNikshay] Store device ID

esoergel:
eNikshay wants some information about the current device available to the app as part of `user.user_data`.  The method `update_device_id_last_used` provides the necessary information for that to get updated, however, that's only called in `fetch_key_records` at the moment.

https://github.com/dimagi/commcare-hq/commit/71e699ec90594adb6199780f58c0368d166bf361 has changes to `corehq/apps/ota/views.py` that I'm unsure of.
@kaapstorm do you know why that is?  Is it correct to also update `user.devices` when users restore?
@snopoke pinging for suggestions on where this logic belongs, or a place to inject something custom if it doesn't.

@nickpell code buddy

### DIFF
--- a/corehq/apps/ota/views.py
+++ b/corehq/apps/ota/views.py
@@ -56,6 +56,11 @@ def restore(request, domain, app_id=None):
     """
     couch_user = CouchUser.from_django_user_include_anonymous(domain, request.user)
     assert couch_user is not None, 'No couch user to use for restore'
+    device_id = request.GET.get('device_id')
+    if device_id and isinstance(couch_user, CommCareUser):
+        if not couch_user.is_demo_user:
+            couch_user.update_device_id_last_used(device_id)
+            couch_user.save()
     response, _ = get_restore_response(domain, couch_user, app_id, **get_restore_params(request))
     return response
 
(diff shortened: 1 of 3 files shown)


corehq/apps/ota/views.py
```diff
@@ -56,6 +56,11 @@ def restore(request, domain, app_id=None):
     """
     couch_user = CouchUser.from_django_user_include_anonymous(domain, request.user)
     assert couch_user is not None, 'No couch user to use for restore'
+    device_id = request.GET.get('device_id')
+    if device_id and isinstance(couch_user, CommCareUser):
+        if not couch_user.is_demo_user:
+            couch_user.update_device_id_last_used(device_id)
+            couch_user.save()
```
snopoke: looks good but is identical to to code in `fetch_key_records` - good candidate for function

snopoke (commented):
Looks ok to me.

snopoke:
> do you know why that is? 
I assume you're asking why this doesn't get updated during restore and the answer is because it hasn't been needed.

esoergel:
> the answer is because it hasn't been needed

Perfect, that's what I was hoping for.  Thanks, will update this into my main branch.  (And convert that thing to a function)

esoergel:
Closing in favor of https://github.com/dimagi/commcare-hq/pull/15540

PR #15538, opened 2017-03-27
remove generator.instantiate_accounting() from more tests

nickpell:
removes all uses outside of accounting tests, where this is needed for creating the "test" plans which have lower user/SMS thresholds.


nickpell:
@esoergel

PR #15534, opened 2017-03-27
eNikshay ID generation

esoergel:
[spec](https://docs.google.com/document/d/11Nxk3XMuae9S4L3JZc4FCVTocLz6bOC-glclrgxnQ5o/edit#)
The compression code and tests are translated from https://github.com/dimagi/commcare-core/pull/539 (although I did pythonify the compression code)
@ctsims @nickpell 
:classical_building:

PR #15540, opened 2017-03-27
eNikshay ID generation

esoergel:
Travis is being stupid and won't let me see the test failure on https://github.com/dimagi/commcare-hq/pull/15534 so I'm opening a new PR to try to get that to work.


esoergel:
(most of this is stuff you already reviewed in https://github.com/dimagi/commcare-hq/issues/15534)

nickpell (approved):

corehq/tests/noseplugins/dividedwerun.py
```diff
@@ -84,7 +84,7 @@ def skip():
                         skip.__name__ = "*"
                     skip.__module__ = name
             else:
-                desc = test.test
+                desc = test.test.descriptor
```
millerdev: I wonder if this is necessary in all cases. Wonder if it might be more correct to do:
```py
if hasattr(test.test, "__name__"):
    desc = test.test
else:
    desc = test.test.descriptor
```

Shame on me for not writing tests for this 😞

esoergel: I'm not convinced - nose says this:
```
* descriptor -- the function, other than the test, that should be used
  to construct the test name. This is to support generator functions.
```
So it says that it shouldn't be the test itself, and it seems logical to just pass through the test descriptor, since this is the "descriptor" param.

millerdev: I'm fine with always using `test.test.descriptor` as long as it always exists and it has a descriptive (i.e., not generic) `__name__` for the test

PR #15544, opened 2017-03-28
send timestamp metric to datadog

sravfeyn:


corehq/ex-submodules/pillowtop/pillow/interface.py
```diff
@@ -206,7 +206,8 @@ def __record_change_metric_in_datadog(self, metric, change, timer=None):
                 u'document_type:{}'.format(change.metadata.document_type),
                 u'domain:{}'.format(change.metadata.domain),
                 u'is_deletion:{}'.format(change.metadata.is_deletion),
-                u'pillow_name:{}'.format(self.get_name())
+                u'pillow_name:{}'.format(self.get_name()),
+                u'publist_timestamp:{}'.format(change.metadata.publish_timestamp),
```
snopoke: This won't work - you need a new guage metric which records the timestamp as an epoch

snopoke (changes requested):

sravfeyn:
@snopoke Updated

corehq/ex-submodules/pillowtop/pillow/interface.py
```diff
@@ -206,10 +207,13 @@ def __record_change_metric_in_datadog(self, metric, change, timer=None):
                 u'document_type:{}'.format(change.metadata.document_type),
                 u'domain:{}'.format(change.metadata.domain),
                 u'is_deletion:{}'.format(change.metadata.is_deletion),
-                u'pillow_name:{}'.format(self.get_name())
+                u'pillow_name:{}'.format(self.get_name()),
             ]
             datadog_counter(metric, tags=tags)
 
+            change_lag = (datetime.utcnow() - change.metadata.publish_timestamp).seconds
+            datadog_gauge('commcare.change_feed.change_lag', change_lag, tags=tags)
```
snopoke: I don't think all these tags make sense for this metric. I think we could limit it to:
* pillow_name
* topic

I think you may have to add the topic to the change though. You can add it here: https://github.com/dimagi/commcare-hq/blob/be5c8ce9ea6f9cf1c16894570505172911ccbb7c/corehq/apps/change_feed/consumer/feed.py#L179
```
topic = message.topic
```

snopoke (changes requested):

corehq/ex-submodules/pillowtop/pillow/interface.py
sravfeyn: https://github.com/dimagi/commcare-hq/pull/15544/commits/33f627431237443bdc2b4a45aa35f8d6925e2d67

corehq/apps/change_feed/consumer/feed.py
```diff
@@ -181,6 +181,7 @@ def change_from_kafka_message(message):
         deleted=change_meta.is_deletion,
         metadata=change_meta,
         document_store=document_store,
+        topic=message.topic,
```
snopoke: You'll need to add this kwarg to the `Change` class as well

snopoke (changes requested):

corehq/apps/change_feed/consumer/feed.py
sravfeyn: Sorry, thought it was already there

Updated now

snopoke (approved):

sravfeyn:
Is there anything else needed to be done on datadog UI for this metric to reflect there?

snopoke:
@sravfeyn you'll need to add a graph to see the data but you don't have to do anything for it to be collected

PR #15546, opened 2017-03-28
Use es scroll when getting case ids

gcapalbo:
@proteusvacuum

### DIFF
--- a/corehq/apps/data_interfaces/models.py
+++ b/corehq/apps/data_interfaces/models.py
@@ -82,7 +82,7 @@ class AutomaticUpdateRule(models.Model):
                  .exclude_source()
                  .size(chunk_size))
 
-        if boundary_date is not None:
+        if boundary_date:
             query = query.server_modified_range(lte=boundary_date)
 
         result = []
@@ -96,7 +96,7 @@ class AutomaticUpdateRule(models.Model):
                 yield result
                 result = []
 
-        if len(result) > 0:
+        if result:
             yield result
 
     def rule_matches_case(self, case, now):
--- a/corehq/apps/data_interfaces/models.py
+++ b/corehq/apps/data_interfaces/models.py
@@ -73,36 +73,31 @@ class AutomaticUpdateRule(models.Model):
         Retrieves the case ids in chunks, yielding a list of case ids each time
         until there are none left.
         """
-        offset = 0
-        chunk_size = 10000
+        chunk_size = 100
 
-        # We have to sort on _uid; _id is not sortable. See:
-        # https://www.elastic.co/guide/en/elasticsearch/reference/current/mapping-uid-field.html
-        # https://www.elastic.co/guide/en/elasticsearch/reference/current/mapping-id-field.html
+        query = (CaseES()
+                 .domain(domain)
+                 .case_type(case_type)
+                 .is_closed(closed=False)
+                 .exclude_source()
+                 .size(chunk_size))
 
-        while True:
-            query = (CaseES()
-                     .domain(domain)
-                     .case_type(case_type)
-                     .is_closed(closed=False)
-                     .exclude_source()
-                     .sort('_uid')
-                     .start(offset)
-                     .size(chunk_size))
+        if boundary_date is not None:
+            query = query.server_modified_range(lte=boundary_date)
 
-            if boundary_date is not None:
-                query = query.server_modified_range(lte=boundary_date)
+        result = []
 
-            results = query.run()
-            ids = results.doc_ids
+        for case_id in query.scroll():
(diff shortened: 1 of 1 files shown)


corehq/apps/data_interfaces/models.py
```diff
@@ -73,36 +73,31 @@ def get_case_ids(cls, domain, case_type, boundary_date=None):
...
+        if boundary_date is not None:
+            query = query.server_modified_range(lte=boundary_date)
 
-            if boundary_date is not None:
-                query = query.server_modified_range(lte=boundary_date)
+        result = []
 
-            results = query.run()
-            ids = results.doc_ids
+        for case_id in query.scroll():
```
emord: Was this due to the ES issues we've had on prod recently? 100 seems pretty low.

The code here looks very similar, but if there's not a scale issue, you could use `get_ids` https://github.com/dimagi/commcare-hq/blob/3dbdc2dd6d9dab3ecdc29093b991945d71bd06d1/corehq/apps/es/es_query.py#L429

Alternatively we could change get_ids to use scroll if it's large.

gcapalbo: so there's a few things going on here - we get these case ids to process rules against the cases, and so whatever the chunk size is we need to be able to process them all within the scroll timeout otherwise the next scroll query will fail.  since the iter_cases methods on CaseAccessors only do 100 cases at a time anyway, i figured it's a good place to start.

also i had first started off with using .run().doc_ids but that was timing out for domains with large numbers of cases so i had to take this approach.

emord: Ah. interesting I missed the yield in the for loop.

I think the timeout is 5 minutes? if you think the processing time wouldbe larger than that, this could be generate a list of lists? I'm not sure what the failure case is with update rules so if it fails gracefully then thsi should be fine

corehq/apps/data_interfaces/models.py
```diff
@@ -73,36 +73,31 @@ def get_case_ids(cls, domain, case_type, boundary_date=None):
...
-        while True:
-            query = (CaseES()
-                     .domain(domain)
-                     .case_type(case_type)
-                     .is_closed(closed=False)
-                     .exclude_source()
-                     .sort('_uid')
-                     .start(offset)
-                     .size(chunk_size))
+        if boundary_date is not None:
```
emord: why not `if boundary_date:`

corehq/apps/data_interfaces/models.py
```diff
@@ -73,36 +73,31 @@ def get_case_ids(cls, domain, case_type, boundary_date=None):
...
+            result.append(case_id)
+            if len(result) >= chunk_size:
+                yield result
+                result = []
 
-            if len(ids) < chunk_size:
-                return
-
-            offset += chunk_size
+        if len(result) > 0:
```
emord: I think this could be `if result`

emord (approved):

corehq/apps/data_interfaces/models.py
gcapalbo: updated https://github.com/dimagi/commcare-hq/pull/15546/commits/4dbbc51925313567fe91a98f9b21a61c14c7f8f1

corehq/apps/data_interfaces/models.py
gcapalbo: updated https://github.com/dimagi/commcare-hq/pull/15546/commits/4dbbc51925313567fe91a98f9b21a61c14c7f8f1

corehq/apps/data_interfaces/models.py
emord: if the function that it yields to takes > scroll timeout, the scrolling will cease to work right?

If we're worried that that function will take that long, you could instead get all the ids in a list of lists then yield that in a separate for loop.

With that said. I think this is fine the way it is

gcapalbo:
On Tue, Mar 28, 2017 at 11:48 AM, Jonathan Emord <notifications@github.com>
wrote:

> *@emord* commented on this pull request.
> ------------------------------
>
> In corehq/apps/data_interfaces/models.py
> <https://github.com/dimagi/commcare-hq/pull/15546#discussion_r108460168>:
>
> >
> -            results = query.run()
> -            ids = results.doc_ids
> +        for case_id in query.scroll():
>
> if the function that it yields to takes > scroll timeout, the scrolling
> will cease to work right?
>
yup, that's right


> If we're worried that that function will take that long, you could instead
> get all the ids in a list of lists then yield that in a separate for loop.
>
i see what you mean now - another thing i'm trying to avoid for large scale
is keeping all the ids in memory

PR #15547, opened 2017-03-28
ucr server environment

emord:
@snopoke icds-cas and icds-test exist in prod so it'll still try to load those configs. This adds an optional server environment to data source and report configs

@sravfeyn

### DIFF
--- a/corehq/apps/userreports/models.py
+++ b/corehq/apps/userreports/models.py
@@ -495,6 +495,7 @@ class StaticDataSourceConfiguration(JsonObject):
     """
     _datasource_id_prefix = STATIC_PREFIX
     domains = ListProperty()
+    server_environment = ListProperty(default=["production"])
     config = DictProperty()
 
     @classmethod
@@ -525,6 +526,9 @@ class StaticDataSourceConfiguration(JsonObject):
     @classmethod
     def all(cls):
         for wrapped, path in cls._all():
+            if settings.SERVER_ENVIRONMENT not in wrapped.server_environment:
+                continue
+
             for domain in wrapped.domains:
                 yield cls._get_datasource_config(wrapped, domain)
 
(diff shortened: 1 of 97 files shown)


corehq/apps/userreports/models.py
```diff
@@ -525,6 +526,10 @@ def _all(cls):
     @classmethod
     def all(cls):
         for wrapped, path in cls._all():
+            if (wrapped.server_environment and
```
emord: Eventually this should be required, but just want to get the icds reports off prod ES

corehq/apps/userreports/pillow.py
```diff
@@ -115,9 +114,6 @@ def get_filtered_configs(self, configs=None):
         elif self.ucr_division:
             configs = _filter_by_hash(configs, self.ucr_division)
 
-        if self.filter_missing_domains:
-            configs = _exclude_missing_domains(configs)
```
snopoke: can delete this function too

snopoke (approved):

corehq/apps/userreports/pillow.py
emord: https://github.com/dimagi/commcare-hq/pull/15547/commits/80946e83ebb8ca306c6a73993bbb4e400cceba4b

corehq/apps/userreports/models.py
```diff
@@ -495,6 +495,7 @@ class StaticDataSourceConfiguration(JsonObject):
     """
     _datasource_id_prefix = STATIC_PREFIX
     domains = ListProperty()
+    server_environment = ListProperty()
```
sravfeyn: (Minor), if we expect this to contain multiple environments, `server_environments` might be more suitable name.

sravfeyn (approved):

corehq/apps/userreports/models.py
emord: will make this change once the rest of hte ucrs get filled out with environments

PR #15550, opened 2017-03-28
Increase shards

benrudolph:
We may want to combine this with @calellowitz [PR](https://github.com/dimagi/commcare-hq/pull/15514), but I never love combining changes, so would rather not if we think it's doable to just run the AppSubmissionTracker pillow reindex on its own. I looked at our disk usage and I think we'll actually be fine not reallocating another node. Our most full disk still has around 120 GB of free space. hqcases is 61gb and xforms is 110gb. That will roughly add 60GB - 70GB to each node conservatively. 

1. `fab <softlayer|production> setup_release:keep_days=10 --set code_branch=br/more-shards`
1.5. disable shard reallocation
2. `./manage.py ptop_preindex --reset` (in a screen)
3. Merge this PR
4. Deploy
5. Run `./manage.py prune_elastic_indices` once we feel good that everything is settled

es head should then look something like this except distributed amongst 3 nodes
screen shot 2017-03-28 at 1 20 44 pm

### DIFF
--- a/corehq/pillows/mappings/case_mapping.py
+++ b/corehq/pillows/mappings/case_mapping.py
@@ -1,10 +1,11 @@
+from copy import deepcopy
 from corehq.pillows.base import DEFAULT_META
 from corehq.pillows.core import DATE_FORMATS_ARR, DATE_FORMATS_STRING
 from corehq.pillows.mappings import NULL_VALUE
 from corehq.util.elastic import es_index
 from pillowtop.es_utils import ElasticsearchIndexInfo
 
-CASE_INDEX = es_index("hqcases_2016-03-04")
+CASE_INDEX = es_index("hqcases_2017-03-28")
 CASE_ES_TYPE = 'case'
 
 CASE_MAPPING = {
@@ -104,10 +105,17 @@ CASE_MAPPING = {
 
 CASE_ES_ALIAS = "hqcases"
 
+CASE_INDEX_META = deepcopy(DEFAULT_META)
+CASE_INDEX_META.update({
+    'settings': {
+        'number_of_shards': 10,
+    },
+})
+
 CASE_INDEX_INFO = ElasticsearchIndexInfo(
     index=CASE_INDEX,
     alias=CASE_ES_ALIAS,
     type=CASE_ES_TYPE,
-    meta=DEFAULT_META,
+    meta=CASE_INDEX_META,
     mapping=CASE_MAPPING
 )
--- a/corehq/pillows/mappings/case_mapping.py
+++ b/corehq/pillows/mappings/case_mapping.py
@@ -1,6 +1,5 @@
 from copy import deepcopy
 from django.conf import settings
-from corehq.pillows.base import DEFAULT_META
 from corehq.pillows.core import DATE_FORMATS_ARR, DATE_FORMATS_STRING
 from corehq.pillows.mappings import NULL_VALUE
 from corehq.util.elastic import es_index
@@ -106,19 +105,9 @@ CASE_MAPPING = {
 
 CASE_ES_ALIAS = "hqcases"
 
-CASE_INDEX_META = deepcopy(DEFAULT_META)
-
-if settings.SERVER_ENVIRONMENT == 'production':
-    CASE_INDEX_META.update({
-        'settings': {
-            'number_of_shards': 10,
-        },
-    })
-
 CASE_INDEX_INFO = ElasticsearchIndexInfo(
     index=CASE_INDEX,
     alias=CASE_ES_ALIAS,
     type=CASE_ES_TYPE,
-    meta=CASE_INDEX_META,
     mapping=CASE_MAPPING
 )
--- a/corehq/pillows/mappings/case_mapping.py
+++ b/corehq/pillows/mappings/case_mapping.py
@@ -1,5 +1,3 @@
-from copy import deepcopy
-from django.conf import settings
 from corehq.pillows.core import DATE_FORMATS_ARR, DATE_FORMATS_STRING
 from corehq.pillows.mappings import NULL_VALUE
 from corehq.util.elastic import es_index
--- a/corehq/pillows/mappings/case_mapping.py
+++ b/corehq/pillows/mappings/case_mapping.py
@@ -3,7 +3,7 @@ from corehq.pillows.mappings import NULL_VALUE
 from corehq.util.elastic import es_index
 from pillowtop.es_utils import ElasticsearchIndexInfo
 
-CASE_INDEX = es_index("hqcases_2017-03-28")
+CASE_INDEX = es_index("hqcases_2016-03-04")
 CASE_ES_TYPE = 'case'
 
 CASE_MAPPING = {
--- a/corehq/pillows/mappings/case_mapping.py
+++ b/corehq/pillows/mappings/case_mapping.py
@@ -1,4 +1,5 @@
 from copy import deepcopy
+from django.conf import settings
 from corehq.pillows.base import DEFAULT_META
 from corehq.pillows.core import DATE_FORMATS_ARR, DATE_FORMATS_STRING
 from corehq.pillows.mappings import NULL_VALUE
@@ -106,11 +107,13 @@ CASE_MAPPING = {
 CASE_ES_ALIAS = "hqcases"
 
 CASE_INDEX_META = deepcopy(DEFAULT_META)
-CASE_INDEX_META.update({
-    'settings': {
-        'number_of_shards': 10,
-    },
-})
+
+if settings.SERVER_ENVIRONMENT == 'production':
+    CASE_INDEX_META.update({
+        'settings': {
+            'number_of_shards': 10,
+        },
+    })
 
 CASE_INDEX_INFO = ElasticsearchIndexInfo(
     index=CASE_INDEX,
--- a/env_settings.py
+++ b/env_settings.py
@@ -88,7 +88,6 @@ ES_META = {
         },
         'hqcases': {
             'settings': {
-                'number_of_shards': 10,
                 "analysis": {
                     "analyzer": {
                         "default": {
--- a/env_settings.py
+++ b/env_settings.py
@@ -1,0 +1,91 @@
+ES_META = {
+    # Default settings for all indexes on ElasticSearch
+    'default': {
+        "settings": {
+            "analysis": {
+                "analyzer": {
+                    "default": {
+                        "type": "custom",
+                        "tokenizer": "whitespace",
+                        "filter": ["lowercase"]
+                    },
+                    "sortable_exact": {
+                        "type": "custom",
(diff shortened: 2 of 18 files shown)


benrudolph:
arguably might be better to have 9 shards

emord:
What if we made number of shards dependent on `SERVER_ENVIRONMENT`? I imagine we don't want the overhead on monoliths like swiss.

Also we should think about icds and thing about how many shards we expect to need

benrudolph:
that's a good point, it's a little unfortunate that all of this gets set in the django code. i think it may be best to specify in the localsettings, what do you think?

emord:
Yeah. I'm pro moving anything like that to localsettings

benrudolph:
actually doing this in the localsettings makes it really hard to reindex in a private release

benrudolph:
https://github.com/dimagi/commcare-hq/pull/15550/commits/60b6717d38bbbf83b01753b93d2bb46d8e936179 -- this makes me cringe, but if we move this to localsettings we have to make some infrastructure for ansible to be able to deploy to a private release which seems non-intuitive for changes in the future

emord:
ouch. that's gross, but agreed we don't have a great way to do that with ansible yet.

emord (approved):
Looks good. Maybe it'd be worthwhile to set https://github.com/dimagi/commcare-hq/blob/d26d796c9c50edac166fb01bc83719381482e0ce/corehq/ex-submodules/pillowtop/es_utils.py#L10 to -1 and index it once the reindex finishes? or at least bump it past 15 min?

millerdev (commented):
the code changes look fine, but I agree would be better to have this in env-specific configuration like localsettings. I didn't understand the above-mentioned problem with using localsettings, but that's probably because I don't have enough context on ES in general.

benrudolph:
ok doubled it to 30 mins.

@millerdev it's hard with ansible since ansible always deploys to the `current` release. however we want to do this in a private release so we can build up the indices separately, and then when we merge this PR the switch over to the new indices will be fairly painless

corehq/pillows/mappings/case_mapping.py
```diff
@@ -104,10 +106,19 @@
...
 CASE_ES_ALIAS = "hqcases"
 
+CASE_INDEX_META = deepcopy(DEFAULT_META)
+
+if settings.SERVER_ENVIRONMENT == 'production':
+    CASE_INDEX_META.update({
+        'settings': {
+            'number_of_shards': 10,
+        },
+    })
```
snopoke: could we make use of the `environments.yml` file for this kind of thing?

snopoke:
```
FAIL: test_instantiate_all (testapps.test_pillowtop.tests.test_settings.PillowtopSettingsTest)
  File "/mnt/commcare-hq-ro/testapps/test_pillowtop/tests/test_settings.py", line 30, in test_instantiate_all
    self.assertEqual(expected_meta[pillow_config.name], _pillow_meta_from_config(pillow_config))
AssertionError: {u'change_feed_type': u'KafkaChangeFeed', u'full_class_name': u'pillowtop.pillow [truncated]... != {'change_feed_type': 'KafkaChangeFeed', 'checkpoint_id': u'CaseToElasticsearchPi [truncated]...
- {u'advertised_name': u'CaseToElasticsearchPillow',
?  -                   -
+ {'advertised_name': 'CaseToElasticsearchPillow',
-  u'change_feed_type': u'KafkaChangeFeed',
?  -                    -
+  'change_feed_type': 'KafkaChangeFeed',
-  u'checkpoint_id': u'CaseToElasticsearchPillow-test_hqcases_2016-03-04',
?  -                                                             ^    ^^
+  'checkpoint_id': u'CaseToElasticsearchPillow-test_hqcases_2017-03-28',
?                                                               ^    ^^
-  u'full_class_name': u'pillowtop.pillow.interface.ConstructedPillow',
?  -                   -
+  'full_class_name': 'pillowtop.pillow.interface.ConstructedPillow',
-  u'name': u'CaseToElasticsearchPillow'}
?  -        -
+  'name': 'CaseToElasticsearchPillow'}
```

dannyroberts:
@benrudolph for those of us who weren't following closely on slack#pillows but want to understand, can you go over some basics such as

- how many shards there were before/are by default
- whether shards are a per-index thing (looks like they are)
- basic mental model I can use of what shards in ES are
- what we expect to happen by increasing the number of shards
- what the tradeoff is (i.e. why 10 is better than 1 but worse than 1000)

Thanks, and sorry to make you summarize, but hopefully the explanation will be useful to others as well.

corehq/pillows/mappings/case_mapping.py
benrudolph: we could do that, i'm a little hesitant because then our ES settings would live in ansible, HQ, and commcare-hq-deploy. i was already confused initially about where to make this change

benrudolph:
@dannyroberts sure

> - how many shards there were before/are by default


By default there are 5

> - whether shards are a per-index thing (looks like they are)


Yes

> - basic mental model I can use of what shards in ES are


Shards are under the hood Lucene indexes. Elasticsearch indexes are made up of many Lucene indexes in order to scale horizontally. Elasticsearch tries to balance the shards evenly across the number of nodes (servers). Currently we have 5 shards that is attempting to be balanced across 3 nodes, so it's a bit uneven at the moment.

> - what we expect to happen by increasing the number of shards


Two things. 1) a more even distribution of shards/shard size. Since shards are the ~same size, have 2 20GB shards and 1 node and 1 20GB shard on another is a fairly big difference. However if we have 3 shards at 10GB on one node and 4 shards of 10GB on another that's not as large as a difference. 2) We think that when ES tries to rebalance shards, the form and case shards are too big and causing it to fall over. We hope that with smaller shards it will be able to more easily to balance the shards

> - what the tradeoff is (i.e. why 10 is better than 1 but worse than 1000)


My understanding, which may not be totally correct, is that there is overhead to maintaining a larger number of shards and querying across all of them, so we don't want an outrageously high number if we can help. At the same time we want our shards to be evenly balanced amongst our nodes so that we can distribute the load more evenly, which is why 10 is better than 1.

Let me know if any of that makes (or @emord / @snopoke correct me if I've misunderstood something)

snopoke:
That's true. I think localsettings is the right place for this to live with
our current workflows. The issue that you mentioned earlier is that we're
changing the value Pre deploy which doesn't really work right now.

If we changed the workflow to make Ansible create the indexes (instead of
HQ) then this wouldn't be an issue but it would mean an extra step anytime
we added an index which isn't very often.

benrudolph:
I would be very pro ansible managing our ES indexes. it's weird to me as is that they are so closely tied to pillows

emord:
One thing that might make that hard is that UCR data sources can create ES indexes. Although I think there could be a workaround for that if the datasources are refactored

snopoke:
I was only thinking of the static indexes - fine to have UCR's dynamically
create indexes.

dannyroberts:
Thanks @benrudolph https://github.com/dimagi/commcare-hq/pull/15550#issuecomment-290129280 was helpful

emord:
@benrudolph not sure where your experimentation with different date formats left off, but if it seemed promising or maybe inconclusive due to too little data, it might be a good time to add some extra columns for formats on received on so we can profile

benrudolph:
i stopped with the `epoch_second` (figured it wasn't the root cause to our pain, so stopped looking into it). agree would be fairly easy to add some new formats here. i'm a bit torn on this PR as i'm not sure how to move forward.

1. moving all of this to ansible feels out of scope for this change
2. move this configuration to commcare-hq-deploy environments.yml 
3. continue the hacky way forward
4. we create a `prod_settings.py` file (similar to `dev_settings.py`) where we configure all of this

emord:
I'm ok with 3, but maybe 4 would be better.

what if we had some structure:
```json
{
    "production": {
         "xforms_index": {"settings": {"num_shards": 20}}
     },
    "softlayer": {... },
    ....
}
```
and instead of having the if statement in random pillow mapping, we just grabbed the corresponding index settings in all mapping files and merged the dictionaries?

that's slightly less hacky and still keeps the differences in one place

benrudolph:
yeah that's exactly what i'm thinking. ok cool i'll do that

benrudolph:
there are some gotcha's with that approach. i'm going to bring this up at the scale meeting tomorrow discuss

dimagimon:
There are 5 errors:

* corehq/ex-submodules/pillowtop/es_utils.py, line 74 - F821 undefined name 'index_info'
* corehq/pillows/mappings/case_mapping.py, line 1 - F401 'deepcopy' imported but unused
* corehq/pillows/mappings/case_mapping.py, line 2 - F401 'settings' imported but unused
* corehq/pillows/mappings/xform_mapping.py, line 1 - F401 'deepcopy' imported but unused
* corehq/pillows/mappings/xform_mapping.py, line 2 - F401 'settings' imported but unused

benrudolph:
Ok this went through some churn so probably better to review all at once. However this commit https://github.com/dimagi/commcare-hq/pull/15550/commits/a7531734c46a4e680c81f4a1353855b6d4213b80 is pretty important because it specifies the change we are making to the prod indices

snopoke (approved):

env_settings.py
```diff
@@ -0,0 +1,109 @@
...
+                            "type": "custom",
+                            "tokenizer": "keyword",
+                            "filter": ["lowercase"]
+                        },
+                    },
+                },
+            },
+        }
+    },
+}
```
snopoke: Need to add ICDS to this as well. With the new infra we'll have 4 ES nodes. Not sure how many shards we want.

emord (approved):

emord:
@benrudolph do you plan on temporarily disabling shard reallocation while doing this?

benrudolph:
> do you plan on temporarily disabling shard reallocation while doing this?

yes, we may also want to permanently disable that for prod considering the cluster always goes down when that happens.

env_settings.py
benrudolph: 👍 i may do that in another PR unless the timing works out well

benrudolph:
https://github.com/dimagi/commcare-hq/pull/15550/commits/13e710d6b0abd74ed2bf9ffb3cbbb04e6afcd786 - @sravfeyn i don't have context on this change, is it fine for me to just have this commit? or do i need to include hq changes?

emord:
I think depending on the answer to https://github.com/dimagi/commcare-hq/pull/14535/files#r109478389 You may be able to not do this

sravfeyn:
The additional change required is to update `CASE_INDEX` per emord's comment.

benrudolph:
Ok this now only will affect the form index

emord:
looks good from my end. Do you want to pull out everything except for the index changes into another PR? might get annoying if there are conflicts in the next 2 weeks while the reindex runs

benrudolph:
good thought (not pressing though, can do that at any time)

PR #15551, opened 2017-03-28
ICDS: Update case history expression and data sources

sheelio:
@emord 

Updated the data sources and also changed your case history expression slightly.


custom/icds_reports/ucr/expressions.py
```diff
@@ -966,6 +970,9 @@ def get_case_history_by_date(spec, context):
         "case_id_expression": case_id_expression
     }
 
+    if spec['xmlns'] is not None and len(spec['xmlns']) > 0:
```
emord: I think this can just be `if spec['xmlns']`

custom/icds_reports/ucr/data_sources/ccs_record_cases_monthly_tableau.json
```diff
@@ -1982,41 +1981,35 @@
...
-              "http://openrosa.org/formdesigner/2864010F-B1B1-4711-8C59-D5B2B81D65DB"
-            ],
-            "start_date": {
-              "type": "icds_month_start"
-            },
-            "end_date": {
-              "type": "icds_month_end"
-            }
-          }
+          "type": "icds_get_case_forms_in_date",
```
emord: This needs a `case_id_expression`. Can either add it here or make it not required in ucr/expressions

emord (changes requested):

emord:
@sheelio I added https://github.com/dimagi/commcare-hq/pull/15551/commits/fb6baae7434fd5f3318be06329a5a270a9d94831 I used from_date_expression/to_date_expression as that was the terminology used in the core ucr code.

There's still an issue with your refactor to `nutrition_status_last_recorded` and `current_month_nutrition_status` that causes tests to fail. It looks like a great simplification but not sure I have enough context to debug that. I tried reverting those two changes and the tests pass, so I think that's all that's blocking this PR

emord (approved):

emord:
@sheelio I added https://github.com/dimagi/commcare-hq/pull/15551/commits/445296a2f3d9f0ba988de88cf8b6f6b3d15d3137 I decided to rely on get_case_forms cache because this class was mostly a pass through for the json. In between  runs the start_date and end_date that are passed here are the same `{end_date: icds_month_end}` which didn't look as easy to refactor

sheelio:
Awesome thanks!

PR #15552, opened 2017-03-28
ucr ext use new case forms

emord:
first commit is where the real change is. second commit is just changing line endings from windows to unix cause git (or maybe one of my hooks?) yelled at me

@sravfeyn


sravfeyn:
What is `ucr_ext` for?

emord:
@sravfeyn I think that it's just a place to have expressions that are made from other expressions, and don't have any db hits or calculations of their own

PR #15554, opened 2017-03-28
Ce/login as

calellowitz:
http://manage.dimagi.com/default.asp?250262#1309207
@benrudolph @biyeun cc: @snopoke


benrudolph (approved):
yay for another priv

corehq/apps/hqwebapp/templatetags/hq_shared_tags.py
```diff
@@ -257,7 +257,7 @@ def can_use_restore_as(request):
     return (
         request.couch_user.can_edit_commcare_users() and
         has_privilege(request, privileges.DATA_CLEANUP) and
```
benrudolph: realizing that `DATA_CLEANUP` is only for pro plan right? so this pretty much means the `LOGIN_AS` priv isn't helpful. should we remove this as well?

calellowitz: seems fine to me, unless there is a reason that was put there in the first place. seems odd that we had two conditions, both of which really just check if the domain is on the pro plan

benrudolph: can't remember why that was added, but seems silly in retrospect

benrudolph: i mean i understand why it's there, it's because with login as you can technically do a lot of data cleanup that you can't do normally on standard plan. but if we want to make it available for standard plan then we have to remove it...

PR #15559, opened 2017-03-29
Added one-off command to export form impact data

kkrampa:
@calellowitz 
http://manage.dimagi.com/default.asp?250613#1309541

### DIFF
--- a/corehq/apps/data_analytics/management/commands/export_form_impact_data.py
+++ b/corehq/apps/data_analytics/management/commands/export_form_impact_data.py
@@ -1,0 +1,83 @@
+from collections import defaultdict
+
+import csv
+
+import dateutil
+from dateutil.relativedelta import relativedelta
+
+from dimagi.utils.couch.database import iter_docs
+from django.core.management.base import BaseCommand
+
+from corehq.apps.data_analytics.esaccessors import get_forms_for_users
+from corehq.apps.domain.models import Domain
+from corehq.apps.users.models import CommCareUser
+
+
+class Command(BaseCommand):
+    """
+        Export form impact data to given csv file
+        e.g. ./manage.py export_form_impact_data February-2017 example.csv
+    """
+    help = 'Adds data to MALT table from given files'
+
+    def add_arguments(self, parser):
+        parser.add_argument(
+            'month_year'
+        )
+        parser.add_argument(
+            'file_path'
+        )
+
+    def handle(self, month_year, file_path, **options):
+        month_year_parsed = dateutil.parser.parse('1-' + month_year)
+        start_date = month_year_parsed.replace(day=1)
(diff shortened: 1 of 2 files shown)


kkrampa:
@fyniac

fyniac:
@calellowitz please review

corehq/apps/data_analytics/management/commands/export_form_impact_data.py
```diff
@@ -0,0 +1,85 @@
...
+                'user id',
+                'total number of forms submitted in a month',
+                'used management case?',
+                'multiple_form_types?'
+            ])
+
+            for domain_object in Domain.get_all():
+                domain_name = domain_object.name
+                user_ids = CommCareUser.ids_by_domain(domain=domain_name)
+                for hit in get_forms_for_users(domain_object.name, user_ids, start_date, end_date).hits:
```
calellowitz: this section would be more efficient if you used user aggregation on the es query like in the form users part of `active_mobile_users` in the esaccessors file

corehq/apps/data_analytics/management/commands/export_form_impact_data.py
```diff
@@ -0,0 +1,85 @@
...
+                    try:
+                        form = hit['form']
+                        user_id = form['meta']['userID']
+                        case = form.get('case')
+                        xmlns = form['@xmlns']
```
calellowitz: you can replace all these with `.get(key)` so that one `KeyError` wont cause you to lose the whole form submission. it would also make it easier to read since it would eliminate the try except block

corehq/apps/data_analytics/management/commands/export_form_impact_data.py
calellowitz: you can just parse the result of that and will not need to create the secondary `defaultdict` or do the large couch query at the end.

calellowitz (changes requested):

corehq/apps/data_analytics/management/commands/export_form_impact_data.py
kkrampa: Is it possible to get doc using aggregation? I thought it will give me pair (user_id, forms_count)

calellowitz: it can when combined with a `TopHitsAggregation` as in https://github.com/dimagi/commcare-hq/blob/master/corehq/apps/reports/analytics/esaccessors.py#L40 You can pass a size parameter to the TopHitsAggregation of 10000 or a similarly high number that would be greater than the number of forms for a user in a month.

kkrampa:
@calellowitz Sorry for late response but I was offline. I updated PR.

kkrampa:
@calellowitz bump

calellowitz (approved):

PR #15563, opened 2017-03-29
Remove app dashboard from v2

orangejenny:
v2 doesn't use template apps, doesn't need the apps dashboard. The places that redirect to the app dashboard in v1 (e.g., after you delete an app) will instead redirect to the main domain dashboard.

@biyeun / @NoahCarnahan

### DIFF
--- a/corehq/apps/dashboard/templates/dashboard/dashboard_new_user.html
+++ b/corehq/apps/dashboard/templates/dashboard/dashboard_new_user.html
@@ -12,83 +12,33 @@
 {% block dashboard-content %}
 {% initial_page_data 'templates' templates %}
 <div class="page-header">
-    {% if request|toggle_enabled:"APP_MANAGER_V2" %}
-        <h1>{% trans 'Would you like to work on an existing app or create a new one?' %}</h1>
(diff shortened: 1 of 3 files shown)


corehq/apps/dashboard/templates/dashboard/dashboard_new_user.html
```diff
@@ -12,83 +12,33 @@
...
 <div class="page-header">
-    {% if request|toggle_enabled:"APP_MANAGER_V2" %}
-        <h1>{% trans 'Would you like to work on an existing app or create a new one?' %}</h1>
-    {% else %}
-        <h1>{% trans 'What type of app would you like to build?' %}</h1>
-    {% endif %}
+    <h1>{% trans 'What type of app would you like to build?' %}</h1>
 </div>
 <div class="row">
-    {% if request|toggle_enabled:"APP_MANAGER_V2" %}
```
orangejenny: This diff came out weird; all I'm doing here is removing the v2 part of the `if` block.

NoahCarnahan (approved):

PR #15565, opened 2017-03-29
Add index to device_id and username

benrudolph:
There's a pretty brutal query that is happening on ICDS right now in the device log filters:

https://github.com/dimagi/commcare-hq/blob/master/corehq/apps/reports/filters/api.py#L252-L258

For now I'm just adding these indexes to see if that is enough to speed up the filter query to something reasonable. Otherwise if that still doesn't work, we can remove the distinct clause from the username filter and do a join with the user table instead.


emord:
Are you going to add these before the migration so that it doesn't lengthen the deploy?

benrudolph:
i didn't think it would take that long, i know we had 26GB of device logs, before, but we deleted all of them on NIC. we should only keep 3 months of logs around on other envs. i'll check and see the size. if it's in the 100s of thousands of rows i think we're fine to just do it on deploy

benrudolph:
NIC - 25,903,624
Softlayer - 654,954
Swiss - 37,064
Prod - 63,165,600

Pretty big... think i'll run it before hand. I don't think anticipate it being a big deal to run the index synchronously. I don't think that table is heavily used anyways.

emord (approved):

millerdev (approved):

emord:
Sorry if you've already run this index, but did you profile this vs using a multi column index with domain? I think most times this is run it'll be on one domain

benrudolph:
i didn't profile it with multi column index. i think that sounds better long term. however for icds it won't make a difference. guessing there will be a world where we have mulitple domains with many many users

benrudolph:
ah wait realized we already do use `index_together`. this won't help at all guessing the issue here is indeed the `distinct` portion of this clause. will run an explain on NIC

PR #15568, opened 2017-03-29
add management command to delete migrated cases by episode id

nickpell:
Also stops adding drtb-hiv-referral cases if the patient's treatment has already ended ([here](https://github.com/dimagi/commcare-hq/commit/482afdce1faf87e1f8beb2e1c04067442055dcf9)).

@proteusvacuum cc: @esoergel

### DIFF
--- a/custom/enikshay/nikshay_datamigration/management/commands/delete_migrated_cases.py
+++ b/custom/enikshay/nikshay_datamigration/management/commands/delete_migrated_cases.py
@@ -73,8 +73,29 @@ class Command(BaseCommand):
             'episode_case_id',
             nargs='+',
         )
+        parser.add_argument(
+            '--noinput',
+            action='store_true',
+            default=False,
+            help='For unit tests only. Skips user confirmation.'
+        )
 
     def handle(self, domain, episode_case_id, **options):
+        if not options.get('noinput'):
+            confirm = raw_input(
+                u"""
+                Are you sure you want to delete all these episodes, and their associated
+                person, occurrence, and drtb-hiv-referral cases?
+
+                Type DELETE and press Enter to confirm.
+                """
+            )
+            if confirm != "DELETE":
+                print("\n\t\tCancelled.")
+                return
+            else:
+                print("\n\t\tDeleting...")
+
         migrated_case_ids = _concatenate_list_of_lists([
             _get_related_case_ids(domain, episode_id) for episode_id in episode_case_id
         ])
--- a/custom/enikshay/nikshay_datamigration/management/commands/delete_migrated_cases.py
+++ b/custom/enikshay/nikshay_datamigration/management/commands/delete_migrated_cases.py
@@ -38,8 +38,14 @@ def _concatenate_list_of_lists(list_of_lists):
 
 
 def _get_related_case_ids(domain, episode_case_id):
+    case_accessor = CaseAccessors(domain)
+
+    episode_case = case_accessor.get_case(episode_case_id)
+    assert episode_case.dynamic_case_properties().get('migration_created_case') == 'true'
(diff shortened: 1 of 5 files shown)


custom/enikshay/nikshay_datamigration/management/commands/delete_migrated_cases.py
```diff
@@ -0,0 +1,75 @@
...
+from corehq.form_processor.interfaces.dbaccessors import CaseAccessors
+from custom.enikshay.case_utils import (
+    CASE_TYPE_DRTB_HIV_REFERRAL,
+    CASE_TYPE_EPISODE,
+    CASE_TYPE_OCCURRENCE,
+    CASE_TYPE_PERSON,
+)
+
+
+def _get_parent_case_by_type(domain, child_case_id, parent_case_type):
```
proteusvacuum: not super important here, but these methods already exist in the `case_utils` for enikshay

custom/enikshay/nikshay_datamigration/management/commands/delete_migrated_cases.py
```diff
@@ -0,0 +1,75 @@
...
+        [drtb_hiv_referral_case.case_id for drtb_hiv_referral_case in all_drtb_hiv_referral_cases],
+    ])
+
+
+class Command(BaseCommand):
+
+    def add_arguments(self, parser):
+        parser.add_argument('domain')
+        parser.add_argument(
+            'episode_case_id',
```
proteusvacuum: nit: might want to call this `episode_case_ids` since you can have multiple?

custom/enikshay/nikshay_datamigration/management/commands/delete_migrated_cases.py
```diff
@@ -0,0 +1,75 @@
...
+            nargs='+',
+        )
+
+    def handle(self, domain, episode_case_id, **options):
+        migrated_case_ids = _concatenate_list_of_lists([
```
proteusvacuum: doesn't this just delete all related cases to the `episode` you pass in? Seems like this could be really destructive... might be worth adding a check in here that the user knows what they are doing.

proteusvacuum (commented):
Looks fine. Bit confused where the "only migrated cases" part comes from? Do you have another script that does that?

custom/enikshay/nikshay_datamigration/management/commands/delete_migrated_cases.py
nickpell: added a check: https://github.com/dimagi/commcare-hq/pull/15568/commits/f98eff3151aa53b739ef3b3f440eb137e29f7d41. Also now fails if the episode wasn't created in a migration.

custom/enikshay/nikshay_datamigration/management/commands/delete_migrated_cases.py
nickpell: It looked like those methods only select open cases - I wanted to delete closed cases too.

custom/enikshay/nikshay_datamigration/management/commands/delete_migrated_cases.py
nickpell: updating using metavar so help still shows corrrectly:
```
usage: manage.py delete_migrated_cases [-h] [--version] [-v {0,1,2,3}]
                                       [--settings SETTINGS]
                                       [--pythonpath PYTHONPATH] [--traceback]
                                       [--no-color] [--noinput]
                                       domain episode_case_id
                                       [episode_case_id ...]
```

nickpell:
> Bit confused where the "only migrated cases" part comes from? Do you have another script that does that?

nickpell:
> Bit confused where the "only migrated cases" part comes from? Do you have another script that does that?

This is on a case by case basis - most of the time I'm using exports + Excel to figure that out.

PR #15569, opened 2017-03-29
use TestCase and remove instantiate_accounting()

nickpell:
also adds a missing test case and general code cleanup

@esoergel

### DIFF
--- a/corehq/apps/accounting/tests/test_stripe_payment.py
+++ b/corehq/apps/accounting/tests/test_stripe_payment.py
@@ -5,6 +5,7 @@ from django.test import TransactionTestCase
 from django.test.client import RequestFactory
 
 import stripe
+from stripe.resource import StripeObject
 
 from corehq.apps.accounting.models import PaymentRecord, PaymentMethod, BillingAccount
 from corehq.apps.accounting.payment_handlers import CreditStripePaymentHandler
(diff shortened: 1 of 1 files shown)


corehq/apps/accounting/tests/test_stripe_payment.py
```diff
@@ -1,22 +1,28 @@
...
 
-class TestCreditStripePaymentHandler(TransactionTestCase):
+class MockFailingStripeObject(object):
+
+    @property
+    def id(self):
+        raise Exception
+
+
+class TestCreditStripePaymentHandler(TestCase):
```
esoergel: Does this `setUp` need to happen for each test?  Could it go in a `setUpClass`?

esoergel: I wonder if `PaymentRecord.objects.all().delete()` is the only cleanup needed between tests.

esoergel (approved):

corehq/apps/accounting/tests/test_stripe_payment.py
nickpell: > Could it go in a setUpClass?

updated

> I wonder if PaymentRecord.objects.all().delete() is the only cleanup needed between tests.

I don't think this is needed since the default ```self.tearDown()``` should rollback any SQL objects created.

esoergel: awesome, thanks!

> I don't think this is needed since the default self.tearDown() should rollback any SQL objects created.

Fancy!  So that's anything created during the test, but leaving stuff created in `setUpClass` alone?

nickpell: yeah - the stuff created in setUpClass then gets rolled back in tearDownClass (new as of django 1.8)

PR #15571, opened 2017-03-30
Override getHash and stop FireFox double request bug

benrudolph:
this was a fun one! https://manage.dimagi.com/default.asp?250644

the stack exchange answer i linked to is a little misleading. it surmises that this error happens when you don't url encode your routes. however i don't think this is the case here. the error is coming from the fact that Chrome and FireFox return different values

### DIFF
--- a/corehq/apps/cloudcare/static/cloudcare/js/formplayer/app.js
+++ b/corehq/apps/cloudcare/static/cloudcare/js/formplayer/app.js
@@ -453,3 +453,35 @@ FormplayerFrontend.on('navigateHome', function() {
         FormplayerFrontend.navigate("/apps", { trigger: true });
     }
 });
+
+/**
+ * This is a hack to ensure that routing works properly on FireFox. Normally,
+ * location.href is supposed to return a url decoded string. However, FireFox's
+ * location.href returns a url encoded string. For example:
+ *
+ * Chrome:
+ * > location.hash
+ * > "#{"appId"%3A"db732ce1735229da84b451cbd7cfa7ac"}"
+ *
+ * FireFox:
+ * > location.hash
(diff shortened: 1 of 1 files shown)


corehq/apps/cloudcare/static/cloudcare/js/formplayer/app.js
```diff
@@ -453,3 +453,35 @@ FormplayerFrontend.on('navigateHome', function() {
...
+ * when the hash changes.
+ *
+ * Additional explanation here: http://stackoverflow.com/a/25849032/835696
+ *
+ * https://manage.dimagi.com/default.asp?250644
+ */
+_.extend(Backbone.History.prototype, {
+    getHash: function(window) {
+        var match = (window || this).location.href.match(/#(.*)$/);
+        return match ? decodeURI(match[1]) : '';
```
benrudolph: the change here is calling `decodeURI` on the string

proteusvacuum (approved):
Thanks!

corehq/apps/cloudcare/static/cloudcare/js/formplayer/app.js
```diff
@@ -453,3 +453,35 @@ FormplayerFrontend.on('navigateHome', function() {
...
     }
 });
+
+/**
+ * This is a hack to ensure that routing works properly on FireFox. Normally,
+ * location.href is supposed to return a url decoded string. However, FireFox's
+ * location.href returns a url encoded string. For example:
+ *
+ * Chrome:
+ * > location.hash
```
millerdev: above in this comment you talk about `location.href`, but here it's `location.hash`. That's confusing. Why the difference?

benrudolph: it was just a smaller output so it was an easier example. ill update it to `href` and then truncate the output

PR #15580, opened 2017-03-30
always take newest label for export schema items

snopoke:
http://manage.dimagi.com/default.asp?249342

buddy @mkangia

### DIFF
--- a/corehq/apps/export/models/new.py
+++ b/corehq/apps/export/models/new.py
@@ -183,6 +183,7 @@ class ExportItem(DocumentSchema):
     @classmethod
     def merge(cls, one, two):
         item = one
+        item.label = two.label  # always take the newest label
         item.last_occurrences = _merge_dicts(one.last_occurrences, two.last_occurrences, max)
         item.inferred = one.inferred or two.inferred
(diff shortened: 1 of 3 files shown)


corehq/apps/export/models/new.py
```diff
@@ -183,6 +183,7 @@ def create_from_question(cls, question, app_id, app_version, repeats):
     @classmethod
     def merge(cls, one, two):
         item = one
+        item.label = two.label  # always take the newest label
```
benrudolph: i don't think we can technically know if `one` or if `two` is the newest. this feels like we're getting lucky based on how we're passing the items into the merge function. it's not obvious to me though how we can smartly resolve this besides doing something really inefficient like looking up the last_occurrences. that said i don't think this is really that big of a deal and mostly just a bonus feature, so i'm fine with it if it works

benrudolph (approved):

corehq/apps/export/models/new.py
snopoke: We do pass in the current schema in first (at least for forms) so it will work but agree with you that it's not exactly clear that `two` should take priority over `one`

PR #15587, opened 2017-03-30
ICDS: Choose indicator language based on state

gcapalbo:
@proteusvacuum @emord

### DIFF
--- a/custom/icds/tasks.py
+++ b/custom/icds/tasks.py
@@ -5,6 +5,7 @@ from corehq.apps.locations.dbaccessors import (
     generate_user_ids_from_primary_location_ids,
     get_location_ids_with_location_type,
 )
+from corehq.apps.locations.models import SQLLocation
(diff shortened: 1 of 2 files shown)


custom/icds/tasks.py
```diff
@@ -23,9 +24,16 @@
 AWC_LOCATION_TYPE_CODE = 'awc'
 SUPERVISOR_LOCATION_TYPE_CODE = 'supervisor'
 
+ANDHRA_PRADESH_SITE_CODE = '28'
+MAHARASHTRA_SITE_CODE = ''
```
gcapalbo: we don't have this yet so for now it's ignored by being left blank. we'll just fill it in once we have it

emord (approved):
Any ideas on how long this is going to need to run?

gcapalbo:
@emord not sure, but the permanent fix of putting language code on the user case is being included with the next app rollout

PR #15589, opened 2017-03-30
Improvements to async ucr indicators

emord:
* adds date_queued to the row
* emails if something's been queued in celery for 24 hours (could probably come down)
* adds postgres model to the admin site
* adds a checkbox on the data source to enable
* removes the dependency on the pillow name

@snopoke @sravfeyn


corehq/apps/userreports/models.py
```diff
@@ -707,35 +708,32 @@ class AsyncIndicator(models.Model):
...
-    pillow = models.CharField(max_length=126, null=False)
     indicator_config_ids = ArrayField(
         models.CharField(max_length=126, null=True, blank=True),
         null=False
     )
     date_created = models.DateTimeField(auto_now_add=True, db_index=True)
+    date_queued = models.DateTimeField(null=True, db_index=True)
 
     class Meta(object):
-        unique_together = ('doc_id', 'pillow',)
```
emord: Removing a unique constraint seems a little sketchy but I think it's ok in this case as only my test domain has ever tried this feature. 

The reason I want to do this is that we can have multiple pillows add indicators for the same doc_id which should keep the queue smaller at the expense of making each task a little longer.

Also in the future I want to refactor these processors to base the cache on the doc currently being processed instead of having it in the table. This would allow N tables to have a really expensive operation (like `get_case_forms`) and only hit the db once per case change instead of N times.

corehq/apps/userreports/tasks.py
```diff
@@ -203,22 +212,27 @@ def queue_async_indicators():
...
-                if not lock.acquire(blocking=False):
-                    continue
+def _queue_indicator(redis_client, indicator):
+    lock_key = _get_indicator_queued_lock_key(indicator)
+    lock = redis_client.lock(lock_key, timeout=60 * 60 * 24)
+    if not lock.acquire(blocking=False):
+        return
 
-                save_document.delay(indicator.id, indicator.doc_id, indicator.pillow)
+    indicator.date_queued = datetime.utcnow()
```
emord: I initially had this update after queueing the document, but that doesn't work when `CELERY_ALWAYS_EAGER=True` (tests) because it will process the save (which deletes the indicator) and then inserts the indicator again

corehq/apps/userreports/ui/forms.py
```diff
@@ -160,6 +161,14 @@ class ConfigurableDataSourceEditForm(DocumentFormBase):
         label=_("Backend"),
         initial=UCR_SQL_BACKEND
     )
+    asynchronous = forms.BooleanField(
+        initial=False,
+        required=False,
+        label="",
+        widget=BootstrapCheckboxInput(
+            inline_label="Asynchronous processing"
```
emord: didn't translate because this is a superuser feature

sravfeyn:
Nice! Looks good to me, will let Simon review

corehq/apps/userreports/models.py
snopoke: NIce, so you'd pass in some caching context to the UCR engine which would be common among all the tables processed for a particular doc.

corehq/apps/userreports/tasks.py
```diff
@@ -194,6 +195,14 @@ def queue_async_indicators():
     cutoff = start + ASYNC_INDICATOR_QUEUE_TIME
     time_for_crit_section = ASYNC_INDICATOR_QUEUE_TIME.seconds - 10
 
+    yesterday = datetime.utcnow() - timedelta(days=1)
+    old_indicators = AsyncIndicator.objects.filter(date_queued__lt=yesterday).count()
+    if old_indicators:
+        notify_exception(
+            None,
+            message="{} indicators have been queued but did not complete in the past day".format(old_indicators)
+        )
```
snopoke: could we not just make this a metric in datadog like:
```
oldest_indicator_date = AsyncIndicator.objects.aggregate(max_date=Max('date_queued'))['max_date']
lag = (datetime.utcnow() - oldest_indicator_date).total_seconds
datadog_guage('metric_name', lag)
```

snopoke (commented):
Just one commend about metric vs email

corehq/apps/userreports/tasks.py
emord: yeah that's way better gotta switch my thinking to using datadog first https://github.com/dimagi/commcare-hq/pull/15589/commits/4d928a55e586c752541c07c5f8b6547be44a9630

corehq/apps/userreports/tasks.py
```diff
@@ -1,5 +1,5 @@
 from __future__ import absolute_import
-from datetime import datetime
+from datetime import datetime, timedelta
```

corehq/apps/userreports/models.py
emord: Yeah. The downside is that it makes it harder to reason about which data sources are slow, but i'll probably just randomize the order they are processed so that there are some logs on the slowness

snopoke (approved):

PR #15590, opened 2017-03-30
simplify tests + accounting setup

nickpell:
recommend 🐡 
@esoergel


corehq/apps/accounting/tests/generator.py
```diff
@@ -36,7 +38,8 @@
 
 
 @unit_testing_only
-def instantiate_accounting():
+@nottest
```
esoergel: dammit, nose

esoergel (approved):
:skull:

PR #15591, opened 2017-03-30
Change test output to be copypastable test name

esoergel:
@millerdev
This builds off your "uniformresult" plugin, which already does copypastable names for everything except function tests.  This adds it in for function tests too, and defaults that plugin to enabled.  Does that sound reasonable?

Regarding the names for function tests, `test_compression` is a generator test, and `test_no_overlap` is a regular function test.  What do you think of this format?
```
custom.enikshay.tests.test_id_compression:test_compression args (11, 'T', 0)
custom.enikshay.tests.test_id_compression:test_compression args (12, 'LA', 0)
custom.enikshay.tests.test_id_compression:test_compression args (119, 'VT', 0)
custom.enikshay.tests.test_id_compression:test_no_overlap args ()
```
I'm not sure whether args are ever relevant for regular function tests.  For generator tests, I don't believe you can run just a single case, but you do want the info on the args in order to know which case failed, so I put them both in there, but separated for convenience.

### DIFF
--- a/corehq/tests/noseplugins/uniformresult.py
+++ b/corehq/tests/noseplugins/uniformresult.py
@@ -27,7 +27,7 @@ def uniform_description(test):
         return "%s:%s" % (test.__module__, test.__name__)
     if isinstance(test, FunctionTestCase):
         descriptor = test.descriptor or test.test
-        return "%s:%s args %s" % (
+        return "%s:%s %s" % (
             descriptor.__module__,
             descriptor.__name__,
             test.arg,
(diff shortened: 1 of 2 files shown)


corehq/tests/noseplugins/uniformresult.py
```diff
@@ -26,7 +26,12 @@ def uniform_description(test):
     if isinstance(test, type):
         return "%s:%s" % (test.__module__, test.__name__)
     if isinstance(test, FunctionTestCase):
-        return str(test)
+        descriptor = test.descriptor or test.test
+        return "%s:%s args %s" % (
```
millerdev: Nitpick: I'd change this to `return "%s:%s %s" % (` – shorter is better, and there's not much ambiguity when the args tuple is surrounded by parens anyway.

millerdev (approved):
Excellent!

corehq/tests/noseplugins/uniformresult.py
esoergel: Fair point.  Updated.

PR #15594, opened 2017-03-30
Don't fail hard on InvalidFormatException

gcapalbo:
http://manage.dimagi.com/default.asp?247396
@proteusvacuum


snopoke (approved):
Build failure doesn't seem related. :boot:

PR #15595, opened 2017-03-30
move agg task to background queue

emord:
@snopoke @benrudolph 

i think this should be on the background queue instead of the main one

fyi @sheelio @sravfeyn


benrudolph (approved):
👍 background queue is getting a lot of added tasks, we should just be careful not to overwhelm it

emord:
Happy to change it to another one if we think there's a better queue. It takes about 15 min at midnight UTC

PR #15603, opened 2017-03-31
Use `user_es_query` in Submit History

benrudolph:
@snopoke this optimizes the submit history report. given the complexity of this portion it's hard to tell if this is the _exact_ same functionality, however if we believe that `user_es_query` gets all relevant user ids (which is an assumption we make elsewhere), then this should be the same functionality. ran a little profiling as well:

```
In [7]: timeit.timeit('from corehq.apps.reports.filters.users import LocationRestrictedMobileWorkerFilter as EMWF; EMWF.pull_users_and_groups("aspace", ["t__0","t__4"])', number=100)
Out[7]: 8.12283182144165
```

```
In [9]: timeit.timeit('from corehq.apps.reports.filters.users import LocationRestrictedMobileWorkerFilter as EMWF; EMWF.user_es_query("aspace", ["t__0","t__4"]).values_list("_id")', number=100)
Out[9]: 0.4410839080810547
```

This is on my local domain which doesn't even have 100 users. `pull_users_and_groups` is a _horrible_ function that should never ever be called. it uses the dreaded `get_all_users_by_domain` not once, not twice, but 3 times (depending on what's passed in). in addition the submit history make _another_ call to `get_all_users_by_domain`. so in sum it can be called up to 4 times during a call to submit history, and the function is not cached. this PR removes all those calls

to test this i ran some basic queries on this branch vs. master branch

cc: @millerdev

### DIFF
--- a/corehq/apps/reports/standard/inspect.py
+++ b/corehq/apps/reports/standard/inspect.py
@@ -62,28 +62,8 @@ class SubmitHistoryMixin(ElasticProjectInspectionReport,
         return datespan_from_beginning(self.domain_object, self.timezone)
 
     def _get_users_filter(self, mobile_user_and_group_slugs):
-        truthy_only = functools.partial(filter, None)
-        users_data = EMWF.pull_users_and_groups(
-            self.domain,
-            mobile_user_and_group_slugs,
-            include_inactive=True
-        )
-        selected_user_types = EMWF.selected_user_types(mobile_user_and_group_slugs)
-        all_mobile_workers_selected = HQUserType.REGISTERED in selected_user_types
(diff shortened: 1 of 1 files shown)


benrudolph:
(label because i want to do a little more testing)

snopoke (commented):
:heart:

millerdev (approved):
Wow, nice rewrite!

benrudolph:
Ok i'm happy with this. I've single selectors (user types, users, groups). I've tested unioning of selectors (groups + user types, groups + single users, user types + single users). From what i can tell the functionality is the same

snopoke:
🚀

NoahCarnahan:
🏎

PR #15604, opened 2017-03-31
allow all users to access App API

snopoke:
http://manage.dimagi.com/default.asp?250786


mkangia (approved):
It would be nice to check if there are some redundant test cases now that can be removed if not checked already.

snopoke:
I don't think the permissions are testsed.

PR #15606, opened 2017-03-31
add @method_decorator to toggle decorator in class-based views

dannyroberts:
this works totally by chance if the toggle is enabled, which is why we
haven't noticed this until now.

Basically in the decorator inner function the first param is called
`request`, but what's passed (without method_decorator) is the
class-based view object, which also happens to respond to `.domain`.
So everything's fine through `toggle.enabled(request.domain)`, but if
that returns `False`, then it fails later on `request.user`, since the
CBV object doesn't respond to `.user`.


dannyroberts:
cc: @kaapstorm buddy

emord (approved):

PR #15607, opened 2017-03-31
share eval context with doc

emord:
as mentioned here: https://github.com/dimagi/commcare-hq/pull/15589#discussion_r109036212

When a change comes in it'll create an evaluation context which is what contains the cache. it passes it into any more data sources that process the same docs.

@snopoke @sravfeyn


corehq/apps/userreports/tests/test_pillow.py
```diff
@@ -254,6 +258,75 @@ def test_process_doc_from_sql_stale(self):
...
+        with self.assertNumQueries(15):
+            pillow.process_changes(since=since, forever=False)
+
+    def test_reuse_cache(self):
+        since1 = self.pillow1.get_change_feed().get_latest_offsets()
+        since2 = self.pillow2.get_change_feed().get_latest_offsets()
+        form, cases = self._post_case_blocks()
+
+        self._test_pillow(self.pillow1, since1)
+        self._test_pillow(self.pillow2, since2)
```
snopoke: Could do with a comment explaining how this test works - pillow2 is processing 2 configs but still makes the same number of queries as pillow1 which on has 1 config.

corehq/apps/userreports/tests/test_pillow.py
```diff
@@ -254,6 +258,75 @@ def test_process_doc_from_sql_stale(self):
...
+                    case_name='child-name',
+                    case_type='bug-child',
+                    index={'parent': ('bug', 'parent-id')},
+                    update={'update-prop-child': iteration}
+                ).as_xml()
+            ], domain=self.domain
+        )
+
+    def _test_pillow(self, pillow, since):
+        with self.assertNumQueries(15):
```
snopoke: This seems like a lot of queries. Running it locally I only got 5 per case:
```
SELECT * from get_case_by_id('child-id')
SELECT * FROM get_case_indices('bug-domain', 'child-id')
SELECT * from get_case_transactions('child-id')
SELECT * from get_case_attachments('child-id')
SELECT form_id FROM get_case_transactions_by_type('child-id', 1)
```

snopoke: Reduced this to 4 per case here: https://github.com/dimagi/commcare-hq/pull/15611

emord: Agreed it's a lot. the data source tests using related doc to get the parent case from the child case, so you have 3 case gets.

1) get the parent case and it is not processed by the UCR
2) get the child case, it is processed by UCR
3) while processing it gets the parent case

corehq/apps/userreports/tests/test_pillow.py
emord: https://github.com/dimagi/commcare-hq/pull/15607/commits/726bdd01845a8094a3c9a2a4503f07d6303b91b5#diff-67129b077cd07e96bf8f4f4ffdd44b55R315

corehq/apps/userreports/tests/test_pillow.py
```diff
@@ -254,6 +258,76 @@ def test_process_doc_from_sql_stale(self):
...
+                    case_name='child-name',
+                    case_type='bug-child',
+                    index={'parent': ('bug', 'parent-id')},
+                    update={'update-prop-child': iteration}
+                ).as_xml()
+            ], domain=self.domain
+        )
+
+    def _test_pillow(self, pillow, since):
+        with self.assertNumQueries(12):
```
snopoke: Odd, this still fails for me locally with only 8 queries.

snopoke: Figured it out - I'd created an extra partition on the `case-sql` topic so this test was only getting one of the cases.

PR #15512, opened 2017-03-24
Release vellum onboarding changes

orangejenny:
Deliberately doing this in a brain-dead way just to get the release out; will remove the feature flag, the `vellum_beta` branch, etc. once this has been in production for a week or two.

@millerdev / @NoahCarnahan

PR #15608, opened 2017-03-31
Vellum alpha flag

orangejenny:
Since https://github.com/dimagi/commcare-hq/pull/15512 there hasn't been a way to turn off the new vellum, which I'd like to be able to do for debugging purposes.

A little bit confusing, but the plan is for this flag (along with vellum_beta) to be short-lived, just for the next week or two, until old vellum is fully removed from the codebase.

@NoahCarnahan / @millerdev

### DIFF
--- a/corehq/apps/app_manager/views/formdesigner.py
+++ b/corehq/apps/app_manager/views/formdesigner.py
@@ -136,7 +136,7 @@ def form_designer(request, domain, app_id, module_id=None, form_id=None):
 
     vellum_base = 'corehq/apps/app_manager/static/app_manager/js/'
     vellum_dir = 'vellum'
-    if isdir(join(vellum_base, 'vellum_beta')):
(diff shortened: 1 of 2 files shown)


millerdev (approved):
I'm a little confused about what is the difference between vellum_alpha and vellum_beta. Assuming neither of those are referring to the "old version" of vellum, but the message on this PR lead me to believe this was adding a way to switch back to "old vellum" temporarily for debugging. I'm trusting that you know what you're doing and that I'm just being dense.

orangejenny:
@millerdev 

I'm planning to keep some parts of the `vellum_beta` code (the parts that will be useful for future soft rollouts) and remove the rest of it, but I didn't have time to do that when it was time to release, so instead I just added some `False and` conditions in https://github.com/dimagi/commcare-hq/pull/15512 Now toggling the `vellum_beta` flag does nothing.

`vellum_alpha` does actually control which version of vellum the user gets.

Going to merge this, but let me know if it seems too sloppy, even for temporary code, and I can take another pass at it.

PR #15609, opened 2017-03-31
Get rid of 'active_tab' workaround

dannyroberts:
There were a bunch of places that for example `ProjectSettingsTab` was being put into the view context as 'active_tab' as a hack to get the sidebar to render. At some point this became necessary because`ProjectSettingsTab` and `MySettingsTab` were removed from the list of tabs, because they were moved into a menu rather than a top-level tab in the UI. Instead, I'm putting them back in the list of tabs, but just setting `_is_viewable` to `False`, so they'll never show up, but will be automatically set as the `active_tab`.


dannyroberts:
cc: @kaapstorm

kaapstorm (approved):

kaapstorm:
(I found this [template tag](https://github.com/dimagi/commcare-hq/blob/2455f6851de0726a6e8d775e318ef1d53cacabe7/corehq/tabs/templatetags/menu_tags.py#L101-L101) useful for context when reviewing.)

PR #15610, opened 2017-03-31
Ability to expire blobs

benrudolph:
@millerdev went over this plan with the scale team and everyone is onboard


benrudolph:
fyi: @NoahCarnahan this is another piece for making it possible to do large paginated export. i think the plan will be to temporarily store the export in blob storage. also will be useful for expiring images on icds

corehq/blobs/models.py
```diff
@@ -4,3 +4,17 @@
...
+    This models records when temporary blobs should be deleted
+    '''
+    bucket = models.CharField(max_length=255)
+    identifier = models.CharField(max_length=255)
+    expires_on = models.DateTimeField(db_index=True)
+    created_on = models.DateTimeField(auto_now=True)
+    length = models.IntegerField()
+
+    # This is set to True when the blob associated has been successfully deleted
+    deleted = models.BooleanField(default=False, db_index=True)
```
emord: Do these get cleaned up? what's the benefit of keeping these around after they're deleted?

corehq/blobs/tasks.py
```diff
@@ -0,0 +1,27 @@
...
+@transaction.atomic
+def delete_expired_blobs():
+    blob_expirations = BlobExpiration.objects.filter(expires_on__lt=_utcnow(), deleted=False)
+
+    db = get_blob_db()
+    paths = []
+    for blob_expiration in blob_expirations:
+        paths.append(db.get_path(blob_expiration.identifier, blob_expiration.bucket))
+        blob_expiration.deleted = True
+        blob_expiration.save()
```
emord: I think you could do a bulk update instead of saving in the loop

corehq/blobs/models.py
benrudolph: so this is for a couple things, 1) if we ever want to be able to clean up these records it'll be easy 2) there's an edge case that someone could reuse a `bucket`, `identifier` pair and i want to make sure not to grab the old blobexpiration record

corehq/blobs/tasks.py
benrudolph: i had thought the django orm magically created this into one database query when using `atomic`. however i think `bulk_update` is probably more intuitive / less magical. will update

corehq/blobs/models.py
benrudolph: i'll make a test case for this so it's obvious why we need it

benrudolph: thinking about it more, we may want to actually update the blob expiration record when the same bucket/identifier is used rather than create a new one

millerdev: I would hope it's pretty rare for a key to be reused since the edge cases around object existence/versioning in Riak can get a bit gnarly when reusing keys for different objects, depending on timing and what you're doing. If the object content is the same then it's not as much of an issue.

corehq/blobs/models.py
```diff
@@ -4,3 +4,17 @@
...
+
+class BlobExpiration(models.Model):
+    '''
+    This models records when temporary blobs should be deleted
+    '''
+    bucket = models.CharField(max_length=255)
+    identifier = models.CharField(max_length=255)
+    expires_on = models.DateTimeField(db_index=True)
+    created_on = models.DateTimeField(auto_now=True)
+    length = models.IntegerField()
```
millerdev: What is this field used for? Seems like it isn't needed/doesn't belong here.

millerdev (commented):
cool feature!

millerdev:
Test failure looks unrelated:

```
FAIL: casexml.apps.phone.tests.test_cleanliness:OwnerCleanlinessTest.test_hint_invalidation
  File "corehq/ex-submodules/casexml/apps/phone/tests/test_cleanliness.py", line 197, in test_hint_invalidation
    self._set_owner(self.parent.case_id, self.owner_id)
  File "corehq/ex-submodules/casexml/apps/phone/tests/test_cleanliness.py", line 104, in _set_owner
    self.assertEqual(owner_id, case.owner_id)
AssertionError: 'e8a59b8cf76b4ac295ff101e9c9d016c' != u'e6918c4c603b4b45bc9cbd132ee2ba66'
corehq.form_processor.tests.utils: DEBUG: Deleting all Couch cases for domain None
corehq.form_processor.tests.utils: DEBUG: Deleting all SQL cases for domain None
corehq.form_processor.tests.utils: DEBUG: Deleting all Couch xforms for domain None
corehq.form_processor.tests.utils: DEBUG: Deleting all SQL xforms for domain None
corehq.form_processor.tests.utils: DEBUG: Deleting all synclogs
```

corehq/blobs/models.py
benrudolph: this was discussed in the scale meeting as a suggestion. it's way to easily keep track of how much temp storage we have (it's been an issue on icds recently)

corehq/blobs/models.py
benrudolph: that's interesting, i did not know that. i would also hope it's rare, but figured if it's possible, then it'll happen. given that information, do you think it's worth disabling it for identical bucket / identifiers?

corehq/blobs/models.py
millerdev: ah, that certainly could be useful 👍

Incidentally, short of looking at used disk space, I do not know of a way to query the size of all objects stored in Riak, not to mention subsets of that data.

corehq/blobs/models.py
benrudolph: https://github.com/dimagi/commcare-hq/pull/15610/commits/9926d613222b87a522e037906e3e0c7fa45e3d8a the fsdb blocks blobs from being updated if the path already exists. so i doubt we'll ever have to worry about this at this level. however i made the change just to future proof it in case we ever want to allow blobs to be updated and because it wasn't that difficult to do

benrudolph (commented):
this is good to go from my end

NoahCarnahan:
👍 looks good!

corehq/blobs/models.py
snopoke: I don't think this discussion answers the question of why we keep these records around rather than just deleting them. I don't see a benefit to having them.

millerdev: > the fsdb blocks blobs from being updated if the path already exists

this [just changed today](https://github.com/dimagi/commcare-hq/pull/19936), btw. s3db never prevented that because it couldn't be done efficiently.

> I don't think this discussion answers the question of why we keep these records around rather than just deleting them. I don't see a benefit to having them.

I agree it doesn't really answer that question, but I can see some benefit from a historical record/logging perspective. Answering the question: why was this blob deleted? However, it comes up pretty rarely and I've never actually looked here to answer that question (silly mistake on my part, nothing more). I think ideally I'd like to delete them after 30 days or something, but that requires more machinery and code to maintain, so maybe not worth the extra effort over just deleting the record immediately? Maybe we should simply log that the blob was deleted (along with it's `BlobExpiration` object) and let log rotation take care of the rest?

millerdev: @snopoke I've added a task to my todo list. I'll delete them and add some logging.

PR #15612, opened 2017-04-02
Add meta settings to prod

benrudolph:
@emord made this separately so on my other shard PR it's very obvious that we aren't accidentally changing any of the index settings that we don't want. realize this only tests the prod settings, but i think that's ok tradeoff given how annoying this test usually ends up being (would rather not write out the indexes for all the envs)

cc: @millerdev


emord:
> would rather not write out the indexes for all the envs

You could do something like have a default and then merge the changes for prod though. Then you could have a test for all envs

PR #15622, opened 2017-04-03
Ability to distinguish when group has no nested questions

benrudolph:
@wpride this is getting increasingly complex. this ensures that if the group doesn't have any nested questions that we don't show the group header: https://manage.dimagi.com/default.asp?251043

cc: @millerdev


wpride:
@benrudolph would it be cleaner to solve this on the formplayer side (IE not send down empty groups?)

benrudolph:
if that's easy than definitely, not sure if we'd be overlooking anything. that would happen in `Walker.java` right?

benrudolph:
@wpride i think  we should actually just get this in. i think we have our hands full optimizing formplayer and this seems like a lower priority thing to worry about

PR #15633, opened 2017-04-04
save riak calls icds ucr

emord:
@NoahCarnahan not sure if you might be able to use this for some enikshay UCRs

@sheelio @sravfeyn


NoahCarnahan (approved):
nice

PR #15635, opened 2017-04-04
Add case list menu item back into app manager v2

orangejenny:
Would be great to integrate this better into app manager, but doing so isn't a priority for getting users onboarded, so for now it's just going to be one of the "advanced" app manager features.

@biyeun / @esoergel

### DIFF
--- a/corehq/apps/app_manager/templates/app_manager/v2/partials/module_view_settings.html
+++ b/corehq/apps/app_manager/templates/app_manager/v2/partials/module_view_settings.html
@@ -126,6 +126,8 @@
             </div>
             <div class="panel-body">
               {% include 'app_manager/v2/partials/case_list_form_setting.html' %}
+
+              {% include 'app_manager/v1/partials/case_list_setting.html' with LABEL="Case List Menu Item" DESC="An item in the module's menu that lets you browse the case list without moving on to fill out a form." SLUG="case_list" case_list=module.case_list %}
             </div>
           </div>
           {% endif %}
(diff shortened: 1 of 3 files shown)


corehq/apps/app_manager/templates/app_manager/v2/partials/module_view_settings.html
```diff
@@ -126,6 +126,8 @@ <h4 class="panel-title panel-title-nolink">{% trans "Advanced Settings" %}</h4>
             </div>
             <div class="panel-body">
               {% include 'app_manager/v2/partials/case_list_form_setting.html' %}
+
+              {% include 'app_manager/v1/partials/case_list_setting.html' with LABEL="Case List Menu Item" DESC="An item in the module's menu that lets you browse the case list without moving on to fill out a form." SLUG="case_list" case_list=module.case_list %}
```
esoergel: should this be `v1` or `v2`?

biyeun: yeah this should be a v2 template

biyeun: seems more verbose right now, but will make people more aware of making changes and make things easier when we move over to v2

orangejenny: makes sense, updated

biyeun (approved):

PR #15639, opened 2017-04-04
Put existing reg from case list logic back

orangejenny:
The original v2 design includes making registration from the case list the default setup for new apps. Onboarding team discussed and decided to punt on this for now - it's a better experience, but it isn't enough of an issue to block the v2 release.

@biyeun / @esoergel


biyeun:
@orangejenny test failure looks like it might be real? bulk ui translations

orangejenny:
tests passing now

orangejenny:
ping, this should be quick to review

PR #15643, opened 2017-04-05
Use latest available file for translations test

orangejenny:
see https://github.com/dimagi/commcare-translations/pull/41

@dannyroberts / @esoergel


orangejenny:
https://github.com/dimagi/commcare-translations/pull/41 is merged and the submodule ref is updated; this is good to go

PR #15645, opened 2017-04-05
Fix migration crash on blobs != external_blobs

millerdev:
This state should not exist, but there is at least one app on staging that has it, and I didn't see another way to get around it. Fixing that app on staging doesn't seem worth the trouble. The owner even granted permission to delete it, although that's not trivial either.

@NoahCarnahan

### DIFF
--- a/corehq/blobs/migrate.py
+++ b/corehq/blobs/migrate.py
@@ -267,6 +267,7 @@ class BlobDbBackendMigrator(BaseDocMigrator):
         self.db = get_blob_db()
         self.total_blobs = 0
         self.not_found = 0
+        self.bad_blobs_state = 0
         if not isinstance(self.db, MigratingBlobDB):
             raise MigrationError(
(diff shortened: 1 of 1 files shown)


corehq/blobs/migrate.py
```diff
@@ -277,8 +278,16 @@ def _backup_doc(self, doc):
...
         obj = self.blob_helper(doc, self.couchdb)
         bucket = obj._blobdb_bucket()
-        assert obj.external_blobs and obj.external_blobs == obj.blobs, doc
-        for name, meta in obj.blobs.iteritems():
+        if obj.external_blobs != obj.blobs:
+            self.bad_blobs_state += 1
+            super(BlobDbBackendMigrator, self)._backup_doc({
+                "doc_type": obj.doc_type,
+                "doc_id": obj._id,
+                "blob_identifier": meta.id,
```
dimagimon: F821 undefined name 'meta'

PR #15646, opened 2017-04-05
Optimize Forms by Submission

benrudolph:
This should help out the forms by submission report quite a bit for ICDS. Just removing `pull_users_and_groups` is always a big win. This still does do some pretty expensive queries for each user though (getting the form counts for each user). Hard to paginate because we need to sort by the forms submitted.

@proteusvacuum @dannyroberts

### DIFF
--- a/corehq/apps/reports/standard/monitoring.py
+++ b/corehq/apps/reports/standard/monitoring.py
@@ -684,12 +684,12 @@ class SubmissionsByFormReport(WorkerMonitoringFormReportTableBase,
 
     @property
     @memoized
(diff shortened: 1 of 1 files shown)


corehq/apps/reports/standard/monitoring.py
```diff
@@ -7,6 +7,8 @@
 from corehq.util.quickcache import quickcache
 import pytz
 
+from dimagi.utils.couch.database import iter_docs
```
dimagimon: F401 'iter_docs' imported but unused

PR #15647, opened 2017-04-05
UCR improvements for icds

emord:
Chipping away at some UCR stuff

Now caches named expressions (only for the current iteration). Also changes the ES query in ICDS to not depend on the date or xmlns arguments. Results in 8 or so less queries

@NoahCarnahan @sheelio @sravfeyn

### DIFF
--- a/corehq/apps/userreports/expressions/specs.py
+++ b/corehq/apps/userreports/expressions/specs.py
@@ -84,11 +84,12 @@ class NamedExpressionSpec(JsonObject):
 
     def __call__(self, item, context=None):
         key = self._context_cache_key()
-        if context.exists_in_cache(key):
+        if context and context.exists_in_cache(key):
             return context.get_cache_value(key)
 
         result = self._context.named_expressions[self.name](item, context)
-        context.set_iteration_cache_value(key, result)
+        if context:
+            context.set_iteration_cache_value(key, result)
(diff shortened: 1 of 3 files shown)


emord:
forgot new code buddies @kaapstorm

corehq/apps/userreports/expressions/specs.py
```diff
@@ -79,8 +79,18 @@ def configure(self, context):
             raise BadSpecError(u'Name {} not found in list of named expressions!'.format(self.name))
         self._context = context
 
+    def _context_cache_key(self):
+        return 'named_expression-{}'.format(self.name)
```
NoahCarnahan: It looks like this key would need more information in it, but maybe not because this uses the iteration cache? I'm not super familiar with how named expressions work, but what would happen if you used a named expression as some sort of sub-expression. Would the "item" be different?

emord: From my understanding named expressions are either the root_doc or the item from `base_item_expression`. and anything inside name expressions will be consistent throughout the doc/iteration as they don't take arguments.

With that said it does feel a little hacky to rely on a second cache, but item can be an arbitrary dictionary which could include something that is not json serializeable (like datetimes) which makes making this vary on item difficult

PR #15648, opened 2017-04-05
ucr es settings

emord:
Enables specifying index settings per UCR. This should cut down on any shards on prod @benrudolph 

@kaapstorm


benrudolph (approved):
nice that this is configurable. looks good to me. we may want to think about using 6 shards instead? i doubt it'll make much of a difference and also depends on how many ES machines we have, but 6 splits evenly across two machines and three machines

emord:
@benrudolph the default was 5 shards, and requires a rebuild to change so if we do want to do that we should do one at a time

benrudolph:
👍 right

PR #15649, opened 2017-04-05
Fix broken custom location safe reports (eNikshay)

NoahCarnahan:
The custom enikshay reports weren't fully location safe. They restricted users from selecting locations outside of their hierarchy, but they didn't filter the data at all if no location was selected to filter by. This change should ensure that if a user is location restricted, the user's locations will be added to the filter.
@esoergel cc @millerdev

### DIFF
--- a/custom/enikshay/reports/filters.py
+++ b/custom/enikshay/reports/filters.py
@@ -47,6 +47,8 @@ class EnikshayLocationFilter(BaseMultipleOptionFilter):
     @classmethod
     def get_value(cls, request, domain):
         selected = super(EnikshayLocationFilter, cls).get_value(request, domain)
+        if not request.couch_user.has_permission(domain, 'access_all_locations'):
+            selected += request.couch_user.get_location_ids(domain)
         choice_provider = LocationChoiceProvider(StubReport(domain=domain), None)
         choice_provider.configure({'include_descendants': True})
         return [
--- a/custom/enikshay/reports/filters.py
+++ b/custom/enikshay/reports/filters.py
@@ -47,14 +47,18 @@ class EnikshayLocationFilter(BaseMultipleOptionFilter):
     @classmethod
     def get_value(cls, request, domain):
         selected = super(EnikshayLocationFilter, cls).get_value(request, domain)
-        if not request.couch_user.has_permission(domain, 'access_all_locations'):
-            selected += request.couch_user.get_location_ids(domain)
         choice_provider = LocationChoiceProvider(StubReport(domain=domain), None)
         choice_provider.configure({'include_descendants': True})
-        return [
+        selected_locations = [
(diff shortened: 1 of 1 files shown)


custom/enikshay/reports/filters.py
```diff
@@ -47,6 +47,8 @@ def selected(self):
     @classmethod
     def get_value(cls, request, domain):
         selected = super(EnikshayLocationFilter, cls).get_value(request, domain)
+        if not request.couch_user.has_permission(domain, 'access_all_locations'):
+            selected += request.couch_user.get_location_ids(domain)
```
esoergel: Won't this sometimes include more locations than are selected?  If I'm assigned to Suffolk and I select "Boston", I'd expect to see only results in Boston, but I think with this I'd see results in both Boston AND Suffolk.  I think the user's location ids should only be added if they haven't selected anything, and if they have selected something, we should make sure the selections are accessible (in case they modified the request params or the choice provider was broken).

NoahCarnahan: Or should this be `couch_user.get_location_id()`? (singular)

esoergel: Multiple is correct.

NoahCarnahan: Ah yes, should only be added if selected is empty

NoahCarnahan: Fixed here: https://github.com/dimagi/commcare-hq/pull/15649/commits/cec4d5d1ab27ca111d9b51fa0316933573799630.

`LocationChoiceProvider.get_choices_for_known_values()` already makes sure that selected locations are accessible. Also, I moved this check to *after* `get_choices_for_known_values()`, because `get_choices_for_known_values()` could return an empty list `selected` contains only non-accessible locations.

custom/enikshay/reports/filters.py
```diff
@@ -47,14 +47,18 @@ def selected(self):
...
-        if not request.couch_user.has_permission(domain, 'access_all_locations'):
-            selected += request.couch_user.get_location_ids(domain)
         choice_provider = LocationChoiceProvider(StubReport(domain=domain), None)
         choice_provider.configure({'include_descendants': True})
-        return [
+        selected_locations = [
             choice.value
             for choice in choice_provider.get_choices_for_known_values(selected, request.couch_user)
         ]
+        if len(selected_locations) == 0 and not request.couch_user.has_permission(domain, 'access_all_locations'):
```
millerdev: nitpick: the idiomatic way to test for list emptiness according to PEP8 is `if not selected_locations and ...`

millerdev (approved):
Looks OK to me, but will let Ethan review as well since he has more context.

esoergel (approved):

PR #15650, opened 2017-04-05
ICDS SMS: Add Telugu Translations

gcapalbo:
@emord 
buddy @mkangia


emord (commented):
Do you need to copy over the aggregate performance ones, or will they default to hindi if the file doesn't exist?

gcapalbo:
yeah we're still waiting on those so i didn't upload them, but for now they will just default to hindi

emord (approved):

PR #15653, opened 2017-04-05
Export download to use blobdb

benrudolph:
@millerdev @NoahCarnahan this converts the export code to use the blobdb instead of redis. this has the huge added benefit of not having to load the entire file into memory. Decided to put this behind a flag before rollout just in case there are some quirks with production riak that I didn't foresee

buddies @dannyroberts @proteusvacuum

### DIFF
--- a/corehq/apps/export/tasks.py
+++ b/corehq/apps/export/tasks.py
@@ -8,8 +8,10 @@ from corehq.apps.export.system_properties import MAIN_CASE_TABLE_PROPERTIES
 from corehq.util.decorators import serial_task
 from corehq.util.files import safe_filename_header
 from corehq.util.quickcache import quickcache
+from corehq.blobs import get_blob_db
+from corehq.toggles import BLOBDB_EXPORTS
 from couchexport.models import Format
-from soil.util import expose_cached_download
+from soil.util import expose_cached_download, expose_blob_download
 
 logger = logging.getLogger('export_migration')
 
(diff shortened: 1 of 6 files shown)


corehq/apps/export/tasks.py
```diff
@@ -26,16 +28,32 @@ def populate_export_download_task(export_instances, filters, download_id, filena
...
+            db.put(open(export_file.path, 'r'), download_id, timeout=expiry)
+
+            expose_blob_download(
+                download_id,
+                mimetype=file_format.mimetype,
+                content_disposition=safe_filename_header(filename, file_format.extension),
+                download_id=download_id,
+            )
+        else:
+            payload = export_file.file.payload
```
benrudolph: calling "`.payload`" loads the entire payload into memory

corehq/ex-submodules/soil/__init__.py
```diff
@@ -257,3 +258,60 @@ def create(cls, path, **kwargs):
...
+        file_obj = None
+        try:
+            file_obj = self.blobdb.get(self.identifier, self.bucket).read()
+            return file_obj.read()
+        finally:
+            if file_obj is not None:
+                file_obj.close()
+
+    def toHttpResponse(self):
+        file_obj = self.blobdb.get(self.identifier, self.bucket)
```
benrudolph: the `FileDownload` doesn't worry about closing the file and I'm not sure why yet or if it was just an oversight. much of this api is to conform to the `DownloadBase` object

corehq/apps/export/tasks.py
```diff
@@ -26,16 +28,32 @@ def populate_export_download_task(export_instances, filters, download_id, filena
...
-        mimetype=file_format.mimetype,
-        content_disposition=safe_filename_header(filename, file_format.extension),
-        download_id=download_id,
-    )
-    export_file.file.delete()
+    domain = export_instances[0].domain
+
+    try:
+        if BLOBDB_EXPORTS.enabled(domain):
+            db = get_blob_db()
```
benrudolph: @millerdev is this something that is better down at the module level rather than inside of a function?

NoahCarnahan (commented):
Looks ok to me, but I'll let Daniel take a pass too since I'm not super familiar with the blobdb.

corehq/ex-submodules/soil/__init__.py
```diff
@@ -257,3 +258,60 @@ def create(cls, path, **kwargs):
...
+        self.blobdb = get_blob_db()
+
+    def get_filename(self):
+        return self.identifier
+
+    def get_content(self):
+        file_obj = None
+        try:
+            file_obj = self.blobdb.get(self.identifier, self.bucket).read()
+            return file_obj.read()
```
millerdev: This reads the entire file into memory. I don't see where `get_content` it is used. Is it needed?

millerdev (approved):
Do you think it's worth writing some tests?

corehq/ex-submodules/soil/__init__.py
benrudolph: yeah... i only implemented it because the API defined it. i think i'm actually going to just raise a notimplementederror because we shouldn't be calling `get_content` with the blobdownload

benrudolph:
good call on the test, i did actually look to see if soil had tests and there were none, but that's not a good excuse not to write a test

millerdev (approved):

PR #15655, opened 2017-04-06
export timings

benrudolph:
@NoahCarnahan want to get some stats on this. i have this weird feeling that processing 500 docs is taking a _very_ long time

https://github.com/dimagi/commcare-hq/pull/15655/files?w=1
cc: @dannyroberts @proteusvacuum

### DIFF
--- a/corehq/apps/export/export.py
+++ b/corehq/apps/export/export.py
@@ -375,8 +375,8 @@ def _write_export_instance(writer, export_instance, documents, progress_tracker=
                 DownloadBase.set_progress(progress_tracker, row_number + 1, documents.count)
 
             datadog_gauge('export_iteration', context.duration, tags=[
-                'iteration:{}'.format(row_number / 500),
-                'domain:{}'.format(domain),
+                u'iteration:{}'.format(row_number / 500),
+                u'domain:{}'.format(domain),
             ])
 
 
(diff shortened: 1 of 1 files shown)


corehq/apps/export/export.py
```diff
@@ -354,20 +356,28 @@ def _write_export_instance(writer, export_instance, documents, progress_tracker=
...
+        with TimingContext('write_export') as context:
+            for table in export_instance.selected_tables:
+                rows = table.get_rows(
+                    doc,
+                    row_number,
+                    split_columns=export_instance.split_multiselects,
+                    transform_dates=export_instance.transform_dates,
+                )
+                for row in rows:
+                    # It might be bad to write one row at a time when you can do more (from a performance perspective)
```
dimagimon: E501 line too long (118 > 115 characters)

NoahCarnahan (approved):
Looks good. Test failure looks related though. Just to confirm, we're interested in the timing for 500 documents because that's how big each scroll query is?

benrudolph:
Yes and I'm curious if there is any relation with the number of iterations and the slowness of processing the request. It's weird to me that the scroll query times out sometimes (means we couldn't process 500 forms in 5 mins). I'm kind of wondering if somehow the previous iterations' forms don't get cleaned up or if that only happens when there is a lot of load on the machine. Shot in the dark, but I think worth it

PR #15662, opened 2017-04-06
ICDS: Add all indicators to monthly tables

sheelio:
@emord 

This PR makes sure we're saving all of our monthly indicators to the monthly Tableau tables.  This will allow us to download data if we need to share with the government.


emord (approved):

sheelio:
Test failure doesn't look related.  @emord, I'll merge this if you think that's ok.

emord:
@sheelio I restarted the build looked like a network error

PR #15664, opened 2017-04-06
Remove tab refresh behavior from app manager

orangejenny:
https://manage.dimagi.com/default.asp?251221

@esoergel / @proteusvacuum


proteusvacuum (approved):
Thanks!

PR #15674, opened 2017-04-07
Include host as a "parent" type

benrudolph:
@proteusvacuum @dannyroberts this just registers host cases so that exports knows to add the Parent Cases table

### DIFF
--- a/corehq/apps/export/models/new.py
+++ b/corehq/apps/export/models/new.py
@@ -1788,7 +1788,7 @@ class CaseExportDataSchema(ExportDataSchema):
             app.copy_of or app._id,  # If not copy, must be current app
             app.version,
         ))
-        if any(map(lambda relationship_tuple: relationship_tuple[1] == 'parent', parent_types)):
+        if any(map(lambda relationship_tuple: relationship_tuple[1] in ['parent', 'host'], parent_types)):
             case_schemas.append(cls._generate_schema_for_parent_case(
                 app.copy_of or app._id,
                 app.version,


benrudolph:
https://manage.dimagi.com/default.asp?250827

proteusvacuum:
@benrudolph this looks fine to solve that ticket. However I don't think this is a tenable way of getting all of the "hosts" or "parents" of a case. In the [casexml](https://github.com/dimagi/commcare-core/wiki/casexml20) spec, the `tag` for an index is user defined so you could have something like the following:
```
<mother relationship="child" case_type="person">{case_id}</mother>
<master relationship="extension" case_type="person">{case_id}</master>
```
I don't think this code wouldn't pick up either of those case indices (they would be `mother` and `master`, not `parent` or `host`). Is there a reason that `in` clause is needed?

proteusvacuum:
(also, we've been using different tags in phase 2 to make things clearer in app building, so this would fail for us later down the road...)

corehq/apps/export/models/new.py
```diff
@@ -1788,7 +1788,7 @@ def _process_app_build(cls, current_schema, app, case_type):
             app.copy_of or app._id,  # If not copy, must be current app
             app.version,
         ))
-        if any(map(lambda relationship_tuple: relationship_tuple[1] == 'parent', parent_types)):
+        if any(map(lambda relationship_tuple: relationship_tuple[1] in ['parent', 'host'], parent_types)):
```
proteusvacuum: is there a test for this anywhere?

benrudolph:
May have to offline with you a bit to fully understand, but this just checks the value returned from the app manager unil that gets the case types related cases. From what I understand that relationship type can only be a certain set of things. 

Yes there is a test testing the parent type, I didn't make a test for the host because it's really the responsibility of app manager to return the correct relationship type of the case type. It seemed overkill to ensure that every relationship type worked (imagining if there were more I think it's sufficient to ensure one relationship type correctly register because the code path is exactly the same for any additional relationship type)

benrudolph:
*util that checks the case type's related case types and relationship

PR #15679, opened 2017-04-07
Don't error on UCR transform

NoahCarnahan:
If the value is `""`, this will error. Not certain if `""` is the best return value, or returning the value is better.
@nickpell cc @millerdev

### DIFF
--- a/corehq/apps/userreports/transforms/custom/date.py
+++ b/corehq/apps/userreports/transforms/custom/date.py
@@ -4,7 +4,10 @@ from dimagi.utils.dates import force_to_datetime
 
 
 def get_month_display(month_index):
-    return calendar.month_name[int(month_index)]
+    try:
+        return calendar.month_name[int(month_index)]
+    except Exception as e:
+        return ""
 
 
 def days_elapsed_from_date(date):


corehq/apps/userreports/transforms/custom/date.py
```diff
@@ -4,7 +4,10 @@
 
 
 def get_month_display(month_index):
-    return calendar.month_name[int(month_index)]
+    try:
+        return calendar.month_name[int(month_index)]
+    except Exception as e:
```
dimagimon: F841 local variable 'e' is assigned to but never used

nickpell: nit: this is a broad exception case.  Could be more explicit by catching ValueError if month_index== "", KeyError if month_index == 13, etc.

NoahCarnahan: 👍

NoahCarnahan: https://github.com/dimagi/commcare-hq/pull/15682

PR #15684, opened 2017-04-07
app manager v2: styling for custom properties

orangejenny:
Before
screen shot 2017-04-07 at 2 14 57 pm

After
screen shot 2017-04-07 at 2 14 15 pm

The custom properties section doesn't behave as a part of the accordion (it can stay open even if one of the other sections is also open). I think that's fine for a feature flag.

`w=1`

@esoergel / @biyeun


esoergel (approved):
Looks good.  Thanks for the screenshot!

biyeun (commented):
dope

biyeun (approved):

PR #15687, opened 2017-04-07
App manager v2: support adding additional module types

orangejenny:
@biyeun feel free to replace icons or change text (I wouldn't have bothered with text, since they're feature flags, but the alignment gets out of whack if there isn't text).

screen shot 2017-04-07 at 5 18 58 pm

cc @esoergel


biyeun:
LOL these icons are great! Love the shadow menu HAHAHA

PR #15690, opened 2017-04-07
Django has a 30 char limit on these two fields

esoergel:
Looks like they're planning on lifting it to 150 eventually:
https://docs.djangoproject.com/en/dev/ref/contrib/auth/#django.contrib.auth.models.User.last_name
For now we should probably catch this
@proteusvacuum


proteusvacuum:
I guess all our users just have very short names. I only have 21 instances of this error in my email since November 2015.

proteusvacuum (approved):

PR #15695, opened 2017-04-08
Fix unicode column names

emord:
@biyeun looks like I missed a column a couple of places in my first PR, and had an incorrect test for table name

Going to hotfix this on prod (update: hotfix worked and ucrs are processing again)

@kaapstorm


emord:
I hot fixed this on prod yesterday morning. will go ahead and self merge so there's no confusion for what's on prod

PR #15698, opened 2017-04-08
more indexes on person case

emord:
@benrudolph best I could come up with after looking at the logs. May want to look into adding multi-column indexes to ucrs

buddy @kaapstorm

### DIFF
--- a/custom/icds_reports/ucr/data_sources/awc_locations.json
+++ b/custom/icds_reports/ucr/data_sources/awc_locations.json
@@ -56,6 +56,7 @@
         "is_nullable": false,
         "is_primary_key": false,
         "column_id": "supervisor_id",
+        "create_index": true,
         "expression": {
           "location_id_expression": {
             "datatype": null,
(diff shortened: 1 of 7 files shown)


benrudolph:
@emord i added a few more indexes. do these automatically get added on deploy? i'm curious because some of the tables had `supervisor_id` as `create_index: true`, but i didn't see it when i was running queries on the server

emord:
@benrudolph Thanks for looking throuhg that. they should be created during the bootstrap process, but maybe there's a bug. If you rebuild the table they're definitely created, but maybe the pillow doesn't detect index changes? I'll look into it tomorrow

I'll go ahead and merge this since, it's what's running on NIC

benrudolph:
👍 i wasn't sure how it does the index checking process, but i doubt i did it in a way that the ucr will detect (i just used the naming scheme that was in the doc). unless it does something clever to detect if a column has index. in any case, i can rename the indexes once we want to deploy

emord:
It uses sqlalchemy/alembic to check the difference in tables. not sure why it's not picking up indexes though

emord:
Index additions followed up https://github.com/dimagi/commcare-hq/pull/15710

PR #15701, opened 2017-04-10
App manager v2: more intelligently pick case type for new module

orangejenny:
The old logic assigned a new case type for each module: `case` for the first, `case-1` for the second, etc. Thought about this and decided it doesn't really make sense...people are reasonably likely to want a second module with the same case type as an existing module, but they're never going to want a second module called `case-1`, so we might as well default to an existing case type. People will still often need to change it, but at least they won't *always* need to change it.

@biyeun / @esoergel

### DIFF
--- a/corehq/apps/app_manager/views/modules.py
+++ b/corehq/apps/app_manager/views/modules.py
@@ -994,14 +994,11 @@ def new_module(request, domain, app_id):
                 followup.actions.update_case = UpdateCaseAction(condition=FormActionCondition(type='always'))
 
                 # make case type unique across app
-                app_case_types = set(
-                    [module.case_type for module in app.modules if
-                     module.case_type])
-                module.case_type = 'case'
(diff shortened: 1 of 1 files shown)


corehq/apps/app_manager/views/modules.py
```diff
@@ -994,14 +994,11 @@ def new_module(request, domain, app_id):
...
                 # make case type unique across app
-                app_case_types = set(
-                    [module.case_type for module in app.modules if
-                     module.case_type])
-                module.case_type = 'case'
-                suffix = 0
-                while module.case_type in app_case_types:
-                    suffix = suffix + 1
-                    module.case_type = 'case-{}'.format(suffix)
+                app_case_types = [module.case_type for module in app.modules if module.case_type]
```
dimagimon: F812 list comprehension redefines 'module' from line 974

PR #15702, opened 2017-04-10
Use time.time instead of timer

benrudolph:
@NoahCarnahan misread how to use the TimerContext. i would've needed to use another `with` clause to get it to work. it's not really meant for this kind of thing, so opted for using `time.time`


benrudolph:
https://github.com/dimagi/commcare-hq/pull/15702/files?w=1

proteusvacuum (approved):

PR #15705, opened 2017-04-10
Blob restore

benrudolph:
(opening for tests)

### DIFF
--- a/corehq/ex-submodules/casexml/apps/phone/restore.py
+++ b/corehq/ex-submodules/casexml/apps/phone/restore.py
@@ -841,6 +841,6 @@ class RestoreConfig(object):
 
 def _get_restore_class(domain):
     return BlobRestoreResponse
-    if BLOBDB_RESTORE.enabled(self.domain):
+    if BLOBDB_RESTORE.enabled(domain):
         return BlobRestoreResponse
     return FileRestoreResponse
--- a/corehq/ex-submodules/casexml/apps/phone/restore.py
+++ b/corehq/ex-submodules/casexml/apps/phone/restore.py
@@ -18,7 +18,7 @@ from casexml.apps.phone.exceptions import (
     BadStateException, RestoreException, DateOpenedBugException,
 )
 from casexml.apps.phone.tasks import get_async_restore_payload, ASYNC_RESTORE_SENT
-from corehq.toggles import LOOSE_SYNC_TOKEN_VALIDATION, EXTENSION_CASES_SYNC_ENABLED
+from corehq.toggles import LOOSE_SYNC_TOKEN_VALIDATION, EXTENSION_CASES_SYNC_ENABLED, BLOBDB_RESTORE
 from corehq.util.soft_assert import soft_assert
 from corehq.util.timer import TimingContext
 from dimagi.utils.decorators.memoized import memoized
@@ -460,7 +460,12 @@ class RestoreState(object):
     This allows the providers to set values on the state, for either logging or performance
     reasons.
     """
-    restore_class = FileRestoreResponse
+
+    @property
+    def restore_class(self):
+        if BLOBDB_RESTORE.enabled(self.domain) or BLOBDB_RESTORE.enabled(self.restore_user.username):
+            return BlobRestoreResponse
+        return FileRestoreResponse
 
     def __init__(self, project, restore_user, params, async=False):
         if not project or not project.name:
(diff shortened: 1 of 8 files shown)


corehq/ex-submodules/casexml/apps/phone/restore.py
```diff
@@ -735,3 +837,10 @@ def _set_cache_in_redis(self, cache_payload):
     def delete_cached_payload_if_necessary(self):
         if self.overwrite_cache and self.cache.get(self._initial_cache_key):
             self.cache.delete(self._initial_cache_key)
+
+
+def _get_restore_class(domain):
+    return BlobRestoreResponse
+    if BLOBDB_RESTORE.enabled(self.domain):
```
dimagimon: F821 undefined name 'self'

benrudolph:
this is actually a bit trickier than i thought. we do this weird optimization where we replace the sync id in the cached restore payload. it makes a lot of assumptions about the cached payload being a file that are hard to convert over to working with a blob

benrudolph:
@emord @calellowitz @sravfeyn ok this about ready to go. waiting for tests to pass on travis, then will need to remove this line: https://github.com/dimagi/commcare-hq/pull/15705/files#diff-668e726338b412d3a9f2631438979a57R850. should be fairly safe to merge since it should only do anything on toggle

sravfeyn (approved):
Build failure looks related though

corehq/ex-submodules/casexml/apps/phone/restore.py
```diff
@@ -229,6 +249,89 @@ def get_http_response(self):
...
+    def get_payload(cls, identifier):
+        try:
+            return get_blob_db().get(identifier)
+        except NotFound:
+            return None
+
+    def as_string(self):
+        try:
+            blob = self.blobdb.get(self.get_filename())
+            return blob.read()
```
millerdev: is this actually used? it reads the entire file into memory (obviously, since there's no other way to get as string)

benrudolph: this one is actually used in tests i believe. might make sense to wrap in unit_testing_only though

millerdev: hmm, why not do `.as_file().read()` in tests? I know it doesn't close the file, which is not ideal, but I hate mixing test code with live/production code. Especially when the test code should never be used in production.

If closing the file is a concern, then a little more effort is required in tests:
```py
with resp.as_file() as fh:
    do_thing_with_string(fh.read())
```

benrudolph: i think i did it more for simplicity since there's another object that implements `as_string` and it seemed the easiest way to get tests working. i'll take a pass and see if it's that hard to change

benrudolph:
i think this is OK to merge as is once tests pass. it won't actually affect anything until toggle is turned on. want to keep the scope of the PR somewhat small. have to do some synclog refactoring in order to get this to totally work. @millerdev i'm going to take that pass on tests in the followup PR

emord (approved):

corehq/ex-submodules/casexml/apps/phone/restore.py
```diff
@@ -329,7 +329,8 @@ def as_string(self):
             blob.close()
 
     def get_http_response(self):
-        return stream_response(self.blobdb.get(self.get_filename()))
+        headers = {'Content-Length': self.blobdb.size(self.get_filename())}
```
millerdev: It is unfortunate that we need to make two requests to Riak CS, one for content and one for size. Presumably, there is a content-length header returned from the original request for content, which could be reused.

Relatedly, it would be even better to use a proxy header like [`XSendfile`](https://www.nginx.com/resources/wiki/start/topics/examples/xsendfile/) to serve the blob directly from Riak CS (s3db) or file system (fsdb) rather than relaying through Django like this.

benrudolph: it's true, the content-length is returned when we get the object. i wasn't sure how to incorporate that without changing the api pretty significantly. i guess we can use some caching. or return a BlobInfo object on `get`

i think we can definitely use XSendfile. we already do for some things. on my roadmap for phase 2. essentially just wanted to get this working as soon as possible to alleviate the disk space issue we're having on icds

corehq/ex-submodules/casexml/apps/phone/restore.py
benrudolph: also btw, i have a commit that does this, but it's literally used _all_ over, so i left it out of my followup PR in order to not block on fixing tests. the killer one is the fact that calling `__str__` on the response uses `as_string` which is used implicitly in lots of tests and is hard to track down

millerdev: ugh, that's the worst.

PR #15706, opened 2017-04-10
Changes in Block 3 in Case Finding Report

kkrampa:
@NoahCarnahan 
http://manage.dimagi.com/default.asp?251350

* Use date_reported instead of date_of_diagnosis
* Changed headers Patients -> Cases


kkrampa:
@NoahCarnahan Could you merge also this PR?

PR #15696, opened 2017-04-08
Location safety for UCR and Logo; Preselect User Location in UCR filter

sravfeyn:
@biyeun

PR #15707, opened 2017-04-10
Location safe

sravfeyn:
https://manage.dimagi.com/default.asp?251580#edit_0_251580 https://manage.dimagi.com/default.asp?251578

@esoergel

### DIFF
--- a/corehq/apps/domain/views.py
+++ b/corehq/apps/domain/views.py
@@ -54,6 +54,7 @@ from corehq.apps.dhis2.forms import (
 )
 from corehq.apps.dhis2.models import JsonApiLog
 from corehq.apps.hqwebapp.templatetags.hq_shared_tags import toggle_js_domain_cachebuster
+from corehq.apps.locations.permissions import location_safe
 from corehq.apps.locations.forms import LocationFixtureForm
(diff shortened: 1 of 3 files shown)


sravfeyn:
On location-restriction, I think instead of a cryptic 403 to the user, we should display an error message that convey that it's a project setup organizations setup issue. @sheelio

sravfeyn:
Closing this as these changes are required in another PR https://github.com/dimagi/commcare-hq/pull/15696

esoergel:
@sravfeyn We currently show [this message](https://github.com/dimagi/commcare-hq/blob/master/corehq/apps/locations/middleware.py#L6-L11) if the user doesn't have a location assigned, or in other location restriction things, we show [this message](https://github.com/dimagi/commcare-hq/blob/master/corehq/apps/locations/permissions.py#L95-L99).  I'd definitely be in support of a clearer message or more precise messages in various circumstances if you wanna take a crack at it.  For example, we could distinguish between a page being unavailable because it's not location-safe and a page being unavailable because it IS location-safe, but the data requested is out of scope.

sravfeyn:
My bad, it already is showing the message for non-location-safe page as well. Thanks for explanation!

PR #15710, opened 2017-04-10
add indexes

emord:
@benrudolph this will actually add the indices without removing the ones that are already there

buddy @kaapstorm

### DIFF
--- a/corehq/apps/userreports/tests/test_pillow.py
+++ b/corehq/apps/userreports/tests/test_pillow.py
@@ -7,12 +7,14 @@ from kafka.common import KafkaUnavailableError
 from mock import patch, MagicMock
 from datetime import datetime, timedelta
 from six.moves import range
+from sqlalchemy.engine import reflection
 
 from casexml.apps.case.mock import CaseBlock
 from casexml.apps.case.models import CommCareCase
 from casexml.apps.case.signals import case_post_save
 from casexml.apps.case.tests.util import delete_all_cases, delete_all_xforms
 from casexml.apps.case.util import post_case_blocks
+
 from corehq.apps.change_feed import topics
 from corehq.apps.change_feed.producer import producer
 from corehq.apps.userreports.const import UCR_SQL_BACKEND, UCR_ES_BACKEND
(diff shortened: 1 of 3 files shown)


corehq/ex-submodules/fluff/signals.py
```diff
@@ -136,6 +137,53 @@ def get_tables_to_rebuild(diffs, table_names):
...
+        indexes_by_table_and_col[(index.table.name, column_names)].append((action, index))
+
+    indexes_to_change = []
+
+    for indexes in indexes_by_table_and_col.values():
+        if len(indexes) == 1:
+            indexes_to_change.append(indexes[0])
+        else:
+            # tests when alembic tries to remove/add the same index
+            assert len(indexes) == 2
```
benrudolph: how does this happen? could it ever be more than 2?

corehq/ex-submodules/fluff/signals.py
```diff
@@ -136,6 +137,53 @@ def get_tables_to_rebuild(diffs, table_names):
...
+            actions = [action for action, index in indexes]
+            assert DiffTypes.ADD_INDEX in actions
+            assert DiffTypes.REMOVE_INDEX in actions
+
+    return indexes_to_change
+
+
+def apply_index_changes(engine, raw_diffs, table_names):
+    indexes = _get_indexes_to_change(raw_diffs, table_names)
```
benrudolph: nit would be nice if this returned a list of named tuples

corehq/apps/userreports/tests/test_pillow.py
```diff
@@ -520,3 +521,43 @@ def _save_sql_case(doc):
...
+        adapter = get_indicator_adapter(config)
+
+        # mock rebuild table to ensure the table isn't rebuilt when adding index
+        pillow = get_kafka_ucr_pillow()
+        pillow.processors[0].rebuild_table = MagicMock()
+        pillow.bootstrap([config])
+        self.assertFalse(pillow.processors[0].rebuild_table.called)
+        engine = adapter.engine
+        insp = reflection.Inspector.from_engine(engine)
+        self.assertEqual(len(insp.get_indexes(table_name)), 1)
```
benrudolph: nice test

benrudolph (approved):
👍

corehq/ex-submodules/fluff/signals.py
emord: > how does this happen?

I'm not sure why this is but when alembic generates an index it can generate a list like:

[(add_index, index_with_name_a_on_col),(remove_index, index_with_name_b_on_col)]

But for som reason when if you add the index with name a, the name b will get added consistently. It also makes sure we don't remove/add any indexes that were added manually on icds

> could it ever be more than 2

I don't think so, but that's why I added the assertions. I think since this is based on table and column that the only thing that could happen would be 1 add and one remove. I'd rather something like this fail hard before altering the database, but happy to look into any other solutions you might have

benrudolph: that makes sense. thanks for the explanation

corehq/ex-submodules/fluff/signals.py
emord: https://github.com/dimagi/commcare-hq/pull/15710/commits/b0ef6deabeccb1042c28a3dc9b2ff230c64021e1

emord:
I think that this is safe to deploy whenever but good to know that it will cause some delays for UCRs on india and NIC (not like they're up to date anyways...) We should do this whenever NIC isn't being used as much because it'll kick off a bunch at once.

benrudolph (approved):

PR #15711, opened 2017-04-10
ICDS: Add Views for Tableau

sheelio:
@emord 

I'm moving some of the SQL queries we have out of Tableau into our code base.  This just adds views for some of the custom SQL queries so they can be re-used and/or optimized when we move away from Tableau.


emord:
So much SQL. curious if you investigated using SQLalchemy to do some of these?

sheelio:
Have not used SQL Alchemy - once we switch to away from Tableau to build the ICDS Dashboard, we can maybe start using SQL alchemy or something else.

PR #15716, opened 2017-04-11
Save case search config if there are only default properties

proteusvacuum:
@benrudolph @dannyroberts 
We are going to have the mobile skip the search screen if there is only default properties at some point. This allows HQ to generate the required XML for this.


corehq/apps/app_manager/tests/test_suite_remote_request.py
```diff
@@ -91,3 +91,24 @@ def test_case_search_action_relevant_condition(self):
         suite = self.app.create_suite()
         suite = parse_normalize(suite, to_string=False)
         self.assertEqual(condition, suite.xpath('./detail[1]/action/@relevant')[0])
+
+    def test_only_default_properties(self):
+        self.module.search_config = CaseSearch(
+            default_properties=[
+                DefaultCaseSearchProperty(
+                    property=u'ɨŧsȺŧɍȺᵽ',
```
benrudolph: 🐭 :cheese: 🔫

benrudolph (approved):

dannyroberts (approved):

PR #15718, opened 2017-04-11
catch unknown vhnsd date

emord:
@gcapalbo there were a lot errors from icds this weekend https://sentry.io/dimagi/commcarehq/issues/249602018/

I looked at the form and noticed that there's a display condition on that question so not all forms will have it

buddy @kaapstorm


kaapstorm (approved):

gcapalbo:
thanks @emord , was tracking this one [here](https://manage.dimagi.com/default.asp?250472) but didn't get a chance to look at it yet 

i'll confirm the behavior with the icds team, but going to merge this in the meantime as it's better than what happens in the current state

PR #15721, opened 2017-04-11
Add payload id to repeat record report

proteusvacuum:
Adds the payload_id to the repeat record report.
@benrudolph @dannyroberts


proteusvacuum:
FYI @mkangia

benrudolph (approved):
would it make sense to have this behind the support toggle? i don't immediately see how this is useful to the user

dannyroberts (approved):

mkangia (approved):
Thanks for adding this. I think this would be useful in debugging issues. I often generate payload to fetch the case id. And in case the payload errors out, this would be a good place to start.

benrudolph (approved):

PR #15722, opened 2017-04-11
fix compare ucr for static reports

emord:
@benrudolph @calellowitz @sravfeyn for NIC laboratory ucr testing

@kaapstorm

### DIFF
--- a/corehq/apps/userreports/tasks.py
+++ b/corehq/apps/userreports/tasks.py
@@ -20,7 +20,7 @@ from corehq.apps.userreports.models import (
     DataSourceConfiguration,
     StaticDataSourceConfiguration,
     id_is_static,
-    ReportConfiguration,
+    get_report_config,
 )
 from corehq.apps.userreports.reports.factory import ReportFactory
 from corehq.apps.userreports.util import get_indicator_adapter, get_async_indicator_modify_lock_key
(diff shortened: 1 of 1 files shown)


benrudolph (approved):

corehq/apps/userreports/tasks.py
```diff
@@ -170,7 +170,7 @@ def _run_report(backend_to_use):
             json_response["total_row"] = total_row
         return json_response
 
-    spec = get_document_or_not_found(ReportConfiguration, domain, report_config_id)
```
sravfeyn: Isn't there any code that is expecting an exception if config_id is not found, does this change handle that?

kaapstorm (approved):

corehq/apps/userreports/tasks.py
kaapstorm: It seems as if this task is only used in one place, [`ConfigurableReport.get_ajax()`](https://github.com/dimagi/commcare-hq/blob/871586315e7c155cf70382f4f91ff204fff7b188/corehq/apps/userreports/reports/view.py#L429-L429). `DocumentNotFound` exceptions are not caught, and the return result is not used.

emord: @sravfeyn This code never gets called unless a report has already been grabbed in the main HQ. I hadn't really thought about it before, but seems like enough of an edge case that it's not worth fixing at the moment (especially since I control laboratory UCRs)

PR #15723, opened 2017-04-11
Switch app manager v2 flag from domain to user

orangejenny:
To simplify the eventual migration.

The only people with this turned on in prod are me, Biyeun, and Jeremy.

@esoergel / @biyeun


orangejenny:
Pinging on review, this is in QA now and it'd be good to have the flag configuration settled.

PR #15727, opened 2017-04-11
pillow partition

emord:
Building off of simon's pr

@benrudolph this safely migrates any checkpoints that use kafka to a new model. I'm picturing the rollout like this (dev is still needed on #3):

1) Deploy Simon's and this change
2) Run one off management command
3) Add ability to specify topics and partitions to pillows 1 by 1, when this is done, it should only be using KafkaCheckpoint instead of DjangoPillowCheckpoint. This will be a local settings change in LOCAL_PILLOWTOPS. Eventually this should be moved to commcare-hq-deploy similar to the celery options
The initial implementation will build off of the ucr params and will look like:
```json
{
    "topics_and_partitions": {
        "case-sql": [0,1,2,3],
        "form-sql": [0,1,2,3],
        "location": [0]
    }
}
```
4) Add checkpoints to high use topics like cases, using the provided management command.
5) split expensive pillows into multiple pillows using work from #3. I think the first candidates will be the UCR pillows and form submission metadata.

I think this work should be done to fill up our 16 cores and then moving the config to commcare-hq-deploy so that we can use two pillowtop machines.

The reasoning for using a new model is so that we can arbitrarily split up partitions in the future among multiple workers without any thought

@kaapstorm


kaapstorm:
LGTM, but I don't have a lot of context on this. Will leave it to @benrudolph to Approve.

benrudolph:
awesome. one question about the kafka and partition dict. `"case-sql": [0,1,2,3]` that would specify that for whichever pillow is subscribed to that topic, it would be subscribed to partitions 0,1,2,3 correct?

if we add the kafka checkpoint model, that means we have to change this right? https://github.com/dimagi/commcare-hq/blob/master/corehq/ex-submodules/pillowtop/checkpoints/manager.py#L18

emord:
@benrudolph I just realized how fragile that partition doc will be. Instead it'll be a change to say how many processes to run and the pillow will handle how to split up partitions

That function should change once we have pillows that rely on KafkaCheckpoint for the source of truth. for now it will still rely on DjangoPillowCheckpoint and only write to KafkaCheckpoint. I want to do it this way so that we can lazily migrate over to KafkaCheckpoint pillow by pillow instead of needing to change 20 pillows on prod at once (and i want to save custom code for last)

benrudolph:
sounds good

benrudolph (approved):
whenever you're ready

PR #15728, opened 2017-04-11
enable deleting migrated cases by person

nickpell:
recommend 🐡 

@proteusvacuum @mkangia cc: @calellowitz


mkangia (approved):
Will let farid weigh in for context. But otherwise looks good.

proteusvacuum:
cool

PR #15731, opened 2017-04-11
App manager v2: show advanced features used in simple apps

orangejenny:
To simplify migration (make it easier to flip toggle on/off and also reduce problems if one user in a domain is on v1 and another is on v2), make the various "advanced" features display if they're in use, even if this isn't an advanced app.

Easiest to review by commit.

@esoergel / @biyeun

### DIFF
--- a/corehq/apps/app_manager/templates/app_manager/v2/partials/case_list_setting.html
+++ b/corehq/apps/app_manager/templates/app_manager/v2/partials/case_list_setting.html
@@ -21,7 +21,7 @@
     </div>
     <div class="form-group" id="{{ SLUG }}-label">
         <label class="col-sm-2 control-label">
-            Label for Case List Menu Item
+            {% trans "Label for Case List Menu Item" %}
         </label>
         <div class="col-sm-4">
             {% input_trans case_list.label langs %}
--- a/corehq/apps/app_manager/templates/app_manager/v2/partials/case_list_setting.html
+++ b/corehq/apps/app_manager/templates/app_manager/v2/partials/case_list_setting.html
@@ -1,60 +1,65 @@
 {% load xforms_extras %}
 {% load hq_shared_tags %}
 {% load i18n %}
-<div class="form-group">
-    <label class="col-sm-2 control-label">
-        {% trans LABEL %}
-        <span class="hq-help-template"
-              data-title="{% trans LABEL %}"
-              data-content="{% trans DESC %}"
-        ></span>
-    </label>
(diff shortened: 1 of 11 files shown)


corehq/apps/app_manager/templates/app_manager/v2/partials/case_list_setting.html
```diff
@@ -1,60 +1,65 @@
...
+            </select>
+        </div>
     </div>
-    <div class="col-sm-2">
-        <span class="label label-danger" id="case_list_label_error" style="display: none;">
-            {% trans "Case list menu item requires a label" %}
-        </span>
+    <div class="form-group" id="{{ SLUG }}-label">
+        <label class="col-sm-2 control-label">
+            Label for Case List Menu Item
```
biyeun: trans

corehq/apps/app_manager/templates/app_manager/v2/partials/module_view_settings.html
```diff
@@ -119,17 +119,23 @@ <h4 class="panel-title panel-title-nolink">{% trans "Multimedia Settings" %}</h4
...
+           {% if not module.is_surveys %}
+             {% if app.advanced_app_builder or module.case_list_form.form_id or module.case_list.show or module.task_list.show %}
+             <div class="panel panel-appmanager">
+              <div class="panel-heading">
+                <h4 class="panel-title panel-title-nolink">{% trans "Advanced Settings" %}</h4>
+              </div>
+              <div class="panel-body">
+                {% include 'app_manager/v2/partials/case_list_form_setting.html' %}
+
+                {% include 'app_manager/v2/partials/case_list_setting.html' with LABEL="Case List Menu Item" DESC="An item in the module's menu that lets you browse the case list without moving on to fill out a form." SLUG="case_list" case_list=module.case_list %}
```
biyeun: trans?

corehq/apps/app_manager/templates/app_manager/v2/partials/case_list_setting.html
orangejenny: fixed

corehq/apps/app_manager/templates/app_manager/v2/partials/module_view_settings.html
orangejenny: I don't think so...that is, I don't think there's a straightforward way to do it in the template...that is, we don't do it in other, similar places.

PR #15739, opened 2017-04-12
Location info user tiles

benrudolph:
Adds location info to login as feature in preparation to do some more custom stuff for nikshay.

screen shot 2017-04-12 at 4 24 44 pm
screen shot 2017-04-12 at 4 24 50 pm

### DIFF
--- a/corehq/apps/users/views/mobile/users.py
+++ b/corehq/apps/users/views/mobile/users.py
@@ -638,7 +638,7 @@ class MobileWorkerListView(JSONResponseMixin, BaseUserSettingsView):
     def query(self):
         return self.request.GET.get('query')
 
-    def _format_user(self, user_json):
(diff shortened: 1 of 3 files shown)


corehq/apps/users/views/mobile/users.py
```diff
@@ -650,6 +650,7 @@ def _format_user(self, user_json):
             'last_name': user.last_name,
             'phoneNumbers': user.phone_numbers,
             'user_id': user.user_id,
+            'location': user.sql_location.to_json() if include_location and user.sql_location else None,
```
benrudolph: i went back and forth on whether or not to batch this location get by all the users returned. however since it's only ever 10 users and only used in the login as feature, i opted for doing the lookup on each individual user for readability

dannyroberts (approved):

PR #15741, opened 2017-04-12
fix change feed page

emord:
Fixes the hq/admin/system page for pillow checkpoints

@kaapstorm

### DIFF
--- a/corehq/apps/change_feed/consumer/feed.py
+++ b/corehq/apps/change_feed/consumer/feed.py
@@ -9,7 +9,8 @@ from corehq.apps.change_feed.data_sources import get_document_store
 from corehq.apps.change_feed.exceptions import UnknownDocumentStore
 from corehq.apps.change_feed.topics import get_multi_topic_offset, validate_offsets
 from dimagi.utils.logging import notify_error
-from pillowtop.checkpoints.manager import PillowCheckpointEventHandler, DEFAULT_EMPTY_CHECKPOINT_SEQUENCE
+from pillowtop.checkpoints.manager import PillowCheckpointEventHandler
+from pillowtop.models import kafka_seq_to_str
 from pillowtop.feed.interface import ChangeFeed, Change, ChangeMeta
 
 MIN_TIMEOUT = 100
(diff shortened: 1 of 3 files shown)


corehq/apps/change_feed/consumer/feed.py
```diff
@@ -101,6 +102,9 @@ def get_processed_offsets(self):
     def get_latest_offsets(self):
         return get_multi_topic_offset(self.topics)
 
+    def get_latest_offsets_json(self):
+        return json.loads(kafka_seq_to_str(self.get_latest_offsets()))
```
emord: The fact this returns a dict instead of json is some tech debt. the calling function needs a dict that is processed by `json_response`

kaapstorm: If you moved `seq = {'{},{}'.format(*key): val for key, val in seq.items()}` into its own function in [pillowtop/models.py](https://github.com/dimagi/commcare-hq/blob/718bc0c1f5e8b2d98ddb5fe7016e23fd7b935747/corehq/ex-submodules/pillowtop/models.py#L23-L23), say, `seq_to_kafka_seq()`, then you wouldn't need to call `json.loads()` on something that's just been returned from `json.dumps()`.

corehq/ex-submodules/pillowtop/utils.py
```diff
@@ -169,7 +169,7 @@ def get_pillow_json(pillow_config):
     def _seq_to_int(checkpoint, seq):
         from pillowtop.models import kafka_seq_to_str
         if checkpoint.sequence_format == 'json':
-            return kafka_seq_to_str(seq)
+            return json.loads(kafka_seq_to_str(seq))
```
kaapstorm: (ditto previous comment)

kaapstorm:
Just the one comment. Otherwise all good.

emord:
Models for kafka based checkpoints is about to change with https://github.com/dimagi/commcare-hq/pull/15727 so this will likely change in a future PR. going to go ahead and merge since the current page is broken

PR #15746, opened 2017-04-13
Motech openmrs auth

dannyroberts:
@proteusvacuum @benrudolph 

This PR connects a simple OpenMRS server settings page to the authproxy service I wrote a few weeks ago, so that openmrs credentials will be stored encrypted. (See [How we will store passwords to external services](https://docs.google.com/document/d/1gH7ETG_pzi6N6hJ6h9mccGml1dJz4Hr30wc-HOjNrwk/edit) for the concept). Unfortunately the code for that is all in the motechhub repo, which we're probably going to scrap for now. I didn't want to get into a big code reorganization right now, because how we want to organize the code hasn't totally coalesced for me yet, and I mostly just want to proof-of-concept these moving parts. So instead, I just added the motechhub repo as a submodule for now, and reference the code I'm interested in directly from it (as a file path) in the requirements.txt file. Let me know if you have other ideas for how to do it, but I think this mostly just affects me for now (though it does mean a new submodule for everyone else), and I'll probably end up changing it.


corehq/apps/motech/connceted_accounts.py
```diff
@@ -0,0 +1,49 @@
+import os
```
benrudolph: filename here is misspelled

corehq/apps/motech/connceted_accounts.py
```diff
@@ -0,0 +1,49 @@
...
+
+    if account:
+        return authproxy_client.requests(account.token)
+    else:
+        return None
+
+
+def get_openmrs_account(domain):
+    try:
+        return ConnectedAccount.objects.get(domain=domain, server_type='openmrs')
```
benrudolph: nit: may want to make this a constant

benrudolph (approved):
schweet

PR #15750, opened 2017-04-13
Run reminder rules periodically

gcapalbo:
https://manage.dimagi.com/default.asp?251705

The only reminders currently in ICDS are using the old framework and send emails in response to cases created or updated from the issue tracker, so the volume of these cases is low.

There's currently a big backlog of case updates to process (which I'm working on working down), so this is a workaround which will process the issue tracker reminders periodically every hour so that the emails will go out and won't be affected by the backlog.

@mkangia

### DIFF
--- a/corehq/apps/hqcase/esaccessors.py
+++ b/corehq/apps/hqcase/esaccessors.py
@@ -1,0 +1,29 @@
+from corehq.apps.es.cases import CaseES
+
+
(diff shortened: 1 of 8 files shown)


gcapalbo:
the only new commit to review for this PR is https://github.com/dimagi/commcare-hq/pull/15750/commits/9e325c05c68a58e315ae2652eab5a28ba31462fb.

the rest is in https://github.com/dimagi/commcare-hq/pull/15747

emord (approved):

PR #15755, opened 2017-04-13
App manager v2: Ability to turn off v2 in ui

orangejenny:
To make the eventual migration more user-friendly.

I put little thought into the design of this, would welcome suggestions. Would be nice for it to be somewhere non-app-manager specific, so it could be used for other rollouts.

screen shot 2017-04-13 at 2 13 19 pm

@esoergel / @biyeun 
fyi @millerdev


millerdev (approved):
Looks like there is no prompt for confirmation on click, is that correct? Is there a way to undo this if clicked accidentally?

biyeun:
how does this look with the Advanced App Builder tag? That's what that UI was meant for, and I didn't want it to be clickable (hence the dark background). otherwise I would have made it blue

biyeun:
Maybe use the toggle switch ui?

corehq/apps/hqwebapp/static/hqwebapp/js/rollout_modal.js
```diff
@@ -33,5 +36,27 @@ hqDefine("hqwebapp/js/rollout_modal.js", function() {
...
+
+        $("#rollout-revert").click(function() {
+            var slug = $(this).data("slug"),
+                redirect = $(this).data("redirect");
+            $.post({
+                url: hqImport("hqwebapp/js/urllib.js").reverse("toggle_" + slug),
+                data: {
+                    on_or_off: "off",
+                },
+                success: function(data) {
```
dimagimon: 'data' is defined but never used (no-unused-vars)

orangejenny:
@biyeun What do you think of this?

screen shot 2017-04-14 at 5 43 16 pm

@millerdev If you mis-click, you'd go back to v1, and then (currently 3 days later, the snooze period) you'd get the modal allowing you to get back to v2. Think it's worth doing a confirmation modal?

orangejenny:
Added modal confirmation before turning off flag.

millerdev:
Sorry, was offline/traveling so didn't see the question until now. I see you added a confirmation. Looks good to me.

orangejenny:
The act of writing out that question made me realize I should just go ahead and implement the modal ;)

PR #15757, opened 2017-04-14
make CaseDetailsView @location_safe

biyeun:
@proteusvacuum 

related: http://manage.dimagi.com/default.asp?251959


biyeun:
@proteusvacuum updated

biyeun:
@dannyroberts buddy @kaapstorm

dannyroberts (approved):
Looks good to me—should I let someone with more `@location_safe` experience like @esoergel review it?

corehq/apps/reports/views.py
```diff
@@ -1242,6 +1244,12 @@ def dispatch(self, request, *args, **kwargs):
                           "Sorry, we couldn't find that case. If you think this "
                           "is a mistake please report an issue.")
             return HttpResponseRedirect(CaseListReport.get_url(domain=self.domain))
+        if not user_can_access_location_id(
+            self.domain,
+            request.couch_user,
+            self.case_instance.owner_id
+        ):
```
esoergel: I think this is overly restrictive - these users should also be able to access cases owned by users at accessible locations.  That is, I think you'd need to figure out if the case is owned by a user or a location.
The `CaseInfo` class would help with that, something like:
```python
info = CaseInfo(report, case)
if info.owner_type == 'location':
    return user_can_access_location_id(domain, user, info.owner_id)
elif info.owner_type == 'user':
    owning_user = CommCareUser.get(info.owner_id)
    return user_can_access_other_user(domain, user, owning_user)
else:
    return False
```
If the case is owned by a group, it shouldn't ever be accessible to restricted users.

biyeun: @esoergel what's `report` referencing in this case? (realized my comment was pending, as I'd started a review) :cry:

esoergel: It looks to be the report instance itself, so in this case - `self`.  That's kind of a weird dependency, IMO, but from a quick perusal, it looks like the only place it might give you trouble is [here](https://github.com/dimagi/commcare-hq/blob/ab6b8bb384539d479da6299452c407016d08f605/corehq/apps/reports/standard/cases/data_sources.py#L120), which expects the report to be a subclass of `ProjectReportParametersMixin`, which defines `individual`.  It looks like adding `individual = ''` [would work](https://github.com/dimagi/commcare-hq/blob/ab6b8bb384539d479da6299452c407016d08f605/corehq/apps/reports/standard/__init__.py#L110).  If not, you may just have to pull out that logic or duplicate it to figure out case ownership.

corehq/apps/reports/views.py
```diff
@@ -1244,14 +1246,23 @@ def dispatch(self, request, *args, **kwargs):
...
+        domain = self.domain
+        user = self.request.couch_user
+        info = CaseInfo(self, case)
+        if info.owner_type == 'location':
+            return user_can_access_location_id(domain, user, info.owner_id)
+        elif info.owner_type == 'user':
+            owning_user = CommCareUser.get(info.owner_id)
+            return user_can_access_other_user(domain, user, owning_user)
+        else:
+            return False
```
esoergel: Sweet, yeah, this looks good to me.

esoergel (approved):

dannyroberts:
I should say, I haven't actually tested this locally, and I've passed this on (again) to the new interrupt team. Added the hold off on merge label. Sorry for the regrettably slow progress on this.

calellowitz:
added one more commit and i think we finally got this. @biyeun @dannyroberts @esoergel

proteusvacuum:
should we add a test for that new function?

corehq/apps/reports/views.py
```diff
@@ -1348,6 +1340,7 @@ def form_to_json(form):
     ])
 
 
+@location_safe
 class CaseAttachmentsView(CaseDetailsView):
```
esoergel: Does this for sure call the above restriction logic somewhere?

corehq/apps/reports/util.py
```diff
@@ -471,3 +472,17 @@ def resync_case_to_es(domain, case):
...
+    from corehq.apps.reports.standard.cases.data_sources import CaseInfo
+
+    info = CaseInfo(None, case)
+    if info.owner_type == 'location':
+        return user_can_access_location_id(domain, user, info.owner_id)
+    elif info.owner_type == 'user':
+        owning_user = CouchUser.get_by_user_id(info.owner_id)
+        return user_can_access_other_user(domain, user, owning_user)
+    else:
+        return False
```
esoergel: I like the idea of pulling this out into a reusable function.  It might fit better in `locations/permissions.py`, though.  There are a several `user_can_access_X` functions in there, and this would work well as `user_can_access_case` (maybe with the params rearranged to `domain, user, case` for consistency)

corehq/apps/reports/views.py
calellowitz: yeah, the dispatch function calls the `CaseDetailsView` dispatch function which has that logic in it

corehq/apps/reports/util.py
esoergel: I offlined with Cal, but wanted to note that this feedback is subjective and non-blocking.

calellowitz:
added test and moved util @esoergel @proteusvacuum

proteusvacuum (approved):

esoergel:
:balloon:

PR #15759, opened 2017-04-14
fix domain delete mgmt command

biyeun:
can't mix optional and position-based arguments and specify a dest for the position arg. throws `ValueError('dest supplied twice for positional argument')` otherwise

@calellowitz // @sravfeyn

### DIFF
--- a/corehq/apps/domain/management/commands/delete_domain.py
+++ b/corehq/apps/domain/management/commands/delete_domain.py
@@ -10,7 +10,7 @@ class Command(BaseCommand):
 
     def add_arguments(self, parser):
         parser.add_argument(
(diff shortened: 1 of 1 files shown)


corehq/apps/domain/management/commands/delete_domain.py
```diff
@@ -10,7 +10,7 @@ class Command(BaseCommand):
 
     def add_arguments(self, parser):
         parser.add_argument(
-            'domain',
+            '--domain',
```
calellowitz: it might be nicer to just remove the `dest` and rename the argument to `domain_name`, since this isnt actually an optional argument

PR #15761, opened 2017-04-14
Update deploy sub

benrudolph:
@proteusvacuum @dannyroberts update deploy module to add formplayer back to swiss


benrudolph:
killed build

PR #15762, opened 2017-04-14
Fix export sorting

benrudolph:
Still puzzling over why the export sorting isn't quite working for a few domains. This is definitely more correct. My theory is that people have been changing the question types and it thus creates a false order since the ordering only checks the path not the doc_type

https://manage.dimagi.com/default.asp?250350

### DIFF
--- a/corehq/apps/export/models/new.py
+++ b/corehq/apps/export/models/new.py
@@ -147,6 +147,19 @@ class ExportItem(DocumentSchema):
     inferred = BooleanProperty(default=False)
     inferred_from = SetProperty(default=set)
 
+    def __key(self):
+        return'{}:{}:{}'.format(
+            _path_nodes_to_string(self.path),
+            self.doc_type,
+            self.transform,
+        )
+
+    def __hash__(self):
+        return hash(self.__key())
+
+    def __eq__(self, other):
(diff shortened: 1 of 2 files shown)


corehq/apps/export/models/new.py
```diff
@@ -1425,11 +1438,7 @@ def _merge_schemas(cls, *schemas):
         def resolvefn(group_schema1, group_schema2):
 
             def keyfn(export_item):
-                return'{}:{}:{}'.format(
-                    _path_nodes_to_string(export_item.path),
-                    export_item.doc_type,
-                    export_item.transform,
-                )
```
benrudolph: this code is well tested, so as long as tests pass, i think we should be fine

corehq/apps/export/tests/test_export_data_schema.py
```diff
@@ -1239,3 +1241,33 @@ def test_partial_match_ordering(self):
                 ExportItem(path=[PathNode(name='three')]),
             ],
         )
+
+    def test_different_doc_types_ordering(self):
```
benrudolph: previous tests never tested if the doc type changed

NoahCarnahan:
This looks like it was a tricky one!

PR #15773, opened 2017-04-14
parallel pillow processing

emord:
opening for the tests. it's not quite ready for prime time and will need to be cleaned up.

The usage will be:

`./manage.py run_ptop --pillow-name pillowname --num-process 4 --process-num 0`

and we'll run four of those with the pillowname. it'll split up the topics among themselves. not the prettiest solution, but it contains the least configuration while allowing us to split them up 

@benrudolph @kaapstorm @snopoke


corehq/apps/userreports/pillow.py
```diff
@@ -301,10 +321,16 @@ def get_kafka_ucr_pillow(pillow_id='kafka-ucr-main', ucr_division=None,
     )
 
 
-def get_kafka_ucr_static_pillow(pillow_id='kafka-ucr-static', ucr_division=None,
-                                include_ucrs=None, exclude_ucrs=None, topics=None):
+def get_kafka_ucr_static_pillow(pillow_id='kafka-ucr-static', params=None):
```
benrudolph: how come you went with passing in one `params` object over individual kwargs? looks like that's how it was previously

corehq/ex-submodules/pillowtop/models.py
```diff
@@ -76,3 +76,19 @@ class KafkaCheckpoint(models.Model):
 
     class Meta:
         unique_together = ('checkpoint_id', 'topic', 'partition')
+
+    @classmethod
+    def create_for_checkpoint_id(cls, checkpoint_id, topics):
+        tps = list(
+            cls.objects
+            .filter(topic__in=topics)
+            .distinct('topic', 'partition')
```
benrudolph: what happens when we alter the partitions? guessing it shouldn't affect anything, but wonder if there should be any cleanup or if there are any edge cases

benrudolph (commented):
cool looking good

corehq/apps/change_feed/consumer/feed.py
```diff
@@ -20,12 +20,13 @@ def normalize_topics(topics, group_id):
...
-        .filter(checkpoint_id=group_id, topic__in=topics)
-        .values_list('topic', 'partition')
-        .order_by('topic', 'partition')
+    consumer = KafkaConsumer(
+        *topics,
+        group_id=group_id,
+        bootstrap_servers=[settings.KAFKA_URL]
     )
+    offsets = consumer.offsets('fetch')
+    return offsets.keys()
```
snopoke: Could use `get_all_offsets().keys()`

snopoke (commented):
Haven't done a full review, still trying to understand how this will work.

corehq/ex-submodules/pillowtop/pillow/interface.py
```diff
@@ -89,6 +89,7 @@ def process_changes(self, since, forever):
         """
         Process changes from the changes stream.
         """
+        print("processing changes from: {}", since)
```
snopoke: print

corehq/ex-submodules/pillowtop/utils.py
```diff
@@ -74,10 +75,15 @@ def __hash__(self):
     def get_class(self):
         return _import_class_or_function(self.class_name)
 
-    def get_instance(self):
+    def get_instance(self, num_processes=1, process_num=0):
```
snopoke: why are these passed in separately from `params`

corehq/pillows/app_submission_tracker.py
```diff
@@ -10,21 +10,23 @@
...
+def get_form_submission_metadata_tracker_pillow(pillow_id='FormSubmissionMetadataTrackerProcessor', params=None):
     """
     This gets a pillow which iterates through all forms and marks the corresponding app
     as having submissions. This could be expanded to be more generic and include
     other processing that needs to happen on each form
     """
-    change_feed = KafkaChangeFeed(topics=[topics.FORM, topics.FORM_SQL], group_id='form-processsor')
-    checkpoint = PillowCheckpoint('form-submission-metadata-tracker', change_feed.sequence_format)
+    checkpoint_id = 'form-submission-metadata-tracker'
+    kafka_topics = [topics.FORM, topics.FORM_SQL]
```
snopoke: got quite confused as to whether these should be tuples or not. Would be nice to standardise from some point down.

snopoke (commented):
Looking good. Was mostly confused before because I'd missed the first PR that introduced the `KafkaPillowCheckpoint`

emord:
This PR is unwieldy and hard to follow. Going to do a bit of a refactor and split it up into smaller PRs so that it's easier to review

PR #15779, opened 2017-04-15
start tracking restores

benrudolph:
@snopoke want to record some metrics before rolling out blob db temp restore

### DIFF
--- a/corehq/apps/ota/views.py
+++ b/corehq/apps/ota/views.py
@@ -24,6 +24,7 @@ from corehq.apps.domain.decorators import (
     check_domain_migration,
     login_or_digest_or_basic_or_apikey_or_token,
 )
+from corehq.util.datadog.gauges import datadog_counter, datadog_gauge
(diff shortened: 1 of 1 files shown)


dannyroberts:
Do you think this won't have much of an impact on performance, even given the restore load on ICDS? That seems plausible to me if it just checks the time before and after and sends the difference to datadog (via UDP), but want to make sure you thought about it.

benrudolph:
👍 yes i think in general we've been assuming datadog queries are negligible. we already running the timing on the restore regardless of whether or not we report on it. i also think the insight into restore timings is worth the very slight performance hit. currently restores are around 40 seconds to a minute. we've already had two instances of restores jumping up above 5 mins and we had to wait until someone from the field reported it. it'd be great if we could alert if restore breach a certain threshold. granted our alerts are very noisy right now, but maybe soon

dannyroberts:
Yeah, that sounds great to me

PR #15783, opened 2017-04-17
Motech openmrs concepts

dannyroberts:
screen shot 2017-04-17 at 3 04 48 pm

I'd review commit by commit. Note that one of the early commits is moving code over from the other repo.


corehq/apps/motech/static/motech/js/motech_openmrs_concepts.js
```diff
@@ -0,0 +1,33 @@
...
+        var self = this;
+        self.concepts = ko.observableArray([]);
+        self.syncButton = {
+            inProgress: ko.observable(false),
+            hasError: ko.observable(false),
+            sync: function () {
+                self.concepts([]);
+                self.syncButton.inProgress(true);
+                self.syncButton.hasError(false);
+                $.get(urllib.reverse('openmrs_sync_concepts')).done(function (data) {
```
dimagimon: 'data' is defined but never used (no-unused-vars)

corehq/apps/motech/static/motech/js/motech_openmrs_concepts.js
```diff
@@ -0,0 +1,33 @@
...
+                    self.refreshConcepts();
+                    self.syncButton.inProgress(false);
+                    self.syncButton.hasError(false);
+                }).fail(function () {
+                    self.syncButton.inProgress(false);
+                    self.syncButton.hasError(true);
+                });
+            }
```
dimagimon: Missing trailing comma. (comma-dangle)

corehq/apps/motech/openmrs/concepts/sync.py
```diff
@@ -0,0 +1,113 @@
+import json
+from django.db import IntegrityError
```
dimagimon: F401 'IntegrityError' imported but unused

corehq/apps/motech/openmrs/tests.py
```diff
@@ -0,0 +1,3 @@
+from django.test import TestCase
```
dimagimon: F401 'TestCase' imported but unused

corehq/apps/motech/openmrs/concepts/views.py
```diff
@@ -0,0 +1,50 @@
...
+            return JsonResponse({'concepts': []})
+        else:
+            return JsonResponse({'concepts': [
+                openmrs_concept_json_with_answers_from_concept(concept).to_json()
+            ]})
+    elif len(search) > 2:
+        all_openmrs_concepts = OpenmrsConcept.objects.filter(Q(account=account) & ~Q(answers=None))
+        openmrs_concepts = all_openmrs_concepts.filter(names__icontains=search)
+        first_50 = openmrs_concepts[:50]
+        return JsonResponse({'concepts': [
```
benrudolph: does jsonresponse not already automatically call `to_json` on all objects?

corehq/apps/motech/templates/motech/motech_openmrs_concepts.html
```diff
@@ -0,0 +1,47 @@
...
+{% endblock %}
+{% block page_content %}
+    {% registerurl 'all_openmrs_concepts' domain %}
+    {% registerurl 'openmrs_sync_concepts' domain %}
+    <div id="openmrs_concepts_template" class="ko-template">
+        <button class="btn btn-default" data-bind="
+            disable: syncButton.inProgress,
+            click: syncButton.sync
+        ">
+            <!--ko ifnot: syncButton.inProgress-->Sync from OpenMRS<!--/ko-->
```
benrudolph: not sure if these need to be translate, but flaggin in case

benrudolph (approved):

corehq/apps/motech/openmrs/concepts/sync.py
```diff
@@ -0,0 +1,112 @@
...
+            answer_concepts = OpenmrsConcept.objects.filter(account=account, uuid__in=answers).all()
+            if set(answer_concept.uuid for answer_concept in answer_concepts) == set(answers):
+                concept.answers = answer_concepts
+                concept.save()
+            else:
+                delayed.append((concept, answers))
+
+        if len(answers_relationships) == len(delayed):
+            # this is going to be an infinite loop
+            raise Exception(delayed)
```
proteusvacuum: nit: do you want to have a better named exception?

corehq/apps/motech/openmrs/concepts/sync.py
```diff
@@ -0,0 +1,112 @@
...
+    api = OpenmrsListApi(requests, 'concept')
+    answers_relationships = []
+    for concept in api.get_all():
+        concept = openmrs_concept_json_from_api_json(concept)
+        concept, answers = openmrs_concept_from_concept_json(account, concept)
+        if answers:
+            answers_relationships.append((concept, answers))
+        concept.save()
+
+    while answers_relationships:
```
proteusvacuum: I don't know what OpenMRS concepts are. Can you explain what this loop is doing? it looks to me as though it just looks in the database to see if all the `answers` are the same as the ones you received from the API. Is that correct? I might just be a bit slow this afternoon.

corehq/apps/motech/openmrs/patients/views.py
```diff
@@ -0,0 +1,43 @@
...
+            identifier['identifier'] == patient_id
+            and identifier['identifierType']['uuid'] == patient_identifier_type
+            for identifier in patient['identifiers']
+        )
+    ]
+    return HttpResponse(
+        json.dumps([{
+            'uuid': p['uuid'],
+            'display': p['display'],
+        } for p in patients]), content_type='text/json')
```
proteusvacuum: I think the content type for json is `application/json`.

corehq/apps/motech/openmrs/patients/views.py
```diff
@@ -0,0 +1,43 @@
...
+
+@require_motech_permissions
+def all_patient_identifier_types(request, domain):
+    requests = get_openmrs_requests_object(domain)
+    api = OpenmrsListApi(requests, 'patientidentifiertype')
+    patient_identifier_types = [
+        openmrs_patient_identifier_type_json_from_api_json(patient_identifier_type).to_json()
+        for patient_identifier_type in api.get_all()
+    ]
+    return HttpResponse(json.dumps(patient_identifier_types), content_type='text/json')
```
proteusvacuum: `application/json`

corehq/apps/motech/openmrs/patients/views.py
```diff
@@ -0,0 +1,43 @@
...
+@require_motech_permissions
+def search_patients(request, domain):
+    patient_identifier_type = request.GET.get('idtype')
+    patient_id = request.GET.get('id')
+    if not patient_id or not patient_identifier_type:
+        return HttpResponse(json.dumps({
+            'badrequest': 'Must include params idtype and id'
+        }), status=400)
+    requests = get_openmrs_requests_object(domain)
+    patients = requests.get('/openmrs/ws/rest/v1/patient', params={'q': patient_id, 'v': 'full'}).json()['results']
```
proteusvacuum: what if this fails?

corehq/apps/motech/openmrs/urls.py
```diff
@@ -0,0 +1,23 @@
...
+        views.all_openmrs_concepts,
+        name='all_openmrs_concepts'),
+    url(r'^rest/concept/search/$',
+        views.concept_search,
+        name='openmrs_concept_search'),
+    url(r'^rest/concept/sync/$',
+        views.sync_concepts,
+        name='openmrs_sync_concepts'),
+
+    # patients
```
proteusvacuum: are these ever called?

corehq/apps/motech/openmrs/urls.py
```diff
@@ -0,0 +1,23 @@
...
+from corehq.apps.motech.openmrs import views
+
+urlpatterns = [
+    # concepts
+    url(r'^rest/concept/$',
+        views.all_openmrs_concepts,
+        name='all_openmrs_concepts'),
+    url(r'^rest/concept/search/$',
+        views.concept_search,
+        name='openmrs_concept_search'),
```
proteusvacuum: where is this called from?

corehq/apps/motech/openmrs/concepts/views.py
dannyroberts: It doesn't; `JsonResponse` is a Django built-in that takes in a `dict` and returns a Django `HttpResponse` with the desired status code, a JSON body based on your dictionary, and content type set to the correct JSON mimetype (`application/json`). `.to_json` is a `JsonObject` method that converts from an instance of a `JsonObject` (subclass) to a JSON-serializable `dict` (not actually to JSON). I think it's a little confusing because at some point, possibly before Django's `JsonResponse` existed, we had defined a `json_response` helper that used our custom JSON handler that actually _does_ call the `.to_json()` method on the object-to-be-serialized, if it exists. This method is still used in some places, but I now prefer using the built-in `JsonResponse`. (Sorry for the characteristic over-explanation here!)

corehq/apps/motech/templates/motech/motech_openmrs_concepts.html
dannyroberts: That's fair. I was in kind of "this isn't really going to be live yet so it doesn't matter" mode, but you're right—it's way easier to do this kind of thing as you go along than it is to do them in bulk after the fact.

- [ ] Translate all easily-translatable strings in this PR's diff
- [ ] Add a todo comment next to any hard-to-translate string that should eventually be translated (if any).

corehq/apps/motech/openmrs/concepts/sync.py
dannyroberts: Concepts correspond to a few different things as we think of them in CommCare, including
- questions
- answers
- case properties (e.g. patient properties)
- case property values (e.g. patient property values)

A concept that is... _conceptually_... a "question", will be defined with "answers" which are other concepts.

The key lines in here is

```python
concept.answers = answer_concepts
concept.save()
```

Conceptually this is really just part of the the previous loop—for every concept that comes in you want to save the corresponding answers in the question & answer table, which is a many-to-many table from concepts (questions) to concepts (answers). But you can't save to that table until both the question and answer involved in a row has been saved. Hence, I defer populating that table until each concept has been saved.

corehq/apps/motech/openmrs/concepts/sync.py
dannyroberts: OK fine :)

- [ ] Raise a better exception than `Exception`

corehq/apps/motech/openmrs/patients/views.py
dannyroberts: True

corehq/apps/motech/openmrs/urls.py
dannyroberts: I think you're right and I can just get rid of the whole `patients` portion. It's copied from an earlier prototype that did use it, but I can just add it back once it becomes a thing.

corehq/apps/motech/openmrs/urls.py
dannyroberts: It's not actually called here yet, but it's setting up an endpoint that pages can use in an async call. Tempted to just leave it here, since I'm planning to use it soon.

corehq/apps/motech/openmrs/concepts/sync.py
dannyroberts: Actually I think this approach isn't necessary at all—at one point I think I was trying to get each concept to save only once, so I basically had to feel out the dependency graph using that method. Now that they're all saved up front and then they're re-saved after adding answers (which may only actually update the concept-to-answer table) I think it's no longer necessary.

dannyroberts:
@proteusvacuum updated

proteusvacuum:
Motech! Gotech!

PR #15784, opened 2017-04-17
Add seconds to repeater report

proteusvacuum:
@benrudolph @dannyroberts


benrudolph (approved):
getting down to the nitty gritty!

PR #15785, opened 2017-04-17
Account for other column sizes

benrudolph:
failed to account for the fact that other screen sizes have different number of columns

### DIFF
--- a/corehq/apps/cloudcare/templates/formplayer/menu_list.html
+++ b/corehq/apps/cloudcare/templates/formplayer/menu_list.html
@@ -71,7 +71,11 @@
         </div>
     </div>
     <% if (menuIndex % 2 === 1) { %>
-    <div class="clearfix"></div>
+      <div class="clearfix visible-xs"></div>
+    <% } else if (menuIndex % 3 === 2) { %>
+      <div class="clearfix visible-sm"></div>
+    <% } else if (menuIndex % 4 === 3) { %>
+      <div class="clearfix visible-lg"></div>
     <% } %>
 </script>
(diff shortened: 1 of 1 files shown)


corehq/apps/cloudcare/templates/formplayer/menu_list.html
```diff
@@ -71,7 +71,11 @@ <h1 class="page-title"><%- title %></h1>
         </div>
     </div>
     <% if (menuIndex % 2 === 1) { %>
-    <div class="clearfix"></div>
+      <div class="clearfix visible-xs"></div>
```
proteusvacuum: why wouldn't you just have a `<div class="clearfix">` for everything? (i.e. without the if statements)

proteusvacuum (approved):

corehq/apps/cloudcare/templates/formplayer/menu_list.html
benrudolph: clearfix forces a clean break in the columns. so if you have a xs screen size, with normal column layout, you get:
```
A B
C D
```
However if the A box has a higher height than the B box (happens when there is more text in the module) then you get:
```
A B
  C
D
```
Which is annoying. the way to fix it, is to add a clear fix element after the first two elements. or after the first 3, or the first 4. if you were to do it after every item, your screen would look like:
```
A
B
C
D
```

proteusvacuum: Aha! Thanks for the explanation.

PR #15788, opened 2017-04-18
for legacy forms, check if open case exists as an attribute in case actions

biyeun:
@orangejenny 
fix for ? https://manage.dimagi.com/default.asp?252177#1318653


orangejenny (approved):

orangejenny:
thank you for fixing

PR #15790, opened 2017-04-18
Pull out all case names from the extension cases test json

proteusvacuum:
Because @ctsims  made me feel bad it didn't do this already.

@dannyroberts @benrudolph


benrudolph (approved):
curious how this even came up

PR #15792, opened 2017-04-18
Add no_result_task

gcapalbo:
Adds a decorator to define a task which won't create any records in celery_taskmeta, by specifying that the DisabledBackend results backend be used for that task.

@dimagi/scale-team 
buddy @NoahCarnahan


emord (approved):

benrudolph (approved):
nice

PR #15800, opened 2017-04-19
NIC gateway: catch invalid number errors

gcapalbo:
https://manage.dimagi.com/default.asp?250476

@emord 
buddy @NoahCarnahan

### DIFF
--- a/corehq/messaging/smsbackends/icds_nic/models.py
+++ b/corehq/messaging/smsbackends/icds_nic/models.py
@@ -13,6 +13,8 @@ from dimagi.utils.logging import notify_exception
 
 ERR_INVALID_DESTINATION = '-410'
 
+INDIA_COUNTRY_CODE = '91'
+
 ERROR_CODES = {
     "-2": "Invalid credentials",
     "-3": "Empty mobile number",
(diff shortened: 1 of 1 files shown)


corehq/messaging/smsbackends/icds_nic/models.py
```diff
@@ -84,6 +88,12 @@ def get_generic_name(cls):
     def get_form_class(cls):
         return ICDSBackendForm
 
+    def destination_number_is_valid(self, phone_number):
+        """
+        phone_number is not expected to contain the leading + for this validation to work.
+        """
+        return phone_number.startswith(INDIA_COUNTRY_CODE) and phone_number != INDIA_COUNTRY_CODE
```
emord: Would a regex be more concise here? `/\+91.*/` or something similar?

emord (approved):

corehq/messaging/smsbackends/icds_nic/models.py
gcapalbo: not sure, i kind of like the readability this way. the main issue is that there are a lot of numbers that aren't filled in and only consist of just the country code, so i wanted to be explicit that we're ignoring those

PR #15801, opened 2017-04-19
don't use insert for export sort

benrudolph:
sometimes the dumbest mistakes are the hardest to find...

@calellowitz


corehq/apps/export/tests/test_export_data_schema.py
```diff
@@ -1145,8 +1145,7 @@ def _create_schema(self, items):
             group_schemas=[
                 ExportGroupSchema(
                     path=[],
-                    items=[
-                    ]
+                    items=items,
```
benrudolph: when do we start writing tests for our tests? 😳

PR #15802, opened 2017-04-19
preserve user group ids and group names in ES on user save

dannyroberts:
previously saving the user would wipe its group membership by excluding __group_ids and __group_names

Fixes https://manage.dimagi.com/default.asp?251023.

I should test this on staging to make sure it works the way I think it does, but what do you think of the approach? The other option is to pull these from what's currently in elasticsearch, but pulling from source of truth seems better to me, even if it's somewhat redundant.

### DIFF
--- a/corehq/apps/userreports/tests/test_choice_provider.py
+++ b/corehq/apps/userreports/tests/test_choice_provider.py
@@ -232,6 +232,7 @@ class UserChoiceProviderTest(SimpleTestCase, ChoiceProviderTestMixin):
         return user
 
     @classmethod
+    @mock.patch('corehq.apps.groups.models.Group.by_user', mock.Mock(return_value=[]))
     def setUpClass(cls):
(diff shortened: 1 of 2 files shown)


benrudolph (approved):
this seems like the right and in the most foolproof place to do it to me

biyeun (approved):

biyeun:
@dannyroberts build fail potentially real?

dannyroberts:
yeah, pushed a fix, but there may be other test failures. Basically this adds a db call where there wasn't one before, which incidentally affects `SimpleTestCase`s. Easy solution is to mock patch them to have `Group.by_user` just always return `[]` for those tests that aren't specifically testing that functionality.

dannyroberts:
I've also discovered that there's no mechanism to _remove_ group membership from users in ES when a group saved after removing a user. But tested that the thing I was trying to fix is now fixed on staging, so I'm going to merge.

PR #15803, opened 2017-04-19
Revert "Optimize Forms by Submission"

biyeun:
Reverts dimagi/commcare-hq#15646

Likely causing issues described here https://manage.dimagi.com/default.asp?251852
@proteusvacuum // @dannyroberts


benrudolph:
don't think this is necessary anymore

PR #15804, opened 2017-04-19
Only send to users in Andhra Pradesh in first rollout

gcapalbo:
https://manage.dimagi.com/default.asp?251379

@emord 
buddy @NoahCarnahan

### DIFF
--- a/custom/icds/tasks.py
+++ b/custom/icds/tasks.py
@@ -111,6 +111,8 @@ def run_weekly_indicators():
 
         for user_id in generate_user_ids_from_primary_location_ids(domain, get_awc_location_ids(domain)):
             language_code = get_language_code(user_id, telugu_user_ids, marathi_user_ids)
+            if language_code != TELUGU:
+                continue
 
             if first_week_of_month_result:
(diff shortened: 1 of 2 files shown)


custom/icds/tasks.py
```diff
@@ -111,6 +111,8 @@ def run_weekly_indicators():
 
         for user_id in generate_user_ids_from_primary_location_ids(domain, get_awc_location_ids(domain)):
             language_code = get_language_code(user_id, telugu_user_ids, marathi_user_ids)
+            if language_code != TELUGU:
```
emord: Would it be possible to use the location instead of language code? Will all states have different languages?

gcapalbo: yeah only Andhra Pradesh will use Telugu so this is fine for now (we derive Telugu language code by checking state location), and based on the rollout plan it seemed like the next step will be the Hindi speaking states so we can probably just keep this simple and use language code as the way to do our rollout

emord (approved):

emord (approved):

PR #15770, opened 2017-04-14
asyncindicator

emord:
One bug fix and a better admin page

@kaapstorm

PR #15807, opened 2017-04-19
async queues for ucr

emord:
This will enable asynchronous processing on the two child case monthly data sources should be merged after https://github.com/dimagi/commcare-hq/pull/15770

Want to hold off until we're sure that celery and postgres machines are not going to go down

@millerdev


millerdev (approved):

emord:
:shipit:

PR #15809, opened 2017-04-19
Duplicate scripts

orangejenny:
Noticed a couple of places where we're including `main.js` in templates that already inherit from `style/base.html`

@dannyroberts


dannyroberts (approved):
nice

PR #15818, opened 2017-04-20
eNikshay custom reports: Display location restricted users' locations in location filter

NoahCarnahan:
When a user loads a custom eNikshay report with a location filter, if that user is location restricted, their location will be displayed as a selected value in that report. (The report is already filtered by the users location, but previously there had been no visible indication of this).
@esoergel 
buddy @gcapalbo

### DIFF
--- a/custom/enikshay/reports/filters.py
+++ b/custom/enikshay/reports/filters.py
@@ -33,7 +33,8 @@ class EnikshayLocationFilter(BaseMultipleOptionFilter):
         """
         location_ids = self.request.GET.getlist(self.slug)
 
-        if len(location_ids) == 0 and not self.request.couch_user.has_permission(self.request.domain, 'access_all_locations'):
+        if len(location_ids) == 0 \
+                and not self.request.couch_user.has_permission(self.request.domain, 'access_all_locations'):
             # Display the user's location in the filter if none is selected
             location_ids = self.request.couch_user.get_location_ids(self.request.domain)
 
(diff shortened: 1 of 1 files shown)


custom/enikshay/reports/filters.py
```diff
@@ -32,6 +32,11 @@ def selected(self):
         :return: Selected locations without their descendants
         """
         location_ids = self.request.GET.getlist(self.slug)
+
+        if len(location_ids) == 0 and not self.request.couch_user.has_permission(self.request.domain, 'access_all_locations'):
```
dimagimon: E501 line too long (126 > 115 characters)

dimagimon: E501 line too long (126 > 115 characters)

NoahCarnahan: dimagimon is confused, this is not what the diff looks like any more

custom/enikshay/reports/filters.py
```diff
@@ -32,6 +32,12 @@ def selected(self):
         :return: Selected locations without their descendants
         """
         location_ids = self.request.GET.getlist(self.slug)
+
+        if len(location_ids) == 0 \
+                and not self.request.couch_user.has_permission(self.request.domain, 'access_all_locations'):
+            # Display the user's location in the filter if none is selected
+            location_ids = self.request.couch_user.get_location_ids(self.request.domain)
+
```
esoergel: Nice, I like this idea.  `request.can_access_all_locations` is also available as long as it's a real request (not like, a test mock)

PR #15819, opened 2017-04-20
Make new user number easy to disable

dannyroberts:
and disable it

quick response to https://manage.dimagi.com/default.asp?252377

in the future to enable just change

```
NewUserNumberAbTestMixin = NewUserNumberAbTestMixin__Disabled
```

back to

```
NewUserNumberAbTestMixin = NewUserNumberAbTestMixin__Enabled
```

For context I'm doing this because it's already been added, reverted, and then re-added (see https://github.com/dimagi/commcare-hq/pull/14557). The chance that there's some more back and forth on this reasonable, and this is a tiny little gift to whoever eventually enables it again.

### DIFF
--- a/corehq/apps/registration/views.py
+++ b/corehq/apps/registration/views.py
@@ -79,7 +79,7 @@ class NewUserNumberAbTestMixin__Disabled(object):
         pass
 
 
-NewUserNumberAbTestMixin = NewUserNumberAbTestMixin__Enabled
+NewUserNumberAbTestMixin = NewUserNumberAbTestMixin__Disabled
 
 
 class ProcessRegistrationView(JSONResponseMixin, NewUserNumberAbTestMixin, View):
--- a/corehq/apps/registration/views.py
+++ b/corehq/apps/registration/views.py
@@ -48,7 +48,7 @@ def registration_default(request):
     return redirect(UserRegistrationView.urlname)
 
 
-class NewUserNumberAbTestMixin(object):
+class NewUserNumberAbTestMixin__Enabled(object):
     @property
     @memoized
     def _ab(self):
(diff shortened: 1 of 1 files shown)


dannyroberts:
@calellowitz how does this look? I tested the page (/register/user) after each commit, and tested that the re-enabled step (not committed here) will work. cc: @orangejenny buddy

orangejenny (approved):

calellowitz (commented):
In general, looks good and I would be fine merging. My one concer is that I think its pretty unlikely it will be enabled again. I think we did an a/b test and it failed. Tried another and it failed again. So I doubt we will want this again and there is a lot of code that may just sit around. May be cleaner to just kill all this code, as its easy enough to re add. 
Additionally, if what you are going for is an easy switch to turn the A/B test on and off, what we have done in the past is just set both options to off, so we didnt need to change view code. Then when we wanted to turn it on we just changed one of the options to showing the phone number. However, I dont have a strong opinion over whether that solution is better than this, so if you prefer this solution that is fine by me.

biyeun:
@calellowitz I think we should revert and remove the A/B test code for the sake of cleanliness and simplicity.

dannyroberts:
Okay, I was working off of what Audrey said in the ticket

> (I hope it does not take too much time to remove: we might have to put it back in a few weeks, but for now we don't have enough bandwith to use it and it is decreasing the number of sign-ups without the additional benefit of using the phone nummber). 
>
> It might be quicker to keep the 2 A/B group - even if they see the exact same page, so that we can re-use that structure when we want to re-launch the test?

biyeun:
@dannyroberts ah ok, then in that case I'll keep this PR in favor of https://github.com/dimagi/commcare-hq/pull/15836

calellowitz (approved):

PR #15820, opened 2017-04-20
Show case list missing warning when appropriate

orangejenny:
https://manage.dimagi.com/default.asp?232239

@dannyroberts

### DIFF
--- a/corehq/apps/app_manager/views/modules.py
+++ b/corehq/apps/app_manager/views/modules.py
@@ -111,7 +111,7 @@ def get_module_view_context(app, module, lang=None):
         'lang': lang,
         'langs': app.langs,
         'module_type': module.module_type,
-        'requires_case_details': bool(module.requires_case_details),
+        'requires_case_details': bool(module.requires_case_details()),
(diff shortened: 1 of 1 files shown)


corehq/apps/app_manager/views/modules.py
```diff
@@ -111,7 +111,7 @@ def get_module_view_context(app, module, lang=None):
         'lang': lang,
         'langs': app.langs,
         'module_type': module.module_type,
-        'requires_case_details': bool(module.requires_case_details),
+        'requires_case_details': bool(module.requires_case_details()),
```
dannyroberts: Wonder why the `bool` is needed at all. My guess is in the original implementation it was 500ing because there was this method value where a boolean value was expected, and the author interpreted that as `requires_case_details` returning something weird, casted it to a `bool` (!), saw it worked, and called it a day. And so I bet the `bool`'s not necessary. Might have to look at the commit history of that line to confirm that story or replace it with a better one, but up to you whether you think it's worth it.

PR #15828, opened 2017-04-20
throttle errors sent to sentry

snopoke:
budy @nickpell


nickpell (approved):

corehq/util/sentry.py
```diff
@@ -0,0 +1,42 @@
...
+        return RATE_LIMITED_EXCEPTIONS[exc_name]
+
+
+def _is_rate_limited(rate_limit_key):
+    exponential_backoff_key = '{}_down'.format(rate_limit_key)
+    ExponentialBackoff.increment(exponential_backoff_key)
+    if ExponentialBackoff.should_backoff(exponential_backoff_key):
+        datadog_counter('commcare.sentry.errors.rate_limited', tags=[
+            'service:{}'.format(rate_limit_key)
+        ])
```
benrudolph: love it

corehq/util/sentry.py
```diff
@@ -0,0 +1,42 @@
+from raven.contrib.django import DjangoClient
+
+from corehq.util.cache_utils import ExponentialBackoff
+from corehq.util.datadog.gauges import datadog_counter
+
+RATE_LIMITED_EXCEPTIONS = {
+    'botocore.vendored.requests.packages.urllib3.exceptions.ProtocolError': 'riak',
+    'OperationalError': 'postgres',  # could be psycopg2._psycopg or django.db.utils
+    'socket.error': 'rabbitmq',
+    'redis.ConnectionError': 'redis'
```
benrudolph: nice

benrudolph (approved):

corehq/util/sentry.py
dannyroberts: nit: this counts how many times `_is_rate_limited` is called, but the function has a name that implies no side effects. I might have put this in the function that calls `_is_rate_limited` instead.

corehq/util/sentry.py
```diff
@@ -0,0 +1,42 @@
...
+    ExponentialBackoff.increment(exponential_backoff_key)
+    if ExponentialBackoff.should_backoff(exponential_backoff_key):
+        datadog_counter('commcare.sentry.errors.rate_limited', tags=[
+            'service:{}'.format(rate_limit_key)
+        ])
+        return True
+    return False
+
+
+class HQSentryClient(DjangoClient):
```
dannyroberts: brilliant

corehq/form_processor/submission_post.py
```diff
@@ -163,9 +162,6 @@ def run(self):
                 if instance.xmlns == DEVICE_LOG_XMLNS:
                     try:
                         process_device_log(self.domain, instance)
-                    except ProtocolError:
-                        # if riak is down/struggling, the caller will catch and notify
-                        raise
```
dannyroberts: doesn't this still mean that the `notify_exception` below gets called, which calls sentry, which won't catch it because it's not a raised error? Or does `notify_exception` still end up passing in `exc_info` to `should_capture` properly below? I'm now guessing that that's the case, but I'll leave this comment to make sure that was the reasoning

dannyroberts (commented):
Awesome, this is a much better approach

corehq/form_processor/submission_post.py
snopoke: Good question, I checked and as long as `notify_exception` is called within an `except` block then the `exec_info` should get passed through (also tested locally)

dannyroberts: Nice. Thanks!

PR #15829, opened 2017-04-20
use datadog_counter to log repeater errors

proteusvacuum:
@emord 

cc: @sravfeyn

### DIFF
--- a/corehq/apps/repeaters/models.py
+++ b/corehq/apps/repeaters/models.py
@@ -9,7 +9,7 @@ from requests.exceptions import Timeout, ConnectionError
 from corehq.apps.cachehq.mixins import QuickCachedDocumentMixin
 from corehq.form_processor.exceptions import XFormNotFound
 from corehq.util.datadog.metrics import REPEATER_ERROR_COUNT
-from corehq.util.datadog.utils import log_counter
+from corehq.util.datadog.gauges import datadog_counter
 from corehq.util.quickcache import quickcache
 from utils import get_repeater_auth_header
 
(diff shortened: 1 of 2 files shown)


corehq/apps/repeaters/models.py
```diff
@@ -608,10 +608,13 @@ def _fail(self, reason, response):
             self.last_checked = datetime.utcnow()
             self.cancel()
         self.failure_reason = reason
-        log_counter(REPEATER_ERROR_COUNT, {
-            '_id': self._id,
-            'reason': reason,
-            'target_url': self.url,
+        datadog_counter(REPEATER_ERROR_COUNT, {
```
sravfeyn: Curious if there will be lot of these or if there are any datadog limits/throttling

sravfeyn (approved):

corehq/util/datadog/metrics.py
```diff
@@ -1,5 +1,5 @@
 JSERROR_COUNT = 'jserror.count'
 ERROR_COUNT = 'error.count'
-REPEATER_ERROR_COUNT = 'repeater.error.count'
+REPEATER_ERROR_COUNT = 'commcare.repeater.error.count'
```
emord: @proteusvacuum It looks like there's a metric called `commcare.repeaters.overdue` could we change this to be `commcare.repeaters.error_count` or something similar so they can have the same prefix?

emord: Also I think error_count may be redundant. could probably just be commcare.repeaters.error

emord (approved):

PR #15539, opened 2017-03-27
Print case detail

orangejenny:
https://docs.google.com/document/d/18SpmZDc5GvFR2VcYY4KtevIgQQ1aHqDsD0rYMbZLTR0

This code feels atrocious, but it's possible that's just how all multimedia uploader code is.

It also unfortunately reintroduces inline js into `module_view.html`, which I'll deal with at a later date, because it's not straightforward to define `printTemplateUploader` using `initial_page_data`.

Easiest to review all at once.

@NoahCarnahan 
@biyeun can you also look this over? I think you've got the most expertise with the multimedia uploader.

PR #15830, opened 2017-04-20
Add print template for case detail to multimedia suite

orangejenny:
https://github.com/dimagi/commcare-hq/pull/15539#issuecomment-290354030

Verified that the template now has a reference in `media_suite.xml` and that the html file appears in `commcare/text`.

@dannyroberts 

fyi @amstone326


dannyroberts (approved):
Looks good!

PR #15833, opened 2017-04-21
prevent dictionary lookup by passing in tuples

snopoke:
https://sentry.io/dimagi/commcarehq/issues/239319451/

Figured this one out - Django does dictionary lookups in templates so `get.items` was returning the value of the `items` key in the dict and not the list of tuples.

https://docs.djangoproject.com/en/1.10/ref/templates/api/#variables-and-lookups


snopoke:
patched to prod as well - going to merge

PR #15834, opened 2017-04-21
turn off blob restore for now

snopoke:
https://manage.dimagi.com/default.asp?252506


wpride (approved):

benrudolph:
thanks, starting to think this is caused from restores that used to use the FileRestoreResponse, and then switched to the new BlobRestore and still have the old file path to the restore cache. will investigate

PR #15837, opened 2017-04-21
claim as user

proteusvacuum:
Case claim was broken when using login as.
This adds a param `commcare_login_as` to the request when a user is acting as someone else.
https://manage.dimagi.com/default.asp?252448#1321156

@wspride 
cc: @sravfeyn


proteusvacuum:
requires: https://github.com/dimagi/formplayer/pull/368

wpride:
@proteusvacuum :man_student: man stud

PR #15839, opened 2017-04-21
Shadow Forms

NoahCarnahan:
This PR adds a new type of form called a "Shadow Form". A Shadow form inherits its form xml from an advanced form, but can define its own case actions. The shadow form can introduce totally new actions, or modify existing ones.

Note that this feature is *not* wholly implemented for app manager v2. The UI for choosing the type of the form when adding a new one to a module was not specified, and this feature is needed ASAP for a project, so I would prefer to to wait until that component has been designed.

Best reviewed commit-by-commit
buddy @gcapalbo


corehq/apps/app_manager/models.py
```diff
@@ -2974,6 +2976,106 @@ def update_app_case_meta(self, app_case_meta):
                 meta.add_closer(self.unique_id, action.close_condition)
 
 
+class ShadowForm(AdvancedForm):
+    form_type = 'shadow_form'
+    shadow_parent_form_id = StringProperty(required=False, default=None)  # The unqiue id of the form we are shadowing.
```
dimagimon: E501 line too long (119 > 115 characters)

gcapalbo (approved):
looks good to me. is there anyone else you'd like to review that has more context?

corehq/apps/app_manager/models.py
```diff
@@ -2989,6 +2991,108 @@ def update_app_case_meta(self, app_case_meta):
...
+
+    # form actions to be merged with the parent actions
+    extra_actions = SchemaProperty(AdvancedFormActions)
+
+    @property
+    def shadow_parent_form(self):
+        if not self.shadow_parent_form_id:
+            return None
+        else:
+            app = self.get_app()
```
orangejenny: Would it be less expensive to use `FormBase.get_form` instead?

corehq/apps/app_manager/models.py
```diff
@@ -2989,6 +2991,108 @@ def update_app_case_meta(self, app_case_meta):
...
+            return app.get_form(self.shadow_parent_form_id)
+
+    @property
+    def source(self):
+        if self.shadow_parent_form_id:
+            return self.shadow_parent_form.source
+        from corehq.apps.app_manager.views.utils import get_blank_form_xml
+        return get_blank_form_xml("Untitled Form")
```
orangejenny: trans

corehq/apps/app_manager/templates/app_manager/v1/managed_app.html
```diff
@@ -175,14 +175,36 @@
                                 {% endfor %}
                             {% endwith %}
                             <li class="sort-disabled">
-                                <form action="{% url "new_form" domain app.id module.id %}"
+                                <form id="new-form-form-{{ module.id }}" action="{% url "new_form" domain app.id module.id %}"
```
orangejenny: nitpick, but `form-form` is a weird part of a name

orangejenny:
It's a pretty edgy edge case, but you might want to update https://github.com/dimagi/commcare-hq/blob/master/corehq/apps/app_manager/templates/app_manager/v1/partials/form_tab_settings.html#L73-L89 to disable the `no_vellum` checkbox if the form is a shadow form.

corehq/apps/app_manager/models.py
```diff
@@ -3012,6 +3012,12 @@ def actions(self):
             shadow_parent_actions = self.shadow_parent_form.actions
         return self._merge_actions(shadow_parent_actions, self.extra_actions)
 
+    def get_shadow_parent_options(self):
```
orangejenny: Would be nice for this to also handle the case where the form has a shadow parent but it isn't in `app.get_forms` giving it a name like `Unknown, please change`. Someone's going to set up a shadow form, delete the parent form accidentally, and then have trouble fixing the build error.

corehq/apps/app_manager/templates/app_manager/v1/partials/shadow_parent_select.html
```diff
@@ -1,10 +1,21 @@
...
-</head>
-<body>
+{% load i18n %}
+{% load hq_shared_tags %}
 
-</body>
-</html>
+<div class="form-group" xmlns="http://www.w3.org/1999/html">
+    <label class="control-label col-sm-2">
+        {% trans "Shadow parent" %}
```
orangejenny: Nomenclature: shadow modules uses `source` not `parent`, does it matter if this matches? We use `parent` for a lot of things. Although I suppose forms already have a `source`, too.

corehq/apps/app_manager/models.py
```diff
@@ -2989,6 +2991,108 @@ def update_app_case_meta(self, app_case_meta):
...
+
+    @staticmethod
+    def _merge_actions(source_actions, extra_actions):
+        new_actions = OrderedDict(
+            (action.case_tag, LoadUpdateAction.wrap(action.to_json()))
+            for action in source_actions.load_update_cases
+        )
+        for action in extra_actions.load_update_cases:
+            new_action = new_actions.get(action.case_tag, LoadUpdateAction(case_tag=action.case_tag))
+            overwrite_properties = [
```
orangejenny: You've probably thought about this already, as new properties are added to actions in the future, is it likely that they'll also need to be added to this list? Just wondering whether it would make sense to make the hard-coded list the properties *not* to overwrite.

corehq/apps/app_manager/templates/app_manager/v1/form_view_base.html
```diff
@@ -318,17 +322,20 @@
...
-        </a>
-        {% else %}
-        <button class="btn btn-primary disabled" disabled="disabled"
-                title="{% trans "Your administrator has locked this form from edits through the form builder"|force_escape %}">
-            <i class="fa fa-pencil"></i>
-            {% trans "Edit" %}
-        </button>
+        {% if not form.form_type == "shadow_form" %}
+            {% if not form.no_vellum %}
+            <span>{{ form.no_vellum }}</span>
```
orangejenny: Is this supposed to be here?

corehq/apps/app_manager/templates/app_manager/v1/partials/form_tab_advanced.html
```diff
@@ -2,6 +2,7 @@
 {% load xforms_extras %}
 {% load hq_shared_tags %}
 <div class="tab-pane" id="advanced">
+    {% if form.form_type != "shadow_form" %}
```
orangejenny: This should be moved to hide the entire advanced tab for shadow forms (since it looks like `allow_form_copy` is only advanced forms, which will make this tab always blank for shadow forms).

corehq/apps/app_manager/suite_xml/post_process/workflow.py
```diff
@@ -461,7 +461,7 @@ def get_case_type_created_by_form(form, target_module):
                 return form.get_module().case_type
             else:
                 return reg_action.case_type
-        elif form.form_type == 'advanced_form':
+        elif form.form_type == 'advanced_form' or form.form_type == "shadow_form":
```
orangejenny: Maybe I'm not clear on what this code does, but I thought shadow forms don't have reg actions, is this needed?

orangejenny:
Looks good, just minor comments.

corehq/apps/app_manager/models.py
NoahCarnahan: Ah didn't realize that method exists. Although looks like internally `FormBase.get_form()` fetches the app as well, plus it does another query on couch to check if the id is unique, so getting the app directly will be slightly faster.

corehq/apps/app_manager/models.py
NoahCarnahan: I think I should actually just remove that string, as this empty form xml should never be rendered or shown to the user. (It's just a placeholder until the user selects a shadow parent)

corehq/apps/app_manager/templates/app_manager/v1/partials/shadow_parent_select.html
NoahCarnahan: I was using "parent" to avoid the "source" collision, but I think for the UI it's better to say "source".

corehq/apps/app_manager/models.py
NoahCarnahan: I hadn't thought too hard about it, but I think I'm ok with the white list because I think it might be safer to have someone make an explicit choice about the new property being included in shadows.

corehq/apps/app_manager/templates/app_manager/v1/form_view_base.html
NoahCarnahan: whoops!

corehq/apps/app_manager/suite_xml/post_process/workflow.py
NoahCarnahan: They can't have their "own" registration actions, but they do inherit them from their source form, so I think this is still needed.

NoahCarnahan:
Thank you both for the review. Just pushed some updates.

corehq/apps/app_manager/models.py
orangejenny: ha, bad intuition on my part, thanks for investigating

corehq/apps/app_manager/models.py
orangejenny: makes sense

orangejenny (approved):

PR #15841, opened 2017-04-21
Allow retries for failing 99dots records

proteusvacuum:
We've been running into issues where their server is 400'ing for no reason. For now, retry these records later to try again later until we figure out why it is happening.


proteusvacuum:
@NoahCarnahan

PR #15842, opened 2017-04-21
Revert "Optimize Forms by Submission"

biyeun:
Reverts dimagi/commcare-hq#15646


benrudolph:
(only until we reindex users to fix the way `__group_ids` is set on the user

biyeun:
@benrudolph on it https://manage.dimagi.com/default.asp?252552

PR #15844, opened 2017-04-21
Make worker activity report location safe (if flag is enabled)

NoahCarnahan:
@esoergel 
cc @gcapalbo


corehq/apps/reports/standard/monitoring.py
```diff
@@ -1461,8 +1470,18 @@ def get_admins_and_demo_users(self):
     @property
     def users_to_iterate(self):
         if toggles.EMWF_WORKER_ACTIVITY_REPORT.enabled(self.request.domain):
-            user_query = EMWF.user_es_query(self.domain, self.request.GET.getlist(EMWF.slug))
-            return util.get_simplified_users(user_query)
+            user_query = LocationRestrictedMobileWorkerFilter.user_es_query(
+                self.domain, self.request.GET.getlist(LocationRestrictedMobileWorkerFilter.slug)
+            )
+            users = util.get_simplified_users(user_query)
```
gcapalbo: could you move these 4 lines to an else clause below? seems like `users` can be overridden below and we could save a hit to es by not running these

gcapalbo:
@NoahCarnahan just one comment, also looks like there are conflicts

corehq/apps/reports/standard/monitoring.py
```diff
@@ -1461,8 +1470,18 @@ def get_admins_and_demo_users(self):
...
+            if not self.request.couch_user.has_permission(self.domain, 'access_all_locations'):
+                accessible_location_ids = (SQLLocation.active_objects.accessible_location_ids(
+                    self.request.domain,
+                    self.request.couch_user)
+                )
+                accessible_user_ids = set(user_ids_at_locations(accessible_location_ids))
+                users = [u for u in users if u.user_id in accessible_user_ids]
```
esoergel: `user_query` above is an unevaluated ESQuery instance, so you could chain this filter logic in there, something like:
```python
if not has_permission:
  user_query = user_query.filter(accessible_location_ids)
```
which would let ES handle the filtering

NoahCarnahan:
@gcapalbo @esoergel fixed to filter in ES

esoergel (approved):

corehq/apps/reports/standard/monitoring.py
```diff
@@ -1473,15 +1473,14 @@ def users_to_iterate(self):
...
             if not self.request.couch_user.has_permission(self.domain, 'access_all_locations'):
                 accessible_location_ids = (SQLLocation.active_objects.accessible_location_ids(
                     self.request.domain,
                     self.request.couch_user)
                 )
-                accessible_user_ids = set(user_ids_at_locations(accessible_location_ids))
-                users = [u for u in users if u.user_id in accessible_user_ids]
-            return users
+                accessible_user_ids = user_ids_at_locations(accessible_location_ids)
+                user_query.filter(accessible_user_ids)
```
gcapalbo: do you know how big `accessible_user_ids` can be? i don't think this will perform well if we're passing large lists of user ids to es

NoahCarnahan: ah right. Could be quite long (like could be every user in the domain) . Do you think it's better to move the filtering back to python, or perhaps filter in python only if length of users is over a certain length? Any idea what threshold should be if I went that route?

corehq/apps/reports/standard/monitoring.py
```diff
@@ -1453,7 +1461,16 @@ def get_admins_and_demo_users(self):
...
+            user_query = LocationRestrictedMobileWorkerFilter.user_es_query(
+                self.domain, self.request.GET.getlist(LocationRestrictedMobileWorkerFilter.slug)
+            )
+            if not self.request.couch_user.has_permission(self.domain, 'access_all_locations'):
+                accessible_location_ids = (SQLLocation.active_objects.accessible_location_ids(
+                    self.request.domain,
+                    self.request.couch_user)
+                )
+                accessible_user_ids = user_ids_at_locations(accessible_location_ids)
+                user_query.filter(accessible_user_ids)
```
esoergel: Oh right, just saw @gcapalbo 's comment - I think you can drop the `user_ids_at_locations` call here and just do:
```python
user_query = user_query.location(accessible_location_ids)
```
The `filter` method you're using here expects a json filter, and it (along with the other filter methods) doesn't mutate the query - it returns a new query, so you need to assign it to something.

NoahCarnahan: 👍 updated: https://github.com/dimagi/commcare-hq/pull/15844/commits/ae61bf441028009416cde1144a816bc1214a2a38

corehq/apps/reports/standard/monitoring.py
NoahCarnahan: Going to go with filtering by locations which should be much shorter: https://github.com/dimagi/commcare-hq/pull/15844/commits/ae61bf441028009416cde1144a816bc1214a2a38

The length of this could still be as large as there are locations in the domain though.

gcapalbo (approved):

PR #15847, opened 2017-04-21
Cache fixture items by datatype

emord:
@dimagi/scale-team 

This caches the data items. I think this is sufficient as none of the fixtures use fixture ownership

### DIFF
--- a/corehq/apps/fixtures/dbaccessors.py
+++ b/corehq/apps/fixtures/dbaccessors.py
@@ -27,7 +27,7 @@ def get_fixture_data_types_in_domain(domain):
     ))
 
 
-@quickcache(['domain', 'data_type_id'], timeout=60, skip_arg='bypass_cache')
+@quickcache(['domain', 'data_type_id'], timeout=5 * 60, skip_arg='bypass_cache')
 def get_fixture_items_by_data_type(domain, data_type_id, bypass_cache=False):
     from corehq.apps.fixtures.models import FixtureDataItem
     return list(FixtureDataItem.view(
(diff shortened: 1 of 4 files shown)


corehq/apps/fixtures/dbaccessors.py
```diff
@@ -27,6 +27,18 @@ def get_fixture_data_types_in_domain(domain):
     ))
 
 
+@quickcache(['domain', 'data_type_id'], timeout=60, skip_arg='bypass_cache')
+def get_fixture_items_by_data_type(domain, data_type_id, bypass_cache=False):
```
snopoke: This timeout feels short, thinking something like 5 minutes at least.

emord: Changed it. I was a worried that it would affect other domains negatively, but 5 should be fine

benrudolph: this also worries me a bit. maybe we should make a feature flag? i already know this is going to confuse people using app preview

emord: Actually thinking about it do you know if the skip_arg busts the cache? if so then it should be fine because when you update items it passes the bypass_cache=True

benrudolph: oh nice that would work. without looking into it, i'm not sure off the top of my head. it would make sense if that was the case though

emord: Yeah. I think that's how it works https://github.com/dimagi/quickcache/blob/master/quickcache/quickcache_helper.py#L147 nothing to worry about

snopoke:
Yes, quickcache clears and recaches when you specify the skip arg. So this
should recache anytime the UI is loaded which I think is an OK mechanism
for busting the cache.

PR #15848, opened 2017-04-22
Yet another UCR backend

emord:
@dimagi/scale-team Realized that we don't actually need to wait for refactoring the aggregation scripts to start using ES for restores. This is a new backend that will write to ES and SQL (like the laboratory) but it defaults to using ES for reads.

This should allow us to move the read load over to ES data source by data source. I also need to write something that compares UCRs generated on mobile with the compare_ucr task. Right now that only works for web reports. (it will be rate limited so celery doesn't get hammered)


benrudolph (approved):
sweet 🍬

snopoke (approved):
Nice. So this will move the load from mobile reports from SQL to ES. Seems good to me and gives us a way to scale those horizonally which is what we need. Will be interesting to see how ES copies with the load.

emord:
Added a commit to compare reports based on mobile ucr requests. This does it once in 10,000 reports which because icds has 141 reports per restore (no wonder it takes 40s in the restore) it'll do one report per ~800 restores which I don't think will overwhelm celery too much.

This is a little more high stakes so put a do not merge until i can do more testing

emord:
looks like just an error with a test mock. Will look into it tomorrow

dannyroberts:
neat :)

emord:
I think this is good to go pending review

snopoke (approved):

PR #15849, opened 2017-04-22
Fix blobdb restore

benrudolph:
@dimagi/scale-team this fixes the blobdb restore. when we were storing the cache payloads paths in couch, we didn't say which restore class it was for. when we swapped from using file restore to blob restore, s3 would complain about a bad filepath. i didn't notice this earlier because my blob db uses the filesystem so it never complained. here's an example of what i'm talking about:

screen shot 2017-04-22 at 6 42 39 pm

i don't think we need to store the restore payload paths on the sync log, but i haven't quite wrapped my head around the implications of that if any, so i didn't do it. but something to think about to reduce some of the restore complexity

buddy @calellowitz

### DIFF
--- a/corehq/ex-submodules/casexml/apps/phone/models.py
+++ b/corehq/ex-submodules/casexml/apps/phone/models.py
@@ -295,6 +295,10 @@ class AbstractSyncLog(SafeSaveDocument, UnicodeMixIn):
             ret.strict = False
         return ret
 
+    @property
+    def response_class(self):
+        return get_restore_response_class(self.domain)
+
     def case_count(self):
         """
         How many cases are associated with this. Used in reports.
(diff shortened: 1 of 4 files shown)


emord (approved):

snopoke (approved):
Don't we store the path for caching purposes? Though I agree that it doesn't seem like a great reason. In fact once there has been a successfully sync it didn't seem necessary to store / cache the payload at all (except perhaps for an edge case or 2)

benrudolph:
we used to store the entire file, then i changed to the path, now i'm wondering if we can just cache in redis like we do with restores that don't have a sync log: https://github.com/dimagi/commcare-hq/blob/master/corehq/ex-submodules/casexml/apps/phone/restore.py#L806-L814. good to know, will investigate a bit

benrudolph:
there may be one more issue here. when we do this hacky thing where we replace the cached payload's sync log id, https://github.com/dimagi/commcare-hq/blob/master/corehq/ex-submodules/casexml/apps/phone/restore.py#L395-L397, we have to use temp files, and we end up changing the payload_path to a file path when we need a s3 path

PR #15850, opened 2017-04-22
Cache get_last_synclog_for_user

benrudolph:
@dimagi/scale-team Realized one of the ways couch is getting hammered is via this call. It's only used in one place, the Application Status report. Essentially we loop through every user and get the latest sync log for that user. And when that's 100k users... that's a lot of hits. This does two things, 1) cache that call, and 2) limit the number users on the App Status to 5000

screen shot 2017-04-22 at 7 33 01 pm

buddy: @calellowitz

### DIFF
--- a/corehq/apps/reports/standard/deployments.py
+++ b/corehq/apps/reports/standard/deployments.py
@@ -35,7 +35,8 @@ from corehq.apps.reports.datatables import DataTablesHeader, DataTablesColumn, D
 from corehq.apps.reports.filters.select import SelectApplicationFilter
 from corehq.apps.reports.generic import GenericTabularReport
 from corehq.apps.reports.standard import ProjectReportParametersMixin, ProjectReport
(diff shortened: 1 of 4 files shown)


snopoke (approved):

benrudolph:
(for now reverted the cache commit because it's gonna break a ton of tests and it'll take me a bit to refactor, unless we just don't cache that function on tests)

snopoke:
👍

Looking at the screenshot, how fine we're querying `phone-tmp`? Isn't that an artifact of a prereindex?

dannyroberts:
@snopoke I also don't know the answer to that question. If the two are the same (`phone` and `phone-tmp`) then couch uses the same underly index for them, so I've seen couch sometimes arbitrarily display e.g. `phone-tmp` instead of `phone` in the list of things it's indexing. But that doesn't explain this, which I believe is logging the urls at the application (django) levels that we're calling. I don't know why we'd ever call a -tmp view.

PR #15851, opened 2017-04-22
save the sync log only once

benrudolph:
@dimagi/scale-team how do you guys feel about only saving the sync log once per sync? this would mean that we'd only get a record of a sync log if the restore successfully finishes, which i feel like is fine. i also noticed that under load, puts put strain on the couch database. this should reduce them (there are a couple other saves):

screen shot 2017-04-22 at 7 57 57 pm
 buddy: @calellowitz

### DIFF
--- a/corehq/ex-submodules/casexml/apps/phone/restore.py
+++ b/corehq/ex-submodules/casexml/apps/phone/restore.py
@@ -602,6 +602,7 @@ class RestoreState(object):
         previous_log_rev = None if self.is_initial else self.last_sync_log._rev
         last_seq = str(get_db().info()["update_seq"])
         new_synclog = SyncLog(
+            _id=uuid4().hex,
             domain=self.restore_user.domain,
             build_id=self.params.app_id,
             user_id=self.restore_user.user_id,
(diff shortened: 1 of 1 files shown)


snopoke (approved):
This seems fine to me - I don't see why we need to save it for incomplete restores.

snopoke:
build failure related

calellowitz:
@benrudolph This seems like a good move in the right direction, but have you done any analysis of the puts or slow requests? I noticed that the users db seemed to also have a lot of writes going on (or at least a lot of reindex activity), during periods of high load and im curious how much of the couch writes were synclogs vs user writes. Particulary since we get synclogs by id (so shouldnt hit the view), but we fetch users via a view so we need to wait for the reindex to finish each time. I would have imagined that had much greater load impacts and I am curious if you think its worth continuing investigation down that trail or whether you think this should be sufficient for a while

benrudolph:
so i decided to do this because 1. it was simple and 2. i want to reduce as many requests to couch as possible. right now we can save a sync log up to 5 times per sync which is a little silly and adds a lot of cruft to our database. i agree that it probably won't affect the retrieval of sync logs, but i think it will reduce the overall strain on couch which is something i think is useful.also makes sense to look into reducing user saves / load on db

PR #15852, opened 2017-04-22
add db level constraint for at most one active subscription per domain

nickpell:
Should run without issue on all production environments, since we monitor and remove duplicate active subscriptions.


corehq/apps/accounting/migrations/0006_unique_active_domain_subscription.py
```diff
@@ -0,0 +1,27 @@
...
+
+    dependencies = [
+        ('accounting', '0005_automatic_downgrade_adjustment_method'),
+    ]
+
+    operations = [
+        HqRunSQL(
+            """
+            CREATE UNIQUE INDEX accounting_subscription_active_subscriber
+            ON accounting_subscription(subscriber_id) WHERE (is_active = TRUE and is_hidden_to_ops = FALSE);
```
snopoke: Unless you're very sure this won't fail on any env you should do it manually beforehand and the use something like this to allow the migration to go ahead even with the index created: https://github.com/dimagi/commcare-hq/blob/c8a1d42bf1de3adace8cbebfb1559d302a05caae/corehq/util/django_migrations.py#L33-L33

corehq/apps/accounting/migrations/0006_unique_active_domain_subscription.py
```diff
@@ -0,0 +1,27 @@
...
+            """,
+            reverse_sql=
+            """
+            DROP INDEX accounting_subscription_active_subscriber;
```
snopoke: Use `DROP INDEX IF EXISTS` to be safe

nickpell: updated

corehq/apps/accounting/migrations/0006_unique_active_domain_subscription.py
nickpell: updated so it can be run manually beforehand

snopoke (approved):

snopoke:
Added `don't merge` label till you've had a chance to run this manually.

nickpell:
ran on all envs

nickpell:
@dimagi/team-commcare-hq If running this migration gives you problems locally, you'll want to delete any duplicate active subscriptions for the same domain (offline with me on how to do this)

PR #15853, opened 2017-04-22
Add metric for restore cache hits vs misses

benrudolph:
anyone / @calellowitz. curious to see what our ratio is


corehq/ex-submodules/casexml/apps/phone/restore.py
```diff
@@ -695,8 +696,12 @@ def get_payload(self):
         self.delete_cached_payload_if_necessary()
 
         cached_response = self.get_cached_response()
+        tags = [u'domain:{}'.format(self.domain)]
         if cached_response:
+            datadog_counter('commcare.restores.cache_hits.count', tags=tags)
             return cached_response
+        datadog_counter('commcare.restores.cache_misses.count', tags=tags)
```
snopoke: would be good to tag this with `is_initial`

PR #15856, opened 2017-04-23
add icds to toggles

emord:
@gcapalbo liked your idea and was in the UCR code when i noticed mobile ucr would be hit on every restore.

@millerdev

### DIFF
--- a/corehq/toggles.py
+++ b/corehq/toggles.py
@@ -588,6 +588,7 @@ MOBILE_UCR = StaticToggle(
      'through the app builder'),
     TAG_EXPERIMENTAL,
     namespaces=[NAMESPACE_DOMAIN],
+    always_enabled={'icds-cas'}
 )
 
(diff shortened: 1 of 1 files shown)


benrudolph (approved):

corehq/toggles.py
```diff
@@ -164,8 +168,6 @@ def _get_identifier(self, item):
     def enabled(self, item, **kwargs):
         if settings.UNIT_TESTING:
             return False
-        elif item in self.always_disabled:
-            return False
```
millerdev: Removing this means that the predictably random calculation could enable this toggle for a domain that is `always_disabled` since it comes first in the `or` condition below. Is that expected/intentional?

gcapalbo:
:+1: thanks @emord, looks good. agree with Daniel on that one point, seems like it's not the behavior we want there

emord:
Thanks for the close review @millerdev I added that check back to predictably random

gcapalbo (approved):

PR #15857, opened 2017-04-23
ICDS: Add filter for state to IT Issues Report

sheelio:
@sravfeyn 

Is this change enough to make this report location-safe and prevent users from seeing data outside their assigned states?


sravfeyn:
Yeah, that's enough to make it location-safe.

PR #15858, opened 2017-04-24
Soft assert when username is null

benrudolph:
@snopoke have some guesses where it could be coming from, but feel like this is the only way to be sure

cc: @calellowitz

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,7 +32,7 @@ script:
   - export TRAVIS_HQ_USERNAME=$TRAVIS_HQ_USERNAME
   - export TRAVIS_HQ_PASSWORD=$TRAVIS_HQ_PASSWORD
   - export DATADOG_API_KEY=$DATADOG_API_KEY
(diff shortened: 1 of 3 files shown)


snopoke (approved):

.travis.yml
```diff
@@ -32,7 +32,7 @@ script:
   - export TRAVIS_HQ_USERNAME=$TRAVIS_HQ_USERNAME
   - export TRAVIS_HQ_PASSWORD=$TRAVIS_HQ_PASSWORD
   - export DATADOG_API_KEY=$DATADOG_API_KEY
-  - "scripts/docker test --noinput --stop --verbosity=2 --divide-depth=1 --with-timing --threshold=10"
+  - "scripts/docker test --noinput --verbosity=2 --divide-depth=1 --with-timing --threshold=10"
```
snopoke: why?

benrudolph: just temporary so i can see all the test failures

PR #15860, opened 2017-04-24
count form submission types

benrudolph:
@snopoke couldn't really think of any other useful tags to include, let me know if you have any ideas. ran this with a few regular submissions

buddy: @calellowitz

### DIFF
--- a/corehq/form_processor/submission_post.py
+++ b/corehq/form_processor/submission_post.py
@@ -149,7 +149,7 @@ class SubmissionPost(object):
 
         cases = []
         ledgers = []
-        submission_type = 'normal'
+        submission_type = 'unknown'
         with result.get_locked_forms() as xforms:
             from casexml.apps.case.xform import get_and_check_xform_domain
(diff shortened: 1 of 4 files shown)


snopoke (approved):

snopoke:
pitty we can't get it in a single metric with tags

benrudolph:
actually we almost can, just one early return statement

benrudolph:
that might be better, will update

corehq/form_processor/submission_post.py
```diff
@@ -18,6 +18,7 @@
     CaseValueError
 from casexml.apps.case.xml import V2
 from corehq.toggles import ASYNC_RESTORE
+from corehq.util.datadog.gauges import datadog_counter
```
dimagimon: F401 'datadog_counter' imported but unused

benrudolph:
added `submission_type` to the FormProcessingResult

snopoke (approved):

snopoke (approved):

PR #15862, opened 2017-04-24
shortcircuit DB lookup

snopoke:
@dimagi/scale-team 

This will only help if all the app reports are static. There is no caching here so this get's called at least once per restore request.

### DIFF
--- a/settings.py
+++ b/settings.py
@@ -1277,6 +1277,8 @@ APPS_DB = NEW_APPS_DB
 
 SYNCLOGS_DB = 'synclogs'
 
+META_DB = 'meta'
+
 
 COUCHDB_APPS = [
     'api',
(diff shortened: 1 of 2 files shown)


nickpell (approved):

emord (approved):

settings.py
```diff
@@ -1277,6 +1277,8 @@
 
 SYNCLOGS_DB = 'synclogs'
 
+META_DB = 'meta'
+
```
dannyroberts: Is this related to the other commit or did it just kind of sneak its way in here?

dannyroberts (commented):
I love that a `get_db()` call (which I think makes just a `HEAD` request to couch?) might matter now.

dannyroberts:
(Also we might want to patch `get_db()` somehow to not make a request to couch if it currently does. That request is (if my memory serves) is just to see if the database exists, but of course all of these databases exist!)

settings.py
snopoke: yea, snuck it in

snopoke:
@dannyroberts this isn't trying to get rid of calls to `get_db` but rather to `get_doc`. 

Also I don't think `Document.get_db` will do a HEAD request except for the first time for that DB since couchdbkit keeps a global list of databases. If we use `dimagi.utils.couch.database.get_db` or `couch_config.get_db` then it will do a HEAD request

dannyroberts:
`get_docs`, right? I thought you just changed `get_docs` in another PR to short circuit if it was passed an empty list?

snopoke:
I did, I guess this is a double short circuit then

PR #15865, opened 2017-04-24
generate uuid's in python

snopoke:
@dimagi/scale-team 

I did a quick log analysis and because couchdbkit fetches UUID's in batches of 1000 there aren't that many of these requests (~1400 today)

### DIFF
--- a/corehq/apps/app_manager/models.py
+++ b/corehq/apps/app_manager/models.py
@@ -5143,7 +5143,7 @@ class ApplicationBase(VersionedDoc, SnapshotMixin,
         if not copy._id:
             # I expect this always to be the case
             # but check explicitly so as not to change the _id if it exists
-            copy._id = copy.get_db().server.next_uuid()
+            copy._id = uuid.uuid4().hex
 
         force_new_forms = False
         if previous_version and self.build_profiles != previous_version.build_profiles:
--- a/corehq/apps/app_manager/tests/test_app_manager.py
+++ b/corehq/apps/app_manager/tests/test_app_manager.py
@@ -1,5 +1,7 @@
 # coding: utf-8
 import json
+import uuid
+
 from mock import patch
 from corehq.apps.app_manager.tests.util import add_build, patch_default_builds
 from corehq.apps.app_manager.util import (add_odk_profile_after_build,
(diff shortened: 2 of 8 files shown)


emord (approved):

benrudolph (approved):

snopoke:
also https://github.com/dimagi/dimagi-utils/pull/403

corehq/apps/app_manager/models.py
```diff
@@ -5143,7 +5143,7 @@ def make_build(self, comment=None, user_id=None, previous_version=None):
         if not copy._id:
             # I expect this always to be the case
             # but check explicitly so as not to change the _id if it exists
-            copy._id = copy.get_db().server.next_uuid()
+            copy._id = uuid.uuid4().hex
```
dannyroberts: I'm guessing you thought of this, but did you check that these are the same format ("look the same") as the ones coming from couch. Would be mildly annoying to have a visible change in the ids like them previously having dashes and now not having dashes or something.

dannyroberts (approved):
Ha! It's amazing that we're at the point were this kind of thing matters—or even might plausibly matter!

corehq/apps/app_manager/models.py
snopoke: The format is the same:
```
aef4b1631098306c1d7239808c000871 # couch
29a07f566f854b089f5a1bda057dba9d # python
```
but the algorithm is different. Couch uses a sequential algorithm to generate unique ID's per couch instance: http://docs.couchdb.org/en/2.0.0/config/misc.html#uuids

dannyroberts: Ah, right. Cool. Saw this right below it in the docs you linked, wonder if it'll have an impact—at this scale it's not clear to me it wouldn't

> Impact of UUID choices: the choice of UUID has a significant impact on the layout of the B-tree, prior to compaction.
>
> For example, using a sequential UUID algorithm while uploading a large batch of documents will avoid the need to rewrite many intermediate B-tree nodes. A random UUID algorithm may require rewriting intermediate nodes on a regular basis, resulting in significantly decreased throughput and wasted disk space space due to the append-only B-tree design.
> 
> It is generally recommended to set your own UUIDs, or use the sequential algorithm unless you have a specific need and take into account the likely need for compaction to re-balance the B-tree and reclaim wasted space.

snopoke: Thanks, that is interesting.

I just realized that this is a bit pointless because if you save a doc without an ID couchdbkit will still call `next_uuid` to get the doc ID. 

Either way, without forms and cases in couch we aren't generating THAT many new docs so I don't think this should be an issue

PR #15866, opened 2017-04-24
Couch user id in djago user

benrudolph:
@dimagi/scale-team best read 🐟 . this is the first part of moving couch user ids to a new table. next PR will be a migration that converts existing users to have this field. 

cc: @calellowitz


corehq/apps/users/models.py
```diff
@@ -2445,3 +2451,9 @@ def can_edit_locations(self):
 
     def can_edit_web_users(self):
         return False
+
+
+class DjangoUserCommCareFields(models.Model):
+    user = models.OneToOneField(User, on_delete=models.CASCADE)
+
+    couch_user_id = models.CharField(max_length=255)
```
benrudolph: decided not to index because i don't think there's a real use case for looking up the django user by couch id since we store the django id in the couch model

emord (approved):

corehq/apps/users/models.py
```diff
@@ -2445,3 +2451,9 @@ def can_edit_locations(self):
 
     def can_edit_web_users(self):
         return False
+
+
+class DjangoUserCommCareFields(models.Model):
```
benrudolph: will update with some docs

corehq/apps/users/models.py
```diff
@@ -1287,6 +1287,11 @@ def create(cls, domain, username, password, email=None, uuid='', date='',
         else:
             couch_user.created_on = datetime.utcnow()
 
+        DjangoUserCommCareFields.objects.create(
+            user=django_user,
+            couch_user_id=couch_user._id,
+        )
```
snopoke: just looking at the code above it seems like `django_user` could already exists - that seems quite an edge case but what would happen if 2 of these models got created for one user?

benrudolph: yea good point. i'll write something for that

benrudolph:
thinking about holding off on this and seeing if the caching of usernames helps enough

PR #15869, opened 2017-04-24
(Revert Revert) Optimize Forms by Submission

dannyroberts:
@benrudolph @biyeun, cleaning up after http://manage.dimagi.com/default.asp?252504, reindexing like Ben suggested. (Sorry for overlooking that in the first place, I should have known better and included a reindex in the original fix.)

### DIFF
--- a/corehq/pillows/mappings/user_mapping.py
+++ b/corehq/pillows/mappings/user_mapping.py
@@ -3,7 +3,7 @@ from corehq.pillows.core import DATE_FORMATS_ARR, DATE_FORMATS_STRING
 
 from pillowtop.es_utils import ElasticsearchIndexInfo
 
-USER_INDEX = es_index("hqusers_2017-04-24")
+USER_INDEX = es_index("hqusers_2017-03-22")
 USER_MAPPING = {'_all': {'analyzer': 'standard'},
  '_meta': {'created': None},
  'date_detection': False,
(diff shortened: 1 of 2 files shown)


corehq/pillows/mappings/user_mapping.py
```diff
@@ -3,7 +3,7 @@
 
 from pillowtop.es_utils import ElasticsearchIndexInfo
 
-USER_INDEX = es_index("hqusers_2017-03-22")
+USER_INDEX = es_index("hqusers_2017-04-24")
```
dannyroberts: @calellowitz you were the last one to do a user reindex so I'll ask you; is this all that's needed to trigger a reindex? (And is it the preferred way?)

dannyroberts:
@orangejenny buddy

corehq/pillows/mappings/user_mapping.py
calellowitz: if you are not changing the mapping file you should not need to change the index name. i think you can just run the management command `./manage.py ptop_reindexer_v2 user --in-place` which will reindex all the docs without  creating an entirely new index. The other will create a whole new index (and hold up deploy i believe). Let me know if thats not actually what you were going for. If you do proceed with changing the name, I think there are a few other places it appears in the code base (particularly test code) that you will have to change.

biyeun (approved):

biyeun:
thanks, @dannyroberts!

corehq/pillows/mappings/user_mapping.py
dannyroberts: Yeah ok that seems better. I'll revert that and run the reindex manually on production, and softlayer. And coordinate w/ scale team on NIC

dannyroberts:
Manually reindex production, softlayer, and NIC like Cal suggested

PR #15870, opened 2017-04-24
Bust repeater cache and other stories

proteusvacuum:
We were having many issues with the 99DOTS repeater that I was sure was a ghost.

We thought we were receiving responses from their server for requests that they thought they were never receiving. 

In reality, we were never sending the requests because... whenever we encountered _any_ error for a particular URL, we would just return a cached response for the next hour. Sweet.
So, bust that cache: https://github.com/dimagi/commcare-hq/commit/062872859856268e9cb0736bef42cd7401c044ca
I'm surprised that this hasn't ever been reported. Maybe no one ever looks at their own repeat record logs.

We were also hammering the 99DOTS servers with multiple requests for the same case.

Before, we did this [by design](https://github.com/dimagi/commcare-hq/commit/f1e0670c2d4d6963cc2b8eb554a2962b205eaae2). 

I turned that off, but it kept happening.

Turns out, we were just making a new repeat record whenever we processed any form that was updating a case that hadn't successfully been registered on the 99DOTS server, but was scheduled to be. This included forms that were submitted by the system against that case to log that the 99DOTS endpoint had sent us an error. 

So, only make a [single](https://github.com/dimagi/commcare-hq/commit/ee5e813494ce9d557c9076b2f8855ee865d2052e) repeat record.

@sravfeyn @NoahCarnahan 🐡 
@benrudolph if you are interested.


proteusvacuum:
@dannyroberts FYI

custom/enikshay/integrations/ninetyninedots/repeater_generators.py
```diff
@@ -92,8 +93,16 @@ def create(cls, person_case, episode_case):
         )
 
 
+class NinetyNineDotsBasePayloadGenerator(BasePayloadGenerator):
+    def handle_exception(self, exception, repeat_record):
+        if isinstance(exception, RequestConnectionError):
+            update_case(repeat_record.domain, repeat_record.payload_id, {
+                "dots_99_error": u"RequestConnectionError: {}".format(unicode(exception))
+            })
```
dannyroberts: we handle network exceptions in the payload generator? Weird.

corehq/apps/repeaters/models.py
```diff
@@ -50,7 +50,7 @@
 def simple_post_with_cached_timeout(data, url, expiry=60 * 60, force_send=False, *args, **kwargs):
     # no control characters (e.g. '/') in keys
     key = hashlib.md5(
-        '{0} timeout {1}'.format(__name__, url)
+        u'{0} timeout {1} {2}'.format(__name__, url, data).encode('utf-8')
```
dannyroberts: this is saying cache the error when sending the same data to the same url—that basically means we're deciding not cache to the fact that an endpoint is timing out, right? What we were doing before was definitely too broad, but just checking that we're making that decision.

dannyroberts (approved):

corehq/apps/repeaters/models.py
proteusvacuum: Would be worth a discussion, actually. I'm not sure why we cache failure responses at all. 

In general repeaters currently do the following:
- If you hit any error
   - [immediately retry 3 times](https://github.com/dimagi/commcare-hq/blob/master/corehq/apps/repeaters/models.py#L589-L593) (I call this the "pummel the other server into compliance" phase) (Now that I think of it, the pummel phase is actually useless because whenever there is an error and this method is called again, it just returns the cached response.)
   - if all 3 of those fail, [requeue](https://github.com/dimagi/commcare-hq/blob/master/corehq/apps/repeaters/models.py#L605-L606) the repeater to try again in an hour

The caching *might* be useful if you hit a timeout, but there is no way of knowing whether the timeout was due to the particular data you submitted or if it is because the server is unavailable. If it is because of the data that you submitted, and we cache that timeout, we are just delaying all other requests to that server. 

I'd like to propose:
- remove the (useless) pummel phase
- remove the cache completely

This means:
- If the server is timing out because it is unavailable, we try once for each repeat record, then requeue it an hour later (then an increasing amount of time).
- If the server is timing out because of the data you submitted, it will requeue that particular record up to 3 times, then put that record into the [cancelled state](https://github.com/dimagi/commcare-hq/blob/master/corehq/apps/repeaters/models.py#L607-L609), without affecting any of the other requests to that server

benrudolph: I would be keen to have a conversation about this. FWIW those timeouts have ended up killing us in the past. However there have been significant changes since then so we may want to reconsider

dannyroberts: I offlined with @proteusvacuum, but basically we agreed to just cut those two things out entirely (the pummeling and the error caching), and replace the error caching with a datadog counter of timeouts so that the next there's a pileup in the queue, we can check the timeouts graph to see if there's been a spike in timeouts to rule that out or identify it as the cause.

dannyroberts: Our guess is there's a chance that this just won't be an issue now with all the significant changes you mentioned @benrudolph

custom/enikshay/integrations/ninetyninedots/repeater_generators.py
proteusvacuum: We handle the exception in the repeater code, but we write back to the case that this error occurred when this case was being submitted. 
It would be great if we had a way of pulling a report that showed you each payload that failed and the reason why. I didn't have time to build that when making the repeaters the first time around, so this is our hacky way of logging it.

dannyroberts: Ah, thanks for the context.

proteusvacuum:
✅

corehq/apps/repeaters/models.py
benrudolph: Ok great, that sounds fine to me

mkangia:
@proteusvacuum great that this is figured.

> In reality, we were never sending the requests because... whenever we encountered any error for a particular URL, we would just return a cached response for the next hour. Sweet.
So, bust that cache: 0628728

Not sure what would be the indentation behind a cached response for a request. Thanks for fixing this. 

> We were also hammering the 99DOTS servers with multiple requests for the same case.
This would be 3 tries per attempt right? I think its come handy if service times out but succeeds on reattempt instead of waiting for getting tried again later. Might want to re consider that change.

Seems like the underlying issue was multiple repeat records (N) for a single case which would amplify the tries to 3 x N times, which you have fixed later along with fixing the reason for 3 attempts due to url issue.
Thanks for tracking this down.

benrudolph:
> Not sure what would be the indentation behind a cached response for a request. Thanks for fixing this.

We used to pile up hundreds of thousands of repeater requests to servers that would just time out. They would cause a huge back log in the queue. Caching was actually pretty critical and it was when we broke the caching that the backup happened. Now that we cancel repeaters though, I think it'll be much better

mkangia:
Thank you @benrudolph

snopoke:
Great story!

PR #15871, opened 2017-04-24
use username_to_user_id

benrudolph:
@dimagi/scale-team this uses a function that maps a username to a couch user id and then aggressively caches it. hopefully this'll provide short term relief while we work on a longer term solution. thanks @emord for the suggestion

cc: @calellowitz

### DIFF
--- a/corehq/apps/users/models.py
+++ b/corehq/apps/users/models.py
@@ -44,6 +44,7 @@ from corehq.apps.domain.models import Domain, LicenseAgreement
 from corehq.apps.users.util import (
     user_display_string,
     user_location_data,
+    username_to_user_id,
 )
 from corehq.apps.users.tasks import tag_forms_as_deleted_rebuild_associated_cases, \
     tag_cases_as_deleted_and_remove_indices, tag_system_forms_as_deleted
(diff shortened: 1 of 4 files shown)


corehq/apps/users/models.py
```diff
@@ -1216,6 +1214,7 @@ def clear_quickcache_for_user(self):
 
         self.get_by_username.clear(self.__class__, self.username)
         self.get_by_user_id.clear(self.__class__, self.user_id)
+        username_to_user_id.clear(self.username)
```
snopoke: does this get run on bulk upload?

benrudolph: like the `.clear`? no, but i don't think we have to worry about that too much. the cache returns incorrectly when a user deletes a mobile worker and then creates a new one with the same username

snopoke: I was thinking about adding users in a bulk restore but I guess it's quite unlikely that a user will try to auth before they've been created.

dannyroberts: There's also no reason we couldn't call `user.clear_quickcache_for_user()` for each user during bulk upload if we wanted to make sure that bulk uploads don't create an inconsistent cache situation.

PR #15874, opened 2017-04-24
make web_user_lookup available to superusers

dannyroberts:
even if they do not have is_staff


dannyroberts:
@orangejenny @calellowitz

dannyroberts:
tested locally for with and without is_staff

benrudolph (approved):

PR #15878, opened 2017-04-25
bump lock timeout on forms

snopoke:
Getting some errors like this https://sentry.io/dimagi/commcarehq/issues/250116028/ where the phone times out and re-submits a form. I think this should hopefully stop it.

(sneaking in the first commit as a tiny cleanup)


benrudolph (approved):
dang we're taking longer than 2 mins to process a form

PR #15879, opened 2017-04-25
user server uuid

benrudolph:
@snopoke from you and @dannyroberts 's conversation


snopoke:
comment for future would be useful

PR #15881, opened 2017-04-25
cache blob content on attachment during write

snopoke:
@dimagi/scale-team this should reduce our riak reads by a lot. At the moment we read the form out of riak during form processing instead of keeping it in memory (dumb).

buddy @nickpell

### DIFF
--- a/corehq/form_processor/models.py
+++ b/corehq/form_processor/models.py
@@ -413,14 +413,31 @@ class AbstractAttachment(DisabledDbMixin, models.Model, SaveStateMixin):
         if not self.name:
             raise InvalidAttachment("cannot save attachment without name")
 
+        content_readable = content
+        if isinstance(content, basestring):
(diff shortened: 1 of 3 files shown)


corehq/form_processor/models.py
```diff
@@ -413,14 +413,31 @@ def write_content(self, content):
         if not self.name:
             raise InvalidAttachment("cannot save attachment without name")
 
+        content_readable = content
+        if isinstance(content, basestring):
+            content_readable = StringIO(content)
```
benrudolph: do you know when is this a string vs not a string?

benrudolph (approved):

corehq/form_processor/models.py
snopoke: this should actually always be a string so could potentially just more the `if`. I just added it as a safeguard

PR #15882, opened 2017-04-25
Reminder Calendar Fixes

gcapalbo:
first commit: addresses the soft asserts
second commit: addresses an issue where times were being displayed in the wrong timezone
third commit: addresses https://manage.dimagi.com/default.asp?252582

@orangejenny @dannyroberts 
buddy @NoahCarnahan

### DIFF
--- a/corehq/apps/reminders/views.py
+++ b/corehq/apps/reminders/views.py
@@ -102,7 +102,7 @@ class ScheduledRemindersCalendarView(BaseMessagingSectionView):
         today = timezone_now.date()
 
         def adjust_next_fire_to_timezone(reminder_utc):
-            return ServerTime(reminder_utc.next_fire).user_time(timezone).done()
+            return ServerTime(reminder_utc.next_fire).user_time(timezone).done().replace(tzinfo=None)
 
         if reminders:
             start_date = adjust_next_fire_to_timezone(reminders[0]).date()
--- a/corehq/apps/reminders/static/reminders/js/scheduled_reminders.js
+++ b/corehq/apps/reminders/static/reminders/js/scheduled_reminders.js
@@ -2,7 +2,7 @@
 hqDefine("reminders/js/scheduled_reminders.js", function() {
     $(function () {
         var reminders = _.map(hqImport("hqwebapp/js/initial_page_data.js").get("reminder_data"), function(data) {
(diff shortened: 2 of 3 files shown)


NoahCarnahan (approved):
Looks good to me

orangejenny (approved):

corehq/apps/reminders/views.py
```diff
@@ -102,7 +102,7 @@ def page_context(self):
         today = timezone_now.date()
 
         def adjust_next_fire_to_timezone(reminder_utc):
-            return ServerTime(reminder_utc.next_fire).user_time(timezone).done()
+            return ServerTime(reminder_utc.next_fire).user_time(timezone).done().replace(tzinfo=None)
```
dannyroberts: I cringe a little to see this, even though looking at the PR as a whole I don't have an obviously better option for you. I say I cringe, because it kind of breaks the only sane-keeping standard we have about datetimes and timezones which are that

1. datetimes without timezones == server time == UTC
2. datetimes with timezones == user time (== probably not utc)

I don't know what the best solution is. Maybe we decide that timezones are such a mess and so different between implementations that we want to only ever do timezone conversions in python; in that case, the thing to do would be do the full conversion to a user display string here in python, and pass that through. Or we decide when we're passing data from python to js, we want to pass the underlying data and let js decide how to display it, in which case we should probably use something like [Moment Timezone](http://momentjs.com/timezone/), pass the server time (UTC) plus the user's preferred timezone to js, and then do the conversion and formatting in js.

dannyroberts (approved):
Thanks for the quick fix. Have one comment that's a bit of a nit/pedantic, but I thought important to raise given how messy timezone-related problems can be when we don't follow strict conventions around them.

corehq/apps/reminders/views.py
gcapalbo: thanks, i agree that we should make it a standard that naive datetimes are always utc.

i think for this particular case though, it's probably not worth coming up with a more complex solution since this view is going away in a few months and will be replaced with a report where we do follow better timezone conventions.

dannyroberts: Ah, nice. Yeah I'm totally comfortable with that then. Wasn't a deal-breaker to begin with, but even less so in code we're actively replacing.

PR #15883, opened 2017-04-25
Enikshay fix and debug view

sravfeyn:
@proteusvacuum

### DIFF
--- a/custom/enikshay/tasks.py
+++ b/custom/enikshay/tasks.py
@@ -239,10 +239,14 @@ class EpisodeUpdate(object):
             }
         """
 
+        debug_data = []
         adherence_schedule_date_start = self.get_adherence_schedule_start_date()
+        debug_data.append("adherence_schedule_date_start: {}".format(adherence_schedule_date_start))
+        debug_data.append("purge_date: {}".format(self.case_updater.purge_date))
+
         if not adherence_schedule_date_start:
             # adherence schedule hasn't been selected, so no update necessary
-            return {}
+            return {'update': {}, 'debug_data': debug_data}
 
         default_update = {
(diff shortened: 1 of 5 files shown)


custom/enikshay/tasks.py
```diff
@@ -179,6 +182,24 @@ def is_dose_taken(cases):
...
+        # return property 'adherence_schedule_date_start' of episode case, cast to datetime
+        raw_date = self.get_property('adherence_schedule_date_start')
+        if not raw_date:
+            return None
+        if parse_datetime(raw_date):
+            return parse_datetime(raw_date)
+        elif parse_date(raw_date):
+            return datetime.datetime.combine(parse_date(raw_date), datetime.datetime.min.time())
+        else:
+            logger.error(
```
proteusvacuum: this is the only place where the logger is being used, is that intentional?

custom/enikshay/tasks.py
```diff
@@ -179,6 +182,24 @@ def is_dose_taken(cases):
 
         return dose_taken_by_date
 
+    def get_adherence_schedule_start_date(self):
+        # return property 'adherence_schedule_date_start' of episode case, cast to datetime
+        raw_date = self.get_property('adherence_schedule_date_start')
```
proteusvacuum: I know you didn't change this here, but I think every call to `dynamic_case_properties` does a DB lookup. You might want to cache that in some way (e.g. `self.properties`) in the `get_property` method.

proteusvacuum (approved):
couple of questions, otherwise g2g

custom/enikshay/tasks.py
sravfeyn: Good catch, it was mistake. I meant to raise an exception here, updated now

PR #15884, opened 2017-04-25
Cache restore payload paths in redis

benrudolph:
@snopoke this refactors restores to cache restore payload paths in redis. best read 🐟 

_Previous world_
On initial sync we would cache the payload path in redis if it took a long time. When the user was in steady state sync, we would store the cache path on the sync log token

_New world_
All syncs store the path to the cached restore in redis

cc: @calellowitz

### DIFF
--- a/corehq/ex-submodules/casexml/apps/phone/restore.py
+++ b/corehq/ex-submodules/casexml/apps/phone/restore.py
@@ -66,14 +66,15 @@ from xml.etree import ElementTree
 logger = logging.getLogger(__name__)
 
 
-def restore_cache_key(domain, prefix, user_id, version=None, sync_log_id=None):
+def restore_cache_key(domain, prefix, user_id, version=None, sync_log_id=None, device_id=None):
     response_class = get_restore_response_class(domain)
-    hashable_key = '{response_class}-{prefix}-{user}-{version}-{sync_log_id}'.format(
+    hashable_key = '{response_class}-{prefix}-{user}-{version}-{sync_log_id}-{device_id}'.format(
         response_class=response_class.__name__,
         prefix=prefix,
         user=user_id,
         version=version or '',
         sync_log_id=sync_log_id or '',
+        device_id=device_id or '',
     )
     return hashlib.md5(hashable_key).hexdigest()
 
@@ -406,14 +407,22 @@ class RestoreParams(object):
     :param version:             The version of the restore format
     :param state_hash:          The case state hash string to use to verify the state of the phone
     :param include_item_count:  Set to `True` to include the item count in the response
+    :param device_id:           The Device id of the device restoring
     """
 
-    def __init__(self, sync_log_id='', version=V1, state_hash='', include_item_count=False, app=None):
+    def __init__(self,
+            sync_log_id='',
+            version=V1,
+            state_hash='',
+            include_item_count=False,
+            device_id=None,
+            app=None):
         self.sync_log_id = sync_log_id
         self.version = version
         self.state_hash = state_hash
         self.include_item_count = include_item_count
         self.app = app
+        self.device_id = device_id
 
     @property
     def app_id(self):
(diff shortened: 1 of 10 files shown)


corehq/ex-submodules/casexml/apps/phone/tests/test_ota_restore.py
```diff
@@ -131,6 +131,11 @@ def testOverwriteCache(self):
         self.assertIsInstance(restore_config_cached.get_payload(), CachedResponse)
         self.assertNotIsInstance(restore_config_overwrite.get_payload(), CachedResponse)
 
+        # even cached responses change the sync log id so they are not the same
```
benrudolph: here is the crux of the `finalize` problem in the `CachedPayload`. this will continue to fail because we don't replace the sync log id in the restore. i really want to avoid having to do that. i think we discussed that we can just add the device_id to the restore_cache_key and that will alleviate the issue (have not done that yet). it's a bit confusing to me still cause i'm not exactly sure why it's bad when a user is logged into multiple devices and gets the same restore payload

corehq/ex-submodules/casexml/apps/phone/tests/test_sync_logs.py
```diff
@@ -78,29 +78,6 @@ def test_archive(self):
         self.assertEqual(0, len(log.get_footprint_of_cases_on_phone()))
 
 
-class CachingResponseTest(TestCase):
-
-    def testCachingResponse(self):
```
benrudolph: i no longer think this test is necessary

benrudolph:
I think this is g2g barring this one question here: https://github.com/dimagi/commcare-hq/pull/15884#discussion_r113209497. That determines how I fix the last failing test

corehq/ex-submodules/casexml/apps/phone/tests/test_ota_restore.py
snopoke: It's bad because then their submissions from different forms will be attributed to the same synclog and the state will diverge from what is on the phone causing the phone to 412.

snopoke: It is an edge case which probably doesn't happen that often (though I don't have data to support that assumption). If we include the device ID in the key then I think we can go ahead.

benrudolph: > It's bad because then their submissions from different forms will be attributed to the same synclog and the state will diverge from what is on the phone causing the phone to 412.

Ah! that was the lightbulb I needed 💡. thanks, will update with device id

benrudolph: https://github.com/dimagi/commcare-hq/pull/15884/commits/bfbb1e37a02767ac64e513bae9ef3c993186b29c

corehq/ex-submodules/casexml/apps/phone/restore.py
```diff
@@ -436,14 +407,22 @@ class RestoreParams(object):
...
+    :param device_id:           The Device id of the device restoring
     """
 
-    def __init__(self, sync_log_id='', version=V1, state_hash='', include_item_count=False, app=None):
+    def __init__(self,
+            sync_log_id='',
+            version=V1,
+            state_hash='',
+            include_item_count=False,
+            device_id=None,
```
snopoke: not seeing where this gets passed in from the view

benrudolph: forgot to commit the `views.py` file :man_facepalming:

snopoke: code review ftw :grin:

snopoke (approved):

corehq/ex-submodules/casexml/apps/phone/restore.py
benrudolph: eagle eyes 🐦

PR #15885, opened 2017-04-25
don't fetch all case IDs just to validate one

snopoke:


corehq/apps/zapier/views.py
```diff
@@ -155,7 +156,9 @@ def post(self, request, *args, **kwargs):
         if not couch_user.is_member_of(domain):
             return HttpResponseForbidden("This user does not have access to this domain.")
 
-        if case_id not in CaseAccessors(domain).get_case_ids_in_domain(case_type):
+        try:
+            CaseAccessors(domain).get_case(case_id)
```
emord: I think you may want to assert case_type as well

nickpell: +1

snopoke: turns out there was a test for that too, nice

snopoke: 06f2a14

emord (approved):

PR #15886, opened 2017-04-25
Cache get_casedb_schema

orangejenny:
Band-aid for https://manage.dimagi.com/default.asp?252644

`w=1`

@dannyroberts

### DIFF
--- a/corehq/apps/app_manager/util.py
+++ b/corehq/apps/app_manager/util.py
@@ -418,27 +418,27 @@ def get_casedb_schema(form):
         related = builder.get_parent_type_map(case_types, allow_multiple_parents=True)
         map = builder.get_case_property_map(case_types, include_parent_properties=False)
         descriptions_dict = get_case_property_description_dict(app.domain)
-    
+
         if base_case_type:
             # Generate hierarchy of case types, represented as a list of lists of strings:
             # [[base_case_type], [parent_type1, parent_type2...], [grandparent_type1, grandparent_type2...]]
             # Vellum case management only supports three levels
             generation_names = ['case', 'parent', 'grandparent']
             generations = [[] for g in generation_names]
-    
+
             def _add_ancestors(ctype, generation):
                 if generation < len(generation_names):
                     generations[generation].append(ctype)
                     for parent in related.get(ctype, {}).get('parent', []):
                         _add_ancestors(parent, generation + 1)
-    
+
             _add_ancestors(base_case_type, 0)
-    
+
             # Remove any duplicate types or empty generations
             generations = [set(g) for g in generations if len(g)]
         else:
             generations = []
-    
+
         subsets = [{
             "id": generation_names[i],
             "name": "{} ({})".format(generation_names[i], " or ".join(ctypes)) if i > 0 else base_case_type,
@@ -451,7 +451,7 @@ def get_casedb_schema(form):
                 "key": "@case_id",
             }} if i < len(generations) - 1 else None,
         } for i, ctypes in enumerate(generations)]
-    
+
         if is_usercase_in_use(app.domain) and toggles.USER_PROPERTY_EASY_REFS.enabled(app.domain):
             subsets.append({
                 "id": USERCASE_TYPE,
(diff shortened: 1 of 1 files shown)


dimagimon:
There are 7 errors:

* corehq/apps/app_manager/util.py, line 421 - W293 blank line contains whitespace
* corehq/apps/app_manager/util.py, line 428 - W293 blank line contains whitespace
* corehq/apps/app_manager/util.py, line 434 - W293 blank line contains whitespace
* corehq/apps/app_manager/util.py, line 436 - W293 blank line contains whitespace
* corehq/apps/app_manager/util.py, line 441 - W293 blank line contains whitespace
* corehq/apps/app_manager/util.py, line 454 - W293 blank line contains whitespace
* corehq/apps/app_manager/util.py, line 462 - W293 blank line contains whitespace

corehq/apps/app_manager/util.py
```diff
@@ -408,64 +408,68 @@ def get_casedb_schema(form):
...
-            p: {"description": descriptions_dict.get(t, {}).get(p, '')}
-            for t in ctypes for p in map[t]},
-        "related": {"parent": {
-            "hashtag": "#case/" + generation_names[i + 1],
-            "subset": generation_names[i + 1],
-            "key": "@case_id",
-        }} if i < len(generations) - 1 else None,
-    } for i, ctypes in enumerate(generations)]
+    @quickcache(['app.id', 'app.version'])
+    def _inner(app, form):
```
dannyroberts: Do you want this to be cached across different calls to `get_casedb_schema`? I've never tried the decorator on an inner function like this, and I can't see why it wouldn't work—the cache key should still match up. (`@memoized` definitely would not work here, since its cache is stored on the function object itself, which is what made this catch my attention.

Anyway I think this is fine despite my initial side-glance, especially if you tried it and it's clearly caching.

dannyroberts (commented):
Just checking my understanding here—this function would return a stale cache for the default timeout (five minutes I think?) when _another_ app changes, since other apps' versions are not in the cache key.

If I read it correctly and that's what'll happen, are you comfortable with the cost/benefit of caching it for 5 minutes? Since you don't often reload a form within 5 minutes, I'm guessing the benefit would be for when you open a bunch of different forms at the same time in different tabs? Even that may not get the benefit—they'll likely all hit a cold cache, since the time to compute it form one form may be longer than the time between opening forms. (Can offline if I didn't describe that clearly.)

Being stale for 5 minutes may more may not be a big downside, but I'm trying to see whether we're really getting the upside we want from this to compensate for whatever downside there is to a stale cache.

I think the ideal solution would cache only steps that depend on a single app, but cache them for much longer, like a couple hours. (When I quickly looked over the code, though, I agree finding such an entry point was actually a lot tricker than I thought.)

dannyroberts:
possibly due to the same stale cache issue I mentioned?

```
FAIL: corehq.apps.app_manager.tests.test_util:SchemaTest.test_get_casedb_schema_form_without_cases
  File "/mnt/commcare-hq-ro/corehq/apps/app_manager/tests/test_util.py", line 116, in test_get_casedb_schema_form_without_cases
    "subsets": [],
  File "/mnt/commcare-hq-ro/corehq/apps/app_manager/tests/test_util.py", line 355, in assert_has_kv_pairs
    self.assertEqual(test_dict[key], value)
AssertionError: Lists differ: [{'structure': {'case_name': {... != []
First list contains 2 additional elements.
First extra element 0:
{'structure': {'case_name': {'description': ''}}, 'id': 'case', 'related': {'parent': {'subset': 'parent', 'key': '@case_id', 'hashtag': '#case/parent'}}, 'name': u'referral'}
+ []
- [{'id': 'case',
-   'name': u'referral',
-   'related': {'parent': {'hashtag': '#case/parent',
-                          'key': '@case_id',
-                          'subset': 'parent'}},
-   'structure': {'case_name': {'description': ''}}},
-  {'id': 'parent',
-   'name': 'parent (pregnancy or child)',
-   'related': None,
-   'structure': {'case_name': {'description': ''}}}]
```

orangejenny:
Closing in favor of https://github.com/dimagi/commcare-hq/pull/15906

PR #15888, opened 2017-04-25
use simpler call to determine if case exists

snopoke:
buddy @nickpell @dimagi/scale-team 

Change from a view call to a simpler ID check.


corehq/form_processor/tests/utils.py
```diff
@@ -135,20 +103,34 @@ def delete_all_sql_forms(domain=None, user_id=None):
...
+
+            deleted = set()
+            for row in db.view(view, reduce=False, **view_kwargs):
+                doc_id = row['id']
+                if doc_id not in deleted:
+                    try:
+                        safe_delete(db, doc_id)
+                        deleted.add(doc_id)
+                    except ResourceNotFound:
+                        pass
```
snopoke: Previously this wasn't deleting 'deleted' docs because it was using views that excluded them.

corehq/form_processor/backends/couch/dbaccessors.py
```diff
@@ -126,6 +126,10 @@ def get_cases(case_ids, ordered=False):
         ]
 
     @staticmethod
+    def case_exists(case_id):
+        return CommCareCase.get_db().doc_exist(case_id)
```
snopoke: This is a broader check since it's not hitting a view that only includes cases. I think that's OK since we shouldn't have any docs with duplicate ID's anyway.

dannyroberts: Yeah that seems fine to me. Looks like `get_case_with_lock` was also doing a naive `CommCareCase.get_db().get(case_id)` without checking its `case_type`, so whether or not this is super "correct" (which is a harder question), it does seem to be what we've always been doing.

dannyroberts (approved):
Looks good to me! What we were doing before was pretty roundabout 🔀

nickpell:
tests are failing

nickpell:
still getting some test failures

custom/ewsghana/tests/handlers/handler.py
```diff
@@ -17,16 +17,25 @@ def inner(*args, **kwargs):
...
+    @classmethod
+    def setUpClass(cls):
+        super(HandlerTest, cls).setUpClass()
+        cls.domain = 'ews-handler-test'
+        bootstrap_domain(cls.domain)
+        cls.create_account_and_subscription(cls.domain)
+        cls.domain_obj = Domain.get_by_name(cls.domain)
+        cls.loc = make_loc(code="garms", name="Test RMS", type="Regional Medical Store", domain=cls.domain)
+        cls.user = bootstrap_user(username='testuser', phone_number='323232', domain=cls.domain,
+                                   home_loc=cls.loc)
```
dimagimon: E127 continuation line over-indented for visual indent

PR #15891, opened 2017-04-25
Fix saving UCRs

orangejenny:
https://manage.dimagi.com/default.asp?252710

@dannyroberts

### DIFF
--- a/corehq/apps/reports/static/reports/js/saved_reports.js
+++ b/corehq/apps/reports/static/reports/js/saved_reports.js
@@ -36,7 +36,7 @@ var ReportConfig = function (data) {
 
     self.unwrap = function () {
         var data = ko.mapping.toJS(self);
-        var standardHQReport = hqImport("userreports/js/configurable_report.js").get();
+        var standardHQReport = hqImport("userreports/js/configurable_report.js").getStandardHQReport();
         if (null !== standardHQReport.slug) {
             data['report_slug'] = standardHQReport.slug;
         }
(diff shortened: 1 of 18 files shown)


corehq/apps/userreports/static/userreports/js/configurable_report.js
```diff
@@ -80,6 +84,11 @@ hqDefine("userreports/js/configurable_report.js", function() {
         }
         var standardHQReport = new HQReport(reportOptions);
         standardHQReport.init();
+        return standardHQReport;
+    };
+
+    $(function() {
+       getStandardHQReport();
```
dimagimon: Expected indentation of 8 space characters but found 7. (indent)

corehq/apps/reports/static/reports/js/saved_reports.js
```diff
@@ -36,6 +36,7 @@ var ReportConfig = function (data) {
 
     self.unwrap = function () {
         var data = ko.mapping.toJS(self);
+        var standardHQReport = hqImport("userreports/js/configurable_report.js").getStandardHQReport();
```
dannyroberts: offlined but wanted to leave a note on the convo.

It doesn't feel quite right for the framework (reports, saved reports) to be importing this from userreports. While on that particular page `standardHQReport`'s referent is the `HQReport` instance created in `userreports/js/configurable_report.js`, this is loaded on a number of pages, and on those other pages, its referent certainly _won't_ be the userreports `HQReport` instance. In other words, you might say that the referent of `standardHQReport` is not statically defined—it's different depending on the page—so any attempt to import the instance directly from a particular module isn't going to work here.

The thing I'm suggesting is to have an intermediate module in reports/js that defines `getStandardHQReport` and `setStandardHQReport` that acts as PO box for the instantiating model and these js files to communicate about the value of `standardHQReport`. Both the benefit and the drawback of this approach is that it's close to what's presently done.

kaapstorm:
I chatted with Claire to gauge level of urgency. I'm going to merge this, and open a second ticket to implement Danny's suggestion.

PR #15894, opened 2017-04-26
exclude root node from metrics

snopoke:
not sure why this is necessary since root node should never be a leaf
but perhaps in cases where there are errors or timeouts none of the
nested timers get executed


benrudolph (approved):
strange

PR #15895, opened 2017-04-26
Call get_short_odk_url to reset app_code

kaapstorm:
[FB 251649](http://manage.dimagi.com/default.asp?251649)

The `app_code` observable was getting set with the Java short code, but wasn't getting reset back to the Android short code.

### DIFF
--- a/corehq/apps/app_manager/static/app_manager/js/releases.js
+++ b/corehq/apps/app_manager/static/app_manager/js/releases.js
@@ -161,6 +161,7 @@ hqDefine('app_manager/js/releases.js', function () {
             ga_track_event('App Manager', 'Deploy Button', self.id());
             analytics.workflow('Clicked Deploy');
             $.post(releasesMain.options.urls.hubspot_click_deploy);
+            self.get_short_odk_url();
         };
 
         self.clickScan = function() {
--- a/corehq/apps/app_manager/static/app_manager/js/releases.js
+++ b/corehq/apps/app_manager/static/app_manager/js/releases.js
@@ -93,7 +93,10 @@ hqDefine('app_manager/js/releases.js', function () {
             if (!(ko.utils.unwrapObservable(self[url_type])) || self.build_profile()) {
                 return self.generate_short_url(url_type);
             } else {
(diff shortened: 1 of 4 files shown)


corehq/apps/app_manager/static/app_manager/js/releases.js
```diff
@@ -161,6 +161,7 @@ hqDefine('app_manager/js/releases.js', function () {
             ga_track_event('App Manager', 'Deploy Button', self.id());
             analytics.workflow('Clicked Deploy');
             $.post(releasesMain.options.urls.hubspot_click_deploy);
+            self.get_short_odk_url();
```
dannyroberts: I'm not sure I understand why this is necessary.

Also, and maybe this is related, I noticed `get_short_odk_url` is just a function and not a computed observable, so it won't get re-run when, say, `self.include_media()`, which it depends on, changes. You can change that by using

```javascript
self.get_short_odk_url = ko.computed(function() {
};)
```
instead.

kaapstorm: This resolves the workflow where the user clicks "Deploy", then clicks "Send to phone via SMS", then clicks "Close". 

`get_short_code_url()` calls `generate_short_url()` (if necessary) with the right `url_type` (i.e. Java / Android / Android + media) which in turn sets the Observable `app_code()`.

Toggling "Include media" will also set `app_code()` correctly.

It's `app_code()` that gets rendered.

But the "Deploy" button (in the case of this line of code) and the "Download to Android" accordion panel don't reset `app_code()` if it is storing a URL for Java. This PR should resolve that.

I should have written all that up in the PR comment. I hope this clarifies it.

kaapstorm:
(Thanks for build_app_manager_v2_diffs!)

dannyroberts:
Am I still holding off on merge @kaapstorm

kaapstorm:
Ready. "Hold off" dropped.

kaapstorm:
Forgot buddy @biyeun ... better late than never. :grimacing:

biyeun (approved):

PR #15897, opened 2017-04-26
standard_hq_report.js

orangejenny:
https://manage.dimagi.com/default.asp?252807

@dannyroberts 

I did implement a module just to keep track of `standardHQReport`, but only use it to get, not to set. Tested locally with UCRs, standard reports (the WAR), and fixtures.

### DIFF
--- a/corehq/apps/reports/static/reports/js/standard_hq_report.js
+++ b/corehq/apps/reports/static/reports/js/standard_hq_report.js
@@ -1,3 +1,4 @@
+/* globals COMMCAREHQ_MODULES, standardHQReport */
 /*
     Ugly half-measure, because reports and UCR traditionally depend on a global standardHQReport
     variable that's defined in several different places. The UCR version of standardHQReport now
--- a/corehq/apps/reports/static/reports/js/standard_hq_report.js
+++ b/corehq/apps/reports/static/reports/js/standard_hq_report.js
@@ -1,0 +1,23 @@
+/*
+    Ugly half-measure, because reports and UCR traditionally depend on a global standardHQReport
+    variable that's defined in several different places. The UCR version of standardHQReport now
+    lives in userreports/js/configurable_report.js, whereas the other standardHQReport vars are
+    still defined globally. This module's sole purpose is to fetch the correct standardHQReport,
+    if it exists at all.
+*/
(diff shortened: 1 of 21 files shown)


corehq/apps/reports/static/reports/js/standard_hq_report.js
```diff
@@ -0,0 +1,23 @@
...
+    Ugly half-measure, because reports and UCR traditionally depend on a global standardHQReport
+    variable that's defined in several different places. The UCR version of standardHQReport now
+    lives in userreports/js/configurable_report.js, whereas the other standardHQReport vars are
+    still defined globally. This module's sole purpose is to fetch the correct standardHQReport,
+    if it exists at all.
+*/
+hqDefine("reports/js/standard_hq_report.js", function() {
+    var get = function() {
+        if (typeof standardHQReport !== 'undefined') {
+            return standardHQReport;
```
dimagimon: 'standardHQReport' is not defined. (no-undef)

corehq/apps/reports/static/reports/js/standard_hq_report.js
```diff
@@ -0,0 +1,23 @@
...
+        }
+        var ucr = "userreports/js/configurable_report.js";
+        if (typeof COMMCAREHQ_MODULES[ucr] !== 'undefined') {
```
dimagimon: 'COMMCAREHQ_MODULES' is not defined. (no-undef)

dannyroberts (approved):
Great, looks good to me. Just looking at it I can't reason with whether this is loaded before or after the js that sets the global variables in other pages, but if it works when you test it locally on non-ucr pages, then this is great.

benrudolph:
hot fix merging

orangejenny:
@dannyroberts I think loading order is okay because any attempts to access `standardHQReport` should be after the global is defined (or the previous system wouldn't have worked), and the new `standard_hq_report.js` module is always defined by the time that global is defined (because the global is defined in the js-inline block, which is after the js block that includes standard_hq_report.js...the exception to this is `report_loader.html`, where I put the standard_hq_report.js include immediately before the `standardHQReport` global is defined). Convoluted, I admit.

PR #15898, opened 2017-04-26
update text for New App Builder rollback

biyeun:
@orangejenny 
Was reading New Form Builder instead of New App Builder
now looks like
screen shot 2017-04-26 at 11 30 16 am


orangejenny (approved):

orangejenny:
thanks!

PR #15798, opened 2017-04-19
share case types and their properties across all apps

dannyroberts:
not just "Case Sharing" apps

https://manage.dimagi.com/default.asp?251731

Team approved (#product)
screen shot 2017-04-19 at 11 57 13 am

PR #15901, opened 2017-04-26
get_casedb_schema: don't load cases unless needed

orangejenny:
https://manage.dimagi.com/default.asp?252644

@dannyroberts / @millerdev

### DIFF
--- a/corehq/apps/app_manager/util.py
+++ b/corehq/apps/app_manager/util.py
@@ -409,7 +409,7 @@ def get_casedb_schema(form):
     This lists all case types and their properties for the given app.
     """
     app = form.get_app()
-    base_case_type = form.get_module().case_type
+    base_case_type = form.get_module().case_type if form.requires_case() else None
     case_types = app.get_case_types() | get_shared_case_types(app)
     per_type_defaults = get_per_type_defaults(app.domain, case_types)
     builder = ParentCasePropertyBuilder(app, ['case_name'], per_type_defaults)


corehq/apps/app_manager/util.py
```diff
@@ -409,7 +409,7 @@ def get_casedb_schema(form):
     This lists all case types and their properties for the given app.
     """
     app = form.get_app()
-    base_case_type = form.get_module().case_type
+    base_case_type = form.get_module().case_type if form.requires_case() else None
```
millerdev: This looks like a good change, but wondering if it will hide case properties from people who were previously seeing/using them?

Also, I'm wondering if this will fix the issue or if it would be better to roll back #15798 and give that a second try that doesn't have so much impact on performance? @dannyroberts may want to chime in on that.

dannyroberts: Yeah, I'm a bit mixed. It's kind of a dumb reason that things weren't slow before—it was excluding non-case-sharing apps and it turns out that most dummy apps happen to be non-case-sharing because that's the default when you create an app. Seems like there should be a better way than relying on that accident. That said, if we think this is a widespread problem—that is, that it's common to have many, many (dummy) apps in your domain—then it's possible the right thing to do is in fact to revert

orangejenny: > This looks like a good change, but wondering if it will hide case properties from people who were previously seeing/using them?

I believe the strict answer to this question is "no." `get_casedb_schema` is only used by `get_form_data_schema`, which is only used by vellum to display app properties. Forms that don't require cases don't have the app properties window, so they wouldn't have displayed case properties at all. (I'm actually a bit confused by that - even if usercase in the form builder turned on, the app properties window doesn't seem to show unless the form requires a case. Am I not using user properties in the form builder correctly?)

That said, I do think it's worth considering rolling back or further discussing the case sharing change, because this PR will improve performance for forms that don't require cases, but forms that do require cases have presumably also had a performance hit and they won't benefit from this change.

millerdev: > even if usercase in the form builder turned on, the app properties window doesn't seem to show unless the form requires a case

That might be a bug in [`_form_uses_case`](https://github.com/dimagi/commcare-hq/blob/a59656f1285374f3474eb82bcbfd3e43fd1b981e/corehq/apps/app_manager/views/formdesigner.py#L57), which is used to [enable the data browser](https://github.com/dimagi/commcare-hq/blob/master/corehq/apps/app_manager/views/formdesigner.py#L98-L99). `_form_uses_case` should probably do

```py
return (
    (module and module.case_type and form.requires_case())
    or is_usercase_in_use(domain)
)
```

millerdev: > > This looks like a good change, but wondering if it will hide case properties from people who were previously seeing/using them?
>
> I believe the strict answer to this question is "no."

In that case, I approve this PR. I'm still in favor of continuing to evaluate whether #15798 can be further optimized or should be reverted.

millerdev:
test failure looks real
```
ERROR: corehq.apps.app_manager.tests.test_util:SchemaTest.test_get_case_sharing_hierarchy
  File "/mnt/commcare-hq-ro/corehq/apps/app_manager/tests/test_util.py", line 302, in test_get_case_sharing_hierarchy
    subsets["parent"]["name"]))
KeyError: 'parent'
```

corehq/apps/app_manager/util.py
emord: Any reason to not make this return from the data dictionary if it exists? I think that would be the preferred way going forward

dannyroberts: Maybe I just need to know more about that. How is it kept up to date? Is it recalculated every time an app is saved?

emord: Yeah. it's updated everytime an app saves a new property or a new property is imported from case import.

dannyroberts: chatted with emord, seems like data dictionary is still missing parent/child relationships, so that may not be so simple.

dannyroberts: backing up though, that is sort of the whole point of (or at least one of the main points of) the data dictionary, that it could then be used to add richness to vellum's knowledge of case properties. I think we're just not quite ready yet

dannyroberts: As for reverting, it may unfortunately be worth it. Timings for the vellum schema URL (my change was merged Apr. 19): https://app.datadoghq.com/metric/explorer?live=true&page=0&is_auto=false&from_ts=1490650456272&to_ts=1493242456272&tile_size=m&exp_metric=nginx.timings&exp_scope=url%3A%2Fa%2F_%2Fapps%2Fschema%2Fform%2F_%2F&exp_agg=max&exp_row_type=metric.

dannyroberts:
:'( closing this in favor of reverting the other pr: https://github.com/dimagi/commcare-hq/pull/15906

orangejenny:
😢 indeed

millerdev:
This seems like it might still be worthwhile, even if #15798 is reverted. Or am I missing something?

orangejenny:
@millerdev Yeah, I was being lazy by deleting this, it'll be a genuine performance improvement for forms that use usercase but not other cases. Will re-open and fix test.

PR #15908, opened 2017-04-26
Add BETS payment confirmation endpoints

esoergel:
https://docs.google.com/document/d/1RPPc7t9NhRjOOiedlRmtCt3wQSjAnWaj69v2g7QRzS0/edit 
@mkangia
@proteusvacuum @NoahCarnahan FYI

### DIFF
--- a/custom/enikshay/integrations/bets/views.py
+++ b/custom/enikshay/integrations/bets/views.py
@@ -6,19 +6,13 @@ import json
 from django.views.decorators.http import require_POST
 from dimagi.utils.web import json_response
 from dimagi.ext import jsonobject
-from jsonobject.exceptions import BadValueError, WrappingAttributeError
-# import JsonObject, StringProperty, ListProperty, DictProperty
-# from dimagi.utils.decorators.memoized import memoized
+from jsonobject.exceptions import BadValueError
 from corehq.apps.domain.decorators import login_or_digest_or_basic_or_apikey
 
-# TODO does this auth check domain?
-
-
-from corehq.form_processor.interfaces.dbaccessors import CaseAccessors
-from casexml.apps.case.const import CASE_INDEX_EXTENSION, UNOWNED_EXTENSION_OWNER_ID
 from corehq.form_processor.exceptions import CaseNotFound
-from casexml.apps.case.mock import CaseFactory, CaseStructure, CaseIndex
+from corehq.form_processor.interfaces.dbaccessors import CaseAccessors
 from corehq.apps.hqcase.utils import update_case
+from .const import BETS_EVENT_IDS
 
 
 class ApiError(Exception):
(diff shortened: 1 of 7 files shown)


custom/enikshay/integrations/bets/views.py
```diff
@@ -0,0 +1,130 @@
...
+from jsonobject.exceptions import BadValueError
+from corehq.apps.domain.decorators import login_or_digest_or_basic_or_apikey
+
+from corehq.form_processor.exceptions import CaseNotFound
+from corehq.form_processor.interfaces.dbaccessors import CaseAccessors
+from corehq.apps.hqcase.utils import update_case
+from .const import BETS_EVENT_IDS
+
+
+class ApiError(Exception):
```
NoahCarnahan: Looks like this class isn't used anywhere

custom/enikshay/integrations/bets/views.py
```diff
@@ -0,0 +1,130 @@
...
+        )
+
+    update_case(domain, update.case_id, case_properties=update.properties)
+
+    return json_response({'status': "success"})
+
+
+@require_POST
+@csrf_exempt
+@login_or_digest_or_basic_or_apikey()
```
NoahCarnahan: Does `login_or_digest_or_basic_or_apikey` check that the domain matches? Also, seems like maybe only users with specific permissions should be able to access these URLs? Otherwise any user on the domain could muck things up in here.

custom/enikshay/tests/test_bets_updates.py
```diff
@@ -0,0 +1,144 @@
...
+        cls.web_user.delete()
+        super(TestBetsUpdates, cls).tearDownClass()
+
+    def make_request(self, view, data):
+        factory = RequestFactory()
+        request = factory.post("/a/enikshay/bets/{}".format(view.__name__),
+                               data=json.dumps(data),
+                               content_type='application/json')
+        request.user = self.web_user.get_django_user()
+        return view(request, self.domain)
```
NoahCarnahan: (non-blocking)
I generally prefer using django [`TestClient`](https://docs.djangoproject.com/en/1.10/topics/testing/tools/#the-test-client) instead of creating the request and view separately, as it creates a closer to production scenario, since the request will be passed through middleware and all that.

NoahCarnahan (changes requested):

custom/enikshay/integrations/bets/views.py
esoergel: Oh yeah, thanks - I changed direction and don't need this anymore.

custom/enikshay/integrations/bets/views.py
esoergel: It does - I had the same thought, but it calls "login_and_domain_required" somewhere down the stack.  This seems to be what we generally use for API auth, but I agree that some role validation would be a good idea.

custom/enikshay/integrations/bets/views.py
```diff
@@ -0,0 +1,130 @@
...
+        super(ApiError, self).__init__(msg)
+
+
+class VoucherUpdate(jsonobject.JsonObject):
+    voucher_id = jsonobject.StringProperty(required=True)
+    payment_status = jsonobject.StringProperty(required=True, choices=['success', 'failure'])
+    payment_amount = jsonobject.IntegerProperty(required=False)
+    failure_description = jsonobject.StringProperty(required=False)
+
+    case_type = 'voucher'
```
proteusvacuum: nit, we have constants for the case types in `enikshay.case_utils`

custom/enikshay/integrations/bets/views.py
```diff
@@ -0,0 +1,130 @@
...
+                'reason_rejected': self.failure_description,
+                'date_rejected': datetime.datetime.utcnow().date().isoformat(),
+            }
+
+
+class IncentiveUpdate(jsonobject.JsonObject):
+    beneficiary_id = jsonobject.StringProperty(required=True)
+    episode_id = jsonobject.StringProperty(required=True)
+    payment_status = jsonobject.StringProperty(required=True, choices=['success', 'failure'])
+    payment_amount = jsonobject.IntegerProperty(required=False)
```
proteusvacuum: In theory I think they can pay decimal amounts?

custom/enikshay/integrations/bets/views.py
NoahCarnahan: 👍

custom/enikshay/integrations/bets/views.py
```diff
@@ -0,0 +1,130 @@
...
+def get_case(domain, case_id):
+    case_accessor = CaseAccessors(domain)
+    return case_accessor.get_case(case_id)
+
+
+def _update_case_from_request(request, domain, update_model):
+    try:
+        request_json = json.loads(request.body)
+    except ValueError:
+        return json_response({"error": "Malformed JSON"}, status_code=400)
```
proteusvacuum: do you think we should log these errors somewhere? from our experience with 99DOTS and others, it is useful to monitor what has been going on since we can't necessarily rely on their processes to alert us when things have gone sour.

proteusvacuum (commented):
Very nice! Couple of comments, but looks great.

custom/enikshay/integrations/bets/views.py
esoergel: I think the server logs should store the requests and the status code returned, right?  What did you have in mind?

proteusvacuum: something we receive daily in "report style" that we can send to BETS and say "hey look, we got all these bad requests from you". Even just a separate log that doesn't have all the other requests in it could be a start.

NoahCarnahan (approved):
Thanks for the changes!

PR #15909, opened 2017-04-27
App Manager v2: show build profiles

orangejenny:
https://manage.dimagi.com/default.asp?252840

I had hidden this a while ago just to get it out of the way for user testing.

@dannyroberts / @biyeun

### DIFF
--- a/corehq/apps/app_manager/templates/app_manager/v2/partials/releases.html
+++ b/corehq/apps/app_manager/templates/app_manager/v2/partials/releases.html
@@ -61,7 +61,7 @@
         var app_langs = {{ app.langs|JSON }};
         var app_profiles = {{ app.build_profiles|JSON }}
 
-        {% if build_profile_access %}
+        {% if not intro_only and build_profile_access %}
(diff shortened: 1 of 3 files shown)


corehq/apps/app_manager/templates/app_manager/v2/partials/releases.html
```diff
@@ -61,7 +61,7 @@
         var app_langs = {{ app.langs|JSON }};
         var app_profiles = {{ app.build_profiles|JSON }}
 
-        {% if build_profile_access %}
+        {% if not intro_only and build_profile_access %}
```
dannyroberts: For the uninitiated, what does `intro_only` mean?

dannyroberts (approved):

corehq/apps/app_manager/templates/app_manager/v2/partials/releases.html
orangejenny: It's when your app is new/empty and instead of the deploy page you get the fancy new help content:
screen shot 2017-04-27 at 12 13 02 pm

PR #15911, opened 2017-04-27
App manager v2: fix usercase

orangejenny:
https://manage.dimagi.com/default.asp?252785

v2 wasn't including case config-related templates unless the form used cases, so forms with usercase but no regular cases weren't working. Also made a couple of small unrelated fixes.

@dannyroberts / @biyeun


corehq/apps/app_manager/templates/app_manager/v2/form_view.html
```diff
@@ -8,7 +8,7 @@
 {% endblock %}
 
 {% block case_management_js %}{{ block.super }}
-  {% if form.uses_cases %}
+  {% if form.source %}
```
orangejenny: @biyeun `master` uses `form.source` but you changed it in https://github.com/dimagi/commcare-hq/commit/c8403b6d074170dea37ebc132b866a370cf45b0f - do you remember why?

biyeun: @orangejenny I changed this because the surrounding logic that required it used the same requirements. I guess I didn't take usercase into account

orangejenny: cool, this change sounds safe, then

PR #15915, opened 2017-04-27
App Manager v2: Change logic for forms added to new case lists

orangejenny:
This came out of conversation with Clark, who found it weird that every case module has a reg form. I think it makes sense that it's more common for people to want multiple modules because they want multiple menus on the phone, rather than because they want multiple reg+followups units. So this updates the new module logic to only add a reg form if it's the first case list in the app (or if there are no other case modules that have a reg form...don't want people to get into a situation where they delete the module that has the reg form and then can't get it back).

@biyeun how does that sound? can offline if easier.
cc @dannyroberts

### DIFF
--- a/corehq/apps/app_manager/views/modules.py
+++ b/corehq/apps/app_manager/views/modules.py
@@ -980,23 +980,31 @@ def new_module(request, domain, app_id):
         form_id = None
         if toggles.APP_MANAGER_V2.enabled(request.user.username):
             if module_type == 'case':
-                # registration form
-                register = app.new_form(module_id, "Register", lang)
-                register.actions.open_case = OpenCaseAction(condition=FormActionCondition(type='always'))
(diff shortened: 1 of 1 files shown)


biyeun:
@orangejenny will offline, but I think we should run some user tests with this in place

orangejenny:
@biyeun 

Chatted with @ctsims and decided to revert this. Main point was that having multiple case types (each with a reg form) is of much greater value than having multiple modules/menus for a single case type, so we should facilitate that, regardless of whether or not it's the more common workflow for new users. 

Also, in a future (and better!) world, module/case configuration won't be so tightly coupled with the phone's menu structure, and we might as well design with that world in mind, where those two things are more independent, rather than constrain how people set up their modules/cases based on expectations about what they want the phone menu to look like.

Happy to offline further with you or @nsclark if desired.

PR #15917, opened 2017-04-27
Ability to disable sorting and log scroll times

benrudolph:
Edit: completely changed this PR

This adds the ability to disable sorting and also logs time it takes for a scroll query in datadog. The docs specifically say that `scan` should disable sorting, but I'm a little skeptical. This PR will help be figure out if scroll queries are taking longer and longer time as well. hoping this will help figure out if sorting affects it at all

cc: @calellowitz

### DIFF
--- a/corehq/toggles.py
+++ b/corehq/toggles.py
@@ -376,6 +376,13 @@ USER_CONFIGURABLE_REPORTS = StaticToggle(
     help_link='https://confluence.dimagi.com/display/RD/User+Configurable+Reporting', 
 )
 
+EXPORT_NO_SORT = StaticToggle(
+    'export_no_sort',
+    'Do not sort exports',
+    TAG_ONE_OFF,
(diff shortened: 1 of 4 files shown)


corehq/toggles.py
```diff
@@ -376,6 +376,13 @@ def toggle_values_by_name(username=None, domain=None):
     help_link='https://confluence.dimagi.com/display/RD/User+Configurable+Reporting', 
 )
 
+EXPORT_NO_SORT = StaticToggle(
+    'export_no_sort',
+    'Do not sort exports',
+    TAG_ONE_OFF,
+    [NAMESPACE_DOMAIN],
```
emord: should you add `always_enabled` ?

benrudolph: i want to be able to turn it on and off for icds in order to test the difference first

PR #15919, opened 2017-04-27
Make sure reports are loaded so location_safe is applied

kkrampa:
@esoergel @dannyroberts 
https://manage.dimagi.com/default.asp?251837
Reports are not loaded by django on server start so location safe is not applied to ICDS reports. I think this might cause the problem with permissions for users described in the case above.

### DIFF
--- a/custom/icds_reports/__init__.py
+++ b/custom/icds_reports/__init__.py
@@ -1,0 +1,10 @@
+from django.apps import AppConfig
+
+
+class ICDSReportsAppConfig(AppConfig):
+    name = 'custom.icds_reports'
+
+    def ready(self):
+        import custom.icds_reports.reports.reports  # noqa
+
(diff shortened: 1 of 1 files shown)


custom/icds_reports/__init__.py
```diff
@@ -0,0 +1,10 @@
+from django.apps import AppConfig
+
+
+class ICDSReportsAppConfig(AppConfig):
+    name = 'custom.icds_reports'
+
+    def ready(self):
+        import custom.icds_reports.reports.reports
```
dimagimon: F401 'custom' imported but unused

dannyroberts (approved):

kkrampa:
Can we merge this and deploy to icds to check if this will fix the problem?

dannyroberts:
Yup, sorry for the wait!

PR #15920, opened 2017-04-27
Only update for nikshay

benrudolph:
missed this one @calellowitz


benrudolph:
@emord this one would be nice. most likely will conflict with the other couch user one

emord:
Maybe we should add something like an `only_enabled` to toggles so it doesn't hit couch for the toggle every time

benrudolph:
ah i see, yea that would be nice especially for toggles like these. i think for now this is still a win because i think it'll be better to hit the toggle db than the user db

PR #15922, opened 2017-04-27
don't sync mobile reports on every sync

snopoke:
Allow setting sync interval as domain default and also per user.

buddy @nickpell


corehq/apps/app_manager/fixtures/mobile_ucr.py
```diff
@@ -18,6 +18,11 @@
 from corehq.apps.app_manager.dbaccessors import get_apps_in_domain
 
 
+def _should_sync(last_sync_log):
+    sync_interval = 4 * 60 * 60
```
snopoke: needs to be configurable per domain

corehq/apps/app_manager/fixtures/mobile_ucr.py
```diff
@@ -1,5 +1,5 @@
 from collections import defaultdict
-from datetime import datetime
+from datetime import datetime, timedelta
```
dimagimon: F401 'timedelta' imported but unused

corehq/apps/app_manager/fixtures/mobile_ucr.py
snopoke: Unfortunately since all mobile reports are under a single XML element they can't be timed individually.

benrudolph: might be difficult but would also be nice if it respected the `overwrite_cache` param

NoahCarnahan:
Not sure if you already have a plan for this, but I think we'll need some way to be able to force the restore to happen while developing and testing new reports, otherwise the app building process will be super painful.

corehq/apps/domain/models.py
```diff
@@ -371,6 +371,9 @@ class Domain(QuickCachedDocumentMixin, Document, SnapshotMixin):
     requested_report_builder_trial = StringListProperty()
     requested_report_builder_subscription = StringListProperty()
 
+    # seconds between sending mobile UCRs to users. Can be overridden per user
+    defaul_mobile_ucr_sync_interval = IntegerProperty()
```
NoahCarnahan: typo in this property name. Missing "t" in "default".

corehq/apps/app_manager/fixtures/mobile_ucr.py
```diff
@@ -18,18 +17,29 @@
...
+    def __call__(self, restore_state):
         """
         Generates a report fixture for mobile that can be used by a report module
         """
-        assert isinstance(restore_user, OTARestoreUser)
-
-        if not toggles.MOBILE_UCR.enabled(restore_user.domain):
+        restore_user = restore_state.restore_user
+        if not toggles.MOBILE_UCR.enabled(restore_user.domain) or not _should_sync(restore_state):
             return []
```
NoahCarnahan: Will the mobile keep around the previously synced report fixtures if the latest sync doesn't include any?

NoahCarnahan (approved):
Looks great

nickpell:
conflicts with master

corehq/apps/app_manager/fixtures/mobile_ucr.py
snopoke: Yes it should - we use a similar pattern for other fixtures

NoahCarnahan:
Looks like just an import error:
```
ERROR: corehq.apps.fixtures.tests.test_fixture_data:FixtureDataTest.test_empty_data_types
  File "/mnt/commcare-hq-ro/corehq/apps/fixtures/tests/test_fixture_data.py", line 191, in test_empty_data_types
    fixtures = fixturegenerators.item_lists(self.user.to_ota_restore_user(), V2)
NameError: global name 'V2' is not defined
```

snopoke:
:white_check_mark:

PR #15927, opened 2017-04-27
Drop app manager specific parts of USER_TESTING_SIMPLIFY

orangejenny:
Left the flag itself and the logic for removing the global nav, so we can use it in the future for non-app-manager-v2 user testing.

@biyeun / @dannyroberts 

`w=1`


dannyroberts:
(merge conflicts)

orangejenny:
fixed conflicts, it was all conflicted v2 diff files

biyeun (approved):

PR #15930, opened 2017-04-27
fix _determine_couch_databases for production

dannyroberts:
where localsettings will not have `from dev_settings import *`

found this while chatting through it with @emord (thanks!)


dannyroberts:
cc: @snopoke

dannyroberts:
@orangejenny buddy

PR #15931, opened 2017-04-27
Increase case search limit to 100 results

proteusvacuum:
@sravfeyn 
@NoahCarnahan 

FYI @ctsims, since we talked about this.


sravfeyn (approved):

sravfeyn:
Test fail looks like it's related

PR #15934, opened 2017-04-27
App manager v2: reset menu links after moving form between modules

orangejenny:
https://manage.dimagi.com/default.asp?252555

The "from" module's links weren't being updated, and module ids in form links weren't getting updated when moving between modules.

@biyeun / @dannyroberts


biyeun:
yikes. thanks

biyeun (approved):

PR #15937, opened 2017-04-27
App Manager V2: Hide Case Settings in survey module

biyeun:
@orangejenny // buddy @kaapstorm 

fixes https://manage.dimagi.com/default.asp?252950#1324073

### DIFF
--- a/corehq/apps/app_manager/templates/app_manager/v2/partials/module_view_settings.html
+++ b/corehq/apps/app_manager/templates/app_manager/v2/partials/module_view_settings.html
@@ -10,13 +10,15 @@
         {% block settings_fields %}
 
 
-          {% if not module.is_surveys %}
+          {% if child_module_enabled or module.module_type == 'shadow' or app.advanced_app_builder or module.put_in_root or app.grid_display_for_some_modules or value or ko_value or not module.is_surveys %}
           <div class="panel panel-appmanager">
             <div class="panel-heading">
               <h4 class="panel-title panel-title-nolink">{% trans "Case Settings" %}</h4>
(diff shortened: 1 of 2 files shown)


orangejenny (changes requested):
This looks like it's going to hide a bunch of legitimate settings from survey modules: parent module, source module, menu mode, etc.

biyeun:
@orangejenny so maybe what needs to happen if parent, source, and menu module are not showing up, hide the div. I think the issue in the ticket was it was just an empty panel

orangejenny:
Yeah, the panel needs to be hidden if nothing is inside of it, some kind of {% if not shadow and not child and and not advanced_app_builder ... blah blah %}

biyeun:
@orangejenny  responded to your feedback btw

corehq/apps/app_manager/templates/app_manager/v2/partials/module_view_settings.html
```diff
@@ -9,14 +9,15 @@
     <fieldset>
         {% block settings_fields %}
 
+
+          {% if child_module_enabled or module.module_type == 'shadow' or app.advanced_app_builder or module.put_in_root or app.grid_display_for_some_modules or value or ko_value or not module.is_surveys %}
```
orangejenny: I think instead of `value or ko_value` you want to check for `module.module_filter`

orangejenny (changes requested):

biyeun:
@orangejenny hey responded btw

orangejenny (approved):

PR #15938, opened 2017-04-28
For the sake of discussion ...

kaapstorm:
Don't merge. This is just to help unpack [FB 252987](https://manage.dimagi.com/default.asp?252987).

I can't see why a `CachedResponse` instance has `self.payload` set to anything other than a `CachedPayload` instance. But I can imagine how it might happen. This file confuses me. It seems the words "response" and "payload" get used interchangeably. e.g. Methods called "get_payload" return Response instances and `Response.as_string()` returns a payload.

I haven't tested this code locally. It might resolve the error, but **it's not the right approach**. I need to figure out how `CachedResponse` ends up with its `payload` set to a file instead of a `Payload` instance. I'll keep digging, but all hints and feedback welcome.

@dannyroberts  cc @snopoke @proteusvacuum buddy @biyeun

### DIFF
--- a/corehq/ex-submodules/casexml/apps/phone/restore.py
+++ b/corehq/ex-submodules/casexml/apps/phone/restore.py
@@ -413,11 +413,26 @@ class CachedResponse(object):
         return self.payload.as_string()
 
     def get_http_response(self):
+
+        def get_payload_content_length():
+            if isinstance(self.payload, file):
+                offset = self.payload.tell()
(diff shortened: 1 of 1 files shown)


kaapstorm:
cc @amstone326

dannyroberts:
Going to close this in favor of https://github.com/dimagi/commcare-hq/pull/15938

PR #15939, opened 2017-04-28
test couch 2.0

snopoke:
@dimagi/scale-team

### DIFF
--- a/docker/hq-compose.yml
+++ b/docker/hq-compose.yml
@@ -35,7 +35,8 @@ postgres:
     - ${VOLUME_PREFIX}postgresql:/var/lib/postgresql/data
 
 couch:
-  image: klaemo/couchdb:1.6.1
+  image: klaemo/couchdb:2.0-dev
+  command: ["--with-haproxy", "--with-admin-party-please", "-n", "1"]
   expose:
     - "5984"
   volumes:
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,8 +7,7 @@ env:
   matrix:
     # need node installs for app_manager.XpathValidatorTest
     - TEST=python NOSE_DIVIDED_WE_RUN=05 JS_SETUP=yes
-    - TEST=python NOSE_DIVIDED_WE_RUN=6a JS_SETUP=yes
-    - TEST=python NOSE_DIVIDED_WE_RUN=bf JS_SETUP=yes
+    - TEST=python NOSE_DIVIDED_WE_RUN=6f JS_SETUP=yes
     - TEST=python-sharded-and-javascript JS_SETUP=yes
   global:
     # TRAVIS_HQ_USERNAME
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,7 +7,8 @@ env:
   matrix:
     # need node installs for app_manager.XpathValidatorTest
     - TEST=python NOSE_DIVIDED_WE_RUN=05 JS_SETUP=yes
-    - TEST=python NOSE_DIVIDED_WE_RUN=6f JS_SETUP=yes
+    - TEST=python NOSE_DIVIDED_WE_RUN=6a JS_SETUP=yes
+    - TEST=python NOSE_DIVIDED_WE_RUN=bf JS_SETUP=yes
     - TEST=python-sharded-and-javascript JS_SETUP=yes
   global:
     # TRAVIS_HQ_USERNAME
--- a/corehq/ex-submodules/couchforms/tests/test_meta.py
+++ b/corehq/ex-submodules/couchforms/tests/test_meta.py
@@ -4,11 +4,10 @@ from datetime import date, datetime
 from django.test import TestCase
 from django.conf import settings
 
+from casexml.apps.case.tests.util import delete_all_xforms
+from corehq.form_processor.tests.utils import use_sql_backend, post_xform
 from corehq.util.test_utils import TestFileMixin
 from couchforms.datatypes import GeoPoint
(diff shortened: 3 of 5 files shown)


docker/hq-compose.yml
```diff
@@ -35,7 +35,8 @@ postgres:
     - ${VOLUME_PREFIX}postgresql:/var/lib/postgresql/data
 
 couch:
-  image: klaemo/couchdb:1.6.1
+  image: klaemo/couchdb:2.0-dev
```
emord: They moved this to an official apache repo https://github.com/apache/couchdb-docker. haven't looked into moving it yet

emord (commented):
@snopoke I mentioned this on slack but I think this is due to most of our documents not using `SafeSaveDocument`. Because tests save and check the same document very quickly we'll probably need to either have docker only use one node, or convert every doc to safe saves

EDIT: This still uses the proxy and cluster setup. The only difference is that it's using one node instead of 3 in the backend

emord:
@snopoke I think you mentioned that you could not reproduce these errors locally. That may be due to the way we run the docker script. You should be able to reproduce them by running 

```
./scripts/docker down # this will stop and totally remove any docker processes
git checkout sk/couch2
./scripts/docker up # this recreates the containers
./manage.py test failing.test
```
You may also need to manually drop the postgres test db as reusedb wasn't successfully flushing for me

If you want to keep your data you'll have to have a couch 1.6.1 docker running alongside the 2.0 and replicate it over. You can do that by changing the ports that are exposed and going through the replication process. For now I've just been switching back and forth between 1.6.1 and 2.0.

As for the two failing tests that have come up so far:

```
FAIL: testapps.test_pillowtop.tests.test_changes:ChangeFeedDbTest.test_couch_filter
  File "/mnt/commcare-hq-ro/testapps/test_pillowtop/tests/test_changes.py", line 49, in test_couch_filter
    self.assertEqual(form._id, change['id'])
AssertionError: u'61faeabb63a89b6ce1dd7a8904bae97f' != '5cc7f3fe058b488dbdaa95704e13cd76'
```

This fails intermittently for me. We'll need to change this test, but I'm not sure what it should look like at the moment. Due to the sharding of databases change feeds are no longer guaranteed to be sequential. reference: https://blog.couchdb.org/2016/08/17/migrating-to-couchdb-2-0/

```
FAIL: corehq.apps.consumption.tests:GetDefaultConsumptionTestCase.testGetForId
  File "/mnt/commcare-hq-ro/corehq/apps/consumption/tests.py", line 76, in testGetForId
    self.assertEqual(5, self.consumption_method(domain, product_id, type_id, supply_point_id))
AssertionError: 5 != Decimal('4.0')
```
The `corehq.apps.consumption.tests:GetDefaultConsumptionTestCase` fails 3 tests consistently for me. I haven't dug into consumption enough to understand why, but these are 3 year old tests so they're hopefully working with big couch backends.

I've removed --stop from the travis config so that we can see all the test failures. will pick this back up tomorrow if you don't get to it first

emord:
I think I tracked down the consumption issue. This was a weird one.

http://docs.couchdb.org/en/2.0.0/couchapp/views/intro.html#reversed-results says that it is sorting by key, not value. It looks like 1.6.1 did the same thing.

https://github.com/dimagi/commcare-hq/blob/sk/couch2/corehq/apps/consumption/shortcuts.py#L14 says it should return the most specific key.

The keys start with the most specific key, so I think that that should be returned first. If that's the case, descending would actually return it last. I have no idea how this would have been working on 1.6

(also adding a commit to kill one of the travis workers since these tests are really fast)

snopoke:
I didn't run `./scripts/docker down`, I just removed the couch 1.6 container and started up a 2.0 container. It was definitely using 2.0 in the tests so not sure why I wasn't able to reproduce. Anyway, great that you got this working!

docker/hq-compose.yml
snopoke: doesn't look like the official 2.0 image is published yet: https://hub.docker.com/r/library/couchdb/

benrudolph:
any harm in merging this? feel like we could use the test performance boost!

snopoke:
Fine by me. Rest of the team would need to download the 2.0 image next time
they restart their docker containers. It would also result in data loss
locally unless they migrate.

benrudolph:
hmm ok, perhaps best to come up with a plan to help people migrate their data

emord:
I'll send out some instructions on migrating data when I get a chance (probably later today)

emord:
sent instructions out to dev list. there was an intermittent failing test which i've hopefully solved in the last commit

testapps/test_pillowtop/tests/test_form_pillow.py
```diff
@@ -3,6 +3,7 @@
 
 from corehq.apps.app_manager.tests.app_factory import AppFactory
 from corehq.apps.app_manager.models import Application
+from corehq.apps.builds.models import CommCareBuildConfig
```
dimagimon: F401 'CommCareBuildConfig' imported but unused

emord:
I'm reverting the change to go down to two test jobs. I can't explain it but there's some issue whenever tests hashing with 1 and 6 are included in the same test. I tried to run all tests hashed with 1 and the test failing on 6 together locally but couldn't get it to fail.

We could go with tests hashed on 05 and 6f, but it introduces a big disparity int ests  where one is twice as long

emord:
Changes are only in tests. self merging as all changes were reviewed at some point before my rebase

PR #15942, opened 2017-04-28
fix async restore

dannyroberts:
@benrudolph agrees this looks like just a typo—
self.payload is a file and we wouldn't expect it to respond to
`.get_content_length()`, whereas self has a specific method defined for
that (`get_content_length`)

### DIFF
--- a/corehq/ex-submodules/casexml/apps/phone/restore.py
+++ b/corehq/ex-submodules/casexml/apps/phone/restore.py
@@ -388,13 +388,7 @@ class CachedResponse(object):
         content_length = self.get_content_length()
         if content_length is not None:
             headers['Content-Length'] = content_length
-        return stream_response(self.payload.as_file(), headers)
-
-    def as_file(self):
-        if self.payload:
-            return self.payload.as_file()
-        else:
-            return None
+        return stream_response(self.payload, headers)
 
 
 class RestoreParams(object):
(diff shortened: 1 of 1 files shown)


dannyroberts:
@snopoke @kaapstorm cc: @orangejenny, @amstone326

benrudolph:
thank you danny!

orangejenny (approved):

corehq/ex-submodules/casexml/apps/phone/restore.py
```diff
@@ -385,7 +385,7 @@ def as_file(self):
 
     def get_http_response(self):
         headers = {}
-        content_length = self.payload.get_content_length()
+        content_length = self.get_content_length()
         if content_length is not None:
             headers['Content-Length'] = content_length
         return stream_response(self.payload.as_file(), headers)
```
kaapstorm: Shouldn't this be `self.as_file()` then?

kaapstorm (changes requested):
This makes sense. But should `self.payload.as_file()` in line 391 then be `self.as_file()` if `self.payload` is potentially a file?

corehq/ex-submodules/casexml/apps/phone/restore.py
dannyroberts: Yeah I think you're right. And `self.as_file()` is now just `self.payload` (Actually it's defined twice but I think the second time is a mistake.)

proteusvacuum:
I'm a bit confused. Does this mean that this code never worked and that particular code path was never hit? Or did the recent change to the restore change the way this was reacting?

PR #15945, opened 2017-04-28
Treat case properties as date objects

sravfeyn:
@proteusvacuum

### DIFF
--- a/custom/enikshay/tasks.py
+++ b/custom/enikshay/tasks.py
@@ -193,9 +193,9 @@ class EpisodeUpdate(object):
         if not raw_date:
             return None
         if parse_datetime(raw_date):
-            return parse_datetime(raw_date)
+            return (parse_datetime(raw_date)).replace(tzinfo=pytz.utc)
         elif parse_date(raw_date):
-            return datetime.datetime.combine(parse_date(raw_date), datetime.datetime.min.time())
+            return (datetime.datetime.combine(parse_date(raw_date), datetime.datetime.min.time())).replace(tzinfo=pytz.utc)
         else:
             raise EnikshayTaskException(
                 "Episode case {case_id} has invalid format for 'adherence_schedule_date_start' {date}".format(
--- a/custom/enikshay/tasks.py
+++ b/custom/enikshay/tasks.py
@@ -24,6 +24,7 @@ from .const import (
     DAILY_SCHEDULE_ID,
     SCHEDULE_ID_FIXTURE,
     HISTORICAL_CLOSURE_REASON,
+    ENIKSHAY_TIMEZONE,
 )
 from .exceptions import EnikshayTaskException
 from .data_store import AdherenceDatastore
@@ -52,7 +53,9 @@ class EpisodeAdherenceUpdater(object):
 
     def __init__(self, domain):
         self.domain = domain
-        self.purge_date = pytz.UTC.localize(datetime.datetime.today() - datetime.timedelta(days=60))
+        # set purge_date to 60 days back
+        self.purge_date = datetime.datetime.now(
+            pytz.timezone(ENIKSHAY_TIMEZONE)).date() - datetime.timedelta(days=60)
         self.adherence_data_store = AdherenceDatastore(domain)
 
     def run(self):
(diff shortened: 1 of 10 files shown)


custom/enikshay/tasks.py
```diff
@@ -193,9 +193,9 @@ def get_adherence_schedule_start_date(self):
         if not raw_date:
             return None
         if parse_datetime(raw_date):
-            return parse_datetime(raw_date)
+            return (parse_datetime(raw_date)).replace(tzinfo=pytz.utc)
         elif parse_date(raw_date):
-            return datetime.datetime.combine(parse_date(raw_date), datetime.datetime.min.time())
+            return (datetime.datetime.combine(parse_date(raw_date), datetime.datetime.min.time())).replace(tzinfo=pytz.utc)
```
dimagimon: E501 line too long (123 > 115 characters)

custom/enikshay/tasks.py
```diff
@@ -193,9 +193,9 @@ def get_adherence_schedule_start_date(self):
         if not raw_date:
             return None
         if parse_datetime(raw_date):
-            return parse_datetime(raw_date)
+            return (parse_datetime(raw_date)).replace(tzinfo=pytz.utc)
```
proteusvacuum: can you explain what outcome you are expecting here? I'm a bit confused by what is happening. I think `raw_date` is actually potentially in IST as it is coming from the phone? I don't think that is encoded in the date itself, so you might need to first tell python that the TZ of the date is IST before converting it to UTC. 

I believe the dates we've been using in this project are in IST, since that is what gets saved to the case. (for [example](https://github.com/dimagi/commcare-hq/blob/7e86b09be40e9bd642dd97a495a49d940aba2dcc/custom/enikshay/integrations/ninetyninedots/utils.py#L88-L97))

sravfeyn: Thanks for the catch!

The `episode.adherence_schedule_start_date` comes in a string format, I thought it would be in UTC, so I added UTC tzinfo so that date comparisons between timezone-aware and timezone-naive don't break. I am changing it to IST

sravfeyn: Actually both `episode.adherence_schedule_start_date` and `adherence_date` are just IST date objects (example [case](https://india.commcarehq.org/a/enikshay/reports/case_data/2822c551-01c8-47d4-b187-c87651e807e0/)). So I have changed the code to treat them as date objects

proteusvacuum:
build fail real:
```  File "/mnt/commcare-hq-ro/custom/enikshay/urls.py", line 7, in <module>

    url(r'^99dots/', include("custom.enikshay.integrations.ninetyninedots.urls")),

  File "/vendor/lib/python2.7/site-packages/django/conf/urls/__init__.py", line 50, in include

    urlconf_module = import_module(urlconf_module)

  File "/usr/local/lib/python2.7/importlib/__init__.py", line 37, in import_module

    __import__(name)

  File "/mnt/commcare-hq-ro/custom/enikshay/integrations/ninetyninedots/urls.py", line 2, in <module>

    from custom.enikshay.integrations.ninetyninedots.views import (

  File "/mnt/commcare-hq-ro/custom/enikshay/integrations/ninetyninedots/views.py", line 13, in <module>

    from custom.enikshay.integrations.ninetyninedots.utils import (

  File "/mnt/commcare-hq-ro/custom/enikshay/integrations/ninetyninedots/utils.py", line 6, in <module>

    from casexml.apps.case.const import CASE_INDEX_EXTENSION, UNOWNED_EXTENSION_OWNER_ID, ENIKSHAY_TIMEZONE

ImportError: cannot import name ENIKSHAY_TIMEZONE```

File: custom/enikshay/tests/test_adherence_updates.py [hunk 3]
```diff
@@ -144,15 +143,15 @@ def _create_episode_case(self, adherence_schedule_date_start, adherence_schedule
     def _create_adherence_cases(self, adherence_cases):
         self.factory.create_or_update_cases([
             get_adherence_case_structure(
-                adherence_date.strftime('%Y-%m-%d-%H-%M'),
+                "adherence{}".format(i),
```
proteusvacuum: why do you do this? this will name the adherence cases `adherence0`, `adherence1` etc, right? 
Wouldn't that lead to duplicate names each time this task is run?

proteusvacuum (commented):
one question, otherwise looks good

custom/enikshay/tests/test_adherence_updates.py
sravfeyn: Because I want to test adherences that have same adherence_date and there is no time info, so they will get same case_id with previous name format, which means only one case will get created, so I name them to have unique names. The cases are cleaned up after every individual test case, so there is no collision.

sravfeyn: These are just test cases, the task will have real case_ids.

proteusvacuum: oh gotcha.

proteusvacuum: oh gotcha.

proteusvacuum: gotcha

PR #15957, opened 2017-04-30
Always ensure that SQLLocation.location_id is not blank when saved

gcapalbo:
https://manage.dimagi.com/default.asp?253058

SQLLocation.location_id has a unique constraint on it, and within the sql <-> couch sync processes, it was being initially saved as empty string every time a SQLLocation was created, even though it was later being updated in the same transaction block with the actual id value.

This was causing an issue because when you have lots of uncommitted inserts being referenced in separate transactions that all conflict on the same (blank) location_id with a unique constraint, it basically causes deadlock. Postgres has to wait for the inserts to happen in order, and only lets the next block continue once the previous one has committed or rolled back.

We probably haven't noticed this yet because the deadlock must have normally cleared in a timely manner, but when this gets combined with a huge location import (which itself has all of its writes entirely wrapped in an outer transaction), it results in a long-lasting deadlock for anyone who's trying to create a location, which is what happened in this ticket.

So the main purpose of this PR is to just make sure that every time we save SQLLocation, that we already have the right location_id set so that we don't have inserts in different transaction blocks conflicting on the same value for location_id.

This PR also removes the use of the SyncCouchToSQL and SyncSQLToCouch mixins because it was getting very difficult to follow everything given that this sync process is so custom. By doing this, we also introduce some performance improvements - previously the sync process was saving each postgres and couch model twice (and syncing the supply point twice). Now each save only happens once, whether it's a create or an edit.

@esoergel @kaapstorm @snopoke 
buddy @NoahCarnahan

### DIFF
--- a/corehq/apps/locations/models.py
+++ b/corehq/apps/locations/models.py
@@ -9,7 +9,6 @@ from dimagi.ext.couchdbkit import *
 import itertools
 from corehq.apps.cachehq.mixins import CachedCouchDocumentMixin
 from dimagi.utils.couch.database import iter_docs
-from dimagi.utils.couch.migration import SyncSQLToCouchMixin, SyncCouchToSQLMixin
 from dimagi.utils.decorators.memoized import memoized
 from datetime import datetime
 from django.db import models, transaction
@@ -369,11 +368,10 @@ class OnlyUnarchivedLocationManager(LocationManager):
         return list(self.accessible_to_user(domain, user).location_ids())
 
 
-class SQLLocation(SyncSQLToCouchMixin, MPTTModel):
+class SQLLocation(MPTTModel):
     domain = models.CharField(max_length=255, db_index=True)
     name = models.CharField(max_length=255, null=True)
     location_id = models.CharField(max_length=100, db_index=True, unique=True)
-    _migration_couch_id_name = "location_id"  # Used for SyncSQLToCouchMixin
     location_type = models.ForeignKey(LocationType, on_delete=models.CASCADE)
     site_code = models.CharField(max_length=255)
     external_id = models.CharField(max_length=255, null=True, blank=True)
@@ -399,34 +397,54 @@ class SQLLocation(SyncSQLToCouchMixin, MPTTModel):
     active_objects = OnlyUnarchivedLocationManager()
 
     @classmethod
-    def _migration_get_fields(cls):
+    def get_sync_fields(cls):
         return ["domain", "name", "site_code", "external_id",
                 "metadata", "is_archived"]
 
-    @classmethod
-    def _migration_get_couch_model_class(cls):
-        return Location
+    def sync_couch_location(self):
+        if self.location_id:
+            couch_obj = self.couch_location
+        else:
+            couch_obj = Location()
+            couch_obj._id = uuid.uuid4().hex
+            self.location_id = couch_obj._id
 
-    def _migration_do_sync(self):
-        couch_obj = self._migration_get_or_create_couch_object()
         couch_obj._sql_location_type = self.location_type
         couch_obj.latitude = float(self.latitude) if self.latitude else None
         couch_obj.longitude = float(self.longitude) if self.longitude else None
-        self._migration_sync_to_couch(couch_obj)
+
+        for field_name in self.get_sync_fields():
+            value = getattr(self, field_name)
+            setattr(couch_obj, field_name, value)
+
+        return couch_obj
 
     @transaction.atomic()
     def save(self, *args, **kwargs):
         from corehq.apps.commtrack.models import sync_supply_point
         from .document_store import publish_location_saved
-        self.supply_point_id = sync_supply_point(self)
 
         sync_to_couch = kwargs.pop('sync_to_couch', True)
-        kwargs['sync_to_couch'] = False  # call it here
-        super(SQLLocation, self).save(*args, **kwargs)
+
         if sync_to_couch:
-            self._migration_do_sync()
+            couch_obj = self.sync_couch_location()
+
+        self.supply_point_id = sync_supply_point(self)
+
+        if not self.location_id:
+            raise ValueError("Expected non-empty location_id")
+
+        super(SQLLocation, self).save(*args, **kwargs)
+
+        if sync_to_couch:
+            couch_obj.save(sync_to_sql=False)
 
         publish_location_saved(self.domain, self.location_id)
+
+    def delete(self, *args, **kwargs):
+        if kwargs.pop('sync_to_couch', True):
+            self.couch_location.delete(sync_to_sql=False)
+        super(SQLLocation, self).delete(*args, **kwargs)
 
     def to_json(self):
         return {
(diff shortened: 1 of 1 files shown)


gcapalbo:
(Tests for the sync are [here](https://github.com/dimagi/commcare-hq/blob/6f463a9b92a767f96fe2797f96cb606548426751/corehq/apps/locations/tests/test_location_sync.py))

kaapstorm (approved):

kaapstorm:
Nicely done!

NoahCarnahan (commented):
Can the definitions of `SyncSQLToCouchMixin` and `SyncCouchToSQLMixin` also be removed? I see those class names are referenced in migrations, so perhaps that complicates things...

NoahCarnahan:
Wow this sounds like it was probably hard to track down! Thanks!

gcapalbo:
@NoahCarnahan We may want to keep the syncing mixins since they can help when migrating models from couch to postgres (I used them heavily in the messaging migrations). They're good for when the syncing logic is more simple.

snopoke:
@gcapalbo would be great to do a retro on how you debugged this.

corehq/apps/locations/models.py
```diff
@@ -399,35 +397,58 @@ class SQLLocation(SyncSQLToCouchMixin, MPTTModel):
...
-    @classmethod
-    def _migration_get_couch_model_class(cls):
-        return Location
+    def sync_couch_location(self):
+        if self.location_id:
+            couch_obj = self.couch_location
+        else:
+            couch_obj = Location()
+            couch_obj._id = uuid.uuid4().hex
+            self.location_id = couch_obj._id
```
esoergel: We have a lot of instances where we save something to the db to get an ID, even though it would be simpler to save later.  We should really get more comfortable setting the ID during creation when it makes sense to do so.

corehq/apps/locations/models.py
```diff
@@ -846,24 +867,30 @@ def set_location_type(self, location_type_name):
...
+    def get_sync_fields(cls):
         return ["domain", "name", "site_code", "external_id", "metadata",
                 "is_archived", "latitude", "longitude"]
 
-    @classmethod
-    def _migration_get_sql_model_class(cls):
-        return SQLLocation
+    def sync_sql_location(self):
+        if not self._id:
+            self._id = uuid.uuid4().hex
```
esoergel: This seems like an odd place to set the UUID, what's the reason for putting it here rather than the save method?

esoergel (commented):
I'd definitely be interested in hearing more about how you debugged this.

The location models haven't been on my radar for a little while, but this is a good reminder that we really need to get rid of couch locations.  I've got a little more breathing room now, and I'll push through some steps this week.

corehq/apps/locations/models.py
gcapalbo: I guess it doesn't matter too much, but if it was in the save method then you couldn't call sync_sql_location independently, so my approach here was to be explicit about putting it in the context where it was needed. But yeah since we never use sync_sql_location independently I guess it could have also gone in save, thought it wouldn't change the current behavior functionally. I tried to avoid setting it in __init__ because once you have an ._id it kind of implies you have a SQLLocation too so I didn't know if there was any other code that would break if I did that.

corehq/apps/locations/models.py
gcapalbo: :+1: Agreed - to be clear, this is setting a random uuid instead of hitting the db.  There's also some info here related to performance based on how ids are generated: http://docs.couchdb.org/en/1.6.1/maintenance/performance.html?highlight=id#document-s-id

Right now we just do random ids which doesn't seem like the best approach, but generating our own sequential ids might increase the chances of conflicts unless we find a good way to do it.

gcapalbo:
@snopoke @esoergel Debugging this was a bit difficult. The symptom was that creating a location on prod would just hang, but it was not affecting editing locations.

It felt like some kind of deadlock was happening so my first thought was that something was happening with the redis locking that's used to prevent simultaneous updates to a domain's location structure. So I first looked at the lock key but confirmed it was properly taking the domain into account so it wasn't a global lock.

But I also realized that those locks are non-blocking so it couldn't be the redis locking that was causing the issue. However I did notice an issue there where locks weren't being set with a timeout, which was causing a separate issue with them timing out properly: https://github.com/dimagi/commcare-hq/pull/15955

After realizing it wasn't an issue with the redis locking, I then tried replicating the issue locally without success. I spent a lot of time trying to step through the process locally manually in various different ways and examining the code to see where it might be getting stuck. Throughout this process, I noticed a couple of things that seemed weird: 1) each time I created a location, the save() method was being called twice on SQLLocation and Location models, and 2) two supply point cases were being updated each time I created a location: a single supply point case with a blank location id that was getting updated on every location create and a supply point case with the right location id that was being created.

I then realized that I could try to examine on prod the state of the supply point cases to see how far the process was getting before it got stuck.  So I tried to create a location on prod and noticed that it was updating the supply point case with the blank location id, but never creating the supply point case with the right location id.

This meant it was getting past [this point](https://github.com/dimagi/commcare-hq/blob/2f8bcbb1470ca9bc8046f434a92ba43d02124d55/corehq/apps/locations/models.py#L421). Also, since the syncing of the couch location would call SQLLocation.save() again as I had noticed earlier, and since the supply point case with the right location id wasn't being created I knew it wasn't getting past [here](https://github.com/dimagi/commcare-hq/blob/2f8bcbb1470ca9bc8046f434a92ba43d02124d55/corehq/apps/locations/models.py#L427).

So this meant the deadlock was happening within [this block](https://github.com/dimagi/commcare-hq/blob/2f8bcbb1470ca9bc8046f434a92ba43d02124d55/corehq/apps/locations/models.py#L423-L427).

At this point I figured it must be deadlocking within postgres since that's the only service it was hitting in that block that I could think of where more locking might be happening. I tried to think of ways in which postgres might use locking and realized it might be happening on inserting records for conflicting values on fields with unique constraints. So I started looking at the unique constraints on SQLLocation and once I noticed that location_id had a unique constraint I realized that the SQLLocation must be saved with a blank location_id at first on each create, which would explain the single supply point case with no location id.

I then checked the code locally and confirmed SQLLocation was indeed being saved with a blank location id on the first save on each location create, and to confirm that this would cause a locking issue, I ran this locally in two different psql shells:

the first psql shell:

    commcarehq_sharded=# create table test(id integer unique);
    CREATE TABLE
    commcarehq_sharded=# start transaction;
    START TRANSACTION
    commcarehq_sharded=# insert into test values (1);
    INSERT 0 1
    commcarehq_sharded=# update test set id = 2 where id = 1;
    UPDATE 1
    commcarehq_sharded=# 

the second psql shell:

    commcarehq_sharded=# start transaction;
    START TRANSACTION
    commcarehq_sharded=# insert into test values (1);

and the second shell would not complete the insert - it just kept waiting, until i issued a commit or rollback on the first shell.

So then I realized all I had to do was just make sure location_id had a unique value initially and then threads wouldn't interfere with each other on location create.

One weird side effect of this issue is that for domains that use supply point cases, there is a single supply point case in the domain (that never gets used anywhere) with a blank location id that has a ton of updates, since it was getting its name, etc. updated on every location create.

snopoke:
Thanks for the writeup!

PR #15958, opened 2017-04-30
[WIP] Mk enikshay private api

mkangia:
[Spec](https://docs.google.com/document/d/1kHEzXLq-8RvPRcRXbYaSNgFKeD3OFP7DR31VytNxxl8/edit#heading=h.svb16cn9xwxo)

### DIFF
--- a/corehq/apps/repeaters/models.py
+++ b/corehq/apps/repeaters/models.py
@@ -45,16 +45,30 @@ from .const import (
 from .exceptions import RequestConnectionError
 from .utils import get_all_repeater_types
 
+def log_timeout_and_raise(domain, error):
+    datadog_counter('commcare.repeaters.timeout', tags=[
+            u'domain:{}'.format(domain),
+        ])
+        raise RequestConnectionError(error)
 
 def simple_post_with_logged_timeout(domain, data, url, *args, **kwargs):
     try:
         response = simple_post(data, url, *args, **kwargs)
     except (Timeout, ConnectionError) as error:
-        datadog_counter('commcare.repeaters.timeout', tags=[
-            u'domain:{}'.format(domain),
-        ])
-        raise RequestConnectionError(error)
+        log_timeout_and_raise(domain, error)
     return response
+
+def simple_xml_post_with_logged_timeout(domain, data, url, operation, *args, **kwargs):
+    try:
+        client = Client(url)
+        # for response object use
+        # with client.options(raw_response=True):
+        #     response = client.service[self.repeater.operation](**post_info.payload)
+        response = client.service[operation](**data)
+    except (Timeout, ConnectionError) as error:
+        log_timeout_and_raise(domain, error)
+    return response
+
 
 DELETED = "-Deleted"
 
@@ -538,9 +552,10 @@ class RepeatRecord(Document):
         try:
             generator = self.get_payload_generator(self.format_or_default_format())
             if generator.format_label == 'XML' and hasattr(self.repeater, 'operation'):
-                client = Client(self.url)
-                response = client.service[self.repeater.operation](**post_info.payload)
-                return self.handle_success(response)
+                response = simple_xml_post_with_logged_timeout(
+                    self.domain, post_info.payload, self.url, self.repeater.operation
+                )
+                return self.handle_success(response)  # soap operations always throw exceptions when they fail
             else:
                 response = simple_post_with_logged_timeout(
                     self.domain,
--- a/corehq/apps/repeaters/models.py
+++ b/corehq/apps/repeaters/models.py
@@ -1,5 +1,6 @@
 from collections import namedtuple
 from datetime import datetime, timedelta
+from zeep import Client
 import logging
 import urllib
 import urlparse
@@ -535,13 +536,19 @@ class RepeatRecord(Document):
     def post(self, post_info, tries=0):
         tries += 1
         try:
-            response = simple_post_with_logged_timeout(
-                self.domain,
-                post_info.payload,
-                self.url,
-                headers=post_info.headers,
-                timeout=POST_TIMEOUT,
-            )
+            generator = self.get_payload_generator(self.format_or_default_format())
+            if generator.format_label == 'XML' and self.repeater.operation:
+                client = Client(self.url)
+                response = client.service[self.repeater.operation](**post_info.payload)
+                return self.handle_success(response)
+            else:
+                response = simple_post_with_logged_timeout(
+                    self.domain,
+                    post_info.payload,
+                    self.url,
+                    headers=post_info.headers,
+                    timeout=POST_TIMEOUT,
+                )
         except Exception as e:
             self.handle_exception(e)
         else:
--- a/corehq/apps/repeaters/models.py
+++ b/corehq/apps/repeaters/models.py
@@ -552,8 +552,7 @@ class RepeatRecord(Document):
     def post(self, post_info, tries=0):
         tries += 1
         try:
-            generator = self.get_payload_generator(self.format_or_default_format())
-            if generator.format_label == 'XML' and hasattr(self.repeater, 'operation'):
+            if 'WSDL' in self.repeater.url.upper() and self.repeater.operation.strip():
                 response = simple_xml_post_with_logged_timeout(
                     self.domain, post_info.payload, self.url, self.repeater.operation
                 )
--- a/corehq/apps/repeaters/models.py
+++ b/corehq/apps/repeaters/models.py
@@ -48,8 +48,8 @@ from .utils import get_all_repeater_types
 
 def log_timeout_and_raise(domain, error):
     datadog_counter('commcare.repeaters.timeout', tags=[
-            u'domain:{}'.format(domain),
-        ])
+        u'domain:{}'.format(domain),
+    ])
     raise RequestConnectionError(error)
 
 
--- a/corehq/apps/repeaters/models.py
+++ b/corehq/apps/repeaters/models.py
@@ -63,10 +63,11 @@ def simple_post_with_logged_timeout(domain, data, url, *args, **kwargs):
 def simple_xml_post_with_logged_timeout(domain, data, url, operation, *args, **kwargs):
     try:
         client = Client(url)
-        # for response object use
-        # with client.options(raw_response=True):
-        #     response = client.service[self.repeater.operation](**post_info.payload)
-        return client.service[operation](**data)
+        # for just response message use
+        # with client.options(raw_response=False) # explicitly set it to avoid caching mysteries
+        #     return client.service[operation](**data)
+        with client.options(raw_response=True):
+            return client.service[operation](**data)
     except (Timeout, ConnectionError) as error:
         log_timeout_and_raise(domain, error)
 
@@ -540,6 +541,12 @@ class RepeatRecord(Document):
         if reraise:
             raise
 
+    def get_xml_message_response(self, response):
+        # parse the xml content on response object to return just the message
+        client = Client(self.url)
+        binding = client.service[self.repeater.operation]._proxy._binding
+        return binding.process_reply(client, binding.get(self.repeater.operation), response)
+
     def fire(self, force_send=False):
         headers = self.repeater.get_headers(self)
         if self.try_now() or force_send:
@@ -556,7 +563,6 @@ class RepeatRecord(Document):
                 response = simple_xml_post_with_logged_timeout(
                     self.domain, post_info.payload, self.url, self.repeater.operation
                 )
-                return self.handle_success(response)  # soap operations always throw exceptions when they fail
             else:
                 response = simple_post_with_logged_timeout(
                     self.domain,
--- a/corehq/apps/repeaters/models.py
+++ b/corehq/apps/repeaters/models.py
@@ -150,6 +150,7 @@ class Repeater(QuickCachedDocumentMixin, Document, UnicodeMixIn):
     use_basic_auth = BooleanProperty(default=False)
     username = StringProperty()
     password = StringProperty()
+    operation = StringProperty()
     friendly_name = _("Data")
 
     @classmethod
--- a/corehq/apps/repeaters/models.py
+++ b/corehq/apps/repeaters/models.py
@@ -537,7 +537,7 @@ class RepeatRecord(Document):
         tries += 1
         try:
             generator = self.get_payload_generator(self.format_or_default_format())
-            if generator.format_label == 'XML' and self.repeater.operation:
+            if generator.format_label == 'XML' and hasattr(self.repeater, 'operation'):
                 client = Client(self.url)
                 response = client.service[self.repeater.operation](**post_info.payload)
                 return self.handle_success(response)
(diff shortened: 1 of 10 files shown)


custom/enikshay/integrations/nikshay/repeater_generator.py
```diff
@@ -337,6 +344,92 @@ def handle_exception(self, exception, repeat_record):
...
+        person_case_properties = person_case.dynamic_case_properties()
+
+        person_locations = get_person_locations(person_case)
+        tu_choice = person_case_properties.get('tu_choice')
+        try:
+            tu_location = SQLLocation.objects.get(location_id=tu_choice)
+        except SQLLocation.DoesNotExist:
+            raise NikshayLocationNotFound(
+                "Location with id {location_id} not found. This is the tu_choide for person with id: {person_id}"
+                    .format(location_id=tu_choice, person_id=person_case.case_id)
```
dimagimon: E131 continuation line unaligned for hanging indent

mkangia:
@proteusvacuum Its still a work in progress. Waiting on permissions to actually hit the API and develop this further for correct request and update code to parse responses.

custom/enikshay/integrations/nikshay/repeater_generator.py
```diff
@@ -337,6 +344,92 @@ def handle_exception(self, exception, repeat_record):
                                 "followup_nikshay_registered", "followup_nikshay_error")
 
 
+@RegisterGenerator(NikshayRegisterPrivatePatientRepeater, 'case_xml', 'XML', is_default=True)
```
proteusvacuum: I think this is very misleading, since this doesn't actually return XML. Is there a reason you wouldn't make a new `SOAP` type like I had in my WIP?

custom/enikshay/integrations/nikshay/repeater_generator.py
```diff
@@ -337,6 +344,92 @@ def handle_exception(self, exception, repeat_record):
...
+        episode_case_properties = episode_case.dynamic_case_properties()
+        person_case_properties = person_case.dynamic_case_properties()
+
+        person_locations = get_person_locations(person_case)
+        tu_choice = person_case_properties.get('tu_choice')
+        try:
+            tu_location = SQLLocation.objects.get(location_id=tu_choice)
+        except SQLLocation.DoesNotExist:
+            raise NikshayLocationNotFound(
+                "Location with id {location_id} not found. This is the tu_choide for person with id: {person_id}"
```
proteusvacuum: spelling `tu_choice`

corehq/apps/repeaters/models.py
```diff
@@ -535,13 +536,19 @@ def fire(self, force_send=False):
...
         try:
-            response = simple_post_with_logged_timeout(
-                self.domain,
-                post_info.payload,
-                self.url,
-                headers=post_info.headers,
-                timeout=POST_TIMEOUT,
-            )
+            generator = self.get_payload_generator(self.format_or_default_format())
+            if generator.format_label == 'XML' and hasattr(self.repeater, 'operation'):
```
proteusvacuum: Is there a reason you didn't want to explicitly set a `repeater.method`? This seems to be a bit hacky.

corehq/apps/repeaters/models.py
```diff
@@ -535,13 +536,19 @@ def fire(self, force_send=False):
     def post(self, post_info, tries=0):
```
proteusvacuum: I'd added a few comments in my original commit that might be useful here. Also, it looks like you are only adding the `operation` property to the NikshayRepeater rather than the parent class. What's your rationale behind that? I think we might want to be able to use that later for other repeaters as it is generic behaviour.

custom/enikshay/integrations/nikshay/repeater_generator.py
```diff
@@ -337,6 +344,92 @@ def handle_exception(self, exception, repeat_record):
...
+                .format(location_id=tu_choice, person_id=person_case.case_id)
+            )
+        try:
+            tu_code = tu_location.metadata['nikshay_code']
+        except (KeyError, AttributeError) as e:
+            raise NikshayCodeNotFound("Nikshay codes not found: {}".format(e))
```
proteusvacuum: can you add the ID for the `tu_location` in the error message.

custom/enikshay/integrations/nikshay/repeater_generator.py
```diff
@@ -337,6 +344,92 @@ def handle_exception(self, exception, repeat_record):
...
+            "usersid": settings.ENIKSHAY_PRIVATE_API_USERS.get(person_locations.sto, ''),
+            "password": settings.ENIKSHAY_PRIVATE_API_PASSWORD,
+            "source": ENIKSHAY_ID,
+        }
+
+    def handle_success(self, response, payload_doc, repeat_record):
+        # A success would be getting a nikshay_id for the patient
+        # without it this would actually be a failure
+        try:
+            nikshay_id = _get_nikshay_id_from_response(response)
```
proteusvacuum: I think the SOAP API just returns the Nikshay ID directly

custom/enikshay/integrations/nikshay/repeater_generator.py
```diff
@@ -337,6 +344,92 @@ def handle_exception(self, exception, repeat_record):
...
+                    "nikshay_registered": "true",
+                    "nikshay_id": nikshay_id,
+                    "nikshay_error": "",
+                },
+                external_id=nikshay_id,
+            )
+        except NikshayResponseException as e:
+            _save_error_message(payload_doc.domain, payload_doc.case_id, unicode(e.message))
+
+    def handle_failure(self, response, payload_doc, repeat_record):
```
proteusvacuum: This isn't going to work. You explicitly call `handle_success` in the `fire` method, so it will never hit this.

custom/enikshay/integrations/nikshay/repeaters.py
```diff
@@ -186,6 +176,41 @@ def allowed_to_forward(self, test_case):
             return False
 
 
+class NikshayRegisterPrivatePatientRepeater(BaseNikshayRepeater):
+    operation = 'InsertHFIDPatient_UATBC'
+    white_listed_case_types = ['episode']
```
proteusvacuum: I don't think this will work because when you save the view, it will overwrite this property.

requirements/requirements.txt
```diff
@@ -98,4 +98,5 @@ email_validator==1.0.1
 dnspython==1.13.0
 raven==6.0.0
 quickcache==0.0.1
+git+https://github.com/mvantellingen/python-zeep.git@46653fac2a92a99bc93d20c772cd7d797c301105
```
proteusvacuum: is there a reason we aren't using a specific version of this library?

custom/enikshay/integrations/nikshay/repeater_generator.py
```diff
@@ -337,6 +344,92 @@ def handle_exception(self, exception, repeat_record):
...
+            "lno": person_case_properties.get('phone', ''),
+            "mno": '0',
+            "tbdiagdate": str(episode_date),
+            "tbstdate": episode_case_properties.get(TREATMENT_START_DATE, str(datetime.date.today())),
+            "Type": disease_classification.get(episode_case_properties.get('pulmonary_extra_pulmonary', ''), ''),
+            "B_diagnosis": episode_case_properties.get('case_definition', ''),
+            "D_SUSTest": episode_case_properties.get('drug_susceptibility_test_status', ''),
+            "Treat_I": episode_case_properties.get('treatment_initiation_status', ''),
+            "usersid": settings.ENIKSHAY_PRIVATE_API_USERS.get(person_locations.sto, ''),
+            "password": settings.ENIKSHAY_PRIVATE_API_PASSWORD,
```
proteusvacuum: I think the password is different for each user?

corehq/apps/repeaters/models.py
```diff
@@ -535,13 +536,19 @@ def fire(self, force_send=False):
...
-                post_info.payload,
-                self.url,
-                headers=post_info.headers,
-                timeout=POST_TIMEOUT,
-            )
+            generator = self.get_payload_generator(self.format_or_default_format())
+            if generator.format_label == 'XML' and hasattr(self.repeater, 'operation'):
+                client = Client(self.url)
+                response = client.service[self.repeater.operation](**post_info.payload)
+                return self.handle_success(response)
```
proteusvacuum: Can you explicitly move the `self.handle_response` for the POST out of the `else` clause so that the code flow is easier to follow? Also, I'd added a comment alongside this line so that it was more obvious why we were calling `handle_success` instead of letting this class handle the response.

custom/enikshay/integrations/nikshay/repeater_generator.py
mkangia: Right! i used xml because this eventually sends out an xml request.
It looks like a better way out instead of adding another variable unless it conflicts with the implementation in Base payload generator.

custom/enikshay/integrations/nikshay/repeater_generator.py
mkangia: Its the same as seen from details in email "Phase 2 Nikshay Integration"

custom/enikshay/integrations/nikshay/repeater_generator.py
mkangia: A success response would look something like
```
<?xml version="1.0" encoding="utf-8"?>
<soap:Envelope xmlns:soap="http://www.w3.org/2003/05/soap-envelope" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns:xsd="http://www.w3.org/2001/XMLSchema">
    <soap:Body>
        <InsertHFIDPatient_UATBCResponse xmlns="http://tempuri.org/">
            <InsertHFIDPatient_UATBCResult>00003</InsertHFIDPatient_UATBCResult>
        </InsertHFIDPatient_UATBCResponse>
    </soap:Body>
</soap:Envelope>
```
I just added the code as placeholder. Will now modify it according to the response.

corehq/apps/repeaters/models.py
mkangia: From what i understand, currently `handle_response` would get called only when the request is sent successfully. Moving it out of else would invoke it in case of failures/exceptions as well.

custom/enikshay/integrations/nikshay/repeater_generator.py
mkangia: Improving the failure case.

requirements/requirements.txt
mkangia: This is the same version that successfully worked when i used zeep for the integration i did earlier. So just used the trusted version. Checking with latest version now.

corehq/apps/repeaters/models.py
```diff
@@ -45,15 +46,30 @@
 from .utils import get_all_repeater_types
 
 
+def log_timeout_and_raise(domain, error):
+    datadog_counter('commcare.repeaters.timeout', tags=[
+            u'domain:{}'.format(domain),
```
dimagimon: E126 continuation line over-indented for hanging indent

corehq/apps/repeaters/models.py
```diff
@@ -45,15 +46,30 @@
...
+        ])
```
dimagimon: E121 continuation line under-indented for hanging indent

custom/enikshay/integrations/nikshay/repeater_generator.py
```diff
@@ -337,6 +344,91 @@ def handle_exception(self, exception, repeat_record):
...
+                )
+            else:
+                self.handle_failure(response, payload_doc, repeat_record)
+        except NikshayResponseException as e:
+            _save_error_message(payload_doc.domain, payload_doc.case_id, unicode(e.message))
+
+    def handle_failure(self, response, payload_doc, repeat_record):
+        _save_error_message(payload_doc.domain, payload_doc.case_id, unicode(response),
+                            "private_nikshay_registered", "private_nikshay_error"
+        )
```
dimagimon: E124 closing bracket does not match visual indentation

custom/enikshay/integrations/nikshay/repeater_generator.py
```diff
@@ -337,6 +344,91 @@ def handle_exception(self, exception, repeat_record):
...
+
+    def handle_exception(self, exception, repeat_record):
+        if isinstance(exception, RequestConnectionError):
+            update_case(repeat_record.domain, repeat_record.payload_id, {"private_nikshay_error": unicode(exception)})
```
dimagimon: E501 line too long (118 > 115 characters)

custom/enikshay/integrations/nikshay/repeater_generator.py
dimagimon: E124 closing bracket does not match visual indentation

custom/enikshay/integrations/nikshay/repeater_generator.py
dimagimon: E501 line too long (118 > 115 characters)

corehq/apps/repeaters/models.py
mkangia: Cool.. Adding it on repeater

custom/enikshay/integrations/nikshay/repeater_generator.py
mkangia: I have updated the code to not rely on the format but on the url mentioned for the repeater, checking if its a WSDL link or not and presence of an operator.

So can just update this back to case_json and JSON but i feel its still confusing to keep both JSON/CASE JSON and SOAP together. Still feel keeping it XML makes more sense. let me know if its still confusing.

requirements/requirements.txt
mkangia: latest version works as well.

custom/enikshay/integrations/nikshay/repeaters.py
mkangia: removed

corehq/apps/repeaters/models.py
mkangia: updated to rely on url and operation now.

proteusvacuum:
@dannyroberts FYI, this is the PR where Manish is adding SOAP functionality. Figured you might want to weigh in on how this happens with your recent refactor.

@mkangia sorry, you might need to rejig things a bit to follow Danny's new pattern. It should make things easier in the long run though, as you can override the `fire` method in the repeater now, which means you don't have to do the awkward if statement here: https://github.com/dimagi/commcare-hq/pull/15958/commits/fc197fc29a679c01dc5d7a0ca604b9a024215f67

Also, would be great to see some tests for this.

mkangia:
Thanks for flagging the changes @proteusvacuum.

i thought we wanted to make this generic so we would still need to modify the `fire` now `fire_for_record` method in the `Repeater` itself?

mkangia:
Was better to start a [new PR](https://github.com/dimagi/commcare-hq/pull/16149) instead of working on the conflicts. A lot had changed.
Sorry.

PR #15960, opened 2017-05-01
log celery task exceptions via django logger on NIC

sravfeyn:
@gcapalbo

### DIFF
--- a/corehq/util/signals.py
+++ b/corehq/util/signals.py
@@ -1,0 +1,10 @@
+import logging
+from celery.signals import task_failure
+from django.conf import settings
+
+
+@task_failure.connect
+def log_celery_task_exception(task_id, exception, traceback, einfo, *args, **kwargs):
+    if settings.environment == 'icds':
+        django_logger = logging.getLogger('django.request')
+        django_logger.error(u"Exception - {}".format(exception.message))
--- a/corehq/util/signals.py
+++ b/corehq/util/signals.py
@@ -1,10 +1,9 @@
-import logging
+from dimagi.utils.logging import notify_exception
 from celery.signals import task_failure
 from django.conf import settings
 
 
 @task_failure.connect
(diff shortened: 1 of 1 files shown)


sravfeyn:
By default it sends task-exception emails. We use our internal loggers when we catch exceptions inside a task, but when celery task itself fails, celery uses its own logging wrapper. By default it's configured to send emails. 

As part of disabling emails on NIC, I am going to [disable](https://github.com/dimagi/commcarehq-ansible/pull/723) `CELERY_SEND_TASK_ERROR_EMAILS` on NIC and use this PR to hijack logging from  celery wrapper onto django logger. Since django logger on NIC uses sentry, these exceptions will come into sentry now.

I don't see any reason not to do this on all envs

corehq/util/signals.py
```diff
@@ -0,0 +1,10 @@
+import logging
+from celery.signals import task_failure
+from django.conf import settings
+
+
+@task_failure.connect
+def log_celery_task_exception(task_id, exception, traceback, einfo, *args, **kwargs):
+    if settings.environment == 'icds':
```
snopoke: should be `settings.SERVER_ENVIRONMENT`

corehq/util/signals.py
```diff
@@ -0,0 +1,10 @@
...
+        django_logger = logging.getLogger('django.request')
+        django_logger.error(u"Exception - {}".format(exception.message))
```
snopoke: Can do this which will give the correct traceback etc:
```
notify_exception('Celery task failure', exec_info=einfo.exc_info)
```

snopoke (dismissed):

sravfeyn:
Updated

gcapalbo (approved):
looks good to me. just confirming that notify exception emails go to sentry?

sravfeyn:
Yeah, they go sentry

PR #15966, opened 2017-05-01
Remove old `FLAT_LOCATION_FIXTURE` toggle and migration

esoergel:
@mkangia  This feature flag isn't used any more so we can just drop it, right?


mkangia (approved):
From the code seems like its irrelevant now.

Would be nice to get rid of the tongue twister `HIERARCHICAL_LOCATION_FIXTURE` as well and just have single fixture format. Some day!

PR #15969, opened 2017-05-01
App manager v2: Fix v1/v2 include mixup

orangejenny:
@dannyroberts / @proteusvacuum 

Used "question" in the text because the column is titled "question". Open to further discussion on what the best copy is.

### DIFF
--- a/corehq/apps/app_manager/templates/app_manager/v2/partials/case_config_ko_templates.html
+++ b/corehq/apps/app_manager/templates/app_manager/v2/partials/case_config_ko_templates.html
@@ -130,7 +130,7 @@
 <script type="text/html" id="case-config:case-transaction:case-properties">
     <div class="panel-heading">
       <h4 class="panel-title panel-title-nolink">
-        <i class="fa fa-save"></i> {% trans "Save Question Answers to Case Properties" %}
+        <i class="fa fa-save"></i> {% trans "Save Questions to Case Properties" %}
       </h4>
     </div>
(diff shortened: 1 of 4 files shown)


corehq/apps/app_manager/templates/app_manager/v2/partials/case_config_ko_templates.html
```diff
@@ -130,7 +130,7 @@ <h4 class="panel-title panel-title-nolink">
 <script type="text/html" id="case-config:case-transaction:case-properties">
     <div class="panel-heading">
       <h4 class="panel-title panel-title-nolink">
-        <i class="fa fa-save"></i> {% trans "Save Question Answers to Case Properties" %}
```
proteusvacuum: Can you also change the column in the table below?

orangejenny: Oh, nevermind, that's not what you're saying at all.

orangejenny: Yeah, I'll update the column name.

orangejenny: https://github.com/dimagi/commcare-hq/pull/15969/commits/1e510e57811c0bb34ed78506560a792bf9ab886a

proteusvacuum (approved):
Thanks!

PR #15970, opened 2017-05-01
Remove redundant "delete_all_locations" calls

esoergel:
@mkangia
`domain.delete()` is usually being called anyways, and it includes
"delete_all_locations".  These calls actually have a noticeable cost to them too, so hopefully we'll see some minor test speed-up from this.


esoergel:
@biyeun new code buddy ping

PR #15975, opened 2017-05-02
cache case_search_enabled_for_domain

snopoke:
This get's called for every change that goes to the case search pillow.

buddy @nickpell


proteusvacuum:
Youch! Thanks @snopoke

PR #15980, opened 2017-05-02
Make focused select 2 have blue box shadow

biyeun:
@orangejenny 
screen shot 2017-05-02 at 11 36 29 am

re: https://manage.dimagi.com/default.asp?249761#1325860


orangejenny:
Super-nitpicky, but it would be nice to have these grouped with the other `select2` styles in this file.

PR #15988, opened 2017-05-02
Revert "Show HTML5 validation alert for select2 in report builder config"

NoahCarnahan:
Reverts dimagi/commcare-hq#15867

This PR prevented valid report builder forms from being submitted. This doesn't solve the problem of providing front end validation to this field, but until we can figure out how to do front end validation correctly, I think the previous state was better.

buddy @gcapalbo


NoahCarnahan:
@calellowitz There are 3 or 4 open interrupt tickets about this issue, so would be nice to get this in before today's deploy, although seems like tests might not run before then. I'd be in favor of merging this before tests pass because I don't believe this has any test coverage.

gcapalbo (approved):

PR #15991, opened 2017-05-02
Fix repeat record report's total docs count

NoahCarnahan:
Replicates the view key logic in `get_paged_repeat_records()`, the function that is actually returning the individual records that we are counting.
Fixes http://manage.dimagi.com/default.asp?252801

@benrudolph @proteusvacuum 
buddy @sravfeyn

### DIFF
--- a/corehq/apps/repeaters/tests/test_dbaccessors.py
+++ b/corehq/apps/repeaters/tests/test_dbaccessors.py
@@ -94,7 +94,7 @@ class TestRepeatRecordDBAccessors(TestCase):
 
     def test_get_repeat_record_count_with_state_and_no_repeater(self):
         count = get_repeat_record_count(self.domain, state=RECORD_PENDING_STATE)
-        self.assertEqual(count, 6)
+        self.assertEqual(count, 3)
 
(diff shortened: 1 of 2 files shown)


proteusvacuum (approved):

corehq/apps/repeaters/tests/test_dbaccessors.py
```diff
@@ -94,7 +94,7 @@ def test_get_failure_repeat_record_count(self):
 
     def test_get_repeat_record_count_with_state_and_no_repeater(self):
         count = get_repeat_record_count(self.domain, state=RECORD_PENDING_STATE)
-        self.assertEqual(count, 6)
+        self.assertEqual(count, 3)
```
NoahCarnahan: Ok this should actually be 3, as that matches `get_paged_repeat_records()` with the same arguments:
https://github.com/dimagi/commcare-hq/pull/15991/commits/9a90ee2d715ffaec7314dca7492cb06c0d966ea3#diff-a556cb8edac1a5f46167e86c50f0eeafR117

PR #15993, opened 2017-05-02
Tell mobile which fixture fields to index

kaapstorm:
Allows users to mark fixture fields as to be indexed on CommCare (mobile), using a column in Excel with a title like "field N: is_indexed?" (where N is the number of the field to be indexed) (similar to a field properties column).

These fixtures will have an attribute 'indexed="true"', and a schema giving CommCare the names of the columns to index in the sync payload.

This does not affect suite.xml or other fixtures that are not determined by the user.

[Spec](https://docs.google.com/document/d/1l7ffsMwduGi9CNfs7Sc3GW78RslSgMwmRouNwzkd9Tw/edit#). This feature is supported from [CommCare 2.33+](https://github.com/dimagi/commcare-core/wiki/fixtures#fixture-schemas).

@dannyroberts cc @snopoke @ctsims buddy @mkangia


dannyroberts (commented):
@kaapstorm awesome feature!

~I'm wondering whether you want this new XML to show up for all apps as they get built—which include changing apps' XML out from under existing projects when they make their next build—or you want this to only be included in CommCare 2.33+ apps. I'd think you'd want the latter, but I don't see anything in this PR restricting it to those apps.~

~Try adding an "enable_fixture_indexed_fields" method (or your better name) in `corehq/apps/app_manager/feature_support.py`, and then checking that flag in this code.~

ctsims:
I'm a bit confused as to why there would be any changes to people's app
XML, shouldn't this code should be ~entirely maintained in the User
Restore, or are there apps that use fixtureXML internally? (I know the
scheduler does, but it likely wouldn't need this treatment since there's no
clear way to define the indexing)

-Clayton

corehq/apps/fixtures/upload/workbook.py
```diff
@@ -113,7 +120,8 @@ def is_number(text):
         fields = [
             FixtureTypeField(
                 field_name=field,
-                properties=_get_field_properties('field {count}'.format(count=i + 1))
+                properties=_get_field_properties('field {count}'.format(count=i + 1)),
+                is_indexed=_get_field_is_indexed('field {count}'.format(count=i + 1)),
```
snopoke: could extract the `prop_key` to a var

snopoke (commented):
Looks good but could you add a test for the new upload column.

snopoke:
@dannyroberts @ctsims there are no changes to app build here, only change to the item lists fixture generator. 

To the question of whether all restores should get this or only ones from 2.33+, the CC version isn't available during restore so there's no way to make that decision. @ctsims I assume this won't break older versions, correct?

ctsims:
Thanks for doing the final pass on this PR to get it pushed out, Simon!
This will be a huge win for some of our programs for sure.

dannyroberts:
I just added strike-though to my comments above—the things I was saying didn't make sense because I'd misinterpreted this as an app builder change, which was just a mistake on my part. Sorry for the confusion!

PR #15995, opened 2017-05-03
metrics updates

snopoke:
* record request counts for auth_keys endpoint
* record form counts for all form submission requests
* a bit of refactoring

review by commit :fish:

### DIFF
--- a/corehq/apps/mobile_auth/views.py
+++ b/corehq/apps/mobile_auth/views.py
@@ -1,4 +1,6 @@
 import datetime
+
+from corehq.util.datadog.utils import count_by_response_code
 from django.http import HttpResponse, HttpResponseNotFound
 from django.views.decorators.http import require_GET
 from corehq.apps.domain.decorators import login_or_digest_or_basic_or_apikey, domain_admin_required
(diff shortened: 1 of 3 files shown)


corehq/apps/receiverwrapper/views.py
```diff
@@ -41,12 +41,26 @@
 
 def _process_form(request, domain, app_id, user_id, authenticated,
                   auth_cls=AuthContext):
+    metic_tags = [
```
gcapalbo: not blocking but did you mean `metric_tags`

gcapalbo (approved):

corehq/apps/receiverwrapper/views.py
snopoke: 891b8ac

corehq/apps/mobile_auth/views.py
```diff
@@ -60,6 +62,7 @@ def get_payload(self):
 
 @login_or_digest_or_basic_or_apikey()
 @require_GET
+@count_by_response_code('commcare.auth_keys.fetch.count')
```
benrudolph: can we get this metric from `nginx.requests` by filtering for a url?

snopoke: we can, though we may have to get rid of the url tagging if we don't want to pay for custom metrics

PR #15964, opened 2017-05-01
Externalize JavaScript: form designer

orangejenny:
This gets rid of all inline js in `form_designer.html` except for the guided tour, which I'll deal with in another PR. I've done some sanity testing in both v1 and v2, but marking "don't merge" until I do a bit more testing.

@millerdev / @dannyroberts

PR #15997, opened 2017-05-03
Revert notifications changes to form designer

orangejenny:
Reverts changes made in https://github.com/dimagi/commcare-hq/pull/15964/commits/3670937c4f49585c7b3ca9023326434e085aed0e which is causing notifications to appear repeatedly. Will re-do at a later date.

@nickpell / @millerdev 
fyi @proteusvacuum


millerdev (approved):
😢

orangejenny:
ha, agreed that this is sad...i'm hoping it's something small (maybe the fact that the `websockets_override` is now called twice?), i just won't have a chance to look at it for a couple of days

PR #15998, opened 2017-05-03
Don't forward archived cases to 99DOTS or Nikshay

proteusvacuum:
@NoahCarnahan @mkangia 

https://sentry.io/dimagi/commcarehq/issues/265692476/


mkangia (approved):
Thanks for fixing this @proteusvacuum.

PR #16001, opened 2017-05-03
private sector datamigration

nickpell:
This is a work in progress, but figured it would be good to merge with master and get feedback.

The management command ```create_cases_from_beneficiary``` is run to populate person, occurrence, episode, and adherence cases based off of the data dumps that we get from UATBC (imported into postgres).  Prescription and test cases are included for tests but I'm not populating those case properties yet, since they have low priority.

@proteusvacuum @mkangia cc: @orangejenny

PR #16015, opened 2017-05-04
follow up to #16001

nickpell:
#16001 

@proteusvacuum @mkangia cc: @orangejenny 

recommend 🐡


custom/enikshay/private_sector_datamigration/models.py
```diff
@@ -229,6 +229,16 @@ def current_patient_type_choice(self):
         return ''  # TODO - handle
 
     @property
+    def diabetes_status(self):
+        return {
+            'Yes': 'diabetic',
+            'No': 'non-diabetic',
```
proteusvacuum: `non_diabetic`

proteusvacuum (commented):
👍

custom/enikshay/private_sector_datamigration/models.py
nickpell: thanks, updated

PR #16017, opened 2017-05-04
owner id case search

proteusvacuum:
Adds the ability to search for `owner_id` to case search

Also adds case indices to the debug view and removes the raw response (which you can look up in the network panel anyway)

@NoahCarnahan cc: @benrudolph

### DIFF
--- a/corehq/apps/case_search/static/case_search/js/case_search.js
+++ b/corehq/apps/case_search/static/case_search/js/case_search.js
@@ -41,21 +41,6 @@ hqDefine('case_search/js/case_search.js', function(){
                 )},
                 success: function(data){
                     self.results(data.values);
-                    var values = JSON.stringify(data.values, null, '    ');
-                    if (self.codeMirror === null){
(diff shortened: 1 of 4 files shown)


corehq/apps/case_search/static/case_search/js/case_search.js
```diff
@@ -1,11 +1,12 @@
-/* globals hqDefine, ko, $, CodeMirror */
+/* globals hqDefine, ko, $ */
 
 hqDefine('case_search/js/case_search.js', function(){
     'use strict';
     return function(case_data_url){
         var self = this;
         self.codeMirror = null;
```
NoahCarnahan: remove this property too?

NoahCarnahan (commented):
Looks good, just one small thing.

corehq/apps/case_search/static/case_search/js/case_search.js
proteusvacuum: 👍

PR #16018, opened 2017-05-04
clean up topic usage

emord:
Cleans up places where we were using document types instead of kafka topics to pass into pillows

@dannyroberts


corehq/pillows/case.py
```diff
@@ -43,7 +43,7 @@ def get_case_to_elasticsearch_pillow(pillow_id='CaseToElasticsearchPillow', **kw
         index_info=CASE_INDEX_INFO,
         doc_prep_fn=transform_case_for_elasticsearch
     )
-    kafka_change_feed = KafkaChangeFeed(topics=[topics.CASE, topics.CASE_SQL], group_id='cases-to-es')
+    kafka_change_feed = KafkaChangeFeed(topics=topics.CASE_TOPICS, group_id='cases-to-es')
```
dannyroberts: I assume this accepts either [lists] or (tuples,) for `topics`?

dannyroberts (approved):
Approve

corehq/pillows/case.py
emord: Yeah. we expand it before passing to the library https://github.com/dimagi/commcare-hq/blob/je/kafka-checkpoint-only/corehq/apps/change_feed/consumer/feed.py#L119

dannyroberts: 👍

PR #16019, opened 2017-05-04
UCRs support partitioned kafka change feed

emord:
Sets up partitioned processing for the UCR pillows. Just doing UCR right now as it's a bit more of a refactor to get all pillows to support it (mostly just some tests failing).

@dannyroberts @dimagi/scale-team


dannyroberts (commented):
Does this kind of replace part of the PR you closed? Or is this orthogonal to anything I've seen so far?

snopoke (approved):
Test failures related to broken tests, not broken functionality.

corehq/apps/cleanup/migrations/0012_add_es_index_to_checkpoint_ids.py
```diff
@@ -59,6 +59,7 @@ class Migration(migrations.Migration):
 
     dependencies = [
         ('cleanup', '0011_merge_couch_sql_pillows'),
+        ('pillowtop', '0003_auto_20170411_1957'),
```
emord: I think we could actually no-op migrations in the cleanup directory after they've been run, but will leave that for another day

emord:
@dannyroberts this replaces a large part of that PR. I think the only thing that is left out is making the other pillows support kafka partitions which will happen here https://github.com/dimagi/commcare-hq/pull/16020

I split them out because UCRs should be ready to go and that'll be the biggest win for ICDS

dannyroberts (approved):

emord:
is is now deployed to NIC. There were 2 commits since official approval so not going to self merge

https://github.com/dimagi/commcare-hq/pull/16019/commits/d05be4e58c89035fb1411618fe382f60331db5b9
https://github.com/dimagi/commcare-hq/pull/16019/commits/b1700394db02c1dd468a54283d46d19696807c27

It does "break" part of the hq/admin/system page. All pillows show up as unknown because they are now named for the process number. for example instead of kafka-ucr-special-icds-ucr it is now kafka-ucr-special-icds-ucr-0. Not a big deal for right now as there are other ways to make sure pillows are running (like checking the time since last checkpoint on that same page), but will put on my backlog to fix that

snopoke (approved):

PR #16022, opened 2017-05-05
add sync ID and device ID to async restore cache key

snopoke:
- device ID added to prevent edge case where the same user syncs from different devices
- synclog ID doesn't seem likely to have any effect but also doesn't harm

### DIFF
--- a/corehq/ex-submodules/casexml/apps/phone/restore.py
+++ b/corehq/ex-submodules/casexml/apps/phone/restore.py
@@ -636,8 +636,6 @@ class RestoreConfig(object):
             self.cache_settings.overwrite_cache
         )
 
-        self.asyc = async
-
         self.force_cache = self.cache_settings.force_cache or self.async
         self.cache_timeout = self.cache_settings.cache_timeout
         self.overwrite_cache = self.cache_settings.overwrite_cache
--- a/corehq/ex-submodules/casexml/apps/phone/restore.py
+++ b/corehq/ex-submodules/casexml/apps/phone/restore.py
@@ -653,7 +653,13 @@ class RestoreConfig(object):
 
     @property
     def async_cache_key(self):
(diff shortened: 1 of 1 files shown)


corehq/ex-submodules/casexml/apps/phone/restore.py
```diff
@@ -655,7 +653,13 @@ def sync_log(self):
 
     @property
     def async_cache_key(self):
-        return restore_cache_key(self.domain, ASYNC_RESTORE_CACHE_KEY_PREFIX, self.restore_user.user_id)
+        return restore_cache_key(
```
benrudolph: i thought about doing this as well but held off because it changed this: https://github.com/dimagi/commcare-hq/blob/br/replicate-couch/corehq/form_processor/submission_post.py#L228 (the invalidation). i wasn't sure in what ways async restore needed that. i'm guessing it doesn't make much of a difference. in regular syncs it doesn't matter because it deletes the caches if there is no device id / sync log id and it's irrelevant if the cache has the sync log id since the cache won't get used again. i wasn't sure if i could make the same assumptions about async restore. guessing you can, but wanted to flag just in case

benrudolph (approved):

corehq/ex-submodules/casexml/apps/phone/restore.py
proteusvacuum: async restore stores the ID of running tasks in the cache. If there is a task currently running with that ID it returns a 202 with the current progress of the sync. 
Adding device ID here means that if you start a restore on one device and then try and restore on a different one while the task is running, it will create 2 tasks, which I think might be what you want. If the cache of that key is deleted, it won't find the task, so will kick off a new one. 
It should be fine to add the synclog_id too.

proteusvacuum (approved):

PR #16025, opened 2017-05-05
multi get es

emord:
@esoergel @dannyroberts 

This makes the tests a bit more sketchy, but I think this will get rid of the ES timeouts we've been seeing on ICDS

### DIFF
--- a/custom/icds_reports/ucr/tests/test_expressions.py
+++ b/custom/icds_reports/ucr/tests/test_expressions.py
@@ -13,7 +13,7 @@ es_form_cache = []
 
 def mget_query_fake(index, ids, source):
     return [
-        form
+        {"_id": form['_id'], "_source": form}
         for form in es_form_cache
         if form['_id'] in ids
     ]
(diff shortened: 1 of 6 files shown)


custom/icds_reports/ucr/tests/test_expressions.py
```diff
@@ -1,36 +1,48 @@
...
 from django.test import TestCase, override_settings
 import mock
 from casexml.apps.case.mock import CaseStructure, CaseFactory
 from casexml.apps.case.tests.util import delete_all_cases, delete_all_xforms
-from corehq.apps.es.fake.forms_fake import FormESFake
 from corehq.apps.userreports.expressions.factory import ExpressionFactory
 from corehq.apps.userreports.specs import EvaluationContext
 from corehq.form_processor.interfaces.dbaccessors import FormAccessors
 
+es_form_cache = []
```
esoergel: I'm not crazy about these test globals - seems like they should be attached to the test class.  Doing so sounds a bit tricky, though, and it may end  up more complicated anyways, so I'm cool with it.

esoergel (approved):

custom/icds_reports/ucr/tests/test_expressions.py
emord: Yeah. I really didn't like it either. I wasn't sure on how to mock this without a lot of extra work. I'm happy to refactor if there's a better pattern I can use

dannyroberts (approved):

PR #16026, opened 2017-05-05
fix 500 at /a/{domain}/reports/case_data/{case_id}/attachments/ when …

nickpell:
…not authenticated

@orangejenny


nickpell:
https://sentry.io/dimagi/commcarehq/issues/266337148/

orangejenny (approved):

PR #16028, opened 2017-05-05
use simplejson instead of json in laboratory

emord:
Found errors going through celery logs

@dannyroberts


dannyroberts (approved):
ha, that's a little annoying :)

PR #16031, opened 2017-05-05
The marriage of mobile UCR and case detail graphing

orangejenny:
I've wanted to do this for a long time; doing it now because Ali wants to modify the series x value for one of her projects, which mobile UCR's UI doesn't allow. As a positive side effect, mobile UCR now also supports locale-specific configuration (axis titles, etc.) and annotations. This will also make it easier to support custom threshold lines in the future (although there's some mobile work that needs to be done first, because mobile doesn't support mixed bar/line graphs yet).

There's a lazy migration in here for `ReportAppConfig`, to convert its graph configs from `ReportGraphConfig` to `GraphConfiguration`. I'm planning to follow this up with a real migration and then deprecate `ReportGraphConfig` but didn't want to block this PR.

screen shot 2017-05-05 at 3 08 57 pm

screen shot 2017-05-05 at 3 11 31 pm

The "default" values that mobile UCR uses for x, y, and nodeset are populated in the UI as placeholders. If the user leaves them alone, they'll continue to be populated during suite creation, based on the UCR's current chart configuration. If the user overrides any of these values, they'll get a warning that value will not be updated if they change the report's chart config in the future:

screen shot 2017-05-05 at 3 35 28 pm

Easiest to review by commit. Would be nice to review soon-ish, because Ali's waiting on this.

code buddy / mobile UCR @nickpell 
app manager @dannyroberts / @snopoke

### DIFF
--- a/corehq/apps/app_manager/app_strings.py
+++ b/corehq/apps/app_manager/app_strings.py
@@ -129,6 +129,16 @@ def _create_custom_app_strings(app, lang, for_default=False):
                         id_strings.report_column_header(config.uuid, column.column_id),
                         column.get_header(lang)
                     )
+                for chart_id, graph_config in config.complete_graph_configs.iteritems():
+                    for index, item in enumerate(graph_config.annotations):
+                        yield id_strings.mobile_ucr_annotation(module, config.uuid, index), trans(item.values)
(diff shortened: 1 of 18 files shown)


nickpell (approved):
👏

snopoke:
This is great. Thanks!

PR #16037, opened 2017-05-05
Add user and location repeaters for BETS

esoergel:
Built on top of https://github.com/dimagi/commcare-hq/pull/15957


custom/enikshay/integrations/bets/models.py
```diff
@@ -0,0 +1,2 @@
+# Something bets.repeater_generators must be imported to allow repeater generators to be registered
+from custom.enikshay.integrations.bets import repeater_generators
```
dimagimon: F401 'repeater_generators' imported but unused

PR #16040, opened 2017-05-06
cast row to list

nickpell:
https://manage.dimagi.com/default.asp?253373


orangejenny (approved):
Code looks fine. Do you think this is a long-standing bug? The code being changed looks like it's been around for a while (though I didn't dig around to see if related code has changed recently).

nickpell:
It only fails when the data source is backed by elastic search, which is a relatively new addition.  The code was originally written for a postgesql backed data source, which does not error.

orangejenny:
cool, thanks for explaning

PR #16041, opened 2017-05-08
properly handling signature header

calellowitz:
@millerdev 
This fixes an error in the code that handles our greenhouse web hook. The actual header format is `hash signature` not `signature`

### DIFF
--- a/corehq/apps/analytics/views.py
+++ b/corehq/apps/analytics/views.py
@@ -5,8 +5,8 @@ import json
 from django.http import HttpResponse
 from django.views.decorators.csrf import csrf_exempt
 from django.views.generic import View
+from django.conf import settings
 
-import settings
 from corehq.apps.analytics.tasks import track_clicked_deploy_on_hubspot, track_job_candidate_on_hubspot
 from corehq.apps.analytics.utils import get_meta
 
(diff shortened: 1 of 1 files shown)


corehq/apps/analytics/views.py
```diff
@@ -31,7 +31,7 @@ def post(self, request, *args, **kwargs):
         digester = hmac.new(settings.GREENHOUSE_API_KEY, request.body, hashlib.sha256)
         calculated_signature = digester.hexdigest()
 
-        if str(calculated_signature) == str(request.META.get('HTTP_SIGNATURE', '')):
+        if str(calculated_signature) == str(request.META.get('HTTP_SIGNATURE'.split()[1], '')):
```
millerdev: This is splitting the string `'HTTP_SIGNATURE'`, which I'm pretty sure is not what you want and will result in an index error due to the `[1]`.

It seems possible that the value of the `HTTP_SIGNATURE` header (?) may not have two parts when split, which means `[1]` may throw an index error. What do you think of making that code more fault-tolerant? (probably can't do it in one line and keep it readable)

millerdev (approved):

PR #16046, opened 2017-05-08
App manager v2: case label

orangejenny:
https://manage.dimagi.com/default.asp?253565

@biyeun any objections to moving the case label up above save so it's more like a title? I think it makes sense conceptually, and it's simpler to implement (because everything else on the page is part of the detail, but case_label is a module attribute, saving gets messy).

Before:
screen shot 2017-05-08 at 9 50 46 am

After:
screen shot 2017-05-08 at 9 46 08 am

code buddy @nickpell


biyeun (approved):

orangejenny:
forgot to run v2 diffs

nickpell (approved):

PR #16048, opened 2017-05-08
Remove CodeMirror from admin restore pages

proteusvacuum:
Because that page contains useful information, and codemirror kills your browser.

The restore is still well indented, it just isn't foldable or syntax highlighted. If you want that, click "Raw" and your browser will do that for you in a much nicer way. 

@benrudolph


benrudolph (approved):
1000x yes

orangejenny:
yay, good call

PR #16051, opened 2017-05-08
update django otp

emord:
django otp gets installed implicitly by django-two-factor, so it's whatever's latest at time of install. doing this here to get rid of some of the deprecation warnings

@dannyroberts

### DIFF
--- a/requirements/requirements.txt
+++ b/requirements/requirements.txt
@@ -73,6 +73,7 @@ git+git://github.com/smartfile/django-transfer.git@6e0dc94c3341c358fca8eb2bf74e2
 Pygments==2.0.2
 tinys3==0.1.11
 django-formtools==2.0
+django-otp==0.3.13
 django-two-factor-auth==1.6.0
 datadog==0.15.0
 django-websocket-redis==0.4.6


requirements/requirements.txt
```diff
@@ -73,6 +73,7 @@ git+git://github.com/smartfile/django-transfer.git@6e0dc94c3341c358fca8eb2bf74e2
 Pygments==2.0.2
 tinys3==0.1.11
 django-formtools==2.0
+django-otp==0.3.13
```
dannyroberts: this is up from 0.3.7 on mine. That seems fine / intended, but just checking—the point here is to upgrade to a (point-release) version that does not suffer from these deprecation warnings?

dannyroberts (approved):

requirements/requirements.txt
emord: That was the main reason, but after looking at it I also noticed django-two-factor-auth specifies >=0.3.4,<0.99, which means prod environments could have very different versions even if there aren't functional differences. 0.3.13 is the latest release

PR #16054, opened 2017-05-08
.track-usage-link

orangejenny:
Use a `track-usage-link` class and a single event handler to handle `gaTrackLink`/`trackUsageLink` calls, rather than sprinkling inline js all over. Same style as `.hq-help-template`.

Replaced all `gaTrackLink` calls in app manager, will do the rest of HQ shortly.

@nickpell / @calellowitz


nickpell:
very nice!

PR #16055, opened 2017-05-08
update MediaUploader submodule

orangejenny:
For https://github.com/dimagi/MediaUploader/pull/23

fyi @nickpell


orangejenny:
Merging. MediaUplaoder doesn't have tests.

PR #16057, opened 2017-05-08
make repeat record view safer

dannyroberts:
- check domain of record
- require login (and domain)

Fun demo, try opening https://www.commcarehq.org/a/hahahahahaha-not-really-a-domain/settings/project/repeat_record/?record_id=8f0906c8511a126d09dcab5cea7b67c0 in an incognito window. (Don't worry, it's fake data.) To exploit this vulnerability you would need to know a repeat record id (but don't have to even be logged in!), which may not be trivial to discover if you're not in the know, but definitely not treated by our team as a "secret" (we paste each other links, etc.).

@benrudolph

### DIFF
--- a/corehq/apps/domain/urls.py
+++ b/corehq/apps/domain/urls.py
@@ -10,7 +10,6 @@ from django.conf import settings
 from django.views.generic import RedirectView
 
 from corehq.apps.callcenter.views import CallCenterOwnerOptionsView
-from corehq.apps.domain.decorators import login_and_domain_required
 from corehq.apps.domain.forms import ConfidentialPasswordResetForm, HQSetPasswordForm
 from corehq.apps.domain.views import (
     ActivateTransferDomainView,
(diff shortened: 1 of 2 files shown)


dannyroberts:
@emord buddy

corehq/apps/domain/urls.py
```diff
@@ -154,7 +155,7 @@
         name=InternalSubscriptionManagementView.urlname),
     url(r'^billing_information/$', EditExistingBillingAccountView.as_view(),
         name=EditExistingBillingAccountView.urlname),
-    url(r'^repeat_record/', RepeatRecordView.as_view(), name=RepeatRecordView.urlname),
+    url(r'^repeat_record/', login_and_domain_required(RepeatRecordView.as_view()), name=RepeatRecordView.urlname),
```
emord: Why doesn't this use `LoginAndDomainMixin`?

dannyroberts: Even better. I'm just a luddite who never learned to use class-based views :)

corehq/apps/repeaters/views.py
```diff
@@ -33,18 +36,30 @@ def make_repeater(self):
...
+
     def get(self, request, domain):
-        record = RepeatRecord.get(request.GET.get('record_id'))
+        record = self.get_record_or_404(request, domain)
         content_type = record.repeater.get_payload_generator(
             record.repeater.format_or_default_format()
         ).content_type
         try:
-            payload = record.get_payload()
+            payload = record.get_payload(record)
```
emord: Was it intentional to change this? looks like it only takes save_failures https://github.com/dimagi/commcare-hq/blob/master/corehq/apps/repeaters/models.py#L511

benrudolph (commented):
yikes that's not good. nice catch. looks good except for the one comment emord already made

corehq/apps/repeaters/views.py
dannyroberts: Ah, no! Totally unintentional change, thanks for catching in review! (I was playing with some different changes on another branch, which I ended up not going with, and I just didn't properly/fully revert it.)

emord (approved):

PR #16058, opened 2017-05-08
eNikshay user and location repeaters

esoergel:
@NoahCarnahan @proteusvacuum 
@biyeun buddy
These repeaters will send user and location updates to BETS whenever their post_save methods are called.  They use the same serialization format as the public APIs.  Note that this is a PR into Noah's branch - I'm just trying to get a jump start on any feedback.

### DIFF
--- a/custom/enikshay/integrations/bets/repeaters.py
+++ b/custom/enikshay/integrations/bets/repeaters.py
@@ -294,5 +294,8 @@ class BETSLocationRepeater(Repeater):
 
 
 @receiver(post_save, sender=SQLLocation, dispatch_uid="create_bets_location_repeat_records")
-def create_location_repeat_records(sender, **kwargs):
+def create_location_repeat_records(sender, raw=False, **kwargs):
+    if raw:
+        return
+
     create_repeat_records(BETSLocationRepeater, kwargs['instance'])
--- a/custom/enikshay/integrations/bets/repeaters.py
+++ b/custom/enikshay/integrations/bets/repeaters.py
@@ -248,7 +248,7 @@ def create_case_repeat_records(sender, case, **kwargs):
 case_post_save.connect(create_case_repeat_records, CommCareCaseSQL)
 
 
-class UserRepeater(Repeater):
+class BETSUserRepeater(Repeater):
     friendly_name = _("Forward Users")
 
     class Meta(object):
@@ -267,7 +267,7 @@ class UserRepeater(Repeater):
 
 
 def create_user_repeat_records(sender, couch_user, **kwargs):
-    create_repeat_records(UserRepeater, couch_user)
+    create_repeat_records(BETSUserRepeater, couch_user)
 
 
 commcare_user_post_save.connect(create_user_repeat_records)
--- a/custom/enikshay/integrations/bets/repeaters.py
+++ b/custom/enikshay/integrations/bets/repeaters.py
@@ -1,7 +1,11 @@
 from django.utils.translation import ugettext_lazy as _
 from casexml.apps.case.signals import case_post_save
-from corehq.apps.repeaters.models import CaseRepeater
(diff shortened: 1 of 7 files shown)


custom/enikshay/integrations/bets/repeaters.py
```diff
@@ -242,3 +250,52 @@ def create_case_repeat_records(sender, case, **kwargs):
...
+
+    @classmethod
+    def available_for_domain(cls, domain):
+        return BETS_INTEGRATION.enabled(domain)
+
+    def __unicode__(self):
+        return "forwarding users to: %s" % self.url
+
+
+@receiver(commcare_user_post_save, dispatch_uid="create_bets_user_repeat_records")
```
proteusvacuum: nice.

custom/enikshay/integrations/bets/repeaters.py
```diff
@@ -242,3 +250,52 @@ def create_case_repeat_records(sender, case, **kwargs):
     create_repeat_records(BETSAYUSHReferralRepeater, case)
 
 case_post_save.connect(create_case_repeat_records, CommCareCaseSQL)
+
+
+class BETSUserRepeater(Repeater):
```
proteusvacuum: Any reason not to make these part of the product? Not sure if that is something other people might find useful.

custom/enikshay/integrations/bets/tests/test_repeaters.py
```diff
@@ -235,3 +241,119 @@ def test_trigger(self):
...
+            self.domain,
+            "{}@{}.commcarehq.org".format(username, self.domain),
+            "123",
+        )
+        self.addCleanup(user.delete)
+        return user
+
+    def test_trigger(self):
+        self.assertEqual(0, len(self.repeat_records().all()))
+        user = self.make_user("bselmy")
```
proteusvacuum: do you want to add a test that updating the user also triggers this?

custom/enikshay/integrations/bets/tests/test_repeaters.py
```diff
@@ -235,3 +241,119 @@ def test_trigger(self):
...
+            name=name,
+            site_code=name,
+            location_type=self.location_type,
+        )
+        self.addCleanup(location.delete)
+        return location
+
+    def test_trigger(self):
+        self.assertEqual(0, len(self.repeat_records().all()))
+        location = self.make_location('kings_landing')
```
proteusvacuum: should we add a test that when you edit a location, it also gets forwarded?

proteusvacuum (commented):
Looks great to me. Had a few non-blocking queries. Would be willing to merge as-is though - what was the reason for the Hold off flag?

custom/enikshay/integrations/bets/repeaters.py
esoergel: Yeah, I was thinking about that too - I wanted to start them this way so I wouldn't have a mental block from doing something BETS-specific if it would've been the smarter approach.  After having finished writing them, they are still generic, so maybe I will genericize them.

esoergel:
I stuck the hold-off flag there mostly 'cause I didn't wanna unnecessarily clutter up Noah's PR, but from the looks of it we could call that a "passed reviewed" branch, since he's going to PR the additional changes to it as well.

Thanks for the good suggestions, too - I'll take a pass at those.

custom/enikshay/integrations/bets/repeaters.py
proteusvacuum: cool. we can keep it as is for now, and you can update later if you want.

proteusvacuum:
Great. Is there anything in here that depends on Noah's branch? I think it might be a few days until that one gets completed, so it might be easiest if you cherry-picked whatever commits you need and we merge yours into master first.

PR #16059, opened 2017-05-08
Refactor repeater fire

dannyroberts:
The main thrust of the change is to move more of the controlling logic into the `Repeater`. It generally seems like a better place for them regardless, but this may help make it easier to write custom repeaters that handle the action portion differently (multiple requests, say, or something else).

Easiest to review commit by commit (🐡 if you will)

### DIFF
--- a/corehq/apps/repeaters/models.py
+++ b/corehq/apps/repeaters/models.py
@@ -315,9 +315,25 @@ class Repeater(QuickCachedDocumentMixin, Document, UnicodeMixIn):
             log_repeater_timeout_in_datadog(self.domain)
             raise RequestConnectionError(error)
         except Exception as e:
-            repeat_record.handle_exception(e)
+            self.handle_response(e, repeat_record)
         else:
-            repeat_record.handle_response(response)
+            self.handle_response(response, repeat_record)
+
+    def handle_response(self, result, repeat_record):
+        """
+        route the result to the success, failure, or exception handlers
+
+        result may be either a response object or an exception
+        """
+        if isinstance(result, Exception):
+            repeat_record.handle_exception(result)
+            self.handle_exception(result, repeat_record)
+        elif 200 <= result.status_code < 300:
+            repeat_record.handle_success(result)
+            self.handle_success(result, repeat_record)
+        else:
+            repeat_record.handle_failure(result)
+            self.handle_failure(result, repeat_record)
 
     def handle_success(self, response, repeat_record):
         """handle a successful post
@@ -553,19 +569,12 @@ class RepeatRecord(Document):
     def fire(self, force_send=False):
         self.repeater.fire_for_record(self, force_send=force_send)
 
-    def handle_response(self, response):
-        if 200 <= response.status_code < 300:
-            self.handle_success(response)
-        else:
-            self.handle_failure(response)
-
     def handle_success(self, response):
         """Do something with the response if the repeater succeeds
         """
         self.last_checked = datetime.utcnow()
         self.next_check = None
         self.succeeded = True
-        self.repeater.handle_success(response, self)
 
     def handle_failure(self, response):
         """Do something with the response if the repeater fails
@@ -574,13 +583,11 @@ class RepeatRecord(Document):
             u'{}: {}. {}'.format(response.status_code, response.reason, getattr(response, 'content', None)),
             response
         )
-        self.repeater.handle_failure(response, self)
 
     def handle_exception(self, exception):
         """handle internal exceptions
         """
         self._fail(unicode(exception), None)
-        self.repeater.handle_exception(exception, self)
 
     def _fail(self, reason, response):
         if self.repeater.allow_retries(response) and self.overall_tries < self.max_possible_tries:
--- a/corehq/apps/repeaters/models.py
+++ b/corehq/apps/repeaters/models.py
@@ -298,12 +298,6 @@ class Repeater(QuickCachedDocumentMixin, Document, UnicodeMixIn):
             return HTTPDigestAuth(self.username, self.password)
         return None
 
-    def fire_for_record(self, repeat_record, force_send):
-        if repeat_record.try_now() or force_send:
-            repeat_record.overall_tries += 1
-            self.post_for_record(repeat_record)
-            repeat_record.save()
-
     def post_for_record(self, repeat_record):
         headers = self.get_headers(repeat_record)
         auth = self.get_auth()
@@ -567,7 +561,10 @@ class RepeatRecord(Document):
         self.save()
 
     def fire(self, force_send=False):
-        self.repeater.fire_for_record(self, force_send=force_send)
+        if self.try_now() or force_send:
+            self.overall_tries += 1
+            self.repeater.post_for_record(self)
+            self.save()
 
     def handle_success(self, response):
         """Do something with the response if the repeater succeeds
--- a/corehq/apps/repeaters/models.py
+++ b/corehq/apps/repeaters/models.py
@@ -3,6 +3,7 @@ from datetime import datetime, timedelta
 import logging
 import urllib
 import urlparse
+import warnings
 from django.utils.translation import ugettext_lazy as _
 from requests.auth import HTTPBasicAuth, HTTPDigestAuth
 
@@ -295,7 +296,7 @@ class Repeater(QuickCachedDocumentMixin, Document, UnicodeMixIn):
         headers = self.get_headers(repeat_record)
         auth = self.get_auth()
         payload = repeat_record.get_payload()
-        url = repeat_record.url  # todo: does this ever not equal self.url?
+        url = self.get_url(repeat_record)
         domain = repeat_record.domain  # todo: does this ever not equal self.domain?
         try:
             response = simple_post_with_logged_timeout(domain, payload, url, headers=headers, timeout=POST_TIMEOUT,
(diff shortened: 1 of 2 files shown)


dannyroberts:
@proteusvacuum @emord

corehq/apps/repeaters/models.py
```diff
@@ -316,8 +310,10 @@ def post_for_record(self, repeat_record):
         payload = self.get_payload_and_handle_exception(repeat_record)
         url = self.get_url(repeat_record)
         try:
-            response = simple_post_with_logged_timeout(self.domain, payload, url, headers=headers,
-                                                       timeout=POST_TIMEOUT, auth=auth)
+            response = simple_post(payload, url, headers=headers, timeout=POST_TIMEOUT, auth=auth)
+        except (Timeout, ConnectionError) as error:
+            log_repeater_timeout_in_datadog(self.domain)
+            raise RequestConnectionError(error)
```
proteusvacuum: I don't think `repeat_record.handle_exception` is called if you move this here. I think it should still get called with this exception.

corehq/apps/repeaters/models.py
```diff
@@ -175,6 +168,24 @@ def get_payload(self, repeat_record):
         generator = self.get_payload_generator(self.format_or_default_format())
         return generator.get_payload(repeat_record, self.payload_doc(repeat_record))
 
+    def get_payload_and_handle_exception(self, repeat_record, save_failure=True):
+        try:
+            return self.get_payload(repeat_record)
+        except ResourceNotFound as e:
```
proteusvacuum: I guess this doesn't handle if the payload comes from SQL...

corehq/apps/repeaters/models.py
```diff
@@ -287,6 +298,37 @@ def get_auth(self):
...
+            log_repeater_timeout_in_datadog(self.domain)
+            raise RequestConnectionError(error)
+        except Exception as e:
+            self.handle_response(e, repeat_record)
+        else:
+            self.handle_response(response, repeat_record)
+
+    def handle_response(self, result, repeat_record):
+        """
+        route the result to the success, failure, or exception handlers
```
proteusvacuum: 👍

corehq/apps/repeaters/models.py
```diff
@@ -509,85 +552,39 @@ def try_now(self):
...
-                ))
-
-            if save_failure:
-                self._payload_exception(e, reraise=False)
-        except Exception as e:
-            if save_failure:
-                self._payload_exception(e, reraise=True)
-            else:
-                raise
+        warnings.warn("RepeatRecord.get_paylod is deprecated. Use Repeater._get_payload instead.")
```
proteusvacuum: sp: `get_payload`

corehq/apps/repeaters/models.py
```diff
@@ -509,85 +552,39 @@ def try_now(self):
...
 
     def fire(self, force_send=False):
-        headers = self.repeater.get_headers(self)
-        auth = self.repeater.get_auth()
         if self.try_now() or force_send:
             self.overall_tries += 1
-            tries = 0
-            post_info = PostInfo(self.get_payload(), headers, auth)
-            self.post(post_info, tries=tries)
+            self.repeater.post_for_record(self)
```
proteusvacuum: what if we want to define other methods for our repeater instead of POST?
e.g. SOAP or GET or some other thing? would you name this something different (e.g. `fire_for_record`)?

corehq/apps/repeaters/models.py
```diff
@@ -334,7 +376,10 @@ def get_url(self, repeat_record):
             # adapted from http://stackoverflow.com/a/2506477/10840
             url_parts = list(urlparse.urlparse(url))
             query = urlparse.parse_qsl(url_parts[4])
-            query.append(("app_id", self.payload_doc(repeat_record).app_id))
+            try:
+                query.append(("app_id", self.payload_doc(repeat_record).app_id))
+            except (XFormNotFound, ResourceNotFound):
+                return None
```
proteusvacuum: why return `None` here?

corehq/apps/repeaters/models.py
emord: Also I think it should be `Repeater.get_payload`

corehq/apps/repeaters/models.py
dannyroberts: Yeah a lot of these checks seemed pretty arbitrary and possibly unnecessary at this point. I limited myself to trying to just keep things equivalent though, and also make it clearer what exceptions are caught where so it's easier to reason with (and maybe do something about later)

corehq/apps/repeaters/models.py
dannyroberts: This is intended to match exactly previous behavior. It was moved in from `Repeater.url` in https://github.com/dimagi/commcare-hq/pull/16059/commits/12b6288bfc7a3c656aac9b0c1515f2f0a1f3f6e4. I also checked that this function (`get_url`) was not called anywhere other than in `Repeater.url`, (and now after this change, it's not called anywhere except from places that _used_ to call the now-deprecated `Repeater.url`).

Why it was originally chosen to catch these particular errors and return `None`, I didn't try to deeply re-examine. On the face of it, it's a reasonable contract for `get_url` not to expose whatever the underlying errors are and just return `None` to mean "I couldn't get that", but of course the caller would need to check for that, and as you probably guessed, they don't. This again, just makes that clearer but doesn't try to fix it.

corehq/apps/repeaters/models.py
dannyroberts: Ah you're both right, thanks. It should say "Use Repeater.get_payload_and_handle_exception instead" now.

corehq/apps/repeaters/models.py
dannyroberts: That's a good suggestion. I like `fire_for_record`, now that I ended up moving `fire` back into `RepeatRecord` (in an earlier commit, I'd moved it into `Repeater`, so I needed to differentiate the two methods and kind of kept the same vocabulary that we'd been using previously). I'll make this change.

corehq/apps/repeaters/models.py
dannyroberts: Oh dear, you're totally right. Thanks for the keen eye.

corehq/apps/repeaters/models.py
```diff
@@ -307,7 +307,7 @@ def post_for_record(self, repeat_record):
             response = simple_post(payload, url, headers=headers, timeout=POST_TIMEOUT, auth=auth)
         except (Timeout, ConnectionError) as error:
             log_repeater_timeout_in_datadog(self.domain)
-            raise RequestConnectionError(error)
+            self.handle_response(RequestConnectionError(error), repeat_record)
```
dannyroberts: What do you think about this to address your (correct) comment?

The alternative is nesting another try/except in the try block. I'm leaning towards this because it keeps it flat and has the same effect, the only downside being the somewhat surprising choice to pass a newly created exception directly to a function instead of raising it.

corehq/apps/repeaters/models.py
```diff
@@ -287,6 +298,37 @@ def get_auth(self):
...
+    def fire_for_record(self, repeat_record):
+        headers = self.get_headers(repeat_record)
+        auth = self.get_auth()
+        payload = self.get_payload_and_handle_exception(repeat_record)
+        url = self.get_url(repeat_record)
+        try:
+            response = simple_post(payload, url, headers=headers, timeout=POST_TIMEOUT, auth=auth)
+        except (Timeout, ConnectionError) as error:
+            log_repeater_timeout_in_datadog(self.domain)
+            self.handle_response(RequestConnectionError(error), repeat_record)
```
dannyroberts: @proteusvacuum What do you think about this to address your (correct) comment?

The alternative is nesting another try/except in the try block. I'm leaning towards this because it keeps it flat and has the same effect, the only downside being the somewhat surprising choice to pass a newly created exception directly to a function instead of raising it.

dannyroberts:
@proteusvacuum @emord thanks for the review, I've tried to address all your feedback.

corehq/apps/repeaters/models.py
proteusvacuum: The alternative is to just let the next `except` block handle this type of error, which I think might be fine too. Our custom Repeaters currently log a `RequestConnectionError` to the case, but you could just as easily change those references to `if isinstance(Timeout, ConnectionError)` which is a little more verbose, but remains consistent. 

Or, just have the custom repeaters log every exception to the case, which I'd be in favor of too.

corehq/apps/repeaters/models.py
proteusvacuum: Sounds like the current behaviour is incorrect? If I call `get_url` and it returns `None`, what will the RepeatRecord do when it fires?

dannyroberts: Yeah I agree it's or at the very least not obviously correct. Mind if we just make a note of it and address this outside this PR though? My thinking is I want to separate as much as possible organizational changes and behavior changes, so that if any bug does come up, I know it's a bug in my reorganization, as opposed to a consequence of an intentional change. Unless of course I'm misunderstanding and you're saying this change makes things _worse_?

corehq/apps/repeaters/models.py
dannyroberts: Hadn't thought of that, but makes sense. I think to me it sounds like kind of a wash, so all things equal I think I'll just leave it how it is.

Agree we should improve how we log errors (and other things) in a repeat attempt, but I think we'll handle that in another PR (or more likely, other PRs).

proteusvacuum: Sure.

corehq/apps/repeaters/models.py
proteusvacuum: Sure. I don't think this makes things worse, just the current situation seems silly.

dannyroberts:
Offlined with @proteusvacuum and got his verbal approval to merge.

proteusvacuum:
🎉 
@mkangia FYI, this is going to cause conflicts in your branch, but will be a nicer API to work with.

PR #15846, opened 2017-04-21
App manager v2: Add module/form name tooltips to menu

orangejenny:
Let's see if this gets annoying.

screen shot 2017-04-21 at 4 17 10 pm

Also fixing injection vulnerability (see https://github.com/dimagi/commcare-hq/pull/14313#pullrequestreview-14182930).

@biyeun / @dannyroberts

PR #16061, opened 2017-05-09
Overflow with ellipsis

benrudolph:
@wpride @dimagi/ux 

Overflow long breadcrumbs with ellipsis

screen shot 2017-05-09 at 10 30 25 am

cc: @proteusvacuum @ctsims


orangejenny:
Looks good to me. One thing to consider is whether or not it'd be helpful to have a way to see the full text. https://github.com/dimagi/commcare-hq/pull/15846 has an example of this.

benrudolph:
👍 good idea

PR #16062, opened 2017-05-09
Added test to prove bug in caching in named expression

kkrampa:


kkrampa:
```======================================================================
FAIL: corehq.apps.userreports.tests.test_expressions:NestedExpressionTest.test_caching
  File "/mnt/commcare-hq-ro/corehq/apps/userreports/tests/test_expressions.py", line 595, in test_caching
    self.assertEqual(expression2(doc, evaluation_context), 'my_parent_id2')
AssertionError: 'my_parent_id' != 'my_parent_id2'
```

@emord It seems caching data doesn't work properly if we don't work on root doc itself. I'm not sure what's the best way to fix this. Can we get rid of caching for now? It causes problems in enikshay reports.

PR #16063, opened 2017-05-09
ReportConfigs might use old slug

kaapstorm:
Context: [FB 251023](https://manage.dimagi.com/default.asp?251023)

Saved or e-mailed reports use ReportConfig instances that store their filters keyed by old slugs. This allows the Application Status report to accept users specified with either the new or the old slug. (Changing the filter will use the new slug.)

@mkangia @calellowitz


kaapstorm:
@calellowitz I'm not sure of the implications of this change. An alternative would be to migrate ReportConfigs that have an "emw" filter to "location_restricted_mobile_worker" filter instead.

mkangia:
Would be nice to update the reports instead of adding support for old filter unless you are trying to avoid it for some reason.

mkangia (changes requested):
@kaapstorm Are you thinking of revisiting this?

kaapstorm:
Hey @mkangia. I've added a soft-migrate to change existing ReportConfigs from using `ExpandedMobileWorkerFilter` ("emw") filter to `LocationRestrictedMobileWorkerFilter` ("location_restricted_mobile_worker") filter instead. But I haven't tested this yet.

corehq/apps/reports/models.py
```diff
@@ -218,6 +219,18 @@ class ReportConfig(CachedCouchDocumentMixin, Document):
...
 
+    @classmethod
+    def wrap(cls, data):
+        if 'filters' in data and ExpandedMobileWorkerFilter.slug in data['filters']:
+            data['filters'][LocationRestrictedMobileWorkerFilter.slug] = \
+                data['filters'][ExpandedMobileWorkerFilter.slug]
+            del data['filters'][ExpandedMobileWorkerFilter.slug]
+        try:
+            self = super(ReportConfig, cls).wrap(data)
+        except AttributeError:
```
kaapstorm: (Catching `AttributeError` because parent `wrap()` method may not be found by method resolution.)

PR #16064, opened 2017-05-09
adding location multi column index

calellowitz:
@millerdev @snopoke 
This should speed up location uploads (and all location queries)


millerdev (approved):

snopoke (approved):
Assuming you verified that the index name will match the one you created manually.

calellowitz:
I used sqlmigrate to print the command this would run and it matched. im
assuming the hash is consistent but dont know how its generated so cant be
sure.

snopoke:
Based on my previous testing of this the hashes are consistent so that
should be fine.

PR #16066, opened 2017-05-09
Fact Table warehouse model and states

benrudolph:
@dimagi/scale-team Here's my initial go for the fact table state models. Definitely not fully fleshed out and probably won't be until I begin writing the code to generate the tables. I decided to use the django-fsm library to manage the states for this. Looks like a pretty solid library: https://github.com/kmmbvnr/django-fsm. And I think it'll provide some constraints and clarity over what state the fact table is in. And it can generate a sweet state diagram:

warehouse_transitions

cc: @proteusvacuum


corehq/warehouse/models.py
```diff
@@ -0,0 +1,64 @@
+from django.db import models
+from django_fsm import FSMField, transition, RETURN_VALUE
```
dimagimon: F401 'RETURN_VALUE' imported but unused

localsettings.example.py
```diff
@@ -204,6 +204,7 @@
 #    'dimagi.utils',
 #    'testapps.test_elasticsearch',
 #    'testapps.test_pillowtop',
+#    'django_fms', # Adds the ability to generate state diagrams for models using django-fsm
```
dimagimon: E122 continuation line missing indentation or outdented

benrudolph:
The graph is a bit confusing cause every state can lead to failed or decommissioned

corehq/warehouse/models.py
```diff
@@ -0,0 +1,64 @@
...
+    FACT_TABLE_FAILED,
+    FACT_TABLE_DECOMMISSIONED,
+)
+
+
+class FactTableState(models.Model):
+    domain = models.CharField(max_length=255)
+    report_slug = models.CharField(max_length=255)
+    state = FSMField(default=FACT_TABLE_NEEDS_UPDATING)
+    last_modified = models.DateTimeField()
```
emord: You could use auto_now

localsettings.example.py
emord: should that be django_fsm?

benrudolph: yes, yes it should

dimagimon: E122 continuation line missing indentation or outdented

snopoke:
Can you give a bit more info about each state.

localsettings.example.py
```diff
@@ -204,6 +204,7 @@
 #    'dimagi.utils',
 #    'testapps.test_elasticsearch',
 #    'testapps.test_pillowtop',
+#    'django_fsm', # Adds the ability to generate state diagrams for models using django-fsm
```
dimagimon: E122 continuation line missing indentation or outdented

benrudolph:
Added some comments here: https://github.com/dimagi/commcare-hq/pull/16066/commits/42c829cb30f2ab9d2948feecdbe9e83a951111f4. let me know if you would like more explanation

corehq/warehouse/states.py
```diff
@@ -0,0 +1,34 @@
...
+'''
+When a fact table is in a failed state, it means that it
+has failed to progress during one of the transitions due
+to an error.
+'''
+FACT_TABLE_FAILED = 'failed'
+
+'''
+When a fact table is in a decommissioned state, it means
+that it no process will attempt to re-process it unless
```
emord: > that it no process will attempt to re-process it unless

...that it will make no attempt to reprocess it unless... ?

corehq/warehouse/states.py
```diff
@@ -0,0 +1,34 @@
+'''
+When a fact table is in the "needs_updating" state, it will be
+selected for processing during the task that determines which
+fact tables need updating.
+'''
+FACT_TABLE_NEEDS_UPDATING = 'needs_updating'
+
+'''
+When a fact table has reached the batch dumped state, it will
```
emord: nitpicky: Can you change this to start with "When a fact table is in"? Or change the rest of the comments os that these are in the same format?

benrudolph: 👍

corehq/warehouse/states.py
benrudolph: > there is manual intervention

benrudolph: https://github.com/dimagi/commcare-hq/pull/16066/commits/f2337ab34a9498d399150e3a38591000a4afc59a

corehq/warehouse/states.py
```diff
@@ -6,7 +6,7 @@
 FACT_TABLE_NEEDS_UPDATING = 'needs_updating'
 
 '''
-When a fact table has reached the batch dumped state, it will
+When a fact table in in the batch dumped state, it will
```
emord: > in in

benrudolph: omg wow

gcapalbo:
this is an interesting approach i've never seen. the only thing i'd be concerned about is how hard it might be to debug if something goes wrong.

gcapalbo:
(sorry didn't mean to close, hit the wrong button. second time that's happened today)

benrudolph:
> this is an interesting approach i've never seen. the only thing i'd be concerned about is how hard it might be to debug if something goes wrong.

i tried to choose this approach based on what we discussed about batching. i also used the framework in order to make it easier if something goes wrong (make it simpler to reason about the state the fact table is in and also has logging when we make a transition to a new state). how have you seen it implemented in the past and do you have any suggestions for making the parts easier to debug?

gcapalbo:
i guess it's just not clear to me from the state machine diagram what would be the right next action if we're ever in a failed or decommissioned state.

usually the transitions are just simple one-way transitions in a step-wise manner, so the only way you move to the next state is by completing the current one, and there aren't any loops that take you back.

also it's important for all the transitions to be idempotent, so if something fails then you know you can just fix whatever the error was and then restart the process from the failed step.

maybe we should have a quick chat about this in case i'm missing how the etl is planned to work, but i think there's some overlap here between job states and the state of the current running etl process that we'd want to split out.

so usually we'll track where we are in the etl process with one value (are we loading a staging table from the source, are we calculating the delta with the dimension, etc.), and then track the state of that current job (running, succeeded, failed) separately.

this makes it clear when there's a failure and you always know that you can restart the current running etl process again when a failure happens.

benrudolph:
> i guess it's just not clear to me from the state machine diagram what would be the right next action if we're ever in a failed or decommissioned state.

I think my initial intention for failed state was to be able to track when a fact table fails to update. A failed state gets reprocessed (somewhat similar to how pillow errors can be reprocessed). decommissioned state is for removing a fact table from being updated. either because it keeps erroring or because we no longer need to update it. it was more of a state that i thought might be useful, but i don't have any real plans for it beyond that.

> usually the transitions are just simple one-way transitions in a step-wise manner, so the only way you move to the next state is by completing the current one, and there aren't any loops that take you back.

interesting so there'd be no different workflow if the fact table fails to update? i opted not to do this because if a fact table fails somewhere along the line, we don't know for certain what caused it and we have no choice but to start anew. for example if there was a bug in dumping to the intermediate table but did not surface until the ETL stage, we would never be able to update that fact table unless we completely started it over.

> also it's important for all the transitions to be idempotent, so if something fails then you know you can just fix whatever the error was and then restart the process from the failed step.

agree being idempotent is key, though i still would argue that depending on the bug, we could have to start from scratch

> so usually we'll track where we are in the etl process with one value (are we loading a staging table from the source, are we calculating the delta with the dimension, etc.), and then track the state of that current job (running, succeeded, failed) separately.

interesting, would be happy to chat about this. in general this seems like a good idea; i also have some questions about it

gcapalbo:
On Thu, May 11, 2017 at 2:48 PM, Ben Rudolph <notifications@github.com>
wrote:

> i guess it's just not clear to me from the state machine diagram what
> would be the right next action if we're ever in a failed or decommissioned
> state.
>
> I think my initial intention for failed state was to be able to track when
> a fact table fails to update. A failed state gets reprocessed (somewhat
> similar to how pillow errors can be reprocessed). decommissioned state is
> for removing a fact table from being updated. either because it keeps
> erroring or because we no longer need to update it. it was more of a state
> that i thought might be useful, but i don't have any real plans for it
> beyond that.
>
> usually the transitions are just simple one-way transitions in a step-wise
> manner, so the only way you move to the next state is by completing the
> current one, and there aren't any loops that take you back.
>
> interesting so there'd be no different workflow if the fact table fails to
> update? i opted not to do this because if a fact table fails somewhere
> along the line, we don't know for certain what caused it
>
Not sure I agree - if we don't know what caused the error how would we stop
it from happening again? Failures should be rare but usually mean that it
requires manual intervention to fix the problem which is why we usually
stop the job processing chain, fix the error, and restart it

> and we have no choice but to start anew. for example if there was a bug in
> dumping to the intermediate table but did not surface until the ETL stage,
> we would never be able to update that fact table unless we completely
> started it over.
>
> also it's important for all the transitions to be idempotent, so if
> something fails then you know you can just fix whatever the error was and
> then restart the process from the failed step.
>
> agree being idempotent is key, though i still would argue that depending
> on the bug, we could have to start from scratch
>
Agree, starting an incremental over from the beginning is also fine since
all the intermediate tables can be cleared out, and might be necessary if
the bug is in an earlier step

> so usually we'll track where we are in the etl process with one value (are
> we loading a staging table from the source, are we calculating the delta
> with the dimension, etc.), and then track the state of that current job
> (running, succeeded, failed) separately.
>
> interesting, would be happy to chat about this. in general this seems like
> a good idea; i also have some questions about it
>
  Yeah would be good to briefly sync about our plans for the etl process

corehq/warehouse/models/__init__.py
```diff
@@ -0,0 +1,12 @@
+from corehq.warehouse.models.dimensions import (
```
dimagimon: F401 'DomainDim' imported but unused
F401 'GroupDim' imported but unused
F401 'LocationDim' imported but unused
F401 'UserDim' imported but unused
F401 'UserGroupDim' imported but unused
F401 'UserLocationDim' imported but unused

corehq/warehouse/models/__init__.py
```diff
@@ -0,0 +1,12 @@
...
+    UserDim,
+    GroupDim,
+    LocationDim,
+    DomainDim,
+    UserLocationDim,
+    UserGroupDim,
+)
+
+from corehq.warehouse.models.meta import (
```
dimagimon: F401 'FactTableState' imported but unused

benrudolph:
Just to update the rest of you:

Gio and I had a meeting on Friday to discuss some of the points mentioned above. We talked through some edge case scenarios and how best to approach the batching infrastructure. Instead of updating dimensions via pillows, we're going to update them using the batching code. This is because, using pillows leads to some hard solve/track down edge cases where data can go missing.

This will now cause FactTables to have dependencies on UserDimensions being updated. This dependency graph can get quite complex, but I've found an open source tool that I think can help manage this complexity: https://medium.com/airbnb-engineering/airflow-a-workflow-management-platform-46318b977fd8

We also discussed the state diagram a bit. We agreed for now to just send the fact table back to the beginning for processing if it fails. Down the line we can get smarter about just retrying the current stage that it is on as an optimization.

We also discussed the complexity of attempting to update fact tables by domain, and decided that it actually introduces a lot of complexity unless we completely siloed each domain's warehouse in its own database. For now I think we are going to move forward with processing all domains at once with equal priority.

corehq/sql_db/routers.py
```diff
@@ -43,6 +44,8 @@ def allow_migrate(db, app_label):
         )
     elif app_label == SQL_ACCESSORS_APP:
         return db in partition_config.get_form_processing_dbs()
+    elif app_label == WAREHOUSE_APP:
```
gcapalbo: i think we'll need to add `hasattr(settings, "WAREHOUSE_DATABASE_ALIAS")` to this condition as well, otherwise if that setting is not present then it would always return False on the next line and the migration wouldn't run in any db

corehq/warehouse/models/dimensions.py
```diff
@@ -0,0 +1,86 @@
...
+
+    location_last_modified = models.DateTimeField()
+    location_created_on = models.DateTimeField()
+
+
+class DomainDim(BaseDim):
+    domain_id = models.CharField(max_length=255)
+    default_timezone = models.CharField(max_length=255)
+    hr_name = models.CharField(max_length=255)
+    creating_user_id = models.CharField(max_length=255)
```
gcapalbo: i think it would be preferable to have a foreign key to the user dim here

corehq/warehouse/models/meta.py
```diff
@@ -0,0 +1,61 @@
...
+from ..states import (
+    FACT_TABLE_NEEDS_UPDATING,
+    FACT_TABLE_BATCH_DUMPED,
+    FACT_TABLE_READY,
+    FACT_TABLE_FAILED,
+    FACT_TABLE_DECOMMISSIONED,
+)
+
+
+class FactTableState(models.Model):
```
gcapalbo: i don't remember if we discussed this or not but if we end up taking the batched approach for dims we should probably just have one generic table that stores the state of each batch for each dim/fact

corehq/warehouse/states.py
```diff
@@ -0,0 +1,34 @@
...
+to an error.
+'''
+FACT_TABLE_FAILED = 'failed'
+
+'''
+When a fact table is in a decommissioned state, it means
+that no process will attempt to re-process it unless
+there is manual intervention.
+'''
+FACT_TABLE_DECOMMISSIONED = 'decommissioned'
```
gcapalbo: if we end up using the batch processing tool i think it will end up tracking some of these states for us - maybe we can briefly chat again once you've had a chance to evaluate it a bit

corehq/sql_db/routers.py
benrudolph: yeah not necessarily a bad thing though. this is just used in the partition router, and i don't think we ever want the warehouse db to be in the main db. can update

corehq/warehouse/states.py
benrudolph: yeah i agree. i have some ideas i'm going to play around with today

corehq/warehouse/models/__init__.py
dimagimon: F401 'TableState' imported but unused

corehq/warehouse/models/dimensions.py
benrudolph: (offlined and we're going to leave as is to not create a dependency on the UserDim being up to date in order for the DomainDim to be up to date. in addition this property isn't really used much in reports)

corehq/warehouse/models/meta.py
benrudolph: i think this table is very much in flux, so hard right to now to determine what exactly we'll need. it'll be easy to add and remove states as well

corehq/warehouse/models/__init__.py
dimagimon: F401 'TableState' imported but unused

benrudolph:
ok ended up leaving the TableState model and will redefine/modify when the time comes. i also ensured that we do not migrate when the warehouse db is not defined so we can merge this without having the warehouse db in place

PR #16067, opened 2017-05-09
v2: missing dependency on hqmediauploaders.js

orangejenny:
Has a brain dead moment on https://github.com/dimagi/commcare-hq/pull/15971 and ran `build_app_manager_v2_diffs` without having actually updated the v2 template.

@nickpell / @dannyroberts


dannyroberts (approved):
This highlights also that the reviewer (me in this case!) should be more careful about reviewing changes that include these app diffs!

PR #16072, opened 2017-05-09
remove sync_logs_by_date

emord:
@dimagi/scale-team This is one of the views that always takes forever to preindex when trying to deploy on NIC. I don't think it's being used on master:

```
ag sync_logs_by_date
corehq/couchapps/__init__.py:24:    'sync_logs_by_date': settings.SYNCLOGS_DB,
```

The other two are

1.  commcarehq__synclogs, _design/phone-tmp
2. _design/by_domain_doc_type_date-tmp

They are both look to be used in a few different places, but likely aren't being hit by any endpoints with regularity.

Doubt this will make the deploy faster though, just seemed like some easy cleanup if it's not being used

@dannyroberts


emord:
I think that since this deletes the entire folder it can be done without a reindex, so should be safe to merge whenever without worrying about timing.

Going to put a hold off on merge tag so that we can confirm it's not needed

dannyroberts:
I think the only reason we were holding onto that view is out of the thought that we'd use it to go back and delete sync logs older than some date on a schedule. (Can't really do with the other view, since that's by user _then_ date.) But if it's causing problems, we should probably delete it and just reintroduce it or something else later.

emord:
That makes sense. There's actually an item on the icds trello to do that exact thing although on nic we could use by domain doc type date since there's only one domain.

I don't think that this will actually help deploy since there are those other two views. happy to close if we think this isn't good to merge now

dannyroberts:
Are those two other views also in the commcarehq__synclogs db, or just the first one?

I'm happy to have this merged, it may not remove the delay entirely, but it'll probably slow it and take load off couch.

dannyroberts (approved):

emord:
@dannyroberts the other two are in synclogs as well

dannyroberts:
Cool, up to you whether/when to merge based on deploy, but looks good to me and approve of merging.

emord:
going to use this in #16139

PR #16075, opened 2017-05-09
ICDS: redirect to domain-specific login

orangejenny:
https://trello.com/c/JKeZXJiw/153-change-default-login-page-of-icds

Non-domain login page no longer accessible. Can still access other domains by replacing the domain name in the login url.

Very open to suggestions if there is a less gross way to accomplish this.

@nickpell / @calellowitz

### DIFF
--- a/corehq/apps/hqwebapp/views.py
+++ b/corehq/apps/hqwebapp/views.py
@@ -351,6 +351,11 @@ def login(req):
     # this view, and the one below, is overridden because
     # we need to set the base template to use somewhere
     # somewhere that the login page can access it.
+
+    if settings.SERVER_ENVIRONMENT == 'icds':
+        login_url = reverse('domain_login', kwargs={'domain': 'icds-cas'})
+        return HttpResponseRedirect(login_url)
+
     req_params = req.GET if req.method == 'GET' else req.POST
     domain = req_params.get('domain', None)
     return _login(req, domain, "login_and_password/login.html")


corehq/apps/hqwebapp/views.py
```diff
@@ -351,6 +351,11 @@ def login(req):
     # this view, and the one below, is overridden because
     # we need to set the base template to use somewhere
     # somewhere that the login page can access it.
+
+    if settings.SERVER_ENVIRONMENT == 'icds':
```
nickpell: it might be slightly less gross to have a boolean variable in ```settings``` that gets used here, but that might be unnecessarily complicated.

nickpell (approved):

corehq/apps/hqwebapp/views.py
orangejenny: not a bad thought, though I think it's about equivalent, complexity-wise. i'm hoping this code will be the only thing that's strictly for icds, as opposed to the `enterprise_mode` setting.

snopoke: pretty sure you can make this use `settings.ENTERPRISE_MODE` for this. Currently ICDS is the only env with that set True but I imagine we want the same thing for future enterprise clusters.

orangejenny: But future clusters won't redirect specifically to `icds-cas`, so I'd also need some kind of server_environment/domain mapping, we'll be updating this code in some way for future clusters anyway.

PR #16076, opened 2017-05-09
Parse adherence dates properly

proteusvacuum:
https://sentry.io/dimagi/commcarehq/issues/269856877/
We store adherence_date as a date, not a datetime.

@NoahCarnahan 
cc: @benrudolph


NoahCarnahan (approved):
Looks good. Have you already checked for other places where we parse this case property?

PR #16079, opened 2017-05-09
don't save inferred_schema in loop

emord:
@snopoke I pulled this off of https://github.com/dimagi/commcare-hq/pull/15939. Seems safe to merge it before that

@benrudolph @dannyroberts


benrudolph:
👍 thanks

PR #16084, opened 2017-05-10
don't fail hard for missing date prop

snopoke:
https://sentry.io/dimagi/commcarehq/issues/269856877/


proteusvacuum:
I already fixed that Sentry error here: https://github.com/dimagi/commcare-hq/pull/16076. 

I'm going to make a slightly different change to the repeater for the missing date prop case (that you fixed here) as we don't actually want to send this to 99DOTS if there isn't a date property.

PR #16085, opened 2017-05-10
Dimension tables

benrudolph:
@gcapalbo this is my first pass at defining the dimension tables for the warehouse. have a couple of questions that i will comment about

fyi: @dimagi/scale-team cc: @proteusvacuum


corehq/warehouse/migrations/0002_domaindim_groupdim_locationdim_userdim_usergroupdim_userlocationdim.py
```diff
@@ -0,0 +1,134 @@
+# -*- coding: utf-8 -*-
+# Generated by Django 1.10.7 on 2017-05-10 10:21
+from __future__ import unicode_literals
+
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
```
benrudolph: have not added indexes yet, going to think about that in another pass

corehq/warehouse/models.py
```diff
@@ -62,3 +62,88 @@ def fail(self):
     )
     def decommission(self):
         return
+
+
+class BaseDim(models.Model):
+    domain = models.CharField(max_length=255)
```
benrudolph: i like the idea of always including the domain with each dim to ensure we never mix up data and make it easier to reason about. however it's probably not entirely necessary, let me know if you think we should remove

corehq/warehouse/models.py
```diff
@@ -62,3 +62,88 @@ def fail(self):
...
+class LocationDim(BaseDim):
+    location_id = models.CharField(max_length=100)
+    name = models.CharField(max_length=255)
+    site_code = models.CharField(max_length=255)
+    external_id = models.CharField(max_length=255)
+    supply_point_id = models.CharField(max_length=255, null=True)
+
+    location_type_id = models.CharField(max_length=255)
+    location_type_name = models.CharField(max_length=255)
+    location_type_code = models.CharField(max_length=255)
```
benrudolph: i've de-normalized the relationship with location type here

corehq/warehouse/models.py
```diff
@@ -62,3 +62,88 @@ def fail(self):
...
+    group_id = models.CharField(max_length=255)
+    name = models.CharField(max_length=255)
+
+    case_sharing = models.BooleanField()
+    reporting = models.BooleanField()
+
+    group_last_modified = models.DateTimeField()
+
+
+class LocationDim(BaseDim):
```
benrudolph: my biggest question for modeling the location dimension is how we should model the hierarchy. in the kimball guide, it talks about using a "Slightly ragged" hierarchy for models that have a short depth and are variable. it'd be great to talk through whether you think this is a good approach or not. my other concern is that we actually don't place any limits on how deep the location hierarchy can go. perhaps we can start enforcing that?

corehq/warehouse/models.py
```diff
@@ -62,3 +62,88 @@ def fail(self):
...
+    is_archived = models.BooleanField()
+
+    latitude = models.DecimalField(max_digits=20, decimal_places=10, null=True)
+    longitude = models.DecimalField(max_digits=20, decimal_places=10, null=True)
+
+    location_last_modified = models.DateTimeField()
+    location_created_on = models.DateTimeField()
+
+
+class DomainDim(BaseDim):
```
benrudolph: the domain model has a ton of attributes. i chose the ones that i thought were relevant. curious if you have any thoughts on how exhaustive the dimension table should be. i added ones that i thought might be useful for reports

benrudolph (commented):
i'm also planning on having the warehouse live in its own database `commcarehq_warehouse`, but that will also come in a later PR

corehq/warehouse/models.py
snopoke: This is definitely the biggest question I have. We're already experiencing quite a few scale issues with the current model but I think a simpler model is all we need: https://explainextended.com/2009/09/24/adjacency-list-vs-nested-sets-postgresql/

corehq/warehouse/models/dimensions.py
```diff
@@ -0,0 +1,86 @@
...
+    dim_last_modified = models.DateTimeField(auto_now=True)
+    dim_created_on = models.DateTimeField(auto_now_add=True)
+
+    class Meta:
+        abstract = True
+
+
+class UserDim(BaseDim):
+    user_id = models.CharField(max_length=255)
+    username = models.CharField(max_length=255)
```
emord: I think username, first and last name can all be reduced. https://docs.djangoproject.com/en/1.11/ref/contrib/auth/ (I think first and last may be getting an increase in 2.0)

Unsure about email though. maybe it's a text field in django? Either way we should keep in consistent

corehq/warehouse/models/dimensions.py
```diff
@@ -0,0 +1,86 @@
...
+    first_name = models.CharField(max_length=255)
+    last_name = models.CharField(max_length=255)
+    email = models.CharField(max_length=255)
+    user_type = models.CharField(max_length=255)
```
emord: If this is a choice between mobile and web, should it jsut be a boolean?

corehq/warehouse/models/dimensions.py
```diff
@@ -0,0 +1,86 @@
...
+    location_restriction_for_users = models.BooleanField()
+    use_sql_backend = models.BooleanField()
+    first_domain_for_user = models.BooleanField()
+
+    domain_last_modified = models.DateTimeField()
+    domain_created_on = models.DateTimeField()
+
+
+class UserLocationDim(BaseDim):
+    location_id = models.CharField(max_length=100)
```
emord: What's the reasoning behind not making htese foreign keys to location and user dims? All of these dimension tables will be replicated to all shards, so I think foreign keys would work right?

corehq/warehouse/models/dimensions.py
benrudolph: it's based on these: https://github.com/dimagi/commcare-hq/blob/master/corehq/pillows/utils.py#L23

corehq/warehouse/models/dimensions.py
benrudolph: yes, i think that should work. will update that

corehq/warehouse/models/dimensions.py
benrudolph: 👍 was taking from the `DjangoUserMixin` model which was a couch document

corehq/warehouse/models.py
gcapalbo: i think that's probably fine, it is an attribute on all of our source models

corehq/warehouse/models.py
gcapalbo: curious how come?

corehq/warehouse/models.py
gcapalbo: i think that's fine, we definitely only need the attributes that are worth filtering on / reporting on

corehq/warehouse/models/dimensions.py
```diff
@@ -0,0 +1,86 @@
...
+
+    latitude = models.DecimalField(max_digits=20, decimal_places=10, null=True)
+    longitude = models.DecimalField(max_digits=20, decimal_places=10, null=True)
+
+    location_last_modified = models.DateTimeField()
+    location_created_on = models.DateTimeField()
+
+
+class DomainDim(BaseDim):
+    domain_id = models.CharField(max_length=255)
```
gcapalbo: what's domain_id?

corehq/warehouse/models/dimensions.py
gcapalbo: yeah we should also make all of our joins use integer keys

corehq/warehouse/models.py
gcapalbo: yeah agree this is a big question. i think we also need to figure out what kind of filtering we're going to allow for location

corehq/warehouse/models.py
```diff
@@ -62,3 +62,88 @@ def fail(self):
     )
     def decommission(self):
         return
+
+
+class BaseDim(models.Model):
+    domain = models.CharField(max_length=255)
+
```
gcapalbo: we might want to add an identifier to track which batch a dim record came from, but we can do that later when we figure out our etl strategy

corehq/warehouse/models/dimensions.py
benrudolph: just the id of the couch doc

corehq/warehouse/models.py
benrudolph: the current filtering by location allows you to see your location + all those underneath you. we may also want to support just data at a certain location (not below it) though that is easy.

corehq/warehouse/models.py
benrudolph: for simplicity since it'll reduce the number of joins we have to do. can make it its own dim too

corehq/warehouse/models.py
benrudolph: 👍

corehq/warehouse/models.py
gcapalbo: i guess it's a trade off, by denormalizing it that means any time a location type's info changes we need to update every location that has that location type

benrudolph:
merging this into `br/warehouse`

PR #16087, opened 2017-05-10
Ce/loc upload

calellowitz:
@snopoke @sravfeyn @esoergel  cc: @millerdev 
context https://django-mptt.github.io/django-mptt/mptt.managers.html
location insertions update anywhere from 1 to all the other locations in a tree. this delays those updates to only occur every 100 insertions to speed up the bulk upload process


corehq/apps/commtrack/tasks.py
```diff
@@ -5,7 +5,7 @@
 from corehq.util.workbook_json.excel_importer import MultiExcelImporter
 from django.conf import settings
 
-LOCK_LOCATIONS_TIMEOUT = 60 * 60  # seconds
+LOCK_LOCATIONS_TIMEOUT = 60 * 60 * 10 # seconds
```
dimagimon: E261 at least two spaces before inline comment

corehq/apps/locations/tests/test_bulk_management.py
```diff
@@ -1144,3 +1184,17 @@ def test_partial_delete_children_invalid(self):
...
+            self.basic_tree
+        )
+        result = self.bulk_update_locations(
+            FLAT_LOCATION_TYPES,
+            BIG_LOCATION_TREE
+        )
+        self.assertEqual(result.errors, [])
+        self.assertLocationTypesMatch(FLAT_LOCATION_TYPES)
+        self.assertLocationsMatch(self.as_pairs(self.basic_tree + BIG_LOCATION_TREE))
+        self.assertCouchSync()
```
dimagimon: W292 no newline at end of file

corehq/apps/locations/views.py
```diff
@@ -818,18 +818,18 @@ def post(self, request, *args, **kwargs):
         domain = args[0]
 
         # stash this in soil to make it easier to pass to celery
-        ONE_HOUR = 1*60*60
+        TEN_HOURS = 10*60*60
```
dimagimon: E226 missing whitespace around arithmetic operator

corehq/apps/locations/bulk_management.py
```diff
@@ -870,31 +871,55 @@ def append_at_bottom(parent_type):
...
     for l in reversed(to_be_deleted):
         # Deletion has to happen bottom to top, otherwise mptt complains
         #   about missing parents
-        l.delete()
+        locs_to_delete.append(l)
+        if len(locs_to_delete) > chunk_size:
+            with transaction.atomic():
+                with SQLLocation.objects.delay_mptt_updates():
+                    for l in locs_to_delete:
+                        l.delete()
```
millerdev: Too bad we have to run a query per object. Would probably be a lot more efficient to run a query like `SQLLocation.objects.filter(id__in=[loc.id for loc in locs_to_delete]).delete()`. Does mptt know how to handle queries like that? Could probably do all deletes with a single query instead of using `chunk_size` if it does.

corehq/apps/locations/bulk_management.py
```diff
@@ -870,31 +871,55 @@ def append_at_bottom(parent_type):
...
+    locs_to_process = []
+    for location in top_to_bottom_locations:
+        locs_to_process.append(location)
+        if len(locs_to_process) > chunk_size:
+            with transaction.atomic():
+                with SQLLocation.objects.delay_mptt_updates():
+                    _process_locations(locs_to_process)
+            locs_to_process = []
+    if locs_to_process:
+        _process_locations(locs_to_process)
```
millerdev: Should this be inside
```py
with transaction.atomic():
    with SQLLocation.objects.delay_mptt_updates():
```
?

calellowitz:
also tagging @gcapalbo because of the work he did here recently

corehq/apps/locations/bulk_management.py
snopoke: this should also be inside an atomic block and a delay mptt updates block

corehq/apps/locations/bulk_management.py
```diff
@@ -870,31 +871,55 @@ def append_at_bottom(parent_type):
...
-                loc_object.parent = SQLLocation.objects.get(
-                    domain=domain,
-                    site_code__iexact=loc.parent_code
-                )
-            else:
-                loc_object.parent = None
-            loc.db_object.save()
-
+    locs_to_process = []
+    for location in top_to_bottom_locations:
```
snopoke: could use `dimagi.utils.chunked` which would simplify the code a bit

snopoke:
@calellowitz can you confirm that we don't rely on intermediate queries to the tree to during the import. i.e. confirm that this isn't a problem: `NOTE that this context manager causes inconsistencies! MPTT model methods are not guaranteed to return the correct results until the end of the context block.`

snopoke:
Side note: In the long term should we consider switching to adjacency lists to represent the location hierarchy?

* https://explainextended.com/2009/09/24/adjacency-list-vs-nested-sets-postgresql/
* https://django-treebeard.readthedocs.io/en/latest/al_tree.html

corehq/apps/locations/bulk_management.py
```diff
@@ -870,31 +872,55 @@ def append_at_bottom(parent_type):
...
+        locs_to_process.append(location)
+        if len(locs_to_process) > chunk_size:
+            with transaction.atomic():
+                with SQLLocation.objects.delay_mptt_updates():
+                    _process_locations(locs_to_process)
```
gcapalbo: i don't think we should wrap this in transaction.atomic() for the reasons i outlined here: https://github.com/dimagi/commcare-hq/pull/16034

corehq/apps/locations/bulk_management.py
```diff
@@ -870,31 +872,55 @@ def append_at_bottom(parent_type):
...
+        _process_locations(locs_to_process)
+
+    locs_to_delete = []
     for l in reversed(to_be_deleted):
         # Deletion has to happen bottom to top, otherwise mptt complains
         #   about missing parents
-        l.delete()
+        locs_to_delete.append(l)
+        if len(locs_to_delete) > chunk_size:
+            with transaction.atomic():
```
gcapalbo: same comment about wrapping these operations in a transaction

corehq/apps/locations/bulk_management.py
gcapalbo: i don't think we have utils to handle bulk operations on locations, since updating/deleting a location also requires syncing to the couch location, the supply point, and publishing to kafka which right now all happen within save()

gcapalbo:
(hit the wrong button, didn't mean to close)

esoergel:
> i've always been curious why we need the mptt model - do we ever query the location hierarchy for anything more than what a simple parent foreign key would give us?

@gcapalbo the parent key relationship does store all the relationship information we need - and the MPTT stuff can be regenerate based on parent keys alone (`SQLLocation.objects.rebuild()`).  We do do a number of queries that extend further than that, though.  For instance, we do things like `location.get_descendants()`, `location.get_ancestors()`, and `SQLLocation.objects.get_queryset_descendants()`. Some of the more complex queries can be found on `LocationManager` and `LocationQueriesMixin`.  We might be able to reproduce that functionality using some sort of recursive set of SQL queries, but I'm not sure.

Another approach would be to store the complete ancestry along with each location - I'm pretty confident that would allow us to meet our querying needs without MPTT.  The downside is that we'd then have to roll our own consistency model to make sure to update all descendant locations if parentage changes.

esoergel:
@calellowitz when @sravfeyn was first implementing this he looked in to using `delay_mptt_updates`, but decided it didn't work for some reason or another.  I'm not sure why that was, and it does seem like the best approach, but I'd want his input if possible.

calellowitz:
ran into some issues getting one test to pass which i will look more into tomorrow, but to answer @snopoke question about that note, the atomic transaction is supposed to ensure that, since nothing is committed to the database until the rebuild at the exit of the context manager. we also lock most location transactions during upload (and this pr extends the timeout on that lock since it was much shorter than many of the recent uploads).

corehq/apps/locations/bulk_management.py
calellowitz: this is needed because `delay_mptt_updates` leaves the locations in an inconsistent state while it is modifying the table. I chunked those transactions in batches of 100 which fixes the progress bar issue (at least to some extent), and the transaction is after updates to task meta so it shouldnt mess with celery, but this does address your concerns about rollback leaving extra couch locations. Not sure how big a deal that is, since it can only leave a few (less than 100) but will defer to @esoergel or you if you think that is a dealbreaker

esoergel:
> Side note: In the long term should we consider switching to adjacency lists to represent the location hierarchy?
>
>    https://explainextended.com/2009/09/24/adjacency-list-vs-nested-sets-postgresql/
>    https://django-treebeard.readthedocs.io/en/latest/al_tree.html

@snopoke That first article is great, thanks for sharing.  I also briefly skimmed the treebeard [api docs](https://django-treebeard.readthedocs.io/en/latest/api.html) and it looks like it has what we need, including most of the same methods provided by [MPTT](https://django-mptt.github.io/django-mptt/mptt.models.html).  That certainly seems like it'd work if it's faster for our typical operations.  The most visible pain point right now is bulk upload, so if it dramatically speeds up writes, I could see that being worth it.  My preferred plan of approach would be to see where we get after dropping couch (I'm picking this back up now) and properly delaying mptt updates first, then if needed do some performance testing on MPTT vs treebeard with typical use cases.

As far as `delay_mptt_updates` goes - I recently ran a full rebuild on each of our environments, and it was on the order of a couple of minutes for prod and india.  This suggests to me that if we can get the import down to: normal db writes and bookkeeping, followed by tree model stuff (for just the tree_ids in the domain), the tree model stuff wouldn't be a big blocker.

corehq/apps/locations/bulk_management.py
gcapalbo: i see - yeah i was less worried about the progress bar, but more concerned about rolling back and leaving couch locations and supply point cases in an inconsistent state with the sql locations. but i'll defer to @esoergel on that

corehq/apps/locations/bulk_management.py
```diff
@@ -276,6 +279,11 @@ def _needs_save(self):
 
         return False
 
+    def _moved_to_root(self):
+        if not self.is_new and self.parent_code == ROOT_LOCATION_TYPE and self.db_object.parent.site_code !=  ROOT_LOCATION_TYPE:
```
dimagimon: E222 multiple spaces after operator
E501 line too long (129 > 115 characters)

corehq/apps/locations/tests/test_bulk_management.py
```diff
@@ -1144,3 +1184,43 @@ def test_partial_delete_children_invalid(self):
...
+
+    def test_new_root(self):
+        # new locations can be added without having to specify all of old ones
+        self.bulk_update_locations(
+            FLAT_LOCATION_TYPES,
+            self.basic_tree
+        )
+
+        upload = [
+             ('S1', 's1', 'city', 'county11', '', False) + extra_stub_args,
```
dimagimon: E126 continuation line over-indented for hanging indent

corehq/apps/locations/tests/test_bulk_management.py
```diff
@@ -1144,3 +1184,43 @@ def test_partial_delete_children_invalid(self):
...
+        ]
+
+        result = self.bulk_update_locations(
+            FLAT_LOCATION_TYPES,
+            upload
+        )
+        self.assertEqual(result.errors, [])
+        self.assertLocationTypesMatch(FLAT_LOCATION_TYPES)
+        self.assertLocationsMatch(self.as_pairs(upload))
+        self.assertCouchSync()
```
dimagimon: W292 no newline at end of file

corehq/apps/locations/bulk_management.py
```diff
@@ -276,6 +279,12 @@ def _needs_save(self):
 
         return False
 
+    def _moved_to_root(self):
+        old_parent = self.db_object.parent.site_code if self.db_object.parent else ROOT_LOCATION_TYPE
+        if not self.is_new and self.parent_code == ROOT_LOCATION_TYPE and old_parent !=  ROOT_LOCATION_TYPE:
```
dimagimon: E222 multiple spaces after operator

calellowitz:
@snopoke @millerdev I addressed the missing atomic transaction and cleaned up the code a bit. I also added a few exceptions in which uploads will not delay mptt updates because it was leaving the tree in an inconsistent state that a partial rebuild couldnt fix. In particular, it left two root nodes with the same tree_id, so i exempted uploads in which the location types change and ones in which an existing location is moved to the root level. That blocks more than is strictly necessary but the edge cases on when the tree is left inconsistent were complex enough that a simple heuristic seemed better, and they both seem like unusual cases.

corehq/apps/locations/bulk_management.py
```diff
@@ -421,7 +431,11 @@ def bulk_commit(self, type_stubs, location_stubs):
             loc.lookup_old_collection_data(self.old_collection)
 
         type_objects = save_types(type_stubs, self.excel_importer)
-        save_locations(location_stubs, type_objects, self.domain, self.excel_importer)
+        types_changed = any(loc_type.needs_save for loc_type in type_stubs)
+        moved_to_root = any(loc.moved_to_root for loc in location_stubs)
+        delay_updates = not (types_changed or moved_to_root)
```
millerdev: Does this PR make any improvements for situations where this expression evaluates to false? If not, how often does that happen? If often, maybe we should look harder at moving to adjacency list + recursive queries? In case you haven't read the articles linked by @snopoke, we do not need to change the data model to do that; we could skip all the MPTT updates and just put bogus values in `lft` and `rght`.

millerdev (commented):
I'm not sure if @gcapalbo's comment was answered:

> concerned about rolling back and leaving couch locations and supply point cases in an inconsistent state with the sql locations. but i'll defer to @esoergel on that

Other than that I'm 👍 on this for now. Seems like it would be a big win to move to adjacency list + recursive queries in the longer term.

corehq/apps/locations/bulk_management.py
calellowitz: it makes no improvents for those situations but i think they are quite rare. one if them is changes to the types hierarchy, which is easy enough to do from the ui since few people have more than a 10 if any, and if we warn people to make those changes first they easily can. the second exemption is a location moving from its current type to the root type, and it seems unlikely people will do that since locations should rarely change types, but again if that is necessary, they can move that type, then proceed with the rest of the upload.

calellowitz:
@millerdev i was waiting on @esoergel to answer that since he seems best positioned to know the impact of the extra couch locations or the supply point cases (he may have said this on a fogbugz ticket but will let him repeat it here). this pr does limit those to less than 100 because of the chunk size but giovanni is correct that some can be made

esoergel:
Here's what I wrote on fogbugz:
> The extra couch locations shouldn't cause any problems, as SQLLocations are the source of truth and there are no remaining couch location views, they are only retrieved by id (usually pulled from SQL).  The extra supply points might show up in reports, I'm not sure.  Those are only relevant for Supply domains, though.  Still, it would be good to clean up both of those things.

@calellowitz would it be possible to do something like the pseudocode:

```python
with transaction:
    supply_points_created = []
    for location in locations:
        location.save()
        if location.supply_point_id and supply_point_didnt_already_exist:
            supply_points_created.append(location.supply_point_id)
if transaction aborted:
    delete_cases(supply_points_created)
```

calellowitz:
@esoergel I looked into this a bit and django does not have any sort of transaction aborted handler, which will make doing that kind of cleanup tricky. We could add another clause to the if statement so that it does not use the delayed updates for commtrack domains (or at least non administrative locations), but i dont know if we are starting to defeat the purpose at that point and we still have the couch locations lying around.

esoergel:
@calellowitz I guess that's okay. This shouldn't happen anyways, so I suppose we could always do manual clean-up.

As an aside, django does have `transaction.on_commit()`, which let's you specify things to be called only when a transaction completes successfully, so in theory that could be used to set a state variable which is then checked to determine whether or not the transaction succeeded.  This does mean there'd have to be exception handling and all that, so I could see it getting out of hand.

calellowitz:
@esoergel yeah, that seemed a little nasty and was not at all the recommended way of using that hook, and it was unclear how robust an approach it is though it seems like it should be.

more generally, this pr seems like a big win and im wondering if you or @gcapalbo need anything else from me before we can push this out, because i would hate to see it stall in pr forever

esoergel (approved):

corehq/apps/locations/bulk_management.py
```diff
@@ -870,31 +884,43 @@ def append_at_bottom(parent_type):
...
-            else:
-                loc_object.parent = None
-            loc.db_object.save()
-
-    for l in reversed(to_be_deleted):
+    if delay_updates:
+        for locs in chunked(top_to_bottom_locations, chunk_size):
+            with transaction.atomic():
+                with SQLLocation.objects.delay_mptt_updates():
+                    _process_locations(locs, to_be_deleted)
```
sravfeyn: I am (gladly) surprised that this is working. When I tried this originally, I ran into test case failures involving mptt node inconsistencies.

PR #16089, opened 2017-05-10
fix ucr named expression cache

emord:
@kkrampa Thanks for catching this and for the test! I cherry-picked your commit from #16062 

Named expressions use the iteration cache and that cache is reset on every iteration through a doc, so I'm not quite sure this is a bug. Could you link me to a fogbugz case or the UCR you found this on?

Either way, it seems like having a the item as part of the cache is an improvement so went ahead and implemented it.

@dannyroberts

### DIFF
--- a/corehq/apps/userreports/expressions/specs.py
+++ b/corehq/apps/userreports/expressions/specs.py
@@ -1,4 +1,7 @@
 from __future__ import absolute_import
+import hashlib
+import json
+
 from jsonobject.base_properties import DefaultProperty
 from simpleeval import InvalidExpression
 import six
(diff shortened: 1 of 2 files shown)


kkrampa:
http://manage.dimagi.com/default.asp?253776#1329041

Those two indicators are always equal: 
https://github.com/dimagi/commcare-hq/blob/master/custom/enikshay/ucr/data_sources/episode.json#L1020L1036
https://github.com/dimagi/commcare-hq/blob/master/custom/enikshay/ucr/data_sources/episode.json#L1071L1087

kkrampa:
@emord Not sure if it's important but we don't use iterator in episode case at all.

emord:
Thanks @kkrampa i see the issue now. This should be fixed by my change. it will be deployed today

corehq/apps/userreports/expressions/specs.py
```diff
@@ -79,11 +83,12 @@ def configure(self, context):
             raise BadSpecError(u'Name {} not found in list of named expressions!'.format(self.name))
         self._context = context
 
-    def _context_cache_key(self):
-        return 'named_expression-{}'.format(self.name)
+    def _context_cache_key(self, item):
+        item_hash = hashlib.md5(json.dumps(item, cls=DjangoJSONEncoder, sort_keys=True)).hexdigest()
```
snopoke: @emord why is it necessary to hash the full object here. Can we not just use the ID? for cases where the item is a repeat or something like that it may not have an ID so the we could use the hash. Also wondering if we can't use the object ID `id(item)` rather than generating a hash?

Reason for asking is that this eagerly evaluates any lazy fields in the item.

emord: That seems reasonable to me. Its likely not necessary to hash the whole object, just what I thought to do at the time. as long as tests pass with `id` i'm happy with the change

PR #16090, opened 2017-05-10
using djangos json encoder

emord:
@dannyroberts found this when fixing #16089 Seems like a better solution as it's more robust (can handle datetimes)


dannyroberts (approved):
Cool, nice

PR #16097, opened 2017-05-10
Refactor repeater error handling

dannyroberts:
@proteusvacuum 

This lightly simplifies some of the error handling around `get_payload`, and in the process also gets rid of the special treatment of `ResourceNotFound`—I looked in email and this has happened once, in 2015, so I think it's fine not to treat it any special way at this point if it'll get in the way of reworking this code to make way for upcoming changes.

Good to review "🐡"

### DIFF
--- a/corehq/apps/repeaters/models.py
+++ b/corehq/apps/repeaters/models.py
@@ -168,7 +168,13 @@ class Repeater(QuickCachedDocumentMixin, Document, UnicodeMixIn):
         generator = self.get_payload_generator(self.format_or_default_format())
         return generator.get_payload(repeat_record, self.payload_doc(repeat_record))
 
-    def get_payload_and_handle_exception(self, repeat_record, save_failure=True):
+    def get_payload_or_none(self, repeat_record):
+        try:
+            return self.get_payload(repeat_record)
+        except ResourceNotFound:
+            return None
+
+    def get_payload_and_save_exception(self, repeat_record):
         try:
             return self.get_payload(repeat_record)
         except ResourceNotFound as e:
@@ -179,11 +185,9 @@ class Repeater(QuickCachedDocumentMixin, Document, UnicodeMixIn):
                     repeat_record._id, self.domain,
                 ))
 
-            if save_failure:
-                repeat_record.handle_payload_exception(e)
+            repeat_record.handle_payload_exception(e)
         except Exception as e:
-            if save_failure:
-                repeat_record.handle_payload_exception(e)
+            repeat_record.handle_payload_exception(e)
             raise
 
     def register(self, payload, next_check=None):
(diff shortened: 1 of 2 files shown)


corehq/apps/repeaters/models.py
```diff
@@ -551,12 +541,13 @@ def try_now(self):
...
-        warnings.warn("RepeatRecord.get_payload is deprecated. Use Repeater.get_payload_and_handle_exception.",
-                      DeprecationWarning)
-        return self.repeater.get_payload_and_handle_exception(self, save_failure=save_failure)
+    def get_payload(self):
+        return self.repeater.get_payload(self)
 
     def handle_payload_exception(self, exception):
+        # todo: this seems to just say that it failed but not actually cancel it
+        # todo:   or even change the last_checked or next_check
+        # todo:   so this will just get requeued indefinitely, right?
```
dannyroberts: ...and here

corehq/apps/repeaters/models.py
```diff
@@ -301,7 +283,15 @@ def get_auth(self):
     def fire_for_record(self, repeat_record):
         headers = self.get_headers(repeat_record)
         auth = self.get_auth()
-        payload = self.get_payload_and_handle_exception(repeat_record)
+
+        try:
+            payload = self.get_payload(repeat_record)
+        except Exception as e:
+            # todo: seems like this just does everything that handle_exception does but not as well.
+            # todo:   seems like they could be combined
```
dannyroberts: curious if you have thoughts here...

corehq/apps/repeaters/models.py
dannyroberts: Not indefinitely, sorry, 3x. `self.overall_tries += 1` a few lines below.

corehq/apps/repeaters/models.py
proteusvacuum: Yeah, I think previously the `handle_payload_exception` was also calling `raise`. It might not have gotten updated when we added the `cancelled` state.

corehq/apps/repeaters/models.py
proteusvacuum: we could just immediately cancel it I suppose.

dannyroberts: 👍

corehq/apps/repeaters/models.py
dannyroberts: OK, I'm not sure I follow (this calls `raise` right below `handle_payload_exception`, maybe there was a typo in what you wrote or I'm misunderstanding). I'm also leaning towards leaving this like this with the todo comment and addressing later, since it feels like anything that addresses it is likely to (possibly significantly) change behavior. I imagine this `except` block is mostly here to catch errors in custom repeaters' get_payload call, which may be more complicated than the built-ins which just fetch a case/form

proteusvacuum: You recently changed the behaviour of that method: https://github.com/dimagi/commcare-hq/pull/16059/commits/96539922436a57d1ec137cb93480b0260e41bde6#diff-7a94e70a3056d09745068e953a4847f6L558
It used to reraise the exception based on a flag, which the other handle_exception method didn't do. 

the `handle_payload_exception` method also didn't bubble up to call any `repeater` (or `repeater_generator`) handling methods. 

You are correct about the intention of this code to catch errors in the `get_payload` calls.

PR #16105, opened 2017-05-11
do not save blobdb to obj

benrudolph:
So we can serialize the obj

cc: @proteusvacuum


proteusvacuum:
I think you can also set this as a `@property` to allow serializing. This works too though.

benrudolph:
thanks, that's good to know

PR #16106, opened 2017-05-11
Show Historical Adherence report in web user permissions

NoahCarnahan:
@proteusvacuum 
buddy @sravfeyn 
This needs to be deployed today

### DIFF
--- a/corehq/apps/reports/generic.py
+++ b/corehq/apps/reports/generic.py
@@ -695,9 +695,9 @@ class GenericReportView(object):
         return True
 
     @classmethod
-    def show_in_webuser_permissions(cls, domain=None, project=None, user=None):
+    def show_in_user_roles(cls, domain=None, project=None, user=None):
         """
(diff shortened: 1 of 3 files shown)


corehq/apps/reports/generic.py
```diff
@@ -695,6 +695,14 @@ def show_in_navigation(cls, domain=None, project=None, user=None):
         return True
 
     @classmethod
+    def show_in_webuser_permissions(cls, domain=None, project=None, user=None):
```
proteusvacuum: Does this show in that modal when you create or edit a "Role"? If so, these are more than just webuser permissions, as they can also be used for mobile workers (confusing, because it is on the same page as webusers).
I'd rename this to `show_in_user_roles` or something similar.

NoahCarnahan: updated, thanks

proteusvacuum (approved):

PR #16109, opened 2017-05-11
app manager v2: Float case properties refresh button right

orangejenny:
Looks a little awkward still on the usercase screen, but less so than when it was on the left.

screen shot 2017-05-11 at 11 54 42 am

screen shot 2017-05-11 at 11 54 49 am

@biyeun / @nickpell


nickpell (approved):

nickpell:
(kicked build)

PR #16111, opened 2017-05-11
add Subscription.get_active_subscription_by_domain()

nickpell:
This is the second half to https://github.com/dimagi/commcare-hq/pull/15852 (wanted to wait one invoicing period to make sure nothing unexpectedly broke).

Since we can now know that there is at most one active subscription per domain, we're able to clean up the query to find a domain's active subscription (https://github.com/dimagi/commcare-hq/commit/39188fd4f0b3e45922649b67365bfb1485981b42).

@orangejenny

### DIFF
--- a/corehq/apps/accounting/models.py
+++ b/corehq/apps/accounting/models.py
@@ -1558,43 +1558,34 @@ class Subscription(models.Model):
             self.account.save()
 
     @classmethod
-    def _get_plan_by_subscriber(cls, subscriber):
-        active_subscriptions = cls.objects\
-            .filter(subscriber=subscriber, is_active=True)\
-            .order_by('-date_created')[:2]\
-            .select_related('plan_version__role')
-
-        if not active_subscriptions:
-            return None, None
-
-        if len(active_subscriptions) > 1:
-            log_accounting_error(
-                "There seem to be multiple ACTIVE subscriptions for the "
-                "subscriber %s. Odd, right? The latest one by "
-                "date_created was used, but consider this an issue."
-                % subscriber
+    def get_active_subscription_by_domain(cls, domain_name):
+        try:
+            return cls.objects.select_related(
+                'plan_version__role'
+            ).get(
+                is_active=True,
+                subscriber__domain=domain_name,
             )
-        current_subscription = active_subscriptions[0]
-        return current_subscription.plan_version, current_subscription
+        except cls.DoesNotExist:
+            return None
 
     @classmethod
     def get_subscribed_plan_by_domain(cls, domain):
         """
-        Returns SoftwarePlanVersion, Subscription for the given domain.
+        Returns SoftwarePlanVersion for the given domain.
         """
         domain_obj = ensure_domain_instance(domain)
         if domain_obj is None:
             try:
-                plan_version = DefaultProductPlan.get_default_plan_version()
-                return plan_version, None
+                return DefaultProductPlan.get_default_plan_version()
             except DefaultProductPlan.DoesNotExist:
                 raise ProductPlanNotFoundError
-        domain = domain_obj
-        subscriber = Subscriber.objects.safe_get(domain=domain.name)
-        plan_version, subscription = cls._get_plan_by_subscriber(subscriber) if subscriber else (None, None)
-        if plan_version is None:
-            plan_version = DefaultProductPlan.get_default_plan_version()
-        return plan_version, subscription
+        else:
+            active_subscription = cls.get_active_subscription_by_domain(domain_obj.name)
+            if active_subscription is not None:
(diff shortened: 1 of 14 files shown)


corehq/apps/accounting/models.py
```diff
@@ -1558,43 +1558,34 @@ def set_billing_account_entry_point(self):
...
+                return DefaultProductPlan.get_default_plan_version()
             except DefaultProductPlan.DoesNotExist:
                 raise ProductPlanNotFoundError
-        domain = domain_obj
-        subscriber = Subscriber.objects.safe_get(domain=domain.name)
-        plan_version, subscription = cls._get_plan_by_subscriber(subscriber) if subscriber else (None, None)
-        if plan_version is None:
-            plan_version = DefaultProductPlan.get_default_plan_version()
-        return plan_version, subscription
+        else:
```
orangejenny: nitpick, but this doesn't need to be inside an `else`

corehq/apps/accounting/models.py
```diff
@@ -1558,43 +1558,34 @@ def set_billing_account_entry_point(self):
...
+            active_subscription = cls.get_active_subscription_by_domain(domain_obj.name)
+            if active_subscription is not None:
+                return active_subscription.plan_version
+            else:
```
orangejenny: same here

corehq/apps/api/domain_metadata.py
```diff
@@ -47,15 +47,13 @@ def dispatch(self, request_type, request, **kwargs):
...
-            Subscription.get_subscribed_plan_by_domain(domain)
-        )
+        subscription = Subscription.get_active_subscription_by_domain(domain.name)
         return {
             "date_start": (subscription.date_start
                            if subscription is not None else None),
             "date_end": (subscription.date_end
                          if subscription is not None else None),
-            "plan_version": plan_version,
+            "plan_version": subscription.plan_version,
```
orangejenny: is it guaranteed that `subscription` won't be `None` here?

corehq/apps/domain/views.py
```diff
@@ -677,7 +677,8 @@ def can_purchase_credits(self):
     @property
     @memoized
     def plan(self):
-        plan_version, subscription = Subscription.get_subscribed_plan_by_domain(self.domain_object)
+        subscription = Subscription.get_active_subscription_by_domain(self.domain)
+        plan_version = subscription.plan_version
```
orangejenny: This should probably check for `subscription` being `None` (there are similar checks in this function)

corehq/apps/accounting/models.py
```diff
@@ -1558,43 +1558,34 @@ def set_billing_account_entry_point(self):
...
-        current_subscription = active_subscriptions[0]
-        return current_subscription.plan_version, current_subscription
+        except cls.DoesNotExist:
+            return None
 
     @classmethod
     def get_subscribed_plan_by_domain(cls, domain):
         """
-        Returns SoftwarePlanVersion, Subscription for the given domain.
+        Returns SoftwarePlanVersion for the given domain.
```
orangejenny: so much nicer!

corehq/apps/userreports/views.py
```diff
@@ -301,8 +301,7 @@ def page_context(self):
     @property
     @memoized
     def plan_name(self):
-        plan_version, _ = Subscription.get_subscribed_plan_by_domain(self.domain)
-        return plan_version.plan.name
+        return Subscription.get_subscribed_plan_by_domain(self.domain).plan.name
```
orangejenny: should this check for `None`?

orangejenny (changes requested):
There are a couple of places where it looks like you should check for `None`. Otherwise looks good.

corehq/apps/api/domain_metadata.py
nickpell: updated

corehq/apps/domain/views.py
nickpell: updated

corehq/apps/userreports/views.py
nickpell: ```get_subscribed_plan_by_domain``` won't return ```None```, so this one is all set

nickpell:
updated ```None``` checks

orangejenny:
New code looks good, but test failure looks real:
```
File "/mnt/commcare-hq-ro/corehq/apps/users/util.py", line 175, in can_add_extra_mobile_workers
    user_limit = request.plan.user_limit
AttributeError: 'WSGIRequest' object has no attribute 'plan'```

nickpell:
✅

orangejenny (approved):

PR #16114, opened 2017-05-11
Repeater generator callbacks

dannyroberts:
Pulling this out of a longer refactor since it's pretty self contained.

Just reducing verbosity and indirection, but does exactly the same thing.

### DIFF
--- a/corehq/apps/repeaters/models.py
+++ b/corehq/apps/repeaters/models.py
@@ -315,25 +315,13 @@ class Repeater(QuickCachedDocumentMixin, Document, UnicodeMixIn):
         """
         if isinstance(result, Exception):
             repeat_record.handle_exception(result)
-            self.handle_exception(result, repeat_record)
+            self.generator.handle_exception(result, repeat_record)
         elif 200 <= result.status_code < 300:
             repeat_record.handle_success(result)
-            self.handle_success(result, repeat_record)
+            self.generator.handle_success(result, self.payload_doc(repeat_record), repeat_record)
         else:
             repeat_record.handle_failure(result)
-            self.handle_failure(result, repeat_record)
-
-    def handle_success(self, response, repeat_record):
-        """handle a successful post"""
-        return self.generator.handle_success(response, self.payload_doc(repeat_record), repeat_record)
-
-    def handle_failure(self, response, repeat_record):
-        """handle a failed post"""
-        return self.generator.handle_failure(response, self.payload_doc(repeat_record), repeat_record)
-
-    def handle_exception(self, exception, repeat_record):
-        """handle an exception during a post"""
-        return self.generator.handle_exception(exception, repeat_record)
+            self.generator.handle_failure(result, self.payload_doc(repeat_record), repeat_record)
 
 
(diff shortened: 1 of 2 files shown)


corehq/apps/repeaters/models.py
```diff
@@ -163,12 +163,16 @@ def get_payload_generator(self, payload_format):
         gen = RegisterGenerator.generator_class_by_repeater_format(self.__class__, payload_format)
         return gen(self)
 
+    @property
+    @memoized
```
dannyroberts: I said it does "exactly the same thing". The only really unlikely thing I can think of that's different is that with this memoized property, every time we want to use _a_ payload generator, we'll now be using _the same_ one (as in the same instance). I can't imagine that being a problem; if it is, we can easily remove the `@memoized` decorator to fix that.

dannyroberts:
@emord buddy

corehq/apps/repeaters/models.py
```diff
@@ -313,31 +315,13 @@ def handle_response(self, result, repeat_record):
         """
         if isinstance(result, Exception):
             repeat_record.handle_exception(result)
-            self.handle_exception(result, repeat_record)
+            self.generator.handle_exception(result, repeat_record)
```
proteusvacuum: This might be a conversation to have, but does it make more sense to have these on the `Repeater` or the `PayloadGenerator`? I think I maybe made the wrong decision about handling this in the payload generators, since those should actually just generate the payload.
Instead, if these were handled by the Repeater, we could have some series of `RepeaterCallback` classes that could encapsulate handling the responses (e.g. updating cases, or whatever else)

dannyroberts: Yeah, I think it's not super clear where the right place is to put these callbacks, but on the repeater kind of makes more sense to me too. While we're on the topic, this thing of linking repeater payload generators to repeaters through a decorator is totally insane. (I probably wrote it). It'd make a lot more sense for the repeater to have `payload_generators` dict that mapped formats to payload generators.

Anyway, I think all of that can wait for another PR, but I think that's a good suggestion.

proteusvacuum: Yes, please get rid of the decorator. please.

dannyroberts:
(Does that mean you approve of this PR pending build? :))

proteusvacuum:
I guess so. Will you plan on moving it to the repeater at a later date?

dannyroberts:
Oh, yes totally! I'm going to work on isolating state changes into `RepeatRecordAttempt` objects first after this, but reorganizing the callbacks, payload generators, and this decorator business is definitely on the radar

corehq/apps/repeaters/models.py
snopoke: I think the decorator was my fault :disappointed:

dannyroberts: :) well if it was, I've certainly also used that pattern before in other code then, because it seems very familiar...

snopoke:
Yea, I think I copied from some of your code in app manager...

PR #16116, opened 2017-05-11
Create 24 Hour Registration reminder email

biyeun:
@orangejenny // cc: @esoergel 

Per request from comms to resend the confirmation reminder email after 24 hours and no activation has been made.

### DIFF
--- a/corehq/apps/registration/models.py
+++ b/corehq/apps/registration/models.py
@@ -42,6 +42,27 @@ class RegistrationRequest(Document):
         return result[0]['value']
 
     @classmethod
+    def get_requests_24hrs_ago(cls):
+        today = datetime.datetime.utcnow()
+        yesterday = today - datetime.timedelta(1)
+        join_on_start = datetime.datetime(
+            yesterday.year, yesterday.month, yesterday.day, yesterday.hour, 0, 0, 0)
+        join_on_end = datetime.datetime(
+            yesterday.year, yesterday.month, yesterday.day, yesterday.hour, 59, 59, 59)
+        result = cls.view(
+            "registration/requests_by_time",
+            startkey=join_on_start.isoformat(),
+            endkey=join_on_end.isoformat(),
+            include_docs=True,
+            reduce=False
+        ).all()
+        return filter(
+            lambda doc: (doc.new_user_username == doc.requesting_user_username
+                         and doc.confirm_time is None),
+            result
+        )
+
+    @classmethod
     def get_request_for_username(cls, username):
         result = cls.view("registration/requests_by_username",
             key=username,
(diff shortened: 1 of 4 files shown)


corehq/apps/registration/templates/registration/email/confirm_account_reminder.txt
```diff
@@ -0,0 +1,23 @@
+Dear {{ full_name }},
```
orangejenny: Shouldn't this file also be translated?

corehq/apps/registration/tasks.py
```diff
@@ -0,0 +1,50 @@
...
+
+    for request in request_reminders:
+        user = WebUser.get_by_username(request.new_user_username)
+        registration_link = 'http://' + DNS_name + reverse(
+            'registration_confirm_domain') + request.activation_guid + '/'
+        email_context = {
+            "domain": request.domain,
+            "registration_link": registration_link,
+            "full_name": user.full_name,
+            'url_prefix': '' if settings.STATIC_CDN else 'http://' + DNS_name,
```
orangejenny: Doesn't look like this is used in the email templates.

corehq/apps/registration/models.py
```diff
@@ -42,6 +42,27 @@ def get_requests_today(cls):
         return result[0]['value']
 
     @classmethod
+    def get_requests_24hrs_ago(cls):
+        today = datetime.datetime.utcnow()
+        yesterday = today - datetime.timedelta(1)
+        join_on_start = datetime.datetime(
+            yesterday.year, yesterday.month, yesterday.day, yesterday.hour, 0, 0, 0)
+        join_on_end = datetime.datetime(
+            yesterday.year, yesterday.month, yesterday.day, yesterday.hour, 59, 59, 59)
```
orangejenny: Nitpicky, but you might just leave off the microseconds or else use 999999 instead of 59

orangejenny (approved):
A couple of comments, but nothing blocking.

corehq/apps/registration/tasks.py
biyeun: it is actually. in the parent template

corehq/apps/registration/templates/registration/email/confirm_account_reminder.txt
biyeun: how will they pick the language to translate to, tbh?

biyeun: the original confirm account email lacks translation too but in all practical terms how do we determine what language to send the confirm account email in?

orangejenny: Yeah, good point. I imagine when they sign up we could store whether they're looking at prelogin in English or French...which would be kind of nice, actually...but doubt we're doing that now.

corehq/apps/registration/tasks.py
orangejenny: ah, I see, didn't notice

esoergel (approved):

PR #16117, opened 2017-05-11
Nikshay login as

benrudolph:
This implements the functionality necessary to be able to query login as users by phone number and custom user data which is needed for enikshay. separate PR coming that deals with just updating the user es index
by 🏠 
@proteusvacuum 

fyi: @ctsims

### DIFF
--- a/corehq/apps/cloudcare/esaccessors.py
+++ b/corehq/apps/cloudcare/esaccessors.py
@@ -1,5 +1,3 @@
-from corehq import toggles
-
 from corehq.apps.es import filters, queries, UserES
 from corehq.apps.locations.models import SQLLocation
 
--- a/corehq/apps/cloudcare/esaccessors.py
+++ b/corehq/apps/cloudcare/esaccessors.py
@@ -10,6 +10,22 @@ def login_as_user_query(
         offset,
         can_access_all_locations=False,
         user_data_fields=None):
+    '''
+    Takes in various parameters to determine which users to populate the login as screen.
+
+    :param domain: String domain
+    :param couch_user: The CouchUser that is using the Login As feature
+    :param search_string: The query that filters the users returned. Filters based on the
+        `search_fields` as well as any fields defined in `user_data_fields`.
+    :param limit: The max amount of users returned.
+    :param offset: From where to start the query.
+    :param can_access_all_locations: When set to False, will limit the users returned by the
+        locations the user has visibility into.
+    :param user_data_fields: A list of custom user data fields that should also be searched
+        by the `search_string`
+
+    :returns: An EsQuery instance.
+    '''
     search_fields = ["base_username", "last_name", "first_name", "phone_numbers"]
 
     should_criteria_query = [
@@ -43,6 +59,10 @@ def login_as_user_query(
             queries.BOOL_CLAUSE(
                 queries.SHOULD_CLAUSE(
                     should_criteria_query,
+                    # It should either match on the search fields like username or it
+                    # should match on the custom user data fields. If this were 2, then
+                    # it would require the search string to match both on the search fields and
+                    # the custom user data fields.
                     minimum_should_match=1,
                 ),
             )
--- a/corehq/apps/cloudcare/esaccessors.py
+++ b/corehq/apps/cloudcare/esaccessors.py
@@ -1,0 +1,53 @@
+from corehq import toggles
+
+from corehq.apps.es import filters, queries, UserES
+from corehq.apps.locations.models import SQLLocation
+
+
+def login_as_user_query(domain, couch_user, search_string, limit, offset, can_access_all_locations=False):
+    search_fields = ["base_username", "last_name", "first_name", "phone_numbers"]
+
+    should_criteria_query = [
+        queries.search_string_query(search_string, search_fields),
+    ]
+
+    if toggles.ENIKSHAY.enabled(domain):
+        should_criteria_query.append(
+            queries.nested_filter(
+                'user_data_es',
+                filters.OR(
+                    filters.AND(
+                        filters.term('user_data_es.key', 'id_issuer_body'),
+                        filters.term('user_data_es.value', search_string),
+                    ),
+                    filters.AND(
+                        filters.term('user_data_es.key', 'id_issuer_number'),
+                        filters.term('user_data_es.value', search_string),
+                    ),
+                )
+            )
+        )
+
+    user_es = (
+        UserES()
+        .domain(domain)
+        .start(offset)
+        .size(limit)
+        .sort('username.exact')
+        .set_query(
+            queries.BOOL_CLAUSE(
+                queries.SHOULD_CLAUSE(
+                    should_criteria_query,
+                    minimum_should_match=1,
+                ),
+            )
+        )
+    )
+
+    if not can_access_all_locations:
+        loc_ids = SQLLocation.objects.accessible_to_user(
+            domain, couch_user
+        ).location_ids()
+        user_es = user_es.location(list(loc_ids))
+
+    return user_es.mobile_users()
--- a/corehq/apps/cloudcare/esaccessors.py
+++ b/corehq/apps/cloudcare/esaccessors.py
@@ -4,27 +4,34 @@ from corehq.apps.es import filters, queries, UserES
 from corehq.apps.locations.models import SQLLocation
 
 
-def login_as_user_query(domain, couch_user, search_string, limit, offset, can_access_all_locations=False):
+def login_as_user_query(
+        domain,
+        couch_user,
+        search_string,
+        limit,
+        offset,
+        can_access_all_locations=False,
+        user_data_fields=None):
     search_fields = ["base_username", "last_name", "first_name", "phone_numbers"]
 
     should_criteria_query = [
         queries.search_string_query(search_string, search_fields),
     ]
 
-    if toggles.ENIKSHAY.enabled(domain):
+    if user_data_fields:
+        or_criteria = []
+        for field in user_data_fields:
+            or_criteria.append(
+                filters.AND(
+                    filters.term('user_data_es.key', field),
+                    filters.term('user_data_es.value', search_string),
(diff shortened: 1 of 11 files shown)


corehq/apps/cloudcare/esaccessors.py
```diff
@@ -0,0 +1,60 @@
+from corehq import toggles
```
dimagimon: F401 'toggles' imported but unused

corehq/apps/cloudcare/views.py
```diff
@@ -14,7 +14,7 @@
 from django.utils.translation import ugettext as _, ugettext_noop
 from django.views.decorators.cache import cache_page
 from django.views.generic import View
-from django.views.generic.base import TemplateView
+from django.views.generic.base import TemplateView, View
```
dimagimon: F811 redefinition of unused 'View' from line 16

corehq/apps/cloudcare/views.py
```diff
@@ -72,6 +74,8 @@
     use_jquery_ui,
 )
 from corehq.apps.users.models import CouchUser, CommCareUser
+from corehq.apps.users.analytics import get_search_users_in_domain_es_query
```
dimagimon: F401 'get_search_users_in_domain_es_query' imported but unused

corehq/pillows/mappings/user_mapping.py
```diff
@@ -95,7 +96,20 @@
...
+                        'key': {
+                            'type': 'string',
+                            'index': 'not_analyzed',
+                        },
+                        'value': {
+                            'type': 'string',
+                            'index': 'not_analyzed',
+                        }
+                    }
+                },
```

corehq/apps/cloudcare/static/cloudcare/js/formplayer/users/collections.js
```diff
@@ -9,7 +9,7 @@ FormplayerFrontend.module("Users.Collections", function(Collections, FormplayerF
             if (!this.domain) {
                 throw new Error('Cannot instantiate collection without domain');
             }
-            return '/a/' + this.domain + '/settings/users/commcare/';
+            return '/a/' + this.domain + '/cloudcare/api/login_as/users/';
```
benrudolph: this used to hit the mobile workers route. this now hits a separate view which frees us up to add functionality without affecting the users page. the big one being changing permissions

corehq/apps/cloudcare/views.py
```diff
@@ -406,6 +410,76 @@ def get(self, request, *args, **kwargs):
...
+                'itemList': map(lambda user: self._format_user(user), users_data.hits),
+                'total': users_data.total,
+                'page': page,
+                'query': query,
+                'total_records': total_records
+            },
+        })
+
+    def _user_query(self, search_string, page, limit):
+        return login_as_user_query(
```
benrudolph: @ctsims once we figure out what custom user data property users should be allowed to query this should be trivial to update

corehq/apps/cloudcare/esaccessors.py
```diff
@@ -0,0 +1,60 @@
...
+        .sort('username.exact')
+        .set_query(
+            queries.BOOL_CLAUSE(
+                queries.SHOULD_CLAUSE(
+                    should_criteria_query,
+                    minimum_should_match=1,
+                ),
+            )
+        )
+    )
```
benrudolph: fyi @esoergel this is how i ended up implementing the query

corehq/apps/cloudcare/views.py
```diff
@@ -406,6 +409,76 @@ def get(self, request, *args, **kwargs):
...
+            page = 1
+        query = request.GET.get('query')
+
+        users_query = self._user_query(query, page - 1, limit)
+        total_records = users_query.count()
+        users_data = users_query.run()
+
+        return json_response({
+            'response': {
+                'itemList': map(lambda user: self._format_user(user), users_data.hits),
```
proteusvacuum: nit, any reason not to use a list comprehension instead of these `map`?

corehq/apps/cloudcare/esaccessors.py
```diff
@@ -0,0 +1,58 @@
...
+
+
+def login_as_user_query(
+        domain,
+        couch_user,
+        search_string,
+        limit,
+        offset,
+        can_access_all_locations=False,
+        user_data_fields=None):
```
benrudolph: going to add docs here

corehq/apps/cloudcare/views.py
proteusvacuum: this might be putting the cart before the horse, but how will you deal with other domains that don't search the same properties as eNikshay?

corehq/pillows/mappings/user_mapping.py
```diff
@@ -95,7 +96,20 @@
...
+                    'type': 'nested',
+                    'dynamic': False,
+                    'properties': {
+                        'key': {
+                            'type': 'string',
+                            'index': 'not_analyzed',
+                        },
+                        'value': {
+                            'type': 'string',
+                            'index': 'not_analyzed',
```
proteusvacuum: not sure why you wouldn't leave this as being analyzed for if we want to run real queries on these values in the future. I guess we can just rewrite the mapping at that point?

corehq/apps/cloudcare/views.py
benrudolph: this should actually be `map(self._format_user, user_data.hits)`. no reason just preference

corehq/apps/cloudcare/views.py
benrudolph: i'm just going to conditionally pass in an array of custom user data fields based on whether or not the ENIKSHAY toggle is enabled. if we ever need to search certain data props for other domains we can make another toggle

corehq/pillows/mappings/user_mapping.py
benrudolph: yeah i'm not sure either to be honest. kind of an arbitrary decision based on the requirement that we didn't need fuzzy searching. can make it analyzed

corehq/pillows/mappings/user_mapping.py
```diff
@@ -95,7 +96,20 @@
...
+                        'key': {
+                            'type': 'string',
+                            'index': 'not_analyzed',
+                        },
+                        'value': {
+                            'type': 'string',
+                            'index': 'analyzed',
+                        }
+                    }
+                },
```

corehq/apps/cloudcare/views.py
benrudolph: https://github.com/dimagi/commcare-hq/pull/16117/commits/03c375060cb414af3ffaf4eb571e702dfdc3051d

corehq/apps/cloudcare/esaccessors.py
benrudolph: https://github.com/dimagi/commcare-hq/pull/16117/commits/ef16eaaf2d8b33c5160134b5450a0b6fc557676e

proteusvacuum (approved):

benrudolph:
Ah wait we can't merge this

corehq/apps/cloudcare/tests/test_esaccessors.py
```diff
@@ -0,0 +1,115 @@
...
+                'child',
+                10,
+                0,
+                user_data_fields=['wild'],
+                can_access_all_locations=True,
+            ).count(),
+            1
+        )
+
+        # Do not fuzzy match
```
esoergel: It's a little odd that this query doesn't fuzzy match against user_data fields, but it DOES fuzzy match against username and the other fields.  Is there a reason to behave differently for the one than the other?

corehq/pillows/mappings/user_mapping.py
```diff
@@ -95,7 +96,20 @@
                                        },
                                        'type': 'object'},
                 'status': {'type': 'string'},
-                'user_data': {'type': 'object', 'enabled': False},
+                'user_data_es': {
```
esoergel: OOC, why user_data_***es***?

esoergel (commented):
A few comments, but nothing blocking.

corehq/pillows/mappings/user_mapping.py
benrudolph: just cause i was changing the way we store the data. i think we usually assume that the couch model maps directly to the ES model. so i would expect if fetching a couch user from ES that the `user_data` field would be a dictionary. if i overwrote that and made that an array i felt that would be confusing

corehq/apps/cloudcare/tests/test_esaccessors.py
benrudolph: just the requirements of nikshay. i don't think it'll be hard to change though

corehq/apps/cloudcare/tests/test_esaccessors.py
```diff
@@ -0,0 +1,115 @@
...
+
+    def setUp(self):
+        super(TestCloudcareESAccessors, self).setUp()
+        self.username = 'superman'
+        self.first_name = 'clark'
+        self.last_name = 'kent'
+        self.doc_type = 'CommCareUser'
+        self.domain = 'user-esaccessors-test'
+        self.es = get_es_new()
+        ensure_index_deleted(USER_INDEX)
```
benrudolph: true, we probably don't. was on autopilot

corehq/apps/cloudcare/esaccessors.py
```diff
@@ -0,0 +1,78 @@
...
+    Takes in various parameters to determine which users to populate the login as screen.
+
+    :param domain: String domain
+    :param couch_user: The CouchUser that is using the Login As feature
+    :param search_string: The query that filters the users returned. Filters based on the
+        `search_fields` as well as any fields defined in `user_data_fields`.
+    :param limit: The max amount of users returned.
+    :param offset: From where to start the query.
+    :param can_access_all_locations: When set to False, will limit the users returned by the
+        locations the user has visibility into.
```
benrudolph: good point

PR #16118, opened 2017-05-11
User mapping update

benrudolph:
@proteusvacuum

### DIFF
--- a/corehq/pillows/mappings/user_mapping.py
+++ b/corehq/pillows/mappings/user_mapping.py
@@ -3,7 +3,7 @@ from corehq.pillows.core import DATE_FORMATS_ARR, DATE_FORMATS_STRING
 
 from pillowtop.es_utils import ElasticsearchIndexInfo
 
-USER_INDEX = es_index("hqusers_2017-04-13")
+USER_INDEX = es_index("hqusers_2017-05-11")
 USER_MAPPING = {'_all': {'analyzer': 'standard'},
  '_meta': {'created': None},
  'date_detection': False,
--- a/corehq/pillows/mappings/user_mapping.py
+++ b/corehq/pillows/mappings/user_mapping.py
@@ -96,6 +96,7 @@ USER_MAPPING = {'_all': {'analyzer': 'standard'},
                                        },
                                        'type': 'object'},
                 'status': {'type': 'string'},
+                'user_data': {'type': 'object', 'enabled': False},
                 'user_data_es': {
                     'type': 'nested',
                     'dynamic': False,
--- a/corehq/ex-submodules/pillowtop/pillow/interface.py
+++ b/corehq/ex-submodules/pillowtop/pillow/interface.py
@@ -100,7 +100,7 @@ class PillowBase(object):
         """
         context = PillowRuntimeContext(changes_seen=0, do_set_checkpoint=True)
         try:
-            for change in self.get_change_feed().iter_changes(since=since, forever=forever):
+            for change in self.get_change_feed().iter_changes(since=since or None, forever=forever):
                 if change:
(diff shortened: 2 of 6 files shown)


corehq/pillows/mappings/user_mapping.py
```diff
@@ -95,7 +96,20 @@
...
+                        'key': {
+                            'type': 'string',
+                            'index': 'not_analyzed',
+                        },
+                        'value': {
+                            'type': 'string',
+                            'index': 'not_analyzed',
+                        }
+                    }
+                },
```

proteusvacuum (approved):

benrudolph:
i will be running:
```
./manage.py ptop_reindexer_v2 user
```
on

- [ ] prod
- [ ] icds
- [ ] softlayer
(the other envs i think are small enough to do on deploy)

then i will merge

corehq/pillows/mappings/user_mapping.py
```diff
@@ -95,7 +96,20 @@
...
+                        'key': {
+                            'type': 'string',
+                            'index': 'not_analyzed',
+                        },
+                        'value': {
+                            'type': 'string',
+                            'index': 'analyzed',
+                        }
+                    }
+                },
```

emord:
@benrudolph in general users are small enough to do during a deploy (could even just run `preindex_views` from deploy)

yesterday we ran into an issue when this index was changed and it broke group to user and unknown users see https://manage.dimagi.com/default.asp?253934 for more details 

fyi @NoahCarnahan

benrudolph:
(offlined a bit, not going to do anything until i fully understand what will happen)

benrudolph:
https://github.com/dimagi/commcare-hq/pull/16118/commits/93b7afb6e8881d302b05d11a800b40f5efa580d0 -- @emord / @snopoke i don't see why these pillows are tied to the ES index name. i've changed it to they do not. it seems odd that we reprocess the last 30 days of changes for those pillows when we update the user mapping. let me know if you think i'm overlooking something.

https://github.com/dimagi/commcare-hq/pull/16118/commits/e72c92fffcdc2c8f325d05a5ca89ccc5fed5ba12 -- this enables me to start the pillows without an error complaining about the `since` param being `{}`.

also fyi @calellowitz i found this while grepping and realized it's probably broken since the change to `DEFAULT_EMPTY_CHECKPOINT_SEQUENCE`: https://github.com/dimagi/commcare-hq/blob/br/user-mapping-update/corehq/apps/es/management/commands/restore_es_snapshot.py#L91. i didn't address in this PR

snopoke:
@benrudolph I think that's OK though it does mean that you should stop the pillows prior to reindexing so that any changes that happen between reindex and switching to the new index get processed.

Fixed the `DEFAULT_EMPTY_CHECKPOINT_SEQUENCE` issue here: https://github.com/dimagi/commcare-hq/pull/16155

benrudolph:
>  it does mean that you should stop the pillows prior to reindexing so that any changes that happen between reindex and switching to the new index get processed.

why does that imply that? those pillows should be reading from the user topic in kafka. as long as all changes get pushed to that topic and none are skipped, then i don't see how changing the ES index affects those pillows

benrudolph:
AH! i understand now. those pillows directly modify the user es index. will revert that commit

corehq/pillows/mappings/user_mapping.py
```diff
@@ -96,6 +96,7 @@
                                        },
                                        'type': 'object'},
                 'status': {'type': 'string'},
+                'user_data': {'type': 'object', 'enabled': False},
```
benrudolph: this was accidentally removed in the previous PR

emord (approved):
Looks reasonable to me. I wonder how hard it would be to combine these into one pillow

PR #16123, opened 2017-05-11
Make repeaters fire for record side effect free

dannyroberts:
This is what I've been working towards. Not 100% about my approach: `self.cancel()` and `self.set_next_try()` were a bit more suggestive than then the `return RepeatRecordAttempt(...)` lines they've been replaced with, but I think the present change—isolating the side effects of a single attempt into a returnable object—is pretty necessary for keeping track of multiple attempts better.

This repeater still doesn't actually _change_ anything fundamentally behavior-wise; the next step—which should now be pretty easy—is to actually store these `RepeatRecordAttempt`s as a list on `RepeatRecord`. (And the step after that is to do something with that in the UI.)

### DIFF
--- a/corehq/apps/repeaters/models.py
+++ b/corehq/apps/repeaters/models.py
@@ -600,17 +600,19 @@ class RepeatRecord(Document):
     def handle_failure(self, response):
         """Do something with the response if the repeater fails
         """
-        self._fail(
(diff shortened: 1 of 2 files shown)


dannyroberts:
@proteusvacuum @emord buddy

proteusvacuum (approved):

PR #16125, opened 2017-05-11
Convert ewsghana to use only SQLlocations

esoergel:


custom/ewsghana/models.py
```diff
@@ -95,7 +95,7 @@ class EWSExtension(models.Model):
     def supply_point(self):
         if not self.location_id:
             return
-        return Location.get(doc_id=self.location_id).linked_supply_point()
+        return get_location(doc_id=self.location_id).linked_supply_point()
```
gcapalbo: i think this will need to change from a kwarg to just be the first positional arg

esoergel:
@gcapalbo fixed it

PR #16129, opened 2017-05-12
Login as claim

benrudolph:
@proteusvacuum @wpride @ctsims i checked on the server when trying to claim the case in https://manage.dimagi.com/default.asp?253902 that the session said i had already claimed it. i realized that we were storing the session id of the web user who claimed the case, not the restore user. so if a web user was logged in as a different user and claimed a case, and then logged in as a different user and claimed that same case, it would fail. see test case for example

### DIFF
--- a/corehq/apps/ota/tests/test_search_claim_endpoints.py
+++ b/corehq/apps/ota/tests/test_search_claim_endpoints.py
@@ -3,6 +3,7 @@ from uuid import uuid4
 
 from django.urls import reverse
 from django.test import TestCase, Client
+from dimagi.utils.couch.cache.cache_core import get_redis_default_cache
 from mock import patch
 
 from casexml.apps.case.mock import CaseBlock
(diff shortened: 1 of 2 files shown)


snopoke (approved):

wpride:
@benrudolph 💵 thanks man

corehq/apps/ota/tests/test_search_claim_endpoints.py
```diff
@@ -141,6 +144,43 @@ def test_claim_restore_as(self):
         claim_case = CaseAccessors(DOMAIN).get_case(claim_ids[0])
         self.assertEqual(claim_case.owner_id, other_user._id)
 
+    def test_claim_restore_as_proper_cache(self):
```
proteusvacuum: probably not a huge issue, but the other tests here `@run_with_all_backends`

proteusvacuum:
Thanks @benrudolph

corehq/apps/ota/tests/test_search_claim_endpoints.py
benrudolph: yeah i thought about it, but decided i wasn't testing that functionality with this test and didn't want to add the overhead

PR #16131, opened 2017-05-12
check connection to all databases

javierwilson:


corehq/apps/hqadmin/service_checks.py
```diff
@@ -157,10 +159,27 @@ def check_heartbeat():
 
 
 def check_postgres():
+    connected = True
+    status_str = ""
+    for db in settings.DATABASES:
+        db_conn = connections[db]
+        try:
+            c = db_conn.cursor()
```
dimagimon: F841 local variable 'c' is assigned to but never used

javierwilson: cheeky!

emord (approved):
Looks good. You may want to add some newlines though. Also is there a way we could use this for the standbys as well? Are they in settings.DATABASES?

esoergel (approved):
Looks good.  I agree with Emord that you'll probably want some newlines in there, though.  You could also make these two separate checks - one to test the connections, and one to perform a dummy query.

javierwilson:
I am going to merge this now, pending improvements.  Standbys are not listed anywhere in the Django environment :( But it is kind of urgent to test standys and backups. I will come up with a solution for this.

PR #16137, opened 2017-05-12
Don't process bad case in UCR or ES pillows

proteusvacuum:
Not sure if this is a good approach or not.
I can revert this once the changes have been processed, and the offending forms against that case have been archived.
@emord @NoahCarnahan


emord (approved):
Looks reasonable. It would probably be best to rebase this on master (if it isn't already) and deploy india from this branch so that it doesn't get into any other envs and we don't have to remember to remove it

proteusvacuum:
This is currently live on softlayer and seems to be working. Closing this out so it doesn't get merged.
I'll archive the bad forms on Monday.
@benrudolph FYI.

PR #16139, opened 2017-05-12
function to prune synclogs

emord:
https://trello.com/c/ylowaGGK/90-purge-old-synclogs

Sync logs are growing pretty quickly these days. it's > 150 GB on icds after compaction.

I wasn't sure about the correct thresholds to use, but this will allow us to do some pruning manually before making a periodic celery task.

When is it safe to start deleting synclogs? A day ago, week ago, month ago?
How many docs can you safely delete at once in couch without causing too much load on view reindexes? (we likely won't know that until we hit it)
Should these parameters be configurable based on environment? (would likely need to happen in localsettings)
Is it worth it to do something smarter? Like don't delete if this is the users latest sync? I think that would add a lot of load to couch or would at least take a lot of requests

@snopoke @dannyroberts

### DIFF
--- a/corehq/ex-submodules/casexml/apps/phone/tests/test_sync_mode.py
+++ b/corehq/ex-submodules/casexml/apps/phone/tests/test_sync_mode.py
@@ -20,11 +20,10 @@ from corehq.form_processor.tests.utils import (
     FormProcessorTestUtils,
     use_sql_backend,
 )
-from corehq.toggles import LOOSE_SYNC_TOKEN_VALIDATION
 from corehq.util.test_utils import flag_enabled
 from casexml.apps.case.tests.util import (
     check_user_has_case, assert_user_doesnt_have_case,
-    assert_user_has_case, TEST_DOMAIN_NAME, assert_user_has_cases, check_payload_has_cases,
+    assert_user_has_case, TEST_DOMAIN_NAME, assert_user_has_cases,
     check_payload_has_case_ids, assert_user_doesnt_have_cases)
 from casexml.apps.phone.tests.utils import create_restore_user, has_cached_payload
 from casexml.apps.phone.models import SyncLog, get_properly_wrapped_sync_log, SimplifiedSyncLog, \
@@ -2070,14 +2069,6 @@ class SyncTokenReprocessingTestSQL(SyncTokenReprocessingTest):
 
 class LooseSyncTokenValidationTest(SyncBaseTest):
 
-    def test_submission_with_bad_log_default(self):
-        with self.assertRaises(ResourceNotFound):
-            post_case_blocks(
-                [CaseBlock(create=True, case_id='bad-log-default').as_xml()],
-                form_extras={"last_sync_token": 'not-a-valid-synclog-id'},
-                domain='some-domain-without-toggle',
-            )
-
     def test_submission_with_bad_log_toggle_enabled(self):
         domain = 'submission-domain-with-toggle'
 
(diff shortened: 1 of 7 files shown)


dannyroberts (approved):

dannyroberts:
So, I believe the recovery if a needed sync log is missing is "graceful"; it may take a lot longer to do from scratch, but I think it will fall back to that rather than failing. Worth actually making sure about that though before we do this.

In terms of doing something "smarter", Cory laid out his understanding of the exact condition under which you can "safely" delete a sync log (i.e. without it causing full restore down the line): https://trello.com/c/LomMo8WF/11-find-and-delete-unused-documents-synclogs.

One thing you could do is every time someone syncs, you arrange the sync log _previous to the one being synced against_. You can't delete the one for the actual sync token sent, because it's not guaranteed that the restore "took" on the phone, and it may try to sync with that token again; but you can delete all previous ones in its chain. Looking up this whole chain would be one request per sync log, so maybe not worth it. Also, there would inevitably be some "abandoned" sync logs, so you'd still have to do the bulk delete of everything older than 6 months or whatever. So, not totally sure that's worth it.

emord:
@dannyroberts I get an error if I try to go to that trello card. Maybe I don't have permissions for that?

dannyroberts:
Derp. OK just tried adding you

snopoke:
In order to make the failure mode graceful the `LOOSE_SYNC_TOKEN_VALIDATION` toggle has to be enabled. Before we go ahead with removing sync logs we'll need to remove the toggle and always fail gracefully.

dannyroberts:
Ah, okay, makes sense @snopoke

corehq/ex-submodules/casexml/apps/phone/tests/test_sync_mode.py
```diff
@@ -2088,25 +2079,9 @@ def _test():
                 domain=domain,
             )
 
-        LOOSE_SYNC_TOKEN_VALIDATION.set(domain, False, namespace='domain')
-        with self.assertRaises(ResourceNotFound):
-            _test()
-
-        LOOSE_SYNC_TOKEN_VALIDATION.set(domain, True, namespace='domain')
         # this is just asserting that an exception is not raised after the toggle is set
```
snopoke: nit: comment still refers to toggle

corehq/ex-submodules/casexml/apps/phone/utils.py
```diff
@@ -7,3 +7,11 @@ def get_restore_response_class(domain):
     if BLOBDB_RESTORE.enabled(domain):
         return BlobRestoreResponse
     return FileRestoreResponse
+
+
+def delete_sync_logs(before_date, limit=1000):
```
snopoke: Is the plan to put this in a celery task? Should this not be called recursively (or do an iteration) to ensure that all logs prior to `date` are removed?

snopoke (commented):
I'm fine to go ahead with this approach using a fairly high date threshold (e.g. 60+ days).

I think there are some other cases that need to be considered though. Since we're deleting synclogs it means that forms and case transactions / actions will now point to deleted docs and will raise `ResourceNotFound` if they try to fetch them. I'm sure there are places where that isn't handled correctly e.g. https://github.com/dimagi/commcare-hq/blob/83b4d9627d7c25e64c9cffd87c3acf05d1b29c22/corehq/apps/commtrack/processing.py#L68

corehq/ex-submodules/casexml/apps/phone/utils.py
emord: I haven't put it in a celery task because I thought we should do some manual testing first. Deleting 5 years of synclogs from prod at once seemed scary

PR #10718, opened 2016-03-11
AncestorLocationTypeFilter is not an AutoFilter

kaapstorm:
I should revert the commit that added `_filter_by_ancestor_location_type_id` and `_filter_by_parent_location_id`, and then rebase this on top.

I got the original design wrong: This is not supposed to be an AutoFilter, in that the device user doesn't get to choose the ancestor location type. The report creator sets it.

The UI is currently producing a JS error. (It seems that `filter.selectedValue[value]` ([report-module.js:201](https://github.com/dimagi/commcare-hq/blob/e85d4351431ec8704d4fde56a87b6f953db941d4/corehq/apps/app_manager/static/app_manager/js/report-module.js#L201-L201)) is not an observable for this filter. I don't understand why -- it is for all the other filters.)

Context: [FB 220703](http://manage.dimagi.com/default.asp?220703), [FB 215658](http://manage.dimagi.com/default.asp?215658), [Trello](https://trello.com/c/q1MDLHl7/99-mobile-ucr-custom-location-id-options)

Anyway. Open for review. 

@benrudolph @calellowitz cc @czue @dannyroberts @nickpell

PR #16140, opened 2017-05-12
Fix `AncestorLocationTypeFilter.get_filter_value`

NoahCarnahan:
`AncestorLocationTypeFilter.get_filter_value()` should use the corresponding ui filter to get a value of the correct type. The other mobile UCR filters already do this e.g. the [location_id AutoFilter](https://github.com/dimagi/commcare-hq/blob/cce1537355777a89c69abb0a8a97c0a4d42e708a/corehq/apps/app_manager/models.py#L3938).

This function was returning strings, but when the `AncestorLocationTypeFilter` is used with a `DynamicChocieListFilter` for instance, functions further down the line were expecting [`Choice`](https://github.com/dimagi/commcare-hq/blob/cce1537355777a89c69abb0a8a97c0a4d42e708a/corehq/apps/reports_core/filters.py#L303) objects. In fact, I'm not sure I understand how this feature was ever working. @kaapstorm, does it seem like I'm missing something about how this is supposed to work? Do you know if other projects used this mobile ucr filter successfully somewhere?

buddy @sravfeyn

### DIFF
--- a/corehq/apps/app_manager/models.py
+++ b/corehq/apps/app_manager/models.py
@@ -4160,14 +4160,18 @@ class AncestorLocationTypeFilter(ReportAppFilter):
 
     def get_filter_value(self, user, ui_filter):
         from corehq.apps.locations.models import SQLLocation
+        from corehq.apps.reports_core.filters import REQUEST_USER_KEY
 
+        kwargs = {REQUEST_USER_KEY: user}
         try:
             ancestor = user.sql_location.get_ancestors(include_self=True).\
                 get(location_type__name=self.ancestor_location_type_name)
+            kwargs[ui_filter.name] = ancestor.location_id
         except (AttributeError, SQLLocation.DoesNotExist):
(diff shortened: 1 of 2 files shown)


kaapstorm:
Hey @NoahCarnahan It was originally developed for ICDS, and used by Matt Theis. See [Trello](https://trello.com/c/q1MDLHl7/99-mobile-ucr-custom-location-id-options) for more details. @sheelio may have more context if you need it.

Judging from the [original PR](#10718) I also thought that this was supposed to be an AutoFilter where the user can choose the ancestor location type. But actually the location type is set by the report creator, which is why it only returns the value to be matched.

NoahCarnahan:
Checked with Sheel and review ICDS app. I don't think this change will have any affect on them.

NoahCarnahan:
@nickpell would you mind reviewing this? I would like to have it in before today's deploy. Thanks!

PR #16141, opened 2017-05-12
don't require Episode.beneficiaryID to be foreign key

nickpell:
makes life easier so I don't have to clean the data dump. Episodes with invalid beneficiaryIDs will get ignored anyway.


custom/enikshay/private_sector_datamigration/migrations/0001_initial.py
```diff
@@ -138,6 +138,7 @@ class Migration(migrations.Migration):
                 ('associatedFO', models.CharField(max_length=255, null=True)),
                 ('bankName', models.CharField(max_length=255, null=True)),
                 ('basisOfDiagnosis', models.CharField(max_length=255, null=True)),
+                ('beneficiaryID', models.CharField(max_length=18)),
```
nickpell: I want this column to be in the right place to make migrating from the data dumps easier.  Since this data is only used in the data migration, its ok to rollback private_sector_datamigration to 0000, then re-apply migrations, and re-import into postgresql.

PR #16006, opened 2017-05-03
private sector 99 dots updates

proteusvacuum:
Bunch of updates to the 99DOTS payload for 1.1 and phase 2a of eNikshay

@NoahCarnahan @mkangia 🐡 / 🐡 

Hold off on merge because I think it will make a bunch of conflicts in https://github.com/dimagi/commcare-hq/pull/15963 and that one is slightly more pressing (also I want to verify with 99DOTS that they are OK with us changing the format)

PR #16149, opened 2017-05-14
Mk enikshay private api legit

mkangia:
[Spec](https://docs.google.com/document/d/1kHEzXLq-8RvPRcRXbYaSNgFKeD3OFP7DR31VytNxxl8/edit#heading=h.svb16cn9xwxo)


custom/enikshay/integrations/nikshay/repeater_generator.py
```diff
@@ -337,6 +343,98 @@ def handle_exception(self, exception, repeat_record):
...
+            "password": settings.ENIKSHAY_PRIVATE_API_PASSWORD,
+            "Source": ENIKSHAY_ID,
+        }
+
+    def handle_success(self, response, payload_doc, repeat_record):
+        # A successful response returns a Nikshay ID like 00001
+        # Failures also return with status code 200 and some message like
+        # Dublicate Entry or Invalid data format
+        # (Dublicate is not a typo)
+        message = get_message_from_xml_response(repeat_record.repeater.url, repeat_record.repeater.operation, response)
```
dimagimon: E501 line too long (119 > 115 characters)

dannyroberts:
Hey @mkangia great, seems like a new PR is the right idea. I also wanted to say sorry for the overlapping changes—hopefully the cleanup made things easier in some ways even if it made things harder in others.

I just talked this over with @proteusvacuum, and I think he and I are both on the same page. I wanted to take up a conversation from the last thread (https://github.com/dimagi/commcare-hq/pull/15958#issuecomment-300271583) and continue it here.

> i thought we wanted to make this generic so we would still need to modify the `fire` now `fire_for_record` method in the `Repeater` itself?

I think this is a response to @proteusvacuum's suggestion to override (rather than modify) `Repeater`'s `fire` method.

At the core of that question, in my mind, is the question of what you mean by "generic". I can think of two possible answers (though maybe you can better articulate it if there's a third):
- Possibility \#1: "By generic, we mean that non-custom users should be able to set up form, case, and app forwarding to SOAP endpoints without custom code"
- Possibility \#2: "By generic, we mean that the next time we need to add a custom repeater to a SOAP endpoint, either for eNikshay or for another project, it should be easy switch that on by following a particular pattern, with minimal code duplication."

Assuming those are the possibilities, my understanding is that what we mean by generic is \#2—specifically, we _don't_ think it's plausible that someone will be able to set up a (non-custom) case or form repeater to a SOAP endpoint entirely in the public UI with no custom code. I say that, because, well I guess, how would they? They'd have to be able to specify what fields show up in the payload, etc., since it's very unlikely that the fields we send would map in any obvious pre-determined way to the fields that their SOAP API is set up to receive.

My understanding of this PR, however, is that it's implicitly assuming \#1—that we should expose this to non-custom repeaters as well: that everyone using the public/non-custom UI will see a new *SOAP operation* field, and that the help text under the *URL to forward to* field will say _Please enter the full url, like http://www.example.com/forwarding/. Enter wsdl link for SOAP_. With no further ability to configure the SOAP payload, however, users will be led to believe that they can do something which they cannot actually reasonably do, so these new fields will only be confusing. Let me know though if I'm misreading the PR.

If what I've said so far is true, then I would suggest the following changes:
- break out this new code into a `SOAPRepeaterMixin`. It's not the perfect abstraction, but I think it's good enough—pretty good actually—and also the best we can do given how repeaters are currently structured.
- have your custom repeater inherit from that mixin. This will be the "on switch" to the "generic" SOAP repeater ability.
- into the mixin, move the new properties
- you may also want to make a corresponding `SOAPRepeaterFormMixin` or something similar, for the UI
- finally, either override the `fire_for_record` entirely, or start by refactoring it so that
  ```python
        headers = self.get_headers(repeat_record)
        auth = self.get_auth()
        payload = self.get_payload(repeat_record)
        url = self.get_url(repeat_record)
        response = simple_post(payload, url, headers=headers, timeout=POST_TIMEOUT, auth=auth)
  ```
  lives inside a separate function `deliver_for_record`, and then override just that.

Thanks for reading through this rather large comment, but hopefully it's helpful. Let me know if you want to discuss more—I'm always happy to set up a call.

mkangia:
Hey @dannyroberts 
Much appreciate the detailed feedback.

Everything you said makes sense. Thanks for taking the time to write this down. I missed the part about "non-custom" repeaters.

Will make the change.

`I just talked this over with @proteusvacuum`
Also devs at DSI never get to talk things over. Time to change timezone.. #timezone_no_talks

custom/enikshay/integrations/nikshay/tests/test_repeaters.py
```diff
@@ -44,6 +54,25 @@ def json(self):
...
 
+class MockSoapResponse(object):
+    def __init__(self, status_code, content):
+        self.status_code = status_code
+        self.content = content
+        self.headers = {'Content-Type': 'text/xml; charset=utf-8'}
+
+FAILURE_RESPONSE = (
+    '<?xml version="1.0" encoding="utf-8"?><soap:Envelope xmlns:soap="http://schemas.xmlsoap.org/soap/envelope/" '
+    'xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns:xsd="http://www.w3.org/2001/XMLSchema"><soap:Body>'
```
dimagimon: E501 line too long (116 > 115 characters)

custom/enikshay/integrations/nikshay/tests/test_repeaters.py
```diff
@@ -44,6 +54,25 @@ def json(self):
...
+    '<InsertHFIDPatient_UATBCResponse xmlns="http://tempuri.org/"><InsertHFIDPatient_UATBCResult>Invalid data format'
```
dimagimon: E501 line too long (117 > 115 characters)

custom/enikshay/integrations/nikshay/tests/test_repeaters.py
```diff
@@ -44,6 +54,25 @@ def json(self):
...
+    '</InsertHFIDPatient_UATBCResult></InsertHFIDPatient_UATBCResponse></soap:Body></soap:Envelope>')
+
+SUCCESSFUL_SOAP_RESPONSE = (
+    '<?xml version="1.0" encoding="utf-8"?><soap:Envelope xmlns:soap="http://schemas.xmlsoap.org/soap/envelope/" '
+    'xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns:xsd="http://www.w3.org/2001/XMLSchema"><soap:Body>'
```
dimagimon: E501 line too long (116 > 115 characters)

mkangia:
@dannyroberts changes made.
1. Extract `send_request` on repeater and override that. I have just extracted the code to send request. Was it better to extract all arguments as well as you suggested?
2. Added `SOAPCaseRepeaterForm` and `AddCustomSOAPCaseRepeaterView`. Added `class` instead of a mixin.

Reverted all changes from `Repeater` and `CaseRepeaterForm`

custom/enikshay/integrations/nikshay/repeaters.py
```diff
@@ -186,6 +178,42 @@ def allowed_to_forward(self, test_case):
...
+        return (
+            episode_case_properties.get('nikshay_registered', 'false') == 'false' and
+            not episode_case_properties.get('nikshay_id') and
+            case_properties_changed(episode_case, [PRIVATE_PATIENT_EPISODE_PENDING_REGISTRATION]) and
+            episode_case_properties.get(PRIVATE_PATIENT_EPISODE_PENDING_REGISTRATION, 'yes') == 'no' and
+            is_valid_person_submission(person_case)
+        )
+
+    def send_request(self, payload, url, headers, auth):
+        return simple_xml_post(payload, url, self.operation)
```
dannyroberts: I noticed you still added `operation` property to the `Repeater` base class; again I'd suggest having a `SOAPRepeaterMixin` (a mixin just being a `class` that you don't intend to instantiate directly, but to supplement a second base class) that has the operation property and this `send_request` method.

(That's because for other repeaters, the `operation` property is meaningless, so it shouldn't go on the base class.)

I would then have `NikshayRegisterPrivatePatientRepeater` subclass both `SOAPRepeaterMixin` and `BaseNikshayRepeater`

corehq/apps/repeaters/models.py
```diff
@@ -284,14 +285,17 @@ def get_auth(self):
...
+
     def fire_for_record(self, repeat_record):
         headers = self.get_headers(repeat_record)
         auth = self.get_auth()
         payload = self.get_payload(repeat_record)
         url = self.get_url(repeat_record)
 
         try:
-            response = simple_post(payload, url, headers=headers, timeout=POST_TIMEOUT, auth=auth)
+            response = self.send_request(payload, url, headers, auth)
```
dannyroberts: to your question about whether to also pull out the other params: this is probably fine for the time being. My thinking was basically that since depending on what you actually do in `send_request`, you'll use different information and since it's all determined off of `self` and `repeat_record` we could just let each `send_request` gather its own headers, auth, etc as necessary. (For example, when you override it below, you don't use all the params.) That said, it's a tad tricky because for example `self.get_payload(repeat_record)` might raise an one of the exceptions handled in the except block below (it probably doesn't but I haven't thought too hard about it), so putting it inside would theoretically change the behavior in a way we don't want, etc.

I'm fine leaving it the way you have it now, and I'll likely be revisiting it in a later refactor down the line anyway.

dannyroberts:
> > I just talked this over with @proteusvacuum
> 
> Also devs at DSI never get to talk things over. Time to change timezone.. #timezone_no_talks

Haha 😢

custom/enikshay/integrations/nikshay/repeaters.py
mkangia: Mixins in django have not yet sunk in for me. I feel ruby does it slightly better.

Anyway, that change makes sense.

corehq/apps/repeaters/models.py
mkangia: I don't have a preference to either of those approaches as of now. Just did the one which was less code changes. Making that change if that sounds better.

dannyroberts:
Nice, it's looking really clean now, I like it!

dannyroberts (approved):

custom/enikshay/integrations/nikshay/repeater_generator.py
```diff
@@ -337,6 +343,102 @@ def handle_exception(self, exception, repeat_record):
...
+            "Address": person_case_properties.get('current_address', '').replace(',', ''),
+            "pin": person_case_properties.get('current_address_postal_code', ''),
+            "lno": person_case_properties.get('phone', ''),
+            "mno": '0',
+            "tbdiagdate": str(episode_date),
+            "tbstdate": episode_case_properties.get(TREATMENT_START_DATE, str(datetime.date.today())),
+            "Type": disease_classification.get(episode_case_properties.get('pulmonary_extra_pulmonary', ''), ''),
+            "B_diagnosis": episode_case_properties.get('case_definition', ''),
+            "D_SUSTest": episode_case_properties.get('drug_susceptibility_test_status', ''),
+            "Treat_I": episode_case_properties.get('treatment_initiation_status', ''),
```
proteusvacuum: this should be `treatment_initiated`. Also there is a mapping to perform here.

custom/enikshay/integrations/nikshay/repeater_generator.py
```diff
@@ -337,6 +343,102 @@ def handle_exception(self, exception, repeat_record):
...
+            # API for Address with char ',' returns Invalid data format error
+            "Address": person_case_properties.get('current_address', '').replace(',', ''),
+            "pin": person_case_properties.get('current_address_postal_code', ''),
+            "lno": person_case_properties.get('phone', ''),
+            "mno": '0',
+            "tbdiagdate": str(episode_date),
+            "tbstdate": episode_case_properties.get(TREATMENT_START_DATE, str(datetime.date.today())),
+            "Type": disease_classification.get(episode_case_properties.get('pulmonary_extra_pulmonary', ''), ''),
+            "B_diagnosis": episode_case_properties.get('case_definition', ''),
+            "D_SUSTest": episode_case_properties.get('drug_susceptibility_test_status', ''),
```
proteusvacuum: this should be `dst_status` and there is a mapping that needs to happen

custom/enikshay/integrations/nikshay/repeater_generator.py
```diff
@@ -337,6 +343,102 @@ def handle_exception(self, exception, repeat_record):
...
+            "gender": person_case_properties.get('sex', '').capitalize(),
+            # API for Address with char ',' returns Invalid data format error
+            "Address": person_case_properties.get('current_address', '').replace(',', ''),
+            "pin": person_case_properties.get('current_address_postal_code', ''),
+            "lno": person_case_properties.get('phone', ''),
+            "mno": '0',
+            "tbdiagdate": str(episode_date),
+            "tbstdate": episode_case_properties.get(TREATMENT_START_DATE, str(datetime.date.today())),
+            "Type": disease_classification.get(episode_case_properties.get('pulmonary_extra_pulmonary', ''), ''),
+            "B_diagnosis": episode_case_properties.get('case_definition', ''),
```
proteusvacuum: There's a mapping that needs to happen here.

custom/enikshay/integrations/nikshay/repeater_generator.py
```diff
@@ -337,6 +343,102 @@ def handle_exception(self, exception, repeat_record):
...
+            "age": person_case_properties.get('age', ''),
+            "gender": person_case_properties.get('sex', '').capitalize(),
+            # API for Address with char ',' returns Invalid data format error
+            "Address": person_case_properties.get('current_address', '').replace(',', ''),
+            "pin": person_case_properties.get('current_address_postal_code', ''),
+            "lno": person_case_properties.get('phone', ''),
+            "mno": '0',
+            "tbdiagdate": str(episode_date),
+            "tbstdate": episode_case_properties.get(TREATMENT_START_DATE, str(datetime.date.today())),
+            "Type": disease_classification.get(episode_case_properties.get('pulmonary_extra_pulmonary', ''), ''),
```
proteusvacuum: where is `disease_classification` defined? 
This property should be `episode.disease_classification`. There's also a mapping.

proteusvacuum (changes requested):
there were a few changes to the app that I didn't add to the spec (sorry). Just a few case property changes. You can see an up-to-date list here: 
https://india.commcarehq.org/a/enikshay/apps/view/9340429733463e58ae0e1518defee221/summary/#/cases

custom/enikshay/integrations/nikshay/repeater_generator.py
mkangia: `disease_classification` is in the field mappings
```
disease_classification = {
    'pulmonary': 'P',
    'extra_pulmonary': 'EP',
}
```
Added this taking hints from the doc
Type -> Type of TB (P = Pulmonary; EP = Extra Pulmonary) -> episode.pulmonary_extra_pulmonary

custom/enikshay/integrations/nikshay/repeater_generator.py
mkangia: from the doc it seemed just the case property would do.
Seeing Case summary, seems like this would be a combination of Basis of Diagnosis and Case definition.

Can that mapping logic be configured in the App itself and just saved on a case property that we can pick?

custom/enikshay/integrations/nikshay/repeater_generator.py
mkangia: Got that name from the doc. 
That mapping seems straight forward. Will add.

Sorry, Can you confirm this the correct doc https://docs.google.com/spreadsheets/d/1WWsxUrt24zYR-IVNh7SmrcOgfpEjK9t-ZFWfM2YR5WE/edit#gid=0

custom/enikshay/integrations/nikshay/repeater_generator.py
mkangia: Got that name from the doc.

Will add mapping.

custom/enikshay/integrations/nikshay/repeater_generator.py
```diff
@@ -337,6 +343,102 @@ def handle_exception(self, exception, repeat_record):
...
+            "password": settings.ENIKSHAY_PRIVATE_API_PASSWORD,
+            "Source": ENIKSHAY_ID,
+        }
+
+    def handle_success(self, response, payload_doc, repeat_record):
+        # A successful response returns a Nikshay ID like 00001
+        # Failures also return with status code 200 and some message like
+        # Dublicate Entry or Invalid data format
+        # (Dublicate is not a typo)
+        message = get_message_from_xml_response(
```
mkangia: ya... thats not a good name..

message from a SOAP action would be closest. `parse_SOAP_response` sounds good.

custom/enikshay/integrations/nikshay/repeater_generator.py
```diff
@@ -337,6 +343,102 @@ def handle_exception(self, exception, repeat_record):
...
+            repeat_record.repeater.url,
+            repeat_record.repeater.operation,
+            response
+        )
+        try:
+            if isinstance(message, basestring) and message.isdigit():
```
mkangia: I have always seen just numbers in that response while testing.. though i have not done enough to certain it. Also as you shared
``Private Sector Nikshay ID format is: {6 digit number that represents the agency}-{incrementing integer}`

custom/enikshay/integrations/nikshay/repeater_generator.py
mkangia: mapping added.

custom/enikshay/integrations/nikshay/repeater_generator.py
mkangia: Could not find this one the form linked on the doc. Can you please help where i can find the values for mapping.

custom/enikshay/integrations/nikshay/field_mappings.py
```diff
@@ -119,3 +119,11 @@
...
     '3+': 13,
 }
+
+drug_susceptibility_test_status = {
+    'pending': 'N',
+    'not_done': 'N',
+    'rif_sensitive': 'S',
+    'rif_resistant': 'R',
+    'xdr': 'N'
+}
```
dimagimon: W292 no newline at end of file

custom/enikshay/integrations/nikshay/repeater_generator.py
```diff
@@ -337,6 +348,102 @@ def handle_exception(self, exception, repeat_record):
...
+            "fhname": sanitize_text_for_xml(person_case_properties.get('husband_father_name', '')),
+            "age": person_case_properties.get('age', ''),
+            "gender": person_case_properties.get('sex', '').capitalize(),
+            # API for Address with char ',' returns Invalid data format error
+            "Address": sanitize_text_for_xml(person_case_properties.get('current_address', '').replace(',', '')),
+            "pin": person_case_properties.get('current_address_postal_code', ''),
+            "lno": person_case_properties.get('phone', ''),
+            "mno": '0',
+            "tbdiagdate": _format_date(str(episode_date)),
+            "tbstdate": _format_date(episode_case_properties.get(TREATMENT_START_DATE, str(datetime.date.today()))),
```
dimagimon: E501 line too long (116 > 115 characters)

mkangia:
Hey @dannyroberts thanks for the reviews

I had to make this [one change](https://github.com/dimagi/commcare-hq/pull/16149/commits/5c85cbfdcff4235d470573e7515fd3235eb8a145) after you approved changes. Not sure if thats the best way to do this but seemed the way to do it for now. 

This change was done to allow exception in case of payload fetch fails which was working earlier but now failed due to changes made for extracting the request to overwrite it. For details [failing build](https://travis-ci.org/dimagi/commcare-hq/jobs/232597153)

corehq/apps/repeaters/models.py
```diff
@@ -284,16 +284,16 @@ def get_auth(self):
...
         return simple_post(payload, url, headers=headers, timeout=POST_TIMEOUT, auth=auth)
 
     def fire_for_record(self, repeat_record):
+        payload = self.get_payload(repeat_record)
         try:
-            response = self.send_request(repeat_record)
+            response = self.send_request(repeat_record, payload)
         except (Timeout, ConnectionError) as error:
             log_repeater_timeout_in_datadog(self.domain)
             return self.handle_response(RequestConnectionError(error), repeat_record)
```
dannyroberts: Aha! I missed that the lines right below this are 

```python
except Exception as e:
    return self.handle_response(e, repeat_record)
```

It makes sense that that would break the error handling. This looks like a good fix for now!

custom/enikshay/integrations/nikshay/repeater_generator.py
proteusvacuum: The case property here is `episode_case_properties.get('basis_of_diagnosis')`
The doc linked to the form where the values are. I didn't list them all out in the document since they were already listed int the form.
If you want to add extra case properties onto the case, feel free. You can call it `basis_of_diagnosis_nikshay` or something. 

Otherwise, here is the mapping:
- `clinical_chest` -> `X`
- `clinical_other` -> `E`
- `microbiological_smear` -> `S`
- `microbiological_cbnaat` -> `M`
- `microbiological_culture` -> `C`
- `microbiological_pcr` -> `M`
- `microbiological_other` -> `M`

custom/enikshay/integrations/nikshay/field_mappings.py
proteusvacuum: 👍

custom/enikshay/integrations/nikshay/repeater_generator.py
proteusvacuum: Yup, correct doc. The case property names changed from when I made it. You can see them all in the form that I had linked there.
The name of the case property is `dst_status`

custom/enikshay/integrations/nikshay/repeater_generator.py
```diff
@@ -337,6 +348,102 @@ def handle_exception(self, exception, repeat_record):
...
+            "Type": disease_classification.get(episode_case_properties.get('pulmonary_extra_pulmonary', ''), ''),
+            "B_diagnosis": episode_case_properties.get('case_definition', ''),
+            "D_SUSTest": drug_susceptibility_test_status.get(episode_case_properties.get('dst', '')),
```
proteusvacuum: this should be `dst_status` as I mentioned in the previous comment.

custom/enikshay/integrations/nikshay/repeater_generator.py
proteusvacuum: I added this as a property in the form so you don't have to do a mapping here. Can leave it as `treatment_initiation_status`

custom/enikshay/integrations/nikshay/repeater_generator.py
```diff
@@ -337,6 +348,102 @@ def handle_exception(self, exception, repeat_record):
...
+            "TBUcode": tu_code,
+            "HFIDNO": person_locations.phi,
+            "pname": sanitize_text_for_xml(person_case.name),
+            "fhname": sanitize_text_for_xml(person_case_properties.get('husband_father_name', '')),
+            "age": person_case_properties.get('age', ''),
+            "gender": person_case_properties.get('sex', '').capitalize(),
+            # API for Address with char ',' returns Invalid data format error
+            "Address": sanitize_text_for_xml(person_case_properties.get('current_address', '').replace(',', '')),
+            "pin": person_case_properties.get('current_address_postal_code', ''),
+            "lno": person_case_properties.get('phone', ''),
```
proteusvacuum: this should be `phone_number`

proteusvacuum (commented):
Looking great. 

Sorry that I wasn't more explicit in the spec doc, I had assumed that you could get all of additional information you needed from the case summary and form itself when building this.

custom/enikshay/integrations/nikshay/repeater_generator.py
```diff
@@ -337,6 +348,102 @@ def handle_exception(self, exception, repeat_record):
                                 "followup_nikshay_registered", "followup_nikshay_error")
 
 
+@RegisterGenerator(NikshayRegisterPrivatePatientRepeater, 'case_xml', 'XML', is_default=True)
```
proteusvacuum: This probably doesn't matter as it isn't going to get shown to the user anywhere, but I'm still hung up on the fact that this payload generator doesn't generate XML, but instead generates a dict to be consumed by the `SOAP` library. @dannyroberts what do you think?

dannyroberts: Yeah, luckily, I think those names are all arbitrary, so you can just change it to, say,

```python
@RegisterGenerator(NikshayRegisterPrivatePatientRepeater, 'episode_soap_data', 'SOAP', is_default=True)
```

and I think that'll be a less surprising name, and it should work just fine.

custom/enikshay/integrations/nikshay/repeater_generator.py
mkangia: This won't fail since the response does not return that format. The response has just a 5 digit code in case of success, otherwise it has an error message. You can check failure and success response expected in the tests [here](https://github.com/dimagi/commcare-hq/pull/16149/files#diff-f2705cd281a2209df30e219fc1e93016R63)

mkangia: Confirmed this. 
Nikshay would return a 5 digit number only which is to be prefixed with health establishment number to get this final nikshay ID `{6 digit number that represents the agency}-{incrementing integer}` and that should be stored for further reference.

mkangia:
@proteusvacuum 

> Sorry that I wasn't more explicit in the spec doc, I had assumed that you could get all of additional information you needed from the case summary and form itself when building this.

I did not assume that mapping on that sheet was to be done. I thought it was just for just documentation. Sorry for the miss. I will keep that in mind for further API integrations, i am sure there will be more.

custom/enikshay/integrations/nikshay/repeater_generator.py
mkangia: changing this to `soap_xml_data` and `SOAP`

mkangia:
> The case property here is episode_case_properties.get('basis_of_diagnosis')
The doc linked to the form where the values are. I didn't list them all out in the document since they were already listed int the form.
If you want to add extra case properties onto the case, feel free. You can call it basis_of_diagnosis_nikshay or something.

Otherwise, here is the mapping:

clinical_chest -> X
clinical_other -> E
microbiological_smear -> S
microbiological_cbnaat -> M
microbiological_culture -> C
microbiological_pcr -> M
microbiological_other -> M

I was looking at case definition for this coz the doc said "Basis of Diagnosis (Now changed to ‘Case definition’ as per new TOG)". This is a simple mapping now.. i thought it was a combination of both. Thank You, will add this.

custom/enikshay/integrations/nikshay/repeaters.py
```diff
@@ -197,6 +188,41 @@ def allowed_to_forward(self, test_case):
             return False
 
 
+class NikshayRegisterPrivatePatientRepeater(SOAPRepeaterMixin, BaseNikshayRepeater):
+
+    payload_generator_classes = (NikshayRegisterPrivatePatientPayloadGenerator,)
```
dannyroberts: Added in merge.

custom/enikshay/integrations/nikshay/repeater_generator.py
```diff
@@ -327,6 +337,104 @@ def handle_exception(self, exception, repeat_record):
                                 "followup_nikshay_registered", "followup_nikshay_error")
 
 
+class NikshayRegisterPrivatePatientPayloadGenerator(BaseNikshayPayloadGenerator):
```
dannyroberts: Removed `@RegisterGenerator` in merage

dannyroberts (commented):
@mkangia I merged https://github.com/dimagi/commcare-hq/pull/16214 today, which changes how we declare payload generators; that conflicted with this branch so I merged master into this PR, and in the process also changed this PR to use the new style.

I left comments by the important changes

mkangia:
Thank You @dannyroberts. Much appreciated 👍

custom/enikshay/integrations/nikshay/field_mappings.py
```diff
@@ -119,3 +119,21 @@
...
+
+basis_of_diagnosis = {
+    'clinical_chest': 'X',
+    'clinical_other': 'E',
+    'microbiological_smear': 'S',
+    'microbiological_cbnaat': 'M',
+    'microbiological_culture': 'C',
+    'microbiological_pcr': 'M',
+    'microbiological_other': 'M',
+}
```
dimagimon: W292 no newline at end of file

custom/enikshay/integrations/nikshay/repeater_generator.py
```diff
@@ -327,6 +337,104 @@ def handle_exception(self, exception, repeat_record):
...
+            "fhname": sanitize_text_for_xml(person_case_properties.get('husband_father_name', '')),
+            "age": person_case_properties.get('age', ''),
+            "gender": person_case_properties.get('sex', '').capitalize(),
+            # API for Address with char ',' returns Invalid data format error
+            "Address": sanitize_text_for_xml(person_case_properties.get('current_address', '').replace(',', '')),
+            "pin": person_case_properties.get('current_address_postal_code', ''),
+            "lno": person_case_properties.get('phone_number', ''),
+            "mno": '0',
+            "tbdiagdate": _format_date(str(episode_date)),
+            "tbstdate": _format_date(episode_case_properties.get(TREATMENT_START_DATE, str(datetime.date.today()))),
```
dimagimon: E501 line too long (116 > 115 characters)

dimagimon: E501 line too long (116 > 115 characters)

mkangia:
@proteusvacuum 
i am not sure how we are adding Health establishment IDs. We need to use that here along with ID (5 digit number) returned by Nikshay to generate the final Nikshay ID. Is it the same as the "HFIDNO" in the payload. Also will that owner be PHI or some other type.. I think i need to update the way i am fetching state, district etc based on who the owner of the person case is.

custom/enikshay/integrations/nikshay/repeater_generator.py
```diff
@@ -327,6 +337,105 @@ def handle_exception(self, exception, repeat_record):
...
+        if episode_case_date:
+            episode_date = datetime.datetime.strptime(episode_case_date, "%Y-%m-%d").date()
+        else:
+            episode_date = datetime.date.today()
+
+        return {
+            "Stocode": person_locations.sto,
+            "Dtocode": person_locations.dto,
+            "TBUcode": tu_code,
+            "HFIDNO": person_locations.phi,
```
proteusvacuum: Hey, this should be from the private sector `pcp` location:
https://github.com/dimagi/commcare-hq/pull/16006/files#diff-f9aa24b679e7af67fd33a5d7a75f1485R290
Sorry I didn't catch that earlier. 

You might want to cherry-pick https://github.com/dimagi/commcare-hq/pull/16006/commits/414731b556eab9b3ed7a9a0bb8c3edaf48c4b457  into your branch, since the other one can't be merged yet because 99DOTS isn't ready:

proteusvacuum:
@mkangia the HFID is the `pcp` Nikshay ID. This is what should be used for the Health Establishment ID to generate the final ID.

mkangia:
Thank you @proteusvacuum.

custom/enikshay/integrations/nikshay/repeater_generator.py
```diff
@@ -24,7 +25,11 @@
     get_person_case_from_occurrence,
     get_lab_referral_from_test)
 from custom.enikshay.integrations.nikshay.exceptions import NikshayResponseException
-from custom.enikshay.exceptions import NikshayLocationNotFound, NikshayRequiredValueMissing
+from custom.enikshay.exceptions import (
```
dimagimon: F401 'NikshayCodeNotFound' imported but unused

custom/enikshay/integrations/nikshay/repeater_generator.py
```diff
@@ -327,6 +338,93 @@ def handle_exception(self, exception, repeat_record):
                                 "followup_nikshay_registered", "followup_nikshay_error")
 
 
+class NikshayRegisterPrivatePatientPayloadGenerator(BaseNikshayPayloadGenerator):
+    format_name = 'case_xml'
+    format_label = 'XML'
+
+    @memoized
```
mkangia: @dannyroberts any thoughts? @proteusvacuum 
From the payload that was sent in request, i need a field(HFIDNO) from it in `handle_success` of the corresponding repeater.
So i have called `repeat_record.get_payload` in handle_success, which eventually calls this and memoized it assuming that it would all happen in one request.
**Pros**: Computation/SQL-Queries skipped, same value as sent was used just in case recomputing gives a different value though highly unlikely

Does it look a good idea? Or it would be harmful?
Should it be done on the parent class?

Other way would be to recompute the value needed or pass the payload as an argument which would need me to update all handle_success.

corehq/apps/repeaters/models.py
```diff
@@ -348,7 +348,7 @@ def __unicode__(self):
         return "forwarding cases to: %s" % self.url
 
 
-class SOAPRepeaterMixin(object):
+class SOAPRepeaterMixin(Repeater):
```
proteusvacuum: why this change?

custom/enikshay/integrations/nikshay/repeater_generator.py
proteusvacuum: what if you just added:
```py
@memoized
def _get_person_locations(self, person_case):
    return person_locations(person_case)
```
to this payload generator? 
Then in your `handle_success` method you'd do:
```py
health_facility_id = self._get_person_locations(person_case).pcp
```
That would serve the same purpose and be a bit less confusing. 

If you go that route, you could also cache the `get_person_from_episode` call in the same way.

mkangia: Thank you @proteusvacuum . Was just thinking of making something more generic which can be used in case payload is needed by any other repeater. That change for just `_get_person_locations` does do the needful.

corehq/apps/repeaters/models.py
mkangia: Inheriting from object didn't let me use the `operation` property. Moving this to `Document` or `Repeater` solved it. `Repeater` seemed better option.

dannyroberts:
```
ERROR: custom.enikshay.integrations.nikshay.tests.test_repeaters:TestNikshayRegisterPrivatePatientPayloadGenerator.test_payload_properties
  File "/mnt/commcare-hq-ro/custom/enikshay/integrations/nikshay/tests/test_repeaters.py", line 1023, in test_payload_properties
    payload = NikshayRegisterPrivatePatientPayloadGenerator(None).get_payload(None, episode_case)
  File "/mnt/commcare-hq-ro/custom/enikshay/integrations/nikshay/repeater_generator.py", line 358, in get_payload
    person_locations = self._get_person_locations(person_case)
  File "submodules/dimagi-utils-src/dimagi/utils/decorators/memoized.py", line 10, in _inner
    return m(*args, **kwargs)
  File "submodules/dimagi-utils-src/dimagi/utils/decorators/memoized.py", line 127, in __call__
  File "/mnt/commcare-hq-ro/custom/enikshay/integrations/nikshay/repeater_generator.py", line 346, in _get_person_locations
    person_case = self._get_person_case(episode_case)
  File "submodules/dimagi-utils-src/dimagi/utils/decorators/memoized.py", line 10, in _inner
    return m(*args, **kwargs)
  File "submodules/dimagi-utils-src/dimagi/utils/decorators/memoized.py", line 127, in __call__
  File "/mnt/commcare-hq-ro/custom/enikshay/integrations/nikshay/repeater_generator.py", line 351, in _get_person_case
    return get_person_case_from_episode(episode_case.domain, episode_case.get_id)
  File "/mnt/commcare-hq-ro/custom/enikshay/case_utils.py", line 99, in get_person_case_from_episode
    get_occurrence_case_from_episode(domain, episode_case_id).case_id
  File "/mnt/commcare-hq-ro/custom/enikshay/case_utils.py", line 86, in get_occurrence_case_from_episode
    return get_first_parent_of_case(domain, episode_case_id, CASE_TYPE_OCCURRENCE)
  File "/mnt/commcare-hq-ro/custom/enikshay/case_utils.py", line 76, in get_first_parent_of_case
    "Couldn't find any {} cases for id: {}".format(parent_case_type, case_id)
ENikshayCaseNotFound: Couldn't find any occurrence cases for id: person
```

mkangia:
Thank You @dannyroberts. Should be fixed now

mkangia:
🎆

dannyroberts:
🎉

PR #16151, opened 2017-05-14
Migration for graph configs in report modules

orangejenny:
Followup for https://github.com/dimagi/commcare-hq/pull/16031

@nickpell


corehq/apps/app_manager/management/commands/migrate_report_app_configs.py
```diff
@@ -0,0 +1,41 @@
+import logging
+import re
+from django.core.management import BaseCommand, call_command
```
dimagimon: F401 'call_command' imported but unused

PR #16152, opened 2017-05-15
Externalize JavaScript: form_view_*.html

orangejenny:
Ended up deleting `form_view.html`, `form_view_advanced.html`, and `form_view_careplan.html` since most of the differences between them were inline javascript, replaced them with a couple of conditionals in `form_view_base.html` (now renamed `form_view.html`).

Review commits individually; renaming form_view_base.html mucked up the overall diff.

@nickpell / @millerdev


millerdev (commented):
❤️ this

PR #16155, opened 2017-05-15
fix after change to DEFAULT_EMPTY_CHECKPOINT_SEQUENCE

snopoke:


corehq/apps/es/management/commands/restore_es_snapshot.py
```diff
@@ -88,6 +88,6 @@ def rewind_pillows(date):
                                                                     date_updated=date)
                 seq = checkpoint.seq
             except HistoricalPillowCheckpoint.DoesNotExist:
-                seq = DEFAULT_EMPTY_CHECKPOINT_SEQUENCE
+                seq = DEFAULT_EMPTY_CHECKPOINT_SEQUENCE[checkpoint.sequence_format]
```
benrudolph: i don't think this works because the checkpoint model in this case is a  `HistoricalPillowCheckpoint`. though i do think you can use `pillow.checkpoint.sequence_format`

snopoke: since this is in the exception handler for the case where the `HistoricalPillowCheckpoint` doesn't exists `checkpoint` should still point to `pillow.checkpoint`. But I agree that it's not clear. I'll update.

snopoke: ceec514

benrudolph (approved):
ah good catch i didn't think about that

benrudolph:
also this command seems like something we'd want tests for. it's one of those things that'll break and we'll never know cause we don't use it. when we do go to use it and it's broken it could cause worse headaches

benrudolph:
(most likely out of scope for this PR)

PR #16170, opened 2017-05-15
Use urllib.js on deploy page

orangejenny:
Now we have a standard way to do this kind of thing!

@dannyroberts / @nickpell

### DIFF
--- a/corehq/apps/app_manager/static/app_manager/js/releases.js
+++ b/corehq/apps/app_manager/static/app_manager/js/releases.js
@@ -143,8 +143,8 @@ hqDefine('app_manager/js/releases.js', function () {
         };
 
         self.get_odk_install_url = ko.computed(function() {
-            var slug = self.include_media() ? 'odk_media' : 'odk';
-            return releasesMain.url(slug, self.id());
+            var slug = self.include_media() ? 'odk_media_install' : 'odk_install';
+            return releasesMain.reverse(slug, self.id());
         });
 
         self.full_odk_install_url = ko.computed(function() {
@@ -162,7 +162,7 @@ hqDefine('app_manager/js/releases.js', function () {
                 }
             }
         };
-        
+
         self.download_application_zip = function (multimedia_only, build_profile) {
             releasesMain.download_application_zip(self.id(), multimedia_only, build_profile);
         };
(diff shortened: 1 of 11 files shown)


corehq/apps/app_manager/static/app_manager/js/releases.js
```diff
@@ -253,13 +253,15 @@ hqDefine('app_manager/js/releases.js', function () {
...
             for (var i = 1; i < arguments.length; i++) {
-                template = template.replace('___', ko.utils.unwrapObservable(arguments[i]));
+                arguments[i] = ko.utils.unwrapObservable(arguments[i]);
             }
-            return template;
+            return hqImport("hqwebapp/js/urllib.js").reverse.apply(null, arguments);
         };
+        self.app_error_url = function(app_id, version) {
+            return self.reverse('project_report_dispatcher') + '?app=' + app_id + '&version_number=' + version;
+        }
```
dimagimon: Missing semicolon. (semi)

corehq/apps/app_manager/static/app_manager/js/releases.js
```diff
@@ -253,12 +253,14 @@ hqDefine('app_manager/js/releases.js', function () {
             var lastApp = self.savedApps()[0];
             self.lastAppVersion(lastApp ? lastApp.version() : -1);
         });
-        self.url = function (name) {
-            var template = self.options.urls[name];
+        self.reverse = function() {
             for (var i = 1; i < arguments.length; i++) {
-                template = template.replace('___', ko.utils.unwrapObservable(arguments[i]));
+                arguments[i] = ko.utils.unwrapObservable(arguments[i]);
```
dannyroberts: Ha! That's kind of annoying that it does that unwrap thing rather than having the caller know whether their object needs to be unwrapped. I suppose you could push this out and just add in a `ko.utils.unwrapObservable` call wherever it's ambiguous (and there are many places where it's obvious, like when the arg is a string), and then you wouldn't need this self.reverse function at all—but probably you thought of that and thought against it as just not that much better for the amount of hassle, and I won't fault you for that :) Does just seem really pesky either way.

dannyroberts (approved):
Nice! Using conventions. Looks great

nickpell (approved):

corehq/apps/app_manager/static/app_manager/js/releases.js
orangejenny: It does seem kind of annoying. I think I'd still end up having a function here, though, if only because of the places where `reverse` is called from within a `data-bind`, it seems clunky to do an entire `hqImport(...).reverse(...)` in html. So I'm inclined to leave it alone.

dannyroberts: 👍

PR #16174, opened 2017-05-15
initialize checkpoint to 0

emord:
May or may not fix the issue @nickpell ran into on softlayer today 

Old pillows would initalize checkpoints to 0, so I think the pillow should fix the  checkpoint if necessary. Still need to do mroe testing though https://github.com/dimagi/commcare-hq/blob/je/pillowcheckpoints/corehq/ex-submodules/pillowtop/checkpoints/manager.py#L26

### DIFF
--- a/corehq/ex-submodules/pillowtop/models.py
+++ b/corehq/ex-submodules/pillowtop/models.py
@@ -96,7 +96,7 @@ class KafkaCheckpoint(models.Model):
         for tp, offset in all_offsets.items():
             if tp not in already_created:
                 to_create.append(
-                    cls(checkpoint_id=checkpoint_id, topic=tp[0], partition=tp[1], offset=offset)
+                    cls(checkpoint_id=checkpoint_id, topic=tp[0], partition=tp[1], offset=0)
                 )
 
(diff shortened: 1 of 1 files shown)


emord:
should be safe to merge after tests run

buddy @dannyroberts

snopoke:
Could you give more info about the issue that @nickpell had.

emord:
@snopoke The following happened on pillows that are in settings.py but aren't used. So that's kafka-ucr-main-08 and kafka-ucr-main-9f on softlayer and kafka-ucr-main on prod

whenever you visit the hq/admin/system page it calls `get_or_create_checkpoint` on the pillow. for ucr pillows that now use `KafkaCheckpoint` it would be initialized to the earliest offsets. That caused `validate_kafka_checkpoints` to fail on deploy because those early offsets would be deleted from kafka. 

When that checkpoint is 0 then the check does not fail and if the pillow is started at a later date it resets the checkpoint to the earliest offset available.

`DjangoPillowCheckpoint` initializes checkpoints to either {} or 0 so this is consistent witht hat behvaior

snopoke:
👍 thanks for the explanation

PR #16180, opened 2017-05-16
Added Worker Activity Report

kkrampa:
@benrudolph @snopoke @sheelio 
http://manage.dimagi.com/default.asp?252766

### DIFF
--- a/custom/enikshay/reports/worker_activity.py
+++ b/custom/enikshay/reports/worker_activity.py
@@ -1,0 +1,55 @@
+from corehq.apps.es import UserES
+from corehq.apps.es.users import mobile_users
+from corehq.apps.reports.analytics.esaccessors import get_submission_counts_by_user
+from corehq.apps.reports.datatables import DataTablesHeader, DataTablesColumn, DTSortType
+from corehq.apps.reports.filters.dates import DatespanFilter
+from corehq.apps.reports.standard import CustomProjectReport
+from corehq.apps.reports.standard.monitoring import WorkerActivityReport
+from corehq.apps.reports.util import get_simplified_users, numcell
+from custom.enikshay.reports.filters import EnikshayLocationFilter
+
+from django.utils.translation import ugettext as _
+
+
+class EnikshayWorkerActivityReport(WorkerActivityReport, CustomProjectReport):
+    slug = 'enikshay_worker_activity_report'
+    num_avg_intervals = 4
+
+    @property
+    def fields(self):
+        return DatespanFilter, EnikshayLocationFilter
+
+    @property
+    def headers(self):
+        return DataTablesHeader(
+            DataTablesColumn(_("User")),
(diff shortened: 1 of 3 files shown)


custom/enikshay/reports/worker_activity.py
```diff
@@ -0,0 +1,55 @@
...
+        user_query = UserES().domain(self.domain).filter(mobile_users())
+        if location_ids:
+            user_query = user_query.location(location_ids)
+        return get_simplified_users(user_query)
+
+    @property
+    def rows(self):
+        rows = []
+        submissions_by_user = get_submission_counts_by_user(self.domain, self.datespan)
+        avg_submissions_by_user = get_submission_counts_by_user(self.domain, self.avg_datespan)
```
snopoke: Running these queries over the whole domain is going to be prohibitively expensive. You should add in the `user_ids` kwarg to limit it to only those users being displayed.

snopoke (changes requested):
This report will need to be paginated since it's going to be used to report on 100k users.

kkrampa:
@snopoke updated

custom/enikshay/reports/worker_activity.py
```diff
@@ -0,0 +1,87 @@
...
+        )
+
+    @property
+    def locations_id(self):
+        return EnikshayLocationFilter.get_value(self.request, self.domain)
+
+    @property
+    def user_query(self):
+        user_query = UserES().domain(self.domain).filter(mobile_users())
+        location_ids = self.locations_id
```
benrudolph: nit: weird that this is called `location_id` and then you set it to var called `location_ids`

custom/enikshay/reports/worker_activity.py
```diff
@@ -0,0 +1,87 @@
...
+        return user_query
+
+    @property
+    @memoized
+    def users(self):
+        return get_simplified_users(
+            self.user_query
+                .start(self.pagination.start)
+                .size(self.pagination.count)
+                .sort('username', desc=self.pagination.desc)
```
benrudolph: guessing this report will have to be able to be sorted on other columns. not a super easy way to do that at the moment

snopoke: I don't think you could sort on anything other than username and still do performant pagination since ES doesn't support pagination on aggregations.

snopoke (approved):
pending Ben's comment on variable name

kkrampa:
@snopoke @benrudolph Updated

custom/enikshay/reports/worker_activity.py
benrudolph: i agree, just flagging that this most likely won't be that useful to the team once they have thousands of users

benrudolph: they can always filter the results down by a lot though

snopoke: One approach could be to check the total number of users selected and if it's below some threshold then we just pull all the data for those users and sort in python.

PR #16186, opened 2017-05-16
use named values for translatable string

nickpell:
@millerdev code buddy


millerdev (approved):

nickpell:
rebased on master because of failing tests

PR #16189, opened 2017-05-16
Add custom UCR expressions for enikshay Referral Report

NoahCarnahan:
To be used in a UCR I'm building for enikshay.
@sravfeyn @proteusvacuum


custom/enikshay/tests/utils.py
```diff
@@ -332,6 +351,11 @@ def create_voucher_case(self, prescription_id, extra_update=None):
         )[0]
 
 
+    def create_referral_case(self, case_id):
```
dimagimon: E303 too many blank lines (2)

settings.py
```diff
@@ -1935,6 +1935,10 @@ def _determine_couch_databases(couch_databases):
     ('first_case_form_with_xmlns', 'custom.enikshay.expressions.first_case_form_with_xmlns_expression'),
     ('count_case_forms_with_xmlns', 'custom.enikshay.expressions.count_case_forms_with_xmlns_expression'),
     ('month_expression', 'custom.enikshay.expressions.month_expression'),
+    ('referred_to', 'custom.enikshay.expressions.referred_to_expression'),
+    ('referred_by', 'custom.enikshay.expressions.referred_by_expression'),
+    ('date_of_referral', 'custom.enikshay.expressions.date_of_referral_expression'),
+    ('date_of_acceptance', 'custom.enikshay.expressions.referred_to_expression'),
```
proteusvacuum: should this be `date_of_acceptance`?

custom/enikshay/expressions.py
```diff
@@ -119,6 +121,113 @@ def concatenate_strings_expression(spec, context):
...
+
+    def _handle_referral_case(self, referral):
+        raise NotImplementedError
+
+    def _handle_trail_case(self, trail, domain):
+        raise NotImplementedError
+
+
+class ReferredBy(ReferralExpressionBase):
+    type = TypeProperty("referred_by")
```
proteusvacuum: do we want to namespace these custom properties? `enikshay_referred_by` or something?

custom/enikshay/case_utils.py
```diff
@@ -170,6 +172,39 @@ def get_open_episode_case_from_person(domain, person_case_id):
...
+
+def get_latest_trail_case_from_person(domain, person_case_id):
+    episode = get_open_episode_case_from_person(domain, person_case_id)
+    case_accessor = CaseAccessors(domain)
+    reverse_indexed_cases = case_accessor.get_reverse_indexed_cases([episode.case_id])
+    trail_cases = [
+        case for case in reverse_indexed_cases
+        if case.type == CASE_TYPE_TRAIL
+    ]
+    trail_cases.sort(key=lambda c: c.opened_on)
```
proteusvacuum: wow, it would be nice if we could do a query into the case_json at some point so we can avoid doing things like this...

benrudolph: both opened_on and case type aren't stored in the case_json so it'd definitely be feasible to write a query

custom/enikshay/expressions.py
```diff
@@ -119,6 +121,113 @@ def concatenate_strings_expression(spec, context):
...
+    def configure(self, person_id_expression):
+        self._person_id_expression = person_id_expression
+
+    def __call__(self, item, context=None):
+        person_id = self._person_id_expression(item, context)
+        domain = context.root_doc['domain']
+        if not person_id:
+            return None
+
+        referral = get_open_referral_case_from_person(domain, person_id)
```
emord: Are these going to be called multiple times or exist in multiple data sources? It may be worth caching them

settings.py
NoahCarnahan: good catch thanks

custom/enikshay/expressions.py
NoahCarnahan: That's not currently the pattern we've been using, but I imagine at some point we will want some sort of namespacing. I guess it can't hurt to change now.

custom/enikshay/case_utils.py
NoahCarnahan: What do you think of this for the a new case accessors method?
```python
    @staticmethod
    def get_newest_reverse_indexed_case(domain, case_ids, case_types):
        raise NotImplementedError
```
I'll probably also add `case_types` as an arg to `get_reverse_indexed_cases()`

Alternatively we could go very generic and add a `limit` and `order_by` param to `get_reverse_indexed_cases() `. Thoughts?

benrudolph: maybe have the sql function be parameterized by an order column since it's a pain to write those, but having a method that doesn't require you to pass in the order_by would be nice. minor nit: i think django normally uses `latest` rather than `newest`

benrudolph: it also may mean we need a btree index on `opened_on` in order for it to be performant

NoahCarnahan: Thanks for the feedback Ben.
I'm thinking it might not be worth adding the column index. For one, I don't think this will be a super common query, as it's specific to one project, and also I think the number of extension cases that need to be sorted will be less than 100, so doesn't seem like the index would provide such a big improvement.
Also, without the ordered_by index we will still have the advantage of returning less content, and also making the subsequent `get_multiple_cases_indices` call smaller.

Another thing to consider is that the couch version of this method will have to do all the filtering and sorting in the application layer, unless we change the couch view. Do you think it's misleading to add parameters to CaseAccessor functions that won't be used at the db level?

custom/enikshay/expressions.py
NoahCarnahan: Yeah these will be called multiple times per document. Added cacheing.

custom/enikshay/case_utils.py
benrudolph: ah that's good context. didn't realize it was less than 100. i meant add a parameter to the actual sql function. see `get_forms_by_state.sql` for an example. you could potentially make a function that is get_ordered_cases.sql that accepts a column name to order by. then the CaseAccessor would pass in `opened_on`. then the couch and sql version would accept the same params. 

it's also probably fine to just hardcode opened_on in the sql function. i've just rewritten a few of those sql functions wishing i had made them more flexible. it doesn't seem super likely we'll ever need to order cases by anything else (maybe modified_on?)

custom/enikshay/expressions.py
```diff
@@ -133,14 +133,35 @@ def __call__(self, item, context=None):
...
+        context.set_cache_value(cache_key, referral)
+        return referral
+
+    @staticmethod
+    def _get_trail(context, domain, person_id):
+        cache_key = (ReferralExpressionBase.__name__, "_get_trail", person_id)
+
+        if context.get_cache_value(cache_key, False) is not False:
```
proteusvacuum: curious why you needed to add the `False` part? couldn't you just also have checked for `None` here?

NoahCarnahan: Ah yeah meant to explain that. The cached value could be `None` (because `get_latest_trail_case_from_person()` can return `None`), so I used `False` to indicate "we have not attempted to calculate this value yet".

PR #16192, opened 2017-05-16
Remove TestTemplateAppViews

orangejenny:
App manager v2, which doesn't use template apps, is now on for all new users, so these tests are failing. Template apps are going to be killed once v2 is rolled out, so I'm not worried about losing these tests.

code buddy @nickpell


nickpell (approved):
Seems fine to merge now if tests are failing already

PR #16198, opened 2017-05-17
Nh/253389/excel2016

kaapstorm:
Context: [FB 253389](https://manage.dimagi.com/default.asp?253389)

Although previous versions of Excel work fine, Excel 2016 is not convinced that the heading which we don't want in the dashboard, and the table which we do want, are separate, *even though* when you create the dashboard in Excel 2016 (as with previous versions) it allows you to select the table and exclude the heading.

This is an attempt to convince Excel 2016 to exclude the heading.

I don't have Excel 2016 readily available. Do not merge until it has been tested successfully. I'll follow up on testing in FogBogz, and drop the label on success.

@NoahCarnahan buddy @snopoke


corehq/ex-submodules/couchexport/templates/couchexport/html_export.html
```diff
@@ -8,8 +8,17 @@
 {% endif %}
 
 {% if section == "table_begin" %}
-    <h2>{{ name }}</h2>
-    <table>
+    <div>
+      <table>
+        <tr>
+          <td>
+      <h2>{{ name }}</h2>
```
NoahCarnahan: Indentation is a little weird but non-blocking

NoahCarnahan (approved):

corehq/ex-submodules/couchexport/templates/couchexport/html_export.html
kaapstorm: Sometimes I try to keep the diff as small as possible, and sometimes my OCD takes over and reformats huge chunks. One day I'll be able to find a balance. :balance_scale:

kaapstorm:
Excel 2016 does not take the hint. :nauseated_face: It allows you to select the table and exclude the heading, as before ... and then it ignores your instructions and imports everything anyway, as before. 

Back to the drawingboard.

PR #16202, opened 2017-05-17
Match base class method signature

NoahCarnahan:
`AbstractCaseAccessor. get_case_ids_in_domain()` doesn't have a deleted kwarg, and this method doesn't do anything with it anyways.
@dannyroberts


dannyroberts (approved):
legit

PR #16204, opened 2017-05-17
Staging tables

benrudolph:
@gcapalbo first go at an example for group staging. going to have to do some work to get all the source models to have last modified. also a lot of our models have wide ranging data since we've never been too strict about data constraints. i think i can iterate to make better abstractions when i start implementing more of the staging tables. hold off on merge, want to get `br/warehouse` in first

cc: @proteusvacuum


corehq/couchapps/group_last_modified/views/by_last_modified/map.js
```diff
@@ -0,0 +1,5 @@
+function(doc){
```
dimagimon: Parsing error: Unexpected token (

corehq/warehouse/models/__init__.py
```diff
@@ -10,3 +10,7 @@
 from corehq.warehouse.models.meta import (
     TableState,
 )
+
+from corehq.warehouse.models.staging import (
```
dimagimon: F401 'GroupStagingTable' imported but unused

corehq/warehouse/models/staging.py
```diff
@@ -0,0 +1,64 @@
...
+
+from dimagi.utils.couch.database import iter_docs
+
+from corehq.apps.groups.dbaccessors import get_group_ids_by_last_modified
+from corehq.apps.groups.models import Group
+
+
+class StagingTable(models.Model):
+
+    staged_on = models.DateTimeField(auto_now_add=True)
```
gcapalbo: just curious where this value would end up? i think we'd want to include batch id on this table as well

corehq/warehouse/models/staging.py
```diff
@@ -0,0 +1,64 @@
...
+
+    @classmethod
+    def raw_record_iter(cls, start_datetime, end_datetime):
+        group_ids = get_group_ids_by_last_modified(start_datetime, end_datetime)
+
+        return iter_docs(Group.get_db(), group_ids)
+
+    @classmethod
+    def clear_records(cls):
+        cls.objects.all().delete()
```
gcapalbo: i think ideally we'd want to make this a truncate so it's faster, but that also means using raw sql

corehq/warehouse/models/staging.py
```diff
@@ -0,0 +1,64 @@
...
+        records = []
+        for raw_record in cls.raw_record_iter(start_dateime, end_datetime):
+            records.append(GroupStagingTable(
+                group_id=raw_record.get('_id'),
+                name=raw_record.get('name'),
+                case_sharing=raw_record.get('case_sharing'),
+                reporting=raw_record.get('reporting'),
+                group_last_modified=raw_record.get('last_modified'),
+            ))
+        cls.objects.bulk_create(records)
```
gcapalbo: this is a good idea. for the bigger dims we may want to do 1000 inserts at a time or something like that

corehq/warehouse/management/commands/stage_table.py
```diff
@@ -0,0 +1,60 @@
...
+            help='Specifies the last modified datetime at which records should stop being included',
+            type=_valid_date
+        )
+
+    def handle(self, slug, **options):
+        start = options.get('start')
+        end = options.get('end')
+
+        model = SLUG_TO_STAGING_TABLE.get(slug)
+        model.stage_records(start, end)
```
gcapalbo: i forgot to mention this in our previous chats but it's a best practice any time we're moving data from one step to another to do a balancing step at the end where we make sure the count of records in the source matches the count of records in the destination, and if that doesn't match we should fail the process

corehq/warehouse/models/staging.py
benrudolph: yep that sounds right. i was reading this post too: http://stefano.dissegna.me/django-pg-bulk-insert.html. purposely didn't do anything too crazy, but may be worth looking into some of the ideas proposed in that article (using postgres `COPY` or using raw SQL) in addition to batching

corehq/warehouse/management/commands/stage_table.py
benrudolph: 👍 good call. in some of the example etl setups i've seen while googling, they have an entire task in the progress called data validation, that does checks like that to ensure that the staged data is ready to be moved to the production tables. probably a bit overkill right now, but something to consider

corehq/warehouse/models/staging.py
benrudolph: this is actually inherited by all tables that are staging tables. so for each group record in the staging table it will also include the date that it was added. just thought it might be useful meta data. will add a column for batch id, but probably won't address it in this pr until i sort out how the batch state will manifest

corehq/warehouse/models/staging.py
benrudolph: https://github.com/dimagi/commcare-hq/pull/16204/commits/d9185d2d5b111d74692d96bacf0f87d6cf812281

corehq/warehouse/models/staging.py
gcapalbo: ok, i think batch id might replace this data element since we would know when the batch ran

gcapalbo (approved):

gcapalbo:
looks like `test_get_group_ids_by_last_modified` is failing, maybe because other tests don't clean up the groups they create

gcapalbo:
one thing that just occurred to me is we'll need to figure out how to deal with deleted data. we might want to maintain deleted groups, for example, with a deleted flag

benrudolph:
we could also just make it another staging table and iterate through all `Group-Deleted` items (`X-Deleted`) and then update the dim accordingly.

deleted data is going to be much harder to deal with for case or form deletes if we've aggregated the data for fact tables...

corehq/couchapps/last_modified/views/by_last_modified/map.js
```diff
@@ -0,0 +1,7 @@
+function(doc){
```
dimagimon: Parsing error: Unexpected token (

corehq/apps/groups/tests/test_groups.py
```diff
@@ -1,3 +1,4 @@
+from datetime import datetime, timedelta
```
dimagimon: F401 'datetime' imported but unused
F401 'timedelta' imported but unused

corehq/warehouse/models/__init__.py
```diff
@@ -10,3 +10,8 @@
...
```
dimagimon: F401 'DomainStagingTable' imported but unused
F401 'GroupStagingTable' imported but unused

benrudolph:
Added the domain staging table and added some more tests and a few more abstractions. i think g2g once we are happy with it. enough for now, going to look at making an `update_warehouse_table` that pulls from the staging tables that are necessary, and upserts them into the dim/fact table

corehq/warehouse/models/__init__.py
dimagimon: F401 'DomainStagingTable' imported but unused
F401 'GroupStagingTable' imported but unused

corehq/warehouse/models/staging.py
```diff
@@ -15,12 +16,16 @@ class Meta:
...
+    def raw_record_iter(cls, start_datetime, end_datetime):
         raise NotImplementedError
 
     @classmethod
-    def raw_record_iter(cls, start_datetime, end_datetime):
-        raise NotImplementedError
+    @transaction.atomic
+    def stage_records(cls, start_dateime, end_datetime):
+        cls.clear_records()
+        record_iter = cls.raw_record_iter(start_dateime, end_datetime)
```
gcapalbo: `start_datetime`

corehq/warehouse/models/staging.py
```diff
@@ -40,21 +45,15 @@ class GroupStagingTable(StagingTable):
...
-                group_id=raw_record.get('_id'),
-                name=raw_record.get('name'),
-                case_sharing=raw_record.get('case_sharing'),
-                reporting=raw_record.get('reporting'),
-                group_last_modified=raw_record.get('last_modified'),
-            ))
-        cls.objects.bulk_create(records)
+    # Map source model fields to staging table fields
+    # ( <source field>, <staging field> )
+    FIELD_MAPPING = [
```
gcapalbo: it might be a good idea to make this a property or be returned from a classmethod, so that we can put it on the superclass and raise a NotImplementError, just so it's clear what needs to be implemented on a new staging table model

corehq/couchapps/last_modified/views/by_last_modified/map.js
```diff
@@ -1,5 +1,7 @@
 function(doc){
     if (doc.doc_type == "Group") {
         emit(doc.last_modified, null);
+    } else if (doc.doc_type == "Domain") {
+        emit(doc.last_modified, null);
```
gcapalbo: we need to emit doc type now right, and filter for it in the staging process

gcapalbo: (i realize this might be working right now because they're in different dbs, but i don't think we want to rely on that long term)

gcapalbo:
i think the right way to handle the deleted data for these two dims would actually be to pull in doc type on the staging table, and then transform it to a boolean deleted flag on the dim later on, that way we pull it all at once and don't have to duplicate tables.

agree it's going to be more complicated for the facts when dealing with deleted forms.

benrudolph:
> i think the right way to handle the deleted data for these two dims would actually be to pull in doc type on the staging table, and then transform it to a boolean deleted flag on the dim later on, that way we pull it all at once and don't have to duplicate tables.

couch is annoying in that it'll take multiple queries to get the set that we want, but agree i like the idea of transforming the doc_type

corehq/couchapps/last_modified/views/by_last_modified/map.js
benrudolph: 👍 will add users now as well so we can avoid the additional reindex

corehq/warehouse/models/staging.py
benrudolph: 👍 will do. i added it as a required param to the bulk create method method, but i think explicitly requiring it in the abstract class is better

corehq/warehouse/models/__init__.py
dimagimon: F401 'DomainStagingTable' imported but unused
F401 'GroupStagingTable' imported but unused

corehq/couchapps/last_modified/views/by_last_modified/map.js
```diff
@@ -1,7 +1,12 @@
...
-    } else if (doc.doc_type == "Domain") {
-        emit(doc.last_modified, null);
+    if (doc.doc_type == "Group" || doc.doc_type == "Group-Deleted") {
+        emit([doc.doc_type, doc.last_modified], null);
+    } else if (doc.doc_type == "Domain" || doc.doc_type == "Domain-Deleted") {
+        emit([doc.doc_type, doc.last_modified], null);
+    } else if (doc.doc_type == "WebUser" ||
+               doc.doc_type == "WebUser-Deleted" ||
+               doc.doc_type == "CommCareUser" ||
+               doc.doc_type == "CommCareUser-Deleted") {
```
benrudolph: potentially could use base_doc_type here

corehq/warehouse/models/__init__.py
dimagimon: F401 'DomainStagingTable' imported but unused
F401 'GroupStagingTable' imported but unused

dimagimon: F401 'DomainStagingTable' imported but unused
F401 'GroupStagingTable' imported but unused

dimagimon: F401 'DomainStagingTable' imported but unused
F401 'GroupStagingTable' imported but unused

dimagimon: F401 'DomainStagingTable' imported but unused
F401 'GroupStagingTable' imported but unused

corehq/couchapps/last_modified/views/by_last_modified/map.js
gcapalbo: actually we'll have to use base doc, at least for the deleted suffix: https://github.com/dimagi/commcare-hq/blob/8063cd3ff1f889d58a997c00557be092bda8e214/corehq/apps/users/models.py#L1611

benrudolph: ah indeed

gcapalbo:
just one last comment about base_doc for users, otherwise this is looking good

corehq/warehouse/models/__init__.py
dimagimon: F401 'DomainStagingTable' imported but unused
F401 'GroupStagingTable' imported but unused

dimagimon: F401 'DomainStagingTable' imported but unused
F401 'GroupStagingTable' imported but unused

dimagimon: F401 'DomainStagingTable' imported but unused
F401 'GroupStagingTable' imported but unused

dimagimon: F401 'DomainStagingTable' imported but unused
F401 'GroupStagingTable' imported but unused

corehq/couchapps/last_modified/views/by_last_modified/map.js
```diff
@@ -1,12 +1,9 @@
...
         emit([doc.doc_type, doc.last_modified], null);
-    } else if (doc.doc_type == "Domain" || doc.doc_type == "Domain-Deleted") {
-        emit([doc.doc_type, doc.last_modified], null);
-    } else if (doc.doc_type == "WebUser" ||
-               doc.doc_type == "WebUser-Deleted" ||
-               doc.doc_type == "CommCareUser" ||
-               doc.doc_type == "CommCareUser-Deleted") {
+    } else if (doc.doc_type === "Domain" || doc.doc_type === "Domain-Deleted") {
         emit([doc.doc_type, doc.last_modified], null);
+    } else if (doc.base_doc && (doc.base_doc === "CouchUser" || doc.doc_type === "CouchUser-Deleted")) {
```
gcapalbo: probably just a typo but we'll want to check doc.base_doc for the deleted type too

corehq/warehouse/models/__init__.py
dimagimon: F401 'DomainStagingTable' imported but unused
F401 'GroupStagingTable' imported but unused

corehq/couchapps/last_modified/views/by_last_modified/map.js
benrudolph: :man_facepalming: clearly can't write code when there are no tests. making some for when i create the user staging table

corehq/couchapps/last_modified/views/by_last_modified/map.js
```diff
@@ -0,0 +1,9 @@
+function(doc){
+    if (doc.doc_type === "Group" || doc.doc_type === "Group-Deleted") {
+        emit([doc.doc_type, doc.last_modified], null);
+    } else if (doc.doc_type === "Domain" || doc.doc_type === "Domain-Deleted") {
+        emit([doc.doc_type, doc.last_modified], null);
```
snopoke: Could be simpler to use a switch: https://github.com/dimagi/commcare-hq/blob/db598e5b053c741f753cdb1238500bd4aca7b83a/corehq/couchapps/by_domain_doc_type_date/views/view/map.js#L4-L4

gcapalbo (approved):

corehq/couchapps/last_modified/views/by_last_modified/map.js
benrudolph: something tells me i'm going to have to revisit this view... i think i'm going to opt to merge this now and when i touch this again, i will update to use a switch. there are a couple things i'd like to work on and PR independently that depend on this PR

PR #16206, opened 2017-05-17
App manager v2: missing feature flagged content

orangejenny:
https://github.com/dimagi/commcare-hq/pull/16203 inspired me to skim through the v2 diffs looking for any other features that might have slipped through the cracks. Just found two small things: the ability to add graphs to case list/detail, and the form linking workflow. Also dropped a couple of diff files for templates that no longer exist.

@biyeun / @esoergel


corehq/apps/app_manager/templates/app_manager/v2/partials/form_workflow.html
```diff
@@ -96,25 +94,25 @@ <h4 class="panel-title panel-title-nolink">
...
-            <div class="col-sm-12" data-bind="visible: formLink.manualDatums">
+            <div class="col-sm-11" data-bind="visible: formLink.manualDatums">
                 <span data-bind="click: disableManualDatums" class="btn btn-danger btn-sm">
                     {% trans "Disable manual linking" %}
                 </span>
             </div>
         </div>
         <div class="row">
-            <div class="col-sm-12"  data-bind="visible: formLink.showLinkDatums()">
+            <div class="col-sm-11"  data-bind="visible: formLink.showLinkDatums()">
```
biyeun: where's the sm-1 column complementing this? why did you decide to go with sm-12?

orangejenny: sm-12 causes the dropdown to spill out of the container (there's extra left margin on a select inside of a .row inside of a .form-group, didn't dig into why).

The alignment still isn't ideal, but I don't think it's any worse than in v1.

v1
screen shot 2017-05-17 at 2 22 31 pm

v2
screen shot 2017-05-17 at 2 22 39 pm

orangejenny:
@biyeun is this good to merge?

corehq/apps/app_manager/tests/data/v2_diffs/templates/form_view.html.diff.txt
```diff
@@ -7,48 +7,24 @@
...
-         {% endif %}
- 
--        {% if allow_cloudcare %}
-+        {% if can_preview_form and allow_cloudcare %}
-             // tag the 'preview in cloudcare' button with the right url
-             // unfortunately, has to be done in javascript
-             var getCloudCareUrl = function(urlRoot, appId, moduleId, formId, caseId) {
-@@ -291,75 +294,58 @@
-     </script>
+ {% if request.guided_tour %}{% include request.guided_tour.template %}{% endif %}
```
biyeun: caution here. the guided tour is going to be different for v2. we shouldn't just pull from the guided tours that exist now because they won't point to the same places

biyeun:
@orangejenny had a question about guided tour content

corehq/apps/app_manager/tests/data/v2_diffs/templates/form_view.html.diff.txt
orangejenny: we handle which guided tour is provided on the server side: https://github.com/dimagi/commcare-hq/blob/master/corehq/apps/app_manager/views/forms.py#L655-L656

biyeun (approved):

PR #16208, opened 2017-05-17
Use unique form id more often in urls

orangejenny:
Haven't nailed down the cause of http://manage.dimagi.com/default.asp?254212 yet, but one way this kind of thing happens is by using index-based form and module IDs, which can cause problems when items are being edited and re-ordered at the same time. Don't think that's the cause of the bug, but might as well reduce that risk, at least in urls that are only used to POST.

@esoergel / @NoahCarnahan


corehq/apps/app_manager/urls.py
```diff
@@ -70,7 +70,7 @@
 urlpatterns = [
     url(r'^$', view_app, name='default_app'),
     url(r'^xform/(?P<form_unique_id>[\w-]+)/$', xform_display, name='xform_display'),
-    url(r'^browse/(?P<app_id>[\w-]+)/modules-(?P<module_id>[\w-]+)/forms-(?P<form_id>[\w-]+)/source/$',
+    url(r'^browse/(?P<app_id>[\w-]+)/(?P<form_unique_id>[\w-]+)/source/$',
```
emord: I think this will break for anyone who has tab open during deploy, and they'll just get 400/500s if they try to do anything without reloading. It's probably worth having some way to support both endpoints for the next couple of days

orangejenny: They'll get an error if they have form view open before deploy and then after deploy try to either download the form or view it in a popup; it doesn't break the whole page. But you're right that there's no reason not to support both.

https://github.com/dimagi/commcare-hq/pull/16208/commits/956d76675a5afd1464acbccb254de88e75aa9021

emord: I actually meant that as a general comment for all of these. Endpoints like `edit_form_actions` would give anyone editing an app a weird error when they try to save and lose any case management changes.

corehq/apps/app_manager/views/schedules.py
```diff
@@ -39,10 +39,17 @@ def edit_schedule_phases(request, domain, app_id, module_id):
...
 @require_can_edit_apps
-def edit_visit_schedule(request, domain, app_id, module_id, form_id):
+def edit_visit_schedule_by_form_index(request, domain, app_id, module_id, form_id):
     app = get_app(domain, app_id)
-    module = app.get_module(module_id)
-    form = module.get_form(form_id)
+    form = app.get_module(module_id).get_form(form_id)
+    return edit_visit_schedule(request, domain, app_id, form.unique_id)
+
+@no_conflict_require_POST
```
dimagimon: E302 expected 2 blank lines, found 1

corehq/apps/app_manager/urls.py
orangejenny: fixed

NoahCarnahan (commented):
Nice, I like these changes.

NoahCarnahan (approved):
(meant to approve)

PR #16213, opened 2017-05-17
Rename unique_form_id

orangejenny:
The inconsistency was bothering me when working on https://github.com/dimagi/commcare-hq/pull/16208  We use `form_unique_id` more.

@esoergel


orangejenny:
nevermind, i'm going to add this on top of https://github.com/dimagi/commcare-hq/pull/16208 otherwise it's going to conflict

PR #16214, opened 2017-05-17
Simplify repeater payload generators

dannyroberts:
@proteusvacuum this is kind of what we'd talked about.

This PR just simplifies the payload configuration so it's inside the repeater rather than registered through a decorator, but it also retains support for the decorator-style matching. (~If you try to use both in the same repeater though, that behavior's not well defined.~ Update: Ended up supporting this for test.) In a later PR I can gut the code around this which I think can be greatly simplified as well.

### DIFF
--- a/corehq/apps/repeaters/models.py
+++ b/corehq/apps/repeaters/models.py
@@ -68,6 +68,12 @@ class Repeater(QuickCachedDocumentMixin, Document, UnicodeMixIn):
     password = StringProperty()
     friendly_name = _("Data")
 
+    _generators = None
+
+    @classmethod
+    def get_generators(cls):
+        return cls._generators
+
     @classmethod
     def get_custom_url(cls, domain):
         return None
@@ -260,8 +266,7 @@ class FormRepeater(Repeater):
 
     """
 
-    class Formats(object):
-        formats = [FormRepeaterXMLPayloadGenerator, FormRepeaterJsonPayloadGenerator]
+    _generators = [FormRepeaterXMLPayloadGenerator, FormRepeaterJsonPayloadGenerator]
 
     include_app_id_param = BooleanProperty(default=True)
     white_listed_form_xmlns = StringListProperty(default=[])  # empty value means all form xmlns are accepted
@@ -309,8 +314,7 @@ class CaseRepeater(Repeater):
 
     """
 
-    class Formats(object):
-        formats = [CaseRepeaterXMLPayloadGenerator, CaseRepeaterJsonPayloadGenerator]
+    _generators = [CaseRepeaterXMLPayloadGenerator, CaseRepeaterJsonPayloadGenerator]
 
     version = StringProperty(default=V2, choices=LEGAL_VERSIONS)
     white_listed_case_types = StringListProperty(default=[])  # empty value means all case-types are accepted
@@ -354,8 +358,7 @@ class ShortFormRepeater(Repeater):
     version = StringProperty(default=V2, choices=LEGAL_VERSIONS)
     friendly_name = _("Forward Form Stubs")
 
-    class Formats(object):
-        formats = [ShortFormRepeaterJsonPayloadGenerator]
+    _generators = [ShortFormRepeaterJsonPayloadGenerator]
 
     @memoized
     def payload_doc(self, repeat_record):
@@ -378,8 +381,7 @@ class ShortFormRepeater(Repeater):
 class AppStructureRepeater(Repeater):
     friendly_name = _("Forward App Schema Changes")
 
-    class Formats(object):
-        formats = [AppStructureGenerator]
+    _generators = [AppStructureGenerator]
 
     def payload_doc(self, repeat_record):
         return None
@@ -388,8 +390,7 @@ class AppStructureRepeater(Repeater):
 class UserRepeater(Repeater):
     friendly_name = _("Forward Users")
 
-    class Formats(object):
-        formats = [UserPayloadGenerator]
+    _generators = [UserPayloadGenerator]
 
     @memoized
     def payload_doc(self, repeat_record):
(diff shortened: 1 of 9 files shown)


dannyroberts:
@NoahCarnahan 🅱️

corehq/apps/repeaters/models.py
```diff
@@ -54,67 +52,6 @@ def log_repeater_timeout_in_datadog(domain):
...
-
-    def get_all_formats(self, for_domain=None):
-        """returns all the formats added to this repeater collection"""
-        return [(name, format.label) for name, format in self.format_generator_map.iteritems()
-                if not for_domain or format.generator_class.enabled_for_domain(for_domain)]
-
-    def get_generator_by_format(self, format):
-        """returns generator class given a format"""
-        return self.format_generator_map[format].generator_class
-
```
dannyroberts: Just had to move this code because of circular imports

corehq/apps/repeaters/models.py
```diff
@@ -324,6 +261,13 @@ class FormRepeater(Repeater):
 
     """
 
+    class Formats(object):
+        formats = {
+            'form_xml': (FormRepeaterXMLPayloadGenerator, _('XML')),
+            'form_json': (FormRepeaterJsonPayloadGenerator, _('JSON')),
+        }
+        default_format = 'form_xml'
```
dannyroberts: This is the config style for specifying multiple formats

corehq/apps/repeaters/models.py
```diff
@@ -412,6 +363,9 @@ class ShortFormRepeater(Repeater):
     version = StringProperty(default=V2, choices=LEGAL_VERSIONS)
     friendly_name = _("Forward Form Stubs")
 
+    class Formats(object):
+        generator = ShortFormRepeaterJsonPayloadGenerator
```
dannyroberts: this is the (simpler) config style for specifying a single format, since in that case you don't need to specify a slug, a display string, or a default.

corehq/apps/repeaters/repeater_generators.py
```diff
@@ -75,6 +67,68 @@ def handle_exception(self, exception, repeat_record):
         return True
 
 
+FormatInfo = namedtuple('FormatInfo', 'name label generator_class')
+
+
+class GeneratorCollection(object):
```
dannyroberts: Again this is all just moved here.

dannyroberts:
✅ (but still worth reviewing of course)

corehq/apps/repeaters/models.py
proteusvacuum: would it make sense to let the `PayloadGenerator` handle what it is called / what it's format is?
That way you could just have: 
```py
formats = [PayloadGenerator1, PayloadGenerator2]
```
and have it be implicit that the first generator that is in the list be the default (i.e. it is the first one that shows up in the UI). 
This also makes the single-format case much easier, as it would just be a single thing in the list (and you don't need to remember that when there is only one format you use the word `generator` instead of `formats`).

NoahCarnahan: What's the advantage of using this inner class as opposed to adding a `formats` and `default_format` case property?

dannyroberts: That's a good idea.

NoahCarnahan: (didn't see Farid's comment before making mine)

dannyroberts: @NoahCarnahan So I tried going with that first, but it turns out that couchdbkit.Document does not let you have top-level, non-private (i.e. non-`_`-prefixed) properties that are dicts or lists. That's because if it's a type that it naively detects as JSON-serializable (ints, strings, lists, dicts, etc.) then it treats it as a *Property with a constant value (with the effect that it saves these to the db). This is behavior I inherited from couchdbkit when I rewrote its schema library to be jsonobject-based, and I intentionally kept it for backwards compatibility (we do rely on that behavior here and there).

Anyway, so as a result, I'm sticking it inside this class. Open to better ideas!

NoahCarnahan: ah ha, that makes sense, thanks.

proteusvacuum: can you do this instead?

```py
class Repeater(Document):
    _formats = [Formats]

    @property
    def formats(self):
        return _formats

    @formats.setter(self, formats)
        self._formats = formats
```
I guess that doesn't work with class instantiation.

dannyroberts: @proteusvacuum Yeah, so I thought of some version of that too earlier. Not sure what you mean by

> I guess that doesn't work with class instantiation.

exactly, but the thing that broke it for me is that this `.formats` property we're trying to create is primarily a _class_ property and not an instance property. `@property` only works to create _instance_ properties; `Repeater.formats` in the example above would be just the property object that `property` returns.

We could totally just make it `_`-prefixed and be fine referencing it outside the class or through a `get_formats()` classmethod.

In addition but separately, it'd kind of make more sense to call the method `generators` (or `_generators`) rather than `formats`. Not sure what you think.

corehq/apps/repeaters/models.py
```diff
@@ -132,6 +68,8 @@ class Repeater(QuickCachedDocumentMixin, Document, UnicodeMixIn):
     password = StringProperty()
     friendly_name = _("Data")
 
+    payload_generator_classes = ()
```
dannyroberts: @proteusvacuum OK I think I finally nailed it. As @NoahCarnahan suggested, it'd ideally just be a property—and I then realized/remembered that `tuple`s don't trigger the same overzealous response in a `couchdbkit.Document`, so just using `tuple`s is an easy work around I think.

proteusvacuum: woo. very nice.

NoahCarnahan (approved):
nice

dannyroberts:
✅

PR #16215, opened 2017-05-17
DHIS2 integration UI improvements

kaapstorm:
This PR ...
* Allows multiple UCRs to be forwarded to DHIS2
* Allows data set maps and data value maps to be added and removed easily
* Allows data to be forwarded to DHIS2 on demand, as opposed to only on schedule

e.g.
datavalue_map

@snopoke cc @sheelio


corehq/apps/domain/static/domain/js/dhis2_map_settings.js
```diff
@@ -0,0 +1,137 @@
...
+            };
+        };
+    };
+
+    module.Dhis2MapSettings = function (dataSetMaps, sendDataUrl) {
+        var self = this;
+
+        self.frequencyOptions = [
+            {"value": "monthly", "text": "Monthly"},
+            {"value": "quarterly", "text": "Quarterly"}
```
dimagimon: Missing trailing comma. (comma-dangle)

corehq/apps/domain/static/domain/js/dhis2_map_settings.js
```diff
@@ -0,0 +1,137 @@
...
+        self.submit = function (form) {
+            var dataSetMaps = [];
+            for (var i = 0; i < self.dataSetMaps().length; i++) {
+                var dataSetMap = self.dataSetMaps()[i];
+                dataSetMaps.push(dataSetMap.serialize());
+            }
+            $.post(
+                form.action,
+                {'dataset_maps': JSON.stringify(dataSetMaps)},
+                function (data) { alert_user(data, "success", true); }
```
dimagimon: 'alert_user' is not defined. (no-undef)

corehq/apps/domain/static/domain/js/dhis2_map_settings.js
```diff
@@ -0,0 +1,137 @@
...
+            );
+        };
+
+        self.sendData = function () {
+            $.post(
+                sendDataUrl,
+                {},
+                function (data) { alert_user(data, "success", true); }
```
dimagimon: 'alert_user' is not defined. (no-undef)

corehq/apps/domain/static/domain/js/dhis2_map_settings.js
```diff
@@ -0,0 +1,137 @@
...
+            );
+        };
+    }
```
dimagimon: Missing semicolon. (semi)

corehq/apps/dhis2/tasks.py
```diff
@@ -52,4 +53,4 @@ def send_datavalues(domain_name):
 )
 def send_datasets_for_all_domains():
     for domain_name in find_domains_with_toggle_enabled(toggles.DHIS2_INTEGRATION):
-        send_datavalues(domain_name)
+        send_datasets.apply(args=(domain_name,))
```
snopoke: Do you want this to run synchronously? If so you can still just call the function normally, if not you should use `.delay`

corehq/apps/domain/templates/domain/admin/dhis2/dataset_map.html
```diff
@@ -4,16 +4,48 @@
 {% load crispy_forms_tags %}
 {% load i18n %}
 
+{% block js-inline %}{{ block.super }}
+    {% include 'domain/admin/partials/dhis2_dataset_map_templates.html' %}
+    <script src="{% static 'domain/js/dhis2_map_settings.js' %}"></script>
```
snopoke: Don't think these 2 should be in js-inline block. The templates should probably just live outside the block and the script should live in the `js` block.

corehq/apps/domain/templates/domain/admin/dhis2/dataset_map.html
```diff
@@ -4,16 +4,48 @@
...
+    <script>
+    (function () {
```
snopoke: Should this be `$(function...)`

corehq/apps/dhis2/dbaccessors.py
```diff
@@ -1,7 +1,7 @@
 from corehq.util.quickcache import quickcache
 
 
-@quickcache(['domain_name'])
+@quickcache(['domain_name'], timeout=10)
```
kaapstorm: I had a bug where after saving connection details or dataset maps, the user would still get the cached version. Considering that usually those functions are going to be called daily, I reckoned that if the timeout is longer than the call that needs the cache but less than the amount of time it takes the user to change and save the details, then it would resolve the bug, and still be no less effective when it comes to sending data to DHIS2 24 hours later.

I saw the same pattern being used in [accounting.models](https://github.com/dimagi/commcare-hq/blob/master/corehq/apps/accounting/models.py#L732) and [reports.util](https://github.com/dimagi/commcare-hq/blob/master/corehq/apps/reports/util.py#L479).

Is there a better way to do this?

corehq/apps/dhis2/tasks.py
kaapstorm: Calling it normally makes sense. [The docs](http://docs.celeryproject.org/en/latest/reference/celery.app.task.html) gave me the impression that using `apply` was the right way to do this, but they don't explain the difference between using `apply` and calling it normally. If there's no difference, or if it's not important, calling it normally is simpler. I'll change it.

corehq/apps/domain/templates/domain/admin/dhis2/dataset_map.html
kaapstorm: I'm not sure. I originally thought I should use `$(document).ready()`, but then realised that is not a thing any more, and noticed this pattern in [quite][1] [a][2] [few][3] [places][4]. 

But now that I search for `$(function...)` I'm seeing it everywhere, so I'll change this. 

I understand `$(function...)` is jQuery shorthand for `$(document).ready(function...)` and `(function...)()` just calls the function, but when should I use the second format? (Or are some of those instances I linked to above just the same mistake I made? :slightly_smiling_face: ) ... 

... After some googling, it seems at least some of those instances are definitely mistakes. (And the others are probably just logical collections of commands to be executed immediately, which is fair enough.)

  [1]: https://github.com/dimagi/commcare-hq/blob/master/corehq/apps/data_interfaces/templates/data_interfaces/interfaces/archive_forms.html#L23
  [2]: https://github.com/dimagi/commcare-hq/blob/master/corehq/apps/domain/templates/domain/billing_statements.html#L12
  [3]: https://github.com/dimagi/commcare-hq/blob/master/corehq/apps/data_interfaces/templates/data_interfaces/interfaces/archive_forms.html#L23
  [4]: https://github.com/dimagi/commcare-hq/blob/master/custom/m4change/templates/m4change/selectTemplate.html#L14

corehq/apps/dhis2/dbaccessors.py
kaapstorm: I'll answer my own question: I'll clear the cache on `save()`

corehq/apps/domain/templates/domain/admin/dhis2/dataset_map.html
```diff
@@ -4,16 +4,49 @@
 {% load crispy_forms_tags %}
 {% load i18n %}
 
+{% block js-inline %}{{ block.super }}
+    <script src="{% static 'domain/js/dhis2_map_settings.js' %}"></script>
```
kaapstorm: Sorry, I thought you meant just the templates.

PR #16217, opened 2017-05-17
Filter out cases belonging to specific owners in case search

proteusvacuum:
Allows adding either an XPath expression or a space separated list of valid owner ids to filter out of search results. 
screenshot from 2017-05-17 19-48-11

@sheelio @ctsims FYI
@benrudolph Buddy 🐟 / 🐡

### DIFF
--- a/corehq/apps/ota/views.py
+++ b/corehq/apps/ota/views.py
@@ -35,7 +35,7 @@ from corehq.apps.es.case_search import CaseSearchES, flatten_result
 from corehq.apps.hqwebapp.views import BaseSectionPageView
 from corehq.apps.ota.forms import PrimeRestoreCacheForm, AdvancedPrimeRestoreCacheForm
 from corehq.apps.ota.tasks import queue_prime_restore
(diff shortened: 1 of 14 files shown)


corehq/apps/ota/views.py
```diff
@@ -120,6 +121,11 @@ def search(request, domain):
     if owner_id:
         search_es = search_es.owner(owner_id)
 
+    blacklisted_owner_ids = criteria.pop(CASE_SEARCH_BLACKLISTED_OWNER_ID_KEY, None)
+    if blacklisted_owner_ids is not None:
+        for blacklisted_owner_id in blacklisted_owner_ids.split(' '):
+            search_es = search_es.blacklist_owner_id(blacklisted_owner_id)
```
benrudolph: could potentially add a test for this

benrudolph (approved):
looking good

benrudolph (approved):

benrudolph:
thanks, realize that's pretty much the same as your other test. i had thought there was already some setup to test the actual query. looks good

PR #16218, opened 2017-05-18
Add SaveButton to Enable Case Search page

proteusvacuum:
I'm about to add some more things to this page. I updated the view so that the UI is a little friendlier and will be easier to add things to later.

screenshot from 2017-05-17 23-50-43

Also added `SaveButton`, which made the JS a little more complicated, but it shows when there are errors, etc.

@benrudolph

### DIFF
--- a/corehq/apps/domain/static/domain/js/case-search-config.js
+++ b/corehq/apps/domain/static/domain/js/case-search-config.js
@@ -1,4 +1,4 @@
-/* globals hqDefine, ko, $, _ */
+/* globals hqDefine, ko, $, _, COMMCAREHQ, hqImport */
 
 hqDefine('domain/js/case-search-config.js', function () {
     'use strict';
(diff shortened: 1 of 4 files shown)


corehq/apps/domain/static/domain/js/case-search-config.js
```diff
@@ -50,14 +50,33 @@ hqDefine('domain/js/case-search-config.js', function () {
...
+        };
+        self.fuzzyProperties.subscribe(self.change);
+
         self.addCaseType = function () {
             self.fuzzyProperties.push(new CaseTypeProps('', ['']));
+            self.change();
         };
         self.removeCaseType = function (caseType) {
             self.fuzzyProperties.remove(caseType);
+            self.change();
```
benrudolph: doesn't the SaveButton listen to all changes on `input` elements and then fire the change event for you? i would've expected it to just work

corehq/apps/domain/templates/domain/admin/partials/case_search_templates.html
```diff
@@ -1,40 +1,37 @@
...
+    </div>
 </div>
 
 <div id="case_type_template" class="hide">
-    <fieldset style="padding: 0 0 0 2em;">
-        <p>
-            <label>{% trans "Case type" %}</label>
+    <div class="panel panel-default">
+        <div class="panel-heading clearfix">
+            <strong>{% trans "Case type" %}: </strong>
```
benrudolph: 👗

benrudolph (approved):

corehq/apps/domain/static/domain/js/case-search-config.js
proteusvacuum: yeah, me too. It wasn't working for me - I think because it doesn't subscribe to new elements that are created on the page.

PR #16224, opened 2017-05-18
App manager v2: issues with shadow modules

orangejenny:
https://manage.dimagi.com/default.asp?254260

The conditionals in module view and form view are getting difficult to maintain. Making a mental note to revisit them once we've figured out what to do about simple vs advanced apps, since the many checks for `app.advanced_app_builder` checks aren't helping the situation.

@esoergel / @biyeun


biyeun (approved):
oof yeah these conditionals are definitely dirtying the templates and are in need of some cleaning.

PR #16225, opened 2017-05-18
Add TimestampDocument to couch models

benrudolph:
@gcapalbo how do you feel about this approach to adding last modified? on one hand i like that this is just a mixin to add. however i hate subclassing `Document`. other option could be to just define the two properties on each model and override the save function. (we don't have to add created_on, but i've always wanted it and this felt like a good excuse)

cc: @proteusvacuum

### DIFF
--- a/corehq/apps/users/models.py
+++ b/corehq/apps/users/models.py
@@ -32,6 +32,7 @@ from dimagi.utils.decorators.memoized import memoized
 from dimagi.utils.make_uuid import random_hex
 from dimagi.utils.modules import to_function
 from corehq.util.quickcache import quickcache
+from corehq.util.couch_helpers import TimestampDocument
 from casexml.apps.case.mock import CaseBlock
 from corehq.form_processor.interfaces.dbaccessors import CaseAccessors
 from corehq.form_processor.exceptions import CaseNotFound
@@ -807,7 +808,7 @@ class ReportingMetadata(DocumentSchema):
     last_submissions = SchemaListProperty(LastSubmission)
 
 
-class CouchUser(Document, DjangoUserMixin, IsMemberOfMixin, UnicodeMixIn, EulaMixin):
+class CouchUser(TimestampDocument, DjangoUserMixin, IsMemberOfMixin, UnicodeMixIn, EulaMixin):
     """
     A user (for web and commcare)
     """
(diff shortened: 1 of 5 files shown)


benrudolph:
if ok with approach i'll add some tests (works locally)

benrudolph:
requires: https://github.com/dimagi/dimagi-utils/pull/408

corehq/apps/users/models.py
```diff
@@ -1311,11 +1311,6 @@ def create(cls, domain, username, password, email=None, uuid='', date='',
         else:
             couch_user = cls()
 
-        if date:
-            couch_user.created_on = force_to_datetime(date)
```
benrudolph: looked into this gem a bit. couldn't immediately find any usages. it was added back in 2011 for [this](https://github.com/dimagi/commcare-hq/commit/e0073d1d5#diff-16ce9f2eb3be0695a91eca75714bb3b7R359) which is no longer in the code base

corehq/util/couch_helpers.py
```diff
@@ -170,6 +172,20 @@ def _get_key_kwargs(self, key):
...
 
+class TimestampDocument(Document):
+
+    last_modified = DateTimeProperty()
+    created_on = DateTimeProperty(default=datetime(year=1900, month=1, day=1))
+
+    def save(self, *args, **kwargs):
+        utcnow = datetime.utcnow()
+        self.last_modified = utcnow
+        if not self._id:
```
benrudolph: this isn't foolproof. if you init a doc with an _id this'll fail. another option is to remove the default value on `created_on` and make the condition `if not self.created_on`. however that will then create false datums (i.e. the created_on would be set on save for some old models and that value would be incorrect)

benrudolph: or we can say this isn't worth it and can just drop

emord:
Since we maintain our own fork of couchdbkit, would it be possible to implement an `auto_add_now` on the property similar to django? That would mean that you don't need to subclass the document at all and it becomes much easier to add those

benrudolph:
that's not a bad thought, i was thinking it'd be nice to have what django has. i'll look into that. though it still may not be foolproof in terms of figuring out if a doc has been created without making a trip to the db

gcapalbo:
one thought that occurred to me is we'll also need to update our bulk save methods in order to support this.

i think what Emord suggested is probably the cleanest way to do it if we can figure out a way to. it might not be that hard actually to just inspect the list of field definitions in the save and bulk save methods, check for auto_add_now, and update accordingly.

regarding created_on - i don't think we have a use for that yet right? i'd be ok with dropping it if it's too much trouble

benrudolph:
https://github.com/dimagi/couchdbkit/pull/23 -- will reopen with auto_now_add

PR #16227, opened 2017-05-18
Revert "attempt fix docker biuld"

emord:
Reverts dimagi/commcare-hq#16223 didn't work, but it's more obvious to me why it doesn't work.

Will likely need to change https://github.com/dimagi/commcare-hq/blob/master/requirements/requirements.txt#L99 to a git endpoint. @dannyroberts why did you choose to install it locally from submodule instead of using a git endpoint?


dannyroberts:
@emord I can get rid of that for now. Ended up starting what I was doing from a different point, so I'm not using that yet anyway.

PR #16229, opened 2017-05-18
prune synclogs task

emord:
@snopoke buddy @calellowitz 

Adds the celery task.

I looked at the break down of synclogs on icds and it seems like more than half of them were created in the past 2 weeks (almost 1 million). and this task won't delete many for the next couple of weeks. so the db may still grow out of control at scale. 

What do you think about making this configurable based on environment? (would require some ansible work in localsettings) I think we could do 30 days (maybe less?) for icds since they sync so often?

### DIFF
--- a/corehq/ex-submodules/casexml/apps/phone/tasks.py
+++ b/corehq/ex-submodules/casexml/apps/phone/tasks.py
@@ -1,8 +1,10 @@
+from datetime import date, timedelta
 from celery import current_task, current_app
 from celery.schedules import crontab
 from celery.task import periodic_task, task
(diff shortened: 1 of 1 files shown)


snopoke:
I'd prefer to implement the other pruning method which removes previous synclogs after a successful form submission. This task would just be responsible for cleaning up any that remain. We do already keep the ID of the previous log in the SyncLog so it should be fairly straight forward. I'll take a quick crack at that.

snopoke:
https://github.com/dimagi/commcare-hq/pull/16238

PR #16230, opened 2017-05-18
Output when import_app runs successfully

orangejenny:
Tiny.

@esoergel

### DIFF
--- a/corehq/apps/app_manager/management/commands/import_app.py
+++ b/corehq/apps/app_manager/management/commands/import_app.py
@@ -78,4 +78,5 @@ class Command(BaseCommand):
         app_source = resp.json()
         if not name:
             name = app_source['name']
-        import_app(app_source, target_domain, {'name': name})
+        app = import_app(app_source, target_domain, {'name': name})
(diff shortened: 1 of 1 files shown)


corehq/apps/app_manager/management/commands/import_app.py
```diff
@@ -78,4 +78,5 @@ def handle(self, domain, app_id, **options):
         app_source = resp.json()
         if not name:
             name = app_source['name']
-        import_app(app_source, target_domain, {'name': name})
+        app = import_app(app_source, target_domain, {'name': name})
+        return "Created app '{}' at /a/{}/apps/view/{}/".format(app.name, app.domain, app.id)
```
esoergel: Should this be a `print` rather than `return` statement?

esoergel: Just looked at the Django source for `BaseCommand` and apparently if `handle` returns something, it writes it to `sys.stdout`.  Learned something new today.

esoergel (approved):

PR #16232, opened 2017-05-18
Remove motech app for now

dannyroberts:
@emord This removes (among other things) authproxy_client dependency. We can add this back when I start working in that directory. (Right now iterating in a smaller way.)


dannyroberts:
@NoahCarnahan 🅱️ this comes from @emord's request on https://github.com/dimagi/commcare-hq/pull/16227

emord (approved):

PR #16235, opened 2017-05-18
Move fuzzy properties config to model

proteusvacuum:
I started down this path then couldn't get off it...
I'm not sure why we were ever using such an awkward config structure with that JSON document.

@benrudolph 🐡 / 🐠


corehq/apps/case_search/models.py
```diff
@@ -62,6 +63,28 @@ def get_or_none(self, **kwargs):
             return None
 
 
+class FuzzyProperties(models.Model):
```
benrudolph: 🐻

corehq/apps/domain/static/domain/js/case-search-config.js
```diff
@@ -70,26 +70,33 @@ hqDefine('domain/js/case-search-config.js', function () {
...
-                    ),
-                });
+                    );
+
+                if (fuzzyProperties[caseType]){
+                    for (var propIdx in properties){
+                        fuzzyProperties[caseType].push(properties[propIdx]);
+                    }
+                } else {
+                    fuzzyProperties[caseType] = properties;
```
benrudolph: could just change this if statemen to:
`fuzzyProperties[caseType] = (fuzzyProperties[caseType] || []).concat(properties)`

corehq/apps/case_search/models.py
```diff
@@ -62,6 +63,28 @@ def get_or_none(self, **kwargs):
...
+        blank=False,
+        db_index=True,
+    )
+    properties = ArrayField(
+        models.TextField(null=True, blank=True),
+        null=True,
+    )
+
+    class Meta(object):
+        unique_together = ('domain', 'case_type')
```
benrudolph: probably not that necessary depending on the record but seems like index_together for `case_type` and `domain` would be the most efficient

benrudolph: (actually realizing this would be very small. not worth it)

benrudolph (approved):
tests still failing

PR #16241, opened 2017-05-19
set case server_modified_on to edited_on date during edit form workflow

snopoke:
There is a known issue that after editing a form that touches cases those cases won't get re-synced to a users phone. The reason for this is that the cases `server_modified_on` date doesn't get changed by the form edit.

This change will set `case.server_modified_on` to the time of the edit (which seems more correct anyway).

### DIFF
--- a/corehq/ex-submodules/couchforms/tests/test_edits.py
+++ b/corehq/ex-submodules/couchforms/tests/test_edits.py
@@ -165,12 +165,13 @@ class EditFormTest(TestCase, TestFileMixin):
                 'property': 'edited value'
             }
         ).as_string()
-        submit_case_blocks(case_block, domain=self.domain, form_id=form_id)
+        xform, _ = submit_case_blocks(case_block, domain=self.domain, form_id=form_id)
 
         case = self.casedb.get_case(case_id)
         self.assertEqual(case.type, 'newtype')
         self.assertEqual(case.dynamic_case_properties()['property'], 'edited value')
         self.assertEqual([form_id], case.xform_ids)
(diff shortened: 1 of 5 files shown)


benrudolph (approved):
yeah this is an interesting one. can see arguments both ways, but sounds like this'll fix a limitation

kaapstorm (approved):

snopoke:
@benrudolph curious to hear your arguments against

benrudolph:
the only argument i was thinking of is the fact that if you view edit forms as a invisible swap of forms (as if the user had actually submitted the form the way you edited it) then it makes sense that the last time the case was modified will stay the same. for instance:

Form A, Form B modify Case C on 1/1/17

if you edit Form A, you will still process the forms in order Form A then Form B, however Case C now has a modified_on date that is later than when Form B was received which could be confusing.

still think the practical use case of cases not getting re-sync probably out weighs the above

snopoke:
Mmm, I think we're currently in an inconsistent place. `modified_on` consistently refers to the latest 'timeEnd' from forms that were submitted against the case. `server_modified_on` currently means the latest `received_on` date from forms that were submitted against the case. 

There are two scenarios that make it confusing which are case rebuilds and editing forms. So what I'm proposing is to change the definition of `server_modified_on` to be the max data of any case transactions which include rebuild (triggered either by form edit or manually).

I've added a commit to update the date for manually triggered rebuilds as well. Let me know what you think.

benrudolph (approved):
👍 i like that. maybe a comment above `server_modified_on` so people know the definition?

kaapstorm (approved):

PR #16243, opened 2017-05-19
Allow specifying the es instance you want to use

benrudolph:
Some more robust configuration settings for exports. i think the ultimate goal would be able to detect a failure and the resume if we can, however i think that's going to be exceedingly difficult with the way es scroll queries work

### DIFF
--- a/corehq/elastic.py
+++ b/corehq/elastic.py
@@ -74,9 +74,9 @@ ES_INSTANCES = {
 }
 
 
-def get_es_instance(es_instance=ES_DEFAULT_INSTANCE):
-    assert es_instance in ES_INSTANCES
-    return ES_INSTANCES[es_instance]()
+def get_es_instance(es_instance_alias=ES_DEFAULT_INSTANCE):
+    assert es_instance_alias in ES_INSTANCES
+    return ES_INSTANCES[es_instance_alias]()
 
 
 def doc_exists_in_es(index_info, doc_id_or_dict):
@@ -179,7 +179,7 @@ class ESError(Exception):
     pass
 
 
-def run_query(index_name, q, debug_host=None, es_instance=ES_DEFAULT_INSTANCE):
+def run_query(index_name, q, debug_host=None, es_instance_alias=ES_DEFAULT_INSTANCE):
     # the debug_host parameter allows you to query another env for testing purposes
     if debug_host:
         if not settings.DEBUG:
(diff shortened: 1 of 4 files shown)


corehq/elastic.py
```diff
@@ -144,7 +179,7 @@ class ESError(Exception):
     pass
 
 
-def run_query(index_name, q, debug_host=None):
+def run_query(index_name, q, debug_host=None, es_instance=ES_DEFAULT_INSTANCE):
```
emord: Instead of passing in a string that looks in a dictionary to get this, why not pass in the actual instance?

That would allow you to have the export ES instance defined in corehq/apps/export and the default value could just be `None` which defaults to the old behavior of calling get_es_new

benrudolph: i kind of wanted all the EXPORT_INSTANCES defined in one place in case this ever gets reused it's less likely someone creates the same exact es instance with the same params. that param name is a bit misleading though. will rename

snopoke (approved):

corehq/elastic.py
```diff
@@ -23,27 +23,62 @@
...
         get_es_new._es_client = Elasticsearch(hosts)
     return get_es_new._es_client
 
 
+def get_es_export():
+    """
+    Get a handle to the configured elastic search DB with settings geared towards exports.
+    Returns an elasticsearch.Elasticsearch instance.
+    """
+    if not getattr(get_es_export, '_es_client', None):
```
esoergel: FYI `@memoized` also works on module-level functions

esoergel:
:+1:

corehq/elastic.py
```diff
@@ -23,27 +23,62 @@
...
+ES_DEFAULT_INSTANCE = 'default'
+ES_EXPORT_INSTANCE = 'export'
+
+ES_INSTANCES = {
+    ES_DEFAULT_INSTANCE: get_es_new,
+    ES_EXPORT_INSTANCE: get_es_export,
+}
+
+
+def get_es_instance(es_instance=ES_DEFAULT_INSTANCE):
```
NoahCarnahan: When I first read this PR I thought that `es_instance` referred to a completely different es cluster, but it looks like it's actually a different `Elasticsearch` object that points to the same cluster, but with different settings. I wonder if it would be more clear to future readers of the code if these kwargs were called `es_client` instead? (I don't feel strongly about this name change but just thought I would throw it out there in case you agree it's worth changing.)

benrudolph: ah i see i'm kind of torn because it is an instance of an elasticsearch object, but can see where that's confusing

corehq/elastic.py
benrudolph: that seems cleaner

PR #16247, opened 2017-05-19
Add static label to static report/data source names

NoahCarnahan:
This makes it easier to tell at a glance which are static and which are not without having to add (static) to the name of ever data source and report. These labels only show on the internal UCR pages, not on the general report pages.

@proteusvacuum 
buddy @dannyroberts 

Screen shot:
screen shot 2017-05-19 at 1 40 26 pm


dannyroberts:
💣

PR #16250, opened 2017-05-19
Better error message

emord:
@dannyroberts @benrudolph  Thoughts on this error message?

http://manage.dimagi.com/default.asp?252848#1332770

buddy @calellowitz

### DIFF
--- a/corehq/apps/reports/views.py
+++ b/corehq/apps/reports/views.py
@@ -1756,11 +1756,14 @@ class EditFormInstance(View):
         instance = _get_location_safe_form(domain, request.couch_user, instance_id)
         context = _get_form_context(request, domain, instance)
         if not instance.app_id or not instance.build_id:
-            return _error(_(
-                "This submission was made from the App Preview."
-                "As a result it is not possible to edit this submission\n\n"
-                "If this was not submitted from App Preview, please report an issue"
-            ))
+            deviceID = instance.metadata.deviceID
+            if deviceID and deviceID == 'Formplayer':
+                return _error(_(
+                    "Could not detect the application or form for this submission. "
+                    "A common cause is that the form was submitted via App or Form preview"
+                ))
+            else:
+                return _error(_('Could not detect the application or form for this submission.'))
 
         user = CouchUser.get_by_user_id(instance.metadata.userID, domain)
         if not user:
--- a/corehq/apps/reports/views.py
+++ b/corehq/apps/reports/views.py
@@ -1756,7 +1756,11 @@ class EditFormInstance(View):
         instance = _get_location_safe_form(domain, request.couch_user, instance_id)
         context = _get_form_context(request, domain, instance)
         if not instance.app_id or not instance.build_id:
-            return _error(_('Could not detect the application/form for this submission.'))
+            return _error(_(
+                "This submission was made from the App Preview."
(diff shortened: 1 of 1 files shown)


corehq/apps/reports/views.py
```diff
@@ -1756,7 +1756,11 @@ def _error(msg):
         instance = _get_location_safe_form(domain, request.couch_user, instance_id)
         context = _get_form_context(request, domain, instance)
         if not instance.app_id or not instance.build_id:
-            return _error(_('Could not detect the application/form for this submission.'))
+            return _error(_(
+                "This submission was made from the App Preview."
+                "As a result it is not possible to edit this submission\n\n"
+                "If this was not submitted from App Preview, please report an issue"
```
dannyroberts: Ha! I guess pushing our uncertainty on the user is one way to do this...

@benrudolph Are there any other markers of the form we could use to determine whether it was submitted via App Preview?

emord: Hmm. that's a good point. I didn't mean to push uncertainty to the users, just meant to give a more precie message.

We could also assert that the deviceID is Formplayer and/or it was submitted by a webuser. If not, we could show the previous error.

benrudolph: Yea I think App Preview is the main use case, but certainly not the only. Form Preview also creates un-editable forms, as does any sort of form submitted to commcare via api that isn't attached to an app. guessing system forms act similarly.

We could say "Could not detect the application/form for this submission. A common cause is that the form was submitted via App or Form preview." or something like that. i like the idea of checking the deviceID

corehq/apps/reports/views.py
```diff
@@ -1756,7 +1756,14 @@ def _error(msg):
         instance = _get_location_safe_form(domain, request.couch_user, instance_id)
         context = _get_form_context(request, domain, instance)
         if not instance.app_id or not instance.build_id:
-            return _error(_('Could not detect the application/form for this submission.'))
+            deviceID = instance.metadata.deviceID
+            if deviceID and deviceID == 'Formplayer':
```
emord: @benrudolph I couldn't find where this was set. Is there a constant I can pull from for this?

emord (commented):
Went with two different error messages. If we detect it's from formplayer we'll hint that it's from the preview

dannyroberts (approved):
This certainly looks like a strict improvement :)

@benrudolph down the road might be useful for formplayer to leave a clear trail of what feature the user was using to submit it. That way we can check some field that positively declares the form was submitted through 'App Preview', or through 'Web Apps', or through 'Edit Form' (or any other thing that uses Formplayer in the backend), and we can reliably mark them as such. Just placing a little seed in your brain.

corehq/apps/reports/views.py
benrudolph: unfortunate not :/ https://github.com/dimagi/formplayer/blob/f47d3bf7a749f4ee13ad2e92745906794ae0088f/src/main/java/session/FormplayerInstanceInitializer.java#L79

benrudolph:
i definitely agree. we've talked about it before, but have never done it cause it didn't make the priority cut. i think the paper trail will be nice

benrudolph:
those seeds 🌱  grow into trees 🌳

PR #16252, opened 2017-05-19
show error message when too many columns are selected

emord:
@benrudolph @NoahCarnahan http://manage.dimagi.com/default.asp?253745

not the sexiest or most robust solution, but it's something and I think this is a pretty far edge case

buddy @calellowitz


benrudolph (approved):
👍 the only semi edge case is the fact that when you do a bulk it automatically switches the format to xlsx, but i like the idea of validation on save, so i think it's worth it

PR #16257, opened 2017-05-22
Changes in TB Notification Register report

kkrampa:
* Added health_facility_for_treatment column
* Changed logic of treatment_status_value column


kkrampa:
@NoahCarnahan

PR #16259, opened 2017-05-22
Log _validate_case_removal soft asserts to a file instead of email

NoahCarnahan:
This soft assert is generating like 500 MB of email messages per day, which would fill an empty inbox in 60 days. This PR logs those messages to a file instead.

@dannyroberts because soft assert/buddy
@snopoke @millerdev because I think you two might be working on `_validate_case_removal`?

### DIFF
--- a/corehq/util/soft_assert/api.py
+++ b/corehq/util/soft_assert/api.py
@@ -1,3 +1,5 @@
+import logging
+
 from django.conf import settings
 
 from corehq.apps.hqwebapp.tasks import send_mail_async, mail_admins_async
@@ -5,6 +7,7 @@ from corehq.util.log import get_sanitized_request_repr
 from corehq.util.global_request import get_request
 from corehq.util.soft_assert.core import SoftAssert
 
+logger = logging.getLogger('soft_asserts')
 
 def _send_message(info, backend):
     request = get_request()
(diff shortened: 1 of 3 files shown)


snopoke:
Yikes, thanks @NoahCarnahan. I was planning on digging into these but I think I have plenty in my email to look at. Probably worth just commenting it out completely for now.

emord (approved):

snopoke (approved):

NoahCarnahan:
I'm going to merge before tests pass because latest commit just comments out some lines and it would be nice to have this in before deploy.

corehq/util/soft_assert/api.py
```diff
@@ -87,20 +91,21 @@ def send_to_admins(subject, message):
...
-            _send_message(info, backend=send_to_recipients)
-    elif to:
-        def send(info):
-            _send_message(info, backend=send_to_recipients)
-    elif notify_admins:
-        def send(info):
-            _send_message(info, backend=send_to_admins)
-    else:
+    def write_to_log_file(subject, message):
+        logger.debug(message)
```
millerdev: What do you think of increasing the logging level to WARNING or ERROR? (it is an assertion failure after all)

PR #16260, opened 2017-05-22
Start re-externalizing javascript in form view

orangejenny:
Re-applying https://github.com/dimagi/commcare-hq/pull/16152 but doing it in smaller pieces because I still don't have a cause for https://manage.dimagi.com/default.asp?254212 and that PR looked related (although there was a report of the same issue after it was rolled back).

This PR just starts pulling the various case config options over to the server.

@esoergel / @millerdev

### DIFF
--- a/corehq/apps/app_manager/templates/app_manager/v1/form_view_base.html
+++ b/corehq/apps/app_manager/templates/app_manager/v1/form_view_base.html
@@ -118,6 +118,7 @@
             return result;
         }
 
+        var form_requires = ko.observable({{ form.requires|JSON }});
         var putInRoot = {{ module.put_in_root|BOOL }};
         var isUserCaseInUse = {{ is_usercase_in_use|BOOL }};
         var allOtherFormsRequireACase = {{ form.all_other_forms_require_a_case|BOOL }};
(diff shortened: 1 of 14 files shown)


corehq/apps/app_manager/views/forms.py
```diff
@@ -722,7 +726,9 @@ def commtrack_programs():
             "app_manager/v1/form_view.html",
             "app_manager/v2/form_view.html",
         )
-        return template, context
+
+    context.update({'case_config_options': case_config_options})
```
esoergel: OOC, why is this way down at the bottom of the function rather than in the initial dict, near where it's defined?  Also, this function looks like a great candidate for the next refactorathon

corehq/apps/app_manager/templates/app_manager/v1/form_view_base.html
```diff
@@ -118,6 +118,7 @@
             return result;
         }
 
+        var form_requires = ko.observable({{ form.requires|JSON }});
```
esoergel: What's this guy for?  Looks like elsewhere you've been using `initial_page_data` for it.

esoergel (approved):

corehq/apps/app_manager/views/forms.py
orangejenny: In future PRs, there's going to be more added to that dict in the module-type-based conditionals above this.

> Also, this function looks like a great candidate for the next refactorathon
yeah, tell me about it...

corehq/apps/app_manager/templates/app_manager/v1/form_view_base.html
orangejenny: It's used in `FormFilter` a bit below this, all I'm doing is moving it closer to where it's used. The `{{ form.requires|JSON }}` here is indeed equivalent to using `initial_page_data('form_requires')`. No particular reason for not using initial_page_data here, probably just copied/pasted before I added the initial_page_data code. I'll come back to this eventually, planning on getting this whole file externalized over the next week or two.

PR #16262, opened 2017-05-22
Replace `Application.get_form_by_xmlns()` (shadow forms related)

NoahCarnahan:
`Application.get_form_by_xmlns()` doesn't make sense any more now that shadow forms allow for multiple forms with the same xmlns. This PR replaces that method with `get_forms_by_xmlns()`.
@orangejenny 
@dannyroberts

### DIFF
--- a/corehq/apps/app_manager/models.py
+++ b/corehq/apps/app_manager/models.py
@@ -6033,6 +6033,8 @@ class Application(ApplicationBase, TranslationMixin, HQMediaMixin):
         if not forms:
             return None
         else:
+            # If there are multiple forms with the same xmlns, then all but one are shadow forms, therefore they
+            # all have the same xform.
             return forms[0].wrapped_xform()
 
 
--- a/corehq/apps/app_manager/models.py
+++ b/corehq/apps/app_manager/models.py
@@ -6007,28 +6007,43 @@ class Application(ApplicationBase, TranslationMixin, HQMediaMixin):
             xmlns_map[form.xmlns].append(form)
         return xmlns_map
 
-    def get_form_by_xmlns(self, xmlns, log_missing=True):
+    def get_forms_by_xmlns(self, xmlns, log_missing=True):
+        """
+        Return the forms with the given xmlns.
+        This function could return multiple forms if there are shadow forms in the app.
+        """
         if xmlns == "http://code.javarosa.org/devicereport":
-            return None
+            return []
         forms = self.get_xmlns_map()[xmlns]
-        if len(forms) != 1:
-            if log_missing or len(forms) > 1:
+        if len(forms) < 1:
+            if log_missing:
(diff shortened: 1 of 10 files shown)


dannyroberts:
Can you help me understand why this is important? I kind of though the point of shadow modules was to add secondary (or n-ary) ways to access the same form. In what context do you need to differentiate between the primary form and its shadow forms? It's obviously not inconceivable to me, but I think I just I don't understand the concrete/current use case.

NoahCarnahan:
Shadow forms and their parent forms are represented by the same xform, but different instances of `Form`. Shadow forms can have different names, display conditions, icons, form linking, and actions than their source forms (full spec [here](https://docs.google.com/document/d/1EYckqcOF9ZIHLm706dQxJ9Y_vnc7QY3BEE1pRdCNb-k/edit#)). This method, `Application.get_form_by_xmlns()` returns a `Form` instance. Before this PR, it would raise an exception if shadow forms were present, because there would be more than one `Form` object in the app that corresponds to the same xmlns (which was not allowed before shadow forms). This changes the function to instead return all `Form` instances that match the given xmlns.

corehq/apps/app_manager/models.py
```diff
@@ -6007,28 +6007,45 @@ def get_xmlns_map(self):
...
+
+    def get_xform_by_xmlns(self, xmlns, log_missing=True):
+        forms = self.get_forms_by_xmlns(xmlns, log_missing)
+        if not forms:
             return None
         else:
-            form, = forms
-        return form
+            # If there are multiple forms with the same xmlns, then all but one are shadow forms, therefore they
+            # all have the same xform.
```
benrudolph: is this assumption always true? based on the previous code it seems like this could happen when weird things arise. maybe we should raise an error if the md5 of the xforms don't match

benrudolph (commented):
thanks!

corehq/apps/app_manager/models.py
emord: I'm pro this extra check, but every form in an app should have a unique xmlns and if it doesn't we change it

https://github.com/dimagi/commcare-hq/blob/master/corehq/apps/app_manager/util.py#L125-L132

NoahCarnahan: @benrudolph 
`get_forms_by_xmlns()`, which this function calls, performs the check that confirms that all but one of the forms are shadow forms, so I think a second check here would be redundant. But we could check the xfrom itself here if we want to be extra sure there hasn't been any sort of corruption. Although it sounds like given the check that emord has also mentioned, we're probably covered.

dannyroberts (approved):
Thanks for the elaboration @NoahCarnahan, that makes lots of sense.

Looks good to me!

corehq/apps/app_manager/models.py
benrudolph: ah ok makes sense. fine to leave as is

PR #16270, opened 2017-05-22
add db_indexes for migration performance

nickpell:
@mkangia @millerdev 

Normally I would squash the two migration into one file, but they've already been run on softlayer this weekend.


custom/enikshay/private_sector_datamigration/migrations/0006_auto_20170520_1638.py
```diff
@@ -0,0 +1,30 @@
...
+# Generated by Django 1.10.7 on 2017-05-20 16:38
+from __future__ import unicode_literals
+
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('private_sector_datamigration', '0005_auto_20170520_1530'),
```
millerdev: Can't you just put both migrations in this file and set this to `('private_sector_datamigration', '0004_migratedbeneficiarycounter'),`?

I don't think the migration framework requires that every migration file is present as long as the most recent one is there and its dependencies are set correctly.

nickpell: updated

millerdev:
Looks like you moved all migrations into `0005_auto_20170520_1530.py`. It wasn't obvious (sorry for that), but I had suggested that they be moved into `0006_auto_20170520_1638.py`. I'm not 100% sure, but I thought the migration framework stores the leading number (`0006`) somewhere so it knows the last migration that was applied. I just looked at the documentation and realized I'm probably making too many assumptions about how this works, so maybe you should ignore that advice. It would be safer to use [`./manage.py squashmigrations`](https://docs.djangoproject.com/en/dev/topics/migrations/#migration-squashing) to squash them into a single file (and follow the instructions on that page about applying it to environments that have had those migrations already applied). Or if that's too much work just revert 01712d7 and keep them separate.

nickpell:
going to take the easy route and leave as is

millerdev (approved):

PR #16274, opened 2017-05-22
Fix accidental tuple

NoahCarnahan:
https://sentry.io/dimagi/commcarehq/issues/278343349/
This tuple prevents case list configs from being saved if the "Enable synchronous mobile searching and case claiming" feature flag is enabled.
@proteusvacuum 
buddy @dannyroberts


proteusvacuum:
https://github.com/dimagi/commcare-hq/pull/16272

NoahCarnahan:
nice

PR #16276, opened 2017-05-22
Cmitfb migration: skip non-Application apps

orangejenny:
https://manage.dimagi.com/default.asp?254349

Will now skip remote apps and linked apps.

@millerdev / @esoergel

### DIFF
--- a/corehq/apps/app_manager/management/commands/migrate_app_to_cmitfb.py
+++ b/corehq/apps/app_manager/management/commands/migrate_app_to_cmitfb.py
@@ -74,8 +74,8 @@ class Command(BaseCommand):
                             self.migrate_app(app)
                     else:
                         logger.info("Skipping %s/%s because it is a %s", domain, app_id, app.doc_type)
-                except Exception as err:
-                    logger.error("skipping app %s/%s: %s", domain, app_id, err)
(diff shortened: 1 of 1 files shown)


corehq/apps/app_manager/management/commands/migrate_app_to_cmitfb.py
```diff
@@ -66,14 +66,16 @@ def handle(self, **options):
...
+                        else:
+                            self.migrate_app(app)
                     else:
-                        self.migrate_app(app_id)
-                except SkipApp as err:
+                        logger.info("Skipping %s/%s because it is a %s", domain, app_id, app.doc_type)
+                except Exception as err:
                     logger.error("skipping app %s/%s: %s", domain, app_id, err)
-                except Exception:
-                    logger.exception("skipping app %s/%s", domain, app_id)
```
millerdev: I would prefer to bring back this `except` clause instead of the one above it. That is, I think it should use `logger.exception`, which logs the stack trace, and don't put the error name in the message since it's in the stack trace.

orangejenny: updated

PR #16277, opened 2017-05-22
Don't error on module filters that use dots in acceptable places

orangejenny:
https://manage.dimagi.com/default.asp?254632

@esoergel


esoergel (approved):
Seems like a foolproof approach - I assume it's overly complex to make `DOT_INTERPOLATE_PATTERN` not throw false positives?

orangejenny:
Yeah, making `DOT_INTERPOLATE_PATTERN` handle quotes is a mess.

PR #16279, opened 2017-05-23
Add last_modified to user model

benrudolph:
@gcapalbo @emord i talked to danny a bit yesterday about the `auto_now_add` and it's going to be a little longer process. i have to make some updates to the jsonobject library and change the way we wire up the couchdbkit models in hq. it's not terrible, but given all i need is this one property to move forward, i'm writing this as an interim solution

buddy: @proteusvacuum

### DIFF
--- a/corehq/apps/users/models.py
+++ b/corehq/apps/users/models.py
@@ -1340,6 +1340,13 @@ class CouchUser(Document, DjangoUserMixin, IsMemberOfMixin, UnicodeMixIn, EulaMi
         self.username = username
         self.save()
 
+    @classmethod
+    def save_docs(cls, docs, **kwargs):
+        utcnow = datetime.utcnow()
+        for doc in docs:
+            doc['last_modified'] = utcnow
+        super(CouchUser, cls).save_docs(docs, **kwargs)
+
     def save(self, **params):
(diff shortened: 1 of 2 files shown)


emord (approved):
Seems reasonable. Honestly didn't expect it to be super simple, but it'll be a cleaner solution eventually

benrudolph:
yeah i agree too. i just didn't want to get derailed at the moment, still on my radar though

corehq/apps/users/models.py
```diff
@@ -1339,7 +1340,15 @@ def change_username(self, username):
         self.username = username
         self.save()
 
+    @classmethod
+    def save_docs(cls, docs, **kwargs):
+        utcnow = datetime.utcnow()
+        for doc in docs:
+            doc['last_modified'] = utcnow
+        super(CouchUser, cls).save_docs(docs, **kwargs)
+
```
gcapalbo: i think we should also make sure `bulk_save` gets updated too since i think it will still point to the old version: https://github.com/dimagi/couchdbkit/blob/jsonobject/couchdbkit/schema/base.py#L159

benrudolph: wow that is uber tricky and great catch

PR #16283, opened 2017-05-23
TB Lab Register change

kkrampa:
* show only records with date reported filled


kkrampa:
@NoahCarnahan Need test data source rebuild

PR #16284, opened 2017-05-23
application last modified

benrudolph:
@gcapalbo 
cc: @proteusvacuum

### DIFF
--- a/corehq/apps/app_manager/models.py
+++ b/corehq/apps/app_manager/models.py
@@ -5364,11 +5364,21 @@ class ApplicationBase(VersionedDoc, SnapshotMixin,
             response_json=response_json, increment_version=increment_version, **params)
 
     @classmethod
-    def save_docs(cls, docs, **kwargs):
+    def _update_last_modified(cls, docs):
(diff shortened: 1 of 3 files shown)


corehq/apps/app_manager/models.py
```diff
@@ -5361,6 +5363,13 @@ def save(self, response_json=None, increment_version=None, **params):
         super(ApplicationBase, self).save(
             response_json=response_json, increment_version=increment_version, **params)
 
+    @classmethod
+    def save_docs(cls, docs, **kwargs):
+        utcnow = datetime.datetime.utcnow()
+        for doc in docs:
+            doc['last_modified'] = utcnow
+        super(ApplicationBase, cls).save_docs(docs, **kwargs)
+
```
gcapalbo: i think we should also make sure `bulk_save` gets updated too since i think it will still point to the old version: https://github.com/dimagi/couchdbkit/blob/jsonobject/couchdbkit/schema/base.py#L159

gcapalbo (approved):

PR #16285, opened 2017-05-23
Added way to filter on more than one field

kkrampa:


corehq/apps/userreports/reports/filters/factory.py
```diff
@@ -86,6 +86,22 @@ def _build_dynamic_choice_list_filter(spec, report):
...
 
+def _build_multi_field_dynamic_choice_list_filter(spec, report):
+    wrapped = MultiFieldDynamicChoiceFilterSpec.wrap(spec)
+    choice_provider_spec = wrapped.get_choice_provider_spec()
+    choice_provider = MultiFieldChoiceProviderFactory.from_spec(choice_provider_spec)(report, wrapped.slug)
+    choice_provider.configure(choice_provider_spec)
+    return DynamicChoiceListFilter(
+        name=wrapped.slug,
+        datatype=wrapped.datatype,
+        field=wrapped.fields,
```
kkrampa: I know that this looks weird but this field is used here https://github.com/dimagi/commcare-hq/pull/16285/files#diff-fe5cacb99e1e5c5f67991de667b352cdR195

custom/enikshay/ucr/reports/tb_notification_register.json
```diff
@@ -41,9 +41,10 @@
             {
                 "compare_as_string": false,
                 "datatype": "string",
-                "type": "dynamic_choice_list",
+                "type": "multi_field_dynamic_choice_list",
                 "slug": "village",
                 "field": "person_owner_id",
```
kkrampa: @NoahCarnahan I can't make field property not required due to this line https://github.com/dimagi/commcare-hq/blob/master/corehq/apps/userreports/reports/factory.py#L22. Is there any reason to not use report filter factory in that line?

corehq/apps/userreports/reports/filters/factory.py
emord: Instead of returning `DynamicChoiceListFilter` could you create a `MultiFieldDynamicChoiceLIstFilter` so that this can be `fields`? It would be much less confusing that way

kkrampa:
@emord I made changes that you've suggested.

I think @NoahCarnahan is offline. Could you merge this? it's pretty urgent for Enikshay project. I can add tests in further PR.

emord:
@kkrampa I'll deploy this today

PR #16287, opened 2017-05-23
Added way to clone data source

kkrampa:
@emord @nickpell 
It might be useful e.g. when we want to create data source based on static data source.

### DIFF
--- a/corehq/apps/userreports/views.py
+++ b/corehq/apps/userreports/views.py
@@ -1110,6 +1110,30 @@ class CreateDataSourceView(BaseEditDataSourceView):
     page_title = ugettext_lazy("Create Data Source")
 
 
+class CloneDataSourceView(CreateDataSourceView):
+    urlname = 'clone_configurable_data_source'
+
+    @property
+    def config(self):
+        data_source_to_clone = get_datasource_config_or_404(self.config_id, self.domain)[0]
+        return DataSourceConfiguration(
+            domain=self.domain,
+            referenced_doc_type=data_source_to_clone.referenced_doc_type,
+            table_id=data_source_to_clone.table_id,
+            description=data_source_to_clone.description,
+            base_item_expression=data_source_to_clone.base_item_expression,
+            configured_filter=data_source_to_clone.configured_filter,
+            configured_indicators=data_source_to_clone.configured_indicators,
+            named_expressions=data_source_to_clone.named_expressions,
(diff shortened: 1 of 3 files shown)


nickpell (approved):

nickpell:
How would I access this view?  Would I have to enter the url with the config id of the datasource I want to copy?  If so, is there a way to hook this up into a UI?

kkrampa:
@nickpell updated

corehq/apps/userreports/views.py
```diff
@@ -1110,6 +1110,37 @@ class CreateDataSourceView(BaseEditDataSourceView):
...
+    def page_url(self):
+        return reverse(self.urlname, kwargs=dict(domain=self.domain, config_id_to_clone=self.config_id_to_clone))
+
+    @property
+    def config(self):
+        data_source_to_clone = get_datasource_config_or_404(self.config_id_to_clone, self.domain)[0]
+        return DataSourceConfiguration(
+            domain=self.domain,
+            referenced_doc_type=data_source_to_clone.referenced_doc_type,
+            table_id=data_source_to_clone.table_id,
```
nickpell: will the new data source point to the same table?  If so that seems problematic.

kkrampa: ahh, you're right. That's an obvious mistake

kkrampa:
Updated

corehq/apps/userreports/views.py
```diff
@@ -1110,6 +1110,36 @@ class CreateDataSourceView(BaseEditDataSourceView):
...
+            configured_filter=data_source_to_clone.configured_filter,
+            configured_indicators=data_source_to_clone.configured_indicators,
+            named_expressions=data_source_to_clone.named_expressions,
+            named_filters=data_source_to_clone.named_filters,
+            display_name='{} - cloned {}'.format(
+                data_source_to_clone.display_name,
+                datetime.datetime.utcnow().strftime('%Y-%m-%d')
+            ),
+            backend_id=data_source_to_clone.backend_id
+        )
```
snopoke: the better approach to cloning json objects is to do this `new = DataSourceConfiguration.wrap(old.to_json())`

and then change any values on `new` that need changing.

snopoke (commented):
Not really sure about the need for this. @kkrampa is this something you've needed in the past?

I'm cautious to add functionality because it 'might' be useful in the future.

PR #16294, opened 2017-05-23
Make a new build when reverting an app

orangejenny:
@emord


emord (approved):

snopoke:
Interesting, I had thought this would make a build of the version prior to reverting as a safeguard. I've definitely seen users who have reverted and lost changes in the process.

I would be very pro making a pre-revet build as well as this.

orangejenny:
@snopoke good idea, added in https://github.com/dimagi/commcare-hq/pull/16314

PR #16295, opened 2017-05-23
Update enikshay adherence calendar report dose logic

NoahCarnahan:
QA Ticket: https://manage.dimagi.com/default.asp?254727#1336080

This PR updates the enikshay historical adherence report to use the same logic that the enikshay `EpisodeUpdater` task uses to calculate total doses.
It also updates the logic used to find the most relevant adherence case for purposes of displaying an icon on the calendar.

buddy @dannyroberts 
cc: @proteusvacuum

### DIFF
--- a/custom/enikshay/reports/historical_adherence.py
+++ b/custom/enikshay/reports/historical_adherence.py
@@ -231,8 +231,10 @@ class HistoricalAdherenceReport(EnikshayReport):
     def get_primary_adherence_case(self, adherence_cases):
         """
         Return the case who's adherence value should be used.
-        Cases with adherence_source == enikshay take precedence over other sources, and cases with a later
-        modified_on take precedence over earlier.
+        Cases with adherence_source == enikshay take precedence over other sources
+        Then open cases tak precedence over other cases
+        Then cases with a later modified_on take precedence over earlier cases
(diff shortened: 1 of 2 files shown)


custom/enikshay/reports/historical_adherence.py
```diff
@@ -240,7 +243,7 @@ def _source_is_enikshay(case):
             return case.dynamic_case_properties().get('adherence_source') in ('enikshay', '')
 
         return sorted(
-            adherence_cases, key=lambda c: (_source_is_enikshay(c), c.modified_on)
+            adherence_cases, key=lambda c: (_source_is_enikshay(c), not c.closed, c.modified_on, c.opened_on)
```
NoahCarnahan: If a second adherence for a given day is created through the app, the previously open case will be closed. Therefore both cases will have the same `modified_on` value, so we need to defer to `opened_on` to determine which case is older. We can't simply rely on open/close status because the cases could later both be closed by a periodic task that closes adherence cases older than some threshold.

NoahCarnahan: And the `not closed` part of the key was added because the mobile version of the calendar gives preference to open cases (because of course only open cases are synced to the device).

proteusvacuum (approved):

PR #16297, opened 2017-05-23
Don't allow create for bad passwords on icds

emord:
http://manage.dimagi.com/default.asp?254738#1336149

buddy @calellowitz

### DIFF
--- a/corehq/apps/users/static/users/js/mobile_workers.ng.js
+++ b/corehq/apps/users/static/users/js/mobile_workers.ng.js
@@ -379,6 +379,13 @@
                 ctrl.$validators.validatePassword = function (password) {
                     if (!password) {
                         return false;
+                    } else if (!(
+                        password.length >= 8 &&
+                        /\W/.test(password) &&
+                        /\d/.test(password) &&
+                        /[A-Z]/.test(password)
+                    )) {
+                        return false;
                     }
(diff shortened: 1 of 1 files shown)


emord:
bump for interrupt @benrudolph

corehq/apps/users/static/users/js/mobile_workers.ng.js
```diff
@@ -379,6 +379,13 @@
                 ctrl.$validators.validatePassword = function (password) {
                     if (!password) {
                         return false;
+                    } else if (!(
+                        password.length >= 8 &&
+                        /\W/.test(password) &&
+                        /\d/.test(password) &&
+                        /[A-Z]/.test(password)
```
benrudolph: is the requirement that it needs a lower case and an upper case? weird that it just wants an upper case letter

benrudolph (approved):

corehq/apps/users/static/users/js/mobile_workers.ng.js
emord: just uppercase according to https://github.com/dimagi/commcare-hq/blob/master/corehq/apps/domain/forms.py#L1210-L1222 probably just assume that there are lower case in the password as a default

PR #16299, opened 2017-05-23
Require aggregation columns

emord:
https://manage.dimagi.com/default.asp?254254#1333104

@NoahCarnahan @nickpell buddy @calellowitz


nickpell (approved):

calellowitz:
test failures are related. i think you will need a migration or a default value to deal with existing ones without migrations

emord:
Yeah. this was to fix an edge case. Realistically it's not worth writing a migration because thsi can only be hit with custom UCRs

NoahCarnahan:
Ah that's annoying that it will require a migration. It would definitely nice to make this required eventually.

PR #16300, opened 2017-05-23
Fixes to List Apps

proteusvacuum:
A few fixes to the new install workflow in 2.36:
- Don't pull deleted apps
https://manage.dimagi.com/default.asp?253790
- Use `profile.ccpr` instead of `profile.xml`
https://manage.dimagi.com/default.asp?254496

@amstone326 
@benrudolph buddy

### DIFF
--- a/corehq/apps/app_manager/tests/data/list_apps/list_apps.xml
+++ b/corehq/apps/app_manager/tests/data/list_apps/list_apps.xml
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <apps>
-  <app profile="http://onering.com/a/rohan/apps/download/app_id/profile.xml?latest=true" domain="rohan" name="Untitled Application" environment="production" media-profile="http://onering.com/a/rohan/apps/download/app_id/media_profile.xml?latest=true"/>
-  <app profile="http://onering.com/a/angmar/apps/download/app_id/profile.xml?latest=true" domain="angmar" name="Untitled Application" environment="production" media-profile="http://onering.com/a/angmar/apps/download/app_id/media_profile.xml?latest=true"/>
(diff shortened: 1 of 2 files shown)


benrudolph (approved):

corehq/apps/app_manager/tests/data/list_apps/list_apps.xml
```diff
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <apps>
-  <app profile="http://onering.com/a/rohan/apps/download/app_id/profile.xml?latest=true" domain="rohan" name="Untitled Application" environment="production" media-profile="http://onering.com/a/rohan/apps/download/app_id/media_profile.xml?latest=true"/>
```
amstone326: Hey, just curious why you decided to remove the `?latest=true` part? We never discussed which version of an app (latest saved vs. build vs. starred) it would be best for this to return, so there's definitely not a clear answer. I was envisioning that in a later iteration of this I would allow users to configure that anyway, so not a big deal for now. Does this mean that right now, this will return the latest starred version of the app?

proteusvacuum: hey, the way I set this up, it always returns only latest starred versions. 
I'm not actually sure why the previous url contains `latest=true`, as I don't think that parameter does anything. The thing that gets the "version" in the profile url is confusingly the `app_id` (which here is just mocked out as `app_id`).

At a later date, we can definitely let the user select which build they want, and we can discuss how we want that to happen.

proteusvacuum: the `app_id` _is_ the version. 
We create a new `Application` document in the database whenever we make a build which is a "copy of" the actual `Application` at that point in time. This new document has its own `app_id`, which corresponds to a particular version of the app.

PR #16307, opened 2017-05-24
update form name in form navigation upon change of form name within formbuilder

biyeun:
@orangejenny


orangejenny (approved):

orangejenny:
Can't you use `updateDOM` here? It'd be more standard.

biyeun:
@orangejenny wasn't aware of `updateDOM`

PR #16308, opened 2017-05-24
App V2: Update Look and Feel of Starred Builds to use Toggle Switch

biyeun:
@orangejenny cc @gcapalbo

fyi @proteusvacuum

screen shot 2017-06-05 at 4 16 38 pm

screen shot 2017-06-05 at 4 21 19 pm

### DIFF
--- a/corehq/apps/app_manager/static/app_manager/js/app_manager.js
+++ b/corehq/apps/app_manager/static/app_manager/js/app_manager.js
@@ -1,4 +1,4 @@
-/* globals hqDefine COMMCAREHQ django */
+/* globals hqDefine COMMCAREHQ django hqLayout */
 hqDefine('app_manager/js/app_manager.js', function () {
     'use strict';
     var module = eventize({});
@@ -92,7 +92,7 @@ hqDefine('app_manager/js/app_manager.js', function () {
         } else {
             $(hqLayout.selector.publishStatus).fadeOut();
         }
-    }
+    };
 
     module.init = function (args) {
 
@@ -178,7 +178,7 @@ hqDefine('app_manager/js/app_manager.js', function () {
                 url: currentAppVersionUrl,
                 success: function (data) {
                     module.setPublishStatus((!data.latestRelease && data.currentVersion > 1) || (data.latestRelease !== null && data.latestRelease < data.currentVersion));
-                }
+                },
             });
         };
         _checkPublishStatus();
--- a/corehq/apps/app_manager/static/app_manager/js/app_manager.js
+++ b/corehq/apps/app_manager/static/app_manager/js/app_manager.js
@@ -86,6 +86,14 @@ hqDefine('app_manager/js/app_manager.js', function () {
         }
     };
 
+    module.setPublishStatus = function (isOn) {
+        if (isOn) {
+            $(hqLayout.selector.publishStatus).fadeIn();
+        } else {
+            $(hqLayout.selector.publishStatus).fadeOut();
+        }
+    }
+
     module.init = function (args) {
 
         _initCommcareVersion(args);
@@ -97,6 +105,7 @@ hqDefine('app_manager/js/app_manager.js', function () {
         if (COMMCAREHQ.toggleEnabled('APP_MANAGER_V2')) {
             _initResponsiveMenus();
             _initAddItemPopovers();
+            _initPublishStatus();
         } else {
             // legacy JS
             $('#form-tabs').show();
@@ -160,6 +169,34 @@ hqDefine('app_manager/js/app_manager.js', function () {
             });
         });
         module.setCommcareVersion(args.commcareVersion);
+    };
+
+    var _initPublishStatus = function () {
+        var currentAppVersionUrl = hqImport('hqwebapp/js/initial_page_data.js').get('current_app_version_url');
+        var _checkPublishStatus = function () {
+            $.ajax({
+                url: currentAppVersionUrl,
+                success: function (data) {
+                    module.setPublishStatus((!data.latestRelease && data.currentVersion > 1) || (data.latestRelease !== null && data.latestRelease < data.currentVersion));
+                }
+            });
+        };
+        _checkPublishStatus();
+        // check publish status every 20 seconds
+        setInterval(_checkPublishStatus, 20000);
+
+        // sniff ajax calls to other urls that make app changes
+        $(document).ajaxComplete(function(e, xhr, options) {
+            if (/edit_form_attr/.test(options.url) ||
+                /edit_module_attr/.test(options.url) ||
+                /edit_module_detail_screens/.test(options.url) ||
+                /edit_app_attr/.test(options.url) ||
+                /edit_form_actions/.test(options.url) ||
+                /edit_commcare_settings/.test(options.url) ||
+                /patch_xform/.test(options.url)) {
+                module.setPublishStatus(true);
+            }
+        });
     };
 
     /**
@@ -392,6 +429,8 @@ hqDefine('app_manager/js/app_manager.js', function () {
                             }
                             $form.submit();
                         }
+
+                        module.setPublishStatus(true);
                     }
                 };
                 if (sorting_forms) {
--- a/corehq/apps/app_manager/static/app_manager/js/apps_base.js
+++ b/corehq/apps/app_manager/static/app_manager/js/apps_base.js
@@ -19,4 +19,16 @@ $(function () {
     if (v2) {
         hqLayout.utils.setIsAppbuilderResizing(true);
     }
+
+    $(document).ajaxComplete(function(e, xhr, options) {
+        if (/edit_form_attr/.test(options.url) ||
+            /edit_module_attr/.test(options.url) ||
+            /edit_module_detail_screens/.test(options.url) ||
+            /edit_app_attr/.test(options.url) ||
+            /edit_form_actions/.test(options.url) ||
+            /edit_commcare_settings/.test(options.url) ||
+            /patch_xform/.test(options.url)) {
+            $(hqLayout.selector.publishStatus).fadeIn();
+        }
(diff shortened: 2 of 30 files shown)


corehq/apps/app_manager/static/app_manager/js/releases.js
```diff
@@ -334,6 +334,17 @@ hqDefine('app_manager/js/releases.js', function () {
             });
         };
 
+
+
+        self.handleAfterRender = function (elem, savedApp) {
+            $(elem).find('.js-auto-update-checkbox').bootstrapSwitch({
+                state: savedApp.is_released(),
+                onSwitchChange: function () {
+                    self.toggleRelease(savedApp)
```
dimagimon: Missing semicolon. (semi)

corehq/apps/app_manager/static/app_manager/js/releases.js
```diff
@@ -334,6 +334,17 @@ hqDefine('app_manager/js/releases.js', function () {
...
+                }
```
dimagimon: Missing trailing comma. (comma-dangle)

orangejenny:
Hey, I'm a bit confused on this. I thought the feedback on the previous iteration of this was just that the legend in the top right looked clickable, not that the general interaction of clicking on a star, as people do in v1, is confusing. The idea, and the language, of "starring" a build is a rather convenient shorthand, and replacing it with "auto update" language seems clunky.

proteusvacuum:
Hey,
I somewhat agree with @orangejenny , however I do think this is an overall improvement over a "star" which has no meaning really. 

I have a few comments:
- I think in general with those switches it is much nicer to also show what will happen when you click it, so have `On|Off`, where off is greyed out, so that you are clicking on the action that you want to have happen (e.g. https://ghinda.net/css-toggle-switch/bootstrap.html)
- That top bar is very busy, and adding more words to it seems like a shame. I also think being able to scan down the side to see which are starred and aren't is nice to be able to do. What if you changed "On|Off" to something like "Released | Test Build" or something (wording to be worked out). This way we don't need a legend (I think if we need a legend, the UI element we've chosen isn't clear enough).

biyeun:
@orangejenny @proteusvacuum updated this to Farid's recommendation. The starred build is a lot more ambiguous and requires way more documentation than a released / test switch imo. This feels more straightforward.

orangejenny:
Code looks good.

On the interaction: I agree with Farid that it's nicer to be able to see what both options are. The bootstrap switch is sleek, but the basic `.btn-group` applied to radio buttons would be more usable.

On release: can you get Clark/Kevin's opinion on whether we should time documentation updates to happen at the same time that this is released? (Or if they're fine with the docs continuing to reference stars/starring for now? I don't have a good sense of how often that shows up in docs.)

proteusvacuum:
Agreed with @orangejenny re the button group (or just showing what the other option is) also, can we make "released" be green, or some other positive looking colour? (the ⭐️ is nice because it is bright and yellow)

Like the updates so far!

snopoke:
nice, I like the outcome here. One note on the 'released' help text - it should also mention something like `or if the user clicks 'Update App'`

proteusvacuum:
I was thinking about this while brushing my teeth this morning. What if we generate the bitly app code when a build is switched to "Released" and show it on this page? I think that would be a nice lower-friction way of getting the app on your phone. (Or even having a mini QR code that you can click either scan or click to enlarge). 

Also, if this is going to be the "Publish" page, maybe the action should be "Published" instead of "Released".

biyeun:
Hi all, got back from vacation yesterday and still catching up with some stuff and performance evals, but I'm really loving the suggestions here. Thank you! Will be playing around with some of the ideas presented and will hopefully have something to look at shortly. @proteusvacuum @orangejenny @snopoke

biyeun:
@proteusvacuum I love that you were thinking of the deploy page while brushing your teeth. :joy: :heart:

biyeun:
@proteusvacuum there was some chatter about the release panels already being too high. Adding the barcode / bitly code might just be too much.

@orangejenny  I did the thing with hiding the comments unless you fill it in, removing one line of vertical space.
screen shot 2017-06-05 at 4 12 55 pm

biyeun:
@orangejenny @proteusvacuum @snopoke see previous comment for the updated button group version of Released/In Test

biyeun:
@snopoke here is the new help copy 
screen shot 2017-06-05 at 4 21 19 pm

orangejenny:
ooh, I like that in test vs released are in different colors (it's the little details...), and the more compressed spacing when there aren't comments

orangejenny (approved):
Code looks good. Still think it's worth checking in with @dimagi/product about documentation changes.

biyeun:
added the hold off on merge as we're going to bundle this with an additional feature to auto detect when a new build is possible and highlight the app name as a clickable region

biyeun:
@orangejenny 

updated
- removed publish button
- added indication to publish when changes are available
- added edit form button on form settings pages

screen shot 2017-06-13 at 12 58 12 am
screen shot 2017-06-13 at 12 46 36 am

proteusvacuum:
yay! Edit Form!

I'm wondering if it will be hard to find how you download the app you've already published to your phone if that little banner isn't there?

orangejenny:
Nice! I really like the banner, I'd had trouble imagining what a good indicator would look like.

corehq/apps/app_manager/static/app_manager/js/apps_base.js
```diff
@@ -19,4 +19,16 @@ $(function () {
     if (v2) {
         hqLayout.utils.setIsAppbuilderResizing(true);
     }
+
+    $(document).ajaxComplete(function(e, xhr, options) {
+        if (/edit_form_attr/.test(options.url) ||
```
orangejenny: The url matching seems potentially fragile as we add new urls (or switch existing urls to use ajax). What would you think of doing something like firing an additional request to check the version number? It'd be another request, but a short one, and would be more future-proof.

orangejenny:
What would you think of having the new changes indicator also show on page load if there have been changes since the last build?

biyeun:
@orangejenny I like the idea of having the indicator show up on page load. I'll work on that in a few. Thanks!

corehq/apps/app_manager/static/app_manager/js/apps_base.js
biyeun: Yes, this sounds like a very good idea. I was following the lead from app preview, mainly.

benrudolph: 👍 sorry i definitely wrote the first iteration of this because it was the easiest thing to get out the door. the one issue i see with the additional request is that there will likely be a lot of false positives if the request is made here since this triggers after any ajax request is made. essentially we want this to update when someone hits the save button, but it felt really hacky to splice that in there

orangejenny: not quite following - what's the case where false positives would happen? i was thinking of the version-checking because it's what we do on the releases page to verify that you can/can't make a new build.

benrudolph: any other ajax call on the page would trigger this callback, so we'd check the version regardless if it were a save or not. (may naively trigger an infinite loop, the call to check the version will trigger this callback method again)

orangejenny: Oh, I see. Seems like a question then of whether to store the set of "saving" urls or the set of "non-saving" urls (at least the version check url).

biyeun:
this now detects publish changes on page load (and every 20 seconds) @orangejenny

biyeun:
@orangejenny fyi, the publish button was in fact removed from formdesigner here

corehq/apps/app_manager/static/app_manager/js/app_manager.js
```diff
@@ -141,6 +150,34 @@ hqDefine('app_manager/js/app_manager.js', function () {
...
+            $.ajax({
+                url: currentAppVersionUrl,
+                success: function (data) {
+                    module.setPublishStatus((!data.latestRelease && data.currentVersion > 1) || (data.latestRelease !== null && data.latestRelease < data.currentVersion));
+                }
+            });
+        };
+        _checkPublishStatus();
+        // check publish status every 20 seconds
+        setInterval(_checkPublishStatus, 20000);
```
orangejenny: This could probably be less frequent given that the code is also checking for known app-modifying ajax calls below.

orangejenny:
@biyeun Nice. Code looks fine, looks like v2 diffs are out of date.

biyeun:
@orangejenny do you think this is g2g? I don't want to keep piling on the features here... :P

orangejenny:
It's good from my perspective. I think @krsdimagi wanted to communicate with the field team specifically about this one, would check that he's okay with merging.

biyeun:
@orangejenny moved the app preview button to a turquiose bar along the left side of app preview 
screen shot 2017-06-19 at 1 36 24 pm
screen shot 2017-06-19 at 1 36 30 pm
screen shot 2017-06-19 at 1 39 44 pm
screen shot 2017-06-19 at 1 39 47 pm
screen shot 2017-06-19 at 1 39 50 pm

orangejenny (approved):

biyeun:
@proteusvacuum 
took care of full screen formbuilder and preview

screen shot 2017-06-19 at 2 29 49 pm

orangejenny (approved):

biyeun:
Should be set to go @orangejenny 
cc @proteusvacuum 

This is the notification
screen shot 2017-07-14 at 1 25 26 pm

biyeun:
merging since all prior changes were approved...

orangejenny:
looks great!

PR #16309, opened 2017-05-24
fix call to get_form_by_xmlns

benrudolph:
fyi @NoahCarnahan breaks old exports. may be worth an early deploy?


NoahCarnahan:
Darn... not sure how I missed this. The shadow forms have different actions than their parents, so I've modified the PR a little to check the actions of each `Form`.

NoahCarnahan (approved):
(I approve your part, but perhaps you should approve my part)

benrudolph:
can't approve since i made the PR, but I also approve your part. makes sense ✅

PR #16312, opened 2017-05-24
Add always_enabled to enikshay toggles

NoahCarnahan:
Re: https://github.com/dimagi/commcare-hq/commit/1d630ddc5563d7fef297893a912786b24f765e48#commitcomment-22263747

buddy @dannyroberts 
@proteusvacuum @esoergel let me know if you think any of the flags shouldn't remain dynamic.

### DIFF
--- a/corehq/toggles.py
+++ b/corehq/toggles.py
@@ -1072,7 +1072,6 @@ ENIKSHAY = StaticToggle(
     "Enable custom enikshay functionality: additional user and location validation",
     TAG_ONE_OFF,
     namespaces=[NAMESPACE_DOMAIN],
-    always_enabled={"enikshay"},
 )
 
 DATA_DICTIONARY = StaticToggle(
--- a/corehq/toggles.py
+++ b/corehq/toggles.py
@@ -770,7 +770,8 @@ ENIKSHAY_API = StaticToggle(
     'enikshay_api',
     'Enable access to eNikshay api endpoints',
     TAG_ONE_OFF,
-    [NAMESPACE_USER]
+    [NAMESPACE_USER],
+    always_enabled={"enikshay"},
 )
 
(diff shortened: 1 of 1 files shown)


corehq/toggles.py
```diff
@@ -1070,6 +1072,7 @@ def _commtrackify(domain_name, toggle_is_enabled):
     "Enable custom enikshay functionality: additional user and location validation",
     TAG_ONE_OFF,
     namespaces=[NAMESPACE_DOMAIN],
+    always_enabled={"enikshay"},
```
esoergel: Actually I'd like this one dynamic at least for now, so we can do UAT on prod and turn off if it's overly restrictive.

NoahCarnahan: updated

snopoke (approved):

snopoke:
Really feels like feature flags aren't the right thing to use here in the long term and that a static domain check would be better. We really don't want random domains getting added to these toggles which is possible at the moment. Nothing to do for this PR, just musing.

esoergel (approved):

proteusvacuum:
curious what the impetus for this is?

NoahCarnahan:
Based on this suggestion: https://github.com/dimagi/commcare-hq/commit/1d630ddc5563d7fef297893a912786b24f765e48#commitcomment-22263747

proteusvacuum:
Do these still show up in the list of feature flagged domains? I'm a bit worried that this will become confusing in the long run.

snopoke:
I don't think they do though it wouldn't be hard to add then to the UI. The
big advantage of having these here is that they don't require a DB / cache
hit. In the long run I think we should have a different type for flag that
isn't configurable via the UI for these 'one-off' flags that are only
needed on a specific set of domains.

PR #16314, opened 2017-05-24
Make build before reverting

orangejenny:
@snopoke / @esoergel

### DIFF
--- a/corehq/apps/app_manager/views/releases.py
+++ b/corehq/apps/app_manager/views/releases.py
@@ -257,16 +257,6 @@ def revert_to_copy(request, domain, app_id):
     """
     app = get_app(domain, app_id)
     copy = get_app(domain, request.POST['saved_app'])
-
-    errors = app.validate_app()
-    if not errors:
-        before = app.make_build(
-            comment=_("Auto-generated before reverting to version %s") % copy.version,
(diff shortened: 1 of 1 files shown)


snopoke:
I confess I didn't think this through very carefully so apologies for that. I think this is actually slightly worse than the current situation since a user may expect that a build will get created but then it doesn't because their app has errors but there's no feedback to them about it.

I see 4 ways forward:

1. Revert to current state and don't do anything
2. Try to create the pre-revert build if the app has errors inform the user and require them to accept the consequences (loss of changes) or cancel the revert.
3. Don't auto create a build but before reverting check if there are changes in the app since the last build and if so inform the user that going ahead with the revert will result in losing their changes since the last build.
4. Save the app source in some other record which would allow restoring it if necessary (like a  `DeleteRecord` doc) but wouldn't care if there were errors in the app.

Feel free to declare this out of scope for now (option 1).

orangejenny:
@snopoke I see your point. Option 3 seems like a decent interaction and possibly not a lot of effort, I'll take a stab at it.

orangejenny:
Took a look at warning the user when there have been changes, found the knockout on the deploy page is a bit finicky, don't think it's worth spending time on right now. Closing this for the time being.

PR #16316, opened 2017-05-24
add deprecated_format_names to payload generators

dannyroberts:
to store aliases for backwards compatability

I made this change on master but decided to cherry pick this commit directly on top of the last commit in https://github.com/dimagi/commcare-hq/pull/16214 so that you can see the "whole" change together (i.e. that PR + this PR): [16214 + 16316](https://github.com/dimagi/commcare-hq/compare/79ac1b9746de1ffd51138c127d736cfd969f4c93...deprecated_format_names?diff=unified). That view was helpful in convincing me that I had in fact found all of the places where I needed to set `deprecated_format_names` because the diff includes what they were set to before.

### DIFF
--- a/corehq/apps/repeaters/repeater_generators.py
+++ b/corehq/apps/repeaters/repeater_generators.py
@@ -29,6 +29,9 @@ class BasePayloadGenerator(object):
     format_name = ''
     format_label = ""
 
+    # if you ever change format_name, add the old format_name here for backwards compatability
+    deprecated_format_names = ()
+
     def __init__(self, repeater):
         self.repeater = repeater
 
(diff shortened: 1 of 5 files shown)


dannyroberts:
@proteusvacuum @mkangia @NoahCarnahan

dannyroberts:
Once this is merged, I'm planning on running this in the production shell

```
from corehq.apps.repeaters.dbaccessors import iter_repeat_records_by_domain
records = iter_repeat_records_by_domain('enikshay', state='CANCELLED')
records = [record for record in records if record.failure_reason == "u'case_json'"]

for record in records:
    record.requeue()
    record.save()
```

That should be enough for that domain. Seems like this is common enough a thing though that I should write a management command to do it, where I can search on all domains.

mkangia:
Thanks @dannyroberts 
Do we need to add `case_xml` for `NikshayRegisterPrivatePatientPayloadGenerator`?

dannyroberts:
@mkangia it's new right? If it was never deployed before https://github.com/dimagi/commcare-hq/pull/16214, then it won't have this problem. The problem arises when existing `Repeater`s have `self.format` set to one thing (e.g. `case_json`) but then the `format_name` of the payload generator is changed. In this case, I changed all of the `format_name`s to `''`, but existing repeaters were referencing `'case_json'`. Kinda confusing, let me know if you want more!

mkangia:
@dannyroberts it definitely was merged after this change that deprecated formats. Was just checking.

proteusvacuum:
cool beans. This was a tricksy one.

PR #16320, opened 2017-05-24
Add repeater dbaccessors that make it easy to find and requeue cancelled records

dannyroberts:


dannyroberts:
`sniffer -x --failfast -x corehq.apps.repeaters` is passing locally, I'm going to sneak this in to deploy

PR #16321, opened 2017-05-24
Don't include fullstory if domain restricts superusers

orangejenny:
@esoergel / @calellowitz


esoergel (approved):

calellowitz:
Is the request guaranteed to have a project (and a `restrict_superusers` attribute). I am wondering about admin pages and the login page.

orangejenny:
It isn't guaranteed, but because it's in a template if `project` isn't there, there's no error thrown and `request.project.restrict_superusers` is falsy.

PR #16322, opened 2017-05-24
Use issuer ID as username

esoergel:
@orangejenny @proteusvacuum

### DIFF
--- a/custom/enikshay/user_setup.py
+++ b/custom/enikshay/user_setup.py
@@ -7,13 +7,14 @@ assumptions laid out here.
 import math
 import re
 import uuid
+from django.forms import ValidationError
 from django.utils.text import slugify
 from django.utils.translation import ugettext as _
 from dimagi.utils.decorators.memoized import memoized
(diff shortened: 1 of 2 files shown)


custom/enikshay/user_setup.py
```diff
@@ -271,13 +273,59 @@ def __init__(self, *args, **kwargs):
         super(ENikshayUserDataEditor, self).__init__(*args, **kwargs)
 
 
+def get_new_username_and_id(domain, attempts_remaining=3):
+    if attempts_remaining <= 0:
+        raise AssertionError(
+            "3 IssuerIds were created, but they all corresponded to existing "
+            "users.  Are there a bunch of users with usernames matching "
+            "possible compressed ids?  Better investigate.")
```
proteusvacuum: would it be useful to know what the conflicting issuer ids were to help debug?

esoergel: Yeah, probably - I'll add something in.

PR #16323, opened 2017-05-24
do not try to adjust repeat record attempt numbers

dannyroberts:
this was originally for repeat records that had been tried a few times
but were created before attempts were all saved.
This logic did not quite work, because overall_tries is reset to 0 when you requeue.
Decided it was not really that important, so just removing it.

Prompted because I was seeing weird numbering for records that had been requeued:
screen shot 2017-05-24 at 4 59 45 pm


dannyroberts:

proteusvacuum:
👢 👷

proteusvacuum (approved):

PR #16326, opened 2017-05-25
User staging table

benrudolph:
@gcapalbo needed for the initial draft of the app status fact table (requires the user dim)

cc: @proteusvacuum


corehq/warehouse/models/staging.py
```diff
@@ -74,6 +76,49 @@ def raw_record_iter(cls, start_datetime, end_datetime):
         return iter_docs(Group.get_db(), group_ids)
 
 
+class UserStagingTable(StagingTable):
+    slug = 'user_staging'
+
```
gcapalbo: i think it would be good to bring in base_doc on the staging table, and transform to a deleted flag on the dim

gcapalbo:
:+1: just one comment otherwise looks good

corehq/warehouse/models/staging.py
benrudolph: 👍 good call

gcapalbo (approved):

gcapalbo:
(build failure looks related)

benrudolph:
merging because this was approved and last commit was just test cleanup code

PR #16327, opened 2017-05-25
Initial POC of full warehouse cycle

benrudolph:
This isn't a completely perfect PR and I know there is some slop that I'm going to go back and fix up. Before diving too deeply into perfecting this, I wanted to get general thoughts on the process and ask some questions. The basic process for updating a warehouse table is:

```
./manage.py stage_table <staging_slug> -s 2012-05-05 -e 2017-06-05 (for all deps)
./manage.py commit_table <dim/fact>
```

fyi: @dimagi/scale-team cc: @proteusvacuum


corehq/warehouse/models/shared.py
```diff
@@ -0,0 +1,86 @@
...
+
+from corehq.sql_db.routers import db_for_read_write
+
+from corehq.warehouse.transforms import get_transform
+from corehq.warehouse.const import (
+    DJANGO_MAX_BATCH_SIZE,
+)
+
+
+class WarehouseTableMixin(object):
```
benrudolph: I would say the heart of the logic lies within this class

corehq/warehouse/models/shared.py
```diff
@@ -0,0 +1,86 @@
...
+    @classmethod
+    def sql_query_template_name(cls):
+        return None
+
+    @classmethod
+    def dependencies(cls):
+        raise NotImplementedError
+
+    @classmethod
+    def load(cls):
```
benrudolph: loading is going to be annoying. it seems far away to upgrade to postgres 9.5 for upsert. the alternative is to implement something like this: https://www.postgresql.org/docs/9.4/static/plpgsql-control-structures.html#PLPGSQL-UPSERT-EXAMPLE

corehq/warehouse/models/shared.py
```diff
@@ -0,0 +1,86 @@
...
+        '''
+        Bulk loads records for a dim or fact table from
+        their corresponding dependencies
+        '''
+        for record in cls.record_iter():
+            print record
+
+    @classmethod
+    def _table_context(cls):
```
benrudolph: we need a dictionary to map slugs to table names so we can test queries in django tests

corehq/warehouse/models/shared.py
```diff
@@ -0,0 +1,86 @@
...
+
+    @classmethod
+    def sql_query_template(cls, template_name):
+        path = os.path.join(
+            settings.BASE_DIR,
+            'corehq',
+            'warehouse',
+            'transforms',
+            'sql',
+            '{}.sql'.format(template_name),
```
benrudolph: this dynamically assumes that there will be a sql file named the same thing as the slug to extract all the records it needs. i tend to prefer convention over configuration

corehq/warehouse/models/shared.py
```diff
@@ -0,0 +1,86 @@
...
+
+    @classmethod
+    def record_iter(cls):
+        '''
+        Returns an iterator over all records to be updated in
+        the table.
+        '''
+        database = db_for_read_write(cls)
+        with connections[database].cursor() as cursor:
+            cursor.execute(cls.sql_query_template(cls.slug))
```
benrudolph: this'll execute the sql query defined to extract the records from the dependent tables and then run each row over a python transform function. this'll be enough for app status and all the dims, but i'm not sure if it'll be enough customization for every fact table (i.e. maybe some need two sql queries). for now YAGNI

corehq/warehouse/models/staging.py
```diff
@@ -167,3 +177,70 @@ def raw_record_iter(cls, start_datetime, end_datetime):
...
+    app_id = models.CharField(max_length=255, null=True)
+    xmlns = models.CharField(max_length=255, default=None)
+    user_id = models.CharField(max_length=255, null=True)
+
+    # The time at which the server has received the form
+    received_on = models.DateTimeField(db_index=True)
+
+    build_id = models.CharField(max_length=255, null=True)
+
+    # TODO: Should we be storing state or will reports only deal with Normal forms?
```
benrudolph: right now i'm assumption is we don't need to think about anything but normal forms

corehq/warehouse/transforms/__init__.py
```diff
@@ -0,0 +1,17 @@
+from corehq.warehouse.const import (
+    APP_STATUS_FACT_SLUG,
+)
+
+
+def app_status_transform(row):
+    print row
```
benrudolph: this will actually fetch things like the commcare version that are hard to get in SQL

corehq/warehouse/transforms/sql/app_status_fact.sql
```diff
@@ -0,0 +1,45 @@
+SELECT * FROM
```
benrudolph: this is a fairly complex sql query that i still need to tweak a bit. for one i need to ensure its accuracy and two need to join it with the user dim. i'll also need to document this sql query to make it easier to understand, but these are the types of queries we're looking at...

this currently returns:
```
{'build_id': None, 'user_id': u'a191f7eba69a44c9bf80e0f8ec4c7923', 'app_id': u'be5192c04e0ff690cd6de44fdae9531b', 'last_sync': datetime.datetime(2017, 2, 1, 10, 11, 4, 182407), 'last_submission': datetime.datetime(2017, 2, 1, 14, 13, 51, 283805), 'last_submission_id': u'399929e1-458c-4be0-a933-64570c03864e'}
```

corehq/warehouse/models/facts.py
```diff
@@ -0,0 +1,44 @@
...
+
+from .dimensions import UserDim
+from .shared import WarehouseTableMixin
+
+
+class ApplicationStatusFact(models.Model, WarehouseTableMixin):
+    '''
+    Application Status Report Fact Table
+
+    Grain: app_id, user_id
```
gcapalbo: great to see this! we should include the grain in our docs for all facts and dims

corehq/warehouse/models/facts.py
```diff
@@ -0,0 +1,44 @@
...
+    '''
+    slug = APP_STATUS_FACT_SLUG
+
+    # app_dim = models.CharField(max_length=255)
+
+    # TODO is CASCADE the functionality that we want?
```
gcapalbo: good question. i don't think it will actually matter since we won't be using orm to populate the dims and facts probably, but i think PROTECT might be better to prevent accidental deletion of data

corehq/warehouse/models/facts.py
```diff
@@ -0,0 +1,44 @@
...
+    last_form_submission_date = models.DateTimeField(null=True)
+    last_sync_log_date = models.DateTimeField(null=True)
+
+    last_form_app_build_version = models.CharField(max_length=255)
+    last_form_app_commcare_version = models.CharField(max_length=255)
+    last_form_app_source = models.CharField(max_length=255)
+
+    last_sync_log_app_build_version = models.CharField(max_length=255)
+    last_sync_log_app_commcare_version = models.CharField(max_length=255)
+    last_sync_log_app_source = models.CharField(max_length=255)
```
gcapalbo: i see you denormalized a lot of this - i think typically we'd try to point to the relevant dims instead, since we want to make it as easy as possible to build new reports or new filters from the same set of tables

corehq/warehouse/models/shared.py
gcapalbo: we should avoid updating/inserting one row at a time. we can actually load the dim in just a few bulk queries using an extra intermediate table which has the same exact structure as the final dim and also includes one extra boolean column: is_new, which is set to true by default.  the process would be:

1) truncate intermediate table and insert data from staging table into intermediate table, using an insert ... select statement, and perform any transformations along with the select (e.g., doc_type -> deleted flag)

2) update the intermediate table using an update ... from statement, joining with the dim on doc id, and setting the is_new flag to false for anything that matches

3) update the dim using an update ... from statement, joining with the intermediate table on doc id

4) insert into the dim selecting from the intermediate table where is_new is true

5) vacuum analyze the dim (this should be the last step of every dim / fact loading process)

this should suffice for almost all our dims. how we load our facts will be dependent on the complexity of the fact table, and each one might be different

corehq/warehouse/models/shared.py
gcapalbo: i alluded to this above, but it would be more efficient to apply transformations in an insert ... select statement

corehq/warehouse/models/staging.py
gcapalbo: it might be a good idea to actually build a form_fact table, which stores all forms (in all states). this will make it much easier to build any new reports that are based off forms since we'll always have the full history

corehq/warehouse/models/staging.py
```diff
@@ -167,3 +177,70 @@ def raw_record_iter(cls, start_datetime, end_datetime):
...
+            ('xmlns', 'xmlns'),
+            ('user_id', 'user_id'),
+
+            ('received_on', 'received_on'),
+            ('build_id', 'build_id'),
+        ]
+
+    @classmethod
+    def raw_record_iter(cls, start_datetime, end_datetime):
+        return get_forms_by_submission_date(start_datetime, end_datetime)
```
gcapalbo: i noticed we're pulling on received_on. is that still going to work if a form is archived, deleted, or edited?

corehq/warehouse/transforms/__init__.py
gcapalbo: can you explain what this transform is a bit more? once we get past the staging of the data we should really try to do everything in bulk sql statements

corehq/warehouse/transforms/sql/app_status_fact.sql
gcapalbo: we should almost always avoid subqueries, and especially avoid filtering or joining on a result from a subquery. to get around that, we can define any number of intermediate tables which we first load and then join to each other or do any other filtering from

corehq/warehouse/models/shared.py
benrudolph: interesting, good to know. and the reason we couldn't do something like a `bulk_create` is that we do not know which records to update vs which records to insert? is it then true that if we did have the upsert command we could bypass the intermediate table?

corehq/warehouse/models/staging.py
benrudolph: do you think it might be too expensive to create an entire `form_fact` table? i think that would mean we'd need to immediately setup sharding for icds

corehq/warehouse/transforms/__init__.py
benrudolph: so i definitely assumed we'd have a portion where we'd be able to run some python function over a row in order to perform transformations. some of the transforms get quite tricky, for example of getting the commcare version, you can see this function: https://github.com/dimagi/commcare-hq/blob/master/corehq/apps/receiverwrapper/util.py#L146. maybe this is me just misunderstanding the way these things work, but couldn't we still do bulk sql statements if we iterate over the result set and then do another bulk update?

corehq/warehouse/transforms/sql/app_status_fact.sql
benrudolph: gotcha, heh that's gonna make for a lot more staging tables. would make any sense to use materialized views for our temporary tables?

corehq/warehouse/models/facts.py
benrudolph: 👍

corehq/warehouse/models/facts.py
benrudolph: so getting this info is a bit tricky and for that reason i thought it would make sense to just do it in python. it's possible we don't even need some of these fields like `last_sync_log_app_source`. to attach this to a dim, i think we'd need an applicationdim, and then connect both form and sync log to it via `sync_build_dim` and `form_build_dim`

corehq/warehouse/models/staging.py
benrudolph: yea, i noticed that too, but we don't have a last_modified field on forms (surprise!). we could almost use a combination of `received_on`, `deleted_on`, and `edited_on` to get the last time it was modified, however that still doesn't cover the case where the form is archived

benrudolph:
👍 thanks for the feedback. might be good to discuss this in the scale meeting today

corehq/warehouse/models/shared.py
gcapalbo: i haven't used upsert before but it looks like it would solve this problem without the need for another intermediate table

corehq/warehouse/models/staging.py
gcapalbo: i think it's a trade off. it looks like we're currently not storing form data in the warehouse so every time we needed to build a new report that depended on it we'd need to pull all form data again from the source

corehq/warehouse/transforms/__init__.py
gcapalbo: if we really need to do python transforms i think we could select from the staging table, perform the transformations in python, and then do a bulk insert into an intermediate table, and then update the final table from there. we basically want to avoid updating one record at a time

corehq/warehouse/transforms/sql/app_status_fact.sql
gcapalbo: performance-wise i think it wouldn't matter. doing truncate..insert just makes it more explicit i think, and makes it easier to make changes to the process

corehq/warehouse/models/facts.py
gcapalbo: yeah i think if those are dims that we would need at some other point it might be good to build them out now, but again is a trade off

corehq/warehouse/models/staging.py
gcapalbo: shoot, we probably have to give this some more thought then

corehq/form_processor/backends/sql/dbaccessors.py
```diff
@@ -292,6 +292,18 @@ def get_attachments(form_id):
         return list(XFormAttachmentSQL.objects.raw('SELECT * from get_form_attachments(%s)', [form_id]))
 
     @staticmethod
+    def iter_form_ids_by_submission_date(start_datetime, end_datetime):
+        with get_cursor(XFormInstanceSQL) as cursor:
+            cursor.execute(
+                'SELECT * from get_form_ids_by_type_and_date(%s, %s, %s)',
+                [start_datetime, end_datetime, XFormInstanceSQL.NORMAL],
+            )
+            # TODO make this a real iterator
```
snopoke: Something like this is what you want: https://github.com/dimagi/commcare-hq/blob/b7d91efe32231953df9d67097bf600ec3e8bc1f4/custom/hki/management/commands/hki_case_migration.py#L48-L53

corehq/sql_accessors/sql_templates/get_form_ids_by_type_and_date.sql
```diff
@@ -0,0 +1,13 @@
...
+-- return SETOF so that we get 0 rows when no form matches otherwise we'll get an empty row
+CREATE FUNCTION get_form_ids_by_type_and_date(start_date timestamp without time zone, end_date timestamp without time zone, p_state INTEGER DEFAULT 1) RETURNS TABLE (form_id VARCHAR(255)) AS $$
+BEGIN
+    RETURN QUERY
+    SELECT form_table.form_id FROM form_processor_xforminstancesql as form_table
+    WHERE form_table.state = p_state
+        AND form_table.received_on >= start_date
+        AND form_table.received_on < end_date;
+END;
+$$ LANGUAGE plpgsql;
```
snopoke: No need to have this SQL function since we can now run sharded queries in Django

corehq/warehouse/migrations/0004_app_status_fact_form_and_sync_staging_and_deleted_field.py
```diff
@@ -0,0 +1,103 @@
...
+            fields=[
+                ('id', models.AutoField(auto_created=True, primary_key=True, serialize=False, verbose_name='ID')),
+                ('last_form_submission_date', models.DateTimeField(null=True)),
+                ('last_sync_log_date', models.DateTimeField(null=True)),
+                ('last_form_app_build_version', models.CharField(max_length=255)),
+                ('last_form_app_commcare_version', models.CharField(max_length=255)),
+                ('last_form_app_source', models.CharField(max_length=255)),
+                ('last_sync_log_app_build_version', models.CharField(max_length=255)),
+                ('last_sync_log_app_commcare_version', models.CharField(max_length=255)),
+                ('last_sync_log_app_source', models.CharField(max_length=255)),
```
snopoke: how does this get linked to the user_dim?

corehq/warehouse/migrations/0004_app_status_fact_form_and_sync_staging_and_deleted_field.py
```diff
@@ -0,0 +1,103 @@
...
+            name='FormStagingTable',
+            fields=[
+                ('id', models.AutoField(auto_created=True, primary_key=True, serialize=False, verbose_name='ID')),
+                ('form_id', models.CharField(max_length=255, unique=True)),
+                ('domain', models.CharField(default=None, max_length=255)),
+                ('app_id', models.CharField(max_length=255, null=True)),
+                ('xmlns', models.CharField(default=None, max_length=255)),
+                ('user_id', models.CharField(max_length=255, null=True)),
+                ('received_on', models.DateTimeField(db_index=True)),
+                ('build_id', models.CharField(max_length=255, null=True)),
```
snopoke: `time_start`, `time_end` and probably others?

corehq/warehouse/models/dimensions.py
```diff
@@ -56,8 +86,14 @@ class LocationDim(BaseDim):
     location_last_modified = models.DateTimeField()
     location_created_on = models.DateTimeField()
```
snopoke: shouldn't this have a link to parent?

corehq/warehouse/models/staging.py
snopoke: I think this is a discussion point. For most reports we would only be considering 'normal' forms. There are some reports that look at other form types but they are not aggregated reports and so can probably be run off the raw data.

corehq/warehouse/transforms/__init__.py
snopoke: Can we not do the transform prior to putting the data in the staging table since we'll already need to be iterating over each app doc to extract the info?

snopoke (commented):
Did a high level review. Looks pretty good so far but agree with @gcapalbo's comments.

corehq/sql_accessors/sql_templates/get_form_ids_by_type_and_date.sql
benrudolph: nice!

corehq/warehouse/migrations/0004_app_status_fact_form_and_sync_staging_and_deleted_field.py
```diff
@@ -0,0 +1,103 @@
...
+        ),
+        migrations.AddField(
+            model_name='userlocationdim',
+            name='deleted',
+            field=models.BooleanField(default=False),
+        ),
+        migrations.AddField(
+            model_name='applicationstatusfact',
+            name='user_dim',
+            field=models.ForeignKey(on_delete=django.db.models.deletion.CASCADE, to='warehouse.UserDim'),
```
benrudolph: @snopoke the foreignkey is down here strangely. i'll autogenerate the migration from scratch again once it's good to go

corehq/warehouse/migrations/0004_app_status_fact_form_and_sync_staging_and_deleted_field.py
benrudolph: yea those seem like good props to have. not necessary for app status, but will be needed for the form fact table

corehq/warehouse/models/dimensions.py
benrudolph: still haven't figured out how to best model that relationship so i haven't done it yet

corehq/warehouse/transforms/__init__.py
benrudolph: that's a good question. i had kind of wanted to keep the E, T, and L parts very distinct. one concern would be that this could prolong the strain on the main database (instead of the warehouse database) since we take longer to process each record. not sure how valid that is, and it does seem like a good place to do the python transform. i'm hesitant we'll be able to get everything we want without some python

benrudolph (commented):
comments make sense, going to jump back on this when i'm off interrupt

snopoke:
I would consider the process of populating the staging table an entire ETL
cycle so if there are any transforms that make sense to put there I don't
see an issue with it. What we're doing is essentially a multi-stage ETL
where each stage can be it's own ETL.

gcapalbo:
i don't have strong feelings about this, but the reason Ben gave is typically the reason for making staging be an exact copy of source data:

> this could prolong the strain on the main database (instead of the warehouse database) since we take longer to process each record

if we make our etl process retryable from a failed step, keeping them separate would also allow us to avoid restaging the data in case of a failure during the transform step. but all of this is a trade off with adding a new intermediate table and extra steps to the process.

corehq/warehouse/models/shared.py
```diff
@@ -0,0 +1,86 @@
+import os
+
+from django.db import connections
+from django.conf import settings
+from django.template import Context, engines
+
+from corehq.sql_db.routers import db_for_read_write
+
+from corehq.warehouse.transforms import get_transform
+from corehq.warehouse.const import (
```
dimagimon: F401 'DJANGO_MAX_BATCH_SIZE' imported but unused

corehq/warehouse/migrations/0004_app_status_fact_form_and_sync_staging_and_deleted_field.py
benrudolph: this actually might be a bit unrealistic right now... it means we'd have to fetch the form from riak to get the time_start and time_end since we don't save it to the form

emord: Depending on how you're grabbing them you can also get them from ES in bulk. that's how i do it in icds reports UCR

benrudolph: hmm that's not a bad thought. the risk is running on query on ES for a certain time period and the index not being up to date

corehq/warehouse/tests/test_integration.py
```diff
@@ -0,0 +1,104 @@
...
+    def setUpClass(cls):
+        cls.domain_records = [
+            Domain(name=cls.domain, hr_name='One', creating_user_id='abc', is_active=True),
+        ]
+
+        for domain in cls.domain_records:
+            domain.save()
+
+        cls.user_records = [
+            # TODO: Handle WebUsers who have multiple domains
```
benrudolph: this is going to be a pain. since webusers have multiple domains, it means we most likely have to create these records differently when first fetching them from couch

corehq/warehouse/transforms/sql/app_status_fact.sql
benrudolph: I did not touch this query for this PR

corehq/warehouse/transforms/sql/domain_dim.sql
```diff
@@ -0,0 +1,55 @@
+INSERT INTO {{ domain_dim }} (
```
benrudolph: I realize this shouldn't be `INSERT`, but I'm talking with javier and lu about the possibility of upgrading PG in order to take advantage of `UPSERT`. I think that `UPSERT` will make the code simpler and easier to understand, so if the LOE isn't too high, i'd vote to leave these as insert until we upgrade

corehq/warehouse/transforms/sql/form_fact.sql
```diff
@@ -0,0 +1,44 @@
...
+    CASE
+        WHEN
+            (received_on > deleted_on OR deleted_on IS NULL) AND
+            (received_on > edited_on OR edited_on IS NULL)
+        THEN received_on::timestamp AT TIME ZONE default_timezone
+        WHEN
+            (deleted_on > received_on OR received_on IS NULL) AND
+            (deleted_on > edited_on OR edited_on IS NULL)
+        THEN deleted_on::timestamp AT TIME ZONE default_timezone
+        ELSE edited_on::timestamp AT TIME ZONE default_timezone
```
benrudolph: I believe this is what Nabil was suggesting for how we handle timezones. i haven't yet tested out querying the `last_modified` field

corehq/warehouse/transforms/sql/user_dim.sql
```diff
@@ -0,0 +1,55 @@
...
+    domain,
+    user_id,
+    username,
+    CASE
+        WHEN user_id = 'system' THEN 'system'
+        WHEN user_id = 'demo_user' THEN 'demo'
+        WHEN user_id = 'commtrack-system' THEN 'supply'
+        WHEN doc_type = 'WebUser' THEN 'web'
+        WHEN doc_type = 'CommCareUser' THEN 'mobile'
+        ELSE 'unknown'
```
benrudolph: possible i should allow for dims to insert extra context vars so we don't have to hard code these

benrudolph (commented):
I made some fairly significant changes from the last commit as I had some time to think about structuring the models. I essentially think warehouse models need to be customizable on two axis. one axis is whether or not it is staging, or dim/fact. this will allow for customizing behavior of committing the data (i.e. staging tables wipe the records). The other axis, is for how they transfer data. all three types of tables should be able to utilize different types of ETL strategies, for now there are just two Couch to SQL and SQL to SQL.

corehq/form_processor/backends/sql/dbaccessors.py
```diff
@@ -292,6 +293,28 @@ def get_attachments(form_id):
...
+        :returns: An iterator of XFormInstanceSQL objects
+        '''
+        from corehq.sql_db.util import run_query_across_partitioned_databases
+
+        return run_query_across_partitioned_databases(
+            XFormInstanceSQL,
+            (Q(received_on__gt=start_datetime) & Q(received_on__lte=end_datetime)) |
+            (Q(deleted_on__gt=start_datetime) & Q(deleted_on__lte=end_datetime)) |
+            (Q(edited_on__gt=start_datetime) & Q(edited_on__lte=end_datetime))
+        )
```
gcapalbo: it looks like deleted_on and edited_on don't have indexes in the source - we'll likely need to add them

gcapalbo: also, if we use OR statements like this i don't think it will be able to use indexes. we may need to split this into three queries

gcapalbo: also just confirming that all of these timestamps from the source are UTC?

corehq/warehouse/dbaccessors.py
```diff
@@ -40,3 +42,24 @@ def _get_ids_by_last_modified(cls, doc_types, start_datetime, end_datetime):
...
+
+    results = SyncLog.view(
+        "sync_logs_by_date/view",
+        startkey=[json_format_datetime(start_datetime)],
+        endkey=[json_format_datetime(end_datetime)],
+        reduce=False,
+        include_docs=False
+    )
+    for result in results:
+        yield result['id']
```
gcapalbo: i don't think couch has a way to make this query be exclusive of the start date the way you did for forms. is that going to be a problem on the off chance we pull the same record in two separate runs? or should we just make this check the key to make sure the date doesn't equal the start date before we yield

corehq/warehouse/etl.py
```diff
@@ -0,0 +1,103 @@
...
+        :returns: Dict of slug to table_name
+        {
+            <slug>: <table_name>,
+            ...
+        }
+        '''
+        from corehq.warehouse.models import get_cls_by_slug
+
+        context = {cls.slug: cls._meta.db_table}
+        for dep in cls.dependencies():
```
gcapalbo: should dependencies be a method that raises NotImplementedError on this class?

corehq/warehouse/etl.py
```diff
@@ -0,0 +1,103 @@
...
+        # ( <source field>, <staging field> )
+        raise NotImplementedError
+
+    @classmethod
+    def record_iter(cls, start_datetime, end_datetime):
+        raise NotImplementedError
+
+    @classmethod
+    def load(cls, start_datetime, end_datetime):
+        record_iter = cls.raw_record_iter(start_datetime, end_datetime)
```
gcapalbo: should this call `cls.record_iter`?

corehq/warehouse/management/commands/commit_table.py
```diff
@@ -53,10 +44,10 @@ def handle(self, slug, **options):
         start = options.get('start')
         end = options.get('end')
 
-        model = SLUG_TO_STAGING_TABLE.get(slug)
+        model = get_cls_by_slug(slug)
         if not model:
             raise CommandError('{} is not a valid slug. \n\n {}'.format(slug, USAGE))
-        model.stage_records(start, end)
+        model.commit(start, end)
```
gcapalbo: so does committing handle clearing the staging table, staging the data, and then moving to the final table?

corehq/warehouse/models/__init__.py
```diff
@@ -11,8 +11,35 @@
...
+        UserDim.slug: UserDim,
+        GroupDim.slug: GroupDim,
+        LocationDim.slug: LocationDim,
+        DomainDim.slug: DomainDim,
+        UserLocationDim.slug: UserLocationDim,
+        UserGroupDim.slug: UserGroupDim,
+
+        ApplicationStatusFact.slug: ApplicationStatusFact,
+        FormFact.slug: FormFact,
+    }.get(slug)
```
gcapalbo: should this use `[slug]` so that it fails hard if it's missing?

corehq/warehouse/models/dimensions.py
```diff
@@ -1,33 +1,83 @@
...
+    @classmethod
+    @transaction.atomic
+    def commit(cls, start_datetime, end_datetime):
+        cls.load(start_datetime, end_datetime)
 
     class Meta:
         abstract = True
 
+    @classmethod
+    @unit_testing_only
```
gcapalbo: good idea!

corehq/warehouse/models/dimensions.py
```diff
@@ -36,8 +86,19 @@ class GroupDim(BaseDim):
 
     group_last_modified = models.DateTimeField()
 
+    @classmethod
+    def dependencies(cls):
+        return [GROUP_STAGING_SLUG]
```
gcapalbo: probably a good idea. not sure where we'd do it - maybe if we had a test that breaks the build if it finds a circular dependency?

corehq/warehouse/models/dimensions.py
```diff
@@ -72,15 +144,33 @@ class DomainDim(BaseDim):
...
 
-class UserGroupDim(BaseDim):
+class UserGroupDim(BaseDim, CustomSQLETLMixin):
+    '''
+    Dimension for User and Group mapping
+
+    Grain: user_id, group_id
+    '''
+    slug = USER_GROUP_DIM_SLUG
+
```
gcapalbo: should UserLocationDim and UserGroupDim have dependencies specified?

corehq/warehouse/models/facts.py
```diff
@@ -0,0 +1,108 @@
...
+from corehq.sql_db.routers import db_for_read_write
+from corehq.util.test_utils import unit_testing_only
+from corehq.warehouse.etl import CustomSQLETLMixin
+from corehq.warehouse.models.shared import WarehouseTableMixin
+
+
+class BaseFact(models.Model, WarehouseTableMixin):
+
+    @classmethod
+    @transaction.atomic
```
gcapalbo: since this stuff happens in the non-default db we'll actually need to do `with transaction.atomic(using=database):`

corehq/warehouse/models/facts.py
```diff
@@ -0,0 +1,108 @@
...
+
+    class Meta:
+        abstract = True
+
+    @classmethod
+    @unit_testing_only
+    def clear_records(cls):
+        database = db_for_read_write(cls)
+        with connections[database].cursor() as cursor:
+            cursor.execute("TRUNCATE {}".format(cls._meta.db_table))
```
gcapalbo: maybe we can pull this out into its own function? seems like it's used in a couple places

corehq/warehouse/models/facts.py
```diff
@@ -0,0 +1,108 @@
...
+
+
+class FormFact(BaseFact, CustomSQLETLMixin):
+    '''
+    Represents the staging table to dump data before loading into the FormFact
```
gcapalbo: is this the staging table or the actual fact?

corehq/warehouse/models/facts.py
```diff
@@ -0,0 +1,108 @@
...
+
+    Grain: form_id
+    '''
+    slug = FORM_FACT_SLUG
+
+    form_id = models.CharField(max_length=255, unique=True)
+
+    domain = models.CharField(max_length=255, default=None)
+    app_id = models.CharField(max_length=255, null=True)
```
gcapalbo: curious if you think we need an app dim

corehq/warehouse/models/staging.py
```diff
@@ -8,46 +8,52 @@
...
-        raise NotImplementedError
-
-    @classmethod
-    def field_mapping(cls):
-        # Map source model fields to staging table fields
-        # ( <source field>, <staging field> )
-        raise NotImplementedError
-
-    @classmethod
     @transaction.atomic
```
snopoke: Is it necessary to do this as an atomic operation? Considering it's size it would be quite expensive.

corehq/warehouse/models/facts.py
```diff
@@ -0,0 +1,108 @@
...
+    xmlns = models.CharField(max_length=255, default=None)
+    user_id = models.CharField(max_length=255, null=True)
```
gcapalbo: do we need user_id if we have user_dim as a foreign key?

corehq/warehouse/models/staging.py
```diff
@@ -70,14 +76,23 @@ def field_mapping(cls):
         ]
```
snopoke: not part of this PR but curious why you renamed `last_modified` to `group_last_modified`. Same for others below.

corehq/warehouse/models/facts.py
```diff
@@ -0,0 +1,108 @@
...
+    # The time at which the server has received the form
+    received_on = models.DateTimeField(db_index=True)
+    deleted_on = models.DateTimeField(null=True)
+    edited_on = models.DateTimeField(null=True)
+    last_modified = models.DateTimeField(null=True)
+
+    build_id = models.CharField(max_length=255, null=True)
+    state = models.PositiveSmallIntegerField(
+        choices=XFormInstanceSQL.STATES,
+        default=XFormInstanceSQL.NORMAL
```
gcapalbo: it might be better to not set a default since we'd be assuming something about the source

corehq/warehouse/models/facts.py
```diff
@@ -0,0 +1,108 @@
...
+    '''
+    Application Status Report Fact Table
+
+    Grain: app_id, user_id
+    '''
+    slug = APP_STATUS_FACT_SLUG
+
+    # TODO: Add app dimension
+    # app_dim = models.CharField(max_length=255)
+
```
gcapalbo: do we need domain or domain_dim on here?

corehq/warehouse/models/shared.py
```diff
@@ -0,0 +1,14 @@
+from django.db import transaction
+
+
+class WarehouseTableMixin(object):
+
+    @classmethod
+    @transaction.atomic
```
gcapalbo: same comment about being careful with transaction.atomic

corehq/warehouse/transforms/sql/domain_dim.sql
```diff
@@ -0,0 +1,55 @@
...
+    CASE is_test
+        WHEN 'none' THEN false
+        WHEN 'true' THEN true
+        WHEN 'false' THEN false
+    END,
+    location_restriction_for_users,
+    -- Ensures all the boolean values are either true or false
+    CASE use_sql_backend
+        WHEN use_sql_backend THEN true
+        ELSE false
```
snopoke: isn't this already a boolean? also `first_domain_for_user`

corehq/warehouse/transforms/sql/form_fact.sql
snopoke: I think you could use `GREATEST` here instead of the case statement. Also aren't these server times so wouldn't need to store timezone?

corehq/warehouse/models/dimensions.py
```diff
@@ -1,33 +1,83 @@
...
+    DOMAIN_STAGING_SLUG,
+)
+
+from corehq.sql_db.routers import db_for_read_write
+from corehq.util.test_utils import unit_testing_only
+from corehq.warehouse.etl import CustomSQLETLMixin
+from corehq.warehouse.models.shared import WarehouseTableMixin
+
+
+class BaseDim(models.Model, WarehouseTableMixin):
```
gcapalbo: we can handle in another pr but we should probably track batch_id on all our tables

corehq/warehouse/tests/test_integration.py
gcapalbo: hmm, yeah we need to think about this a bit

corehq/warehouse/transforms/sql/app_status_fact.sql
gcapalbo: ok. before this goes live i think we're really going to want to split this into intermediate tables - we shouldn't ever join results from subqueries

corehq/warehouse/transforms/sql/form_fact.sql
gcapalbo: so this stores the timestamp in the domain's local timezone? also is this last_modified field the only one we'll be querying on later on when doing daily aggregations?

corehq/warehouse/transforms/sql/form_fact.sql
```diff
@@ -0,0 +1,44 @@
...
+    END
+FROM
+    {{ form_staging }} AS form_table
+JOIN {{ domain_dim }} AS domain_table ON form_table.domain = domain_table.domain
+-- TODO: Should we allow records to be inserted if not found in the UserDim? Unknown Users?
```
gcapalbo: probably ok to keep them - there's no harm right? they'd just drop out of queries when we join form fact to user dim

corehq/warehouse/transforms/sql/user_dim.sql
gcapalbo: should we just allow WebUser, CommCareUser, and unknown, or are there plans to use the other ones

corehq/warehouse/transforms/sql/user_dim.sql
```diff
@@ -0,0 +1,55 @@
...
+    END,
+    first_name,
+    last_name,
+    email,
+    doc_type,
+    CASE is_active
```
gcapalbo: is this valid syntax?

corehq/warehouse/transforms/sql/user_dim.sql
```diff
@@ -0,0 +1,55 @@
...
+        WHEN is_active OR is_active IS NULL THEN true
+        ELSE false
+    END,
+    CASE is_staff
```
gcapalbo: aren't these booleans on the staging table?

corehq/warehouse/utils.py
```diff
@@ -6,7 +6,10 @@ def django_batch_records(cls, record_iter, field_mapping):
     for index, raw_record in enumerate(record_iter):
         record = {}
         for source_key, destination_key in field_mapping:
-            record[destination_key] = raw_record.get(source_key)
+            if isinstance(raw_record, dict):
+                record[destination_key] = raw_record.get(source_key)
```
gcapalbo: `getattr` will fail hard below but this one won't. maybe we should make them consistently either fail hard or default to None?

corehq/warehouse/transforms/sql/form_fact.sql
benrudolph: ah yes `GREATEST ` exactly what i was looking for.

yes this stores the form in the domain's local timezone. as for if this will be the only one we'll be querying on, that's a good question. i wonder if we need to even worry about `deleted_on` and `edited_on`. Mostly likely we'll want to aggregate always by `received_on`. even for reports like "submissions by day" i don't think we actually care if the form was archived or edited

corehq/warehouse/transforms/sql/user_dim.sql
benrudolph: yea these are used in the report filters. it's what's used when you use the filter "CommCare Supply", etc

benrudolph: i suppose i was just going for parity, but we may not want to do that anymore

corehq/form_processor/backends/sql/dbaccessors.py
benrudolph: what do you mean by yielding the same form multiple times? are you talking about in a given time range? not sure how that would happen

yes `greatest` seems better. thanks for that tip. i would certainly like to refactor to have one field

corehq/warehouse/dbaccessors.py
benrudolph: interesting. right that was definitely something i didn't think about for this one. is it actually that bad if we reprocess a  sync log? shouldn't it just correctly update instead of insert?

corehq/warehouse/etl.py
benrudolph: yeah they're not defined on this model, but rather on the warehousemixin. i want all warehouse tables to have dependencies regardless of whether they use custom sql to do their transformation. perhaps the correct thing to do here is to pass in those two variables

corehq/warehouse/etl.py
benrudolph: yes oops, i need to update all these function names

corehq/warehouse/management/commands/commit_table.py
benrudolph: yes that's the idea. for staging tables, `commit` is a `clear` then `load` command. for facts and dims, `commit` is just a `load` call

corehq/warehouse/models/dimensions.py
benrudolph: 👍 agreed, will handle on a later PR. i think it'll fit in well

corehq/warehouse/models/staging.py
benrudolph: just to specify that the `last_modified ` referred to the model. while in staging tables it doesn't matter too much, but in dim and fact tables there's also a `last_modified` field for when that row was last modified vs. when the model was. thought it might be useful to know. could be a bit over engineered

corehq/warehouse/transforms/sql/app_status_fact.sql
benrudolph: yep agreed. i wanted to leave this here because it's useful for me to think about the query / tables we'll need in order to get the desired app status report

corehq/warehouse/transforms/sql/domain_dim.sql
benrudolph: could also be null. probably not necessary and i can make this field accept nulls

corehq/warehouse/transforms/sql/form_fact.sql
benrudolph: true

corehq/warehouse/transforms/sql/user_dim.sql
benrudolph: yes, but realized i must've changed this in a second pass and it's not necessary (might even cause unexpected value) https://www.postgresql.org/docs/9.4/static/functions-conditional.html

corehq/warehouse/utils.py
benrudolph: 👍 given the data we have in couch, i'll make it default to `None`

benrudolph (commented):
thanks for the feedback will take another pass to address the comments

corehq/warehouse/models/facts.py
benrudolph: indeed we do, will add it on next iteration. for now making a TODO comment

corehq/warehouse/etl.py
benrudolph: https://github.com/dimagi/commcare-hq/pull/16327/commits/349b67c7fb467ddce2eb9d269e3312f1aefd0987 for now i've assert that it's of the proper class. i can offline if you'd like to know the abstraction issue i'm dealing with and why this is somewhat tricky to solve cleanly. open to more ideas

corehq/warehouse/etl.py
benrudolph: https://github.com/dimagi/commcare-hq/pull/16327/commits/548894959c361bf3b3d343b02b23db0c8f189b8d

corehq/warehouse/models/__init__.py
benrudolph: https://github.com/dimagi/commcare-hq/pull/16327/commits/8a5a2ac695e37f228d8ffd60007dfa32ed0d7e88

corehq/warehouse/models/dimensions.py
benrudolph: https://github.com/dimagi/commcare-hq/pull/16327/commits/d58a57dbc6c1e0c05fd32a83b9acb009877f1ffb

corehq/warehouse/models/facts.py
benrudolph: https://github.com/dimagi/commcare-hq/pull/16327/commits/f63163687d2a1262f51f1b7ec9fcb2aaf7f1320f

corehq/warehouse/models/facts.py
benrudolph: https://github.com/dimagi/commcare-hq/pull/16327/commits/2a822603188d05046dbfb3a76032271ca1093a29

corehq/warehouse/models/facts.py
benrudolph: https://github.com/dimagi/commcare-hq/pull/16327/commits/5a314aa69ff53531cd70b6554b946c24c7904940

corehq/warehouse/models/facts.py
benrudolph: https://github.com/dimagi/commcare-hq/pull/16327/commits/adfb909cbfbc6b5eb8364d280e4a598bf0622c31

corehq/warehouse/models/facts.py
benrudolph: https://github.com/dimagi/commcare-hq/pull/16327/commits/8b074b90bfcdaa3a5c214631e9192c879d24a636

corehq/warehouse/models/shared.py
```diff
@@ -0,0 +1,14 @@
+from django.db import transaction
+
+
+class WarehouseTableMixin(object):
```
benrudolph: can't actually do this because we set the meta class in the subclass

corehq/warehouse/transforms/sql/form_fact.sql
benrudolph: https://github.com/dimagi/commcare-hq/pull/16327/commits/7518a7921c39c66e1c21c8d0f66117226071d3cf

corehq/warehouse/transforms/sql/user_dim.sql
benrudolph: https://github.com/dimagi/commcare-hq/pull/16327/commits/4420f4c9ad5c265464ea49acc461ab171bc1173a

corehq/warehouse/utils.py
benrudolph: https://github.com/dimagi/commcare-hq/pull/16327/commits/323d99166ea7ca38314bc291b8b3029e4f7bcb2b

benrudolph:
given these tests pass, do you think we can merge this? i've responded to most of the feedback. i still have a couple open questions, but for the sake of review-ability it may be best to merge and then follow up on the rest of the open questions?

benrudolph:
✅ at last

corehq/warehouse/transforms/sql/user_dim.sql
gcapalbo: i see

corehq/warehouse/dbaccessors.py
gcapalbo: yeah it depends on the etl process. if it's a dim it wouldn't matter,if it's a fact it might cause us to increment a count one too many or insert one too many records. might be good to just make it a standard that we always make start date exclusive and end date inclusive

corehq/warehouse/management/commands/commit_table.py
gcapalbo: cool

corehq/warehouse/dbaccessors.py
benrudolph: right that makes sense. i was just thinking in terms of the synclog fact, but it could lead to some hard to figure out bugs. i like the idea of making it a standard. updated here: https://github.com/dimagi/commcare-hq/pull/16327/commits/f3758145405586f6855a318a031a343611b6f6b9

gcapalbo:
yeah i think it's fine to merge and address the rest of the outstanding stuff in separate prs

corehq/form_processor/backends/sql/dbaccessors.py
snopoke: I mean that the same for could fall into multiple date ranges

e.g. 
form.received_on = 2017-01-10
form.edited_on = 2017-02-10
form.deleted_on = 2017-04-01

range 1 = 2017-01-01 to 2017-01-31
range 2 = 2017-02-01 to 2017-02-28
range 3 = 2017-04-01 to 2017-04-30

corehq/warehouse/dbaccessors.py
```diff
@@ -39,4 +53,43 @@ def _get_ids_by_last_modified(cls, doc_types, start_datetime, end_datetime):
             reduce=False,
         )
         for result in results:
+            result_modified_datetime = result['key'][1]
+            # Skip the record if the datetime is equal to the start because this should return
+            # records with an exclusive start date.
+            if result_modified_datetime == json_format_datetime(start_datetime):
```
snopoke: better to call `json_format_datetime` outside the loop

corehq/warehouse/etl.py
snopoke: Sounds interesting.

corehq/warehouse/models/staging.py
```diff
@@ -115,16 +134,22 @@ def field_mapping(cls):
...
+class DomainStagingTable(StagingTable, CouchToDjangoETLMixin):
+    '''
+    Represents the staging table to dump data before loading into the DomainDim
+
+    Grain: domain_id
+    '''
+    slug = DOMAIN_STAGING_SLUG
 
     domain_id = models.CharField(max_length=255)
+    domain = models.CharField(max_length=100)
```
snopoke: should this just be name?

corehq/warehouse/transforms/sql/form_fact.sql
snopoke: I agree with that - I don't think we'll every query on the other dates (though we will need completion time at some point).

corehq/warehouse/models/staging.py
benrudolph: it could be, but since every dimension has `domain` on it, it'll be converted to `domain` anyways: https://github.com/dimagi/commcare-hq/blob/br/prep-appstatus/corehq/warehouse/models/dimensions.py#L23. thought it'd be simpler to have the domain be named the same thing on every model, even if it's the domain dim

corehq/form_processor/backends/sql/dbaccessors.py
benrudolph: i see i see, you're right this doesn't work great as a last modified if you're querying in the past

corehq/warehouse/dbaccessors.py
benrudolph: https://github.com/dimagi/commcare-hq/pull/16327/commits/632241687a063ba44ed5ccc5b32950b1e4a3dff8

benrudolph:
✅ at last. @snopoke requested an architecture review of the models i'm using so i'm going to work on writing something up. in the meantime i think we can merge. i'm happy to do any refactoring after the review, just don't love having giant long running PRs

benrudolph:
🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 

slow and steady baby

PR #16331, opened 2017-05-25
Proper ajax fn

benrudolph:
i think this is confounding the issue described here: https://manage.dimagi.com/default.asp?254777.


benrudolph:
actually it's not confounding much, but still good to fix

dannyroberts (approved):

PR #16332, opened 2017-05-25
update doctypemigrations

nickpell:
I'm guessing this was broken when we upgraded to django 1.10.

@millerdev


millerdev (approved):
Huh, how did we miss that?

PR #16333, opened 2017-05-25
Tweak web apps anonymous login page

orangejenny:
@benrudolph what do you think of this? My main thought was to make the login look more like web apps in general. I'm also guessing that people who hit this page are just going to be confused by the option to log in with a real account and also that they won't know what "CommCare" means, so I removed those elements.

before
screen shot 2017-05-25 at 1 35 22 pm

after
screen shot 2017-05-25 at 1 35 28 pm

cc @esoergel


benrudolph:
sweet nice and simple 💰

PR #16334, opened 2017-05-25
fix deploy view headers with app changes

orangejenny:
very minor, but now that I've noticed it I can't un-see it

screen shot 2017-05-25 at 2 00 09 pm

@esoergel


esoergel (approved):
ooh, fair point

PR #16341, opened 2017-05-25
App manager v2: reconcile inline edit widgets

orangejenny:
Realized that inline_edit.js and inline_edit_v2.js have gotten out of sync, decided to merge them rather than risk them getting out sync again (we don't touch this widget often, but when we do it's issues like html injection). The code is ugly, but we're in the home stretch of v2 so it shouldn't last long.

Easiest to review by commit.

@esoergel / @biyeun


orangejenny:
pinging old & new code buddies: @esoergel @emord

corehq/apps/style/templates/style/includes/ko.html
```diff
@@ -14,5 +14,163 @@
 
 <script src="{% static 'style/js/components/inline_edit.js' %}"></script>
 <script src="{% static 'style/js/components/inline_edit_v2.js' %}"></script>
-<script src="{% static 'style/js/components.ko.js' %}"></script>
 {% endcompress %}
+
```
esoergel: Just curious - what's the advantage of putting this in an HTML file rather than a separate JS file?

corehq/apps/style/templates/style/includes/ko.html
```diff
@@ -123,9 +123,8 @@
                     self.hasError(false);
                 };
             },
-            {% if request|toggle_enabled:"APP_MANAGER_V2" %}
```
esoergel: Nice, I was just gonna ask about this.

esoergel (approved):
Cool, looks good!

corehq/apps/style/templates/style/includes/ko.html
orangejenny: I think you answered your own question, but let me know if not :)  (I need access to toggles, and the js version of toggles is deliberately loaded much later than this file.)

PR #16343, opened 2017-05-26
Use different property in location filter: referring_facility_id -> d…

kkrampa:
…iagnosing_facility_id


kkrampa:
@NoahCarnahan 
@proteusvacuum If Noah is offline, could you make sure it will be merged and deployed? It needs episode data source rebuild.

PR #16344, opened 2017-05-26
Remove name in excel web query

benrudolph:
@NoahCarnahan @emord this thing has been driving me nuts. in the end i decided to remove the name after trying various number of things with this html page: https://benrudolph.github.io/excel-web-query/. i still have to ensure it won't break old excel queries

http://manage.dimagi.com/default.asp?253389
http://manage.dimagi.com/default.asp?254400

### DIFF
--- a/corehq/apps/export/tests/test_get_export_file.py
+++ b/corehq/apps/export/tests/test_get_export_file.py
@@ -487,10 +487,10 @@ class WriterTest(SimpleTestCase):
         with writer.open([export_instance]):
             _write_export_instance(writer, export_instance, docs)
         with ExportFile(writer.path, writer.format) as export:
-            exported_tables = [table for table in re.findall('<h2>(.*)</h2>', export.read())]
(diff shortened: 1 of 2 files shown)


benrudolph:
ok tested going from having a header to not having a header and excel updates fine. other solution if we're worried about older version of excel not handling this correctly is to pass a param in and conditionally exclude the excel header and then only pass that param in on new dashboard links. that isn't my preferred solution as it will take a lot of parameter passing and i'm not super keen to introduce that tech debt

emord:
test failure looks real

NoahCarnahan (approved):
I'm guessing excel still imports the entire page, regardless of what is selected, right? This would still be a problem who have multiple tables in an export, although it's probably a less common workflow, so I think this fix makes sense while excel is broken.

benrudolph:
yes that is true and i thought about if there was a way to handle that. but at least in that case we can tell people the work around is to only select the table you want to import into excel :/ not great

PR #16345, opened 2017-05-26
Allow Named Filters to Reference Each Other

sheelio:
This is following the same model as for named expressions.


sheelio:
@NoahCarnahan 

Can you take a quick look at this?  Want it for some ICDS reports.

PR #16346, opened 2017-05-26
Raise error when xform will create a new case

benrudolph:
@wpride this'll hopefully save us from ourselves when people try to edit a form that will create a new case instead of updating the one it already created
https://manage.dimagi.com/default.asp?251677
cc: @proteusvacuum interrupt @emord


wpride (approved):
awesome

PR #16347, opened 2017-05-26
smaller chunk size for apps

emord:
interrupt-ish @benrudolph 

Trying to load 1000 apps at once doesn't do great things for memory

buddy @calellowitz


benrudolph (approved):
ouf ran into this for exports

PR #16348, opened 2017-05-26
Fix assert for tzmigration

benrudolph:
@dannyroberts


benrudolph:
other args get passed into `options`. must've changed since the upgrade of django

dannyroberts (approved):

PR #16349, opened 2017-05-26
Kill log_counter

benrudolph:
@proteusvacuum @emord http://manage.dimagi.com/default.asp?252494

tested the js error one, but admittedly didn't run the other ones due to being complicated to repro and it being a fairly easy change. i also rationalized that datadog swallows any errors if they occur. will update ddog dashboards once this is in

🏠


emord (approved):

snopoke:
:+1:

PR #13731, opened 2016-10-26
Convert knockout typeahead binding from jQuery UI to atwho

orangejenny:
Same rationale as https://github.com/dimagi/langcodes/pull/8

@calellowitz  / @emord

PR #16351, opened 2017-05-26
Bring atwho into main HQ and use for case config

orangejenny:
I did this a while ago and ended up rolling it back because of compatibility issues with jQuery's caret plugin. We don't use the caret plugin anymore, and there are some slight usability differences between select2 and atwho (http://manage.dimagi.com/default.asp?248233 is the catalyst for this PR), so let's have both widgets available in HQ.

Mostly re-applying changes from https://github.com/dimagi/commcare-hq/pull/13731

@esoergel / @emord

### DIFF
--- a/corehq/apps/style/static/style/js/atwho.js
+++ b/corehq/apps/style/static/style/js/atwho.js
@@ -1,0 +1,60 @@
+hqDefine('style/js/atwho', function () {
+    var _init = function($input, options, afterInsert) {
+        $input.atwho(options).on("inserted.atwho", function(event, $li, otherEvent) {
+            $input.val($input.data("selected-value")).change();
+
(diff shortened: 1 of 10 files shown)


corehq/apps/style/static/style/js/atwho.js
```diff
@@ -0,0 +1,60 @@
+hqDefine('style/js/atwho', function () {
+    var _init = function($input, options, afterInsert) {
```
emord: How hard would this be to get vellum to use this?

emord (approved):

corehq/apps/style/static/style/js/atwho.js
orangejenny: will offline

PR #16267, opened 2017-05-22
kill case label and referral label

orangejenny:
https://manage.dimagi.com/default.asp?254069

Hold off on merge until I've finished confirming with mobile team that this is okay to kill.

code buddy @esoergel 
app manager @snopoke / @dannyroberts

PR #16352, opened 2017-05-26
Empty text node, not blank xpath function

orangejenny:
http://manage.dimagi.com/default.asp?255007

Caused by https://github.com/dimagi/commcare-hq/pull/16267 where I should have used `<xpath function="''"/>` not `<xpath function=""/>` which breaks. But using an empty text node seems to work fine.


orangejenny:
@emord this is for today's deploy

orangejenny:
code buddy @esoergel

esoergel (approved):

kaapstorm (approved):

PR #16353, opened 2017-05-26
Add SMS backend for vertex

mkangia:
Adds new SMS backend for eNikshay.

best review as 🏛 

https://trello.com/c/v1kmyTv3/143-sms-backend-wiring?menu=filter


corehq/messaging/smsbackends/vertex/models.py
```diff
@@ -0,0 +1,32 @@
...
+
+
+class SQLVertexBackend(SQLSMSBackend):
+    class Meta:
+        app_label = 'sms'
+        proxy = True
+
+    @classmethod
+    def get_api_id(cls):
+        return 'Vertex'
```
gcapalbo: just for convention, we usually make this all uppercase

corehq/messaging/smsbackends/vertex/forms.py
```diff
@@ -0,0 +1,36 @@
+from corehq.apps.sms.forms import BackendForm
+from django import forms
+from crispy_forms import layout as crispy
+from django.utils.translation import ugettext_lazy, ugettext as _
+
+
+class VertexBackendForm(BackendForm):
+    _cchq_domain = "enikshay"
```
gcapalbo: this isn't necessary - we'll just create an instance of this gateway in the domain we need so it's only available there

corehq/messaging/smsbackends/vertex/models.py
```diff
@@ -0,0 +1,32 @@
+from corehq.messaging.smsbackends.http.models import SQLSMSBackend
+from corehq.messaging.smsbackends.vertex.forms import VertexBackendForm
+
+
+class SQLVertexBackend(SQLSMSBackend):
```
gcapalbo: we can remove the SQL from the beginning of SQLVertexBackend - it's on other classes because I haven't had a chance to rename them from when I moved everything over from couch

corehq/messaging/smsbackends/vertex/models.py
```diff
@@ -0,0 +1,74 @@
...
+    @classmethod
+    def get_api_id(cls):
+        return 'VERTEX'
+
+    @classmethod
+    def get_generic_name(cls):
+        return "Vertex"
+
+    @classmethod
+    def get_available_extra_fields(self):
```
gcapalbo: `cls` instead of `self`

corehq/messaging/smsbackends/vertex/models.py
```diff
@@ -0,0 +1,74 @@
...
+        return [
+            'username',
+            'password',
+            'senderid',
+            'response',
+        ]
+
+    @classmethod
+    def get_form_class(self):
```
gcapalbo: `cls`

corehq/messaging/smsbackends/vertex/forms.py
```diff
@@ -0,0 +1,34 @@
+from corehq.apps.sms.forms import BackendForm
+from django import forms
+from crispy_forms import layout as crispy
+from django.utils.translation import ugettext_lazy, ugettext as _
+
+
+class VertexBackendForm(BackendForm):
+    username = forms.CharField(
```
gcapalbo: i usually like to makes these `dimagi.utils.django.fields.TrimmedCharField`s

corehq/messaging/smsbackends/vertex/forms.py
```diff
@@ -0,0 +1,34 @@
...
+    )
+    password = forms.CharField(
+        label=ugettext_lazy("password"),
+        required=True,
+    )
+    senderid = forms.CharField(
+        label=ugettext_lazy("senderid"),
+        required=True,
+    )
+    response = forms.ChoiceField(
```
gcapalbo: this shouldn't be a configurable option - since we're depending on the return value to be present from the api response we should just always send this value as Y

corehq/messaging/smsbackends/vertex/models.py
```diff
@@ -0,0 +1,74 @@
...
+            'response': config.response,
+            # It must include the country code appended before the mobile number
+            # The mobile number should contain only numbers and no symbols like "+", "-" etc.
+            'dest_mobileno': msg.phone_number,
+            'msgtype': 'TXT'  # TXT/UNI/FLASH/WAP
+        }
+        message = self.sanitize_message(msg)
+        params['message'] = message
+        url = '%s?%s' % (VERTEX_URL, urllib.urlencode(params))
+        resp = urllib2.urlopen(url, timeout=settings.SMS_GATEWAY_TIMEOUT).read()
```
gcapalbo: i've been trying to use the requests library for all new backends. an example is here: https://github.com/dimagi/commcare-hq/blob/master/corehq/messaging/smsbackends/smsgh/models.py

corehq/messaging/smsbackends/vertex/models.py
```diff
@@ -0,0 +1,74 @@
...
+
+    def sanitize_message(self, message):
+        # sanitize message using the rules as suggested by the api
+        # to avoid collision with reserved HTTP characters
+        return (message
+                .replace('&', 'amp;')
+                .replace('#', ';hash')
+                .replace('+', 'plus;')
+                .replace(',', 'comma;')
+                )
```
gcapalbo: i realize this is what the docs say to do, but this seems strange to me. can we test out just using the python requests library and not doing these conversions? requests will automatically url encode all parameters

corehq/messaging/smsbackends/vertex/models.py
```diff
@@ -0,0 +1,74 @@
...
+            'username': config.username,
+            'pass': config.password,
+            'senderid': config.senderid,
+            'response': config.response,
+            # It must include the country code appended before the mobile number
+            # The mobile number should contain only numbers and no symbols like "+", "-" etc.
+            'dest_mobileno': msg.phone_number,
+            'msgtype': 'TXT'  # TXT/UNI/FLASH/WAP
+        }
+        message = self.sanitize_message(msg)
```
gcapalbo: msg is actually an instance of corehq.apps.sms.models.QueuedSMS; msg.text includes the text

corehq/messaging/smsbackends/vertex/models.py
```diff
@@ -0,0 +1,74 @@
...
+    def send(self, msg, *args, **kwargs):
+        config = self.config
+        params = {
+            'username': config.username,
+            'pass': config.password,
+            'senderid': config.senderid,
+            'response': config.response,
+            # It must include the country code appended before the mobile number
+            # The mobile number should contain only numbers and no symbols like "+", "-" etc.
+            'dest_mobileno': msg.phone_number,
```
gcapalbo: all entries in QueuedSMS / SMS include the + on the phone number, so we should call `corehq.apps.sms.util.strip_plus` on the phone number before adding it here

corehq/messaging/smsbackends/vertex/models.py
```diff
@@ -0,0 +1,74 @@
...
+                .replace(',', 'comma;')
+                )
+
+    def send(self, msg, *args, **kwargs):
+        config = self.config
+        params = {
+            'username': config.username,
+            'pass': config.password,
+            'senderid': config.senderid,
+            'response': config.response,
```
gcapalbo: (we should just always set this to Y)

corehq/messaging/smsbackends/vertex/models.py
```diff
@@ -0,0 +1,74 @@
+import urllib2, urllib
+import re
+from django.conf import settings
+from corehq.messaging.smsbackends.http.models import SQLSMSBackend
+from corehq.messaging.smsbackends.vertex.forms import VertexBackendForm
+
+VERTEX_URL = "http://www.smsjust.com/sms/user/urlsms.php"
```
gcapalbo: we should ask if there's an https endpoint we can connect to

corehq/messaging/smsbackends/vertex/models.py
```diff
@@ -0,0 +1,74 @@
...
+SUCCESS_RESPONSE_REGEX = r'^([0-9]+)-(20[0-9]{2})_([0-9]{2})_([0-9]{2})$'  # 570737298-2017_05_27
```
gcapalbo: this might read a little easier if we used \d instead of [0-9]:

`^\d+-\d{4}_\d{2}_\d{2}$`

corehq/messaging/smsbackends/vertex/models.py
```diff
@@ -0,0 +1,74 @@
...
+        return resp
+
+    def handle_response(self, msg, resp):
+        # in case of success store the complete response msg which should be like
+        # 570737298-2017_05_27 AS-IS. This complete id referred to as "Scheduleid"
+        # can be then used to enquire for status of this particular SMS
+        if SUCCESS_RESPONSE_REGEX_MATCHER.match(resp):
+            msg.backend_message_id = resp
+        else:
+            msg.set_system_error(resp)
```
gcapalbo: we should try to use our defined constants when calling set_system_error. i looked through the error messages and what we should do here is:

when the response is `ES1009 Sorry unable to process request`, we should `msg.set_system_error(SMS.ERROR_INVALID_DESTINATION_NUMBER)`

when the response is `Message is blank`, we should `msg.set_system_error(SMS.ERROR_TOO_MANY_UNSUCCESSFUL_ATTEMPTS)`

for all other error messages, we should raise a `VertexBackendException` with the http response code and the returned error message.  this will let us be notified of it, and also will let the framework retry it.

gcapalbo:
this is looking good @mkangia. i added some comments, also we should do the following:

- we should add a test for this in `corehq.apps.sms.tests.test_backends.AllBackendTest.test_outbound_sms`. can see how the other backends are tested for outbound in that method

- should run an end-to-end local test of this by creating an instance of the gateway locally and sending an sms through the compose interface

corehq/messaging/smsbackends/vertex/models.py
```diff
@@ -0,0 +1,74 @@
...
+        config = self.config
+        params = {
+            'username': config.username,
+            'pass': config.password,
+            'senderid': config.senderid,
+            'response': config.response,
+            # It must include the country code appended before the mobile number
+            # The mobile number should contain only numbers and no symbols like "+", "-" etc.
+            'dest_mobileno': msg.phone_number,
+            'msgtype': 'TXT'  # TXT/UNI/FLASH/WAP
```
gcapalbo: we will need to support unicode eventually as well, but the docs aren't clear on how to do that. i'm adding a checklist item for this in the trello.

for now, we should be explicit about this by testing the message for ascii and only sending if it is ascii. so we should do something like:

    try:
        params['message'] = str(msg.text)
        params['msgtype'] = 'TXT'
    except UnicodeEncodeError:
        raise VertexBackendException('Only ascii messages are currently supported')

corehq/messaging/smsbackends/vertex/forms.py
mkangia: Makes sense.. Will remove it.

corehq/messaging/smsbackends/vertex/models.py
mkangia: There is. Testing it.

corehq/messaging/smsbackends/vertex/models.py
mkangia: Tested that a message is sent properly even without that sanitisation. Removing it. Thank you!

corehq/messaging/smsbackends/vertex/models.py
mkangia: Cool. Thanks

corehq/messaging/smsbackends/vertex/models.py
mkangia: I checked this.
For unicode we need to send msgtype as "UNI".
`msg.encode('utf-8')` worked. So i think its safe to default to "UNI" for all messages.

corehq/messaging/smsbackends/vertex/models.py
mkangia: i am not sure about this change. None of the error responses seem like a situation where the framework should re try it so not sure why these should be treated differently.
For notifying it makes sense to me where service is permanently barred like authentication or expired/exhausted account.

corehq/messaging/smsbackends/vertex/models.py
mkangia: their certificate is not great. It still needs a manual confirmation when accessed from browser and fails with SSLError from code.
```
SSLError: ("bad handshake: Error([('SSL routines', 'tls_process_server_certificate', 'certificate verify failed')],)",)
```

mkangia:
@gcapalbo 

> we should add a test for this in corehq.apps.sms.tests.test_backends.AllBackendTest.test_outbound_sms. can see how the other backends are tested for outbound in that method

Done

> should run an end-to-end local test of this by creating an instance of the gateway locally and sending an sms through the compose interface

Done this to ensure https and utf-8 encoded messages in local language(hindi)

sravfeyn:
Looks good to me, will let @gcapalbo review

corehq/messaging/smsbackends/vertex/models.py
```diff
@@ -0,0 +1,68 @@
...
+        config = self.config
+        params = {
+            'username': config.username,
+            'pass': config.password,
+            'senderid': config.senderid,
+            'response': 'Y',
+            # It must include the country code appended before the mobile number
+            # The mobile number should contain only numbers and no symbols like "+", "-" etc.
+            'dest_mobileno': strip_plus(msg_obj.phone_number),
+            'msgtype': 'UNI'  # TXT/UNI/FLASH/WAP
```
gcapalbo: we shouldn't default to unicode - when you send a message as unicode, the message is charged in 70-character chunks (since each character takes up 16 bits). we should instead do something like:

    try:
        params['message'] = str(msg_obj.text)
        params['msgtype'] = 'TXT'
    except UnicodeEncodeError:
        params['message'] = msg_obj.text.encode('utf-8')
        params['msgtype'] = 'UNI'

corehq/messaging/smsbackends/vertex/models.py
```diff
@@ -0,0 +1,68 @@
...
+        }
+        params['message'] = msg_obj.text.encode('utf-8')
+        return params
+
+    def send(self, msg_obj, *args, **kwargs):
+        params = self.populate_params(msg_obj)
+        resp = requests.get(VERTEX_URL, params=params)
+        self.handle_response(msg_obj, resp.content)
```
gcapalbo: when the response is text i think `resp.text` is the preferred way to access it

corehq/messaging/smsbackends/vertex/models.py
```diff
@@ -0,0 +1,68 @@
...
+        return resp
```
gcapalbo: (we don't need to return anything here)

corehq/messaging/smsbackends/vertex/models.py
```diff
@@ -0,0 +1,68 @@
...
+
+    def handle_response(self, msg_obj, resp):
+        # in case of success store the complete response msg which should be like
+        # 570737298-2017_05_27 AS-IS. This complete id referred to as "Scheduleid"
+        # can be then used to enquire for status of this particular SMS
+        if SUCCESS_RESPONSE_REGEX_MATCHER.match(resp):
+            msg_obj.backend_message_id = resp
+            msg_obj.save()
```
gcapalbo: no need to save the object here - it will be saved by the framework automatically later

corehq/messaging/smsbackends/vertex/models.py
mkangia: Oh! Didn't know that. Makes sense now. Will update.

corehq/messaging/smsbackends/vertex/models.py
```diff
@@ -0,0 +1,68 @@
...
+        else:
+            msg_obj.set_system_error(resp)
```
gcapalbo: i agree that most of the errors aren't retryable, but we should be notified of these errors otherwise we can't act on them. i think this would be the best way to handle:

- when the response is `ES1009 Sorry unable to process request`, `msg.set_system_error(SMS.ERROR_INVALID_DESTINATION_NUMBER)`

- when the response is any of the other documented error responses, `msg.set_system_error(SMS.ERROR_TOO_MANY_UNSUCCESSFUL_ATTEMPTS)`, and notify exception with the error message. this way it's not retried but we can get notified of the error

- when the response is anything else, raise a `VertexBackendException('Unrecognized response from Vertex gateway with {response_status_code} status code, response {response_text}')`. this way it's retried and we get notified about it.

corehq/messaging/smsbackends/vertex/models.py
mkangia: whats the reason for this being preferred though?

gcapalbo: http://docs.python-requests.org/en/master/user/quickstart/#response-content

.content is binary content, .text converts to text based on the encoding of the response

corehq/messaging/smsbackends/vertex/models.py
mkangia: @gcapalbo changes [here](https://github.com/dimagi/commcare-hq/pull/16353/commits/16fa76faf502e4cce9e67d4fd41a9630ec032df7)

corehq/messaging/smsbackends/vertex/models.py
```diff
@@ -61,13 +67,31 @@ def populate_params(self, msg_obj):
...
+    def handle_failure(self, msg_obj, resp_status_code, error_message):
+        error_code, text = GATEWAY_ERROR_MESSAGE_REGEX_MATCHER.match(error_message).groups()
+        if error_code and error_code == INCORRECT_MOBILE_NUMBER_CODE:
+            msg_obj.set_system_error(SMS.ERROR_INVALID_DESTINATION_NUMBER)
+        elif (
+            error_code in GATEWAY_ERROR_CODES or
+            error_message in GATEWAY_ERROR_MESSAGES
+        ):
+            msg_obj.set_system_error(SMS.ERROR_TOO_MANY_UNSUCCESSFUL_ATTEMPTS)
+            raise VertexBackendException(error_message)
```
gcapalbo: raising the exception causes it to be retried which we can avoid for these known errors by just notifying exception like we do here: https://github.com/dimagi/commcare-hq/blob/master/corehq/messaging/smsbackends/icds_nic/models.py#L118

corehq/messaging/smsbackends/vertex/tests/test_response.py
```diff
@@ -14,11 +19,33 @@ def setUp(self):
         self.queued_sms = QueuedSMS()
 
     def handle_success(self):
-        self.vertex_backend.handle_response(self.queued_sms, TEST_SUCCESSFUL_RESPONSE)
+        self.vertex_backend.handle_response(self.queued_sms, 200, TEST_SUCCESSFUL_RESPONSE)
         self.assertEqual(self.queued_sms.backend_message_id, TEST_SUCCESSFUL_RESPONSE)
         self.assertFalse(self.queued_sms.error)
 
     def handle_failure(self):
```
gcapalbo: the test method names should start with test_ otherwise i don't think they get picked up by the runner

mkangia: Aha! so that is what was happening. I had to run this one by one. Thanks for flagging this.

corehq/messaging/smsbackends/vertex/models.py
mkangia: cool. Updated.

corehq/messaging/smsbackends/vertex/tests/test_response.py
```diff
@@ -18,28 +19,30 @@ def setUp(self):
...
-        self.queued_sms.error = False
-        with self.assertRaisesMessage(VertexBackendException, TEST_FAILURE_RESPONSE):
-            self.vertex_backend.handle_response(self.queued_sms, 200, TEST_FAILURE_RESPONSE)
-        self.assertEqual(self.queued_sms.system_error_message, SMS.ERROR_TOO_MANY_UNSUCCESSFUL_ATTEMPTS)
-        self.assertTrue(self.queued_sms.error)
+            self.assertEqual(self.queued_sms.system_error_message, SMS.ERROR_TOO_MANY_UNSUCCESSFUL_ATTEMPTS)
+            self.assertTrue(self.queued_sms.error)
+            exception_notifier.assert_called_once_with(None, TEST_NON_CODE_MESSAGES)
+
+            with mock.patch('corehq.messaging.smsbackends.vertex.models.notify_exception') as exception_notifier:
```
gcapalbo: did you mean to nest this block or should it be unindented by one?

corehq/messaging/smsbackends/vertex/models.py
```diff
@@ -87,7 +88,7 @@ def handle_failure(self, msg_obj, resp_status_code, error_message):
             error_message in GATEWAY_ERROR_MESSAGES
         ):
             msg_obj.set_system_error(SMS.ERROR_TOO_MANY_UNSUCCESSFUL_ATTEMPTS)
-            raise VertexBackendException(error_message)
+            notify_exception(None, error_message)
```
gcapalbo: would be good to also prepend something like "Error with the Vertex SMS Backend: "

corehq/messaging/smsbackends/vertex/tests/test_response.py
mkangia: oops.. fixing..

corehq/messaging/smsbackends/vertex/models.py
mkangia: Done.

gcapalbo (approved):

mkangia:
Thank you @gcapalbo

PR #16355, opened 2017-05-26
default to since if a variable doesn't exist

emord:
interrupt @benrudolph buddy @calellowitz 

if the variable doesn't exist then the export doesn't get filtered at all so it will be since project start. this makes the popup super confusing.

I don't know why some exports don't have this initialized. All new exports have this value from my testing.


benrudolph:
huh interesting

benrudolph:
could be the migrated exports?

PR #16356, opened 2017-05-26
eNikshay Agency Registration

esoergel:
https://docs.google.com/spreadsheets/d/1R1rv3bpJxPRQSeTWDMU3IXX4lVhWMfZvLnxye3riizQ/edit#gid=1458284637

### DIFF
--- a/custom/enikshay/user_setup.py
+++ b/custom/enikshay/user_setup.py
@@ -7,6 +7,7 @@ assumptions laid out here.
 import math
 import re
 import uuid
+from crispy_forms import layout as crispy
 from django import forms
 from django.utils.text import slugify
 from django.utils.translation import ugettext as _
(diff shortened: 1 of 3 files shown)


esoergel:
@proteusvacuum @orangejenny

custom/enikshay/user_setup.py
```diff
@@ -267,50 +269,104 @@ def __init__(self, *args, **kwargs):
...
 
-class ENikshayUserDataEditor(CustomDataEditor):
-    """Custom User Data (everywhere it turns up)"""
-    def __init__(self, *args, **kwargs):
-        super(ENikshayUserDataEditor, self).__init__(*args, **kwargs)
+# pcp -> MBBS
+# pac -> AYUSH/other
+# plc -> Private Lab
+# pcc -> pharmacy / chemist
+# dto -> Field officer??
```
nickpell: I think field officer is pdr?

nickpell (approved):

custom/enikshay/user_setup.py
esoergel: Nice, thanks - [the spec](https://docs.google.com/spreadsheets/d/1R1rv3bpJxPRQSeTWDMU3IXX4lVhWMfZvLnxye3riizQ/edit#gid=1458284637) said dto, but seemed unsure

proteusvacuum: that column in the spec was detailing the location type that should be created when someone selects that particular user type. For the field officer, I don't think they get a new location created, but instead are assigned to the DTO level (AFAIK, I think we were going to check this with Erin, which is why I put it in parens). 

I don't know where in this code that happens, if at all, so I'm assuming this is resolved.

PR #16357, opened 2017-05-28
adherence calcs; skip unnecessary update

sravfeyn:
@mkangia

### DIFF
--- a/custom/enikshay/tasks.py
+++ b/custom/enikshay/tasks.py
@@ -336,7 +336,23 @@ class EpisodeAdherenceUpdate(object):
                         True,
                         "No fixture item found with schedule_id {}".format(adherence_schedule_id)
                     )
-        return {'update': update, 'debug_data': debug_data}
+        if self.check_if_needs_update(update):
+            return {'update': update, 'debug_data': debug_data}
+        else:
+            return {'update': None, 'debug_data': debug_data}
+
+    def check_if_needs_update(self, case_properties_expected):
+        """
+        Args:
+            case_properties_expected: dict of case property name to values
+
+        Returns:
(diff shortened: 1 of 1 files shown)


custom/enikshay/tasks.py
```diff
@@ -63,32 +73,36 @@ def run(self):
...
-                logger.error(
-                    "Error calculating updates for episode case_id({}): {}".format(
-                        episode.case_id,
-                        e
-                    )
-                )
         logger.info(
-            "Summary of enikshay_task: Updated {updates}, errored on {errors} and {noupdates} "
-            "cases didn't need update ".format(updates=update_count, errors=error_count, noupdates=noupdate_count)
+            "Summary of enikshay_task: domain: {domain}, duration (sec): {duration} "
```
mkangia: Whats the need for using `domain` here? Can this be used for a domain except `enikshay`?

custom/enikshay/tasks.py
```diff
@@ -45,6 +45,16 @@ def enikshay_task():
         updater.run()
 
 
+class Timer:
```
mkangia: Nice!

mkangia (approved):
Overall looks good. Just in context of domain, if this can't be used for any domain other than 'enikshay' we might want to restrict it to that. Can skip it if that is unnecessary.

custom/enikshay/tasks.py
sravfeyn: This is run for only adherence_task [toggle enabled domains](https://github.com/dimagi/commcare-hq/blob/73c71b6cc753c0ea949596f290fbd6d650a158ab/custom/enikshay/tasks.py#L42), there are couple of enikshay test domains that will also need to have this run

PR #16360, opened 2017-05-29
Reset on abort

benrudolph:
interrupt @emord, buddy @proteusvacuum


emord:
Is this for http://manage.dimagi.com/default.asp?254937#1338304 ?

This solution looks pretty hacky, but at the same time, I couldn't figure out how to do this without a major refactor.

👍 from me if the build passes. one of them failed, but it didn't look related

benrudolph:
yea agree it's hacky. i tried to find a better solution. but started spending way too much time on it because it definitely does need a refactor. felt like this was easier. feel free to refactor if you have something in the works.

yes that was the ticket. i couldn't find it, but that makes sense because it was assigned to you heh sorry about that

PR #16361, opened 2017-05-29
Remove user from group

benrudolph:
@dannyroberts @emord what do you think about this approach? i don't love adding this property to the group model, but i also wasn't a huge fan of having the caller be responsible for saving the user removed

https://manage.dimagi.com/default.asp?252367

### DIFF
--- a/corehq/pillows/groups_to_user.py
+++ b/corehq/pillows/groups_to_user.py
@@ -3,6 +3,7 @@ from corehq.apps.change_feed.consumer.feed import KafkaChangeFeed, KafkaCheckpoi
 from corehq.apps.change_feed import topics
 from corehq.apps.groups.models import Group
 from corehq.elastic import stream_es_query, get_es_new, ES_META
+from corehq.apps.es import UserES
 from corehq.pillows.mappings.user_mapping import USER_INDEX, USER_INDEX_INFO
 from pillowtop.checkpoints.manager import get_checkpoint_for_elasticsearch_pillow
 from pillowtop.pillow.interface import ConstructedPillow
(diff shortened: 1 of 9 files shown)


corehq/pillows/groups_to_user.py
```diff
@@ -67,16 +68,29 @@ def update_es_user_with_groups(group_doc, es_client=None):
...
 def stream_user_sources(user_ids):
-    q = {"filter": {"and": [{"terms": {"_id": user_ids}}]}}
-    for result in stream_es_query(es_index='users', q=q, fields=["__group_ids", "__group_names"]):
-        group_ids = result.get('fields', {}).get("__group_ids", [])
+    results = (
+        UserES()
+        .user_ids(user_ids)
+        .fields(['_id', '__group_ids', '__group_names'])
+        .scroll()
+    )
```
benrudolph: note this query is _slightly_ different from the original. this one excludes deleted and inactive users, which i think is fine, but maybe for consistency sake we should keep them up to date?

corehq/pillows/groups_to_user.py
```diff
@@ -67,16 +68,29 @@ def update_es_user_with_groups(group_doc, es_client=None):
             doc = {"__group_ids": list(user_source.group_ids), "__group_names": list(user_source.group_names)}
             es_client.update(USER_INDEX, ES_META['users'].type, user_source.user_id, body={"doc": doc})
 
+    for user_source in stream_user_sources(group_doc.get("removed_users", [])):
+        if group_doc["name"] in user_source.group_names or group_doc["_id"] in user_source.group_ids:
+            user_source.group_ids.remove(group_doc["_id"])
+            user_source.group_names.remove(group_doc["name"])
+            doc = {"__group_ids": list(user_source.group_ids), "__group_names": list(user_source.group_names)}
+            es_client.update(USER_INDEX, ES_META['users'].type, user_source.user_id, body={"doc": doc})
```
emord: Should the user be removed from `group.removed_users` at this point?

emord (commented):
Curious why you used the group to do this? Does the user not get saved when they are removed from a group?

Really wish users/groups were in a relational db...

corehq/pillows/groups_to_user.py
benrudolph: yes it should, good catch

benrudolph:
> Curious why you used the group to do this? Does the user not get saved when they are removed from a group?
users do not get saved when we remove them from a group 😞 one solution is to always save when removing a user from a group and then also save the users.
> Really wish users/groups were in a relational db...
yes, yes indeed

corehq/pillows/groups_to_user.py
benrudolph: err sorry, i think it's functionally fine as is. if we want to remove the user from that property it'd require us to do a couch fetch which seems meh. i don't see that property getting too large

dannyroberts:
Clever. 👍

PR #16365, opened 2017-05-29
Mk app builder condition support

mkangia:
[Ticket](https://trello.com/c/wekRPsTb/137-app-builder-changes-for-enikshay-phase-2b?menu=filter) # 3

When configuring a case list or detail, allow specifying a calculation as the "key" in the ID mapping (similar to how we do that for the icon preview)

[test](https://docs.google.com/document/d/1nWSK2m0sRcrIvgoLkT3oZS0XtPiDTXSaUAhqrV0ZLsw/edit)


corehq/apps/app_manager/suite_xml/sections/details.py
```diff
@@ -512,17 +513,23 @@ def _get_column_context(self, column):
                 )
             )
 
+        context['variables'] = ''
         if column.format == "enum":
-            context["enum_keys"] = self._get_enum_keys(column)
+            context["variables"] = self._get_enum_keys(column)
         return context
 
     def _get_enum_keys(self, column):
```
orangejenny: minor: i'd rename this

orangejenny:
Hey Manish, this looks like a good direction to me.

Is this going to auto-detect whether the key is a plain string or a calculation based on [this regex](https://github.com/dimagi/commcare-hq/blob/master/corehq/apps/app_manager/models.py#L1842)? You might already be thinking about this, but it'd be good for the case list page to give users some kind of feedback when CommCare is going to interpret their key as an expression, since that might not be what they intend. I'm a little worried about novice users using a key with a period or apostrophe and then breaking their apps.

Alternatively, this could be a different display property format ("Calculation Mapping"?). What do you think?

mkangia:
@orangejenny I actually didnt think of that since this seems like already present [here](https://github.com/dimagi/commcare-hq/blob/78e60240f2f218131b5d7f3a77ea99e4602234ac/corehq/apps/app_manager/detail_screen.py#L382). I thought this was ok to do.

sheelio:
@orangejenny  @mkangia 

The Icon format option doesn't distinguish between calculations and values (but it is a feature preview).  I'm ok with moving this to a feature preview and making it its own Format type if that is cleaner / easier.

orangejenny:
@sheelio @mkangia 

Yeah, thinking about it more, I'm in favor of a separate format option. I think otherwise we'll end up with people who want to use keys like `5.0` or `don't know` in regular ID mappings and we'd have to tell them to use something like `. == "5.0"` which seems like a crazy workaround if you're not used to xpath.

mkangia:
Thank you for input @orangejenny @sheelio .

I have updated to add a new format instead "Conditional Enum" and placed it as feature preview.

corehq/apps/app_manager/suite_xml/case_tile_templates/tdh.txt
```diff
@@ -58,12 +58,17 @@
       </header>
       <template>
         <text>
-          &lt;b&gt;{header[prefix]} &lt;/b&gt; <xpath function="{header[xpath_function]}"/>
+          &lt;b&gt;{header[prefix]} &lt;/b&gt;
+          <xpath function="{header[xpath_function]}">
+            {header[variables]}
```
dannyroberts: This looks like a real change (not just whitespace). Was that intended? I wouldn't expect this new feature to change the output for any existing apps or tests

dannyroberts (commented):
Looks really good. Just had this one comment.

corehq/apps/app_manager/suite_xml/case_tile_templates/tdh.txt
mkangia: @dannyroberts definitely a real change. There was no option to add xpath variables till now for these fields. So added that ability for all fields and also there sort options.

dannyroberts: Ah, that makes lots of sense. Looking again now, I had misinterpreted this as a _test_ file, but it's actually a production file for the custom case tiles "feature" we wrote for tdh.

Thanks for the explanation.

dannyroberts (approved):

snopoke (commented):
Just one question at this stage: is it necessary to make this a new enum option? Can we not just make the current enum option support calculations? If we did that would it break any existing apps?

mkangia:
@snopoke Jenny raised concerns about the way we would detect conditions over text. Details [here](https://github.com/dimagi/commcare-hq/pull/16365#issuecomment-304913775)

snopoke:
Right, sorry, just read the comment history now.

corehq/apps/app_manager/static/app_manager/js/detail-screen-config.js
```diff
@@ -756,6 +756,7 @@ hqDefine('app_manager/js/detail-screen-config.js', function () {
                         property_name: that.field,
                         multimedia: that.screen.config.multimedia,
                         values_are_icons: that.original.format == 'enum-image',
+                        values_are_conditions: that.original.format == 'conditional-enum'
```
dimagimon: Expected '===' and instead saw '=='. (eqeqeq)
Missing trailing comma. (comma-dangle)

corehq/apps/app_manager/static/app_manager/js/detail-screen-config.js
```diff
@@ -824,9 +825,9 @@ hqDefine('app_manager/js/detail-screen-config.js', function () {
                         that.filter_xpath_extra.ui.detach();
                         that.calc_xpath_extra.ui.detach();
                         that.time_ago_extra.ui.detach();
-
-                        if (this.val() === "enum" || this.val() === "enum-image") {
+                        if (this.val() === "enum" || this.val() === "enum-image" || this.val() === 'conditional-enum') {
                             that.enum_extra.values_are_icons(this.val() === 'enum-image');
+                            that.enum_extra.values_are_conditions(this.val() === 'conditional-enum')
```
dimagimon: Missing semicolon. (semi)

corehq/feature_previews.py
```diff
@@ -76,6 +76,14 @@ def preview_values_by_name(domain):
     help_link='https://help.commcarehq.org/display/commcarepublic/Adding+Icons+in+Case+List+and+Case+Detail+screen'
 )
 
+CONDITIONAL_ENUM = FeaturePreview(
+    slug='conditional_enum',
+    label=_('Conditional Enum in Case List'),
+    description=_(
+        "Conditional ID mapping in case tiles. "
```
snopoke: how about this: `Specify a custom xpath expression to calculate a lookup key in the case list or case detail screen enum column.`

snopoke (commented):
main thing is the question about the sort_node

corehq/apps/app_manager/detail_screen.py
```diff
@@ -352,8 +352,39 @@ def variables(self):
...
 
 
+@register_format_type('conditional-enum')
+class ConditionalEnum(Enum):
+    @property
+    def sort_node(self):
+        node = super(ConditionalEnum, self).sort_node
+        if node:
+            node.order = "-2"
+            node.direction = "ascending"
```
mkangia: Sort nodes are not added with variables. This overrides the sort node to add variables for the sort node. Plus adds the order and direction which i found was used for sort nodes as default for example in `tdh.txt` we have `<sort type="string" order="-2" direction="ascending">`

snopoke: I don't think that's correct re the sort order etc. If you look at this you'll see that `-2` means index: https://github.com/dimagi/commcare-hq/blob/f390477b6ac24c0c0321d915201abf5dbf820161/corehq/apps/app_manager/detail_screen.py#L175-L175

mkangia: Thank You Simon. Removed the order and direction assignment. So now the sort node is added without any order and direction. Working well now.

PR #16367, opened 2017-05-30
Temporarily disable the custom setup rules

esoergel:
@proteusvacuum
we're pushing this back, and it's confusing in it's current state, so I figure it's best to just disable
https://trello.com/c/R2ZKRcmc/84-orgs-custom-user-location-management-rules

### DIFF
--- a/custom/enikshay/tests/test_user_setup.py
+++ b/custom/enikshay/tests/test_user_setup.py
@@ -26,6 +26,7 @@ from ..models import IssuerId
 
 
 @flag_enabled('ENIKSHAY')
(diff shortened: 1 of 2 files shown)


proteusvacuum:
want to just make a different toggle for this, so that we can turn on again without another deploy?

esoergel:
I've also got another branch going to move users off signals to a subclass, so I don't wanna turn this on until that's merged.  I do want to be able to turn this off and on independently of the other stuff, though, so I'll create a different flag on that branch.

proteusvacuum (approved):

PR #16373, opened 2017-05-30
Externalize JavaScript: App settings

orangejenny:
Review by commit.

@emord / @millerdev


corehq/apps/app_manager/static/app_manager/js/app_logos.js
```diff
@@ -0,0 +1,60 @@
+hqDefine("app_manager/js/app_logos.js", function() {
+    var manager = function() {
```
millerdev: This closure seems unnecessary since the `hqDefine` is already a closure and this one is called exactly once immediately after it is defined. When removing it, I think the only change needed is to do `var manager = self = {};` instead of `var self = this;`

corehq/apps/app_manager/static/app_manager/js/app_logos.js
```diff
@@ -0,0 +1,60 @@
...
+        var self = this;
+        var HQMediaUploaders = hqImport("hqmedia/js/hqmediauploaders.js").get(),
+            initial_page_data = hqImport("hqwebapp/js/initial_page_data.js").get;
+        var refs = initial_page_data('media_refs');
+        var media_info = initial_page_data('media_info');
+
+        self.image_refs = {};
```
millerdev: I think this should be defined as `var image_refs = {};` instead of exposed as an attribute of the manager. Below on line 17 and lower it is accessed as `image_refs`, not `self.image_refs` and it seems unnecessary to expose this private internal object publicly.

millerdev (commented):
Looking good. Found a couple small things, one (`self.image_refs`) looks like a bug.

PR #16375, opened 2017-05-30
remove bihar pillow

emord:
https://github.com/dimagi/commcare-hq-deploy/pull/269

Removes it from settings here. Will save deleting the pillow code for when we're able to totally kill that directory

@benrudolph @orangejenny


orangejenny (approved):

benrudolph:
yessss

PR #16382, opened 2017-05-31
Added RNTCP PMDT Treatment Register report

kkrampa:
@sheelio


sheelio:
@NoahCarnahan 

Have you been reviewing this eNikshay reports?

NoahCarnahan:
@kkrampa, do these changes modify any of the existing indicators in the episode data source? (The diff is pretty big, so sorta hard for me to tell)

kkrampa:
@NoahCarnahan 
I modified one indicator but yeah diff looks messy.

I've just figured out that in this report we need only treatment public records. Do you think it's worth to extract this into new data source? Episode is really big right now and it takes a lot of time to rebuild it.

kkrampa:
I made not static version of this on Sheel's domain: https://india.commcarehq.org/a/sheel-enikshay/reports/configurable/4a6b943664b07e0ef351d97a759d7474/

NoahCarnahan:
Ok PR looks good then. I think creating a separate data source will speed up the build time for either one of the new data sources individually, but if you're rebuilding both it will probably take the same amount of time. I'll merge this PR as is, and you can separate them in a different PR if you think it's worth it.

PR #16383, opened 2017-05-31
Run new-style messaging rules on case-post-save

gcapalbo:
@emord 
buddy @sravfeyn


snopoke (commented):
Didn't review closely but looks good

emord (approved):

PR #16385, opened 2017-05-31
Notify on slow changes

benrudolph:
What do you think about adding this logging to sentry? i'm a little worried that this'll just spam a lot of events. It should rollup by pillow, but it's still unclear how sentry country events. i'm wondering if we can tinker with the sampling rate.

Example: https://sentry.io/dimagi/commcarehq/issues/283932782/


mkangia (approved):

corehq/ex-submodules/pillowtop/pillow/interface.py
```diff
@@ -119,6 +123,21 @@ def process_changes(self, since, forever):
...
+                        extra = change.metadata.to_json() if change.metadata else {}
+                        extra.update({
+                            'duration': '{} seconds'.format(timer.duration),
+                            'threshold': self.slow_threshold,
+                        })
+                        client.captureMessage(
+                            u'{} change slow to process'.format(self.pillow_id),
+                            extra=extra,
+                            time_spent=int(timer.duration * 1000),  # Convert to milliseconds
+                            fingerprint=[self.pillow_id],
```
benrudolph: could potentially add the topic here to rollow up by pillow id and topic

corehq/ex-submodules/pillowtop/pillow/interface.py
```diff
@@ -119,6 +123,21 @@ def process_changes(self, since, forever):
                             self._record_checkpoint_in_datadog()
                         self._record_change_success_in_datadog(change)
                     self._record_change_in_datadog(change, timer)
+
+                    if timer.duration > self.slow_threshold:
```
emord: Why not put this in `_record_change_in_datadog`?

corehq/ex-submodules/pillowtop/pillow/interface.py
emord: Would also be nice to have the domain as well.

corehq/ex-submodules/pillowtop/pillow/interface.py
benrudolph: heh, cause it's not datadog. but i'll put it in `_record_change_in_sentry`

corehq/ex-submodules/pillowtop/pillow/interface.py
benrudolph: adding domain will cause a separate entry in sentry for each domain which might get quite large. are you thinking that there actually won't be that many docs from different domains that break the threshold?

corehq/ex-submodules/pillowtop/pillow/interface.py
emord: d'oh. obviously.

corehq/ex-submodules/pillowtop/pillow/interface.py
emord: that makes sense. my reasoning is that domains have extremely differently structured data so I think it would be helpful when something comes up to see the same domain causing issues. I think it should be in the change.metadata so it'll be under extra. Is there a way in sentry to see that in a nice way?

benrudolph: at the bottom of the error message it'll show you `extra`: https://sentry.io/dimagi/commcarehq/issues/283929017/. however when it rolls up from multiple domains, i believe it'll overwrite the extra with the most current event

snopoke: re billing every event that we send to sentry is billed unless if get's filtered on their side (we don't have any filters set up though) or it's over our configured rate limits.

snopoke (commented):
I really like the idea of tracking this but not sure sentry is the right place. @benrudolph what action do you expect to be able to take with this info?

From what I've seen this is mostly an issue with UCR's which are domain specific. I'd love to have a way of reporting back to users that their UCR's are slow and even throttling / ignoring UCR's that are too slow.

emord:
> I'd love to have a way of reporting back to users that their UCR's are slow and even throttling / ignoring UCR's that are too slow.

These are logged right now at 0.5 seconds and sent as soft asserts at 5 seconds. My strategy has been to enable them as async processing UCRS if they hit the threshold too much

benrudolph:
> what action do you expect to be able to take with this info?
i think i was just hoping to get a sense of what documents/domains are very slow and then be able to optimize them. right now that isn't super easy to do. the impetus for this was today reportcase and casetoelasticsearch got really backed up and we were processing 100 changes in 3 mins

unfortunately i don't think the inbound filters will be able to help us 😞 

agree this might blow up our events too much even if the sampling rate is low. are you suggesting we just log this to a file or something?

snopoke:
My main question is what you expect to do with this info. We already have
timing info in Datadog which we could use to create alerts from but still,
what will we do if the alert triggers or we get there sentry events?

benrudolph:
Once i know the document that is slow i was hoping to write code to prevent it from being slow. One other time this was an issue was when we had the case that had 45k forms submitted to it and it slowed our change feed to a halt. More I think about it though the more I think we'll stop actually using that information... closing for now

PR #16390, opened 2017-05-31
don't allow setting checkpoints ahead of the latest possible

emord:
This one was pretty annoying. the code was incrementing each offset even if it was past the latest available. The reasoning was good (it's covered in the initial comment in this function), but if it increments it to an offset that doesn't exist it gets reset to the earliest available offset (a change from one month ago)

http://manage.dimagi.com/default.asp?255255#1340267

@benrudolph @snopoke 

buddy @orangejenny


snopoke (approved):
Nice find

benrudolph:
wow nice 👏

PR #16391, opened 2017-05-31
Update expected parameter names from BETS

esoergel:
@nickpell @proteusvacuum


proteusvacuum (approved):

proteusvacuum:
(assuming you made these changes based on testing their endpoint, not just the doc they sent?)

PR #16396, opened 2017-05-31
Model reconcilliation task: prescription refill due dates

proteusvacuum:
https://docs.google.com/document/d/1s1-MHKS5I8cvf0_7NvcqeTsYGsXLy9YGArw2L7cZSRM/edit#

@esoergel @NoahCarnahan 
🐟 / 🐡

### DIFF
--- a/custom/enikshay/case_utils.py
+++ b/custom/enikshay/case_utils.py
@@ -437,7 +437,7 @@ def get_prescription_vouchers_from_episode(domain, episode_case_id):
     ]
 
 
-def get_approved_prescription_vouchers_from_episode(domain, episode_case_id):
+def get_fulfilled_prescription_vouchers_from_episode(domain, episode_case_id):
     return [
         voucher for voucher in get_prescription_vouchers_from_episode(domain, episode_case_id)
         if (voucher.get_case_property("voucher_type") == CASE_TYPE_PRESCRIPTION
--- a/custom/enikshay/tasks.py
+++ b/custom/enikshay/tasks.py
@@ -406,20 +406,18 @@ class EpisodeVoucherUpdate(object):
         return sorted(relevant_vouchers, key=lambda v: v.get_case_property('date_issued'))
 
     @staticmethod
-    def _is_an_update(existing_properties, new_properties):
+    def _updated_fields(existing_properties, new_properties):
+        updated_fields = {}
         for prop, value in new_properties.items():
             if unicode(existing_properties.get(prop, '--')) != unicode(value):
-                return True
+                updated_fields[prop] = value
+        return updated_fields
 
     def update_json(self):
         output_json = {}
         output_json.update(self.get_prescription_total_days())
         output_json.update(self.get_prescription_refill_due_dates())
-
-        if not self._is_an_update(self.episode.dynamic_case_properties(), output_json):
-            return {}  # Don't trigger a case update
-
-        return output_json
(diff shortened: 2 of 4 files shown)


NoahCarnahan:
Looks good

custom/enikshay/case_utils.py
```diff
@@ -437,7 +437,7 @@ def get_prescription_vouchers_from_episode(domain, episode_case_id):
     ]
 
 
-def get_approved_prescription_vouchers_from_episode(domain, episode_case_id):
+def get_fulfilled_prescription_vouchers_from_episode(domain, episode_case_id):
```
esoergel: :+1:

custom/enikshay/tasks.py
```diff
@@ -382,17 +382,20 @@ def __init__(self, domain, episode_case):
...
+
+    def _get_fulfilled_vouchers(self):
         relevant_vouchers = [
-            voucher for voucher in all_vouchers
+            voucher for voucher in self._get_all_vouchers()
             if (voucher.get_case_property('voucher_type') == 'prescription'
                 and voucher.get_case_property('state') == 'fulfilled')
         ]
-        return sorted(relevant_vouchers, key=self._get_voucher_date)
+        return sorted(relevant_vouchers, key=self._get_fulfilled_voucher_date)
```
esoergel: What's this about?  Isn't this the same logic as `get_fulfilled_prescription_vouchers_from_episode` except sorted?

esoergel: Ah, nevermind, I see you fork the logic in a later commit

custom/enikshay/tasks.py
```diff
@@ -397,6 +397,14 @@ def _get_fulfilled_vouchers(self):
         ]
         return sorted(relevant_vouchers, key=self._get_fulfilled_voucher_date)
 
+    def _get_fulfilled_available_vouchers(self):
+        relevant_vouchers = [
+            voucher for voucher in self._get_all_vouchers()
+            if (voucher.get_case_property('voucher_type') == 'prescription'
+                and voucher.get_case_property('state') in ['fulfilled', 'available'])
```
esoergel: Flagging that there are other voucher states that may be applicable here, such as `paid` and `rejected`.  It kinda feels like we want a boolean version of `state`, something like `is_approved`.

esoergel: The full set of states is available here: https://docs.google.com/spreadsheets/d/1MCG205FOcsYsmKXHoSTjZ6A1iuqCIAESyleBl7G7PR8/edit#gid=0

proteusvacuum: Hmm, I'm wondering if it is better if we added those as a "payment_status" property? Otherwise, I've made that change here: https://github.com/dimagi/commcare-hq/pull/16411

esoergel: Yeah, that also sounds reasonable.  `state` feels overloaded, anyways.  It's easy enough to handle in HQ, so I think whatever makes the most sense from an app-building perspective.  The payment confirmation status is set [here](https://github.com/dimagi/commcare-hq/blob/fd8e4b65e86023d37c2482894b45f01be300efca/custom/enikshay/integrations/bets/views.py#L71-L71).
I'm not planning on making any changes right now, but if you or @ctsims would like, I'd be happy to set `payment_status` instead of `state`.

PR #16397, opened 2017-06-01
Index subcase to parent even if form doesn't have a non-subcase action

orangejenny:
http://manage.dimagi.com/default.asp?246906


orangejenny:
@snopoke Does this seem reasonable? The line changed has been around for 5 years, so I'm a little wary of changing it, but tests do pass.

code buddy @emord

snopoke:
Pretty sure a form like this would error on the phone:
```
<bind calculate="/data/case/@case_id" nodeset="/data/subcase_0/case/index/parent"/>
```
This references a node that doesn't exist (`/data/case/@case_id`).

The reason that check is there is because the code blow blindly references the case block to create the index reference.

The behaviour in the ticket is expected since the form must have a reference to the parent case in order to create the child case index (unless we change the way the case management is done). There is no need to load / save any case properties in the parent case block.

I think the most correct approach would be to not include the parent case block if it doesn't save any properties and just reference the parent ID directly (but only if the user has specified that the form updates / closes a case otherwise we'll break existing forms).

orangejenny:
@snopoke Going to close this out - the original ticket is and edge case, and unless I'm misunderstanding, updating this PR to work correctly isn't trivial. Don't think this is high enough priority to dig into more.

snopoke:
👍

PR #16399, opened 2017-06-01
Add case type to case changes tab

proteusvacuum:
When you have lots of cases in the case changes tab, it is hard to figure out what each of them is. 

screenshot from 2017-06-01 10-42-34

@NoahCarnahan 
@orangejenny for feedback


NoahCarnahan (approved):
Nice, I just had this issue looking at an enikshay case 10 minutes ago.

orangejenny:
Nice! Two minor things:
- how about `label-default` (grey) instead? HQ trains people to click on bright blue elements.
- is it feasible to not show these unless there are multiple case types on the page (or the app uses multiple types)? not a big deal if not.

orangejenny (approved):

PR #16400, opened 2017-06-01
App manager v2 copy: 'publish' not 'deploy'

orangejenny:
Based on conversation in #product

@emord / @biyeun / @proteusvacuum

### DIFF
--- a/corehq/apps/app_manager/static/app_manager/js/releases.js
+++ b/corehq/apps/app_manager/static/app_manager/js/releases.js
@@ -398,7 +398,7 @@ hqDefine('app_manager/js/releases.js', function () {
                     } else if (self.lastAppVersion() !== self.currentAppVersion()) {
                         self.actuallyMakeBuild();
                     } else {
-                        window.alert(gettext("No new changes to deploy!"));
+                        window.alert(gettext("No new changes!"));
                     }
                 },
                 error: function () {
--- a/corehq/apps/app_manager/templates/app_manager/v2/partials/releases_deploy_modal.html
+++ b/corehq/apps/app_manager/templates/app_manager/v2/partials/releases_deploy_modal.html
@@ -7,7 +7,7 @@
             <button type="button" class="close" data-dismiss="modal">
                 <span aria-hidden="true">&times;</span>
             </button>
-            <h4 class="modal-title">{% trans "Making new version" %} <span data-bind="text: version"></span></h4>
+            <h4 class="modal-title">{% trans "Version" %} <span data-bind="text: version"></span></h4>
         </div>
         <div class="modal-body">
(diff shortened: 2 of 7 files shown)


corehq/apps/app_manager/templates/app_manager/v2/partials/releases_deploy_modal.html
```diff
@@ -7,14 +7,14 @@
             <button type="button" class="close" data-dismiss="modal">
                 <span aria-hidden="true">&times;</span>
             </button>
-            <h4 class="modal-title">{% trans "Deploying version" %} <span data-bind="text: version"></span></h4>
+            <h4 class="modal-title">{% trans "Making new version" %} <span data-bind="text: version"></span></h4>
```
proteusvacuum: At this point, the version is already made. Maybe "Download version {xyz}"?

corehq/apps/app_manager/static/app_manager/js/releases.js
```diff
@@ -398,7 +398,7 @@ hqDefine('app_manager/js/releases.js', function () {
                     } else if (self.lastAppVersion() !== self.currentAppVersion()) {
                         self.actuallyMakeBuild();
                     } else {
-                        window.alert(gettext("No new changes to deploy!"));
+                        window.alert(gettext("No new changes!"));
```
proteusvacuum: gahh, this alert is so annoying. Not saying you should change it here, just venting 😝

orangejenny: really? what would you want to happen? suppose we could not show the "Make New Version" button if you don't have changes...but this is one of the few pages that works reasonably well if you're making changes in other tabs.

corehq/apps/app_manager/templates/app_manager/v2/partials/releases_deploy_modal.html
orangejenny: Oops, I confused this with the message that shows while the new version is building. Updated to just say `Version {xyz}`.

corehq/apps/app_manager/static/app_manager/js/releases.js
proteusvacuum: something other than an alert box, which is quite jarring. Never thought exactly how this should be shown, but since there is already the "getting latest versions" and "building latest version" loading text, it could probably be a transient message that goes in that same spot.

proteusvacuum (approved):

corehq/apps/app_manager/static/app_manager/js/releases.js
orangejenny: got it, yeah, alert boxes are awful. will make it better at some point.

PR #16401, opened 2017-06-01
Create custom repeater for beneficiary registration in BETS

esoergel:
@nickpell @proteusvacuum


nickpell (approved):

proteusvacuum:
Don't you need a view for this? Or does the generic view just pick this one up?

proteusvacuum:
Also, you don't have a payload generator attached to this, so I think this will have undefined behaviour.

esoergel:
It works with the generic view, actually.  It's a subclass of CaseRepeater, so it uses the default CaseRepeaterJsonPayloadGenerator.  They didn't want a modified payload, so I figured we'd just stick with that.

esoergel:
The way payload generators are linked to repeaters changed recently, so now it's specified as a [class attr](https://github.com/dimagi/commcare-hq/blob/master/corehq/apps/repeaters/models.py#L312-L312), meaning they'll be inherited.  Not sure what would've happened with the old decorator syntax.

PR #16405, opened 2017-06-01
assign properties to prescription case

nickpell:
@esoergel


nickpell:
✅ @proteusvacuum @mkangia

PR #16406, opened 2017-06-01
adding varchar_pattern_ops index to devicelog table

calellowitz:
@emord @snopoke @benrudolph 
This creates an index with the proper options so that it can do comparisons on varchar columns in postgres. Django wont create these automatically for multicolumn indexes and that was preventing the device log filter from using the index when doing the wildcard match it uses in https://github.com/dimagi/commcare-hq/blob/master/corehq/apps/reports/filters/api.py#L274


calellowitz:
Note, this is only a partial fix. It works when you include most of the device id but not when its only the first few characters. I spent some time looking around and I dont think the distinct part really hurts that much. Once it can scan on the index, the values are already sorted so it just needs to keep going until it gets the first 10 unique ones which adds a tiny bit of time to scan but takes no more operations remove the distinct part.

corehq/ex-submodules/phonelog/migrations/0008_devicelog_varchar_index.py
```diff
@@ -0,0 +1,30 @@
...
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('phonelog', '0007_devicelog_indexes'),
+    ]
+
+    operations = [
+        HqRunSQL(
+            add_if_not_exists(
```
emord: I think this is the wrong way to use this migration. I think it should be like https://github.com/dimagi/commcare-hq/blob/master/corehq/ex-submodules/phonelog/migrations/0007_devicelog_indexes.py

It was used incorrectly in accounting and I think it cause a headache on deploy or locally although I don't remember exactly what

calellowitz: I don't think i can use `AlterIndexIfNotExists` since that relies on django to create the sql and it will not create the right sql.

benrudolph:
interesting and nice find. why is it that this doesn't work with few characters? i feel like if the index was sorted then it would be very quick to find the first 10 items. have you tried running the query without distinct to see if it makes any difference?

PR #16408, opened 2017-06-01
Fix string comparison in EpisodeAdherenceUpdate

NoahCarnahan:
I believe that this faulty comparison is causing this web adherence report QA bug:
https://manage.dimagi.com/default.asp?254727

We may want to consider kicking off the enikshay task that updates adherence dose counts early, as this could be affecting other cases.

buddy @proteusvacuum 
FYI @sravfeyn

### DIFF
--- a/custom/enikshay/tasks.py
+++ b/custom/enikshay/tasks.py
@@ -215,7 +215,7 @@ class EpisodeAdherenceUpdate(object):
             else:
                 valid_cases = filter(
                     lambda case: (
-                        case['adherence_source'] is 'enikshay' and
+                        case['adherence_source'] == 'enikshay' and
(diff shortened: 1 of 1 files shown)


proteusvacuum:
Agree we should re-try the adherence task @sravfeyn is there a way we can kick this off manually once this is deployed?

sravfeyn:
I have kicked off the re-try. There is no UI to trigger it, since it's a celery task I have been doing `episode_task.apply()` in a prod Django shell (use a gnu screen since the update can take time). It's not difficult to add some kind of admin UI for this if required.

proteusvacuum:
OK. I think that's fine for now.

Thanks!

PR #16412, opened 2017-06-01
Ce/impact report

calellowitz:
@kaapstorm @emord 
Improvements so this command wont break es every time we try to run it

### DIFF
--- a/corehq/apps/data_analytics/esaccessors.py
+++ b/corehq/apps/data_analytics/esaccessors.py
@@ -73,20 +73,9 @@ def get_forms_for_users(domain, user_ids, start, end):
         .domain(domain)
         .submitted(gte=start, lte=end)
         .user_id(user_ids)
-        .aggregation(
-            TermsAggregation('user_id', 'form.meta.userID').aggregation(
-                TopHitsAggregation(
-                    name='top_hits_user_submissions',
-                    size=1000000,
-                    include=['form.case', 'form.@xmlns']
-                )
-            )
-        )
-        .size(0)
     )
 
-    aggregations = query.run().aggregations
-    return aggregations.user_id.buckets_dict
+    return query.scroll()
 
 
 def get_possibly_experienced(domain, start):
--- a/corehq/apps/data_analytics/esaccessors.py
+++ b/corehq/apps/data_analytics/esaccessors.py
@@ -73,6 +73,7 @@ def get_forms_for_users(domain, user_ids, start, end):
         .domain(domain)
         .submitted(gte=start, lte=end)
         .user_id(user_ids)
+        .source(['form.meta.userID', 'form.case', 'form.@xmlns'])
     )
 
     return query.scroll()
--- a/corehq/apps/data_analytics/management/commands/export_form_impact_data.py
+++ b/corehq/apps/data_analytics/management/commands/export_form_impact_data.py
@@ -1,4 +1,5 @@
 import csv
+from collections import defaultdict
 
 import dateutil
 from dateutil.relativedelta import relativedelta
@@ -8,6 +9,7 @@ from django.core.management.base import BaseCommand
 from corehq.apps.data_analytics.esaccessors import get_forms_for_users
 from corehq.apps.domain.models import Domain
 from corehq.apps.users.models import CommCareUser
+from corehq.util.log import with_progress_bar
 
 
 class Command(BaseCommand):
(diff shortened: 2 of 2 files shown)


corehq/apps/data_analytics/management/commands/export_form_impact_data.py
```diff
@@ -40,28 +43,33 @@ def handle(self, month_year, file_path, **options):
...
-                        ]
-                    ) >= 2
-                    writer.writerow([
-                        domain_name,
-                        user_id,
-                        bucket.doc_count,
-                        has_case,
-                        has_two_or_more_different_forms_submitted
-                    ])
+                        has_two_or_more_different_forms_submitted = len(
```
kaapstorm: If the user submits the same form twice, this will be true. Is that correct? If not, you'll want to use a set instead of a list.

corehq/apps/data_analytics/management/commands/export_form_impact_data.py
```diff
@@ -40,28 +43,33 @@ def handle(self, month_year, file_path, **options):
...
+                    user_dict = defaultdict(list)
+                    for form in forms:
+                        user_id = form['form']['meta']['userID']
+                        user_dict[user_id].append(form)
+                    for user_id, forms in user_dict.iteritems():
+                        has_case = bool(
+                            filter(
+                                lambda h: h.get('form', {}).get('case'),
+                                forms
+                            )
```
kaapstorm: If you use 
```python
has_case = any(h.get('form', {}).get('case') for h in forms)
``` 
then `any` will stop iterating as soon as it finds the first one, but `filter` will iterate all of them. Obviously not a biggy unless there are users with LOADS of forms.

kaapstorm (approved):
This is great. Just that one question.

corehq/apps/data_analytics/management/commands/export_form_impact_data.py
```diff
@@ -40,28 +43,33 @@ def handle(self, month_year, file_path, **options):
...
+                domain_name = domain['key']
                 user_ids = CommCareUser.ids_by_domain(domain=domain_name)
-                buckets_dict = get_forms_for_users(domain_object.name, user_ids, start_date, end_date)
-                for user_id, bucket in buckets_dict.iteritems():
-                    has_case = bool(
-                        filter(
-                            lambda h: h.get('form', {}).get('case'),
-                            bucket.top_hits_user_submissions.hits
+                for users in chunked(user_ids, 100):
+                    forms = get_forms_for_users(domain_name, users, start_date, end_date)
```
emord: I think some forms can get pretty long. Can you add a parameter to specify the fields you need?

corehq/apps/data_analytics/management/commands/export_form_impact_data.py
emord: Similar to an earlier comment, you can put this in a for loop and break once you get to two forms submitted so you don't need to process all of them

emord (commented):
I think this query could be written almost entirely in ES using one query per chunked user list and no python code.

Looks like this is just a managment command so no biggie to change that. happy to offline if you want to talk about the query i'm thinking of

calellowitz:
@emord @kaapstorm 
Optimized the loops (into one big one for fun) and got to all the feedback

corehq/apps/data_analytics/management/commands/export_form_impact_data.py
```diff
@@ -53,23 +53,24 @@ def handle(self, month_year, file_path, **options):
...
+                        for form in forms:
+                            if has_case and has_two_forms_submitted:
+                                break
+                            if not has_case and form.get('form', {}).get('case'):
+                                has_case = True
+                            if not has_two_forms_submitted:
+                                xmlns = form.get('form', {}).get('@xmlns')
+                                if xmlns:
+                                    unique_forms.add(xmlns)
+                                    if len(unique_forms) > 2:
```
emord: `>=`

corehq/apps/data_analytics/esaccessors.py
```diff
@@ -73,6 +73,7 @@ def get_forms_for_users(domain, user_ids, start, end):
         .domain(domain)
         .submitted(gte=start, lte=end)
         .user_id(user_ids)
+        .source(['form.meta.userID', 'form.case', 'form.@xmlns'])
```
emord: offlined and this is the only place this is called

emord (changes requested):

emord (approved):

PR #16421, opened 2017-06-02
Trigger episode adherence update when the data is received; tests refactor

sravfeyn:
@proteusvacuum This implements what you have asked for

cc @gcapalbo


custom/enikshay/integrations/ninetyninedots/utils.py
```diff
@@ -141,3 +142,20 @@ def update_adherence_confidence_level(domain, person_id, start_date, end_date, n
 
 def update_default_confidence_level(domain, person_id, new_confidence):
     return AdherenceCaseFactory(domain, person_id).update_default_confidence_level(new_confidence)
+
+
+def update_episode_adherence_properties(domain, person_id):
+    try:
+        episode_case = get_open_episode_case_from_person(domain, person_id)
```
sravfeyn: This could potentially be plugged into `AdherenceFactory` to avoid another case lookup, but these are functionally different, we could do some refactoring so that the view that is triggering various updates keep track of cases in memory - but wasn't sure if that is worth. In terms of where this fits into the project, is this critical enough for a refactor?

proteusvacuum: Good question... I wonder if instead we should just cache those functions?

PR #16423, opened 2017-06-02
Remove bad domain

proteusvacuum:
@emord


emord:
this was deployed on softlayer and  through the queue. I think it's ok to close this out

PR #16424, opened 2017-06-02
Remove unneeded domains from ucrs

proteusvacuum:
@nickpell @emord 
@kkrampa do you use any of these domains for testing?


nickpell:
I might need to add back a couple of these domains at some point for testing how migrated data appears in the UCRs, but fine to delete for now to reduce load.

emord (approved):

PR #16426, opened 2017-06-02
better ES service check

emord:
@esoergel @orangejenny 

This forces the elasticsearch index to refresh instead of waiting a second so it should be more consistent. had some issues recently with prod/india returning failures


emord:
@javierwilson I think you asked about this error earlier this week

benrudolph:
nice

PR #16430, opened 2017-06-03
App manager v2: toggle panels on settings pages

orangejenny:
Add widget to app/module/form settings pages to show/hide sections. This looks and works the same way it does in vellum, storing user's preferences for which sections are expanded.

screen shot 2017-06-02 at 9 32 59 pm

@emord / @biyeun


corehq/apps/app_manager/static/app_manager/js/commcaresettings.js
```diff
@@ -245,12 +246,15 @@ hqDefine('app_manager/js/commcaresettings.js', function () {
...
                 });
             });
-            section.reallyCollapse = ko.computed(function () {
-                var el = document.getElementById(section.id);
-                return section.collapse &&
-                    (!el || !el.classList.contains("in")) &&
-                    !_(section.settings).some(function (setting) { return setting.hasError(); });
-            });
+            section.collapse = hqImport("app_manager/js/section_changer.js").shouldCollapse("app-settings", section.id, section.collapse);
+            if (!COMMCAREHQ.toggleEnabled('APP_MANAGER_V2')) {
```
dimagimon: 'COMMCAREHQ' is not defined. (no-undef)

corehq/apps/app_manager/static/app_manager/js/section_changer.js
```diff
@@ -0,0 +1,115 @@
...
+            page: page,
+            section: section,
+        });
+    };
+
+    // Determine if section should be shown or not, based on localStorage and given default
+    var shouldCollapse = function(page, section, defaultCollapse) {
+        var key = getKey(page, section);
+        return localStorage.hasOwnProperty(key) ? localStorage.getItem(key) : defaultCollapse;
+    }
```
dimagimon: Missing semicolon. (semi)

corehq/apps/app_manager/static/app_manager/js/commcaresettings.js
orangejenny: This is old accordion logic that forces accordion sections on the app settings page to be open if they have any errors. It's good logic, it isn't used consistently (isn't in forms / modules / vellum) and I had trouble getting this to play nicely with the localStorage preferences combined with the default collapse settings from `commcare-settings-layout.yaml`, decided it wasn't worth spending more time getting it to work right now.

snopoke:
What's the impetus for doing this?

orangejenny:
@snopoke Short answer is to get more intermediate/advanced content out of the way of early users but still accessible to more seasoned users. App settings is the biggest example of where this is an issue, where there are currently 6 sections (2 of them specific to J2ME...) that new users like to explore but that don't have much value until you're more experienced.

The broader context is that the original v2 spec included a "simple" vs "advanced" mode, where advanced included about a dozen features that are unlikely to be relevant to early SaaS users (form/module filtering, ability to change form actions, child cases, etc.). Some concerns came up about what that simple => advanced transition will actually feel like (it's one more checkbox buried in app settings, plus the set of advanced features isn't really cohesive), so we decided to kill "advanced" mode and instead explore ways to give users finer-grained control over what features they see. One suggestion that came up was adding this kind of context menu, with semi-advanced features specific to the page you're on, so this PR adds that UI. I'm also starting to work on an "app manager labs" area to turn on/off access to individual advanced features - like feature previews, but app-specific and contained in app manager instead of domain settings so they're easier to discover. It's not set in stone, but that's the direction we've been discussing.

snopoke:
Thanks for the detailed response and I think it's good to test out
different options. This doesn't feel that much different from what we
already have since the sections are all already collapsed when you access
the advanced tab, we're just hiding them behind another button (that makes
it 4 levels deep). I also don't really see an issue with new users
'exploring' these settings as long as they don't get stuck here (which I
don't think is an issue).

Not blocking this PR though, just expressing my scepticism that this is
actually that useful here but fine to go ahead and test it out. Are we
going to do any measurement to see if there is any improvement in user
behaviour?

emord (approved):
Looks fine to me. 

general question, obviously out of scope for this PR. how hard do you think it will be to add tests to app manager js?

orangejenny:
@snopoke Fair skepticism. v2 is already downplaying the settings pages; I don't know how much of an additional gain, if any, this will be. I do think it's a promising pattern because it tested well in vellum, in terms of users working faster and expressing less confusion. Not worried about the extra click because it's a one-time thing and then the preference for displaying that section is remembered. We don't have plans to test this change specifically, but will be measuring how user behavior changes in v2 overall.

@emord Do you mean `app_manager.js` in particular? App manager in general has some js tests, although I think they're more oriented towards testing pure js functionality, not DOM interaction.

biyeun:
@orangejenny those should go in some panels to get more visual separation.

orangejenny:
offlined with @biyeun , this is good to go

PR #16432, opened 2017-06-03
Fix check for persist_tile_on_forms only when tiles are actually present

mkangia:
If `detail.persist_case_context` is True and `persist_tile_on_forms` is set to True 
but no case tiles are present (for ex when a user selects to not use case tiles but does not uncheck the persist tile on forms option) 

The persistent context block is not added which causes app to break.

### DIFF
--- a/corehq/apps/app_manager/suite_xml/sections/details.py
+++ b/corehq/apps/app_manager/suite_xml/sections/details.py
@@ -89,7 +89,9 @@ class DetailContributor(SectionContributor):
                                     )
                                     if d:
                                         r.append(d)
-                        if detail.persist_case_context and not detail.persist_tile_on_forms:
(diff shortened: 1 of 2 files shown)


corehq/apps/app_manager/suite_xml/sections/details.py
```diff
@@ -89,7 +89,9 @@ def get_section_elements(self):
                                     )
                                     if d:
                                         r.append(d)
-                        if detail.persist_case_context and not detail.persist_tile_on_forms:
+                        # add the persist case context if needed and if
+                        # case tiles are present and have their own persistent block
+                        if detail.persist_case_context and not (detail.use_case_tiles and detail.persist_tile_on_forms):
```
dimagimon: E501 line too long (120 > 115 characters)

PR #16434, opened 2017-06-04
App manager v2: fix usercase not showing on non-case forms

orangejenny:
Case config, including usercase, doesn't run unless `has_form_source` is true, and in one of the PRs externalizing javascript in form view, I moved it into the case management tab, which isn't included in v2 forms that don't have cases.

@emord


snopoke (approved):
Has conflicts

orangejenny:
thanks, fixed conflicts

emord (approved):

PR #16435, opened 2017-06-05
purge date is 30 days ago not 60 days ago

sravfeyn:
https://manage.dimagi.com/default.asp?254980#edit_0_254980

@gcapalbo

### DIFF
--- a/custom/enikshay/data_store.py
+++ b/custom/enikshay/data_store.py
@@ -1,4 +1,4 @@
-from dateutil.parser import parse
+from datetime import datetime
 
 from corehq.apps.es import filters
 from corehq.apps.userreports.models import StaticDataSourceConfiguration
@@ -32,11 +32,13 @@ class AdherenceDatastore(object):
         if result:
             # the result is sorted on 'adherence_date'
             latest_date = result[0].get('adherence_date')
-            parsed_date = parse(latest_date).date()
-            if not latest_date or not parsed_date:
-                raise EnikshayTaskException("Adherence row {} does not or has invalid 'adherence_date'".format(
-                    result[0]))
-            else:
-                return parsed_date
+            try:
+                return datetime.strptime(latest_date, '%Y-%m-%d').date()
+            except ValueError:
+                try:
+                    return datetime.strptime(latest_date, '%Y-%m-%dT%H:%M:%S').date()
+                except ValueError:
+                    raise EnikshayTaskException("Adherence row {} does not or has invalid 'adherence_date'".format(
+                        result[0]))
         else:
             return None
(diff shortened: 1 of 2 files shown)


custom/enikshay/data_store.py
```diff
@@ -32,10 +32,11 @@ def latest_adherence_date(self, episode_id):
         if result:
             # the result is sorted on 'adherence_date'
             latest_date = result[0].get('adherence_date')
-            if not latest_date or not parse_date(latest_date):
+            parsed_date = parse(latest_date).date()
```
gcapalbo: `dateutil.parser.parse` tries to fill in a lot of blanks, so for example, if you call `parse('')` with an empty string it returns the current date.

so it would be good to first validate the format of the string value to make sure it matches any of the patterns you want to support.

also note `parse()` itself can raise an exception if it can't parse the given format.

sravfeyn: Yeah, that's needed here . The empty argument issue is handled in the next line  `if not latest_date`. `parse_date` was not handling date strings with timezone info on them

gcapalbo: ok - depending on the source data it's usually still good to do validation first. for example, if the value given is a time string, it will just fill in the current date and not raise an error (`parse('02:00')`).

also if parse() raises an exception if a bad format is given will that cause an issue here?

sravfeyn:
@gcapalbo 
I have done a more strict parsing. The data comes from ES UCR, so I didn't bother about validation, but I think it's better to do more validation than less. 

Also Exceptions in this block are fine, they are caught elsewhere

custom/enikshay/data_store.py
```diff
@@ -32,10 +32,13 @@ def latest_adherence_date(self, episode_id):
...
-            if not latest_date or not parse_date(latest_date):
-                raise EnikshayTaskException("Adherence row {} does not or has invalid 'adherence_date'".format(
-                    result[0]))
-            else:
-                return parse_date(latest_date)
+            try:
+                return datetime.strptime(latest_date, '%Y-%m-%d').date()
+            except ValueError:
+                try:
+                    return datetime.strptime(latest_date, '%Y-%m-%dT%H:%M:%S').date()
```
gcapalbo: i think this is still going to have an issue if the string has timezone info on it. what i was saying before was that it's ok to use `parse()`, we should just first make sure that the string matches a regex like `^\d{4}-\d{2}-\d{2}.+$` first

sravfeyn: It's desired that it errors when a timezone-aware string comes up, the data is not supposed to have timezone info on it. So when it does we do want to investigated.

gcapalbo: i see, looks good then

gcapalbo (approved):

PR #16444, opened 2017-06-05
preindex couch periodically

emord:
@dimagi/scale-team preindexing couch takes quite a while when we don't deploy often (sometimes hours), so this will make sure it's up to date every hour. I think this will have little to none effect on prod or softlayer or swiss, but big gains on nic and maybe l10k/enikshay in the future

buddy @orangejenny

### DIFF
--- a/corehq/preindex/accessors.py
+++ b/corehq/preindex/accessors.py
@@ -53,10 +53,11 @@ def copy_design_doc(design, temp=None, delete=True):
     )
 
 
-def sync_design_doc(design, temp=None):
+def sync_design_doc(design, temp=None, force_index=False):
     sync_docs.sync_design_docs(
         db=design.db,
         design_dir=design.design_path,
         design_name=design.app_label,
         temp=temp,
+        force_index=force_index,
     )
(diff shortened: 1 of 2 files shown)


benrudolph (approved):

snopoke:
I don't think this will trigger an index. I think you just need this portion: https://github.com/dimagi/dimagi-utils/blob/dbc52f6ef32f376524c17bb65c41d5a26c6de4e0/dimagi/utils/couch/sync_docs.py#L26-L37

emord:
Thanow dependent on https://github.com/dimagi/dimagi-utils/pull/410

snopoke (commented):
Do we actually want to push the design docs to couch at all? That doesn't seem useful and may incur some penalty.

emord:
I don't think that pushing will have much of an effect, but I changed it to pull out index_design_doc

(cancelled tests until the dimagi-utils pr is merged)

snopoke:
merged dimagi/dimagi-utils#410 and updated the submodule ref in this branch

PR #16448, opened 2017-06-05
return json payload to avoid conflicting with repeat record view

mkangia:


benrudolph (approved):
a bit unfortunate, but sounds like this is necessary

proteusvacuum:
Hey, since this is going to be true for all `SOAP` repeaters, can't you just handle that case here:
https://github.com/dimagi/commcare-hq/blob/969f356089f79d6492687a0361f48eb2451d10aa/corehq/apps/repeaters/views.py#L69-L79 ?
You could add an extra check there so that it is handled correctly. 
Otherwise, this seems pretty hacky...

mkangia:
@proteusvacuum the change is just in the mixin so that keeps it in one place for all `SOAP` repeaters. Plus i don't think the repeater framework should be modified to accommodate for repeater specific changes. It expects a payload to be a `JSON` from all generators, so it seems better to just keep that.

If you still think otherwise, happy to make that change.

proteusvacuum:
Yeah, since the thing I linked to is actually in the view, you aren't overriding anything for general `Repeaters`. That view already has a bunch of conditionals based on the content type of the repeater, which is why this is happening in the first place. My preference would be to add another statement to the view. 
You could do something like this starting on line 79, to allow it to be more "generic" and not need to know about `SOAP` at all: 
```py
try: 
    json_dict = json.loads(payload)
except ValueError:
    json_dict = payload

payload = json.dumps(json_dict, indent=4)
```

mkangia:
@proteusvacuum Made that change. Used `TypeError` instead of `ValueError` as seen in the exception email

proteusvacuum (approved):

PR #16449, opened 2017-06-05
PhoneLoadBalancingMixin - Use hashed load balancing instead of round robin

gcapalbo:
https://trello.com/c/44fIudWN/691-sms-load-balancing%3A-allow-hashed-load-balancing-as-an-alternative-to-round-robin

@sravfeyn @snopoke

### DIFF
--- a/corehq/apps/sms/models.py
+++ b/corehq/apps/sms/models.py
@@ -1,5 +1,6 @@
 #!/usr/bin/env python
 import base64
+import hashlib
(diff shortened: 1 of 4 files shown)


snopoke (approved):
curious what made you change the approach

gcapalbo:
@snopoke details in the trello, but basically it makes for a better user experience since the recipient always gets messages from the same number rather than one that's always changing

PR #16453, opened 2017-06-05
better case import warn

emord:
http://manage.dimagi.com/default.asp?255365#1344646

using case id doesn't make sense when creating a case. so always using the row seems like a reasonable user experience

buddy @orangejenny interrupt @benrudolph


benrudolph (approved):
cool i like it

PR #16456, opened 2017-06-05
eNikshay reconcilliation task: adherence counts

proteusvacuum:
https://docs.google.com/document/d/1TG9YWSdccgKeKj0mVIAsoq9LthcZfw5_OebSrkYCF3A/edit#heading=h.vqi0rbfe3ifo
🐡 (first 2 commits are a refactor, new functionality in https://github.com/dimagi/commcare-hq/commit/ac033a1d842d27e335a99bd166ad809d2608d46e)

@benrudolph @NoahCarnahan


proteusvacuum:
✅

custom/enikshay/tests/test_adherence_updates.py
```diff
@@ -671,3 +671,53 @@ def test_update_by_person(self):
             {key: episode.dynamic_case_properties()[key] for key in expected_update},
             {key: str(val) for key, val in expected_update.iteritems()}  # convert values to strings
         )
+
+    def test_adherence_score_start_date_month(self):
+        # If the start date is more than a month ago, calculate the last month's scores
+        self.case_updater.date_today_in_india = datetime.date(2016, 1, 31)
+        self.assert_update(
```
NoahCarnahan: I know you're not introducing the function here, but these tests would be way more readable if `assert_update()` used kwargs instead of this massive nested tuple. e.g.
```python
def assert_update(
    self,
    purge_date=None, 
    adherence_schedule_date_start=None,
    adherence_schedule_id=None
    adherence_cases=None,
)
```

custom/enikshay/tasks.py
```diff
@@ -308,19 +309,54 @@ def update_json(self):
...
             return None
 
+    def get_adherence_scores(self, doses_taken_by_date):
+        today = self.case_updater.date_today_in_india
+        start_date = self.get_adherence_schedule_start_date()
+        one_week_score_count_taken = self.count_doses_taken(
+            doses_taken_by_date,
+            start_date=today - datetime.timedelta(days=7),
+            end_date=today,
+        ) if today - datetime.timedelta(days=7) >= start_date else 0
```
NoahCarnahan: I think the `x = ... if ... else ...` construct is kinda hard to read when the clauses get long, breaking it up like this instead might be more readable:
```python
one_week_ago = today - datetime.timedelta(days=7)
if one_week_ago >= start_date:
    one_week_score_count_taken = self.count_doses_taken(
        doses_taken_by_date,
        start_date=one_week_ago,
        end_date=today,
    )
else:
    one_week_score_count_taken = 0
```

NoahCarnahan (approved):
Looks good. I've got two little style comments, but I don't think they are blocking so feel free to merge now if you're in a hurry.

custom/enikshay/tests/test_adherence_updates.py
proteusvacuum: I totally agree, but didn't have time to update all the other test cases. It was very confusing to write them.

custom/enikshay/tasks.py
proteusvacuum: good point. will update.

PR #16461, opened 2017-06-05
use integer instead of string

emord:
https://sentry.io/dimagi/commcarehq/issues/287633022/

I manually fixed the supervisor config to get around this

@snopoke @benrudolph @orangejenny


benrudolph (approved):
ah annoying

snopoke:
Updated this to make all values string (the JS doesn't handle 0 well). Also added a 'null' default.

emord:
@snopoke was it some false-y check that was failing in JS? I didn't notice anything, but I only really checked that it would update properly

snopoke:
yea, it also only failed after I added a new default option.

Simon Kelly
Director of Server Engineer | Dimagi

PR #16462, opened 2017-06-05
support new lines in comment of form designer edit name modal

biyeun:
@orangejenny @millerdev

### DIFF
--- a/corehq/apps/app_manager/static/app_manager/js/form_designer.js
+++ b/corehq/apps/app_manager/static/app_manager/js/form_designer.js
@@ -147,7 +147,7 @@ hqDefine("app_manager/js/form_designer.js", function() {
                 reverse("edit_form_attr", "name")
             );
             editDetails.initComment(
-                initial_page_data("form_comment"),
+                initial_page_data("form_comment").replace(/(?:\\[n])+/g, "\n"),
                 reverse("edit_form_attr", "comment")
             );
             editDetails.setUpdateCallbackFn(function (name) {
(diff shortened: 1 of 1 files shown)


corehq/apps/app_manager/static/app_manager/js/form_designer.js
```diff
@@ -147,7 +147,7 @@ hqDefine("app_manager/js/form_designer.js", function() {
                 reverse("edit_form_attr", "name")
             );
             editDetails.initComment(
-                initial_page_data("form_comment"),
+                initial_page_data("form_comment").replace(/(?:\\[n])+/g, "\n"),
```
orangejenny: Doesn't this collapse multiple newlines into one? I'd think we'd want to preserve however many newlines were in the original comment.

biyeun: It shouldn't. It just replaces `\\n` with `\n`

millerdev: @biyeun that doesn't seem right. The `+` at the end of the regex will cause it to match one or more contiguous `\\n`, and the entire matched sequence will be replaced with a single `\n`. I think it will do what you want if you remove the `+`, and you can also remove the non-grouping parens since they are not needed. In case that wasn't entirely clear, I think it should be changed to this:
```js
.replace(/\\n/g, "\n")
```
Aside: I think you know this, but it will replace `\n` (that is, a single literal slash followed by the character `n`) with a newline character.

biyeun: thanks! responded

millerdev (approved):

PR #16467, opened 2017-06-05
Upgrade dropbox

benrudolph:
horrible experience, but i think this should get us most of the way there until they decide to completely rewrite their api again. still need to look at the oauth flow, but that should be much more straight forward. i removed the retry functionality because i thought it was fairly pointless and adding complexity

https://manage.dimagi.com/default.asp?254739

interrupt: @emord buddy: @mkangia

### DIFF
--- a/corehq/apps/dropbox/tasks.py
+++ b/corehq/apps/dropbox/tasks.py
@@ -21,7 +21,6 @@ def upload(dropbox_helper_id, access_token, size, max_retries):
     from .models import DropboxUploadHelper
     helper = DropboxUploadHelper.objects.get(id=dropbox_helper_id)
     dbx = Dropbox(access_token)
-    retries = 0
 
     try:
         with open(helper.src, 'rb') as f:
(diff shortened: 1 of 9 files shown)


corehq/apps/dropbox/tasks.py
```diff
@@ -10,43 +13,68 @@
...
+                    UploadSessionCursor(
+                        upload_session.session_id,
+                        offset,
+                    ),
+                )
+                offset += len(chunk)
 
-            upload = uploader.finish(helper.dest)
+            file_metadata = dbx.files_upload_session_finish(
+                b'',
```
mkangia: why is this needed? an empty string?

mkangia (approved):
Looks good. 
Also, we have a dropbox integration!! Didn't know.

corehq/apps/dropbox/tasks.py
benrudolph: in order to finish the upload you must pass some bytes in i believe. honestly though the documentation is pretty bad and the way i figured out how this worked was looking at the source code, so there could be a better way. i was able to sync a ccz to my dropbox, download it and unzip the ccz and it seemed fine

mkangia: Cool. Thanks for the details. 👍

emord (approved):

benrudolph:
ok added one more commit that should take care of the oauth flow. also more gracefully handles when a token becomes invalid

emord (approved):
will this affect users that have already synced with dropbox? like force them to resync or anything

benrudolph:
no i don't believe so (the only time it made me reset my token was when i deleted the app from my dropbox account)

benrudolph:
✅

PR #16469, opened 2017-06-06
Fix stickler.yml file.

markstory:
YAML parsers assume strings starting with `*` are aliases, which these are not.


emord:
Thanks @markstory !

PR #16471, opened 2017-06-06
Changed calculation for referring facility id

kkrampa:
@proteusvacuum @NoahCarnahan


kkrampa:
Needs test data source rebuild

PR #16473, opened 2017-06-06
Update scaling pillows doc

emord:
@dimagi/scale-team 

The management command had a majority of the docs on what to run already. adding some thought process  docs to it

I imagine we'll want 8 partitions on form-sql and case-sql for icds with same number of processes for each pillow that's behind

buddy @orangejenny

### DIFF
--- a/corehq/ex-submodules/pillowtop/management/commands/add_kafka_partition.py
+++ b/corehq/ex-submodules/pillowtop/management/commands/add_kafka_partition.py
@@ -29,7 +29,7 @@ class Command(BaseCommand):
         )
         added_partition = raw_input("have you run {} ? [y/n]".format(kafka_command))
         if added_partition not in ['y', 'yes']:
-            print("then run it")
+            print("then run it on the kafka machine")
 
         for checkpoint in DjangoPillowCheckpoint.objects.filter(sequence_format='json'):
             try:
--- a/docs/pillows.rst
+++ b/docs/pillows.rst
@@ -92,12 +92,24 @@ profile the expensive data sources.
 Parallel Processors
 ~~~~~~~~~~~~~~~~~~~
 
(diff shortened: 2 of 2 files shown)


snopoke:
killing build

corehq/ex-submodules/pillowtop/management/commands/add_kafka_partition.py
```diff
@@ -29,7 +29,7 @@ def handle(self, topic, num_partitions, **options):
         )
         added_partition = raw_input("have you run {} ? [y/n]".format(kafka_command))
         if added_partition not in ['y', 'yes']:
-            print("then run it")
+            print("then run it on the kafka machine")
 
         for checkpoint in DjangoPillowCheckpoint.objects.filter(sequence_format='json'):
```
snopoke: Doesn't this command need to change to use the `KafkaCheckpoint`  model?

emord: yes. this will need to change i'll do that in a followup pr.

It shouldn't block running this command right now. I think the only error will be if we add a new kafka pillow (which we haven't so far)

snopoke: Perhaps I'm confused about what this does but I thought the kafka pillows don't use the `DjangoPillowCheckpoint` at all anymore so the offset values in any `DjangoPillowCheckpoints` would be old. Wouldn't that result in a rewind?

emord: https://github.com/dimagi/commcare-hq/blob/b0b2b8d53ae15caa1a74544923452550ffc9d566/corehq/ex-submodules/pillowtop/management/commands/add_kafka_partition.py#L61-L62 will protect against a rewind.

PR #16475, opened 2017-06-06
Revert "Revert "Minor reminder rule updates for existing framework""

gcapalbo:
Reverts dimagi/commcare-hq#16474


gcapalbo:
(this was already merged, self-merging as tests now pass)

PR #16476, opened 2017-06-06
lazy load project in tabs

emord:
lazy load project if it doesn't get passed in

http://manage.dimagi.com/default.asp?255694#1345269

@benrudolph @orangejenny

### DIFF
--- a/corehq/tabs/uitab.py
+++ b/corehq/tabs/uitab.py
@@ -67,7 +67,7 @@ class UITab(object):
 
     @property
     def project(self):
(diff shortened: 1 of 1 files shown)


benrudolph (approved):

emord:
Doesn't solve the root problem of why this doesn't have a project specified. very strange, but I think this is an ok fix for now

benrudolph:
i think i've seen this before. it's fairly weird. i saw it for some users when accessing a certain hidden url on formplayer

PR #16477, opened 2017-06-06
Ability to use case list tile persisted from another module

mkangia:
[trello card](https://trello.com/c/wekRPsTb/137-app-builder-changes-for-enikshay-phase-2b?menu=filter)

This adds ability to use a different module for
1. case tiles in case list
2. case tiles persisted over the forms

🏛


mkangia:
Going to extract the changes for just the persistent case tile configuration since thats what is needed on priority, Will add for case tile in case list later

PR #16483, opened 2017-06-06
check shadow form string

emord:
http://manage.dimagi.com/default.asp?254122

this is gross and should really use a constant or something, but if you look around appmanager lots of places just use a string 'shadow_form' or 'advanced_form'

You can't use `ShadowForm.form_type` because when loading a couch doc that is interpreted as a stringproperty

@benrudolph @NoahCarnahan @orangejenny


benrudolph (approved):
👍 nice find

NoahCarnahan:
oof, thanks

PR #16490, opened 2017-06-06
Revert "Externalize JavaScript: form_view.html"

orangejenny:
Reverts dimagi/commcare-hq#16482 which breaks on display conditions that have newlines.

@proteusvacuum


orangejenny:
don't wait on tests, going to deploy this now

PR #16496, opened 2017-06-07
clone only location types

calellowitz:
@snopoke cc: @kaapstorm 
For the icds ucr work

### DIFF
--- a/corehq/apps/hqadmin/management/commands/clone_domain.py
+++ b/corehq/apps/hqadmin/management/commands/clone_domain.py
@@ -73,7 +73,8 @@ class Command(BaseCommand):
             self.copy_fixtures()
 
         if self._clone_type(options, 'locations'):
-            self.copy_locations()
+            types_only = self._clone_type(options, 'types_only')
+            self.copy_locations(types_only)
 
         if self._clone_type(options, 'products'):
             self.copy_products()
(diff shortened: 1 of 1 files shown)


corehq/apps/hqadmin/management/commands/clone_domain.py
```diff
@@ -73,7 +73,8 @@ def handle(self, existing_domain, new_domain, **options):
             self.copy_fixtures()
 
         if self._clone_type(options, 'locations'):
-            self.copy_locations()
+            types_only = self._clone_type(options, 'types_only')
```
snopoke: this option breaks the patter of include / exclude options. The alternative of splitting locations and location types completely is more complex though since there is a dependency between the two.

I'm torn between the two options but I think let's stick with this to keep it simple.

snopoke (approved):

corehq/apps/hqadmin/management/commands/clone_domain.py
calellowitz: Yeah, i thought about breaking it into two functions where the location types one returned the mapping that we could pass to the second, but that required sharing state between two functions, which we dont do now and would create other restrictions on how you could pass in arguments that we would need to guard against

kaapstorm: Fair enough.

kaapstorm (approved):

PR #16497, opened 2017-06-07
convert case to dict

emord:
http://manage.dimagi.com/default.asp?255694#1345943

Likely broken for a while http://manage.dimagi.com/default.asp?255694#1345943

from https://github.com/dimagi/commcare-hq/blob/master/corehq/apps/reports/standard/cases/data_sources.py#L21 should have been a dictionary

@benrudolph @orangejenny


benrudolph (approved):
crazy

PR #16500, opened 2017-06-07
Store success_response on repeat attempts

esoergel:
@dannyroberts  @nickpell

@proteusvacuum  FYI

### DIFF
--- a/corehq/apps/repeaters/models.py
+++ b/corehq/apps/repeaters/models.py
@@ -419,6 +419,7 @@ class RepeatRecordAttempt(DocumentSchema):
     cancelled = BooleanProperty(default=False)
     datetime = DateTimeProperty()
     failure_reason = StringProperty()
+    success_response = StringProperty()
     next_check = DateTimeProperty()
     succeeded = BooleanProperty(default=False)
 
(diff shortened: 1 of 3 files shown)


corehq/apps/repeaters/models.py
```diff
@@ -419,6 +419,7 @@ class RepeatRecordAttempt(DocumentSchema):
     cancelled = BooleanProperty(default=False)
     datetime = DateTimeProperty()
     failure_reason = StringProperty()
+    success_response = StringProperty()
```
proteusvacuum: would it make more sense to just have a generic `response` property instead? I guess then you'd have to write a migration...

esoergel: Yeah, I agree that'd be a better state, but it didn't seem super straightforward to implement.  `failure_reason` can also be a string version of an exception that happens, and it's stored on the repeater class in addition to the attempts.  Its presence is also used as a boolean in some cases (`if repeater.failure_reason:`), so that would also require a bit of a refactor.

proteusvacuum: oh yeah. it also gets used in the couch view to output the failed state...

PR #16502, opened 2017-06-07
Conditional queue

benrudolph:
@gcapalbo this overrides the export queue to use the `saved_exports_queue` when processing them daily (i believe this is how it should've always been). i left the task decorator to specify `background_queue` so that when a user clicks "Update Data" it doesn't get thrown in the same queue.

reasoning here: https://manage.dimagi.com/default.asp?255560

buddy: @mkangia

### DIFF
--- a/corehq/apps/reports/tasks.py
+++ b/corehq/apps/reports/tasks.py
@@ -18,6 +18,7 @@ from celery.utils.log import get_task_logger
 
 from casexml.apps.case.xform import extract_case_blocks
 from corehq.apps.export.dbaccessors import get_all_daily_saved_export_instance_ids
+from corehq.apps.export.const import SAVED_EXPORTS_QUEUE
 from corehq.dbaccessors.couchapps.all_docs import get_doc_ids_by_class
 from corehq.form_processor.interfaces.dbaccessors import FormAccessors
 from corehq.util.dates import iso_string_to_datetime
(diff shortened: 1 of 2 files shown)


corehq/apps/reports/tasks.py
```diff
@@ -135,7 +136,15 @@ def saved_exports():
...
         last_access_cutoff = datetime.utcnow() - timedelta(days=settings.SAVED_EXPORT_ACCESS_CUTOFF)
-        rebuild_export_task.delay(daily_saved_export_id, last_access_cutoff)
+        rebuild_export_task.apply_async(
+            args=[
+                daily_saved_export_id, last_access_cutoff
+            ],
+            # Normally the rebuild_export_task uses the background queue,
+            # however we want to override it to use its own queue so that it does
+            # not disrupt other actions.
+            queue=SAVED_EXPORTS_QUEUE,
```
benrudolph: @snopoke there's another options here that specifies priority as an integer from 0 to 9. i thought it'd be interesting to have a mapping that takes the domain's plan level and assigns a priority

snopoke: interesting, not very clear from the docs exactly how that works

snopoke (approved):

corehq/apps/reports/tasks.py
benrudolph: i believe rabbit mq handles the prioritization: https://www.rabbitmq.com/priority.html

gcapalbo:
:+1:

snopoke:
Yes I read that but it's still not exact clear how that works.

PR #16505, opened 2017-06-07
Ignore row limit

benrudolph:
allows for ignoring export limit...

### DIFF
--- a/corehq/apps/export/views.py
+++ b/corehq/apps/export/views.py
@@ -10,7 +10,7 @@ from django.template.defaultfilters import filesizeformat
 from django.views.decorators.csrf import csrf_exempt
 from django.views.decorators.http import require_GET
 
-from corehq.toggles import MESSAGE_LOG_METADATA, PAGINATED_EXPORTS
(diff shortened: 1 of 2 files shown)


benrudolph:
https://manage.dimagi.com/default.asp?249411

emord (commented):
not really excited about htis. I don't think we should allow users to complain so they take more server time...

should probably wait and see if we can push them to commcare-export or a better way to do this

benrudolph:
totally agree. unfortunately this client is a fairly big one, so we may be between a rock and hard place. talbot is going to continue to press them on using daily saved exports

mkangia (approved):
Looks good. Just one small suggestion

corehq/apps/export/views.py
```diff
@@ -2212,6 +2212,9 @@ def _check_deid_permissions(self, export_instances):
 
     def _check_export_size(self, export_instances, filters):
         count = 0
+        if IGNORE_ROW_LIMIT.enabled(self.domain):
+            return
```
benrudolph: so if it every gets called in another place it will always return correctly

mkangia: Seems like its called from that method only, as of now.

benrudolph:
laid down the law

PR #16508, opened 2017-06-07
Trigger case claim submission on clicking Enter

wpride:


benrudolph:
the way to do this is by setting up the html to be a form and the submit button be of input type submit. the browser will automatically handle the enter button

wpride:
@benrudolph thanks bud!

benrudolph (approved):
😎

PR #16509, opened 2017-06-07
Make case lists displayed in details unclickable

wpride:
for https://manage.dimagi.com/default.asp?252775#1346183

Case lists in details shouldn't be clickable

### DIFF
--- a/corehq/apps/cloudcare/static/cloudcare/js/formplayer/menus/views.js
+++ b/corehq/apps/cloudcare/static/cloudcare/js/formplayer/menus/views.js
@@ -206,9 +206,20 @@ FormplayerFrontend.module("Menus.Views", function (Views, FormplayerFrontend, Ba
         return view;
     };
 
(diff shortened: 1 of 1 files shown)


corehq/apps/cloudcare/static/cloudcare/js/formplayer/menus/views.js
```diff
@@ -232,6 +222,17 @@ FormplayerFrontend.module("Menus.Views", function (Views, FormplayerFrontend, Ba
         },
     });
 
+    Views.CaseView = Views.CaseViewUnclickable.extend({
```
benrudolph: feels weird that CaseView inherits from CaseViewUnclickable, can it be the other way around?

wpride: Made the switch - not sure if there's a more elegant way to override parent functions with no-ops

benrudolph (approved):
cool looks good 👍

PR #16510, opened 2017-06-07
Some debugger refactoring

benrudolph:
@wpride @proteusvacuum this rips the debugger out into its own module and sets it up to be easily used in the menu sessions. there's no change in functionality. i refactored a bit, but there's definitely way more we could do to clean this up

cc: @mkangia 
definitely best read commit by commit 🐟


benrudolph:
this is fairly large, but i think it's a good refactor and the way forward to getting menu debugger. happy to offline about it

benrudolph:
can i get a review on this? will walk through with anyone

corehq/apps/cloudcare/static/cloudcare/js/debugger/debugger.js
```diff
@@ -0,0 +1,407 @@
...
+    /**
+     * These define tabs that are availabe in the debugger.
+     * {
+     *   id: <id of the tab HTML element>,
+     *   tab: <link to corresponding tab content. sets the href attribute>
+     *   tabTemplate: <id of the tab's template>
+     *   label: <label of the tab being displayed>
+     * }
+     */
+    var DebuggerTabs = {
```
orangejenny: Could be more succinct if you took for granted that, say, the tab is always `id + '-tab'`, template is always `id + '-template'`, etc., though it would also be a little more magical.

corehq/apps/cloudcare/static/cloudcare/js/debugger/debugger.js
```diff
@@ -0,0 +1,407 @@
...
+                },
+                insertTpl: '${value}',
+                callbacks: {
+                    matcher: self.matcher,
+                },
+            });
+        };
+    };
+
+    var getIconFromType = function(type) {
```
orangejenny: Stylistic, but this would be a lot more concise if you just used an object like `{'Trigger': 'fcc fcc-fd-text', ...}`

orangejenny (approved):
looks good!

benrudolph:
thanks jenny, i'm going to be taking another pass at this when i add the menu debugger route and will address those comments there. mostly those things were already existing, but i agree with all of them!

PR #16512, opened 2017-06-07
Revert "update loading indicator due to performance issues"

orangejenny:
Reverts dimagi/commcare-hq#16492

This breaks modals in v2 (see https://manage.dimagi.com/default.asp?255837) because of the `z-index: 2`, but I figure it's there for a reason so I'm reverting the full PR rather than deleting that one line.

The new image also doesn't always appear vertically centered in the page, it jumps around.

code buddy @emord 
fyi @biyeun


biyeun:
@orangejenny 
thanks for taking care of this. apologies for this slipping through.

The vertically centered thing is a bit annoying then...the loading indicator is dependent on knowing the height of its parent. And it seems the js is being served to the client too slowly on page load so the width and height changes are very noticeable. All the options besides the gif background image approach are going to struggle with this. This is why I went with the gif background in the first place, as it was the most reliable with centering. The way around this is to always have it be a fixed distance from the top based on the `vh` unit, but it won't be perfectly vertically centered. That's the tradeoff.

orangejenny:
Hmm, that sounds aggravating (the centering). Fixed distance from the top sounds like it might be the best compromise...I've seen the image jump around, and also sometimes be visible after the page has loaded, in the dark blue space above form builder but below the top nav, and both of those are pretty weird-looking.

PR #16514, opened 2017-06-08
ICDS: Ability to auto-escalate tech issue cases

gcapalbo:
https://trello.com/c/uRbaeC3v/160-automatic-case-update-and-case-creation-for-issue-tracker

Creates a custom action to be integrated with the rule running framework for automatically escalating tech issue cases.

@sravfeyn @snopoke


custom/icds/rules/custom_actions.py
```diff
@@ -0,0 +1,114 @@
...
+            not subcase.closed
+        ):
+            return subcase
+
+    return None
+
+
+def escalate_tech_issue(case, rule):
+    if case.type != 'tech_issue':
+        return CaseRuleActionResult()
```
snopoke: exception messages could be helpful if these ever get raised

custom/icds/tests/test_custom_rules.py
```diff
@@ -0,0 +1,128 @@
...
+            self.assertEqual(len(subcases), 1)
+            [tech_issue_delegate] = subcases
+
+            self.assertEqual(tech_issue_delegate.type, 'tech_issue_delegate')
+            self.assertEqual(tech_issue_delegate.name, tech_issue.name)
+            self.assertEqual(tech_issue_delegate.owner_id,
+                tech_issue.get_case_property('%s_location_id' % to_level))
+            self.assertEqual(tech_issue_delegate.get_case_property('change_in_level'), '1')
+
+    @run_with_all_backends
```
snopoke: I don't think it's necessary to run these with against couch backend since ICDS is SQL. Probably better to mark the test class with `@use_sql_backend`

custom/icds/rules/custom_actions.py
gcapalbo: this isn't raising an exception - all we're doing is just returning an empty result to the calling process so that it knows no updates were made. this is a scenario that shouldn't happen but we want to make sure we're only updating tech_issue cases, and if not we just do nothing

custom/icds/tests/test_custom_rules.py
gcapalbo: good point, updated

custom/icds/rules/custom_actions.py
snopoke: my bad - didn't read it properly

snopoke (approved):

PR #16515, opened 2017-06-08
Fixed patient type property

kkrampa:
* for enikshay app use patient_type_choice
* for enikshay private app use patient_type


kkrampa:
@NoahCarnahan It turned out two apps use different names for the same property.
We need to rebuild episode data source again.

kkrampa:
@proteusvacuum

PR #16516, opened 2017-06-08
move ICDS task by 12 hours

sravfeyn:
@snopoke @sshah @emord

Today we found that this task is causing lot of processes to be sprung up on postgres, which in turn makes Tableau queries slow, resulting in timeout errors for Tableau dashboard users. I am not sure how much time the task is taking now, and why has this become a problem now given that it is running since a while. May be it's to do with increasing data quantity? Any case, moving this task to IST evening time makes sense. If it doesn't improve things we can review the task further for performance.

### DIFF
--- a/custom/icds_reports/tasks.py
+++ b/custom/icds_reports/tasks.py
@@ -10,7 +10,7 @@ from django.db import connections
 celery_task_logger = logging.getLogger('celery.task')
 
 
-@periodic_task(run_every=crontab(minute=0, hour=0), acks_late=True, queue='background_queue')
+@periodic_task(run_every=crontab(minute=0, hour=12), acks_late=True, queue='background_queue')
 def move_ucr_data_into_aggregation_tables():
 
     if hasattr(settings, "ICDS_UCR_DATABASE_ALIAS") and settings.ICDS_UCR_DATABASE_ALIAS:


custom/icds_reports/tasks.py
```diff
@@ -10,7 +10,7 @@
 celery_task_logger = logging.getLogger('celery.task')
 
 
-@periodic_task(run_every=crontab(minute=0, hour=0), acks_late=True, queue='background_queue')
+@periodic_task(run_every=crontab(minute=0, hour=12), acks_late=True, queue='background_queue')
```
sravfeyn: This is 5:30pm IST

sravfeyn: Is this intentional or a bug with stickler?

snopoke: The version of master you branched off didn't have the stickler config in it. You should always try branch off latest master.

snopoke (approved):
I'm OK with this but also wondering if we shouldn't rather just make celery run on IST instead of UTC: http://docs.celeryproject.org/en/latest/userguide/configuration.html#std:setting-timezone

sravfeyn:
I thought there was a reason to make celery run specifically on UTC (the server is also UTC), no?

snopoke:
The servers are on UTC but I think it makes sense to run celery on IST since most of our schedules are setup to run things at times like midnight with the expectation that it's a quieter period

emord:
@snopoke I think for things like the tasks in the custom icds directories we should look into a way to specify the timezone in the periodic task. does that sound more preferable?

snopoke:
We could do that but I don't see a down side to moving all tasks to run on IST. The only reason they run on UTC now is that it's the default and because it makes logical sense for prod because there isn't a single timezone that is 'off peak'. I did a quick audit of periodic tasks (excluding custom ones that aren't ICDS): https://docs.google.com/spreadsheets/d/11yOwU10q2RCYcvH5r5pHCuy6T7F0HFi0U_CfqeI3VhU/edit#gid=0

PR #16517, opened 2017-06-08
add ability to pick persisted case tile from another module

mkangia:
extracted from #16477 
[trello card](https://trello.com/c/wekRPsTb/137-app-builder-changes-for-enikshay-phase-2b?menu=filter)

This adds ability to use a different module for
1. case tiles in case list
2. case tiles persisted over the forms

🏛

### DIFF
--- a/corehq/apps/app_manager/templates/app_manager/v1/partials/case_list.html
+++ b/corehq/apps/app_manager/templates/app_manager/v1/partials/case_list.html
@@ -58,7 +58,7 @@
     </div>
     {% endif %}
 
-    <div data-bind="visible: (useCaseTiles() == 'yes' && COMMCAREHQ.toggleEnabled('CASE_LIST_TILE'))|| COMMCAREHQ.toggleEnabled('CASE_LIST_CUSTOM_XML')">
+    <div data-bind="visible: useCaseTiles() == 'yes' || COMMCAREHQ.toggleEnabled('CASE_LIST_CUSTOM_XML')">
         <div class="checkbox">
             <label>
                 <input type="checkbox" data-bind="checked: persistTileOnForms">
--- a/corehq/apps/app_manager/templates/app_manager/v1/partials/case_list.html
+++ b/corehq/apps/app_manager/templates/app_manager/v1/partials/case_list.html
@@ -76,7 +76,6 @@
     </div>
     {% endif %}
 
-    {% if request|feature_preview_enabled:"SHARE_CASE_TILE_CONTEXT" %}
     {{ request|toggle_tag_info:"CASE_LIST_TILE" }}
     {% if request|toggle_enabled:'CASE_LIST_TILE' %}
     {{ request|toggle_tag_info:"SHOW_PERSIST_CASE_CONTEXT_SETTING" }}
@@ -100,7 +99,6 @@
         </div>
         {% endif %}
         {% endwith %}
-    {% endif %}
     {% endif %}
     {% endif %}
     {% include 'app_manager/v1/partials/case_list_properties.html' %}
--- a/corehq/apps/app_manager/templates/app_manager/v1/partials/case_list.html
+++ b/corehq/apps/app_manager/templates/app_manager/v1/partials/case_list.html
@@ -80,10 +80,10 @@
         <div class="row">
             <div class="col-sm-6">
                 <label>
-                    {% trans "Use persistent case list tile from module" %}
+                    {% trans "Use persistent case tile from module" %}
                 </label>
                 <div>
-                    <select name='persistent_case_context_from_module' class='form-control' data-bind="value: persistentCaseContextFromModule">
+                    <select name='persistent_case_tile_from_module' class='form-control' data-bind="value: persistentCaseTileFromModule">
                         <option value></option>
                         {% for other_mod in available_modules %}
                             <option value={{ other_mod.unique_id }}>{{ other_mod.name|html_trans:langs }}</option>
--- a/corehq/apps/app_manager/templates/app_manager/v1/partials/case_list.html
+++ b/corehq/apps/app_manager/templates/app_manager/v1/partials/case_list.html
@@ -58,9 +58,7 @@
     </div>
     {% endif %}
 
-    {{ request|toggle_tag_info:"CASE_LIST_TILE" }}
-    {% if request|toggle_enabled:'CASE_LIST_TILE' %}
-    <div data-bind="visible: useCaseTiles() == 'yes' || COMMCAREHQ.toggleEnabled('CASE_LIST_CUSTOM_XML')">
+    <div data-bind="visible: (useCaseTiles() == 'yes' && COMMCAREHQ.toggleEnabled('CASE_LIST_TILE'))|| COMMCAREHQ.toggleEnabled('CASE_LIST_CUSTOM_XML')">
         <div class="checkbox">
             <label>
                 <input type="checkbox" data-bind="checked: persistTileOnForms">
@@ -74,7 +72,6 @@
             </label>
         </div>
     </div>
-    {% endif %}
 
     {{ request|toggle_tag_info:"CASE_LIST_TILE" }}
     {% if request|toggle_enabled:'CASE_LIST_TILE' %}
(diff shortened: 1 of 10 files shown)


corehq/apps/app_manager/templates/app_manager/v1/partials/case_list.html
```diff
@@ -56,6 +58,8 @@
     </div>
     {% endif %}
 
+    {{ request|toggle_tag_info:"CASE_LIST_TILE" }}
+    {% if request|toggle_enabled:'CASE_LIST_TILE' %}
```
snopoke: how come this is going behind a toggle now? Seems like it's already behind the `CASE_LIST_CUSTOM_XML` toggle.

corehq/feature_previews.py
```diff
@@ -127,3 +127,12 @@ def enable_callcenter(domain_name, checked):
     ),
     privilege=LOOKUP_TABLES,
 )
+
+
+SHARE_CASE_TILE_CONTEXT = FeaturePreview(
```
snopoke: Don't think this should be a feature preview since there are other toggles that are required to make this work.

snopoke: In fact since this is basically an extension to the case tiles feature I don't think it needs a separate toggle at all and can just piggy back on the `CASE_LIST_TILE` toggle.

corehq/apps/app_manager/templatetags/app_manager_extras.py
```diff
@@ -16,3 +16,9 @@ def get_available_modules_for_case_list_configuration(app, module):
             and m.module_type not in disallowed_module_types
             and m.case_type == module.case_type)
     ]
+
+
+@register.filter
+def get_available_modules_for_case_tile_configuration(app, module):
```
snopoke: rename `module` to `exclude_module`

corehq/apps/app_manager/templates/app_manager/v2/partials/case_list.html
```diff
@@ -76,6 +79,34 @@ <h4 class="panel-title panel-title-nolink">
                 </label>
             </div>
         </div>
+        {% endif %}
+
+        {% if request|feature_preview_enabled:"SHARE_CASE_TILE_CONTEXT" %}
+        {{ request|toggle_tag_info:"CASE_LIST_TILE" }}
+        {% if request|toggle_enabled:'CASE_LIST_TILE' %}
+        {% if request|toggle_enabled:'SHOW_PERSIST_CASE_CONTEXT_SETTING' %}
```
snopoke: having something under 2 toggles seems pretty weird. Can we not clean up the toggles so that there's only 1?

snopoke (changes requested):

corehq/apps/app_manager/templates/app_manager/v1/partials/case_list.html
mkangia: the `CASE_LIST_TILE` toggle allows user to set `useCaseTiles` property. If someone sets that to 'yes' and then the toggle is removed for the domain, the container below is still visible since `useCaseTiles` was set to `yes`.

Oh! i see. this now makes it unavailable in case just `CASE_LIST_CUSTOM_XML` toggle is enabled. Would need to add check for that toggle above instead of `data-bind` attr.

corehq/apps/app_manager/templatetags/app_manager_extras.py
mkangia: using `for_module` sounds better?

snopoke: if you read the function as a sentence "get available modules for case tile configuration [for] app [but exclude this] module"

I don't know not really that important

corehq/feature_previews.py
mkangia: i am not completely sure about how feature previews are related to toggles, but it seems alright to move this under `SHOW_PERSIST_CASE_CONTEXT_SETTING` since we are talking about persistent case tiles.

corehq/apps/app_manager/templates/app_manager/v2/partials/case_list.html
mkangia: i am not sure if that would be good to do. Those two toggles might be used by different set of domains.

Though just the check for `SHOW_PERSIST_CASE_CONTEXT_SETTING` is sufficient but if somebody enables that without enabling `CASE_LIST_TILE`, this might give unexpected results in case the case tiles are configured and then the `CASE_LIST_TILE` was removed.

corehq/feature_previews.py
mkangia: also `SHOW_PERSIST_CASE_CONTEXT_SETTING` seems like being used only by a couple of active real domains and a bunch of test domains so this sounds safe enough to move under `SHOW_PERSIST_CASE_CONTEXT_SETTING` only.

corehq/apps/app_manager/templatetags/app_manager_extras.py
mkangia: i read this as [for] app [for] module. 
The logic is not just about getting all modules except this one.  There is a method above that already does that. This method gives the module that are valid options for the configuration.

snopoke: Yes but the only reason you pass in the current module is so that it doesn't get included in the returned list. Anyway, not worth wasting airtime on :)

benrudolph (approved):
don't have a lot of context on this change, but the code looks good (looks like still one comment pending). toggle soup! 🍚

mkangia:
@snopoke sorry misinterpreted `SHOW_PERSIST_CASE_CONTEXT_SETTING`. Please don't review further yet. Need to fix this.

corehq/apps/app_manager/templates/app_manager/v1/partials/case_list.html
mkangia: Fixed this by moving the toggle check in the data-bind attr.

corehq/feature_previews.py
mkangia: Removed Feature Preview.

corehq/apps/app_manager/templates/app_manager/v2/partials/case_list.html
mkangia: check for `SHOW_PERSIST_CASE_CONTEXT_SETTING` was not needed. I misinterpreted the context of that toggle. Removed it now.

mkangia:
@snopoke fixed now. Good to review.

corehq/apps/app_manager/templates/app_manager/v2/partials/case_list.html
```diff
@@ -62,7 +63,7 @@ <h4 class="panel-title panel-title-nolink">
             </div>
         </div>
         {% endif %}
-        <div data-bind="visible: useCaseTiles() == 'yes' || COMMCAREHQ.toggleEnabled('CASE_LIST_CUSTOM_XML')">
+        <div data-bind="visible: (useCaseTiles() == 'yes' && COMMCAREHQ.toggleEnabled('CASE_LIST_TILE')) || COMMCAREHQ.toggleEnabled('CASE_LIST_CUSTOM_XML')">
```
snopoke: Why is this change necessary? Changing this doesn't seem like part of the feature you're adding and I think the original logic is fine.

snopoke (commented):
Looks good. Just comment on toggle logic.

corehq/apps/app_manager/templates/app_manager/v2/partials/case_list.html
mkangia: Yup! its not a part of the feature that i am adding but an improvement in what is present

the reason
```
the CASE_LIST_TILE toggle allows user to set useCaseTiles property. If someone sets that to 'yes' and then the toggle is removed for the domain, the container below is still visible since useCaseTiles was set to yes.
```

snopoke:
Yes that's correct and I think that's how it should be since it is still
set in the app and hence will still appear on the phone. If we hide it when
the toggle is removed then there is no way to remove it from the app.

Simon Kelly
Director of Server Engineer | Dimagi

mkangia:
@snopoke cool. Removing that change.

snopoke (approved):

PR #16519, opened 2017-06-08
Updated RNTCP PMDT Treatment Register report

kkrampa:
@NoahCarnahan @proteusvacuum


kkrampa:
@NoahCarnahan it doesn't touch any property from live reports but it still demands data source rebuilding. Do you think it's worth to move this into another data source to avoid rebuilding episode data source?

NoahCarnahan:
Let's keep it as one since that would be less work. Although If you find during the development process that it would be more convenient to have two data sources, then feel free to make that change.

PR #16520, opened 2017-06-08
Line breaks come before operators

proteusvacuum:
@esoergel @millerdev

### DIFF
--- a/.stickler.yml
+++ b/.stickler.yml
@@ -8,6 +8,6 @@ files:
 linters:
   flake8:
     max-line-length: 115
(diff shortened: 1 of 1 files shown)


millerdev (approved):

emord:
>  In the default configuration, the checks E121, E123, E126, E133, E226, E241, E242, E704 and W503 are ignored because they are not rules unanimously accepted, and PEP 8 does not enforce them.
http://pep8.readthedocs.io/en/latest/intro.html#error-codes
should we remove these others? it appears stickler doesn't use the default config?

esoergel (approved):
killing build

PR #16525, opened 2017-06-08
paging device ids for infinite scroll without counts

calellowitz:
@kaapstorm cc: @benrudolph @emord @NoahCarnahan 
This should fix the device log filters once and for all. The count call was taking forever since it was counting over subquery that did a `SELECT DISTINCT` on all device ids.


emord (approved):

snopoke (approved):

benrudolph (approved):
rough..

kaapstorm (approved):

PR #16526, opened 2017-06-08
Fix loading indicator so that it doesn't use breaking z-index

biyeun:
@orangejenny // @millerdev 

Tested with the edit name modal in formbuilder too. seemed to work fine.


orangejenny (approved):

orangejenny:
Today I learned about `vh` and `vw`. Nice :)

PR #16534, opened 2017-06-08
fixing month start and end dates in icds ucrs

calellowitz:
@sheelio cc: @snopoke @kaapstorm 
This creates copies of the existing monthly ucr data sources with the new month configuration. It also copies all ucrs using those data sources so there is a copy using the new ones. (This diff is best seen as a diff between original and v2 versions of each of the data sources but I do not know how to show that)


snopoke:
Would it be possible to split this into 2 commits where the first just copies the UCRs and the second makes the changes. Then it would be easy to see the diff.

calellowitz:
@snopoke that didnt help as much as i thought it would but i broke it into a couple commits

emord (approved):

PR #16535, opened 2017-06-08
Add validation conditions to app summary

orangejenny:
https://manage.dimagi.com/default.asp?251826

@emord

### DIFF
--- a/corehq/apps/app_manager/xform.py
+++ b/corehq/apps/app_manager/xform.py
@@ -579,6 +579,8 @@ def validate_xform(domain, source):
         )
 
 
+ControlNode = collections.namedtuple('ControlNode', ['node', 'path', 'repeat', 'group', 'items',
+                                     'is_leaf', 'data_type', 'relevant', 'required', 'constraint'])
 class XForm(WrappedNode):
     """
     A bunch of utility functions for doing certain specific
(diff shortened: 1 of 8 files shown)


corehq/apps/app_manager/xform.py
```diff
@@ -954,7 +954,7 @@ def get_questions(self, langs, include_triggers=False,
         control_nodes = self.get_control_nodes()
         leaf_data_nodes = self.get_leaf_data_nodes()
 
-        for node, path, repeat, group, items, is_leaf, data_type, relevant, required in control_nodes:
+        for node, path, repeat, group, items, is_leaf, data_type, relevant, required, constraint in control_nodes:
```
emord: This is easily the most values i've seen in a for loop. (not that you need to fix that here)

emord (approved):

corehq/apps/app_manager/xform.py
orangejenny: Yeah, it's a bit much. Switched to a namedtuple, which is at least a little bit more readable.

proteusvacuum:
Hey, 
Think it would be worthwhile to add these to the download too?

orangejenny:
@proteusvacuum adding them to `get_questions` adds them to the download, too

proteusvacuum:
@orangejenny oh, cool. I thought you would have needed to define an extra column in the excel sheet, but maybe not?

orangejenny:
haven't looked at the code...maybe the excel sheet config would let me do things like rename the column? but it does work, adds a `relevant` column.

orangejenny:
Offlined with Farid; I was being dumb. Validation conditions should be in download for real now.

PR #16538, opened 2017-06-08
virtual location users should have 'Default Mobile Worker' role

nickpell:
re: https://manage.dimagi.com/default.asp?254690#BugEvent.1339013

@esoergel


custom/enikshay/private_sector_datamigration/management/commands/create_dtos_and_agencies.py
```diff
@@ -82,10 +82,15 @@ def create_user(self, agency_loc, user_level):
...
 
         agency_loc_id = agency_loc.location_id
+        domain = agency_loc.domain
 
         user = make_location_user(agency_loc)
         user.user_location_id = agency_loc_id
         user.set_location(agency_loc, commit=False)
+        user.set_role(
+            domain,
+            UserRole.by_domain_and_name(domain, 'Default Mobile Worker')[0].get_qualified_id()
```
esoergel: Matching on the name of a role is scary, but I don't think there's another way.

esoergel (approved):

PR #16542, opened 2017-06-08
making sure not to miss anyone confirming account

calellowitz:
@kaapstorm 
some users who confirm their accounts are not getting marked as such in hubspot. this is to fix a possible cause, and alert me if it turned out to be the cause


kaapstorm (approved):
Good luck

PR #16545, opened 2017-06-09
Clear Fixture quickcache after upload (and a test with refactor of old tests)

sravfeyn:
https://manage.dimagi.com/default.asp?255251#1347058

### DIFF
--- a/corehq/apps/fixtures/tests/test_upload.py
+++ b/corehq/apps/fixtures/tests/test_upload.py
@@ -175,6 +175,7 @@ class TestFixtureUpload(TestCase):
             return get_workbook(f)
 
     def get_fixture_items(self, attribute):
+        # return list of 'attribute' values of fixture table 'things'
         fixtures = []
         for fixture in FixtureDataItem.get_item_list(self.domain, 'things'):
             fixtures.append(fixture.fields.get(attribute).field_list[0].field_value)
@@ -202,7 +203,8 @@ class TestFixtureUpload(TestCase):
         )
 
         # reupload with additional row
-        new_rows = initial_rows + [(None, 'N', 'orange')]
+        apple_id = FixtureDataItem.get_item_list(self.domain, 'things')[0]._id
+        new_rows = [(apple_id, 'N', 'apple'), (None, 'N', 'orange')]
         workbook = self._get_workbook_from_data(self.headers, self.make_rows(new_rows))
         _run_fixture_upload(self.domain, workbook)
         self.assertItemsEqual(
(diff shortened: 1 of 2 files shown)


sravfeyn:
@gcapalbo @snopoke 1st commit is just refactor.

corehq/apps/fixtures/upload/run_upload.py
```diff
@@ -189,4 +190,22 @@ def _update_progress(table_count, item_count, items_in_table):
                         old_data_item.add_location(location_cache.location,
                                                    transaction=transaction)
 
+    clear_fixture_quickcache(data_types)
     return return_val
+
+
+def clear_fixture_quickcache(data_types):
+    """
+    Clears quickcache for fixtures.dbaccessors
```
sravfeyn: @emord FYI, this wasn't properly busted before

corehq/apps/fixtures/tests/test_upload.py
```diff
@@ -119,13 +128,84 @@ def test_comprehensiveness(self):
...
     def test_indexed_field(self):
+        # test indexed fields are ready correctly
         workbook = self._get_workbook('ok')
         type_sheets = workbook.get_all_type_sheets()
         indexed_field = type_sheets[0].fields[0]
         self.assertEqual(indexed_field.field_name, 'name')
         self.assertTrue(indexed_field.is_indexed)
+
+    def test_row_addition(self):
+        # upload and then reupload with addition of a new fixture-item should work
```
snopoke: comment could be expanded to say what `should work` means. i.e. that fetching the list of fixtures after an upload should include newly uploaded items

sravfeyn: Updated

corehq/apps/fixtures/upload/run_upload.py
```diff
@@ -189,4 +190,22 @@ def _update_progress(table_count, item_count, items_in_table):
...
+        List of FixtureDataType objects in a domain
+    """
+    if not data_types:
+        return
+    type_ids = []
+    for data_type in data_types:
+        data_type.clear_caches()
+        type_ids.append(data_type.get_id)
+    from corehq.apps.fixtures.dbaccessors import get_fixture_items_for_data_types
+    get_fixture_items_for_data_types.clear(data_types[0].domain, type_ids)
```
gcapalbo: it might be good to make domain be an arg to this function, just to make it clear that all the data_types have to belong to the same domain

corehq/apps/fixtures/upload/run_upload.py
emord: Yikes. thanks for catching that

corehq/apps/fixtures/upload/run_upload.py
sravfeyn: I thought about it, but this would really never be called with a mix of domains. Even if it's called, there is no effect in that the cache key won't exist and nothing happens.

PR #16546, opened 2017-06-09
Only send to Indian numbers from vertex

gcapalbo:
@mkangia

### DIFF
--- a/corehq/messaging/smsbackends/vertex/models.py
+++ b/corehq/messaging/smsbackends/vertex/models.py
@@ -65,7 +65,17 @@ class VertexBackend(SQLSMSBackend):
         }
         return params
 
+    def phone_number_is_valid(self, phone_number):
+        phone_number = strip_plus(phone_number)
+        # Phone number must be an Indian phone number
+        # Also avoid processing numbers that are obviously too short
+        return phone_number.startswith('91') and len(phone_number) > 3
+
     def send(self, msg_obj, *args, **kwargs):
+        if not self.phone_number_is_valid(msg_obj.phone_number):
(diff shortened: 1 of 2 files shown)


corehq/messaging/smsbackends/vertex/models.py
```diff
@@ -65,7 +65,17 @@ def populate_params(self, msg_obj):
         }
         return params
 
+    def phone_number_is_valid(self, phone_number):
+        phone_number = strip_plus(phone_number)
+        # Phone number must be an Indian phone number
+        # Also avoid processing numbers that are obviously too short
+        return phone_number.startswith('91') and len(phone_number) > 3
```
mkangia: A valid number would have length 10 exclusive of `91` if we want to make this strict.

mkangia (commented):
Worth adding test for this?

mkangia (approved):

corehq/messaging/smsbackends/vertex/models.py
gcapalbo: i don't think we need to be too strict here since the gateway will also reject invalid destination numbers. what this does is just catch 95% of invalid numbers so that we avoid processing them all together

gcapalbo:
@mkangia added test

corehq/messaging/smsbackends/vertex/models.py
mkangia: the check for `91` seems a good add. I am not sure if length of 4 digits around will happen.. but a length of 9 is possible considering someone mistypes or skips a digit.

PR #16549, opened 2017-06-09
ucr slow log

snopoke:
Trying to get some info about domains that are slow. Curious what you guys think of this approach.

Basically trying to answer the question: "What's causing these UCR pile ups". 
image

@dimagi/scale-team 

buddies @gcapalbo @sravfeyn

### DIFF
--- a/corehq/apps/userreports/pillow.py
+++ b/corehq/apps/userreports/pillow.py
@@ -1,6 +1,6 @@
 from __future__ import absolute_import
 
-from collections import defaultdict
+from collections import defaultdict, Counter
 from datetime import datetime, timedelta
 import hashlib
 
@@ -20,7 +20,9 @@ from corehq.apps.userreports.sql import metadata
 from corehq.apps.userreports.tasks import rebuild_indicators
 from corehq.apps.userreports.util import get_indicator_adapter, get_backend_id
 from corehq.sql_db.connections import connection_manager
+from corehq.util.datadog.gauges import datadog_histogram
 from corehq.util.soft_assert import soft_assert
+from corehq.util.timer import TimingContext
 from fluff.signals import (
     apply_index_changes,
     get_migration_context,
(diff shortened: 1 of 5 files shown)


emord:
was this prod or softlayer? If it's prod, I think we just need to provision a new pillow machine

Did you take a look at the slow logs on the pillow machine? it'll give you an idea of the particular UCRs (but not domain)

emord (approved):

snopoke:
@emord yea this is for prod, agree that we need a new machine but also would be good to know where the load is coming from because there isn't a corresponding trend in forms submissions or case changes which means there must be submissions from some domains with slow UCRs.

corehq/apps/userreports/pillow.py
```diff
@@ -233,19 +237,40 @@ def process_change(self, pillow_instance, change):
...
+
+        self.domain_timing_context.update(**{
+            domain: timer.duration
+        })
+
+    def checkpoint_updated(self):
+        total_duration = sum(self.domain_timing_context.values())
+        duration_seen = 0
+        top_half_domains = {}
+        for domain, duration in self.domain_timing_context.most_common():
```
benrudolph: cool usage

benrudolph (approved):
👍 looks good. i still sort of think we're gonna need to log actual slow docs to get a sense. once we have domains that are slow we're still going to need to look through their data to figure out which docs are causing the slow down. not sure the best strategy for that. (was recently thinking of doing something similar for saved exports)

emord (approved):
> i still sort of think we're gonna need to log actual slow docs

@benrudolph I offlined with @snopoke and I've already implemented that https://github.com/dimagi/commcare-hq/blob/master/corehq/apps/userreports/pillow.py#L53-L65 It logs to the pillow's log file.

We can then use https://github.com/dimagi/commcare-hq/blob/master/corehq/apps/userreports/management/commands/profile_data_source.py to profile with more granularity

PR #16550, opened 2017-06-09
App manager v2 copy change

orangejenny:
As discussed in #product 

@emord / @proteusvacuum


proteusvacuum:
👍 
Don't want to start a discussion about every single thing, but is there a reason this isn't in a `logic` section like in forms? I think having consistency of where I would find similar actions is a good thing.

proteusvacuum (approved):

PR #16552, opened 2017-06-09
BETS repeater: correct case property name

proteusvacuum:
@esoergel @NoahCarnahan


proteusvacuum:
https://sentry.io/dimagi/commcarehq/issues/290637134/

esoergel (approved):

PR #16553, opened 2017-06-09
Fix custom enikshay ucr expressions

NoahCarnahan:
Whoops, looks like `trail` and `referral` cases are related to `person` cases, not `episode`s, so these expressions weren't working.
@proteusvacuum


proteusvacuum:
d'oh.

proteusvacuum (approved):

PR #16556, opened 2017-06-09
Add SMS alert fields

esoergel:
@gcapalbo @nickpell

### DIFF
--- a/custom/enikshay/const.py
+++ b/custom/enikshay/const.py
@@ -105,7 +105,9 @@ AGENCY_USER_FIELDS = [
     ("address_line_1", "Address Line 1", []),
     ("address_line_2", "Address Line 2", []),
     ("pincode", "Pincode", []),
-    ("mobile_no_1", "Mobile No. 1", []),
+    # TODO add format validation (91 + 10 digits, no leading zeros)
+    ("contact_phone_number", "Mobile number for SMS alerts (please only enter 91 and ten digits, not beginning with 0)", []),
+    ("language_code", "Preferred language for SMS alerts", ["en", "hin", "mar", "bho", "guj"]),
     ("mobile_no_2", "Mobile No. 2", []),
     ("landline_no", "Landline No.", []),
     ("email", "Email", []),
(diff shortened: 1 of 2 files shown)


esoergel:
Ended up adding the language code and phone number validation stuff too - it was actually pretty straightforward!

custom/enikshay/user_setup.py
```diff
@@ -318,6 +319,31 @@ def init_form(self, post_dict=None):
...
+                    ('', _('Select one')),
+                    ("en", "English"),
+                    ("hin", "Hindi"),
+                    ("mar", "Marathi"),
+                    ("bho", "Bhojpuri"),
+                    ('guj', "Gujarati"),
+                ],
+            )
+        if field.slug == 'contact_phone_number':
+            regexp = "^91[1-9][0-9]{9}$"
```
gcapalbo: sorry, when i said no leading zero i just meant that we don't want the zero before the 10 digits (so adding a leading zero makes it 11 digits).

i'm not sure if the first digit should actually be restricted to 1-9. to be safe maybe we should just make it be [0-9]{10}

custom/enikshay/const.py
```diff
@@ -105,7 +105,9 @@
     ("address_line_1", "Address Line 1", []),
     ("address_line_2", "Address Line 2", []),
     ("pincode", "Pincode", []),
-    ("mobile_no_1", "Mobile No. 1", []),
+    # TODO add format validation (91 + 10 digits, no leading zeros)
+    ("contact_phone_number", "Mobile number for SMS alerts (please only enter 91 and ten digits, not beginning with 0)", []),
```
gcapalbo: maybe we should say:

Please enter only digits. Enter 91 followed by the 10-digit national number.

custom/enikshay/user_setup.py
esoergel: I checked in the app, and that validation condition did require the leading number to be 1-9

gcapalbo (approved):

PR #16560, opened 2017-06-09
Update device id on correct user during restore

NoahCarnahan:
@proteusvacuum 
FYI @ctsims


NoahCarnahan:
So I didn't write a test for this because it sounded urgent and not sure if I'll be able to have on by the end of the day... do you think this needs one?

benrudolph (approved):
looks good

benrudolph:
i think fine for now as it's just under the nikshay toggle

PR #16562, opened 2017-06-09
Ce/overwrite app

calellowitz:
@snopoke cc: @kaapstorm @sheelio 
New management command for copying the icds apps back after the ucr migration. This can easily be expanded to allow linked apps to work with static ucrs so I am going to work on adding that in after this.

(This assumes all relevant UCRs are static, if that is not the case I figure out the rest)

### DIFF
--- a/corehq/apps/app_manager/management/commands/overwrite_app.py
+++ b/corehq/apps/app_manager/management/commands/overwrite_app.py
@@ -28,6 +28,5 @@ class Command(BaseCommand):
     @property
     def report_map(self):
         if not self._report_map:
-            self._report_map = {}
-            self._report_map = get_static_report_mapping(self.from_domain, self.to_domain, self._report_map)
+            self._report_map = get_static_report_mapping(self.from_domain, self.to_domain, {})
         return self._report_map
--- a/corehq/apps/app_manager/management/commands/overwrite_app.py
+++ b/corehq/apps/app_manager/management/commands/overwrite_app.py
@@ -1,0 +1,47 @@
+from django.core.management.base import BaseCommand
+
+from corehq.apps.app_manager.dbaccessors import get_current_app_doc, get_app, wrap_app
+from corehq.apps.app_manager.models import Application, ReportModule
+from corehq.apps.userreports.util import copy_static_reports
+
+
+class Command(BaseCommand):
+    """
+    Overwrites target application with other application
+    """
+
+    args = "<from_domain> ><from_app_id> <to_domain> <to_app_id>"
+
+    def add_arguments(self, parser):
+        parser.add_argument('from_domain')
+        parser.add_argument('from_app_id')
+        parser.add_argument('to_domain')
+        parser.add_argument('to_app_id')
+
+    def handle(self, from_domain, from_app_id, to_domain, to_app_id, *args, **options):
+        self.from_domain = from_domain
+        self.to_domain = to_domain
+        app = get_current_app_doc(self.to_domain, to_app_id)
+        latest_master_build = get_app(None, from_app_id, latest=True)
+        excluded_fields = set(Application._meta_fields).union(
+            ['date_created', 'build_profiles', 'copy_history', 'copy_of', 'name', 'comment', 'doc_type']
+        )
+        master_json = latest_master_build.to_json()
+        for key, value in master_json.iteritems():
+            if key not in excluded_fields:
+                app[key] = value
+        app['version'] = master_json['version']
+        wrapped_app = wrap_app(app)
+        for module in wrapped_app.modules:
+            if isinstance(module, ReportModule):
+                for config in module.report_configs:
+                    config.report_id = self.report_map[config.report_id]
+        wrapped_app.copy_attachments(latest_master_build)
+        wrapped_app.save(increment_version=False)
+
+    @property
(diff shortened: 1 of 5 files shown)


corehq/apps/userreports/util.py
```diff
@@ -191,3 +188,28 @@ def get_ucr_class_name(id):
 
 def get_async_indicator_modify_lock_key(doc_id):
     return 'async_indicator_save-{}'.format(doc_id)
+
+
+def copy_static_reports(from_domain, to_domain, report_map):
```
snopoke: should just return a new dict instead of modifying one from input

snopoke: Also the name is a bit weird since it's not copying anything. How about `get_static_report_mapping`

corehq/apps/hqadmin/management/commands/clone_domain.py
```diff
@@ -282,15 +265,6 @@ def copy_ucr_datasources(self):
...
             datasource_map[old_id] = new_id
-        for static_datasource in StaticDataSourceConfiguration.by_domain(self.existing_domain):
-            table_id = static_datasource.get_id.replace(
-                StaticDataSourceConfiguration._datasource_id_prefix + self.existing_domain + '-',
-                ''
-            )
-            new_id = StaticDataSourceConfiguration.get_doc_id(self.new_domain, table_id)
-            # check that new datasource is in new domain's list of static datasources
-            StaticDataSourceConfiguration.by_id(new_id)
-            datasource_map[static_datasource.get_id] = new_id
```
snopoke: how come this got removed?

corehq/apps/app_manager/management/commands/overwrite_app.py
```diff
@@ -0,0 +1,47 @@
...
+            if key not in excluded_fields:
+                app[key] = value
+        app['version'] = master_json['version']
+        wrapped_app = wrap_app(app)
+        for module in wrapped_app.modules:
+            if isinstance(module, ReportModule):
+                for config in module.report_configs:
+                    config.report_id = self.report_map[config.report_id]
+        wrapped_app.copy_attachments(latest_master_build)
+        wrapped_app.save(increment_version=False)
```
snopoke: This code is almost identical to what's in the `pull_master_app` view function. I think it would make sense to reuse the code.

snopoke (changes requested):

corehq/apps/hqadmin/management/commands/clone_domain.py
calellowitz: It appeared to do nothing. The `datasource_map` was never used in the static reports part of `copy_ucrs` which is the only time that map is used at all. I had assumed this is because the datasources on the config dont change for static reports. It now looks to me like the domain check is relevant so maybe only the last line should be deleted?

snopoke: You still need it because there could be dynamic reports that reference static data sources.

calellowitz:
@snopoke I think I got to all the feedback (and set the stage for mobile ucr support in linked apps)

corehq/apps/app_manager/management/commands/overwrite_app.py
```diff
@@ -0,0 +1,33 @@
...
+        self.from_domain = from_domain
+        self.to_domain = to_domain
+        app = get_current_app_doc(self.to_domain, to_app_id)
+        latest_master_build = get_app(None, from_app_id, latest=True)
+        overwrite_app(app, latest_master_build, include_ucrs=True, report_map=self.report_map)
+
+    @property
+    def report_map(self):
+        if not self._report_map:
+            self._report_map = {}
```
snopoke: unnecessary

corehq/apps/app_manager/views/apps.py
```diff
@@ -912,13 +898,12 @@ def pull_master_app(request, domain, app_id):
...
             messages.success(request,
                              _('Your linked application was successfully updated to the latest version.'))
-        wrapped_app.copy_attachments(latest_master_build)
-        wrapped_app.save(increment_version=False)
     else:
         messages.error(request, _(
             'This project is not authorized to update from the master application. '
             'Please contact the maintainer of the master app if you believe this is a mistake. ')
         )
+    params = {}
```
snopoke: nit: no need for variable

corehq/apps/app_manager/views/utils.py
```diff
@@ -125,3 +125,28 @@ def get_blank_form_xml(form_name):
...
+    app['version'] = master_json['version']
+    wrapped_app = wrap_app(app)
+    errors = False
+    for module in wrapped_app.modules:
+        if isinstance(module, ReportModule):
+            if include_ucrs and report_map is not None:
+                for config in module.report_configs:
+                    config.report_id = report_map[config.report_id]
+            else:
+                errors = True
```
snopoke: Just a style point here. There are two better options:
1. `return True` which short circuits the rest of the function and means you don't have to check `errors` later
2. Raise an exception. This is usually the best approach when a function wants to fail. It also makes it clear that the function has failed and what the error is.

snopoke (changes requested):

calellowitz:
@snopoke fixed

PR #16566, opened 2017-06-09
Only store strings

proteusvacuum:
@NoahCarnahan @esoergel


proteusvacuum:
tests passed locally, and build is broken, so merging

PR #16571, opened 2017-06-12
don't 500, 404

sravfeyn:
http://manage.dimagi.com/default.asp?255970#edit_0_255970

### DIFF
--- a/custom/enikshay/reports/historical_adherence.py
+++ b/custom/enikshay/reports/historical_adherence.py
@@ -13,6 +13,7 @@ from corehq.apps.reports.datatables import DataTablesHeader
 from corehq.apps.reports.filters.base import BaseReportFilter
 from corehq.apps.reports.filters.dates import DatespanFilter
 from corehq.form_processor.interfaces.dbaccessors import CaseAccessors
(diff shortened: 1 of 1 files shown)


custom/enikshay/reports/historical_adherence.py
```diff
@@ -63,7 +64,10 @@ class HistoricalAdherenceReport(EnikshayReport):
 
     def __init__(self, *args, **kwargs):
         super(HistoricalAdherenceReport, self).__init__(*args, **kwargs)
-        self.episode = CaseAccessors(self.domain).get_case(self.episode_case_id)
+        try:
+            self.episode = CaseAccessors(self.domain).get_case(self.episode_case_id)
+        except:
```
snopoke: This should only catch `CaseNotFound`

snopoke (changes requested):

custom/enikshay/reports/historical_adherence.py
sravfeyn: Yeah, just updated

sravfeyn:
Updated

PR #16572, opened 2017-06-12
Submit updates in batches of 100 cases

sravfeyn:
@gcapalbo
FYI @proteusvacuum


proteusvacuum:
Build failure looks real:
```======================================================================

FAIL: custom.enikshay.tests.test_voucher_counts:TestVoucherCounts.test_case_updates

  File "/mnt/commcare-hq-ro/custom/enikshay/tests/test_voucher_counts.py", line 81, in test_case_updates

    {}

AssertionError: {'first_voucher_drugs': None} != {}

- {'first_voucher_drugs': None}

+ {}```

File: custom/enikshay/tasks.py [hunk 1]
```diff
@@ -99,6 +98,7 @@ def run(self):
                             e
                         )
                     )
+            bulk_update_cases(self.domain, updates)
```
gcapalbo: would be good to only perform this if len(updates) > 0

custom/enikshay/tasks.py
```diff
@@ -418,18 +396,28 @@ def get_aggregated_scores(self, latest_adherence_date, adherence_schedule_date_s
...
-        return any([
+        needs_update = any([
             self.get_property(k) != v
-            for (k, v) in case_properties_expected.iteritems()
+            for (k, v) in update_dict.iteritems()
         ])
+        if needs_update:
+            # cast any date objects to strings, so that Django template rendering
+            # used by update_case formats dates as YYYY:MM:DD instead of (Month Date, Year)
+            return {key: str(val) if isinstance(val, datetime.date) else val
```
gcapalbo: i just made a [PR](https://github.com/dimagi/commcare-hq/pull/16577) which makes the update_case utils use CaseBlock so that can just reuse the formatting done within the CaseBlock util (something i've wanted to do for some time).

once that's merged i think we can just merge master into here and remove any special formatting done so that the formatting is done in one place only

sravfeyn: Nice, thanks! The build failure is also related to the same issue

sravfeyn:
Updated

custom/enikshay/tasks.py
gcapalbo: @sravfeyn i think it would still be good to check for len(updates) > 0 here otherwise i think it will perform an unnecessary empty form submission

sravfeyn: Sorry, missed that. I have updated now

gcapalbo (approved):

PR #16573, opened 2017-06-12
Resolve bugs updating DHIS2 settings

kaapstorm:
Resolves a type casting bug and an arithmetic error when saving DHIS2 config, and provides more useful error messages from HQ.

Not a big change, but :fish: might be easier.

@calellowitz cc @sheelio

### DIFF
--- a/corehq/apps/domain/views.py
+++ b/corehq/apps/domain/views.py
@@ -3169,22 +3169,22 @@ class DataSetMapView(BaseAdminProjectSettingsView):
                 instance[key] = value
 
         try:
-            dataset_maps = json.loads(request.POST['dataset_maps'])
+            new_dataset_maps = json.loads(request.POST['dataset_maps'])
             current_dataset_maps = get_dataset_maps(request.domain)
             i = -1
             for i, dataset_map in enumerate(current_dataset_maps):
-                if i < len(dataset_maps):
+                if i < len(new_dataset_maps):
                     # Update current dataset maps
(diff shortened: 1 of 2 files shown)


corehq/apps/domain/views.py
```diff
@@ -3168,25 +3168,28 @@ def update_dataset_map(instance, dict_):
...
+                    dataset_map.delete()
+            if i + 1 < len(new_dataset_maps):
+                # Insert new dataset maps
+                for j in range(i + 1, len(new_dataset_maps)):
+                    dataset_map = DataSetMap(domain=request.domain)
+                    update_dataset_map(dataset_map, new_dataset_maps[j])
+                    dataset_map.save()
+            get_dataset_maps.clear(request.domain)
+            return json_response({'success': _('DHIS2 DataSet Maps saved')})
+        except Exception as err:
```
calellowitz: Do we have any idea what these exceptions may be? might be nice to have a more specific error response if there are specific problems we foresee

calellowitz (approved):
one comment. otherwise looks good

corehq/apps/domain/views.py
kaapstorm: The exception I was looking for was from trying to save a string as an integer. It was resolved by commit "Cast dayOfMonth as a number" (f4ac8c3). I'm not expecting exceptions in the future, but I figured it would be useful to leave this in, at least until we have one non-demo integration up and running.

PR #16574, opened 2017-06-12
remove old style args from management commands

snopoke:


corehq/apps/analytics/management/commands/update_hubspot_properties.py
```diff
@@ -11,11 +11,11 @@
 
 
 class Command(BaseCommand):
-    args = '<property_name_1> <property_name_2> ...'
     help = ("Updates given Hubspot properties for all users active within last 6 months. "
             "Only subscription, domain-membership, and A/B Test properties are supported")
 
     def add_arguments(self, parser):
+        parser.add_argument('property_names', nargs='+')
```
gcapalbo: looks like this was already here (next line)?

corehq/apps/app_manager/management/commands/find_broken_builds.py
```diff
@@ -71,9 +71,9 @@ def broken_suite_files(build):
 
 class Command(BaseCommand):
     help = "Print a list of broken builds"
-    args = '<check_function>'
 
     def add_arguments(self, parser):
+        parser.add_argument('check_function')
```
gcapalbo: this also was already here with a destination variable of `check_fn_name`

corehq/apps/data_pipeline_audit/management/commands/compare_doc_counts.py
```diff
@@ -8,9 +8,9 @@
 
 class Command(BaseCommand):
     help = "Print doc counts in ES & Primary database split by Doc Type."
-    args = '<domain1>..<domainN>'
 
     def add_arguments(self, parser):
+        parser.add_argument('domains', nargs='*')
```
gcapalbo: if i'm reading [this example](https://docs.djangoproject.com/en/1.11/howto/custom-management-commands/#module-django.core.management) right, this means the list of domains will now be available in `options['domains']`, instead of args, in the handle() method

corehq/apps/callcenter/management/commands/rebuild_call_center_datasources.py
```diff
@@ -13,7 +13,6 @@
 
 
 class Command(BaseCommand):
-    args = '<domain domain ...>'
```
gcapalbo: this one has the same issue where the list of domains will be in options['domains']

corehq/apps/data_pipeline_audit/management/commands/compare_doc_ids.py
```diff
@@ -13,9 +13,10 @@
 
 class Command(BaseCommand):
     help = "Print doc IDs that are in the primary DB but not in ES. Use in conjunction with 'raw_doc' view."
-    args = '<domain> <doc_type>'
 
     def add_arguments(self, parser):
+        parser.add_argument('domain')
+        parser.add_argument('doc_type')
```
gcapalbo: (same thing, i think the positional args passed to handle() are just empty now)

gcapalbo:
@snopoke made a few comments.  i think my comments apply to all these management commands unless i've misread the docs

corehq/apps/callcenter/management/commands/rebuild_call_center_datasources.py
snopoke: this does work because the handle function doesn't specify *args:
```
def handle(self, domains, **options):
```
So the kwarg `domains` will get pulled out of `options` and supplied as a positional arg.

corehq/apps/data_pipeline_audit/management/commands/compare_doc_ids.py
snopoke: Same as above, this works fine.

snopoke (commented):
Updated

corehq/apps/callcenter/management/commands/rebuild_call_center_datasources.py
gcapalbo: ah, that's right

corehq/apps/app_manager/management/commands/find_broken_builds.py
```diff
@@ -101,8 +100,8 @@ def add_arguments(self, parser):
             help='End date',
         )
 
-    def handle(self, check_fn_name, **options):
-        check_fn = CHECK_FUNCTIONS[check_fn_name]
+    def handle(self, check_function, **options):
```
gcapalbo: i think the destination variable will still be named `check_fn_name`

corehq/apps/tour/management/commands/unmark_tour_as_seen.py
```diff
@@ -12,7 +12,10 @@
 
 class Command(BaseCommand):
     help = "marks a tour with tour_slug as not seen for all existing users"
-    args = "tour_slug <specific user>"
+
+    def add_arguments(self, parser):
+        parser.add_argument('tour_slug')
+        parser.add_argument('user', nargs='?')
```
gcapalbo: doesn't this one still not work for `user`?

gcapalbo:
left two comments otherwise :+1:

corehq/apps/app_manager/management/commands/find_broken_builds.py
snopoke: thanks, updated to remove `dest` option from argument

corehq/apps/tour/management/commands/unmark_tour_as_seen.py
snopoke: fixed

snopoke:
@gcapalbo updated. Thanks for the close review.

gcapalbo (approved):

PR #16582, opened 2017-06-12
Add some metrics for export duration

benrudolph:
@snopoke couldn't think of any reasonable tags to add here. it'd be kind of nice to correlate with size of document. worth bucketing the average doc size as a tag? otherwise i think domain will generate too many metrics

cc: @mkangia

### DIFF
--- a/corehq/apps/export/export.py
+++ b/corehq/apps/export/export.py
@@ -360,10 +360,11 @@ def _write_export_instance(writer, export_instance, documents, progress_tracker=
     start = _time_in_milliseconds()
     total_bytes = 0
     total_rows = 0
+    compute_total = 0
+    write_total = 0
 
     for row_number, doc in enumerate(documents):
-        doc_bytes = sys.getsizeof(doc)
-        total_bytes += doc_bytes
+        total_bytes += sys.getsizeof(doc)
         for table in export_instance.selected_tables:
             compute_start = _time_in_milliseconds()
             rows = table.get_rows(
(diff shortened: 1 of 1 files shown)


snopoke (approved):
Could normalize by total size as well - I think that would be more useful than tagging by avg size.

Could also be worth timing `table.get_rows` separately from `writer.write` (unless you know what the breakdown there is).

benrudolph:
that's a good idea, will add those

benrudolph:
✔️

corehq/apps/export/export.py
```diff
@@ -356,26 +357,70 @@ def _write_export_instance(writer, export_instance, documents, progress_tracker=
...
+            _record_datadog_export_compute_rows(compute_start, compute_end, doc_bytes, len(rows))
+
+            write_start = _time_in_milliseconds()
             for row in rows:
                 # It might be bad to write one row at a time when you can do more (from a performance perspective)
                 # Regardless, we should handle the batching of rows in the _Writer class, not here.
                 writer.write(table, row)
+            write_end = _time_in_milliseconds()
+
+            _record_datadog_export_write_rows(write_start, write_end, doc_bytes, len(rows))
```
snopoke: is it necessary to record these inside the loop? Could just total them up and then record once at the end.

benrudolph: probably not, originally thought it might be nice to have the histogram stats on a row level for better accuracy, but thinking about it again i actually don't think there will be a lot of variance in these timings for one export

benrudolph: will update

corehq/apps/export/export.py
```diff
@@ -354,21 +357,71 @@ def _write_export_instance(writer, export_instance, documents, progress_tracker=
...
+
+            total_rows += len(rows)
+
         if progress_tracker:
             DownloadBase.set_progress(progress_tracker, row_number + 1, documents.count)
 
+    end = _time_in_milliseconds()
+    _record_datadog_export_write_rows(write_total, total_bytes, total_rows)
+    _record_datadog_export_compute_rows(compute_total, total_bytes, total_rows)
+    _record_datadog_export_write_rows(end - start, total_bytes, total_rows)
```
benrudolph: oops need to change this function name

snopoke (commented):
It just occurred to me that a very useful tag would be the output format. I imagine that writing to CSV is much faster than to Excel.

benrudolph:
💡

PR #16586, opened 2017-06-12
clean up es facets

emord:
When looking at the ES cluster I remembered that something blocking us from upgrading to 2.x is facets. I got rid of a couple of easy wins. I think the  remaining parts that would need to migrate are:

- admin reports
- exchange
- pact reports
- bihar reports (I think this project spun down though, so wecan remove that?)

admin reports will probably be the worst to migrate, but I don't see anything at the moment that would block us 

buddy @orangejenny

### DIFF
--- a/corehq/apps/domain/calculations.py
+++ b/corehq/apps/domain/calculations.py
@@ -377,18 +377,14 @@ def calced_props(dom, id, all_stats):
     }
 
 
-def total_distinct_users(domains=None):
+def total_distinct_users(domain):
     """
-    Get total number of users who've ever submitted a form.
+    Get total number of users who've ever submitted a form in a domain.
     """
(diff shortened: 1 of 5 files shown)


corehq/apps/domain/calculations.py
```diff
@@ -377,18 +377,14 @@ def calced_props(dom, id, all_stats):
     }
 
 
-def total_distinct_users(domains=None):
+def total_distinct_users(domain):
```
orangejenny: Don't you also need to change the call? https://github.com/dimagi/commcare-hq/blob/master/corehq/apps/domain/calculations.py#L342

emord: Actually it did work because `.domain` can accept a list or a string, but agreed that that is not obvious

snopoke (approved):
nice

PR #16587, opened 2017-06-12
store location data to beneficiary

nickpell:
Adds current_address_postal_code and current_address_village_town_city

I'm skipping assigning other current_address_* because the UATBC location hierarchy is inconsistent and we should have most of the data in current_address anyway.  I do save all of the UATBC legacy location data just in case we want to use it later.

@proteusvacuum cc: @esoergel


custom/enikshay/private_sector_datamigration/factory.py
```diff
@@ -120,6 +126,12 @@ def get_person_case_structure(self):
 
         kwargs['attrs']['update']['facility_assigned_to'] = self._location_owner_id
 
+        if self.beneficiary.pincode:
+            kwargs['attrs']['update']['current_address_postal_code'] = self.beneficiary.pincode
+
+        if self.beneficiary.villageTownCity is not None:
+            kwargs['attrs']['update']['current_address_village_town_city'] = self.beneficiary.villageTownCity
```
esoergel: just wondering - why are these two only set if they're not null, when the other ones are all set unconditionally?

esoergel (approved):

custom/enikshay/private_sector_datamigration/factory.py
nickpell: for the legacy_* properties, I'd like to explicitly mark them as null if that's the case, for my own record keeping.

For the current_address_* properties, I like to follow the convention that if they weren't set in UATBC, then the case should look like they were never entered in the app.  But it's a somewhat arbitrary convention.

esoergel: makes sense, thanks for the explanation!

PR #16588, opened 2017-06-12
Diagnostic investigations results task

proteusvacuum:
@NoahCarnahan @sravfeyn 

Attempt 2


custom/enikshay/case_utils.py
```diff
@@ -349,7 +349,6 @@ def _get_private_locations(person_case):
         raise NikshayCodeNotFound("Nikshay codes not found: {}".format(e))
 
 
-@nottest
```
NoahCarnahan: Did you not add `hqnottest` here on purpose?

NoahCarnahan (approved):

custom/enikshay/case_utils.py
proteusvacuum: mostly because it is monday.

PR #16575, opened 2017-06-12
be specific when casting boolean-string urlparams

sravfeyn:
https://manage.dimagi.com/default.asp?255822#1347786
@gcapalbo

PR #16591, opened 2017-06-12
Don't cast boolean filter values to booleans

proteusvacuum:
@NoahCarnahan 

Requires: https://github.com/dimagi/dimagi-utils/pull/411

### DIFF
--- a/corehq/apps/userreports/reports/view.py
+++ b/corehq/apps/userreports/reports/view.py
@@ -104,7 +104,7 @@ def query_dict_to_dict(query_dict, domain):
     :param domain:
     :return: a dict
     """
-    request_dict = json_request(query_dict)
+    request_dict = json_request(query_dict, booleans_as_strings=True)
     request_dict['domain'] = domain
     return request_dict
 


corehq/apps/userreports/reports/view.py
```diff
@@ -104,7 +104,7 @@ def query_dict_to_dict(query_dict, domain):
     :param domain:
     :return: a dict
     """
-    request_dict = json_request(query_dict)
+    request_dict = json_request(query_dict, booleans_as_strings=True)
```
NoahCarnahan: Does this work when filtering both string and boolean typed columns? Is it worth writing some tests?

proteusvacuum: I think yes.
Booleans in UCR are modeled as `0` or `1`, so filters actually need to be of type `integer`. There is no `boolean` filter type.

proteusvacuum:
✅

corehq/apps/userreports/reports/view.py
sravfeyn: I changed this in favour of https://github.com/dimagi/commcare-hq/pull/16575/ (that does more explicit casting and has tests)

PR #16593, opened 2017-06-12
ICDS: New Child Monthly Data Source

sheelio:
@emord 
This should get added to the existing monthly queue.  (Probably best reviewed commit by commit).


custom/icds_reports/ucr/data_sources/child_health_cases_monthly_tableau2.json
```diff
@@ -2415,6 +2622,201 @@
...
+        "type": "switch",
+        "switch_on": {
+          "type": "nested",
+          "argument_expression": {
+            "aggregation_fn": "last_item",
+            "type": "reduce_items",
+            "items_expression": {
+              "type": "icds_get_case_forms_by_date",
+              "xmlns": [
+                "http://openrosa.org/formdesigner/7a55754119359466b1951d7251068bd4f45e73c3"
```
emord: How many times does this form get submitted per case? If it's a lot, this is probably going to be really expensive

custom/icds_reports/ucr/data_sources/child_health_cases_monthly_tableau2.json
```diff
@@ -102,25 +102,24 @@
...
         "display_name": null,
         "expression": {
-          "type": "nested",
-          "argument_expression": {
-            "type": "ancestor_location",
-            "location_id": {
-              "expression": {
-                "property_name": "owner_id",
-                "type": "property_name"
+          "location_id_expression": {
```
emord: Why did you need to revert ancestor location?

emord (commented):
Need to do figure out how many shards to give this index. How many docs do you anticipate this to create at scale?

sheelio:
@emord 
This is the same as the other child health monthly (which we will eventually remove).  Approximately 300 per AWC, so 51 million for 170K AWCs.

I reverted the ancestor location changes since they weren't working when I tested them locally (and the previous configuration using location_parent_id was working).

custom/icds_reports/ucr/data_sources/child_health_cases_monthly_tableau2.json
sheelio: About once a month.  Over a child's life this form should be filled out approximately 30 times.

custom/icds_reports/ucr/data_sources/child_health_cases_monthly_tableau2.json
```diff
@@ -2373,6 +2371,35 @@
           }
         ]
       },
+      "height_eligible": {
+        "type": "and",
+        "filters": [
+          {
+            "type": "boolean_expression",
+            "operator": "gte",
```
sheelio: There isn't a between filter unfortunately (at least not that I know of).

PR #16594, opened 2017-06-12
ICDS: Add New CCS Record Monthly

sheelio:
@emord 
Also added to the existing queue and reviewed commit by commit


emord (commented):
Looks fine pending answer on reverting ancestor location and deciding on # of shards. How many docs do you expec this to have?

sheelio:
Expecting about 100-150 per AWC.  So about 17 million at 170 million users.

PR #16598, opened 2017-06-12
populate contact_phone_number in migrations

nickpell:
@proteusvacuum @gcapalbo cc: @esoergel 

Since there is real public sector data that uses phone_number instead of contact_phone_number, that's going to have to be migrated over and all apps in the field updated.  I think that task will require coordination with the rollout team, so I'm adding an exception to make the public sector migration fail until that migration has been completed.  I don't expect needing to run the public sector migration until the next rollout (during the fall) so hopefully the migration of phone_number -> contact_phone_number will be complete by then.


custom/enikshay/nikshay_datamigration/factory.py
```diff
@@ -143,6 +143,7 @@ def get_person_case_structure(self):
                 'update': {
                     'age': self.patient_detail.page,
                     'age_entered': self.patient_detail.page,
+                    'contact_phone_number': '91' + validate_phone_number(self.patient_detail.pmob),
```
gcapalbo: what happens if `validate_phone_number` raises an exception when converting to int, or raises an AssertionError?

also could we add validation in `validate_phone_number` to check that the phone number is positive after the int conversion

gcapalbo:
thanks, taking conversation to email

custom/enikshay/nikshay_datamigration/factory.py
nickpell: I'd see a message in the migration logs that migrating the particular user failed due to a phone number issue.  fwiw that hasn't happened yet in the 6 districts we've migrated.

nickpell: Added validation that phone number is a positive integer.

custom/enikshay/nikshay_datamigration/factory.py
```diff
@@ -31,6 +31,7 @@ def validate_phone_number(string_value):
     if string_value is None or string_value.strip() in ['', '0']:
         return ''
     else:
+        assert int(string_value) > 0
         phone_number = str(int(string_value))
```
esoergel: @nickpell I'm not sure it's a good idea to coerce to an int, since that might do things like strip valid leading zeroes.  Python strings have a builtin `isdigit` method which you could use:
```python
In [1]: "abc".isdigit()
Out[1]: False

In [2]: "123".isdigit()
Out[2]: True

In [3]: "123abc".isdigit()
Out[3]: False
```
Or you could just call `int(string_value)` to see if it fails and do nothing with the result.

esoergel: Oh, that apparently would also suffice for the `> 0` check:
```python
In [1]: "-123".isdigit()
Out[1]: False
```
>  Return True if all characters in the string are digits and there is at least one character, False otherwise.

nickpell: good point @esoergel - updated

esoergel (approved):

custom/enikshay/nikshay_datamigration/factory.py
gcapalbo: removing the leading zero is actually required for putting it into international format (for example if it's an 11-digit number with a leading zero)

gcapalbo: if we use isdigit() we should also make sure we trim whitespace

PR #16599, opened 2017-06-12
Only apply module filter bindings if UI exists

orangejenny:
@emord

### DIFF
--- a/corehq/apps/app_manager/static/app_manager/js/module-view.js
+++ b/corehq/apps/app_manager/static/app_manager/js/module-view.js
@@ -94,7 +94,10 @@ $(function () {
 
     // Module filter
     $(function () {
-        $('#module-filter').koApplyBindings({xpath: initial_page_data("module_filter") || ''});
+        var $moduleFilter = $('#module-filter');
+        if ($moduleFilter.length) {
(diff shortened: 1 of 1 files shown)


corehq/apps/app_manager/static/app_manager/js/module-view.js
```diff
@@ -94,7 +94,10 @@ $(function () {
 
     // Module filter
     $(function () {
-        $('#module-filter').koApplyBindings({xpath: initial_page_data("module_filter") || ''});
+        var $moduleFilter = $('#module-filter');
+        if ($moduleFilter.length) {
```
emord: Could this be `if ($moduleFilter)`? I forget what's pythonic vs normal in JS

emord (approved):

corehq/apps/app_manager/static/app_manager/js/module-view.js
orangejenny: have to use `.length`, because `$(...)` always returns the jquery object. since the jquery object acts something like an array, i suppose it's at least consistent with the fact that empty arrays are truthy in javascript...

PR #16600, opened 2017-06-12
ICDS: Add Person Case Table 

sheelio:
@emord   This is the new person cases table.  Easiest to review commit by commit.  The first commit just copies the existing UCR.


custom/icds_reports/ucr/data_sources/person_cases_v2.json
```diff
@@ -2211,6 +2211,48 @@
...
+        "datatype": "date",
+        "expression": {
+          "type": "nested",
+          "argument_expression": {
+            "type": "reduce_items",
+            "aggregation_fn": "first_item",
+            "items_expression": {
+              "type": "filter_items",
+              "items_expression": {
+                "type": "icds_get_case_history_by_date",
```
emord: This could be really expensive, if there are a lot of these forms submitted. How many do you expect?

May be worht writing a icds_get_first_case_property or something like that so that it doesnt hit riak too mcuh

emord (commented):
Same questions about number of shards. How many number of docs do you expect this data source to have?

sheelio:
@emord 
We expect 1000 per AWC (so about 170 million at 170K users).

sheelio:
@emord 
The add household / edit member forms are filled out infrequently (like once or twice per person), so don't expect that this will be too slow if we can filter forms by XMLNS

custom/icds_reports/ucr/data_sources/person_cases_v2.json
sheelio: The add household / edit member forms are filled out infrequently (like once or twice per person), so don't expect that this will be too slow if we can filter forms by XMLNS

PR #16601, opened 2017-06-12
Add all ancestors with include_without_expand

proteusvacuum:
@esoergel @NoahCarnahan 
https://trello.com/c/V79NXuc5/107-update-sync-but-not-expand-options-to-choose-a-level-of-the-hierarchy

### DIFF
--- a/corehq/apps/locations/fixtures.py
+++ b/corehq/apps/locations/fixtures.py
@@ -237,18 +237,18 @@ def _get_include_without_expanding_locations(domain, location_type):
     """
     include_without_expanding = location_type.include_without_expanding
     if include_without_expanding is not None:
-        # add in other "forced" locations the user should have, along with their ancestors
-        forced_locations = set(SQLLocation.active_objects.filter(
-            domain__exact=domain,
-            location_type=include_without_expanding
-        ))
-        forced_ancestors = {
-            ancestor
-            for location in forced_locations
-            for ancestor in location.get_ancestors()
-        }
-
-        return forced_locations | forced_ancestors
+        forced_location_level = set(
+            SQLLocation.active_objects.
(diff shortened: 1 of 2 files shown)


corehq/apps/locations/fixtures.py
```diff
@@ -237,18 +237,18 @@ def _get_include_without_expanding_locations(domain, location_type):
...
-            for location in forced_locations
-            for ancestor in location.get_ancestors()
-        }
-
-        return forced_locations | forced_ancestors
+        forced_location_level = set(
+            SQLLocation.active_objects.
+            filter(domain__exact=domain, location_type=include_without_expanding).
+            values_list('level', flat=True)
+        ) or None
```
esoergel: This'll return the level value for each location of that type - you could instead use a `distinct` query to get only the unique values (of which there should only be one).  Actually you could also probably  just slap a `.first()` on the end of what you've got here and use that, since we expect them all to be the same anyways.

corehq/apps/locations/tests/test_location_fixtures.py
```diff
@@ -286,6 +285,85 @@ def test_index_location_fixtures(self):
...
+        location_metadata = {'is_test': 'no', 'nikshay_code': 'nikshay_code'}
+        setup_location_types_with_structure(self.domain, location_type_structure),
+        self.locations = setup_locations_with_structure(self.domain, location_structure, location_metadata)
+
+    def tearDown(self):
+        super(ForkedHierarchiesTest, self).tearDown()
+        delete_all_users()
+
+    def test_include_without_expanding_includes_all_ancestors(self):
+        self.user._couch_user.set_location(self.locations['DTO'].couch_location)
```
esoergel: `set_location` now also accepts SQL locations, actually

esoergel (approved):

PR #16605, opened 2017-06-13
device logs cleanup

snopoke:


emord (approved):
I can deploy this today

corehq/ex-submodules/phonelog/tasks.py
```diff
@@ -2,10 +2,13 @@
...
+from phonelog.models import DeviceReportEntry, UserErrorEntry, ForceCloseEntry, UserEntry
 
 
 @periodic_task(run_every=crontab(minute=0, hour=0), queue=getattr(settings, 'CELERY_PERIODIC_QUEUE', 'celery'))
 def purge_old_device_report_entries():
     max_age = datetime.utcnow() - timedelta(days=60)
     DeviceReportEntry.objects.filter(server_date__lt=max_age).delete()
+    UserErrorEntry.objects.filter(server_date__lt=max_age).delete()
+    ForceCloseEntry.objects.filter(server_date__lt=max_age).delete()
+    UserEntry.objects.filter(server_date__lt=max_age).delete()
```
gcapalbo: i think this won't delete the records with server_date null, is that intended?

emord: I think that makes sense. This made the migration really fast. we can fill those with today's date at a later time.

gcapalbo: ok, could even wait a little while and delete all the ones with nulls

snopoke:
Yes that was what I was intending to do.

PR #16609, opened 2017-06-13
DHIS2 integration UI improvements

kaapstorm:
A few small changes to make the DHIS2 config UI suck a little less.

The first commit makes it a bit more obvious that DataValues belong to DataSets. 

The second and third commits group related fields together, and use radio buttons to select/disable the fields that need to be completed. 

@calellowitz cc @sheelio


kaapstorm:
Radio buttons and grouped fields look like this:
dhis2_ui

PR #16611, opened 2017-06-13
mgmt command to queue rebuilding of UCR table in async queue

emord:
Wanted to put this in the normal rebuild, but this has a number of limitiations.

- Only supports sql cases. The couch case accessors don't have a filter by case type
- doesn't support xforms. will need to add a filter for xmlns to form accessors classes

@calellowitz @sheelio this is how your UCR changes will need to be rebuilt

buddy @orangejenny

### DIFF
--- a/corehq/apps/userreports/management/commands/async_rebuild_table.py
+++ b/corehq/apps/userreports/management/commands/async_rebuild_table.py
@@ -1,0 +1,45 @@
+from __future__ import print_function
+from collections import namedtuple
+
+from django.core.management.base import BaseCommand
+
+from corehq.apps.userreports.document_stores import get_document_store
+from corehq.apps.userreports.models import AsyncIndicator, get_datasource_config
+from corehq.apps.userreports.util import get_indicator_adapter
+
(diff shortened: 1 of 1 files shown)


corehq/apps/userreports/management/commands/async_rebuild_table.py
```diff
@@ -0,0 +1,45 @@
...
+        num_case_ids = len(case_ids)
+        print("inserting %d docs" % num_case_ids)
+        adapter = get_indicator_adapter(config, can_handle_laboratory=True)
+        adapter.build_table()
+        # normally called after rebuilding finishes
+        adapter.after_build_table()
+
+        for i, case_id in enumerate(case_ids):
+            change = FakeChange(case_id, fake_change_doc)
+            AsyncIndicator.update_indicators(change, [config._id])
```
emord: Ideally this would be batched, but this function does a safe check that will protect us if the celery queue is currently processing the record when it gets updated.

calellowitz (approved):

PR #16615, opened 2017-06-13
filtering out unregistered cases

calellowitz:
@sheelio @kaapstorm

### DIFF
--- a/custom/icds_reports/ucr/data_sources/ccs_record_cases_monthly_v2.json
+++ b/custom/icds_reports/ucr/data_sources/ccs_record_cases_monthly_v2.json
@@ -1377,6 +1377,41 @@
               }
             },
             "property_value": 183
+          },
+          {
+            "type": "or",
+            "filters": [
+              {
+                "operator": "not_eq",
+                "type": "boolean_expression",
+                "expression": {
+                  "doc_id_expression": {
+                    "type": "icds_parent_id"
+                  }
+                },
+                "related_doc_type": "CommCareCase",
+                "value_expression": {
+                  "type": "property_name",
+                  "property_name": "registered_status"
+                },
+                "property_value": "not_registered"
+              },
+              {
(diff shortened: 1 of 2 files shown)


kaapstorm (approved):

custom/icds_reports/ucr/data_sources/ccs_record_cases_monthly_v2.json
```diff
@@ -1377,6 +1377,43 @@
               }
             },
             "property_value": 183
+          },
+          {
+            "type": "or",
```
sheelio: @calellowitz 

Shouldn't this be an "and" (the person is not not_registered AND not migrated)?

sheelio: (guessing this should apply wherever you've added this check)

calellowitz: I think this is right. It includes them (which is my understanding of a filter, let me know if thats wrong) when the person is NOT `registered_status=not_registered` OR NOT `migrated=yes`, which is to say it includes them if either of the two does not match the exclusion condition. My read of the spec is that both of those properties need to be met to exclude them but it is a bit ambiguous

sheelio: @calellowitz I'm pretty sure that the rules are:
"valid" = registered_status != not_registered and migrated != yes

Basically as soon as you're migrated OR as soon as you're marked as "not registered" you're no longer considered a valid beneficiary.  Either should exclude them.

calellowitz:
@sheelio confirmed that you are correct with srikant. have updated the filter

sheelio (approved):

snopoke (approved):

PR #16618, opened 2017-06-13
Soft assert if there are multiple triggers

proteusvacuum:
@NoahCarnahan @esoergel 
I don't think we need to hard assert here, since we already only pluck the first one to send.


esoergel (approved):

proteusvacuum:
Thanks @esoergel

PR #16621, opened 2017-06-14
smaller chunks for pruning synclogs

emord:
This fails trying to post to nginx. will test it out tomorrow to make sure that 5 is small enough. currently still have synclogs dating back to october

@orangejenny


orangejenny (approved):

snopoke (approved):

emord:
confirmed chunk size is small enough

PR #16624, opened 2017-06-14
Allow module with custom xml for the configuration

mkangia:
~~I am not sure if its okay to still keep this available under the case list tile feature flag but removing the check to allow only modules that have case tiles available for the configuration.
This would allow user to also add modules with custom xml and context that is present but is not made a case tile.~~

Updated to include modules with custom case xml for the configuration as well.

cc: @sheelio


mkangia:
I have not tested this yet but sorted out tests so keeping the label. Good to review though.

mkangia:
tested. Works fine.

PR #16626, opened 2017-06-14
Total adherence score

kkrampa:
@NoahCarnahan As I wrote in the email, there was a problem with the solution that you've suggested. I'm proposing hiding column needed only in eval expression.

First, we need to merge #16623 

Please look only at last three commits.


NoahCarnahan:
This approach looks good, thanks

PR #16637, opened 2017-06-14
enikshay: Add tests for custom ucr expressions

NoahCarnahan:
@proteusvacuum

### DIFF
--- a/corehq/form_processor/abstract_models.py
+++ b/corehq/form_processor/abstract_models.py
@@ -215,14 +215,6 @@ class AbstractCommCareCase(CaseToXMLMixin):
     def set_case_id(self, case_id):
         raise NotImplementedError()
 
-    @property
-    def server_opened_on(self):
-        try:
-            open_action = self.actions[0]
-            return open_action.server_date
-        except Exception:
-            pass
-
     def _resolve_case_property(self, property_name, result):
         CasePropertyResult = collections.namedtuple('CasePropertyResult', 'case value')
 
--- a/corehq/form_processor/abstract_models.py
+++ b/corehq/form_processor/abstract_models.py
@@ -215,6 +215,14 @@ class AbstractCommCareCase(CaseToXMLMixin):
     def set_case_id(self, case_id):
         raise NotImplementedError()
 
+    @property
+    def server_opened_on(self):
+        try:
+            open_action = self.actions[0]
+            return open_action.server_date
+        except Exception:
+            pass
+
     def _resolve_case_property(self, property_name, result):
         CasePropertyResult = collections.namedtuple('CasePropertyResult', 'case value')
 
--- a/corehq/form_processor/serializers.py
+++ b/corehq/form_processor/serializers.py
@@ -154,7 +154,7 @@ class CommCareCaseSQLAPISerializer(serializers.ModelSerializer):
     date_modified = serializers.DateTimeField(source='modified_on')
     properties = serializers.JSONField(source='get_properties_in_api_format')
     server_date_modified = serializers.DateTimeField(source='server_modified_on')
-    server_date_opened = serializers.DateTimeField(source='opened_on')
+    server_date_opened = serializers.DateTimeField(source='server_opened_on')
     indices = serializers.JSONField(source='get_index_map')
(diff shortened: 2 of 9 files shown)


custom/enikshay/case_utils.py
```diff
@@ -197,7 +197,7 @@ def get_latest_trail_case_from_person(domain, person_case_id):
         case for case in reverse_indexed_cases
         if case.type == CASE_TYPE_TRAIL
     ]
-    trail_cases.sort(key=lambda c: c.opened_on)
+    trail_cases.sort(key=lambda c: c.server_opened_on)
```
NoahCarnahan: `opened_on` is just year-month-day, need a full timestamp.

NoahCarnahan:
adding hold-off label because github makes it looks like tests have passed, but they haven't yet. (Need to look at second to last commit).

corehq/form_processor/serializers.py
```diff
@@ -154,7 +154,7 @@ class CommCareCaseSQLAPISerializer(serializers.ModelSerializer):
     date_modified = serializers.DateTimeField(source='modified_on')
     properties = serializers.JSONField(source='get_properties_in_api_format')
     server_date_modified = serializers.DateTimeField(source='server_modified_on')
-    server_date_opened = serializers.DateTimeField(source='opened_on')
+    server_date_opened = serializers.DateTimeField(source='server_opened_on')
```
snopoke: I don't think this is needed in this serializer

corehq/form_processor/abstract_models.py
```diff
@@ -215,6 +215,14 @@ def to_api_json(self):
     def set_case_id(self, case_id):
         raise NotImplementedError()
 
+    @property
+    def server_opened_on(self):
+        try:
+            open_action = self.actions[0]
```
snopoke: One reason not to do this is that it will load all the case transactions from the DB

corehq/form_processor/serializers.py
NoahCarnahan: Shall we rename this serialized property then to accurately reflect the value ("server_date_opened" -> "opened_on"), or are you suggesting that we remove it altogether?

corehq/form_processor/abstract_models.py
NoahCarnahan: Right. Ok I guess in that case I'll just calculate the `server_opened_on` for my particular report, and not make it available on the model in general. In my case, the cases should almost always have just a single case action.

snopoke:
I wasn't suggesting you remove that. I just wanted to point on the
performance implications in case there was some other way around it for
your use case

corehq/form_processor/serializers.py
snopoke: I think it's fine to remove it altogether

NoahCarnahan:
Oh got it. Yeah looks like we could continue to use a `@property` that queries the `CaseTransactions` by date, and only fetches one.

We could also actually store this value in the case table. I'm thinking the first option is probably better since this property doesn't appear to be widely used.

snopoke:
👍 of performance becomes an issue we can revisit adding it to the table

custom/enikshay/case_utils.py
```diff
@@ -197,9 +197,15 @@ def get_latest_trail_case_from_person(domain, person_case_id):
...
-        return trail_cases[-1]
+    trails_with_server_opened_on = []
+    for trail in trail_cases:
+        server_opened_on = trail.actions[0].server_date
+        trails_with_server_opened_on.append((server_opened_on, trail))
+
+    trails_with_server_opened_on.sort()
+    if trails_with_server_opened_on:
+        # Return the latest trail case
+        return trails_with_server_opened_on[-1][1]
```
NoahCarnahan: Decided to just implement the existing approach here for my report, and not give sql cases a server_modified_on property for now because I'm a bit pressed for time, and also don't need it yet.

NoahCarnahan:
Ah whoops, I pushed some commits to this branch I wasn't intending to include. I'm going to close this PR and reopen a new one to avoid rebasing remote changes.

PR #16644, opened 2017-06-15
truncate really long question ids in question properties select 2

biyeun:
fixes https://manage.dimagi.com/default.asp?256090#1348537

@orangejenny // @sravfeyn

### DIFF
--- a/corehq/apps/app_manager/static/app_manager/js/case-knockout-bindings.js
+++ b/corehq/apps/app_manager/static/app_manager/js/case-knockout-bindings.js
@@ -8,7 +8,7 @@
         },
         getDisplay: function (question, MAXLEN) {
             return utils.getIcon(question) + utils.getLabel(question, MAXLEN)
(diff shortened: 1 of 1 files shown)


corehq/apps/app_manager/static/app_manager/js/case-knockout-bindings.js
```diff
@@ -19,7 +19,11 @@
             var MAXLEN = MAXLEN || 40,
                 maxlen = MAXLEN - suffix.length;
             return ((label.length <= maxlen) ? (label) : (label.slice(0, maxlen) + "...")) + suffix;
-        }
+        },
+        truncateValue: function (value, MAXLEN) {
+            MAXLEN = MAXLEN || 40;
```
orangejenny: Would be nice to just have one `maxlen || 40` in this file.

orangejenny (approved):

PR #16645, opened 2017-06-15
don't error if synclog is missing

snopoke:
Since we started deleting synclogs it is expected that they won't always exist

Fixes COMMCAREHQ-21C
https://sentry.io/dimagi/commcarehq/issues/263323774/

### DIFF
--- a/corehq/form_processor/abstract_models.py
+++ b/corehq/form_processor/abstract_models.py
@@ -124,10 +124,7 @@ class AbstractXFormInstance(object):
             try:
                 return get_properly_wrapped_sync_log(self.last_sync_token)
             except ResourceNotFound:
-                logging.exception('No sync token with ID {} found. Form is {} in domain {}'.format(
-                    self.last_sync_token, self.form_id, self.domain,
-                ))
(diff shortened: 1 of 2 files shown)


corehq/form_processor/abstract_models.py
```diff
@@ -124,10 +124,7 @@ def get_sync_token(self):
             try:
                 return get_properly_wrapped_sync_log(self.last_sync_token)
             except ResourceNotFound:
-                logging.exception('No sync token with ID {} found. Form is {} in domain {}'.format(
-                    self.last_sync_token, self.form_id, self.domain,
-                ))
-                raise
+                pass
```
gcapalbo: it looks like this now returns None rather than failing hard. just confirming that the code which uses this can handle the None return value?

snopoke: yes all callers check for None which was also a possible return type before.

snopoke:
@gcapalbo added on more commit to remove some unnecessary error handling

gcapalbo (approved):

PR #16646, opened 2017-06-15
Add custom messaging recipient for ICDS

gcapalbo:
Allows us to send reminders to the mother person case when the child_health extension case is the one that spawns the alert.

@sravfeyn @snopoke

### DIFF
--- a/custom/icds/messaging/custom_recipients.py
+++ b/custom/icds/messaging/custom_recipients.py
@@ -1,0 +1,19 @@
+from corehq.form_processor.interfaces.dbaccessors import CaseAccessors
+from corehq.form_processor.models import CommCareCaseIndexSQL
+
+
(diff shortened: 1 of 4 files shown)


custom/icds/messaging/custom_recipients.py
```diff
@@ -0,0 +1,19 @@
+from corehq.form_processor.interfaces.dbaccessors import CaseAccessors
+from corehq.form_processor.models import CommCareCaseIndexSQL
+
+
+def mother_person_case_from_child_health_case(case_schedule_instance):
+    child_health_case = CaseAccessors(case_schedule_instance.domain).get_case(case_schedule_instance.case_id)
```
snopoke: should this handle `CaseNotFound` exception?

PR #16648, opened 2017-06-15
escape special chars; add test

sravfeyn:
http://manage.dimagi.com/default.asp?256190#edit_0_256190
@gcapalbo 
https://sentry.io/dimagi/commcarehq/issues/284514522/

### DIFF
--- a/corehq/apps/ota/views.py
+++ b/corehq/apps/ota/views.py
@@ -184,7 +184,8 @@ def _add_case_property_queries(domain, case_type, search_es, criteria):
             case_property=key,
         )
         for removal_regex in remove_char_regexs:
-            value = re.sub(removal_regex.regex, '', value)
+            to_remove = re.escape(removal_regex.regex)
+            value = re.sub(to_remove, '', value)
         search_es = search_es.case_property_query(key, value, fuzzy=(key in fuzzies))
 
     return search_es
--- a/corehq/apps/ota/tests/test_search_claim_endpoints.py
+++ b/corehq/apps/ota/tests/test_search_claim_endpoints.py
@@ -100,10 +100,21 @@ class CaseSearchTests(TestCase, ElasticTestMixin):
         rc.save()
         config.ignore_patterns.add(rc)
         config.save()
+        rc = IgnorePatterns(
+            domain=DOMAIN,
+            case_type='case_type',
+            case_property='phone_number',
+            regex='+',
+        )                       # remove '+' from the phone_number case property
+        rc.save()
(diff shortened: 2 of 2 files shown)


corehq/apps/ota/views.py
```diff
@@ -184,7 +184,8 @@ def _add_case_property_queries(domain, case_type, search_es, criteria):
             case_property=key,
         )
         for removal_regex in remove_char_regexs:
-            value = re.sub(removal_regex.regex, '', value)
+            to_remove = re.escape(removal_regex.regex)
```
gcapalbo: are we sure that we want to escape all the characters in `removal_regex`? that would mean that we [lose](https://docs.python.org/2/library/re.html#re.escape) all pattern matching capabilities.

if `removal_regex` is configurable, i think it would be preferable to just configure the plus as `'\+` in the regex when trying to match an actual plus so that we can still use the `+` for pattern matching.

sravfeyn: It looks like `regex` attribute on IgnorePatterns is a misnomer, it means a substring to ignore (at least as it's described in [the UI](https://india.commcarehq.org/a/enikshay/settings/project/case_search/) and not a regex. This is something that's configured by users who we assume don't know regex and that input is not treated as regex (for e.g. it's not valdiated if it's a valid regex or not). Should we be treating the input as regex?

```
>>> re.sub("+", '', "+91")
  File "<stdin>", line 1, in <module>
  File "/Users/sravfeyn/HQ/virtualenvs/hqenv/lib/python2.7/re.py", line 155, in sub
    return _compile(pattern, flags).sub(repl, string, count)
  File "/Users/sravfeyn/HQ/virtualenvs/hqenv/lib/python2.7/re.py", line 251, in _compile
    raise error, v # invalid expression
sre_constants.error: nothing to repeat
>>> cut = lambda x, y: re.sub(re.escape(x), '', y)
>>> cut(".exe", "python.exe")
'python'
>>> cut("+", "+91")
'91'
>>> cut("-", "011-91")
'01191'
>>> cut("word", "some word")
'some '
>>> cut("/", "some/word")
'someword'
```

proteusvacuum: hey, I think let's leave it as a regex for now. I've updated the app config for now, and will update the docs / do some validation when I get back on Tuesday.

gcapalbo: @sravfeyn i see, maybe this requires input from product then on whether this should be treated as a regex or substring match

proteusvacuum:
merged this, because I think this is actually want we want to have happen.

PR #16651, opened 2017-06-15
Np/tb/language code

nickpell:
@gcapalbo


gcapalbo (approved):

gcapalbo:
just realized, does https://github.com/dimagi/commcare-hq/blob/master/custom/enikshay/integrations/bets/repeater_generators.py#L457 need to be updated too?

PR #16656, opened 2017-06-15
Add separate path for expanding inline details

wpride:
When clicking a case tile while on a case list this action was indistinguishable from clicking a case list row (IE both made a `get_details` call with the same step args). This fixes that by adding a slightly different path for persistent case tile expansion.

Fixes: https://manage.dimagi.com/default.asp?252775
cross-request: https://github.com/dimagi/formplayer/pull/413

### DIFF
--- a/corehq/apps/cloudcare/static/cloudcare/js/formplayer/menus/api.js
+++ b/corehq/apps/cloudcare/static/cloudcare/js/formplayer/menus/api.js
@@ -59,6 +59,7 @@ FormplayerFrontend.module("Menus", function (Menus, FormplayerFrontend, Backbone
                 "previewCommand": params.previewCommand,
                 "installReference": params.installReference,
                 "oneQuestionPerScreen": displayOptions.oneQuestionPerScreen,
+                "isPersistent": params.isPersistent,
             });
             options.url = formplayerUrl + '/' + route;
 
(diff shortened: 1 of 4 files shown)


corehq/apps/cloudcare/static/cloudcare/js/formplayer/menus/views.js
```diff
@@ -248,6 +248,13 @@ FormplayerFrontend.module("Menus.Views", function (Views, FormplayerFrontend, Ba
         },
     });
 
+    Views.PersistentCaseTileView = Views.CaseTileView.extend({
+        rowClick: function (e) {
+            e.preventDefault();
+            FormplayerFrontend.trigger("menu:show:detail", this.options.model.get('id'), 0, true);
```
benrudolph: nit i don't think you actually need to use `this.options`, backbone i believe saves it as `this.model`

benrudolph (approved):

corehq/apps/cloudcare/static/cloudcare/js/formplayer/menus/api.js
```diff
@@ -76,7 +76,8 @@ FormplayerFrontend.module("Menus", function (Menus, FormplayerFrontend, Backbone
         return Menus.API.queryFormplayer(options, 'navigate_menu');
     });
 
-    FormplayerFrontend.reqres.setHandler("entity:get:details", function (options) {
+    FormplayerFrontend.reqres.setHandler("entity:get:details", function (options, isPersistent) {
```
benrudolph: is `isPersistent` something we should store in the URL?

wpride: No, I think it needs only be a one off

PR #16657, opened 2017-06-15
Remove DetailScreenConfig.messages

orangejenny:
Simplify.

@benrudolph


benrudolph (approved):
hooray. just verifying, i don't remember if we ever found out if `gettext` ever actually worked, do you?

orangejenny:
Not sure, I haven't looked into it. Would check with Biyeun if there's a question of it working.

PR #16658, opened 2017-06-15
Add multiple UCRs to aysnc table at once

emord:
@calellowitz So that we can batch the management command for yours and sheels rebuild together

### DIFF
--- a/corehq/apps/userreports/management/commands/async_rebuild_table.py
+++ b/corehq/apps/userreports/management/commands/async_rebuild_table.py
@@ -1,4 +1,5 @@
 from __future__ import print_function
+import argparse
 from collections import namedtuple
 
 from django.core.management.base import BaseCommand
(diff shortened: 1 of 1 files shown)


corehq/apps/userreports/management/commands/async_rebuild_table.py
```diff
@@ -16,30 +17,35 @@ class Command(BaseCommand):
...
-        adapter = get_indicator_adapter(config, can_handle_laboratory=True)
-        adapter.build_table()
-        # normally called after rebuilding finishes
-        adapter.after_build_table()
 
+        for config in configs:
+            adapter = get_indicator_adapter(config, can_handle_laboratory=True)
+            adapter.build_table()
+            # normally called after rebuilding finishes
+            adapter.after_table_build()
```
emord: I was calling the wrong function here before

calellowitz (approved):

PR #16661, opened 2017-06-15
enikshay: Add tests for custom ucr expressions (take 2)

NoahCarnahan:
@nickpell @proteusvacuum 

This is a fixed up version of https://github.com/dimagi/commcare-hq/pull/16637/


nickpell (approved):
Didn't look too closely, but looks reasonable

PR #16664, opened 2017-06-16
ENikshay model migration: episode facility id migrations

proteusvacuum:
https://docs.google.com/document/d/1w6T1ShTdZCB1UZjogckh1uBUYynMUOY-OfOi5d4IFt8/edit#

This ended up being much harder than I had expected... We don't have a good way of saying "get this case property at this date", so I had to hack it a little. 

This should only get run once per case, only if the case doesn't already have the property we are looking for.
This task will get deleted at some point, after everyone updates their apps to the latest version, which sets this property automatically... 😈 

buddy @snopoke  (also, if you can think of a better way of doing this, I'm all 👂 )
@sravfeyn


custom/enikshay/tasks.py
```diff
@@ -86,10 +88,12 @@ def run(self):
                 adherence_update = EpisodeAdherenceUpdate(episode, self)
                 voucher_update = EpisodeVoucherUpdate(self.domain, episode)
                 test_update = EpisodeTestUpdate(self.domain, episode)
+                episode_facility_id_migration = EpisodeFacilityIDMigration(self.domain, episode)
```
sravfeyn: Isn't this a one time migration, why not put this in a management command instead of a periodic task that repeats?

custom/enikshay/model_migration_sets/episode_facility_id_migration.py
```diff
@@ -0,0 +1,122 @@
...
+        return self.person.actions
+
+    @memoized
+    def _episode_actions(self):
+        return self.episode.actions
+
+    def update_json(self):
+        if not self.should_update:
+            return {
+                'facility_id_migration_complete': 'true',
```
sravfeyn: Will this not repeatedly submit this update for cases that already have this?

custom/enikshay/model_migration_sets/episode_facility_id_migration.py
```diff
@@ -0,0 +1,122 @@
...
+        for i, transactions in enumerate(actions):
+            property_changed_info = self._property_changed_in_action(transactions, case_property)
+            if property_changed_info and property_changed_info.new_value == value:
+                # get the date that case_property changed
+                date_of_change = parse_datetime(property_changed_info.modified_on)
+                break
+
+        return date_of_change
+
+    def get_owner_id_at_date(self, date):
```
sravfeyn: docstrings will be useful in future when deciphering specs like these.

custom/enikshay/model_migration_sets/episode_facility_id_migration.py
snopoke: agree this logic should change to make the migration idempotent

snopoke (changes requested):
I don't see a better way of doing this but it's going to be slow and put quite a bit of load on riak.

proteusvacuum:
✅

proteusvacuum:
✅

custom/enikshay/tasks.py
sravfeyn: Hey @proteusvacuum this is still unclear to me

snopoke (approved):

custom/enikshay/tasks.py
proteusvacuum: because users won't all migrate to the new app version at the same time, so cases without these case properties will still get created.

PR #16666, opened 2017-06-16
Use getattr

sravfeyn:
http://manage.dimagi.com/default.asp?256360#1350063
@nickpell 

May have been introduced [here](https://github.com/dimagi/commcare-hq/pull/16354/files#diff-b3fe9f03e6ae57a909134fabd8bc228dR685) FYI @orangejenny


orangejenny (approved):

orangejenny:
thank you @sravfeyn !

PR #16667, opened 2017-06-16
Update continuous indexing frequency

gcapalbo:
Seems like the couch timings spike on the hour, which might be related to this task.

hourly-couch-slowness

This makes the task run every 5 minutes which should reduce the amount of indexing needed.

Alternatively, we could also only run this task off-hours.

@dimagi/scale-team

### DIFF
--- a/corehq/preindex/tasks.py
+++ b/corehq/preindex/tasks.py
@@ -11,7 +11,7 @@ def run_continuous_indexing_task():
     preindex_couch_views.delay()
 
 
-@serial_task('couch-continuous-indexing', timeout=60 * 60, queue='background_queue', max_retries=0)
+@serial_task('couch-continuous-indexing', timeout=60 * 60, queue=settings.CELERY_PERIODIC_QUEUE, max_retries=0)
 def preindex_couch_views():
     for design in get_preindex_designs():
         index_design_doc(design)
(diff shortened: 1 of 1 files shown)


corehq/preindex/tasks.py
```diff
@@ -2,9 +2,16 @@
...
+from django.conf import settings
 
 
-@periodic_task(run_every=crontab(minute=0), queue='background_queue')
+@periodic_task(run_every=crontab(minute='*/5'), queue=settings.CELERY_PERIODIC_QUEUE)
+def run_continuous_indexing_task():
+    preindex_couch_views.delay()
+
+
+@serial_task('couch-continuous-indexing', timeout=60 * 60, queue=settings.CELERY_PERIODIC_QUEUE, max_retries=0)
```
emord: What's the benefit of changing this to a serial task and using a periodic task to call delay?

emord (approved):
No opinion on whether we should run this only on off hours or just more often. Seems good to run it more often though

corehq/preindex/tasks.py
benrudolph: guessing it's to try and prevent a reindex from starting before the previous one has finished. i think that was still a possibility in the previous version, and now more of a possibility since the interval is shorter

gcapalbo: yeah, using serial_task with max_retries=0 makes it so that we never have two of these tasks running at the same time, and if a task is spawned while another is running, it will be ignored (within the timeout window, so within a 1 hour period in this case).

we don't have a serial_periodic_task util yet, so that's the only reason i had to split them into two tasks.

emord: Ah cool. that makes sense. Thanks for catching this

PR #16675, opened 2017-06-16
Add xpath query right after it's executed

benrudolph:
@proteusvacuum you drive me to be a better person! this should add the xpath query right after it's executed not when you close / reopen the debugger

cc: @orangejenny


orangejenny:
nice usability improvement

orangejenny (approved):

proteusvacuum:
Thanks !

PR #16680, opened 2017-06-16
enikshay: Add ucr episode_from_person expression

NoahCarnahan:
Adds another expression needed by the referral report. I don't believe there is way to access reverse indexed cases with existing ucr expressions, so this custom expression is needed to fetch episode from person.
@nickpell


nickpell:
nice!

nickpell (approved):

PR #16683, opened 2017-06-16
Cleanup app_manager.js

biyeun:
mainly split the `init` method into smaller methods that will make this a little easier to follow

@orangejenny


corehq/apps/app_manager/static/app_manager/js/app_manager.js
```diff
@@ -74,9 +74,6 @@ hqDefine('app_manager/js/app_manager.js', function () {
             }
             }
             }
-        function getVar(name) {
```
orangejenny: hooray, always glad to see functions get dropped

orangejenny (approved):
this is great

PR #16685, opened 2017-06-16
Include inline and persistent details when gathering xpaths

orangejenny:
https://manage.dimagi.com/default.asp?256148

@benrudolph / @sravfeyn

### DIFF
--- a/corehq/apps/app_manager/suite_xml/post_process/instances.py
+++ b/corehq/apps/app_manager/suite_xml/post_process/instances.py
@@ -33,6 +33,8 @@ class EntryInstances(PostProcessor):
         for datum in entry.datums:
             detail_ids.add(datum.detail_confirm)
             detail_ids.add(datum.detail_select)
+            detail_ids.add(datum.detail_inline)
+            detail_ids.add(datum.detail_persistent)
             xpaths.add(datum.nodeset)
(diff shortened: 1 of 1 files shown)


benrudolph (approved):

corehq/apps/app_manager/suite_xml/post_process/instances.py
```diff
@@ -33,6 +33,8 @@ def _get_all_xpaths_for_entry(self, entry):
         for datum in entry.datums:
             detail_ids.add(datum.detail_confirm)
             detail_ids.add(datum.detail_select)
+            detail_ids.add(datum.detail_inline)
+            detail_ids.add(datum.detail_persistent)
```
sravfeyn: Would be good to have a test that reproduces the bug.

sravfeyn:
Thanks Jenny!

corehq/apps/app_manager/suite_xml/post_process/instances.py
sravfeyn: Would be good to have a test that reproduces the bug.

orangejenny: thanks for the nudge! https://github.com/dimagi/commcare-hq/pull/16713

PR #16687, opened 2017-06-16
don't fail when setting start or limit

nickpell:
need to query before taking a slice of the queryset

@NoahCarnahan @sravfeyn


nickpell:
✅

NoahCarnahan (approved):

PR #16689, opened 2017-06-17
Only run indexing tax before work day

emord:
@dimagi/scale-team Seems strange to me that running this often would cause a lot of load, but it would be good to get rid of as many variables as possible

### DIFF
--- a/corehq/preindex/tasks.py
+++ b/corehq/preindex/tasks.py
@@ -6,7 +6,7 @@ from corehq.util.decorators import serial_task
 from django.conf import settings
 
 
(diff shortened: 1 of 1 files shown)


benrudolph (approved):

snopoke (commented):
If we're only doing it at night then it's probably not necessary to run it every 5 mins.

emord:
@snopoke made it every 30 minutes. want to keep it up to date because I think some of the pillows change couch docs. (synclog and form submission metadata maybe?)

PR #16690, opened 2017-06-17
pick inline detail from same module as persistent case tiles

mkangia:
An extension to #16517

cc: buddy @biyeun @sheelio

### DIFF
--- a/corehq/apps/app_manager/tests/test_suite.py
+++ b/corehq/apps/app_manager/tests/test_suite.py
@@ -634,20 +634,23 @@ class SuiteTest(SimpleTestCase, TestXmlMixin, SuiteMixin):
 
         # configured to use pull down tile from the other module
         module1.case_details.short.pull_down_tile = True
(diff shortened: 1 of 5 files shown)


corehq/apps/app_manager/tests/test_suite.py
```diff
@@ -586,6 +586,74 @@ def test_case_tile_pull_down(self):
             "./entry/session"
         )
 
+    def test_inline_case_detail_from_another_module(self):
+        def ensure_module_session_datum_xml(detail_inline_attr, detail_persistent_attr):
```
snopoke: This is basically identical to the one in `test_persistent_case_tiles_from_another_module`. Why not make it a reusable function.

snopoke (changes requested):

corehq/apps/app_manager/tests/test_suite.py
mkangia: Extracting a common method.
They just had different arguments since other didnt need `detail_inline_attr` so just kept it under the particular test.

snopoke (approved):

PR #16691, opened 2017-06-17
fix js error for safari browser

mkangia:
from #16657

cc: @biyeun


orangejenny:
Oops. Didn't realize that errors in Safari, slightly surprised the linter doesn't catch it. Thanks, Manish!

mkangia:
> slightly surprised the linter doesn't catch it
@orangejenny No one cares about Safari much :)

PR #16693, opened 2017-06-19
Import Fixes

sravfeyn:
@nickpell 
Fixes reports-fixture missing from restore and adds a test.

There are some unknowns (see comments) but the issue gets resolved after these import arrangements

### DIFF
--- a/corehq/apps/reports_core/filters.py
+++ b/corehq/apps/reports_core/filters.py
@@ -4,9 +4,6 @@ from datetime import datetime, time
 from corehq.apps.locations.models import SQLLocation
 from corehq.apps.locations.util import load_locs_json, location_hierarchy_config
 from corehq.apps.reports_core.exceptions import FilterValueException
-from corehq.apps.userreports.expressions.getters import transform_from_datatype
-from corehq.apps.userreports.reports.filters.values import SHOW_ALL_CHOICE, CHOICE_DELIMITER, \
-    LocationDrilldownFilterValue
 from corehq.apps.userreports.util import localize
 from corehq.util.dates import iso_string_to_date, get_quarter_date_range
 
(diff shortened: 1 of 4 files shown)


custom/bihar/reports/indicators/indicators.py
```diff
@@ -275,6 +274,7 @@ def get_indicator_set(self, slug):
 class IndicatorSet(object):
 
     def __init__(self, spec):
+        from django.utils.datastructures import SortedDict
```
sravfeyn: This fix is not really related to the bug, but I am pretty sure, bihar fixture items are missing due to an import error here. So this fixes that issue (I am not sure if bihar is active or not)

I do not understand why this is an issue though, it just got fixed with this.

corehq/ex-submodules/casexml/apps/phone/fixtures.py
```diff
@@ -52,7 +52,7 @@ def __init__(self):
         if hasattr(settings, "FIXTURE_GENERATORS"):
             for group, func_paths in settings.FIXTURE_GENERATORS.items():
                 self._generator_providers[group] = filter(None, [
-                    to_function(func_path) for func_path in func_paths
+                    to_function(func_path, failhard=True) for func_path in func_paths
```
sravfeyn: Because `failhard` was set to False, `to_function('corehq.apps.app_manager.fixtures.report_fixture_generator')` was returning None and so this was missing from the restore. Same issue happens for `"custom.bihar.reports.indicators.fixtures.generator"`

nickpell (approved):
Nice, thanks for fixing!

corehq/apps/reports_core/filters.py
```diff
@@ -184,6 +181,7 @@ def year_param_name(self):
 
     @property
     def years(self):
+        from corehq.apps.userreports.reports.filters.values import SHOW_ALL_CHOICE
```
sravfeyn: These are bunch of circular import issues, as `userreports` module uses functions defined in `reports_core` module. This is the only place where `reports_core` uses functions defined in `userreports`

sravfeyn:
I am interested to dig a bit more to understand what's going on, I will come back to this latter though (I spent lot of time on this)

corehq/ex-submodules/casexml/apps/phone/fixtures.py
benrudolph: great call

benrudolph (approved):

PR #16695, opened 2017-06-19
Don't show blank form error for shadow forms

orangejenny:
Noticed that when a shadow form is missing a parent, it shows the missing parent error but also the blank form error, which is kind of a red herring:
screen shot 2017-06-19 at 9 58 52 am

Removed the second error.
screen shot 2017-06-19 at 9 58 59 am

@NoahCarnahan / @benrudolph


benrudolph (approved):

NoahCarnahan:
Thanks!

PR #16697, opened 2017-06-19
updates to location migration

nickpell:
Recommend 🐡 
@sravfeyn


custom/enikshay/private_sector_datamigration/management/commands/create_dtos_and_agencies.py
```diff
@@ -25,12 +25,13 @@ def handle(self, domain, state_code, district_code, parent_loc_id, user_level, o
         dto_parent = SQLLocation.active_objects.get(location_id=parent_loc_id)
         for org_id in org_ids:
             dto = self.create_dto(domain, state_code, district_code, dto_parent, org_id)
+            counter = 1
```
sravfeyn: Why not use `enumerate`?

nickpell: good call: https://github.com/dimagi/commcare-hq/pull/16751

PR #16700, opened 2017-06-19
label for max rows

benrudolph:
for icds

cc: @orangejenny

### DIFF
--- a/corehq/apps/export/templates/export/partial/export_list_controller.html
+++ b/corehq/apps/export/templates/export/partial/export_list_controller.html
@@ -8,7 +8,9 @@
      ng-cloak>
     {% if request|toggle_enabled:"PAGINATED_EXPORTS" %}
     <div class="alert alert-info">
-        These exports are paginated. That means exports that have over {{ max_exportable_rows|intcomma }} rows will be split into multiple files.
+        {% blocktrans with max_rows=max_exportable_rows|intcomma %}
+        These exports are paginated. That means exports that have over {{ max_rows }} rows will be split into multiple files.
+        {% endblocktrans %}
     </div>
     {% endif %}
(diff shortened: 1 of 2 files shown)


corehq/apps/export/templates/export/partial/export_list_controller.html
```diff
@@ -2,9 +2,17 @@
...
 {% load hq_shared_tags %}
 {% load djangular_tags %}
+{% load humanize %}
 
 <div ng-controller="ListExportsController"
      ng-cloak>
+    {% if request|toggle_enabled:"PAGINATED_EXPORTS" %}
+    <div class="alert alert-info">
+        {% blocktrans with max_rows=max_exportable_rows|intcomma %}
+        These exports are paginated. That means exports that have over {{ max_rows }} rows will be split into multiple files.
```
orangejenny: Should this be "over {{ max rows }} rows, *they* will be split"?

orangejenny:
Curious why this is behind a flag - are there domains that have exports this big but do need them to be in a single file?

benrudolph:
normally you can't export over 100k rows, and we don't want to make that a thing. however we have an exception for icds, and since it is greater than 100k rows we need to split it into multiple files

corehq/apps/export/templates/export/partial/export_list_controller.html
benrudolph: i think i'd have to change the beginning of the sentence for that, perhaps it doesn't read that clearly 😞

orangejenny: oh, I misread this, looking again it sounds fine

orangejenny (approved):

PR #16711, opened 2017-06-20
celery requires kombu app

sravfeyn:
@nickpell


gcapalbo:
@sravfeyn how come this is needed?

sravfeyn:
I was running into a migration error when I didn't have this app, after I had this app and ran migrations, it worked. (I had this from beginning, but I had to rediscover this when I resetup my dev env recently)

PR #16713, opened 2017-06-20
Test for #16685

orangejenny:
https://github.com/dimagi/commcare-hq/pull/16685#pullrequestreview-44750957

@sravfeyn / @benrudolph


benrudolph (approved):
test fixtures seem very large, assuming it has to be that way?

orangejenny:
Substantially easier that way, yeah. I don't love using app json files for tests, but I think it's a  pain to set up case tiles programatically.

PR #16715, opened 2017-06-20
Move prime restore tasks off of main queue

gcapalbo:
I wasn't sure if these should be moved to a different queue (like background_queue or async_restore_queue) or a new queue. For now this moves to a new one, but happy to change that if we can reuse another one we already have.

@calellowitz @snopoke @benrudolph


gcapalbo:
depends on https://github.com/dimagi/commcare-hq-deploy/pull/295 if we want to add a new one

benrudolph:
i think this in general is rarely used, as long as we think the overhead of running a new queue is worth it, i'm fine putting this in its own queue

benrudolph (approved):

PR #16720, opened 2017-06-20
Location dimension

benrudolph:
@gcapalbo @snopoke @calellowitz here's a first draft of the location dimension. still planning on writing tests, but wanted to see if you were still down with the flattened location structure

buddy: @orangejenny

can be reviewed all at once 🏠 

💩 is getting real


corehq/warehouse/models/staging.py
```diff
@@ -45,6 +48,70 @@ def clear_records(cls):
...
+        return []
+
+    @classmethod
+    def additional_sql_context(cls):
+        return {
+            'sqllocation_table': SQLLocation._meta.db_table
+        }
+
+
+class LocationTypeStagingTable(StagingTable, CustomSQLETLMixin):
```
benrudolph: I kept this table de-normalized on the LocationDim because I didn't think it was worth it to normalize. Open to push back on that though, my only real deciding factor was the fact that this table was small and it reduced some complexity

corehq/warehouse/transforms/sql/location_dim.sql
```diff
@@ -0,0 +1,126 @@
...
+    location_hierarchy.location_created_on,
+
+    now(),
+    now(),
+    false
+FROM
+    location_hierarchy
+INNER JOIN
+    {{ location_type_staging }} as lt_table
+ON location_hierarchy.site_code = lt_table.code AND location_hierarchy.domain = lt_table.domain
```
benrudolph: there might be a bug on this join, it's filtering out a lot of records for me locally. need to look into it

corehq/warehouse/models/staging.py
gcapalbo: ok, if it reduces complexity in other places then it's probably fine, i don't think we'll need to join to it explicitly for anything

corehq/warehouse/models/dimensions.py
```diff
@@ -98,26 +100,40 @@ class LocationDim(BaseDim, CustomSQLETLMixin):
...
+    # then this should have levels 0, 1, 2, 3 populated. Each level contains
+    # the sql id of the location.
+    location_level_0 = models.IntegerField()
+    location_level_1 = models.IntegerField(null=True)
+    location_level_2 = models.IntegerField(null=True)
+    location_level_3 = models.IntegerField(null=True)
+    location_level_4 = models.IntegerField(null=True)
+    location_level_5 = models.IntegerField(null=True)
+    location_level_6 = models.IntegerField(null=True)
+    location_level_7 = models.IntegerField(null=True)
```
gcapalbo: i think it would be good to index these fields if we'll be joining on them

corehq/warehouse/tests/test_dimensions.py
```diff
@@ -90,3 +94,47 @@ def test_user_types(self):
...
+class TestLocationDim(TestCase):
+
+    domain = 'location-dim-test'
+
+    @classmethod
+    def teardDownClass(cls):
+        LocationStagingTable.clear_records()
+        LocationTypeStagingTable.clear_records()
+        LocationDim.clear_records()
+        super(TestLocationDim, cls).teardDownClass()
```
gcapalbo: `tearDownClass`

corehq/warehouse/tests/test_dimensions.py
```diff
@@ -90,3 +94,47 @@ def test_user_types(self):
...
             'beeboobop',
         )
+
+
+class TestLocationDim(TestCase):
+
+    domain = 'location-dim-test'
+
+    @classmethod
+    def teardDownClass(cls):
```
gcapalbo: `tearDownClass`

corehq/warehouse/transforms/sql/location_dim.sql
```diff
@@ -0,0 +1,130 @@
...
+    location_type_name,
+    location_type_code,
+    location_type_id,
+
+    location_last_modified,
+    location_created_on,
+
+    dim_last_modified,
+    dim_created_on,
+    -- TODO: Figure out how to handle deletes when we actually hard deleted the model
```
gcapalbo: yeah this is also complicated by the fact that there might be form submissions or cases (with owner_id) pointing to that location id

corehq/warehouse/transforms/sql/location_dim.sql
```diff
@@ -0,0 +1,130 @@
...
+    deleted
+)
+WITH RECURSIVE location_hierarchy(
```
gcapalbo: interesting, i've never seen this syntax. when you had asked about a recursive query i thought you mean just joining the table to itself.

also does this work when we pull in a child location to the staging table but not its parent if the parent wasn't modified in the window?

gcapalbo: curious if you think it might be easier to first insert/update the dim with all the single location's info, and then update all the foreign keys later. i think we could use one update query to refresh all the foreign key ids using an update...from query since the table should be small enough.

corehq/warehouse/etl.py
```diff
@@ -54,8 +62,9 @@ def _table_context(cls, start_datetime, end_datetime):
         for dep in cls.dependencies():
             dep_cls = get_cls_by_slug(dep)
             context[dep] = dep_cls._meta.db_table
-        context['start_datetime'] = start_datetime
-        context['end_datetime'] = start_datetime
+        context['start_datetime'] = start_datetime.isoformat()
```
gcapalbo: does this format work when comparing?

benrudolph: this does work, however i can see where it'd be confusing since the text doesn't look exactly the same as the text that is stored in postgres. my guess is pg auto converts it. do you have a better practice?

screen shot 2017-06-22 at 10 41 23 am
screen shot 2017-06-22 at 10 41 30 am

corehq/warehouse/tests/test_dimensions.py
benrudolph: :man_facepalming:

corehq/warehouse/transforms/sql/location_dim.sql
benrudolph: > i thought you mean just joining the table to itself.

ah how is that possible without some sort of recursive syntax? you won't know how many joins to do? or are you suggesting just join 8 times?

> also does this work when we pull in a child location to the staging table but not its parent if the parent wasn't modified in the window?

going to play around with this today. in regards to doing foreign keys in a separate query, it might be, though i would imagine the query to update the foreign keys would look similar to the current one without all the meta data. will experiment

corehq/warehouse/etl.py
gcapalbo: ok, no that's fine. wasn't sure if the T throws it off, but as long as it works i think it's fine

corehq/warehouse/transforms/sql/location_dim.sql
gcapalbo: (offlined this)

benrudolph:
Ok refactored the SQL script to properly update the dimension when new data comes in. Added a test that specifically tests for that as well

corehq/warehouse/tests/test_dimensions.py
```diff
@@ -90,3 +95,85 @@ def test_user_types(self):
...
+
+        self.assertEqual(
+            home_location.location_level_2,
+            LocationDim.objects.filter(name='Illinois').first().sql_location_id,
+        )
+        self.assertEqual(
+            home_location.location_level_1,
+            LocationDim.objects.filter(name='Naperville').first().sql_location_id,
+        )
+        self.assertEqual(home_location.location_level_0, home_location.sql_location_id)
```
snopoke: Interesting, I had assumed the levels would count from the root down so level_0 would always be the root node and location_N would be the leaf node.

With it like this are you able to have a query that get's all locations below a specific location?

benrudolph: i think the query would essentially look something like this:
```
select * from location_dim
where
location_level_0 = <location_id> OR
location_level_1 = <location_id>
...
```
i know this uses `OR` syntax, but i don't actually imagine this being the bottle neck of a query

corehq/warehouse/transforms/sql/location_dim.sql
```diff
@@ -34,97 +28,72 @@ INSERT INTO {{ location_dim }} (
...
+LEFT JOIN {{ location_dim }} l1 ON l0.location_level_1 = l1.location_level_0
+LEFT JOIN {{ location_dim }} l2 ON l1.location_level_1 = l2.location_level_0
+LEFT JOIN {{ location_dim }} l3 ON l2.location_level_1 = l3.location_level_0
+LEFT JOIN {{ location_dim }} l4 ON l3.location_level_1 = l4.location_level_0
+LEFT JOIN {{ location_dim }} l5 ON l4.location_level_1 = l5.location_level_0
+LEFT JOIN {{ location_dim }} l6 ON l5.location_level_1 = l6.location_level_0
+LEFT JOIN {{ location_dim }} l7 ON l6.location_level_1 = l7.location_level_0
+
+WHERE
+    location_dim_target.sql_location_id = l0.sql_location_id;
```
gcapalbo: :+1: this query looks nice

corehq/warehouse/transforms/sql/location_dim.sql
```diff
@@ -0,0 +1,97 @@
+-- First insert all records into location dim from staging tables without foreign keys
+INSERT INTO {{ location_dim }} (
```
gcapalbo: we'll need to handle updates here as well eventually, or is that coming for all dims in a later pr

gcapalbo (approved):

corehq/warehouse/transforms/sql/location_dim.sql
benrudolph: yeah coming in a later PR. right now it's blocking on `UPSERT`. i think it'll be fairly easy to convert

corehq/warehouse/tests/test_dimensions.py
snopoke: Is there an advantage to having the hierarchy this way round? If the hierarchy was inverted we could just do a query like this:
```
select * from locatoin_dim where location_level_N = <location_id>;
```
It would mean we'd need to know what level the location with id <location_id> is.

gcapalbo: don't we still have that ability? we just calculate the N in location_level_N a different way now before we write our select query

i think it's easier to populate from the update query this way

benrudolph: agree this is easier to update the table. so you're saying that if i wanted to find the root node, in the original way it'd just be `location_level_0 = <location_id>`. in the current model, that `N` is different for every location. it'd have to be `location_level_(current_location_level - N) = <location_id>`. i'm not positive we can create a dynamic column like that, but if we can it should work

snopoke: perhaps we should go to voice. Just so we're on the same page I threw together this sheet: https://docs.google.com/spreadsheets/d/1tlLomOjk0r1GoTqlVMwu0pp46yqNKm6R7uEU4BuO1nM/edit#gid=0

gcapalbo: i see what you mean now @snopoke , yeah that is not ideal for querying then. maybe we can add a step which flips them around

corehq/warehouse/transforms/sql/location_dim.sql
gcapalbo: does upsert require upgrade? i think we can also get around using upsert (and doing an upgrade everywhere) by just using some combination of `update from...` and `delete from using...` queries

benrudolph: upsert does require an upgrade. yea i think we can definitely work around if we don't want to upgrade. thought it was a good excuse to bump our postgres version, but may be more hassle than it's worth, especially on icds

PR #16722, opened 2017-06-21
only update user device once per day

snopoke:
Waiting on confirmation from ICDS and eNikshay project teams


benrudolph (approved):
cool, looks good to me pending approval from project teams

snopoke:
@benrudolph  :white_check_mark: a few changes since last review

corehq/apps/users/models.py
```diff
@@ -2088,19 +2088,25 @@ def get_usercase_id(self):
...
+
+        :returns: True if user was updated and needs to be saved
         """
         when = when or datetime.utcnow()
+
         for user_device_id_last_used in self.devices:
             if user_device_id_last_used.device_id == device_id:
-                user_device_id_last_used.last_used = when
-                break
+                if when.date() > user_device_id_last_used.last_used.date():
```
calellowitz: What is the advantage of doing this over storing just the date and using `date.today()` (keeping everything in dates instead of always converting from datetimes)

calellowitz (commented):
looks good. one question but its not blocking

corehq/apps/users/models.py
snopoke: the only thing I can think of is that this doesn't require a data model change.

I agree that storing it as a date makes more sense. I'll look at updating it.

snopoke: I think let's get this in as is since the question of only storing the date may take longer to validate with the team

PR #16305, opened 2017-05-24
Deprecate ReportGraphConfig

orangejenny:
Followup for https://github.com/dimagi/commcare-hq/pull/16031

Do not merge until migration is run on prod, softlayer, and NIC.

@esoergel / @nickpell

PR #16727, opened 2017-06-21
500 when creating new report module

orangejenny:
https://manage.dimagi.com/default.asp?256634

Introduced in https://github.com/dimagi/commcare-hq/pull/16305

@benrudolph


benrudolph (approved):
wow! didn't realize this had been live for so long. thanks for the quick fix. this unblocks the l10k team

PR #16733, opened 2017-06-21
Replace ignore patterns in custom query variables

proteusvacuum:
Removes "ignore patterns" from user generated input in custom queries.
We need this for eNikshay where we are doing some funky "or" search clause against an id case property that we also want to remove certain characters from (namely, the `-` character). 

First commit is a refactor that moves the compilation of the case search es query into its own class. 
Second commit adds the extra functionality. 

@snopoke


snopoke:
Would be good to update the doc string.

PR #16735, opened 2017-06-22
only reverse indices once

calellowitz:
for the tests

### DIFF
--- a/corehq/ex-submodules/casexml/apps/phone/models.py
+++ b/corehq/ex-submodules/casexml/apps/phone/models.py
@@ -771,7 +771,7 @@ class SimplifiedSyncLog(AbstractSyncLog):
     def primary_case_ids(self):
         return self.case_ids_on_phone - self.dependent_case_ids_on_phone
 
-    def purge(self, case_id, quiet_errors=False):
+    def purge(self, case_id, quiet_errors=False, cached_child_map=None, cached_extension_map=None):
         """
         This happens in 3 phases, and recursively tries to purge outgoing indices of purged cases.
         Definitions:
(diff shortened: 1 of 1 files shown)


calellowitz:
@proteusvacuum cc: @ctsims 
This should fix the purge algorithm to not take forever by stopping it from constantly recalculating the reversed indices.

ctsims:
Structurally this approach looks correct to me, but am probably not the best person for a deep review.

corehq/ex-submodules/casexml/apps/phone/models.py
```diff
@@ -750,6 +750,14 @@ class SimplifiedSyncLog(AbstractSyncLog):
     closed_cases = SetProperty(unicode)
     extensions_checked = BooleanProperty(default=False)
 
+    _purged_cases = set()
```
millerdev: Is this meant to be a class-level mutable set? Would it work to set this to `None` instead so the class attribute is immutable? Seems like bad things would happen if this set instance somehow got an item added to it, and as far as I can tell this instance is never used.

millerdev (commented):
Would like to see the class attribute changed to an immutable value, but if this is really urgent it appears that it will work as is, so could be merged.

corehq/ex-submodules/casexml/apps/phone/models.py
```diff
@@ -750,6 +750,14 @@ class SimplifiedSyncLog(AbstractSyncLog):
     closed_cases = SetProperty(unicode)
     extensions_checked = BooleanProperty(default=False)
 
+    _purged_cases = None
+
+    @property
+    def purged_cases(self):
+        if not self._purged_cases:
```
millerdev: nitpick: could now change this to `if self._purged_cases is None:` so it does not continually create a new `set` each time `self.purged_cases` is accessed while the set is empty.

millerdev (approved):

PR #16736, opened 2017-06-22
soft asserts for form ordering bug

snopoke:
These updates should help validate that the form ordering bug isn't happening in CC > 2.35. They'll also help to identify any 'missing' form issues.

### DIFF
--- a/corehq/ex-submodules/casexml/apps/case/xform.py
+++ b/corehq/ex-submodules/casexml/apps/case/xform.py
@@ -14,6 +14,7 @@ from corehq.toggles import EXTENSION_CASES_SYNC_ENABLED
 from corehq.apps.users.util import SYSTEM_USER_ID
 from corehq.form_processor.interfaces.processor import FormProcessorInterface
 from corehq.form_processor.interfaces.dbaccessors import CaseAccessors
+from corehq.util.soft_assert import soft_assert
 from couchforms.models import XFormInstance
 from casexml.apps.case.exceptions import (
     NoDomainProvided,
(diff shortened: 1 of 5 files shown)


corehq/ex-submodules/casexml/apps/case/xform.py
```diff
@@ -264,6 +266,18 @@ def _validate_indices(case_db, cases):
...
+                    from distutils.version import LooseVersion
+                    if case_db.cached_xforms:
+                        xform = case_db.cached_xforms[0]
+                        if xform.metadata and xform.metadata.commcare_version:
+                            commcare_version = xform.metadata.commcare_version
+                            _soft_assert(
+                                commcare_version < LooseVersion("2.35"),
+                                "Invalid Case Index in CC version >= 2.35", {
+                                    'domain': case_db.domain,
+                                    'version': str(commcare_version)
```
amstone326: Could include the form id here like you are below, and maybe also the ID of the index that was invalid, in case that's helpful for investigating?

amstone326 (changes requested):
Thanks for adding this! Few thoughts

corehq/form_processor/casedb_base.py
```diff
@@ -150,12 +150,19 @@ def get_case_from_case_update(self, case_update, xform):
...
-                    'case_id': case_update.id,
-                    'domain': xform.domain
-                }
-            )
+            if xform.metadata and xform.metadata.commcare_version:
+                from distutils.version import LooseVersion
+                if xform.metadata and xform.metadata.commcare_version:
+                    commcare_version = xform.metadata.commcare_version
+                    _soft_assert(
+                        case_update.creates_case() or commcare_version < LooseVersion("2.35"),
```
snopoke: This asserts that the condition is True, if not it will notify. So this says don't notify if case is being created or version < 2.35.

proteusvacuum:
Oops, merged this before seeing @amstone326's comments. Sounds like none of them were blocking, but would be good to address.

PR #16737, opened 2017-06-22
Data File Downloads

kaapstorm:
This PR is for HSPH/FIGO Data Transfer. See [the spec][1] for more context. TLDR? HSPH would like to clean and anonymize exported form data, upload it, and allow other users to download it. This is a simple page to allow admin users upload files, and allow other users to download them.

There is a little left to do, but I'm opening this PR just to confirm some choices I've made. See comments below.

Here is what an admin user sees:
figo_admin_user

This is what a lowly web user sees:
figo_web_user

And this is the Default Landing setting for a web user's role to send them to the right place.
figo_role_config

@gcapalbo cc @snopoke 

  [1]: https://docs.google.com/document/d/1P91t9sdJ9-i8mppCCcw2poDJoh6LviLpIWFHRybvJ0A/edit


corehq/apps/export/views.py
```diff
@@ -1530,11 +1530,9 @@ class DataFileDownloadList(BaseProjectDataView):
 
     def get_context_data(self, **kwargs):
         context = super(DataFileDownloadList, self).get_context_data(**kwargs)
-        perms = self.request.user.get_all_permissions()
         context.update({
             'data_files': DataFile.objects.filter(domain=self.domain).order_by('filename').all(),
-            'can_add': 'export.add_datafile' in perms,
-            'can_delete': 'export.delete_datafile' in perms,
+            'is_admin': self.request.couch_user.is_domain_admin(self.domain),
```
kaapstorm: I just want to confirm that this is the right way to go about this. I realised my first approach wasn't used anywhere else, and this should do what I need.

corehq/apps/export/views.py
```diff
@@ -1521,6 +1522,58 @@ def get_saved_exports(self):
...
+        context.update({
+            'data_files': DataFile.objects.filter(domain=self.domain).order_by('filename').all(),
+            'can_add': 'export.add_datafile' in perms,
+            'can_delete': 'export.delete_datafile' in perms,
+        })
+        return context
+
+    def post(self, request, *args, **kwargs):
+        # TODO: Use a form for validation
+        # TODO: Limit file size
```
kaapstorm: Any suggestions on how big we should allow uploaded files to be?

corehq/apps/export/templates/export/download_data_files.html
```diff
@@ -0,0 +1,130 @@
...
+                                        if (copied) {
+                                            alert_user('{% trans 'Data file URL copied to clipboard.' %}', 'success', true);
+                                        } else {
+                                            showCopyDialog();
+                                        }
+                                    } catch (err) {
+                                        // Older browsers
+                                        showCopyDialog();
+                                    }
+                                }())">
```
benrudolph: would be nice to extract this into a separate JS file if you have a chance. same with css

corehq/apps/export/views.py
benrudolph: im guessing we should look at the hsph domain and see what their requirements are. can't imagine this being larger than 100MB or so, though hsph never ceases to surprise me

corehq/apps/export/views.py
```diff
@@ -1521,6 +1521,56 @@ def get_saved_exports(self):
...
+
+
+class DataFileDownloadDetail(BaseProjectDataView):
+    urlname = 'download_data_file'
+
+    def get(self, request, *args, **kwargs):
+        try:
+            data_file = DataFile.objects.filter(domain=self.domain).get(pk=kwargs['pk'])
+            blob = data_file.get_blob()
+            response = HttpResponse(blob.read(), content_type=data_file.content_type)
```
benrudolph: do you think we could stream this?

corehq/apps/export/models/new.py
```diff
@@ -2470,6 +2477,48 @@ def user_to_be_notified(cls, domain, user):
...
+            blob = db.get(self.blob_id)
+        except (KeyError, NotFound) as err:
+            raise self.DoesNotExist(str(err))
+        return blob
+
+    def save_blob(self, uploaded_file):
+        content_string = StringIO()
+        for chunk in uploaded_file.chunks():
+            content_string.write(chunk)
+        self.content_type = magic.from_buffer(content_string.read(1024))
```
benrudolph: cool, that is magic

benrudolph (commented):
cool looks good, only a couple comments

corehq/apps/export/models/new.py
```diff
@@ -2470,6 +2477,48 @@ def user_to_be_notified(cls, domain, user):
...
+
+    class Meta(object):
+        app_label = 'export'
+
+    def get_blob(self):
+        db = get_blob_db()
+        try:
+            blob = db.get(self.blob_id)
+        except (KeyError, NotFound) as err:
+            raise self.DoesNotExist(str(err))
```
gcapalbo: might be better to raise a different exception otherwise it might get confused with when a `DataFile` record itself doesn't exist

corehq/apps/export/templates/export/download_data_files.html
```diff
@@ -0,0 +1,130 @@
...
+            <tr id="row_{{ data_file.id }}">
+                <td>
+                    <a href="{% url 'download_data_file' domain data_file.id data_file.filename %}">
+                        <i class="fa fa-file text-primary"></i>
+                        {{ data_file.filename }}
+                    </a>
+                </td>
+                <td>
+                    {{ data_file.description }}
+                    <textarea id="url_{{ data_file.id }}" style="display: none">{% if request.is_secure %}https://{% else %}http://{% endif %}{{ request.get_host }}{% url 'download_data_file' domain data_file.id data_file.filename %}</textarea>
```
gcapalbo: i think we usually use `dimagi.utils.web.get_url_base` for this unless this is different

corehq/apps/export/views.py
gcapalbo: :+1: looks good

corehq/apps/export/views.py
kaapstorm: Great idea!

kaapstorm:
Thanks for the feedback @gcapalbo and @benrudolph.

I went with a 100MB limit, pending Meryn getting back to me with a bigger number.

I had a "TODO" to use a form for validation, but decided against it because it would just be boilerplate for what actually just took an `if` clause in the view.

benrudolph (approved):

corehq/apps/export/models/new.py
```diff
@@ -2470,6 +2477,48 @@ def user_to_be_notified(cls, domain, user):
...
+        try:
+            blob = db.get(self.blob_id)
+        except (KeyError, NotFound) as err:
+            raise NotFound(str(err))
+        return blob
+
+    def save_blob(self, uploaded_file):
+        content_string = StringIO()
+        for chunk in uploaded_file.chunks():
+            content_string.write(chunk)
```
snopoke: Is it necessary to read the whole thing into a StringIO?

corehq/apps/export/models/new.py
```diff
@@ -2470,6 +2477,48 @@ def user_to_be_notified(cls, domain, user):
...
+        self.content_type = magic.from_buffer(content_string.read(1024))
```
snopoke: Since this is already set from the request, what's the need to check it here, especially if we aren't going to validate it at all.

snopoke (commented):
Looking great. 2 small comments. 

Also I think we should add a limit on how many a files a domain can have - or maybe the max data size a domain can have - or both. I know we don't need that now but if we ever release it we'll need to add it.

corehq/apps/export/models/new.py
kaapstorm: The content type that we get from the request is whatever the user's browser tells us it is. I'm going with "trust but verify" here by checking the file ourselves and updating the value. But you make an important point -- if we don't use the value, does it matter what the content type is? I'm undecided but I'll look up what can go wrong, and consider the fact that at the moment only admins can upload files. It will be nice to be able to take this out.

kaapstorm: I also wanted to mention why I don't validate the file type against a set of acceptable document types. When I looked at some of the possible values of `content_type` I realised how fine-grained it can be -- like the file type given for a "docx" file created by LibreOffice is different from the file type of a "docx" file created by MS Office. I realised it was going to be a support issue if we didn't have all possible permutations, and getting that list might be hard. So I went with Ben's suggestion and just limited the file size.

(It gets even trickier when it comes to zip files, because you have to choose not to accept zip files, which users might not like, or to accept all zip files, so you can effectively upload anything you want (so why bother checking content types at all (this is the option WhatsApp went with, btw)) ... or you inspect inside unencrypted zip files, in which case you're doubling the number of acceptable content types ... and then what do you do with encrypted zip files (which is what a password-protected docx file is)? ... A rabbit hole I didn't to fall down.)

snopoke:
Thanks, I figured that was the case and I'm fine with not validating the
content type against a whitelist.

kaapstorm:
Maximum space allowance is very wise. And I think trusting the browser to set the content type should be fine -- it makes no sense that an admin would try to mess with their own web users ... by manipulating the content type of their data files?! :stuck_out_tongue_closed_eyes: 

Limiting the number of files might also be a good idea. The current limitation would just be poor UX because the list of data files isn't paginated. That's not a great approach. If we're looking at more than 20 or 30 I think we should paginate and have a max number.

corehq/apps/export/views.py
```diff
@@ -1546,11 +1547,23 @@ def post(self, request, *args, **kwargs):
             )
             return self.get(request, *args, **kwargs)
 
+        file_size_total = (
+            sum(f.content_length for f in DataFile.objects.filter(domain=self.domain).all()) +
```
snopoke: You can do this in SQL: https://docs.djangoproject.com/en/1.11/topics/db/aggregation/#cheat-sheet

corehq/apps/export/views.py
```diff
@@ -1546,11 +1547,23 @@ def post(self, request, *args, **kwargs):
...
+            request.FILES['file'].size
+        )
+        if file_size_total > MAX_DATA_FILE_SIZE_TOTAL:
+            messages.warning(
+                request,
+                _('Uploading this data file would exceed the total allowance of {} GB for this project space'
```
snopoke: add `Please remove some files in order to upload new files.`

snopoke (changes requested):

corehq/apps/export/views.py
kaapstorm: Doh. Thank you.

snopoke (approved):

PR #16738, opened 2017-06-22
Fixes in Beneficiary Register report

kkrampa:
* Add logic for date_of_registration
* Set Treating Provider Name to person owner name
* date_of_diagnosis2 -> date_of_diagnosis


kkrampa:
@NoahCarnahan

PR #16740, opened 2017-06-22
Select2 not returning expected results

orangejenny:
https://manage.dimagi.com/default.asp?256667

Seems that the results text itself can't be truncated without interfering with search.

Introduced in https://github.com/dimagi/commcare-hq/pull/16644

@benrudolph / @biyeun


biyeun:
😱

biyeun:
now we're really going to have wrapping issues. UGH this select2 widget.

orangejenny:
> now we're really going to have wrapping issues

?? This shouldn't affect how the widget is displayed, just the search.

biyeun:
ohhhhh i see what you did here. ok cool

PR #16742, opened 2017-06-22
add ability to configure sort_nodeset_fields

mkangia:
https://manage.dimagi.com/default.asp?255102

Good to review 🏛 . 

"Hold off on merge" for any feedback on changes. Deployed on staging.
cc: buddy @biyeun @sheelio

### DIFF
--- a/corehq/apps/app_manager/detail_screen.py
+++ b/corehq/apps/app_manager/detail_screen.py
@@ -101,7 +101,7 @@ class FormattedDetailColumn(object):
         self.id_strings = id_strings
         self.parent_tab_nodeset = parent_tab_nodeset
 
-    def add_sort_node_for_nodeset_field(self):
+    def has_sort_node_for_nodeset_column(self):
         return False
 
     @property
@@ -138,7 +138,7 @@ class FormattedDetailColumn(object):
     @property
     def sort_node(self):
         if not (self.app.enable_multi_sort and
-                (self.detail.display == 'short' or self.add_sort_node_for_nodeset_field())
+                (self.detail.display == 'short' or self.has_sort_node_for_nodeset_column())
                 ):
             return
 
@@ -262,7 +262,7 @@ class HideShortHeaderColumn(FormattedDetailColumn):
 
     @property
     def header(self):
-        if self.detail.display == 'short' or self.add_sort_node_for_nodeset_field():
+        if self.detail.display == 'short' or self.has_sort_node_for_nodeset_column():
             header = sx.Header(
                 text=sx.Text(),
                 width=self.template_width
@@ -276,7 +276,7 @@ class HideShortColumn(HideShortHeaderColumn):
 
     @property
     def template_width(self):
-        if self.detail.display == 'short' or self.add_sort_node_for_nodeset_field():
+        if self.detail.display == 'short' or self.has_sort_node_for_nodeset_column():
             return 0
 
 
@@ -441,7 +441,7 @@ class LateFlag(HideShortHeaderColumn):
 @register_format_type('invisible')
 class Invisible(HideShortColumn):
 
-    def add_sort_node_for_nodeset_field(self):
+    def has_sort_node_for_nodeset_column(self):
         return self.parent_tab_nodeset and self.detail.sort_nodeset_columns_for_detail()
 
     @property
--- a/corehq/apps/app_manager/detail_screen.py
+++ b/corehq/apps/app_manager/detail_screen.py
@@ -137,9 +137,8 @@ class FormattedDetailColumn(object):
 
     @property
     def sort_node(self):
-        if not (self.app.enable_multi_sort and (
-                    self.detail.display == 'short' or
-                    self.add_sort_node_for_nodeset_field())
+        if not (self.app.enable_multi_sort and
+                (self.detail.display == 'short' or self.add_sort_node_for_nodeset_field())
                 ):
             return
 
--- a/corehq/apps/app_manager/detail_screen.py
+++ b/corehq/apps/app_manager/detail_screen.py
@@ -1,7 +1,7 @@
 from corehq.apps.app_manager import id_strings
 from corehq.apps.app_manager.suite_xml import xml_models as sx
 from corehq.apps.app_manager.suite_xml import const
-from corehq.apps.app_manager.util import sort_nodeset_fields_for_detail
+from corehq.apps.app_manager.util import sort_nodeset_columns_for_detail
 from corehq.apps.app_manager.xpath import (
     CaseXPath,
     CommCareSession,
@@ -443,7 +443,7 @@ class LateFlag(HideShortHeaderColumn):
 class Invisible(HideShortColumn):
 
     def add_sort_node_for_nodeset_field(self):
-        return self.parent_tab_nodeset and sort_nodeset_fields_for_detail(self.detail_type, self.detail)
+        return self.parent_tab_nodeset and sort_nodeset_columns_for_detail(self.detail_type, self.detail)
 
     @property
     def header(self):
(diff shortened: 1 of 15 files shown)


biyeun:
looks ok but flagging @orangejenny too

corehq/apps/app_manager/templates/app_manager/v1/partials/case_list.html
```diff
@@ -116,6 +116,16 @@
 
 {% include 'app_manager/v1/partials/case_list_filtering.html' %}
 
+{{ request|toggle_tag_info:"DETAIL_LIST_TAB_NODESETS" }}
+{% if request|toggle_enabled:'DETAIL_LIST_TAB_NODESETS' %}
+<div data-bind="with: shortScreen">
+    <input data-bind="checked: sortNodesetColumns" type="checkbox">
+    <span>
+        {% trans "Sort entries in the Nodeset tabs" %}
```
orangejenny: Minor, but it'd be a ux benefit to make this a label so users can click on the text to check the checkbox. [This](http://getbootstrap.com/css/#forms-controls) has the typical markup for a label+checkbox in bootstrap.

corehq/apps/app_manager/templates/app_manager/v1/partials/case_list.html
```diff
@@ -116,6 +116,16 @@
 
 {% include 'app_manager/v1/partials/case_list_filtering.html' %}
 
+{{ request|toggle_tag_info:"DETAIL_LIST_TAB_NODESETS" }}
+{% if request|toggle_enabled:'DETAIL_LIST_TAB_NODESETS' %}
+<div data-bind="with: shortScreen">
```
orangejenny: Surprised to see this in the case list tab, I'd imagined adding a "Sorting" section to the case detail and putting it there, since it affects long details not short ones. What would you think of moving it?

corehq/apps/app_manager/detail_screen.py
```diff
@@ -98,6 +100,10 @@ def __init__(self, app, module, detail, column, sort_element=None,
         self.sort_element = sort_element
         self.order = order
         self.id_strings = id_strings
+        self.parent_tab_nodeset = parent_tab_nodeset
+
+    def add_sort_node_for_nodeset_field(self):
```
orangejenny: Minor, but the name here confused me, `add` sounds like this function alters the column. Maybe `should_sort_nodeset_detail_tab` or `has_sort_node_for_nodeset_detail`?

corehq/apps/app_manager/templates/app_manager/v1/partials/case_list.html
orangejenny: This would also be a good place for a help icon, to remind people how this works (that it sorts by the first column, and in what order it sorts).

corehq/apps/app_manager/models.py
```diff
@@ -2030,6 +2030,10 @@ def from_json(cls, data):
                     item.value[lang] = interpolate_media_path(path)
         return to_ret
 
+    @property
```
orangejenny: I don't think this is adding much value, since we already expose the string values for detail column formats outside of models.py (app_strings.py and detail-screen-config.js use them).

orangejenny:
@mkangia Could you summarize what the final functionality of this is? [This](https://github.com/dimagi/commcare-hq/pull/16742/files#diff-5fb1ec6df6f30bf5f8223273482f0704R365) looks like it'll be sorted by any search-only fields in the detail, but I thought there was also going to be something about sorting by the first column.

This would also be good info to have in a docstring, probably in `get_nodeset_sort_elements`.

corehq/apps/app_manager/util.py
```diff
@@ -572,3 +572,11 @@ def get_form_data(domain, app):
         module_meta['forms'] = forms
         modules.append(module_meta)
     return modules, errors
+
+
+def sort_nodeset_columns_for_detail(detail_type, detail):
```
orangejenny: Could this be a method on the detail class? It's pretty specific to be in a utils file.

corehq/apps/app_manager/suite_xml/sections/details.py
```diff
@@ -358,6 +362,23 @@ def get_detail_id_safe(self, module, detail_type):
         return detail_id if detail_id in self.active_details else None
 
 
+def get_nodeset_sort_elements(detail):
+    from corehq.apps.app_manager.models import SortElement
+    sort_elements = []
+    tab_spans = detail.get_tab_spans()
+    for tab in detail.get_tabs():
```
orangejenny: This behaves correctly if there are multiple nodeset tabs, right? (I don't know the sorting code well.)

corehq/apps/app_manager/models.py
mkangia: it usually prefer having such checks as methods on objects instead of doing the string comparison at various places. So in case the value is to be changed someday, it needs to be done only at one place. For ex here if 'invisible' is changed to 'hidden'.
This made more sense in ruby where method name would actually be 'invisible?' but i think it still holds value here.

mkangia:
> @mkangia Could you summarize what the final functionality of this is? This looks like it'll sort by any search-only fields in the detail, but I thought there was also going to be something about sorting by the first column.

@orangejenny it will sort for any 'Search Only' fields added under nodeset. The check for nodeset is when it is called [here](https://github.com/dimagi/commcare-hq/pull/16742/files/eeb94c9b2e30b822be622397dc29a581f9e739bd#diff-ae61842a4e76ffee74843e118b718a68R446) and not in `get_nodeset_sort_elements` since we don't have nodeset available in that method.

From what i saw in the code base, Sorting by the first column happens in case there are no sort columns by [this method](https://github.com/dimagi/commcare-hq/pull/16742/files/eeb94c9b2e30b822be622397dc29a581f9e739bd#diff-5fb1ec6df6f30bf5f8223273482f0704L361). I don't think we need that here.

corehq/apps/app_manager/suite_xml/sections/details.py
mkangia: Yes it does. I did test it. Why do you doubt it?

orangejenny: Was just asking, I like to double-check things :)  Happy you tested it.

mkangia: Oh yes.. no problem.. wanted to just check if something didn't look right. :)

corehq/apps/app_manager/models.py
orangejenny: Okay, this is basically a stylistic difference of opinion, works for me.

orangejenny:
@mkangia cool, thanks for the explanation

corehq/apps/app_manager/templates/app_manager/v1/partials/case_list.html
mkangia: i moved this under Case Details. The changes in v1 and v2 are slightly different. Had to add a spacer div under v2 and may be we have discontinued using that for spacing now.

corehq/apps/app_manager/detail_screen.py
mkangia: I used "add" because it decides if sort node is to be "added" for the column under nodeset. In column under nodeset we would want to add sort node if needed. So its not for the tab or the detail but for the column under it, hence referring to tab or nodeset detail in that name does not look right.
Can use `add_sort_node_for_nodeset_column` or `has_sort_node_for_nodeset_column` if that makes it better.

mkangia:
@orangejenny All changes done. Good for review.

corehq/apps/app_manager/detail_screen.py
orangejenny: `has_sort_node_for_nodeset_column` sounds good to me

orangejenny (approved):
Looks good, thanks for polishing up the UI!

PR #16745, opened 2017-06-22
App manager v2: reload page on rearrange forms/modules

orangejenny:
https://manage.dimagi.com/default.asp?256730

For once I'm unhappy to be deleting code. Rearranging forms/modules without refreshing the page is a better experience. But there are quite a few urls that use the index-based id, and missing updating one of them has rather severe consequences (overwriting a module/form's settings).

@benrudolph / @biyeun

### DIFF
--- a/corehq/apps/app_manager/static/app_manager/js/app_manager.js
+++ b/corehq/apps/app_manager/static/app_manager/js/app_manager.js
@@ -276,40 +276,6 @@ hqDefine('app_manager/js/app_manager.js', function () {
                 $(related).data(name, value);
             });
         }
-        function resetIndexes($sortable) {
-            if (COMMCAREHQ.toggleEnabled('APP_MANAGER_V2')) {
-                var parentVar = $sortable.data('parentvar');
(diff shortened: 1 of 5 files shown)


benrudolph (commented):
oh man that is a big bummer. code looks fine. is it easy to detect when things aren't going to work?

orangejenny:
@benrudolph The real fix would be to update the various URLs that take `module_id` and instead pass in unique module id. It's not a huge task; I'm admittedly being kind of lazy.

biyeun:
@orangejenny I can take that on over the weekend.

orangejenny:
@biyeun okay, cool, will leave this open for now in case you have other stuff come up, but won't merge

orangejenny:
Biyeun's fix has turned out to be more involved than expected, I'd like to deploy this in the interim, will revert when the more robust fix is ready.

new code buddies @kaapstorm @gcapalbo

biyeun:
@orangejenny noted. thanks.

PR #16741, opened 2017-06-22
app manager v2: update page title when updating form name

orangejenny:
https://manage.dimagi.com/default.asp?256330

Debated whether to change `updateDOM` instead of updating `document.title` all over the place, but don't think it's significantly cleaner.

@benrudolph / @biyeun

PR #16746, opened 2017-06-22
Make sure inline edits of App Name, Form Name, and Module Name update doc Title

biyeun:
Builds off of https://github.com/dimagi/commcare-hq/pull/16741
@orangejenny // @benrudolph 

taking into account suggestion from @proteusvacuum 
- "{form name} - Form Settings - CommCare HQ"
- "{form name} - Edit Form - CommCareHQ"
- "{module name} - Module Settings - CommcareHQ"

### DIFF
--- a/corehq/apps/app_manager/static/app_manager/js/app_manager.js
+++ b/corehq/apps/app_manager/static/app_manager/js/app_manager.js
@@ -2,9 +2,30 @@
 hqDefine('app_manager/js/app_manager.js', function () {
     'use strict';
     var module = eventize({});
+    var _p = {};
+    _p.appendedPageTitle = "";
+    _p.prependedPageTitle = "";
 
     module.setCommcareVersion = function (version) {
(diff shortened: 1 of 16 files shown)


orangejenny (approved):
Looks good, though I'm not a fan of `_p` as a variable name.

biyeun:
@orangejenny I use it as a shorthand for `_private`...would you prefer I use that instead?

orangejenny:
yeah - it makes sense when you explain it, but I don't think `_p` alone is a convention in our codebase

biyeun:
@orangejenny resolved conflicts and responded to request

orangejenny (approved):

biyeun:
@proteusvacuum I just like having some sort of concrete signal that the variable is intended for use throughout the module rather than variables inside a smaller scope. I always did like that distinction between public / private ...

PR #16748, opened 2017-06-22
async indicator diagram

emord:
@nickpell I felt like I wasn't super clear when I was explaining this workflow to you. hoping this diagram helps illustrate it in the future. does this make the workflow make more sense?


nickpell:
The diagram is good.  And your explanation earlier made sense too

nickpell (approved):

emord:
just docs so self merging

PR #16750, opened 2017-06-22
soft rollout ucr

emord:
@snopoke the other day you mentioned a soft rollout option for reports. This will allow us to specify a soft rollout per report. works by 1/n will go to ES

### DIFF
--- a/corehq/apps/app_manager/fixtures/mobile_ucr.py
+++ b/corehq/apps/app_manager/fixtures/mobile_ucr.py
@@ -13,7 +13,7 @@ from corehq.apps.app_manager.suite_xml.features.mobile_ucr import is_valid_mobil
 from corehq.apps.userreports.reports.filters.factory import ReportFilterFactory
 from corehq.util.xml_utils import serialize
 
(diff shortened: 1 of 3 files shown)


corehq/apps/app_manager/fixtures/mobile_ucr.py
```diff
@@ -135,6 +135,9 @@ def report_config_to_fixture(report_config, restore_user):
     def _get_report_and_data_source(report_id, domain):
         report = get_report_config(report_id, domain)[0]
         data_source = ReportFactory.from_spec(report, include_prefilters=True)
+        if report.soft_rollout != 0 and data_source.config.backend_id == UCR_LABORATORY_BACKEND:
+            if random.randint(0, report.soft_rollout) == report.soft_rollout:
```
snopoke: To make this 1/N the range should be from 1 so that 1/1 = 100%, 1/2 = 50% etc.

It could also make it easier to think about if you inverted the soft rollout so that a value of 0.1 would yield 10% , 0.5 yields 50% etc.

emord:
changed to a decimal

snopoke (approved):

PR #16752, opened 2017-06-22
Externalize js: edit scheduled report

orangejenny:
@benrudolph / @millerdev


millerdev (approved):
approving, but would be nice to add a globals declaration at the top for `django`

orangejenny:
linted, then realized that edit_scheduled_report.js is only used on this one page & combined the two js files

millerdev (approved):
yay!

PR #16755, opened 2017-06-23
default export date filter to ALL data

snopoke:
https://manage.dimagi.com/default.asp?253764

### DIFF
--- a/corehq/apps/export/forms.py
+++ b/corehq/apps/export/forms.py
@@ -102,8 +102,13 @@ class DateSpanField(forms.CharField):
     def clean(self, value):
         date_range = super(DateSpanField, self).clean(value)
         dates = date_range.split(DateRangePickerWidget.separator)
-        startdate = dateutil.parser.parse(dates[0])
-        enddate = dateutil.parser.parse(dates[1])
+        if len(dates) != 2:
+            return
+        try:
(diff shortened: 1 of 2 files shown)


corehq/apps/export/forms.py
```diff
@@ -102,8 +102,13 @@ class DateSpanField(forms.CharField):
...
         dates = date_range.split(DateRangePickerWidget.separator)
-        startdate = dateutil.parser.parse(dates[0])
-        enddate = dateutil.parser.parse(dates[1])
+        if len(dates) != 2:
+            return
+        try:
+            startdate = dateutil.parser.parse(dates[0])
+            enddate = dateutil.parser.parse(dates[1])
+        except ValueError:
+            raise ValueError(_("Date range invalid. Pleaes select a valid date range."))
```
benrudolph: is this supposed to raise a `ValidationError`?

benrudolph (approved):

corehq/apps/export/forms.py
snopoke: I don't know. All I wanted to do here was change the message which get's displayed to the user.

benrudolph: ah i see i think the `clean` functions are supposed to raise ValidationErrors, though not sure how this form interacts with that. also small typo `Pleaes`

PR #16756, opened 2017-06-23
reduce set of case properties that aren't allowed

snopoke:
https://manage.dimagi.com/default.asp?255777


benrudolph (approved):
good solution

PR #16759, opened 2017-06-23
Externalize JavaScript: reports home

orangejenny:
@benrudolph / @millerdev


benrudolph (approved):
looks good, though i'd be very careful with the way this stuff gets loaded. reports has some of the craziest time dependencies and event handling

orangejenny:
@benrudolph ha, yes, I have trepidation about mucking around here

PR #16761, opened 2017-06-23
Drop couch locations

esoergel:
Couch `Location` is dead, long live `SQLLocation`.

I recently removed the last references to couch locations outside of the locations module.  This PR drops the sync code and the code testing it.

I'll later write a mgmt command to delete the documents from couch, but I figure there's no rush on that, and it'd be better to be cautious.


benrudolph:
holy 💩

corehq/apps/dump_reload/couch/dump.py
```diff
@@ -14,7 +14,6 @@
 from dimagi.utils.couch.database import iter_docs
 
 DOC_PROVIDERS = {
-    DocTypeIDProvider(['Location']),
```
esoergel: @snopoke as best as I could tell, this is used to determine what data to save to disk when a particular management command is run.  My assumption is that it's safe to just drop this entirely given that `Location` is being killed, and `SQLLocation` is being dumped anyways, does that sound correct?

snopoke:
Wow, this is awesome!

esoergel:
I think this is good to go.  @snopoke @gcapalbo got any reservations about removing couch locations at this point?  UCR location access has been migrated off couch, as has all code which uses the python `Location` model.  `copy_domain` and similar features also copy SQL locations, so dropping couch locations shouldn't be an issue there.  Are there any places I'm missing which may be accessing the documents in couch directly without going through the model?

snopoke:
Looks good to me. I was wondering about publishing to kafka but looks like that's already set to go.

snopoke (approved):

gcapalbo (approved):
@esoergel looks good to me, thanks for doing this.

confirming that the pillow code is using SQLLocations now as well?

esoergel:
@gcapalbo Yup, this gets published to kafka via the SQLLocation, so the UCR pillow reads from that.

PR #16762, opened 2017-06-23
Additional locations clean-up

esoergel:
Some followup I had after https://github.com/dimagi/commcare-hq/pull/16761

:blowfish:

### DIFF
--- a/corehq/apps/locations/models.py
+++ b/corehq/apps/locations/models.py
@@ -16,8 +16,6 @@ from corehq.apps.products.models import SQLProduct
 from corehq.toggles import LOCATION_TYPE_STOCK_RATES
 from mptt.models import MPTTModel, TreeForeignKey, TreeManager
 
-LOCATION_REPORTING_PREFIX = 'locationreportinggroup-'
-
 
 class LocationTypeManager(models.Manager):
 
@@ -562,35 +560,6 @@ class SQLLocation(MPTTModel):
         return '/'.join(self.get_ancestors(include_self=True)
                             .values_list('name', flat=True))
 
-    def _make_group_object(self, user_id, case_sharing):
-        from corehq.apps.groups.models import UnsavableGroup
-
-        g = UnsavableGroup()
-        g.domain = self.domain
-        g.users = [user_id] if user_id else []
-        g.last_modified = datetime.utcnow()
-
-        if case_sharing:
-            g.name = self.get_path_display() + '-Cases'
-            g._id = self.location_id
-            g.case_sharing = True
-            g.reporting = False
-        else:
-            # reporting groups
-            g.name = self.get_path_display()
-            g._id = LOCATION_REPORTING_PREFIX + self.location_id
-            g.case_sharing = False
-            g.reporting = True
-
-        g.metadata = {
-            'commcare_location_type': self.location_type.name,
-            'commcare_location_name': self.name,
-        }
-        for key, val in self.metadata.items():
-            g.metadata['commcare_location_' + key] = val
-
-        return g
-
     def get_case_sharing_groups(self, for_user_id=None):
         if self.location_type.shares_cases:
             yield self.case_sharing_group_object(for_user_id)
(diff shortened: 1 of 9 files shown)


esoergel:
@emord

esoergel:
bump

corehq/apps/locations/models.py
```diff
@@ -463,63 +463,41 @@ def _close_case_and_remove_users(self):
...
+            loc.save()
+
+            # reopen supply point case if needed
+            sp = loc.linked_supply_point()
+            # sanity check that the supply point exists and is not open.
+            # this is important because if you unarchive a child, then try
+            # to unarchive the parent, we don't want to try to open again
+            if sp and sp.closed:
+                for action in sp.actions:
+                    if action.action_type == 'close':
```
snopoke: I don't think SQL case 'actions' (`CaseTransaction`) have and `action_type` property.

esoergel: Oh yikes, this must've been broken for a while, then.  Do you have a recommendation for a backend-agnostic way to get the form which closed a case?

snopoke: Sorry, should have included that in the previous comment. You can use `action.is_case_close` which is implemented for both backends.

esoergel: So the reason this wasn't failing was because the supply point was being updated during `.save()` which happens before this is called.  This means it was never closed at this point.  I pushed all supply_point sync code into `supply_point_sync`, which simplifies that logic a bit.

snopoke:
test failure could be related
```
ERROR: corehq.apps.reminders.tests.test_recipient:ReminderRecipientTest.test_recipient_case_owner_location_parent
  File "/mnt/commcare-hq-ro/corehq/util/test_utils.py", line 267, in inner
  File "/mnt/commcare-hq-ro/corehq/util/test_utils.py", line 238, in __call__
  File "/mnt/commcare-hq-ro/corehq/util/test_utils.py", line 253, in call_with_settings
    fn(*args, **kwargs)
  File "/mnt/commcare-hq-ro/corehq/util/test_utils.py", line 234, in fn_with_pre_and_post
    self.fn(*args, **kwargs)
  File "/mnt/commcare-hq-ro/corehq/apps/reminders/tests/test_recipient.py", line 75, in test_recipient_case_owner_location_parent
    self.user.unset_location()
  File "/mnt/commcare-hq-ro/corehq/apps/users/models.py", line 1870, in unset_location
    self.save()
  File "/mnt/commcare-hq-ro/corehq/apps/users/models.py", line 1528, in save
    super(CommCareUser, self).save(**params)
  File "/mnt/commcare-hq-ro/corehq/apps/users/models.py", line 1394, in save
    super(CouchUser, self).save(**params)
  File "/vendor/lib/python2.7/site-packages/couchdbkit/schema/base.py", line 134, in save
    db.save_doc(doc, **params)
  File "/vendor/lib/python2.7/site-packages/couchdbkit/client.py", line 516, in save_doc
    **params).json_body
  File "/vendor/lib/python2.7/site-packages/restkit/resource.py", line 153, in put
    headers=headers, params_dict=params_dict, **params)
  File "/vendor/lib/python2.7/site-packages/couchdbkit/resource.py", line 141, in request
    response=e.response)
ResourceConflict: Document update conflict.
-------------------- >> begin captured stdout << ---------------------
<function test_recipient_case_owner_location_parent at 0x7f2feed13398> failed with the following settings:
settings.TESTS_SHOULD_USE_SQL_BACKEND = False
--------------------- >> end captured stdout << ----------------------
```

corehq/apps/reminders/tests/test_recipient.py
```diff
@@ -65,9 +65,9 @@ def test_recipient_case_owner_location_parent(self):
                 self.assertEqual(reminder.recipient, [parent_location])
 
             # Remove parent location
-            parent_location.delete()
             child_location.parent = None
             child_location.save()
+            parent_location.delete()
```
esoergel: @gcapalbo can you confirm that this fits the intent of the test?  `.delete()` is a cascading delete, so it also deletes the child location.  This would automatically unassign the user from the location, saving them and creating a document update conflict when the user is saved again below.

gcapalbo: ok, if .delete() is a cascading delete then i think we can just remove this line altogether, remove the child_location.delete() line below, and add a line for self.addCleanup(parent_location.delete) right after creating parent_location

corehq/apps/locations/models.py
```diff
@@ -389,10 +387,24 @@ def save(self, *args, **kwargs):
         if not self.location_id:
             self.location_id = uuid.uuid4().hex
         set_site_code_if_needed(self)
-        self.supply_point_id = sync_supply_point(self)
+        sync_supply_point(self)
         super(SQLLocation, self).save(*args, **kwargs)
         publish_location_saved(self.domain, self.location_id)
 
+    def delete(self, *args, **kwargs):
```
gcapalbo: i realize it may already behave this way, but is doing a hard delete of a location a desired feature? seems like some cases  or other export data still might point to that location and its information would be lost

esoergel: When we built that out, we were thinking of it as analogous to deleting a user.  The associated cases still exist, and you can dig them up, but they're hidden from normal views.  There had been an archival workflow around locations for a while, but that was confusing, since people would frequently make mistakes in setting up their project, then try to archive things and start over.  Archiving only hides a location, so it prevents you from making another location with the same site code, or with the same name and parent.

corehq/apps/reminders/tests/test_recipient.py
esoergel: Just updated.  I also put in a `self.addCleanup(parent_location.delete)`, as by the time cleanup happens, child_location will no longer be a child location of parent, so it won't be picked up in the cascading delete.

gcapalbo (approved):

PR #16763, opened 2017-06-23
sync device_id

benrudolph:
for icds

### DIFF
--- a/corehq/apps/callcenter/utils.py
+++ b/corehq/apps/callcenter/utils.py
@@ -178,7 +178,7 @@ def _get_user_case_fields(commcare_user):
         'email': commcare_user.email,
         'language': commcare_user.language or '',
         'phone_number': commcare_user.phone_number or '',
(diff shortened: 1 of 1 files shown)


corehq/apps/callcenter/utils.py
```diff
@@ -177,7 +177,8 @@ def valid_element_name(name):
         'username': commcare_user.raw_username,
         'email': commcare_user.email,
         'language': commcare_user.language or '',
-        'phone_number': commcare_user.phone_number or ''
+        'phone_number': commcare_user.phone_number or '',
+        'device_id': commcare_user.devices[0].device_id if commcare_user.devices else '',
```
benrudolph: curious if you think this should be `last_device_used` or maybe it should join all the device ids?

(the user case updates get inferred by the export and the key is used to name the column)

snopoke (commented):
I think fine to just be last device ID. Perhaps good to name the property in that way as well.

benrudolph:
✅

PR #16764, opened 2017-06-23
Do less work in location syncs

proteusvacuum:
@esoergel 
Probably isn't the full solution to the slow location fixtures, but is a start.

@snopoke buddy


proteusvacuum:
@calellowitz builds on slow sync stuff

snopoke (approved):

PR #16766, opened 2017-06-23
Fix error message

benrudolph:
@wpride @orangejenny this was so broken haha. it would just show "Case"


orangejenny (approved):
haha, much better

PR #16767, opened 2017-06-23
adding username parameter to async builds

calellowitz:
@emord cc: @amstone326 
This adds a username to the build comment for builds triggered by update to latest saved when the phone passes a username parameter.


millerdev (approved):

emord (approved):

amstone326:
Thank you @calellowitz! In the instances that mobile doesn't send the username parameter (which will be rare on 2.37 and above, but will remain true across the board for any versions below that), am I understanding correctly that the message would then read `Auto-generated by a phone update from user `? If so it might be nice to alter the message slightly in that case so that it doesn't end awkwardly like that. Could just use the old message, or say it's unknown or something?

calellowitz:
@amstone326 i changed the comment to read `Auto-generated by a phone update from <username>` if the user exists or `Auto-generated by a phone update from unknown user` otherwise. let me know if that still seems awkward.

amstone326:
@calellowitz That sounds great, thanks!

PR #16768, opened 2017-06-23
ICDS: Update to Beneficiary Indicator #2

gcapalbo:
- allows for custom content handlers in new framework
- makes second beneficiary indicator use a custom content handler to inspect latest growth monitoring form for content
- allows switching schedules to be live instead of always being in test mode
- starts capturing the icds indicator slug along with each SMS sent

@emord


custom/icds/messaging/custom_content.py
```diff
@@ -0,0 +1,74 @@
...
+    forms_under_consideration = []
+
+    db_names = get_db_aliases_for_partitioned_query()
+    for db_name in db_names:
+        result = XFormInstanceSQL.objects.using(db_name).filter(
+            domain=domain,
+            form_id__in=form_ids,
+            xmlns=GROWTH_MONITORING_XMLNS,
+            state=XFormInstanceSQL.NORMAL,
+        ).order_by('-received_on')[0:1]
```
emord: Could you use `.first()`  here?

corehq/apps/sms/api.py
```diff
@@ -55,17 +55,17 @@ def __init__(self, *args, **kwargs):
...
 def add_msg_tags(msg, metadata):
     if msg and metadata:
         fields = ('workflow', 'xforms_session_couch_id', 'reminder_id', 'chat_user_id',
-                  'ignore_opt_out', 'location_id', 'messaging_subevent_id')
+                  'ignore_opt_out', 'location_id', 'messaging_subevent_id', 'custom_metadata')
         for field in fields:
             value = getattr(metadata, field)
             if value is not None:
                 setattr(msg, field, value)
-        msg.save()
```
emord: why did you remove this save?

custom/icds/messaging/custom_content.py
gcapalbo: yeah, that's cleaner, updated

corehq/apps/sms/api.py
gcapalbo: i noticed this when adding the metadata, it's an unnecessary save as the message is always saved later on after processing, and this prevents one hit to the main db

emord (approved):

PR #16769, opened 2017-06-23
App manager v2: allow rearranging of registration forms

orangejenny:
https://manage.dimagi.com/default.asp?256712

@biyeun Could turn this into an add-on instead, but after mulling it over, I don't think making reg forms come first is likely to enhance people's understanding of case management (or that they'll understand why they can't rearrange the reg forms), I think it's enough that case lists are created with the reg form first. Let me know if you think this is an important constraint for us to have.

cc @benrudolph

### DIFF
--- a/corehq/apps/app_manager/templates/app_manager/v2/partials/form_link_primary.html
+++ b/corehq/apps/app_manager/templates/app_manager/v2/partials/form_link_primary.html
@@ -26,9 +26,7 @@
           {% else %}appnav-title-survey
           {% endif %}
           appnav-responsive">
-    {% if form.get_action_type != 'open' %}
         <i class="drag_handle appnav-drag-icon"></i>
-    {% endif %}
     <i
     {% if form.get_action_type == 'open' %}
         class="fcc fcc-app-createform appnav-primary-icon appnav-primary-icon-lg"


biyeun:
@orangejenny Doesn't allowing people to drag the reg forms also enable them to drop it into another module? Won't that confuse things?

orangejenny:
It does allow them to drop reg forms into another module, which will strip the case management if the other module doesn't use cases. Don't think it's any more or less confusing that being able to drag followup forms into survey modules.

My guess is that it isn't going to be obvious to new users why they can drag some forms around but not others - I don't actually think I can clearly articulate what this constraint is adding.

benrudolph (approved):
👍 agree that it won't make much sense to the user. maybe if you wanted to go down that route you could show a user an error message for why they can't move it

biyeun:
ok cool

biyeun:
I do like ben's idea though of alerting the user when dragging the reg form to another module. I think reg forms should be paired with their caselists...otherwise the whole point of the case list falls apart, no?

orangejenny:
@biyeun let's offline at the onboarding meeting tomorrow

orangejenny:
Offlined, should be good to merge. I'll do the other reg form-related changes in another PR, would like to get this one out since it was reported as a bug.

PR #16773, opened 2017-06-23
Ce/icds orphan cases

calellowitz:
@snopoke @emord

### DIFF
--- a/custom/icds/management/commands/export_orphaned_cases.py
+++ b/custom/icds/management/commands/export_orphaned_cases.py
@@ -12,6 +12,7 @@ from corehq.util.log import with_progress_bar
 CHILD_PROPERTIES = ['case_id', 'owner_id', 'opened_on', 'server_modified_on',
                     'name', 'indices', 'aadhar_number', 'dob', 'died']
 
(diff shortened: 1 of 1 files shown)


custom/icds/management/commands/export_orphaned_cases.py
```diff
@@ -0,0 +1,65 @@
...
+            'child_file',
+            help="File path child case output file",
+        )
+
+    def handle(self, hh_file, child_file, **options):
+        hh_cases = self._get_closed_hh_cases()
+        with open(hh_file, 'w') as hh_csv, open(child_file, 'w') as child_csv:
+            hh_writer = csv.writer(hh_csv)
+            child_writer = csv.writer(child_csv)
+            for cases in chunked(with_progress_bar(hh_cases), 500):
```
snopoke: This will fail because `with_progress_bar` need to know the total. Options are to pass in the total `hh_cases.count` or implement `__length__` on `ScanResult`

snopoke (changes requested):
One comment otherwise looks good

PR #16777, opened 2017-06-24
Add support to add condition on nodeset detail tabs to control when they are shown

mkangia:
https://manage.dimagi.com/default.asp?254746

Functionality complete. Need to add tests.
Review best 🏛 

"Hold off on merge" for any feedback on changes. Deployed on staging.

cc: buddy @biyeun @sheelio

### DIFF
--- a/corehq/apps/app_manager/static/app_manager/js/detail-screen-config.js
+++ b/corehq/apps/app_manager/static/app_manager/js/detail-screen-config.js
@@ -731,7 +731,7 @@ hqDefine('app_manager/js/detail-screen-config.js', function () {
                         }
 
                         // Observe condition values for the sake of validation
-                        if (that.condition) {
+                        if (that.original.condition) {
                             that.condition.observableVal = ko.observable(that.original.condition);
                             that.condition.on("change", function(){
                                 that.condition.observableVal(that.condition.val());
@@ -1026,7 +1026,7 @@ hqDefine('app_manager/js/detail-screen-config.js', function () {
                         0,
                         _.extend({
                             hasNodeset: tabs[i].has_nodeset,
-                        }, _.pick(tabs[i], ["header", "nodeset", "isTab"]))
+                        }, _.pick(tabs[i], ["header", "nodeset", "isTab", "condition"]))
                     );
                 }
                 if (this.columnKey === 'long') {
--- a/corehq/apps/app_manager/static/app_manager/js/detail-screen-config.js
+++ b/corehq/apps/app_manager/static/app_manager/js/detail-screen-config.js
@@ -719,7 +719,7 @@ hqDefine('app_manager/js/detail-screen-config.js', function () {
                         setTimeout(function () {
                             that.header.ui.addClass('input-group').prepend($('<span class="input-group-addon">Tab</span>'));
                             that.nodeset.ui.addClass('input-group').prepend($('<span class="input-group-addon">Nodeset</span>'));
-                            that.relevant.ui.addClass('input-group').prepend($('<span class="input-group-addon">DisplayCondition</span>'));
+                            that.relevant.ui.addClass('input-group').prepend($('<span class="input-group-addon">Display Condition</span>'));
                         }, 0);
 
                         // Observe nodeset values for the sake of validation
--- a/corehq/apps/app_manager/static/app_manager/js/detail-screen-config.js
+++ b/corehq/apps/app_manager/static/app_manager/js/detail-screen-config.js
@@ -730,7 +730,6 @@ hqDefine('app_manager/js/detail-screen-config.js', function () {
                             });
                         }
 
-                        // Observe condition values for the sake of validation
                         if (that.original.relevant) {
                             that.relevant.observableVal = ko.observable(that.original.relevant);
                             that.relevant.on("change", function(){
--- a/corehq/apps/app_manager/static/app_manager/js/detail-screen-config.js
+++ b/corehq/apps/app_manager/static/app_manager/js/detail-screen-config.js
@@ -671,7 +671,7 @@ hqDefine('app_manager/js/detail-screen-config.js', function () {
                     isTab: false,
                     hasNodeset: false,
                     nodeset: "",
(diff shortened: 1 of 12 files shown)


corehq/apps/app_manager/static/app_manager/js/detail-screen-config.js
```diff
@@ -712,11 +713,13 @@ hqDefine('app_manager/js/detail-screen-config.js', function () {
                     that.header.setVisibleValue(visibleVal);
 
                     that.nodeset = uiElement.input().val(that.original.nodeset);
+                    that.relevant = uiElement.input().val(that.original.relevant);
                     if (that.isTab) {
                         // hack to wait until the input's there to prepend the Tab: label.
                         setTimeout(function () {
                             that.header.ui.addClass('input-group').prepend($('<span class="input-group-addon">Tab</span>'));
                             that.nodeset.ui.addClass('input-group').prepend($('<span class="input-group-addon">Nodeset</span>'));
+                            that.relevant.ui.addClass('input-group').prepend($('<span class="input-group-addon">DisplayCondition</span>'));
```
orangejenny: "DisplayCondition" => "Display Condition"

corehq/apps/app_manager/templates/app_manager/v1/partials/case_list_properties.html
```diff
@@ -64,11 +64,24 @@
...
                             <div data-bind="jqueryElement: nodeset.ui"></div>
                             <span class="help-block"
                                   data-bind="text: 'A nodeset must be specified.', visible: showWarning"></span>
                         </td>
+                        {{ request|toggle_tag_info:"DISPLAY_CONDITION_ON_NODESET" }}
+                        {% if request|toggle_enabled:'DISPLAY_CONDITION_ON_NODESET' %}
+                        <td class="col-sm-3 form-group">
+                            <div data-bind="jqueryElement: relevant.ui"></div>
+                            <span class="help-block"
+                                  data-bind="text: 'A display condition can be specified.'"></span>
```
orangejenny: I'd delete this, don't think it's adding much value.

orangejenny:
Code looks good, just a couple minor comments.

What would you think about **not** feature flagging this? Nodesets in case detail is already an advanced feature, I'd think anyone who has it turned on is going to understand the display condition pretty easily. Alternatively, might be interesting to feature flag this separately from the nodeset part, so that even non-detail tabs could have display conditions. (also would welcome input from @sheelio )

mkangia:
@orangejenny i added the feature flag because it was suggested [here](https://manage.dimagi.com/default.asp?254746#BugEvent.1335974)

> Alternatively, might be interesting to feature flag this separately from the nodeset part

Its a new feature flag only. Separate from nodeset which is a different feature flag. 
I am okay to keep this under the existing nodeset feature flag itself.

orangejenny:
I'd been imagining that with the feature flag turned on, you'd be able to add a display condition to any tab. But it looks like the display condition is only available for tabs that have nodesets:
screen shot 2017-06-26 at 3 55 54 pm

I think if this feature is only for tabs with nodesets, an additional feature flag isn't necessary. Sorry for not being clearer in the ticket.

sheelio:
SOrry, I probably messed up writing the ticket, but we do want this available on both nodeset and normal tabs (see the top of the FB ticket).
@mkangia

mkangia:
i think i lost track of the requirements. Sorry for the miss and thanks for catching this @orangejenny. I will update to have this on tabs as well. I think then the new FF is alright.

mkangia:
@orangejenny updated. Should work for just tabs also now.

corehq/apps/app_manager/templates/app_manager/v1/partials/case_list_properties.html
```diff
@@ -66,23 +66,29 @@
...
-                        <td class="col-sm-3 form-group">
-                            <div data-bind="jqueryElement: relevant.ui"></div>
                             <span class="help-block"
-                                  data-bind="text: 'A display condition can be specified.'"></span>
+                                  data-bind="text: 'A nodeset must be specified.', visible: showWarning"></span>
                         </td>
-                        {% endif %}
                     <!--/ko-->
                     <!--ko ifnot: $data.hasNodeset -->
+                        {{ request|toggle_tag_info:"DISPLAY_CONDITION_ON_NODESET" }}
```
orangejenny: Couldn't this section - the header ui, toggle_tag_info, and relevant ui - move before `if: $data.hasNodeset`? Looks like it's duplicated in the two if blocks.

corehq/toggles.py
```diff
@@ -1211,3 +1211,10 @@ def _commtrackify(domain_name, toggle_is_enabled):
     [NAMESPACE_DOMAIN],
     always_enabled={'icds-cas'}
 )
+
+DISPLAY_CONDITION_ON_NODESET = StaticToggle(
```
orangejenny: Would rename this now that it applies to tabs whether or not they have nodesets.

orangejenny (approved):
Left a couple of comments, but nothing blocking.

corehq/apps/app_manager/templates/app_manager/v1/partials/case_list_properties.html
mkangia: @orangejenny there is a difference in their css.  One way to combine them to one would be to put if conditions for the styling. I thought i was better this way.

corehq/toggles.py
mkangia: Cool. Updated to `DISPLAY_CONDITION_ON_TABS`

corehq/apps/app_manager/templates/app_manager/v1/partials/case_list_properties.html
orangejenny: Ah, I did miss that. I'd lean toward doing ifs for the styling - it'd be ugly, but a good deal shorter - but leave it to whatever you prefer.

orangejenny (approved):

PR #16781, opened 2017-06-26
adding static ucrs to linked apps

calellowitz:
@snopoke cc: @emord 
Took a stab at this over the weekend as we think about next steps for icds app building.

### DIFF
--- a/corehq/apps/userreports/util.py
+++ b/corehq/apps/userreports/util.py
@@ -16,7 +16,6 @@ from corehq.apps.userreports.const import (
 from django_prbac.utils import has_privilege
 
 from corehq.apps.userreports.dbaccessors import get_all_es_data_sources
-from corehq.apps.userreports.exceptions import UserReportsError
 
 
 def localize(value, lang):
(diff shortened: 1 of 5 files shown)


corehq/apps/userreports/util.py
```diff
@@ -201,12 +202,14 @@ def get_static_report_mapping(from_domain, to_domain, report_map):
...
             is_custom_report = False
-        else:
+        elif static_report.get_id.startswith(CUSTOM_REPORT_PREFIX):
             report_id = static_report.get_id.replace(
                 CUSTOM_REPORT_PREFIX + from_domain + '-',
                 ''
             )
             is_custom_report = True
+        else:
+            raise UserReportsError("Found dynamic UCR when static expected")
```
snopoke: I don't think this will ever get raised since this loop is only iterating over static UCRs.

snopoke (changes requested):

snopoke (approved):
Would be great to have a test for this.

emord (approved):

PR #16782, opened 2017-06-26
domain admin restore security

snopoke:


corehq/apps/hqadmin/views.py
```diff
@@ -467,6 +468,11 @@ def get(self, request, *args, **kwargs):
         if not self.user:
             return HttpResponseNotFound('User %s not found.' % full_username)
 
+        if 'domain' in kwargs:
+            # domain check when this get's called as domain level admin restore.
```
proteusvacuum: Feels weird to have a check for the subclass be here. What if you made a `validate_user` method here which checks all the user properties which you could then override in the subclass?

proteusvacuum (commented):
Looks fine. One non-blocking comment.

benrudolph:
i know this wasn't you, but i feel like if we're going to have this be a route for some people it may be useful to ensure that the version param is by default set to `2.0` so that when we call `get_restore_params` it doesn't default to `1.0`. https://github.com/dimagi/commcare-hq/blob/master/corehq/apps/ota/views.py#L175

corehq/apps/hqadmin/views.py
snopoke: dd8c6e8

snopoke:
@benrudolph c077e4b

proteusvacuum (approved):

benrudolph (approved):

benrudolph:
(@amstone326 is going to double check this)

benrudolph:
ok we got the go ahead to give this a go

benrudolph:
(will revert if it causes a bunch of issues). the play store says there are 0 installs of it, but aliza says that's usually an approximate

PR #16783, opened 2017-06-26
ICDS: Fix indicators to handle the seeking_services flag

sheelio:
This fixes up a number of indicators so they don't count the case if the case is not seeking services at the AWC center.  

@emord for review

### DIFF
--- a/custom/icds_reports/ucr/data_sources/ccs_record_cases_monthly_tableau2.json
+++ b/custom/icds_reports/ucr/data_sources/ccs_record_cases_monthly_tableau2.json
@@ -2002,6 +2002,15 @@
                 }
               }
             }
+          },
+          {
+            "type": "boolean_expression",
+            "operator": "eq",
+            "expression": {
+              "type": "named",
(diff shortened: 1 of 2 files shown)


emord:
Build failure looks real. Also it looks like you're adding a column and renaming a column. this will require a full rebuild. is that ok?

sheelio:
@emord 

Pretty sure that I'm not adding a column (I think that's the diff being weird).   Fixed the build failure though.

sheelio:
Ah I see that re-arranged the columns.  Will fix.

sheelio:
After looking again, don't think the columns were re-arranged (the order is still bp3_complete, pnc_complete and postnatal).

emord (approved):
Yeah just a weird diff. looks good. is it important to deploy today?

sheelio:
Nope, not important to deploy today.  Just wanted to wrap up the data source changes (this shouldn't have any impact on the existing data since the app hasn't been updated to negate the seeking_services indicator).

PR #16784, opened 2017-06-26
make get_current_sequence_is always return a string

snopoke:


corehq/apps/hqadmin/tests/test_utils.py
```diff
@@ -101,6 +103,20 @@ def test_get_historical_max(self):
         self.assertIsNone(store)
 
 
+class TestHistoricalPillowCheckpoint(TestCase):
+
+    @real_pillow_settings()
+    def test_all_pillows(self):
+        for pillow in get_all_pillow_instances():
+            checkpoint = pillow.checkpoint
+            current_seq = checkpoint.get_current_sequence_id()
```
snopoke: This value gets saved to `HistoricalPillowCheckpoint.seq` so it should always be a string.

PR #16788, opened 2017-06-26
is_demo_user in ES user index

sravfeyn:
I am using this in this [PR](https://github.com/dimagi/commcare-hq/compare/sr-practice-mode), that will take time to get in. So, I am separating this out, so I don't have to deal with conflicts. The other PR has a util and test cases for this 

@nickpell

### DIFF
--- a/corehq/apps/userreports/es/adapter.py
+++ b/corehq/apps/userreports/es/adapter.py
@@ -18,6 +18,9 @@ import six
 
 # These settings tell ES to not tokenize strings
 UCR_INDEX_SETTINGS = {
+    "settings": {
+        "number_of_replicas": 0,
+    },
     "mappings": {
         "indicator": {
             "dynamic": "true",
@@ -216,6 +219,7 @@ class IndicatorESAdapter(IndicatorAdapter):
         except NotFoundError:
             pass
 
+
 def build_es_mapping(data_source_config):
     properties = {}
     for indicator in data_source_config.configured_indicators:
@@ -228,6 +232,6 @@ def build_es_mapping(data_source_config):
         if datatype == 'string':
             properties[indicator['column_id']]['index'] = 'not_analyzed'
     mapping = deepcopy(UCR_INDEX_SETTINGS)
-    mapping.update(data_source_config.get_es_index_settings())
+    mapping['settings'].update(data_source_config.get_es_index_settings()['settings'])
     mapping['mappings']['indicator']['properties'] = properties
     return mapping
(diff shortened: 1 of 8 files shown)


emord:
@benrudolph I think this is  a good opportunity to change this to 1 or 2 shards instead of 5. maybe with 1 replica? I think that the change would be somewhere here https://github.com/dimagi/commcare-hq/pull/15550/files#diff-014e7b677fc4d0d566f1c2e7da059c92R54

thoughts?

benrudolph:
2 shards sounds good to me with 1 replica. yes that is where you'd make the change

nickpell (approved):
Looks good pending questions from @emord and @benrudolph

sravfeyn:
@emord I've updated the shards/replicas conf. I have already deployed this yesterday on staging without sharding conf, assuming nothing special needs to happen for staging, yeah?

emord:
@sravfeyn @benrudolph I had to add 3 more commits to this because the way our ES settings were setup the replica setting was being overwritten during indexing.

@benrudolph would be good to get your review on that since I think you wrote tihs code originally

corehq/apps/userreports/es/adapter.py
```diff
@@ -228,6 +232,6 @@ def build_es_mapping(data_source_config):
         if datatype == 'string':
             properties[indicator['column_id']]['index'] = 'not_analyzed'
     mapping = deepcopy(UCR_INDEX_SETTINGS)
-    mapping.update(data_source_config.get_es_index_settings())
+    mapping['settings'].update(data_source_config.get_es_index_settings()['settings'])
```
benrudolph: this was necessary because it'd overwrite settings defined elsewhere? i don't think i wrote this portion

corehq/ex-submodules/pillowtop/es_utils.py
```diff
@@ -12,7 +12,6 @@
         "merge.policy.merge_factor": 20,
         "store.throttle.max_bytes_per_sec": "1mb",
         "store.throttle.type": "merge",
-        "number_of_replicas": "0"
```
benrudolph: sneaky.

benrudolph: do you have any idea why this specified number of replicas?

emord: The default in ES is 1

corehq/apps/userreports/es/adapter.py
emord: Yeah. that one is on me. This is the UCR settings I made

benrudolph:
👍 looks good

sravfeyn:
Much appreciated @emord, thanks!

sravfeyn:
Is this good to go? @emord Do I need to run anything special when a deploy happens on ICDS (or any other env)?

sravfeyn:
Offlined with Emord.

PR #16789, opened 2017-06-26
don't recast to set every time

proteusvacuum:
@snopoke


corehq/ex-submodules/casexml/apps/phone/models.py
```diff
@@ -689,7 +689,7 @@ def traverse_incoming_extensions(case_id, extension_index_tree, closed_cases):
 
     @memoized
     def get_cases_that_directly_depend_on_case(self, case_id):
-        return set(self.reverse_indices.get(case_id, []))
+        return self.reverse_indices.get(case_id, set([]))
```
snopoke: Nit. You can omit the empty list when creating a set

snopoke (approved):

PR #16790, opened 2017-06-26
User Group Dim

benrudolph:
@gcapalbo @snopoke @calellowitz sets up for the user group dim (goes into the location dim PR)

cc; @orangejenny


corehq/warehouse/transforms/sql/user_group_dim.sql
```diff
@@ -0,0 +1,43 @@
...
+FROM
+(
+(
+    SELECT
+        UNNEST(user_ids) AS user_id,
+        group_id AS group_id,
+        false AS deleted
+    FROM
+        {{ group_staging }}
+)
```
benrudolph: curious what you think of this query. i know subqueries like this aren't advisable, but i think this will be fast and small in almost all cases

corehq/warehouse/transforms/sql/user_group_dim.sql
```diff
@@ -0,0 +1,43 @@
...
+
+UNION
+
+(
+    SELECT
+        UNNEST(removed_user_ids) AS removed_user_id,
```
snopoke: Should this be `AS user_id`

benrudolph: it could be, union just matches on index and then takes the alias of the first query:

screen shot 2017-06-26 at 4 11 28 pm

it was just a semantic name

corehq/warehouse/transforms/sql/group_dim.sql
```diff
@@ -0,0 +1,26 @@
...
+    group_id,
+    name,
+    domain,
+    case_sharing,
+    reporting,
+    group_last_modified,
+    CASE doc_type
+        WHEN 'Group' THEN false
+        WHEN 'Group-Deleted' THEN true
+    END,
```
gcapalbo: i've sometimes seen weird source data for some models where the -Deleted suffix gets applied twice in some docs. i think we may want to use an else clause for all of our deleted flags

benrudolph: that's a good point, repeaters had this issue for sure

corehq/warehouse/transforms/sql/user_group_dim.sql
gcapalbo: do we need to keep track of removed users for any reason on the user group dim? i think i'd be in favor of not keeping track of them since a deleted record here isn't really referenced anywhere on its own in the source.

if we don't keep track of them, then we just need to load the user group dim in two steps:

1) delete from user_group_dim using group_staging ... to delete any existing records for the staged groups
2) insert into user_group_dim select from group_staging join group_dim join user_dim ... to insert the current ones

and we should be able to get rid of the subquery

benrudolph:
(can't get tests to run, closing and making new PR)

PR #16792, opened 2017-06-26
Broadcast message view

emord:
@gcapalbo This implements the first view of the spec. I used the old Broadcast Messages view as a template to make this one

I also blocked the old "Broadcast Messages" UI, but I think that "Compose a Message" and maybe "Reminders" should be blocked as well for the framework. Is that correct?

image

buddy @calellowitz


gcapalbo (approved):
:+1: looks good

gcapalbo:
@emord yeah agree, in addition to hiding the old broadcast pages, we should also hide compose, reminders, and reminder calendar when the new framework is enabled

emord:
> we should also hide compose, reminders, and reminder calendar when the new framework is enabled

will do this in a followup PR

PR #16794, opened 2017-06-26
Partition device log table

emord:
@dimagi/scale-team curious what you think about this. It was really easy to implement. This is pretty much copied straight from their docs.

This would partition the device log entry table by day and it's totally transparent to django and implemented in postgres. The table definition looks like:

```
commcarehq=# \d phonelog_partitioneddevicereportentry 
                                    Table "public.phonelog_partitioneddevicereportentry"
   Column    |           Type           |                                     Modifiers                                      
-------------+--------------------------+------------------------------------------------------------------------------------
 id          | integer                  | not null default nextval('phonelog_partitioneddevicereportentry_id_seq'::regclass)
 xform_id    | character varying(50)    | not null
 i           | integer                  | not null
 msg         | text                     | not null
 type        | character varying(32)    | not null
 date        | timestamp with time zone | not null
 server_date | timestamp with time zone | 
 domain      | character varying(100)   | not null
 device_id   | character varying(50)    | 
 app_version | text                     | 
 username    | character varying(100)   | 
 user_id     | character varying(50)    | 
Indexes:
    ...
Triggers:
    after_insert_phonelog_partitioneddevicereportentry_trigger AFTER INSERT ON phonelog_partitioneddevicereportentry FOR EACH ROW EXECUTE PROCEDURE phonelog_partitioneddevicereportentry_delete_master()
    before_insert_phonelog_partitioneddevicereportentry_trigger BEFORE INSERT ON phonelog_partitioneddevicereportentry FOR EACH ROW EXECUTE PROCEDURE phonelog_partitioneddevicereportentry_insert_child()
Number of child tables: 2 (Use \d+ to list them.)

```

I think the pros of this would be that we could drop an entire table at a time instead of looking through rows and we'll have many smaller tables (~10 GB) instead of one large one (~500 GB). I think this would make things like vacuuming faster. (the current one has been running for the past 5 days)

Migration could probably just be a management command to bulk save them into the new table after calls to this table were changed.

Library docs: http://architect.readthedocs.io/features/partition/index.html


corehq/ex-submodules/phonelog/models.py
```diff
@@ -30,6 +31,34 @@ def __repr__(self):
         return u"DeviceReportEntry(domain='{}', msg='{}')".format(self.domain, self.msg)
 
 
+@architect.install('partition', type='range', subtype='date', constraint='day', column='server_date')
+class PartitionedDeviceReportEntry(models.Model):
```
emord: This is probably confusing naming since it's not using our partitions in the same way as forms/cases/sms. Might want to change it if you guys want to go through with this change

snopoke:
Nice, this is great. I just added this to the [tech-debt trello](https://trello.com/c/2iXm7fNQ) yesterday. Agree that the table name should be different (or else we change the other references to be `Sharded` or something).

emord:
Looks like this test is exposing an improperly torn down test. I'm going to have to go through the tests to make sure they call super's teardown/setup class methods to get these to pass :(

gcapalbo:
:+1: great thing to do for this table

corehq/ex-submodules/phonelog/__init__.py
```diff
@@ -0,0 +1,34 @@
...
+    paths = {model.__module__ for model in sender.get_models()}
+    for path in paths:
+        try:
+            partition.run(dict(module=path))
+        except ProgrammingError:
+            # Possibly because models were just un-migrated or
+            # fields have been changed that effect Architect
+            print("Unable to apply partitions for module '{}'".format(path))
+        else:
+            print("Applied partitions for module '{}'".format(path))
```
gcapalbo: so we need to apply the partitions on every migration, or just the first time the new table is created?

corehq/ex-submodules/phonelog/__init__.py
```diff
@@ -0,0 +1,34 @@
...
+    in the app and ensure the partitions have been setup
+    """
+    paths = {model.__module__ for model in sender.get_models()}
+    for path in paths:
+        try:
+            partition.run(dict(module=path))
+        except ProgrammingError:
+            # Possibly because models were just un-migrated or
+            # fields have been changed that effect Architect
+            print("Unable to apply partitions for module '{}'".format(path))
```
gcapalbo: should this fail hard?

emord:
Ignore the second commit for now. I will pull it out once I verify that tests pass with it

corehq/ex-submodules/phonelog/__init__.py
emord: I'm not sure. I just copied this from their docs so that I could quickly verify it and show you guys what it may look like. I'll do some more digging since it looks like we're all in agreement that this looks to be a good thing to move towards

corehq/ex-submodules/phonelog/__init__.py
emord: It's hard to say. when this runs it means the migration has already occured (it happens post migrate after the couchdb ddoc sync). It certainly needs a harder way of warning than just a print statement that will be lost in our deploy

corehq/ex-submodules/phonelog/__init__.py
```diff
@@ -0,0 +1,34 @@
...
+            print("Applied partitions for module '{}'".format(path))
+
+
+class PhoneLogConfig(AppConfig):
+    name = 'phonelog'
+
+    def ready(self):
+        super(PhoneLogConfig, self).ready()
+        # Hook up Architect to the post migrations signal
+        post_migrate.connect(create_partitions, sender=self)
```
gcapalbo: also just remembered, i think it's good to give a dispatch_uid here to make sure this only gets connected once: https://docs.djangoproject.com/en/1.11/topics/signals/#preventing-duplicate-signals

emord:
You should only look at the last two commits, i'll rebase once my other PR is approved/merged.

I removed the app config and added the partitioning to the migration https://github.com/dimagi/commcare-hq/pull/16794/commits/e0609e7b5df908cfc8907b164055b4bbfda5845b#diff-86ceb68fafaa819c46b946772c81979bR12

My change renames the former `DeviceLogEntry` to `OldDeviceLogEntry`, and reintroduceds `DeviceLogEntry` as a partitioned model. 

I think the migration should just be going to postgres and running `insert into phonelog_daily_partitioned_devicereportentry select * from phonelog_devicereportentry`

In a separate PR we can drop the old device logs and change the pruning to dropping tables instead of deleting records.

snopoke:
Considering the size of this table I'm not sure we could migrate all the data in one go.  Another option would be to keep using the old table but put a trigger on it to insert new records into the new table. Then once we're past 60 days (or if we migrate some old data over manually) we can just switch to using the new table.

gcapalbo (approved):
looks good. is it going to be an issue that there will be some period of time where the new table doesn't have the full set of records?

emord:
>  is it going to be an issue that there will be some period of time where the new table doesn't have the full set of records?

@gcapalbo I think these are mostly used for auditing issues so I don't think that it would be the worst thing if there is some downtime for these.

> Considering the size of this table I'm not sure we could migrate all the data in one go. Another option would be to keep using the old table but put a trigger on it to insert new records into the new table.

@snopoke Yeah. that makes sense. What if I created a management command that ran something like:

```
for day in days:
    INSERT INTO phonelog_daily_partitioned_devicereportentry 
    SELECT * FROM phonelog_devicereportentry 
    WHERE server_date BETWEEN day@midnight AND day@11:59PM
```
where we start at today and then go back through the history? Do you think that would be a sufficient migration strategy?

emord:
I've updated with a management command to migrate over to the new table. It chunks by hour

corehq/ex-submodules/phonelog/management/commands/migrate_device_entry.py
```diff
@@ -0,0 +1,39 @@
...
+        while current > oldest_date:
+            hour_ago = current - timedelta(hours=1)
+            with connection.cursor() as cursor:
+                cursor.execute(
+                    "INSERT INTO " + partitioned_table +
+                    " (" + ','.join(COLUMNS) + ") " +
+                    "SELECT " +
+                    ','.join(COLUMNS) + " " +
+                    "FROM " + old_table + " " +
+                    "WHERE server_date BETWEEN %s AND %s",
```
gcapalbo: between does inclusive on both ends of the range, might be better to do > start and <= end

corehq/ex-submodules/phonelog/migrations/0012_server_date_not_null.py
```diff
@@ -0,0 +1,22 @@
...
+
+    dependencies = [
+        ('phonelog', '0011_partition_devicelogentry'),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name='devicereportentry',
+            name='server_date',
+            field=models.DateTimeField(db_index=True, default=datetime.datetime(2018, 6, 30, 0, 0, 0, 0)),
```
gcapalbo: why 2018 for the default?

gcapalbo (approved):

corehq/ex-submodules/phonelog/management/commands/migrate_device_entry.py
```diff
@@ -0,0 +1,39 @@
...
+
+from phonelog.models import OldDeviceReportEntry, DeviceReportEntry
+
+COLUMNS = (
+    "xform_id", "i", "msg", "type", "date", "server_date", "domain",
+    "device_id", "app_version", "username", "user_id"
+)
+
+
+class Command(BaseCommand):
```
gcapalbo: before we run this we may also want to just double check we have enough space to fit it

corehq/ex-submodules/phonelog/management/commands/migrate_device_entry.py
emord: thanks makes sense. https://github.com/dimagi/commcare-hq/pull/16794/commits/83bf41585ef5a997207f2675375284a23f40c079

corehq/ex-submodules/phonelog/migrations/0012_server_date_not_null.py
emord: Ah. I meant to comment on this. This shouldn't be hit because it's a new table, but by making it that far ahead i can easily go check after the migration.

in the current device log table on NIC we don't have any records taht are null. on staging we had a lot, but I think that would have been an earlier bug. that shouldn't happen any longer

snopoke (approved):

emord:
Confirmed that prod and softlayer have more than enough diskspace for this migration.

NIC has 1 TB and this table should be < 500 GB. I also removed a useless index on xform_id on these tables. Any queries to that will be covered by the unique constraint of (xform_id, i). I think this is good to go. and I can handle migrating on wednesday

PR #16800, opened 2017-06-27
Check aggregate['total_size'] is not None

kaapstorm:
If there are no DataFiles, then `aggregate['total_size']` will be `None`, not 0. Check this before assuming it's an integer.

@snopoke cc @gcapalbo @benrudolph

### DIFF
--- a/corehq/apps/export/views.py
+++ b/corehq/apps/export/views.py
@@ -1550,7 +1550,10 @@ class DataFileDownloadList(BaseProjectDataView):
             return self.get(request, *args, **kwargs)
 
         aggregate = DataFile.objects.filter(domain=self.domain).aggregate(total_size=Sum('content_length'))
-        if aggregate['total_size'] + request.FILES['file'].size > MAX_DATA_FILE_SIZE_TOTAL:
+        if (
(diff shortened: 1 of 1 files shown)


corehq/apps/export/views.py
```diff
@@ -1550,7 +1550,10 @@ def post(self, request, *args, **kwargs):
             return self.get(request, *args, **kwargs)
 
         aggregate = DataFile.objects.filter(domain=self.domain).aggregate(total_size=Sum('content_length'))
-        if aggregate['total_size'] + request.FILES['file'].size > MAX_DATA_FILE_SIZE_TOTAL:
+        if (
+            aggregate['total_size'] and
+            aggregate['total_size'] + request.FILES['file'].size > MAX_DATA_FILE_SIZE_TOTAL
+        ):
```
gcapalbo: if i understand this correctly, i think the way this is right now it would always let you upload any size you want the first time, since this statement would always evaluate to False?

gcapalbo: i see, i didn't notice that line above, thanks

PR #16803, opened 2017-06-27
"domain" -> "project space"

kaapstorm:
Apparently they're called "project spaces" now. :stuck_out_tongue_winking_eye: 

Noticed this, and figured I'd change it.

@gcapalbo

### DIFF
--- a/corehq/apps/users/views/__init__.py
+++ b/corehq/apps/users/views/__init__.py
@@ -744,7 +744,7 @@ class UserInvitationView(object):
 
     @property
     def success_msg(self):
-        return 'You have been added to the "%s" project space.' % self.domain
+        return _('You have been added to the "%s" project space.') % self.domain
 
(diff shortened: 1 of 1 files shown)


corehq/apps/users/views/__init__.py
```diff
@@ -744,7 +744,7 @@ def inviting_entity(self):
 
     @property
     def success_msg(self):
-        return "You have been added to the %s domain" % self.domain
+        return 'You have been added to the "%s" project space.' % self.domain
```
gcapalbo: nice catch. does this need internationalization too?

orangejenny:
keep fighting the good fight!

corehq/apps/users/views/__init__.py
kaapstorm: Good point!

PR #16807, opened 2017-06-27
save external ID for new cases

snopoke:
https://manage.dimagi.com/default.asp?253767


dannyroberts:
Wow nice find, "save on update but not create" bugs are one of those things that tend to test someone's confidence in their own sanity.

dannyroberts (approved):

PR #16808, opened 2017-06-27
fix test setup and teardown

emord:
opening for tests

### DIFF
--- a/corehq/blobs/tests/test_migratingdb.py
+++ b/corehq/blobs/tests/test_migratingdb.py
@@ -1,13 +1,11 @@
 from __future__ import unicode_literals
 from StringIO import StringIO
 
-from django.conf import settings
 from testil import replattr
 
 import corehq.blobs.migratingdb as mod
 from corehq.blobs import DEFAULT_BUCKET
-from corehq.blobs.tests.util import get_id, TemporaryS3BlobDB, TemporaryFilesystemBlobDB
-from corehq.util.test_utils import trap_extra_setup
+from corehq.blobs.tests.util import get_id, TemporaryFilesystemBlobDB, TemporaryMigratingBlobDB
 
 
 def get_base_class():
@@ -20,16 +18,11 @@ class TestMigratingBlobDB(get_base_class()):
 
     @classmethod
     def setUpClass(cls):
-        with trap_extra_setup(AttributeError, msg="S3_BLOB_DB_SETTINGS not configured"):
-            config = settings.S3_BLOB_DB_SETTINGS
-        cls.s3db = TemporaryS3BlobDB(config)
-        cls.fsdb = TemporaryFilesystemBlobDB()
-        cls.db = mod.MigratingBlobDB(cls.s3db, cls.fsdb)
+        super(TestMigratingBlobDB, cls).setUpClass()
 
-    @classmethod
-    def tearDownClass(cls):
-        cls.fsdb.close()
-        cls.s3db.close()
+        cls.s3db = cls.db
+        cls.fsdb = TemporaryFilesystemBlobDB()
+        cls.db = TemporaryMigratingBlobDB(cls.s3db, cls.fsdb)
 
     def test_fall_back_to_fsdb(self):
         info = self.fsdb.put(StringIO(b"content"), get_id())
--- a/corehq/blobs/tests/test_migratingdb.py
+++ b/corehq/blobs/tests/test_migratingdb.py
@@ -5,7 +5,12 @@ from testil import replattr
 
 import corehq.blobs.migratingdb as mod
 from corehq.blobs import DEFAULT_BUCKET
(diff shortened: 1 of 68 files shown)


corehq/blobs/migratingdb.py
```diff
@@ -45,3 +45,7 @@ def get_path(self, *args, **kw):
 
     def copy_blob(self, content, info, bucket):
         self.new_db.copy_blob(content, info, bucket)
+
+    def close(self):
```
emord: @millerdev Is there any reason the migrating db didn't have a close method? I added it so that I could change the tests around to properly call `setUpClass` and `tearDownClass`

millerdev: Yes, it's not part of the blob db interface. It's a convenience method on test blob db implementations only. I think it would be better to use [`TemporaryMigratingBlobDB`](https://github.com/dimagi/commcare-hq/blob/2553af0662c033080df4fd72083222836992a921/corehq/blobs/tests/util.py#L95), which has a `close()` method.

corehq/blobs/tests/test_migratingdb.py
```diff
@@ -20,17 +20,16 @@ class TestMigratingBlobDB(get_base_class()):
...
     def setUpClass(cls):
+        super(TestMigratingBlobDB, cls).setUpClass()
+
         with trap_extra_setup(AttributeError, msg="S3_BLOB_DB_SETTINGS not configured"):
             config = settings.S3_BLOB_DB_SETTINGS
+
+        # these are all closed in the base class
         cls.s3db = TemporaryS3BlobDB(config)
         cls.fsdb = TemporaryFilesystemBlobDB()
         cls.db = mod.MigratingBlobDB(cls.s3db, cls.fsdb)
```
millerdev: Looks like this overwrites [an attribute created by `super...setupClass()`](https://github.com/dimagi/commcare-hq/blob/2553af0662c033080df4fd72083222836992a921/corehq/blobs/tests/test_s3db.py#L90) before that db has been cleaned up properly. I think it was intentional that this method did not call the super method because of that. However, you can get around that by replacing `cls.s3db = TemporaryS3BlobDB(config)` with `cls.s3db = self.db`, and can also remove the `with trap_extra_setup...` block.

millerdev (changes requested):

millerdev:
FWIW, sometimes I prefer to override without calling a super method if there is not much shared/reused code. It makes it easier to quickly understand the local implementation without referencing another class or file, and obviates comments like `# these are all closed in the base class`.

However, it appears that there is some other reason to call `super.setUpClass/tearDownClass()` that I didn't know about. The title of this PR is "fix test setup and teardown," but it's not clear what this is fixing. What's the reason for adding all these super calls?

emord:
@millerdev https://docs.djangoproject.com/en/1.11/topics/testing/tools/#django.test.SimpleTestCase.allow_database_queries

> SimpleTestCase and its subclasses (e.g. TestCase, …) rely on setUpClass() and tearDownClass() to perform some class-wide initialization (e.g. overriding settings). If you need to override those methods, don’t forget to call the super implementation:

It seems like this works in most cases without calling this, but it sometimes pops up when making other unrelated changes. This current motivation to clean up everything came from the test failure https://github.com/dimagi/commcare-hq/pull/16794, but it also hit me when trying to combine test jobs on travis (we could reduce the # of jobs by 1, but there's some bad test teardown/setup that prevented that)

corehq/blobs/migratingdb.py
emord: Thanks. I think the correct setup is now used https://github.com/dimagi/commcare-hq/pull/16808/commits/969fe76a65d396fa98d0a9796b1a71c5891324ce

corehq/blobs/tests/test_migratingdb.py
```diff
@@ -20,16 +18,11 @@ class TestMigratingBlobDB(get_base_class()):
...
-        cls.s3db = TemporaryS3BlobDB(config)
-        cls.fsdb = TemporaryFilesystemBlobDB()
-        cls.db = mod.MigratingBlobDB(cls.s3db, cls.fsdb)
+        super(TestMigratingBlobDB, cls).setUpClass()
 
-    @classmethod
-    def tearDownClass(cls):
-        cls.fsdb.close()
-        cls.s3db.close()
+        cls.s3db = cls.db
```
millerdev: nitpick: maybe add `assert isinstance(cls.db, TemporaryS3BlobDB), cls.db` here to protect against the base class changing in some unanticipated way?

millerdev (approved):

corehq/blobs/tests/test_migratingdb.py
emord: added https://github.com/dimagi/commcare-hq/pull/16808/commits/be2a8370f205d525188cc038062c4ce2b00f04e5

emord:
:white_check_mark: (not going to fix that stickler comment as it was prexisting)

emord:
buddy @esoergel

PR #16809, opened 2017-06-27
Only generate xml for required cases

proteusvacuum:
@snopoke 
FYI @ctsims


snopoke:
looks good but haven't done detailed review - I can look later today or
tomorrow. Also @millerdev will be interested in this.

millerdev:
I had seen that the XML was generated early and thought it was a potential optimization to defer that, but hadn't yet determined if the functions that create the XML had other side effects so wasn't confident in just doing it. Seems like a win.

proteusvacuum:
@snopoke @millerdev we were hoping to get this in today if possible

snopoke (approved):

PR #16811, opened 2017-06-27
enikshay: Add referral report 

NoahCarnahan:
@nickpell 
Opening this PR in favor of https://github.com/dimagi/commcare-hq/pull/16771
I would like to get this PR in by wed EOD if possible.
Probably not worth reviewing this too closely, as most of it is UCR config

### DIFF
--- a/custom/enikshay/management/commands/print_referral_report_indicators.py
+++ b/custom/enikshay/management/commands/print_referral_report_indicators.py
@@ -1,0 +1,287 @@
+import json
+
+from django.core.management.base import BaseCommand
+
+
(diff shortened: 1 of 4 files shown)


custom/enikshay/management/commands/print_referral_report_indicators.py
```diff
@@ -0,0 +1,287 @@
+import json
```
NoahCarnahan: This is just a little script to make writing data source indicators less painful. This script was used to generate several of the indicators in `person_for_referral_report.json`, and can be used again if changed need to be made.

nickpell (approved):

PR #16008, opened 2017-05-04
Revert "Revert "Mk nikshay audit cookie safe""

mkangia:
Reverts dimagi/commcare-hq#16005

This fixes the issue introduced due to changes in app preview.
Adding hold off on merge to add a notification in case csrftoken is missed anywhere to better tracking.

PR #16816, opened 2017-06-28
remove csrf assert notification

mkangia:
Removing this assert as it is no longer needed

Was added here https://github.com/dimagi/commcare-hq/pull/16008


sravfeyn (commented):
Looks good, but build failed

mkangia:
the build failed to run.. restarted it.

PR #16817, opened 2017-06-28
add edited_on system property for form exports

snopoke:
https://manage.dimagi.com/default.asp?257055

### DIFF
--- a/corehq/apps/export/system_properties.py
+++ b/corehq/apps/export/system_properties.py
@@ -178,7 +178,7 @@ BOTTOM_MAIN_FORM_TABLE_PROPERTIES = [
         tags=[PROPERTY_TAG_SERVER],
         label="edited_on",
         item=ExportItem(path=[PathNode(name="edited_on")]),
-        selected=True,
+        is_advanced=True,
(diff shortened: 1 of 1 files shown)


corehq/apps/export/system_properties.py
```diff
@@ -176,6 +176,13 @@
     ),
     ExportColumn(
         tags=[PROPERTY_TAG_SERVER],
+        label="edited_on",
+        item=ExportItem(path=[PathNode(name="edited_on")]),
+        selected=True,
+        help_text=_("The time at which this form was last edited."),
```
benrudolph: can this also have `is_advanced=True`? otherwise looks good

snopoke: f5798a1

benrudolph (approved):

krsdimagi:
Thanks, @snopoke!  Similar to https://github.com/dimagi/commcare-hq/pull/16883, is there a way to expose user_id here as well?

snopoke:
user_id is already available as `userID` in form exports

PR #16818, opened 2017-06-28
Added celery task to clear expired sessions.

emord:
@dimagi/scale-team turns out our django_sessions table is huge for no reason. this [command] will clear out any expired sessions from our db. From prod (reduces the table by 98%....):

```
commcarehq=> select count(*) from django_session where expire_date < now();
  count   
----------
 65369587
(1 row)

commcarehq=> select count(*) from django_session where expire_date > now();
  count  
---------
 1621639
(1 row)
```

[command]: https://github.com/django/django/blob/stable/1.10.x/django/contrib/sessions/management/commands/clearsessions.py

### DIFF
--- a/corehq/apps/cleanup/tasks.py
+++ b/corehq/apps/cleanup/tasks.py
@@ -132,3 +132,8 @@ def pprint_stats(stats, outstream):
         )
         for user in sorted(list(users)):
             outstream.write("        {}\n".format(user))
+
+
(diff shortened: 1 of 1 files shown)


snopoke (approved):
Nice, I was going to mention that in the next meeting.

benrudolph:
👍 we also may want to move away from sql sessions

PR #16819, opened 2017-06-28
don't pass user object to celery task

sravfeyn:
https://sentry.io/dimagi/commcarehq/issues/242460043/ 

@nickpell


sravfeyn:
@snopoke ✅

PR #16822, opened 2017-06-28
Reduce ASYNC_RETRY_AFTER

proteusvacuum:
If your sync takes 41 seconds long to compute, then it will actually take 55s to return (10+15*3), which seems silly. This just tells the phone to ping the server to see if it is complete more frequently

@snopoke 

FYI @amstone326 @ctsims


snopoke (approved):

amstone326:
woooo baby steps toward this being fully rolled out!

PR #16823, opened 2017-06-28
Restore optimization: "livequery"

millerdev:
This is a new implementation of restore logic that attempts to minimize database queries and thereby optimize restores. The couch implementation is functional and passes all tests, but it is less efficient than SQL. Currently the new calculation is only accessible via the admin API by specifying `livequery=true`:

https://www.commcarehq.org/hq/admin/phone/restore/?version=2.0&as=USER_ID&livequery=true

In other words, this PR should not affect existing production sync logic in any way.

The implementation changed to accommodate sharding since I wrote the documentation. However, I believe the algorithm in the documentation is still basically correct. It depends on a `case_index` data structure that does not currently exist. Creating and maintaining the `case_index` envisioned there would require changes to sync writes (i.e., when persisting incoming case changes from phone) to pre-calculate metrics that could be used to speed up the liveness calculation on read/restore. That is left as a future exercise. But before we go there we should likely do other simpler changes, such as writing parent (`referenced_id`) closed status on `CommCareCaseIndexSQL`, to further reduce the number of queries and optimize the restore calculation.

@proteusvacuum @dannyroberts @snopoke @ctsims


corehq/ex-submodules/casexml/apps/phone/data_providers/case/livequery.py
```diff
@@ -0,0 +1,208 @@
...
+                enliven(case_id)
+
+        # open root nodes and their extensions -> live
+        # TODO store referenced_id open/closed status in case index to avoid this query
+        root_ids = [r for r in extensions_by_host if r not in hosts_by_extension]
+        debug('roots: %r', root_ids)
+        for case_id in accessor.filter_open_case_ids(root_ids):
+            enliven(case_id)
+
+        debug('live: %r', live_ids)
```
millerdev: - Is it necessary to save updated sync logs somewhere around here?
- Possibly related: should this tell the phone to remove no-longer-relevant cases?

corehq/ex-submodules/casexml/apps/phone/data_providers/case/livequery.py
```diff
@@ -0,0 +1,208 @@
...
+                for item in get_xml_for_response(update, restore_state))
+
+    return response
+
+
+def get_commtrack_elements(cases, restore_state):
+    return get_stock_payload(
+        restore_state.project,
+        restore_state.stock_settings,
+        [CaseStub(case.case_id, case.type) for case in cases],
```
millerdev: Is it required to pass in a list of `CaseStub`s here, or would it work just as well to pass in `cases` (which is a list of real cases)?

proteusvacuum: you can also pass in a list of cases (a change I recently made actually does that)

corehq/ex-submodules/casexml/apps/phone/data_providers/case/providers.py
```diff
@@ -8,5 +9,7 @@ class CasePayloadProvider(FullResponseDataProvider):
     """
 
     def get_response(self, restore_state):
+        if not self.async_task and restore_state.do_livequery:
```
proteusvacuum: you probably know this already, but we're going to want this in async restore when this goes live

millerdev:
...well, it worked before I rebased my changes on master. Working on test failures.

corehq/ex-submodules/casexml/apps/phone/data_providers/case/utils.py
```diff
@@ -40,8 +40,9 @@ def _get_required_updates(self):
 
 def get_case_sync_updates(domain, cases, last_sync_log):
     """
-    Given a domain, list of cases, and sync log representing the last sync, return a list
-    of CaseSyncUpdate objects that should be applied to the next sync.
+    Given a domain, list of cases, and sync log representing the last
+    sync, return a dict of CaseSyncUpdate objects (keyed by case_id)
+    that should be applied to the next sync.
```
proteusvacuum: oops, thanks

proteusvacuum (approved):

millerdev:
Hmm, tests failed on the sharded setup:
```
ProgrammingError: function get_related_indices(unknown, text[], text[]) does not exist
LINE 1: SELECT * FROM get_related_indices('test-domain', ARRAY['d'],...
                      ^
HINT:  No function matches the given name and argument types. You might need to add explicit type casts.
```
Maybe I didn't setup the migration correctly for sharding? @snopoke do you know what I did wrong?

millerdev:
Figured it out. Hopefully fixed this time.

PR #16824, opened 2017-06-28
Externalize js: case details

orangejenny:
And fix some small annoyances.

@benrudolph

### DIFF
--- a/corehq/apps/reports/templates/reports/reportdata/case_details.html
+++ b/corehq/apps/reports/templates/reports/reportdata/case_details.html
@@ -7,56 +7,14 @@
 
 {% block head %} {{ block.super }}
     {% include "imports/proptable.html" %}
-    <style type="text/css">
-        .dataTables_length select {
-            display: inline-block;
-        }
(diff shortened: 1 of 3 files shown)


corehq/apps/reports/templates/reports/reportdata/case_details.html
```diff
@@ -7,74 +7,32 @@
...
-    <style type="text/css">
-        .dataTables_length select {
-            display: inline-block;
-        }
-    </style>
-{% endblock %}
-
-{% block js-inline %}{{ block.super }}
-    <script>
-    $(function() {
```
orangejenny: This whole section appears to not be in use; I couldn't find any references to `form_lookups` or to the classes `xform_ref` and `case_ref`. Dug in git a bit but didn't find where they were removed.

benrudolph: interesting, have no idea what this is either

benrudolph (approved):
out with old in with the new

PR #16825, opened 2017-06-28
Drop static/reports/javascripts

orangejenny:
There aren't any references to daterangepicker.min.js, and the only references to moment.min.js are in cloudcare/formplayer, which pull it from bower.

@benrudolph


benrudolph (approved):
👏

PR #16827, opened 2017-06-28
BatchRecord

benrudolph:
@gcapalbo @snopoke @calellowitz the long awaited Batch model! here's a first go to see how you guys feel about the direction. (i haven't forgotten about the other comments you've made in the other PRs, i'm tracking them in a trello). note this PR goes into the user group dim PR

cc: @orangejenny


corehq/warehouse/management/commands/commit_table.py
```diff
@@ -15,39 +15,40 @@
...
             raise CommandError('{} is not a valid slug. \n\n {}'.format(slug, USAGE))
-        model.commit(start, end)
+
+        commit_record = CommitRecord(
+            slug=slug,
+            batch_record=batch_record,
+            verified=False,
+        )
+        try:
+            commit_record.verified = model.commit(batch_record)
```
benrudolph: gio i'm thinking that the `commit` function can return True or False based on whether or not the commit is valid. this is based on the conversation we had some time ago about counting the number of records entered and checking to ensure it was the expected amount. (this functionality has not been implemented)

corehq/warehouse/models/meta.py
```diff
@@ -1,61 +1,25 @@
...
-    @transition(
-        field=state,
-        source='*',
-        target=FACT_TABLE_DECOMMISSIONED,
-    )
-    def decommission(self):
-        return
+
+class BatchRecord(models.Model):
+    batch_id = models.UUIDField(unique=True, db_index=True, primary_key=True)
```
benrudolph: The reason i'm going with a UUID instead of using the auto incrementing field is that I'm imagining that if we have a separate system to run the warehouse, it will call `create_batch` with some UUID that it can use throughout the warehouse process.

corehq/warehouse/transforms/sql/user_group_dim.sql
```diff
@@ -13,7 +14,8 @@ SELECT
     gd.id,
     now(),
     now(),
-    combined_user_group_table.deleted
+    combined_user_group_table.deleted,
+    '{{ batch_id }}'
```
benrudolph: i know we've talked about some optimizations in regards to the location dim using the batch_id. i'm tracking that, but haven't implemented it in this PR

benrudolph (commented):
The process for updating the warehouse now looks something like this:
```
./manage.py create_batch 222617b9-8cf0-40a2-8462-7f872e1f1344 -s 2012-05-05 -e 2018-06-05
./manage.py commit_table user_staging 222617b9-8cf0-40a2-8462-7f872e1f1344
./manage.py commit_table user_dim 222617b9-8cf0-40a2-8462-7f872e1f1344
```

corehq/warehouse/management/commands/commit_table.py
gcapalbo: ok, so you mean if the balance fails then it returns False? and then does that fail the job

corehq/warehouse/models/meta.py
gcapalbo: what if we just had the create batch process take no args and return the auto-incrementing id to use throughout the rest of the process? for example, it would just insert a record into this table and return the auto-incrementing pk that postgres gave it.

i think if we use uuids it might make it more difficult to debug / quickly identify which records belong to a given batch.

corehq/warehouse/management/commands/clear_staging_records.py
```diff
@@ -0,0 +1,21 @@
+from django.core.management import BaseCommand
+from corehq.warehouse.const import STAGING_TABLES
+from corehq.warehouse.models import get_cls_by_slug
+
+
+USAGE = """Usage: ./manage.py clear_staging_records
+
+Removes all records from staging tables
```
gcapalbo: will this be called by the warehouse load process? it might be better to clear one staging table at a time right before loading the respective dim/fact, in case we ever need to restart any individual jobs

corehq/warehouse/management/commands/commit_table.py
```diff
@@ -15,39 +15,40 @@
...
+        except BatchRecord.DoesNotExist:
+            raise CommandError('Invalid batch ID: {}'.format(batch_id))
 
         try:
             model = get_cls_by_slug(slug)
         except KeyError:
             raise CommandError('{} is not a valid slug. \n\n {}'.format(slug, USAGE))
-        model.commit(start, end)
+
+        commit_record = CommitRecord(
```
gcapalbo: :+1: good idea to keep track of this. it might also help for debugging if we keep track of the number of records pulled from the source and loaded into the target if it's not too hard to do that. that can help drive the balancing process

corehq/warehouse/management/commands/commit_table.py
```diff
@@ -15,39 +15,40 @@
...
+            slug=slug,
+            batch_record=batch_record,
+            verified=False,
```
gcapalbo: would it be useful to store the current timestamp on this record too? just for logging purposes

corehq/warehouse/management/commands/create_batch.py
```diff
@@ -0,0 +1,51 @@
...
+from corehq.warehouse.models import BatchRecord
+
+
+USAGE = """Usage: ./manage.py create_batch <batch_id> -s <start_datetime> -e <end_datetime>
+"""
+
+
+class Command(BaseCommand):
+    """
+    Example: ./manage.py create_batch 222617b9-8cf0-40a2-8462-7f872e1f1344 -s 2017-05-01 -e 2017-06-01
```
gcapalbo: i think the start timestamp would always just be the end timestamp of the previous batch run, right? (since we always pull start date exclusive, end date inclusive) otherwise it might be hard for an external process to keep track of the last successful run in the event the system is down or something

similarly, should we be setting end date automatically?

corehq/warehouse/models/meta.py
```diff
@@ -1,61 +1,25 @@
...
+
+    start_datetime = models.DateTimeField()
+    end_datetime = models.DateTimeField()
+
+    created_on = models.DateTimeField(auto_now_add=True)
```
gcapalbo: would it be useful to also track the start and end timestamps of the entire batch run itself? just so we can track run times over time

corehq/warehouse/tests/test_dimensions.py
```diff
@@ -28,6 +32,16 @@
 )
 
 
+def setup_module():
```
gcapalbo: cool, i didn't know about this

corehq/warehouse/models/meta.py
```diff
@@ -1,61 +1,25 @@
...
+
+
+class CommitRecord(models.Model):
+    '''
+    A CommitRecord records meta data about a certain warehouse table's
+    batch.
+    '''
+    batch_record = models.ForeignKey('BatchRecord', on_delete=models.PROTECT)
```
gcapalbo: small thing but could we call this `batch` to match the rest of the foreign keys to `BatchRecord`

gcapalbo:
:+1: lookin good

corehq/warehouse/management/commands/clear_staging_records.py
benrudolph: this is actually just for convenience purposes. i have to do this a lot, so felt like just adding a mgmt command to do it

corehq/warehouse/management/commands/commit_table.py
benrudolph: yep i think so, 👍

corehq/warehouse/management/commands/commit_table.py
benrudolph: right now it doesn't fail the job, but that would make sense to do

corehq/warehouse/management/commands/create_batch.py
benrudolph: yeah this is a good point and may need some further investigation. i was under the impression that airflow handles / stores all that information. on one hand it's quite nice to be able to specify the dates because it makes the system less stateful, on the other hand it could introduce complexities on the airflow side of things. we could also make assertions about allowed date ranges based on the previous batch to make sure we don't do anything funny like skip dates

corehq/warehouse/models/meta.py
benrudolph: how are you envisioning the management command returns the number? i had thought that management commands just return a certain exit code

gcapalbo: yeah good point. ideally if we can figure out a way so that we have auto incrementing batch ids that would be preferable over uuids i think.

benrudolph: we may be able to do it by making a web route that is something like `get_latest_batch_id`

corehq/warehouse/models/meta.py
benrudolph: actually i have to think about this a bit more. i think we'll have to have a separate mgmt command that is something like `batch_completed` to record the time

benrudolph:
OK tried to clean this up as much as I could before heading out. The [commit](https://github.com/dimagi/commcare-hq/pull/16827/commits/d7ebb6dccfefc3e92b2dec714305a4802a69952d) that changes `clear_records` to use `all().delete()` is a bit sketchy. Didn't have time to get to the bottom of it, but my hunch is that it was causing tests to fail to use truncate because of the way that django tests handle transactions. The error message was: 
```
OperationalError: cannot TRUNCATE "warehouse_groupstagingtable" because it has pending trigger events
```
In addition I wasn't able to address all the outstanding feedback so I made 4 additional tasks to handle when I get back (or Cal can tackle if he gets the chance).

1. Make Batch use auto incrementing IDs
1. Relook at how verified works for committing records
1. Create management to mark the finish of a batch
1. Discuss auto date setting for batch creation

corehq/warehouse/utils.py
```diff
@@ -20,6 +21,9 @@ def django_batch_records(cls, record_iter, field_mapping, batch_id):
 
 
 def truncate_records_for_cls(cls, cascade=False):
+    if settings.UNIT_TESTING:
+        cls.objects.all().delete()
+        return
```
gcapalbo: why is this needed?

gcapalbo: never mind, saw your comment

gcapalbo (approved):
i'm ok with merging and tracking the outstanding things separately

corehq/warehouse/tests/test_staging.py
```diff
@@ -26,6 +28,7 @@ def teardown_module():
     Batch.objects.all().delete()
 
 
+@partitioned
```
gcapalbo: maybe we should make a BaseWarehouseTestCase so that we don't have to set the partitioned decorator on each TestCase

corehq/warehouse/tests/test_staging.py
```diff
@@ -79,6 +82,7 @@ class TestGroupStagingTable(BaseStagingTableTest, StagingRecordsTestsMixin):
 
     @classmethod
     def setUpClass(cls):
+        cls.using = 'warehouse' if settings.USE_PARTITIONED_DATABASE else 'default'
```
gcapalbo: might be better if this referenced `WAREHOUSE_DATABASE_ALIAS` instead of `'warhouse'`

this would also be a good thing to add to the BaseWarehouseTestCase

gcapalbo (approved):

calellowitz:
@gcapalbo i finally got travis tests passed on the last commit. is this good to go?

gcapalbo:
thanks @calellowitz . yeah the outstanding items from this PR are being tracked in trello

PR #16830, opened 2017-06-28
fix Repeater caching

dannyroberts:
This hasn't yet been a problem, but only because we don't have interfaces for editing repeaters. Once we do (pretty necessary for a more complex configuration like the one we'll need for openmrs), having this PR's fix in will make it so that changes on save go into effect immediately (rather than remaining stale in the cache for some time).


dannyroberts:
@proteusvacuum (small) @NoahCarnahan

NoahCarnahan (approved):

PR #16831, opened 2017-06-28
App manager v2: reg form changes

orangejenny:
Add ability to add a new reg form, and remove the restriction on deleting reg forms.

@biyeun / @benrudolph


orangejenny:
Added a commit to allow rearranging reg forms, which will fix https://manage.dimagi.com/default.asp?256712

I had done this in https://github.com/dimagi/commcare-hq/pull/16769 but rolled it back before deploying because it wasn't working; turned out I'd missed removing a `sort-disabled` class.

orangejenny:
pinging again for review @biyeun or new code buddies @kaapstorm @gcapalbo

biyeun (approved):

biyeun:
@orangejenny looks good.

PR #16834, opened 2017-06-28
ICDS: Remove elasticsearch from the process that spawns run_indicator tasks

gcapalbo:
ES is having trouble fetching the results for this part of the process. I don't think it will be too bad for couch to handle it, and it's done serially anyway.

@emord 
cc @kaapstorm


emord (approved):
Looks good from me.I don't think there was a particular reason I chose to use ES other than there were other ES methods already written

> I don't think it will be too bad for couch to handle it

Can't be too confident these days.

PR #16836, opened 2017-06-28
fix experience editing case import comment while in progress

dannyroberts:
before it would kick you out of the box when the upload list refreshed

fixes https://manage.dimagi.com/default.asp?256011

### DIFF
--- a/corehq/apps/case_importer/static/case_importer/js/import_history.js
+++ b/corehq/apps/case_importer/static/case_importer/js/import_history.js
@@ -18,8 +18,11 @@ hqDefine('case_importer/js/import_history.js', function () {
         };
         var taskStatusesInDataMatchCurrent = function (data) {
             return (
-                _.chain(self.case_uploads()).pluck('task_status').map(function (task_status) {return task_status()})
-                            .isEqual(_(data).pluck('task_status').map(function (task_status) {return task_status()})).value()
+                _.chain(self.case_uploads()).pluck('task_status').map(function (task_status) {
+                    return task_status();
+                }).isEqual(_(data).pluck('task_status').map(function (task_status) {
+                    return task_status();
+                })).value()
             );
         };
         self.updateCaseUploads = function (data) {
--- a/corehq/apps/case_importer/static/case_importer/js/import_history.js
+++ b/corehq/apps/case_importer/static/case_importer/js/import_history.js
@@ -22,28 +22,23 @@ hqDefine('case_importer/js/import_history.js', function () {
                             .isEqual(_(data).pluck('task_status').map(function (task_status) {return task_status()})).value()
             );
         };
-        var shouldUpdate = function (data) {
-            // do not update DOM
-            // if we're not either adding new uploads or updating the status
-            // this prevents some jumpiness when not necessary
-            // and is particularly bad if you're in the middle of editing a comment
-            return !(uploadIdsInDataMatchCurrent(data) && taskStatusesInDataMatchCurrent(data));
-        };
         self.updateCaseUploads = function (data) {
-            if (uploadIdsInDataMatchCurrent(data)) {
-                // in the easy case, update just the essential information (task_status) in place
-                // this prevents some jumpiness when not necessary
-                // and is particularly bad if you're in the middle of editing a comment
-                _.each(_.zip(self.case_uploads(), data), function (pair) {
-                    var case_upload = pair[0];
-                    var new_case_upload = pair[1];
-                    if (case_upload.upload_id !== new_case_upload.upload_id) {
-                        throw {message: "Somehow even after checking, the case upload lists didn't line up."};
-                    }
-                    case_upload.task_status(new_case_upload.task_status());
-                });
-            } else {
-                self.case_uploads(data);
+            if (!uploadIdsInDataMatchCurrent(data) || !taskStatusesInDataMatchCurrent(data)) {
+                if (uploadIdsInDataMatchCurrent(data)) {
+                    // in the easy case, update just the essential information (task_status) in place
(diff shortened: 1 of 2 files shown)


corehq/apps/case_importer/static/case_importer/js/import_history.js
```diff
@@ -19,7 +19,25 @@ hqDefine('case_importer/js/import_history.js', function () {
             // this prevents some jumpiness when not necessary
             // and is particularly bad if you're in the middle of editing a comment
             return !(_.chain(self.case_uploads()).pluck('upload_id').isEqual(_(data).pluck('upload_id')).value() &&
-                _.chain(self.case_uploads()).pluck('task_status').isEqual(_(data).pluck('task_status')).value());
+                _.chain(self.case_uploads()).pluck('task_status').map(function (task_status) {return task_status()})
```
dannyroberts: Thanks, stickler, good point.

dannyroberts:
(I think something's whack with those stickler comments; one's an expression that continues on the next line, and the other... has a semicolon.)

dannyroberts:
@NoahCarnahan

corehq/apps/case_importer/static/case_importer/js/import_history.js
```diff
@@ -19,7 +19,25 @@ hqDefine('case_importer/js/import_history.js', function () {
...
+                    .isEqual(_(data).pluck('task_status').map(function (task_status) {return task_status()})).value());
+        };
+        self.updateCaseUploads = function (data) {
```
NoahCarnahan: Is there a more descriptive name you could use for this argument? Would `newCaseUploads` or something like that be accurate?

corehq/apps/case_importer/static/case_importer/js/import_history.js
```diff
@@ -19,7 +19,25 @@ hqDefine('case_importer/js/import_history.js', function () {
             // this prevents some jumpiness when not necessary
             // and is particularly bad if you're in the middle of editing a comment
             return !(_.chain(self.case_uploads()).pluck('upload_id').isEqual(_(data).pluck('upload_id')).value() &&
-                _.chain(self.case_uploads()).pluck('task_status').isEqual(_(data).pluck('task_status')).value());
+                _.chain(self.case_uploads()).pluck('task_status').map(function (task_status) {return task_status()})
+                    .isEqual(_(data).pluck('task_status').map(function (task_status) {return task_status()})).value());
```
NoahCarnahan: This is a pretty gnarly statement to parse. When I have a big boolean statement like this, I sometimes like to break it up into smaller parts and use variable names to help the reader understand what's going on. For instance, you could do something like this:
```js
var shouldUpdate = function (data) {
	var uploadIdsAreEqual = _.chain(self.case_uploads())
		.pluck('upload_id')
		.isEqual(_(data).pluck('upload_id'))
		.value()
	var newTaskStatuses = _(data).pluck('task_status').map(function (task_status) {
		return task_status()
	})
	var taskStatusesAreEqual = _.chain(self.case_uploads())
		.pluck('task_status')
		.map(function (task_status) {
			return task_status()
		})
		.isEqual(newTaskStatuses)
		.value());
	return !(uploadIdsAreEqual && taskStatusesAreEqual);
```
It's much easier to read `return !(uploadIdsAreEqual && taskStatusesAreEqual);` than it is the whole statement at once.

dannyroberts: Thanks, very fair. I definitely let this line grow out of control. Fix coming right up!

corehq/apps/case_importer/static/case_importer/js/import_history.js
dannyroberts: The idea in my head at least is that this function is _conceptually_ equivalent to just `self.case_uploads()` (which it calls when it can't do the easy optimization), but one that has better behavior in terms of updating the DOM in place instead of replacing it entirely.

That's where the name comes from, but I agree the name leaves the purpose behind its existence a bit of a mystery. Maybe something like `self.updateCaseUploadsWithMinimalEffectOnDOM` would be clearer, which may be worth the verbosity?

NoahCarnahan: Oh sorry I think I was a little unclear. I think `updateCaseUploads` is a fine name for the function, but I think the name `data` for the argument could be replaced with something that describes what sort of data is being passed to the function.

corehq/apps/case_importer/static/case_importer/js/import_history.js
```diff
@@ -13,13 +13,33 @@ hqDefine('case_importer/js/import_history.js', function () {
...
-            // and is particularly bad if you're in the middle of editing a comment
-            return !(_.chain(self.case_uploads()).pluck('upload_id').isEqual(_(data).pluck('upload_id')).value() &&
-                _.chain(self.case_uploads()).pluck('task_status').isEqual(_(data).pluck('task_status')).value());
+        var uploadIdsInDataMatchCurrent = function (data) {
+            return _.chain(self.case_uploads()).pluck('upload_id').isEqual(_(data).pluck('upload_id')).value();
+        };
+        var taskStatusesInDataMatchCurrent = function (data) {
+            return (
+                _.chain(self.case_uploads()).pluck('task_status').map(function (task_status) {return task_status()})
+                            .isEqual(_(data).pluck('task_status').map(function (task_status) {return task_status()})).value()
```
dannyroberts: oh i get it. it's complaining about the missing semicolon after `return task_status()`. Tricky. Fixed.

NoahCarnahan (approved):
Thanks for the update!

PR #16839, opened 2017-06-29
User Group Dim 2

benrudolph:


benrudolph:
TRAVIS!

corehq/warehouse/transforms/sql/user_group_dim.sql
```diff
@@ -13,31 +23,18 @@ SELECT
...
-    SELECT
-        UNNEST(removed_user_ids) AS removed_user_id,
-        group_id AS group_id,
-        true AS deleted
+        group_id AS group_id
     FROM
         {{ group_staging }}
-)
-) combined_user_group_table
+    WHERE doc_type = 'Group'
```
gcapalbo: i don't think the subquery is needed - looks like we can join normally and then unnest() in the main select

gcapalbo: (never mind, my idea won't work, fine to leave as is)

benrudolph: (offlined, going to leave as is for now and address later with intermediate table)

gcapalbo (approved):

PR #16219, opened 2017-05-18
pillow work for new app status report

calellowitz:
@snopoke @benrudolph 
This adds a synclog pillow and fixes up the form metadata pillow to make the new app status report work. it must be merged before that pr

PR #16840, opened 2017-06-29
fix docupdate conflict

sravfeyn:
https://sentry.io/dimagi/commcarehq/issues/242460043/

This issue started occurring after app status report was [recently updated](https://github.com/dimagi/commcare-hq/pull/16219/files#diff-fa0e8af5f7775e2be761383c7e5b33c3R85). I am not sure if there is a better solution than the fix I made
 
FYI @benrudolph @calellowitz

### DIFF
--- a/corehq/apps/ota/utils.py
+++ b/corehq/apps/ota/utils.py
@@ -2,6 +2,8 @@ import time
 from django.utils.translation import ugettext as _
 from casexml.apps.case.xml import V2
 from casexml.apps.phone.restore import RestoreConfig, RestoreParams
+from couchdbkit import ResourceConflict
+
 from corehq.apps.domain.models import Domain
 from corehq.apps.users.util import format_username
 from corehq.apps.users.models import CommCareUser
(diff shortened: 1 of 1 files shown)


sravfeyn:
Wait for the build https://travis-ci.org/dimagi/commcare-hq/builds/248358886

benrudolph:
do you think we could use either locking (something like `CriticalSection`) or attempt to save the doc and catch the document update error and when that happens refetch the doc and try to save again?

sravfeyn:
That's much better, I have updated it

corehq/apps/ota/utils.py
```diff
@@ -62,9 +63,15 @@ def reset_demo_user_restore(commcare_user, domain):
...
-    commcare_user.demo_restore_id = demo_restore.id
-    commcare_user.save()
+    # Set reference to new restore
+    try:
+        commcare_user.demo_restore_id = demo_restore.id
+        commcare_user.save()
+    except ResourceConflict:
+        # If CommCareUser.report_metadata gets updated by sync log pillow, after a restore is
+        #   generated or for any other reason, there will be a DocumentUpdate conflict
+        commcare_user = CommCareUser.get(commcare_user.get_id)
```
benrudolph: sweet, i think you'll still need to set the `demo_restore_id`

sravfeyn: Sorry, thanks.

benrudolph (approved):

PR #16842, opened 2017-06-29
Tweak display conditions

orangejenny:
Allow better resizing now that these don't use the value or non ui anymore.

@benrudolph / @proteusvacuum

### DIFF
--- a/corehq/apps/app_manager/templates/app_manager/v1/partials/form_filter.html
+++ b/corehq/apps/app_manager/templates/app_manager/v1/partials/form_filter.html
@@ -10,7 +10,7 @@
     </label>
     <div id="form-filter" class="col-sm-4 row">
         <div data-bind="css: {'has-error': !allowed()}">
-            <textarea name="form_filter" class="form-control vertical-resize" data-bind="
+            <textarea name="form_filter" class="form-control" data-bind="
                     xpathValidator: {text: formFilter, allowCaseHashtags: true,
                                      errorHtml: $('#formFilterXpathErrorHtml').html()}
             "></textarea>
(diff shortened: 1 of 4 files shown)


corehq/apps/app_manager/templates/app_manager/v2/partials/form_filter.html
```diff
@@ -12,7 +12,7 @@
     </label>
     <div id="form-filter" class="col-sm-4 row">
         <div data-bind="css: {'has-error': !allowed()}">
-            <textarea name="form_filter" class="form-control vertical-resize" data-bind="
+            <textarea name="form_filter" class="form-control" data-bind="
```
proteusvacuum: does this need spellcheck?

corehq/apps/app_manager/templates/app_manager/v1/partials/form_filter.html
```diff
@@ -10,7 +10,7 @@
     </label>
     <div id="form-filter" class="col-sm-4 row">
         <div data-bind="css: {'has-error': !allowed()}">
-            <textarea name="form_filter" class="form-control vertical-resize" data-bind="
+            <textarea name="form_filter" class="form-control" data-bind="
```
proteusvacuum: does this need spellcheck?

proteusvacuum (approved):
yay

corehq/apps/app_manager/templates/app_manager/v1/partials/form_filter.html
orangejenny: oops, thought i took care of that, clearly not

orangejenny: took a second look, this is already taken care of, because form filter uses the xpath validator, which turns off spellcheck.

proteusvacuum: that is very confusing, thanks for looking into it!

PR #16844, opened 2017-06-29
ICDS - Update Aggregation and View for Dashboard v2

sheelio:
@emord 

This does a few things:
- Adds new columns to the existing aggregate tables
- Updates the views for those aggregate tables
- Adds a new aggregate table (for daily tracking)

One open TODO is purging any daily tables > 15 days old.  Not sure the best way to do that.


sheelio:
@emord 
It looks like @kkrampa has added a migration to specify views that are already present in the aggregation scripts (they're on master, not in this PR).  Is that going to create issues with the views that are being manually created?

emord:
@sheelio do you mean these? https://github.com/dimagi/commcare-hq/blob/master/custom/icds_reports/migrations/0016_aggawcmonthly_aggccsrecordmonthly_aggchildhealthmonthly_aggdailyusageview_aggthrmonthly_awclocation_.py

It looks like managed = False so it should be fine https://docs.djangoproject.com/en/1.11/ref/models/options/#managed

custom/icds_reports/migrations/0016_add_v2_indicators.py
```diff
@@ -0,0 +1,21 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import migrations, models
+from corehq.sql_db.operations import RawSQLMigration
+
+migrator = RawSQLMigration(('custom', 'icds_reports', 'migrations', 'sql_templates'))
+
+
+class Migration(migrations.Migration):
```
emord: You'll probably need to rebase branch this onto master and rename this migration and change the dependency

emord (changes requested):
I think I was wrong actually. Since you're add columns you'll probably want to change the model definitions that he wrote. and make another migration

PR #16845, opened 2017-06-29
Fix validation error on form-based report builder configs

esoergel:
Getting a validation error here: https://manage.dimagi.com/default.asp?257168
Introduced in https://github.com/dimagi/commcare-hq/pull/16602

@biyeun, can you verify that this is appropriate?

@NoahCarnahan @biyeun cc @emord

### DIFF
--- a/corehq/apps/userreports/reports/builder/forms.py
+++ b/corehq/apps/userreports/reports/builder/forms.py
@@ -1490,7 +1490,7 @@ class ConfigureMapReportForm(ConfigureListReportForm):
             report_name, app_id, source_type, report_source_id, existing_report, *args, **kwargs
         )
         if self.source_type == "form":
-            self.fields['group_by'].widget = QuestionSelect(
+            self.fields['location'].widget = QuestionSelect(
                 attrs={'class': 'input-large'},
                 ko_value='groupBy'
             )
--- a/corehq/apps/userreports/reports/builder/forms.py
+++ b/corehq/apps/userreports/reports/builder/forms.py
@@ -86,16 +86,14 @@ class Select2(Widget):
         self.choices = list(choices)
 
     def render(self, name, value, attrs=None, choices=()):
-        value = '' if value is None else value
+        self.value = '' if value is None else value
         final_attrs = self.build_attrs(attrs, name=name)
 
         return format_html(
             '<input{0} type="text" data-bind="select2: {1}, {2}">',
             flatatt(final_attrs),
             json.dumps(self._choices_for_binding(choices)),
-            'value: {}'.format(
-                self.ko_value
-            ) if self.ko_value else ""
+            'value: {}'.format(self.ko_value or self.value)
         )
 
     def _choices_for_binding(self, choices):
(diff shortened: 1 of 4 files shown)


NoahCarnahan (changes requested):
Hey, `group_by` is actually relevant for some form based reports. However it is not relevant for form/case list reports, or for chart reports (this type is stored in `report_type`).

corehq/apps/userreports/templates/userreports/reportbuilder/configure_report.html
```diff
@@ -32,6 +32,7 @@
         var reportColumnOptions = {{ report_column_options|JSON }};
         var dateRangeOptions = {{ date_range_options|JSON }};
         var isGroupByRequired = {{ is_group_by_required|JSON }};
+        var groupByInitialValue = {{ form.fields.group_by.widget.value|JSON}};
```
esoergel: This widget's initial value was no longer being populated, so I added that in in this somewhat hacky fashion

esoergel:
@NoahCarnahan ready for review again

NoahCarnahan (approved):
Nice, looks good to me.

corehq/apps/userreports/reports/builder/forms.py
```diff
@@ -1488,7 +1490,10 @@ def __init__(self, report_name, app_id, source_type, report_source_id, existing_
             report_name, app_id, source_type, report_source_id, existing_report, *args, **kwargs
         )
         if self.source_type == "form":
-            self.fields['location'].widget = QuestionSelect(attrs={'class': 'input-large'})
+            self.fields['group_by'].widget = QuestionSelect(
```
kaapstorm: Should this be `self.fields['location'].widget`, considering this is a Map Report, or should line 1498 also be `self.fields['group_by'].widget`?

esoergel: you're absolutely right - I must've just copypasted this.  Thanks!

corehq/apps/userreports/reports/builder/forms.py
```diff
@@ -86,16 +86,14 @@ def __init__(self, attrs=None, choices=(), ko_value=None):
...
         final_attrs = self.build_attrs(attrs, name=name)
 
         return format_html(
             '<input{0} type="text" data-bind="select2: {1}, {2}">',
             flatatt(final_attrs),
             json.dumps(self._choices_for_binding(choices)),
-            'value: {}'.format(
-                self.ko_value
-            ) if self.ko_value else ""
+            'value: {}'.format(self.ko_value or self.value)
```
biyeun: hey, this isn't right. value is specifically the KO variable, not the text value of whatever is contained in the field

esoergel: Right, but when `ko_value` isn't provided, there is no knockout variable to bind to.  This widget is called both in contexts where it's bound to a model, and contexts where it isn't.  This falls back to the old behavior when it's not a knockout-bound widget.  That is, https://github.com/dimagi/commcare-hq/pull/16845/files#diff-66cf616ff719298fed0e5e58ca5165ccL123
Actually I just realized that this should be
```python
'value: {}'.format(self.ko_value) if self.ko_value else 'value: "{}"'.format(self.value) 
```
(I forgot the quotes when it's a literal)  Does that sound right to you?

biyeun: @esoergel if there is no knockout variable to bind to, it needs to be passed to the `value` attr, otherwise `data-bind` has no meaning....

esoergel: https://github.com/dimagi/commcare-hq/pull/16889

PR #16847, opened 2017-06-29
Add new states to user indicator task

gcapalbo:
@emord 
@kaapstorm


custom/icds/tasks.py
```diff
@@ -30,6 +30,8 @@
 BIHAR_SITE_CODE = '10'
 CHHATTISGARH_SITE_CODE = '22'
 JHARKHAND_SITE_CODE = '20'
+RAJASTHAN_SITE_CODE = '08'
+UTTAR_PRADESH_SITE_CODE = 'state9'
```
emord: Seems like a strange site code. assuming this is intentional

emord (approved):

custom/icds/tasks.py
gcapalbo: yeah this one was different from the rest, but matches the source

PR #16849, opened 2017-06-29
Release livequery case sync algorithm behind flag

millerdev:
Bunch of refactoring at the beginning to make livequery function easier to read.

Best reviewed by commit 🐠 

@proteusvacuum @ctsims @dannyroberts @snopoke

### DIFF
--- a/corehq/ex-submodules/casexml/apps/phone/data_providers/case/livequery.py
+++ b/corehq/ex-submodules/casexml/apps/phone/data_providers/case/livequery.py
@@ -109,12 +109,7 @@ def get_payload(timing_context, restore_state):
         return CIRCULAR_REF
 
     CIRCULAR_REF = object()
-    log = logging.getLogger(__name__)
-    if log.isEnabledFor(logging.DEBUG):
-        debug = log.debug
-    else:
-        def debug(*args):
-            return None
+    debug = logging.getLogger(__name__).debug
     live_ids = set()
     extensions_by_host = defaultdict(set)  # host_id -> extension_ids
     hosts_by_extension = defaultdict(set)  # extension_id -> host_ids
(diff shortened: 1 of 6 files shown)


ctsims:
@calellowitz and @esoergel Might also be able to do some review.

corehq/ex-submodules/casexml/apps/phone/data_providers/case/livequery.py
```diff
@@ -58,6 +58,51 @@ def enliven(case_id):
             for next_id in chain(ext_ids, host_ids):
                 enliven(next_id)
 
+    def classify(index):
+        """Classify index as either live or extension with live status pending
+
+        This closuer mutates `next_ids`, `extensions_by_host`,
```
dannyroberts: sometimes closures can be replaced with helper classes (in the case `do_livequery(*args, **kwargs)` would be a wrapper like `return LiveQueryHelper(*args, **kwargs).main()`, and `next_ids` would then be `self.next_ids`), but that doesn't always end up feeling more elegant/easier to reason with. Since there are a few inner functions here and they're somewhat long and involved, I thought I'd at least suggest it and leave it up to you to decide whether it would help with organization or not.

nickpell (approved):

dannyroberts (approved):
@millerdev looking good! Just had that one comment, but certainly not blocking.

corehq/ex-submodules/casexml/apps/phone/data_providers/case/livequery.py
millerdev: That's true, if you're into class-oriented thinking, which we often do at Dimagi. In this case you can think of `do_livequery` sort of like a class, with its local variables being members that would be bound to `self` in a class. The inner closures `enliven` and `classify` could be thought of like private methods that mutate state.

I personally prefer the syntax of function closures to helper classes like the one you described because there is less boilerplate syntax involved: a `LiveQueryHelper` instance has no useful purpose other than to call it's `main()` method, so why not just make it a function?

corehq/ex-submodules/casexml/apps/phone/data_providers/case/livequery.py
```diff
@@ -179,9 +192,28 @@ def take(n, iterable):
         yield accessor.get_cases(next_ids)
 
 
-def compile_response(timing_context, batches, restore_state):
+def init_progress(async_task, total):
+    if not async_task:
+        return lambda done: None
+
+    def update_progress(done):
```
proteusvacuum: I really wonder if this will ever be useful now that this is so fast...

proteusvacuum:
🎉

PR #16852, opened 2017-06-29
using server time and including hh names

calellowitz:
@emord 
Last one for sure


millerdev (approved):

emord (commented):
This looks like you aren't using server time. Is the commit message wrong?

calellowitz:
@emord yep the commit message is definitely wrong. this was a purposeful change to phone time. is there a good way to change that

emord:
That's fine. just wanted to make sure it was intentional

PR #16853, opened 2017-06-29
Store report_class in redis for permissions check

esoergel:
The current behavior checks specifically for access to the case list report...
https://manage.dimagi.com/default.asp?257116


esoergel:
@emord

emord (approved):
Would be nice if you could use namedtuple in redis, but I don't think that's possible right?

esoergel:
I think you can use anything pickle-able, although there's some business about pickle needing to be able to find the class, which I think makes it more complicated.

PR #16855, opened 2017-06-30
allow revoking both is_staff and is_superuser in SuperuserManagement

dannyroberts:
previously if you selected nothing
the form would fail (shallowly) with a validation error


dannyroberts:
@krsdimagi this should get you what you need, once it's deployed

dannyroberts:
@NoahCarnahan

NoahCarnahan (approved):

sravfeyn:
Thanks!

PR #16856, opened 2017-06-30
archive old tables, create new ones

nickpell:
@NoahCarnahan 

I want to import a new data dump, but don't want to drop the existing tables, since I will need to do a diff between them.  So I'm renaming the existing tables and creating new ones.

I would have kept the same names for the most recent tables, but https://code.djangoproject.com/ticket/23577 got in my way, so instead I'm renaming ```Beneficiary``` to ```Beneficiary_Jun30``` etc.


nickpell:
choosing to ignore stickler comments

NoahCarnahan (approved):

PR #16858, opened 2017-06-30
ICDS: Add management command to aggregate sms counts by state and indicator

gcapalbo:
To be used as a stop-gap until we build the report.
@emord 
@kaapstorm

### DIFF
--- a/custom/icds/management/commands/get_icds_sms_usage.py
+++ b/custom/icds/management/commands/get_icds_sms_usage.py
@@ -1,0 +1,102 @@
+import csv
+from corehq.apps.sms.models import SMS
+from corehq.apps.locations.models import SQLLocation
+from corehq.apps.users.models import CommCareUser
+from corehq.form_processor.utils import is_commcarecase
+from corehq.messaging.smsbackends.icds_nic.models import SQLICDSBackend
+from corehq.util.argparse_types import utc_timestamp
+from django.core.management.base import BaseCommand
+
+
+class Command(BaseCommand):
+    help = ""
+
+    def add_arguments(self, parser):
+        parser.add_argument('domain')
+        parser.add_argument('start_timestamp', type=utc_timestamp)
+        parser.add_argument('end_timestamp', type=utc_timestamp)
+
+    def get_location_id(self, sms):
+        location_id = None
+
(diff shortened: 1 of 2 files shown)


custom/icds/management/commands/get_icds_sms_usage.py
```diff
@@ -0,0 +1,102 @@
...
+        return sms.custom_metadata.get('icds_indicator', 'unknown')
+
+    def handle(self, domain, start_timestamp, end_timestamp, **options):
+        self.recipient_id_to_location_id = {}
+        self.location_id_to_location = {}
+        self.location_id_to_state_code = {}
+
+        data = {}
+
+        for sms in SMS.objects.filter(
```
emord: Does this need to be paginated in some way? It'll return 50k rows right?

custom/icds/management/commands/get_icds_sms_usage.py
```diff
@@ -0,0 +1,102 @@
...
+            return 'unknown'
+
+        self.location_id_to_state_code[location.location_id] = state.site_code
+        return state.site_code
+
+    def get_indicator_slug(self, sms):
+        if not isinstance(sms.custom_metadata, dict):
+            return 'unknown'
+
+        return sms.custom_metadata.get('icds_indicator', 'unknown')
```
emord: is it important to distinguish between not having custom_metadata and the custom_metadata not having icds_indicator? May be nice to know if there are a lot of 'unknown' showing up

emord (approved):
looks good from me unless you want to paginate that query

custom/icds/management/commands/get_icds_sms_usage.py
gcapalbo: there actually shouldn't be any without custom_metadata since monday (there will be a lot of existing ones before that without it which i'm planning to update in a separate script at some point). the unknown tag here is just a way for me to be alerted to this so i can investigate further if it happens

custom/icds/management/commands/get_icds_sms_usage.py
gcapalbo: the volume should be low enough that we should be ok - this is only to be used for the next month or so until we build the report

PR #16859, opened 2017-06-30
LiveQuery for Formplayer

wpride:
Pass the `LIVEQUERY_ENABLED` flag into Formplayer

cross-request: https://github.com/dimagi/formplayer/pull/426

cc @millerdev @ctsims


corehq/toggles.py
```diff
@@ -849,6 +849,13 @@ def _commtrackify(domain_name, toggle_is_enabled):
     [NAMESPACE_DOMAIN],
 )
 
+USE_LIVE_QUERY = StaticToggle(
```
proteusvacuum: Can you call this `FORMPLAYER_LIVE_QUERY` or something? And maybe put it below the other `LIVEQUERY` flag? This will be confusing otherwise.

proteusvacuum (changes requested):

corehq/toggles.py
wpride: good call

proteusvacuum (approved):

corehq/toggles.py
```diff
@@ -849,6 +849,13 @@ def _commtrackify(domain_name, toggle_is_enabled):
     [NAMESPACE_DOMAIN],
 )
 
+FORMPLAYER_USE_LIVEQUERY = StaticToggle(
+    'use_live_query',
```
millerdev: not sure if this is blocking or not, but I think it's convention to have the slug be the lowercase version of the variable name, so this should be `'formplayer_use_livequery'`

wpride: Hey is that Daniel Miller or Stickler cI

millerdev (approved):

wpride:
Failure was due to skipped build, merging

PR #16860, opened 2017-06-30
enable phone number but no AB test

biyeun:
@calellowitz // cc @dannyroberts 
comms wants the phone number back and wants everyone to see it right now. still not a permanent change though.


dannyroberts:
👍

PR #16861, opened 2017-06-30
Isolate dhis2

dannyroberts:
@kaapstorm 

I wanted to start grouping integration code together, and saw that a lot of the dhis2 UI straddled `dhis2` and `domain` apps. I took the liberty in this PR of moving the parts that were in `domain` back into `dhis2` so it could act more like a unit. (The actual user-facing navigation, UIs, urls, etc. don't change at all, just where in the code those things are defined.)

Let me know what you think.


kaapstorm (approved):

kaapstorm:
Yes! :+1: :+1: I think this is the right approach.

proteusvacuum:
Looks good! The repeater report is also in `domain` for some reason. Would be good to get that out of there.

dannyroberts:
Right, noted. I'll file that away for a later move.

PR #16863, opened 2017-06-30
Motech package

dannyroberts:
this is on top of https://github.com/dimagi/commcare-hq/pull/16861


kaapstorm (approved):
Strongly approve of where this is going. I also appreciate that the repeaters app is no longer called "receiverwrappers". :upside_down_face:

corehq/motech/repeaters/dbaccessors.py
```diff
@@ -41,7 +41,7 @@ def get_overdue_repeat_record_count(overdue_threshold=datetime.timedelta(minutes
     from .models import RepeatRecord
     overdue_datetime = datetime.datetime.utcnow() - overdue_threshold
     results = RepeatRecord.view(
-        "receiverwrapper/repeat_records_by_next_check",
+        "repeaters/repeat_records_by_next_check",
```
proteusvacuum: lol. curious about the history of this.

proteusvacuum:
tests didn't run.

dannyroberts:
thanks @proteusvacuum ✅

corehq/motech/repeaters/dbaccessors.py
dannyroberts: Yeah, let's see. So we long ago had a submodule that had django views for the form submission API, and it was called `receiver`. Since at the time that code was shared between commcarehq and another project, when we wanted to adapt it for commcarehq, we made an app called `receiverwrapper` that presented slightly different django views but then called the same logic. Since `receiver` already basically wrapped `couchforms`, at some point I decided it was ridiculous and moved everything that was in `receiver` to either `couchforms` or (mostly, I think) `receiverwrapper`. Meanwhile, and I forget the exact timing, we had also introduced repeaters, which were at the time thought of as sort of an add on functionality to receiving form submissions, so they went in `receiverwrapper` as well. When repeaters were moved out, the couch views were moved to their own "couchapp" (our area for views not associated with any django app), and kept the name receiverwrapper, presumably so as not to have to change all of the references to it.

proteusvacuum: very interesting! thanks :)

kaapstorm: :grinning: Thank you.

PR #16864, opened 2017-06-30
Webapps Permissions by user data

proteusvacuum:
This allows you to set which webapps you see by user data.

The primary reason for this is to only sync reports to eNikshay users that are actually pertinent to the apps that they will use. 
For now, behind a flag until we can make this play nicely in the WebApps Login-As UI.

🐡 / 🐡 

@snopoke buddy
FYI @ctsims 
@benrudolph (we talked about this before you went offline)

### DIFF
--- a/corehq/apps/app_manager/fixtures/mobile_ucr.py
+++ b/corehq/apps/app_manager/fixtures/mobile_ucr.py
@@ -57,7 +57,11 @@ class ReportFixturesProvider(FixtureProvider):
         app_aware_sync_app = restore_state.params.app
         if app_aware_sync_app:
             apps = [app_aware_sync_app]
-        elif restore_state.params.device_id and "WebAppsLogin" in restore_state.params.device_id:
+        elif (
+                toggles.USERDATA_WEBAPPS_PERMISSIONS.enabled(restore_user.domain)
+                and restore_state.params.device_id
+                and "WebAppsLogin" in restore_state.params.device_id
+        ):
             # Only sync reports for apps the user has access to if this is a restore from webapps
             access = ApplicationAccess.get_by_domain(restore_user.domain)
             apps = [
--- a/corehq/apps/app_manager/fixtures/mobile_ucr.py
+++ b/corehq/apps/app_manager/fixtures/mobile_ucr.py
@@ -8,6 +8,7 @@ from lxml.builder import E
 
 from casexml.apps.phone.fixtures import FixtureProvider
 from corehq import toggles
+from corehq.apps.cloudcare.models import ApplicationAccess
 from corehq.apps.app_manager.models import ReportModule
 from corehq.apps.app_manager.suite_xml.features.mobile_ucr import is_valid_mobile_select_filter_type
 from corehq.apps.userreports.reports.filters.factory import ReportFilterFactory
@@ -42,7 +43,6 @@ def _should_sync(restore_state):
     )
 
 
-
 class ReportFixturesProvider(FixtureProvider):
     id = 'commcare:reports'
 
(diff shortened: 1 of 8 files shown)


corehq/apps/app_manager/fixtures/mobile_ucr.py
```diff
@@ -54,8 +54,23 @@ def __call__(self, restore_state):
         if not toggles.MOBILE_UCR.enabled(restore_user.domain) or not _should_sync(restore_state):
             return []
 
-        app = restore_state.params.app
-        apps = [app] if app else [a for a in get_apps_in_domain(restore_user.domain, include_remote=False)]
+        app_aware_sync_app = restore_state.params.app
+        if app_aware_sync_app:
+            apps = [app_aware_sync_app]
+        elif restore_state.params.device_id and "WebAppsLogin" in restore_state.params.device_id:
+            # Only sync reports for apps the user has access to if this is a restore from webapps
```
esoergel: just curious - why only webapps?

esoergel (approved):

corehq/apps/app_manager/fixtures/mobile_ucr.py
proteusvacuum: Because the app visibility permission is only for webapps. 
The phone already does "app aware sync".
The fallback case is if you are syncing from the admin restore page or similar.

corehq/apps/app_manager/fixtures/mobile_ucr.py
```diff
@@ -54,8 +54,27 @@ def __call__(self, restore_state):
...
+            apps = [
+                app for app
+                in get_apps_in_domain(restore_user.domain, include_remote=False)
+                if access.user_can_access_app(restore_user, app)
+            ]
+        else:
+            apps = [
+                app for app
+                in get_apps_in_domain(restore_user.domain, include_remote=False)
+            ]
```
proteusvacuum: oh, hah, I just copied it from line 58

corehq/apps/cloudcare/models.py
```diff
@@ -68,34 +87,59 @@ def get_by_domain(cls, domain):
         return self or cls(domain=domain)
 
     def user_can_access_app(self, user, app):
-        user_id = user['_id']
         app_id = app['_id']
-        if not self.restrict or user['doc_type'] == 'WebUser':
+        app_ids = (app_id, app['copy_of'] or ())
```
proteusvacuum: I don't know. It was there before, which isn't a good excuse, but I was trying to make as few changes as possible

corehq/apps/app_manager/fixtures/mobile_ucr.py
esoergel: Oh cool, thanks for the explanation

corehq/apps/cloudcare/models.py
```diff
@@ -68,34 +87,59 @@ def get_by_domain(cls, domain):
...
+
+        return self.group_permitted(app_ids, user) or self.userdata_permitted(app_ids, user)
+
+    def group_permitted(self, app_ids, user):
         app_group = None
         for app_group in self.app_groups:
-            if app_group.app_id in (app_id, app['copy_of'] or ()):
+            if app_group.app_id in app_ids:
                 break
         if app_group:
```
snopoke: If none of the apps match won't `app_group` just be the last one in the list?

snopoke (commented):
Curious why you went with custom user data vs user roles

corehq/apps/cloudcare/models.py
proteusvacuum: Yeah, this is a really weird construct. I didn't actually write this, and it's been in prod since 2012, so I left it as is. But yes..

proteusvacuum:
I'm actually asking myself the same thing now. I haven't turned this on for eNikshay yet, so I'm going to spend some time updating it to use roles instead of userdata.

proteusvacuum:
https://github.com/dimagi/commcare-hq/pull/16878

PR #16865, opened 2017-06-30
Always enable the ENIKSHAY toggle for enikshay

proteusvacuum:
@snopoke 
@esoergel, we don't need to be able to quickly turn this one off, right?


esoergel:
I guess this is okay.  Originally I was thinking this would be nice to toggle this at will, although I think the flag now controls more than just the user setup stuff.

esoergel (approved):

PR #16866, opened 2017-06-30
Externalize JavaScript: reports standard base template

orangejenny:
@millerdev / @dannyroberts 

Review by commit. "hold off on merge" tag so it doesn't get deployed when I'm not around (want to sanity check the custom reports after deploying).


dannyroberts (approved):
Nice work! Certainly hard to find errors through review, but nothing popped out at me. Nice change

PR #16868, opened 2017-07-01
Run async rebuild queue during off hours

emord:
@dimagi/scale-team this will run the the UCR async queue only during off hours


snopoke (commented):
Seems ok. Will the queue be able to keep up?

emord:
@snopoke no it's not able to keep up, but I think this has been adversly affecting pg0. going to offline with sheel today about possible solutions

PR #16871, opened 2017-07-02
add ability to add custom icon for module and form

mkangia:
[trello card](https://trello.com/c/CXPyayOW/153-menu-list-icon-badges)

This is a step 1 of 2 for adding the ability to add custom icons. 
Step1 is the set up for allowing user to set just one custom icon in the front end but still have the ability to have a list of custom icons within the app.
Step2 would be to allow user to actually add multiple custom icons.


corehq/apps/app_manager/templatetags/xforms_extras.py
```diff
@@ -29,7 +29,7 @@ def trans(name, langs=None, include_lang=True, use_delim=True, prefix=False, esc
     else:
         tag = lambda lang: ""
     for lang in langs:
-        if lang in name and name[lang]:
+        if lang in name and name[lang] is not None:
```
mkangia: This one change is not related to the feature but had to do this here. Details in [commit](https://github.com/dimagi/commcare-hq/pull/16871/commits/14453f4482f6580adc9619a8ad3efb39d5cf37a2)

sravfeyn: Might want to add whatever commit note here as well, since it's easy for someone to assume this can be changed to older simpler statement.

corehq/apps/app_manager/views/forms.py
```diff
@@ -348,6 +349,31 @@ def should_edit(attribute):
...
+                return json_response(
+                    {'message': _("Please enter either text body or xpath for custom icon")},
+                    status_code=400
+                )
+            # a form should have just one custom icon for now
+            # so this just adds a new one with params or replaces the existing one with new params
+            form_custom_icon = (form.custom_icon if form.custom_icon else CustomIcon())
+            form_custom_icon.form = request.POST.get("custom_icon_form")
+            form_custom_icon.text[lang] = request.POST.get("custom_icon_text_body")
+            form_custom_icon.xpath = request.POST.get("custom_icon_xpath")
```
sravfeyn: Could replace `request.POST.get(<>)` elements with the variable names defined at the top

corehq/apps/app_manager/views/forms.py
```diff
@@ -348,6 +349,31 @@ def should_edit(attribute):
...
+        if request.POST.get("custom_icon_form"):
+            # validate that only of either text or xpath should be present
+            if (text_body and xpath) or (not text_body and not xpath):
+                return json_response(
+                    {'message': _("Please enter either text body or xpath for custom icon")},
+                    status_code=400
+                )
+            # a form should have just one custom icon for now
+            # so this just adds a new one with params or replaces the existing one with new params
+            form_custom_icon = (form.custom_icon if form.custom_icon else CustomIcon())
```
sravfeyn: nit: `form.custom_icon if form.custom_icon else CustomIcon()` this logic could go into `form.custom_icon` method itself, so you don't have to repeat this in multiple places and so that it will be easy to update this logic in one place latter if need be

corehq/apps/app_manager/views/modules.py
```diff
@@ -427,6 +430,30 @@ def should_edit(attribute):
     module = app.get_module(module_id)
     lang = request.COOKIES.get('lang', app.langs[0])
     resp = {'update': {}, 'corrections': {}}
+    if should_edit("custom_icon_form"):
+        text_body = request.POST.get("custom_icon_text_body")
```
sravfeyn: Might be worth making this into a separate function since the parent function itself is pretty large

sravfeyn:
Would be good to have some unit tests to test text/xpath/id_strings etc.

corehq/apps/app_manager/templatetags/xforms_extras.py
mkangia: Will add the comment.
ya.. it can't be changed to older simpler statement unless there is another way to address the issue i highlighted.

corehq/apps/app_manager/views/forms.py
mkangia: we need to identify that if there is actually a custom icon present or not just above this and changing that method would take away that possibility. One way to have this at one place would be to have a separate method and call that instead.

mkangia:
@sravfeyn all comments addressed. Better to review commit by commit post last review

sravfeyn (approved):
Assuming there are no actual media that is introduced in this PR (that would require adding support for new media in existing bulk media upload and other related features etc)

mkangia:
@orangejenny just confirming that we need not make changes to v1? Just generated app manager diffs for tests.

mkangia:
> Assuming there are no actual media that is introduced in this PR (that would require adding support for new media in existing bulk media upload and other related features etc)

@sravfeyn it would just add text and xpath expressions.

orangejenny:
@mkangia yes, new features don't need to go in v2. we're very close to being able to delete v1.

corehq/apps/app_manager/templates/app_manager/v2/partials/custom_icon.html
```diff
@@ -0,0 +1,31 @@
+{% load xforms_extras %}
+{% load i18n %}
+
+<div>
+    <div class="form-group container-fluid">
+        <label class="control-label col-sm-2">{% trans "Custom Icon" %}</label>
+        <div class="col-sm-10 commcare-feature" data-since-version="1.3">
```
orangejenny: Surprised by this version...Is this really supposed to be 1.3?

corehq/apps/app_manager/templates/app_manager/v2/partials/custom_icon.html
```diff
@@ -0,0 +1,31 @@
...
+            <label>
+                <span>
+                    {% trans "Form" %}
+                </span>
+                <input class="form-control" name="custom_icon_form" id="custom_icon_form"
```
orangejenny: What would you think of making this a hidden input for now, until more forms are supported?

corehq/apps/app_manager/templates/app_manager/v2/partials/custom_icon.html
```diff
@@ -0,0 +1,31 @@
...
+                    {% trans "Text Body" %}
+                </span>
+                <input class="form-control" name="custom_icon_text_body" id="custom_icon_text_body"
+                       value="{{ custom_icon.text|trans:langs }}">
+            </label>
+            <label>
+                <span>
+                    {% trans "Xpath Function" %}
+                </span>
+                <input class="form-control" name="custom_icon_xpath" id="custom_icon_xpath"
```
orangejenny: Can/should this use the xpath widget that's used in form/module display conditions?

corehq/feature_previews.py
```diff
@@ -130,3 +130,12 @@ def enable_callcenter(domain_name, checked):
     ),
     privilege=LOOKUP_TABLES,
 )
+
+
+CUSTOM_ICON_BADGES = FeaturePreview(
```
orangejenny: Feature previews are being deprecated; this should be an add-on instead: https://github.com/dimagi/commcare-hq/blob/master/corehq/apps/app_manager/add_ons.py

orangejenny (changes requested):
Should do this as an add-on instead of a feature preview, so it doesn't need to be migrated later. My other comments are minor and non-blocking.

corehq/apps/app_manager/templates/app_manager/v2/partials/custom_icon.html
mkangia: removing this.. just kept it similar to menu media.

corehq/apps/app_manager/templates/app_manager/v2/partials/custom_icon.html
mkangia: we are not restricting this to avoid change requests on HQ when mobile starts adding more support.

orangejenny: I'd like to push back on that - this is something that'll be visible to all users, and having a text box where you need to type in a specific word, but you don't even know what the options are, is pretty low usability. Should this be behind a feature flag instead?

mkangia: Hey this is behind the add on now.

corehq/apps/app_manager/templates/app_manager/v2/partials/custom_icon.html
mkangia: Will try to incorporate this in the follow up PR where i add the ability to add multiple custom icons

orangejenny (approved):

PR #16877, opened 2017-07-03
App Summary: Hide case type if no properties match filter

NoahCarnahan:
In the app summary > case summary, when you filter by case property, the meta information for all case types is shown, even for case types that don't have any matching case properties. When this information is long and when you have many case types, this prevents one from quickly seeing matches for the searched case property as one must scroll through a considerable amount of information to see if the case property is present or not.

This PR hides that case meta information if no case properties match.

@dannyroberts 
best reviewed commit-by-commit (and with `?w=1`)


dannyroberts (approved):
very nice

PR #16878, opened 2017-07-03
Webapps restriction by role

proteusvacuum:
@snopoke 
I have no idea what I was thinking on Friday... Long week I guess. 
This makes much more sense. Still behind a toggle because I think we should actually be able to deprecate the groups restrictions with this.

🐡 / 🐡 from https://github.com/dimagi/commcare-hq/commit/a06fcfe14f4eb5d5dd9fe489ac9748fc13147ff5


esoergel (approved):
Nice, this is for sure a better system

PR #16879, opened 2017-07-03
Can't create new case export in domains that have a remote app

orangejenny:
@kaapstorm / @gcapalbo


gcapalbo (approved):
build failure looks like it might be related

orangejenny:
odd, that fails for me locally regardless of the code change. will kick & see if it helps.

orangejenny:
looks like build failure is transient, merging

PR #16883, opened 2017-07-03
Add "edited_on" field to Forms API

esoergel:
https://manage.dimagi.com/default.asp?257055
@krsdimagi asked about making this available to commcare-export.  Looks like that uses the API.  @snopoke, would you expect that this is sufficent for that field to be accessible via the commcare-export tool?

@emord


krsdimagi:
@esoergel Thanks for this!  Would it also be possible to make available the username of the user who did the editing for the particular form?

snopoke:
I think the form's `auth_context` would contain the `uesr_id` of the editing user. Not sure how easy that is to add to the API though.

PR #16887, opened 2017-07-03
Add management command to tag older icds sms with indicator metadata

gcapalbo:
We started collecting this data last monday. This fills in the metadata for the SMS entries that don't have it.

@emord 
@kaapstorm @orangejenny


kaapstorm (approved):
I can't check your Hindi :grin: but the rest looks good to me.

PR #16891, opened 2017-07-04
Revert 16888 revert externalize js report base template

orangejenny:
Reverts https://github.com/dimagi/commcare-hq/pull/16888

The error in ews-ghana turned out to be pre-existing, not caused by that PR. The error in ils-gateway was because `#savedReports` had knockout bindings applied twice (once in the [reports base](https://github.com/dimagi/commcare-hq/blob/aeead2d8c98169fd656ad9047a990eb33f6524df/corehq/apps/reports/templates/reports/base_template.html#L26) and once in the [ilsgateway base](https://github.com/dimagi/commcare-hq/blob/aeead2d8c98169fd656ad9047a990eb33f6524df/custom/ilsgateway/templates/ilsgateway/base_template.html#L107).) The second commit fixes that.

@millerdev / @dannyroberts 
code buddies @kaapstorm / @gcapalbo


orangejenny:
"don't merge" just so I'm around when this is merged/deployed and can test after deploy

kaapstorm (approved):

orangejenny:
deployed, tested ewsghana and ilsgateway

PR #16892, opened 2017-07-04
Ce/app dim

calellowitz:
@snopoke @gcapalbo @benrudolph 
Only the last three commits are relevant (ben recommended working off the branch in https://github.com/dimagi/commcare-hq/pull/16827 ) so this should not be merged until after that branch


corehq/couchapps/last_modified/views/by_last_modified/map.js
```diff
@@ -5,9 +5,9 @@ function(doc){
         emit([doc.doc_type, doc.last_modified], null);
     } else if (doc.base_doc && (doc.base_doc === "CouchUser" || doc.base_doc === "CouchUser-Deleted")) {
         emit([doc.base_doc, doc.last_modified], null);
-    } else if (doc.doc_type === "Application" || doc.doc_type === "Application-Deleted" ||
-               doc.base_doc === "LinkedApplication" || doc.base_doc === "LinkedApplication-Deleted" ||
-               doc.base_doc === "RemoteApp" || doc.base_doc === "RemoteApp-Deleted") {
-        emit([doc.base_doc, doc.last_modified], null);
+    } else if (doc.doc_type === "Application" || doc.base_doc === "LinkedApplication" || doc.base_doc === "RemoteApp") {
+        emit(['ApplicationBase', doc.last_modified], null);
```
calellowitz: i was talked out of adding this property as a `base_doc` to the `ApplicationBase` model similar to how we do with `CouchUser` (as you can see in earlier commits) because of the effort in initial migration and the risk if future subclasses will not want that property. (However, I am still open to the idea since i think it would simplify this view and many others that check for all 3 subclasses of application base)

corehq/warehouse/transforms/sql/application_dim.sql
```diff
@@ -0,0 +1,23 @@
...
+    application_id,
+    domain,
+    name,
+    application_last_modified,
+    CASE base_doc
+        WHEN 'ApplicationBase' then false
+        WHEN 'ApplicationBase-Deleted' then true
+    END,
+    now(),
+    now(),
```
calellowitz: I followed the pattern of the other sql of making both `dim_last_modified` and `dim_created_on` equal to `now()` but i am not sure why `created_on` would be

gcapalbo: i think Ben intended that to be the timestamp that the dimension record was created, so they're both the same on initial insert

gcapalbo (approved):
looks good. would be good to add a test for the staging table and dim as Ben has done here:

https://github.com/dimagi/commcare-hq/blob/ce/app-dim/corehq/warehouse/tests/test_staging.py

https://github.com/dimagi/commcare-hq/blob/ce/app-dim/corehq/warehouse/tests/test_dimensions.py

gcapalbo (approved):

snopoke (approved):

snopoke:
build fialing: 
```
FAIL: corehq.warehouse.tests.test_dbaccessors:TestDbAccessors.test_get_app_ids_by_last_modified
  File "/mnt/commcare-hq-ro/corehq/warehouse/tests/test_dbaccessors.py", line 136, in test_get_app_ids_by_last_modified
    {self.test_app._id, self.deleted_app._id, self.linked_app._id},
AssertionError: Items in the first set but not the second:
'14cff2ea244c4a8f9ddd9690e4062ef6'
'4df9d63a4a23485cb7273a0c919a8172'
'394afd2eda9f461fab422139db572536'
'9c6bc86081b2499b876263a2f40c863e'
```

PR #16894, opened 2017-07-04
rescue exception when raw docing missing location

mkangia:
https://manage.dimagi.com/default.asp?257525

Not sure if this is the best possible fix.


snopoke:
more generic fix here: https://github.com/dimagi/commcare-hq/pull/16897

PR #16895, opened 2017-07-04
Change submissions responses for known exceptions and signal errors

sravfeyn:
https://docs.google.com/document/d/1kSZNPNunybhrErF30JZW1q7pJ3CUlmlGm6i2DfkxmQg/edit

@nickpell @snopoke 

Open for review, hold on till some spec comments are resolved.


corehq/blobs/mixin.py
```diff
@@ -51,10 +51,14 @@ class Meta:
...
     def _blobdb_bucket(self):
-        if self._id is None:
+        if self.blob_identifier is None:
             raise ResourceNotFound(
                 "cannot manipulate attachment on unidentified document")
-        return join(_get_couchdb_name(type(self)), safe_id(self._id))
+        return join(_get_couchdb_name(type(self)), safe_id(self.blob_identifier))
+
+    @property
+    def blob_identifier(self):
```
sravfeyn: @snopoke this is overriding of default `self._id` we have talked about.

corehq/blobs/mixin.py
```diff
@@ -279,6 +283,10 @@ def get_external_blobs(doc, default=None):
     def blobs(self):
         return BlobMixin.blobs.fget(self)
 
+    @property
+    def blob_identifier(self):
+        return self._id
```
sravfeyn: BlobHelper seems to be never used for xform-errors, so I didn't have to override it, does this sound correct? @snopoke

corehq/form_processor/submission_post.py
```diff
@@ -160,6 +164,7 @@ def run(self):
             else:
                 case_db_cache = self.interface.casedb_cache(domain=domain, lock=True, deleted_ok=True, xforms=xforms)
 
+            known_submission_error = False
```
snopoke: seems like this should be `None` and not false since later you're setting it to a String

corehq/form_processor/submission_post.py
```diff
@@ -290,21 +308,18 @@ def process_signals(self, instance):
                 ) % (func, instance.form_id, type(resp).__name__, error_message))
                 errors.append(error_message)
         if errors:
-            self.interface.xformerror_from_xform_instance(instance, ", ".join(errors))
+            self.interface.xformerror_from_xform_instance(instance, ", ".join(errors), with_new_id=True)
             self.formdb.update_form_problem_and_state(instance)
+            publish_form_deleted(self.domain, instance.orig_id)
+            publish_form_saved(instance)
```
snopoke: these should be moved to the `FormAccessorSQL` method. We don't need to publish changes for Couch domains since the changes get published from the couch change feed.

corehq/form_processor/submission_post.py
```diff
@@ -290,21 +308,18 @@ def process_signals(self, instance):
                 ) % (func, instance.form_id, type(resp).__name__, error_message))
                 errors.append(error_message)
         if errors:
-            self.interface.xformerror_from_xform_instance(instance, ", ".join(errors))
+            self.interface.xformerror_from_xform_instance(instance, ", ".join(errors), with_new_id=True)
             self.formdb.update_form_problem_and_state(instance)
```
snopoke: This function will have to change since it doesn't support changing the form ID. For couch this isn't too much of an issue since we just save the form but for SQL we'll have to delete the old form and save the new form since it might move to a new shard.

corehq/form_processor/submission_post.py
```diff
@@ -201,7 +208,16 @@ def run(self):
...
             errors = self.process_signals(instance)
-            response = self._get_open_rosa_response(instance, errors)
+            if instance.is_normal and not errors:
+                response = self.get_success_response()
+            elif not self.is_openrosa_version3():
+                response = self.get_failure_response(instance.problem)
+            elif known_submission_error:
+                response = self.get_retry_response(known_submission_error, ResponseNature.KNOWN_PROCESSING_ERROR)
+            else:
+                response = self.get_retry_response(instance.problem, ResponseNature.SUBMIT_ERROR)
```
snopoke: any reason not to just put this code into `_get_open_rosa_response`?

corehq/form_processor/submission_post.py
```diff
@@ -366,7 +392,7 @@ def _transform_instance_to_error(interface, e, instance):
         u"Warning in case or stock processing "
         u"for form {}: {}."
     ).format(instance.form_id, error_message))
```
snopoke: The form ID in this message should be the new ID

snopoke (commented):
Let me know if you want to go to voice about the saving issues for the SQL backend.

corehq/apps/receiverwrapper/tests/test_submit_errors.py
```diff
@@ -52,6 +56,9 @@ def testSubmitDuplicate(self):
 
         file, res = self._submit('simple_form.xml')
         self.assertEqual(201, res.status_code)
+
+        _, res_openrosa3 = self._submit('simple_form.xml', open_rosa_header=OPENROSA_VERSION_3)
+        self.assertEqual(422, res_openrosa3.status_code)
```
sravfeyn: I have first written separate test, but there is bunch of unavoidable duplication, so I added it here.

corehq/blobs/mixin.py
sravfeyn: Done

corehq/blobs/mixin.py
```diff
@@ -138,7 +142,7 @@ def fetch_attachment(self, name, stream=False):
                 u"{model} {model_id} attachment: {name!r}".format(
                     model=type(self).__name__,
                     name=name,
-                    model_id=self._id,
+                    model_id=self.blob_identifier,
```
sravfeyn: Ah, thanks. Updated

corehq/blobs/mixin.py
sravfeyn: Updated, thanks!

corehq/blobs/mixin.py
```diff
@@ -399,7 +407,7 @@ def fetch_attachment(self, name, stream=False):
                 raise ResourceNotFound(
                     u"{model} {model_id} attachment: {name!r}".format(
                         model=type(self).__name__,
-                        model_id=self._id,
+                        model_id=self.blob_identifier,
```
sravfeyn: 👍 updated

corehq/form_processor/submission_post.py
sravfeyn: 👍

corehq/form_processor/submission_post.py
sravfeyn: Updated

corehq/form_processor/submission_post.py
sravfeyn: Ah, for SQL these are already published in `xformerror_from_xform_instance` if `with_new_id` is set to True, so I can just remove this then

sravfeyn: I had to add it, it's not there

corehq/blobs/mixin.py
```diff
@@ -255,9 +259,19 @@ class BlobHelper(object):
...
+            #   will still be orig_id
+            bucket_id = doc.get('orig_id') or doc['_id']
+        else:
+            bucket_id = doc['_id']
+
+        if bucket_id is None:
             raise TypeError("BlobHelper requires a real _id")
-        self._id = doc["_id"]
+
+        self._id = bucket_id
```
millerdev: Should also add
```py
self.blobdb_bucket_id = bucket_id
```
here because lines like 311 call `BlobMixin` methods passing an instance of `BlobHelper` as first argument:
```py
BlobMixin.put_attachment(self, content, name, *args, **kw)
```
This is possible because those methods use the [`document_method` decorator](https://github.com/dimagi/commcare-hq/blob/11edc1be91610c7e79ab0b297ee4bdcd2e14e054/corehq/blobs/util.py#L32).

sravfeyn: Thanks! Updated

corehq/sql_db/util.py
```diff
@@ -121,6 +123,20 @@ def get_db_alias_for_partitioned_doc(partition_value):
...
+    Returns a new partition value that belongs to the same db alias as
+        the given partition value does
+    """
+    old_db_name = get_db_alias_for_partitioned_doc(partition_value)
+    new_db_name = None
+    while old_db_name != new_db_name:
+        # todo; guard against infinite recursion
+        new_partition_value = unicode(uuid.uuid4())
+        new_db_name = get_db_alias_for_partitioned_doc(new_partition_value)
+    return new_partition_value
```
snopoke: I did some test on NIC and got a max of 129 iterations and max timing of 430 ms (did 500k tries to get data). The average is 10 iterations so this approach seems fine.

sravfeyn: Nice, thanks!

sravfeyn:
Removed the hold-off label. I want to do some refactoring (but it's not related to feature), so I will do it separately.  
@snopoke 
buddy @dannyroberts

corehq/form_processor/models.py
```diff
@@ -320,7 +320,10 @@ def to_json(self, include_attachments=False):
 
     def _get_attachment_from_db(self, attachment_name):
         from corehq.form_processor.backends.sql.dbaccessors import FormAccessorSQL
-        return FormAccessorSQL.get_attachment_by_name(self.form_id, attachment_name)
+        try:
```
sravfeyn: @snopoke does this look good?

corehq/form_processor/backends/sql/dbaccessors.py
```diff
@@ -587,12 +587,18 @@ def save_deprecated_form(form):
...
     @staticmethod
     @transaction.atomic
-    def update_form_problem_and_state(form):
-        with get_cursor(XFormInstanceSQL) as cursor:
-            cursor.execute(
-                'SELECT update_form_problem_and_state(%s, %s, %s)',
-                [form.form_id, form.problem, form.state]
-            )
+    def update_form(form):
+        from corehq.form_processor.change_publishers import publish_form_saved
```
sravfeyn: Please validate this implementation

corehq/apps/receiverwrapper/tests/test_submit_errors.py
snopoke: Not sure about this one. @amstone326 do we want to return a retry response for duplicate forms? We've seen this happen a lot on ICDS where a form is submitted but the client times out or is otherwise disconnected during the request. The client then re-submits the form. Since HQ has completed processing the original form this new form is an exact duplicate.

I don't think we should respond with 422 for this case.

corehq/form_processor/models.py
snopoke: you could do this workflow only if the `form_id` has changed (see https://stackoverflow.com/a/1793323/632517 for a good way to determine that)

corehq/apps/receiverwrapper/tests/test_submit_errors.py
amstone326: @snopoke Yea sounds to me like we definitely don't want to send a 422 in that case, since there is no reason that we would want mobile to try re-submitting, and really no meaningful error for us to worry about from a debugging standpoint. I think it makes sense to continue returning a success response in that case, since for all intents and purposes the form _did_ submit successfully

amstone326:
@sravfeyn What spec comments is this waiting on? I just looked through the doc and didn't see anything pending.

sravfeyn:
@amstone326 You have probably meant to ask that on the [forced LTS update PR](https://github.com/dimagi/commcare-hq/pull/17276#issue-246714412) for which [spec comments](https://docs.google.com/document/d/1RhdJPD6ZvRGcUsSgULMt05yXUQG6j_FLCAi1fMMEtyQ/edit?ts=597f56bd#heading=h.q3wyha58n4s0) are pending, not on this one?

corehq/form_processor/backends/sql/dbaccessors.py
```diff
@@ -587,12 +587,18 @@ def save_deprecated_form(form):
...
+        from corehq.sql_db.util import get_db_alias_for_partitioned_doc
+        db_name = get_db_alias_for_partitioned_doc(form.form_id)
+        if not form.is_saved():
+            FormAccessorSQL.save_new_form(form)
```
sravfeyn: So that means the method shouldn't supposed unsaved forms, and I should raise an assertion if otherwise, yeah?

snopoke:
yes

amstone326:
@sravfeyn No I meant to ask it on this one, since you have written in the PR comment above `"Open for review, hold on till some spec comments are resolved."`.

sravfeyn:
I have since then removed the label, currently I am addressing some of PR feedback and it should be good to go after that.

corehq/form_processor/backends/sql/dbaccessors.py
```diff
@@ -587,12 +587,30 @@ def save_deprecated_form(form):
...
+        from corehq.form_processor.change_publishers import publish_form_saved
+        from corehq.sql_db.util import get_db_alias_for_partitioned_doc
+        assert form.is_saved(), "this method doesn't support creating unsaved forms"
+
+        db_name = get_db_alias_for_partitioned_doc(form.form_id)
+        if form.orig_id:
+            old_db_name = get_db_alias_for_partitioned_doc(form.orig_id)
+            assert old_db_name == db_name, "this method doesn't support moving the form to new db"
+
+        form_id_updated = form.orig_id != form.id or form.orig_id is None
```
sravfeyn: Does this heuristic sound good? If not, I will copy the attribute value on object and check if it's updated

corehq/form_processor/backends/sql/dbaccessors.py
```diff
@@ -587,12 +587,30 @@ def save_deprecated_form(form):
...
+        from corehq.form_processor.change_publishers import publish_form_saved
+        from corehq.sql_db.util import get_db_alias_for_partitioned_doc
+        assert form.is_saved(), "this method doesn't support creating unsaved forms"
+
+        db_name = get_db_alias_for_partitioned_doc(form.form_id)
+        if form.orig_id:
+            old_db_name = get_db_alias_for_partitioned_doc(form.orig_id)
+            assert old_db_name == db_name, "this method doesn't support moving the form to new db"
+
+        form_id_updated = form.orig_id != form.form_id and form.orig_id is not None
```
sravfeyn: Does this heuristic sound good? If not, I will copy the attribute value on object and check if it's updated

snopoke: that's not a good general test since it is only true when `orig_id` get's set. If the form get's updated after that (for whatever reason) then it won't hold true.

corehq/form_processor/backends/sql/dbaccessors.py
```diff
@@ -587,12 +587,32 @@ def save_deprecated_form(form):
 
     @staticmethod
     @transaction.atomic
```
snopoke: can drop this decorator

corehq/form_processor/backends/sql/dbaccessors.py
```diff
@@ -587,12 +587,32 @@ def save_deprecated_form(form):
...
+
+        db_name = get_db_alias_for_partitioned_doc(form.form_id)
+        if form.orig_id:
+            old_db_name = get_db_alias_for_partitioned_doc(form.orig_id)
+            assert old_db_name == db_name, "this method doesn't support moving the form to new db"
+
+        form_id_updated = form.orig_id != form.form_id and form.orig_id is not None
+
+        if form_id_updated:
+            attachments = form.get_attachments()
```
snopoke: Sorry my bad, this won't work since it's trying to get the attachments using `form.form_id` which has now been changed. You can do a query directly:
```
XFormAttachmentSQL.objects.using(db_name).filter(form_id=form.orig_id)
```
The same for operations.

sravfeyn:
@snopoke Updated

snopoke (commented):
I've just realized that there is still one instance where we don't want to change the form ID. In the case where there are error post save (in `process_signals`) we can't change the form ID unless we also change the referencing case / ledger transactions. 

I think in that case we don't want to change the ID.

sravfeyn:
Putting `hold-off` label as Aliza wants to run integration tests off of staging first.

corehq/form_processor/backends/couch/processor.py
```diff
@@ -102,8 +102,8 @@ def assign_new_id(cls, xform):
         return xform
 
     @classmethod
-    def xformerror_from_xform_instance(self, instance, error_message, with_new_id=False):
-        return XFormError.from_xform_instance(instance, error_message, with_new_id=with_new_id)
+    def xformerror_from_xform_instance(self, instance, error_message, replace_form_id=True):
+        return XFormError.from_xform_instance(instance, error_message, replace_form_id=replace_form_id)
```
snopoke: could be slightly confusing that this defaults to `True` but the arg in `from_xform_instance` defaults to `False`. I'd recommend making them both default to the same thing.

corehq/form_processor/backends/sql/dbaccessors.py
```diff
@@ -586,13 +586,30 @@ def save_deprecated_form(form):
...
+            with transaction.atomic(db_name):
+                form.save(using=db_name)
+                for a in attachments:
+                    a.form = form
+                    a.save(using=db_name)
+                for op in operations:
+                    op.form = form
+                    op.save(using=db_name)
+        else:
+            form.save(using=db_name)
```
snopoke: if you merge master you can remove the `using=db_name` from the save call since that is now automatic

corehq/form_processor/backends/sql/dbaccessors.py
```diff
@@ -586,13 +586,30 @@ def save_deprecated_form(form):
...
-            cursor.execute(
-                'SELECT update_form_problem_and_state(%s, %s, %s)',
-                [form.form_id, form.problem, form.state]
-            )
+    def update_form(form):
+        from corehq.form_processor.change_publishers import publish_form_saved
+        from corehq.sql_db.util import get_db_alias_for_partitioned_doc
+        assert form.is_saved(), "this method doesn't support creating unsaved forms"
+
+        db_name = get_db_alias_for_partitioned_doc(form.form_id)
```
snopoke: if you merge master this can just become `db_nmae = form.db`

corehq/form_processor/models.py
```diff
@@ -204,6 +204,28 @@ class XFormInstanceSQL(PartitionedModel, models.Model, RedisLockableMixIn, Attac
...
+
+    @property
+    @memoized
+    def original_attachments(self):
+        from corehq.form_processor.backends.sql.dbaccessors import FormAccessorSQL
+        return FormAccessorSQL.get_attachments(self.__original_form_id)
+
+    @property
+    @memoized
+    def original_operations(self):
```
snopoke: docstrings for these would be useful

snopoke (commented):
Nice, this is looking good. A few minor comments

corehq/form_processor/backends/couch/processor.py
sravfeyn: Yea, updated

sravfeyn:
@snopoke The build has passed. Let me know if you have any more feedback otherwise I will put this on staging

corehq/form_processor/backends/sql/processor.py
```diff
@@ -156,23 +156,26 @@ def deduplicate_xform(cls, xform):
...
         instance.problem = error_message
 
-        if with_new_id:
+        if replace_form_id:
             orig_id = instance.form_id
             cls.assign_new_id(instance)
             instance.orig_id = orig_id
             if instance.is_saved():
                 # clear the ID since we want to make a new doc
                 instance.id = None
```
snopoke: I thought we agreed that this shouldn't happen anymore since we're just updating the form.

dannyroberts:
Haven't been very involved in this PR but looked almost complete a week ago at the time of the last comment, so just pinging in case it fell through the cracks.

amstone326:
Thanks @dannyroberts, I think this was waiting on an issue that needed input from @snopoke. Is that still the case?

snopoke:
Yes, sorry, there is an issue I need to look at. Should be able to look
towards the end of the week.

snopoke:
Looked into this briefly - going to need a bit more time to think it through.

amstone326:
@snopoke Do you have an estimated timeline on this? If possible it would be great to keep it moving along so the code doesn't get too stale, although understood if there are higher priorities.

snopoke:
sorry I've been unable to look at it till today. I spent most of this
morning working on it and the more I look into it the more issues I find
with our current error handling. I'm making notes in the spec doc and will
ping once ready to discuss.

amstone326:
@snopoke Ok that sounds good, thanks for the extra digging.

@sravfeyn In the mean time it looks like there are also some conflicts now.

calellowitz:
@snopoke is there any movement on this pr, the same issue may be appearing again

snopoke:
This has been replaced by this: https://github.com/dimagi/commcare-hq/pull/17867

@calellowitz what issue is appearing again?

PR #16899, opened 2017-07-04
Catch SMTPException

kaapstorm:
@proteusvacuum this will catch the exception that the function's docstring refers to ... 

cc @gcapalbo @orangejenny 

... but I'm labelling "hold off" because if we do encounter an SMTPException it's most likely because HQ is not configured properly, and not something the user would be able to do anything about. A Sad Danny might be more appropriate, which I'm sure is what would have been happening up to now.

If we do want to give the user more information about the error, I'm not convinced that we want to show them a traceback;
```
exc_info() -> (type, value, traceback)

Return information about the most recent exception caught by an except
clause in the current stack frame or in an older stack frame.
```

Instead of fixing this `except` clause, I'm inclined to just remove it.

### DIFF
--- a/corehq/apps/domain/urls.py
+++ b/corehq/apps/domain/urls.py
@@ -4,6 +4,7 @@ from django.contrib.auth.views import (
     password_change_done,
     password_reset_complete,
     password_reset_done,
+    password_reset,
 )
 from django.utils.translation import ugettext as _
 from django.conf import settings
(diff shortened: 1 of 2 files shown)


gcapalbo:
yeah i think removing the try/except so that we raise a 500 when we can't send the email seems like the right thing to do (and it has been behaving that way anyway since `except None` should never match any raised exceptions). i guess that also means we can do away with `exception_safe_password_reset`

gcapalbo (approved):

PR #16902, opened 2017-07-04
Add Support for inline UI Notifications

biyeun:
@orangejenny 

for the App Manager v2 release. Test notification is currently under a toggle.

screen shot 2017-07-04 at 12 55 37 pm


corehq/apps/hqwebapp/templatetags/hq_shared_tags.py
```diff
@@ -217,6 +217,29 @@ def toggle_enabled(request, toggle_or_toggle_name):
...
+    import corehq.apps.notifications.ui_notify
+    return _ui_notify_enabled(
+        corehq.apps.notifications.ui_notify,
+        request,
+        ui_notify_instance_or_name
+    )
+
+
+@register.filter
+def ui_notify_slug(ui_notify_instance_or_name):
```
orangejenny: Not following what this is for - why not change releases.html to use either the slug or the name, whichever is easier? (`TEST_UI_NOTIFY` or `first_ui_notify`)  Having both a name and a slug is something I find strange about the existing toggles framework and avoided when doing add_ons.

orangejenny:
Code looks fine. Some general thoughts:

Should we kill the existing guided tour code? This seems to be replacing it.

What would you think of having some notion of expiration, to force the person who's adding a notification to think about how long it should last?

Should there be a flag for whether new users/apps should see these notifications? I'm thinking of the releases page changes, where it doesn't probably doesn't make sense for new users to see the notification, just current users.

Should probably kill the `SHOW_UI_NOTIFICATIONS` flag before merging this.

corehq/apps/hqwebapp/templatetags/hq_shared_tags.py
biyeun: the name is the name of the variable in ui_toggles, but the slug is what's actually saved in the DB. I think the two should be separated. maybe we want to keep a shorthand name, but give a really long slug with more info `app_builder_publish_08_2017_v001`

biyeun:
@orangejenny 

> Should we kill the existing guided tour code? This seems to be replacing it.
this doesn't really replace the guided tour. The guided tour has previous / next functionality, can move you to a new page, etc. This is just standalone, but can be dismissed per user.

> What would you think of having some notion of expiration, to force the person who's adding a notification to think about how long it should last?
This is a great idea. am adding that in.

> Should there be a flag for whether new users/apps should see these notifications? I'm thinking of the releases page changes, where it doesn't probably doesn't make sense for new users to see the notification, just current users.
OK i can add a `visible_for_users_created_before` flag.

corehq/apps/hqwebapp/templatetags/hq_shared_tags.py
orangejenny: I like making a convention of having a date in the slug, would reduce the likelihood of inadvertently reusing slugs.

biyeun:
@orangejenny responded to PR

biyeun:
@orangejenny is mostly offline, but maybe @dannyroberts or @proteusvacuum can take a look?

orangejenny (approved):

PR #16903, opened 2017-07-04
Externalize js: report filters: datespan

orangejenny:
My general plan for report filters is to move data into `data-*` attributes on the widgets and then initialize them once filters are loaded. Dealing with code that's shared with custom reports looks like it's going to be hairy.

Review commit-by-commit.

@millerdev / @kaapstorm / @gcapalbo

### DIFF
--- a/corehq/apps/hqwebapp/static/hqwebapp/js/initial_page_data.js
+++ b/corehq/apps/hqwebapp/static/hqwebapp/js/initial_page_data.js
@@ -28,15 +28,7 @@ hqDefine('hqwebapp/js/initial_page_data.js', function () {
                 if (COMMCAREHQ_INITIAL_PAGE_DATA[data.name] !== undefined) {
                     throw new Error("Duplicate key in initial page data: " + data.name);
                 }
-                var value = data.value;
-
-                // The initial_page_data tag escapes whitespace with \n, etc. for the sake of
-                // printing the value in a data attribute. Unescape here.
-                if (typeof value === "string") {
-                    value = JSON.parse('"' + value + '"');
-                }
-
-                COMMCAREHQ_INITIAL_PAGE_DATA[data.name] = value;
+                COMMCAREHQ_INITIAL_PAGE_DATA[data.name] = data.value;
                 $div.remove();
             });
         });
--- a/corehq/apps/hqwebapp/templatetags/hq_shared_tags.py
+++ b/corehq/apps/hqwebapp/templatetags/hq_shared_tags.py
@@ -631,7 +631,7 @@ def html_attr(value):
         value = json.dumps(value)[1:-1]
     else:
         value = JSON(value)
-    return "{}".format(escape(value))
+    return escape(value)
 
 
 @register.tag
--- a/corehq/apps/hqwebapp/templatetags/hq_shared_tags.py
+++ b/corehq/apps/hqwebapp/templatetags/hq_shared_tags.py
@@ -625,6 +625,15 @@ def registerurl(parser, token):
     return AddToBlockNode(nodelist, 'js-inline')
 
 
+@register.simple_tag
+def html_attr(value):
+    if isinstance(value, basestring):
+        value = json.dumps(value)[1:-1]
+    else:
+        value = JSON(value)
+    return "{}".format(escape(value))
+
+
 @register.tag
 def initial_page_data(parser, token):
     split_contents = token.split_contents()
--- a/corehq/apps/hqwebapp/templatetags/hq_shared_tags.py
+++ b/corehq/apps/hqwebapp/templatetags/hq_shared_tags.py
@@ -627,9 +627,7 @@ def registerurl(parser, token):
 
 @register.simple_tag
 def html_attr(value):
-    if isinstance(value, basestring):
-        value = json.dumps(value)[1:-1]
-    else:
+    if not isinstance(value, basestring):
         value = JSON(value)
     return escape(value)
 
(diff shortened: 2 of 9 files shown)


corehq/apps/reports/static/reports/js/reports.async.js
```diff
@@ -30,11 +30,34 @@ var HQAsyncReport = function (o) {
...
+        }
+        var separator = $filterRange.data('separator');
+        var report_labels = $filterRange.data('reportLabels');
+        var standardHQReport = hqImport("reports/js/standard_hq_report.js").getStandardHQReport();
+
+        $filterRange.createDateRangePicker(
+            report_labels, separator,
+            $filterRange.data('startDate'),
+            $filterRange.data('endDate'),
+        );
```
kaapstorm: Yeah, that final comma might be a problem.

kaapstorm:
I didn't know about [HTML5's Dataset API][1] and that jQuery 3 [aligns with it][2]. That's very cool.

  [1]: https://www.w3.org/TR/html5/dom.html#dom-dataset
  [2]: https://api.jquery.com/data/

kaapstorm (approved):

gcapalbo (approved):

corehq/apps/hqwebapp/templatetags/hq_shared_tags.py
```diff
@@ -625,6 +625,15 @@ def render(self, context):
     return AddToBlockNode(nodelist, 'js-inline')
 
 
+@register.simple_tag
+def html_attr(value):
+    if isinstance(value, basestring):
+        value = json.dumps(value)[1:-1]
```
millerdev: How is this consumed in JS? Are the quotes re-added and HTML entities un-escaped before JSON parsing? I'm thinking of a scenario like this:
```python
>>> import json
>>> from django.utils.html import escape
>>> escape(json.dumps('abc"def')[1:-1])
u'abc\\&quot;def'
```
Not sure what string value that will produce in JS. Will the bashslash be removed? Maybe it would be better to `escape()` the string before `json.dumps()`?

Relevant: https://stackoverflow.com/a/4015380/10840

corehq/apps/reports/static/reports/js/reports.async.js
orangejenny: oops, fixed

corehq/apps/hqwebapp/templatetags/hq_shared_tags.py
orangejenny: You're right, thanks for flagging - the final value will still have an undesired backslash (`abc\"def`). Initial page data uses the same logic but takes care of this issue by doing a [JSON.parse on get](https://github.com/dimagi/commcare-hq/blob/master/corehq/apps/hqwebapp/static/hqwebapp/js/initial_page_data.js#L36). It's going to be confusing if data attributes set with `html_attr` need to be parsed when accessed, but data attributes set normally don't need to be parsed. Will think about this some more.

Escaping before json encoding unfortunately doesn't work, because the original value may not be a string.

orangejenny: Okay, I squinted at this for a bit and concluded that there's no reason to call `json.dumps` when the given value is just a string, which simplifies the code and takes care of the backslash issue. Added a commit.

corehq/apps/hqwebapp/static/hqwebapp/js/initial_page_data.js
```diff
@@ -28,15 +28,7 @@ hqDefine('hqwebapp/js/initial_page_data.js', function () {
                 if (COMMCAREHQ_INITIAL_PAGE_DATA[data.name] !== undefined) {
                     throw new Error("Duplicate key in initial page data: " + data.name);
                 }
-                var value = data.value;
-
-                // The initial_page_data tag escapes whitespace with \n, etc. for the sake of
-                // printing the value in a data attribute. Unescape here.
-                if (typeof value === "string") {
-                    value = JSON.parse('"' + value + '"');
```
millerdev: I think this is still needed to unescape `\n`. Maybe I'm missing something?

corehq/apps/hqwebapp/templatetags/hq_shared_tags.py
```diff
@@ -625,6 +625,13 @@ def render(self, context):
     return AddToBlockNode(nodelist, 'js-inline')
 
 
+@register.simple_tag
+def html_attr(value):
+    if not isinstance(value, basestring):
+        value = JSON(value)
+    return escape(value)
```
millerdev: This seems better, but how does one know when to JSON-parse the value in the attribute and when to just use the raw attribute value? Does the JS code always know the type of things that are stuffed into attributes (at least whenever the thing is a string)? Otherwise, it seems like the JS might try to `JSON.stringify()` an attribute value and get an error because the value that was put in the attribute was a string (and therefore was not JSON-encoded).

Ugh! this is confusing. Does any of that make sense?

corehq/apps/hqwebapp/static/hqwebapp/js/initial_page_data.js
orangejenny: I was wrong when I wrote this, so the comment here was wrong; it's okay to have multi-line values in html attributes.

corehq/apps/hqwebapp/templatetags/hq_shared_tags.py
orangejenny: jQuery's `.data()` figures it out:

> When the data attribute is an object (starts with '{') or array (starts with '[') then jQuery.parseJSON is used to parse the string; it must follow valid JSON syntax including quoted property names. If the value isn't parseable as a JavaScript value, it is left as a string.

In rather un-javascript-like fashion, it also decides whether the value is a string versus a number:
> Every attempt is made to convert the string to a JavaScript value (this includes booleans, numbers, objects, arrays, and null). A value is only converted to a number if doing so doesn't change the value's representation. For example, "1E02" and "100.000" are equivalent as numbers (numeric value 100) but converting them would alter their representation so they are left as strings. The string value "100" is converted to the number 100.

(from https://api.jquery.com/data/)

millerdev: huh, I wouldn't have guessed that kind of ambiguity would be part of jQuery. Surprised! Thanks for pointing that out.

millerdev (approved):

PR #16904, opened 2017-07-04
BlobDB global fixtures

proteusvacuum:
Stores global fixture types in blob db, so they don't have to be regenerated every time. Since these are domain specific, they only ever have to get generated once, unless they get invalidated for some reason.

Opening for tests and for comments.

The downside of this is that the "global types" and "user types" will no longer be in alphabetical order, but I'm not sure why that was being done anyway. 

@millerdev @snopoke

### DIFF
--- a/corehq/apps/fixtures/fixturegenerators.py
+++ b/corehq/apps/fixtures/fixturegenerators.py
@@ -61,7 +61,8 @@ class ItemListsProvider(FixtureProvider):
             global_items = [db.get(domain, FIXTURE_BUCKET).read()]
         except NotFound:
             global_items = self.get_global_items(restore_user, global_types)
-            db.put(StringIO(" ".join(global_items)), domain, FIXTURE_BUCKET)
+            cached_element = [ElementTree.tostring(element, encoding='utf-8') for element in global_items]
+            db.put(StringIO("".join(cached_element)), domain, FIXTURE_BUCKET)
 
         user_items = self.get_user_items(restore_user, all_types, global_types)
         return global_items + user_items
@@ -125,7 +126,7 @@ class ItemListsProvider(FixtureProvider):
             # So we have to add a dummy empty_element child to prevent
             # this element from being empty.
             ElementTree.SubElement(fixture_element, 'empty_element')
-        return ElementTree.tostring(fixture_element, encoding="utf-8")
+        return fixture_element
 
     def _get_schema_element(self, data_type):
         schema_element = ElementTree.Element(
@@ -137,7 +138,7 @@ class ItemListsProvider(FixtureProvider):
             if field.is_indexed:
                 index_element = ElementTree.SubElement(indices_element, 'index')
                 index_element.text = field.field_name
-        return ElementTree.tostring(schema_element, encoding="utf-8")
+        return schema_element
 
 
 item_lists = ItemListsProvider()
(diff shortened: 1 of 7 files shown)


corehq/apps/fixtures/fixturegenerators.py
```diff
@@ -43,37 +47,59 @@ class ItemListsProvider(FixtureProvider):
...
+        if restore_state.overwrite_cache:
+            db.delete(domain, FIXTURE_BUCKET)
+
+        # get global types from the db
+        try:
+            global_items = [db.get(domain, FIXTURE_BUCKET).read()]
+        except NotFound:
+            global_items = self.get_global_items(restore_user, global_types)
+            cached_element = [ElementTree.tostring(element, encoding='utf-8') for element in global_items]
+            db.put(StringIO("".join(cached_element)), domain, FIXTURE_BUCKET)
```
proteusvacuum: Technically, this is storing invalid XML, since it has no root node. The [restore class](https://github.com/dimagi/commcare-hq/blob/master/corehq/ex-submodules/casexml/apps/phone/restore.py#L143-L144) just blindly appends this into the `<OpenRosaResponse>`, so this isn't an issue, however we do have to deal with it when looking up the [fixture by id](https://github.com/dimagi/commcare-hq/blob/master/corehq/ex-submodules/casexml/apps/phone/fixtures.py#L87-L94) (this is only called by cloudcare in very specific instances).

Another solution I explored was saving the cached xml as:
```<cached-fixture>...<cached-fixture>```
and when fetching it from the blobdb:
```py
cached_items = ElementTree.fromstring(db.get(domain, FIXTURE_BUCKET).read())
global_items = [item for item in cached_items]
```
However, this was much slower than just storing it as string and doing no manipulation at all (0.02s vs 0.5s for the eNikshay ItemLists locally)

corehq/apps/fixtures/upload/run_upload.py
```diff
@@ -191,9 +192,20 @@ def _update_progress(table_count, item_count, items_in_table):
...
 
 
+def clear_fixture_blob(domain, data_types):
+    """Clear global fixture cache for the domain f any of the data_types uploaded
+    are global
+
+    """
+    if any([data_type.is_global for data_type in data_types]):
+        db = get_blob_db()
+        db.delete(domain, FIXTURE_BUCKET)
```
snopoke: Tricky edge case but what if one was global before and now isn't? Maybe worth just always clearing this cache.

corehq/ex-submodules/casexml/apps/phone/fixtures.py
```diff
@@ -90,6 +92,14 @@ def get_fixture_by_id(self, fixture_id, restore_user):
         """
         fixtures = self._get_fixtures(restore_user, fixture_id)
         for fixture in fixtures:
+            if isinstance(fixture, basestring):
+                # could be a string if it's coming from cache
```
snopoke: added comment

snopoke:
+1 to this

corehq/apps/fixtures/upload/run_upload.py
```diff
@@ -191,9 +192,20 @@ def _update_progress(table_count, item_count, items_in_table):
...
+    clear_fixture_blob(domain, data_types)
     return return_val
 
 
+def clear_fixture_blob(domain, data_types):
+    """Clear global fixture cache for the domain f any of the data_types uploaded
+    are global
+
+    """
+    if any([data_type.is_global for data_type in data_types]):
```
millerdev: nitpick: rather than creating an intermediate list with a list comprehension, `any()` is marginally more efficient with a generator expression, like this:

```py
any(data_type.is_global for data_type in data_types)
```

proteusvacuum:
@snopoke made the change you suggested. Also removed the label, and tests passed.
Wasn't sure if your +1 was an approval or not.

PR #16905, opened 2017-07-05
Django's UploadedFile has a fileno method, but you can't use it

kaapstorm:
S3BlobDB errors when trying to send a Django UploadedFile instance, because UploadedFile has a fileno() method but it throws an UnsupportedOperation exception if you try to use it.

This change catches that exception, and falls back to using tell().

Error: https://sentry.io/dimagi/commcarehq/issues/305236253/

@snopoke cc @gcapalbo @orangejenny


kaapstorm:
This could use a unit test. I'll PR that separately.

gcapalbo (approved):

snopoke (approved):

snopoke:
:boot: :construction:

millerdev:
looks good to me

PR #16906, opened 2017-07-05
custom enikshay report: Improve referral report filter

NoahCarnahan:
Since this report will be used as a mobile UCR, the filter needs to be able to handle the location of any user that syncs. Therefore it needs to more gracefully handle locations that come from users who are not the intended users of the report.
@dannyroberts


custom/enikshay/ucr_filters.py
```diff
@@ -24,6 +24,13 @@ def get_hierarchy(self, location_id):
         """
         relevant_location_types = ["sto", "cto", "dto", "tu", "phi"]
         location = SQLLocation.objects.get(location_id=location_id)
+
+        if location.location_type.name not in relevant_location_types:
```
emord: relevant_location_types look like thye might be site codes and not names. is name the correct thing here?

emord: offlined

PR #16907, opened 2017-07-05
Split expression change

kkrampa:
* Make index expression not required
* Return whole array when index expression is not defined
* Added test


kkrampa:
@NoahCarnahan

NoahCarnahan:
Looks great, thanks.

PR #16911, opened 2017-07-05
Fixed device log migration.

emord:
@esoergel


esoergel (approved):

emord:
more info: the original migration attempted to rename the table twice, once to olddevicereportentry and then back to devicereportentry, which caused the mgiration to stall on NIC. This causes no sql changes (check with `./manage.py sqlmigrate phonelog 0010`)

emord:
@esoergel sorry for the churn, but I'm also removing an unnecessary migration for now. It just drops an unused index, but it acquires a full lock on the DB. This didn't show up on staging because I deployed staging in stages so the table wasn't used when I deployed that migration. Once the table has been deprecated the index can be removed again without issue

PR #10346, opened 2016-02-17
Show form case refs in summary

emord:
I think this can be merged whenever and not have any ill effects. It won't start doing anything until https://github.com/dimagi/Vellum/pull/611 is merged though

cc: @snopoke @millerdev

PR #16914, opened 2017-07-05
Kill formtranslate

dannyroberts:


corehq/apps/app_manager/models.py
```diff
@@ -987,6 +987,7 @@ def wrapped_xform(self):
     def validate_form(self):
         vc = self.validation_cache
         if vc is None:
+            # todo: now that we don't use formtranslate, does this still apply?
```
esoergel: Is this something you're planning to investigate at this point?

esoergel (approved):
Awesome!

millerdev:
```
ERROR: corehq.apps.app_manager.tests.test_real_build:TestRealBuild.test_real_build
  File "/vendor/lib/python2.7/site-packages/mock.py", line 1216, in patched
    arg = patching.__enter__()
  File "/vendor/lib/python2.7/site-packages/mock.py", line 1273, in __enter__
    self.target = self.getter()
  File "/vendor/lib/python2.7/site-packages/mock.py", line 1393, in <lambda>
    getter = lambda: _importer(target)
  File "/vendor/lib/python2.7/site-packages/mock.py", line 1126, in _importer
    thing = _dot_lookup(thing, comp, import_path)
  File "/vendor/lib/python2.7/site-packages/mock.py", line 1115, in _dot_lookup
    __import__(import_path)
ImportError: No module named FORMTRANSLATE_FORM_VALIDATION
```

corehq/apps/app_manager/models.py
dannyroberts: OK fine, thanks for prodding me.

Looks like this was added in here for a specific reason, to fix a bug @emord was encountering. Commit here: https://github.com/dimagi/commcare-hq/commit/010461d3a7232239ffdc4952388fe883ce342d03, some discussion on the PR here: https://github.com/dimagi/commcare-hq/pull/10346#r61318829.

@emord do you think this is still necessary with formplayer?

emord: This was added because hashtags in the formbuilder write xml binds/setvalues like:

`<bind nodeset="/data/some/quetsion"  calculate="/data/other/question + 1" vellum:calculate="#form/other/question + 1" />`

App validation failed whenever it ran into `vellum:calculate` because #form isn't a valid xpath. it wasn't clear why it cares that it isn't a valid xpath since it would be ignored by the phone. I don't know if that was fixed by formplayer's validation though

dannyroberts:
✅

dannyroberts:
> Will Pride: awesome form-translate got killed - confirming that Swiss was able to get Formplayer running? I think that was the reason this was still around

dannyroberts:
All cleared up, seems like that was a thing of the past, but swiss has formplayer set up and working so this is fine.

corehq/apps/app_manager/models.py
dannyroberts: OK, sounds like it's definitely fine to leave the stripping, and possibly not fine to remove it. I'm  tempted to leave both the original comment and the todo comment, so if at any point someone traces an issue back to this line they can see both why it was put there in the first place and a comment suggesting that it's possibly not needed anymore; then they can figure out if removing it works fine/fixes their issue.

millerdev (approved):

millerdev:
🎉  Yay!

benrudolph:
woah awesome!

PR #16917, opened 2017-07-06
Call add_progress() to give some feedback when starting to import locations

kaapstorm:
Context: [FB 244450](http://manage.dimagi.com/default.asp?244450)

This provides a little feedback to the user when we first pick up the task to import locations. Without this feedback, the first time the progress bar moves is after the task has been picked off the queue, the workbook has been inspected, and the first chunk of locations have been created. Hopefully this will allay the creeping dread that nothing is ever going to happen.

@gcapalbo @orangejenny

### DIFF
--- a/corehq/util/workbook_json/excel_importer.py
+++ b/corehq/util/workbook_json/excel_importer.py
@@ -15,6 +15,7 @@ class ExcelImporter(object):
     def __init__(self, task, file_ref_id):
         self.task = task
         self.progress = 0
+        self.total_rows = 100
 
         if self.task:
             DownloadBase.set_progress(self.task, 0, 100)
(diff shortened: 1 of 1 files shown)


orangejenny (approved):

corehq/util/workbook_json/excel_importer.py
```diff
@@ -15,6 +15,7 @@ class ExcelImporter(object):
     def __init__(self, task, file_ref_id):
         self.task = task
         self.progress = 0
+        self.total_rows = 100
```
gcapalbo: how come the default?

kaapstorm: That's so we can use the `add_progress()` method before we know the value of `total_rows`. ... We could use `DownloadBase.set_progress()` instead, but we're using 100 as a default value two lines down anyway, and I figured using `self.add_progress()` was a tiny bit cleaner.

PR #16920, opened 2017-07-06
ICDS Message Exports

gcapalbo:
@emord 
@kaapstorm @orangejenny

### DIFF
--- a/custom/icds/management/commands/get_icds_sms_export.py
+++ b/custom/icds/management/commands/get_icds_sms_export.py
@@ -1,0 +1,139 @@
+from collections import defaultdict
+from corehq.apps.sms.models import SMS
+from corehq.apps.locations.models import SQLLocation
+from corehq.apps.users.models import CommCareUser
+from corehq.form_processor.utils import is_commcarecase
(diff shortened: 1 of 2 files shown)


custom/icds/management/commands/get_icds_sms_export.py
```diff
@@ -0,0 +1,139 @@
...
+                'State Name',
+                'Text',
+                'AWC Id',
+                'LS Id',
+                'District Id',
+                'Block Id',
+                'State Id',
+            )
+
+            data = tuple(self.get_records(domain, start_timestamp, end_timestamp))
```
emord: How many records would this be? looks like get_records is a generator but this would load it all into memory I think

emord (approved):
Just one question, but it's likely not actionable

custom/icds/management/commands/get_icds_sms_export.py
gcapalbo: yeah i'm going to be running week by week, so it's going to be manageable right now. this is also just a stop gap until we have the report in place

PR #16923, opened 2017-07-06
migrate icds ucr

emord:
@dimagi/scale-team A script to move old data into the new tables. Unfortunately I think we need to do it one case at a time since we need to check that it doesn't exist in the table already.

The way I'm thinking of rolling this out is:

1) run this script for child_health cases
2) change the dashboard/aggregation code/config to point to child_health_tableau_v2 table
3) remove the old config id from the async indicator queue
4) Remove the static config from icds-cas
5) repeat for ccs_record cases and person cases
6) eventually delete the old data.

@sheelio

### DIFF
--- a/corehq/apps/userreports/management/commands/migrate_icds_ucr.py
+++ b/corehq/apps/userreports/management/commands/migrate_icds_ucr.py
@@ -92,7 +92,7 @@ class Command(BaseCommand):
             print(self.get_count(self.new_table))
             with connections['icds-ucr'].cursor() as cursor:
                 cursor.execute(sql_command)
-            print("count of old table %s: " % self.new_table)
+            print("count of new table %s: " % self.new_table)
             print(self.get_count(self.new_table))
 
     def get_count(self, table_name):
--- a/corehq/apps/userreports/management/commands/migrate_icds_ucr.py
+++ b/corehq/apps/userreports/management/commands/migrate_icds_ucr.py
@@ -1,11 +1,13 @@
 from __future__ import print_function
-import argparse
 from collections import namedtuple
 
 from django.core.management.base import BaseCommand
 from django.db import connections
 
-from corehq.apps.userreports.models import AsyncIndicator, get_datasource_config
+from corehq.apps.es import CaseES
+from corehq.apps.userreports.models import get_datasource_config
+from corehq.form_processor.models import CommCareCaseSQL
+from corehq.util.log import with_progress_bar
 
 
 COMMON_PERSON_COLUMNS = ()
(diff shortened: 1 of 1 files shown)


emord:
also important to note that this ignores the ES indexes here. We don't have a way to reindex those quickly (would be possible with an ES upgrade https://www.elastic.co/blog/reindex-is-coming). That's ok because it will be built later

corehq/apps/userreports/management/commands/migrate_icds_ucr.py
```diff
@@ -0,0 +1,103 @@
...
+            with connections['icds-ucr'].cursor() as cursor:
+                cursor.execute(
+                    "INSERT INTO " + new_table + " " +
+                    " (" + column_string + ") " +
+                    "SELECT " + column_string + " from " + old_table + " " +
+                    "WHERE " +
+                    "    doc_id = %s AND " +
+                    "    NOT EXISTS ( " +
+                    "        SELECT doc_id FROM " + new_table + "WHERE doc_id = %s " +
+                    "    ) ",
```
gcapalbo: We should be able to do this in one query without the iteration by doing something like:

    INSERT INTO <new table> (...)
    SELECT (A..., A..., ...)
    FROM <old table> A
    LEFT JOIN <new table> B ON A.doc_id = B.doc_id
    WHERE B.doc_id IS NULL

Essentially we do a left join from old table to new table and if there's no hit then everything on the right side of the join (aliased B) would be NULL. So assuming doc_id is never NULL the only way it's NULL is if the join didn't hit.

But that assumes that:

- nothing else is writing to <new table> while this query is running
- doc_id is a column that is NOT NULL

I also haven't tried bulk inserting into a table while also including that table in the select clause before. I think it should be ok because it should evaluate the select first. But it's worth testing and if it doesn't work then we could just create a temporary table that has a snapshot of new_table's doc ids, and join to that instead.

emord: Oh wow. that's way cleaner. I'll look into using that. I originally wanted to do this while the indicators were running,b ut I think it's fine to stop those while this query runs

snopoke (commented):
Looks good barring @gcapalbo's comment.

corehq/apps/userreports/management/commands/migrate_icds_ucr.py
gcapalbo: It also would be a good idea to have a balance check at the end of this process, so just comparing a count(*) on new table and count(*) on old table, and making sure that the counts are the same.

Or if new table is expected to have extra rows, we could just compare a count(*) from old table to a count(*) from old table join new table.

emord:
@gcapalbo I updated it to use join. the new strategy will be to stop the celery workers before running this script.

emord:
tested this locally on ccs_records:

old table before:

```
commcarehq_icds_testdata=# select doc_id, repeat_iteration, inserted_at, month, case_id, awc_id from "config_report_icds-cas_static-ccs_record_cases_monthly_33021c88";
       doc_id       | repeat_iteration |        inserted_at         |   month    | case_id | awc_id 
--------------------+------------------+----------------------------+------------+---------+--------
 doc_id             |                1 | 2017-07-07 15:52:44.854554 | 2017-06-01 | case_id | awc_id
 another_doc_id     |                1 | 2017-07-07 15:53:17.563868 | 2017-06-01 | case_id | awc_id
 yet_another_doc_id |                1 | 2017-07-07 15:53:30.67563  | 2017-06-01 | case_id | awc_id
(3 rows)
```

new table before:

```
commcarehq_icds_testdata=# select doc_id, repeat_iteration, inserted_at, month, case_id, awc_id from "config_report_icds-cas_static-ccs_record_cases_monthly_d0e2e49e";
     doc_id     | repeat_iteration |        inserted_at         |   month    | case_id |   awc_id    
----------------+------------------+----------------------------+------------+---------+-------------
 doc_id         |                1 | 2017-07-07 15:46:21.390758 | 2017-06-01 | case_id | awc_id
 another_doc_id |                1 | 2017-07-07 15:53:09.3467   | 2017-06-01 | case_id | diff_awc_id
(2 rows)
```

new table after:

```
commcarehq_icds_testdata=# select doc_id, repeat_iteration, inserted_at, month, case_id, awc_id from "config_report_icds-cas_static-ccs_record_cases_monthly_d0e2e49e";
       doc_id       | repeat_iteration |        inserted_at         |   month    | case_id |   awc_id    
--------------------+------------------+----------------------------+------------+---------+-------------
 doc_id             |                1 | 2017-07-07 15:46:21.390758 | 2017-06-01 | case_id | awc_id
 another_doc_id     |                1 | 2017-07-07 15:53:09.3467   | 2017-06-01 | case_id | diff_awc_id
 yet_another_doc_id |                1 | 2017-07-07 15:53:30.67563  | 2017-06-01 | case_id | awc_id
(3 rows)
```

This tested that `another_doc_id` was not updated (kept a different awc_id) and that `yet_another_doc_id` was added to the new table. there were no changes to the old table before/after the script.

corehq/apps/userreports/management/commands/migrate_icds_ucr.py
```diff
@@ -0,0 +1,122 @@
...
+        if options['dry_run']:
+            print(sql_command)
+        else:
+            print("count of old table %s: " % self.old_table)
+            print(self.get_count(self.old_table))
+            print("count of new table %s: " % self.new_table)
+            print(self.get_count(self.new_table))
+            with connections['icds-ucr'].cursor() as cursor:
+                cursor.execute(sql_command)
+            print("count of old table %s: " % self.new_table)
```
gcapalbo: did you mean to say "count of new table"

gcapalbo (approved):
looks good. when you stop celery, it probably would be a good idea to wait for all the worker processes to stop running too, since everything uses the bash runner now and stopping celery only puts a worker into warm shutdown, allowing it to finish any executing tasks.

can just use `ps -ef | grep <queue name>` to see if anything is still running, and kill any old/stale processes

corehq/apps/userreports/management/commands/migrate_icds_ucr.py
emord: Thanks. fixed that here: https://github.com/dimagi/commcare-hq/pull/16923/commits/3bac3603a2f0649b826debb37e143a30e52b7710

PR #16924, opened 2017-07-06
simplify format_angular_error and make log_errors required

dannyroberts:
This should make it harder to by mistake not log errors.

It also replaces all errors that are based on the exception message with the phrase "There was an error." To a user that is all an exception message means anyway, and to a dev just having the exception message is super not useful.

### DIFF
--- a/corehq/apps/export/views.py
+++ b/corehq/apps/export/views.py
@@ -673,13 +673,14 @@ class BaseDownloadExportView(ExportsPermissionsMixin, HQJSONResponseMixin, BaseP
         try:
             download_id = in_data['download_id']
         except KeyError:
-            return format_angular_error(_("Requires a download id"))
+            return format_angular_error(_("Requires a download id"), log_error=False)
         try:
             context = get_download_context(download_id)
         except TaskFailedError:
             return format_angular_error(
                 _("Download Task Failed to Start. It seems that the server "
-                  "might be under maintenance.")
+                  "might be under maintenance."),
+                log_error=False,
             )
         if context.get('is_ready', False):
             context.update({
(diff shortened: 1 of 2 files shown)


esoergel (approved):
Yikes, that's a lot of exception handling

corehq/apps/export/views.py
```diff
@@ -1831,11 +1820,10 @@ def get_app_data_drilldown_values(self, in_data):
         try:
             rmi_helper = ApplicationDataRMIHelper(self.domain, self.request.couch_user)
             response = rmi_helper.get_case_rmi_response()
-        except Exception as e:
+        except Exception:
             return format_angular_error(
-                _("Problem getting Create Export Form: {}").format(e.message),
+                _("Problem getting Create Export Form"),
                 log_error=True,
-                exception=e,
```
NoahCarnahan: Can you say a bit more about why it's advantageous to swallow the error message and not log it? Since this is a generic `Exception` clause, it seems like this exception potentially has a useful error message. (This comment applies to several of the usages in this PR)

dannyroberts: Yeah, so you'd think that removing the exception param would result in not logging the exception. My understanding is that basically passing in the error itself wasn't doing anything. If you look down to https://github.com/dimagi/commcare-hq/pull/16924/files#diff-6071ec21b34b72523fded4adb3cbb516L19, you'll see that `exception` is being passed in _as the message_ to `logging.exception`. I believe the original author just misunderstood `logging.exception`'s function signature. My understanding is that `logging.exception` is just like `logging.error`, but that it includes exc_info (by looking it up from a global place, not by making you pass in an exception object).

NoahCarnahan: Ah, very cool. I didn't know that `logging.exception` did that.

NoahCarnahan (approved):

corehq/apps/export/views.py
dannyroberts: @snopoke what I said here about `logging.exception` is true, and it does result in an error being sent to Sentry, is that all correct?

dannyroberts: Heads up, what I said about `logging.exception` wasn't quite right—it doesn't log to celery. (The other stuff was true.) I changed this in https://github.com/dimagi/commcare-hq/pull/16993 to use `notify_exception`, which will go to sentry. I could have used logging.getLogger('notify').exception instead with similar effect, but that didn't seem to be used anywhere, so I went with our internal convention and used `notify_exception`.

PR #16925, opened 2017-07-06
Fix bugs in get_related_indices

millerdev:
- bug: `get_cases(..., prefetched_indices=...)` did not set prefetched indices on returned cases.
- bug: some indices that should have been included were excluded by `get_related_indices` (added test for this).
- bug: in rare cases could mark a closed extension case as live (was not caught by tests before changing the index exclusion logic)

@proteusvacuum @dannyroberts cc @ctsims @snopoke


corehq/form_processor/backends/sql/dbaccessors.py
```diff
@@ -649,6 +649,10 @@ def get_cases(case_ids, ordered=False, prefetched_indices=None):
             cases = _order_list(case_ids, cases, 'case_id')
 
         if prefetched_indices:
+            # HACK work around bug in RawQuerySetWrapper that causes
+            # next(iter(cases)) is not next(iter(cases))
```
millerdev: This is a really bad bug. Going to have to dig deep to see what other things it affects.

corehq/ex-submodules/casexml/apps/phone/tests/test_sync_mode.py
```diff
@@ -2232,3 +2232,55 @@ class LiveQueryLooseSyncTokenValidationTest(LooseSyncTokenValidationTest):
...
+        payload = generate_restore_payload(
+            self.project, self.user, version=V2, **self.restore_options)
+        check_payload_has_case_ids(
+            self,
+            payload_string=payload,
+            username=self.user.username,
+            case_ids=[child_id, parent_id, other_parent_id]
+        )
+        self.assertIn(branch_index, payload)  # HACK
+        self.assertIn(wave_index, payload)  # HACK
```
millerdev: Because they use dumb string matching instead of structural validation. Following the example set at https://github.com/dimagi/commcare-hq/blob/f90631f7dd1d4af12c2b7a82aa060dbfa9352157/corehq/ex-submodules/casexml/apps/phone/tests/test_sync_mode.py#L1762-L1764

corehq/form_processor/backends/sql/dbaccessors.py
millerdev: Removing `RawQuerySetWrapper` is out of scope for this PR.

snopoke: agreed.

PR #16926, opened 2017-07-06
Remove custom custom query unless a specific key is being searched

proteusvacuum:
Now if you add `include_if: __prop` to the root of a custom query, it conditionally removes that query based on the presence of `__prop` in the search parameters.

These custom queries are getting a bit crazy now, but we haven't figured out a better way of doing "do this or this or not this" through a UI yet. This is also a "very advanced" feature, so I think this is still fine. 

@snopoke


esoergel (approved):

snopoke:
:boot: :construction:

PR #16927, opened 2017-07-07
Fix the Manage Forms report

esoergel:
This was a hell of an issue to debug.  The report wasn't showing up, no js errors, no nothing.  Eventually I figured out that the template for the report wasn't being rendered at all, so I found where the template is rendered (no easy task with our view code), and saw that the base template was being rendered directly.  With asynchronous to False, it will use the correct template.

@biyeun @orangejenny any ideas as to what happened here?  I think there's been some recent report changes, but I couldn't find anything which could've caused this.  Does this look like the correct fix anyways?

### DIFF
--- a/corehq/apps/data_interfaces/interfaces.py
+++ b/corehq/apps/data_interfaces/interfaces.py
@@ -188,6 +188,7 @@ class BulkFormManagementInterface(SubmitHistoryMixin, DataInterface, ProjectRepo
     name = ugettext_noop("Manage Forms")
     slug = "bulk_archive_forms"
     report_template_path = 'data_interfaces/interfaces/archive_forms.html'
+    asynchronous = False
 
     def __init__(self, request, **kwargs):
         super(BulkFormManagementInterface, self).__init__(request, **kwargs)


esoergel:
I just git bisected this, and this was the commit that introduced the bug: https://github.com/dimagi/commcare-hq/commit/a246d190cd

Honestly I'm still not quite sure how _that_ caused _this_ issue, but my guess is that this report was using the wrong convention with the template blocks.  I will note that the report template extends reports/async/tabular.html, but it doesn't actually render unless I set `asynchronous = False`, which seems a little bonkers.  @orangejenny can you make any sense of this?

orangejenny:
This is definitely my fault, apologies for the debugging headache. It was happening because the js to instantiate the report wasn't being included. So it should be fixed in js, I'll close this and open another PR.

biyeun:
Yikes! Yes this is the tricky situation of including new JS in the served HTML after the page has rendered. This is one of the reasons why reports are so terrible. we just shouldn't be serving straight up after the initial dom rendering...ever. Sorry, Ethan. this is a deep dark relic of young Biyeun trying to quickly refactor reports and get them to at least load asynchronously...why i kept it as chunks of html is still mind boggling to this day.

PR #16929, opened 2017-07-07
set couch auth for docker

snopoke:
@dimagi/team-commcare-hq FYI adding default admin user to CouchDB docker container.

### DIFF
--- a/docker/hq-compose.yml
+++ b/docker/hq-compose.yml
@@ -36,7 +36,7 @@ postgres:
 
 couch:
   image: klaemo/couchdb:2.0-dev
-  command: ["--with-haproxy", "--with-admin-party-please", "-n", "1"]
+  command: ["--with-haproxy", "--admin=commcarehq:commcarehq", "-n", "1"]
   expose:
     - "5984"
   volumes:
--- a/docker/localsettings.py
+++ b/docker/localsettings.py
@@ -75,8 +75,8 @@ COUCH_DATABASES = {
         # for production this ought to be set to true on your configured couch instance
         'COUCH_HTTPS': False,
         'COUCH_SERVER_ROOT': 'couch:5984',  # 6984 for https couch
-        'COUCH_USERNAME': '',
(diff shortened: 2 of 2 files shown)


kaapstorm (approved):

proteusvacuum:
yay!

millerdev:
Awesome!

Do we need to rebuild our containers or is there any other action necessary for pre-existing containers? I'm being lazy and not just trying it on my dev setup, so feel free to just ignore me if it will be obvious when I get around to that.

snopoke:
No rebuild required:
```
$ docker rm -f hqservice_couch_1
$ ./scripts/docker up -d
```

emord:
This might be causing tests to fail intermittently. Seeing a lot of errors in travis tests "Waiting for link to COUCH" example: https://travis-ci.org/dimagi/commcare-hq/jobs/251186023

snopoke:
@emord it does seem that way. I looked at the startup script and running with auth definitely does more work that with admin part: https://gitbox.apache.org/repos/asf?p=couchdb.git;a=blob;f=dev/run;h=2f25071a150142cc703cc4a89de0a7a7015a7b9a;hb=refs/heads/master (see `cluster_setup` vs `cluster_setup_with_admin_party`)

I'll look into making it use admin part in travis only.

emord:
@snopoke @dannyroberts @proteusvacuum Do you guys remember why we added this in the first place?

asking because @sheelio had some issues with running docker locally due to the difference between travis having admin party, but locally needing a password.

proteusvacuum:
My comment "yay!" was because at one point formplayer required you to have a couch user to run, and that user was blown away every time I restarted my computer / docker. However, formplayer no longer uses couch so it doesn't matter any more. Not sure if that was the original reason for this though.

emord:
@proteusvacuum Awesome, that's what I remembered. @sheelio looks like you're good to change both of those docker scripts

PR #16932, opened 2017-07-07
Move changes from QA to live Enikshay domain

kkrampa:
@NoahCarnahan @proteusvacuum

### DIFF
--- a/custom/enikshay/ucr/data_sources/episode.json
+++ b/custom/enikshay/ucr/data_sources/episode.json
@@ -6169,6 +6169,36 @@
                 }
             },
             {
+                "column_id": "patients_with_key_populations",
+                "type": "boolean",
+                "filter": {
+                    "type": "and",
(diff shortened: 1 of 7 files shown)


kkrampa:
We need to rebuild three data sources: test, episode and voucher.

emord:
@NoahCarnahan these rebuilds are probably going to fail if they're put in the celery queue. You'll probably need to run the async rebuild that you had to for episode cases before

NoahCarnahan:
@emord 
Ah, interesting. Why do you think they'll fail in the celery queue? There is a task that automatically rebuilds tables when the configuration has changed, right? Do we need to do something to disable that?

emord:
@NoahCarnahan I think they'll start, but it will take a really long time for them to complete because of hte # of cases in enikshay. that will result in only a partial rebuild. similar to what you saw before

PR #16935, opened 2017-07-07
don't remove media reference is there are form errors

snopoke:
https://manage.dimagi.com/default.asp?255929

If there are form errors it shouldn't get to this stage but it could still be the case that the validation fails due to Formplayer error. If that happens any media in the forms that fail will get removed.


orangejenny (approved):

orangejenny:
fascinating, glad to see resolution to this. can't imagine this was a fun one to track down.

PR #16936, opened 2017-07-07
Add score to case search

proteusvacuum:
Adds a property `commcare_search_score` to the case when it is returned from a search so that case lists can be sorted by relevance.

@snopoke 
FYI @wpride

### DIFF
--- a/corehq/apps/case_search/templates/case_search/case_search.html
+++ b/corehq/apps/case_search/templates/case_search/case_search.html
@@ -119,23 +119,24 @@
                             <th>{% trans "Closed" %}</th>
                             <th>{% trans "Properties" %}</th>
                             <th>{% trans "Indices" %}</th>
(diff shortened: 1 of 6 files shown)


millerdev (approved):

mkangia:
Nice!

snopoke (approved):

wpride:
Very cool. How do you see this being configured from an app perspective? Will case claim results show in a different list than the normal case list?

proteusvacuum:
@wspride I think you would just add `commcare_search_score` as another sort param to the regular case list. 
screenshot from 2017-07-07 14-40-08
If I've understood how that works, it will sort by score, then by name, so things with the same score will be ordered by name, right?

wpride:
Cool, didn't realize you could sort by multiple params so easily - very convenient

PR #16937, opened 2017-07-07
deprecate old ccs child health

emord:
@sheelio will remove the full json in a later PR. just doing this so it's easy to revert if necessary


sheelio:
@emord 

I'll need to dig into the test failure (it might just be a slight difference in the definitions of indicators I used).

sheelio:
@emord 

Fixed failures so this should be good to merge after test pass.

emord:
thanks for fixing the tests @sheelio

PR #16941, opened 2017-07-07
Simplify case list form

orangejenny:
Did some cleanup of the ui for registration from case list, fixed some validation/messaging that wasn't working, and made the code more concise, largely by moving logic from js to django, since this widget isn't especially interactive.

Before, when reg from case list wasn't available, you saw a disabled dropdown and needed to click for more info:
screen shot 2017-07-07 at 2 56 34 pm

After, when unavailable, you see only the reasons your particular module can't have reg from case list:
screen shot 2017-07-07 at 2 56 51 pm

Before, the error state (like the reg form being deleted) weren't all that clear:
screen shot 2017-07-07 at 2 50 31 pm

After, more specific error messaging:
screen shot 2017-07-07 at 2 41 45 pm

`w=1`

@kaapstorm / @gcapalbo

### DIFF
--- a/corehq/apps/app_manager/static/app_manager/js/module-view.js
+++ b/corehq/apps/app_manager/static/app_manager/js/module-view.js
@@ -113,12 +113,6 @@ $(function () {
                 formSet = !!data.form_id,
                 formMissing = formSet && !formOptions[data.form_id];
 
-            self.toggleState = function(active) {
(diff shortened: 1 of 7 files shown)


kaapstorm (approved):

gcapalbo (approved):
(restarted the build)

PR #16944, opened 2017-07-07
Format fogbugz tickets tracebacks the way python does

dannyroberts:
It's been annoying me that at the bottom of bug reports in fogbugz, the tracebacks end without the actual exception type or message, the way they do in python. (That information also doesn't seem to be anywhere else directly in the ticket, just the linked sentry report.)

### DIFF
--- a/corehq/apps/hqwebapp/views.py
+++ b/corehq/apps/hqwebapp/views.py
@@ -80,6 +80,22 @@ def is_deploy_in_progress():
     return cache.get(DEPLOY_IN_PROGRESS_FLAG) is not None
 
 
+def format_traceback_the_way_python_does(type, exc, tb):
+    """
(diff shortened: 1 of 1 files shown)


dannyroberts:
@NoahCarnahan

corehq/apps/hqwebapp/views.py
```diff
@@ -80,6 +80,22 @@ def is_deploy_in_progress():
     return cache.get(DEPLOY_IN_PROGRESS_FLAG) is not None
 
 
+def format_traceback_the_way_python_does(type, exc, tb):
```
esoergel: descriptive function name!

esoergel (approved):
Love this idea, will be very happy to see it in action!

NoahCarnahan (approved):

PR #16945, opened 2017-07-07
Externalize js: reports/async/basic.html

orangejenny:
low-hanging fruit

@kaapstorm / @gcapalbo


kaapstorm (approved):

kaapstorm:
I realise you didn't put it in, but do you think we should take out the `console.log`ging?

gcapalbo (approved):

orangejenny:
@kaapstorm I think you're probably right, but going to err on the side of leaving it alone.

PR #16948, opened 2017-07-07
Add edited_by_user_id to forms API

esoergel:
followup from https://github.com/dimagi/commcare-hq/pull/16883
Feel free to propose alternative names or ways to present the field.
@snopoke @emord
@krsdimagi FYI

### DIFF
--- a/corehq/apps/api/resources/v0_4.py
+++ b/corehq/apps/api/resources/v0_4.py
@@ -106,6 +106,11 @@ class XFormInstanceResource(SimpleSortableResourceMixin, HqBaseResource, DomainS
             and bundle.obj.openrosa_headers.get('HTTP_X_OPENROSA_VERSION')
         )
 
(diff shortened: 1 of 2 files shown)


snopoke:
I don't think we can call this field `edited_by_user_id` since it only represents that if the form was edited. For unedited forms it may be `null` if the form wasn't submitted with auth or else it will the ID of the submitting user.

I think calling it `last_modified_by` might be OK but still not great.

esoergel:
@snopoke what do you think of adding a `if edited_on:` to the top, so that it will only be set if the form has been edited?  As you pointed out, the meaning can be confusing otherwise.

snopoke:
Yea, that seems like a good idea.

esoergel:
I made that change, so I think this should be g2m.

emord (approved):

PR #16958, opened 2017-07-10
[WIP] enikshay 2b migration

NoahCarnahan:
The enikshay project requires patient data collected in excel spreadsheets to be imported into HQ. This management command parses those spreadsheets and creates the corresponding cases. There are three different formats, and each format contains some, but not all, of the same data.

There are mappings defined for each, which map a column identifier (string) to an excel column index. The case structures are assembled by using these mappings to grab the appropriate values from the row. If the row or format is missing data for the given column identifier, then the corresponding case property will just not be initialized.

This management command is a work in progress, but I thought it would be good to get parts of it merged sooner rather than later so that I can start making smaller PRs which will be easier to review. My apologies for not doing that sooner.

@dannyroberts @nickpell 
Probably best reviewed with the full diff


custom/enikshay/two_b_datamigration/management/commands/import_drtb_cases.py
```diff
@@ -0,0 +1,765 @@
...
+    @staticmethod
+    def get_city_constants(format):
+        # TODO: Use constants for formats
+        if format in ("mehsana2016", "mehsana2017"):
+            return MehsanaConstants
+        elif format == "mumbai":
+            return MumbaiColumnMapping
+        else:
+            raise Exception("Invalid format. Format must be mehsana2016, mehsana2017, or mumbai")
+
```
NoahCarnahan: This is not the end of the file, ignoring.

custom/enikshay/two_b_datamigration/management/commands/import_drtb_cases.py
```diff
@@ -0,0 +1,765 @@
...
+    REACTIVE = "reactive"
+    if not value:
+        return None
+    if value.startswith("Non Reactive") or value.startswith("NR") or value.startswith("Nr"):
+        return NON_REACTIVE
+    if value.startswith("R ") or value.startswith("Reactive") or value.startswith("Ractive"):
+        return REACTIVE
+    return {
+        "Pos": REACTIVE,  #? TODO
+        "Positive": REACTIVE,  #? TODO
```
NoahCarnahan: ignoring

custom/enikshay/two_b_datamigration/management/commands/import_drtb_cases.py
```diff
@@ -0,0 +1,765 @@
...
+    NON_REACTIVE = "non_reactive"
+    REACTIVE = "reactive"
+    if not value:
+        return None
+    if value.startswith("Non Reactive") or value.startswith("NR") or value.startswith("Nr"):
+        return NON_REACTIVE
+    if value.startswith("R ") or value.startswith("Reactive") or value.startswith("Ractive"):
+        return REACTIVE
+    return {
+        "Pos": REACTIVE,  #? TODO
```
NoahCarnahan: ignoring

NoahCarnahan:
✅ (ignoring the remaining stickler comments)

PR #16962, opened 2017-07-10
Use Unique IDs in APP Manager URLs

biyeun:
@orangejenny 

I did a bunch of local testing on this and was very meticulous. Should we run this by QA?

fyi @proteusvacuum


orangejenny:
QA might be worthwhile just because it's a fairly large change. On a similar note...this is going to break people who have open a page before deploy and then press "Save" after deploy. What would you think of keeping the old URLs around for a short while? https://github.com/dimagi/commcare-hq/pull/16208#pullrequestreview-38774461 has a discussion of this.

orangejenny:
Build failure maybe real? It's from app manager: `corehq.apps.app_manager.tests.test_build_errors:BuildErrorsTest.test_case_list_form_advanced_module_different_case_config`

corehq/apps/app_manager/templates/app_manager/v1/form_view.html
```diff
@@ -107,7 +107,9 @@
     {# CloudCare #}
     {% initial_page_data 'app_id' app.id %}
     {% initial_page_data 'module_id' module.id %}
+    {% initial_page_data 'module_unique_id' module.unique_id %}
```
orangejenny: Don't see why this is necessary to add, but it's v1 only anyway so not concerned about it.

corehq/apps/app_manager/views/modules.py
```diff
@@ -698,13 +698,13 @@ def undo_delete_module(request, domain, record_id):
 
 @no_conflict_require_POST
 @require_can_edit_apps
-def overwrite_module_case_list(request, domain, app_id, module_id):
+def overwrite_module_case_list(request, domain, app_id, module_unique_id):
     app = get_app(domain, app_id)
-    source_module_id = int(request.POST['source_module_id'])
+    source_module_unique_id = int(request.POST['source_module_unique_id'])
```
orangejenny: Isn't `int` going to fail?

corehq/apps/app_manager/views/modules.py
```diff
@@ -1078,7 +1078,13 @@ def _save_case_list_lookup_params(short, case_list_lookup, lang):
...
 @require_GET
 @require_deploy_apps
-def view_module(request, domain, app_id, module_id):
+def view_module(request, domain, app_id, module_unique_id):
+    from corehq.apps.app_manager.views.view_generic import view_generic
+    return view_generic(request, domain, app_id, module_unique_id=module_unique_id)
+
+@require_GET
+@require_deploy_apps
+def view_module_legacy(request, domain, app_id, module_id):
```
orangejenny: Good call keeping these views. Might be worth a comment explaining why, don't think it'll be obvious to a new dev.

corehq/apps/app_manager/static/app_manager/js/app_manager.js
```diff
@@ -296,6 +300,40 @@ hqDefine('app_manager/js/app_manager.js', function () {
                 $(related).data(name, value);
             });
         }
+        function resetIndexes($sortable) {
```
orangejenny: Feel like I'm missing something big, but isn't this unnecessary now that the view module / view form / form builder links are all based on unique id, not on index?

corehq/apps/app_manager/templates/app_manager/v2/partials/form_link_primary.html
```diff
@@ -4,12 +4,14 @@
 
 {# For APP BUILDER PROTOTYPE use only #}
 
-<a id="view_form_{{ module.id }}_{{ form.id }}_sidebar"
+<a id="view_form_{{ module.unique_id }}_{{ form.unique_id }}_sidebar"
```
orangejenny: The v1 equivalent of this in managed_app.html isn't updated to use unique ids.

corehq/apps/app_manager/views/modules.py
orangejenny: (I see that urls.py says "for docs" but I don't really know what that means)

corehq/apps/app_manager/views/view_generic.py
```diff
@@ -92,7 +92,12 @@ def view_generic(request, domain, app_id=None, module_id=None, form_id=None,
                 raise Http404()
             form_id = form.id
 
-    except ModuleNotFoundException:
+        if form is not None and module is None:
```
orangejenny: stylistic, but `if form and module` would be easier to read

corehq/apps/app_manager/views/view_generic.py
```diff
@@ -80,7 +80,7 @@ def view_generic(request, domain, app_id=None, module_id=None, form_id=None,
                 raise Http404()
             module_id = module.id
 
-        if form_id:
+        if form_id and module is not None:
```
orangejenny: This (form_id but no module) should never happen, right? Might be preferable to just fail hard here on the undefined module rather than having unexpected behavior (a redirect to the wrong page?) later on.

corehq/apps/app_manager/static/app_manager/js/app_manager.js
```diff
@@ -240,46 +263,27 @@ hqDefine('app_manager/js/app_manager.js', function () {
...
+                if (!form.data('clicked')) {
+                    form.data('clicked', 'true');
+                    $('.new-module-icon').removeClass().addClass("fa fa-refresh fa-spin");
+                    form.submit();
+                }
+            });
+            $('.new-form').on('click', function (e) {
+                var dataType = $(this).data("type");
+                var moduleId = $(this).data("module");
+                var form = $("#form-to-create-new-form-for-module-" + moduleId);
```
orangejenny: Does this work right? It looks like `.new-form` uses the index-based id (https://github.com/dimagi/commcare-hq/blob/581fce55f27b3677a6701969a06d7cfbac1e06aa/corehq/apps/app_manager/templates/app_manager/v1/managed_app.html#L206) but `#form-to-create-new-form-for-module-X` uses the unique id (https://github.com/dimagi/commcare-hq/blob/581fce55f27b3677a6701969a06d7cfbac1e06aa/corehq/apps/app_manager/templates/app_manager/v1/managed_app.html#L193).

orangejenny (changes requested):
Couple of changes requested.

corehq/apps/app_manager/views/modules.py
biyeun: yeah good point.

corehq/apps/app_manager/views/modules.py
biyeun: ok will add that comment. great points

corehq/apps/app_manager/static/app_manager/js/app_manager.js
biyeun: it's still necessary because the index values have at least been updated. updateprop doesn't have to be in there tho

corehq/apps/app_manager/static/app_manager/js/app_manager.js
biyeun: i guess at this point it doesn't matter because we've switched full on to v2

orangejenny: Yeah, but I'm anticipating we'll keep v1 around for a couple more weeks and don't love the idea of having known bugs in it. Anyway, up to you, agree it's a very low risk of anyone encountering this.

biyeun:
@orangejenny waiting on tests before I push to staging...

biyeun:
@orangejenny did you happen to take a look at the changes here? putting this on staging rn

orangejenny (approved):
latest commits look good

biyeun:
@orangejenny sanity checked and rebased master for a cleaner git history.

biyeun:
@orangejenny QA PASSSED!!!!!!!!! 🎉 🎈 🎂

PR #16967, opened 2017-07-10
Added load based autoscaler.

emord:
@dimagi/scale-team Started work on autoscaling celery tasks based on load. I think we'll also want to have some time filters (like have if it's at night, we should allow higher load)

Currently the only way I see how to use this autoscaler is to specify `CELERYD_AUTOSCALER` in  settings.py, but that would cause all celery tasks to use this scaler. My first thought was it could be passed in an environment variable, but that seems messy. @gcapalbo do you know any way we could have this only apply to the ucr indicator queue?

### DIFF
--- a/corehq/util/celery_utils.py
+++ b/corehq/util/celery_utils.py
@@ -205,8 +205,13 @@ class LoadBasedAutoscaler(Autoscaler):
         if cur > procs and normalized_load < 0.90:
             self.scale_up(cur - procs)
             return True
-        elif cur < procs or normalized_load > 0.90:
+        elif cur < procs:
             self.scale_down((procs - cur) - self.min_concurrency)
+            return True
+        elif normalized_load > 0.90 and procs > self.min_concurrency:
+            # if load is to0 high trying scaling down 1 worker at a time.
+            # if we're already at minimum concurrency let's just ride it out
+            self.scale_down(1)
             return True
 
 
@@ -221,10 +226,9 @@ class OffPeakLoadBasedAutoscaler(LoadBasedAutoscaler):
 
     def _maybe_scale(self, req=None):
         procs = self.processes
-        cur = min(self.qty, self.max_concurrency)
 
-        if not self._is_off_peak():
-            self.scale_down((procs - cur) - self.min_concurrency)
+        if not self._is_off_peak() and procs > self.min_concurrency:
+            self.scale_down(1)
             return True
 
         super(OffPeakLoadBasedAutoscaler, self)._maybe_scale(req)
--- a/corehq/util/celery_utils.py
+++ b/corehq/util/celery_utils.py
@@ -211,14 +211,15 @@ class LoadBasedAutoscaler(Autoscaler):
         if cur > procs and normalized_load < 0.90:
             self.scale_up(cur - procs)
             return True
-        elif cur < procs:
-            self.scale_down((procs - cur) - self.min_concurrency)
-            return True
-        elif normalized_load > 0.90 and procs > self.min_concurrency:
-            # if load is too high trying scaling down 1 worker at a time.
-            # if we're already at minimum concurrency let's just ride it out
-            self.scale_down(1)
-            return True
+        elif procs > self.min_concurrency:
+            if cur < procs:
+                self.scale_down(min(procs - cur, procs - self.min_concurrency))
+                return True
+            elif normalized_load > 0.90:
+                # if load is too high trying scaling down 1 worker at a time.
+                # if we're already at minimum concurrency let's just ride it out
+                self.scale_down(1)
+                return True
 
 
 class OffPeakLoadBasedAutoscaler(LoadBasedAutoscaler):
--- a/corehq/util/celery_utils.py
+++ b/corehq/util/celery_utils.py
@@ -9,6 +9,7 @@ from celery.backends.base import DisabledBackend
 from celery.task import task
 from celery.worker.autoscale import Autoscaler
 from django.conf import settings
+from djcelery.loaders import DjangoLoader
 from datetime import datetime
 from time import sleep, time
 
@@ -232,3 +233,17 @@ class OffPeakLoadBasedAutoscaler(LoadBasedAutoscaler):
             return True
 
         super(OffPeakLoadBasedAutoscaler, self)._maybe_scale(req)
+
+
+class LoadBasedLoader(DjangoLoader):
+    def read_configuration(self):
+        ret = super(LoadBasedLoader, self).read_configuration()
+        ret['CELERYD_AUTOSCALER'] = 'corehq.util.celery_utils:LoadBasedAutoscaler'
+        return ret
+
+
+class OffPeakLoadBasedLoader(DjangoLoader):
+    def read_configuration(self):
+        ret = super(LoadBasedLoader, self).read_configuration()
+        ret['CELERYD_AUTOSCALER'] = 'corehq.util.celery_utils:OffPeakLoadBasedAutoscaler'
+        return ret
--- a/corehq/util/celery_utils.py
+++ b/corehq/util/celery_utils.py
@@ -202,10 +202,10 @@ class LoadBasedAutoscaler(Autoscaler):
         max_avg = max(load_avgs[0], load_avgs[1])
         normalized_load = max_avg / available_cpus
 
-        if cur > procs and normalized_load < 0.95:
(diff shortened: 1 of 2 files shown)


corehq/util/celery_utils.py
```diff
@@ -186,3 +190,21 @@ def get_running_workers(timeout=10):
         worker_names.extend(worker_info.keys())
 
     return worker_names
+
+
+class LoadBasedAutoscaler(Autoscaler):
+    def _maybe_scale(self, req=None):
```
emord: This function in the base class is https://github.com/celery/celery/blob/v3.1.18/celery/worker/autoscale.py#L84-L92

emord:
Also added an offpeak load based auto scaler. My thinking is that celery machines can use the loadbased autoscaler and will run all day, while webworkers and pillow machines and whatever else that has spare cpu could run as off peak through the night.

hoping to be able to specify autoscaler in environments.yml file if possible, so it could potentially be used for other envs

emord:
On second thought it's possible that we could just use the load based as the default. It would ensure that no machine is too highly resourced for very long. 

However, I'm not sure what would happen in the instance where one queue has a lot of work to get through. For example, if there are 10,000 async indicators to process and they are processing to the point where the server has a load of 1.0, and then 1,000 reminders come in, would the reminders be forced to wait on those indicators? that could be undesirable behavior and a little hard to debug.

corehq/util/celery_utils.py
```diff
@@ -186,3 +190,41 @@ def get_running_workers(timeout=10):
...
+
+        available_cpus = multiprocessing.cpu_count()
+        load_avgs = os.getloadavg()
+        max_avg = max(load_avgs[0], load_avgs[1])
+        normalized_load = max_avg / available_cpus
+
+        if cur > procs and normalized_load < 0.90:
+            self.scale_up(cur - procs)
+            return True
+        elif cur < procs or normalized_load > 0.90:
```
gcapalbo: i think we might want to split this into two elif statements, one for cur < procs, and one for normalized_load > 0.90.

otherwise if the load is high but cur > procs we'd end up scaling down a negative number

corehq/util/celery_utils.py
```diff
@@ -186,3 +190,41 @@ def get_running_workers(timeout=10):
...
+
+        return False
+
+    def _maybe_scale(self, req=None):
+        procs = self.processes
+        cur = min(self.qty, self.max_concurrency)
+
+        if not self._is_off_peak():
+            self.scale_down((procs - cur) - self.min_concurrency)
+            return True
```
gcapalbo: is the intention that we'd scale down to min_concurrency when we're not off peak? if so then i think this should scale down procs - self.min_concurrency, but also should check that procs > self.min_concurrency before doing so

gcapalbo:
On Mon, Jul 10, 2017 at 10:32 PM, Jonathan Emord <notifications@github.com>
wrote:

> Also added an offpeak load based auto scaler. My thinking is that celery
> machines can use the loadbased autoscaler and will run all day, while
> webworkers and pillow machines and whatever else that has spare cpu could
> run as off peak through the night.
>
> hoping to be able to specify autoscaler in environments.yml file if
> possible, so it could potentially be used for other envs
>

Yeah I agree it would be ideal to be able to specify the autoscaler to use
on a per-worker basis in environments.yml. We might be able to do that by
setting certain params when starting the worker. I haven't looked into it
much but we might be able to use a custom loader (--loader param) or the
--config param to somehow add this on a per-worker basis.

If not, I think the Autoscaler class has access to the worker
<https://github.com/celery/celery/blob/v3.1.18/celery/worker/autoscale.py#L75>
so
we might be able to inspect that to at least restrict this bases on queue
or something if we need to.

I think it's also important to remember this only applies to workers using
prefork (not gevent) pooling.

corehq/util/celery_utils.py
emord: Thanks for the close review. https://github.com/dimagi/commcare-hq/pull/16967/commits/404b64743608ff6b77a66bae968817f289a92e51

I changed it to slow down 1 process at a time.

emord:
To enable the new loaders we'll need to pass in an environment variable before the manage.py

`CELERY_LOADER='corehq.util.celery_utils.LoadBasedLoader' ./manage.py celery worker`

This is a deprecated format, but it's what djcelery uses and the documentation says that if this is set it will not use `--loader` https://github.com/celery/celery/blob/v3.1.18/celery/app/__init__.py#L39-L42

corehq/util/celery_utils.py
```diff
@@ -186,3 +190,45 @@ def get_running_workers(timeout=10):
...
     return worker_names
+
+
+class LoadBasedAutoscaler(Autoscaler):
+    def _maybe_scale(self, req=None):
+        procs = self.processes
+        cur = min(self.qty, self.max_concurrency)
+
+        available_cpus = multiprocessing.cpu_count()
+        load_avgs = os.getloadavg()
```
snopoke: should this handle an OSError?

snopoke:
You should use the `environment` param in the supervisor config. For
reference see
https://github.com/dimagi/commcare-hq-deploy/blob/master/fab/services/templates/supervisor_django.conf#L4-L4

corehq/util/celery_utils.py
```diff
@@ -186,3 +190,45 @@ def get_running_workers(timeout=10):
...
+                return True
+
+        return False
+
+    def _maybe_scale(self, req=None):
+        procs = self.processes
+
+        if not self._is_off_peak() and procs > self.min_concurrency:
+            self.scale_down(1)
+            return True
```
snopoke: Won't this just oscillate between `min_concurrency + 1` and `min_concurrency` (assuming there are tasks in the queue)

emord:
> You should use the `environment` param in the supervisor config

Yep. I was just testing that out on the setup release when i dropped the device log table
https://github.com/dimagi/commcare-hq-deploy/pull/321

corehq/util/celery_utils.py
```diff
@@ -186,3 +190,45 @@ def get_running_workers(timeout=10):
...
+    def _maybe_scale(self, req=None):
+        procs = self.processes
+        cur = min(self.qty, self.max_concurrency)
+
+        available_cpus = multiprocessing.cpu_count()
+        load_avgs = os.getloadavg()
+        max_avg = max(load_avgs[0], load_avgs[1])
+        normalized_load = max_avg / available_cpus
+
+        if cur > procs and normalized_load < 0.90:
```
snopoke: any reason not to make this 1?

corehq/util/celery_utils.py
emord: https://github.com/dimagi/commcare-hq/pull/16967/commits/1e822e09bd34cad31810e8d41f4e90a8f6ec5e4c

corehq/util/celery_utils.py
emord: https://github.com/dimagi/commcare-hq/pull/16967/commits/a2a2f5a147681cb062c4bd301ed99c9ae91ee81d

corehq/util/celery_utils.py
emord: if the load is ~1 then the machine is totally saturated right? I was thinking that may make it hard to log in and debug issues. 0.90 seemed like pretty good utilization while still allowing for us to log in

snopoke (approved):

corehq/util/celery_utils.py
```diff
@@ -186,3 +190,45 @@ def get_running_workers(timeout=10):
...
+            self.scale_up(cur - procs)
+            return True
+        elif cur < procs:
+            self.scale_down((procs - cur) - self.min_concurrency)
```
gcapalbo: i think we could still end up with a negative number here in some circumstances

we should probably make this something like:

    elif cur < procs and procs > self.min_concurrency:
        self.scale_down(min(procs - cur, procs - self.min_concurrency))

emord: Thanks. took me a minute to get my head around how that could happen. https://github.com/dimagi/commcare-hq/pull/16967/commits/413c07f378903454da212d46c398a914d925cb42

FWIW this bug is also in the default autoscaler (i also checked the latest version of celery 3.1.25 and it wasn't fixed). It's possible we've never run into it because all of our mins are 0

gcapalbo (approved):

PR #16968, opened 2017-07-11
Removed old device logs table.

emord:
@javierwilson to be deployed today so we can ansewr the question about storage for es2.

This table and indexes ar around 500 GB on NIC. The table should no longer be in use. I think that this should return pretty quickly despite the size. Is that correct @dimagi/scale-team ?


corehq/ex-submodules/phonelog/models.py
```diff
@@ -4,37 +4,9 @@
...
-        ]
-
-    def __repr__(self):
-        return u"OldDeviceReportEntry(domain='{}', msg='{}')".format(self.domain, self.msg)
-
-
 @architect.install('partition', type='range', subtype='date', constraint='day', column='server_date')
 class DeviceReportEntry(models.Model):
-    xform_id = models.CharField(max_length=COUCH_UUID_MAX_LEN)
+    xform_id = models.CharField(max_length=COUCH_UUID_MAX_LEN, db_index=True)
```
emord: I messed up in my first PR and this index was added even though it is not technically needed. Will drop it in a later PR. (will probably need to do it outside of a migration on ICDS)

emord:
weird test failure in supply that doesn't seem related. restarted the job

snopoke (approved):

PR #16972, opened 2017-07-11
revert fixture cache PRs

proteusvacuum:
We're getting weird bugs with fixtures that we can't repro, e.g.
https://sentry.io/dimagi/formplayer/issues/278668711/events/6250443091/

This is a combination of reverting (couldn't revert 2 PRs at a time)
https://github.com/dimagi/commcare-hq/pull/16904/
and https://github.com/dimagi/commcare-hq/pull/16951/

I really don't understand why this would only happen on specific users for global fixtures as the cache is domain wide, and should just get sent down with every sync and restore. 

Will look into it further, but for now, we should revert so that we can figure out if it was this change that is causing these errors.

@snopoke @sheelio

### DIFF
--- a/corehq/apps/fixtures/fixturegenerators.py
+++ b/corehq/apps/fixtures/fixturegenerators.py
@@ -1,12 +1,8 @@
 from collections import defaultdict
 from xml.etree import ElementTree
-from StringIO import StringIO
-
-from corehq.blobs import get_blob_db
-from corehq.blobs.exceptions import NotFound
 
 from casexml.apps.phone.fixtures import FixtureProvider
-from corehq.apps.fixtures.models import FixtureDataItem, FixtureDataType, FIXTURE_BUCKET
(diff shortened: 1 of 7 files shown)


millerdev:
Was it intentional that you did not revert https://github.com/dimagi/commcare-hq/pull/16904/commits/869e2d8eac5646744e77a1a64d67a23d36118257 of #16904 ?

snopoke:
Odd, feels like we could benefit from Datadog's Tracing features: https://app.datadoghq.com/apm/intro

proteusvacuum:
https://github.com/dimagi/commcare-hq/pull/16973

proteusvacuum:
@snopoke , yeah that might be useful. Although I think it will be hard to know what is happening unless we see what payload is getting sent down in the restore somehow. Directly running the restore with the URL the phone accesses gives me the right fixtures, so I'm not sure if there is some non-deterministic behaviour with fetching the cache from Riak, or somewhere else.

PR #16939, opened 2017-07-07
Fix bugs in SQL db accessors

millerdev:
- Eliminate RawQuerySetWrapper as [recommended by @snopoke](https://github.com/dimagi/commcare-hq/pull/16925#discussion_r126094790)
- Fix buggy `_order_list`

Review commits individually 🐠 and definitely wait for tests on this one.

@snopoke

PR #16976, opened 2017-07-11
Revert #16809

millerdev:
Revert #16809 to eliminate unnecessary work done by `CaseSyncUpdate` when using livequery. I believe most of the optimizations in #16809 were made irrelevant by #16939. One exception is prematurely generating XML for updates that are later excluded from the sync. We could add that back in if it seems important.

@proteusvacuum


proteusvacuum:
Might be worth not generating the XML until we know the cases are pertinent, but I think since you already pull far fewer cases, it should fine for now.

PR #16977, opened 2017-07-11
adding management command to mark batches complete

calellowitz:
@gcapalbo 
This was one of the followup items on Ben's original batch pr. I think this is all thats needed but let me know if you had something bigger in mind

cc: @benrudolph


corehq/warehouse/management/commands/mark_batch_complete.py
```diff
@@ -0,0 +1,25 @@
...
+
+from django.core.management import BaseCommand
+from django.core.management import CommandError
+
+from corehq.warehouse.models import Batch
+
+
+class Command(BaseCommand):
+    """
+    Example: ./manage.py create_batch 222617b9-8cf0-40a2-8462-7f872e1f1344 -s 2017-05-01 -e 2017-06-01
```
gcapalbo: was this just accidentally copied from the create_batch command?

corehq/warehouse/management/commands/mark_batch_complete.py
```diff
@@ -0,0 +1,25 @@
...
+    def add_arguments(self, parser):
+        parser.add_argument('batch_id')
+
+    def handle(self, batch_id, **options):
+        try:
+            batch = Batch.objects.get(batch_id=batch_id)
+        except Batch.DoesNotExist:
+            raise CommandError('Invalid batch ID: {}'.format(batch_id))
+
+        batch.completed_on = datetime.utcnow()
```
gcapalbo: might be good to also check if completed_on already has a value, and also raise an error if that happens

gcapalbo (approved):
looks like there's a conflicting migration making the build fail

PR #16978, opened 2017-07-11
ICDS: Optimize person cases table

sheelio:
@emord 

This should help speed up some of the person case indicators (reducing calls to get_case_forms if the person doesn't have an aadhar ID).  Once the ICDS team deploys the new app version, this should further reduce the calls since they are planning to save the aadhar date to the case as well.


sheelio:
(I'm assuming the conditional expression won't evaluate the expression_if_false section if the test is true).

emord (approved):
Nice. this was really easy to review with w=1

PR #16980, opened 2017-07-11
Abt custom reports: use custom user data for target numbers

NoahCarnahan:
@nickpell I can't believe I didn't make this change sooner. This allows FMs to configure the target numbers in custom user data instead of having it hard coded into the data source. Target numbers remain in the data source for past deployments though.
Best reviewed with `w=1`.


nickpell (approved):
good work, sir, good work

PR #16981, opened 2017-07-11
Add a minimal superuser-only UI to show domain privileges

esoergel:
I've seen a couple issues recently where it was difficult to determine which
privileges were available on the domain, and I ended up needing to open a shell
on the server to figure it out.  This should make that easier, and allow support
to do it too.
@nickpell @emord FYI

image


nickpell (approved):
nice!

emord:
Nice. I always hate going through the accounting interface to find this. Do you think it would be worthwil to put feature flags and this on the same page?

esoergel:
Yeah, that could be cool.  Feeling motivated to make that change?  Not sure I'll get to it right away.

emord:
I won't be able to for a while either. Just thought it might be nice to have a comprehensive view of all permissions a domain can have. No action necessary from you

PR #16985, opened 2017-07-12
short circuit sql backed check for ICDS

snopoke:
@dimagi/scale-team FYI


corehq/form_processor/utils/general.py
```diff
@@ -26,7 +26,18 @@ def clear_local_domain_sql_backend_override(domain):
...
+    from corehq.apps.domain.models import Domain
+    domain_name = domain_object_or_name
+    if isinstance(domain_object_or_name, Domain):
+        domain_name = domain_object_or_name.name
+
+    return domain_name == 'icds-cas'
+
+
 def should_use_sql_backend(domain_object_or_name):
+    if _short_circuit_sql_check(domain_object_or_name):
```
emord: why didn't you use `ENTERPRISE_OPTIMIZATIONS`? https://github.com/dimagi/commcare-hq/blob/master/corehq/toggles.py#L1248

snopoke: updated: 11821be

emord (approved):

corehq/form_processor/utils/general.py
```diff
@@ -26,7 +26,18 @@ def clear_local_domain_sql_backend_override(domain):
     _thread_local.use_sql_backend = use_sql_backend_dict
 
 
+def _short_circuit_sql_check(domain_object_or_name):
+    from corehq.apps.domain.models import Domain
+    domain_name = domain_object_or_name
+    if isinstance(domain_object_or_name, Domain):
+        domain_name = domain_object_or_name.name
+
+    return ENTERPRISE_OPTIMIZATIONS.enabled(domain_name)
```
gcapalbo: won't this slightly slow down all domains other than icds though? since it will add one extra hit to the cache

proteusvacuum: would be great to set this up for the `enikshay` domain too, although I'm not certain what the consequences of using this feature flag are on softlayer (the `enikshay` domain is on SQL, and we know that for a fact, so maybe it will be inconsequential).

proteusvacuum (approved):

corehq/form_processor/utils/general.py
snopoke: Unfortunately this can't be set for eNikshay since it's still running on a cluster with non-SQL domains.

To avoid confusion I've completely removed the toggle and just use the global setting

proteusvacuum: can we add a special case for eNikshay here then? 
The case search pillow hits this method for every document it processes.

snopoke:
@proteusvacuum @emord updated to include check for enikshay

emord (approved):
Thanks @snopoke is this something that should get deployed today?

snopoke:
@emord I don't think it's urgent for ICDS, not sure about eNikshay

proteusvacuum:
thanks. not super urgent.

gcapalbo:
@snopoke did you push those two commits here by mistake? looks like they belong to other PR

snopoke:
yes, sorry, removed them now

snopoke:
@proteusvacuum can you take a look a the test failure. I took a quick look but didn't see anything obvious:
```
FAIL: custom.enikshay.ucr.tests.test_episode_datasource:TestEpisodeDatasourceES.test_diagnostic_test_after_end_of_ip
  File "/mnt/commcare-hq-ro/custom/enikshay/ucr/tests/test_episode_datasource.py", line 209, in test_diagnostic_test_after_end_of_ip
    self.assertEqual(row.diagnostic_test_after_end_of_ip, 1)
AssertionError: 0 != 1
root: DEBUG: Beginning atomic commit
root: DEBUG: Saving new form: XFormInstance(form_id='d4ec5131d23c48f88037daa199ef5e4d', domain='enikshay')
```

snopoke:
@proteusvacuum bump

proteusvacuum:
Hey,

Sorry, took a quick look last week but didn't see anything obvious either,
but did figure out that for some reason they weren't working at all with
SQL backends. That test isn't really that useful anymore, as I wrote it so
that I wouldn't have to create a bunch of test data to test the datasource
with manually. We could remove it, or I can spend some more time digging
into why it isn't working for SQL backends.

-F

snopoke:
:white_check_mark:

benrudolph:
this does mean that enikshay can't take advantage of the optimizations now (not sure if that matters or not)

snopoke:
@benrudolph this will still short circuit for enikshay

benrudolph:
(offlined, thanks, looks good)

PR #16986, opened 2017-07-12
 remove duplicate form attachment download view

snopoke:
Also add form attachment URL to API

buddy @proteusvacuum


corehq/apps/api/object_fetch_api.py
```diff
@@ -119,17 +120,24 @@ def get(self, request, domain, case_id=None, attachment_id=None):
 class FormAttachmentAPI(View):
 
     @method_decorator(login_or_digest_or_basic_or_apikey())
+    @method_decorator(require_form_view_permission)
+    @location_safe
```
esoergel: This decorator should go on the class, not the `get` method.

esoergel (changes requested):
One comment, looks good otherwise.

corehq/apps/api/object_fetch_api.py
snopoke: thanks, updated: 739a62e

esoergel (approved):

PR #16991, opened 2017-07-12
Update README.md

idiene:
Changed the transform parameter custom_format to format string for rounding expressions.


benrudolph:
🔥 @idiene first PR

PR #16992, opened 2017-07-12
session bypassing for selected URLs

snopoke:
@dimagi/scale-team 

This PR also changes the session store to just be `cache` instead of `cache + DB`. There's no migration so deploying as is will effectively wipe everyone's session.

Having switched to cache I'm not sure it's really necessary to bypass at all. Also a bit concerned this might break web apps (@benrudolph don't they use session auth?)

### DIFF
--- a/corehq/middleware.py
+++ b/corehq/middleware.py
@@ -3,7 +3,12 @@ import logging
 import mimetypes
 import os
 import datetime
+
+import re
 from django.conf import settings
+from django.contrib.auth.middleware import AuthenticationMiddleware
+from django.contrib.messages.middleware import MessageMiddleware
+from django.contrib.sessions.middleware import SessionMiddleware
 from django.core.exceptions import MiddlewareNotUsed
 from django.http import HttpResponseRedirect
 from django.urls import reverse
@@ -11,6 +16,7 @@ from django.contrib.auth.views import logout as django_logout
 from django.utils.deprecation import MiddlewareMixin
 
 from corehq.apps.domain.models import Domain
+from corehq.apps.domain.utils import legacy_domain_re
 from corehq.const import OPENROSA_DEFAULT_VERSION
 
 from dimagi.utils.parsing import json_format_datetime, string_to_utc_datetime
@@ -112,6 +118,9 @@ class TimeoutMiddleware(MiddlewareMixin):
         if not request.user.is_authenticated:
             return
 
+        if not hasattr(request, 'session'):
+            return
+
         secure_session = request.session.get('secure_session')
         domain = getattr(request, "domain", None)
         now = datetime.datetime.utcnow()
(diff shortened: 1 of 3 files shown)


settings.py
```diff
@@ -159,6 +159,16 @@
...
 SESSION_ENGINE = "django.contrib.sessions.backends.cache"
 
+SESSION_BYPASS_URLS = [
+    r'^/a/{domain}/receiver/',
+    r'^/a/{domain}/phone/restore/',
+    r'^/a/{domain}/phone/search/',
+    r'^/a/{domain}/phone/claim-case/',
+    r'^/a/{domain}/phone/heartbeat/',
+    r'^/a/{domain}/phone/keys/',
+    r'^/a/{domain}/phone/admin_keys/',
```
emord: why not just use a/domian/phone? Is there some enpdoint that uses sessions?

corehq/middleware.py
```diff
@@ -195,3 +204,45 @@ def process_view(self, request, view_func, view_args, view_kwargs):
...
+    def __init__(self, get_response=None):
+        super(SessionBypassMiddleware, self).__init__(get_response)
+        regexes = getattr(settings, 'SESSION_BYPASS_URLS', [])
+        self.bypass_re = [
+            re.compile(regex.format(domain=legacy_domain_re)) for regex in regexes
+        ]
+
+    def process_request(self, request):
+        if self._bypass_sessions(request):
+            print request.path_info
```
emord: print

corehq/middleware.py
```diff
@@ -195,3 +204,45 @@ def process_view(self, request, view_func, view_args, view_kwargs):
...
+            return
+
+        super(SessionBypassMiddleware, self).process_request(request)
+
+    def process_response(self, request, response):
+        if self._bypass_sessions(request):
+            print request.path_info
```
emord: print

emord (commented):
looks good. just a few nits. This should be (probably quickly) QA'd on staging right? since it would directly effect submissions and restores

emord:
nevermind my comment about QA. i opened up this tab before I saw the open for review tag

settings.py
snopoke: yea there's `/a/domain/admin_restore`

snopoke:
rebased off master. Going to put this on staging

emord:
We should also verify that this works with the proxy caching

benrudolph:
does this mean we no longer write to `django_sessions` on those URLs? I think that's OK as long as we have a somewhat up to date `django_sessions` table. this is what we use it for: https://github.com/dimagi/formplayer/blob/da471f28e1eab667e929c839b74f7ebfacd985e6/src/main/java/repo/impl/PostgresTokenRepo.java#L31

snopoke:
Going to close this out. I don't think it's really worth it for the small set of URLs we can do it on.

snopoke:
For future reference: https://stackoverflow.com/questions/4052473/disable-session-creation-in-django

emord (commented):
damn it. I didn't post my comment yesterday :/

corehq/middleware.py
```diff
@@ -195,3 +204,43 @@ def process_view(self, request, view_func, view_args, view_kwargs):
...
+        ]
+
+    def process_request(self, request):
+        if self._bypass_sessions(request):
+            return
+
+        super(SessionBypassMiddleware, self).process_request(request)
+
+    def process_response(self, request, response):
+        if self._bypass_sessions(request):
```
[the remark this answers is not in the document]
snopoke: Thanks @emord. Option 1 is similar to something on the SO post where you set the save function to a noop. 
```
request.session.save = lambda: None
```

I think doing something like this is likely the best option since it still allows access to the session object but prevents the hit on the cache. The one thing I don't like is that if the session IS being used somewhere we wouldn't know about it. 

The other thing I was thinking of was wrapping the session in a proxy object that logs access so that we can see where it's being used. Doing that would allow us to see if it would be safe to not set the session on the request at all.

PR #16997, opened 2017-07-12
add SKIP_REMOVE_INDICES

nickpell:
Alternative to revoking celery tasks, without affecting other domains.

@proteusvacuum @emord @calellowitz @ctsims


calellowitz:
might be nice to a test but otherwise 👍

PR #17004, opened 2017-07-13
Revert "default export date filter to ALL data"

snopoke:
Reverts dimagi/commcare-hq#16755

Seems like it could be causing issues: 
https://manage.dimagi.com/default.asp?258031
https://manage.dimagi.com/default.asp?257701

### DIFF
--- a/corehq/apps/export/forms.py
+++ b/corehq/apps/export/forms.py
@@ -102,13 +102,8 @@ class DateSpanField(forms.CharField):
     def clean(self, value):
         date_range = super(DateSpanField, self).clean(value)
         dates = date_range.split(DateRangePickerWidget.separator)
-        if len(dates) != 2:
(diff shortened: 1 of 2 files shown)


dannyroberts:
Worth a shot. I'll look after deploy and let you know if this was the culprit.

dannyroberts:
Reverting fixed it. Is it possible that `dates` (from '`if len(dates) != 2:`') actually had >2 elements? If it had <2 elements, I believe it would be failing with an `IndexError` now.

snopoke:
I guess it could be possible but I don't think so. The reason it doesn't index error now is that the default option is a valid date range. Part of this PR was setting the default option to 'Export all data' which isn't a valid date range. I'm not going to have time to look into this for a while and also don't think it's very important.

PR #17005, opened 2017-07-13
Add Repeat Records on case data page

proteusvacuum:
Adds a new tab to the case data page if there are forwarded records for a particular case. 

@snopoke 
FYI @dannyroberts @mkangia 

screenshot from 2017-07-13 09-16-55


snopoke (approved):
:candy:

mkangia:
Thank You @proteusvacuum. That reindex would be surely heavy but this would be extremely useful..

PR #17010, opened 2017-07-13
Fixed Enikshay date filter

kkrampa:
@orangejenny I believe you made some changes that make DatespanMixin required in order to have date filter working properly.

Could you merge that and deploy to India server? 
@proteusvacuum @emord

### DIFF
--- a/custom/enikshay/reports/generic.py
+++ b/custom/enikshay/reports/generic.py
@@ -48,7 +48,7 @@ class MultiReport(CustomProjectReport, GenericReportView):
         return context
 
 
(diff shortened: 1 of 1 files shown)


orangejenny (approved):

orangejenny:
@kkrampa Do you know more specifically what the issue here was? If I caused it, it's probably something in https://github.com/dimagi/commcare-hq/pull/16903 but reading through that I can't tell what the problem is.

orangejenny:
@dannyroberts is going to merge this as part of today's softlayer maintenance

PR #17011, opened 2017-07-13
Make test data sources compatible with new version of Enikshay App

kkrampa:
@proteusvacuum @emord Notifying you because @NoahCarnahan is offline.

We are trying to make our reports compatible with new version of Enikshay app. I made copies of all reports that can be affected by those changes and we want to test this on test domains.


proteusvacuum:
Hey @kkrampa I've added the hold off on merge flag for now since we are going to wait until the UCR queues are clearer so that we can run the rebuild.

kkrampa:
Those domains have only a couple of cases so it shouldn't be a problem. Manual rebuild takes about 1 minute.

proteusvacuum:
Oh, I see, these are in a separate folder. Good to merge when tests complete.

kkrampa:
@proteusvacuum :white_check_mark:

PR #17015, opened 2017-07-13
fixing longstanding ucr bugs

calellowitz:
@emord @sheelio 
fixes 2 existing bugs in icds UCRs. display was incorrectly using server date instead of phone date, and pse_daily_attendance was supposed to be a count


sheelio:
@calellowitz 
Looks good to me.

emord:
tests passed, but it had issues connecting to datadog

PR #17016, opened 2017-07-13
DHIS2: Management Command to allow datasets for previous dates

sheelio:
@kaapstorm 

This was a quick fix I made with a management command (you have to do one period at a time).

### DIFF
--- a/corehq/motech/dhis2/tasks.py
+++ b/corehq/motech/dhis2/tasks.py
@@ -14,7 +14,7 @@ from toggle.shortcuts import find_domains_with_toggle_enabled
 
 
 @task(queue='background_queue')
-def send_datasets(domain_name, send_now=False):
+def send_datasets(domain_name, send_now=False, send_date=datetime.today()):
     """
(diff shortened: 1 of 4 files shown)


corehq/motech/dhis2/tasks.py
```diff
@@ -14,7 +14,7 @@
 
 
 @task(queue='background_queue')
-def send_datasets(domain_name, send_now=False):
+def send_datasets(domain_name, send_now=False, send_date=datetime.today()):
```
kaapstorm: The default value `datetime.today()` will be evaluated once when the function is defined, not every time it's called, and this could have [unexpected consequences][1].  Better to change this to `send_date=None` and then inside the function add
```python
if send_date is None:
    send_date = datetime.today()
```

  [1]: https://pythonconquerstheuniverse.wordpress.com/2012/02/15/mutable-default-arguments/

kaapstorm (commented):
This is great! Just the one comment.

PR #17023, opened 2017-07-14
First check whether user can access all locations

kaapstorm:
First check whether the user can access all locations, then check whether they can access any locations.

Context: [FB 256351](https://manage.dimagi.com/default.asp?256351)

@sravfeyn cc @gcapalbo


sravfeyn (approved):
Thanks!

PR #17024, opened 2017-07-14
get mobile_ucr_sync_interval from latest app version

snopoke:
@dimagi/scale-team

### DIFF
--- a/corehq/apps/app_manager/dbaccessors.py
+++ b/corehq/apps/app_manager/dbaccessors.py
@@ -174,6 +174,17 @@ def get_brief_apps_in_domain(domain, include_remote=True):
     return apps
 
 
+def get_brief_app(domain, app_id):
+    from .models import Application
+    from corehq.apps.app_manager.util import get_correct_app_class
+    result = Application.get_db().view(
+        'app_manager/applications_brief',
+        key=[domain, app_id],
+    ).one(except_all=True)
+    doc = result['value']
+    return get_correct_app_class(doc).wrap(doc)
+
+
 def get_app_ids_in_domain(domain):
     from .models import Application
     return [row['id'] for row in Application.get_db().view(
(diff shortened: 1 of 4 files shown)


corehq/apps/app_manager/dbaccessors.py
```diff
@@ -174,6 +174,17 @@ def get_brief_apps_in_domain(domain, include_remote=True):
...
 
+def get_brief_app(domain, app_id):
+    from .models import Application
+    from corehq.apps.app_manager.util import get_correct_app_class
+    result = Application.get_db().view(
+        'app_manager/applications_brief',
+        key=[domain, app_id],
+    ).one(except_all=True)
+    doc = result['value']
+    return get_correct_app_class(doc).wrap(doc)
```
gcapalbo: wrapping an app appears to sometimes save it due to some lazy migrations: https://github.com/dimagi/commcare-hq/blob/master/corehq/apps/app_manager/models.py#L4910

is this going to be a problem if this is the brief version of the app?

snopoke: this won't be an issue with the current code since the brief doc doesn't contain any of the fields referenced in the lazy migrations. It could be an issue in future if new lazy migration are added. 

Note that we already do this in the other brief doc views elsewhere in this file.

gcapalbo: ok, this may be fine for now but still seems a little scary. added an item for this here: https://trello.com/c/tQw9NKFg/695-remove-lazy-migrations

gcapalbo (approved):

corehq/apps/app_manager/dbaccessors.py
snopoke: thanks

dannyroberts:
Oof, so it looks like we're at the point where these app_manager migrations now take about an hour to finish—just a heads up that we should start kicking off a preindex as soon as they're merged now

PR #17025, opened 2017-07-14
track exception

sravfeyn:


sravfeyn:
Pretty sure this is the reason why index files are missing from ccz, but since the error is not displayed anywhere we don't know the reason for it

emord (approved):

PR #17027, opened 2017-07-14
Add indexes

emord:
From pgbadger on prod. Looks like someone added indexes to the models file on ilsgateway without actually creating the migration for them :(

Should review the size of these tables beofre deploying so it doesn't block writes for long

@esoergel @dimagi/scale-team


emord:
| table | size |
| --------------- | ------------------------------ |
| locations_sqllocation | 72 MB |
| products_sqlproduct | 2768 kB |
| ilsgateway_alert | 487 MB |
| ilsgateway_organizationsummary | 102 MB |
| ilsgateway_productavailabilitydata | 3462 MB |
| ilsgateway_supplypointstatus | 282 MB |

snopoke (approved):

corehq/apps/locations/models.py
```diff
@@ -353,7 +353,7 @@ class SQLLocation(MPTTModel):
     metadata = jsonfield.JSONField(default=dict, blank=True)
     created_at = models.DateTimeField(auto_now_add=True)
     last_modified = models.DateTimeField(auto_now=True)
-    is_archived = models.BooleanField(default=False)
+    is_archived = models.BooleanField(default=False, db_index=True)
```
esoergel: :+1:

custom/ilsgateway/models.py
```diff
@@ -136,7 +136,7 @@ class SupplyPointStatus(models.Model):
                                    max_length=50)
     status_value = models.CharField(max_length=50,
                                     choices=((c, c) for c in SupplyPointStatusValues.CHOICES))
-    status_date = models.DateTimeField(default=datetime.utcnow)
+    status_date = models.DateTimeField(default=datetime.utcnow, db_index=True)
     location_id = models.CharField(max_length=100, db_index=True)
```
esoergel: @emord is it worth it to take this opportunity to also add a FK reference to `sql_location` here for https://github.com/dimagi/commcare-hq/blob/master/custom/ilsgateway/tanzania/reminders/reports.py#L36-L42 ?

esoergel (approved):

dannyroberts (approved):
nice nice nice

custom/ilsgateway/models.py
gcapalbo: fwiw i don't think an integer foreign key would help that query. queries that have long lists in an IN clause are going to be slow and should be avoided. the alternative approach is to construct a temporary table with a single indexed column, insert the ids into the temp table, and join the temp table to the main result set.

corehq/apps/locations/models.py
gcapalbo: wouldn't this only help if is_archived is the only column we're querying (which i don't think we do often)?

maybe is_archived should be added in a multicolumn index with the other columns (like maybe domain) depending on our use of it

emord:
```
SELECT
    "commtrack_stockstate"."id",
    "commtrack_stockstate"."section_id",
    "commtrack_stockstate"."case_id",
    "commtrack_stockstate"."product_id",
    "commtrack_stockstate"."stock_on_hand",
    "commtrack_stockstate"."daily_consumption",
    "commtrack_stockstate"."last_modified_date",
    "commtrack_stockstate"."last_modified_form_id",
    "commtrack_stockstate"."sql_product_id",
    "commtrack_stockstate"."sql_location_id"
FROM
    "commtrack_stockstate"
    INNER JOIN "products_sqlproduct" ON ("commtrack_stockstate"."sql_product_id" = "products_sqlproduct"."id")
        LEFT OUTER JOIN "locations_sqllocation" ON ("commtrack_stockstate"."sql_location_id" = "locations_sqllocation"."id")
    WHERE (NOT ("products_sqlproduct"."is_archived" = TRUE)
        AND NOT ("locations_sqllocation"."is_archived" = TRUE
            AND "locations_sqllocation"."is_archived" IS NOT NULL)
        AND "commtrack_stockstate"."case_id" IN (...));
```

I'm not sure if this will help, but it looked like it could based on this query. i'm sure there's a better way to write this but, I don't have any idea where this comes from at the moment. it's just hte output from slowlogs in pgbadger.

gcapalbo: yeah i saw that query you had pasted into slack. the biggest factor impacting performance on that is the enormous IN() clause - if we rewrite that to use a temp table with the case ids and join to it, it would perform much better.

i think even just adding domain to the WHERE clause might make a difference there too, since that would cut the result set down by a lot so we don't have to check the IN on the stockstate for practically all locations

esoergel: Spun up a little doc about this query: https://docs.google.com/document/d/1wwrxVa3dXJFhMwvm-ehUbyCIUgFI_1XRdyVjlCLFLX0/edit

emord:
I removed the changes to products and locations, so this is now only on ilsgateway. I plan on adding stat gathering to postgres so that we can have better data about indexes to make these decisions

gcapalbo (approved):

emord:
Just applied to prod in a setup_release directory

PR #17030, opened 2017-07-14
Track sync times by domain

millerdev:
Also do not track so many segments of the restore since most of them are not very useful and make it hard to understand the histogram data in datadog.

@snopoke @benrudolph

### DIFF
--- a/corehq/apps/ota/views.py
+++ b/corehq/apps/ota/views.py
@@ -1,5 +1,6 @@
 from distutils.version import LooseVersion
 
+from django.conf import settings
 from django.http import JsonResponse
 from django.urls import reverse
 from django.shortcuts import redirect
@@ -27,7 +28,6 @@ from corehq.apps.domain.decorators import (
     check_domain_migration,
     login_or_digest_or_basic_or_apikey_or_token,
 )
-from corehq.util.datadog.gauges import datadog_counter, datadog_histogram
 from corehq.apps.domain.models import Domain
 from corehq.apps.domain.views import DomainViewMixin, EditMyProjectSettingsView
 from corehq.apps.es.case_search import flatten_result
@@ -37,6 +37,7 @@ from corehq.apps.ota.tasks import queue_prime_restore
 from corehq.apps.users.models import CommCareUser, CouchUser
 from corehq.apps.locations.permissions import location_safe
 from corehq.form_processor.exceptions import CaseNotFound
+from corehq.util.datadog.gauges import datadog_counter, datadog_histogram
 from dimagi.utils.decorators.memoized import memoized
 from casexml.apps.phone.restore import RestoreConfig, RestoreParams, RestoreCacheSettings
 from django.http import HttpResponse
@@ -45,6 +46,12 @@ from soil import MultipleTaskDownload
 from .utils import (
     demo_user_restore_response, get_restore_user, is_permitted_to_restore,
     handle_401_response, update_device_id)
+
+
+TIMED_RESTORE_SEGMENTS = {
+    "FixtureElementProvider": "fixtures",
+    "CasePayloadProvider": "cases",
+}
 
 
 @location_safe
@@ -61,15 +68,23 @@ def restore(request, domain, app_id=None):
     tags = [
         u'status_code:{}'.format(response.status_code),
     ]
+    env = settings.SERVER_ENVIRONMENT
+    if (env, domain) in settings.RESTORE_TIMING_DOMAINS:
+        tags.append(u'domain:{}'.format(domain))
     datadog_counter('commcare.restores.count', tags=tags)
     if timing_context is not None:
+        datadog_histogram(
+            'commcare.restores.total_time',
+            timing_context.duration,
+            tags=tags,
+        )
         for timer in timing_context.to_list(exclude_root=True):
-            # Only record leaf nodes so we can sum to get the total
-            if timer.is_leaf_node:
+            if timer.name in TIMED_RESTORE_SEGMENTS:
+                segment = TIMED_RESTORE_SEGMENTS[timer.name]
                 datadog_histogram(
                     'commcare.restores.timings',
                     timer.duration,
-                    tags=tags + [u'segment:{}'.format(timer.name)],
+                    tags=tags + [u'segment:{}'.format(segment)],
                 )
 
     return response
--- a/corehq/util/timer.py
+++ b/corehq/util/timer.py
@@ -30,6 +30,7 @@ class NestableTimer(object):
 
     @property
     def duration(self):
+        """Get timer duration in seconds"""
         if self.beginning and self.end:
             return self.end - self.beginning
 
@@ -128,7 +129,7 @@ class TimingContext(object):
 
     @property
     def duration(self):
-        return self.to_dict()['duration']
+        return self.root.duration
 
     def to_list(self, exclude_root=False):
         """Get the list of ``NestableTimer`` objects in hierarchy order"""
--- a/settings.py
+++ b/settings.py
@@ -2142,6 +2142,12 @@ DOMAIN_MODULE_MAP = {
 
 CASEXML_FORCE_DOMAIN_CHECK = True
 
+RESTORE_TIMING_DOMAINS = {
+    # ("env", "domain"),
+    ("production", "rec"),
+    ("softlayer", "enikshay"),
+}
+
 #### Django Compressor Stuff after localsettings overrides ####
 
 # This makes sure that Django Compressor does not run at all


corehq/apps/ota/views.py
```diff
@@ -61,15 +68,23 @@ def restore(request, domain, app_id=None):
...
+    env = settings.SERVER_ENVIRONMENT
+    if (env, domain) in settings.RESTORE_TIMING_DOMAINS:
+        tags.append(u'domain:{}'.format(domain))
     datadog_counter('commcare.restores.count', tags=tags)
     if timing_context is not None:
+        datadog_histogram(
+            'commcare.restores.total_time',
+            timing_context.duration,
+            tags=tags,
+        )
```
millerdev: Track total restore time in addition to segments below. This should make it easier to aggregate and graph the histogram data in datadog. Summing segments was giving me unintelligible results.

corehq/apps/ota/views.py
```diff
@@ -61,15 +68,23 @@ def restore(request, domain, app_id=None):
...
         for timer in timing_context.to_list(exclude_root=True):
-            # Only record leaf nodes so we can sum to get the total
-            if timer.is_leaf_node:
+            if timer.name in TIMED_RESTORE_SEGMENTS:
+                segment = TIMED_RESTORE_SEGMENTS[timer.name]
                 datadog_histogram(
                     'commcare.restores.timings',
                     timer.duration,
-                    tags=tags + [u'segment:{}'.format(timer.name)],
+                    tags=tags + [u'segment:{}'.format(segment)],
```
millerdev: Another idea I had was to add the segment to the metric name (e.g., `commcare.restore.timings.cases`) and remove the segment tag. This might make it easier to do some things in datadog. However, I'm not sure if that would make it impossible to make a stacked graph of all segments?

corehq/apps/ota/views.py
```diff
@@ -61,15 +68,23 @@ def restore(request, domain, app_id=None):
...
+            'commcare.restores.total_time',
+            timing_context.duration,
+            tags=tags,
+        )
         for timer in timing_context.to_list(exclude_root=True):
-            # Only record leaf nodes so we can sum to get the total
-            if timer.is_leaf_node:
+            if timer.name in TIMED_RESTORE_SEGMENTS:
+                segment = TIMED_RESTORE_SEGMENTS[timer.name]
                 datadog_histogram(
```
millerdev: Already discussed this with @snopoke 

After looking at the existing data in datadog for a long time, I'm beginning to wonder if the histogram metric type is very useful here, given the aggregation interval is so short (10s, see [What is the histogram metric type?](https://help.datadoghq.com/hc/en-us/articles/205638045-What-is-the-histogram-metric-type-)). Quote from [datadog metric types docs](https://help.datadoghq.com/hc/en-us/articles/206955236-Metric-types-in-Datadog): `histogram( ... )` "Should be called multiple times during an agent check (otherwise you have no distribution)."

Are we sending enough data to get a meaningful distribution? The avg, 95% and max stats are closer together than I would expect. Maybe the many segments previously tracked were causing a problem?

millerdev:
I'm not sure the build failure is related to this PR?
```
FAIL: corehq.apps.app_manager.tests.test_real_build:TestRealBuild.test_real_build
  File "/mnt/commcare-hq-ro/corehq/apps/app_manager/tests/test_real_build.py", line 41, in test_real_build
    self.fetch_and_build_app('commcare-tests', 'ae3c6e073262360f89d2630cfd220bd3')
  File "/mnt/commcare-hq-ro/corehq/apps/app_manager/tests/test_real_build.py", line 37, in fetch_and_build_app
    app.create_all_files()
  File "/mnt/commcare-hq-ro/corehq/apps/app_manager/models.py", line 5849, in create_all_files
    practice_user_restore = self.create_practice_user_restore(build_profile_id)
  File "/mnt/commcare-hq-ro/corehq/apps/app_manager/models.py", line 5810, in create_practice_user_restore
    if not self.enable_practice_users:
  File "submodules/dimagi-utils-src/dimagi/utils/decorators/memoized.py", line 10, in _inner
    return m(*args, **kwargs)
  File "submodules/dimagi-utils-src/dimagi/utils/decorators/memoized.py", line 127, in __call__
  File "/mnt/commcare-hq-ro/corehq/apps/app_manager/models.py", line 5779, in enable_practice_users
    domain_has_privilege(self.domain, privileges.PRACTICE_MOBILE_WORKERS)
  File "/vendor/lib/python2.7/site-packages/quickcache/quickcache.py", line 40, in inner
  File "/vendor/lib/python2.7/site-packages/quickcache/quickcache_helper.py", line 145, in __call__
    return self.call(*args, **kwargs)
  File "/vendor/lib/python2.7/site-packages/quickcache/quickcache_helper.py", line 60, in call
    content = self.fn(*args, **kwargs)
  File "/mnt/commcare-hq-ro/corehq/apps/accounting/utils.py", line 122, in domain_has_privilege
    plan_version = Subscription.get_subscribed_plan_by_domain(domain)
  File "/mnt/commcare-hq-ro/corehq/apps/accounting/models.py", line 1578, in get_subscribed_plan_by_domain
    domain_obj = ensure_domain_instance(domain)
  File "/mnt/commcare-hq-ro/corehq/apps/accounting/utils.py", line 44, in ensure_domain_instance
    domain = Domain.get_by_name(domain)
  File "/vendor/lib/python2.7/site-packages/quickcache/quickcache.py", line 40, in inner
  File "/vendor/lib/python2.7/site-packages/quickcache/quickcache_helper.py", line 145, in __call__
    return self.call(*args, **kwargs)
  File "/vendor/lib/python2.7/site-packages/quickcache/quickcache_helper.py", line 60, in call
    content = self.fn(*args, **kwargs)
  File "/mnt/commcare-hq-ro/corehq/apps/domain/models.py", line 568, in get_by_name
    _assert(False, '%r is not a valid domain name' % name)
  File "/mnt/commcare-hq-ro/corehq/util/soft_assert/core.py", line 34, in __call__
    return self._call(assertion, msg, obj)
  File "/mnt/commcare-hq-ro/corehq/util/soft_assert/core.py", line 48, in _call
    raise AssertionError(msg)
AssertionError: None is not a valid domain name
```

corehq/apps/ota/views.py
snopoke: You find still do a stacked graph. It would just be more manual. More that we have a total metric as well it feels like the segments are fine as tags.

corehq/apps/ota/views.py
snopoke: I've never really looked at this data on prod. On ICDS we definitely have enough traffic for histogram. We can't use gauge though since that only reports the last metric in a sync interval.

snopoke (approved):

emord:
@millerdev restarted that job on travis. i fixed the build in another PR

corehq/apps/ota/views.py
millerdev: > On ICDS we definitely have enough traffic for histogram.

I'm not sure I agree (given a 10s aggregation window). The graph below is a roughly one-day window of ICDS data from the past week, and includes the busiest time of the day. Except under the very highest spikes, look how closely the average and max times track eachother, and I see almost no difference between 95th percentile and max. This tells me our aggregation window is too small, but I'm not sure we can change that without globally affecting other datadog stats? Maybe ICDS sync times are fairly uniform across all devices?

https://app.datadoghq.com/dash/321184/restore-timing?live=false&page=0&is_auto=false&from_ts=1499814729885&to_ts=1499888539721&tile_size=m&tpl_var_env=icds&fullscreen=220044309

Compare that to the last day of latency data coming in from Riak on prod. This shows latency for both GET and PUT requests. In this case the average is a mostly flat line, 95th percentile is a slightly more jagged line above that and max is most jagged of all. Note that this data is not recorded with datadog's histogram; it's generated internally in riak, but I would expect the general trends to be similar.

https://app.datadoghq.com/dash/175518/riak-cs-system-resources?live=true&page=0&is_auto=false&from_ts=1499971874621&to_ts=1500058274621&tile_size=s&tpl_var_hosts=group%3Ariakcs_new&fullscreen=175562724&priority=all&abstraction_level=1&aggregate_up=false&per_page=30&use_date_happened=true&display_timeline=false&tpl_var_env=production

snopoke:
During normal operation hours ICDS restore rates are between 0.5/s and 3/s
at peak. Are you saying that that's not enough data to get a reasonable
histogram?

What are the request rates for Riak?

millerdev:
So it sounds like for ICDS the distribution is calculated on somewhere between 5 and 30 data points, given datadog uses a 10s aggregation window. That doesn't seem like a very big sample size; probably not big enough to give very meaningful results for 95th percentile vs max.

Looks like Riak's request rate is somewhere **3-12 GET/s** and **2-4 PUT/s** on prod. The [documentation says it uses a 1-minute window](http://docs.basho.com/riak/kv/2.2.3/using/cluster-operations/inspecting-node/#fsm-time) for distribution calculations.

I wonder if there's a way to calculate distributions in datadog on arbitrary time ranges using data in datadog? It seems like there should be a way to do that if we collected the right data. Can we use existing data, or do we need to change what we're collecting? My knowledge of statics is weak.

millerdev:
Answer to my question above: [At the moment, Datadog cannot compute percentiles on the server side.](https://help.datadoghq.com/hc/en-us/articles/211545826-Why-histogram-stats-are-all-the-same-inaccurate-Characteristics-of-Datadog-histograms-) That article lists a lot of caveats of the datadog histogram type. Notably:

> [Datadog] histograms are great if you want to get an idea of how your **10-second** 95th percentile of your **host** request time evolves over time, but not if you need to:
> 
> 1. get this 95th percentiel over a time period different than 10 sec.
> 2. or get the overall 95th percentile of the request time across hosts.

Adding extra tags also has consequences: histograms are calculated individually for each unique combination of tags, not in aggregate for all tags. So the stats on the recent dashboard I made are averages of averages or sums, not true percentiles.

PR #17037, opened 2017-07-14
Catch all exceptions reconcilliation task

proteusvacuum:
I'm going to do some work to make the error handling a lot better in this task, as previously it was just failing silently in celery and we didn't know. 

Right now, just catch all exceptions and carry on until this exception monitoring work can be completed

FYI @sravfeyn 
@snopoke

### DIFF
--- a/custom/enikshay/tasks.py
+++ b/custom/enikshay/tasks.py
@@ -85,11 +85,12 @@ class EpisodeUpdater(object):
             batch_size = 100
             updates = []
             for episode in self._get_open_episode_cases():
-                adherence_update = EpisodeAdherenceUpdate(episode, self)
-                voucher_update = EpisodeVoucherUpdate(self.domain, episode)
-                test_update = EpisodeTestUpdate(self.domain, episode)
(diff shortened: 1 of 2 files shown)


custom/enikshay/tasks.py
```diff
@@ -48,8 +48,11 @@ def enikshay_task():
     # `toggles.UATBC_ADHERENCE_TASK` enabled
     domains = toggles.UATBC_ADHERENCE_TASK.get_enabled_domains()
     for domain in domains:
-        updater = EpisodeUpdater(domain)
-        updater.run()
+        try:
```
sravfeyn: Thanks for adding this. Ideally, try/except blocks must be present for a task run on each case, so that if a case fails the task proceeds on other cases, so at least when I wrote this, I added those blocks per case-run. Apart from adding this domain-level, if you came across any uncuaght exceptions at case-level they should be included in a try/except as well.,

proteusvacuum: yeah, someone had added a bunch of domains without the right datasource into the feature flag, so this had just stopped running after the first failed domain.

PR #17038, opened 2017-07-15
Update shadow form ids when copying app

orangejenny:
https://manage.dimagi.com/default.asp?256513

Looking at another ticket today, I discovered the magic of `FormIdProperty`.

@NoahCarnahan


NoahCarnahan (commented):
Hey, thanks Jenny. I discovered `FormIdProperty` too when I first looked into this bug. Did you get a chance to test this change? I thought that I had tried just adding `FormIdProperty` where you did, but that copying still didn't go smoothly, I might have had the syntax of the first argument wrong or something though.

orangejenny:
Yes, tested. It did take a couple of tries to get the syntax right; it needs to match the structure of the app json.

NoahCarnahan:
Awesome, thanks!

PR #17039, opened 2017-07-15
Add indexes to custom icds agg

emord:
Just realized that this won't wrok because this is in a migration file. Will add it to another migration later.

@dimagi/scale-team @sheelio

### DIFF
--- a/custom/icds_reports/migrations/sql_templates/create_tables_and_views.sql
+++ b/custom/icds_reports/migrations/sql_templates/create_tables_and_views.sql
@@ -55,6 +55,9 @@ CREATE OR REPLACE VIEW awc_location_months AS
 	months.month_name AS month_display
   FROM awc_location awc_location
   CROSS JOIN "icds_months" months;
+CREATE INDEX "awc_location_months_indx1" ON "awc_location_months" (aggregation_level, month);
+CREATE INDEX "awc_location_months_indx2" ON "awc_location_months" (aggregation_level, state_site_code, month);
+CREATE INDEX "awc_location_months_indx3" ON "awc_location_months" (aggregation_level, state_id, district_id, block_id, supervisor_id, awc_id, month);
 
 -- Table: Geographic boundaries for states, districts and blocks
 DROP TABLE IF EXISTS india_geo_data CASCADE;
@@ -194,6 +197,7 @@ CREATE TABLE agg_awc
   infra_medicine_kits integer,
   infra_adequate_space_pse integer
 );
+CREATE INDEX "agg_awc_indx1" ON "agg_awc" (aggregation_level, state_id, district_id, block_id, supervisor_id, awc_id, month);
 
 -- Table: ccs_record_monthly
 DROP TABLE IF EXISTS ccs_record_monthly CASCADE;


custom/icds_reports/migrations/sql_templates/create_tables_and_views.sql
```diff
@@ -55,6 +55,9 @@ CREATE OR REPLACE VIEW awc_location_months AS
 	months.month_name AS month_display
   FROM awc_location awc_location
   CROSS JOIN "icds_months" months;
+CREATE INDEX "awc_location_months_indx1" ON "awc_location_months" (aggregation_level, month);
+CREATE INDEX "awc_location_months_indx2" ON "awc_location_months" (aggregation_level, state_site_code, month);
+CREATE INDEX "awc_location_months_indx3" ON "awc_location_months" (aggregation_level, state_id, district_id, block_id, supervisor_id, awc_id, month);
```
emord: joined here:

https://github.com/dimagi/commcare-hq/blob/637bccb007c7f79ccb76303d6d396b368e728636/custom/icds_reports/migrations/sql_templates/create_datasource_views.sql#L162-L170

custom/icds_reports/migrations/sql_templates/create_tables_and_views.sql
```diff
@@ -55,6 +55,9 @@ CREATE OR REPLACE VIEW awc_location_months AS
 	months.month_name AS month_display
   FROM awc_location awc_location
   CROSS JOIN "icds_months" months;
+CREATE INDEX "awc_location_months_indx1" ON "awc_location_months" (aggregation_level, month);
```
emord:
```
SELECT
    "agg_child_health_monthly"."state_name",
    sum ("agg_child_health_monthly"."nutrition_status_severely_underweight") AS "severely_underweight",
    sum ("agg_child_health_monthly"."nutrition_status_moderately_underweight") AS "moderately_underweight",
    sum ("agg_child_health_monthly"."wer_eligible") AS "valid",
    sum ("agg_child_health_monthly"."nutrition_status_normal") AS "normal"
FROM
    "agg_child_health_monthly"
WHERE ("agg_child_health_monthly"."aggregation_level" = 0
    AND "agg_child_health_monthly"."month" = ''::date)
GROUP BY
    "agg_child_health_monthly"."state_name";
```

custom/icds_reports/migrations/sql_templates/create_tables_and_views.sql
```diff
@@ -55,6 +55,9 @@ CREATE OR REPLACE VIEW awc_location_months AS
...
+CREATE INDEX "awc_location_months_indx2" ON "awc_location_months" (aggregation_level, state_site_code, month);
```
emord:
```
SELECT
    "agg_child_health_monthly"."district_name",
    sum ("agg_child_health_monthly"."nutrition_status_severely_underweight") AS "severely_underweight",
    sum ("agg_child_health_monthly"."nutrition_status_moderately_underweight") AS "moderately_underweight",
    sum ("agg_child_health_monthly"."wer_eligible") AS "valid",
    sum ("agg_child_health_monthly"."nutrition_status_normal") AS "normal"
FROM
    "agg_child_health_monthly"
WHERE ("agg_child_health_monthly"."aggregation_level" = 0
    AND "agg_child_health_monthly"."state_site_code" = ''
    AND "agg_child_health_monthly"."month" = ''::date)
GROUP BY
    "agg_child_health_monthly"."district_name";
```

snopoke: if you change the order of these I think you can combine this with the previous one:
```
aggregation_level, month, state_site_code
```

gcapalbo: +1

custom/icds_reports/migrations/sql_templates/create_tables_and_views.sql
gcapalbo: regarding that join in the view - is it necessary to join on the whole location hierarchy? i think awc id should be enough unless are we supporting awcs moving around or something?

emord: It's been a while since I took a deep look at these scripts, but awc_id isn't actually unique on this table.

Aggregation level is an integer that says if the row of data is an aggregate of data for the state, district, block, supervisor or awc.

so the table  could be

| agglevel | state | district | block | sup | awc | month | # forms |
| --- | -- | -- | -- | -- | -- | -- | -- | 
| 1 | state1 | null | null | null | null | july | # form for the state |
| 5 | state1 | district1 | block1 | sup1 | awc1 | july | # form for the awc |

snopoke:
Seems like an ideal situation for a partial index (or 2)

custom/icds_reports/migrations/sql_templates/create_tables_and_views.sql
gcapalbo: so in sql the result of comparing null to anything is always false, so the result of null = null is false, in which case that view is currently going to be excluding data for all aggregation levels except for the awc level when all the location ids are there.

if that's the way the data is set up think that view should probably be split into separate views for state level aggregation, district level aggregation, etc. where we'd join only on aggregation level and state id, or aggregation level and district id, etc.

emord: I was incorrect. It's an empty string `''`, not null

emord:
Going to close this PR out. The `state_site_code` query is actually a bug in the v2 dashboards. it should be using state_id.

@sheelio is going to split this table based on aggregation_level

PR #17041, opened 2017-07-16
Ce/ucr sql migration

calellowitz:
@gcapalbo

### DIFF
--- a/custom/icds/management/commands/migrate_pse_attendance.py
+++ b/custom/icds/management/commands/migrate_pse_attendance.py
@@ -24,3 +24,5 @@ class Command(BaseCommand):
         with connections['icds-ucr'].cursor() as cursor:
             print('Migrating pse data')
             cursor.execute(migration_query)
+            print('Vacuuming table')
+            cursor.execute("VACUUM ANALYZE {}".format(new_table))
--- a/custom/icds/management/commands/migrate_pse_attendance.py
+++ b/custom/icds/management/commands/migrate_pse_attendance.py
@@ -22,7 +22,7 @@ INSERT INTO {new_table} (pse_daily_attendance, pse_daily_attendance_female)
 SELECT pse_days_attended, pse_days_attended
 FROM {old_table} A
 LEFT JOIN {new_table} B
-ON A.doc_id = B.doc_id and A.month_start = B.month_start
+ON A.doc_id = B.doc_id and A.month = B.month_start
 WHERE B.sex = 'M'
 """.format(new_table=new_table, old_table=old_table)
 
--- a/custom/icds/management/commands/migrate_pse_attendance.py
+++ b/custom/icds/management/commands/migrate_pse_attendance.py
@@ -18,6 +18,7 @@ FROM {old_table} A
 WHERE A.doc_id = B.doc_id and A.month_start = B.month_start
 """.format(new_table=new_table, old_table=old_table)
 
+
 class Command(BaseCommand):
 
     def handle(self, *args, **options):
--- a/custom/icds/management/commands/migrate_pse_attendance.py
+++ b/custom/icds/management/commands/migrate_pse_attendance.py
@@ -42,4 +42,4 @@ class Command(BaseCommand):
                 print('Migrating pse data')
                 cursor.execute(migration_query)
                 print('Vacuuming table')
-                cursor.execute("VACUUM ANALYZE {}".format(new_table))
+                cursor.execute('VACUUM ANALYZE "{}"'.format(new_table))
--- a/custom/icds/management/commands/migrate_pse_attendance.py
+++ b/custom/icds/management/commands/migrate_pse_attendance.py
@@ -9,21 +9,25 @@ new_table = 'config_report_icds-cas_static-child_cases_monthly_v2_198ccc06'
 old_table = 'config_report_icds-cas_static-child_cases_monthly_tabl_551fd064'
 
 male_sql_query = """
-INSERT INTO {new_table} (pse_daily_attendance, pse_daily_attendance_male)
-SELECT pse_days_attended, pse_days_attended
+UPDATE {new_table}
+SET
+pse_daily_attendance = A.pse_days_attended
+pse_daily_attendance_male = A.pse_days_attended
+FROM {old_table} A
+LEFT JOIN {new_table} B
+ON A.doc_id = B.doc_id and A.month_start = B.month_start
+WHERE B.sex = 'M'
+""".format(new_table=new_table, old_table=old_table)
+
+female_sql_query = """
+UPDATE {new_table}
+SET
+pse_daily_attendance = A.pse_days_attended
+pse_daily_attendance_female = A.pse_days_attended
 FROM {old_table} A
 LEFT JOIN {new_table} B
 ON A.doc_id = B.doc_id and A.month_start = B.month_start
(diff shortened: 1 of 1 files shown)


custom/icds/management/commands/migrate_pse_attendance.py
```diff
@@ -0,0 +1,29 @@
...
+old_table = 'config_report_icds-cas_static-child_cases_monthly_tabl_551fd064'
+
+migration_query = """
+UPDATE {new_table} B
+SET
+B.pse_daily_attendance = A.pse_days_attended,
+B.pse_daily_attendance_male = CASE WHEN B.sex = 'M' THEN A.pse_days_attended ELSE NULL END,
+B.pse_daily_attendance_female = CASE WHEN B.sex = 'F' THEN A.pse_days_attended ELSE NULL END
+FROM {old_table} A
+WHERE A.doc_id = B.doc_id and A.month_start = B.month_start
```
gcapalbo: just to confirm - is (doc_id, month_start) what uniquely defines a record on both tables, and is it indexed (or is it defined explicitly as primary key)? if so then we're fine

gcapalbo (approved):

gcapalbo:
just want to reiterate that we should probably stop the processes which write to new_table while this is running

custom/icds/management/commands/migrate_pse_attendance.py
calellowitz: doc_id, month_start defines a unique record. i believe doc_id is indexed and potentially an explicit primary key and there are exactly 7 month_starts for each doc_id. I will want to coordinate this process with emord as he has been managing the ongoing rebuild of these datasources and want to make sure its alright to stop those processes but it sounds like we are good to go

emord: We can likely do this tomorrow if you want.

I think @gcapalbo was asking if it is indexed as (doc.id, month_start), but the primary key is (doc.id, repeat_iteration) There is not a separate index on month_start. It may be worth adding a unique index on (doc.id, month_start) on both tables.

@sheelio would probably know this better, but I believe that (doc_id, month_start) is unique even though its not hte primary key

custom/icds/management/commands/migrate_pse_attendance.py
```diff
@@ -0,0 +1,29 @@
+from __future__ import print_function
+
+from django.core.management.base import BaseCommand
+
+from django.db import connections
+
+
+new_table = 'config_report_icds-cas_static-child_cases_monthly_v2_198ccc06'
+old_table = 'config_report_icds-cas_static-child_cases_monthly_tabl_551fd064'
```
emord: Is this pulling data from a tableau data source to a non tableau data source. I think tha't sfine, just want to make sure it's intentional

emord (approved):

custom/icds/management/commands/migrate_pse_attendance.py
gcapalbo: ok, thanks for the info, i think this will take a lot longer unless there's a multicolumn index for the columns we're joining on.

i don't have much context on this data model - should we instead be joining on (doc_id, repeat_iteration) if that's the primary key?

emord: I don't think so. repeat_iteration is what determines the month. 

So if one of the data sources had processed a case last month and the other data source processed one this month (I don' thtink this will happen), then it could be that in one data source 1 refers to june 2017 and the other 1 refers to july 2017.

May have been better to have a primary key on doc_id, month instead, but I think it's required to have repeat_iteration in UCRs if you use an iterator. would need to look more into it to confirm

custom/icds/management/commands/migrate_pse_attendance.py
```diff
@@ -0,0 +1,29 @@
...
+
+migration_query = """
+UPDATE {new_table} B
+SET
+B.pse_daily_attendance = A.pse_days_attended,
+B.pse_daily_attendance_male = CASE WHEN B.sex = 'M' THEN A.pse_days_attended ELSE NULL END,
+B.pse_daily_attendance_female = CASE WHEN B.sex = 'F' THEN A.pse_days_attended ELSE NULL END
```
gcapalbo: just realized these aliases for `B.` should not be on the left side of the equals sign here (my mistake). the aliases on the right side are fine though.

custom/icds/management/commands/migrate_pse_attendance.py
emord: offlined with cal. month_start doesn't exist on the tableau data source. so the query will need to be changed

custom/icds/management/commands/migrate_pse_attendance.py
```diff
@@ -12,10 +12,10 @@
 UPDATE {new_table} B
```
emord: I think these tables will need quotes around them because tehy have - in the name

custom/icds/management/commands/migrate_pse_attendance.py
emord: why don't you need `B.` in `B.sex` ?

custom/icds/management/commands/migrate_pse_attendance.py
```diff
@@ -0,0 +1,45 @@
...
+old_table = 'config_report_icds-cas_static-child_cases_monthly_tabl_551fd064'
+
+migration_query = """
+UPDATE "{new_table}" B
+SET
+pse_daily_attendance = A.pse_days_attended,
+pse_daily_attendance_male = CASE WHEN sex = 'M' THEN A.pse_days_attended ELSE NULL END,
+pse_daily_attendance_female = CASE WHEN sex = 'F' THEN A.pse_days_attended ELSE NULL END
+FROM "{old_table}" A
+WHERE A.doc_id = B.doc_id and A.month = immutable_date_cast(B.month_start)
```
gcapalbo: is there any other way to do this join, on indexed columns? if we don't do it on indexed columns it's probably going to take a very long time

custom/icds/management/commands/migrate_pse_attendance.py
```diff
@@ -0,0 +1,45 @@
...
+                print('Query is')
+                print(migration_query)
+                print('-----------------------------')
+                explain_result = cursor.execute("EXPLAIN {}".format(migration_query))
+                print(explain_result)
+            else:
+                print('Migrating pse data')
+                cursor.execute(migration_query)
+                print('Vacuuming table')
+                cursor.execute("VACUUM ANALYZE {}".format(new_table))
```
gcapalbo: this will also need quotes around the table name

custom/icds/management/commands/migrate_pse_attendance.py
gcapalbo: it's fine to keep in `B.sex`, the alias just can't be on the left side of the equals in the SET expression when naming the columns to update

custom/icds/management/commands/migrate_pse_attendance.py
calellowitz: we actually added an expression index on (doc_id, immutable_date_cast(month_start)), because the relevant columns in each table were of different types. so there is an index on that, although explain did not appear to use it. perhaps because the column being written isnt in the index. I tested an EXPLAIN on a join on just doc_id and it also did not plan on using the index.

PR #17046, opened 2017-07-17
support editing a form that removes a ledger transaction

snopoke:
https://sentry.io/dimagi/commcarehq/issues/309954268/

Note that this also moves the case saving functionality out of SQL and into Python.


corehq/form_processor/backends/sql/dbaccessors.py
```diff
@@ -829,53 +830,49 @@ def get_case_ids_in_domain_by_owners(domain, owner_ids, closed=None):
...
+                case.save(using=db_name)
+                for case_transaction in transactions_to_save:
+                    case_transaction.save(using=db_name)
+
+                for index in indices_to_save_or_update:
+                    index.domain = case.domain  # ensure domain is set on indices
+                    update_fields = None
+                    if index.is_saved():
+                        # prevent changing identifier
+                        update_fields = ['referenced_id', 'referenced_type', 'relationship_id']
```
proteusvacuum: out of curiosity, when do indices get updated, and not just removed / created?

proteusvacuum (commented):
Looks fine to me. Will let someone with more knowledge of this area of the code approve

corehq/form_processor/backends/sql/dbaccessors.py
```diff
@@ -829,53 +830,49 @@ def get_case_ids_in_domain_by_owners(domain, owner_ids, closed=None):
...
+                        )
+                    attachment.save(using=db_name)
+
+                CaseAttachmentSQL.objects.using(db_name).filter(id__in=attachment_ids_to_delete).delete()
                 for attachment in case.get_tracked_models_to_delete(CaseAttachmentSQL):
                     attachment.delete_content()
+
                 case.clear_tracked_models()
+        except InternalError as e:
+            raise CaseSaveError(e)
```
benrudolph: woah this is huge. big fan of taking this out of sql

benrudolph (approved):
my only concern is that do we want to support ledgers in edit forms? i'm concerned that we'll end up in a state where we are with cases where it sometimes works

snopoke:
In addition to not having to maintain the SQL we're starting to hit the
limits of a single pl_proxy node. We could setup a new node and round robin
between the 2 or we could rewrite more of the queries in python and query
the shard databases directly.

In the long run it would be nice to get rid of pl_proxy so I thought I'd
rather move this now than make new changes to the SQL function.

corehq/form_processor/backends/sql/dbaccessors.py
snopoke: good question, not 100% sure but it's functionality we support now so don't want to remove it just yet.

snopoke:
@benrudolph good point, we don't prevent it now (I don't think). I'm going to leave that for a separate discussion.

PR #17049, opened 2017-07-17
change session backed to just cache

snopoke:
pulled this out of https://github.com/dimagi/commcare-hq/pull/16992

This will force logout everyone who's logged in when it get's deployed. Added 'open for review' flag until we can check if we need to do comms.

### DIFF
--- a/settings.py
+++ b/settings.py
@@ -140,7 +140,6 @@ MIDDLEWARE = [
     'django.middleware.common.CommonMiddleware',
     'django.middleware.csrf.CsrfViewMiddleware',
     'django.contrib.auth.middleware.AuthenticationMiddleware',
-    'django.contrib.auth.middleware.SessionAuthenticationMiddleware',
     'django.contrib.messages.middleware.MessageMiddleware',
     'django.middleware.common.BrokenLinkEmailsMiddleware',
     'django_otp.middleware.OTPMiddleware',
--- a/settings.py
+++ b/settings.py
@@ -157,7 +157,7 @@ MIDDLEWARE = [
     'corehq.apps.locations.middleware.LocationAccessMiddleware',
 ]
 
-SESSION_ENGINE = "django.contrib.sessions.backends.cached_db"
+SESSION_ENGINE = "django.contrib.sessions.backends.cache"
 
 # time in minutes before forced logout due to inactivity
 INACTIVITY_TIMEOUT = 60 * 24 * 14


benrudolph:
chatted with @emord about this at one point. there is a warning in the docs about the cached based backend: https://docs.djangoproject.com/en/1.11/topics/http/sessions/#using-cached-sessions

it's unclear though if the warning is meant as "don't use local-memory cache" or "only use Memcached backend"

snopoke:
The way I read it is that you shouldn't use the local memory cache but
memcache (or any proper cache service) is fine

On 17 Jul 2017 22:40, "Ben Rudolph" <notifications@github.com> wrote:

chatted with @emord <https://github.com/emord> about this at one point.
there is a warning in the docs about the cached based backend:
https://docs.djangoproject.com/en/1.11/topics/http/sessions/#using-cached-
sessions

it's unclear though if the warning is meant as "don't use local-memory
cache" or "only use Memcached backend"

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/dimagi/commcare-hq/pull/17049#issuecomment-315876386>,
or mute the thread

benrudolph (approved):
that's my gut too. whenever you're ready

emord:
> The way I read it is that you shouldn't use the local memory cache but
memcache (or any proper cache service) is fine

I think that's what we decided as well.

Right now this is on the same server as couch0 on icds and hqdb0 on prod (i think), which have both had issues. do you think that we should have different instances of redis for this?

snopoke:
@dannyroberts we need to time this with a deploy. Once we have a maintenance window I'll co-ordinate with you.

dannyroberts:
This breaks the way that formplayer fetches the session, which was directly from postgres. I'm reverting

dannyroberts:
@benrudolph and I briefly tried to fix this by rewriting formplayer to look in redis instead of postgres for the session. The cache backend, unfortunately, serializes it using python's `pickle` format (not something well supported across languages like JSON), so there's no great way for java to read it. We tried just fetching it and pulling out the part we wanted with a regex, but the more we went down that rabbit hole the less good it felt. So we ended up just reverting this PR.

We can discuss where we want to go from here

benrudolph:
i think likely the easiest things to do is to make a web request endpoint that takes a sessionid and returns when it expires and other session metadata like the user id

snopoke:
Ah this sucks, sorry guys. According to the django docs JSON is the default serialization format: https://docs.djangoproject.com/en/1.10/topics/http/sessions/#session-serialization

Except that doesn't apply to the cache backend: https://code.djangoproject.com/ticket/23764

It would be pretty simple to change that (2 lines): https://github.com/django/django/pull/4372

snopoke:
@benrudolph @dannyroberts do you think it would be better to (1) implement an API in HQ or (2) change the format to JSON and have formplayer pull it out directly.

I'm inclined to go with (2) since it's less requests to HQ but (2) is probably a 'more correct' approach.

benrudolph:
> but (2) is probably a 'more correct' approach.
I think you meant (1) is the more correct approach?

I agree that (1) is going to be the most robust and future proof, but will likely add a 500ms to each formplayer request. i'm fine going with 2 since it's not making the solution any less robust

PR #17054, opened 2017-07-17
cache get_practice_mode_mobile_workers

dannyroberts:
This can be slow enough of a query that it will sometimes time out on domains with many users. Let's at least cache it.


dannyroberts:
@proteusvacuum fyi

snopoke (approved):

PR #17061, opened 2017-07-17
Multi column indexes for UCR

emord:
@dimagi/scale-team @sheelio 

Already added the ones that I've added manually.

buddy @esoergel

### DIFF
--- a/corehq/apps/userreports/models.py
+++ b/corehq/apps/userreports/models.py
@@ -14,17 +14,16 @@ from corehq.util.quickcache import quickcache
 from dimagi.ext.couchdbkit import (
     BooleanProperty,
     DateTimeProperty,
-    Document,
-    DocumentSchema,
-    SchemaProperty,
-    StringListProperty,
-)
-from dimagi.ext.couchdbkit import (
     DecimalProperty,
     DictProperty,
+    Document,
+    DocumentSchema,
     IntegerProperty,
     ListProperty,
+    SchemaProperty,
+    SchemaListProperty,
     StringProperty,
+    StringListProperty,
 )
 from dimagi.ext.jsonobject import JsonObject
 from corehq.apps.cachehq.mixins import (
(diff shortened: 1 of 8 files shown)


corehq/apps/userreports/models.py
```diff
@@ -66,6 +65,10 @@ class ElasticSearchIndexSettings(DocumentSchema):
     number_of_shards = IntegerProperty(default=2)
 
 
+class SQLColumnIndexes(DocumentSchema):
+    column_ids = StringListProperty()
```
emord: I made this a separate schema so it could later be modified to do partial queries or maybe a GIN index if we would like

NoahCarnahan (approved):
nice!

corehq/apps/userreports/sql/adapter.py
```diff
@@ -145,16 +148,30 @@ def handle_exception(self, doc, exception):
...
     sql_columns = [column_to_sql(col) for col in indicator_config.get_columns()]
     table_name = get_table_name(indicator_config.domain, indicator_config.table_id)
+    extra_indices = [
+        Index(
+            _custom_index_name(table_name, index.column_ids),
+            *index.column_ids
+        )
+        for index in indicator_config.sql_column_indexes
+    ]
+    columns_and_indices = sql_columns + extra_indices
```
snopoke: not blocking but you should validate that the columns specified in the index are actual columns in the table.

snopoke:
Will these changes cause a rebuild?

emord:
> Will these changes cause a rebuild

@snopoke 

Indexes won't cause a rebuild. It's only changing columns that cause rebuilds https://github.com/dimagi/commcare-hq/blob/master/corehq/ex-submodules/fluff/signals.py#L48

These are indexes I've already added so they should be no-ops. just keeping them here for documentation purposes

snopoke:
I didn't think so, just making sure.

PR #16918, opened 2017-07-06
Externalize js: reports base

orangejenny:
@millerdev / @gcapalbo / @kaapstorm 

Review by commit. Will merge once other my other open reports PRs are merged; I'd rather deploy one of these at a time.

PR #17062, opened 2017-07-17
Fix ilsgateway base report template

orangejenny:
https://github.com/dimagi/commcare-hq/pull/16918 broke the ilsgateway reports because `#savedReports` now has its knockout bindings applied in `base.js` so they don't need to be applied in `ilsgateway/base_template.html`.

screen shot 2017-07-17 at 6 08 56 pm

@gcapalbo / @kaapstorm


benrudolph (approved):
nice

orangejenny:
Merging for the sake of deploying. Coincidentally, there aren't any tests for this.

PR #17063, opened 2017-07-18
Add count to case search admin view

proteusvacuum:
Adds the total number of results to the case search admin view. 
Useful for seeing for e.g. how many cases have a specific case property set to a particular value

### DIFF
--- a/corehq/apps/case_search/views.py
+++ b/corehq/apps/case_search/views.py
@@ -61,4 +61,4 @@ class CaseSearchView(DomainViewMixin, TemplateView):
             new_query = merge_queries(search.get_query(), addition.query_addition)
             search = search.set_query(new_query)
         search_results = search.run().raw_hits
-        return json_response({'values': search_results})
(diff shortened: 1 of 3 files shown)


corehq/apps/case_search/views.py
```diff
@@ -60,5 +60,5 @@ def post(self, request, *args, **kwargs):
             addition = CaseSearchQueryAddition.objects.get(id=query_addition, domain=self.domain)
             new_query = merge_queries(search.get_query(), addition.query_addition)
             search = search.set_query(new_query)
-        search_results = search.run().raw_hits
-        return json_response({'values': search_results})
+        search_results = search.size(CASE_SEARCH_MAX_RESULTS).run().raw_hits
+        return json_response({'values': search_results, 'count': search.count()})
```
snopoke: This will execute the query twice. You can get the count from the some results as the search results:
```
results = search.size(...).run()
search_results = results.raw_hits
count = results.total
```

snopoke (changes requested):

PR #17064, opened 2017-07-18
Private sector DTO is comes from the app, not hierarchy

proteusvacuum:
I'm not sure that this is the right change to make, so hold off on merge. This is a change in requirements.


snopoke (approved):
Not merging due to label

proteusvacuum:
Not required.

PR #17065, opened 2017-07-18
Refactor AdherenceTask

proteusvacuum:
We're about to do some work in here, to add better error handling, so cleaned it up a little before that happens. 

Mostly just moved things into the correct class so that all the sub-tasks have the same signature.


snopoke:
:boot: :construction:

proteusvacuum:
✔️

PR #17073, opened 2017-07-18
more indexes for ICDS

emord:
Things to do:

- [x] add these concurrently to the database
- [ ] merge and deploy this branch
- [ ] remove indexes removed here
- [ ] once we have more stat gathering on pg0 we should re-evaluate these indexes. for instance some of these are indexed on month, but it's possible that (supervisor_id, age_in_months) would be a better index than (supervisor_id, months)

I also need to add columns to the new v2 indexes, but have been too busy adding them to v1 so far.

@dimagi/scale-team buddy @esoergel


snopoke (approved):

emord:
All have been added to icds-cas table

PR #17074, opened 2017-07-18
Python form save

benrudolph:
I believe this is all it takes


corehq/form_processor/backends/sql/dbaccessors.py
```diff
@@ -507,13 +511,24 @@ def save_new_form(form):
...
+                            'XFormAttachmentSQL {} has already been saved'.format(attachment.id)
+                        )
+                    attachment.save(using=db_name)
+
+                for operation in operations:
+                    if operation.is_saved():
+                        raise XFormSaveError(
+                            'XFormOperationSQL {} has already been saved'.format(operation.id)
+                        )
+                    operation.save(using=db_name)
```
snopoke: I know this is in an atomic transaction but seems like we could check the attachments and operations before and only start the transaction of all is good.

I know we have the same pattern in the case save but just noticing it now.

benrudolph: cool, will do

benrudolph: (also just so you kknow this is wrapped in a larger transaction block in `saved_processed_models`

snopoke (approved):

corehq/form_processor/backends/sql/dbaccessors.py
benrudolph: if what you say is true (which is sounds like it is) then saving a form isn't atomic. https://github.com/dimagi/commcare-hq/blob/br/form-save-python/corehq/form_processor/backends/sql/processor.py#L77 we could successfully save the form, then fail on the case save which leaves us in a weird state. wonder if that has happened at all

snopoke:
Correct, it's not quite atomic. What we do though is update the form to
indicate that it failed. I'm sure there are edge cases though that could
leave things in a weird state. For example we don't till back changes to
cases so if once case save goes through but the next fails then the first
case won't get rolled back.

With pl_proxy it was impossible to make it atomic but moving to python I
think we can with nested transactions.

PR #17075, opened 2017-07-18
ICDS: Add single location level indexes

sheelio:
@emord 
This should handle the current dashboard v2 queries


emord (approved):
The indexes on state_id aren't necessary since it's the first one in the multi column index but it can be removed later

PR #17076, opened 2017-07-18
Allow copy of form to module with different case type

orangejenny:
https://manage.dimagi.com/default.asp?258326

Didn't identify exactly where this broke, but the comments say this should work as Jen says it should. Planning to get rid of `ConflictingCaseTypeError` altogether soon, since it's really a warning, not an exception.

@snopoke / @biyeun


snopoke (approved):
kicked build

PR #17077, opened 2017-07-18
Removed unused index.

emord:
This index isn't necessary because there's a unique index on (xform_id, i). It should be instantaneous because it's dropping it from the parent table and not the child tables, but all future child tables won't have this index. With that said I'll deploy to NIC tomorrow and verify

buddy @nickpell


nickpell (approved):

emord:
deployed to NIC

PR #17080, opened 2017-07-18
Bucket restore times instead of histogram

millerdev:
Follow-up on https://github.com/dimagi/commcare-hq/pull/17030

Clayton had the idea of putting counts in time buckets, which is much more useful and flexible for aggregate stats. Any aggregations we were doing in the past were likely close to meaningless because of the number of tags sent to the histogram call.

@snopoke @benrudolph @ctsims

### DIFF
--- a/corehq/apps/ota/views.py
+++ b/corehq/apps/ota/views.py
@@ -37,7 +37,7 @@ from corehq.apps.ota.tasks import queue_prime_restore
 from corehq.apps.users.models import CommCareUser, CouchUser
 from corehq.apps.locations.permissions import location_safe
 from corehq.form_processor.exceptions import CaseNotFound
-from corehq.util.datadog.gauges import datadog_counter, datadog_histogram
+from corehq.util.datadog.gauges import datadog_counter
 from dimagi.utils.decorators.memoized import memoized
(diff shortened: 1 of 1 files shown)


snopoke (approved):
:boot: :construction:

benrudolph:
would love to offline at some point about this. read the previous PR and still not quite sure why we couldn't use the histogram type if we just used it for the `avg`. we've recently switched a  lot of metrics to histogram, are you concerned that those could be wrong as well?

millerdev:
> we've recently switched a lot of metrics to histogram, are you concerned that those could be wrong as well?

Yes. Did you read https://help.datadoghq.com/hc/en-us/articles/211545826-Why-histogram-stats-are-all-the-same-inaccurate-Characteristics-of-Datadog-histograms-? Unless you're graphing per-host (non-aggregated) stats, and likely not using tags, I believe datadog's histogram is not very useful. Especially see the section in that article under **Consequence 2: ...**

benrudolph:
wow, ok missed that article. that definitely seems like our metrics our off then. though the `avg` could be close to correct if we are load balancing correctly and every host seems the same number of metrics per 10 second interval... though that seems unlikely

PR #17082, opened 2017-07-18
Web apps permissions + login as: prepwork

dannyroberts:
https://manage.dimagi.com/default.asp?257563

@benrudolph this PR implements the backend for the approach of loading in app availability data per user along with the original list of users.

You can check for anything you might think is a performance issue, but I think this'll actually be really quick. Notably, we only load 10 users at a time, and the user/group relationships for those users is a single couch query.

### DIFF
--- a/corehq/apps/cloudcare/views.py
+++ b/corehq/apps/cloudcare/views.py
@@ -1,4 +1,5 @@
 import HTMLParser
+from collections import defaultdict
 import json
 from xml.etree import ElementTree
 
@@ -21,6 +22,7 @@ from couchdbkit import ResourceConflict
 from casexml.apps.case.models import CASE_STATUS_OPEN
 from casexml.apps.case.xml import V2
 from casexml.apps.phone.fixtures import generator
+from corehq.apps.groups.dbaccessors import get_groups_by_user
 from corehq.form_processor.utils import should_use_sql_backend
 from corehq.form_processor.utils.general import use_sqlite_backend
 from dimagi.utils.logging import notify_exception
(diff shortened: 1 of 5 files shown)


dannyroberts:
cc @sravfeyn

corehq/apps/cloudcare/views.py
```diff
@@ -421,6 +423,23 @@ def get(self, request, *args, **kwargs):
         })
 
 
+def get_apps_by_user(domain, user_ids):
```
sravfeyn: A docstring is useful for reading this latter

sravfeyn (approved):

sravfeyn:
Build fail
```
FAIL: corehq.apps.users.tests.test_user_model:UserDeviceTest.test_only_update_once_per_day
  File "/mnt/commcare-hq-ro/corehq/apps/users/tests/test_user_model.py", line 109, in test_only_update_once_per_day
    self.assertFalse(user.update_device_id_last_used(device, later))
AssertionError: True is not false
```

corehq/apps/cloudcare/views.py
```diff
@@ -421,6 +423,23 @@ def get(self, request, *args, **kwargs):
...
+    groups_by_user = get_groups_by_user(domain, user_ids)
+
+    app_access = ApplicationAccess.get_by_domain(domain)
```
benrudolph: what if we just sent this down to the browser and did the app filtering in JS using this?

benrudolph: The app_access doc has the group_id and app_id, the user has its group memberships so it should be pretty easy to implement

dannyroberts:
(restarted the failed job of the build—i saw a similar failure on another PR around the same time last night, and on that one restarting the failed job fixed it.)

corehq/apps/cloudcare/views.py
dannyroberts: > the user has its group memberships

Oh? I don't think currently (i.e. outside of this PR) `LoginAsUsers` is including group info

```python
    def _format_user(self, user_json):
        user = CouchUser.wrap_correctly(user_json)
        return {
            'username': user.raw_username,
            'customFields': user.user_data,
            'first_name': user.first_name,
            'last_name': user.last_name,
            'phoneNumbers': user.phone_numbers,
            'user_id': user.user_id,
            'location': user.sql_location.to_json() if user.sql_location else None,
        }
```

Are you suggesting I include group info here, instead of adding alvailable app info to the user?

benrudolph: yes, sorry i meant that the user in ES has its group info, so it'd be easy to add to the dict

dannyroberts: Ah, gotcha, thanks

PR #17083, opened 2017-07-18
survive and log failure to get practice workers

dannyroberts:
Still a temporary fix, but will get people able to use it again when this is an issue

see: https://manage.dimagi.com/default.asp?258463


sravfeyn:
Didn't imagine there will be hundreds of practice users in a project. The long term fix could then be to make the practice-user-selection an async select2 search box

dannyroberts:
Yeah I'm not sure if there are hundreds of practice users, or just the process of filtering them in elasticsearch from the thousands of users is expensive. It's also possible that this will subside also as we improve ES stability on the india environment.

PR #17093, opened 2017-07-19
Kill OPM reports

orangejenny:
Anushree signed off.

@esoergel


emord:
nice. you'll probably need to change settings.py in a few places like pillows https://github.com/dimagi/commcare-hq/blob/master/settings.py#L1740

there will also need to be a pr to remove the pillow from teh deploy directory https://github.com/dimagi/commcare-hq-deploy/blob/master/fab/environments.yml#L376

orangejenny:
@emord Thanks. Added a commit to remove references to the reports and the pillow, and also https://github.com/dimagi/commcare-hq-deploy/pull/335

esoergel (approved):
I think we might also need to manually remove the db, @dannyroberts do you remember how that works?

dannyroberts:
Good point @esoergel, I think you're intuition's right here that our deploys won't ever automatically delete a (couchdb) db, so we'll have to run that manually to reclaim that space

dannyroberts:
Also: 🎉 🎈 🎊

PR #17101, opened 2017-07-19
Handle javascript translations on transifex

esoergel:
We haven't been sending these to transifex, so none of the strings were translated.  I thought for a bit that something else was broken too, 'cause I still wasn't seeing things locally, but it turns out I'd just forgotten to run:
```$ ./manage.py compilejsi18n```

@kaapstorm @orangejenny @biyeun

### DIFF
--- a/scripts/update-translations.sh
+++ b/scripts/update-translations.sh
@@ -37,7 +37,7 @@ echo "Gathering all translation strings.  Note that this will probably take a wh
 ./manage.py makemessages --all --ignore 'corehq/apps/app_manager/tests/data/v2_diffs*' --ignore 'node_modules'
 
 echo "Gathering javascript translation strings.  This will also probably take a while"
-./manage.py makemessages -d djangojs --all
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: scripts/update-translations.sh [hunk 1]
```diff
@@ -37,7 +37,7 @@ echo "Gathering all translation strings.  Note that this will probably take a wh
 ./manage.py makemessages --all --ignore 'corehq/apps/app_manager/tests/data/v2_diffs*' --ignore 'node_modules'
 
 echo "Gathering javascript translation strings.  This will also probably take a while"
-./manage.py makemessages -d djangojs --all
+./manage.py makemessages -d djangojs --all --ignore 'corehq/apps/app_manager/tests/data/v2_diffs*' --ignore 'node_modules' --ignore 'bower_components'
```
esoergel: Right now there are a bunch of `node_modules` strings in the `.po` file, but these will go away after I run `./scripts/update_translations.sh` (after this is merged).
Is there anything else we should be skipping over?  See anything in https://github.com/dimagi/commcare-hq/blob/master/locale/en/LC_MESSAGES/djangojs.po that doesn't belong?

orangejenny (approved):

PR #17105, opened 2017-07-19
Atomic save processed models

benrudolph:
@snopoke @calellowitz quick POC for atomic saves. i think we should remove the other atomic calls in the `saves`. i don't really know how this will affect the load, but it certainly should cut down on only half of the cases being submitted. thoughts?

cc: @proteusvacuum

### DIFF
--- a/requirements/requirements.txt
+++ b/requirements/requirements.txt
@@ -99,3 +99,4 @@ quickcache==0.0.1
 zeep==1.6.0
 ethiopian_date==0.1.1
 architect==0.5.6
+contextlib2=0.5.4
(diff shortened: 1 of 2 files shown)


calellowitz (approved):
looks good to me. and definitely an improvement over the other method

requirements/requirements.txt
```diff
@@ -99,3 +99,4 @@ quickcache==0.0.1
 zeep==1.6.0
 ethiopian_date==0.1.1
 architect==0.5.6
+contextlib2=0.5.4
```
calellowitz: `==`

proteusvacuum (approved):

snopoke:
This is very cool. It's been on my list for ages to figure out a good way of doing the dynamic nested transactions.

PR #17109, opened 2017-07-20
better get_reverse_indexed_cases

snopoke:
this often gets called with a long list of ID's


millerdev (approved):

emord:
I'm not sure if this is the exact cause of this, but this might have sped up some of the ICDS UCRs by more than 2x:

image

PR #17114, opened 2017-07-20
add UATBC data dump from July 19

nickpell:
@emord


emord (approved):
not a lot of context but looks good. guessing you've got a system for this stuff at this point

PR #17120, opened 2017-07-20
ICDS: Combine person queries

sheelio:
@emord 

This combines the 4 person queries into a single query.


emord (approved):
Thanks. I did a quick explain and it looks like this one query is only a bit more expensive than just the first one.

PR #17124, opened 2017-07-21
allow serving app files directly from blobdb

snopoke:
Experiment to see if this helps with large volume of app installs on ICDS

cc @dimagi/scale-team 
buddy @orangejenny

### DIFF
--- a/corehq/apps/app_manager/views/download.py
+++ b/corehq/apps/app_manager/views/download.py
@@ -240,11 +240,17 @@ def download_file(request, domain, app_id, path):
 
     try:
         assert request.app.copy_of
-        obj = CachedObject('{id}::{path}'.format(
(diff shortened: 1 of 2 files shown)


corehq/apps/app_manager/views/download.py
```diff
@@ -240,11 +240,17 @@ def resolve_path(path):
 
     try:
         assert request.app.copy_of
-        obj = CachedObject('{id}::{path}'.format(
-            id=request.app._id,
-            path=full_path,
-        ))
-        if not obj.is_cached():
+        if toggles.NO_CACHE_APP_FILES.enabled(domain):
```
emord: How heavily is htis cached? this will cause more hits to redis/couch, but guessing it's still worth it

PR #17129, opened 2017-07-21
Ce/linked form

calellowitz:
@snopoke @orangejenny cc: @mkangia 
This will ensure that future migrations with the management command on ICDS work until we get linked apps set up there. I am holding off on truly unique ids for linked apps until the new mobile version comes out that fixes updates when a form id has changed. Otherwise it could break all existing linked app deployments. But I will just be able to remove the conditional and extra argument at that point, then add linked apps to the `xforms_index`.


orangejenny:
Looks fine in general, but there's rather a lot of code that overlaps with `update_unique_ids`, any possibility of refactoring?

snopoke:
Agree with @orangejenny about refactoring to reduce code duplication

calellowitz:
refactored a bit @orangejenny

corehq/apps/app_manager/views/utils.py
```diff
@@ -128,10 +131,11 @@ def get_blank_form_xml(form_name):
     })
 
 
-def overwrite_app(app, master_build, report_map=None):
+def overwrite_app(app, master_build, report_map=None, maintain_ids=False):
     excluded_fields = set(Application._meta_fields).union(
         ['date_created', 'build_profiles', 'copy_history', 'copy_of', 'name', 'comment', 'doc_type']
     )
+    id_map = _get_form_id_map(app)
```
snopoke: Could put this under the `if maintain_ids` check

corehq/apps/app_manager/views/utils.py
```diff
@@ -148,11 +152,45 @@ def overwrite_app(app, master_build, report_map=None):
...
+    new_wrapped_app = new_wrapped_app.save_attachments(attachments)
+    return new_wrapped_app
+
+
+def _save_attachments(app, attachments):
+    with app.atomic_blobs():
+        for name, attachment in attachments.items():
+            if re.match(ATTACHMENT_REGEX, name):
+                app.put_attachment(attachment, name)
+    return app
```
snopoke: don't think this is needed anymore

snopoke (changes requested):
small changes

corehq/apps/app_manager/views/utils.py
```diff
@@ -128,10 +131,14 @@ def get_blank_form_xml(form_name):
...
 
-def overwrite_app(app, master_build, report_map=None):
+def overwrite_app(app, master_build, report_map=None, maintain_ids=False):
     excluded_fields = set(Application._meta_fields).union(
         ['date_created', 'build_profiles', 'copy_history', 'copy_of', 'name', 'comment', 'doc_type']
     )
+    if maintain_ids:
+        id_map = _get_form_id_map(app)
+    else:
+        id_map = {}
```
snopoke: nit: the only place I see this getting used is on line 159 where it's passed to `_update_form_ids`. That line is also under the same if check `if maintain_ids` so it seems to me like `id_map` can be defined there:
```
if maintain_ids:
    id_map = _get_from_id_map(app)
    wrapped_app = _update_form_ids(wrapped_app, master_build, id_map)
```

PR #17130, opened 2017-07-21
Set env settings to not use replicas

benrudolph:
@emord i foresee a future state that is a very ugly `env_settings.py`

### DIFF
--- a/env_settings.py
+++ b/env_settings.py
@@ -112,6 +112,7 @@ ES_META = {
         'hqusers': {
             "settings": {
                 "number_of_shards": 2,
+                "number_of_replicas": 0,
                 "analysis": {
                     "analyzer": {
                         "default": {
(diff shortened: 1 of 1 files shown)


env_settings.py
```diff
@@ -107,4 +107,55 @@
             },
         }
     },
+
+    'swiss': {
+        'hqusers': {
+            "settings": {
+                "number_of_shards": 2,
```
emord: If this inherits settings properly couldn't you have something that's like:

```
'swiss': {
    'hqusers': {
        'settings': {
             'number_of_replicas': 0
}}}
` ?

Comment (benrudolph) [inline env_settings.py]: yes, yes it should

Comment (benrudolph) [inline env_settings.py]: wait no, i don't think it matters since we `update`:
```

In [5]: a = {'settings': { 'replicas': 0 }}

In [6]: a.update({'settings': {}})

In [7]: a
Out[7]: {'settings': {}}
```

emord (approved):

env_settings.py
benrudolph: (offlined)

emord:
Actually I think you may need to specify the number of replicas since the default is 1

benrudolph:
mm makes sense

emord (approved):

PR #17131, opened 2017-07-21
Hide shadow forms from app summary

sheelio:


sheelio:
@NoahCarnahan 

We'd like to hide shadow forms from the app summary (this change makes that possible I think)

NoahCarnahan (approved):
Looks good, thanks

sheelio:
For some reason the import failed at the top level, so I included it in the function (similar to other functions inside the same file).

NoahCarnahan:
Great, looks good. That error was probably because there was a circular import. `corehq.apps.app_manager.models` imported this file. But putting it in the function like you did breaks the cycle.

PR #17133, opened 2017-07-21
CAS-RCH Integration

mkangia:
[Spec](https://docs.google.com/document/d/1BKIGss_R1HRERSW6uHtuJ5GMYIF2hj4eBbYtAD6o-T4/edit#)


mkangia:
@snopoke this is ready now. 
Just need to make a couple of design improvements.

custom/rch/forms.py
```diff
@@ -0,0 +1,111 @@
...
+    return a tuple of tuples with value and display text for each unique available option for
+    display field
+    for ex: return all combinations for state code and state name for all unique state names
+    available
+    :param value_field: column for value
+    :param display_field: column for display text
+    :return: tuple of tuples
+    """
+    if settings.UNIT_TESTING:
+        return tuple()
```
snopoke: This doesn't seem right. I don't see anything in this method that would prevent it from working in unit tests.

custom/rch/forms.py
```diff
@@ -0,0 +1,111 @@
...
+    # and the proper values would be set
+    except ProgrammingError:
+        return tuple()
+
+
+class BeneficiariesFilterForm(forms.Form):
+    stcode = forms.ChoiceField(
+        label=_("State"),
+        required=False,
+        choices=get_choices_for('stcode', 'stname'),
```
snopoke: Rather than doing this statically you should do it in the form's `__init__`: https://stackoverflow.com/a/3420588/632517

That will also remove the need to catch `ProgrammingError` in the function

snopoke: Doing it in `__init__` would also mean you could get all the options from one query rather than multiple queries per field

custom/rch/models.py
```diff
@@ -0,0 +1,159 @@
...
+from jsonfield.fields import JSONField
+from custom.rch.const import (
+    MOTHER_RECORD_TYPE,
+    CHILD_RECORD_TYPE,
+    STATE_DISTRICT_MAPPING,
+    RCH_RECORD_TYPE_MAPPING,
+)
+
+
+class RCHRecord(models.Model):
```
snopoke: Seems like there is quite a lot of forking between `mother` and `child`. Did you consider using separate models or [model inheritance](https://docs.djangoproject.com/en/1.11/topics/db/models/#model-inheritance)?

custom/rch/models.py
```diff
@@ -0,0 +1,159 @@
...
+    cas_case_id = models.CharField(null=True, max_length=255)  # ICDS-CAS case that was found as a match
+    details = JSONField(default=dict)  # all details received from RCH for this beneficiary
+    district_id = models.PositiveSmallIntegerField(null=True)
+    state_id = models.PositiveSmallIntegerField(null=True)
+    village_id = models.IntegerField(null=True)
+    village_name = models.CharField(null=True, max_length=255)
+    name = models.CharField(null=True, max_length=255)
+    aadhar_num = models.BigIntegerField(null=True)
+    dob = models.DateTimeField(null=True)
+    rch_id = models.BigIntegerField(null=True)
```
snopoke: should this be unique?

snopoke: Also it looks like you're filtering on it so you'll almost certainly want an index

custom/rch/models.py
```diff
@@ -0,0 +1,159 @@
...
+                    rch_beneficiary.save()
+
+    def associate_matching_cas_record(self):
+        matching_icds_cas_case_id = find_matching_cas_record_id(self.aadhar_num)
+        if matching_icds_cas_case_id:
+            self.cas_case_id = matching_icds_cas_case_id
+            self.save()
+
+
+class AreaMapping(models.Model):
```
snopoke: might need some indexes on some of these fields depending on how big the table will be

custom/rch/utils.py
```diff
@@ -0,0 +1,83 @@
...
+
+def _initiate_client():
+    return Client(RCH_WSDL_URL)
+
+
+def _get_service():
+    return _initiate_client().service.DS_Data
+
+
+def _fetch_beneficiaries(for_date, state_id, beneficiary_type, district_id):
```
snopoke: some docstrings on these functions would be helpful to know what data they return

snopoke (commented):
Haven't completed my review yet

custom/rch/models.py
mkangia: As of now this table has 873 rows but would grow. Adding indexes.

custom/rch/models.py
mkangia: Yes.
Yes.

custom/rch/models.py
mkangia: I started with that path but then ended up merging them as a single entity since that made it easier. Reverting it back to two different models.

custom/rch/forms.py
mkangia: Doing this in `__init__` would mean that those queries would run each time the form is rendered. Should not we just call it once since it won't change anyway?

> Doing it in __init__ would also mean you could get all the options from one query rather than multiple queries per field

How is that?

custom/rch/forms.py
mkangia: The error here https://travis-ci.org/dimagi/commcare-hq/builds/279301373

custom/rch/views.py
```diff
@@ -0,0 +1,335 @@
...
+            # exclude beneficiaries where cas_case_id is null
+            self.beneficiaries = self.beneficiaries.exclude(cas_case_id__isnull=True)
+        elif not include_matched_beneficiaries:
+            # fetch beneficiaries where case_case_id is NOT null
+            self.beneficiaries = self.beneficiaries.exclude(cas_case_id__isnull=False)
+        village_ids = []
+
+        village_code = in_data.get('village_code')
+        if village_code:
+            village_ids = village_ids + [village_code]
```
snopoke: `village_ids` will always be an empty list here so it doesn't seem useful to include it. Just have `village_ids = [village_code]`

custom/rch/views.py
```diff
@@ -0,0 +1,335 @@
...
+        self.beneficiaries = case_es.CaseES().domain(ICDS_CAS_DOMAIN).size(RECORDS_PER_PAGE)
+        if not include_matched_beneficiaries:
+            self.beneficiaries = self.beneficiaries.filter(case_es.missing('rch_record_id'))
+        awc_ids = []
+
+        # find corresponding awc_id(s) for the selected village id to search
+        # cas records by their awc_ids set as owner_id
+        village_code = in_data.get('village_code')
+        if village_code:
+            awc_ids = awc_ids + AreaMapping.fetch_awc_ids_for_village_id(village_code)
```
snopoke: same here re `awc_ids` which will always be empty to no need to include it

custom/rch/views.py
```diff
@@ -0,0 +1,335 @@
...
+
+        # set beneficiaries empty if unable to filter by any awc_id
+        self.beneficiaries = []
+        return False
+
+    def _set_context_for_cas_records(self, context, in_data=None):
+        include_matched_beneficiaries = (in_data.get('matched') == 'on')
+        size = in_data.get('limit')
+        offset = (in_data.get('page') - 1) * in_data.get('limit')
+        if self._set_cas_beneficiaries(include_matched_beneficiaries, in_data=in_data):
```
snopoke: this should rather just return the beneficiaries rather than setting them as an attribute of the class:
```
beneficiaries = self._get_cas_beneficiaries(include_matched_beneficiaries, in_data=in_data)
if beneficiaries:
   ....
```

custom/rch/views.py
```diff
@@ -0,0 +1,335 @@
...
+            cas_records = self.beneficiaries.start(offset).size(size).run()
+            if cas_records:
```
snopoke: I think this will always be True. If you want to test to see if there are results you should use `cas_records.total`

custom/rch/views.py
```diff
@@ -0,0 +1,335 @@
...
+                context['beneficiaries'] = cas_records.hits
+                context['beneficiaries_total'] = cas_records.total
+                context['beneficiaries_count'] = len(cas_records.hits)
+
+    def _set_context_for_rch_records(self, context, in_data=None):
+        include_matched_beneficiaries = (in_data.get('matched') == 'on')
```
snopoke: nit: unnecessary brackets. Also in other places

custom/rch/views.py
```diff
@@ -0,0 +1,335 @@
...
+        if awc_ids:
+            awc_loc_ids = list(
+                SQLLocation.objects.filter(site_code__in=awc_ids).values_list('location_id', flat=True)
+            )
+            user_ids = user_ids_at_locations(awc_loc_ids)
+            # ToDo: Confirm that this check is correct to include both users and locations
+            # ToDo: remove awc_ids added for demo purpose
+            user_ids = user_ids + awc_loc_ids + awc_ids
+            if user_ids:
+                self.beneficiaries = self.beneficiaries.filter(case_es.user(user_ids))
```
snopoke: you can just go `beneficiaries.user(user_ids)`

custom/rch/views.py
```diff
@@ -0,0 +1,335 @@
...
+    @allow_remote_invocation
+    def get_pagination_data(self, in_data):
+        context = {}
+        if in_data.get('present_in'):
+            self._set_beneficiaries(context, in_data)
+
+        if in_data.get('present_in') == 'cas':
+            itemList = context.get('beneficiaries')
+            for item in itemList:
+                item['editUrl'] = reverse(CaseDetailsView.urlname, args=(ICDS_CAS_DOMAIN, item.get('_id')))
```
snopoke: can this not be done in JS? Doesn't seem great to do essentially the same reverse for every item.
```
# in python
context['case_details_view_url'] = reverse(CaseDetailsView.urlname, args=[request.domain, '---'])

# in template
{% initial_page_data 'case_details_view_url' case_details_view_url %}

# in JS
var url = hqImport('hqwebapp/js/initial_page_data').reverse('case_details_view_url', item._id)
```

snopoke: Same in `_format_beneficiary`

custom/rch/views.py
```diff
@@ -0,0 +1,335 @@
...
+    page_title = ugettext_noop("RCH-CAS Beneficiary Details")
+    template_name = 'rch/beneficiary.html'
+
+    def section_url(self):
+        return reverse(self.urlname, args=[ICDS_CAS_DOMAIN, self.beneficiary_type, self.beneficiary_id])
+
+    def page_url(self):
+        return reverse(self.urlname, args=[ICDS_CAS_DOMAIN, self.beneficiary_type, self.beneficiary_id])
+
+    @method_decorator(require_superuser)
```
snopoke: move to class level

settings.py
```diff
@@ -2205,3 +2208,23 @@ def _determine_couch_databases(couch_databases):
...
+    rch_fields = []
+    for case_type in field_mappings:
+        rch_fields = rch_fields + field_mappings[case_type].keys()
+    return rch_fields
+
+
+RCH_PERMITTED_FIELDS = {
+    'mother': extract_rch_fields_from_mapping('mother'),
+    'child': extract_rch_fields_from_mapping('child'),
+}
```
snopoke: would be nice to make these lazy so that they don't slow down the startup of every Django process

custom/rch/forms.py
mkangia: sorry, the correct build for the error is https://travis-ci.org/dimagi/commcare-hq/jobs/279305659

custom/rch/forms.py
mkangia: > You could cache the query result to avoid running the query every time. This would also mean that you could update the values without having to restart the service.

These values are not expected to change very often. It would be after months it would change so there is no need to worry about the need to refresh it often.

I still think loading them just once on load is still a better call instead of moving to `__init__`. The programming error will just happen once during first migration and after that it would just return results(blank in case of local environment). I would rather have 4 queries fired once on load then have them being called each time on form load or checking for cache in case we cache it.
If you still think it would be better, happy to move it to `__init__`

custom/rch/models.py
```diff
@@ -0,0 +1,156 @@
...
+        """
+        beneficiary_type = cls._beneficiary_type  # mother, child etc
+        if beneficiary_type in RCH_RECORD_TYPE_MAPPING.values():
+            return settings.RCH_PERMITTED_FIELDS[beneficiary_type]
+        else:
+            return set()
+
+    @property
+    def rch_id_key(self):
+        return self._rch_id_key
```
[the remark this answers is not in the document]
mkangia: I prefer to use this when i want to have a private property that should not be updated and just expose a method to make it available for read. Probably safe to use `self._rch_id_key` directly if used within class properties but using `self. rch_id_key()` when accessed outside of class.

custom/rch/utils.py
```diff
@@ -0,0 +1,87 @@
...
+    """
+    Fetch beneficiaries of a specific type from RCH API for a certain date and district under a state.
+    :param beneficiary_type: This can be any value from RCH_RECORD_TYPE_MAPPING keys
+    """
+    try:
+        response = send_request_for_beneficiaries(for_date, state_id, beneficiary_type, district_id)
+        data = response['_value_1'].find('NewDataSet')
+        response_content = etree_to_dict(data)['NewDataSet']
+        if (response_content[0].get('Table1', None) and
+                response_content[0]['Table1'][0]['Message'] == 'Please check your URL or Security Code'):
```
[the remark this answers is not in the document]
mkangia: the response code is always 200 so need to rely on the message.
making it a constant.

custom/rch/views.py
```diff
@@ -0,0 +1,335 @@
...
+    # ToDo: Check how to set page title. Looks like this is not working
+    page_title = ugettext_noop("RCH-CAS Beneficiary list")
+    template_name = 'rch/beneficiaries_list.html'
+
+    def section_url(self):
+        return reverse(self.urlname, args=[ICDS_CAS_DOMAIN])
+
+    @method_decorator(require_superuser)
+    @use_select2
+    @use_angular_js
```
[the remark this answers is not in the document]
mkangia: Need to set the cookie as well. So `dispatch` would need to stay. 
You probably added this review before i added the cookie set up.

custom/rch/views.py
mkangia: just added intentionally for better readability.

custom/rch/forms.py
esoergel: Having it this way means a db query on import, which I think means that that can happen whenever we run `./manage.py anything`.  If you move it to the `__init__` method, you can slap a `@quickcache` decorator on it to cache the results across all workers, or if you prefer, a `@memoized` decorator, which will store it in memory similar to how it's done now.  In either case, it'd be lazily evaluated, so it wouldn't impact startup times.

mkangia:
:(

PR #17134, opened 2017-07-21
Hide detail tile continue correctly

wpride:
Fixes https://manage.dimagi.com/default.asp?258686

### DIFF
--- a/corehq/apps/cloudcare/static/cloudcare/js/formplayer/menus/controller.js
+++ b/corehq/apps/cloudcare/static/cloudcare/js/formplayer/menus/controller.js
@@ -106,18 +106,18 @@ FormplayerFrontend.module("Menus", function (Menus, FormplayerFrontend, Backbone
                 },
             });
 
-            if (model.isPersistentDetail) {
-                $('#select-case').hide();
-            } else {
-                $('#select-case').show();
-            }
-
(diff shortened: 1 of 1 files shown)


benrudolph (approved):

corehq/apps/cloudcare/static/cloudcare/js/formplayer/menus/controller.js
```diff
@@ -106,18 +106,18 @@ FormplayerFrontend.module("Menus", function (Menus, FormplayerFrontend, Backbone
...
-
             $('#select-case').off('click').click(function () {
                 FormplayerFrontend.trigger("menu:select", caseId);
             });
             $('#case-detail-modal').find('.js-detail-tabs').html(tabListView.render().el);
             $('#case-detail-modal').find('.js-detail-content').html(menuListView.render().el);
             $('#case-detail-modal').modal('show');
+
+            if (model.isPersistentDetail) {
+                $('#case-detail-modal').find('#select-case').hide();
```
benrudolph: why can't you just use the `#select-case` id? usually if it's an id it should be the only instance on the page

benrudolph: if not it should use a class

PR #17135, opened 2017-07-21
modularize uiElement

biyeun:
@orangejenny i think there's a separate PR in the future for cleaning up some of the syntax in here, but first I want to get it to a working modularized state


orangejenny (approved):
Sweet!

ui-element.js is massive, suspect it'll be good to break it into smaller pieces eventually, but don't need to think about that today. Having fewer globals is enough of a step forward.

biyeun:
@orangejenny broke up the enormous file. wasn't feeling right to me

orangejenny (approved):
Sweet! If you've sanity tested, looks good to merge.

PR #17136, opened 2017-07-21
Allow nginx to cache app downloads.

emord:
@dimagi/scale-team in combination with https://github.com/dimagi/commcarehq-ansible/pull/857 we can let nginx cache app files. I'm pretty sure this is ok since this endpoint isn't protected by any sort of login.


emord:
Actually I'm not convinced this is safe anymore. confused app_id and build_id. will do more investigation on monday

PR #17137, opened 2017-07-21
Remove unused commtrack config options

dannyroberts:
- remove openlmis_config
- remove requisition_config

See: https://trello.com/c/usgtYuDe/28-requisitionactions-and-related-code

### DIFF
--- a/corehq/apps/commtrack/sms.py
+++ b/corehq/apps/commtrack/sms.py
@@ -23,7 +23,9 @@ from xml.etree import ElementTree
 from casexml.apps.case.mock import CaseBlock
 from corehq.apps.commtrack.exceptions import (
     NoDefaultLocationException,
-    NotAUserClassError)
+    NotAUserClassError,
+    RequisitionsHaveBeenRemoved,
+)
 import re
 from corehq.form_processor.parsers.ledgers.helpers import StockTransactionHelper
 
@@ -515,6 +517,8 @@ def to_instance(data):
             *stock_blocks
         )
         return etree.tostring(root, encoding='utf-8', pretty_print=True)
+    else:
+        raise RequisitionsHaveBeenRemoved()
 
 
 def truncate(text, maxlen, ellipsis='...'):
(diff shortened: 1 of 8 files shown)


snopoke (approved):

corehq/apps/commtrack/sms.py
```diff
@@ -124,27 +123,13 @@ def parse(self, text):
             args = args[1:]
 
         action = self.commtrack_settings.action_by_keyword(action_keyword)
-        if action and action.type == 'stock':
```
gcapalbo: i don't know much about this code but how come we're not checking action.type here anymore?

corehq/apps/commtrack/sms.py
```diff
@@ -601,9 +515,6 @@ def to_instance(data):
             *stock_blocks
         )
         return etree.tostring(root, encoding='utf-8', pretty_print=True)
-    # todo: remove?
-    else:
-        return requisition_case_xml(data, stock_blocks)
```
gcapalbo: should this raise an exception on else now? right now it will just return None

corehq/apps/commtrack/sms.py
dannyroberts: this "type" property isn't actually a declared property of the `CommtrackActionConfig` document schema class, but rather something that was being hacked in by `CommtrackConfig.action_by_keyword` to differentiate stock actions from requisition actions, which exist separately in the config but which this method returns together in a single dict. Since there's now only one type (stock), there's no need to hack in the type property. If you look for `action_by_keyword` in the PR diff, you'll see I also removed the part of that method that hacks `type` in.

corehq/apps/commtrack/sms.py
```diff
@@ -124,27 +123,13 @@ def parse(self, text):
...
             # TODO: support single-action by product, as well as by action?
             self.verify_location_registration()
             self.case_id = self.case.case_id
             _tx = self.single_action_transactions(action, args)
-        elif action and action.action in [
-            RequisitionActions.REQUEST,
-            RequisitionActions.FULFILL,
-            RequisitionActions.RECEIPTS
-        ]:
-            _assert(False, 'Someone is actually using requisition actions', {
```
esoergel: :rofl:

esoergel (approved):

PR #17140, opened 2017-07-21
reverse sql location hierarchy

calellowitz:
@benrudolph @gcapalbo 
This query reverses the sql location hierarchy for easier queries against the dim table


corehq/warehouse/transforms/sql/location_dim.sql
```diff
@@ -87,13 +85,19 @@ SET
...
 
 WHERE
-    location_dim_target.sql_location_id = l0.sql_location_id;
+    location_dim_target.sql_location_id = l1.sql_location_id OR
+    location_dim_target.sql_location_id = l2.sql_location_id OR
+    location_dim_target.sql_location_id = l3.sql_location_id OR
+    location_dim_target.sql_location_id = l4.sql_location_id OR
+    location_dim_target.sql_location_id = l5.sql_location_id OR
+    location_dim_target.sql_location_id = l6.sql_location_id OR
+    location_dim_target.sql_location_id = l7.sql_location_id;
```
gcapalbo: i think all the ORs here and the new join criteria will make this query not perform as optimally as it would have before.

i think all we actually need to do to support this is to just write a stored routine called `get_location_level_id` that takes the level to extract, and the 8 location ids, and returns the right one. then we can use that in the SET clause

calellowitz:
closing in favor of https://github.com/dimagi/commcare-hq/pull/17331

PR #17141, opened 2017-07-22
Don't put multimedia files in redis

emord:
@dimagi/scale-team @snopoke similar to https://github.com/dimagi/commcare-hq/pull/17124 we are still putting the media files in redis instead of serving from riak


emord:
May want to deploy this before Monday to NIC. found by `redis-cli --bigkeys`

PR #17148, opened 2017-07-24
save ledgers in django

snopoke:
Since we deployed [Atomic Save Processed Models](https://github.com/dimagi/commcare-hq/pull/17105) we've seen a lot of ledger saving errors (https://sentry.io/dimagi/commcarehq/issues/314820124/)

This is because the ledger saving code is still using a SQL function which get's called outside of the transaction that is saving the case (because of how pl_proxy works) so if the case didn't already exist in the DB then the ledger won't be able to see it until the other transaction is committed.

buddy @orangejenny 
cc @dimagi/scale-team


corehq/form_processor/backends/sql/dbaccessors.py
```diff
@@ -401,10 +401,14 @@ def hard_delete_forms(domain, form_ids, delete_attachments=True):
...
-            results = fetchall_as_namedtuple(cursor)
-            deleted_count = sum([result.deleted_count for result in results])
+        db_form_ids = split_list_by_db_partition(form_ids)
+        deleted_count = 0
+        for db_name, form_ids in db_form_ids.items():
+            # cascade should delete the attachments and operations
+            _, deleted_models = XFormInstanceSQL.objects.using(db_name).filter(
+                domain=domain, form_id__in=form_ids
+            ).delete()
+            deleted_count += deleted_models.get(XFormInstanceSQL._meta.label, 0)
```
snopoke: Also moved this to Django which was the only other non-django query happening `save_processed_models`

corehq/form_processor/backends/sql/dbaccessors.py
```diff
@@ -537,7 +541,7 @@ def save_new_form(form):
             operation.form = form
 
         try:
-            with transaction.atomic(using=db_name):
+            with transaction.atomic(using=db_name, savepoint=False):
```
snopoke: We don't need savepoints since if anything fails we want the whole transaction to fail.

corehq/form_processor/backends/sql/dbaccessors.py
```diff
@@ -1146,28 +1150,31 @@ def get_ledger_value(case_id, section_id, entry_id):
             raise LedgerValueNotFound
 
     @staticmethod
-    def save_ledger_values(ledger_values, deprecated_form=None):
-        if not ledger_values:
+    def save_ledger_values(ledger_values, stock_result=None):
```
benrudolph: was this param just always wrong?

corehq/form_processor/backends/sql/dbaccessors.py
```diff
@@ -1146,28 +1150,31 @@ def get_ledger_value(case_id, section_id, entry_id):
...
+                    LedgerTransaction.objects.using(db_name).filter(
+                        case_id__in=case_ids,
+                        form_id=stock_result.xform.form_id
+                    ).delete()
+
+            for ledger_value in ledger_values:
+                db_name = get_db_alias_for_partitioned_doc(ledger_value.case_id)
+                transactions_to_save = ledger_value.get_live_tracked_models(LedgerTransaction)
+
+                with transaction.atomic(using=db_name, savepoint=False):
```
benrudolph: probably don't need any of these since all the dbs are already in a transaction i believe. couldn't hurt though

benrudolph (approved):
nice, thanks, sorry about that. surprised tests didn't catch it

corehq/form_processor/backends/sql/dbaccessors.py
snopoke: These can get called from outside of `save_processed_models` so still good to have these here.

corehq/form_processor/backends/sql/dbaccessors.py
snopoke: no but it didn't have enough info so passed in the full stock result so that I could get the relevant case IDs affected by the deprecated form

emord:
merged master to get tests to rerun with fixed migration

snopoke:
thanks

emord:
```
FAIL: corehq.apps.commtrack.tests.test_xml:CommTrackOTATest.test_ota_basic
  File "/mnt/commcare-hq-ro/corehq/util/test_utils.py", line 267, in inner
  File "/mnt/commcare-hq-ro/corehq/util/test_utils.py", line 238, in __call__
  File "/mnt/commcare-hq-ro/corehq/util/test_utils.py", line 253, in call_with_settings
    fn(*args, **kwargs)
  File "/mnt/commcare-hq-ro/corehq/util/test_utils.py", line 234, in fn_with_pre_and_post
    self.fn(*args, **kwargs)
  File "/mnt/commcare-hq-ro/corehq/apps/commtrack/tests/test_xml.py", line 100, in test_ota_basic
    util.get_ota_balance_xml(self.domain, self.user)[0],
  File "corehq/ex-submodules/casexml/apps/case/tests/util.py", line 99, in check_xml_line_by_line
    test_case.assertEqual(expected_lines[i], actual_lines[i])
AssertionError: '<ns0:balance xmlns:ns0="http://commcarehq.org/ledger/v1" xmlns="http://openrosa.org/http/response" date="2017-07-24T16:08:02.068895Z" entity-id="6d052c0e0064493aa949899adc7a7adf" section-id="stock">' != '<ns0:balance xmlns:ns0="http://commcarehq.org/ledger/v1" xmlns="http://openrosa.org/http/response" date="2017-07-24T16:08:02.117148Z" entity-id="6d052c0e0064493aa949899adc7a7adf" section-id="stock">'
```

corehq/form_processor/backends/sql/dbaccessors.py
```diff
@@ -506,8 +506,9 @@ def soft_delete_forms(domain, form_ids, deletion_date=None, deletion_id=None):
     @transaction.atomic
     def _archive_unarchive_form(form, user_id, archive):
         from casexml.apps.case.xform import get_case_ids_from_form
+        from corehq.form_processor.parsers.ledgers.form import get_case_ids_from_stock_transactions
         form_id = form.form_id
-        case_ids = list(get_case_ids_from_form(form))
+        case_ids = list(get_case_ids_from_form(form) | get_case_ids_from_stock_transactions(form))
```
snopoke: This was causing a test to fail: https://github.com/dimagi/commcare-hq/blob/b98ec3f6d8063ae65e06c51f95e4fbefd23e24c8/corehq/apps/commtrack/tests/test_xml.py#L529

But it was also failing for me on master so not sure how it every passed

corehq/form_processor/backends/sql/dbaccessors.py
millerdev: @snopoke I'm trying to understand why `savepoint=False` was added here rather than removing `with transaction.atomic(...)`. Do you remember?

Context: this function has a `@transaction.atomic` decorator (with the default `savepoint=True`), so it seems this inner transaction block is not necessary. Also `savepoint=False` has this somewhat confusing caveat:

> This option should only be used if the overhead of savepoints is noticeable. It has the drawback of breaking the error handling described above. [[source](https://docs.djangoproject.com/en/1.11/topics/db/transactions/#django.db.transaction.atomic)]

snopoke: I think the decorator on the function shouldn't be there (it's not even on the write database).

I think the atomic context here is because this function can be called by itself (in the case of error handling) or as part of normal form processing (in which case there is a parent transaction). Fine to move the atomicity out of the function and handle it differently in the error case.

PR #17157, opened 2017-07-24
re-process incomplete submissions

snopoke:
Not yet complete but opening for review

buddy @orangejenny 
cc @dimagi/scale-team


corehq/form_processor/reprocess.py
```diff
@@ -0,0 +1,146 @@
...
+            if stock_result:
+                assert stock_result.populated
+
+            cases = case_stock_result.case_models
+            if should_use_sql_backend(form.domain):
+                cases = _filter_already_processed_cases(form, cases)
+                for case in cases:
+                    CaseAccessorSQL.save_case(case)
+
+                ledgers = _filter_already_processed_ledgers(form, stock_result.models_to_save)
```
orangejenny: Is `stock_result` guaranteed to exist? The `if stock_result` a few lines up makes me think it might be `None` and error here.

orangejenny: nm, just realized this code is moved, not new

orangejenny:
I don't have much context in this area, but this looks reasonable to me.

corehq/form_processor/reprocess.py
```diff
@@ -0,0 +1,146 @@
...
+        form.doc_type = 'XFormInstance'
+
+    form.initial_processing_complete = True
+    form.problem = None
+
+    cache = FormProcessorInterface(form.domain).casedb_cache(
+        domain=form.domain, lock=True, deleted_ok=True, xforms=[form]
+    )
+    with cache as casedb:
+        case_stock_result = SubmissionPost.process_xforms_for_cases([form], casedb)
```
calellowitz: there appear to be a few cases where this can result in an exception, anything that would trigger `_handle_known_error` of `SubmissionPost` could create a stub that needs to be reprocessed but raise an exception here. additionally that should result in a form where `form.initial_processing_complete` remains `False` and the state becomes `error`

corehq/form_processor/reprocess.py
```diff
@@ -0,0 +1,146 @@
...
+    form = _get_form(form_id)
+    if domain and form.domain != domain:
+        raise Exception('Form not found')
+    return reprocess_xform_error(form)
+
+
+def _reprocess_form(form):
+    # reset form state prior to processing
+    if should_use_sql_backend(form.domain):
+        form.state = XFormInstanceSQL.NORMAL
```
calellowitz: I think some errors and duplicate forms could end up with `UnfinishedSubmissionStubs` which this function does not account for

calellowitz (changes requested):
Added a few questions/comments

corehq/form_processor/reprocess.py
snopoke: I don't think so - at least not in the current code. The `UnfinishedSubmissionStubs` are only created for 'NORMAL` forms

corehq/form_processor/reprocess.py
snopoke: If there was an error raised when calling `process_xforms_for_cases` during the initial form processing then there wouldn't be a `stub`. But probably good to catch exceptions anyway.

corehq/form_processor/reprocess.py
snopoke: thanks, not sure why those checks were there since it's never None at this stage: 1052425

calellowitz (approved):
you are good on both comments. i got mixed up between calls to `self.save_processed_models` and `self.instance.save_processed_models`

snopoke:
@calellowitz made some changes - most notably to rebuild cases / ledgers that get updated. I think this is ready to hook up to a task / mgmt command.

snopoke:
@calellowitz This is ready to go. I want to do some manual testing before hooking it up to a task.

calellowitz (approved):
looks good to me. thanks for pushing this through

corehq/form_processor/reprocess.py
```diff
@@ -0,0 +1,206 @@
...
+    parent case has been created).
+
+    See http://manage.dimagi.com/default.asp?250459
+    :param form_id: ID of the error form to process
+    """
+    if not form:
+        raise Exception('Form with ID {} not found'.format(form.form_id))
+
+    if not form.is_error:
+        raise Exception('Form was not an error form: {}={}'.format(form.form_id, form.doc_type))
```
millerdev: Would it make sense to raise a custom subclass of `Exception` so it can be reliably caught without possibly masking bugs?

Also noticed several places below where `Exception` is raised.

PR #17159, opened 2017-07-24
dont use cached object

emord:
Follow up on the other PRs. CachedObject comes from a time when we had to pull attachments from couch instead of riak. We had some issues with redis growing a lot and it may have been putting binary attachments into redis as a cache. This removes CachedObject except for case attachments because that uses a CachedIMage class that caches thumbnails

@dimagi/scale-team @nickpell


benrudolph (approved):
👏 hooray for removing more instances of CachedObject

PR #17160, opened 2017-07-24
Abt report: Fix custom UCR expression for malformed submissions

NoahCarnahan:
This is a change to a custom UCR expression used by an abt report.

Some form submissions are missing an expected hidden value used to determine language, but we can use the domain in this case instead.


NoahCarnahan:
@biyeun

PR #17163, opened 2017-07-24
Formplayer clear user data

benrudolph:
@ctsims @wpride this is the front end to wipe the the user db / application db. refactored the settings so it's easier to add settings in the future. If you're user is `brudolph@dimagi.com` and you're logged in as `ben`, then it will wipe everything in the dir: `<domain>/brudolph@dimagi.com_ben` which is where the user db for `ben` lives as well as all the apps. it will not wipe the user db for `brudolph@dimagi.com` or other dbs of logged in as users (`brudolph@dimagi.com_test` for example). in order to do that, you need to login as that user.

screen shot 2017-07-24 at 1 37 31 pm
screen shot 2017-07-24 at 1 37 27 pm

broken down by commit, but can probably review all at once

needs: https://github.com/dimagi/formplayer/pull/440

cc: @proteusvacuum

### DIFF
--- a/corehq/apps/cloudcare/static/cloudcare/js/formplayer/app.js
+++ b/corehq/apps/cloudcare/static/cloudcare/js/formplayer/app.js
@@ -473,6 +473,12 @@ FormplayerFrontend.on('refreshApplication', function(appId) {
     });
 });
 
+/**
+ * clearUserData
+ *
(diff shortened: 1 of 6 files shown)


proteusvacuum:
Woo

PR #17166, opened 2017-07-24
App builder v2: remove none and open-other case actions

orangejenny:
https://manage.dimagi.com/default.asp?258598

This will make all forms in case list menus either open or update a case, even if they don't actually save any case properties.

@snopoke / @biyeun / @dannyroberts 

before
screen shot 2017-07-24 at 4 14 18 pm

after
screen shot 2017-07-24 at 4 18 26 pm


snopoke (commented):
I haven't thought through all the implications of this but it seems OK.

orangejenny:
@snopoke Fair. Tested out that forms currently set to "register a case for a different module" change to "update a case" without issue, haven't been able to think of anything else to specifically test before releasing. Going to merge into today's deploy.

PR #17169, opened 2017-07-24
do not allow ampersand

benrudolph:
interrupt @mkangia buddy @proteusvacuum 

https://manage.dimagi.com/default.asp?258380

### DIFF
--- a/corehq/apps/users/views/mobile/users.py
+++ b/corehq/apps/users/views/mobile/users.py
@@ -733,6 +733,10 @@ class MobileWorkerListView(JSONResponseMixin, BaseUserSettingsView):
             return {
                 'error': _(u'Username {} cannot contain "@".').format(username)
             }
+        if '&' in username:
+            return {
+                'error': _(u'Username {} cannot contain "&".').format(username)
+            }
         if ' ' in username:
             return {
                 'error': _(u'Username {} cannot contain '


orangejenny:
I'm surprised that we don't just limit usernames to word characters. Any idea why not?

benrudolph:
no not sure, agree would be simpler

orangejenny:
@ctsims do you know if mobile worker usernames need to support special characters?

Seems straightforward to make the creation restrictive, and then loosen it if we get complaints.

proteusvacuum:
I think we would also want to include `.`, as many usernames employ that. 
Why did you choose to do this instead of URL encoding the login-as URL? That seems like a safer, and less restrictive thing to do.

benrudolph:
@proteusvacuum initially i did try to do the encoding, which is not as straightforward as one might expect, because we already do encode the url. we actually have to double encode the username so hte & isn't interpreted as a param break, and then manually decode in the HQ code. after doing this, i got an error in the commcare engine. i didn't think it was worth the time to try and fix that so i just went with this approach, and we already have some precedent for doing it this way

proteusvacuum:
Well, I just tried this, and it turns out that HQ doesn't send back a well formed XML response if the username has an `&` in it (even if you properly urlencode the url).

screenshot from 2017-07-25 20-01-00

I still think this is somewhat of a workaround to fixing the actual problem, but I guess this is the easiest way... 

What do we do with users who currently have `&` in their username?

proteusvacuum:
(I also tried using other xml special characters like `<` but they weren't allowed, however it did fail silently and didn't give me a useable error message when I did that)

benrudolph:
thanks for giving it a go, that malformed xml makes sense as to why the commcare engine failed to parse it

PR #17177, opened 2017-07-25
Iterate StreamingBody request body

kaapstorm:
[FB 258588](https://manage.dimagi.com/default.asp?258588), [Error on Sentry](https://sentry.io/dimagi/commcarehq/issues/313730489/).

This change addresses a bug in the code that allows users to download "data files", which are essentially just blobs, uploaded to the project space by an admin.

Originally this code returned a data file in an HttpResponse, but I changed that to a StreamingHttpResponse for the sake of the resources required to serve a large response, and the experience of the user.

I had a choice to fix this bug by reverting back to returning an HttpResponse, or to make the blob returned by S3BlobDB iterable. I went with the second option because I would prefer to return a StreamingHttpResponse.

I'm not sure whether my implementation is the most elegant way to do that. And the chunk size I chose is arbitrary. Open to suggestions.

@esoergel


corehq/apps/export/views.py
```diff
@@ -1569,10 +1569,22 @@ class DataFileDownloadDetail(BaseProjectDataView):
...
+                yield chunk
+
+        def get_blob_iterator(blob_):
+            return blob_ if hasattr(blob_, '__iter__') else blob_iterator(blob_)
+
         try:
             data_file = DataFile.objects.filter(domain=self.domain).get(pk=kwargs['pk'])
             blob = data_file.get_blob()
-            response = StreamingHttpResponse(blob, content_type=data_file.content_type)
+            response = StreamingHttpResponse(get_blob_iterator(blob), content_type=data_file.content_type)
```
benrudolph: here's an example of a slightly easier way to do what you're doing in case useful: https://github.com/dimagi/commcare-hq/blob/master/corehq/ex-submodules/soil/__init__.py#L297

kaapstorm: Oh nice! That's what I was looking for.

corehq/apps/export/tests/test_export_views.py
```diff
@@ -42,11 +58,60 @@ def setUpClass(cls):
...
+        cls.data_file.save()
+
+    @classmethod
+    def tearDownClass(cls):
+        super(DataFileDownloadDetailTest, cls).tearDownClass()
+        cls.data_file.delete()
+        _db.pop()
+
+    def setUp(self):
+        super(DataFileDownloadDetailTest, self).setUp()
```
esoergel: nit: you can drop this (and the ones in `ExportViewTest`)

esoergel (approved):

corehq/apps/export/tests/test_export_views.py
kaapstorm: lol, thank you.

PR #17182, opened 2017-07-25
Remove cached last claim

proteusvacuum:
If the user claims a case, closes the claim case (through a form submission, for example), then tries to claim the case again, it will fail, even though that's a valid operation. 
I'm not sure if there was a specific reason for this cache... @kaapstorm? There was no test for it. 

FYI @wspride 
@benrudolph buddy. 
Noticed here: https://manage.dimagi.com/default.asp?258053#1369469

Also, I'm confused about why the screenshot in that ticket says "You have already claimed that **case**" when the error message mentions the case type (in that instance, it should say "You have already claimed that **voucher**")

### DIFF
--- a/corehq/apps/ota/views.py
+++ b/corehq/apps/ota/views.py
@@ -163,17 +163,13 @@ def claim(request, domain):
     """
     as_user = request.POST.get('commcare_login_as', None)
     restore_user = get_restore_user(domain, request.couch_user, as_user)
-    cache = get_redis_default_cache()
 
     case_id = request.POST.get('case_id', None)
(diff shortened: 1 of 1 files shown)


benrudolph (approved):
i was confused why it was necessary to cache when i looked at this last

kaapstorm:
Yeah, @proteusvacuum I don't remember the reason for the cache ... did we think that people would try to claim the same case over and over ... a lot? I don't really get it either.

(Regarding "You have already claimed that case": It will fall back to "case" if the case type is not passed in the POST parameters.)

PR #17183, opened 2017-07-25
Create all build files when releasing app

snopoke:
This is to avoid having to create them on demand which can cause issues if you have a lot of users all updating at the same time.


orangejenny:
I'm a little surprised this is done on release instead of on build. I'd expect people would usually want to download and test out their app before they mark it as released.

corehq/apps/app_manager/views/releases.py
```diff
@@ -170,6 +171,8 @@ def release_build(request, domain, app_id, saved_app_id):
     app_post_release.send(Application, application=saved_app)
 
     if is_released:
+        if saved_app.build_profiles and domain_has_privilege(domain, privileges.BUILD_PROFILES):
```
emord: Is it possible to test if all the builds have already been created? although I suppose not many have build profiles

emord (approved):
@orangejenny I think since this is specifically for apps with multiple profiles it's reasonable that someone would test one profile but not all of them (4 for icds)

snopoke:
The task does check if they've been created by looking for the
`profile.xml` file for each build profile

PR #17185, opened 2017-07-25
Fix case list breaking on language change

orangejenny:
https://manage.dimagi.com/default.asp?258874

Side effect of https://github.com/dimagi/commcare-hq/pull/17135 (fyi @biyeun )

Should go in today's deploy.

@snopoke / @benrudolph


orangejenny:
Also reported in https://manage.dimagi.com/default.asp?258878

PR #17186, opened 2017-07-25
Enikshay 2B case migration

esoergel:
https://docs.google.com/spreadsheets/d/1GFpMht-C-0cMCQu8rfqQG9lgW9omfYi3y2nUXHR8Pio/edit#gid=0
https://trello.com/c/BgpHGGiT/166-case-reconciliation-for-11-2b-migration
@sheelio can you do a first pass on code review?


custom/enikshay/two_b_release_1/management/commands/enikshay_2b_case_properties.py
```diff
@@ -0,0 +1,496 @@
+"""
```
esoergel: Don't let github fool you - this file matters

esoergel:
@kaapstorm 
@nickpell @proteusvacuum FYI

custom/enikshay/two_b_release_1/management/commands/enikshay_2b_case_properties.py
```diff
@@ -0,0 +1,496 @@
...
+            'referred_outside_enikshay_by_id': person.get_case_property('date_by_id'),
+        }
+        if episode:
+            props.update({
+                'current_episode_type': episode.get_case_property('episode_type'),
+                'alcohol_history': episode.get_case_property('alcohol_history'),
+                'alcohol_deaddiction': episode.get_case_property('alcohol_deaddiction'),
+                'tobacco_user': episode.get_case_property('tobacco_user'),
+                'occupation': episode.get_case_property('occupation'),
+                'nikshay_id': episode.get_case_property('nikshay_id'),
```
proteusvacuum: @sheelio to confirm. I thought this wasn't happening now? if it is, I think also adding this as the `external_id` of the person case would be useful. If we are doing this, we should be consistent and also add this for all private sector patients too.

custom/enikshay/two_b_release_1/management/commands/enikshay_2b_case_properties.py
```diff
@@ -0,0 +1,496 @@
...
+                'phone_number_other': episode.get_case_property('phone_number_other'),
+            })
+
+        phone_number = person.get_case_property('phone_number')
+        if phone_number:
+            # TODO check if it already begins with 91 or has a certain length?
```
proteusvacuum: `phone_number` also begins with a 0, so this will result in an invalid number. I would suggest parsing this with the `phonenumbers` library to be sure.

custom/enikshay/two_b_release_1/management/commands/enikshay_2b_case_properties.py
```diff
@@ -0,0 +1,496 @@
...
+            if 'dto' in ancestors_by_type:
+                props['dto_name'] = ancestors_by_type['dto'].name
+                props['dto_id'] = ancestors_by_type['dto'].location_id
+
+        return CaseStructure(
+            case_id=person.case_id,
+            walk_related=False,
+            attrs={
+                "create": False,
+                "owner_id": person.owner_id or '-',
```
proteusvacuum: why do you update the owner id?

custom/enikshay/two_b_release_1/management/commands/enikshay_2b_case_properties.py
```diff
@@ -0,0 +1,496 @@
...
+    def migrate_occurrence(self, occurrence, episodes):
+        self.total_occurrences += 1
+        episodes = [(case.opened_on, case) for case in episodes
+                    if case.type == CASE_TYPE_EPISODE
+                    and any([index.referenced_id == occurrence.case_id for index in case.indices])
+                    and case.dynamic_case_properties().get('episode_type') == "confirmed_tb"]
+        if not episodes:
+            return None
+
+        episode = max(episodes)[1]  # Most recently opened episode for the occurrence
```
proteusvacuum: won't this also return closed episodes?

custom/enikshay/two_b_release_1/management/commands/enikshay_2b_case_properties.py
```diff
@@ -0,0 +1,496 @@
...
+            'diagnosis_test_type_label': TEST_TO_LABEL.get(test_type, ""),
+        }
+
+        treatment_status = episode.get_case_property('treatment_status')
+        if treatment_status == 'second_line_treatment':
+            props['treatment_status'] = 'initiated_second_line_treatment'
+        elif treatment_status == 'yes_phi':
+            props['treatment_status'] = 'initiated_first_line_treatment'
+        elif treatment_status == 'yes_private':
+            props['treatment_status'] = 'initiated_outside_facility'
```
proteusvacuum: It looks like `treatment_status` is used in reports in some places. Is this something that is expected? Could be an error in the datasource: https://github.com/dimagi/commcare-hq/blob/master/custom/enikshay/ucr/data_sources/episode.json#L5880-L5888, but otherwise I think you might be overwriting that property here.

custom/enikshay/two_b_release_1/management/commands/enikshay_2b_case_properties.py
```diff
@@ -0,0 +1,496 @@
...
+            'referred_to_name': prop('referred_to_location_name'),
+            'referred_by_name': prop('referred_by'),
+            'referral_rejection_reason_other_detail': prop('reason_for_refusal_other_detail'),
+            'referral_rejection_reason': prop('reason_for_refusal'),
+            'referral_closed_date': prop('acceptance_refusal_date'),
+            'accepted_by_name': prop('phi'),
+        }
+
+        if occurrences:
+            occurrence = max((case.opened_on, case) for case in occurrences)[1]
```
proteusvacuum: looks like you calculate this for each referral and each trail. might be cacheable somewhere?

custom/enikshay/two_b_release_1/management/commands/enikshay_2b_case_properties.py
```diff
@@ -0,0 +1,496 @@
...
+        self.total_secondary_owners += 1
+        if occurrences:
+            occurrence = max((case.opened_on, case) for case in occurrences)[1]
+            index_kwargs = {'indices': [CaseIndex(
+                occurrence,
+                identifier='host',
+                relationship=CASE_INDEX_EXTENSION,
+                related_type=CASE_TYPE_OCCURRENCE,
+            )]}
+        else:
```
proteusvacuum: if this has no index, won't this case just be a stub with no information attached to it? Might be worth not creating it since it won't do anything, and I assume also won't show up anywhere.

custom/enikshay/two_b_release_1/tests.py
```diff
@@ -0,0 +1,247 @@
...
+        self.project = Domain(name=self.domain)
+        self.project.save()
+        self.location_types, self.locations = setup_enikshay_locations(self.domain)
+        self.factory = CaseFactory(domain=self.domain)
+        self.cases = self.setup_cases()
+
+    def tearDown(self):
+        self.project.delete()
+
+    def setup_cases(self):
```
proteusvacuum: curious why you didn't use the `ENikshayCaseStructureMixin`?

proteusvacuum (changes requested):
Wowza. Nice.

custom/enikshay/two_b_release_1/management/commands/enikshay_2b_case_properties.py
sheelio: (y)

custom/enikshay/two_b_release_1/management/commands/enikshay_2b_case_properties.py
sheelio: That's correct - we should remove this line.

custom/enikshay/two_b_release_1/management/commands/enikshay_2b_case_properties.py
```diff
@@ -0,0 +1,496 @@
...
+
+    def migrate_person(self, person, occurrences, episodes):
+        self.total_persons += 1
+        occurrence = self.get_open_occurrence(occurrences)
+
+        if occurrence:
+            episodes = [case for case in episodes
+                        if not case.closed and case.type == CASE_TYPE_EPISODE
+                        and any([index.referenced_id == occurrence.case_id for index in case.indices])
+                        and case.dynamic_case_properties().get('episode_type') == "confirmed_tb"]
```
sheelio: @esoergel 

I don't think you need to filter on confirmed_tb FWIW (this migration will still work even for episodes that are presumptive_tb

custom/enikshay/two_b_release_1/management/commands/enikshay_2b_case_properties.py
sheelio: We want to set it to "-" if its not set (though I'm not sure that is possible).

custom/enikshay/two_b_release_1/management/commands/enikshay_2b_case_properties.py
sheelio: We're updating all occurrences (even older occurrence that have closed episodes) so that reports can be consistent.

custom/enikshay/two_b_release_1/management/commands/enikshay_2b_case_properties.py
```diff
@@ -0,0 +1,496 @@
...
+                "update": props,
+            },
+        )
+
+    def migrate_occurrence(self, occurrence, episodes):
+        self.total_occurrences += 1
+        episodes = [(case.opened_on, case) for case in episodes
+                    if case.type == CASE_TYPE_EPISODE
+                    and any([index.referenced_id == occurrence.case_id for index in case.indices])
+                    and case.dynamic_case_properties().get('episode_type') == "confirmed_tb"]
```
sheelio: I don't think we need the confirmed_tb filter here.

custom/enikshay/two_b_release_1/management/commands/enikshay_2b_case_properties.py
sheelio: I've chatted with Sarvesh about this, and we're creating new data sources for these migrated cases that should handle this.

sheelio: @esoergel 

We should actually set treatment_status to initiated_outside_rntcp if treatment_status = yes_private (after chatting with Sarvesh).  I've also updated the spec.

custom/enikshay/two_b_release_1/management/commands/enikshay_2b_case_properties.py
```diff
@@ -0,0 +1,496 @@
...
+
+        if treatment_status == 'yes_phi':
+            props['treatment_initiated'] = 'yes_phi'
+        elif treatment_status == 'yes_private':
+            props['treatment_initiated'] = 'yes_private'
+        elif treatment_status:
+            props['treatment_initiated'] = 'no'
+
+        if not episode.get_case_property('date_of_diagnosis'):
+            props['date_of_diagnosis'] = episode.get_case_property('date_reported')
```
sheelio: @esoergel 

I think we also need to blank out phone_number_other (according to the spec).

custom/enikshay/two_b_release_1/management/commands/enikshay_2b_case_properties.py
```diff
@@ -0,0 +1,496 @@
...
+            props['rft_dstb_diagnosis'] = 'private_ntm'
+        else:
+            props['rft_general'] = {
+                'diagnostic': 'diagnosis_dstb',
+                'diagnosis_dstb': 'diagnosis_dstb',
+                'follow_up_dstb': 'follow_up_dstb',
+                'diagnosis_drtb': 'diagnosis_drtb',
+            }.get(test.get_case_property('purpose_of_testing'), "")
+            props['rft_dstb_diagnosis'] = test.get_case_property('diagnostic_test_reason')
+            props['rft_dstb_followup'] = test.get_case_property('follow_up_test_reason')
```
sheelio: @esoergel 

I added one more row (row 14) in the doc for episode_case_id to be set

custom/enikshay/two_b_release_1/management/commands/enikshay_2b_case_properties.py
sheelio: (Agree with Farid)

sheelio:
@esoergel 

This looks great pending a few comments I added (and sadly some changes to the spec based on discussion with reports team today).

custom/enikshay/two_b_release_1/management/commands/enikshay_2b_case_properties.py
```diff
@@ -0,0 +1,496 @@
...
+        return ancestors_by_type
+
+    def migrate(self):
+        person_ids = self.get_relevant_person_case_ids()
+        persons = self.get_relevant_person_case_sets(person_ids)
+        for person in with_progress_bar(persons, len(person_ids)):
+            self.migrate_person_case_set(person)
+
+    def get_relevant_person_case_ids(self):
+        location_owners = self.dto.get_descendants(include_self=True).location_ids()
```
esoergel: I definitely could do that - made those two changes too far apart to notice, thanks!

custom/enikshay/two_b_release_1/management/commands/enikshay_2b_case_properties.py
```diff
@@ -0,0 +1,496 @@
...
+                              CASE_TYPE_REFERRAL: 'referrals'}
+            for case in self.accessor.get_reverse_indexed_cases(
+                    [person_id for person_id in all_persons]):
+                bucket = type_to_bucket.get(case.type, None)
+                if bucket:
+                    for index in case.indices:
+                        if index.referenced_id in all_persons:
+                            getattr(all_persons[index.referenced_id], bucket).append(case)
+                            referrals_and_occurrences_to_person[case.case_id] = index.referenced_id
+                            break
```
esoergel: In this block I'm trying to match up occurrence and referral cases with their corresponding person cases (in bulk, which is not super straightforward).  My assumption is that there is at most one person case associated with occurrences and referrals.

custom/enikshay/two_b_release_1/management/commands/enikshay_2b_case_properties.py
esoergel: This is to get the current episode case in order to pull properties to save to the person case.  Is that still true for that?

There's a different place where this gets all the episode cases to update.

custom/enikshay/two_b_release_1/management/commands/enikshay_2b_case_properties.py
esoergel: cool library - I hadn't seen that before.  Does this look right to you?
```python
In [1]: import phonenumbers

In [2]: pn = phonenumbers.parse('1234567890', 'IN')

In [3]: phonenumbers.format_number(pn, phonenumbers.PhoneNumberFormat.E164)
Out[3]: u'+911234567890'
```

custom/enikshay/two_b_release_1/management/commands/enikshay_2b_case_properties.py
esoergel: Same comment as above - is this also the case when retrieving the most recently opened episode for the occurrence?

custom/enikshay/two_b_release_1/management/commands/enikshay_2b_case_properties.py
proteusvacuum: Yup!  you'll need to remove the `+` for manually for messaging though I think. Always confusing to me why we don't include the `+`

custom/enikshay/two_b_release_1/management/commands/enikshay_2b_case_properties.py
esoergel: updated

custom/enikshay/two_b_release_1/management/commands/enikshay_2b_case_properties.py
esoergel: I thought that was optional?  I was scared of deleting it preemptively since it's needed for the person migration.  It should be fine to add if it's actually needed, though.

custom/enikshay/two_b_release_1/management/commands/enikshay_2b_case_properties.py
esoergel: Sounds good

custom/enikshay/two_b_release_1/tests.py
esoergel: Mainly just wanted to have more control over some of the properties set in the cases and how they related to each other.  I probably could've used it, but it felt like I would've ended up needing to override a lot of stuff anyways, and this felt more straightforward.

kaapstorm:
Looks good. (And I discovered that `max` can take tuples without necessarily defining a `key` function.)

custom/enikshay/two_b_release_1/management/commands/enikshay_2b_case_properties.py
gcapalbo: we've adopted the convention of only storing digits in phone number fields (without +, which is a convention set by the users model a long time ago), but it actually doesn't matter to the framework. to be consistent with the other migrations it's probably good to remove the leading +. [this function](https://github.com/dimagi/commcare-hq/blob/a30445d014e7888a9c6db0c024db434f7e406953/corehq/apps/sms/util.py#L50) can be used to do it

custom/enikshay/two_b_release_1/management/commands/enikshay_2b_case_properties.py
```diff
@@ -317,20 +317,21 @@ def migrate_occurrence(self, occurrence, episodes):
...
+    def _get_last_episode_id(indices, episodes):
         def is_episode_of_occurrence(episode, occurrence_id):
             return any(index.referenced_id == occurrence_id for index in episode.indices)
 
-        occurrence_ids = [index.referenced_id for index in episode.indices
+        occurrence_ids = [index.referenced_id for index in indices
                           if index.referenced_type == CASE_TYPE_OCCURRENCE]
         occurrence_id = occurrence_ids[0] if occurrence_ids else None
+        return max((case.opened_on, case.case_id) for case in episodes
+                   if is_episode_of_occurrence(case, occurrence_id) and not case.closed)[1]
```
sheelio: @esoergel 
For this situation, I don't think we need to filter on opened cases (we just need the last opened episode)

custom/enikshay/two_b_release_1/management/commands/enikshay_2b_case_properties.py
sheelio: Ah ok - lets just leave it as is (don't remove it).

custom/enikshay/two_b_release_1/management/commands/enikshay_2b_case_properties.py
sheelio: Yup - I think for updating the person case, we don't really care whether the episode type was confirmed_tb or presumptive_tb.   All we care about is getting the open episode for the open occurrence.

custom/enikshay/two_b_release_1/management/commands/enikshay_2b_case_properties.py
sheelio: Yup, same deal (we don't care about filtering to confirmed_tb here either).

custom/enikshay/two_b_release_1/management/commands/enikshay_2b_case_properties.py
esoergel: That's perfect, thanks!

esoergel: That's perfect, thanks~

custom/enikshay/two_b_release_1/management/commands/enikshay_2b_case_properties.py
esoergel: done

custom/enikshay/two_b_release_1/management/commands/enikshay_2b_case_properties.py
esoergel: updated

custom/enikshay/two_b_release_1/management/commands/enikshay_2b_case_properties.py
esoergel: also updated

custom/enikshay/two_b_release_1/management/commands/enikshay_2b_case_properties.py
esoergel: added

custom/enikshay/two_b_release_1/management/commands/enikshay_2b_case_properties.py
esoergel: https://github.com/dimagi/commcare-hq/pull/17186/commits/1028786490d2fa188cf45db619d4aa9558f0414b

esoergel:
Okay, I think that should be all the feedback and additions.  Can I get a final pass?

custom/enikshay/two_b_release_1/management/commands/enikshay_2b_case_properties.py
```diff
@@ -239,8 +238,7 @@ def migrate_person(self, person, occurrences, episodes):
         if occurrence:
             episodes = [case for case in episodes
                         if not case.closed and case.type == CASE_TYPE_EPISODE
-                        and any([index.referenced_id == occurrence.case_id for index in case.indices])
-                        and case.dynamic_case_properties().get('episode_type') == "confirmed_tb"]
+                        and any([index.referenced_id == occurrence.case_id for index in case.indices])]
```
proteusvacuum: nit: I think `any` works a little faster with a generator since it doesn't compute all the values.

proteusvacuum (approved):

esoergel:
@sheelio are you signed off on this?

sheelio:
I'm signed off.  

I was chatting with Noah and Nick about prior migrations we've done and they recommended adding a __updated_by_migration case property to the cases that specifies the migration that updated them (i.e. something like mumbai_july31).  Is that easy to add, or do we have another way to track that?

esoergel:
This migration does set "case_version" to 20 on the person case, which will allow us to avoid migrating persons twice.  Something like that sounds like a good idea though just as a safeguard.  I'll merge this now as-is so we can run it without `--commit` ASAP, and add that in in a separate PR.

PR #17188, opened 2017-07-25
Prune touchforms api

dannyroberts:
got rid of two more unused cloudcare endpoints

### DIFF
--- a/corehq/apps/cloudcare/views.py
+++ b/corehq/apps/cloudcare/views.py
@@ -4,7 +4,6 @@ from xml.etree import ElementTree
 from django.conf import settings
 from django.urls import reverse
 from django.http import HttpResponseRedirect, HttpResponse, HttpResponseBadRequest, Http404
-from django.shortcuts import get_object_or_404
 from django.views.decorators.csrf import csrf_exempt
 from django.shortcuts import render
 from django.template.loader import render_to_string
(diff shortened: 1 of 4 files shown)


corehq/apps/cloudcare/views.py
```diff
@@ -577,37 +557,6 @@ def post(self, request, domain):
         })
 
 
-@cloudcare_api
-def render_form(request, domain):
```
benrudolph: we still use this from formplayer

benrudolph (commented):
one thing that needs to be added back but the others i think are good to go

corehq/apps/cloudcare/views.py
benrudolph: https://github.com/dimagi/formplayer/blob/master/src/main/java/services/impl/FormattedQuestionsServiceImpl.java#L32

benrudolph: (sorry maybe a red herring)

dannyroberts: we offlined, here's the convo:

> [1:32 PM] 
> Danny Roberts `grep -R 'cloudcare/api/render_form/' .` is showing up, like, a couple of requests from a year ago
> 
> 
> [1:32 PM] 
> Ben Rudolph hmm
> 
> 
> [1:32 PM] 
> Danny Roberts and `ag cloudcare_render_form` on the codebase shows up nothing
> 
> 
> [1:32 PM] 
> Ben Rudolph ah
> 
> 
> [1:33 PM] 
> i thought `/cloudcare/api/readable_questions`
> 
> 
> [1:33 PM] 
> was the same thing
> 
> 
> [1:33 PM] 
> guess it’s not

PR #17191, opened 2017-07-25
use cache as session

snopoke:
buddy @orangejenny 

As a side effect we also need to provide formplayer with a way to do auth based on the session key. This PR also adds an admin view for getting the internal user ID from a session key.

### DIFF
--- a/settings.py
+++ b/settings.py
@@ -154,7 +154,7 @@ MIDDLEWARE = [
     'corehq.apps.locations.middleware.LocationAccessMiddleware',
 ]
 
-SESSION_ENGINE = "django.contrib.sessions.backends.cached_db"
+SESSION_ENGINE = "django.contrib.sessions.backends.cache"
 
 # time in minutes before forced logout due to inactivity
 INACTIVITY_TIMEOUT = 60 * 24 * 14


benrudolph:
nice i will try and give this a go

benrudolph:
this currently returns the following locally:
```
"\x80\x02X<\x02\x00\x00ZTEzOGM5ZWMwYmYzZDExMzBhZTgwOGYxM2Q3YTRjNWZkNGIwYjdlNTp7Imxhc3RfcmVxdWVzdCI6IjIwMTctMDctMjdUMTg6NTc6MDQuMzY4NTg1WiIsImxhc3RfdmlzaXRlZF9kb21haW4iOiJhc3BhY2UiLCJfYXV0aF91c2VyX2lkIjoiMTIiLCJfYXV0aF91c2VyX2JhY2tlbmQiOiJkamFuZ28uY29udHJpYi5hdXRoLmJhY2tlbmRzLk1vZGVsQmFja2VuZCIsIndpemFyZF9jbG91ZF9jYXJlX2xvZ2luX3ZpZXciOnsic3RlcF9maWxlcyI6e30sInN0ZXAiOm51bGwsImV4dHJhX2RhdGEiOnt9LCJzdGVwX2RhdGEiOnt9LCJ2YWxpZGF0ZWRfc3RlcF9kYXRhIjp7fX0sIndzNHJlZGlzOm1lbWJlcm9mIjpbXSwiX2F1dGhfdXNlcl9oYXNoIjoiYjVmMTY3MjVlMjcwYzFlMzMwMGY1OTIzMDI4MzRhMzdhY2VjYmI5MiIsImRqYW5nb19sYW5ndWFnZSI6ImVuIn0=q\x01."
```

snopoke:
Is it not just base64 encoded?

benrudolph:
yeah it is, but i think the base64 string is pickled. which we can hack around, but definitely makes it more brittle:
```
ZTEzOGM5ZWMwYmYzZDExMzBhZTgwOGYxM2Q3YTRjNWZkNGIwYjdlNTp7Imxhc3RfcmVxdWVzdCI6IjIwMTctMDctMjdUMTg6NTc6MDQuMzY4NTg1WiIsImxhc3RfdmlzaXRlZF9kb21haW4iOiJhc3BhY2UiLCJfYXV0aF91c2VyX2lkIjoiMTIiLCJfYXV0aF91c2VyX2JhY2tlbmQiOiJkamFuZ28uY29udHJpYi5hdXRoLmJhY2tlbmRzLk1vZGVsQmFja2VuZCIsIndpemFyZF9jbG91ZF9jYXJlX2xvZ2luX3ZpZXciOnsic3RlcF9maWxlcyI6e30sInN0ZXAiOm51bGwsImV4dHJhX2RhdGEiOnt9LCJzdGVwX2RhdGEiOnt9LCJ2YWxpZGF0ZWRfc3RlcF9kYXRhIjp7fX0sIndzNHJlZGlzOm1lbWJlcm9mIjpbXSwiX2F1dGhfdXNlcl9oYXNoIjoiYjVmMTY3MjVlMjcwYzFlMzMwMGY1OTIzMDI4MzRhMzdhY2VjYmI5MiIsImRqYW5nb19sYW5ndWFnZSI6ImVuIn0=
```
is the base64 string. i don't know what the other bytes are

snopoke:
:angry: aarg. The redis cache lib pickles the data itself so the options now are:
1. Create a separate cache client that is only used by sessions (SESSION_CACHE_ALIAS) that uses JSON serialization (https://niwinz.github.io/django-redis/latest/#_pluggable_serializer)
2. HQ endpoint that formplayer uses to get session info.

Starting to think that maybe (2) is the better option

snopoke:
@benrudolph added an admin view to get user ID from the session key. It's currently allows insecure access. Thoughts on how to secure this?

emord:
Might be overengineering, but we could use django-otp in some way to have a time based password similar to two factor auth? could also limit it to internal IPs, but that might need formplayer to hit the webworkers directly if nginx masks it.

seems like a risky view to have to not have it locked down. Maybe worth having a very small flask app deployed with formplayer just for this.

benrudolph:
yea so i think we were depending on the "trusted" source method of authentication before. we can replicate that by having a secret token that hq and formplayer agree on and then we send that over. still seems fairly risky / not great

snopoke:
moved the session details view to https://github.com/dimagi/commcare-hq/pull/17360

snopoke:
This is now deployed to staging and working so we can re-schedule the deploy for prod.

PR #17192, opened 2017-07-25
Permissions.view_report

benrudolph:
@esoergel was this a typo? I didn't test it out locally, but given the error message and the commit i'm inclined to think it is (let me know if you think i should do some testing)
https://sentry.io/dimagi/commcarehq/issues/304203718/events/6433056443/
https://manage.dimagi.com/default.asp?258890
https://github.com/dimagi/commcare-hq/commit/06702fb7b
interrupt: @mkangia buddy: @proteusvacuum

### DIFF
--- a/corehq/apps/reports/views.py
+++ b/corehq/apps/reports/views.py
@@ -2008,11 +2008,6 @@ def mk_date_range(start=None, end=None, ago=timedelta(days=7), iso=False):
         return start, end
 
 
-def _can_view_report(domain, user, report_class):
-    return (user.has_permission(domain, Permissions.view_report)
-            or user.has_permission(domain, Permissions.view_report, data=report_class))
-
-
 @login_and_domain_required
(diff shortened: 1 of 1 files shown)


esoergel (changes requested):
@benrudolph This isn't a typo - there are two different ways a user may view a report - they have permission to view all reports (`Permissions.view_reports`), or they have permission to view individual reports (`Permissions.view_report(report_class)`).

It looks like `Permissions.view_report` also encapsulates `view_reports`, though, so not sure why I made this change.  Wanna offline?

esoergel:
Hmm, looking at this again, I think the correct way to do this is to use the string `'view_report'` instead of the attribute `Permissions.view_report`.  Then you should be able to revert my change too.  Sounds like it's worth testing, though.

esoergel (approved):

PR #17193, opened 2017-07-25
Kill careplans

dannyroberts:
I've heard some excitement around the prospect of this from some of you, and I did a little final research on prod data to validate. See also: https://trello.com/c/StQ0J7Ci/21-care-plans.

I left around a `CareplanModule` stub that's just a trigger for displaying a deprecation message, and removed all the code that did anything interesting with it.


snopoke (approved):
💓

orangejenny:
FWIW, this 500s when accessing an app that does contain a care plan module:
screen shot 2017-07-25 at 5 12 23 pm

dannyroberts:
Thanks, I fixed that by changing my strategy and filtering out careplan modules entirely so they never appear in the first place. Was also able to then delete a bunch more boilerplate code.

orangejenny:
Nice, I like the new strategy.

orangejenny (approved):

PR #17194, opened 2017-07-25
Don't cache get_cases

gcapalbo:
https://manage.dimagi.com/default.asp?258790

The caching was causing an issue where newly-created cases were not showing up until the cache expired, making the sms surveys for those cases fail. Since I think this endpoint is not being used by formplayer anymore (@benrudolph correct me if I'm wrong), and since touchforms only calls out to hq once at the beginning of an smsforms session, I think we can just remove the caching all together.

cc @dannyroberts @millerdev


benrudolph (approved):
KILL!

dannyroberts (approved):

PR #17195, opened 2017-07-25
Bump daily max auto-update limit for enikshay

gcapalbo:
Just doing this temporarily to help finish running some newly-created rules which have a lot of updates to make.

@millerdev


millerdev (approved):
Not sure if the test failure is related.
```
FAIL: corehq.apps.users.tests.test_user_model:UserDeviceTest.test_only_update_once_per_day
  File "/mnt/commcare-hq-ro/corehq/apps/users/tests/test_user_model.py", line 109, in test_only_update_once_per_day
    self.assertFalse(user.update_device_id_last_used(device, later))
AssertionError: True is not false
```

benrudolph:
should be fixed here: https://github.com/dimagi/commcare-hq/pull/17210

PR #17198, opened 2017-07-25
Je/fix queueing time

emord:
The lock expires in 4 minutes and 50 seconds, but the cutoff has always been 5 minutes. That means there could be  10 seconds of overlap in the queueing which is why I think we've been getting errors like https://sentry.io/dimagi/commcarehq/issues/309028178/

@gcapalbo I changed this to using serial_task. I think that's the preferred way we use to not have tasks step on each other right?

### DIFF
--- a/corehq/apps/userreports/tasks.py
+++ b/corehq/apps/userreports/tasks.py
@@ -210,7 +210,7 @@ def run_queue_async_indicators_task():
     queue_async_indicators.delay()
 
 
-@serial_task('queue-async-indicators', timeout=10 * 60, queue=settings.CELERY_PERIODIC_QUEUE, max_retries=0)
+@serial_task('queue-async-indicators', timeout=30 * 60, queue=settings.CELERY_PERIODIC_QUEUE, max_retries=0)
 def queue_async_indicators():
     oldest_indicator = AsyncIndicator.objects.order_by('date_queued').first()
     if oldest_indicator and oldest_indicator.date_queued:
--- a/corehq/apps/userreports/tasks.py
+++ b/corehq/apps/userreports/tasks.py
@@ -210,7 +210,7 @@ def compare_ucr_dbs(domain, report_config_id, filter_values, sort_column=None, s
 )
 def queue_async_indicators():
     start = datetime.utcnow()
-    cutoff = start + ASYNC_INDICATOR_QUEUE_TIME
+    cutoff = start + ASYNC_INDICATOR_QUEUE_TIME - timedelta(seconds=60)
     time_for_crit_section = ASYNC_INDICATOR_QUEUE_TIME.seconds - 10
 
     oldest_indicator = AsyncIndicator.objects.order_by('date_queued').first()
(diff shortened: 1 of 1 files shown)


emord:
buddy @nickpell

corehq/apps/userreports/tasks.py
```diff
@@ -204,37 +205,31 @@ def _run_report(backend_to_use):
...
-    run_every=crontab(minute="*/5"),
-    queue=settings.CELERY_PERIODIC_QUEUE,
-)
-def queue_async_indicators():
-    start = datetime.utcnow()
-    cutoff = start + ASYNC_INDICATOR_QUEUE_TIME
-    time_for_crit_section = ASYNC_INDICATOR_QUEUE_TIME.seconds - 10
+@periodic_task(run_every=crontab(minute="*/5"), queue=settings.CELERY_PERIODIC_QUEUE)
+def run_queue_async_indicators_task():
+    queue_async_indicators.delay()
```
snopoke: if `queue_async_indicators` takes longer to run than 5 mins won't this just keep queuing up tasks? Not sure that's a big issue unless it happens all the time.

snopoke (approved):
One question but otherwise looks good

corehq/apps/userreports/tasks.py
```diff
@@ -204,37 +205,31 @@ def _run_report(backend_to_use):
...
-        indicators_by_domain_doc_type = defaultdict(list)
-        for indicator in indicators:
-            # don't requeue anything htat's be queued in the past day
-            if not indicator.date_queued or indicator.date_queued < day_ago:
-                indicators_by_domain_doc_type[(indicator.domain, indicator.doc_type)].append(indicator)
-
-        for k, indicators in indicators_by_domain_doc_type.items():
-            now = datetime.utcnow()
-            if now > cutoff:
-                break
```
gcapalbo: do we still need something like this to make sure this doesn't run over the 10 minute timeout, or will this task always finish in under 10 minutes?

gcapalbo (approved):

gcapalbo:
> @gcapalbo I changed this to using serial_task. I think that's the preferred way we use to not have tasks step on each other right?

yup

corehq/apps/userreports/tasks.py
emord: I believe that max_retries on that task means that it will attempt to queue the task, fail because of a lock in serial_task and then no longer try again. @gcapalbo can you confirm that?

corehq/apps/userreports/tasks.py
emord: This should always finish, but I bumped the timeout to 30 min just in case

corehq/apps/userreports/tasks.py
gcapalbo: yeah when we use serial_task with max_retries=0, then even if a lot of tasks get queued up, they just stop running as soon as they start when they see that another of these tasks is currently running.

gcapalbo (approved):

PR #17199, opened 2017-07-25
enabling usercase on linked apps

calellowitz:
@orangejenny cc @mkangia

### DIFF
--- a/corehq/apps/app_manager/models.py
+++ b/corehq/apps/app_manager/models.py
@@ -6501,11 +6501,16 @@ def import_app(app_id_or_source, domain, source_properties=None, validate_source
                 m.valid_domains.append(domain)
                 m.save()
 
-    if not app.is_remote_app() and any(module.uses_usercase() for module in app.get_modules()):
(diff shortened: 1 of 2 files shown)


corehq/apps/app_manager/models.py
```diff
@@ -6501,13 +6501,18 @@ def import_app(app_id_or_source, domain, source_properties=None, validate_source
...
-    if not app.is_remote_app() and any(module.uses_usercase() for module in app.get_modules()):
-        from corehq.apps.app_manager.util import enable_usercase
-        enable_usercase(domain)
+    if not app.is_remote_app():
+        enable_usercase_if_necessary(app, domain)
 
     return app
 
 
+def enable_usercase_if_necessary(app, domain):
```
orangejenny: Nit: if passing app, no need to pass domain

orangejenny (approved):

mkangia (approved):
Would be a +1 if could add some tests as well.

PR #17203, opened 2017-07-26
Move compress into js block

proteusvacuum:
@benrudolph 
Hopefully this will fix this when sending emailed repeat record reports... 
https://sentry.io/dimagi/commcarehq/issues/312574104/events/6440062778/


benrudolph (approved):
will certainly be a good place to try. i hate these errors

PR #17206, opened 2017-07-26
set expires to 60s for latest queries

snopoke:
looking at the logs we were still getting nailed by queries for `media_profile.ccpr?latest=true`

Allowing that to be cached for 60s. I've patched this to NIC for now and it's working well. Might not be OK for general use though.... will have to think it through some more.

### DIFF
--- a/corehq/apps/app_manager/decorators.py
+++ b/corehq/apps/app_manager/decorators.py
@@ -61,7 +61,9 @@ def safe_cached_download(f):
         try:
             request.app = get_app(domain, app_id, latest=latest, target=target)
             response = f(request, *args, **kwargs)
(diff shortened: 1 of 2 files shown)


snopoke:
Merging for deploy

emord:
> Might not be OK for general use though.... will have to think it through some more.

Yeah. I explictily didn't do this because I didn't think it would be safe, but I think 60s is a good compromise.

From what I can tell it just means that it's possible that for 60 s a phone looking for updates may receive an out of date release. That seems fine to me since after a minute it will correct itself.

Also this will have no effect unless we enable nginx caching for prod (which I think we should, but with different settings)

PR #17212, opened 2017-07-26
Keep removing unused cloudcare views

dannyroberts:
I think this is it for unused cloudcare views now

### DIFF
--- a/corehq/apps/cloudcare/views.py
+++ b/corehq/apps/cloudcare/views.py
@@ -68,15 +68,6 @@ def default(request, domain):
     return HttpResponseRedirect(reverse('formplayer_main', args=[domain]))
 
 
-@use_legacy_jquery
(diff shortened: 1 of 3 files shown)


corehq/apps/cloudcare/views.py
```diff
@@ -59,25 +57,15 @@
...
+from corehq.form_processor.exceptions import XFormNotFound
 
 
 @require_cloudcare_access
 def default(request, domain):
     return HttpResponseRedirect(reverse('formplayer_main', args=[domain]))
 
 
-@use_legacy_jquery
-def insufficient_privilege(request, domain, *args, **kwargs):
```
benrudolph: i wonder what gets shown now if you don't have privs...

benrudolph (approved):
seems good to me

PR #17215, opened 2017-07-26
Revert "Revert "replace filtering by site_code to location_id""

emord:
Reverts dimagi/commcare-hq#17208

@lwyszomi We discussed this in a meeting today and this won't work for staging, but it does work on NIC and queries by id are much more efficient. From now on we'll only be testing on the NIC server

@sheelio


emord:
Ah. I see now. this is only for staging. The original PR was not on master. closing out

PR #17218, opened 2017-07-26
zapier tweaks

czue:
some small tweaks to our zapier integration. can be read by commit. the second commit is actually the only change that really matters.

i'm buddyless! @dannyroberts @proteusvacuum @kaapstorm since this is a little "integrationy"

### DIFF
--- a/corehq/apps/zapier/api/v0_5.py
+++ b/corehq/apps/zapier/api/v0_5.py
@@ -107,7 +107,7 @@ class ZapierCustomTriggerFieldFormResource(BaseZapierCustomFieldResource):
             ))
 
         for form_property in MAIN_FORM_TABLE_PROPERTIES:
-            if form_property.is_advanced:
+            if form_property.is_advanced or form_property.item.transform:
                 continue
             custom_fields.append(CustomField(
                 dict(
--- a/corehq/apps/zapier/views.py
+++ b/corehq/apps/zapier/views.py
@@ -19,6 +19,7 @@ from corehq.apps.zapier.consts import EventTypes
 from corehq import privileges
 from corehq.form_processor.exceptions import CaseNotFound
 from corehq.form_processor.interfaces.dbaccessors import CaseAccessors
+from dimagi.utils.web import json_response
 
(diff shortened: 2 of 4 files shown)


corehq/apps/zapier/api/v0_5.py
```diff
@@ -107,7 +107,7 @@ def obj_get_list(self, bundle, **kwargs):
             ))
 
         for form_property in MAIN_FORM_TABLE_PROPERTIES:
-            if form_property.is_advanced:
+            if form_property.is_advanced or form_property.item.transform:
```
czue: this fixes one of the issues highlighted [here](https://cdn.zapier.com/storage/photos/f109159ec8f2cc0cdde9f8ae86a762ab.png) (by removing the field that wasn't getting transformed anyways)

corehq/apps/zapier/views.py
```diff
@@ -69,7 +70,8 @@ def post(self, request, domain, *args, **kwargs):
         else:
             return HttpResponse(status=400)
 
-        return HttpResponse('OK')
+        # respond with the ID so that zapier can use it to unsubscribe
+        return json_response({'id': subscription.id})
```
dannyroberts: There's also a django built-in now called JsonRepsonse that that behaves pretty much the same. No need to change now, but that reminds me we should probably get rid of this (now superfluous) util.

dannyroberts (approved):

corehq/apps/zapier/views.py
dannyroberts: Actually, I guess ours does use a json handler that calls `to_json()` on an object if possible, so I'll keep it for now.

czue:
I see this is approved. do I merge my own PR now or should I be waiting for something else?

proteusvacuum:
👋 @czue 
You can merge your own branch if it gets approved and tests pass!

czue:
:wave: @proteusvacuum 

> You can merge your own branch if it gets approved and tests pass!

:open_mouth: i don't know if i can handle all this change.

PR #17219, opened 2017-07-26
Kill brudolph instances

benrudolph:
buddy: @proteusvacuum

### DIFF
--- a/corehq/apps/export/models/new.py
+++ b/corehq/apps/export/models/new.py
@@ -1370,7 +1370,7 @@ class ExportDataSchema(Document):
                     identifier,
                 )
             except Exception as e:
-                _soft_assert = soft_assert('{}@{}'.format('brudolph', 'dimagi.com'))
+                _soft_assert = soft_assert(notify_admins=True)
                 _soft_assert(False, 'Failed to process app {}. {}'.format(app._id, e))
                 continue
 
@@ -1390,7 +1390,7 @@ class ExportDataSchema(Document):
         try:
             current_schema = cls._reorder_schema_from_app(current_schema, app_id, identifier)
         except Exception as e:
-            _soft_assert = soft_assert('{}@{}'.format('brudolph', 'dimagi.com'))
+            _soft_assert = soft_assert(notify_admins=True)
             _soft_assert(False, 'Failed to process app during reorder {}. {}'.format(app_id, e))
 
         current_schema.domain = domain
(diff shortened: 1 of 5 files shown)


corehq/apps/export/views.py
```diff
@@ -2128,9 +2128,7 @@ def get(self, request, *args, **kwargs):
 
             except ResourceNotFound:
                 raise Http404()
-            except Exception as e:
-                _soft_assert = soft_assert('{}@{}'.format('brudolph', 'dimagi.com'))
-                _soft_assert(False, 'Failed to convert export {}. {}'.format(self.export_id, e))
```
benrudolph: this code isn't really used nearly as often as it once was, so feels fine to remove

corehq/ex-submodules/couchforms/models.py
```diff
@@ -239,15 +239,6 @@ def save(self, **kwargs):
...
                     raise
 
-    def xpath(self, path):
-        """
-        Evaluates an xpath expression like: path/to/node and returns the value
-        of that element, or None if there is no value.
-        """
-        _soft_assert = soft_assert(to='{}@{}'.format('brudolph', 'dimagi.com'))
-        _soft_assert(False, "Reference to xpath instead of get_data")
-        return safe_index(self, path.split("/"))
```
benrudolph: haven't got a soft assert for this in 1.5 years

proteusvacuum:
If I don't approve this PR, will you have to stay?

benrudolph:
shit, i think so 😲

corehq/apps/export/models/new.py
```diff
@@ -1390,7 +1390,7 @@ def generate_schema_from_builds(cls, domain, app_id, identifier, force_rebuild=F
         try:
             current_schema = cls._reorder_schema_from_app(current_schema, app_id, identifier)
         except Exception as e:
-            _soft_assert = soft_assert('{}@{}'.format('brudolph', 'dimagi.com'))
+            _soft_assert = soft_assert(notify_admins=True)
```
proteusvacuum: do people know what to do about these issues? is there a reason you soft assert to yourself here?

benrudolph: usually i treat it as a 500 to fix, but don't want it to crash for a user. but agree probably not that useful for people

benrudolph: ok changed to a log message so it shows up in sentry. https://github.com/dimagi/commcare-hq/pull/17219/commits/d2f97a66c00005d3b055307c80ada1c67e7297a0

these don't happen often (i think i've only received a couple). i think it's good to log somewhere that apps are failing to be processed by exports

proteusvacuum:
👎

PR #17222, opened 2017-07-26
Web apps permissions login as server side

dannyroberts:
fixes: https://manage.dimagi.com/default.asp?257563

Things I've specifically tested:
- Web Apps
  - Visit as web user, see all apps
  - Log-in-as restricted user, see only the one app
  - Click to view as original user, back to both
  - Reload the page while log-in-as'd, see only the one app
  - In any configuration, can enter forms, multimedia works
  - Can reload any page and be brought to the right place with the right context (i.e. works as you'd expect)
- App Preview
  - Still works, can do log-in-as
  - Note: if you log-in-as a user that shouldn't be able to see the app, it won't say anything in app preview

### DIFF
--- a/corehq/apps/cloudcare/static/cloudcare/js/formplayer/app.js
+++ b/corehq/apps/cloudcare/static/cloudcare/js/formplayer/app.js
@@ -172,15 +172,17 @@ FormplayerFrontend.on('startForm', function (data) {
 
 FormplayerFrontend.on("start", function (options) {
     var user = FormplayerFrontend.request('currentUser'),
-        savedDisplayOptions;
+        savedDisplayOptions,
+        self = this;
     user.username = options.username;
-    user.apps = options.apps;
     user.domain = options.domain;
     user.formplayer_url = options.formplayer_url;
     user.debuggerEnabled = options.debuggerEnabled;
     user.environment = options.environment;
     user.useLiveQuery = options.useLiveQuery;
     user.restoreAs = FormplayerFrontend.request('restoreAsUser', user.domain, user.username);
+
+    FormplayerFrontend.Apps.API.primeApps(user.restoreAs, options.apps);
 
     savedDisplayOptions = _.pick(
         Util.getSavedDisplayOptions(),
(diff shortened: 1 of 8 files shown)


wpride (approved):

corehq/apps/cloudcare/views.py
```diff
@@ -98,11 +101,52 @@ def get(self, request, domain):
...
+
+        cookie_name = urllib.quote(
+            'restoreAs:{}:{}'.format(domain, request.couch_user.username))
+        username = request.COOKIES.get(cookie_name)
+        if username:
+            user = CouchUser.get_by_username(format_username(username, domain))
+            if user:
+                return user, set_cookie
+            else:
+                def set_cookie(response):  # overwrite the default noop set_cookie
```
benrudolph: this kind of confused me when i first read it. would it make more sense to just return a boolean that says whether or not to delete the cookie?

benrudolph:
lookin 🔥

corehq/apps/cloudcare/views.py
dannyroberts: you could do that, but then it's the caller's responsibility to know what the cookie key is, etc. I think while it's a little weird, this encapsulates it better. Would you prefer it if I defined

```
def do_delete_cookie(response):
    ...

def do_not_delete_cookie(response):
    return response
```

at the top and then had

```
set_cookie = do_not_delete_cookie
...
    set_cookie = do_delete_cookie
```

or is that even worse?

corehq/apps/cloudcare/static/cloudcare/js/formplayer/app.js
```diff
@@ -510,9 +508,9 @@ FormplayerFrontend.on('navigateHome', function() {
     FormplayerFrontend.regions.breadcrumb.empty();
     if (currentUser.displayOptions.singleAppMode) {
         appId = FormplayerFrontend.request('getCurrentAppId');
-        FormplayerFrontend.navigate("/single_app/" + appId, { trigger: true });
+        FormplayerFrontend.trigger("app:singleApp", appId);
```
dannyroberts: Offlined with ben. I changed this because if you navigate directly,  then the logic to load the new list of apps (or the new app) in this case isn't called, and without that this feature change doesn't work. The handler for `"app:singleApp"` immediately calls this same navigate line above, so there's no problem with the URL; Ben and I also confirmed this by testing locally.

benrudolph (approved):
offlined, g2g from me

corehq/apps/cloudcare/views.py
dannyroberts: Offlined with ben, decided to keep as is.

PR #17223, opened 2017-07-26
Mk nikshay integration updates

mkangia:
Code changes achieve:
1. Notify only for confirmed_tb patients by checking on episode_type
2. differentiate between 1.1 and 2b app by checking on person case_version and set/get fields from different places on basis of [this doc](https://docs.google.com/spreadsheets/d/1YjG9skHVNw055cF6OA4VLhqmKvoOkH5kQ3x6b_APJOo/edit#gid=0)
3. Check for case property `dataset` for 2b app cases instead of relying on locations for checking for test submissions

cc: buddy @calellowitz @sheelio 

Still need to add tests and test it once to ensure i got the properties fetching correctly but otherwise this should be all that needs to be changed for integrations as of now for 2b release


custom/enikshay/const.py
```diff
@@ -169,3 +169,7 @@
 AGENCY_LOCATION_FIELDS = [
     ('private_sector_org_id', "Private Sector Org ID", []),
 ]
+
+DSTB_EPISODE_TYPE = "confirmed_tb"
+NEW_2B_APP_PERSON_CASE_VERSION = 20
```
proteusvacuum: At some point this won't be `NEW` anymore. Might be easier to just call this `V20` or something.

custom/enikshay/integrations/nikshay/repeater_generator.py
```diff
@@ -97,7 +105,10 @@ def get_payload(self, repeat_record, episode_case):
         person_case = get_person_case_from_episode(episode_case.domain, episode_case.get_id)
         episode_case_properties = episode_case.dynamic_case_properties()
         person_case_properties = person_case.dynamic_case_properties()
-
+        occurence_case = None
+        use_new_2b_app_structure = self.use_new_2b_app_structure(person_case, episode_case)
```
proteusvacuum: same comment as for the const, e.g. `use_v20_app_structure`

custom/enikshay/integrations/nikshay/repeater_generator.py
```diff
@@ -243,9 +255,12 @@ def get_payload(self, repeat_record, test_case):
 
         test_case_properties = test_case.dynamic_case_properties()
         episode_case_properties = episode_case.dynamic_case_properties()
+        use_new_2b_app_structure = self.use_new_2b_app_structure(person_case, episode_case)
 
         interval_id, lab_serial_number, result_grade, dmc_code = self._get_mandatory_fields(
-            test_case, test_case_properties)
+            test_case, test_case_properties, occurence_case,
+            use_new_2b_app_structure
```
proteusvacuum: since you made this an instance method, you could avoid passing this variable around and just call `self.use_new_2b_app_structure`

custom/enikshay/integrations/nikshay/repeater_generator.py
```diff
@@ -86,6 +88,12 @@ def _base_properties(self, repeat_record):
             "IP_FROM": server_ip,
         }
 
+    def use_new_2b_app_structure(self, person_case, episode_case):
```
proteusvacuum: do you think it would be useful to just call this `case_version` and do something like:
```py
if (
            episode_case.dynamic_case_properties().get('episode_type') == DSTB_EPISODE_TYPE and
            person_case.dynamic_case_properties().get('case_version') == NEW_2B_APP_PERSON_CASE_VERSION
        ):
    return V2
else:
    return V1
```
It might be a bit more future proof. Willing to be convinced that this is "YAGNI" though.

mkangia: sorry i don't completely understand this. What is V2, V1?

also sheel said we would eventually remove the logic for 1.1 and just keep for 2.0 after a month since everyone would move to that so okay to keep things simple.

custom/enikshay/integrations/nikshay/repeater_generator.py
mkangia: passed this around to avoid calling and calculating things again and again. Can memoize it as well.

sheelio:
Looks ok to me.

custom/enikshay/integrations/nikshay/repeater_generator.py
proteusvacuum: 👍 sounds right

mkangia:
@proteusvacuum added tests. Better to review commit by commit done after your last review. There is one long commit  for better indentation "just some clean up for update_case syntax" you can skip.

proteusvacuum:
Looks good. Will the `is_valid_person_submission` method still work properly? @sheelio told me there is a new value they've added there: `_invalid_` (or something). I think that method might need to be updated.

proteusvacuum (approved):
Looks good pending my question

mkangia:
> Looks good. Will the is_valid_person_submission method still work properly? @sheelio told me there is a new value they've added there: _invalid_ (or something). I think that method might need to be updated.

@proteusvacuum REAL_DATASET_PROPERTY_VALUE is the one change that i made here. Is there anything else as well @sheelio?

proteusvacuum:
@mkangia see this [ticket](https://manage.dimagi.com/default.asp?258498) for more info. If you think the other changes you made satisfy this criteria, that is fine.

mkangia:
thank you @proteusvacuum. From @sheelio 's comments, looks like after these changes that scenario will not happen when the owner will be '_invalid_'. hence no change should be needed here. @sheelio can you please confirm and then i can merge these changes if there is no other modification needed

PR #17224, opened 2017-07-26
Add index schema to products fixture

esoergel:
Add an index schema to the products fixture upon request from TDH Mali.  This should be a substantial speedup for any Supply projects with a large products list.

Here's a snapshot from a restore on my test domain with this change:

```xml
  <schema id="commtrack:products">
    <indices>
      <index>@id</index>
      <index>category</index>
      <index>code</index>
      <index>program_id</index>
    </indices>
  </schema>
  <fixture id="commtrack:products" indexed="true" user_id="858addfda49a77992d96ad64175733b8">
    <products>
      <product id="858addfda49a77992d96ad641775d2df">
        <name>milk</name>
        <unit/>
        <code>milk</code>
        <description/>
        <category/>
        <program_id>f92145f6b9a372f89599cb82a1bce97f</program_id>
        <cost/>
      </product>
    </products>
  </fixture>
```

@kaapstorm
@ctsims FYI

### DIFF
--- a/corehq/apps/ota/views.py
+++ b/corehq/apps/ota/views.py
@@ -17,7 +17,7 @@ from casexml.apps.case.fixtures import CaseDBFixture
 from casexml.apps.case.models import CommCareCase
 from casexml.apps.case.xml import V2
 from corehq import toggles, privileges
-from corehq.const import OPENROSA_VERSION_MAP, OPENROSA_DEFAULT_VERSION
+from corehq.const import OPENROSA_VERSION_MAP
 from corehq.middleware import OPENROSA_VERSION_HEADER
 from corehq.apps.app_manager.dbaccessors import get_app
 from corehq.apps.case_search.models import QueryMergeException
@@ -199,7 +199,7 @@ def get_restore_params(request):
         openrosa_headers = getattr(request, 'openrosa_headers', {})
         openrosa_version = openrosa_headers[OPENROSA_VERSION_HEADER]
     except KeyError:
-        openrosa_version = request.GET.get('openrosa_version', OPENROSA_DEFAULT_VERSION)
+        openrosa_version = request.GET.get('openrosa_version', None)
 
     return {
         'since': request.GET.get('since'),
@@ -220,7 +220,7 @@ def get_restore_response(domain, couch_user, app_id=None, since=None, version='1
                          cache_timeout=None, overwrite_cache=False,
                          force_restore_mode=None,
                          as_user=None, device_id=None, user_id=None,
-                         openrosa_version=OPENROSA_DEFAULT_VERSION,
+                         openrosa_version=None,
                          case_sync=None):
 
     if user_id and user_id != couch_user.user_id:
@@ -260,8 +260,9 @@ def get_restore_response(domain, couch_user, app_id=None, since=None, version='1
     project = Domain.get_by_name(domain)
     app = get_app(domain, app_id) if app_id else None
     async_restore_enabled = (
-        toggles.ASYNC_RESTORE.enabled(domain) and
-        LooseVersion(openrosa_version) >= LooseVersion(OPENROSA_VERSION_MAP['ASYNC_RESTORE'])
+        toggles.ASYNC_RESTORE.enabled(domain)
+        and openrosa_version
+        and LooseVersion(openrosa_version) >= LooseVersion(OPENROSA_VERSION_MAP['ASYNC_RESTORE'])
     )
     restore_config = RestoreConfig(
         project=project,
@@ -273,6 +274,7 @@ def get_restore_response(domain, couch_user, app_id=None, since=None, version='1
             include_item_count=items,
             app=app,
             device_id=device_id,
+            openrosa_version=openrosa_version,
         ),
         cache_settings=RestoreCacheSettings(
             force_cache=force_cache or async_restore_enabled,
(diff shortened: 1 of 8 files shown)


ctsims:
@esoergel Is it set up in code in any way that this can't be enabled as is? @id will overlap with organizations and crash any phone in the field (since you need 2.38 for that to be fixed).

Even setting the "CommCare App Verison" to 2.38 won't fix that because these come down on restore, so unless HQ is setting a field that will make it to the phone to request this format we need a way to guarantee this doesn't go on unless it's explicitly turned on or we can use some sort of version # to prevent it from going to the wrong devices.

esoergel:
@ctsims right, I'd forgotten about that - will set something up.

kaapstorm:
Could we use something like
```xml
<schema for="some_fixture_id"></schema>
<fixture id="some_fixture_id"></fixture>
``` 
instead of using duplicate ID attributes? Even when they "work", duplicate IDs have caused us perennial problems in the past, e.g. submodules.

kaapstorm:
LGTM.

There is a very similar chunk of code for [indexing lookup tables][1]. At some point we should probably think about DRYing these.

  [1]: https://github.com/dimagi/commcare-hq/blob/51a5ea99de073deb085943db88e1969c8f6aa8b2/corehq/apps/fixtures/fixturegenerators.py#L98-L98

corehq/apps/ota/views.py
```diff
@@ -199,7 +199,7 @@ def get_restore_params(request):
         openrosa_headers = getattr(request, 'openrosa_headers', {})
         openrosa_version = openrosa_headers[OPENROSA_VERSION_HEADER]
     except KeyError:
-        openrosa_version = request.GET.get('openrosa_version', OPENROSA_DEFAULT_VERSION)
+        openrosa_version = request.GET.get('openrosa_version', None)
```
esoergel: @proteusvacuum you'd probably be interested in this commit.  I needed openrosa verson for something else (see the rest of the PR if interested), so I saved it as-is to the `RestoreParams` and dropped the business of defaulting to `1.0`.  Now that fallback behavior is defined in the `async_restore_enabled` calculation.  The feature I'm doing wants the opposite approach, so saving it as `None` when not specified will allow (force?) each use case to decide how to handle unspecified versions.

esoergel:
@kaapstorm good call, I added a commit to DRY that up.  Regarding the duplicate indices thing, the issue is that we can't have multiple schemas telling the phone to index fields with the same name prior to 2.38.  That is, the products fixture wants `@id` indexed, as does the locations fixture, but that's not namespaced properly or something, so that errors on older versions.

esoergel:
Just tested this out on staging with both 2.37 and Clayton's 2.38 apk, and I was able to access the products fixture from within a form.  Webapps on staging could load the form and fill it out, but I got errors on submit.  I'm betting that's unrelated though, 'cause it's after this would've been triggered, and staging webapps has problems.

corehq/apps/products/fixtures.py
```diff
@@ -52,8 +53,19 @@ def get_products():
...
 
-        return simple_fixture_generator(
+        fixture_nodes = simple_fixture_generator(
             restore_user, self.id, "product", PRODUCT_FIELDS, get_products, restore_state.last_sync_log
         )
+        if not fixture_nodes:
+            return []
+
+        if restore_state.params.openrosa_version and restore_state.params.openrosa_version < '2.1':
+            # Don't include index schema when openrosa version is specified and below 2.1
```
proteusvacuum: I might be missing something very obvious from your code, but is there a reason you didn't do the inverse:
```py
if LooseVersion(openrosa_version) >= LooseVersion(OPENROSA_VERSION_MAP['INDEXED_PRODUCTS']:
    schema_node = get_index_schema_node(self.id, ['@id', 'code', 'program_id', 'category'])
    fixture_nodes[0].attrib['indexed'] = 'true'
    return [schema_node] + fixture_nodes
else: 
    return fixture_nodes
```
That way you would have maintained the convention, and we will retain a map of "features to openrosa versions" in the constant. 

Also, then you can keep passing the default version around when this isn't specified (although I don't really care about that as much).

esoergel: I initially wrote this as:
```python
if not restore_state or openrosa_version > '2.1':
    # schema
else:
    # no schema
```
(happy to switch back if that's clearer)
The difference between this use-case and async_restore is that here we use the newer version also when openrosa_version isn't provided (Clayton was particular on that point).

I do like the idea of sticking to the map of features to openrosa versions, though - cleaner than just sticking it in app logic.  Think that still make sense if it's a discontinuous version range?

proteusvacuum: I see. I think it still makes sense to do 
```py
if not restore_state or openrosa_version >= OPENROSA_VERSION_MAP['INDEXED_PRODUCTS']:
...
```
You can also write a comment next to where you define that constant saying that this will also be applied if there is _no_ version, that way we keep everything in one place.

I'm curious when an openrosa version isn't sent down, and why you'd add the index in those instances?

esoergel: Just added it to `OPENROSA_VERSION_MAP`.

I think the thought is that the "correct" way to do this is to always include the schema, and older versions will just ignore it.  The first couple versions using the schema have a bug which means we can't provide this optimization, but those are problem versions, the versions older and new versions are fine.  With that framing, I think this logic makes sense, and for that reason I also kept the "bad" `if` branch first, as a special case that shortcuts out early.

proteusvacuum: 👍

PR #17226, opened 2017-07-27
Use base template in eNikshay repeater report

proteusvacuum:
I'm still getting the offline generation error, even after https://github.com/dimagi/commcare-hq/pull/17203
I don't care about all the javascript or anything, and the report looks exactly the same with this template, and I suspect the other template is what is causing the error.

@benrudolph

### DIFF
--- a/custom/enikshay/reports/repeaters.py
+++ b/custom/enikshay/reports/repeaters.py
@@ -32,6 +32,7 @@ class ENikshayRepeaterFilter(RepeaterFilter):
 class ENikshayForwarderReport(DomainForwardingRepeatRecords):
     name = 'eNikshay Forwarder Report'
     base_template = 'reports/base_template.html'
+    asynchronous = True
     section_name = 'Custom Reports'
     slug = 'enikshay_repeater_report'
     dispatcher = CustomProjectReportDispatcher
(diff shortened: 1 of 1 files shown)


custom/enikshay/reports/repeaters.py
```diff
@@ -32,6 +32,7 @@ def _get_repeaters(self):
 class ENikshayForwarderReport(DomainForwardingRepeatRecords):
     name = 'eNikshay Forwarder Report'
     base_template = 'reports/base_template.html'
+    asynchronous = True
```
proteusvacuum: I think this is the golden ticket... maybe this is obvious, but do `emailable` reports also have to be marked `asynchronous`?

benrudolph: interesting... i don't know why that would be the case

benrudolph (approved):
curious how you're testing this

proteusvacuum:
setting up a saved report locally, then seeing what gets generated when I send it manually (now with all these changes there is no javascript included in the report). Obviously I can't get the `OfflineGenerationError`, but am making an educated guess as to what is causing it.

benrudolph:
👍 there are some instructions for settings up django to use compression in the readme, but what you're doing sounds good and most time effective

PR #17227, opened 2017-07-27
Fix unexpected 412 after livequery sync

millerdev:
http://manage.dimagi.com/default.asp?258491

A lot of the new code is testing infrastructure. I hope to retrofit a lot of existing tests with `MockDevice` since that manages devices state more cleanly instead of having it sprinkled all over the test class.

The first commit is just cleanup.

@snopoke @ctsims @proteusvacuum cc @gcapalbo

### DIFF
--- a/corehq/ex-submodules/casexml/apps/case/mock/case_block.py
+++ b/corehq/ex-submodules/casexml/apps/case/mock/case_block.py
@@ -96,16 +96,16 @@ class CaseBlock(object):
         if self.index:
             result['index'] = {}
             for name in self.index.keys():
-                self.case_type = self.index[name].case_type
-                self.case_id = self.index[name].case_id
+                case_type = self.index[name].case_type
+                case_id = self.index[name].case_id
 
                 relationship = self.index[name].relationship
-                _attrib = {'case_type': self.case_type}
+                _attrib = {'case_type': case_type}
                 if relationship != 'child':
                     _attrib['relationship'] = relationship
                 result['index'][name] = {
                     '_attrib': _attrib,
-                    '_text': self.case_id
+                    '_text': case_id
                 }
         return result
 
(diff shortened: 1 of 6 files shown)


corehq/ex-submodules/casexml/apps/case/mock/case_block.py
```diff
@@ -96,16 +96,16 @@ def _to_json(self):
         if self.index:
             result['index'] = {}
             for name in self.index.keys():
-                self.case_type = self.index[name].case_type
-                self.case_id = self.index[name].case_id
+                case_type = self.index[name].case_type
+                case_id = self.index[name].case_id
```
millerdev: Pretty sure this was a bug. It was likely never noticed before since `CaseBlock`s are rarely referenced after being posted.

corehq/ex-submodules/casexml/apps/phone/data_providers/case/livequery.py
```diff
@@ -232,7 +232,7 @@ def update_open_and_deleted_ids(related):
                     live_ids, restore_state, accessor)
         else:
             sync_ids = live_ids
-        restore_state.current_sync_log.case_ids_on_phone = live_ids | sync_ids
+        restore_state.current_sync_log.case_ids_on_phone = live_ids
```
millerdev: This is the bugfix.

corehq/ex-submodules/casexml/apps/phone/tests/utils.py
```diff
@@ -151,3 +153,69 @@ def call_fixture_generator(gen, restore_user, project=None, last_sync=None, app=
         params.sync_log_id = last_sync._id
         restore_state._last_sync_log = last_sync
     return gen(restore_state)
+
+
+class MockDevice(object):
```
proteusvacuum: this is very nice! 👍

corehq/ex-submodules/casexml/apps/phone/data_providers/case/livequery.py
proteusvacuum: 👍

proteusvacuum (approved):
Looks good to me. I really like the test updates you've made. Will make the rest of those tests much easier to read. Thanks!

PR #17228, opened 2017-07-27
Export only visible columns

kkrampa:
@NoahCarnahan @proteusvacuum


NoahCarnahan (approved):
Looks good thanks. I'll merge when the build finishes.

PR #17232, opened 2017-07-27
Update column heading in Abt static UCR

NoahCarnahan:
cc: buddy @biyeun


NoahCarnahan:
Self-merging to be sure this is in before today's deploy

PR #17235, opened 2017-07-27
Delete more utils

dannyroberts:


emord (approved):

sravfeyn:
I have a question, what if someone in 6 months needs some part of these utils again, is there a concept of 'archiving' utils? This is not specific to this PR, but about deleting code in general

dannyroberts:
Well, in a sense, any time you delete code you're really just archiving it. Git and github store our entire repository history including these utils, so if someone wants them and knows they used to exist, they can pull them from the history. That said, in my experience, this very nearly never happens. By the time a new use case comes around, it's often easier to write a 40-line function serving your use-case than it is to repurpose old code for a new use-case.

PR #17239, opened 2017-07-27
2b migration followup

esoergel:
@sheelio @kaapstorm

### DIFF
--- a/custom/enikshay/two_b_release_1/management/commands/enikshay_2b_case_properties.py
+++ b/custom/enikshay/two_b_release_1/management/commands/enikshay_2b_case_properties.py
@@ -330,8 +330,9 @@ class ENikshay2BMigrator(object):
         occurrence_ids = [index.referenced_id for index in indices
                           if index.referenced_type == CASE_TYPE_OCCURRENCE]
         occurrence_id = occurrence_ids[0] if occurrence_ids else None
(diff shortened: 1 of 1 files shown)


custom/enikshay/two_b_release_1/management/commands/enikshay_2b_case_properties.py
```diff
@@ -330,8 +330,9 @@ def is_episode_of_occurrence(episode, occurrence_id):
         occurrence_ids = [index.referenced_id for index in indices
                           if index.referenced_type == CASE_TYPE_OCCURRENCE]
         occurrence_id = occurrence_ids[0] if occurrence_ids else None
-        return max((case.opened_on, case.case_id) for case in episodes
-                   if is_episode_of_occurrence(case, occurrence_id))[1]
+        episodes = ((case.opened_on, case.case_id) for case in episodes
+                    if is_episode_of_occurrence(case, occurrence_id))
+        return max(episodes)[1] if episodes else None
```
NoahCarnahan: I think this might not work as expected. I believe that `episodes` will be a generator object, which will always be truthy.

esoergel: hmm, that's true - will update to a list

NoahCarnahan (approved):

PR #17240, opened 2017-07-27
Fixed error in exports on deleted groups.

emord:
https://manage.dimagi.com/default.asp?258950

https://sentry.io/dimagi/commcarehq/issues/317554853/

@NoahCarnahan @benrudolph 

buddy @nickpell

### DIFF
--- a/corehq/apps/users/cases.py
+++ b/corehq/apps/users/cases.py
@@ -13,7 +13,7 @@ def get_owner_id(case):
     return case.owner_id or case.modified_by
 
 
-def get_wrapped_owner(owner_id):
+def get_wrapped_owner(owner_id, support_deleted=False):
     """
     Returns the wrapped user or group object for a given ID, or None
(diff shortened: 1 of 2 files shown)


corehq/apps/users/cases.py
```diff
@@ -28,6 +28,13 @@ def _get_class(doc_type):
             'Group': Group,
         }.get(doc_type)
 
+    def _get_deleted_class(doc_type):
+        return {
+            'CommCareUser-Deleted': CommCareUser,
+            'WebUser-Deleted': WebUser,
```
benrudolph: I think this is actually `CouchUser-Deleted`: https://github.com/dimagi/commcare-hq/blob/master/corehq/apps/users/models.py#L1675

benrudolph (commented):
one thing to check, otherwise looks good

benrudolph (approved):

PR #17246, opened 2017-07-28
Revert "Add missing lib to bower"

gcapalbo:
Reverts dimagi/commcare-hq#17245


gcapalbo:
reverting this as it breaks deploy

PR #17248, opened 2017-07-28
Ce/async user sync

calellowitz:
this makes the user filter for the sync history report async so it works on domains with many users. i also fixed up the profiling middleware while i was looking around. 
http://manage.dimagi.com/default.asp?258745#1368987
@mkangia @benrudolph


corehq/apps/reports/filters/api.py
```diff
@@ -195,6 +195,43 @@ class MobileWorkersOptionsView(EmwfOptionsView):
...
+    # This endpoint is used by select2 single option filters
+    def post(self, request, domain):
+        self.domain = domain
+        self.q = self.request.POST.get('q', None)
+        try:
+            count, options = self.get_post_options()
+            return self.render_json_response({
+                'items': options,
+                'total': count,
+                'limit': request.POST.get('page_limit', 10),
```
mkangia: Its good to avoid having numbers kept around and rather move them as a meaningful constant like for instance PAGE_LIMIT here. It makes it easier to change it when needed and much easier to figure out related places.

corehq/apps/reports/filters/api.py
```diff
@@ -195,6 +195,43 @@ class MobileWorkersOptionsView(EmwfOptionsView):
...
+                notify_exception(request, e)
+        return self.render_json_response({
+            'results': [],
+            'total': 0,
+        })
+
+    def get_post_options(self):
+        page = int(self.request.POST.get('page', 1))
+        size = int(self.request.POST.get('page_limit', 10))
+        start = size * (page - 1)
```
mkangia: :nit I think you can just take page default as 0.

corehq/apps/reports/filters/api.py
```diff
@@ -195,6 +195,43 @@ class MobileWorkersOptionsView(EmwfOptionsView):
...
+                # contains the user's input.
+                logger.info('ElasticSearch error caused by query "%s": %s',
+                            self.q, e)
+            else:
+                # The error was our fault
+                notify_exception(request, e)
+        return self.render_json_response({
+            'results': [],
+            'total': 0,
+        })
```
mkangia: This happens when there is an error? The user won't get notified with "There was an error in the query .." message?

mkangia (changes requested):

corehq/apps/reports/filters/api.py
calellowitz: i think this has to do with the defaults of our js library that makes the actual ajax call

corehq/apps/reports/filters/api.py
calellowitz: This is the same as the expanded mobile worker filter. I think it is this way because here the call is being used to return possible results to a  select2, so i think a blank result is just as meaningful. it indicates to the user to change what they have typed in the filter.

calellowitz:
@mkangia i think i got to all of your feedback. let me know what you think

mkangia (approved):
Looks good.

PR #17250, opened 2017-07-28
Fix conflict on save form with shadow forms

millerdev:
https://manage.dimagi.com/default.asp?256317

@NoahCarnahan it looks like you [implemented "shadow forms"](https://github.com/dimagi/commcare-hq/pull/16262) earlier this year. You may want to review this code and also my [comments on the fb ticket](https://manage.dimagi.com/default.asp?256317#BugEvent.1371286). Also, @dannyroberts may be interested since he implemented the [XMLNS changing code](https://github.com/dimagi/commcare-hq/commit/64cd6c91eef957165459041feb3b38488931a59d) way back in the day.

cc @gcapalbo

### DIFF
--- a/corehq/apps/app_manager/util.py
+++ b/corehq/apps/app_manager/util.py
@@ -120,6 +120,7 @@ def split_path(path):
 
 def save_xform(app, form, xml):
     from corehq.apps.app_manager.models import ShadowForm
+
     def change_xmlns(xform, old_xmlns, new_xmlns):
         data = xform.data_node.render()
         data = data.replace(old_xmlns, new_xmlns, 1)
@@ -136,7 +137,7 @@ def save_xform(app, form, xml):
             GENERIC_XMLNS = "http://www.w3.org/2002/xforms"
             xmlns = "http://openrosa.org/formdesigner/%s" % form.get_unique_id()
             tag_xmlns = xform.data_node.tag_xmlns
-            if not tag_xmlns or tag_xmlns == GENERIC_XMLNS:  #no xmlns
+            if not tag_xmlns or tag_xmlns == GENERIC_XMLNS:  # no xmlns
                 xml = change_xmlns(xform, GENERIC_XMLNS, xmlns)
             elif tag_xmlns != xmlns:
                 xml = change_xmlns(xform, tag_xmlns, xmlns)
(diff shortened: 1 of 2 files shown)


corehq/apps/app_manager/util.py
```diff
@@ -119,37 +119,33 @@ def split_path(path):
...
     except XFormException:
         pass
     else:
-        duplicates = app.get_xmlns_map()[xform.data_node.tag_xmlns]
-        for duplicate in duplicates:
-            if form == duplicate:
-                continue
-            else:
-                xform, xml = change_xmlns(xform, xform.data_node.tag_xmlns)
-                break
```
millerdev: I'm not even sure how this was supposed to work. It appears to change the XMLNS of the first form it finds that is not the current form, and then puts the XML from that form into the current form... that seems broken to me, but I may be missing something.

corehq/apps/app_manager/util.py
```diff
@@ -119,37 +119,34 @@ def split_path(path):
 
 
 def save_xform(app, form, xml):
```
NoahCarnahan: I think this function should never be called on `ShadowForm`s. For instance, the line
`form.source = xml`
will error because in shadow forms `form.source` is an object `@property`, not a `StringProperty`.

corehq/apps/app_manager/util.py
NoahCarnahan: I agree that seems weird that is changing a _different_ form.

It looks like your version changes the xmlns of the form if it does not match `form.get_unique_id()`, even if there aren't any conflicts with other forms in the app, which looks a little different then what existed before. Not sure if that's a problem or not, but just wanted to point it out.

corehq/apps/app_manager/util.py
NoahCarnahan: Oh you know what, I wonder if this wasn't being called on a `ShadowForm`, but that `change_xmlns()` below _was_ being called on a shadow form, because as you pointed out it seems to try to modify other forms in the app.

corehq/apps/app_manager/util.py
millerdev: Good point. I just removed the thing that changes the XMLNS if it does not match. I'm not sure if that's ever necessary anyway.

corehq/apps/app_manager/util.py
millerdev: Yep, that's what was happening. I removed the check for `ShadowForm` since it sounds like that's not important here.

NoahCarnahan (approved):
Looks good to me

PR #17254, opened 2017-07-28
Fix maps_prefer_canvas

orangejenny:
@snopoke / @biyeun


snopoke (approved):
Tried to merge but got denied

PR #17259, opened 2017-07-28
Migrate export docs

benrudolph:
@esoergel @NoahCarnahan some docs on migrating exports (noah let me know if you think i missed something)

cc: @proteusvacuum

### DIFF
--- a/corehq/apps/export/README.md
+++ b/corehq/apps/export/README.md
@@ -103,6 +103,39 @@ System properties are properties that CommCareHQ adds to the form or case that a
 
 Often times iterating through each build in an application can be time consuming. To reduce the time to process an application, the code only processes applications that have been marked as having submissions.
 
+### Migrating Exports
+
+At the time of writing this HQ is still in the process of migrating exports over to the new export
+infrastructure. In order to migrate an export there are two commands you can use:
+
+```
+./manage.py migrate_exports
+./manage.py migrate_exports_for_domain <domain>
+```
+
+Check out the command for more documentation on the options.
+
+#### Migration strategy
+
+The general approach for migrating exports is:
+
+1. Load the `SavedExportSchema` (old export model)
+2. Generate a new schema of the export using the applications in the project
+3. Generate an `ExportInstance` based off the new schema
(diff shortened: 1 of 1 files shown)


corehq/apps/export/README.md
```diff
@@ -103,6 +103,39 @@ System properties are properties that CommCareHQ adds to the form or case that a
...
+
+There is one further difficulty that has made the last 20 or so domains difficult to migrate. When migrating some of the leftover domains, you may see the error:
+
+```
+Failed parsing 83b6fbfadc1fe0a43bc084fadb456a3f: Form exports require an app id
+```
+
+This is because when parsing the export 83b6fbfadc1fe0a43bc084fadb456a3f, it could not find a match application for the XMLNS found in the export. This usually happens because of `RemoteApplication`s. The new `FormExportInstance` requires that all exports contain an `app_id` in order to exist. In the future, the correct way to resolve this is to drop that requirement.
```
NoahCarnahan: So for `RemoteApplication`s we should drop the `app_id` requirement and just create an `InferredSchema`? Might be worth adding one more sentence to make it clear what happens after the `app_id` requirement is dropped.

NoahCarnahan (approved):

corehq/apps/export/README.md
proteusvacuum: and if there is an easy command to do this for remote apps?
I'm a bit confused by this, is there someone tasked with migrating the last domains? Or is this something that just happens ad-hoc?

I'm also wondering if this is better off as a confluence doc, since this has more to do with an internal process than anything anyone running the code should do.

benrudolph: there is unfortunately no easy command for remote apps. hasn't been a priority to implement so i haven't yet :/ there is no one tasked with migrating the last of the domains, but i thought it'd be nice to leave a note on the current state in case we ever want to finish.

it could be a confluence doc, but right now all the docs are here and i hate confluence's discoverability. there is some precedent for documenting migrations in the [code base](https://github.com/dimagi/commcare-hq/blob/br/migrate-export-docs/corehq/doctypemigrations/README.md).

PR #17261, opened 2017-07-28
Allow colon in form attachment url

emord:
Allowing users to specify their own form ids was probably not our greatest idea.

https://sentry.io/dimagi/commcarehq/issues/282320610/

http://manage.dimagi.com/default.asp?258693#1371489

@nickpell


benrudolph (approved):
👍 until the next uuid with `%` comes in :trollface:

PR #17265, opened 2017-07-28
Externalize JavaScript: style/base.html

orangejenny:
Takes care of the low-hanging fruit in `style/base.html`

@snopoke / @millerdev


millerdev (approved):
Thanks for keeping on with all this cleanup!

PR #17272, opened 2017-07-28
Coverity changes

emord:
I ran part of HQ through coverity scan and it found these issues/cleanup. and some more, but I don't really feel like digging through some of the old javascript. Was going to try to include a link to each issue, but couldn't see a way to make a permalink.

https://scan.coverity.com/projects/dimagi-commcare-hq

@nickpell


nickpell (approved):
Nice!

emord:
I think I let this one sit for too long. Will have to take a look at a later time

PR #17274, opened 2017-07-28
Remove laboratory from ICDS reports.

emord:
@dimagi/scale-team This would remove the laboratory backend from the ICDS reports. I think we might not be close to having a stable ES cluster, so it might be worth taking this off. Thoughts?

### DIFF
--- a/custom/icds_reports/ucr/data_sources/ccs_record_cases.json
+++ b/custom/icds_reports/ucr/data_sources/ccs_record_cases.json
@@ -1750,10 +1750,6 @@
       }
     },
     "engine_id": "icds-ucr",
-    "backend_id": "LABORATORY",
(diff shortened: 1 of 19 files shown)


snopoke:
If we wanted to put it back on it would require rebuilding right? What's the harm in keeping it? Is it just the added overhead?

emord:
I think we could dump the rows from postgres and iterate over them to post to ES. Would likely be a quick script to write.

I'm also wondering if we should even use the ES backend. These reports seem to work well using SQL and now that we're using architect there's likely a way we could partition these tables by state or block and have reports pass those in. For form based UCRs they could be partitioned based on date or something.

snopoke:
:+1:

PR #17275, opened 2017-07-29
ICDS: Fix AWC Supervision Table

sheelio:
@emord

This change fixes some bugs with this table and adds a new column (so will need to be rebuilt).  Its a form based data source with very few forms - is possible to just run this for those forms?


emord:
@sheelio the code to do this well hasn't been written yet. How urgent is this to get in?

emord:
dependent on https://github.com/dimagi/commcare-hq/pull/17309/ so that it doesn't rebuild

sheelio:
I think 17309 is merged, so this should be good to merge?

emord:
@sheelio noah may have found that my earlier PR didn't work as we thought it would. I'm going to do some more testing on it and then we'll see. in the meantime could you resolve the merge conflict?

sheelio:
@emord 

This should do it.

emord (approved):
@sheelio I'll deploy this tomorrow before the aggregation script runs. I'm holding off right now because this changes the table definition needed by the aggregation script so I believe that it would cause a lot of locks on the database

emord:
Deployed htis, recaluclated the tables and restarted the aggregation. Should see this be fixed  tomorrow

PR #17276, opened 2017-07-31
Option to force mobile users to update app

sravfeyn:
[Spec ](https://docs.google.com/document/d/1RhdJPD6ZvRGcUsSgULMt05yXUQG6j_FLCAi1fMMEtyQ/edit#)

@dannyroberts 
@snopoke FYI

There is one UI spec comment that's not resolved, so hold-off on merge.


corehq/apps/ota/views.py
```diff
@@ -377,11 +375,25 @@ def form_valid(self):
...
-        "latest_apk_version": {},
-        "latest_ccz_version": {}
-    })
+def heartbeat(request, domain, app_id):
+    """
+    An endpoint for CommCare mobile to get latest CommCare APK and app version
+        info. (Should serve from cache as it's going to be busy view)
+
+    'app_id' (that comes from URL) can be id of any version of the app
+    'app_id' (urlparam) is usually id of an app that is not a copy
```
snopoke: can we rename these as follows:
`hq_app_id` - ID of app in HQ
`app_id` - ID of app on mobile

corehq/apps/ota/views.py
```diff
@@ -377,11 +375,25 @@ def form_valid(self):
...
+        mobile simply wants it to be resent back in the JSON, and doesn't
+        need any validation.
+    """
+    url_param_app_id = request.GET.get('app_id', '')
+    info = {"app_id": url_param_app_id}
+    try:
+        # mobile will send brief_app_id
+        info.update(LatestAppInfo(url_param_app_id, domain).get_info())
```
snopoke: this should be using `app_id` not `url_param_app_id`

snopoke (changes requested):

corehq/apps/ota/views.py
sravfeyn: `url_param_app_id` is supposed to be the brief app id of the installed app. And `app_id` may not be a brief app id, in which case we will have to lookup brief app id, so I tried to use `url_param_app_id`

corehq/apps/ota/views.py
sravfeyn: Sure. FWIW, both of them come from mobile. Mobile gets `hq_app_id` from the resource URL and  urlparam app_id from other resource xml files

snopoke: I assume you're talking about the value of `@uniqueid` from `profile.xml`. If that's the case its not at all clear from the spec but the docstring here should be updated to say exactly where this is coming from.

corehq/apps/ota/views.py
snopoke: By `brief app id` do you mean the ID of the `dev` version of the app i.e. not a build?

corehq/apps/app_manager/util.py
```diff
@@ -595,3 +597,62 @@ def get_and_assert_practice_user_in_domain(practice_user_id, domain):
...
+        """
+        self.app_id = brief_app_id
+        self.domain = domain
+
+    @property
+    @memoized
+    def app(self):
+        app = get_app(self.domain, self.app_id)
+        # quickache based on a copy app_id will have to be updated too fast
+        assert not app.copy_of, "this class doesn't handle copy app ids"
```
sravfeyn: I am passing `app.copy_of` from view. But before that, I pass the id that comes from mobile (which may or may not be brief app id). This allows us to completely avoid fetching app if the correct brief app-id is provided by mobile.

corehq/apps/ota/views.py
sravfeyn: Yeah, it's not clear from spec I am also not sure whether it comes from `@unique_id`, but Aliza mentioned me that app_id urlparam is going to be the fixed app-id that doesn't change between versions, @amstone326 can you confirm where this urlparam app_id is pulled from when mobile makes this request?

corehq/apps/ota/views.py
sravfeyn: Yeah

corehq/apps/app_manager/util.py
sravfeyn: Generally, since this info will rarely change, I designed the views/utils in such a way that the info is cached based on the brief app id (provided by mobile) and pulling the app is avoided as much possible

corehq/apps/ota/views.py
amstone326: @sravfeyn @snopoke Simon is correct, the app_id urlparam comes from the `uniqueid` in the profile

sravfeyn:
@calellowitz @snopoke

corehq/apps/app_manager/views/settings.py
```diff
@@ -66,3 +71,30 @@ def edit_commcare_profile(request, domain, app_id):
...
+class PromptSettingsUpdateView(FormView, ApplicationViewMixin):
+    form_class = PromptUpdateSettingsForm
+    urlname = 'update_prompt_settings'
+
+    @property
+    def success_url(self):
+        return reverse('release_manager', args=[self.domain, self.app_id])
+
+    @method_decorator(no_conflict_require_POST)
+    @method_decorator(require_can_edit_apps)
```
snopoke: if you want to avoid having to override the method you can do this:
```
@method_decorator(require_can_edit_apps, name='dispatch')
class PromptSettingsUpdateView(...):

```

sravfeyn: Nice, thanks

corehq/apps/app_manager/static/app_manager/json/v2/commcare-app-settings.yaml
```diff
@@ -172,4 +172,4 @@
   toggle: MOBILE_UCR
   values: ['none', '0', '1', '2', '4', '6', '8', '12', '24']
   value_names: ['Use project level setting', 'Update on every sync', '1 hour', '2 hours', '4 hours', '6 hours', '8 hours', '12 hours', '24 hours']
-  default: 'none'
+  default: 'none'
```
snopoke: nit

corehq/couchapps/config_by_app_id/views/view/map.js
```diff
@@ -0,0 +1,5 @@
+function(doc) {
+    if (doc.doc_type == "GlobalAppConfig") {
+        emit([doc.app_id, doc.domain], null);
+    }
+}
```
snopoke: The name of this could be better, how about `global_app_config_by_app_id`?

corehq/apps/app_manager/forms.py
```diff
@@ -63,3 +64,76 @@ def clean(self):
         if self.cleaned_data.get('linked'):
             if not LINKED_APPS.enabled(domain):
                 raise forms.ValidationError("The target project space does not have linked apps enabled.")
+
+
+class PromptUpdateSettingsForm(forms.Form):
```
snopoke: Looking at the UI on staging the save button is in a different place from the rest of app manager. It's also a different color and doesn't change state when a data change is made.

corehq/apps/app_manager/forms.py
```diff
@@ -63,3 +64,76 @@ def clean(self):
...
+        label=ugettext_lazy("Prompt Updates to Latest CommCare Version"),
+        choices=(
+            ('off', ugettext_lazy('Off')),
+            ('on', ugettext_lazy('On')),
+            ('forced', ugettext_lazy('Forced')),
+        ),
+        help_text=ugettext_lazy(
+            "If enabled, users will receive in-app prompts to update "
+            "to the latest version of CommCare, if they are not already "
+            "on it. Selecting 'Forced' will make it so that users cannot "
```
amstone326: @sravfeyn Can you place the `"Selecting 'Forced' ... until they upgrade"` portion of this text in parentheses so that it matches the text above.

corehq/apps/app_manager/forms.py
sravfeyn: I will update the colour to be consistent. Regarding making it update the state, this page will be rarely accessed, and there isn't lot of dynamic settings (unlike in form-builder) that would usually make it difficult for the user to keep track of whether they updated something or not. Given this is a much simpler form for now, I think this is okay. This is also under a feature flag

corehq/couchapps/config_by_app_id/views/view/map.js
sravfeyn: Will update

snopoke (approved):

PR #17278, opened 2017-07-31
Log domain and duration of long restores

millerdev:
@proteusvacuum @ctsims cc @gcapalbo

### DIFF
--- a/corehq/apps/ota/views.py
+++ b/corehq/apps/ota/views.py
@@ -103,10 +103,6 @@ def restore(request, domain, app_id=None):
                     tags=tags + ['duration:%s' % bucket],
                 )
         tags.append('duration:%s' % _get_time_bucket(timing_context.duration))
-        if timing_context.duration > 20 or response.status_code == 412:
-            log = logging.getLogger(__name__)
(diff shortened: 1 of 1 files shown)


millerdev:
Just had a thought: wonder if we log response time in nginx access logs? If we do, then all of this info should be fairly easy to get directly from there.

millerdev:
Looks like we don't have timing info in nginx logs. Too bad.

corehq/apps/ota/views.py
```diff
@@ -102,6 +103,10 @@ def restore(request, domain, app_id=None):
                     tags=tags + ['duration:%s' % bucket],
                 )
         tags.append('duration:%s' % _get_time_bucket(timing_context.duration))
+        if timing_context.duration > 20 or response.status_code == 412:
+            log = logging.getLogger(__name__)
+            log.info("restore: domain=%s status=%s duration=%.2f",
+                domain, response.status_code, timing_context.duration)
```
proteusvacuum: would it be useful to have the synclog_id or some other thing that would allow you to see which specific restores are taking a long time?

proteusvacuum (approved):

PR #17280, opened 2017-07-31
improve Data Preview Evaluate XPath formatting

dannyroberts:
- remove always-same XML heading and <result/> wrapping tag in display
- display "Empty String" (not monospaced) for an empty string

Here's how it looks now:

XML node:
screen shot 2017-07-31 at 12 31 27 pm
Single value:
screen shot 2017-07-31 at 12 31 40 pm
Empty string:
screen shot 2017-07-31 at 12 31 50 pm

And in the history:
screen shot 2017-07-31 at 12 31 56 pm

### DIFF
--- a/corehq/apps/cloudcare/static/cloudcare/js/debugger/debugger.js
+++ b/corehq/apps/cloudcare/static/cloudcare/js/debugger/debugger.js
@@ -198,7 +198,15 @@ hqDefine('cloudcare/js/debugger/debugger.js', function () {
         self.$xpath = null;
         self.codeMirrorResult = null;
         self.result = ko.observable('');
-        self.success = ko.observable(true);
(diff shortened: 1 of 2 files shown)


benrudolph (approved):
:clap: very cool, love it

benrudolph (approved):
also just to let you know, if you ever wanted to write a test for the regex, or any other changes you make to the debugger, there's a spec file here: https://github.com/dimagi/commcare-hq/blob/master/corehq/apps/cloudcare/static/cloudcare/js/formplayer/spec/debugger_spec.js

dannyroberts:
Good tip. I'm going to continue this and do a refactor of this area to make it simpler in a separate PR, so maybe I'll test it there.

PR #17281, opened 2017-07-31
Fix header widthHint format

wpride:
fixes https://manage.dimagi.com/default.asp?259093#1372298

Case details are passed styling data in a different format from lists (unfortunately)

### DIFF
--- a/corehq/apps/cloudcare/templates/formplayer/case_list.html
+++ b/corehq/apps/cloudcare/templates/formplayer/case_list.html
@@ -176,7 +176,7 @@
                 <thead>
                 <tr>
                     <% _.each(headers, function(header, index) { %>
-                    <% if (!(widthHints) || !(widthHints[index] === 0)) { %>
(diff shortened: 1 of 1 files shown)


corehq/apps/cloudcare/templates/formplayer/case_list.html
```diff
@@ -176,7 +176,7 @@
                 <thead>
                 <tr>
                     <% _.each(headers, function(header, index) { %>
-                    <% if (!(widthHints) || !(widthHints[index] === 0)) { %>
+                    <% if (!(styles[index].widthHint === 0)) { %>
```
benrudolph: so is the `widthHints` value just not used?

wpride: Yeah, just did a barebones implementation to make search only fields work

benrudolph (approved):

PR #17282, opened 2017-07-31
Add clear breadcrumb middleware

wpride:
Fixes https://manage.dimagi.com/default.asp?259147

We should clear out the breadcrumb every time we navigate (and should be sent down along with each Formplayer response when valid)


benrudolph (approved):

wpride:
We ignore this sticker imperative

PR #17283, opened 2017-07-31
Should only forward patients to Nikshay

proteusvacuum:
@benrudolph @mkangia


benrudolph (approved):
this has become quite the boolean logic

PR #17284, opened 2017-07-31
Check multimedia case properties flag correctly

orangejenny:
https://manage.dimagi.com/default.asp?259101

Introduced in https://github.com/dimagi/commcare-hq/pull/13516

@snopoke / @mkangia


mkangia:
🤕 thank you Jenny.

PR #17288, opened 2017-07-31
Fr/retrigger nikshay

proteusvacuum:
@esoergel @mkangia 

This will take all the current repeat records in the private sector Nikshay repeater and create new records for them in a _new_ repeater of the same type (that will be made before running this command)

The reason for this:
- We triggered a lot of the migrated cases accidentally, so we have like 100,000 records that shouldn't be there
- All of the cases failed their initial sending to Nikshay because we didn't have the TU ids

cc @benrudolph

### DIFF
--- a/custom/enikshay/integrations/nikshay/repeaters.py
+++ b/custom/enikshay/integrations/nikshay/repeaters.py
@@ -216,7 +216,7 @@ class NikshayRegisterPrivatePatientRepeater(SOAPRepeaterMixin, BaseNikshayRepeat
 
         episode_case_properties = episode_case.dynamic_case_properties()
         return (
-            episode_case_properties.get('nikshay_registered', 'false') == 'false' and
+            episode_case_properties.get('private_nikshay_registered', 'false') == 'false' and
             not episode_case_properties.get('nikshay_id') and
             case_properties_changed(episode_case, [PRIVATE_PATIENT_EPISODE_PENDING_REGISTRATION]) and
             episode_case_properties.get(PRIVATE_PATIENT_EPISODE_PENDING_REGISTRATION, 'yes') == 'no' and
--- a/custom/enikshay/management/commands/retrigger_private_nikshay_records.py
+++ b/custom/enikshay/management/commands/retrigger_private_nikshay_records.py
@@ -1,0 +1,46 @@
+from datetime import datetime
+from django.core.management.base import BaseCommand
+from corehq.util.log import with_progress_bar
+from corehq.form_processor.interfaces.dbaccessors import CaseAccessors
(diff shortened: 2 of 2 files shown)


proteusvacuum:
In theory, we can use this to retrigger all cases again if we set `old_repeater_id` and `new_repeater_id` to be the same.

custom/enikshay/integrations/nikshay/repeaters.py
```diff
@@ -216,7 +216,7 @@ def allowed_to_forward(self, episode_case):
 
         episode_case_properties = episode_case.dynamic_case_properties()
         return (
-            episode_case_properties.get('nikshay_registered', 'false') == 'false' and
+            episode_case_properties.get('private_nikshay_registered', 'false') == 'false' and
```
mkangia: that was a bad miss.. :(
thanks for catching.

Did this result in irrelevant triggers?

proteusvacuum: I don't think so, since none of them have succeeded yet. This would basically just make "dublicate entry" in Nikshay :P

custom/enikshay/management/commands/retrigger_private_nikshay_records.py
```diff
@@ -0,0 +1,46 @@
...
+
+        for record in with_progress_bar(records, length=record_count):
+
+            try:
+                episode = accessor.get_case(record.payload_id)
+            except CaseNotFound:
+                continue
+
+            episode_case_properties = episode.dynamic_case_properties()
+            if(episode_case_properties.get('private_nikshay_registered', 'false') == 'false'
```
mkangia: Do we need to change this to "nikshay_registered" considering it was incorrectly set earlier? Or we rely on this getting set correctly during this command execution?

custom/enikshay/integrations/nikshay/repeaters.py
mkangia: :D ... 👍 berfect!

mkangia:
@proteusvacuum 
Also have you considered just re-triggering the repeat records instead of creating a new repeat record? For this particular case at least.. 
agree that it would make sense in case of a different new repeater id.

proteusvacuum:
@mkangia, yes, that is another option, but since there were so many triggered erroneously, I wanted to clean up the data. This way we'll have a "clean slate" of repeat records that we can look through to understand any new failures.

custom/enikshay/management/commands/retrigger_private_nikshay_records.py
proteusvacuum: good point

mkangia (approved):

PR #17292, opened 2017-07-31
Remove app id requirement

benrudolph:
@NoahCarnahan this remove the `app_id` constraint from the form export. Makes me sad, but this is the only way we'll be able to move forward with the migrations. I think once this is in, we'll at last be able to finish the export migration

cc: @proteusvacuum


benrudolph:
tested this out on a domain, i think i this could be it. was able to successfully migrate a remote app and the export looks the same as the old one 😲

NoahCarnahan (approved):
Looks fine to me. This code isn't so fresh in my mind anymore... what is the downside of removing the app id requirement?

benrudolph:
no real downside, only we wouldn't know if someone introduced an error into the code that didn't attach the app id to the schema for some reason (reminiscent of old exports where the models stopped tracking the domain)

PR #17294, opened 2017-07-31
Episode task updates

proteusvacuum:
@benrudolph 
FYI: @sravfeyn 

The task is getting killed every night due to an OOM. https://github.com/dimagi/commcare-hq/commit/73bff5a33c4c7f2a9e2f3c802ec39ac3f6fdb67e might solve that.

We were running the task on cases that aren't even valid, so stop doing that https://github.com/dimagi/commcare-hq/commit/e382721f9e5beb9f81e310e54194722418fbfa56

Also, it is very hard to check the logs for this. I'm sending them by email to everyone now with a CSV of all the errors. If we don't get an email on a particular day, we'll know that the task was killed. We can also deal with each error individually. This is so that there is a bit more visibility over the errors, and can be dealt with by many people, not just when I remember to check the celery logs. 

https://github.com/dimagi/commcare-hq/commit/5ec75088962ae99f91654858855a179ffbb7da37 has been run for long enough now that everyone has updated their app, so removing it from the task. 

🐡 / 🐠

### DIFF
--- a/custom/enikshay/tasks.py
+++ b/custom/enikshay/tasks.py
@@ -96,8 +96,7 @@ class EpisodeUpdater(object):
                 update_json = {}
                 for updater in self.updaters:
                     try:
-                        updater(self.domain, episode)
-                        update_json.update(updater.update_json())
+                        update_json.update(updater(self.domain, episode).update_json())
                     except Exception as e:
                         errors.append(
                             [episode.case_id, updater.__class__, e]
--- a/custom/enikshay/tasks.py
+++ b/custom/enikshay/tasks.py
@@ -76,6 +76,12 @@ class EpisodeUpdater(object):
 
     def __init__(self, domain):
         self.domain = domain
+        self.updaters = [
+            EpisodeAdherenceUpdate,
+            EpisodeVoucherUpdate,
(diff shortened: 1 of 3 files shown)


custom/enikshay/tasks.py
```diff
@@ -127,10 +149,24 @@ def update_single_case(self, episode_case):
...
         case_ids = case_accessor.get_open_case_ids_in_domain_by_type(CASE_TYPE_EPISODE)
-        return case_accessor.iter_cases(case_ids)
+        episode_cases = case_accessor.iter_cases(case_ids)
+
+        for episode_case in episode_cases:
+            # if this episode is part of a deleted or archived person, don't update
+            try:
+                person_case = get_person_case_from_episode(self.domain, episode_case.case_id)
+            except ENikshayCaseNotFound:
+                continue
```
proteusvacuum: This happens when the person case was deleted during a migration event. At some point, we'll probably want to close / delete all the episodes that fall into this category, but will leave that as a `TODO`, as that process cascades pretty heavily (since there are __many__ related extension cases that are attached to an episode)

sravfeyn (approved):

PR #17305, opened 2017-08-01
jQuery.get and .post don't take "error" anymore.

kaapstorm:
Since [jQuery 3](https://api.jquery.com/jQuery.get/#jQuery-get-settings),
```javascript
    $.get({
        // ...
        error: function () { alert("I never get called."); },
    });
```
is no longer a thing.

@dannyroberts @esoergel 

I think this is the start. It looks like there might be 47 more of these ...
```bash
$ ggr 'error: function' | wc -l
47
```

### DIFF
--- a/corehq/apps/app_manager/static/app_manager/js/releases/releases.js
+++ b/corehq/apps/app_manager/static/app_manager/js/releases/releases.js
@@ -375,17 +375,13 @@ hqDefine('app_manager/js/releases/releases.js', function () {
                 "Please reload your page and try again");
         self.deleteSavedApp = function (savedApp) {
             savedApp._deleteState('pending');
-            $.post({
-                url: self.reverse('delete_copy'),
-                data: {saved_app: savedApp.id()},
-                success: function () {
-                    self.savedApps.remove(savedApp);
-                    savedApp._deleteState(false);
-                },
-                error: function () {
-                    savedApp._deleteState('error');
-                    alert(self.reload_message);
-                },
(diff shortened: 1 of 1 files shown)


orangejenny:
Confused - the [docs for get](https://api.jquery.com/jQuery.get/) say that if you use the `jQuery.get( [settings ] )` signature, the settings object has all the same options that `jQuery.ajax` accepts, which does include `error`. Is this not working?

kaapstorm:
I was also confused! I assumed this has been working fine on production. But it was definitely not working locally for me&mdash;the callback of the `error` settings option was not getting called&mdash;and this change resolved it. 

After reading your comment I rechecked the docs, and I realised the deprecation notice ...

> The jqXHR.success(), jqXHR.error(), and jqXHR.complete() callback methods are removed as of jQuery 3.0. You can use jqXHR.done(), jqXHR.fail(), and jqXHR.always() instead.

... applies to callbacks, not settings options, just as you said.

So I don't know why this resolves my local bug. And I don't know what to do with this PR. Labelling "Hold off on merge", pending suggestions.

kaapstorm:
cc new code buddy @nickpell

nickpell (approved):

orangejenny:
That's weird...the original version works for me locally (when I change `delete_copy` to always 404, the error callback is triggered). Confirming that you've updated bower recently? Thinking about what else to try, don't have any immediate ideas :-/

PR #17306, opened 2017-08-01
Change xmlns on copy form

millerdev:
https://manage.dimagi.com/default.asp?259314

@orangejenny @NoahCarnahan cc @gcapalbo

### DIFF
--- a/corehq/apps/app_manager/util.py
+++ b/corehq/apps/app_manager/util.py
@@ -134,9 +134,11 @@ def save_xform(app, form, xml):
     else:
         GENERIC_XMLNS = "http://www.w3.org/2002/xforms"
         tag_xmlns = xform.data_node.tag_xmlns
+        form_xmlns = "http://openrosa.org/formdesigner/%s" % form.get_unique_id()
         if not tag_xmlns or tag_xmlns == GENERIC_XMLNS:  # no xmlns
-            xmlns = "http://openrosa.org/formdesigner/%s" % form.get_unique_id()
-            xml = change_xmlns(xform, GENERIC_XMLNS, xmlns)
+            xml = change_xmlns(xform, GENERIC_XMLNS, form_xmlns)
+        elif tag_xmlns != form_xmlns:
+            xml = change_xmlns(xform, tag_xmlns, form_xmlns)
 
     form.source = xml
 
--- a/corehq/apps/app_manager/util.py
+++ b/corehq/apps/app_manager/util.py
@@ -133,12 +133,24 @@ def save_xform(app, form, xml):
         pass
     else:
         GENERIC_XMLNS = "http://www.w3.org/2002/xforms"
+        # we assume form.get_unique_id() is unique across all of HQ and
+        # therefore is suitable to create an XMLNS that will not confict
+        # with any other form
+        uid = form.get_unique_id()
         tag_xmlns = xform.data_node.tag_xmlns
-        form_xmlns = "http://openrosa.org/formdesigner/%s" % form.get_unique_id()
+        new_xmlns = form.xmlns or "http://openrosa.org/formdesigner/%s" % uid
         if not tag_xmlns or tag_xmlns == GENERIC_XMLNS:  # no xmlns
-            xml = change_xmlns(xform, GENERIC_XMLNS, form_xmlns)
-        elif tag_xmlns != form_xmlns:
-            xml = change_xmlns(xform, tag_xmlns, form_xmlns)
+            xml = change_xmlns(xform, GENERIC_XMLNS, new_xmlns)
+        else:
+            forms = [form_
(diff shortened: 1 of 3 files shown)


millerdev:
@NoahCarnahan, turns out [the XMLNS change logic](https://github.com/dimagi/commcare-hq/pull/17250/files#r130123694) is necessary. I realize it's still slightly different than before because, as you pointed out, it always changes the XMLNS to match the form's `unique_id` if it does not match, which seems like probably a good thing, but that might have bad side effects I'm not thinking of? Can you think of any?

corehq/apps/app_manager/util.py
```diff
@@ -134,9 +134,11 @@ def change_xmlns(xform, old_xmlns, new_xmlns):
     else:
         GENERIC_XMLNS = "http://www.w3.org/2002/xforms"
         tag_xmlns = xform.data_node.tag_xmlns
+        form_xmlns = "http://openrosa.org/formdesigner/%s" % form.get_unique_id()
         if not tag_xmlns or tag_xmlns == GENERIC_XMLNS:  # no xmlns
-            xmlns = "http://openrosa.org/formdesigner/%s" % form.get_unique_id()
-            xml = change_xmlns(xform, GENERIC_XMLNS, xmlns)
+            xml = change_xmlns(xform, GENERIC_XMLNS, form_xmlns)
```
NoahCarnahan: Will this `change_xmlns()` call work if `tag_xmlns` was falsey?

corehq/apps/app_manager/util.py
```diff
@@ -134,9 +134,11 @@ def change_xmlns(xform, old_xmlns, new_xmlns):
...
+        elif tag_xmlns != form_xmlns:
+            xml = change_xmlns(xform, tag_xmlns, form_xmlns)
```
NoahCarnahan: It has not been my understanding that form XMLNS always matches form.unique_id (but this assumption might be outdated). I wonder if changing the xmlns here could mess up workflows with existing reports or exports or that sort of thing? You've been thinking about it more than I, so feel free to ignore if you don't think this will be an issue, but just something that crossed my mind.

NoahCarnahan (approved):
Had two questions, but feel free to merge if you think they are non-issues.

corehq/apps/app_manager/util.py
millerdev: I'm not sure, but I don't think this has changed from the way it was before.

corehq/apps/app_manager/util.py
millerdev: @dannyroberts, @snopoke or anyone in @dimagi/team-commcare-hq do you know if this will cause problems? Also see https://github.com/dimagi/commcare-hq/pull/17250 for more context.

millerdev: nevermind. looks like there are tests that depend on the old way. Seems bad that xmlns is derived from `form.unique_id` and both are used to reference forms, but they can get out of sync...

emord: This is definitely going to break things. It looks like everytime an xform is saved it will check that the xmlns and form_unique_id are the same and if it isn't, it changes the xmlns. I checked icds and those two are not the same.

emord: @millerdev curious why you think xmlns is derived from form.unique_id (that may be just the case for shadow forms) ?

I think xmlns is usually done randomly from vellum: https://github.com/dimagi/Vellum/blob/747d5b9c2d4054563f831766ad0b0f503e079bc5/src/util.js?utf8=%E2%9C%93#L170

millerdev: @emord I thought xmlns was derived from `form.unique_id` because of this code: https://github.com/dimagi/commcare-hq/blob/46c4645394eae4474a21f00a7028fed8c95273b5/corehq/apps/app_manager/util.py#L124-L125

That's right, Vellum does generate unique IDs too, although it's always conditional:

- https://github.com/dimagi/Vellum/blob/747d5b9c2d4054563f831766ad0b0f503e079bc5/src/parser.js#L196-L198
- https://github.com/dimagi/Vellum/blob/747d5b9c2d4054563f831766ad0b0f503e079bc5/src/writer.js#L83-L86

I get so confused trying to find a source of truth in HQ.

emord: Yeah. I think that's only used if there's a conflict of xmlns. by default xmlns != form_unique_id. I only know that because I've broken that before...

emord: Oh. also I think form_unique_id is unique for all of HQ, but xmlns is only unique for the app. That needs to work for linked/copied apps I think

millerdev:
Wrote a set of tests for various xmlns changing (or not) scenarios and adjusted the implementation to make them all pass.

PR #17309, opened 2017-08-01
Add nullable columns to UCR without rebuild

emord:
@NoahCarnahan 

These tests are going to fail because the metadata is a global variable. I haven't figured it out yet.

Doesn't do anything smart with the rebuild

buddy @biyeun


NoahCarnahan (approved):
👍

snopoke (approved):

emord:
I think I've fixed the tests. They feel pretty hacky and it does make me wonder how well the rebuilding of tables actually works in the pillow.

PR #17313, opened 2017-08-02
JS error on ews-ghana dashboard

kkrampa:


kkrampa:
@orangejenny 
http://manage.dimagi.com/default.asp?257488

orangejenny (approved):
thanks!

PR #17316, opened 2017-08-02
sync log is None on 412

snopoke:
Introduced here: https://github.com/dimagi/commcare-hq/commit/145e527945a1b6a81940e86f913532b6b4cbf9dc


millerdev (commented):
Thanks for fixing, @snopoke

PR #17317, opened 2017-08-02
Analytics: event for submitting form in app preview

orangejenny:
http://manage.dimagi.com/default.asp?259520

@benrudolph

### DIFF
--- a/corehq/apps/cloudcare/static/cloudcare/js/form_entry/fullform-ui.js
+++ b/corehq/apps/cloudcare/static/cloudcare/js/form_entry/fullform-ui.js
@@ -255,7 +255,6 @@ function Form(json) {
     self.submitForm = function(form) {
         window.analytics.workflow("[app-preview] User submitted a form");
         window.analytics.usage("[app-preview] User submitted a form");
-console.log("did stuff (either button or complete? OQPS)");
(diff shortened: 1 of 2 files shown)


benrudolph (commented):
this won't distinguish between web apps, edit forms, and app preview. so `[app-preview]` prefix is kind of misleading

orangejenny:
@benrudolph Thanks, didn't know that, and I do want this to be only for live preview. Is putting the event in preview_app.js the reasonable thing to do, then?

benrudolph:
ok hmm i think what you can do is put the event here: https://github.com/dimagi/commcare-hq/blob/master/corehq/apps/cloudcare/static/cloudcare/js/formplayer/app.js#L143
and then use the `user.environment` to figure out if it is web apps or app preview. see possible values here: https://github.com/dimagi/commcare-hq/blob/master/corehq/apps/cloudcare/static/cloudcare/js/formplayer/constants.js

form entry code should have no knowledge of where it lives, but there is one variable that is passed in to determine if it's in phone mode. though still think it makes sense to do it in the callback

orangejenny:
@benrudolph Cool, thanks. Updated.

benrudolph (approved):

PR #17319, opened 2017-08-02
move more sharding logic into django models

snopoke:
Can review per commit :fishing_pole_and_fish: 

buddy @orangejenny

### DIFF
--- a/corehq/form_processor/backends/sql/dbaccessors.py
+++ b/corehq/form_processor/backends/sql/dbaccessors.py
@@ -392,7 +392,7 @@ class FormAccessorSQL(AbstractFormAccessor):
         for db_name, split_form_ids in split_list_by_db_partition(form_ids):
             # cascade should delete the attachments and operations
             _, deleted_models = XFormInstanceSQL.objects.using(db_name).filter(
-                domain=domain, form_id__in=form_ids
+                domain=domain, form_id__in=split_form_ids
             ).delete()
             deleted_count += deleted_models.get(XFormInstanceSQL._meta.label, 0)
 
--- a/corehq/form_processor/backends/sql/dbaccessors.py
+++ b/corehq/form_processor/backends/sql/dbaccessors.py
@@ -38,26 +38,17 @@ from corehq.form_processor.models import (
     CommCareCaseSQL,
     XFormAttachmentSQL,
     XFormOperationSQL,
-    CommCareCaseIndexSQL_DB_TABLE,
-    CaseAttachmentSQL_DB_TABLE,
-    LedgerTransaction_DB_TABLE,
-    LedgerValue_DB_TABLE,
     LedgerValue,
     LedgerTransaction,
 )
 from corehq.form_processor.utils.sql import (
     fetchone_as_namedtuple,
-    fetchall_as_namedtuple,
-    case_adapter,
-    case_transaction_adapter,
-    case_index_adapter,
-    case_attachment_adapter
+    fetchall_as_namedtuple
 )
 from corehq.sql_db.config import get_sql_db_aliases_in_use, partition_config
 from corehq.sql_db.routers import db_for_read_write
 from corehq.sql_db.util import get_db_alias_for_partitioned_doc, split_list_by_db_partition
 from corehq.util.queries import fast_distinct_in_domain
-from corehq.util.test_utils import unit_testing_only
 from dimagi.utils.chunked import chunked
 
 doc_type_to_state = {
--- a/corehq/form_processor/backends/sql/dbaccessors.py
+++ b/corehq/form_processor/backends/sql/dbaccessors.py
@@ -47,7 +47,7 @@ from corehq.form_processor.utils.sql import (
 )
 from corehq.sql_db.config import get_sql_db_aliases_in_use, partition_config
 from corehq.sql_db.routers import db_for_read_write
-from corehq.sql_db.util import get_db_alias_for_partitioned_doc, split_list_by_db_partition
+from corehq.sql_db.util import split_list_by_db_partition
 from corehq.util.queries import fast_distinct_in_domain
 from dimagi.utils.chunked import chunked
 
@@ -503,12 +503,10 @@ class FormAccessorSQL(AbstractFormAccessor):
     @staticmethod
     @transaction.atomic
     def save_new_form(form):
-        from corehq.sql_db.util import get_db_alias_for_partitioned_doc
         """
         Save a previously unsaved form
         """
 
-        db_name = get_db_alias_for_partitioned_doc(form.form_id)
         assert not form.is_saved(), 'form already saved'
         logging.debug('Saving new form: %s', form)
         unsaved_attachments = getattr(form, 'unsaved_attachments', [])
@@ -529,7 +527,7 @@ class FormAccessorSQL(AbstractFormAccessor):
             operation.form = form
 
         try:
-            with transaction.atomic(using=db_name, savepoint=False):
+            with transaction.atomic(using=form.db, savepoint=False):
                 form.save()
                 for attachment in unsaved_attachments:
                     attachment.save()
@@ -683,9 +681,7 @@ class CaseAccessorSQL(AbstractCaseAccessor):
 
     @staticmethod
     def case_exists(case_id):
-        from corehq.sql_db.util import get_db_alias_for_partitioned_doc
-        db = get_db_alias_for_partitioned_doc(case_id)
-        return CommCareCaseSQL.objects.using(db).filter(case_id=case_id).exists()
+        return CommCareCaseSQL.objects.partitioned_query(case_id).filter(case_id=case_id).exists()
 
     @staticmethod
     def get_case_xform_ids(case_id):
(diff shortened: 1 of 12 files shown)


corehq/form_processor/tests/test_form_dbaccessor.py
```diff
@@ -401,13 +401,10 @@ class FormAccessorsTestsSQL(FormAccessorsTests):
...
         super(DeleteAttachmentsFSDBTests, self).setUp()
-        self.rootdir = mkdtemp(prefix="blobdb")
-        self.db = FilesystemBlobDB(self.rootdir)
+        self.db = TemporaryFilesystemBlobDB()
 
     def tearDown(self):
-        self.db = None
-        rmtree(self.rootdir)
-        self.rootdir = None
+        self.db.close()
```
snopoke: Unrelated to this PR but I changed this while looking at a test failure

corehq/form_processor/backends/sql/dbaccessors.py
```diff
@@ -388,9 +388,8 @@ def hard_delete_forms(domain, form_ids, delete_attachments=True):
         if delete_attachments:
             attachments = list(FormAccessorSQL.get_attachments_for_forms(form_ids))
 
-        db_form_ids = split_list_by_db_partition(form_ids)
         deleted_count = 0
-        for db_name, form_ids in db_form_ids.items():
+        for db_name, split_form_ids in split_list_by_db_partition(form_ids):
             # cascade should delete the attachments and operations
             _, deleted_models = XFormInstanceSQL.objects.using(db_name).filter(
                 domain=domain, form_id__in=form_ids
```
gcapalbo: should this use `split_form_ids`?

corehq/sql_db/routers.py
```diff
@@ -22,6 +22,15 @@ def db_for_write(self, model, **hints):
...
 
+    def allow_relation(self, obj1, obj2, **hints):
+        from corehq.sql_db.models import PartitionedModel
+        obj1_partitioned = isinstance(obj1, PartitionedModel)
+        obj2_partitioned = isinstance(obj2, PartitionedModel)
+        if obj1_partitioned and obj2_partitioned:
+            return obj1.db == obj2.db
+        elif not obj1_partitioned and not obj2_partitioned:
+            return True
+
```
gcapalbo: should this explicitly return False or is returning None ok?

gcapalbo (approved):
a lot of great stuff here!

corehq/form_processor/backends/sql/dbaccessors.py
snopoke: yes it should thanks

corehq/sql_db/routers.py
snopoke: In our case it has the same effect since there are no other routers but better to be explicit

millerdev:
Just looking at the title of this PR makes me wonder... It seems like we're doing something wrong if we can't get our proxy db to perform at the level we need, especially after seeing the citus demo and noting that it has basically the came architecture as plproxy (central proxy node), which is apparently not a problem for them.

Moving sharding logic into python seems wrong because python is quite slow compared to postgres, so I will not be surprised if we don't encounter more/worse performance issues due to this move down the road.

snopoke:
@millerdev the bottleneck is number of connections to PG the way to get round this is to either have a dedicated node for pl_proxy or to have multiple pl_proxy nodes that you load balance across.

The approach (moving sharding logic to python) will keep the infrastructure setup simpler at the cost of adding some complexity to python code (though that complexity is technically already there because of what we have setup with the SQL functions). 

There is another reason for doing it in python which is that it gives us more control over the sharding which we will need when we split the shard databases. 

Happy to discuss this more if you think this is going to cause problems. Re performance what do you think is going be worse? The only thing I can think of is perhaps when we query multiple databases that pl_proxy might do that in parallel.

millerdev:
@snopoke 
> the bottleneck is number of connections to PG

I thought we used pgbouncer to get around that problem. That's even what the citusdb guy suggested when we mentioned the problem of the proxy being a bottleneck.

> gives us more control over the sharding which we will need when we split the shard databases

Not sure what you mean by this. Are you saying we're going to have multiple different sharded databases? Why would we do that? In terms of adding shards and moving them to different machines, if we do that correctly it should be transparent to the clients (python code accessing the database).

> Re performance what do you think is going be worse?

Adding the sharding logic to python means we're moving another layer of our database into our application code, which means it is more likely to become tightly coupled with the application implementation and can be harder to analyze performance bottlenecks. Python is slow, and when we're optimizing for performance we should either be offloading the processing to fast things like postgres or or maybe a C-implementation of some time-critical logical component, or doing it in parallel with more python processes (although this can get heavy and expensive in terms of hardware).

snopoke:
@millerdev 

We do use pgbouncer. We haven't really had any issues with this except on NIC when there are resource conflicts on the hosts that cause the DB to suffer and queries to take longer. This pushes the number of concurrent requests up and hence the number of connections to the DB.

As an example of how this works in practice, on NIC we have 1024 logical shards split across 10 physical databases (5 on each of 2 DB VMs). One of those VMs also has the pl_proxy DB on it. All requests to those 10 databases go through pl_proxy so if we have 25 connections to each of the 10 DBs then the PG instance with pl_proxy has 25*5 + 250 PG connections = 375 connections. Obviously there are ways around this. Firstly moving the pl_proxy node to a separate VM would make a big difference etc.

> Not sure what you mean by this. Are you saying we're going to have multiple different sharded databases? Why would we do that? In terms of adding shards and moving them to different machines, if we do that correctly it should be transparent to the clients (python code accessing the database).

We already have multiple physical shards. In terms of moving shards (e.g. splitting a physical shard) yes, it would be transparent to the client but during the process of actually moving the data the client wouldn't be able to connect at all. If we have the sharding logic in python we have more flexibility in terms of telling the client where to look for the data.

> Adding the sharding logic to python means we're moving another layer of our database into our application code, which means it is more likely to become tightly coupled with the application implementation and can be harder to analyze performance bottlenecks. 

Sure, I agree with that generally though the sharding logic is pretty well isolated and easy to test.

> Python is slow, and when we're optimizing for performance we should either be offloading the processing to fast things like postgres or or maybe a C-implementation of some time-critical logical component, or doing it in parallel with more python processes (although this can get heavy and expensive in terms of hardware).

Considering the only thing we're moving to python is the query routing I don't expect there is going to much loss in performance. The cost of running the query is going to far outweigh the cost of routing the query to the appropriate DB. The function that's doing the 'hashing' as part of the shard logic is already in C.

millerdev:
> We haven't really had any issues with this except on NIC when there are resource conflicts on the hosts that cause the DB to suffer and queries to take longer. This pushes the number of concurrent requests up and hence the number of connections to the DB.

Considering the hardware/virtualization situation there (i.e., way beyond reasonably limited resources) and that this issue may be entirely a symptom of that, it makes me sad that we would be making such a profound architectural change to work around that one very unfortunate situation.

With regard to the difficulty of scaling the number of servers in our pl_proxy cluster, are you saying there's no  way to do move (virtual) shards without downtime? I'm not talking about changing the number of (virtual) shards, which I understand is not something we ever want to do. If that's true that seems like a pretty big limitation of pl_proxy, and a good argument to switch to something like citusdb. The most impressive part of the citus demo was how scaling was completely transparent to clients, and it did not require any downtime.

snopoke:
Pl_proxy has no tools or features to allow moving shard around. It would be
a completely manual process.

corehq/sql_db/models.py
```diff
@@ -57,17 +93,35 @@ class PartitionedModel(models.Model):
...
+        """The partitioned database for this object"""
+        assert self.partition_value, 'Partitioned model must have a partition value'
+        return RequireDBManager.get_db(self.partition_value)
+
     class Meta:
         abstract = True
 
     def save(self, *args, **kwargs):
-        if not kwargs.get('using'):
-            raise_access_restricted()
```
millerdev: 👍 nice to see this go

corehq/sql_db/models.py
```diff
@@ -57,17 +93,35 @@ class PartitionedModel(models.Model):
...
+            assert kwargs['using'] == self.db
+        else:
+            kwargs['using'] = self.db
 
         return super(PartitionedModel, self).save(*args, **kwargs)
 
     def delete(self, *args, **kwargs):
-        if not kwargs.get('using'):
-            raise_access_restricted()
+        if kwargs.get('using'):
```
millerdev: same as above

millerdev (commented):
Finally got around to reviewing the actual code, and it looks like there was a fair amount of cleanup that went into this effort as well, which is very nice.

I'm still not sold on moving partition logic into Python being a good idea, mostly because I think our motivation (proxy seems like bottleneck on badly configured cluster) is not solid.

corehq/sql_db/models.py
```diff
@@ -57,17 +93,35 @@ class PartitionedModel(models.Model):
...
+        assert self.partition_value, 'Partitioned model must have a partition value'
+        return RequireDBManager.get_db(self.partition_value)
+
     class Meta:
         abstract = True
 
     def save(self, *args, **kwargs):
-        if not kwargs.get('using'):
-            raise_access_restricted()
+        if kwargs.get('using'):
```
snopoke: https://github.com/dimagi/commcare-hq/pull/17376

PR #17320, opened 2017-08-02
Allow sending to case owner from new scheduling framework

gcapalbo:
@emord @snopoke

### DIFF
--- a/corehq/messaging/scheduling/scheduling_partitioned/models.py
+++ b/corehq/messaging/scheduling/scheduling_partitioned/models.py
@@ -4,7 +4,9 @@ from corehq.apps.casegroups.models import CommCareCaseGroup
 from corehq.apps.groups.models import Group
 from corehq.apps.locations.dbaccessors import get_all_users_by_location
 from corehq.apps.locations.models import SQLLocation
+from corehq.apps.users.cases import get_owner_id, get_wrapped_owner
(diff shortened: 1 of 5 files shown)


corehq/messaging/scheduling/scheduling_partitioned/models.py
```diff
@@ -246,10 +249,21 @@ class CaseScheduleInstanceMixin(object):
...
+            return CaseAccessors(self.domain).get_case(self.case_id)
+        except (CaseNotFound, ResourceNotFound):
+            return None
+
+    @property
+    @memoized
     def recipient(self):
         if self.recipient_type == 'Self':
-            return CaseAccessors(self.domain).get_case(self.case_id)
+            return self.case
```
emord: Looks like this will return None where before it was throwing an exception. is that an ok thing to do and the callers expect a possible None value?

emord (commented):
Just one question. Approved if it's nothing to worry about

corehq/messaging/scheduling/scheduling_partitioned/models.py
gcapalbo: yep - it was actually a bad thing for it to raise an exception before since the framework would infinitely just keep retrying it. but yeah the framework handles a None return value from a recipient lookup appropriately where it's referenced

PR #17328, opened 2017-08-02
Mark private Nikshay failed records as failed

proteusvacuum:
We were also making requests to the Nikshay servers every time we run our tests.

https://manage.dimagi.com/default.asp?259280

@mkangia 
code buddy @dannyroberts

### DIFF
--- a/custom/enikshay/integrations/nikshay/repeater_generator.py
+++ b/custom/enikshay/integrations/nikshay/repeater_generator.py
@@ -417,36 +417,35 @@ class NikshayRegisterPrivatePatientPayloadGenerator(SOAPPayloadGeneratorMixin, B
         }
 
     def handle_success(self, response, payload_doc, repeat_record):
-        # A successful response returns a Nikshay ID like 00001
-        # Failures also return with status code 200 and some message like
-        # Dublicate Entry or Invalid data format
-        # (Dublicate is not a typo)
         message = parse_SOAP_response(
             repeat_record.repeater.url,
             repeat_record.repeater.operation,
-            response
+            response,
         )
         try:
-            if isinstance(message, basestring) and message.isdigit():
-                health_facility_id = self._get_person_locations(payload_doc).pcp
-                nikshay_id = '-'.join([health_facility_id, message])
-                update_case(
-                    payload_doc.domain,
-                    payload_doc.case_id,
-                    {
-                        "private_nikshay_registered": "true",
-                        "nikshay_id": nikshay_id,
-                        "private_nikshay_error": "",
-                    },
-                    external_id=nikshay_id,
-                )
-            else:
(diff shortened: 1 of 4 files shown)


custom/enikshay/integrations/nikshay/repeater_generator.py
```diff
@@ -417,36 +417,35 @@ def get_payload(self, repeat_record, episode_case):
...
+                    "private_nikshay_registered": "true",
+                    "nikshay_id": nikshay_id,
+                    "private_nikshay_error": "",
+                },
+                external_id=nikshay_id,
+            )
 
     def handle_failure(self, response, payload_doc, repeat_record):
-        _save_error_message(payload_doc.domain, payload_doc.case_id, unicode(response),
+        message = parse_SOAP_response(
```
proteusvacuum: we have to do this basically everywhere we want to deal with the response in these SOAP repeaters. Not sure if it would be best to delegate this to the mixin to do for us?

mkangia:
@proteusvacuum thanks for fixing this. 
just confirming we were sending requests to Nikshay for just the wsdl request and not actually submitting records when running tests?

proteusvacuum:
@mkangia yeah, we were just requesting the WSDL

proteusvacuum:
@mkangia @dannyroberts ✅

custom/enikshay/integrations/nikshay/repeater_generator.py
mkangia: we can also modify the `send_request` on the mixin and the `perform_SOAP_operation` to control when we need raw response or just parsed response. Basically, If we skip `raw_response=True` zeep will parse the SOAP response and give us just the final value. The reason i picked it to keep it raw here so that we store the complete response and not just the text response. If there is no need to store the complete xml response we can just receive the parsed response here.
And then this won't be needed everywhere and can be made repeater specific.

mkangia: Also don't see why we are passing the response instead of just the message since we have parsed the response already in the repeater? Since we are storing just the final message anyway now.

mkangia (changes requested):

custom/enikshay/integrations/nikshay/repeater_generator.py
proteusvacuum: So that the arguments of this method receive the same thing as for all of the other classes that inherit from `RepeaterGenerator`.

mkangia: hmm.. Can we pass the parsed response along with the actual response for SOAP generators? Its not relevant to this PR so won't block on this here but agree that this can be improved.

mkangia (approved):

PR #17217, opened 2017-07-26
enikshay: 2b DRTB migration (update #2)

NoahCarnahan:
The enikshay project requires patient data collected in excel spreadsheets to be imported into HQ. This management command parses those spreadsheets and creates the corresponding cases. There are three different formats, and each format contains some, but not all, of the same data.

This management command is a work in progress, but I've been PRing it in chunks as I go to make it easier to review.

@biyeun 
best reviewed commit-by-commit

(original PR here for reference: https://github.com/dimagi/commcare-hq/pull/16958)

PR #17330, opened 2017-08-02
enikshay: 2b DRTB migration (update #3)

NoahCarnahan:
Follow up to https://github.com/dimagi/commcare-hq/pull/17217

A whole bunch of updates to the management command based on latest requirements.
Still in progress :(

Commits do one thing, but there are a lot of them, so might be easier to review all at once.

@millerdev


custom/enikshay/two_b_datamigration/management/commands/import_drtb_cases.py
```diff
@@ -1319,17 +1612,84 @@ def match_facility(domain, xlsx_facility_name):
...
+def get_drtb_hiv_location(domain, district_id):
+    if not district_id:
+        return None, None
+    drtb_hiv = SQLLocation.get(domain=domain, parent__location_id=district_id, location_type__code="drtb-hiv")
+    return drtb_hiv.name, drtb_hiv.location_id
+
+
+class PersonIdGenerator(object):
+
+    dry_run_counter = 0
```
millerdev: This is interpreter-level global state, which seems like not a great thing. Can you justify that? I suppose since it's part of a management command it's not a big deal?

custom/enikshay/two_b_datamigration/management/commands/import_drtb_cases.py
```diff
@@ -1357,48 +1717,64 @@ def add_arguments(self, parser):
...
-                        logger.info("Creating cases for row {}".format(i))
-
-                        if options['commit']:
-                            case_factory.create_or_update_cases(case_structures)
-                    except Exception:
-                        logger.info("Creating case structures for row {} failed".format(i))
-                        exception_as_string = traceback.format_exc()
-                        import_log_writer.writerow([i, "", exception_as_string])
+            with open(bad_rows_file_name, "w") as bad_rows_file:
+                with open(import_log_file_name, "w") as import_log_file:
```
millerdev: nit/tip: instead of using nested with statements, you can [open both files in the same `with` statement](https://docs.python.org/2/reference/compound_stmts.html#the-with-statement) and avoid the extra level of indentation.

millerdev (approved):

custom/enikshay/two_b_datamigration/management/commands/import_drtb_cases.py
NoahCarnahan: In any given dry run, I want a new number each time _next_serial_count is called.
Also, I don't expect this class to be used outside of the management command. I'll change its name to signify that it should be a private class though to make sure there is no confusion.

NoahCarnahan:
Thanks for the feedback Daniel. Pushed a few more commits.

millerdev (approved):

millerdev (approved):

PR #17333, opened 2017-08-02
Record restore timing for async restores

millerdev:
Should also fix the issue with "commcare.restores.cases" and "commcare.restores.fixtures" datadog counters having no data for enikshay.

@proteusvacuum cc @ctsims

### DIFF
--- a/corehq/apps/ota/views.py
+++ b/corehq/apps/ota/views.py
@@ -1,7 +1,5 @@
-import logging
 from distutils.version import LooseVersion
 
-from django.conf import settings
 from django.http import JsonResponse
 from django.urls import reverse
 from django.shortcuts import redirect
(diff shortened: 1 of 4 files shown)


corehq/apps/ota/views.py
```diff
@@ -277,15 +231,9 @@ def get_restore_response(domain, couch_user, app_id=None, since=None, version='1
...
-    if timing.duration > 20 or response.status_code == 412:
-        sync_log_id = None
-        if restore_config.restore_state.current_sync_log:
-            sync_log_id = restore_config.restore_state.current_sync_log._id
-        log = logging.getLogger(__name__)
-        log.info("restore %s: domain=%s status=%s duration=%.3f",
-                 sync_log_id, domain, response.status_code, timing.duration)
-    return response, timing
+    if not async_restore_enabled:
+        record_restore_timing(restore_config, response.status_code)
```
millerdev: Do not record timings for async restores since since they don't have real restore timing info.

proteusvacuum: except you do, in the async restore method later on right?

millerdev: well yes, but that's done in a separate (celery worker) process, which is fired off by this call.

millerdev:
@proteusvacuum were you expecting me to make more changes to this PR? If not, can you approve?

millerdev:
@proteusvacuum actually I'm adding one more thing here... updated.

proteusvacuum:
@millerdev is there something that needs to be done with the `RestoreException`? I was under the impression that that exception is only thrown when there is a hash mismatch, which happens outside of fetching the restore payload.

millerdev:
@proteusvacuum that make sense. Will amend this. Thanks!

corehq/ex-submodules/casexml/apps/phone/tasks.py
```diff
@@ -37,7 +38,16 @@ def force_update_cleanliness_flags():
...
+    status = 'err'
+    try:
+        with restore_config.timing_context:
+            response = restore_config.generate_payload(async_task=current_task)
+        status = 200
+    except RestoreException:
+        status = 412
+        raise
+    finally:
+        record_restore_timing(restore_config, status)
```
proteusvacuum: @millerdev , thinking about this a bit further, is there a way you can make this track how long the whole process took, from when the task was kicked off, to when a response was received? I'm worried that this still doesn't take into account the fact that celery might not run this task right away, and there is some overhead in there.

millerdev: good question. I could probably do that, but it's going to take a bit of engineering. Can we merge this PR and I'll implement that separately?

proteusvacuum (approved):

PR #17334, opened 2017-08-03
Don't apply owner transform in ASR report

kkrampa:
@emord 
https://manage.dimagi.com/default.asp?255223
I find this as a bottleneck since it needed to hit a database a lot of times to transform owner_id to owner name. Currently, we don't need this in ASR report so I'm removing all this transform before fetching the data.

### DIFF
--- a/custom/icds_reports/utils.py
+++ b/custom/icds_reports/utils.py
@@ -56,10 +56,18 @@ class ICDSData(object):
 
     def __init__(self, domain, filters, report_id):
         report_config = ReportFactory.from_spec(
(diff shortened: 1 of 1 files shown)


emord (approved):
This seems ok. Is there a reason why you did it this way instead of changing the report?

kkrampa:
@emord I don't know if any of those backing UCRs is really used so I can't remove transform directly from UCR.

Is this change deployed to icds-cas server?

emord:
It was deployed yesterday

PR #17335, opened 2017-08-03
function returns list not dict

snopoke:


snopoke:
Merging to fix build on master

PR #17337, opened 2017-08-03
fix edit form with attachments

snopoke:
If a form has attachments e.g. images then when the form is edited the attachments don't get copied over to the new form.

buddy @orangejenny

### DIFF
--- a/corehq/form_processor/parsers/form.py
+++ b/corehq/form_processor/parsers/form.py
@@ -215,15 +215,6 @@ def apply_deprecation(existing_xform, new_xform, interface=None):
 
     interface = interface or FormProcessorInterface(existing_xform.domain)
     interface.copy_attachments(existing_xform, new_xform)
-    if existing_xform.persistent_blobs:
-        for name, meta in existing_xform.persistent_blobs.items():
-            with existing_xform.fetch_attachment(name, stream=True) as content:
-                existing_xform.deferred_put_attachment(
-                    content,
-                    name=name,
-                    content_type=meta.content_type,
-                    content_length=meta.content_length,
-                )
     new_xform.form_id = existing_xform.form_id
     existing_xform = interface.assign_new_id(existing_xform)
     existing_xform.orig_id = new_xform.form_id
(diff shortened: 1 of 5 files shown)


benrudolph (approved):
nice catch. thought for a second it might be related to this ticket: http://manage.dimagi.com/default.asp?258719#1374321. but alas i don't think those forms were edited

corehq/form_processor/parsers/form.py
```diff
@@ -214,16 +214,7 @@ def apply_deprecation(existing_xform, new_xform, interface=None):
     #  - Save the new instance to the previous document to preserve the ID
 
     interface = interface or FormProcessorInterface(existing_xform.domain)
-
-    if existing_xform.persistent_blobs:
-        for name, meta in existing_xform.persistent_blobs.items():
-            with existing_xform.fetch_attachment(name, stream=True) as content:
-                existing_xform.deferred_put_attachment(
```
millerdev: where these calls failing on SQL forms in the past? If not, why didn't it work for SQL forms?

millerdev (commented):
It's a shame that SQL forms don't implement the same interface as Couch forms for attachments.

snopoke:
@millerdev this code was originally added because when a form ID changes (as happens here in the edit form workflow) the references to the blobs is lost because it's based on the form ID. This code never executed for SQL forms since `form.persistent_blobs` is hardcoded as `None` in their models. It also never executed for Couch forms because the code I linked to already puts all the attachments in the `deferred` buffer and so `existing_xform.persistent_blobs` returns an empty dict.

This PR is addressing a different issue altogether where an edited forms loses the references to any attachments that it had.

millerdev:
@snopoke, thanks for the explanation. That makes sense.

> This code never executed for SQL forms since `form.persistent_blobs` is hardcoded as `None` in their models. It also never executed for Couch forms because the code I linked to already puts all the attachments in the `deferred` buffer and so `existing_xform.persistent_blobs` returns an empty dict.

ouch

PR #17339, opened 2017-08-03
Add custom content handlers to render missed visit messages

gcapalbo:
@emord @snopoke


corehq/apps/users/models.py
```diff
@@ -2096,6 +2096,11 @@ def __repr__(self):
             self=self
         ))
 
+    @property
+    @memoized
```
emord: nit: I think we can start using cached_property https://docs.djangoproject.com/en/1.11/ref/utils/#django.utils.functional.cached_property

custom/icds/messaging/custom_recipients.py
```diff
@@ -16,6 +16,21 @@ def get_child_health_host_case(domain, extension_case_id):
     return related[0]
 
 
+def mother_person_case_from_ccs_record_case(case_schedule_instance):
+    if case_schedule_instance.case.type != 'ccs_record':
+        return None
```
emord: Would it be worth soft asserting/sending to sentry if this happens?

emord (approved):

emord (approved):

corehq/apps/users/models.py
gcapalbo: that's a good point, they seem to do the same exact thing. seems like we should just pick one and use it throughout

custom/icds/messaging/custom_recipients.py
gcapalbo: it's probably fine as is, we only spawn this from ccs_record cases

corehq/apps/users/models.py
emord: Yeah. I found it and mentioned it on #web-team a few weeks ago. Haven't gotten around to migrating our current usage over. I think the double decorator just comes from a time before django had implemented that (their implementation is also much much smaller/simpler than memoized since memoized also works for non-properties)

PR #17343, opened 2017-08-03
Rules: Add a match type for when case property does not have a value

gcapalbo:
@emord @snopoke

### DIFF
--- a/corehq/apps/data_interfaces/static/data_interfaces/js/case_rule_criteria.js
+++ b/corehq/apps/data_interfaces/static/data_interfaces/js/case_rule_criteria.js
@@ -6,6 +6,7 @@ hqDefine("data_interfaces/js/case_rule_criteria.js", function() {
         'use strict';
         var self = this;
 
+        self.constants = constants;
         self.case_type = ko.observable();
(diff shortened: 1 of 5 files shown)


corehq/apps/data_interfaces/static/data_interfaces/js/case_rule_criteria.js
```diff
@@ -146,7 +147,8 @@ hqDefine("data_interfaces/js/case_rule_criteria.js", function() {
                 if(
                     value.match_type === constants.MATCH_EQUAL ||
                     value.match_type === constants.MATCH_NOT_EQUAL ||
-                    value.match_type === constants.MATCH_HAS_VALUE
+                    value.match_type === constants.MATCH_HAS_VALUE ||
+                    value.match_type === constants.MATCH_HAS_NO_VALUE
```
emord: Is this a knockout observable? If so I think it should be using `match_type()`

gcapalbo: no, this is just an object that's passed as an initial value

PR #17345, opened 2017-08-03
Add command for resyncing cases for messaging

gcapalbo:
@emord @snopoke

### DIFF
--- a/corehq/messaging/management/commands/resync_case_type.py
+++ b/corehq/messaging/management/commands/resync_case_type.py
@@ -1,0 +1,25 @@
+from corehq.form_processor.interfaces.dbaccessors import CaseAccessors
+from corehq.messaging.tasks import sync_case_for_messaging
+from corehq.util.log import with_progress_bar
+from django.core.management.base import BaseCommand, CommandError
+
+
+class Command(BaseCommand):
+    help = "Sync messaging models for cases"
+
+    def add_arguments(self, parser):
+        parser.add_argument('domain')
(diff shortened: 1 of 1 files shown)


corehq/messaging/management/commands/resync_case_type.py
```diff
@@ -0,0 +1,25 @@
...
+    help = "Sync messaging models for cases"
+
+    def add_arguments(self, parser):
+        parser.add_argument('domain')
+        parser.add_argument('case_type')
+        parser.add_argument('--limit', type=int, default=-1)
+
+    def handle(self, domain, case_type, limit, **options):
+        print("Fetching case ids for %s/%s ..." % (domain, case_type))
+        case_ids = CaseAccessors(domain).get_case_ids_in_domain(case_type)
```
gcapalbo: i've done this query before and it performs decently well, it only takes a few mins for a few million cases

emord: What case is this for? It may be ok, if that is implemented in python, but when I was rebuilding the UCRs for person cases it failed. Should be ok as long as it's not person cases (40 million or so of those I think)

gcapalbo: this is for ccs_record, i've used this approach for child_health in the past (which has more than ccs_record) and it went ok. i haven't had to do it for person yet

emord (approved):

gcapalbo:
(merging, just a management command so won't affect the build. have run it locally)

PR #17351, opened 2017-08-03
Reconcilliation Task Status view

proteusvacuum:
This guy just never finishes. Let's look at the status in realtime! 
Shows the currently elapsed time, and all the errors that have occurred during the currently running task.
@dannyroberts @sravfeyn

### DIFF
--- a/custom/enikshay/tasks.py
+++ b/custom/enikshay/tasks.py
@@ -138,7 +138,7 @@ class EpisodeUpdater(object):
             'mimetype': 'text/csv',
             'file_obj': csv_file,
         }
-        send_html_email_async.delay(
+        send_html_email_async(
             subject, recipient, message, cc=[cc], text_content=message, file_attachments=[attachment]
         )
 
--- a/custom/enikshay/tasks.py
+++ b/custom/enikshay/tasks.py
@@ -625,7 +625,10 @@ class EpisodeTestUpdate(object):
     @property
     @memoized
     def diagnostic_tests(self):
-        return get_private_diagnostic_test_cases_from_episode(self.domain, self.episode.case_id)
+        try:
+            return get_private_diagnostic_test_cases_from_episode(self.domain, self.episode.case_id)
+        except ENikshayCaseNotFound:
+            return None
 
     def update_json(self):
         if self.diagnostic_tests:
--- a/custom/enikshay/tasks.py
+++ b/custom/enikshay/tasks.py
@@ -67,7 +67,7 @@ class Timer:
 
     def __exit__(self, *args):
         self.end = datetime.datetime.now()
-        self.interval = (self.end - self.start).seconds
+        self.interval = (self.end - self.start)
 
 
 class EpisodeUpdater(object):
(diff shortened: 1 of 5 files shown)


custom/enikshay/tasks.py
```diff
@@ -76,31 +91,59 @@ class EpisodeUpdater(object):
     This is applicable to various enikshay domains. The domain can be specified in __init__ method
     """
 
-    def __init__(self, domain):
+    def __init__(self, domain, task_id=None):
         self.domain = domain
+        self.task_id = task_id
         self.updaters = [
             EpisodeAdherenceUpdate,
```
emord: I don't have a lot of context here, but wondering if it would be possible to split this up based on these updaters? Like could you have 3 tasks with one of these?

custom/enikshay/tasks.py
```diff
@@ -147,11 +195,14 @@ def update_single_case(self, episode_case):
         if update_json:
             update_case(self.domain, episode_case.case_id, update_json)
 
-    def _get_open_episode_cases(self):
+    def _get_case_ids(self):
         case_accessor = CaseAccessors(self.domain)
         case_ids = case_accessor.get_open_case_ids_in_domain_by_type(CASE_TYPE_EPISODE)
```
emord: Another way to split this up is that you could have one task per partitioned database. I think we have 10 on NIC/enikshay, but 5 on softlayer? (should check that to be sure though)

custom/enikshay/tasks.py
proteusvacuum: I've been thinking the same, actually. I was concerned that splitting it up would then require us to loop through every single case multiple times, instead of just once, but I'm not sure if that is the bottleneck yet.

custom/enikshay/tasks.py
proteusvacuum: How would that work? We also fetch related cases during the different tasks (which I think aren't guaranteed to be on the same partition), so not sure if we would get any gains from doing that?

emord: I did something like this https://github.com/dimagi/commcare-hq/blob/master/corehq/apps/userreports/management/commands/async_rebuild_table.py#L49-L62 I think you would want to pass in one partition per task and set those tasks up sequentially. 

You can still get the related case from the other partition, but the main benefit is that this would split up the task in multiple different tasks. If this task is failing because of OOM then I think this might solve it since any memory leaks would disappear once the worker is restarted. Also if it's just a couple of cases in one partition causing issues it might help you to narrow it down

sravfeyn (approved):
This task is growing from a simple celery task to a bigger ETL and update operation. Would moving this from celery task to a pillow change processor improve things like OOM, keeping track of status.

proteusvacuum:
@sravfeyn yes! It's getting pretty complicated now.
I think one of the reasons we are doing it this way currently is because we want to update _every_ case that falls into the bucket every night, rather than just updating cases that get changed (like a pillow change processor) would do. I do agree that this has grown out of what the original task set out to do. I can maybe write a few options up for the architecture initiative to take a look at.

PR #17353, opened 2017-08-03
Add unknown and missed adherence scores/counts

esoergel:
@sravfeyn @proteusvacuum @mkangia
https://docs.google.com/document/d/1mQOVfxaKQ1NMhj8DCLSJF6Ufet0Bte4RUEjlX2CKMcY/edit#

Commit by commit should make sense


custom/enikshay/tasks.py
```diff
@@ -313,27 +257,27 @@ def count_doses_taken(dose_taken_by_date, start_date=None, end_date=None):
...
+    def count_doses_taken_by_source(dose_status_by_date, start_date=None, end_date=None):
         """Count all sources of adherence and return the count within the desired timeframe
 
         {'99DOTS': 1, 'MERM': 1, 'treatment_supervisor': 0, ... }
         """
         counts = defaultdict(int)
-        for date, source in doses_taken_by_date.iteritems():
-            if source in VALID_ADHERENCE_SOURCES:
+        for date, status in dose_status_by_date.iteritems():
+            if status.source in VALID_ADHERENCE_SOURCES:
```
mkangia: Do we need to worry about status set to False here in case dose in missed?

custom/enikshay/tasks.py
```diff
@@ -638,3 +582,63 @@ def _get_diagnostic_test_name(self, diagnostic_test):
...
+    Returns: {day: DoseStatus(taken, missed, unknown, source)}
+    """
+
+    adherence_cases_by_date = defaultdict(list)
+    for case in adherence_cases:
+        adherence_date = parse_date(case['adherence_date']) or parse_datetime(case['adherence_date']).date()
+        adherence_cases_by_date[adherence_date].append(case)
+
+    status_by_day = defaultdict(lambda: DoseStatus(taken=False, missed=False, unknown=True, source=False))
+    for d, cases in adherence_cases_by_date.iteritems():
```
mkangia: is `d` day here? Having such names does cause trouble during debugging. There are time where i use ipdb to debug something but can't check the value of a variable because it means something else for the debugger like 'c' for continue. Would be good to keep this actually a name for clarity as well.

custom/enikshay/tasks.py
```diff
@@ -638,3 +582,63 @@ def _get_diagnostic_test_name(self, diagnostic_test):
...
+    1. Find most relevant case
+        if only non-enikshay source cases
+            consider the case with latest_modified - irrespective of case is closed/open
+
+        if only enikshay source cases
+            filter by '(closed and closure_reason == HISTORICAL_CLOSURE_REASON) or open'
+            consider the case with latest modified after above filter
+
+        if mix of enikshay and non-enikshay source cases
+            ignore non-enikshay and apply above enikshay only condition
```
mkangia: Can we reframe this as
```
if any enikshay case
  filter enikshay cases by ...
else
  consider all cases
pick the latest modified
```

custom/enikshay/tasks.py
```diff
@@ -638,3 +582,63 @@ def _get_diagnostic_test_name(self, diagnostic_test):
...
+    2. Check if 'adherence_value' of most relevent case is one of DOSE_TAKEN_INDICATORS
```
mkangia: Where is this being done? Looks like it was present earlier but removed now?

custom/enikshay/tasks.py
```diff
@@ -638,3 +582,63 @@ def _get_diagnostic_test_name(self, diagnostic_test):
...
+                (not case['closed'] or (case['closed'] and
+                    case.get('adherence_closure_reason') == HISTORICAL_CLOSURE_REASON))
+            ),
+            cases
+        )
+    if valid_cases:
+        by_modified_on = sorted(valid_cases, key=lambda case: case['modified_on'])
+        latest_case = by_modified_on[-1]
+        return latest_case
+    return None
```
mkangia: Do we need to return `None`? return would anyway be `None` unless doing it for clarity.

custom/enikshay/tasks.py
```diff
@@ -301,39 +245,44 @@ def get_adherence_schedule_start_date(self):
...
+
+    @staticmethod
+    def count_doses_of_status(status, dose_status_by_date, start_date=None, end_date=None):
+        """status should be 'taken', 'unknown', or 'missed'"""
         if bool(start_date) != bool(end_date):
             raise EnikshayTaskException("Both of start_date and end_date should be specified or niether of them")
 
         if not start_date:
-            return len([dose_taken for dose_taken in dose_taken_by_date.values() if dose_taken])
+            return len([ds for ds in dose_status_by_date.values() if getattr(ds, status)])
```
mkangia: i am confused about what `getattr(ds, status)` should be? isn't `ds` an instance of the `namedtuple`?

mkangia (changes requested):

custom/enikshay/tasks.py
esoergel: It should be `DoseStatus.taken`, `DoseStatus.missed`, or `DoseStatus.unknown`, depending on what parameter is passed in.  It seems I could stand to use some real variable names here, will update.

custom/enikshay/tasks.py
esoergel: This should preserve the original behavior, which is that a source set to `False`  doesn't count towards this metric.

custom/enikshay/tasks.py
esoergel: Good point.  `d` is `day`.  This is moved here from `calculate_doses_taken_by_day`, but I think we can afford the extra two characters for clarity.  Will update.

custom/enikshay/tasks.py
esoergel: This was moved into the function above.  Will update

custom/enikshay/tasks.py
esoergel: updated

custom/enikshay/tasks.py
esoergel: yeah, I just wanted it to be obvious that the latest relevant case could sometimes be `None`

esoergel:
@mkangia updated.  Thanks for the thoughtful feedback!

mkangia (approved):

NoahCarnahan:
Looks like this breaks the historical adherence report. There is a usage of `EpisodeAdherenceUpdate.calculate_doses_taken_by_day()` that needs replacing here: https://github.com/dimagi/commcare-hq/blob/d9ca4ba9173fcb317c0d84fa6500061479fe6a21/custom/enikshay/reports/historical_adherence.py#L133

PR #17354, opened 2017-08-04
Include task-waiting-to-be-run time in async restore timing

millerdev:
As requested in https://github.com/dimagi/commcare-hq/pull/17333#discussion_r131204930

Probably best reviewed by commit 🐠 


NoahCarnahan (approved):
I'm not familiar with this part of the code base, but looks ok to me.

proteusvacuum (approved):

PR #17356, opened 2017-08-04
UCR Export fix

kkrampa:


kkrampa:
@NoahCarnahan 
https://github.com/dimagi/commcare-hq/blob/c617e19d3646ac3ad500d4e3a1e734a032be94d3/corehq/apps/userreports/sql/data_source.py#L146

It seems that columns property doesn't contain calculated columns so expression columns weren't correctly exported. Are expression columns ignored in export on purpose? We have a problem with export in Enikshay since some of the reports use expression columns.

NoahCarnahan:
I believe that was a bug. Thanks for fixing!

PR #17357, opened 2017-08-04
fix ownership cleanliness for extension case of unowned parent

snopoke:
fixes COMMCAREHQ-418

buddy @orangejenny


snopoke:
https://sentry.io/dimagi/commcarehq/issues/322964220/

orangejenny (approved):

PR #17358, opened 2017-08-04
update asserts for 2.38

snopoke:


corehq/ex-submodules/casexml/apps/case/xform.py
```diff
@@ -273,7 +273,7 @@ def _validate_indices(case_db, cases):
                             commcare_version = xform.metadata.commcare_version
                             _soft_assert(
                                 commcare_version < LooseVersion("2.35"),
```
amstone326: this line should change too I assume?

snopoke: ah yes

PR #17359, opened 2017-08-04
ensure that tasks get registered

snopoke:


gcapalbo:
just curious why this was necessary

snopoke:
The task isn't imported anywhere else during normal django startup so
without this it never get's registered with celery. Took me a while to
figure that out.

PR #17360, opened 2017-08-04
session details view

snopoke:
Let me know what you think of this approach for auth

buddy @orangejenny

### DIFF
--- a/corehq/apps/hqadmin/views.py
+++ b/corehq/apps/hqadmin/views.py
@@ -1173,8 +1173,8 @@ class SessionDetialsView(View):
         if not user:
             raise Http404
 
-        couch_user_id = get_user_id_by_username(user.username)
-        if not couch_user_id:
+        couch_user = CouchUser.get_by_username(user.username)
+        if not couch_user:
             raise Http404
 
         try:
@@ -1184,8 +1184,8 @@ class SessionDetialsView(View):
 
         return JsonResponse({
             'username': user.username,
-            'sql_user_id': user.pk,
-            'couch_user_id': couch_user_id,
-            'is_super_user': user.is_superuser,
-            'auth_token': auth_token,
+            'djangoUserId': user.pk,
+            'superUser': user.is_superuser,
+            'authToken': auth_token,
+            'domains': couch_user.domains,
         })
--- a/corehq/apps/hqadmin/views.py
+++ b/corehq/apps/hqadmin/views.py
@@ -1165,7 +1165,7 @@ class SessionDetialsView(View):
         if not data or not isinstance(data, dict):
             return HttpResponseBadRequest()
 
-        session_id = data.get('session_id', None)
+        session_id = data.get('sessionId', None)
         if not session_id:
             return HttpResponseBadRequest()
 
--- a/corehq/apps/hqadmin/views.py
+++ b/corehq/apps/hqadmin/views.py
@@ -1166,16 +1166,20 @@ class SessionDetialsView(View):
             return HttpResponseBadRequest()
 
         session_id = data.get('sessionId', None)
+        domain = data.get('domain', None)
         if not session_id:
             return HttpResponseBadRequest()
 
         user = get_django_user_from_session_key(session_id)
         if not user:
-            raise Http404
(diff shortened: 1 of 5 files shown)


corehq/apps/hqadmin/views.py
```diff
@@ -1134,3 +1137,51 @@ def get(self, request, *args, **kwargs):
...
+        except AttributeError:
+            raise Http404
+
+        user_id = session.get(SESSION_KEY, None)
+        if not user_id:
+            raise Http404
+
+        user_id = get_user_model()._meta.pk.to_python(user_id)
+        return JsonResponse({
+            'user_id': user_id
```
benrudolph: it would be nice on the formplayer side if this returned the couch_id as well. it will eliminate us having to do an additional lookup in the user table

benrudolph (approved):
this is very clever solution. i like it.

fyi @gcapalbo this method may be useful for when we need to do SMS forms authentication from formplayer (instead of using the touchforms API user)

snopoke:
@dannyroberts :white_check_mark:

snopoke:
@dannyroberts  bump

corehq/apps/hqadmin/views.py
```diff
@@ -1134,3 +1137,68 @@ def get(self, request, *args, **kwargs):
             return JsonResponse(data)
         else:
             return HttpResponse('Only web users can access this endpoint', status=400)
+
+
+@method_decorator(csrf_exempt, name='dispatch')
+@method_decorator(validate_request_hmac('FORMPLAYER_INTERNAL_AUTH_KEY', ignore_if_debug=True), name='dispatch')
+class SessionDetialsView(View):
```
dannyroberts: typo (Detials => Details)

dannyroberts (approved):

dannyroberts:
Apologies for the wait!

dannyroberts:
I see there's a corresponding formplayer PR https://github.com/dimagi/formplayer/pull/450. Really like how that simplifies things on that side!

corehq/apps/hqadmin/views.py
```diff
@@ -1134,3 +1137,68 @@ def get(self, request, *args, **kwargs):
             return JsonResponse(data)
         else:
             return HttpResponse('Only web users can access this endpoint', status=400)
+
+
+@method_decorator(csrf_exempt, name='dispatch')
+@method_decorator(validate_request_hmac('FORMPLAYER_INTERNAL_AUTH_KEY', ignore_if_debug=True), name='dispatch')
+class SessionDetailsView(View):
```
dannyroberts: @snopoke just discovered looking at https://rpm.newrelic.com/accounts/366142/applications/33065068/transactions#id=5b225765625472616e73616374696f6e2f537072696e67436f6e74726f6c6c65722f6e617669676174655f6d656e752028504f535429222c22225d and https://rpm.newrelic.com/accounts/366142/applications/5058236/transactions#id=5b225765625472616e73616374696f6e2f46756e6374696f6e2f636f726568712e617070732e687161646d696e2e76696577733a53657373696f6e44657461696c73566965772e706f7374222c22225d that this view is calling `couchdbkit.client:Database.save_doc` twice per call. I stuck a pdb statement in save_doc locally to see when it happens, and I traced it back to auditcare and the presence of `'corehq.apps.hqadmin'` in `AUDIT_MODULES`.

This seems to add a surprising amount to each request time, 80-200ms. If the auditing is helpful, that wouldn't be _that_ high a price to pay, but I'm guessing it was just unintentional and no one's interested in logging every formplayer authentication request. If we can get rid of it, a sizable majority (~90%) of formplayer requests appear to be under a second, making even 100ms off a noticeable improvement.

If you're on board with moving this but not enthused about any place in particular to keep this, I'd suggest the non-admin bucket "app", `hqwebapp`.

snopoke:
Yikes. Thanks for digging that up @dannyroberts. +1 to moving
into`hqwebapp`. I'll do that tomorrow if you haven't already done it

dannyroberts:
Great, thanks for confirming!

PR #17361, opened 2017-08-04
Put new report builder behind a feature flag

NoahCarnahan:
When the new report builder was started, we thought it would be a relatively quick change, so we planned to do all testing and iteration on staging/our local dev environments. It's been clear for a long time that the rollout is going slower than intended. This PR moves the new report builder behind a feature flag, so that it can be merged before it is 100% finished.

Several files have seen significant changes, so rather than attempt to allow both versions of the report builder to use the same files, I've created "v1" versions that will be imported when appropriate. So most of the added lines here are from copying the versions of these files on the master branch into place. Commits that add files do not modify them at all in the same commit. One or two classes/functions do handle requests from both versions of the feature.

My intention is that once we're ready to get rid of V1 of the report builder, we just need to delete the v1 files and the references to them.

Note that this is a PR into the branch for the main new report builder PR: https://github.com/dimagi/commcare-hq/pull/13321/

best reviewed commit-by-commit
buddy @millerdev 
FYI @kaapstorm

### DIFF
--- a/corehq/toggles.py
+++ b/corehq/toggles.py
@@ -458,6 +458,14 @@ REPORT_BUILDER = StaticToggle(
     [NAMESPACE_DOMAIN]
 )
 
+REPORT_BUILDER_V2 = StaticToggle(
+    "report_builder_v2",
+    "Report builder V2",
+    TAG_PRODUCT_PATH,
+    [NAMESPACE_DOMAIN],
+    description='Enables the "new" report builder. Note that the "Report Builde" feature flag must also be enabled'
+)
+
 ASYNC_RESTORE = StaticToggle(
     'async_restore',
     'Generate restore response in an asynchronous task to prevent timeouts',
--- a/corehq/toggles.py
+++ b/corehq/toggles.py
@@ -463,7 +463,7 @@ REPORT_BUILDER_V2 = StaticToggle(
     "Report builder V2",
     TAG_PRODUCT_PATH,
     [NAMESPACE_DOMAIN],
-    description='Enables the "new" report builder. Note that the "Report Builde" feature flag must also be enabled'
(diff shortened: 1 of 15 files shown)


NoahCarnahan:
Stickler comments are on the v1 versions of the files (direct copies from master), so not going to address.

corehq/toggles.py
```diff
@@ -458,6 +458,14 @@ def toggle_values_by_name(username=None, domain=None):
     [NAMESPACE_DOMAIN]
 )
 
+REPORT_BUILDER_V2 = StaticToggle(
+    "report_builder_v2",
+    "Report builder V2",
+    TAG_PRODUCT_PATH,
+    [NAMESPACE_DOMAIN],
+    description='Enables the "new" report builder. Note that the "Report Builde" feature flag must also be enabled'
```
millerdev: typo: "Report Builde" -> "Report Builder"

millerdev (approved):

orangejenny:
Noah, do you have thoughts on testing? For app builder, the tests that compare the v1 and v2 versions, verifying that if you updated one version, you also updated the other, have saved us quite a few headaches.

orangejenny:
(I haven't read the PR, but don't see any files with "test" in the path)

NoahCarnahan:
This is a little different than the app builder PR in that I don't intend any changes to be made to v1 while these two versions exist. The two versions are a bit different in functionality.

orangejenny:
Cool, sounds good.

millerdev:
> I don't intend any changes to be made to v1 while these two versions exist.

not even bug fixes?

NoahCarnahan:
Fair enough. New PR here https://github.com/dimagi/commcare-hq/pull/17380

PR #17362, opened 2017-08-04
Added default user case properties.

emord:
@millerdev http://manage.dimagi.com/default.asp?259366# should these be in by default? The issue was that the user didn't save these in the app. (we have hte same issue with users that upload via case importer and only load, never save)

### DIFF
--- a/corehq/apps/app_manager/app_schemas/case_properties.py
+++ b/corehq/apps/app_manager/app_schemas/case_properties.py
@@ -186,7 +186,8 @@ def get_all_case_properties(app):
 
 def get_usercase_properties(app):
     if is_usercase_in_use(app.domain):
-        default_properties = {'first_name', 'last_name'}
+        # TODO: add name here once it is fixed to concatenate first and last in form builder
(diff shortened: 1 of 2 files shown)


corehq/apps/app_manager/app_schemas/case_properties.py
```diff
@@ -186,7 +186,10 @@ def get_all_case_properties(app):
 
 def get_usercase_properties(app):
     if is_usercase_in_use(app.domain):
-        return get_case_properties(app, [USERCASE_TYPE])
+        default_properties = {'first_name', 'last_name'}
```
millerdev: I think this should also contain `phone_number` and `username`, and eventually when we get it fixed, `name`. See [Accessing Phone Number and the User Name](https://confluence.dimagi.com/display/commcarepublic/Custom+User+Data#CustomUserData-AccessingPhoneNumberandtheUserName) and [User Case - Other Notes](https://confluence.dimagi.com/display/commcarepublic/User+Case#UserCase-OtherNotes). Maybe there are others as well? I couldn't find them.

millerdev (approved):

PR #17366, opened 2017-08-04
Nc/update abt report config

NoahCarnahan:
@millerdev 
Follow up on https://github.com/dimagi/commcare-hq/pull/17341.
I changed the wrong part of the file :(


NoahCarnahan:
Killed the build because there is no test coverage

PR #17368, opened 2017-08-04
Added task to recalculate UCRs up to 10k records.

emord:
This won't help rebuilding UCRs that are very large, but it will help get https://github.com/dimagi/commcare-hq/pull/17275 out the door (it has < 5k forms)

@sheelio @biyeun


emord:
@NoahCarnahan

NoahCarnahan (approved):
Looks good.

This is my understanding of the PR, let me know if this isn't true:
Users can click the new "Recalculate" button on a data source to reprocess the documents already in the table (only works if the table has less than 10k rows).
This has the advantage of not requiring the rebuild process to iterate over every single case in the domain.

My main concern is that that button isn't documented anywhere. Perhaps you can add it to the UCR readme? It would also be cool if it raised an exception if the user tries it on a table with more than 10k documents, but I realize this is a quick fix sort of thing, so not insisting on it.

emord:
Right now this is explicitly for a data source that has <5k rows in it for ICDS. I plan on removing it once a better solution is in place

PR #17370, opened 2017-08-04
update names

nickpell:
https://docs.google.com/document/d/1VoIffMc9lbDBJ8tACd1nCiWneXxo_14BMNTnZvYyHOg/edit#heading=h.2by0vnwy62zt

@proteusvacuum

### DIFF
--- a/custom/enikshay/private_sector_datamigration/management/commands/update_names.py
+++ b/custom/enikshay/private_sector_datamigration/management/commands/update_names.py
@@ -2,7 +2,7 @@ from django.core.management import BaseCommand
 
 from corehq.apps.locations.models import SQLLocation
 from corehq.apps.users.models import CommCareUser
-from custom.enikshay.private_sector_datamigration.models import UserDetail
+from custom.enikshay.private_sector_datamigration.models import UserDetail, Agency
 
 
 class Command(BaseCommand):
@@ -11,7 +11,7 @@ class Command(BaseCommand):
         parser.add_argument('domain')
 
     def handle(self, domain, **options):
-        for loc in SQLLocation.active_objects.filter(
+        locs_matching_type = SQLLocation.active_objects.filter(
             domain=domain,
             location_type__code__in=[
                 'pac',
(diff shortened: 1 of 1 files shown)


custom/enikshay/private_sector_datamigration/management/commands/update_names.py
```diff
@@ -0,0 +1,55 @@
...
+        parser.add_argument('domain')
+
+    def handle(self, domain, **options):
+        locs_matching_type = SQLLocation.active_objects.filter(
+            domain=domain,
+            location_type__code__in=[
+                'pac',
+                'pcp',
+                'plc',
+                'ps-fieldstaff',
```
proteusvacuum: I think this should be `fo`. the `ps-fieldstaff` is the usertype assigned to that location.

proteusvacuum: we also need chemists `pcc`

custom/enikshay/private_sector_datamigration/management/commands/update_names.py
```diff
@@ -0,0 +1,55 @@
...
+        for i, loc in enumerate(locs_matching_type):
+            private_sector_agency_id = loc.metadata.get('private_sector_agency_id', '')
+            if not private_sector_agency_id.isdigit():
+                continue
+            agency_id = int(private_sector_agency_id)
+            try:
+                agency = Agency.objects.get(agencyId=agency_id)
+            except Agency.DoesNotExist:
+                continue
+            loc.name = "%s - %d" % (agency.agencyName, agency_id)
```
proteusvacuum: 👍 
I think you could also have just set this as `"{} - {}".format(loc.name, agency_id)`, but your way is good because it will also filter out all non-migrated agencies.

proteusvacuum (changes requested):
I think you are inadvertently updating the deploy submodule here.
Just 2 little changes to the location types, otherwise looks good.

proteusvacuum (approved):

PR #17377, opened 2017-08-07
pulling referral cases from occurrence cases

calellowitz:
@NoahCarnahan


NoahCarnahan (commented):
Looks good. One thing I might change would be to check for referrals on the occurrence even if you find one on the person. I think it would be a bug with the app if it allowed you to get in that state, but given the complexity of the app, I think it's worth making sure that isn't happening. For example, I could imagine things getting into a state where there is like an "orphaned" opened referral related to the person case that the app is ignoring because it's now looking at the occurrence.

NoahCarnahan (approved):
Looks great. Just going to wait for tests until merging.

PR #17380, opened 2017-08-07
Add diff tests for report builder

NoahCarnahan:
@millerdev  
FYI @orangejenny @kaapstorm


corehq/apps/userreports/management/commands/build_report_builder_v2_diffs.py
```diff
@@ -0,0 +1,79 @@
+from __future__ import print_function
```
NoahCarnahan: This is one of the files that's actually new code

corehq/apps/userreports/tests/test_report_builder_v2_diffs.py
```diff
@@ -0,0 +1,110 @@
+# coding: utf-8
```
NoahCarnahan: This is the other file that is actually new code

NoahCarnahan (commented):
FYI you might be wondering why I didn't inherit from the app_manager test case and management commend. The way I structured my directories is a bit different, which resulted in enough changes to these files that I didn't think it made sense to inherit from them.

orangejenny:
> FYI you might be wondering why I didn't inherit from the app_manager test case and management commend. The way I structured my directories is a bit different, which resulted in enough changes to these files that I didn't think it made sense to inherit from them.

Also convenient because hopefully we'll be deleting the app manager infrastructure soon.

corehq/apps/userreports/management/commands/build_report_builder_v2_diffs.py
```diff
@@ -0,0 +1,77 @@
...
+    "templates/userreports/v1/",
+    "templates/userreports/partials/v1/",
+    "templates/userreports/reportbuilder/v1/",
+]
+
+RELATIVE_DIFF_STORAGE = 'tests/data/report_builder_v2_diffs/'
+
+
+def get_diff_filename(v2_dir, filename):
+    dir_part = ".".join([x for x in v2_dir.split("/") if x and x != ".."])
```
millerdev: nit: could use generator expression instead of list comprehension here (no need to create intermediate list)

corehq/apps/userreports/management/commands/build_report_builder_v2_diffs.py
```diff
@@ -0,0 +1,77 @@
...
+    if dir_part:
+        filename = "{}.{}".format(dir_part, filename)
+    return "{}.diff.txt".format(filename)
+
+
+def get_diff(file_v1, file_v2):
+    with open(file_v1, "r") as fv1:
+        with open(file_v2, "r") as fv2:
```
millerdev: nit: could put both `open` calls in a single `with` statement.

millerdev (approved):

millerdev (approved):

PR #17381, opened 2017-08-07
Nikshay Integration: Use age_entered if age not available

sheelio:
@proteusvacuum 
Temporary until we update the app to use age


proteusvacuum (changes requested):
You did this for the `PrivatePatientPayloadGenerator`, not the public one, so this won't have the effect you are going for...

proteusvacuum (approved):

PR #17387, opened 2017-08-08
migrate attachments

snopoke:
Migration to retroactively copy attachments to edited forms + a fix to the copy over `content_length` and the correct bucket.

### DIFF
--- a/corehq/apps/cleanup/management/commands/copy_form_attachments.py
+++ b/corehq/apps/cleanup/management/commands/copy_form_attachments.py
@@ -1,0 +1,62 @@
+import uuid
+
+from django.core.management.base import BaseCommand
+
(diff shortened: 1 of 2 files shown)


gcapalbo (approved):
would it be better if this wrote to a log file rather than use print? or were you planning on capturing standard out

sheelio:
@gcapalbo

I think Simon is offline - is this good to merge? The customer pinged me and asked me about.

gcapalbo:
yeah my comment is not blocking merge but i'm guessing @snopoke would likely be the one to run this since he has the most context on it and probably would like to monitor its run. i'll merge so it gets into today's deploys

snopoke:
Meant to comment that I'd already run it on all prod, softlayer and ICDS

PR #17388, opened 2017-08-08
Added missing column

kkrampa:
@NoahCarnahan


kkrampa:
@NoahCarnahan We need test drtb data source rebuild. Is it possible to rebuild this in place when nullable column is added? (I saw PR with that feature)

NoahCarnahan:
Yeah I think we can. I'll try that now.

PR #17389, opened 2017-08-08
submission reprocessing queue tweaks

snopoke:


corehq/ex-submodules/couchforms/migrations/0003_auto_20170808_1212.py
```diff
@@ -0,0 +1,22 @@
...
+    dependencies = [
+        ('couchforms', '0002_auto_20170802_1505'),
+    ]
+
+    operations = [
+        AlterFieldCreateIndexIfNotExists(
+            model_name='unfinishedsubmissionstub',
+            name='timestamp',
+            field=models.DateTimeField(db_index=True),
+        ),
```
snopoke: I've already created this index on all envs

snopoke:
https://sentry.io/dimagi/commcarehq/issues/324655500/

corehq/util/django_migrations.py
```diff
@@ -61,3 +61,14 @@ def database_forwards(self, app_label, schema_editor, from_state, to_state):
...
+
+class AlterFieldCreateIndexIfNotExists(migrations.AlterField):
+
+    def database_forwards(self, app_label, schema_editor, from_state, to_state):
+        schema_editor.__class__ = DatabaseSchemaEditorIfNotExists
+        try:
+            super(AlterFieldCreateIndexIfNotExists, self).database_forwards(
+                app_label, schema_editor, from_state, to_state)
+        finally:
+            schema_editor.__class__ = DatabaseSchemaEditor
```
gcapalbo: won't this fail silently if it fails to create the index?

corehq/form_processor/tasks.py
```diff
@@ -9,6 +9,10 @@
 
 @no_result_task(queue=SUBMISSION_REPROCESS_CELERY_QUEUE, acks_late=True)
 def reprocess_submission(submssion_stub_id):
-    stub = UnfinishedSubmissionStub.objects.get(id=submssion_stub_id)
```
gcapalbo: probably worth wrapping this task in a `CriticalSection` to prevent race conditions

corehq/util/django_migrations.py
snopoke: no it shouldn't. any exception that get's raised will still bubble up

corehq/form_processor/tasks.py
snopoke: Will do though I think the reason it wouldn't be found is that it was deleted as part of form processing. I changed the order in which the stubs are queued to be according to date created which should also prevent this.

snopoke: 8299d7f

corehq/util/django_migrations.py
gcapalbo: right, sorry i misread it

gcapalbo (approved):

PR #17394, opened 2017-08-08
ICDS: Fix infrastructure form aggregation

sheelio:
@calellowitz 

This fixes a small bug in the infrastructure form aggregation.  It also re-runs the create_functions everytime the task is run (just avoiding adding a migration each time it changes).


sheelio:
@emord 
(Since you're back online)

PR #17398, opened 2017-08-08
ICDS: Fix agg child view

sheelio:
@calellowitz 

This fixes a bug in the child view (was grabbing data from the wrong column)


sheelio:
@emord 
(Since you're back online)

PR #17399, opened 2017-08-09
Retry partitioned_get on InterfaceError

gcapalbo:
http://manage.dimagi.com/default.asp?259922

@snopoke this was the reason for the pillow retry queue getting so behind, the pillows were generating lots of the errors in the ticket. seems like this started happening with the new partitioned_get when the db has issues, since the persistent pillow processes don't ever close the connections.

hold off on merge for now, going to see if there's a cleaner way to do this tomorrow. for now the remedy was to just restart the pillows, and i temporarily boosted the concurrency on the pillow retry queue so it processes a little faster.

### DIFF
--- a/corehq/sql_db/models.py
+++ b/corehq/sql_db/models.py
@@ -73,11 +73,16 @@ class RequireDBManager(models.Manager):
                 self.model.partition_attr: partition_value
             }
 
+        db_alias = self.get_db(partition_value)
+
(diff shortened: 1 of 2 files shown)


snopoke (approved):

gcapalbo:
closing in favor of https://github.com/dimagi/commcare-hq/pull/17414

PR #17401, opened 2017-08-09
Restored test_2b data source

kkrampa:
It seems this one was accidentally removed. I guess it needs to be rebuilt.


kkrampa:
@NoahCarnahan

kkrampa:
If there are still data in that datasource, can we try to rebuild it in-place?

kkrampa:
After last commit, we need rebuild following data sources: episode_2b, episode_drtb, test_2b, test_drtb.
I think all of them can be rebuild in place.

NoahCarnahan:
Thanks Kamil.

NoahCarnahan:
Data sources are now rebuilt.

PR #17402, opened 2017-08-09
clone command fix

sravfeyn:
@calellowitz

### DIFF
--- a/corehq/apps/hqadmin/management/commands/clone_domain.py
+++ b/corehq/apps/hqadmin/management/commands/clone_domain.py
@@ -70,11 +70,11 @@ class Command(BaseCommand):
 
         copy_locations = self._clone_type(options, 'locations')
         copy_location_types = self._clone_type(options, 'location_types')
-        if copy_locations and not copy_location_types:
-            raise CommandError("You can't copy locations by excluding types")
-        else:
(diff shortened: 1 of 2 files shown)


corehq/apps/hqadmin/management/commands/clone_domain.py
```diff
@@ -68,8 +68,12 @@ def handle(self, existing_domain, new_domain, **options):
...
 
-        if self._clone_type(options, 'locations'):
-            types_only = self._clone_type(options, 'types_only')
+        copy_locations = self._clone_type(options, 'locations')
+        copy_location_types = self._clone_type(options, 'location_types')
+        if copy_locations and not copy_location_types:
+            raise CommandError("You can't copy locations by excluding types")
+        else:
+            types_only = copy_location_types and not copy_locations
             self.copy_locations(types_only)
```
gcapalbo: it looks like this now calls `self.copy_locations` even when both `copy_locations` and `copy_location_types` are both false. should this check if `copy_locations` is true before copying?

sravfeyn: Thanks! Updated

gcapalbo (approved):

PR #17404, opened 2017-08-09
Check if data migration is in progress every minute to halt auto case update task

gcapalbo:
@proteusvacuum


gcapalbo:
:white_check_mark:

PR #17405, opened 2017-08-09
Ce/phone data

calellowitz:
@dannyroberts @emord 
management command for an icds data dump

### DIFF
--- a/custom/icds/management/commands/phone_number_data.py
+++ b/custom/icds/management/commands/phone_number_data.py
@@ -1,0 +1,36 @@
+from __future__ import print_function
+
+import csv
+
+from django.core.management import BaseCommand
+
+from corehq.apps.locations.models import SQLLocation
+from corehq.form_processor.interfaces.dbaccessors import CaseAccessors
+
+
+class Command(BaseCommand):
+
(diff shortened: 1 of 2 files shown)


custom/icds/management/commands/phone_number_data.py
```diff
@@ -0,0 +1,40 @@
...
+                else:
+                    person = self.case_accessor.get_case(case_id)
+                    hh = None
+                row[15] = SQLLocation.objects.get(location_id=owner).name
+                if hh:
+                    row[16] = hh.get_case_property('name')
+                    row[17] = hh.get_case_property('hh_num')
+                row[18] = person.get_case_property('name')
+                row[19] = person.get_case_property('dob')
+                row[20] = person.get_case_property('sex')
```
emord: Does there need to be an assertion that these actually have 20 columns?

emord (approved):

custom/icds/management/commands/phone_number_data.py
calellowitz: This was for 4 data files that i have and verified have 20 columns. In theory yes, but given that we control the input, it seems unnecessary.

PR #17408, opened 2017-08-09
V2 Privilege Generation Code

ctsims:
Update to the mobile privilege code, allows splitting out privileges.

For now just replicates the previous capabilities (turning on all features)
Spec:
https://docs.google.com/document/d/1VKkAjD6-HRoHM7eql3rwy9_vMnsPaobfeNmrz7zx4lw/edit

Mobile PR and Tests are here:
https://github.com/dimagi/commcare-android/pull/1784

Tested to work on both an old phone (with the old processing code) and the new PR processor

### DIFF
--- a/corehq/apps/settings/views.py
+++ b/corehq/apps/settings/views.py
@@ -440,16 +440,27 @@ class EnableMobilePrivilegesView(BaseMyAccountView):
 
     @method_decorator(login_required)
     def get(self, request, *args, **kwargs):
-        message = json.dumps([
+        message_v1 = json.dumps([
             {'username': request.user.username},
             {'flag': MULTIPLE_APPS_UNLIMITED.slug}
         ]).replace(' ', '')
+
+        message_v2 = json.dumps([
+            {'username': request.user.username},
+            {'flags': [MULTIPLE_APPS_UNLIMITED.slug,ADVANCED_SETTINGS_ACCESS.slug]}
+        ]).replace(' ', '')
+
         qrcode_data = json.dumps({
             'username': request.user.username,
+            'version': 2,
             'flag': MULTIPLE_APPS_UNLIMITED.slug,
-            'signature': b64encode(sign(message))
+            'flags': [MULTIPLE_APPS_UNLIMITED.slug,ADVANCED_SETTINGS_ACCESS.slug],
+            'signature': b64encode(sign(message_v1)),
+            'multiple_flags_signature': b64encode(sign(message_v2))
         })
+        
         qrcode = get_qrcode(qrcode_data)
+        
         context = self.get_context_data(**kwargs)
         context['qrcode_64'] = b64encode(qrcode)
         return self.render_to_response(context)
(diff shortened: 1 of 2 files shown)


ctsims:
@amstone326 should be good to go. We'll still need to resource getting the two split out if needed (I can generate manual barcodes for enterprise projects)

corehq/apps/settings/views.py
```diff
@@ -440,16 +440,27 @@ class EnableMobilePrivilegesView(BaseMyAccountView):
...
     def get(self, request, *args, **kwargs):
-        message = json.dumps([
+        message_v1 = json.dumps([
             {'username': request.user.username},
             {'flag': MULTIPLE_APPS_UNLIMITED.slug}
         ]).replace(' ', '')
+
+        message_v2 = json.dumps([
+            {'username': request.user.username},
+            {'flags': [MULTIPLE_APPS_UNLIMITED.slug, ADVANCED_SETTINGS_ACCESS.slug]}
```
ctsims: fixed

corehq/apps/settings/views.py
```diff
@@ -440,16 +440,27 @@ class EnableMobilePrivilegesView(BaseMyAccountView):
...
+        ]).replace(' ', '')
+
         qrcode_data = json.dumps({
             'username': request.user.username,
+            'version': 2,
             'flag': MULTIPLE_APPS_UNLIMITED.slug,
-            'signature': b64encode(sign(message))
+            'flags': [MULTIPLE_APPS_UNLIMITED.slug, ADVANCED_SETTINGS_ACCESS.slug],
```
ctsims: fixed

amstone326:
@ctsims 2 things:

1. What is the point of this PR if all it does is just blindly add both privileges? I thought the point was to separate them now (i.e. make access to them controlled by separate feature flags), so that we can give ICDS users access to MULTIPLE_APPS_UNLIMITED, but not to ADVANCED_SETTINGS_ACCESS

2. We should now add `advanced_settings_access` to the list of known privilege keys in the mobile code.

ctsims:
@amstone326 forgot to tag you in previous comment

amstone326:
@ctsims Gotcha ok, my only concern is that this means the work outlined in that spec is being left only half-done. Would mind just making it clear in the spec which pieces have and haven't been implemented by this PR already, so that whoever picks up the rest of it isn't confused. (Is there a clear plan for someone else doing the rest of this, or is that not prioritized?)

ctsims:
@amstone326 I don't think there's any specific resourcing left after this point. The only group that needs to enable the split barcode (multiple app only) is eNikshay and I already have a barcode for them to do so. 

We'll need solutions to jump in for the need to split out the feature flags for non-enterprise before it makes sense to pick up the more robust feature.

amstone326:
@ctsims OK, I still want it to be clear from the specs what has and hasn't been done though. I decided to just delete from the existing spec the outline for the work that you didn't do, and created a new spec for that here: https://docs.google.com/document/d/1b1csBrtjwAeD4BB6dEGQc32TWM6F4nrgLS8TiTKJt4A/edit. I'll hold onto this so that it's available when we want to do it

biyeun (approved):

PR #17413, opened 2017-08-09
Finish up BETS data dump commands

esoergel:
https://docs.google.com/spreadsheets/d/1yOrEgKyWKpwyIf4NCzY978LD6xzZaIyqu2FhrBKv22g/edit#gid=1696100868
@proteusvacuum @sravfeyn


custom/enikshay/management/commands/get_vouchers_for_bets.py
```diff
@@ -60,10 +60,12 @@ def handle(self, domain, repeater_id, filename, **options):
                     payload = json.loads(record.get_payload())['voucher_details'][0]
 
                     voucher_id = record.payload_id
-                    person_case = get_person_case_from_voucher(domain, voucher_id)
                     voucher_case = accessor.get_case(voucher_id)
-                    payload[u'PersonId'] = person_case.case_id
-                    payload[u'AgencyId'] = voucher_case.get_case_property('voucher_fulfilled_by_id')
+                    # Could also be person_case.get_case_property('person_id')
+                    payload[u'PersonID'] = voucher_case.get_case_property('person_enikshay_id')
```
proteusvacuum: I don't think voucher cases have this property. Where did you find that? 
I think better to use `person_case.get_case_property('person_id')`

custom/enikshay/integrations/bets/repeater_generators.py
```diff
@@ -220,6 +224,15 @@ def create_voucher_payload(cls, voucher_case):
             related_case_id=voucher_case.case_id
         )
 
+        approver_id = voucher_case.get_case_property('voucher_approved_by_id')
+        if approver_id:
+            approver = CommCareUser.get_by_user_id(approver_id)
+            approver_name = approver.name
+            approver_usertype = approver.user_data.get('usertype')
+        else:
```
proteusvacuum: in theory we should only be sending them approved vouchers. If this property isn't set, then something has gone wrong somewhere...

custom/enikshay/management/commands/get_vouchers_for_bets.py
```diff
@@ -67,12 +66,6 @@ def handle(self, domain, repeater_id, filename, **options):
                         voucher_case.get_case_property('voucher_fulfilled_by_id'))
                     payload[u'AgencyId'] = agency_user.raw_username
 
-                    approver_id = voucher_case.get_case_property('voucher_approved_by_id')
-                    approver = CommCareUser.get_by_user_id(approver_id)
-                    payload[u'EnikshayApprover'] = approver.name
-                    payload[u'EnikshayRole'] = approver.user_data.get('usertype')
-                    payload[u'EnikshayApprovalDate'] = voucher_case.get_case_property('date_approved')
-
                     payload['Succeeded'] = record.succeeded
```
proteusvacuum: Looks like they don't want this any more, and instead want `PaymentStatus` (you might have fixed in a later commit)

custom/enikshay/integrations/bets/repeater_generators.py
```diff
@@ -549,4 +548,6 @@ def serialize(person_case):
             for prop in BETSBeneficiaryPayloadGenerator.case_properties
         }
         case_json["properties"]["owner_id"] = person_case.owner_id
+        # This is the "real" phone number
+        case_json["properties"]["phone_number"] = case_properties.get("contact_phone_number", "")
```
proteusvacuum: this is going to be of the form `91123456789`. Just making sure they are aware that it will have the `91` part at the beginning.

proteusvacuum (changes requested):

custom/enikshay/management/commands/get_vouchers_for_bets.py
esoergel: :+1:

custom/enikshay/management/commands/get_vouchers_for_bets.py
esoergel: I think I'll come back to this and update with the rest of the changes - the voucher dump also needs to include the payment confirmation information, but that won't be part of this next set of dumps.

custom/enikshay/integrations/bets/repeater_generators.py
esoergel: I added an example in that format to the spec doc, although we didn't discuss specifically.

custom/enikshay/integrations/bets/repeater_generators.py
esoergel: Good point. Making this fail hard.

custom/enikshay/management/commands/get_users_for_bets.py
```diff
@@ -118,6 +118,7 @@ def _is_relevant_location(location):
             return
 
         user_data = BETSUserPayloadGenerator.serialize(self.domain, user)
+        user_data['migrationCrtdRecordId'] = user.user_data.get('agency_id_legacy', '')
```
proteusvacuum: are they aware that this won't be set on agencies that were created in eNikshay? 
I think they might be looking for something else here?

proteusvacuum (approved):
One question, looks good though

custom/enikshay/integrations/bets/repeater_generators.py
```diff
@@ -195,15 +198,31 @@ def create_diagnosis_and_notification_payload(cls, episode_case):
...
-        location = cls._get_location(
-            episode_case_properties.get("registered_by"),
-            field_name="registered_by",
-            related_case_type="episode",
-            related_case_id=episode_case.case_id,
-        )
-        if not location.user_id:
-            raise NikshayLocationNotFound(
-                "Location {} does not have a virtual location user".format(location.location_id))
+        registered_by_location = episode_case_properties.get("registered_by")
```
proteusvacuum: If it was only test cases that were failing this test, then I'm happy to leave this as it was, as users are only on the newest version of the app, which sets this property correctly.

esoergel: :+1:

PR #17415, opened 2017-08-10
cache path displays to avoid SQL queries

snopoke:
https://manage.dimagi.com/default.asp?260004

### DIFF
--- a/corehq/apps/userreports/reports/filters/choice_providers.py
+++ b/corehq/apps/userreports/reports/filters/choice_providers.py
@@ -285,8 +285,18 @@ class LocationChoiceProvider(ChainableChoiceProvider):
         return [Choice(SHOW_ALL_CHOICE, "[{}]".format(ugettext('Show All')))]
 
     def _locations_to_choices(self, locations):
+        cached_path_display = {}  # works best if self.order_by_hierarchy == True
+
         def display(loc):
-            return loc.get_path_display() if self.show_full_path else loc.display_name
+            if self.show_full_path:
+                if loc.parent_id in cached_path_display:
+                    path_display = '{}/{}'.format(cached_path_display[loc.parent_id], loc.name)
+                else:
+                    path_display = loc.get_path_display()
+                cached_path_display[loc.id] = path_display
+                return path_display
+            else:
+                return loc.display_name
         return [Choice(loc.location_id, display(loc)) for loc in locations]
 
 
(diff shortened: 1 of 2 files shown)


corehq/apps/userreports/reports/filters/choice_providers.py
```diff
@@ -285,8 +285,18 @@ def default_value(self, user):
...
     def _locations_to_choices(self, locations):
+        cached_path_display = {}  # works best if self.order_by_hierarchy == True
+
         def display(loc):
-            return loc.get_path_display() if self.show_full_path else loc.display_name
+            if self.show_full_path:
+                if loc.parent_id in cached_path_display:
+                    path_display = '{}/{}'.format(cached_path_display[loc.parent_id], loc.name)
+                else:
+                    path_display = loc.get_path_display()
```
mkangia: Might be an overkill, but if this runs for all locations under a certain parent, it would still run `get_path_display` for all. 
Would it be better to actually add the parent full path if its missing by modifying something like
```
if loc.parent_id in cached_path_display:
   parent_full_path = cached_path_display[loc.parent_id]
else:
   cached_path_display[loc.parent_id] = parent_full_path = loc.parent.get_path_display()
path_display = '{}/{}'.format(parent_full_path, loc.name)
cached_path_display[loc.id] = path_display
```

esoergel: Nice catch!  If the filter uses "tree order", then parent should always come first, but unfortunately that's not the default, so this seems like a good idea.

snopoke: I'm not convinced that's an improvement since `loc.parent` is an extra SQL query and if `self.include_descendants = False` then you're doing an extra query for every location.

I think if we want to optimize the other paths then we should consider them separately.

snopoke: @esoergel @mkangia let me know if you disagree

esoergel: No, that seems reasonable.  I imagine each combination of these filter options would have its own optimization.

mkangia: i assumed that `loc.get_path_display` would anyway find the parent of the location. and that looks like its that way from the code here. But if you have considered all possibilities and are convinced its not needed, then its fine.

snopoke:
It will include the parent but it won't use a separate query to do it

PR #17420, opened 2017-08-10
updating date property

calellowitz:
@NoahCarnahan 
getting final clarification on one more possible change but wanted to set up the pr now


NoahCarnahan (approved):
Nice. Glad there weren't more properties to change.

PR #17421, opened 2017-08-10
Reg from case list: case not created: don't include dependent datums

orangejenny:
https://manage.dimagi.com/default.asp?259938

@proteusvacuum / @millerdev


proteusvacuum (approved):

orangejenny:
Merge before deploy please @biyeun as this is blocking an enikshay app release

biyeun:
@orangejenny done!

orangejenny:
@biyeun thank you :)

PR #17424, opened 2017-08-10
Add alternate numbers to 99DOTS

proteusvacuum:
These 3 properties have been added to the app that need to be sent to 99DOTS.


proteusvacuum:
@gcapalbo

PR #17425, opened 2017-08-10
add 'server_modified_on' date to form index

snopoke:
This is a temporary measure until a better and more permanent solution can be implemented. 

Note that I'm not changing the index name on purpose to avoid a full reindex. 

https://manage.dimagi.com/default.asp?259782


corehq/pillows/mappings/xform_mapping.py
```diff
@@ -43,6 +43,10 @@
             "type": "date",
             "format": DATE_FORMATS_STRING
         },
+        "server_modified_on": {
```
emord: Do we actually put the mapping every time? For some reason I thought we only did that on index creation

snopoke: We don't but there is a management command we can run to update them: https://github.com/dimagi/commcare-hq/blob/9a7bf86aa0f81c6a52ccc24022a125f912702d9c/corehq/ex-submodules/pillowtop/management/commands/update_es_mapping.py#L11-L11

snopoke: yes, that's the idea. We don't have a way to reindex only one domain's data though.

snopoke: or maybe we do: https://github.com/dimagi/commcare-hq/blob/03251f4ae769f1c584586c800b7e928c059cbd3d/corehq/apps/hqcase/management/commands/reindex_sql_forms_in_domain.py#L22-L22

dannyroberts (approved):

sheelio:
@dannyroberts 
This looks good - can we merge it and run the migration (for the evaluate-2.0 project first?)

dannyroberts:
@sheelio I'm going to take discussion on next steps to the ticket; it looks like @snopoke has a plan for this, but I don't think I'm up to speed enough to do it myself

PR #17426, opened 2017-08-10
2B Migration: Fix treatment status

sheelio:
@NoahCarnahan @esoergel 

This is migration to fix some of the treatment status issues with the migration.  I'd like to get it run before EOD tomorrow if possible.

### DIFF
--- a/custom/enikshay/two_b_release_1/management/commands/enikshay_2b_treatment_status_fix.py
+++ b/custom/enikshay/two_b_release_1/management/commands/enikshay_2b_treatment_status_fix.py
@@ -1,0 +1,77 @@
+import datetime
+import logging
+
+from corehq.apps.es import CaseSearchES
+from django.core.management import BaseCommand
+from casexml.apps.case.mock import CaseStructure, CaseFactory
+
+logger = logging.getLogger('enikshay_2b_reason_for_test_fix')
+
+
+class Command(BaseCommand):
+
+    def add_arguments(self, parser):
+        parser.add_argument(
+            'domain',
+            help="The domain to migrate."
+        )
+        parser.add_argument(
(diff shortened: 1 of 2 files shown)


custom/enikshay/two_b_release_1/management/commands/enikshay_2b_treatment_status_fix.py
```diff
@@ -0,0 +1,79 @@
...
+
+        logger.info("Starting {} migration on {} at {}".format(
+            "real" if commit else "fake", domain, datetime.datetime.utcnow()
+        ))
+
+        cases = (CaseSearchES()
+                 .domain(domain)
+                 .case_type("episode")
+                 .case_property_query("updated_by_migration", "enikshay_2b_case_properties", "must")
+                 .run().hits)
```
NoahCarnahan: Have you confirmed that this is a relatively small number of cases? We might need to paginate the query if not. One way to confirm the size is to do `query.count()` in a shell (instead of `query.run().hits`).

NoahCarnahan (approved):
Looks good

custom/enikshay/two_b_release_1/management/commands/enikshay_2b_treatment_status_fix.py
sheelio: I think this is going to hit approximately 10000 cases (fewer than the test migration you previously ran)

NoahCarnahan: That might put some strain on the server. I've paginated it here with a scroll query: https://github.com/dimagi/commcare-hq/pull/17439

PR #17427, opened 2017-08-10
Hide all info in email from report

proteusvacuum:
Email body was including all the data. Which in some cases was making the email to large to send.
We just want the excel export anyway.
@mkangia 
cc @gcapalbo

### DIFF
--- a/custom/enikshay/reports/repeaters.py
+++ b/custom/enikshay/reports/repeaters.py
@@ -46,6 +46,8 @@ class ENikshayForwarderReport(DomainForwardingRepeatRecords):
     def get_all_rows(self):
         repeater_id = self.request.GET.get('repeater', None)
         state = self.request.GET.get('record_state', None)
+        if self.is_rendered_as_email:
+            return []
         return [self._make_row(record) for record in
                 iter_repeat_records_by_domain(self.domain, repeater_id=repeater_id, state=state)]
 
@@ -58,10 +60,8 @@ class ENikshayForwarderReport(DomainForwardingRepeatRecords):
             DataTablesColumn(_('URL')),
             DataTablesColumn(_('Last sent date')),
             DataTablesColumn(_('Attempts')),
+            DataTablesColumn(_('Payload')),
         ]
-        if not self.is_rendered_as_email:
-            columns.append(DataTablesColumn(_('Payload')))
-
         return DataTablesHeader(*columns)
 
     def _make_row(self, record):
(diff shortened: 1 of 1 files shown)


gcapalbo (approved):

custom/enikshay/reports/repeaters.py
```diff
@@ -46,6 +46,8 @@ class ENikshayForwarderReport(DomainForwardingRepeatRecords):
     def get_all_rows(self):
         repeater_id = self.request.GET.get('repeater', None)
         state = self.request.GET.get('record_state', None)
+        if self.is_rendered_as_email:
+            return []
```
mkangia: It would be better to not even call this method if it is to return []. Or atleast return at first step instead of fetching repeater and state first?

mkangia (changes requested):
@proteusvacuum would it be possible to have the count of cancelled records for each repeater in the email body instead. It would help to take a quick look at that number to have an idea about the cancelled records without needing to open each report.

proteusvacuum:
@mkangia good idea.

custom/enikshay/reports/repeaters.py
```diff
@@ -46,22 +46,32 @@ class ENikshayForwarderReport(DomainForwardingRepeatRecords):
     def get_all_rows(self):
         repeater_id = self.request.GET.get('repeater', None)
         state = self.request.GET.get('record_state', None)
+        if self.is_rendered_as_email:
+            return [
+                [
+                    get_repeat_record_count(self.domain, repeater_id, "SUCCESS"),
+                    get_repeat_record_count(self.domain, repeater_id, "CANCELLED"),
+                ]
```
mkangia: @proteusvacuum I think we should include Failed in this as well.

custom/enikshay/reports/repeaters.py
```diff
@@ -71,7 +71,6 @@ def headers(self):
                 DataTablesColumn(_('URL')),
                 DataTablesColumn(_('Last sent date')),
                 DataTablesColumn(_('Attempts')),
-                DataTablesColumn(_('Payload')),
```
mkangia: 👍

mkangia (approved):
Looks good. 
Just requested to include Failed count as well if that makes sense. Not a blocker

custom/enikshay/reports/repeaters.py
proteusvacuum: Since failed is a transient state, I figured the count doesn't mean much.

PR #17431, opened 2017-08-10
Fold url registration into initial page data

orangejenny:
The current implementation of `registerurl` is based on inline javascript and isn't compatible with a js module system. Refactor to store urls in html elements, similar to initial page data.

@millerdev / @dannyroberts


millerdev (commented):
Looks good, but build failed:
```grunt mocha --noinput --stop --verbosity=2 --divide-depth=1 --with-timing --threshold=10
Running "mocha:app_manager" (mocha) task
Testing: http://localhost:8000/mocha/app_manager/
Warning: Error: The module 'hqwebapp/js/initial_page_data.js' has not yet been defined.
Did you include <script src="hqwebapp/js/initial_page_data.js"></script> on your html page?
    at hqImport (http://localhost:8000/static/hqwebapp/js/hqModules.js:51)
    at undefined (http://localhost:8000/static/hqwebapp/js/urllib.js:2)
    at hqDefine (http://localhost:8000/static/hqwebapp/js/hqModules.js:45)
    at undefined (http://localhost:8000/static/hqwebapp/js/urllib.js:46) Use --force to continue.```

**Comment by orangejenny** [comment]
Added a commit to combine urllib.js and initial_page_data.js, because they're both small files and now urllib depends on initial_page data.

**Review by millerdev — APPROVED** [review]
Looks good. My only slight reservation is that `initial_page_data` does not sound like as logical a place to find URL-related functions as `urllib` did. Naming is hard.

**Comment by orangejenny** [comment]
> Naming is hard.

Agree, definitely open to change in the future if either of us think of something better.

**Comment by emord** [comment]
@orangejenny @lwyszomi @kkrampa 

This broke the ICDS dashboard. It's not obvious to me why that is. The error on that page (/a/icds-cas/icds_dashboard/#) that we get is:

```
Uncaught Error: URL 'icds-ng-template' not found in registry
    at reverse (initial_page_data.js?version=171c510:80)
    at (index):211
reverse @ initial_page_data.js?version=171c510:80
(anonymous) @ (index):211
```

@orangejenny You can test htis out on any domain locally by enabling the ICDS dashboard flag. 

@lwyszomi @kkrampa you know the setup for the dashboard much better. could one of you pull the latest master and attempt to fix it?

orangejenny:
@emord Thanks for flagging. Looking into it, but I have limited connectivity today.

orangejenny:
@emord @lwyszomi @kkrampa Fixing dashboard issue in https://github.com/dimagi/commcare-hq/pull/17473

PR #17434, opened 2017-08-10
Another remote app 500

orangejenny:
http://manage.dimagi.com/default.asp?248127

Introduced in https://github.com/dimagi/commcare-hq/pull/17276

@sravfeyn / @millerdev


calellowitz (approved):

orangejenny:
not waiting on tests because this is blocking pact, want to get it deployed

PR #17437, opened 2017-08-11
log api access for the last 30 days

snopoke:
With increased usage of our APIs I thought it could be useful to log access.

### DIFF
--- a/corehq/apps/api/tasks.py
+++ b/corehq/apps/api/tasks.py
@@ -1,0 +1,10 @@
+import time
+from celery.schedules import crontab
+from celery.task.base import periodic_task
+from tastypie.models import ApiAccess
+
+
+@periodic_task(run_every=crontab(minute=0, hour=0), queue='background_queue')
+def clean_api_access():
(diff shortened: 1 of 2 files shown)


corehq/apps/api/tasks.py
```diff
@@ -0,0 +1,10 @@
+import time
+from celery.schedules import crontab
+from celery.task.base import periodic_task
+from tastypie.models import ApiAccess
+
+
+@periodic_task(run_every=crontab(minute=0, hour=0), queue='background_queue')
+def clean_api_access():
+    accessed = int(time.time()) - 30 * 24 * 3600  # only keep last 30 days
+    ApiAccess.objects.filter(accessed__lt=accessed).delete()
```
emord: With no index this might be a painful delete: https://github.com/django-tastypie/django-tastypie/blob/master/tastypie/models.py#L20

emord (approved):
Looks good to me. Only concern I have is the index, but not sure how easy it is to add one without making a PR

PR #17443, opened 2017-08-11
Add missing class less doc type for reference

mkangia:
The doc type RepeatRecord-Failed does not have a corresponding class and hence `CouchDataLoader` is unable to find the corresponding db for it.

I am not sure if this is the best fix but did work around for now.

### DIFF
--- a/corehq/util/couch.py
+++ b/corehq/util/couch.py
@@ -101,6 +101,8 @@ def get_classes_by_doc_type():
             if klass._doc_type not in classes_by_doc_type:
                 classes_by_doc_type[klass._doc_type] = klass
         queue.extend(klass.__subclasses__())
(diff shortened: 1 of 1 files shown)


snopoke:
This is ok as a short term solution but would be better to detect amended doc types (look for '-') and then get the correct class based on the trimmed doc type:
```
doc_type = doc_type.split('-')[0]
```

mkangia:
@snopoke 

> This is ok as a short term solution but would be better to detect amended doc types (look for '-') and then get the correct class based on the trimmed doc type

So we add a class with "RepeatRecordFailed" in this case and trim "-" when checking for doc type in `get_classes_by_doc_type`. Can make that change now.

PR #17444, opened 2017-08-11
Show location name in page title

esoergel:
@sravfeyn
This should make it a little easier to tell locations apart from tabs or url bars.

I'm also thinking about doing the same for user pages.


proteusvacuum:
<3

PR #17450, opened 2017-08-11
hqDefine: don't allow module names to end in .js

orangejenny:
Most of our hqDefine modules are name `appName/js/fileName.js`, but this doesn't play well with RequireJS, which does not expect the trailing `.js` and interprets it as a sign that the module is an absolute path rather than a relative path (see http://requirejs.org/docs/api.html). This makes for headaches when transitioning to requirejs, particularly for modules that are used in both the old and new world. Since we don't have a reason to include the `.js`, I'm changing the convention to instead exclude it.

The only remotely interesting file to review in this PR is hqModules.js

@millerdev / @dimagi/js-team


calellowitz (approved):

millerdev (approved):

millerdev:
Tests failed:
```
grunt mocha --noinput --stop --verbosity=2 --divide-depth=1 --with-timing --threshold=10
Running "mocha:app_manager" (mocha) task
Testing: http://localhost:8000/mocha/app_manager/
Warning: TypeError: 'undefined' is not a function (evaluating 'path.endsWith(".js")')
    at hqDefine (http://localhost:8000/static/hqwebapp/js/hqModules.js:46)
    at undefined (http://localhost:8000/mocha/app_manager/:217) Use --force to continue.
```

dannyroberts (approved):
Looks good, and seems like a reasonable convention to have if that's what requireJS expects.

Up to you whether to address the stickler comments, which all seem real though unrelated to the changes in the PR.

PR #17451, opened 2017-08-11
Fix prescription_total_days calculation

esoergel:
"approved" vouchers have also been fulfilled
@proteusvacuum

### DIFF
--- a/custom/enikshay/tasks.py
+++ b/custom/enikshay/tasks.py
@@ -31,6 +31,7 @@ from custom.enikshay.exceptions import ENikshayCaseNotFound
 from .const import (
     DOSE_MISSED,
     DOSE_TAKEN_INDICATORS,
+    DATE_FULFILLED,
     DAILY_SCHEDULE_FIXTURE_NAME,
     DAILY_SCHEDULE_ID,
     SCHEDULE_ID_FIXTURE,
@@ -523,7 +524,7 @@ class EpisodeVoucherUpdate(object):
         relevant_vouchers = [
             voucher for voucher in self._get_all_vouchers()
             if (voucher.get_case_property('voucher_type') == 'prescription'
-                and voucher.get_case_property('state') == 'fulfilled')
+                and voucher.get_case_property(DATE_FULFILLED))
         ]
         return sorted(relevant_vouchers, key=self._get_fulfilled_voucher_date)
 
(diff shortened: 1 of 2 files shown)


custom/enikshay/tasks.py
```diff
@@ -523,7 +524,7 @@ def _get_fulfilled_vouchers(self):
         relevant_vouchers = [
             voucher for voucher in self._get_all_vouchers()
             if (voucher.get_case_property('voucher_type') == 'prescription'
-                and voucher.get_case_property('state') == 'fulfilled')
+                and voucher.get_case_property(DATE_FULFILLED))
```
esoergel: @proteusvacuum is this a reasonably property to check?  This could also check for a state within a list of all states "fulfilled" or later

custom/enikshay/tests/test_voucher_counts.py
```diff
@@ -49,7 +59,17 @@ def test_thresholds(self):
...
+        self.assertEqual(
+            EpisodeVoucherUpdate(self.domain, self.cases['episode']).get_prescription_total_days(),
+            {
+                "prescription_total_days": 29 + 1 + 7 + 70,
+                "prescription_total_days_threshold_30": '2017-01-02T00:00:00.000000Z',
+                "prescription_total_days_threshold_60": '2017-01-04T00:00:00.000000Z',
+                "prescription_total_days_threshold_90": '2017-01-04T00:00:00.000000Z',
+            }
+        )
+        self.approve_voucher(voucher)  # approving the voucher shouldn't change anything
```
esoergel: This diff is misleading - the `assertEqual` right below this is the new one.  This test and the other one I added fail without the code change here.

custom/enikshay/tests/test_voucher_counts.py
```diff
@@ -155,12 +181,10 @@ def test_voucher_generation_dates(self):
             'drugs_ordered_readable': "Happy Pills, Sad Pills, Buggy Pillz"
         })
         voucher11 = get_voucher_case_structure(None, prescription1.case_id, {
-            'date_fulfilled': '2013-01-01',
```
esoergel: I think this test data is incorrect - shouldn't vouchers have `date_fulfilled` set only when the `state` is set to `fulfilled`?

custom/enikshay/tasks.py
proteusvacuum: this looks good to me. I just checked, and this property only gets set when the state is set to fulfilled, so this is a good proxy for the voucher ever having been in the "fulfilled" state. 👍

custom/enikshay/tests/test_voucher_counts.py
proteusvacuum: yes. nice catch.

PR #17457, opened 2017-08-14
order data dump correctly

snopoke:


gcapalbo (approved):

mkangia (approved):

mkangia:
I think we should add tests for it if possible .. so that this is caught in case we add/remove anything

PR #17459, opened 2017-08-14
Add docs to import_drtb_cases management command

NoahCarnahan:
Adding docs to make handoff easier.
@mkangia


NoahCarnahan:
Self merging because it's just docs

PR #17460, opened 2017-08-14
Use FileWrapper for both kinds of blobs

kaapstorm:
Although `blob` has an `__iter__` method, it doesn't guarantee it can iterate. :expressionless: In the case of S3 blobs, `blob.__iter__()` calls `iter(self._obj)`, which fails when `self._obj` isn't iterable. 

For simplicity, just use `FileWrapper` for both S3 and file system blobs.

Context: [FB 2158588](http://manage.dimagi.com/default.asp?258588)

@biyeun


biyeun (approved):

biyeun:
Looks good! waiting on tests

kaapstorm:
Thanks ... :white_check_mark: ... merging.

PR #17462, opened 2017-08-14
update internal subscription ending email

nickpell:
@emord


emord (approved):
Not related to this change but it seems weird that the plaintext email is email_name_plaintext.html instead of email_name.txt

PR #17464, opened 2017-08-14
ICDS: Fix Number of Attended Children Indicator

sheelio:
@emord 

This expression previously handled a bug in the application.  The new version of the application has been released that fixes this bug.  

However, don't have a good way to differentiate the old / new version of the application, so relying on a hidden value that was added to the form.  Let me know if if there's a better way to deal with this.


sheelio:
Actually it turns out /data/gps_location existed for while.  Going to try a different way to handle this.

PR #17468, opened 2017-08-15
Refactor test_sync_mode to use MockDevice

millerdev:
Probably best reviewed by commit 🐠 Don't miss the collapsed `test_sync_mode.py` file, which is where the bulk of the work happened, if reviewing all changes at once.

To assist in this refactoring effort, I created utilities (not part of this PR) to verify that...
- the XML posted to HQ by `MockDevice` is equivalent to what was posted before when interacting directly with the `CaseFactory`.
- the restore config and payload generated by `MockDevice.sync()` are equivalent to what was restored before when using the various restore helper/assert functions.

While these utilities were not used to refactor every single test, they gave me a degree of confidence that the refactoring has not changed what is being tested.

@proteusvacuum @snopoke cc @orangejenny

### DIFF
--- a/corehq/ex-submodules/casexml/apps/case/tests/util.py
+++ b/corehq/ex-submodules/casexml/apps/case/tests/util.py
@@ -105,20 +105,24 @@ def check_xml_line_by_line(test_case, expected, actual):
 
 
 def assert_user_has_case(testcase, user, case_id, **kwargs):
+    """DEPRECATED use assertIn(case_id, <MockDevice>.sync().cases)"""
     return assert_user_has_cases(testcase, user, [case_id], return_single=True, **kwargs)
 
 
 def assert_user_has_cases(testcase, user, case_ids, **kwargs):
+    """DEPRECATED use assertTrue(set(case_ids) <= set(<MockDevice>.sync().cases))"""
     case_blocks = [CaseBlock(case_id=case_id).as_xml() for case_id in case_ids]
     return check_user_has_case(testcase, user, case_blocks,
                                should_have=True, line_by_line=False, **kwargs)
 
 
 def assert_user_doesnt_have_case(testcase, user, case_id, **kwargs):
+    """DEPRECATED use assertNotIn(case_id, <MockDevice>.sync().cases)"""
     return assert_user_doesnt_have_cases(testcase, user, [case_id], return_single=True, **kwargs)
 
 
 def assert_user_doesnt_have_cases(testcase, user, case_ids, **kwargs):
+    """DEPRECATED use assertFalse(set(case_ids) & set(<MockDevice>.sync().cases))"""
     case_blocks = [CaseBlock(case_id=case_id).as_xml() for case_id in case_ids]
     return check_user_has_case(testcase, user, case_blocks,
                                should_have=False, **kwargs)
@@ -137,6 +141,7 @@ def extract_caseblocks_from_xml(payload_string, version=V2):
 @contextmanager
 def cached_restore(testcase, user, restore_id="", version=V2,
                    purge_restore_cache=False):
+    """DEPRECATED use <MockDevice>.sync().cases"""
     assert not hasattr(testcase, 'restore_config'), testcase
     assert not hasattr(testcase, 'payload_string'), testcase
 
@@ -158,6 +163,7 @@ def cached_restore(testcase, user, restore_id="", version=V2,
 def check_user_has_case(testcase, user, case_blocks, should_have=True,
                         line_by_line=True, restore_id="", version=V2,
                         purge_restore_cache=False, return_single=False):
+    """DEPRECATED use <MockDevice>.sync().cases"""
 
     try:
         restore_config = testcase.restore_config
@@ -167,7 +173,7 @@ def check_user_has_case(testcase, user, case_blocks, should_have=True,
             restore_config = testcase.restore_config
             payload_string = testcase.payload_string
 
-    return check_payload_has_cases(
+    return _check_payload_has_cases(
         testcase=testcase,
         payload_string=payload_string,
         username=user.username,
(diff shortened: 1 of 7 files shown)


corehq/ex-submodules/casexml/apps/phone/tests/test_sync_mode.py
```diff
@@ -2165,98 +1984,33 @@ class LiveQuerySteadyStateExtensionSyncTestSQL(LiveQuerySteadyStateExtensionSync
...
-            user_id=self.user_id,
-            owner_id=self.user_id,
-            case_type=PARENT_TYPE,
-        ).as_xml()
-        try:
-            self._postFakeWithSyncToken(caseblock, self.sync_log.get_id)
-            self.fail('posting an update to a non-existant case should fail')
-        except AssertionError:
-            # this should fail because it's a true error
-            pass
```
millerdev: Removed this test because `self.fail(...)` raises `AssertionError`, which means the test will never fail. I have no idea what the "true error" comment was supposed to mean.

corehq/ex-submodules/casexml/apps/phone/tests/test_sync_mode.py
```diff
@@ -2165,98 +1984,33 @@ class LiveQuerySteadyStateExtensionSyncTestSQL(LiveQuerySteadyStateExtensionSync
...
-
-        try:
-            post_case_blocks(
-                _get_bad_caseblocks([case_id2, case_id1]),
-                form_extras={ "last_sync_token": self.sync_log._id }
-            )
-            self.fail('posting an update to non-existant cases should fail')
-        except AssertionError:
-            # this should fail because it's a true error
-            pass
```
millerdev: Removed this test because `self.fail(...)` raises `AssertionError`, which means the test will never fail. I have no idea what the "true error" comment was supposed to mean.

millerdev:
@dannyroberts, I'd be grateful if you were able to review this.

corehq/ex-submodules/casexml/apps/phone/tests/test_sync_mode.py
```diff
@@ -90,7 +90,7 @@ def setUp(self):
             self.restore_options,
             default_case_type=PARENT_TYPE,
         )
-        self.sync_log = self.device.last_sync.log
+        self.sync_log = self.device.sync(overwrite_cache=True, version=V1).log
```
dannyroberts: What's the idea behind this?

corehq/ex-submodules/casexml/apps/phone/tests/utils.py
```diff
@@ -171,7 +172,8 @@ def __init__(self, project, user, restore_options, default_case_type="case"):
             },
         )
         self.last_sync = None
-        self.sync(overwrite_cache=True)
+        if sync:
+            self.sync()
```
dannyroberts: Why no `overwrite_cache=True` here?

corehq/ex-submodules/casexml/apps/phone/tests/test_sync_mode.py
```diff
@@ -1977,8 +1977,8 @@ def create_case_graph(num):
         p2, o2, e2 = create_case_graph(2)
 
         # Alice and Bob sync
-        alice = self.get_device()
-        bob = self.get_device(user=self.other_user)
+        alice = self.get_device(sync=True)
```
dannyroberts: Why isn't this just

```python
alice = self.get_device()
alice.sync()
```

I guess it's more lines. But it is more explicit about what's happening.

corehq/ex-submodules/casexml/apps/phone/tests/test_sync_mode.py
```diff
@@ -165,7 +156,23 @@ def _testUpdate(self, sync_log_or_id, case_id_map, dependent_case_id_map=None):
             self._testUpdate(migrated_sync_log, case_id_map, dependent_case_id_map)
 
 
-class SyncTokenUpdateTest(SyncBaseTest):
+class SyncBaseTest(BaseSyncTest):
```
dannyroberts: Maybe change this name to `DeprecatedBaseSyncTest`? (Right now the only difference in their names is transposing Sync and Base, which doesn't really tell you anything about which is preferred.)

corehq/ex-submodules/casexml/apps/case/tests/util.py
```diff
@@ -105,20 +105,24 @@ def check_xml_line_by_line(test_case, expected, actual):
 
 
 def assert_user_has_case(testcase, user, case_id, **kwargs):
+    """DEPRECATED use assertIn(case_id, <MockDevice>.sync().cases)"""
```
dannyroberts: worth using

```python
warnings.warn("use assertIn(case_id, <MockDevice>.sync().cases)", DeprecationWarning)
```

or changing name to `deprecated_assert_user_has_case` instead?

dannyroberts (approved):
Looks great!

corehq/ex-submodules/casexml/apps/phone/tests/test_sync_mode.py
millerdev: This is what it did [before](https://github.com/dimagi/commcare-hq/blob/aa13efd674f130c22b0f0e7ac769832684c524f5/corehq/ex-submodules/casexml/apps/phone/tests/test_sync_mode.py#L80-L83). `RestoreConfig` uses [`version=V1` by default](https://github.com/dimagi/commcare-hq/blob/5bd6c3ba709f380d9f14f6643e73ccb044fe8bef/corehq/ex-submodules/casexml/apps/phone/restore.py#L410).

corehq/ex-submodules/casexml/apps/phone/tests/test_sync_mode.py
millerdev: it's just for convenience.

corehq/ex-submodules/casexml/apps/phone/tests/test_sync_mode.py
millerdev: Good idea. I'll do it in a separate PR. It's used in a lot of places, which is why I did not change it in this PR.

corehq/ex-submodules/casexml/apps/phone/tests/utils.py
millerdev: I don't think it's necessary. Does a sync without a sync token hit the cache?

corehq/ex-submodules/casexml/apps/case/tests/util.py
millerdev: Good idea. Will do this in a separate PR.

PR #17469, opened 2017-08-15
don't load all cases in one query

snopoke:


mkangia (approved):
Looks good!
Thank you for resolving this so quickly.

corehq/form_processor/backends/sql/dbaccessors.py
```diff
@@ -643,12 +649,23 @@ def get_doc(self, doc_id):
...
+
+        # using raw query to avoid having to expand the tuple comparison
+        results = CommCareCaseSQL.objects.using(from_db).raw(
+            """SELECT * FROM {table} as case_table
+            WHERE {domain_clause}
+            deleted = %s AND
+            (case_table.server_modified_on, case_table.id) > (%s, %s)
+            ORDER BY case_table.server_modified_on, case_table.id
+            LIMIT %s;""".format(
+                table=CommCareCaseSQL._meta.db_table, domain_clause=domain_clause
```
proteusvacuum: Is there a way I can name my domain `; DROP ALL TABLES;` or something similar where this might become an issue?

snopoke: um, no. The domain name is passed in as a query parameter so it's not subject to SQL injection attacks. The string formatting is just the case table name and `case_table.domain = %s AND`.

proteusvacuum: oh I see that now. Thanks for clarifying.

PR #17470, opened 2017-08-15
allow limiting reinder to DB

snopoke:
The command argument handling is a bit messy but will have to do for now.


gcapalbo (approved):
nice. are these reindexers still resumable within a specific db run?

PR #17471, opened 2017-08-15
ICDS: Fix aadhar indicator

sheelio:
@emord 

Apparently there is an app bug where the value has_aadhar is not getting correctly set in all situations.  This changes the way that indicator is calculated.

Still figuring out with the app team if we need to retrigger calculating this table for situations where aadhar_number is not null and has_aadhar != yes


emord (approved):
This table takes long enough to rebuild, that it may be worthwhile coming up with a custom rebuild command

PR #17473, opened 2017-08-15
Make url registration available to ICDS dashboard

orangejenny:
https://github.com/dimagi/commcare-hq/pull/17431#issuecomment-322340709

Happening because the ICDS dashboard doesn't inherit from `style/base.html`.

@millerdev / @emord


emord:
Looks good. Thanks for investigating this

PR #17474, opened 2017-08-15
update django-countries

nickpell:
https://github.com/SmileyChris/django-countries/blob/master/CHANGES.rst


nickpell:
@emord tested on staging, there are countries!

emord (approved):

PR #17476, opened 2017-08-15
Allow further partitioning of case reindexes

gcapalbo:
@snopoke


gcapalbo:
no need to merge, just going to use this to make the process easier right now

snopoke:
Do you think that SQL will perform OK? Worth a try I guess.

gcapalbo:
it shouldn't try to use any different indexes i don't think, so it should perform roughly as it is now

PR #17479, opened 2017-08-15
django 1.11

nickpell:


corehq/form_processor/migrations/0043_rename_to_match.py
```diff
@@ -23,12 +23,11 @@ class Migration(migrations.Migration):
         _rename_index('form_processor_xforminstancesql_form_uuid_key', 'form_processor_xforminstancesql_form_id_key'),
         _rename_index('form_processor_xformattachmentsql_attachment_uuid_key', 'form_processor_xformattachmentsql_attachment_id_key'),
         _rename_index('form_processor_xforminstancesql_form_uuid_d939b4d9_like', 'form_processor_xformattachmentsql_d6cba1ad'),
-        _rename_index('form_processor_xformoperationsql_form_id_bfa39e5e_uniq', 'form_processor_xformoperationsql_d6cba1ad'),
+        _rename_index('form_processor_xformoperationsql_form_id_bfa39e5e_like', 'form_processor_xformoperationsql_d6cba1ad'),
```
nickpell: @snopoke Do the changes in this file look ok to you?

snopoke: not sure - I'd have to dig a bit. It does seem odd that we'd go from a `uniq` index to a `like` index. I don't recall why we had to do this in the first place.

emord: I think we should be dropping like indexes on these fields and renaming unique indexes.

@nickpell assuming this only affects fresh installs? can you setup your test database locally and make sure that the indexes that are referenced here are unique constraints and not LIKE indexes?

nickpell: I looked into this further and that the uniq indices had the ```_uniq``` suffix truncated in django 1.11 and updated the migration accordingly.

nickpell:
This is good to go now.  I'll leave the do not merge label on until after the tech team summit

emord (approved):
Looks good assuming there's no way to decouple those atters changes

nickpell:
yeah, I couldn't think of a good way to decouple them

PR #17482, opened 2017-08-15
Allow further partitioning of case reindexes

gcapalbo:
@snopoke

### DIFF
--- a/corehq/form_processor/backends/sql/dbaccessors.py
+++ b/corehq/form_processor/backends/sql/dbaccessors.py
@@ -1,3 +1,4 @@
+import os
 import itertools
 import logging
 import struct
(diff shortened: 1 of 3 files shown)


snopoke (approved):

gcapalbo:
going to close this since we don't want to merge. in case we need to use this again some time, stats for this were:

- it took about 11-12 hours to run the case-search reindex on each db (without the splitting in this pr)
- introducing 5 processes per db with PARTITION_SIZE of 5 and PARTITION_NUM 0-4 reduced it to 3-4 hours per db (10 dbs * 5 processes each meant 50 processes running total)
- so was able to run the case-search and sql-case reindexes in 3-4 hours

PR #17485, opened 2017-08-15
Add command to import offline voucher payments

esoergel:
@proteusvacuum @nickpell
This command will import voucher payments processed offline and update our voucher cases accordingly, as well as reconcile voucher repeat records with those updates.

### DIFF
--- a/custom/enikshay/management/commands/import_voucher_confirmations.py
+++ b/custom/enikshay/management/commands/import_voucher_confirmations.py
@@ -81,9 +81,9 @@ class Command(BaseCommand):
                 print "\nAborting."
                 return
 
-            rows = [r for r in reader]
+            rows = list(reader)
 
-        print "Received info on {} vouchers.  Headers are:".format(len(rows) - 1)
+        print "Received info on {} vouchers.  Headers are:".format(len(rows))
         for header in headers:
             print header
 
@@ -221,13 +221,16 @@ class Command(BaseCommand):
                         iter_db.save(record)
                     else:
                         # mark record as canceled
-                        status = "cancelled"
-                        record.cancelled = True
-                        record.succeeded = False
-                        record.failure_reason = ''
-                        record.overall_tries = 0
-                        record.next_check = None
+                        record.add_attempt(RepeatRecordAttempt(
+                            cancelled=True,
+                            datetime=datetime.datetime.utcnow(),
+                            failure_reason="Cancelled during import_voucher_confirmations",
+                            success_response=None,
+                            next_check=None,
+                            succeeded=False,
+                        ))
                         iter_db.save(record)
+
                     already_seen.add(record.payload_id)
                     rows.append([record._id, record.payload_id, status])
 
--- a/custom/enikshay/management/commands/import_voucher_confirmations.py
+++ b/custom/enikshay/management/commands/import_voucher_confirmations.py
@@ -1,4 +1,6 @@
 import csv
+import datetime
+from mock import MagicMock
 
 from django.core.management.base import BaseCommand
 
@@ -6,12 +8,14 @@ from dimagi.utils.chunked import chunked
 from dimagi.utils.decorators.memoized import memoized
 
 from corehq.apps.hqcase.utils import bulk_update_cases
-from corehq.apps.users.models import CommCareUser
 from corehq.form_processor.interfaces.dbaccessors import CaseAccessors
+from corehq.motech.repeaters.models import RepeatRecord, RepeatRecordAttempt
+from corehq.motech.repeaters.dbaccessors import iter_repeat_records_by_domain
+from corehq.util.couch import IterDB
 from corehq.util.log import with_progress_bar
 
-from custom.enikshay.case_utils import get_person_case_from_voucher, CASE_TYPE_VOUCHER
-from custom.enikshay.const import PERSON_FIRST_NAME, PERSON_LAST_NAME, VOUCHER_ID
+from custom.enikshay.case_utils import CASE_TYPE_VOUCHER
+from custom.enikshay.const import VOUCHER_ID
 from custom.enikshay.integrations.bets.repeater_generators import VoucherPayload
 from custom.enikshay.integrations.bets.views import VoucherUpdate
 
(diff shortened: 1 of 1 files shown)


nickpell:
build is failing - may need to merge master

esoergel:
@nickpell this is ready for review now, happy to walk you through it if you want.
@millerdev code buddy

custom/enikshay/management/commands/import_voucher_confirmations.py
```diff
@@ -0,0 +1,234 @@
...
+        print "updating voucher cases"
+        for chunk in chunked(with_progress_bar(voucher_updates), 100):
+            updates = [
+                (update.case_id, update.properties, False)
+                for update in chunk
+            ]
+            if self.commit:
+                bulk_update_cases(self.domain, updates)
+
+    def reconcile_repeat_records(self, voucher_updates):
```
esoergel: @dannyroberts would you mind reviewing just this method?  We have two custom repeaters for the `voucher` case type, and this is meant to reconcile those repeat records with the vouchers being updated via this case import.  I want "successful" cases to have their repeat records marked as successful, and any records not included in the dump marked as "cancelled", so we can later requeue them.  We have a separate command that checks for duplicates, but we only want each record sent once.

custom/enikshay/management/commands/import_voucher_confirmations.py
```diff
@@ -0,0 +1,234 @@
...
+            for voucher in self.accessor.iter_cases(voucher_ids)
+            if voucher.get_case_property('state') in (
+                'approved', 'paid', 'rejected', 'expired', 'cancelled', 'canceled')
+            or voucher.get_case_property('voucher_approval_status') in ('approved', 'partially_approved')
+        }
+
+    def write_csv(self, filename, headers, rows):
+        filename = "voucher_confirmations-{}.csv".format(filename)
+        print "writing {}".format(filename)
+        with open(filename, 'w') as f:
```
millerdev: do you care if this overwrites an existing file without warning? if so, you might want to check if it exists first.

custom/enikshay/management/commands/import_voucher_confirmations.py
```diff
@@ -0,0 +1,234 @@
...
+        self.write_csv('updates', headers, rows)
+
+    def update_vouchers(self, voucher_updates):
+        print "updating voucher cases"
+        for chunk in chunked(with_progress_bar(voucher_updates), 100):
+            updates = [
+                (update.case_id, update.properties, False)
+                for update in chunk
+            ]
+            if self.commit:
```
millerdev: looks like this could be moved to the beginning of the function rather than inside this loop. I'm envisioning something like this:
```py
def update_vouchers(self, voucher_updates):
    if not self.commit:
        print("dry run: not updating voucher cases")
        return
    print("updating voucher cases")
    ...
```

custom/enikshay/management/commands/import_voucher_confirmations.py
```diff
@@ -0,0 +1,234 @@
...
+                        record.add_attempt(attempt)
+                        iter_db.save(record)
+                    else:
+                        # mark record as canceled
+                        status = "cancelled"
+                        record.cancelled = True
+                        record.succeeded = False
+                        record.failure_reason = ''
+                        record.overall_tries = 0
+                        record.next_check = None
```
dannyroberts: How come you don't add a `RepeatRecordAttempt` here? I'd recommend always doing that as the mechanism for changing the top level vars.

custom/enikshay/management/commands/import_voucher_confirmations.py
dannyroberts: a little hard to wrap my mind around. Are you basically treating repeat records of a certain type as a log of completed payments, and then going through that log and finding the ones that correspond to the same payments you're making with the upload?

> so we can later requeue them

when would you requeue them?

(maybe i just need fuller context)

millerdev (approved):
Few nitpicky comments, nothing blocking. I don't have a lot of context here, so hard for me to evaluate what the code is actually doing, and if it's doing it correctly.

custom/enikshay/management/commands/import_voucher_confirmations.py
```diff
@@ -0,0 +1,234 @@
...
+        )
+
+    def handle(self, domain, filename, **options):
+        self.domain = domain
+        self.accessor = CaseAccessors(domain)
+        self.commit = options['commit']
+
+        with open(filename) as f:
+            reader = csv.reader(f)
+            headers = reader.next()
```
esoergel: oh cool, I didn't know about that, thanks!

custom/enikshay/management/commands/import_voucher_confirmations.py
```diff
@@ -0,0 +1,234 @@
...
+            if missing_headers:
+                print "Missing the following headers:"
+                for header in missing_headers:
+                    print " ", header
+                print "\nAborting."
+                return
+
+            rows = [r for r in reader]
+
+        print "Received info on {} vouchers.  Headers are:".format(len(rows) - 1)
```
esoergel: Yup, will fix, thanks.

custom/enikshay/management/commands/import_voucher_confirmations.py
esoergel: I think that's fine - if a file already exists, it's because I'm running this same command multiple times.  I'm planning to do a no-commit test run, look at the results, then do a real run.  I'd expect to overwrite the files from the test run.

custom/enikshay/management/commands/import_voucher_confirmations.py
esoergel: It was actually originally that, but I wanted the non-commit version to run as much as possible (for instance, in case something crazy happened and `chunk` contained things that weren't these update objects).

custom/enikshay/management/commands/import_voucher_confirmations.py
esoergel: updated.

dannyroberts: 👍 +1

millerdev (approved):

PR #17489, opened 2017-08-15
Fix UI Element Key Val List JS Error

biyeun:
@orangejenny 

noticed this while refactoring the groups widget.

PR also updates the groups page to use the proper include file for `ui-element.js`


orangejenny (approved):

biyeun:
commits got merged with another PR

PR #17490, opened 2017-08-15
upgrade pygooglechart

nickpell:


emord:
It's probably worth killing this and replacing it with something else as I don't think it's supported anymore. I think it's just for QR codes and a scatter plot. QR code should be easy to replace

nickpell:
@emord Agree we should kill it eventually, but for now I'd rather just update it so unblock py3

emord (approved):
makes sense. good with me if you've tested where we use this on staging

PR #17491, opened 2017-08-15
update django-transfer

nickpell:


nickpell:
@emord tested on staging, looks good

emord (approved):
still has don't merge tag

PR #17495, opened 2017-08-16
adding command for second data dump

calellowitz:
@dannyroberts @emord 
management command for a data dump the icds team is requesting


emord (approved):

emord:
the travis build hasn't passed from your intial commit yet. so not merging at the moment. although I don't htink this will have test coverage

PR #17497, opened 2017-08-16
Add dashboard QA domain to UCRs.

emord:
@biyeun would be good to get this in before India deploy

buddy @nickpell


emord:
(looks like appveyor biulds are failing. not sure what those are so can probably be ignored)

PR #17499, opened 2017-08-16
Fix restore cache invalidation

dannyroberts:
Should fix https://manage.dimagi.com/default.asp?260109

Most commits are to
* make it more obvious what the code is doing, splitting `restore_cache_key` into one function per `prefix`, making arguments explicit, etc.

This commit https://github.com/dimagi/commcare-hq/pull/17499/commits/402e87c7f647b618dbc2a8ba473b706185f52193 is the only one that makes a change, to
* invalidate the async task id cache using the same key that's used to fetch it

### DIFF
--- a/corehq/ex-submodules/casexml/apps/phone/restore.py
+++ b/corehq/ex-submodules/casexml/apps/phone/restore.py
@@ -698,8 +698,8 @@ class RestoreConfig(object):
     @property
     def _restore_cache_key(self):
         return restore_payload_path_cache_key(
-            self.domain,
-            self.restore_user.user_id,
+            domain=self.domain,
+            user_id=self.restore_user.user_id,
             version=self.version,
             sync_log_id=self.sync_log._id if self.sync_log else '',
             device_id=self.params.device_id,
--- a/corehq/ex-submodules/casexml/apps/phone/restore.py
+++ b/corehq/ex-submodules/casexml/apps/phone/restore.py
@@ -68,7 +68,7 @@ logger = logging.getLogger(__name__)
 DEFAULT_CASE_SYNC = CLEAN_OWNERS
 
 
-def _restore_cache_key(domain, prefix, user_id, version=None, sync_log_id=None, device_id=None):
+def _restore_cache_key(domain, prefix, user_id, version, sync_log_id, device_id):
     response_class = get_restore_response_class(domain)
     hashable_key = '{response_class}-{prefix}-{user}-{version}-{sync_log_id}-{device_id}'.format(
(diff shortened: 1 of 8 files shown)


dannyroberts:
This change is definitely good (assuming tests pass), but I don't yet have an assurance that it actually solves the problem in https://manage.dimagi.com/default.asp?260109, just a reasonable story in my head as to why it would.

First off, the cache invalidation (attempt) for async restore task id going on in `SubmissionPost` was definitely not working, as it was using `None` for `sync_token_id` and `device_id` in the cache key, whereas the getter and setter are using actual values there.

But why do I think that fixing the cache invalidation for async restore task id will fix that problem? (Or alternatively, how could that problem be caused by bad cache invalidation of the async restore task id?) My understanding is that async restore will try to pull up a currently running task. If none exists (in the cache) then it will start a new one; if one does exist, then it tries to fetch the result of it using `task.get(timeout=...)`. If that times out then it returns a 202. If it doesn't time out, then it returns that result. If my understanding there is right, then it seems pretty clear that pulling up the "wrong" (or stale) async task id will lead to getting a wrong/stale response like we're seeing in that ticket.

If my summary above's not correct, it'd be great to be corrected by one of you folks, even if it's just a minor point. Specifically, if async restore response relies in any way on the (synchronous) restore payload cache then that would be useful and interesting to know. (From what I can tell, that's just a totally parallel bit of caching for initial synchronous restores, solving a totally different problem.)

calellowitz (approved):
This looks good to me. Agree that these changes are definitely positive, but also do not have enough context on async restore to know if your story about how this fixes the bug is true.

dannyroberts:
Great, ok, I'm going to merge it and see what happens!

dannyroberts:
Clayton's integration test still doesn't pass with this so I'm going to continue looking today

corehq/ex-submodules/casexml/apps/phone/restore.py
```diff
@@ -81,6 +81,28 @@ def restore_cache_key(domain, prefix, user_id, version=None, sync_log_id=None, d
...
+        device_id=device_id,
+    )
+
+
+def async_restore_task_id_cache_key(domain, user_id, sync_log_id, device_id):
+    return _restore_cache_key(
+        domain=domain,
+        prefix=ASYNC_RESTORE_CACHE_KEY_PREFIX,
+        user_id=user_id,
+        version=None,
```
millerdev: is this `None` because only V2 restores are possible with async restores? what if we someday make V3?

PR #17500, opened 2017-08-16
let enikshay reports through

nickpell:


corehq/apps/domain/middleware.py
```diff
@@ -72,6 +72,11 @@ def process_view(self, request, view_func, view_args, view_kwargs):
             if getattr(view_func, 'domain_migration_handled', False):
                 return None
             if DATA_MIGRATION.enabled(request.domain):
+                # hack for enikshay
+                url_parts = request.path.split('/')
+                if len(url_parts) > 3 and url_parts[3] in ['reports', 'toggles.js']:
```
mkangia: why do we need toggle.js?

gcapalbo (approved):
seems ok to do for data checkout now, but i think we should revert this once we no longer need it

corehq/apps/domain/middleware.py
nickpell: This gets loaded with reports and triggers the migration in progress page

nickpell:
@gcapalbo Agree, will revert afterwards

corehq/apps/domain/middleware.py
mkangia: okay cool.

mkangia (approved):

PR #17501, opened 2017-08-16
remove unneeded flag

emord:
This has been randomness = 1.0 for 3 months. think it's safe to remove

@nickpell


emord:
@millerdev Hey It looks like the livequery tests are dependent on old restores. It's probably just a test change that's required, but figured you should know that this isn't actually being tested on the correct restores

dannyroberts:
Looks like
```
ERROR: casexml.apps.phone.tests.test_sync_mode:SyncTokenCachingTest.testCacheInvalidationAfterFileDelete
  File "corehq/ex-submodules/casexml/apps/phone/tests/test_sync_mode.py", line 1360, in testCacheInvalidationAfterFileDelete
    os.remove(original_payload.get_filename())
OSError: [Errno 2] No such file or directory: 'restore-response-d122e0de59cd41aa9c12b5ce4d4a2b9e.xml'
```
is the test failure (or an example of the kind of test failure) @emord was referring to. (Mostly leaving this as a note for myself.)

millerdev:
I'm going to take a look at this today and see if I can fix the tests for this PR.

> It looks like the livequery tests are dependent on old restores. It's probably just a test change that's required, but figured you should know that this isn't actually being tested on the correct restores

@emord I'm not sure what you mean. Are you talking about the restore `version` parameter? As far as I know, livequery uses the same restore versions as all other restore tests. Currently `MockDevice.sync()` defaults to `version=V2` (many tests are setup with a `version=V1` initial sync). Yesterday [while refactoring more tests](https://github.com/dimagi/commcare-hq/pull/17549) I noticed that `V3` is a thing. Is that what you mean? I don't know much about the different versions other than I think V1 does not support extension cases while V2 does. Don't know anything about V3.

emord:
@millerdev I was referring to the feature flag I was removing. It's a PredictablyRandomToggle which is always false during tests. 

from danny's comment:

>    os.remove(original_payload.get_filename())

the blob restore does not have get_filename since it isn't a file. I haven't had a chance to look too deeply but I think invalidating the cache from the restore response class is probably the way to go

millerdev:
@emord ok, that's not as complicated as I was thinking. I can fix that.

millerdev (approved):
two of the commits are mine, so someone else should probably review those before merging.

emord:
Oh. excellent. thanks for doing that. it looks good to me

PR #17504, opened 2017-08-16
Only count properties set on the provided case

esoergel:
@proteusvacuum @nickpell


corehq/ex-submodules/casexml/apps/case/tests/test_property_changed_date.py
```diff
@@ -40,3 +46,46 @@ def test_date_case_property_changed(self):
             updated_on.replace(tzinfo=pytz.UTC),
             get_datetime_case_property_changed(case, "abc", "updated")
         )
+
+    def test_multiple_cases_in_update(self):
```
esoergel: This test fails without the change to `get_datetime_case_property_changed`, a case actions can contain updates from multiple cases.

proteusvacuum:
Thanks.

PR #17507, opened 2017-08-17
async queue stop after 20 attempts

snopoke:


emord:
So the reason that I initially didn't do this was because I didn't want to overwhelm the queue with too many tasks. In the case of the monthly indicators on ICDS they are very CPU intensive and take quite a long while to calculate. If this were deployed I think the celery queue would grow until a full day of processing is in the queue. When that happens we will start to have the same tasks queued up multiple times.

I think that the base algorithm for queuing should stay the same, but we should add datadog monitoring with an alert to say "There are too many unsuccessful attempts in queues. please take a look". thoughts?

snopoke:
I see I left of the 'limit'. That was unintentional. I'll add it back.

emord:
> I see I left of the 'limit'. That was unintentional. I'll add it back.

You kept the limit on here. The problem is that limit in SQL is applied after the filter is applied so your query would return the first x records that aren't already queued

snopoke:
@emord why is that an issue?

emord:
The issue is that we make the assumption that if the indicator was queued 24 hours ago, that celery has crashed and we weren't able to null the date_queued and increment unsuccessful_attempts. When that happens we requeue the indicator.

This means the rabbitmq queue needs to always have fewer tasks than can be processed in 24 hours, but the async indicator table can have infinitely many.

I think that if this query were changed to grab as many queue-able indicators as possible the task queue is likely to grow more tasks than is feasible for the queues to process.

I think another solution could be to check how large the queue is and fill the queue to X number of tasks.

snopoke:
updated

PR #17508, opened 2017-08-17
don't requeue already queued stubs

snopoke:
66K items in the DB but 1.2M in task queue....

image

### DIFF
--- a/corehq/form_processor/management/commands/run_submission_reprocessing_queue.py
+++ b/corehq/form_processor/management/commands/run_submission_reprocessing_queue.py
@@ -9,6 +9,8 @@ from couchforms.models import UnfinishedSubmissionStub
 from hqscripts.generic_queue import GenericEnqueuingOperation
 from corehq.form_processor.tasks import reprocess_submission
 
+ENQUEUING_TIMEOUT = 14 * 86400    # 14 days
+
 
 def _record_datadog_metrics():
     count = UnfinishedSubmissionStub.objects.count()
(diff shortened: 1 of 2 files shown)


corehq/form_processor/reprocess.py
```diff
@@ -48,7 +48,14 @@ def reprocess_unfinished_stub(stub, save=True):
         return
 
     assert form.is_normal
-    result = _reprocess_form(form, save)
+    try:
+        result = _reprocess_form(form, save)
+    except:
+        if save:
+            stub.queued = False
+            stub.save()
```
gcapalbo: if there's a db error here, or the task gets killed (WorkerLostError), we can end up with ones that are stuck in a queued state but not actually queued, so won't be processed.

as an alternative to this, we could just update the enqueuing timeout: https://github.com/dimagi/commcare-hq/blob/1c8b1b7ff96d62b05bf76af856d165fb030ad330/hqscripts/generic_queue.py#L71

snopoke: Isn't that basically the same as what we had before where `timeout = 24 hours`.

Just to confirm my understanding, this should be `> (max time to process a batch)`. ie. if we're enqueuing 100 times at a time then this timeout should be `> 100 * (max time to process individual item)`.

Calculating that for this queue is a bit tricky since a case with only a few forms will process quickly but a case with 1000 forms will take a long time to process.

gcapalbo: it's basically representing how often to requeue tasks that fail hard. it will always queue every item the first time it sees it, and if any of them fail hard then it won't requeue them until this amount of time has passed. so we could even set it to a week, meaning in the small chance that something fails hard we just don't retry it until a week later.

i don't think we can reliably use a `queued` flag because we can't guarantee that we can unset it in every failure scenario. a queued timestamp could work, where we always include items that haven't been queued in a long time, but that's the same thing as setting a longer `get_enqueuing_timeout` value.

to use a longer `get_enqueuing_timeout` value we would also have to remove the [:1000] limit on `get_items_to_be_processed`, but we could use a larger value for `get_fetching_interval` so we don't hit the db so often.

snopoke: @gcapalbo I've updated to use `enqueuing_timeout` but have still kept the limit on the query since the query is also filtered by the same parameters.

corehq/form_processor/management/commands/run_submission_reprocessing_queue.py
```diff
@@ -9,6 +9,8 @@
 from hqscripts.generic_queue import GenericEnqueuingOperation
 from corehq.form_processor.tasks import reprocess_submission
 
+ENQUEUING_TIMEOUT = 14 * 86400    # 14 days
```
gcapalbo: enqueuing timeout is in minutes, so should be `14 * 1440`

PR #17512, opened 2017-08-17
PGHero index feedback

emord:
Feedback from setting up stats collection on p6 on pg1. no migrations here (would likely need to do them concurrently), just looking to see your thoughts

I think the biggest change here is that we shouldn't always index domain together as we aren't gaining much. case and form ids are already unique for the domain, so it just adds size/inefficiency to the index.  (perhaps that's not as true for multi-tenant environments, but even without domain, I think index performance would still grow logarithmically). There could possibly be more indexes to remove domain from, but didn't do an audit of this file

@dimagi/scale-team

### DIFF
--- a/corehq/form_processor/models.py
+++ b/corehq/form_processor/models.py
@@ -379,7 +379,7 @@ class XFormInstanceSQL(PartitionedModel, models.Model, RedisLockableMixIn, Attac
         app_label = "form_processor"
         index_together = [
             ('domain', 'state'),
-            ('domain', 'user_id'),
+            ('domain', 'user_id'),  # could possibly remove this as user_id is ~unique. exception is web users
         ]
 
 
@@ -892,7 +892,8 @@ class CommCareCaseSQL(PartitionedModel, models.Model, RedisLockableMixIn,
 
     class Meta:
         index_together = [
-            ["domain", "owner_id", "closed"],
+            ["owner_id", "closed"],  # pghero recommended just indexing owner_id, but that's probably because there's few closed cases
+            ["owner_id", "server_modified_on"],
             ["domain", "external_id", "type"],
         ]
         app_label = "form_processor"
@@ -990,7 +991,7 @@ class CommCareCaseIndexSQL(PartitionedModel, models.Model, SaveStateMixin):
     RELATIONSHIP_MAP = {v: k for k, v in RELATIONSHIP_CHOICES}
 
     case = models.ForeignKey(
-        'CommCareCaseSQL', to_field='case_id', db_index=False,
+        'CommCareCaseSQL', to_field='case_id',
         related_name="index_set", related_query_name="index",
         on_delete=models.CASCADE,
     )
@@ -1048,7 +1049,6 @@ class CommCareCaseIndexSQL(PartitionedModel, models.Model, SaveStateMixin):
 
     class Meta:
         index_together = [
-            ["domain", "case"],
             ["domain", "referenced_id"],
         ]
         db_table = CommCareCaseIndexSQL_DB_TABLE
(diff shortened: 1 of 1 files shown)


corehq/form_processor/models.py
```diff
@@ -892,7 +892,8 @@ def __unicode__(self):
 
     class Meta:
         index_together = [
-            ["domain", "owner_id", "closed"],
+            ["owner_id", "closed"],  # pghero recommended just indexing owner_id, but that's probably because there's few closed cases
+            ["owner_id", "server_modified_on"],
```
dannyroberts: To me this means that we're doing a lot of queries by owner id without including a domain in the query. (If there were a `WHERE domain = '%s'` clause in the query, then `["domain", "owner_id", ...]` would be the right index to use.) Not including a domain in a db query to me seem like an obvious read security vulnerability—i.e. it could easily (in combination with other easy-to-make oversights) result in a simple way for some to gain access to a case in another domain. (Not saying I have an obvious one in mind, just that it seems pretty plausible.)

Thoughts?

emord: Not having domain in the query would definitely be a vulnerability, but that would be in our query code, not the index.

I think that the reason it suggested this is that you want the first part of the index to drill down as much of the rows as possible. There will always be owner_id >> domain no matter the environment which means that when searching on a query that includes both domain and owner_id, owner_id is 
 the much more useful filter. 

I think the difference is:

with [domain, owner_id, closed] the index says, get me the records that match this domain, inside that domain get me the records from this owner_id.

with [owner_id, closed] the index says get me the records that match this owner id, then i'll check those records to make sure they have that domain. Because  owner_id should always be in only one domain (unless it's a web user) then this should keep the index smaller which I believe is more efficient. 

I obviously haven't been able to try this on a real production database, so I'm not sure how much this would improve it

snopoke:
I agree that for ICDS it doesn't necessarily make a lot of sense to have domain in the indexes (though even on ICDS we to have more than one domain with data in it).

I'm quite reticent to remove the domain from the indexes unless there's a bit performance reason (like the index isn't getting used by the query planner).

corehq/form_processor/models.py
snopoke: Agree with this - I'm pretty sure the queries that would use this also include domain.

dannyroberts: Oh, awesome, right. I misinterpreted that (probably because I'm so used to couch which never _only_ ever filters with indexes), and your explanation makes total sense.

emord:
> I'm quite reticent to remove the domain from the indexes unless there's a bit performance reason (like the index isn't getting used by the query planner).

Yeah that makes sense. Another thought that I had was we could have both indexes [domain, owner_id, closed] and [owner_id, closed] and see if the first is ever used again. That would have a definitive performance hit during testing though

snopoke:
Since we'd have to make this new index before dropping the old one it makes
sense to me to have them both for a while and compare the stats (we need to
reset the stats).

corehq/form_processor/models.py
```diff
@@ -990,7 +991,7 @@ class CommCareCaseIndexSQL(PartitionedModel, models.Model, SaveStateMixin):
     RELATIONSHIP_MAP = {v: k for k, v in RELATIONSHIP_CHOICES}
 
     case = models.ForeignKey(
-        'CommCareCaseSQL', to_field='case_id', db_index=False,
+        'CommCareCaseSQL', to_field='case_id',
```
gcapalbo: it's usually a best practice to put indexes on foreign key fields. did it maybe suggest this just because it's not used heavily right now?

emord:
Closing this out. Will come back to it after the dev summit so that it's easier for me to deploy during india's off hours. There are no more egregious queries on the partitioned database after my other PR

PR #17513, opened 2017-08-17
eNikshay: CBNAAT Test Migration

sheelio:
@esoergel 

This management command will fix up the existing test cases (that were already migrated in Mumbai) with the change you made for CBNAAT tests.

### DIFF
--- a/custom/enikshay/two_b_release_1/management/commands/enikshay_2b_cbnaat_fix.py
+++ b/custom/enikshay/two_b_release_1/management/commands/enikshay_2b_cbnaat_fix.py
@@ -64,7 +64,7 @@ class Command(BaseCommand):
                 updated_by_migration = case_props.get('updated_by_migration')
                 if ((updated_by_migration == 'enikshay_2b_case_properties' or
                      updated_by_migration == 'enikshay_2b_reason_for_test_fix')
-                    and case.get_case_property('result_recorded') == 'yes'):
+                        and case.get_case_property('result_recorded') == 'yes'):
 
                     drug_resistance_list = ''
                     drug_sensitive_list = ''
--- a/custom/enikshay/two_b_release_1/management/commands/enikshay_2b_cbnaat_fix.py
+++ b/custom/enikshay/two_b_release_1/management/commands/enikshay_2b_cbnaat_fix.py
@@ -32,10 +32,12 @@ class Command(BaseCommand):
         """get last form that set result_recorded to yes"""
         for action in reversed(test.actions):
             for update in get_case_updates(action.form):
-                if update.id == test.case_id:
-                    case_properties = update.get_update_action().dynamic_properties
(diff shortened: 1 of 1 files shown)


custom/enikshay/two_b_release_1/management/commands/enikshay_2b_cbnaat_fix.py
```diff
@@ -0,0 +1,117 @@
...
+
+    @staticmethod
+    def _get_result_recorded_form(test):
+        """get last form that set result_recorded to yes"""
+        for action in reversed(test.actions):
+            for update in get_case_updates(action.form):
+                if update.id == test.case_id:
+                    case_properties = update.get_update_action().dynamic_properties
+                    if case_properties.get('result_recorded') == 'yes':
+                        return action.form.form_data
```
esoergel: This needs a modification I just added to the original: https://github.com/dimagi/commcare-hq/pull/17514/files

custom/enikshay/two_b_release_1/management/commands/enikshay_2b_cbnaat_fix.py
```diff
@@ -0,0 +1,117 @@
...
+                 .case_property_query("test_type_value", "cbnaat", "must")
+                 .scroll())
+
+        with open(log_path, "w") as f:
+            for case in cases:
+                case_props = {prop['key']: prop['value'] for prop in case['case_properties']}
+                updated_by_migration = case_props.get('updated_by_migration')
+                if ((updated_by_migration == 'enikshay_2b_case_properties' or
+                     updated_by_migration == 'enikshay_2b_reason_for_test_fix')
+                        and case.get_case_property('result_recorded') == 'yes'):
```
esoergel: I think `get_case_property` won't work here, since `case` is a result from elasticsearch, not a real case.  You should be able to do `case_props.get('result_recorded') == 'yes'`

custom/enikshay/two_b_release_1/management/commands/enikshay_2b_cbnaat_fix.py
```diff
@@ -0,0 +1,117 @@
...
+
+                    drug_resistance_list = ''
+                    drug_sensitive_list = ''
+                    resistance_display = None
+
+                    form_data = self._get_result_recorded_form(case)
```
esoergel: This also won't work with the elasticsearch case - it might be easiest to use `CaseAccessors(domain).iter_cases(case_ids)` with the `case_ids` from ES (that'll require a modification to the ES query as well).

esoergel (changes requested):
I think this needs the real cases for a couple of the operations - we can offline if needed.

sheelio:
@esoergel 
Updated

esoergel (approved):

esoergel:
This has been run on enikshay

PR #17517, opened 2017-08-17
Don't replicate on enikshay

emord:
@nickpell @gcapalbo @snopoke @mkangia @gcapalbo (should really get an enikshay team on here)

This will fix the ES cluster being yellow on enikshay. It's yellow because it is trying to replicate users, but only has one node. (this also happened on swiss and staging)

### DIFF
--- a/env_settings.py
+++ b/env_settings.py
@@ -161,4 +161,21 @@ ES_META = {
             }
         },
     },
+    'enikshay': {
+        'hqusers': {
+            "settings": {
+                "number_of_shards": 2,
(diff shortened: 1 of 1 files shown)


gcapalbo:
i also see 5 unassigned duplicated shards for domains and 5 unassigned duplicated shards for apps. are those replica shards too?

emord:
@gcapalbo That's really weird. We have only added replica shards for users. apps and domains shouldn't be replicated, unless someone did that manually

PR #17518, opened 2017-08-17
Management command to put person cases in the queue

emord:
There's an index on state_id and I don't think there's much point in doing an index on (state_id, aadhar_date).

@sheelio aadhar_date is a date field so I think we only need to check for null and not ''. Are there any other error conditions to look out for?

buddy @nickpell

### DIFF
--- a/custom/icds/management/commands/rebuild_person_cases.py
+++ b/custom/icds/management/commands/rebuild_person_cases.py
@@ -1,0 +1,48 @@
+from django.core.management.base import BaseCommand
+
+from corehq.apps.userreports.models import AsyncIndicator, get_datasource_config
+from corehq.apps.userreports.util import get_indicator_adapter
+
+
+DOMAIN = 'icds-cas'
+DATA_SOURCE_NAME = 'static-icds-cas-static-person_cases_v2'
+
+
+STATE_IDS = [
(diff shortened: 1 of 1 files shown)


sheelio:
@emord 
If its a date, sounds good to not check for ''.

custom/icds/management/commands/rebuild_person_cases.py
```diff
@@ -0,0 +1,48 @@
...
+    def handle(self, *args, **options):
+        data_source, is_static = get_datasource_config(DATA_SOURCE_NAME, DOMAIN)
+        assert is_static
+        adapter = get_indicator_adapter(data_source)
+        table = adapter.get_table()
+        for state_id in STATE_IDS:
+            null_query = adapter.session_helper.Session.query(table.columns.doc_id)
+            null_query = null_query.filter(
+                table.columns.state_id == state_id,
+                table.columns.aadhar_date == None
```
emord: Ignoring this because it's how you do sqlalchemy filters

nickpell (approved):

custom/icds/management/commands/rebuild_person_cases.py
nickpell: 👍

PR #17523, opened 2017-08-18
Update restore caching tests to be more consistent with real usage

dannyroberts:
Much of this PR is changing tests to be more in line with actual usage, but there are a few important changes that go along with the test changes, and bring more consistency to the cache keys we use to fetch and invalidate restore payload caches.

Again, I can't say this will fix https://manage.dimagi.com/default.asp?260109, but any one of the small handful of bugs/inconsistencies fixed in this PR could plausibly be responsible for that problem.

### DIFF
--- a/corehq/apps/hqcase/utils.py
+++ b/corehq/apps/hqcase/utils.py
@@ -25,7 +25,7 @@ ALLOWED_CASE_IDENTIFIER_TYPES = [
 ]
 
 
-def submit_case_blocks(case_blocks, domain, username="system", user_id="",
+def submit_case_blocks(case_blocks, domain, username="system", user_id=None,
                        xmlns=SYSTEM_FORM_XMLNS, attachments=None,
                        form_id=None, form_extras=None, case_db=None):
     """
@@ -44,7 +44,7 @@ def submit_case_blocks(case_blocks, domain, username="system", user_id="",
         'time': now,
         'uid': form_id,
         'username': username,
-        'user_id': user_id,
+        'user_id': user_id or "",
     })
     form_extras = form_extras or {}
 
--- a/corehq/apps/hqcase/utils.py
+++ b/corehq/apps/hqcase/utils.py
@@ -27,7 +27,7 @@ ALLOWED_CASE_IDENTIFIER_TYPES = [
 
 def submit_case_blocks(case_blocks, domain, username="system", user_id=None,
                        xmlns=SYSTEM_FORM_XMLNS, attachments=None,
-                       form_id=None, form_extras=None, case_db=None):
+                       form_id=None, form_extras=None, case_db=None, device_id=None):
     """
     Submits casexml in a manner similar to how they would be submitted from a phone.
 
@@ -45,6 +45,7 @@ def submit_case_blocks(case_blocks, domain, username="system", user_id=None,
         'uid': form_id,
         'username': username,
         'user_id': user_id or "",
+        'device_id': device_id,
     })
     form_extras = form_extras or {}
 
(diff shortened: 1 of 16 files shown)


emord:
I haven't looked through this with great detail, but not sure if you saw my comment here: https://github.com/dimagi/commcare-hq/pull/17501#issuecomment-323193075

I think some of the restore tests are actually dependent on using the file restore instead of the blob restore (because of this line: https://github.com/dimagi/commcare-hq/blob/master/corehq/toggles.py#L225)

I think if you're updating the restore caching tests, it may make sense to remove this toggle in this pr as well?

corehq/apps/hqcase/utils.py
```diff
@@ -44,7 +44,8 @@ def submit_case_blocks(case_blocks, domain, username="system", user_id="",
         'time': now,
         'uid': form_id,
         'username': username,
-        'user_id': user_id,
+        'user_id': user_id or "",
```
calellowitz: why do you change the default from `""` to `None`

corehq/ex-submodules/casexml/apps/phone/models.py
```diff
@@ -1127,13 +1107,6 @@ def get_latest_owner_id(case_id, action=None):
                     self.last_submitted = datetime.utcnow()
                     self.rev_before_last_submitted = self._rev
                     self.save()
-                    if case_list:
-                        try:
-                            self.invalidate_cached_payloads()
```
calellowitz: Im probably missing something simple but why do we not need to cache bust here anymore.

calellowitz (approved):
couple non blocking questions but looks good

corehq/ex-submodules/casexml/apps/phone/models.py
dannyroberts: It was cache-busting ineffectively here (i.e. deleting a different key from the one that is used to fetch the payload), and the correct key already gets deleted when you submit a form (even if it doesn't touch cases).

corehq/apps/hqcase/utils.py
dannyroberts: I changed the default from `""` to `None` so that if I call `submit_case_blocks(..., user_id=user_id, ...)` and `user_id` is `None`, the form will still be rendered with the value `""` for `user_id`.

From an interface perspective, this change also makes the function's signature "more correct". The right way to tell a function "I don't have a `user_id`" is to pass in `user_id=None`; `user_id=""` means "I have a `user_id`, and it's `""`"

dannyroberts:
offlined with @emord on his comment. I'll take over the PR he started on https://github.com/dimagi/commcare-hq/pull/17501 since it's related to the refactoring here I'm doing anyway.

corehq/ex-submodules/casexml/apps/phone/tests/test_sync_mode.py
```diff
@@ -1312,17 +1322,19 @@ def testCacheInvalidation(self):
...
             params=RestoreParams(
                 version=V2,
                 sync_log_id=self.sync_log._id,
+                device_id=device_id,
             ),
             **self.restore_options
         ).get_payload().as_string()
         self.sync_log = get_properly_wrapped_sync_log(self.sync_log._id)
-        self.assertTrue(has_cached_payload(self.sync_log, V2))
+        self.assertTrue(has_cached_payload(self.sync_log, device_id))
```
millerdev: How did we not get a merge conflict here when #17468 was merged? Looks like the changes to this test have been overwritten by that PR.

millerdev: Ah, I see, this was reverted in #17530

millerdev (commented):
I see this has since been reverted in #17530. Ignore any comments that are no longer relevant.

PR #17525, opened 2017-08-18
Added view to allow run aggregation script from UI

kkrampa:
@emord 
http://manage.dimagi.com/default.asp?260172
Could you give me a feedback to that solution?

### DIFF
--- a/custom/icds_reports/views.py
+++ b/custom/icds_reports/views.py
@@ -3,15 +3,17 @@ import requests
 from datetime import datetime, date
 
 from dateutil.relativedelta import relativedelta
+from django.contrib import messages
 from django.db.models.query_utils import Q
 from django.http.response import JsonResponse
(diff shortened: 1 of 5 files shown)


custom/icds_reports/views.py
```diff
@@ -1205,3 +1210,27 @@ def get(self, request, *args, **kwargs):
         return JsonResponse(data={
             'report_data': data,
         })
+
+
+@method_decorator(domain_admin_required, name='dispatch')
+class AggregationScriptPage(BaseDomainView):
```
emord: I think this should also use a feature flag (to be added https://github.com/dimagi/commcare-hq/blob/master/corehq/toggles.py) that way we can enable it for specific domains

emord (approved):
This looks good aside from my one comment. I'll approve this and deploy it, but could you implement the feature flag at some point? (not urgent to do)

PR #17527, opened 2017-08-18
Add owner_id, server_modifed_on index to partitioned dbs.

emord:
This adds the most needed index according to PGHero. This should help with a query that has taken 881 min of server time since I turned on stat collection. it has an average of 250 ms and was called a few hundred thousand times.

I needed to do something a little different here. I used `atomic = False` so that this index runs outside of a transaction because I want to do this concurrently.

@dimagi/scale-team

### DIFF
--- a/corehq/form_processor/migrations/0066_auto_20170818_2020.py
+++ b/corehq/form_processor/migrations/0066_auto_20170818_2020.py
@@ -1,0 +1,41 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import migrations
+
+from corehq.sql_db.operations import HqRunSQL
+
+
+TABLE_NAME = 'form_processor_commcarecasesql'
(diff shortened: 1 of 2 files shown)


gcapalbo (approved):
we just need to be aware that if the index creation fails when using `create index concurrently`, that we may need to drop and recreate the index: https://www.postgresql.org/docs/9.4/static/sql-createindex.html#SQL-CREATEINDEX-CONCURRENTLY

snopoke:
Nice, didn't know you could do non-atomic migrations.

emord:
@gcapalbo yeah. unfortunately I don't think there's an automated way to do that at the moment. I'll make a note to go through the partitions in a couple of days and make sure there aren't invalid indexes

emord:
@dimagi/scale-team Update on this. I reset the stats after adding this index. Before as noted the query took 250 ms. This query now averages 5 ms. a 50x speedup. (stats taken from partition 6 on icds)

image

gcapalbo:
nice!

snopoke:
awesome!

PR #17528, opened 2017-08-18
Fix date timestamp

emord:
@nickpell %M is minute %m is month 

@kkrampa fyi


kkrampa:
@nickpell @emord 
Thanks for catching this. Sorry for that mistake.

PR #17530, opened 2017-08-18
Revert "Update restore caching tests to be more consistent with real usage"

dannyroberts:
Reverts dimagi/commcare-hq#17523


dannyroberts:
This appears to cause a 412 in at least one new situation, so I'm going to have to do more testing locally until it's understood.

PR #17531, opened 2017-08-21
allow per-domain reindexing of cases / forms / ledgers

snopoke:
can review per commit :fish:


corehq/form_processor/backends/sql/dbaccessors.py
```diff
@@ -224,22 +225,44 @@ def get_docs(self, from_db, startkey, last_doc_pk=None, limit=500):
         """
         raise NotImplementedError
 
+    @property
+    def count_query(self):
+        """Define a custom query to use to get the doc count (only used for filtered reindex.
```
dannyroberts: nit: no closing `)` in the comment

corehq/form_processor/backends/sql/dbaccessors.py
```diff
@@ -224,22 +225,44 @@ def get_docs(self, from_db, startkey, last_doc_pk=None, limit=500):
...
         """Get the doc count from the given DB
         :param from_db: The DB alias to query
         """
-        from_db = 'default' if from_db is None else from_db
-        sql_query = "SELECT reltuples FROM pg_class WHERE oid = '{}'::regclass"
-        db_cursor = connections[from_db].cursor()
-        with db_cursor as cursor:
-            cursor.execute(sql_query.format(self.model_class._meta.db_table))
-            return int(fetchone_as_namedtuple(cursor).reltuples)
+        if not self.domain:
```
dannyroberts: I'm assuming that this section could not have been expressed as `count_query`? It looks like it takes a different approach, but it'd be a simpler process if the default and "custom" count queries could be expressed in the same form and then executed the same way. Guessing there's a roadblock that makes that not practical, though.

dannyroberts (approved):

corehq/form_processor/backends/sql/dbaccessors.py
snopoke: they could both be done that way but the non-domain way is likely to be a bit more accurate. Not super important in this case so we could just do it the one way.

PR #17546, opened 2017-08-21
bunch of cleanup with product_type

nickpell:
@emord

### DIFF
--- a/corehq/apps/accounting/models.py
+++ b/corehq/apps/accounting/models.py
@@ -2669,6 +2669,7 @@ class CreditLine(ValidateModelMixin, models.Model):
 
     @classmethod
     def get_credits_for_account(cls, account, feature_type=None, product_type=None):
+        assert not (feature_type and product_type)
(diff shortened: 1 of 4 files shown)


corehq/apps/accounting/models.py
```diff
@@ -2627,42 +2627,38 @@ def adjust_credit_balance(self, amount, is_new=False, note=None,
...
-            ):
-                yield credit_line
-
-        for credit_line in cls.get_credits_for_account(
-            line_item.invoice.subscription.account,
-            product_type=product_type,
-            feature_type=feature_type,
-        ):
-            yield credit_line
+        assert is_product or feature_type
```
emord: would it be better to soft assert this for a week or month first?

nickpell: this is more about protecting it against misuse in the future than asserting about it's current behavior (which I am pretty confident is correct) so I'm ok with the hard assert.

emord (approved):

PR #17549, opened 2017-08-22
Deprecate old testing utils

millerdev:
Follow-up to comments on #17468

@dannyroberts


dannyroberts (approved):
nice and clear! it's great that you were able to get rid of some code during this too!

PR #17551, opened 2017-08-22
handle known errors

snopoke:


corehq/form_processor/reprocess.py
```diff
@@ -47,10 +51,13 @@ def reprocess_unfinished_stub(stub, save=True):
         save and stub.delete()
         return
 
-    assert form.is_normal
-    result = _reprocess_form(form, save)
-    save and stub.delete()
-    return result
+    if form.is_deleted:
+        save and stub.delete()
```
gcapalbo: should this return here or does it not matter

gcapalbo (approved):

corehq/form_processor/reprocess.py
snopoke: it could, but doesn't really matter

PR #17552, opened 2017-08-22
use hiredis parser

snopoke:
Recommended redis parser: https://django-redis-cache.readthedocs.io/en/latest/advanced_configuration.html#pluggable-parser-classes

Also has better performance than the default.

See also https://github.com/dimagi/commcare-hq/pull/17552


dannyroberts (approved):
Looks good. Seems like the kind of thing we want to test on staging, but I assume you're on top of that!

snopoke:
Yup will do.

PR #17554, opened 2017-08-22
Restore caching cleanup

dannyroberts:
To fix https://manage.dimagi.com/default.asp?260109

This PR essentially picks up https://github.com/dimagi/commcare-hq/pull/17523 (which got reverted). I start by reverting the revert, which had a number of conflicts with some of @millerdev's recent PRs, which I fixed.

@millerdev I've addressed your comments here https://github.com/dimagi/commcare-hq/pull/17523#pullrequestreview-57380590. Notably, I took your changes to `test_sync_mode.py` over mine and then fixed a couple things in subsequent commits (particularly https://github.com/dimagi/commcare-hq/pull/17554/commits/d5c952d3f5e1856c9b256615fad2418c40c6b069).

In addition to the tests running on travis, I have been testing this using a modified version of a test that clayton wrote for this issue, which is on a not-to-be-merged branch of commcare-core, https://github.com/dimagi/commcare-core/tree/danny-local/async_collison_replication. (You can see what I had to change to get it to run against my local dev setup here https://github.com/dimagi/commcare-core/compare/ctsims/async_collison_replication...danny-local/async_collison_replication?w=1).

The commits that are actually responsible for fixing clayton's test are
- https://github.com/dimagi/commcare-hq/pull/17554/commits/99074d1153fba7b937e65bf3017c29228771488e
- https://github.com/dimagi/commcare-hq/pull/17554/commits/7601a9cb527b013cc19b14623ba33d8f6413dd63
- https://github.com/dimagi/commcare-hq/pull/17554/commits/a7897e965331278db2da869ad3209f220fbf0c08


ctsims:
Thanks, Danny! @millerdev This definitely could use a review ASAP, this is the data collision fix I was mentioning.

dannyroberts:
(cc: @nickpell new buddy)

corehq/ex-submodules/casexml/apps/phone/restore_caching.py
```diff
@@ -0,0 +1,75 @@
...
+
+    def invalidate(self):
+        logger.debug('invalidating {}'.format(self.debug_info))
+        get_redis_default_cache().delete(self.cache_key)
+
+
+class RestorePayloadPathCache(CacheAccessor):
+    def __init__(self, domain, user_id, sync_log_id, device_id):
+        self.cache_key = _restore_payload_path_cache_key(domain, user_id, sync_log_id, device_id)
+        self.debug_info = ('RestorePayloadPathCache', domain, user_id, sync_log_id, device_id)
```
millerdev: Do the values in this cache never expire? Is it possible that some key/value pairs would never be deleted and could that become a problem?

corehq/ex-submodules/casexml/apps/phone/restore_caching.py
```diff
@@ -0,0 +1,75 @@
...
+    )
+    return hashlib.md5(hashable_key).hexdigest()
+
+
+def _restore_payload_path_cache_key(domain, user_id, sync_log_id, device_id):
+    return _restore_cache_key(
+        domain=domain,
+        prefix=RESTORE_CACHE_KEY_PREFIX,
+        user_id=user_id,
+        version='2.0',
```
millerdev: what does this version mean? why is it hard-coded here rather than using the version specified in the restore config?

millerdev (approved):
Few comments, probably not blocking.

corehq/ex-submodules/casexml/apps/phone/restore_caching.py
dannyroberts: do you ask because there's no timeout set? My intention is for that to meant that it defaults back to the django cache's default, which is what it was doing before (at timeout wasn't being passed in when restore payload path was being set in the cache). I thought we had a default timeout in that case, but you might be right that that just means never expire. Anyway, seems reasonable to only cache it for a day like the other one, so I'll make that change.

corehq/ex-submodules/casexml/apps/phone/restore_caching.py
dannyroberts: It's just left over and never passed in, I'll just remove it

corehq/ex-submodules/casexml/apps/phone/restore_caching.py
```diff
@@ -0,0 +1,75 @@
...
+import logging
+from casexml.apps.phone.const import RESTORE_CACHE_KEY_PREFIX, ASYNC_RESTORE_CACHE_KEY_PREFIX
+from dimagi.utils.couch.cache.cache_core import get_redis_default_cache
+
+logger = logging.getLogger(__name__)
+
+
+def _restore_cache_key(domain, prefix, user_id, version, sync_log_id, device_id):
+    # to invalidate all restore cache keys, increment the number below
+    hashable_key = '8-{domain}-{prefix}-{user}-{version}-{sync_log_id}-{device_id}'.format(
```
dannyroberts: Yeah, good point. Honestly I picked 8 because every time I wanted to try it fresh locally i would bump it, and I ended up with 8 there.

I'll change the dashes to commas.

corehq/ex-submodules/casexml/apps/phone/restore_caching.py
```diff
@@ -24,7 +23,6 @@ def _restore_payload_path_cache_key(domain, user_id, sync_log_id, device_id):
         domain=domain,
         prefix=RESTORE_CACHE_KEY_PREFIX,
         user_id=user_id,
-        version='2.0',
```
millerdev: Hmm, that's not quite what I had in mind. It's not clear to me why we don't want the sync protocol version in the cache key. Can you help me understand your reasoning here?

dannyroberts: Yeah, so the reasoning is, it looked to me like it wasn't being done consistently anyway, so having it look like it was working isn't as good as just not having it there at all; as to the option to fix it, V1 is very, very deprecated, so there's really only one version, 2.0. I'm not too worried about the use case of someone creating a new app on their old domain from 2012 and installing that on the phone of one of their users who recently restored.

dannyroberts: (It may be time to get rid of v1 sync entirely, actually)

millerdev: I'm very pro getting rid of V1. However, I thought I saw [references](https://github.com/dimagi/commcare-hq/blob/master/corehq/ex-submodules/casexml/apps/phone/tests/test_ota_restore_v3.py) to [V3](https://github.com/dimagi/commcare-hq/blob/master/corehq/ex-submodules/casexml/apps/case/xml/__init__.py#L7) in the code. Is that not a thing or was that a false start?

dannyroberts: Yeah I saw that. I think it's a false start, but @ctsims can confirm

ctsims: V3 is highly likely to occur at some point, but there are definitely no short term plans to migrate to a new sync format.

PR #17559, opened 2017-08-22
upgrade eulxml

nickpell:
https://github.com/emory-libraries/eulxml/blob/master/CHANGELOG.rst


nickpell:
@dannyroberts do you think any QA is needed, or do travis tests provide sufficient coverage?

dannyroberts:
We really only use this to build the suite file and such, and that's pretty extensively tested. I'm pretty comfortable with that

dannyroberts (approved):

PR #17563, opened 2017-08-23
Change in Payment Register location filter

kkrampa:
* person_owner_id -> voucher_district_id


kkrampa:
@emord 
We need to rebuild voucher data source

emord:
@kkrampa I will handle the rebuild

PR #17565, opened 2017-08-23
ICDS: Aggregation script tests

kkrampa:


custom/icds_reports/migrations/0014_add_datasource_views.py
```diff
@@ -14,5 +14,4 @@ class Migration(migrations.Migration):
     ]
 
     operations = [
-        migrator.get_migration('create_datasource_views.sql'),
```
emord: Why did you need to remove these?

custom/icds_reports/migrations/sql_templates/create_datasource_views.sql
```diff
@@ -452,6 +452,59 @@ CREATE VIEW agg_awc_daily_view AS
         ("awc_location"."doc_id" = "agg_awc"."awc_id")
     );
 
+CREATE TABLE IF NOT EXISTS public."config_report_icds-cas_static-child_health_cases_a46c129f"
```
emord: Why is this needed it should be created by UCRs. Not handled in migrations

custom/icds_reports/tasks.py
```diff
@@ -14,13 +14,14 @@
 
 
 @periodic_task(run_every=crontab(minute=0, hour=21), acks_late=True, queue='background_queue')
-def move_ucr_data_into_aggregation_tables(date=None):
+def move_ucr_data_into_aggregation_tables(date=None, intervals=3):
     date = date or datetime.utcnow().date()
-    now = datetime.utcnow().date()
+    monthly_date = date.replace(day=1)
+    now = datetime.utcnow().date().replace(day=1)
```
emord: Why does `now` have `replace(day=1)` at the end. this changes the definition of these aggregations and means that `now == monthly_date` which I don't think is desired

custom/icds_reports/tests/test_aggregation_script.py
```diff
@@ -0,0 +1,484 @@
...
+
+                with engine.begin() as connection:
+                    connection.execute(table.insert().values(csv_data))
+
+    @classmethod
+    def setUpClass(cls):
+        super(AggregationScriptTest, cls).setUpClass()
+        connection = connections['default']
+        with connection.cursor() as cursor:
+            path = os.path.join(os.path.dirname(__file__), 'sql_templates', 'generate_tables.sql')
```
emord: Instead of using generate_tables.sql, these tables should be generated from the UCRs.

Look at https://github.com/dimagi/commcare-hq/blob/master/custom/icds_reports/ucr/tests/base_test.py?utf8=%E2%9C%93#L57-L67 for an example of doing that

emord (commented):
looks like tests are failing

```
ERROR: custom.icds_reports.tests.test_aggregation_script:AggregationScriptTest.test_agg_awc_2017_04_01_1
  File "/mnt/commcare-hq-ro/custom/icds_reports/tests/test_aggregation_script.py", line 232, in test_agg_awc_2017_04_01_1
    lambda x: (x['state_id'], x['district_id'], x['block_id'], x['supervisor_id'], x['awc_id'])
  File "/mnt/commcare-hq-ro/custom/icds_reports/tests/test_aggregation_script.py", line 132, in _load_and_compare_data
    list(self._load_data_from_db(table_name)),
  File "/mnt/commcare-hq-ro/custom/icds_reports/tests/test_aggregation_script.py", line 73, in _load_data_from_db
    table = metadata.tables[table_name]
KeyError: 'agg_awc_2017-04-01_1'
```

custom/icds_reports/migrations/sql_templates/create_datasource_views.sql
kkrampa: Unfortunately, this table is queried in migrations (https://github.com/dimagi/commcare-hq/pull/17565/files/13d993b82de9c42e7b6b86f48b62928d3452ecd5#diff-1af23fff6ceb126cf3ab59253ffb2b7bR542) which means it should be created before migrations (which I think is not possible)

Do you see better solution to this?

custom/icds_reports/tests/test_aggregation_script.py
kkrampa: will do

custom/icds_reports/migrations/0014_add_datasource_views.py
kkrampa: ```create_datasource_views.sql``` references ```awc_location.aggregation_level``` (https://github.com/lwyszomi/commcare-hq/blob/13d993b82de9c42e7b6b86f48b62928d3452ecd5/custom/icds_reports/migrations/sql_templates/create_datasource_views.sql#L19) which is created in migration ```0016_aggawcmonthly_aggccsrecordmonthly_aggchildhealthmonthly_aggdailyusageview_aggthrmonthly_awclocation_```
and ```num_launched_states``` (https://github.com/lwyszomi/commcare-hq/blob/13d993b82de9c42e7b6b86f48b62928d3452ecd5/custom/icds_reports/migrations/sql_templates/create_datasource_views.sql#L440) which is created ```0017_add_v2_indicators```

Without this fix it raises error that those column don't exist.

kkrampa:
@sheelio  @emord Here is the output from my local machine: https://gist.github.com/kkrampa/e78227ab4e0f83ecfc8b9f41d6a36151

There are a bunch of differences which look like data issues. Could you look at them?

requirements/requirements.txt
```diff
@@ -99,4 +99,5 @@ zeep==1.6.0
 ethiopian_date==0.1.1
 architect==0.5.6
 contextlib2==0.5.4
+sqlalchemy-postgres-copy==0.5.0
```
emord: This hsould go in test-requiremnts.txt

custom/icds_reports/migrations/sql_templates/create_datasource_views.sql
emord: Thanks for the explanation. I'm not sure hwat the best solution to this is at the moment

emord:
@kkrampa I fixed a couple of issues while looking into this https://github.com/dimagi/commcare-hq/commit/aef313b34280723c407df4257e7631f7822ba255 and https://github.com/dimagi/commcare-hq/commit/79b64cfd00dd770b7637a1dcb5d24aff58527d7b can you cherry pick those into this branch

@sheelio I think @kkrampa is correct about these being data issues. it looks like the output is expect 4 infra_cooking_utensils for the state with month <= '2017-04-01', but there are actually many more than that . Can you look into it that? it may be something different with the way you generated these csvs

kkrampa:
I made some fixes and now 11 tests fail (only agg_awc related).
https://gist.githubusercontent.com/kkrampa/28ee1a6ccea83cbeec0869c404c2e7b0/raw/7d6498728421b262755a82cbf6a45b7cc3ecf06e/output2.txt

kkrampa:
@emord 
It seems there is a problem with creating icds-ucr table.

kkrampa:
Nevermind, fixed that problem.

sheelio:
@kkrampa I'll review and update the data today.

sheelio:
@kkrampa 

I've updated the input / output data (to be smaller and fix the bugs).
[aggregation_test.zip](https://github.com/dimagi/commcare-hq/files/1254582/aggregation_test.zip)

millerdev:
Cross-posting from #tech-team on Slack:

**please do not merge** this PR. If you do it will be forever part of our git history unless we rewrite history (not a trivial thing to do).

The dataset should either live in a different repo (preferred in any case) and/or be greatly reduced in size. I don't really care about reducing the size if it lives in a separate repo.

kkrampa:
@sheelio @emord 
Still there are a bunch of issues: https://gist.github.com/kkrampa/a1f5eada917c276ca1ac0ad989335e6c

emord:
@sheelio Looks like there may still be some data issues

@kkrampa Can you rebase this branch on master and remove any commits that have the old data in them?

sheelio:
@kkrampa 

This PR looks like it moves create_functions.sql back into the migrations folder (I think we explicitly pulled that out of migrations).  Can you reverse that change? I'm looking at the data failures now.

sheelio:
@kkrampa 
I think there are some oddities in how update_monthly_aggregate_tables.sql is being called.  I would recommend:
- changing update_monthly_aggregate_tables.sql to just take a date input
- pass in the date provided to the move_ucr_data_into_aggregation_tables function to update_monthly_aggregate_tables and the current interval

kkrampa:
@sheelio Migration was broken due to moving create_functions.sql. It's referenced e.g. https://github.com/dimagi/commcare-hq/blob/master/custom/icds_reports/migrations/0002_fix_functions.py#L17

kkrampa:
@sheelio Updated
Still one my comment is open(https://github.com/dimagi/commcare-hq/pull/17565#issuecomment-325891358)

sheelio:
@kkrampa 
Ok that make sense to move it back.

sheelio:
@kkrampa 

I think the issue was the ordering of how we build the monthly tables (I think the previous task would do it new -> old).  I've updated the task (based on your changes) to do it old to new.  If you merge master, the tests should start to past.

kkrampa:
@sheelio @emord 
:white_check_mark:

emord:
@kkrampa Thanks for pushing this through.

This branch needs to be rebased, and the references to the old data set should be removed. That means that in https://github.com/dimagi/commcare-hq/pull/17565/commits/8e97ed150b2b133fde77ce9de001af011923b371 the old dataset should not be added.

kkrampa:
@emord Done

PR #17567, opened 2017-08-23
support querying read replicas for UCRs

snopoke:
buddy @kaapstorm 

This requires some localsettings changes to work but is backwards compatible with current settings.

### DIFF
--- a/corehq/apps/reports/tests/sql_fixture.py
+++ b/corehq/apps/reports/tests/sql_fixture.py
@@ -1,8 +1,8 @@
 import sqlalchemy
 from sqlalchemy import *
-from django.conf import settings
-from sqlalchemy.engine.url import make_url
 from datetime import date
+
+from corehq.sql_db.connections import connection_manager
 
 metadata = sqlalchemy.MetaData()
 
@@ -27,7 +27,7 @@ region_table = Table("region_report_data",
 
 
 def load_data():
-    engine = create_engine(make_url(settings.SQL_REPORTING_DATABASE_URL))
+    engine = connection_manager.get_engine('default')
     metadata.bind = engine
     user_table.drop(engine, checkfirst=True)
     region_table.drop(engine, checkfirst=True)
--- a/corehq/apps/userreports/sql/connection.py
+++ b/corehq/apps/userreports/sql/connection.py
@@ -5,13 +5,13 @@ from corehq.sql_db.connections import connection_manager
 READ_REPLICA_ROLLOUT_FACTOR = 0.01  # 1%
 
 
-def get_engine_id(an_object):
+def get_engine_id(an_object, allow_read_replicas=False):
     """
     Given an object, get the engine id for it.
     """
     # for now this only deals with data sources.
     from corehq.apps.userreports.models import DataSourceConfiguration
     assert isinstance(an_object, DataSourceConfiguration)
-    if random.random() < READ_REPLICA_ROLLOUT_FACTOR:
+    if allow_read_replicas and random.random() < READ_REPLICA_ROLLOUT_FACTOR:
         return connection_manager.get_read_replica_engine_id(an_object.engine_id)
     return an_object.engine_id
--- a/corehq/apps/userreports/sql/connection.py
+++ b/corehq/apps/userreports/sql/connection.py
@@ -1,8 +1,8 @@
 import random
 
-from corehq.sql_db.connections import connection_manager
+from django.conf import settings
 
-READ_REPLICA_ROLLOUT_FACTOR = 0.01  # 1%
+from corehq.sql_db.connections import connection_manager
 
 
 def get_engine_id(an_object, allow_read_replicas=False):
@@ -12,6 +12,6 @@ def get_engine_id(an_object, allow_read_replicas=False):
     # for now this only deals with data sources.
     from corehq.apps.userreports.models import DataSourceConfiguration
     assert isinstance(an_object, DataSourceConfiguration)
-    if allow_read_replicas and random.random() < READ_REPLICA_ROLLOUT_FACTOR:
-        return connection_manager.get_read_replica_engine_id(an_object.engine_id)
+    if allow_read_replicas:
+        return connection_manager.get_load_balanced_read_engine_id(an_object.engine_id)
     return an_object.engine_id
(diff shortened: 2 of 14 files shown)


corehq/apps/reports/tests/sql_fixture.py
```diff
@@ -27,7 +27,7 @@
 
 
 def load_data():
-    engine = create_engine(make_url(settings.SQL_REPORTING_DATABASE_URL))
+    engine = connection_manager.get_engine('default')
```
emord: Is `SQL_REPORTING_DATABASE_URL` always guarenteed to be the same as the default db? If so, can we deprecate that setting?

kaapstorm (approved):

corehq/sql_db/connections.py
```diff
@@ -113,23 +112,68 @@ def get_connection_string(self, engine_id):
...
+                    self._add_django_db(replica, replica)
+
+        if DEFAULT_ENGINE_ID not in self.db_connection_map:
+            self._add_django_db(DEFAULT_ENGINE_ID, 'default')
+        if UCR_ENGINE_ID not in self.db_connection_map:
+            self._add_django_db(UCR_ENGINE_ID, 'default')
+
+        self._add_django_db_from_settings_key(ICDS_UCR_ENGINE_ID, 'ICDS_UCR_DATABASE_ALIAS')
+        self._add_django_db_from_settings_key(ICDS_TEST_UCR_ENGINE_ID, 'ICDS_UCR_TEST_DATABASE_ALIAS')
         for custom_engine_id, custom_db_url in settings.CUSTOM_DATABASES:
```
emord: I think we can remove this setting

corehq/sql_db/connections.py
```diff
@@ -113,23 +112,68 @@ def get_connection_string(self, engine_id):
...
+                    assert replica not in self.db_connection_map, replica
+                    self._add_django_db(replica, replica)
+
+        if DEFAULT_ENGINE_ID not in self.db_connection_map:
+            self._add_django_db(DEFAULT_ENGINE_ID, 'default')
+        if UCR_ENGINE_ID not in self.db_connection_map:
+            self._add_django_db(UCR_ENGINE_ID, 'default')
+
+        self._add_django_db_from_settings_key(ICDS_UCR_ENGINE_ID, 'ICDS_UCR_DATABASE_ALIAS')
+        self._add_django_db_from_settings_key(ICDS_TEST_UCR_ENGINE_ID, 'ICDS_UCR_TEST_DATABASE_ALIAS')
```
emord: Should also deprecate the "test" ucr engine as now it just points to the real thing

corehq/apps/userreports/sql/connection.py
```diff
@@ -1,3 +1,8 @@
+import random
+
+from corehq.sql_db.connections import connection_manager
+
+READ_REPLICA_ROLLOUT_FACTOR = 0.01  # 1%
```
emord: It would be nice to have this as a localsetting or setting in HQ UI. that way we wouldn't need a deploy to change this and we could have different values for other envs

emord (approved):
This looks good. Made a few comments about cleaning this up. but I don't think anything is blocking

corehq/apps/reports/tests/sql_fixture.py
snopoke: There's django's `default` and then there's the reporting `default`. If `SQL_REPORTING_DATABASE_URL` isn't defined then they are the same. But if it defined then it can be different. 

But here we're getting the reporting `default` which is always the same as `SQL_REPORTING_DATABASE_URL`

corehq/sql_db/connections.py
snopoke: :+1: don't want to complicate this PR - I think it will take a bit of untangling to remove that since there are a bunch of reports that still point to it.

corehq/apps/userreports/sql/connection.py
```diff
@@ -1,10 +1,17 @@
+import random
 
+from corehq.sql_db.connections import connection_manager
 
-def get_engine_id(an_object):
+READ_REPLICA_ROLLOUT_FACTOR = 0.01  # 1%
```
snopoke: Not sure what happened to your comment "It would be nice to have this as a localsetting or setting in HQ UI. that way we wouldn't need a deploy to change this and we could have different values for other envs"

100% agree with this but didn't have the time to implement the UI. I'll move it to settings for now.

snopoke:
@emord made some changes to how the 'load balancing' is done and made it configurable in settings

corehq/sql_db/connections.py
emord: Yeah. I'm going to wait until the dev on those reports slows down before doing that to avoid conflicts

corehq/sql_db/connections.py
```diff
@@ -104,26 +116,74 @@ def dispose_all(self):
...
+
+    def _populate_from_legacy_settings(self):
+        sql_reporting_db_url = getattr(settings, 'SQL_REPORTING_DATABASE_URL', None)
```
emord: It looks like you removed this setting, and it isn't defined anywhere in ansible, so I think it can be removed.

corehq/sql_db/tests/test_connections.py
```diff
@@ -0,0 +1,86 @@
...
+            manager = ConnectionManager()
+            self.assertEqual(manager.db_connection_map, {
+                'default': 'postgresql+psycopg2://:@localhost:5432/default',
+                'ucr': 'postgresql+psycopg2://:@localhost:5432/ucr',
+                'other': 'postgresql+psycopg2://:@localhost:5432/other',
+            })
+
+            # test that load balancing works with a 15% margin for randomness
+            total_requests = 10000
+            randomness_margin = total_requests * 0.015
```
emord: I think this is 1.5% margin for randomness. not 15 as in the comment

emord (approved):
Nothing blocking from me. Will let you handle merging in case there's any ansible PRs you want to do at the same time

corehq/sql_db/connections.py
snopoke: ha, nice: 2da4e1d

emord (approved):

PR #17569, opened 2017-08-23
remove hack to support enikshay reports on softlayer

nickpell:
Confirmed with Erin these are ok to remove.

@dannyroberts @ctsims


ctsims:
I think it's fine for now, although I'd lobby for changing the feature flag to allow unrestricted access to /reports/ and the app builder url's if you are a superuser, since we use it primarily for migrations, and we had multiple occasions where the migration was hard to validate without access to that information.

PR #17572, opened 2017-08-23
Decouple change feed metrics from the pillows running.

emord:
If a pillow isn't running then these won't get reported. this moves the datadog reporting to the periodic celery queue, so that they give accurate numbers even when a pillow is down. Noticed it on icds this morning:

image

@biyeun @proteusvacuum

### DIFF
--- a/corehq/ex-submodules/pillowtop/tasks.py
+++ b/corehq/ex-submodules/pillowtop/tasks.py
@@ -2,7 +2,7 @@ from celery.schedules import crontab
 from celery.task import periodic_task
 from django.conf import settings
 
-from corehq.util.datadog.gauges import datadog_counter, datadog_gauge, datadog_histogram
+from corehq.util.datadog.gauges import datadog_gauge
 from corehq.util.decorators import serial_task
 from pillowtop.utils import get_all_pillows_json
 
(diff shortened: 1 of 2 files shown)


biyeun (approved):

corehq/ex-submodules/pillowtop/tasks.py
```diff
@@ -0,0 +1,63 @@
...
+            else:
+                assert isinstance(pillow['seq'], dict)
+                if not pillow['seq']:
+                    # this pillow has never been initialized.
+                    # (custom pillows on most environments)
+                    continue
+                assert len(pillow['offsets']) == len(pillow['seq'])
+                topic, partition = topic_name.split(',')
+                tags_with_topic = tags + ['topic:{}-{}'.format(topic, partition)]
+                processed_offset = pillow['seq'][topic_name]
```
snopoke: @emord is it useful to have these broken down by partition? I think we might get more use out of it if we combined all the partitions together. What do you think?

emord: I think it would be good to these to two separate tags for topic and partition. We get use use out of the partition because it would allow seeing any one process that is behind. (This would be less of an issue if we upgraded kafka so that we could use it to rebalance partitions to other processes)

PR #17574, opened 2017-08-23
Report unsuccessful attempts to datadog

emord:
@snopoke 

### DIFF
--- a/corehq/apps/userreports/tasks.py
+++ b/corehq/apps/userreports/tasks.py
@@ -375,6 +375,10 @@ def async_indicators_metrics():
         datadog_gauge('commcare.async_indicator.indicator_count', metrics['count'], tags=tags)
         datadog_gauge('commcare.async_indicator.lag', metrics['lag'], tags=tags)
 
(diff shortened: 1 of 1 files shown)


biyeun (approved):

corehq/apps/userreports/tasks.py
```diff
@@ -375,6 +375,10 @@ def async_indicators_metrics():
         datadog_gauge('commcare.async_indicator.indicator_count', metrics['count'], tags=tags)
         datadog_gauge('commcare.async_indicator.lag', metrics['lag'], tags=tags)
 
+    # Don't use ORM summing because it would attempt to get every value in DB
+    unsuccessful_attempts = sum(AsyncIndicator.objects.values_list('unsuccessful_attempts', flat=True).all()[:100])
```
snopoke: why only 100?

emord: Only 100 because unsuccessful_attempts is not indexed and if a record has a lot of attempts it will eventually bubble up to the top of the queue. I plan on making an alert at around 100 unsuccessful attempts in the table.

PR #17581, opened 2017-08-24
fix keyerror

sravfeyn:
https://manage.dimagi.com/default.asp?260603#1382807
@mkangia

### DIFF
--- a/corehq/apps/fixtures/upload/validation.py
+++ b/corehq/apps/fixtures/upload/validation.py
@@ -87,7 +87,7 @@ def _validate_fixture_upload(workbook):
                             _(FAILURE_MESSAGES["sheet_has_extra_property"])
                             .format(tag=tag, property=missing_property, field=field.field_name))
                     # check that fields with properties are numbered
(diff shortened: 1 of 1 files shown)


corehq/apps/fixtures/upload/validation.py
```diff
@@ -87,7 +87,7 @@ def _validate_fixture_upload(workbook):
                             _(FAILURE_MESSAGES["sheet_has_extra_property"])
                             .format(tag=tag, property=missing_property, field=field.field_name))
                     # check that fields with properties are numbered
-                    if type(data_item['field'][field.field_name]) != list:
+                    if type(data_item.get('field', {}).get(field.field_name, None)) != list:
```
mkangia: nit: Can leave the `None` in the end, get would return that anyway.

mkangia (approved):

corehq/apps/fixtures/upload/validation.py
sravfeyn: Being explicit sounded better

PR #17583, opened 2017-08-24
don't attempt to delete unsaved locations during import

snopoke:
Fixes COMMCAREHQ-4EW (https://sentry.io/dimagi/commcarehq/issues/334760767/)


kaapstorm (approved):

kaapstorm:
Do we run the risk of [saving new locations](https://github.com/dimagi/commcare-hq/blob/06aec438817921bc479029af09dcf2340b50cb36/corehq/apps/locations/bulk_management.py#L102) after the deletion is complete -- new locations that would have been deleted if they were saved first?

snopoke:
No, I was going to change that as well but it get's used in a bunch of other places as well.

PR #17584, opened 2017-08-24
don't use translated string as column slug

snopoke:
https://manage.dimagi.com/default.asp?260099


kaapstorm (commented):
Sorry for lag. That's a :white_check_mark: from me. Interesting bug!

PR #17586, opened 2017-08-24
check that django-digest is compatible with 1.10

nickpell:
for tests only


nickpell:
looks like the answer is no

PR #17587, opened 2017-08-24
add static task list data sources

nickpell:
This is so nobody modifies the task list data sources without going through version control.  We recently had a modification kick off a rebuild from scratch, which means task lists will be incomplete until the rebuild finishes.


nickpell:
@emord

emord (approved):

dannyroberts (approved):

PR #17588, opened 2017-08-24
Start running enikshay_task again

gcapalbo:
@nickpell @esoergel @sravfeyn do any of you have context on this task? based on my notes from @proteusvacuum we can enable this on the new cluster as soon as the static-enikshay-adherence data source is rebuilt (which it now is), so i think this is good to start running again.

fyi @mkangia

### DIFF
--- a/custom/enikshay/tasks.py
+++ b/custom/enikshay/tasks.py
@@ -53,6 +53,7 @@ DoseStatus = namedtuple('DoseStatus', 'taken missed unknown source')
 CACHE_KEY = "enikshay-task-id-{}".format(datetime.date.today())
 cache = get_redis_client()
 
+
 @periodic_task(
     bind=True,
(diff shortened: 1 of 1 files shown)


custom/enikshay/tasks.py
```diff
@@ -67,6 +67,10 @@ def enikshay_task(self):
 
     domains = toggles.UATBC_ADHERENCE_TASK.get_enabled_domains()
     for domain in domains:
+        if toggles.DATA_MIGRATION.enabled(domain):
+            # Don't run this on the india cluster anymore
+            continue
+
```
mkangia: Cant we just kill the celery queue on softlayer?

gcapalbo: we should do that too but the messages would still queue up in rabbitmq if that's all we did

gcapalbo:
(offlined with @nickpell and decided it's ok to run)

PR #17590, opened 2017-08-24
Remove dependency on es to retrieve case ids

gcapalbo:
Uses postgres to get the case ids instead of es. Can review with `w=1`

@orangejenny

### DIFF
--- a/corehq/apps/data_interfaces/models.py
+++ b/corehq/apps/data_interfaces/models.py
@@ -19,6 +19,7 @@ from corehq.messaging.scheduling.const import (
     VISIT_WINDOW_END,
     VISIT_WINDOW_DUE_DATE,
 )
+from corehq.form_processor.utils.general import should_use_sql_backend
 from corehq.messaging.scheduling.models import AlertSchedule, TimedSchedule
 from corehq.messaging.scheduling.tasks import (
     refresh_case_alert_schedule_instances,
@@ -209,6 +210,13 @@ class AutomaticUpdateRule(models.Model):
 
     @classmethod
     def get_case_ids(cls, domain, case_type, boundary_date=None):
+        if should_use_sql_backend(domain):
+            return cls._get_case_ids_from_postgres(domain, case_type, boundary_date=boundary_date)
+        else:
+            return cls._get_case_ids_from_es(domain, case_type, boundary_date=boundary_date)
+
+    @classmethod
+    def _get_case_ids_from_postgres(cls, domain, case_type, boundary_date=None):
         q_expression = Q(
             domain=domain,
             type=case_type,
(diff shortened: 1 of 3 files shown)


corehq/apps/data_interfaces/models.py
```diff
@@ -206,35 +209,17 @@ def get_boundary_date(cls, rules, now):
...
-        """
-        chunk_size = 100
-
-        query = (CaseES()
-                 .domain(domain)
-                 .case_type(case_type)
-                 .is_closed(closed=False)
-                 .exclude_source()
-                 .size(chunk_size))
+        q_expression = Q(
```
emord: Does this code only run for SQL domains?

gcapalbo: good catch, thanks https://github.com/dimagi/commcare-hq/pull/17590/commits/5dc6eab27ddbfc14670a779cc3a7bc5e01d5e9bf

corehq/apps/data_interfaces/models.py
```diff
@@ -206,35 +210,42 @@ def get_boundary_date(cls, rules, now):
...
+            domain=domain,
+            type=case_type,
+            closed=False,
+            deleted=False,
+        )
+
+        if boundary_date:
+            q_expression = q_expression & Q(server_modified_on__lte=boundary_date)
+
+        return run_query_across_partitioned_databases(CommCareCaseSQL, q_expression, values=['case_id'])
```
emord: does this return a generator?

corehq/apps/data_interfaces/tasks.py
```diff
@@ -130,25 +130,24 @@ def run_case_update_rules_for_domain(domain, now=None):
...
-                    case_update_result.total_updates >= settings.MAX_RULE_UPDATES_IN_ONE_RUN or
-                    migration_in_progress
-                ):
-                    run_record.done(DomainCaseRuleRun.STATUS_HALTED, cases_checked, case_update_result)
-                    notify_error("Halting rule run for domain %s." % domain)
-                    return
-
-                case_update_result.add_result(run_rules_for_case(case, rules, now))
-                cases_checked += 1
+        case_ids = list(AutomaticUpdateRule.get_case_ids(domain, case_type, boundary_date))
```
emord: Looks like this will load much more into memory than before. how many case ids are expected here?

emord (approved):
looks good to me if you don't think the loss of chunking is an issue

corehq/apps/data_interfaces/models.py
gcapalbo: it does. but for now it ends up converting to a list before processing

corehq/apps/data_interfaces/tasks.py
gcapalbo: it depends on the rule, but probably less than a million, which i've been able to do in management commands before

i think the chunked approach broke down because it required a connection to stay open for the entire time the processing was happening. i'm going to give this a try and see if it's any better. eventually we're probably going to have to split this process by db

gcapalbo:
(last commit doesn't affect tests, merging)

PR #17591, opened 2017-08-24
Update Copy on 500 Page. Catch ES Error on Web Users Page

biyeun:
@proteusvacuum @emord 

response to http://manage.dimagi.com/default.asp?260064#1383081

screen shot 2017-08-24 at 2 50 34 pm

### DIFF
--- a/corehq/apps/users/views/__init__.py
+++ b/corehq/apps/users/views/__init__.py
@@ -469,7 +469,6 @@ class ListWebUsersView(HQJSONResponseMixin, BaseUserSettingsView):
                 not self.can_restrict_access_by_location
                 and not role.permissions.access_all_locations
             )
-        raise Exception()
         if show_es_issue:
             messages.error(
                 self.request,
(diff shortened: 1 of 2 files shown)


corehq/apps/users/views/__init__.py
```diff
@@ -454,13 +457,30 @@ def user_roles(self):
...
+            try:
+                role.hasUsersAssigned = bool(role.ids_of_assigned_users)
+            except TypeError:
+                # when query_result['hits'] returns None due to an ES issue
+                show_es_issue = True
             role.has_unpermitted_location_restriction = (
                 not self.can_restrict_access_by_location
                 and not role.permissions.access_all_locations
             )
+        raise Exception()
```
emord: This will unconditionally raise an exception

biyeun: whoops. that was supposed to be removed. was for testing. thanks

emord (approved):

PR #17596, opened 2017-08-25
location safety on excel reports

sravfeyn:
http://manage.dimagi.com/default.asp?260587#1383161

@mkangia @esoergel

### DIFF
--- a/corehq/apps/reports/views.py
+++ b/corehq/apps/reports/views.py
@@ -9,6 +9,8 @@ from corehq.apps.app_manager.suite_xml.sections.entries import EntriesHelper
 from corehq.apps.domain.utils import get_domain_module_map
 from corehq.apps.domain.views import BaseDomainView
 from corehq.apps.hqwebapp.view_permissions import user_can_view_reports
+from corehq.apps.locations.permissions import conditionally_location_safe, \
+    report_class_is_location_safe
 from corehq.apps.reports.display import xmlns_to_name
 from corehq.apps.users.permissions import FORM_EXPORT_PERMISSION, CASE_EXPORT_PERMISSION, \
     DEID_EXPORT_PERMISSION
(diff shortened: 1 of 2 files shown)


esoergel:
@sravfeyn is my understanding correct?  This will verify that the report class is accessible, but it won't re-verify the report, right?  If user A generated the report, then sent a link to user B, user B would be able to export user A's version of the report, even if their permissions were different.  Is this a URL that's sent in email or just something used internally through javascript (like, you click the "download" button after it's finished generating, and that makes this request).  I think this is probably fine, and checking the locations again would require passing those as context to redis, which might not be straightforward.

@sheelio thoughts on this?

sheelio:
@esoergel 

Its a bit of a weird situation, but I would think that the act of a user emailing a report would essentially "approve" the recipient for its contents (so even if the recipient didn't have permissions, the sender provided them with permission).

sravfeyn:
Ethan, yeah, you are right. I thought about the same issue a bit, but this felt like still acceptable.

corehq/apps/reports/views.py
```diff
@@ -2008,6 +2010,16 @@ def mk_date_range(start=None, end=None, ago=timedelta(days=7), iso=False):
...
+def _is_location_safe_report_class(request, domain, export_hash, format):
+    cache = get_redis_client()
+
+    content = cache.get(export_hash)
+    if content is not None:
+        report_class, report_file = content
+        return report_class_is_location_safe(report_class)
+
+
+@conditionally_location_safe(_is_location_safe_report_class)
```
mkangia: Probably does not make a difference, just the decorator order would look better if this comes after all conditions are satisfied
```
@login_and_domain_required
@require_GET
@conditionally_location_safe(_is_location_safe_report_class)
```

Unless there is a reason for putting it at the top instead.

mkangia (approved):

esoergel (approved):

PR #17604, opened 2017-08-25
Kill app manager v1 NEW_APP tour

orangejenny:
@biyeun This is the only guided tour. Do you think we'll be using this guided tour infrastructure in the future?

code buddy @gcapalbo


biyeun:
@orangejenny I think it's time to see if there are any other options out there of if we want to build something up ourselves.

biyeun (approved):

gcapalbo (approved):

PR #17610, opened 2017-08-25
Switch from assert to soft assert

emord:
https://sentry.io/dimagi/commcarehq/issues/335214165/

a couple of old, no longer running custom pillows messed this up.

@snopoke @biyeun


emord:
introduced here https://github.com/dimagi/commcare-hq/pull/17572

PR #17618, opened 2017-08-28
MockDevice.id

millerdev:
Add `device_id` to forms posted by `MockDevice` as well as `MockDevice` restore requests.

Tangentially related: also added a test for a bug in `clean_owners` restore. This was only tangentially related because I wanted to verify that the bug still existed and had the same failure modes after adding `device_id` to `MockDevice` operations, which it does.

Best reviewed by commit 🐠 

@dannyroberts cc @ctsims

### DIFF
--- a/corehq/apps/hqcase/templates/hqcase/xml/case_block.xml
+++ b/corehq/apps/hqcase/templates/hqcase/xml/case_block.xml
@@ -1,7 +1,7 @@
 <?xml version='1.0' ?>
 <system version="1" uiVersion="1" xmlns="{{ xmlns }}" xmlns:orx="http://openrosa.org/jr/xforms">
     <orx:meta xmlns:cc="http://commcarehq.org/xforms">
-        {% if not device_id %}<orx:deviceID />{% else %}<orx:deviceID>{{ device_id }}</orx:deviceID>{% endif %}
+        <orx:deviceID>{{ device_id }}</orx:deviceID>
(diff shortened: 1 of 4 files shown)


corehq/apps/hqcase/templates/hqcase/xml/case_block.xml
```diff
@@ -1,7 +1,7 @@
 <?xml version='1.0' ?>
 <system version="1" uiVersion="1" xmlns="{{ xmlns }}" xmlns:orx="http://openrosa.org/jr/xforms">
     <orx:meta xmlns:cc="http://commcarehq.org/xforms">
-        {% if not device_id %}<orx:deviceID />{% else %}<orx:deviceID>{{ device_id }}</orx:deviceID>{% endif %}
+        <orx:deviceID>{{ device_id }}</orx:deviceID>
```
millerdev: @dannyroberts I don't think there's any semantic difference between `<element></element>` and `<element />` in XML, so making this like all the rest. Maybe you had a different reason for doing this one differently?

dannyroberts: Nope, I'm cool with this. I was probably just very literally going off an example that had `<orx:deviceID />` in it, and didn't think harder about it. Agree this is more straightforward, and just as good.

dannyroberts (approved):

PR #17620, opened 2017-08-28
fall back to default language in case translation missing

mkangia:
https://manage.dimagi.com/default.asp?260866

### DIFF
--- a/corehq/apps/app_manager/models.py
+++ b/corehq/apps/app_manager/models.py
@@ -1457,7 +1457,8 @@ class NavMenuItemMediaMixin(DocumentSchema):
     def custom_icon_form_and_text_by_language(self, lang):
         custom_icon = self.custom_icon
         if custom_icon:
-            return custom_icon.form, custom_icon.text[lang]
+            custom_icon_text = custom_icon.text.get(lang) or custom_icon.text.get(self.get_app().default_language)
+            return custom_icon.form, custom_icon_text
         return None, None
 
     def _set_media(self, media_attr, lang, media_path):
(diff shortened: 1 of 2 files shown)


nickpell (approved):

corehq/apps/app_manager/models.py
```diff
@@ -1457,7 +1457,8 @@ def audio_by_language(self, lang, strict=False):
     def custom_icon_form_and_text_by_language(self, lang):
         custom_icon = self.custom_icon
         if custom_icon:
-            return custom_icon.form, custom_icon.text[lang]
+            custom_icon_text = custom_icon.text.get(lang) or custom_icon.text.get(self.get_app().default_language)
```
orangejenny: could be more concise: `custom_icon.text.get(lang, self.get_app().default_language)`

orangejenny (approved):

corehq/apps/app_manager/models.py
esoergel: Is `""` valid icon text?  If so, this could be potentially confusing, as the empty string is falsy.

esoergel (approved):

corehq/apps/app_manager/models.py
esoergel: I think you'd want this:
```python
custom_icon.text.get(lang, custom_icon.text.get(self.get_app().default_language))
```

orangejenny: oops, i was sloppy - thanks Ethan :)

mkangia: Actually did the opposite to reduce nesting.

mkangia: looks like the change `custom_icon.text.get(lang, custom_icon.text.get(self.get_app().default_language))` would take care of it. Will make that change.

PR #17623, opened 2017-08-28
make more columns nullable

nickpell:


nickpell:

emord (approved):
Are these the right way to solve these? Wondering if the cases that have null ids should be filtered out?

nickpell:
Right now I think the answer is yes.  I don't think I was particularly thoughtful when assigning ```"is_nullable"``` in the first place and in general we should support it.

PR #17626, opened 2017-08-29
Changes in episode 2b data source

kkrampa:
* test_type_choice -> test_type_value
* Added clause to result of dst


kkrampa:
@emord require episode_2b rebuild

PR #17630, opened 2017-08-29
Don't allow two aggregations to run at once

emord:
@lwyszomi @kkrampa @sravfeyn  I believe this will fix the issue with the aggregation not working correctly for QA.

When I looked through the logs it looked like multiple aggregation tasks were kicked off at one time. Using serial task will set a lock to make sure only one runs at once.


emord:
✔️

PR #17633, opened 2017-08-29
use age according to nikshay api requirements

mkangia:
buddy: @sravfeyn

### DIFF
--- a/custom/enikshay/integrations/nikshay/repeater_generator.py
+++ b/custom/enikshay/integrations/nikshay/repeater_generator.py
@@ -482,6 +482,16 @@ def _get_nikshay_id_from_response(response):
         raise NikshayResponseException("No Nikshay ID received: {}".format(response_json))
 
 
+def _get_person_age(person_case_properties):
+    # Nikshay excepts only integer value so pick floor value or 1 in case age less than 1
+    # 2.3 => 2, 0.5 => 1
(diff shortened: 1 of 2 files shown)


custom/enikshay/integrations/nikshay/repeater_generator.py
```diff
@@ -482,6 +482,16 @@ def _get_nikshay_id_from_response(response):
...
 
 
+def _get_person_age(person_case_properties):
+    # Nikshay excepts only integer value so pick floor value or 1 in case age less than 1
+    # 2.3 => 2, 0.5 => 1
+    person_age = person_case_properties.get('age', '') or person_case_properties.get('age_entered', '')
+    person_age = person_age.split('.')[0]
+    if person_age == '0':
+        person_age = '1'
+    return person_age
```
sravfeyn: Just noting. Not sure if it's acceptable or not for the call, but this could end up returning an empty string

sravfeyn (approved):

custom/enikshay/integrations/nikshay/repeater_generator.py
mkangia: @sravfeyn in case if there is no age or age_entered? That would be a edgy case but as far as i know one of these two would be definitely present.

proteusvacuum (approved):

PR #17634, opened 2017-08-29
Ignore icds-ucr-standby

emord:
https://github.com/dimagi/commcarehq-ansible/pull/899/files#diff-2a475a69e9f47422e69d6e420532df87R146

https://sentry.io/dimagi/commcarehq/issues/338428936/?referrer=slack

@snopoke This change caused icds to try to run the migrations on icds-ucr-standby, and it failed. I deployed using this change to unblock deploy, but it will fail on the next deploy.

I think the proper fix to this will be adding some flag to the databases in local settings, but is it actually necessary to have the standby in the databases?

### DIFF
--- a/corehq/sql_db/management/commands/migrate_multi.py
+++ b/corehq/sql_db/management/commands/migrate_multi.py
@@ -32,6 +32,8 @@ class Command(BaseCommand):
         if migration_name is not None:
             args.append(migration_name)
         for db_alias in settings.DATABASES.keys():
+            if db_alias == 'icds-ucr-standby':
+                continue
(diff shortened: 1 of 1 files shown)


snopoke:
It should be pretty easy to modify the db router to exclude the standbys.

They don't have to be in django settings but it is nice to only have one
place that databases are defined.

snopoke:
I'll make the change to the router tomorrow.

snopoke:
Fixed by https://github.com/dimagi/commcare-hq/pull/17643

Turns out that even with a router that disables migrations Django still tries to write the `django_migrations` table. So I had to remove the read-only DBs from Django settings.

PR #17636, opened 2017-08-29
Allow form filtering for child modules where the parent selects a case

orangejenny:
https://manage.dimagi.com/default.asp?260592

@sravfeyn / @snopoke


emord (approved):

orangejenny:
updated to fix tests

emord (approved):

PR #17637, opened 2017-08-29
recalculate stagnant cases

emord:
@sheelio @gcapalbo As discussed in the email thread.

I'll need to add indexes to inserted_at concurrently before this is deployed, but otherwise I think this is what we discussed.

### DIFF
--- a/custom/icds_reports/tasks.py
+++ b/custom/icds_reports/tasks.py
@@ -1,4 +1,4 @@
-from datetime import datetime
+from datetime import datetime, timedelta
 import logging
 import os
 
@@ -8,6 +8,11 @@ from dateutil.relativedelta import relativedelta
 from django.conf import settings
 from django.db import connections
 
+from corehq.apps.userreports.models import get_datasource_config
+from corehq.apps.userreports.util import get_indicator_adapter
+from corehq.form_processor.interfaces.dbaccessors import CaseAccessors
+from corehq.form_processor.change_publishers import publish_case_saved
+from corehq.form_processor.exceptions import CaseNotFound
 from corehq.util.decorators import serial_task
 
 celery_task_logger = logging.getLogger('celery.task')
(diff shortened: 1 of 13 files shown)


custom/icds_reports/tasks.py
```diff
@@ -58,3 +64,52 @@ def move_ucr_data_into_aggregation_tables(date=None):
                 sql_to_execute = sql_file.read()
                 cursor.execute(sql_to_execute, {"date": date.strftime('%Y-%m-%d')})
             celery_task_logger.info("Ended icds reports update_daily_aggregate_table")
+
+
+@periodic_task(run_every=crontab(minute=0, hour=21), acks_late=True, queue='background_queue')
+def recalculate_stagnant_cases():
+    domain = 'icds-cas'
+    config_ids = [
+        'static-icds-cas-static-ccs_record_cases_monthly',
```
emord: @sheelio you originally said we only need to look at 4 UCRs, but I think we need 6 because of the v1/v2 for child cases and ccs records right?

sheelio: Yup that's right (though at some point we should deprecate the v1 data sources)

sheelio:
@emord 

Does this filter out closed cases (i.e. not trigger them again?)

emord:
@sheelio If the case is closed then it should not be in the UCR table right? So that means that there shouldn't be any closed cases. If there are then this just sends the case back through kafka which will get processed by the pillow

custom/icds_reports/tasks.py
```diff
@@ -58,3 +64,52 @@ def move_ucr_data_into_aggregation_tables(date=None):
...
+
+
+def _find_stagnant_cases(adapter):
+    stagnant_date = datetime.utcnow() - timedelta(days=45)
+    table = adapter.get_table()
+    query = adapter.get_query_object()
+    query = query.with_entities(table.columns.doc_id).filter(
+        table.columns.inserted_at <= stagnant_date
+    ).distinct()
+    return query.all()
```
gcapalbo: based on @sheelio's auto update rules that he had setup, he was using the case's server_modified_on timestamp to select cases that hadn't been modified in 45 days. it looks like this is selecting cases who had records inserted into the ucr table more than 45 days ago. @sheelio does that still fulfill your use case?

emord: This was intentional by me. If we go by server_modified_on we'll need to either change that when queueuing (which seems weird to me as it's comment says its the date of the last transaction) or recalculate all unupdated cases every night (could be a lot of cases)

sheelio:
Should be fine to rely on inserted_at (I was using server modified on to
approximate updating the cases every 45 days since server modified on is
approximately equal to inserted_at).

gcapalbo (approved):

emord:
adding the indexes now. once I verify they have been completed I will merge and deploy this. the task should get run this weekend

emord:
Verified teh indexes were created. Going to merge. the commits are after approval because I rebased to get rid of a conflict

snopoke:
What problem is this solving?

emord:
@snopoke the monthly UCRs currently calculate rows for 7 months (current, 3 before, 3 after). I believe that if a case hasn't been updated then it can fall off of some reports, so this will trigger a rebuild of UCRs for any row that has been inserted 45 days ago

sheelio:
We are also planning to reduce the iterations (so we'd have to run the UCR
for 1 month before, the current month and 1 month after).  This would
require 45 days instead of something larger.

snopoke:
Thanks for the explanation.

PR #17638, opened 2017-08-29
Pulling from OpenMRS

kaapstorm:
This change pulls from OpenMRS using its Reporting API.

At the moment this imports from a single OpenMRS instance per domain. To do next: Support more than one, and associate each with a location.

Labelling "Hold off" pending a bit more testing locally.

@dannyroberts buddy @snopoke


corehq/motech/openmrs/dbaccessors.py
```diff
@@ -0,0 +1,13 @@
...
+
+@quickcache(['domain_name'])
+def get_openmrs_importers_by_domain(domain_name):
+    results = OpenmrsImporter.get_db().view(
+        'by_domain_doc_type_date/view',
+        key=[domain_name, 'OpenmrsImporter', None],
+        include_docs=True,
+        reduce=False,
+    ).all()
+    return [OpenmrsImporter.wrap(result['doc']) for result in results]
```
snopoke: If you use `OpenmrsImporter.view(...)` (without the `get_db()`) then the wrapping is done for you

corehq/motech/openmrs/forms.py
```diff
@@ -13,3 +22,69 @@ def __init__(self, *args, **kwargs):
...
+            importer = importers[0] if importers else None  # TODO: Support multiple
+            if importer is None:
+                importer = OpenmrsImporter(domain=domain_name)
+            importer.server_url = self.cleaned_data['server_url']
+            importer.username = self.cleaned_data['username']
+            if self.cleaned_data['password']:
+                # Don't save it if it hasn't been changed. Use simple symmetric encryption. We don't need it to be
+                # strong, considering we'd have to store the algorithm and the key together anyway; it just
+                # shouldn't be plaintext.
+                importer.password = b64encode(bz2.compress(self.cleaned_data['password']))
```
snopoke: It should be possible to use the Django secret key or else the HQ RSA public key to make this more secure. 
@dannyroberts any thoughts on this?

corehq/motech/openmrs/views.py
```diff
@@ -112,3 +117,38 @@ def openmrs_test_fire(request, domain, repeater_id, record_id):
...
+            get_openmrs_importers_by_domain.clear(request.domain)
+            return HttpResponseRedirect(self.page_url)
+        context = self.get_context_data(**kwargs)
+        return self.render_to_response(context)
+
+    @method_decorator(domain_admin_required)
+    def dispatch(self, request, *args, **kwargs):
+        if not toggles.OPENMRS_INTEGRATION.enabled(request.domain):
+            raise Http404()
+        return super(OpenmrsImporterView, self).dispatch(request, *args, **kwargs)
```
snopoke: You can replace this with class level decorators:
```
@method_decorator(domain_admin_required, name='dispatch')
@method_decorator(toggles.OPENMRS_INTEGRATION.required_decorator(), name='dispatch')
class OpenmrsImporterView(BaseAdminProjectSettingsView):
    ....
```

corehq/motech/openmrs/views.py
```diff
@@ -112,3 +117,38 @@ def openmrs_test_fire(request, domain, repeater_id, record_id):
...
     attempt = repeater.fire_for_record(record)
     return JsonResponse(attempt.to_json())
+
+
+class OpenmrsImporterView(BaseAdminProjectSettingsView):
+    urlname = 'openmrs_importer_view'
+    page_title = ugettext_lazy("OpenMRS Importers")
+    template_name = 'openmrs/importers.html'
+
+    @method_decorator(domain_admin_required)
```
snopoke: this shouldn't be needed

corehq/motech/openmrs/forms.py
kaapstorm: (For a little context, I used the same approach to store DHIS2 and OpenClinica passwords, but repeaters store passwords in plaintext (including the OpenMRS repeater). I guess if unfriendly people are reading our database we have more things to worry about than the OpenMRS password, but I do like the idea of using the Django secret key -- it seems a little safer, and still simple.)

corehq/motech/openmrs/openmrs_config.py
```diff
@@ -41,6 +49,21 @@ def get_value(self, case_trigger_info):
         return self.value
 
 
+class CasePropertyConcept(CaseProperty):
```
dannyroberts: Is this class a separate feature, or is it added for this import-from-openmrs feature? (Fine if this PR contains a couple different things, just trying to understand.)

dannyroberts (commented):
Approach looks good to me!

corehq/motech/openmrs/openmrs_config.py
kaapstorm: You're right @dannyroberts this is separate. I think I'll move it to a separate PR, not just to make this less confusing, but also because it would be useful to have it deployed sooner.

dannyroberts: Ah, okay cool. Makes total sense. Thanks!

dannyroberts:
Saw some work on this a few days ago. Still more work to go or is it ready for another review and merge?

kaapstorm:
Thanks for the ping @dannyroberts. I fixed some bugs, and rebased to clean up the commits a bit. This has been tested on staging for a few days, and I think what's here is good to go.

There is one addition in the near future: An "Import Now" button on the config page that kicks off the task to import patients. I was going to include it in this PR, but I'll PR it separately.

kaapstorm:
:white_check_mark: Fixed bug with len(AES_key) > 32

dannyroberts (approved):

PR #17639, opened 2017-08-29
improve handling of server errors

dannyroberts:
generally, say something went wrong and stay in the same place
so the user can retry later.

This is in preparation for drastically lowering the lock timeout in formplayer: https://github.com/dimagi/formplayer/pull/461.

Here's what it looks like when it errors during answering a question in the form, a case I payed special attention to:

screen shot 2017-08-29 at 5 59 10 pm


snopoke (approved):

snopoke:
Build failure real - looks like a test that needs updating:
```
  1) WebForm WebFormSession Should not block requests:
     AssertionError: expected false to equal undefined
      at http://localhost:8000/static/chai/chai.js:210
      at assertEqual (http://localhost:8000/static/chai/chai.js:752)
      at http://localhost:8000/static/chai/chai.js:3854
      at http://localhost:8000/static/chai/chai.js:2507
      at http://localhost:8000/static/cloudcare/js/form_entry/spec/webformsession_spec.js:160
```

PR #17640, opened 2017-08-29
Tag system initiated case transacitons

millerdev:
Update docstrings and usages of internal case transaction utility functions, making it easier to trace the source of case operations. I chose to use `device_id` as the primary means of providing source information rather than `xmlns`. The default `xmlns` (`SYSTEM_FORM_XMLNS`) tags the operation as a system initiated transaction, and `device_id` is intended to provide the name of the subsystem in HQ from whence it was initiated. The current convention is to pass the dotted function/method/class/module path as `device_id`.

Best reviewed by commit 🐠 

@proteusvacuum @dannyroberts cc @snopoke @ctsims

### DIFF
--- a/corehq/apps/commtrack/util.py
+++ b/corehq/apps/commtrack/util.py
@@ -174,10 +174,9 @@ def submit_mapping_case_block(user, index):
         )
 
     submit_case_blocks(
-        ElementTree.tostring(
-            caseblock.as_xml()
-        ),
+        ElementTree.tostring(caseblock.as_xml()),
         user.domain,
+        device_id=__name__ + ".submit_mapping_case_block"
     )
 
 
--- a/corehq/form_processor/submission_post.py
+++ b/corehq/form_processor/submission_post.py
@@ -192,7 +192,9 @@ class SubmissionPost(object):
                     else:
                         instance.initial_processing_complete = True
                         self.save_processed_models(xforms, case_stock_result)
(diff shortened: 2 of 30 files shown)


corehq/apps/commtrack/util.py
```diff
@@ -174,10 +174,9 @@ def submit_mapping_case_block(user, index):
         )
 
     submit_case_blocks(
-        ElementTree.tostring(
-            caseblock.as_xml()
-        ),
+        ElementTree.tostring(caseblock.as_xml()),
```
snopoke: Not your code I know but this is the same as `caseblock.as_string()`. A few other places too.

corehq/form_processor/submission_post.py
```diff
@@ -192,7 +192,9 @@ def run(self):
                     else:
                         instance.initial_processing_complete = True
                         self.save_processed_models(xforms, case_stock_result)
-                        case_stock_result.case_result.close_extensions(case_db)
+                        device_id = instance.metadata.deviceID if instance.metadata else ""
+                        case_stock_result.case_result.close_extensions(case_db,
+                            "SubmissionPost-close_extensions-%s" % device_id)
```
snopoke: I'm not suggesting you add this to the device ID but it would be great to know in the form which triggered this when looking at the case close form.

snopoke (approved):

PR #17643, opened 2017-08-30
allow non-django databases for reporting DBs

snopoke:
buddy @kaapstorm

### DIFF
--- a/corehq/sql_db/connections.py
+++ b/corehq/sql_db/connections.py
@@ -50,6 +50,12 @@ class SessionHelper(object):
         return session_scope
 
 
+def _validate_db_alias_or_connection_string(db_alias_or_connection_string):
+    if db_alias_or_connection_string not in settings.DATABASES:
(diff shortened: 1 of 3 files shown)


kaapstorm (approved):

emord (approved):
Looks like a reasonable change to me. Is there any reason that you didn't just add a readonly setting to the entry in settings.DATABASES and skip calling migrate if that was set here: https://github.com/dimagi/commcare-hq/blob/je/more-sms-ui/corehq/sql_db/management/commands/migrate_multi.py#L34-L42

snopoke:
@emord I did have that initially but it didn't seem 'safe' enough

snopoke:
Since this only affects ICDS I'll update the settings and deploy tomorrow

snopoke:
Closing in favour of https://github.com/dimagi/commcare-hq/pull/17656

PR #17650, opened 2017-08-30
App manager menu: allow dragging forms into empty menus

orangejenny:
Part of https://manage.dimagi.com/default.asp?259749

This looks like something we've deliberately not allowed ~ever, but I agree with Jen that it should be supported. Also removed some code that I'm fairly certain isn't relevant anymore. Sanity tested locally.

@biyeun 

code buddy @gcapalbo


biyeun:
yikes...this is going to cause a nasty conflict with the PR I have on staging. lemme digest this for a sec so i know what to do...

PR #17654, opened 2017-08-30
Add file extension when downloading CommCareMultimedia

orangejenny:
https://manage.dimagi.com/default.asp?260475

@biyeun / @millerdev know of any better way to do this? So far as I can tell, the filename is stored outside of `CommCareMultimedia`, someplace like `app.multimedia_map` so there isn't a clean, generic way to go from multimedia back to filename.

code buddy @gcapalbo

### DIFF
--- a/corehq/apps/hqmedia/views.py
+++ b/corehq/apps/hqmedia/views.py
@@ -652,7 +652,12 @@ class ViewMultimediaFile(View):
         data, content_type = self.multimedia.get_display_file()
         if self.thumb:
             data = CommCareImage.get_thumbnail_data(data, self.thumb)
-        return HttpResponse(data, content_type=content_type)
+        extension = ''
+        if self.multimedia.aux_media:
+            extension = self.multimedia.aux_media[-1]['uploaded_filename'].split(".")[-1]
+        response = HttpResponse(data, content_type=content_type)
+        response['Content-Disposition'] = 'filename="download.{}"'.format(extension)
+        return response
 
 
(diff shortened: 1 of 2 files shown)


gcapalbo (approved):

corehq/apps/hqmedia/views.py
```diff
@@ -652,7 +652,12 @@ def get(self, request, *args, **kwargs):
         data, content_type = self.multimedia.get_display_file()
         if self.thumb:
             data = CommCareImage.get_thumbnail_data(data, self.thumb)
-        return HttpResponse(data, content_type=content_type)
+        extension = ''
+        if self.multimedia.aux_media:
+            extension = self.multimedia.aux_media[-1]['uploaded_filename'].split(".")[-1]
```
millerdev: > know of any better way to do this?

No, but I'd add a method to `CommCareMultimedia` rather than digging into it's implementation details here. Maybe something like `get_file_extension()`, and make it always return a string (with a leading `.` if there is an extension, otherwise empty string) so you can change line 655 to be `extension = self.multimedia.get_file_extension()` and avoid this ugliness here. If it returns a leading `.` in the string you'll also want to change line 659.

biyeun: +1

orangejenny: good point, updated

biyeun (approved):

biyeun:
thanks @orangejenny!

PR #17656, opened 2017-08-31
configure reporting databases separately

snopoke:
@emord I really do like making extra work for myself. Decided it's better to just have these settings completely separate (but with fallbacks to using Django settings).

I'll clean up the previous incantation once this is rolled out. I'll also remove the django migrations tables from the UCR DB's.

### DIFF
--- a/corehq/sql_db/connections.py
+++ b/corehq/sql_db/connections.py
@@ -119,63 +119,69 @@ class ConnectionManager(object):
         return connection_string or self.db_connection_map['default']
 
     def _populate_connection_map(self):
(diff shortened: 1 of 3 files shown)


snopoke:
I'm really going back and forth on this - I was just thinking that if we start using a standby machine for the main DB then we'll need a spearate mechanism to prevent it from being migrated (probably just a check in `multi-migrate`). If that's OK for that DB then it should be fine for these too. New PR coming.

snopoke:
https://github.com/dimagi/commcare-hq/pull/17659

PR #17657, opened 2017-08-31
lock cases during rebuild

snopoke:
This should prevent errors like this: https://sentry.io/dimagi/commcarehq/issues/246263032/events/6442823141/

I'm pretty sure what happened there is that a user triggered a rebuild and then a form submission came in during the rebuild.

### DIFF
--- a/corehq/form_processor/backends/couch/processor.py
+++ b/corehq/form_processor/backends/couch/processor.py
@@ -150,7 +150,7 @@ class FormProcessorCouch(object):
     def hard_rebuild_case(domain, case_id, detail, save=True, lock=True):
         case, lock_obj = FormProcessorCouch.get_case_with_lock(case_id, lock=lock)
         found = bool(case)
-        if found:
+        if not found:
             case = CommCareCase()
             case.case_id = case_id
             case.domain = domain
@@ -159,7 +159,8 @@ class FormProcessorCouch(object):
 
         assert case.domain == domain
         try:
-            if lock_obj:
+            if not found and lock_obj:
+                # the lock was only aquired if we found a case
                 acquire_lock(lock_obj, degrade_gracefully=False)
             forms = FormProcessorCouch.get_case_forms(case_id)
             filtered_forms = [f for f in forms if f.is_normal]
--- a/corehq/form_processor/backends/couch/processor.py
+++ b/corehq/form_processor/backends/couch/processor.py
@@ -157,8 +157,8 @@ class FormProcessorCouch(object):
             if lock:
                 lock_obj = CommCareCase.get_obj_lock_by_id(case_id)
 
-        assert case.domain == domain
         try:
+            assert case.domain == domain, (case.domain, domain)
             if not found and lock_obj:
                 # the lock was only aquired if we found a case
                 acquire_lock(lock_obj, degrade_gracefully=False)
--- a/corehq/form_processor/backends/couch/processor.py
+++ b/corehq/form_processor/backends/couch/processor.py
@@ -150,7 +150,7 @@ class FormProcessorCouch(object):
     def hard_rebuild_case(domain, case_id, detail, save=True, lock=True):
         case, lock_obj = FormProcessorCouch.get_case_with_lock(case_id, lock=lock)
         found = bool(case)
-        if not case:
+        if found:
             case = CommCareCase()
             case.case_id = case_id
             case.domain = domain
--- a/corehq/form_processor/backends/couch/processor.py
+++ b/corehq/form_processor/backends/couch/processor.py
@@ -154,7 +154,7 @@ class FormProcessorCouch(object):
             case = CommCareCase()
             case.case_id = case_id
             case.domain = domain
-            if lock_obj:
+            if lock:
                 lock_obj = CommCareCase.get_obj_lock_by_id(case_id)
 
         try:
--- a/corehq/form_processor/backends/couch/processor.py
+++ b/corehq/form_processor/backends/couch/processor.py
@@ -148,18 +148,18 @@ class FormProcessorCouch(object):
 
     @staticmethod
     def hard_rebuild_case(domain, case_id, detail, save=True, lock=True):
-        case, lock = FormProcessorCouch.get_case_with_lock(case_id, lock=lock)
+        case, lock_obj = FormProcessorCouch.get_case_with_lock(case_id, lock=lock)
         found = bool(case)
         if not case:
             case = CommCareCase()
             case.case_id = case_id
             case.domain = domain
-            if lock:
-                lock = CommCareCase.get_obj_lock_by_id(case_id)
+            if lock_obj:
+                lock_obj = CommCareCase.get_obj_lock_by_id(case_id)
 
         try:
-            if lock:
-                acquire_lock(lock, degrade_gracefully=False)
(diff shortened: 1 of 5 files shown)


snopoke:
best viewed with [w=1](https://github.com/dimagi/commcare-hq/pull/17657/files?w=1)

corehq/form_processor/backends/couch/processor.py
```diff
@@ -147,39 +147,44 @@ def get_cases_from_forms(case_db, xforms):
...
-    def hard_rebuild_case(domain, case_id, detail, save=True):
-        try:
-            case = CommCareCase.get(case_id)
-            assert case.domain == domain
-            found = True
-        except CaseNotFound:
+    def hard_rebuild_case(domain, case_id, detail, save=True, lock=True):
+        case, lock_obj = FormProcessorCouch.get_case_with_lock(case_id, lock=lock)
+        found = bool(case)
+        if not case:
```
millerdev: nit: would read much nicer to say `if not found:`

corehq/form_processor/backends/couch/processor.py
```diff
@@ -147,39 +147,44 @@ def get_cases_from_forms(case_db, xforms):
...
-
-        # add a "rebuild" action
-        case.actions.append(_rebuild_action())
-        if save:
-            case.save()
-        return case
+        try:
+            if lock_obj:
+                acquire_lock(lock_obj, degrade_gracefully=False)
+            assert case.domain == domain
```
millerdev: Would it make sense to do this assertion before acquiring the lock? Acquiring and releasing the lock potentially creates contention which would be nice to avoid if you know it's not necessary.

corehq/form_processor/backends/sql/processor.py
```diff
@@ -243,23 +244,27 @@ def get_cases_from_forms(case_db, xforms):
...
-            assert case.domain == domain
-            found = True
-        except CaseNotFound:
+    def hard_rebuild_case(domain, case_id, detail, lock=True):
+        case, lock_obj = FormProcessorSQL.get_case_with_lock(case_id, lock=lock)
+        if not case:
             case = CommCareCaseSQL(case_id=case_id, domain=domain)
-            found = False
+            if lock_obj:
+                lock_obj = CommCareCaseSQL.get_obj_lock_by_id(case_id)
```
millerdev: I'm not understanding this. Can the `lock_obj` returned by `FormProcessorSQL.get_case_with_lock` not be used here? Looks like at the very least it [should be released](https://github.com/dimagi/commcare-hq/blob/master/corehq/ex-submodules/dimagi/utils/couch/__init__.py#L180-L182) before being replaced.

snopoke: Oops, I refactored the name didn't review carefully enough.

corehq/form_processor/backends/couch/processor.py
```diff
@@ -147,39 +147,44 @@ def get_cases_from_forms(case_db, xforms):
...
+    def hard_rebuild_case(domain, case_id, detail, save=True, lock=True):
+        case, lock_obj = FormProcessorCouch.get_case_with_lock(case_id, lock=lock)
+        found = bool(case)
+        if found:
             case = CommCareCase()
             case.case_id = case_id
             case.domain = domain
-            found = False
+            if lock:
+                lock_obj = CommCareCase.get_obj_lock_by_id(case_id)
```
millerdev: This is reassigning a name that pointed to a different lock that [has already been acquired](https://github.com/dimagi/commcare-hq/blob/master/corehq/ex-submodules/dimagi/utils/couch/__init__.py#L180-L182). Can that lock be used instead of getting a new one here? If not, the other lock should be released before the new one is assigned to `lock_obj` here.

kaapstorm (approved):
*If I'm reading this right*, it looks like it's possible for `case` to be `None` and `lock_obj` [to be a `LockManager`](https://github.com/dimagi/commcare-hq/blob/master/corehq/ex-submodules/dimagi/utils/couch/__init__.py#L214-L214) that is locking ... nothing? So I think @millerdev is right that it should probably be released first. Other than that, LGTM.

corehq/form_processor/backends/couch/processor.py
snopoke: Apologies, bug in my logic. But if a case isn't found then `lock_obj` will be None so we need to get a new lock and acquire it.

corehq/form_processor/backends/couch/processor.py
```diff
@@ -159,7 +159,8 @@ def hard_rebuild_case(domain, case_id, detail, save=True, lock=True):
 
         assert case.domain == domain
         try:
-            if lock_obj:
+            if not found and lock_obj:
+                # the lock was only aquired if we found a case
                 acquire_lock(lock_obj, degrade_gracefully=False)
```
millerdev: Ah, now I see what you're doing. What do you think of moving this line up to line 159?

corehq/form_processor/backends/couch/processor.py
```diff
@@ -147,39 +147,45 @@ def get_cases_from_forms(case_db, xforms):
...
-        # add a "rebuild" action
-        case.actions.append(_rebuild_action())
-        if save:
-            case.save()
-        return case
+        assert case.domain == domain
+        try:
+            if not found and lock_obj:
+                # the lock was only aquired if we found a case
+                acquire_lock(lock_obj, degrade_gracefully=False)
```
millerdev: Ah, now I see what you're doing. I think it would be clearer if you moved this line up to line 159.

corehq/form_processor/backends/couch/processor.py
```diff
@@ -147,39 +147,45 @@ def get_cases_from_forms(case_db, xforms):
...
-                return None
-            # there were no more forms. 'delete' the case
-            case.doc_type = 'CommCareCase-Deleted'
-
-        # add a "rebuild" action
-        case.actions.append(_rebuild_action())
-        if save:
-            case.save()
-        return case
+        assert case.domain == domain
```
millerdev: This should move into the `try` block. Otherwise you could end up with a lock that never gets released in case of assertion failure. Sorry I didn't catch this before because it wasn't clear to me what was happening with the `lock_obj` on line 151.

Nit: the error message of this assertion (if it fails) will not be very informative. Suggested change:
```py
assert case.domain == domain, (case.domain, domain)
```

millerdev (commented):
Both comments also apply to the same logic in sql/processor.py

Sorry about all the iterations on this. It's taken me a while to wrap my head around what's going on here.

corehq/form_processor/backends/couch/processor.py
snopoke: Ah, yes, forgetting that if we got a lock initially then this won't release it. I'd thought there was a reason I'd had it that way initially but couldn't remember.

snopoke: 6222ce7

corehq/form_processor/backends/couch/processor.py
snopoke: d6d76db

snopoke:
Sorry for the iteration on this - I should stop trying to code in between meetings. I think I've addressed all comments now.

PR #17662, opened 2017-08-31
Je/decouple aggregation tasks

emord:
@sheelio Can you confirm that there are no dependencies between month and daily aggregations? This change would queue up different tasks after the location table and functions are created. This will mean that just because one monthly aggregation fails the other months/daily aggregation won't fail.

@biyeun @proteusvacuum


sheelio:
The daily aggregation just takes a copy of the agg AWC monthly aggregation for that day (so there's a big dependency).

emord:
ack. ok. i forgot about that. will think about this again.

PR #17667, opened 2017-08-31
ICDS: Reverse order of aggregation

sheelio:
@emord 

This pulls out Kamil's change and reverses the order that months are run in.

### DIFF
--- a/custom/enikshay/const.py
+++ b/custom/enikshay/const.py
@@ -97,7 +97,6 @@ VALID_ADHERENCE_SOURCES = (
 )
 
 CASE_VERSION = 'case_version'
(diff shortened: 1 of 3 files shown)


custom/enikshay/const.py
```diff
@@ -97,6 +97,7 @@
 )
 
 CASE_VERSION = 'case_version'
+CASE_MINOR_VERSION = 'case_minor_version'
```
emord: This is unrelated can you remove it?

sheelio:
@emord 
Removed the file that got accidentally included.

emord (approved):

PR #17672, opened 2017-09-01
Don't trigger on migrated persons

esoergel:
@nickpell

Following incentives to be triggered ONLY for new case created in eNikshay. Not
to be triggered for migrated cases
·   MBBS+ Providers: 180 days of private OR govt. FDCs
·   Providers: For diagnosis and notification of TB case
·   AYUSH/Other provider: Registering and referral of a presumptive TB case

### DIFF
--- a/custom/enikshay/integrations/utils.py
+++ b/custom/enikshay/integrations/utils.py
@@ -49,15 +49,7 @@ def is_valid_episode_submission(episode_case):
 
 
 def is_migrated_uatbc_episode(episode_case):
-    try:
-        person_case = get_person_case_from_episode(episode_case.domain, episode_case)
-    except ENikshayCaseNotFound:
-        return False
-    return is_migrated_uatbc_person(person_case)
-
-
-def is_migrated_uatbc_person(person_case):
-    return bool(person_case.get_case_property('migration_created_from_record'))
+    return episode_case.get_case_property('migration_created_case') == 'true'
 
 
 def is_valid_voucher_submission(voucher_case):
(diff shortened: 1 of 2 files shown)


custom/enikshay/integrations/utils.py
```diff
@@ -48,6 +48,18 @@ def is_valid_episode_submission(episode_case):
     return not _is_submission_from_test_location(person_case.case_id, person_case.owner_id)
 
 
+def is_migrated_uatbc_episode(episode_case):
+    try:
+        person_case = get_person_case_from_episode(episode_case.domain, episode_case)
+    except ENikshayCaseNotFound:
+        return False
+    return is_migrated_uatbc_person(person_case)
```
esoergel: @nickpell can you confirm this is the best way to figure out whether an episode originated in UATBC or eNikshay?

nickpell: ```'migration_created_from_record'``` is set on episode as well

custom/enikshay/integrations/utils.py
```diff
@@ -48,6 +48,18 @@ def is_valid_episode_submission(episode_case):
...
+
+
+def is_migrated_uatbc_person(person_case):
+    return bool(person_case.get_case_property('migration_created_from_record'))
```
nickpell: for safety I would do ```person_case.get_case_property('migration_created_from_record' == 'true'``` since ```bool('false') == False'```

nickpell (changes requested):

custom/enikshay/integrations/utils.py
esoergel: oh, fancy!  That's handy.

custom/enikshay/integrations/utils.py
esoergel: nice catch, thanks, will do!

esoergel: oh wait, nevermind, I remember why I used `bool` - isn't `migration_created_from_record` an ID of a the record in UATBC?

nickpell (approved):

PR #17675, opened 2017-09-01
Move hqwebapp/images to style/images

orangejenny:
Friday refactor: merge `hqwebapp` into `style` - they have the same purpose, but `hqwebapp` is legacy. Doing over multiple PRs for reviewers' (and my) sanity.

@gcapalbo / @calellowitz 

fyi @biyeun


orangejenny:
Change of plans, going to think about this but will likely move `style` into `hqwebapp` rather than the other way around.

PR #17683, opened 2017-09-01
Update prelogin again

orangejenny:
Will wait on tests, then merge along with https://github.com/dimagi/commcare-hq/pull/17682


orangejenny:
moving into https://github.com/dimagi/commcare-hq/pull/17682

PR #17684, opened 2017-09-01
include more info about autopay errors

nickpell:


corehq/apps/accounting/payment_handlers.py
```diff
@@ -424,13 +424,19 @@ def _send_payment_receipt(self, invoice, payment_record):
...
+        err = body.get('error', {})
+
         log_accounting_error(
             "[Autopay] An automatic payment failed for invoice: {} "
             "because the card was declined. This invoice will not be automatically paid. "
-            "Not necessarily actionable, but be aware that this happened."
-            .format(invoice.id)
+            "Not necessarily actionable, but be aware that this happened. "
+            "error = {}"
+            .format(invoice.id, err)
```
dannyroberts: `log_accounting_error` has a param called `show_stack_trace`. Does it not work? Or is it not what you want? (Stack trace always seems better to me than just the error message.)

nickpell: In this case what I really want is the json that stripe returns to us - it includes details about the reason the card payment failed to process

nickpell:
cc: @proteusvacuum

proteusvacuum (approved):

PR #17687, opened 2017-09-01
ICDS: SMS Indicators AWW 4 & 5

gcapalbo:
Context on how immunizations are tracked in the AWW app is [here](https://docs.google.com/document/d/1DMLRCeymGUqLUyamcggJW1bBC-Z7ITS4aCgnPnlxpWQ/edit#)

This adds utils for calculating when immunizations are due, and then uses those utils to implement SMS indicator AWW 4.

Also adds indicator AWW 5.

@emord @orangejenny

### DIFF
--- a/custom/icds/rules/immunization.py
+++ b/custom/icds/rules/immunization.py
@@ -1,0 +1,190 @@
+import pytz
+import re
+from corehq.apps.products.models import SQLProduct
+from corehq.form_processor.backends.sql.dbaccessors import LedgerAccessorSQL
+from corehq.form_processor.models import CommCareCaseIndexSQL
+from corehq.util.timezones.conversions import ServerTime
+from datetime import datetime, date, timedelta
+
+
+def _validate_tasks_case_and_immunization_product(tasks_case, immunization_product):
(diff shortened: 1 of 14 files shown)


custom/icds/rules/immunization.py
```diff
@@ -0,0 +1,216 @@
...
+        raise ValueError("Expected one of: 'child', 'pregnancy'")
+
+
+@quickcache(['domain', 'schedule'], timeout=60 * 60)
+def get_immunization_products(domain, schedule):
+    _validate_child_or_pregnancy_type(schedule)
+
+    return [
+        product
+        for product in SQLProduct.active_objects.filter(domain=domain)
```
emord: How many products are expected to be in the domain? Could it be filtered by anything else like code or program id?

emord (approved):

custom/icds/rules/immunization.py
gcapalbo: there's less than 50 total in the domain at the moment. unfortunately the only other thing we could filter on is part of the custom product data which is not possible in the sql query (there are no programs defined). but i cached this result so this lookup in the db shouldn't happen often

PR #17691, opened 2017-09-02
Change amounts back into a string, but rounded

esoergel:
@millerdev
Walking back a component of https://github.com/dimagi/commcare-hq/pull/17680


esoergel:
@proteusvacuum

PR #17692, opened 2017-09-03
Mk add report commands

mkangia:
Adding reports for debugging
1. Generate report for a repeater or a list of records ids
2. Add report for enikshay cases for checking if nikshay notified by passing case ids

### DIFF
--- a/corehq/motech/repeaters/management/commands/generate_repeater_report.py
+++ b/corehq/motech/repeaters/management/commands/generate_repeater_report.py
@@ -12,7 +12,7 @@ from corehq.motech.repeaters.dbaccessors import iter_repeat_records_by_domain, g
 class Command(BaseCommand):
     help = """
     Pass Repeater ID along with domain and State(Optional) or
-    Pass a list of repeat records IDs to get a report
+    Pass a csv file path with a list of repeat records IDs to get a report(xlsx)
     with final state and message for all attempts(if available)
     """
 
--- a/corehq/motech/repeaters/management/commands/generate_repeater_report.py
+++ b/corehq/motech/repeaters/management/commands/generate_repeater_report.py
@@ -100,12 +100,8 @@ class Command(BaseCommand):
             records = self.record_ids
             record_count = len(records)
         elif domain and repeater_id:
-            if state:
-                records = iter_repeat_records_by_domain(domain, repeater_id=repeater_id, state=state)
-                record_count = get_repeat_record_count(domain, repeater_id=repeater_id, state=state)
-            else:
-                records = iter_repeat_records_by_domain(domain, repeater_id=repeater_id)
-                record_count = get_repeat_record_count(domain, repeater_id=repeater_id)
+            records = iter_repeat_records_by_domain(domain, repeater_id=repeater_id, state=state)
+            record_count = get_repeat_record_count(domain, repeater_id=repeater_id, state=state)
         else:
             raise CommandError("Insufficient Arguments")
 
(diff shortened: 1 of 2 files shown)


corehq/motech/repeaters/management/commands/generate_repeater_report.py
```diff
@@ -0,0 +1,123 @@
...
+
+from django.core.management.base import BaseCommand, CommandError
+from corehq.util.log import with_progress_bar
+from corehq.motech.repeaters.models import RepeatRecord
+from corehq.motech.repeaters.dbaccessors import iter_repeat_records_by_domain, get_repeat_record_count
+
+
+class Command(BaseCommand):
+    help = """
+    Pass Repeater ID along with domain and State(Optional) or
```
sravfeyn: Reading just from the help message, it's not clear what the input file should be and whether the output will be a stdout or a file somewhere

mkangia: That is so true.. Updated.

corehq/motech/repeaters/management/commands/generate_repeater_report.py
```diff
@@ -0,0 +1,123 @@
...
+                first_empty_index = i
+                break
+        for i in range(0, self.max_attempts_in_sheet):
+            column_num = first_empty_index + i
+            self.ws.cell(row=1, column=column_num).value = "Attempt {index}".format(index=i + 1)
+
+    def _save_file(self, repeater_id, state):
+        if self.max_attempts_in_sheet > 0:
+            self._add_header_for_attempts()
+        file_name = "repeater_report_{repeater_id}_{state}_{timestamp}.xlsx".format(
```
proteusvacuum: is there a reason you didn't just write this as a csv?

corehq/motech/repeaters/management/commands/generate_repeater_report.py
```diff
@@ -0,0 +1,123 @@
...
+        records_file_path = options.get('records_file_path')
+
+        if records_file_path:
+            self._load_record_ids_from_file(records_file_path)
+            records = self.record_ids
+            record_count = len(records)
+        elif domain and repeater_id:
+            if state:
+                records = iter_repeat_records_by_domain(domain, repeater_id=repeater_id, state=state)
+                record_count = get_repeat_record_count(domain, repeater_id=repeater_id, state=state)
```
proteusvacuum: I think both of these methods work if `state` is `None`, so you could remove this if statement

custom/enikshay/management/commands/generate_nikshay_notification_report.py
```diff
@@ -0,0 +1,214 @@
...
+        parser.add_argument(
+            '--parse_as_person_enikshay_ids',
+            dest="parse_as_person_enikshay_ids",
+            action='store_true',
+            help='pass this to let parser know to look for enikshay ids instead of episode ids in file'
+        )
+
+    def _search_person_using_enikshay_id(self, enikshay_id):
+        search_results = self.search.case_property_query("person_id", enikshay_id).run()
+        return [case['_id'] for case in search_results.raw_hits]
```
proteusvacuum: you can use `self.search.case_property_query("person_id", enikshay_id).values_list(['_id', ], flat=True)` instead of this.

proteusvacuum (approved):

corehq/motech/repeaters/management/commands/generate_repeater_report.py
mkangia: I just find xl better to pass around to others, specially other stakeholders.  Plus, then there are no compatibility issues when they would open this file in excel and no surprises with fields having commas in them etc.

PR #17695, opened 2017-09-04
Ce/update case property

calellowitz:
@orangejenny 
This fixes up an old icds management command to be more broadly applicable and moves it out of custom code.

### DIFF
--- a/corehq/apps/case_importer/management/commands/update_case_property.py
+++ b/corehq/apps/case_importer/management/commands/update_case_property.py
@@ -13,9 +13,9 @@ from dimagi.utils.chunked import chunked
 class Command(BaseCommand):
 
     def add_arguments(self, parser):
-        parser.add_arugment('domain')
-        parser.add_arugment('infile')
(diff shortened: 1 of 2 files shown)


corehq/apps/case_importer/management/commands/update_case_property.py
```diff
@@ -13,11 +13,12 @@
 class Command(BaseCommand):
 
     def add_arguments(self, parser):
-        parser.add_arugment('infile')
-        parser.add_arugment('logfile')
```
orangejenny: did this just never work?

orangejenny (approved):

corehq/apps/case_importer/management/commands/update_case_property.py
calellowitz: I had made a fix in another branch and deployed a private release with that when I ran the command for ICDS. That was the night I left for vacation, and I had forgotten to PR it when I got back.

PR #17698, opened 2017-09-04
adding docs to update case property command

calellowitz:
@orangejenny 
Now that its in the main codebase, a description seemed necessary.


orangejenny (approved):
agreed

PR #17700, opened 2017-09-04
Restored prefilter in tb notif

kkrampa:


kkrampa:
@emord Are you able to deploy this today to Eniksay server.

emord:
@calellowitz is handling enikshay UCRs now

PR #17674, opened 2017-09-01
Excluded test locations from results

kkrampa:

PR #17705, opened 2017-09-05
ICDS UI changes

kkrampa:
@calellowitz Could you merge https://github.com/dimagi/commcare-hq/pull/17674 before this one?


custom/icds_reports/utils.py
```diff
@@ -243,6 +243,30 @@ def apply_exclude(domain, queryset):
     )
 
 
+def get_age_filter(age_value):
+    """
+        When age_value comes from age filter it means first range is chosen 0-6 months.
+        For that range we want to include 0 and 6 in results.
+    """
+    if age_value == '6':
```
calellowitz: can you explain this part a little more, im not sure i understand what is going on here

calellowitz (changes requested):
one question then it looks good to go

custom/icds_reports/utils.py
```diff
@@ -1414,6 +1476,9 @@ def get_data_for(filters, date):
             weighed=Sum('wer_weighed'),
             all=Sum('wer_eligible'),
         )
+        if show_test:
```
emord: I think this should be `not show_test` (there are some others that need to be switched as well)

emord (changes requested):

custom/icds_reports/utils.py
kkrampa: Fixed in all places

custom/icds_reports/utils.py
kkrampa: We have a bunch of values in filter
0 - 6 months, 6 - 12 months, etc.
currently, age_tranche can have a values 0, 6, 12, 18 etc.
To not lose value '0' we need to include it in the first range 0-6. For the rest of ranges, we include only rightmost value.

0 -6 -> 0,6
6-12 -> 12
12-18 -> 18
etc.
As you can see without this clause we lose 0 value at all.

kkrampa:
@emord @calellowitz 
Updated

emord (approved):

emord:
This PR now has a file conflict

kkrampa:
@calellowitz @emord Updated

calellowitz (approved):

PR #17707, opened 2017-09-05
Ce/no patch thread

calellowitz:
@dannyroberts cc: @millerdev 
https://manage.dimagi.com/default.asp?261065
priming the views is no longer relevant since hardly anyone is on couch and `is_bigcouch` is only false for enikshay and swiss (and it will be true for enikshay soon)


calellowitz:
For the record this was failing because `prime_views` was causing gevent to monkey patch threading half way through the upload which changed the thread of the process to change raising a `DatabaseError`

dannyroberts (approved):

PR #17708, opened 2017-09-05
Update regex to be a bit more lenient with characters to allow symbol…

mkangia:
…s after encoding

For some cases the obfuscated password at first level of encoding can end up with symbols which are currently not allowed according to the regex.

Resolves: https://manage.dimagi.com/default.asp?260563

Found one edge case where the above happens when "?" and "@" are together in a password.

cc: @sravfeyn


custom/nic_compliance/utils.py
```diff
@@ -12,7 +12,7 @@
 
 PASSWORD_HASHER = get_hasher()
 # Passwords set with expected padding length and format would respect this regex
-PASSWORD_REGEX = r"^sha256\$([a-z0-9A-Z]{6})([a-zA-Z0-9=]*)([a-z0-9A-Z]{6})=$"
+PASSWORD_REGEX = r"^sha256\$([a-z0-9A-Z]{6})(\S*)([a-z0-9A-Z]{6})=$"
```
mkangia: @snopoke this reverts the change we made [here](https://github.com/dimagi/commcare-hq/pull/15490/files/db98cf7fd67ae0aef06edf407eb9cae3f34737dc#r114754348)

PR #17712, opened 2017-09-05
ICDS: Fix tetanus complete expression

sheelio:
@emord 

The application unfortunately stores tetanus complete on the tasks case instead of the ccs record case.  I'm going to ask the app team to change, but in the meantime we have to check the task subcases to get the information. 

(This may need a rebuild of the ccs record data source after its merged - we should probably time that rebuild with reducing the number of iterations)

### DIFF
--- a/custom/icds_reports/ucr/data_sources/ccs_record_cases_monthly_tableau2.json
+++ b/custom/icds_reports/ucr/data_sources/ccs_record_cases_monthly_tableau2.json
@@ -1982,19 +1982,34 @@
                 }
               },
               "filter_expression": {
-                "type": "not",
-                "filter": {
-                  "operator": "in",
-                  "type": "boolean_expression",
-                  "expression": {
-                    "type": "property_name",
-                    "property_name": "tt_complete_date"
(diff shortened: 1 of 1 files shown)


custom/icds_reports/ucr/data_sources/ccs_record_cases_monthly_tableau2.json
```diff
@@ -1941,6 +1933,77 @@
...
+                "type": "get_subcases",
+                "case_id_expression": {
+                  "type": "root_doc",
+                  "expression": {
+                    "type": "property_name",
+                    "property_name": "_id"
+                  }
+                }
+              },
+              "filter_expression": {
```
emord: I'm not super familiar with the case structure here. Are there multiple task cases for each ccs record case? Should this filter be based on type instead of a property name? or maybe it can be filtered both on type and property name

sheelio:
Added filtering for the case type as well.  (You're right there could be other case types other than the tasks case).

emord (approved):

PR #17714, opened 2017-09-05
optimize answer responses containing long choice lists

dannyroberts:
by not updating the choices if they haven't changed

### DIFF
--- a/corehq/apps/cloudcare/static/cloudcare/js/form_entry/fullform-ui.js
+++ b/corehq/apps/cloudcare/static/cloudcare/js/form_entry/fullform-ui.js
@@ -162,13 +162,19 @@ Container.prototype.fromJS = function(json) {
                 if (ko.utils.unwrapObservable(options.target.serverError)) {
                     options.data.answer = _.clone(options.target.answer());
                 }
+                if (_.isEqual(options.target.choices(), options.data.choices)) {
+                    // replacing the full choice list if it has a few thousand items
+                    // is actually quite expensive and can freeze the page for seconds.
+                    // at the very least we can skip entirely when there's no change.
+                    delete options.data.choices;
+                }
                 return options.target;
             },
             key: function(data) {
                 return ko.utils.unwrapObservable(data.uuid) || ko.utils.unwrapObservable(data.ix);
             }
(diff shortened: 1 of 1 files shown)


ctsims:
@dannyroberts you rule, thanks for getting this out so fast! 

not sure what the knockout equality op does on these sets, the list order should be guaranteed from the DB, since this is in in XML land at this point, but not sure if he widget ordering retains that order specifically since it's an autocomplete/combobox.

corehq/apps/cloudcare/static/cloudcare/js/form_entry/fullform-ui.js
```diff
@@ -162,13 +162,19 @@ Container.prototype.fromJS = function(json) {
                 if (ko.utils.unwrapObservable(options.target.serverError)) {
                     options.data.answer = _.clone(options.target.answer());
                 }
+                if (_.isEqual(options.target.choices(), options.data.choices)) {
```
orangejenny: Is there anything like an id or path in the array elements that you could pluck out to make the comparison quicker?

orangejenny (approved):

corehq/apps/cloudcare/static/cloudcare/js/form_entry/fullform-ui.js
dannyroberts: Ha, so actually choices is an array of strings, not of the richer objects you might expect, so there's nothing to compare on but these string values themselves.

dannyroberts:
@ctsims, yeah this `_.isEqual` is just a utility (part of the underscore.js or lodash.js library) that is there to make up for the fact that javascript equality is rarely ever what you want it to be, specifically  stuff like `_.isEqual([1], [1])` returns `true` where `[1] == [1]` returns `false` (because `==`/`===` in this case uses object equality). Anyway,

```js
> _.isEqual([1, 2], [1, 2]);
true
> _.isEqual([1, 2], [2, 1]);
false
```
so there's nothing to worry about—in the edge case where the order changes, the dropdown will go through the (slow) update process.

gcapalbo (approved):

PR #17717, opened 2017-09-06
Don't ovewrite "fulfilled" properties from app in BETS API

proteusvacuum:
@nickpell 

We were overwriting properties set by the app with other data from BETS.


proteusvacuum:
FYI @esoergel

nickpell (approved):

custom/enikshay/integrations/bets/tests/test_bets_updates.py
```diff
@@ -70,7 +70,7 @@ def test_update_voucher_success(self):
         }]})
```
esoergel: :+1: on the rename

custom/enikshay/integrations/bets/views.py
```diff
@@ -69,8 +69,9 @@ def properties(self):
         if self.status == SUCCESS:
             return {
                 'state': 'paid',
-                'amount_fulfilled': self.amount,
-                'date_fulfilled': self.paymentDate.isoformat(),
+                'amount_paid': self.amount,
+                'date_paid': self.paymentDate.date().isoformat(),
+                'time_paid': self.paymentDate.time().isoformat(),
```
esoergel: This seems odd - what's the deal with splitting this up?  I'm sure you've got your reasons, just curious.

esoergel (approved):

custom/enikshay/integrations/bets/views.py
proteusvacuum: we don't deal with date-times well (or at all) on formplayer, so if this data needs to be shown anywhere / filtered in the app, it's better to just do it on "date" where we can actually cast to a date.

PR #17722, opened 2017-09-06
Base for SMS UI

emord:
@gcapalbo There are a few more things to do here, but wanted to get the base work out before I get pulled into something else. It currently allows you to add an immediate SMS broadcast to a user. Things left to do:

* Support Email/IVR survey
* support Locations and case groups and user groups
* add scheduling UI


corehq/messaging/scheduling/forms.py
```diff
@@ -1,22 +1,169 @@
...
         self.domain = domain
         self.id = id
         self.query_url = reverse('possible_sms_recipients', args=[self.domain])
+        self.readonly = readonly
 
     def render(self, name, value, attrs=None):
+        initial_value = []
+        if value:
+            for doc_type, doc_id in value:
+                user = CommCareUser.wrap(CommCareUser.get_db().get(doc_id))
```
gcapalbo: we should doublecheck domain here

corehq/messaging/scheduling/forms.py
```diff
@@ -1,22 +1,169 @@
...
+    def to_python(self, value):
+        if not value:
+            return []
+        return value.split(',')
+
+    def validate(self, value):
+        # TODO Will need to add more than user ids
+        # TODO batch id verification
+        for user_id in value:
+            CommCareUser.get_db().get(user_id)
```
gcapalbo: also should doublecheck domain here

corehq/messaging/scheduling/forms.py
```diff
@@ -1,22 +1,169 @@
...
+    send_frequency = ChoiceField(
+        required=True,
+        label=_('Send'),
+        choices=(
+            ('immediately', _('Immediately')),
+        )
+    )
+    recipients = MessageRecipientField(
+        label=_("Recipient(s)"),
+        help_text=_("Type a username, group name or 'send to all'")
```
gcapalbo: we can do away with the 'send to all', we won't be supporting that anymore now that we can support locations

corehq/messaging/scheduling/urls.py
```diff
@@ -2,10 +2,14 @@
...
     BroadcastListView,
+    CreateMessageView,
+    EditMessageView,
     possible_sms_recipients,
 )
 
 urlpatterns = [
     url(r'^broadcasts/$', BroadcastListView.as_view(), name=BroadcastListView.urlname),
+    url(r'^broadcasts/add/$', CreateMessageView.as_view(), name=CreateMessageView.urlname),
+    url(r'^broadcasts/edit/(?P<broadcast_id>[\w-]+)/$', EditMessageView.as_view(), name=EditMessageView.urlname),
```
gcapalbo: i'm not sure if i like the name of CreateMessage / EditMessage in the views, the templates, and the js files, since we're really scheduling abstract content (which could also be a survey, ivr survey, etc.).  can we rename the associated files to  create schedule and edit schedule?

corehq/messaging/scheduling/views.py
```diff
@@ -107,6 +117,130 @@ def get(self, request, *args, **kwargs):
...
+                        if key in values:
+                            messages[lang] = values[key]
+                    content = SMSContent(message=messages)
+                else:
+                    content = SMSContent(message={'*': values['non_translated_message']})
+                schedule = AlertSchedule.create_simple_alert(self.domain, content)
+                broadcast = ImmediateBroadcast(
+                    domain=self.domain, name=values['schedule_name'], schedule=schedule
+                )
+                broadcast.save()
```
gcapalbo: would be good to edit the schedule and broadcast in a transaction

corehq/messaging/scheduling/views.py
```diff
@@ -107,6 +117,130 @@ def get(self, request, *args, **kwargs):
...
+            'message': schedule.memoized_events[0].content.message,
+        }
+        return MessageForm(initial=initial, **self.form_kwargs)
+
+    def post(self, request, *args, **kwargs):
+        values = self.message_form.cleaned_data
+        # TODO shouldn't have gotten here. add soft assert
+        if values['send_frequency'] == 'immediately':
+            # TODO 404 probably isn't the right thing
+            raise Http404()
```
gcapalbo: i think we never expect this to happen since we shouldn't be editing the immediate broadcasts, so i'm ok with raising a normal exception and failing hard if this ever happens

corehq/messaging/scheduling/views.py
```diff
@@ -115,6 +249,7 @@ def possible_sms_recipients(request, domain):
     # TODO Add locations
     # TODO Add mobile worker groups
     # TODO will need to know doc type as well
+    # TODO Support "send to all"
```
gcapalbo: (we can leave out support for send to all)

corehq/messaging/scheduling/forms.py
```diff
@@ -1,22 +1,169 @@
...
+    )
+
+    def __init__(self, *args, **kwargs):
+        self.domain = kwargs.pop('domain')
+        initial = kwargs.get('initial')
+        readonly = False
+        if initial:
+            # TODO Recipients isn't set properly
+            readonly = (initial.get('send_frequency') == 'immediately')
+            message = initial.get('message', dict)
```
gcapalbo: should this be `message = initial.get('message', {})` ?

corehq/messaging/scheduling/views.py
```diff
@@ -107,6 +117,130 @@ def get(self, request, *args, **kwargs):
...
+        recipients = [
+            (instance.recipient_type, instance.recipient_id)
+            for instance in schedule_instances
+        ]
+        initial = {
+            'schedule_name': broadcast.name,
+            'send_frequency': 'immediately',
+            'recipients': recipients,
+            'content': 'sms',
+            'message': schedule.memoized_events[0].content.message,
```
gcapalbo: this will only work if content is an instance of `SMSContent`

corehq/apps/smsbillables/async_handlers.py
```diff
@@ -102,7 +102,9 @@ def _fmt_success(self, response):
 
 class PublicSMSRatesAsyncHandler(BaseAsyncHandler):
     slug = 'public_sms_rate_calc'
-    allowed_actions = 'public_rate'
+    allowed_actions = [
+        'public_rate',
+    ]
```
gcapalbo: yeah seems like this was wrong. seems like we got away with it because of how the [validation](https://github.com/dimagi/commcare-hq/blob/3dc00f772650c7fca26b2be074123179e38ab1f9/corehq/apps/hqwebapp/async_handler.py#L73) is done

corehq/messaging/scheduling/views.py
```diff
@@ -229,10 +236,15 @@ def message_form(self):
             (instance.recipient_type, instance.recipient_id)
             for instance in schedule_instances
         ]
+        ret = []
+        if recipients:
+            for doc_type, doc_id in recipients:
+                user = CommCareUser.wrap(CommCareUser.get_db().get(doc_id))
```
gcapalbo: i realize we're only supporting users for this pr, but we'll need to make this conditional based on doc type eventually

gcapalbo:
lookin good. when was the last time this was merged with master? there have been some shuffling of ui components recently between style and hqwebapp (i see you caught one of them in the last commit)

corehq/messaging/scheduling/urls.py
emord: https://github.com/dimagi/commcare-hq/pull/17722/commits/a9f1dca7096f85fe921b5010692a5aec6999ef2b

corehq/messaging/scheduling/views.py
emord: https://github.com/dimagi/commcare-hq/pull/17722/commits/6f3806dd1c4b2be72ffe1ba40404abf50314c719

corehq/messaging/scheduling/views.py
emord: I changed this to using a soft assert and returning a 400 here: https://github.com/dimagi/commcare-hq/pull/17722/commits/91355a2008e150633c79c9cfeea0ca5fe5f5d8c9

Does that seem like an ok solution or should we 500?

corehq/messaging/scheduling/views.py
emord: Thanks I missed this comment earlier https://github.com/dimagi/commcare-hq/pull/17722/commits/5ddb89e5b8a08962c73deb40c9623eaf7a1423fc

corehq/messaging/scheduling/forms.py
emord: https://github.com/dimagi/commcare-hq/pull/17722/commits/eef0c7790f399abc7daf0dccd2171432df17b0cb

corehq/messaging/scheduling/views.py
emord: I added a comment https://github.com/dimagi/commcare-hq/pull/17722/commits/d2a7d2bc780416e0d7463b42b5a52d77f8094ab4. I plan on handling this when adding support for surveys and emails

emord:
I rebased it on master yesterday so it includes the changes ui components.

according to the commit times, it looks like I have  been working on these on and off since june 30. hoping to get this in before diving into the other types of schedules so I don't need to handle that large of rebase again. and i'm not sure how much time i'll get to work on this while I'm in Delhi as dashboards/UCRs are taking priority.

corehq/messaging/scheduling/views.py
gcapalbo: i think it'd prefer a 500 since it would really be a bug that we'd need to fix

gcapalbo:
ok - i'm still seeing a template import error when i access /messaging/broadcasts/add/ locally which i think is related to the ui component shuffling

but besides that i think all that's outstanding are my two comments about validating domain [here](https://github.com/dimagi/commcare-hq/pull/17722#discussion_r137286804) and [here](https://github.com/dimagi/commcare-hq/pull/17722#discussion_r137286894), and making editing an immediate broadcast 500 instead of soft assert. after that i think we're good to merge

corehq/messaging/scheduling/forms.py
emord: This code was removed when I refactored to using the separate async handler

corehq/messaging/scheduling/forms.py
emord: I needed to move this so that it had access to the domain https://github.com/dimagi/commcare-hq/pull/17722/commits/0348c1d6c426985e8f8c5bab5f973a6f30b6f29e

corehq/messaging/scheduling/views.py
emord: https://github.com/dimagi/commcare-hq/pull/17722/commits/3e0fb84ab5d4333c205561c2c2281fd296c4a655

PR #17727, opened 2017-09-06
Cache domain global fixtures

millerdev:
This is another try at #16904 and #16951. The commits are loosely based on that work, but substantially refactored. Best reviewed by commit 🐠 

@proteusvacuum cc @calellowitz


corehq/apps/fixtures/fixturegenerators.py
```diff
@@ -45,42 +48,81 @@ class ItemListsProvider(FixtureProvider):
...
+        restore_user = restore_state.restore_user
+        domain = restore_user.domain
+        db = get_blob_db()
+        if not restore_state.overwrite_cache:
+            try:
+                fixture_data = db.get(domain, FIXTURE_BUCKET).read()
+                return [fixture_data] if fixture_data else []
+            except NotFound:
+                pass
+        global_items = self._get_global_items(global_types, restore_user)
```
millerdev: this will cache, globally for the domain, XML that includes this `restore_user.user_id`. Looks like that issue was reported in [FB257817](https://manage.dimagi.com/default.asp?257817), but not fixed by #16951. It's unclear to me if this is a blocker or not, but I'm guessing it is.

corehq/apps/fixtures/fixturegenerators.py
```diff
@@ -45,42 +56,86 @@ class ItemListsProvider(FixtureProvider):
...
+    def get_global_items(self, global_types, restore_state):
+        restore_user = restore_state.restore_user
+        user_id = restore_user.user_id
+        domain = restore_user.domain
+        db = get_blob_db()
+        if not restore_state.overwrite_cache:
+            global_id = GLOBAL_USER_ID.encode('utf-8')
+            b_user_id = user_id.encode('utf-8')
+            try:
+                data = db.get(domain, FIXTURE_BUCKET).read()
```
millerdev: This uses the same key that was used in #16904 and #16951, which means it may return previously cached data that is likely outdated and definitely contains the wrong `user_id` for most users.

It's pretty easy to delete the `FIXTURE_BUCKET` in the blob db (needs to be run on each env before this is deployed):
```py
from corehq.blobs import get_blob_db
FIXTURE_BUCKET = 'domain-fixtures'
db = get_blob_db()
db.delete(bucket=FIXTURE_BUCKET)
```

corehq/apps/fixtures/fixturegenerators.py
proteusvacuum: oh man, I guess this explains why some fixtures were just not showing up for users - they were assigned to the user who initially created the cached restore.

corehq/apps/fixtures/fixturegenerators.py
```diff
@@ -45,42 +56,85 @@ class ItemListsProvider(FixtureProvider):
...
+        return items
+
+    def get_global_items(self, global_types, restore_state):
+        restore_user = restore_state.restore_user
+        user_id = restore_user.user_id
+        domain = restore_user.domain
+        db = get_blob_db()
+        if not restore_state.overwrite_cache:
+            global_id = GLOBAL_USER_ID.encode('utf-8')
+            b_user_id = user_id.encode('utf-8')
```
proteusvacuum: what does `b` stand for?

corehq/apps/fixtures/fixturegenerators.py
proteusvacuum: should this get added to a migration?

corehq/apps/fixtures/fixturegenerators.py
millerdev: binary/bytes

proteusvacuum (approved):

corehq/apps/fixtures/fixturegenerators.py
millerdev: 6a74915c3f197ef197a34514c7ff313ac0a68d2f

proteusvacuum:
will merge on monday when you are back online so we can monitor error rates

PR #17728, opened 2017-09-06
Add report for private patient notifications

mkangia:
Adding a report for Private app along on same lines as Public App to look at notifications being made to Nikshay

cc: @proteusvacuum @sheelio

### DIFF
--- a/custom/enikshay/ucr/reports/tb_notification_register_private.json
+++ b/custom/enikshay/ucr/reports/tb_notification_register_private.json
@@ -1,0 +1,712 @@
+{
+    "domains": [
+        "enikshay",
+        "enikshay-reports-qa"
+    ],
+    "server_environment": [
+        "enikshay"
+    ],
+    "data_source_table": "episode_2b",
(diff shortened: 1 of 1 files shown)


custom/enikshay/ucr/reports/tb_notification_register_private.json
```diff
@@ -0,0 +1,712 @@
...
+                    "confirmed_drtb"
+                ],
+                "datatype": "string",
+                "display": null,
+                "field": "episode_type",
+                "type": "pre",
+                "slug": "episode_type"
+            },
+            {
+                "pre_value": 1,
```
mkangia: This is the only difference from the public notification report except less domains to built for and only for enikshay environment.

emord (approved):
next time it may be better to have the first commit be the copy to a new file and a second commit showing the changes made

mkangia:
> next time it may be better to have the first commit be the copy to a new file and a second commit showing the changes made

Cool!

PR #17729, opened 2017-09-06
enikshay: V2 tasklist data sources

sheelio:
@nickpell 

These are the V2 versions of the task list data sources.  Once this is deployed, we'll need to rebuild them asynchronously.  Once that's done, we can repoint all the existing reports to these and remove the v1 data source.

### DIFF
--- a/custom/enikshay/ucr/data_sources/episode_tasklist_v2.json
+++ b/custom/enikshay/ucr/data_sources/episode_tasklist_v2.json
@@ -1,0 +1,2694 @@
+{
+  "domains": [
+    "enikshay",
+    "sheel-enikshay",
+    "enikshay-reports-qa",
+    "enikshay-test"
+  ],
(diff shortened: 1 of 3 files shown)


nickpell (approved):
looks good pending tests

sheelio:
@nickpell 

Can you run the async rebuild command for these tables once deployed?

nickpell:
will do

nickpell:
I kicked off the async rebuild.

@sheelio The table id for these static data sources is identical as the dynamic reports, which means that the static and dynamic data sources point to the same tables.  This has the unfortunate affect that deploying the static data sources kicked off a synchronous rebuild of those tables.  I killed the rebuild for tests before it executed, but the episode table got wiped.

When the dynamic data sources get deleted, we should be careful to make sure that we don't accidentally delete the tables.

PR #12704, opened 2016-08-03
Use dynamic variables in PreFilter

kaapstorm:
As demoed yesterday.

The meat of this PR is in ~~the last commit~~ commits 93563bb and 9cd5109.

@NoahCarnahan @czue cc @sheelio

PR #17736, opened 2017-09-07
Add documentation for pre_operator

emord:
@kaapstorm buddy @sravfeyn

### DIFF
--- a/corehq/apps/userreports/README.md
+++ b/corehq/apps/userreports/README.md
@@ -1218,6 +1218,20 @@ operator. e.g.
 (If `pre_value` is an array and `datatype` is not "array", it is assumed that
 `datatype` refers to the data type of the items in the array.)
 
+You can optionally specify the operator that the prevalue filter uses by adding a pre_operator argument. e.g.
+```
+{
+  "type": "pre",
+  "field": "at_risk_field",
+  "slug": "at_risk_slug",
+  "datatype": "array",
+  "pre_value": ["maybe", "yes"],
+  "pre_operator": "between"
+}
+```
+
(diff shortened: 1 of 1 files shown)


corehq/apps/userreports/README.md
```diff
@@ -1218,6 +1218,20 @@ operator. e.g.
...
+  "type": "pre",
+  "field": "at_risk_field",
+  "slug": "at_risk_slug",
+  "datatype": "array",
+  "pre_value": ["maybe", "yes"],
+  "pre_operator": "between"
+}
+```
+
+Note that instead of using `eq`, `gt`, etc, you will need to use `=`, `>`, etc.
```
emord: @kaapstorm from https://github.com/dimagi/commcare-hq/pull/12704/files it appears that we need to use `=` instead of `eq` like the rest of UCRs. any reason for that?

kaapstorm: If I remember correctly ... it seems we use the mapping from [`NumericFilterValue`](https://github.com/dimagi/commcare-hq/blob/master/corehq/apps/userreports/reports/filters/values.py#L140-L140) for scalar values so that PreFilters work the same way as normal [`ReportFilters`](https://github.com/dimagi/commcare-hq/blob/master/corehq/apps/userreports/reports/filters/specs.py#L35-L35). 

(Mapping operators lets us restrict the operators available, and use the same filter for both SQL and ES -- although that probably doesn't answer your question, and I'm sure you knew that already because you wrote some of that code :slightly_smiling_face: .) 

I like your point though. It would be nice if UCRs and ReportFilters looked the same.

emord: Ah. that makes sense. So it's a difference of report filters vs data source expressions. thanks

PR #17740, opened 2017-09-07
fix case_ids query

nickpell:
@sheelio cc: @proteusvacuum

### DIFF
--- a/custom/enikshay/management/commands/run_testing_facility_name_migration.py
+++ b/custom/enikshay/management/commands/run_testing_facility_name_migration.py
@@ -58,17 +58,19 @@ class Command(BaseCommand):
             "real" if commit else "fake", domain, datetime.datetime.utcnow()
         ))
 
-        cases = (CaseSearchES()
-                 .domain(domain)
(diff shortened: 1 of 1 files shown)


sheelio:
Looks good - was the ES query too large?

nickpell:
I'm not sure - I think something might have been off with the syntax

custom/enikshay/management/commands/run_testing_facility_name_migration.py
```diff
@@ -58,17 +58,19 @@ def handle(self, commit, domain, log_path, **options):
             "real" if commit else "fake", domain, datetime.datetime.utcnow()
         ))
 
-        cases = (CaseSearchES()
-                 .domain(domain)
-                 .case_type("test")
-                 .case_property_query("result_recorded", "yes", "must")
-                 .values_list('case_id', flat=True))
+        case_ids = [
+            hit['_id'] for hit in (CaseSearchES()
```
emord: You may be able to use `get_ids` or `source` to specify that you only want the ids returned

PR #17744, opened 2017-09-08
Retry errors for static UCRs

emord:
@proteusvacuum @snopoke buddy @sravfeyn https://manage.dimagi.com/default.asp?260368

I think this presents a happy medium. I don't think that anyone decided on a good way to retry all UCRs as there could have been a bad configuration or other issues in a dynamic UCR. We do have full control over static data sources though, so they should be safe to retry at a later time.


emord:
this is also one more reason to move all of enikshay's UCR into the repo

PR #17749, opened 2017-09-08
More robust aggregation

emord:
@sheelio @sravfeyn 

The first commit keeps the same functionality, but splits it over 5 different tasks. The second catches any database errors and retries the task after a timeout

### DIFF
--- a/custom/icds_reports/tasks.py
+++ b/custom/icds_reports/tasks.py
@@ -1,9 +1,10 @@
+from collections import namedtuple
 from datetime import datetime, timedelta
 import logging
 import os
 
 from celery.schedules import crontab
-from celery.task import periodic_task
+from celery.task import periodic_task, task
 from dateutil.relativedelta import relativedelta
 from django.conf import settings
 from django.db import connections
@@ -16,6 +17,8 @@ from corehq.util.decorators import serial_task
 from dimagi.utils.chunked import chunked
 
 celery_task_logger = logging.getLogger('celery.task')
+
+UCRAggregationTask = namedtuple("UCRAggregationTask", ['type', 'date'])
 
 
 @periodic_task(run_every=crontab(minute=0, hour=21), acks_late=True, queue='background_queue')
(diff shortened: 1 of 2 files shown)


emord:
@calellowitz would be good to get this deployed assuming tests pass. I kicked the build. I think it was an unrelated test failure

sheelio:
@emord 

Are these still run sequentially, even if its different tasks?

emord:
@sheelio yeah

they get appended to hte list in the same order they were running before and only one is run at one time https://github.com/dimagi/commcare-hq/pull/17749/files#diff-b8c5a5e308a203bcaee8ed83e6837edfR54

if there are any more tasks to be run it is queued after the previous task completes https://github.com/dimagi/commcare-hq/pull/17749/files#diff-b8c5a5e308a203bcaee8ed83e6837edfR105

This retry allows us to only retry the specific aggregation that fails https://github.com/dimagi/commcare-hq/pull/17749/files#diff-b8c5a5e308a203bcaee8ed83e6837edfR105

emord:
The main goal here is to recover gracefully whenever the database needs to go down  for config changes  or maintenance which unfortunately happens a lot on icds

emord:
@sheelio is offline this week and tests passed (they would fail if the ordering was wrong) @sravfeyn @calellowitz

sravfeyn (approved):

custom/icds_reports/tasks.py
```diff
@@ -44,27 +49,65 @@ def move_ucr_data_into_aggregation_tables(date=None, intervals=3):
...
 
-            path = os.path.join(os.path.dirname(__file__), 'sql_templates', 'update_monthly_aggregate_tables.sql')
+        aggregation_tasks = []
+
+        for interval in range(intervals - 1, -1, -1):
+            calculation_date = (monthly_date - relativedelta(months=interval)).strftime('%Y-%m-%d')
+            aggregation_tasks.append(UCRAggregationTask('monthly', calculation_date))
+
+        aggregation_tasks.append(UCRAggregationTask('daily', date.strftime('%Y-%m-%d')))
+        aggregate_tables.delay(aggregation_tasks[0], aggregation_tasks[1:])
```
sravfeyn: Is there a concern of aggregate data over daily vs monthly being inconsistent if one task fails but other succeeds?

emord: If it fails then https://github.com/dimagi/commcare-hq/pull/17749/files#diff-b8c5a5e308a203bcaee8ed83e6837edfR96 will raise an exception and it won't queue the following tasks, but it will retry the current tasks, so there should never be a time when one is queued and the previous one did not finish

snopoke:
Looks fine. Any reason not to let celery handle the sequention exection itself: http://docs.celeryproject.org/en/latest/userguide/canvas.html#map-starmap

emord:
@snopoke interesting I haven't seen that before. I think I would want to use http://docs.celeryproject.org/en/latest/userguide/canvas.html#chains because I'd want some error handling, but I need to read that page more closely to verify that

PR #17759, opened 2017-09-08
facility id migration

nickpell:
adds a management command to run ```EpisodeFacilityIDMigration```

also addresses a copy of edge cases around messy data

@proteusvacuum


proteusvacuum:
You could instead use this: https://github.com/dimagi/commcare-hq/blob/es/voucher-confirmations/custom/enikshay/management/commands/base.py 
It batches all the case submissions, and outputs the errors, etc. 
you use it like this: https://github.com/dimagi/commcare-hq/blob/es/voucher-confirmations/custom/enikshay/management/commands/set_redeemed_prescription_voucher_dates.py

I think the EpisodeFacilityIDMigration as the same API so should work out of the box.

nickpell:
thats really good to know - will follow that pattern in the future.

I've already run this migration, so mind merging it as is just for posterity's sake?

PR #17762, opened 2017-09-08
remove unnecessary fallback for ids that come in as indexes

biyeun:
wasn't really working properly in the beginning...likely due to a bad merge.

@orangejenny buddy @kaapstorm


orangejenny:
are we not doing the fallback?

biyeun:
@orangejenny the template containing the indexed module id no longer exists, and I think it's probably safe to assume that no one has left this page open enough to still have that version?

orangejenny:
fair

orangejenny (approved):

kaapstorm (approved):

PR #17764, opened 2017-09-09
also give count for last 24 hours in repeater report

mkangia:
Made this change to highlight differences to flag issues


proteusvacuum (approved):
sweet!

PR #17766, opened 2017-09-11
Add average to demographics section of Progress Report

kkrampa:


emord:
Why are the formats being added?

kkrampa:
Earlier all indicators that have been configured to have average were automatically converted to % values. After that change is no longer true, we also have non-percent indicators which need average so now it's explicitly defined which of them should be displayed as percents.

PR #17767, opened 2017-09-11
Remove user signal

emord:
https://sentry.io/dimagi/commcarehq/issues/223708154/

@esoergel @snopoke When a device id changes we save the user in the restore. This signal causes restores to be dependent on ES responding, so if an ES node goes down it will cause restores to fail.

from https://github.com/dimagi/commcare-hq/pull/4393 it seems like the only reason to have this is to have changes show up when pillowtop is not running. I'm not aware of a time when pillows should not be running for a long time, so I'm assuming this was just because you wanted local changes to show up immediately? If that's the case then I think we should remove this and let pillows handle this.

buddy @sravfeyn

### DIFF
--- a/corehq/apps/users/signals.py
+++ b/corehq/apps/users/signals.py
@@ -31,15 +31,6 @@ def django_user_post_save_signal(sender, instance, created, raw=False, **kwargs)
     return CouchUser.django_user_post_save_signal(sender, instance, created)
 
 
-def update_user_in_es(sender, couch_user, **kwargs):
-    """
-    Automatically sync the user to elastic directly on save or delete
-    """
-    from corehq.pillows.user import transform_user_for_elasticsearch
-    send_to_elasticsearch("users", transform_user_for_elasticsearch(couch_user.to_json()),
-                          delete=couch_user.to_be_deleted())
-
-
 def sync_user_phone_numbers(sender, couch_user, **kwargs):
     from corehq.apps.sms.tasks import sync_user_phone_numbers as sms_sync_user_phone_numbers
     sms_sync_user_phone_numbers.delay(couch_user.get_id)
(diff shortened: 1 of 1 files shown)


emord:
Looks like @dannyroberts was also in the PR

snopoke:
I think the reason for this signal is to make sure that edits to a user from the UI are shown immediately in the user list which is backed by ES.

Perhaps it would be better to make the signal firing optional? This would also be useful in reducing the number of updates to the user case.

emord:
Looks like that will get messy to get correct. It's currently in the generic save method and I don't see a way to view a "diff" of what changed to see if it's ok to not save to ES.

two thought:
1) if this saves all the time, what's the point of having the pillow? The pillow is fault tolerant and puts things in the pillow retry queue, but seems like we should just add that error handling here
2) At some point it's likely going to be worth biting the bullet to put users solely in postgres, so that we don't have this messy workflow where every user save is saving to 3 databases.

I think the dependency on ES is https://github.com/dimagi/commcare-hq/blob/master/corehq/apps/users/analytics.py#L68. All of that info is in django, but it's in their user model. If we used a custom user model we could use postgres's full text search https://docs.djangoproject.com/en/1.10/ref/contrib/postgres/search/#searchvector to remove that dependency

snopoke:
Is it not possible to add a kwarg to the `save` method: `fire_signals=True`
or something?

esoergel:
I think this was in a response to staleness on the user list page, as Simon stated - you'd change a user's name, then get dropped back on the user list page, and see their old name, because ES hasn't updated yet.  With forms and cases, there's rarely such a real-time update, so staleness matters less.  Your followup PR looks fine, and wouldn't affect the issue which triggered https://github.com/dimagi/commcare-hq/pull/4393

PR #17769, opened 2017-09-11
Add error and success status fields in report for monitoring

mkangia:


snopoke (approved):
I don't have any context but looks OK

emord (approved):

mkangia:
Thank You @snopoke and @emord 

@proteusvacuum do you think this would be a value addition and if yes could be done for public notification register as well?

proteusvacuum:
These fields aren't in the datasource as far as I can tell. I think this will break the report.

PR #17770, opened 2017-09-11
Faster UCR case rebuilding

emord:
The only thing missing here is a SQL function to filter by XMLNS, but this should be mergeable now and mostly better than what we have for cases.

@calellowitz likely useful for eniskahy rebuilds

buddy @sravfeyn

### DIFF
--- a/corehq/apps/userreports/document_stores.py
+++ b/corehq/apps/userreports/document_stores.py
@@ -6,12 +6,12 @@ from corehq.util.couch import get_db_by_doc_type
 from pillowtop.dao.couch import CouchDocumentStore
 
 
-def get_document_store(domain, doc_type):
+def get_document_store(domain, doc_type, case_type_or_xmlns=None):
     use_sql = should_use_sql_backend(domain)
     if use_sql and doc_type == 'XFormInstance':
-        return ReadonlyFormDocumentStore(domain)
+        return ReadonlyFormDocumentStore(domain, xmlns=case_type_or_xmlns)
     elif use_sql and doc_type == 'CommCareCase':
-        return ReadonlyCaseDocumentStore(domain)
+        return ReadonlyCaseDocumentStore(domain, case_type=case_type_or_xmlns)
     elif doc_type == LOCATION_DOC_TYPE:
         return ReadonlyLocationDocumentStore(domain)
     else:
--- a/corehq/apps/userreports/document_stores.py
+++ b/corehq/apps/userreports/document_stores.py
@@ -10,7 +10,7 @@ def get_document_store(domain, doc_type, case_type_or_xmlns=None):
     use_sql = should_use_sql_backend(domain)
     if use_sql and doc_type == 'XFormInstance':
         return ReadonlyFormDocumentStore(domain, xmlns=case_type_or_xmlns)
-    elif use_sql and doc_type == 'CommCareCase':
+    elif doc_type == 'CommCareCase':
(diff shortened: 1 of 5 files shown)


corehq/apps/userreports/document_stores.py
```diff
@@ -6,12 +6,12 @@
...
 
-def get_document_store(domain, doc_type):
+def get_document_store(domain, doc_type, case_type_or_xmlns=None):
     use_sql = should_use_sql_backend(domain)
     if use_sql and doc_type == 'XFormInstance':
-        return ReadonlyFormDocumentStore(domain)
-    elif use_sql and doc_type == 'CommCareCase':
-        return ReadonlyCaseDocumentStore(domain)
+        return ReadonlyFormDocumentStore(domain, xmlns=case_type_or_xmlns)
+    elif doc_type == 'CommCareCase':
```
emord: At the moment the SQL based document stores do not support last_id. last_id is used to resume builds, but I think filtering will be much more useful for couch based document stores so removing the check from here

corehq/apps/userreports/models.py
```diff
@@ -350,6 +350,27 @@ def get_es_index_settings(self):
         es_index_settings.pop('doc_type')
         return {"settings": es_index_settings}
 
+    def get_case_type_or_xmlns_filter(self):
+        def _get_property_value(config_filter, prop_name):
+            if (config_filter.get('type') != 'boolean_expression'
+                    or config_filter['operator'] != 'eq'):
+                return None
+            expression = config_filter['expression']
+            if expression['type'] == 'property_name' and expression['property_name'] == prop_name:
```
sravfeyn: Does this rely on the fact that there will only be one boolean filter that checks `property_name` to a value? What if there is an additional filter that's not case_type/xmlns filter?

emord: If it's a more complex filter, it should return `None` and it will not try to filter anything so it would go through all forms or cases (as it does today).

corehq/form_processor/document_stores.py
```diff
@@ -46,7 +49,7 @@ def get_document(self, doc_id):
 
     def iter_document_ids(self, last_id=None):
         # todo: support last_id
-        return iter(self.case_accessors.get_case_ids_in_domain())
+        return iter(self.case_accessors.get_case_ids_in_domain(type=self.case_type))
```
snopoke: not a comment on this PR but getting ALL ids for ICDS / eNikshay doesn't seem like a good idea

snopoke (approved):

corehq/form_processor/document_stores.py
emord: Indeed it does not. which is why i created the async_rebuild_table management command to get around this. This should at least unblock most rebuilds on cases. only ones I would be concerened about at the moment is adherence for enikshay and person for icds

snopoke: could do something like this too: https://github.com/dimagi/commcare-hq/blob/14b16a17c964ca52e1b44937286385833b219e8f/corehq/ex-submodules/pillowtop/reindexer/change_providers/case.py#L30

emord: Actually I was considering changing the databases definition to do that exact same thing. Weren't we planning on doing all the partitioning in python instead of plproxy? Could at least create a different implementation to return an iterator instead of a list

snopoke:
Yea, that's what I was thinking of.

PR #17772, opened 2017-09-11
conditional fire signal

emord:
https://github.com/dimagi/commcare-hq/pull/17767#issuecomment-328481655

@snopoke for some reason I was only thinking about it the opposite way where it was always false and i'd need to fine where it should be true (though I still think it should get moved to postgres since all of the info is already there and it doesn't require any weird workflows)

buddy @sravfeyn

### DIFF
--- a/corehq/apps/ota/utils.py
+++ b/corehq/apps/ota/utils.py
@@ -212,4 +212,4 @@ def update_device_id(user, device_id):
         if not user.is_demo_user:
             updated = user.update_device_id_last_used(device_id)
             if updated:
-                user.save()
+                user.save(fire_signals=False)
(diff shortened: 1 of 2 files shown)


snopoke (changes requested):
I think you'll need to pass this in from `CommCareUser.save`.

snopoke:
+1 to moving to PG. Are you thinking of just dumping the JSON in a TEXT column?

emord:
> +1 to moving to PG. Are you thinking of just dumping the JSON in a TEXT column?

Actually I wasn't even meaning migrating the whole thing right now (although I think that would be nice to have at some point). for this I think we could just migrate away from django's base user class [0] and add some indexes to enable full text search in django. That would remove the ES dependency on the users list page.

If we were moving the entire document i'd probably dump the json into a jsonb field with indexes so that other access still works. would also keep have it as a foreign key to make sure we're not loading some huge record all the time when we don't need it

[0] kind of documented here: https://code.djangoproject.com/ticket/25313#comment:2

snopoke (approved):

PR #17773, opened 2017-09-11
Revert "Add error and success status fields in report for monitoring"

proteusvacuum:
Reverts dimagi/commcare-hq#17769

@mkangia @emord 
Those properties were not in the datasource, so this will likely break that report.

### DIFF
--- a/custom/enikshay/ucr/reports/tb_notification_register_private.json
+++ b/custom/enikshay/ucr/reports/tb_notification_register_private.json
@@ -706,39 +706,6 @@
                 "calculate_total": false,
                 "type": "field",
                 "display": "Treatment Supporter Details - Designation"
-            },
-            {
-                "sortable": false,
-                "description": null,
-                "format": "default",
-                "transform": {},
-                "aggregation": "simple",
-                "field": "private_nikshay_registered",
-                "calculate_total": false,
-                "type": "field",
-                "display": "Nikshay Registered"
-            },
-            {
-                "sortable": false,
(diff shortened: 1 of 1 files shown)


proteusvacuum:
merging since no tests

mkangia:
Thank you @proteusvacuum. This would have triggered a rebuild right? or to make this work we would need to update data source and rebuild instead? Obviously not doing that without proper planning.

proteusvacuum:
You need to add those properties to the "datasource" file. The current process is:
- Make a new datasource with a new name ("e.g. v2")
- Deploy that. Wait until the new datasource has finished building
- Make changes to the report to point to the new datasource, with your added columns.
- Delete the old datasource.

If you wanted to, you could also make a "QA" version of the report that is hidden, then switch to the new report.

emord:
If you're adding new columns you can add those columns to the data source and set `is_nullable: True` in the column. That will be able to add the columns without deleting any of the existing data.

mkangia:
Oh nice! Thanks @proteusvacuum  and @emord. 
But how can one confirm that that column has been added for all rows? It would be added when the column is fetched?

emord:
It's in SQL so it's added to the schema. you'd still need to kick off a rebuild with the async_rebuild_table management command, but this rebuild shouldn't cause any loss in data.

mkangia:
Added them here to data source,
https://github.com/dimagi/commcare-hq/pull/17786

PR #17777, opened 2017-09-11
Ce/app status sorting

calellowitz:
@millerdev @nickpell


corehq/apps/reports/standard/deployments.py
```diff
@@ -88,6 +89,14 @@ def default_sort(self):
         else:
             return {'reporting_metadata.last_submission_for_user.submission_date': 'desc'}
 
+    @property
+    def sort_base(self):
+        return '.'.join(self.primary_sort_prop.split('.')[:2])
```
calellowitz: if there is a better way to do this, please let me know, because this looks gross

nickpell (approved):

PR #17779, opened 2017-09-11
'Menu' not 'Module' when creating form export

orangejenny:
https://manage.dimagi.com/default.asp?261542

Was expecting this to be a one-line change [here](https://github.com/dimagi/commcare-hq/blob/master/corehq/apps/export/forms.py#L181), but that had no effect. This is verbose but works.

@gcapalbo / @dannyroberts

### DIFF
--- a/corehq/apps/app_manager/fields.py
+++ b/corehq/apps/app_manager/fields.py
@@ -160,8 +160,8 @@ class ApplicationDataRMIHelper(object):
     APP_TYPE_NONE = 'no_app'
     APP_TYPE_UNKNOWN = 'unknown'
 
-    def __init__(self, domain, user, as_dict=True, form_labels=None,
-                 form_placeholders=None, case_placeholders=None):
+    def __init__(self, domain, user, as_dict=True, form_placeholders=None,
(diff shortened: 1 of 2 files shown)


corehq/apps/app_manager/fields.py
```diff
@@ -160,10 +160,17 @@ class ApplicationDataRMIHelper(object):
     APP_TYPE_NONE = 'no_app'
     APP_TYPE_UNKNOWN = 'unknown'
 
-    def __init__(self, domain, user, as_dict=True, form_placeholders=None, case_placeholders=None):
+    def __init__(self, domain, user, as_dict=True, form_labels=None,
```
gcapalbo: any reason to add `form_labels` in the middle here instead of at the end? usually it's safer to add a new kwarg to the end unless you know all everything that calls this doesn't use positional args

gcapalbo (approved):

corehq/apps/app_manager/fields.py
orangejenny: makes sense, updated

gcapalbo (approved):

PR #17780, opened 2017-09-11
add script to update user contact_phone_number

nickpell:
Should fix the missing mobile phone number issue.  Only writes to ```contact_phone_number``` if it's blank.

@ctsims @proteusvacuum

### DIFF
--- a/custom/enikshay/private_sector_datamigration/management/commands/update_user_contact_phone_number.py
+++ b/custom/enikshay/private_sector_datamigration/management/commands/update_user_contact_phone_number.py
@@ -26,6 +26,8 @@ class Command(BaseCommand):
         if location_ids:
             locs_matching_type = locs_matching_type.filter(location_id__in=location_ids)
 
+        count = 0
+
         for _, loc in enumerate(with_progress_bar(locs_matching_type)):
             private_sector_agency_id = loc.metadata.get('private_sector_agency_id', '')
             if not private_sector_agency_id.isdigit():
@@ -43,3 +45,7 @@ class Command(BaseCommand):
             if not contact_phone_number and len(user_detail.mobileNumber) == 10:
                 location_user.user_data['contact_phone_number'] = '91' + user_detail.mobileNumber
                 location_user.save()
+                print 'Updated %s' % location_user.user_id
+                count += 1
+
+        print '%d users updated.' % count
--- a/custom/enikshay/private_sector_datamigration/management/commands/update_user_contact_phone_number.py
+++ b/custom/enikshay/private_sector_datamigration/management/commands/update_user_contact_phone_number.py
@@ -1,0 +1,45 @@
+from django.core.management import BaseCommand
+
+from corehq.apps.locations.models import SQLLocation
+from corehq.apps.users.models import CommCareUser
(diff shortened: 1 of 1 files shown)


custom/enikshay/private_sector_datamigration/management/commands/update_user_contact_phone_number.py
```diff
@@ -0,0 +1,45 @@
...
+
+            try:
+                user_detail = UserDetail.objects.get(agencyId=agency_id, isPrimary=True)
+            except UserDetail.DoesNotExist:
+                continue
+
+            location_user = CommCareUser.get_by_user_id(loc.user_id)
+
+            contact_phone_number = location_user.user_data.get('contact_phone_number')
+            if not contact_phone_number and len(user_detail.mobileNumber) == 10:
```
gcapalbo: are the phone numbers on the user detail already cleaned? otherwise we should probably clean any non digit characters before checking for length.

also we can handle when the source number is 11 digits long and it starts with a leading zero by just removing the leading zero.

and do we know that these are always local numbers, or could some of them already be 12 digits with 91 in front?

nickpell: > are the phone numbers on the user detail already cleaned?

yes

> also we can handle when the source number is 11 digits long and it starts with a leading zero by just removing the leading zero.

All phone numbers here have either 10 or 6 digits.  I'm skipping the ones with 6 digits.

custom/enikshay/private_sector_datamigration/management/commands/update_user_contact_phone_number.py
```diff
@@ -0,0 +1,45 @@
...
+                location_user.user_data['contact_phone_number'] = '91' + user_detail.mobileNumber
+                location_user.save()
```
gcapalbo: should this be keeping track of how many get updated just to verify that the update was applied correctly?

gcapalbo:
@nickpell also, see my note in the email thread about language_code. did we already migrate language_code, and if not is preferred language an available data point?

custom/enikshay/private_sector_datamigration/management/commands/update_user_contact_phone_number.py
nickpell: updated

nickpell:
@gcapalbo We don't have a preferred language for agencies in UATBC

custom/enikshay/private_sector_datamigration/management/commands/update_user_contact_phone_number.py
```diff
@@ -43,3 +45,7 @@ def handle(self, domain, location_ids, **options):
             if not contact_phone_number and len(user_detail.mobileNumber) == 10:
                 location_user.user_data['contact_phone_number'] = '91' + user_detail.mobileNumber
                 location_user.save()
+                print 'Updated %s' % location_user.user_id
```
gcapalbo: printing within the loop might mess with the progress bar output, not sure if that's important or not

gcapalbo (approved):

PR #17782, opened 2017-09-11
Ce/ben reg v2

calellowitz:
@proteusvacuum 
Wasnt really sure from the spec what you wanted. Is this correct? If so, we can roll it out once rebuild finishes


millerdev (approved):
don't have a lot of context, so take that into account when considering my review

proteusvacuum (approved):
good to merge when datasource has finished rebuilding

PR #17783, opened 2017-09-11
Map answer concepts, support form questions

kaapstorm:
This change allows MOTECH to map multiple choice answers in CommCare to concepts in OpenMRS, and to forward form questions that are not case properties. 

(`send_openmrs_data` still assumes that it has been able to identify a patient.)

Might be easier to :tropical_fish: 

@dannyroberts cc @snopoke buddy @biyeun

### DIFF
--- a/corehq/motech/openmrs/handler.py
+++ b/corehq/motech/openmrs/handler.py
@@ -1,7 +1,13 @@
 from corehq.motech.openmrs.logger import logger
 from corehq.motech.openmrs.openmrs_config import IdMatcher
-from corehq.motech.openmrs.repeater_helpers import CaseTriggerInfo, get_patient_by_id, \
-    update_person_attribute, create_person_attribute, create_visit, set_person_properties
+from corehq.motech.openmrs.repeater_helpers import (
+    CaseTriggerInfo,
+    get_patient_by_id,
+    update_person_attribute,
+    create_person_attribute,
+    create_visit,
+    set_person_properties,
+)
 from dimagi.utils.parsing import string_to_utc_datetime
 
 
--- a/corehq/motech/openmrs/handler.py
+++ b/corehq/motech/openmrs/handler.py
@@ -5,7 +5,7 @@ from corehq.motech.openmrs.repeater_helpers import CaseTriggerInfo, get_patient_
 from dimagi.utils.parsing import string_to_utc_datetime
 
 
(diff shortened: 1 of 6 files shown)


snopoke:
Looks OK to me. Build failing though.

corehq/motech/openmrs/repeater_helpers.py
```diff
@@ -166,7 +166,40 @@ def get_patient_matching_identifiers(self, patient_identifier_type, patient_iden
...
+            elif isinstance(value, dict):
+                # Group
+                _recurse_form_questions(value, new_path, result_)
+            else:
+                # key is a question and value is its answer
+                question = '/'.join(new_path)
+                result_[question] = value
+
+    result = {}
+    _recurse_form_questions(form_json['form'], ['/data'], result)
```
dannyroberts: `'data'` might technically come from `form_json['form']['type']` or something, but form builder always calls it data. I think fine to leave, maybe worth leaving a comment, though even that may be overkill. Maybe no change required, that just came to mind.

corehq/motech/openmrs/handler.py
```diff
@@ -35,7 +36,7 @@ def send_openmrs_visit(requests, info, form_config, person_uuid, visit_datetime)
         visit_datetime=visit_datetime,
         values_for_concept={obs.concept: [obs.value.get_value(info)]
                             for obs in form_config.openmrs_observations
-                            if obs.value.get_value(info)},
+                            if obs.value.get_value(info)},  # TODO: form_question_values
```
dannyroberts: What's this todo? Naively it seems like form_question_values is already passed in with `info`.

dannyroberts (approved):

corehq/motech/openmrs/handler.py
kaapstorm: Thanks for pointing it out. It's a stale comment. Removed.

corehq/motech/openmrs/repeater_helpers.py
kaapstorm: Good call. Comment added.

kaapstorm:
I've added 2 commits since you reviewed @dannyroberts . If you're still happy, then I think we're good to go.

dannyroberts (approved):

PR #17786, opened 2017-09-12
Update data source to async add nikshay integration fields

mkangia:


mkangia:
@emord i just used the expression for one of the fields and changed the property name attributes. 
Need to be very careful of this not causing a rebuild. Don't know enough about UCRs so requesting a close review.

emord:
Looks fine to me, but I thought there was a moratorium on data source changes? has that been revoked? may want to add the v2 just in case that's still in effect

mkangia:
> Looks fine to me, but I thought there was a moratorium on data source changes? has that been revoked? may want to add the v2 just in case that's still in effect

To be safe, making a new data source instead.
1. So copied over to a new source v2 along with changes done in [first commit](https://github.com/dimagi/commcare-hq/pull/17786/commits/c2ca0e78c37ff68466ab27f82fb5dcaffe05016c)
and reverted changes to existing report
2. made changes to v2 to avoid conflicting with existing report in [commit](https://github.com/dimagi/commcare-hq/pull/17786/commits/6c3f2ea7dfb2a28f9afcf361d5fbfaa0b3943740).

Is there any other change needed to avoid conflicting with existing data source?

@emord

emord (approved):

nickpell (approved):

PR #17797, opened 2017-09-12
Show list of reports a datasource is used by

proteusvacuum:
@calellowitz @nickpell


corehq/apps/userreports/views.py
```diff
@@ -1125,6 +1126,14 @@ def post(self, request, *args, **kwargs):
                 )
         return self.get(request, *args, **kwargs)
 
+    def get_reports(self):
+        if self.read_only:
```
nickpell: It's possible for nonstatic reports to reference a static data source, and technically possible (although rare) for a static report to refernece a nonstatic data source.  We probably want to check nonstatic and static report configurations regardless of where the data source is static or not.

nickpell (approved):

emord:
@proteusvacuum this broke viewing the stati c data sources on enikshay and ICDS. can you look into this https://sentry.io/dimagi/commcarehq/issues/349554812/events/7209856996/

PR #17802, opened 2017-09-12
adding gtm to base template

calellowitz:
@nickpell @millerdev

### DIFF
--- a/corehq/apps/hqwebapp/templates/hqwebapp/base.html
+++ b/corehq/apps/hqwebapp/templates/hqwebapp/base.html
@@ -2,6 +2,14 @@
 {% get_current_language as LANGUAGE_CODE %}
 <html lang="{{ LANGUAGE_CODE }}">
     <head>
+        <!-- Google Tag Manager -->
+        <script>(function(w,d,s,l,i){w[l]=w[l]||[];w[l].push({'gtm.start':
+        new Date().getTime(),event:'gtm.js'});var f=d.getElementsByTagName(s)[0],
+        j=d.createElement(s),dl=l!='dataLayer'?'&l='+l:'';j.async=true;j.src=
+        'https://www.googletagmanager.com/gtm.js?id='+i+dl;f.parentNode.insertBefore(j,f);
+        })(window,document,'script','dataLayer','GTM-52MT7FS');</script>
+        <!-- End Google Tag Manager -->
+
         {% captureas title_block %}{% block title %}{% endblock title %}{% endcaptureas %}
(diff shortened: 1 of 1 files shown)


corehq/apps/hqwebapp/templates/hqwebapp/base.html
```diff
@@ -228,6 +236,10 @@
         {% endif %}
     </head>
     <body>
+        <!-- Google Tag Manager (noscript) -->
+        <noscript><iframe src="https://www.googletagmanager.com/ns.html?id=GTM-52MT7FS"
```
orangejenny: Maybe not necessary, given how dysfunctional HQ is if you don't have scripting enabled.

orangejenny (approved):

millerdev (commented):
If I'm reading the description of GTM correctly, it seems like we could eliminate some code here too. Possibly https://github.com/dimagi/commcare-hq/blob/986584382da8efa1eb43a32a01205e14bf5e531e/corehq/apps/hqwebapp/templates/hqwebapp/includes/analytics_all.html#L6
Is that part of the plan?

calellowitz:
@millerdev Where are you reading that? As far as I know that is not part of the plan. This is being added for some sales test that the marketing team is running. But my understanding was that this supplements google analytics, not replaces it

millerdev:
[Google Tag Manager overview](https://support.google.com/tagmanager/answer/6102821?hl=en)

> Google Tag Manager is a tag management system that allows you to quickly and easily update tags and code snippets on your website or mobile app, such as those intended for traffic analysis and marketing optimization. You can add and update AdWords, Google Analytics, Firebase Analytics, Floodlight, and 3rd party or custom tags from the Tag Manager user interface instead of editing site code.

calellowitz:
Yeah, I read that as you can use google tag manager to do "tags" (whatever that is) in google analytics and other analytics platforms. It adds that new feature, but does not replace or seek to replicate any of the other functionality GA provides.

millerdev:
> "tags" (whatever that is)

I was, perhaps incorrectly, assuming this means GTM allows you to manage (that is, a GTM user could add or remove) various other analytics snippets on the page. Normally these things are added in the form of a `<script>` "tag", which is what made me think that. In this case it sounds like we're at least not using it in that capacity yet, so fine as is.

calellowitz:
Ah. I totally misinterpreted that. I think you may be correct. I was picturing tags as labels you attach to events. I can talk to carrie about moving in that direction longer term but its not on the roadmap for now.

PR #17806, opened 2017-09-12
Only send Users and Locations to BETS when pertinent

proteusvacuum:
@dannyroberts this is what we discussed yesterday

@nickpell 
FYI @esoergel @mkangia 

🐡

### DIFF
--- a/corehq/motech/repeaters/models.py
+++ b/corehq/motech/repeaters/models.py
@@ -132,6 +132,9 @@ class Repeater(QuickCachedDocumentMixin, Document, UnicodeMixIn):
     def get_payload(self, repeat_record):
         return self.generator.get_payload(repeat_record, self.payload_doc(repeat_record))
 
+    def get_attempt_info(self, repeat_record):
+        return None
+
     def register(self, payload, next_check=None):
         if not self.allowed_to_forward(payload):
             return
@@ -440,7 +443,7 @@ class RepeatRecordAttempt(DocumentSchema):
     success_response = StringProperty()
     next_check = DateTimeProperty()
     succeeded = BooleanProperty(default=False)
-    payload = StringProperty()
+    info = StringProperty()     # extra information about this attempt
 
     @property
     def message(self):
@@ -573,6 +576,9 @@ class RepeatRecord(Document):
     def get_payload(self):
         return self.repeater.get_payload(self)
 
+    def get_attempt_info(self):
+        return self.repeater.get_attempt_info(self)
+
     def handle_payload_exception(self, exception):
         now = datetime.utcnow()
         return RepeatRecordAttempt(
@@ -622,7 +628,7 @@ class RepeatRecord(Document):
             success_response=self._format_response(response) if response else None,
             next_check=None,
             succeeded=True,
-            payload=self.get_payload(),
+            info=self.get_attempt_info(),
         )
 
     def handle_failure(self, response):
@@ -650,7 +656,7 @@ class RepeatRecord(Document):
                 success_response=None,
                 next_check=None,
                 succeeded=False,
-                payload=self.get_payload(),
+                info=self.get_attempt_info(),
             )
 
     def cancel(self):
--- a/corehq/motech/repeaters/models.py
+++ b/corehq/motech/repeaters/models.py
@@ -128,6 +128,7 @@ class Repeater(QuickCachedDocumentMixin, Document, UnicodeMixIn):
     def payload_doc(self, repeat_record):
         raise NotImplementedError
 
+    @memoized
     def get_payload(self, repeat_record):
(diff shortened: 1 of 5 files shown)


esoergel (commented):
This is great!  I haven't done a thorough review, but the high-level approach looks solid, and is consistent with what we've previously discussed.  Thanks for getting this out!

custom/enikshay/integrations/bets/repeaters.py
```diff
@@ -411,9 +427,25 @@ class BETSLocationRepeater(BETSRepeaterMixin, LocationRepeater):
...
     def allowed_to_forward(self, location):
+        # if this location is already in the repeater queue, don't forward again
+        if queued_payload(location.domain, location.location_id):
+            return False
+
+        # If this location has already been forwarded without any changes, then
+        # don't send it to BETS again
+        successful_records = location.metadata.get('BETS_location_repeat_record_ids')
+        if successful_records:
+            latest_record_id = successful_records.split(" ")[-1]
```
mkangia: Whats the reason for keeping all attempts and not just the last successful record?

corehq/motech/repeaters/models.py
```diff
@@ -620,6 +622,7 @@ def handle_success(self, response):
             success_response=self._format_response(response) if response else None,
             next_check=None,
             succeeded=True,
+            payload=self.get_payload(),
```
mkangia: Are we doing this for all repeaters then? This would add quite some load on DB. Can we restrict to only those repeaters where this is needed?

mkangia (changes requested):
Functionality wise looks good but just couple of things to flag

custom/enikshay/integrations/bets/repeaters.py
proteusvacuum: I guess we could just save the last one. I figure this will make it easier to pull all the history from the DB if ever we are trying to debug why certain records are sending incorrectly or something, but maybe thats a YAGNI.

corehq/motech/repeaters/models.py
dannyroberts: I have the same conservative instinct here, @proteusvacuum do you think we're making a fuss over something pretty small? I could see storing all form xml/json in duplicate inside repeat records in couch being a non-trivial hit

proteusvacuum: @mkangia @dannyroberts I've turned this into an `info` field on the `RepeatRecordAttempt` which each Repeater is now free to use as they wish. I've currently only enabled it for the repeaters that we need this functionality for. Does that seem like a reasonable tradeoff?

custom/enikshay/integrations/bets/repeaters.py
```diff
@@ -437,10 +456,14 @@ def allowed_to_forward(self, location):
...
-            if current_payload == previous_payload:
-                return False
+            previous_payload = latest_record.attempts[-1].info if latest_record.attempts else "{}"
+            try:
+                previous_payload_json = json.loads(previous_payload)
+                current_payload = get_bets_location_json(location)
+                if current_payload == previous_payload_json:
+                    return False
+            except (TypeError, ValueError):
+                pass
```
calellowitz: this looks identical to https://github.com/dimagi/commcare-hq/pull/17806/commits/f3fca737a963d179b3b65616d7b5048eb3670bb4#diff-93d85770198e309ac7e7506062ffecccR421
any reason it cant be split out into a function

calellowitz (commented):
dont have a lot of context but seems fine only one question

custom/enikshay/integrations/bets/repeaters.py
proteusvacuum: They get the `current_payload` in a different way, but yes I probably could have done it better.

proteusvacuum:
Merging this now, if there are still concerns I can address them in a followup.

custom/enikshay/integrations/bets/repeaters.py
mkangia: Ya. One can also use `get_repeat_records_by_payload_id` and then their attempts to get all attempts when needed to pull history.

corehq/motech/repeaters/models.py
mkangia: Sounds good.

nit: i think we can just rename this to `payload_info` since its payload anyway.

custom/enikshay/integrations/bets/repeaters.py
proteusvacuum: yeah, this way allows you to do it without having to start a shell session (using raw_doc).

mkangia (commented):
Looks good.
Thanks for addressing the changes.

PR #17808, opened 2017-09-13
default user index to 0 replicas

snopoke:


emord (approved):
Eventually I want to change this to use index templates. I think that will be much more maintainable

PR #17816, opened 2017-09-13
Pick unknown value to avoid failure due to missing mandatory value

mkangia:
Note to self: Trigger cases failed under HIV notification repeater due to this.


mkangia:
@proteusvacuum seems like a safe change to merging this now and retriggering cases as we got some issues raised from users for this. Let me know if there are any review changes and i will take care of them.

proteusvacuum:
:+1:

PR #17817, opened 2017-09-13
zapier new/update case support

czue:
this implements changes necessary for https://manage.dimagi.com/default.asp?247716

product / @nsclark asked me to look into this and it turned out to be pretty straightforward so just went ahead and implemented it.

@snopoke and @proteusvacuum you looked at the ticket a long time ago, but in summary this adds two new repeater classes `CreateCaseRepeater` and `UpdateCaseRepeater` which only fire when a case is first created (or the opposite). 

I'm checking for a newly created case by seeing if `xform_ids` has length 1, which, while seemingly hacky I think is actually one of the more consistent ways to do it.

@proteusvacuum you had suggested using `allowed_to_forward` for this but I opted not to because I wanted to avoid creating a bunch of records in the DB that would never be allowed to forward. so instead this handles it by running the logic in the signal.

I'll throw a couple comments in a review now where I think there is possibly controversial changes. can be reviewed by commit or all at once (all at once probably a little easier).

The only user-facing side of this change is adding the following two options to the repeater page:

image

We could also hide them, though this has the downside of then making it tricker to see your config in HQ when you are using the Zapier integration.


corehq/apps/zapier/consts.py
```diff
@@ -1,3 +1,15 @@
+from corehq.motech.repeaters.models import UpdateCaseRepeater, CreateCaseRepeater, CaseRepeater
+
+
 class EventTypes(object):
     NEW_FORM = 'new_form'
     NEW_CASE = 'new_case'
+    UPDATE_CASE = 'update_case'
+    CHANGED_CASE = 'changed_case'  # includes both new and updated
```
czue: Don't love these two names. Couldn't think of a good way to differentiate including the creation as an update and not. Not sure how important it is as long as it's clear in code.

corehq/apps/zapier/tests/test_utils.py
```diff
@@ -0,0 +1,37 @@
+from collections import namedtuple
```
czue: most of this stuff was just extracted from setup/teardown of a test so I could use it in multiple places

corehq/apps/zapier/tests/test_zapier_hooks.py
```diff
@@ -0,0 +1,220 @@
+import json
```
czue: mostly must moved, but some small changes

corehq/motech/repeaters/models.py
```diff
@@ -361,6 +361,28 @@ def __unicode__(self):
         return "forwarding cases to: %s" % self.url
 
 
+class CreateCaseRepeater(CaseRepeater):
```
czue: debated whether to stick these in `zapier` instead of in the `motech.repeaters`. it made slightly more sense to put them here so I could piggyback off the `create_case_repeat_records` signal (and because they felt generic) but I'm happy to move them out if they feel custom.

corehq/motech/repeaters/signals.py
```diff
@@ -19,6 +20,11 @@ def create_form_repeat_records(sender, xform, **kwargs):
 def create_case_repeat_records(sender, case, **kwargs):
     from corehq.motech.repeaters.models import CaseRepeater
     create_repeat_records(CaseRepeater, case)
+    # assume if there's exactly 1 xform_id that modified the case it's being created
```
czue: mentioned in the PR description but this is arguably a hacky check, but am not sure there are better ones. @proteusvacuum had mentioned checking the last action, but I don't think that's actually right since you could (in theory - if there was a bug on mobile?) have a create action after a case was already made?

corehq/motech/repeaters/models.py
```diff
@@ -361,6 +361,28 @@ def __unicode__(self):
...
+    """
+    Just like CaseRepeater but only create records if the case is being created.
+    Used by the Zapier integration.
+    """
+    # note: the logic for how this is managed is handled in signals.create_case_repeat_records
+    # so this class actually does nothing except exist to be able to differentiate the two
+    # at the repeater and repeat record level. otherwise it's exactly the same as CaseRepeater
+    friendly_name = _("Forward Cases on Creation Only")
+    pass
```
millerdev: nit: unnecessary pass. same below

millerdev (commented):
Welcome back @czue!

I reviewed just to see what you were up to, but don't have a lot of context here, so probably best to have someone who's more familiar with this code approve it.

czue:
thanks @millerdev! look for more official news coming through soon.

zapier has been a sabbatical side project of mine and has been fun to be able to help shepherd it out.

good catch on the review, will fix those now

corehq/apps/zapier/tests/test_zapier_forwarding.py
```diff
@@ -0,0 +1,151 @@
...
+        cls.domain_object.delete()
+        cleanup_repeaters_for_domain(cls.domain)
+        super(TestZapierCaseForwarding, cls).tearDownClass()
+
+    def tearDown(self):
+        cleanup_repeat_records_for_domain(self.domain)
+        ZapierSubscription.objects.all().delete()
+
+    @run_with_all_backends
+    def test_create_case_forwarding(self):
```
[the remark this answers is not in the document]
czue: (fixed)

corehq/apps/zapier/tests/test_utils.py
```diff
@@ -20,3 +21,10 @@ def bootrap_domain_for_zapier(domain_name):
     web_user = WebUser.create(domain_name, 'test', '******')
     api_key_object, _ = ApiKey.objects.get_or_create(user=web_user.get_django_user())
     return ZapierDomainConfig(domain_object, web_user, api_key_object.key)
+
+
+def cleanup_repeaters_for_domain(domain):
+    for repeater in FormRepeater.by_domain(domain):
+            repeater.delete()
+    for repeater in CaseRepeater.by_domain(domain):
+        repeater.delete()
```
snopoke: There is already a `delete_all_repeaters` function which could be parameterized with the domain

snopoke (approved):
Looks great, thanks for the useful review comments.

proteusvacuum:
Hey, I'm not sure I agree with this part:

> @proteusvacuum you had suggested using allowed_to_forward for this but I opted not to because I wanted to avoid creating a bunch of records in the DB that would never be allowed to forward. so instead this handles it by running the logic in the signal.

The `allowed_to_forward` method is exactly the thing that checks whether it should add a RepeatRecord to the db or not: https://github.com/dimagi/commcare-hq/blob/master/corehq/motech/repeaters/models.py#L135-L147

This is called by the `create_case_repeat_records` method that you have edited: https://github.com/dimagi/commcare-hq/blob/master/corehq/motech/repeaters/signals.py#L30-L35

Maybe I've missed something though. 

I think it would be better to move the logic into the `allowed_to_forward` method, then you don't need the comment you've added in those new classes, and all `Repeater` classes then have a central place for that logic.

czue:
whoops, thanks for the comment. not sure where I got it in my head that was something that happened after the doc had been created but I can see you're totally right. 

https://github.com/dimagi/commcare-hq/pull/17841

PR #17822, opened 2017-09-13
Avoid decimal age

mkangia:
Note to self: Once merged, retrigger cases failing with `Character are   not allowed.......`


mkangia:
@proteusvacuum seems like a safe change to merging this now and retriggering cases as we got some issues raised from users for this. Let me know if there are any review changes and i will take care of them.

proteusvacuum:
👍

PR #17827, opened 2017-09-14
Fix ucr edit error

emord:
@proteusvacuum @nickpell @calellowitz 

This should fix the issue where we can't view data sources on enikshay or icds. Looks like some reports point to deleted data sources which caused this

Introduced https://github.com/dimagi/commcare-hq/pull/17797
sentry https://sentry.io/dimagi/commcarehq/issues/349554812/events/7209914929/

buddy @sravfeyn

### DIFF
--- a/corehq/apps/userreports/views.py
+++ b/corehq/apps/userreports/views.py
@@ -116,6 +116,7 @@ from corehq.apps.userreports.reports.util import has_location_filter
 from corehq.apps.users.decorators import require_permission
 from corehq.apps.users.models import Permissions
 from corehq.util.couch import get_document_or_404
+from corehq.util.soft_assert import soft_assert
 from pillowtop.dao.exceptions import DocumentNotFoundError
 import six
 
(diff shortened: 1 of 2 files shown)


corehq/apps/userreports/views.py
```diff
@@ -1127,7 +1128,20 @@ def post(self, request, *args, **kwargs):
...
-        return [report for report in reports if report.table_id == self.config.table_id]
+        ret = []
+        for report in reports:
+            try:
+                if report.table_id == self.config.table_id:
+                    ret.append(report)
+            except DataSourceConfigurationNotFoundError:
+                _soft_assert = soft_assert(to=[
+                    '{}@{}'.format(name, 'dimagi.com')
+                    for name in ['jemord', 'cellowitz', 'npellegrino', 'frener']
```
emord: I'm sending to you guys as well so thatyou can handle the enikshay errors. I'll take care of the icds ones and can own it indefinitely

nickpell (approved):

proteusvacuum:
Thanks for fixing. Do we think it would be important to prevent this from happening in the first place (deleting a datasource that has dependencies)?

emord:
Ideally yes I think when we delete a data source the reports should be deleted as well.

It's going to be a little tricky since non static reports can be connected to static data sources and we'd need to catch it in a few places

PR #17829, opened 2017-09-14
Avoid trial banner added via google analytics for enterprise projects

mkangia:
https://manage.dimagi.com/default.asp?261895

### DIFF
--- a/corehq/util/context_processors.py
+++ b/corehq/util/context_processors.py
@@ -99,6 +99,12 @@ def enterprise_mode(request):
     }
 
 
+def skip_google_tag_manager(request):
+    return {
+        'skip_google_tag_manager': settings.SERVER_ENVIRONMENT in ['icds', 'enikshay']
+    }
+
+
 def commcare_hq_names(request):
     return {
         'commcare_hq_names': {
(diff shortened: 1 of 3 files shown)


settings.py
```diff
@@ -918,7 +918,7 @@
 DAYS_TO_KEEP_DEVICE_LOGS = 60
 
 MAX_RULE_UPDATES_IN_ONE_RUN = 10000
-
+SKIP_GOOGLE_ANALYTICS = False
```
emord: This isn't skipping google analytics. it's only skipping google tag manager. and we want google analytics on eniksay/icds anyways

emord (approved):

corehq/util/context_processors.py
```diff
@@ -99,6 +99,12 @@ def enterprise_mode(request):
     }
 
 
+def skip_google_tag_manager(request):
+    return {
+        'skip_google_tag_manager': settings.SERVER_ENVIRONMENT in ['icds', 'enikshay']
```
emord: @proteusvacuum @snopoke  is there a reason why enikshay isn't in ENTERPRISE_MODE ? seemed like there was some error https://github.com/dimagi/commcare-hq/pull/16985#issuecomment-315352298 any idea if that was resolved?

snopoke: enikshay can use enterprise mode. We couldn't when we were on softlayer since it applies to the entire HQ instance so we add a specific check from the domain name: https://github.com/dimagi/commcare-hq/pull/16985/commits/60d4bda62c94ea93867c4992ef05bbd06c2c7ec2

Now that it's on a separate cluster we can set enterprise mode

PR #17830, opened 2017-09-14
bucket submission metrics instead of using histogram

snopoke:


corehq/util/datadog/utils.py
```diff
@@ -82,3 +82,20 @@ def get_url_group(url):
...
+    """Get value bucket for the given value
+
+    Bucket values because datadog's histogram is too limited
+
+    Basically frequency is not high enough to have a meaningful
+    distribution with datadog's 10s aggregation window, especially
+    with tags. More details:
+    https://help.datadoghq.com/hc/en-us/articles/211545826
+    """
+    unit = unit or ''
```
millerdev: strings are immutable. can just set `''` as the default value rather than `None`.

corehq/util/datadog/utils.py
```diff
@@ -82,3 +82,20 @@ def get_url_group(url):
 def update_datadog_metrics(metrics):
     for metric, value in metrics.items():
         statsd.gauge(metric, value)
+
+
+def bucket_value(value, buckets, unit=None):
```
millerdev: might be nice to wrap this in a factory function so it could be used like this:
```py
_get_time_bucket = make_bucket_func([5, 20, 60, 120], 's')

class RestoreConfig(object):
    def _record_timing(self, duration):
        bucket = _get_time_bucket(duration)
        ...
```

millerdev (approved):

PR #17831, opened 2017-09-14
Use correct field person_id instead of external_id [skip ci]

mkangia:
Fixes for https://manage.dimagi.com/default.asp?260027

In progress:
1. Correct use of person id instead of external id for enikshay id.


proteusvacuum:
Oops, just saw the tag after merging. This looks like a fine change though.

mkangia:
No worries. It was fine to let this roll out without waiting for other changes.

PR #17832, opened 2017-09-14
Revert "Avoid trial banner added via google analytics for enterprise projects"

mkangia:
Reverts dimagi/commcare-hq#17829

hold off on merge till ansible changes done to add Enterprise mode for enikshay
https://github.com/dimagi/commcarehq-ansible/pull/945

### DIFF
--- a/corehq/apps/hqwebapp/templates/hqwebapp/base.html
+++ b/corehq/apps/hqwebapp/templates/hqwebapp/base.html
@@ -2,13 +2,15 @@
 {% get_current_language as LANGUAGE_CODE %}
 <html lang="{{ LANGUAGE_CODE }}">
     <head>
-        <!-- Google Tag Manager -->
-        <script>(function(w,d,s,l,i){w[l]=w[l]||[];w[l].push({'gtm.start':
-        new Date().getTime(),event:'gtm.js'});var f=d.getElementsByTagName(s)[0],
-        j=d.createElement(s),dl=l!='dataLayer'?'&l='+l:'';j.async=true;j.src=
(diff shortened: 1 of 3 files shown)


emord (approved):

corehq/apps/hqwebapp/templates/hqwebapp/base.html
```diff
@@ -2,7 +2,7 @@
 {% get_current_language as LANGUAGE_CODE %}
 <html lang="{{ LANGUAGE_CODE }}">
     <head>
-        {% if not skip_google_tag_manager %}
+        {% if not enterprise_mode %}
```
snopoke: I think a better flag to use would be the `is_saas_environment` flag which is only True on prod.

mkangia: @snopoke but we would want this on india server as well where we would have non enterprise projects?

snopoke: I don't think we want this on india server since we aren't marketing india server at all. If that changes we can always add it back.

mkangia: Updated.

snopoke:
conflicts with https://github.com/dimagi/commcare-hq/pull/17833

mkangia:
@calellowitz 
Looks like you made a similar change #17833 
so is it better to add `GTM_ID` or use the `is_saas_enviroment`. I think both are doing the same thing.

emord:
I think it's better to use GTM_ID. that way it's a little more specific and whitelisted

PR #17835, opened 2017-09-14
benreg expr

calellowitz:
@proteusvacuum 
This should do the trick


proteusvacuum:
I wish we had unit tests for reports. This looks fine to me, but I guess there's only one way to find out if it worked?

proteusvacuum (approved):

emord:
@proteusvacuum We have some of those in ICDS. here's a pretty simple example in the UCR code https://github.com/dimagi/commcare-hq/blob/master/corehq/apps/userreports/tests/test_report_data.py#L171

PR #17837, opened 2017-09-14
send caught api exceptions to datadog

proteusvacuum:
I'm not sure the best way of going about this.
I want to capture all the 500 issues that come from hitting these endpoints so we can make sure that we have 0 of them, or send apologies to the affected parties whenever we do. 
This doesn't catch if anything happens before this code is actually run though (i.e. if the db fails in one of the decorator methods, or something else causes a 500). Anyone else have ideas?

@dannyroberts @nickpell

### DIFF
--- a/custom/enikshay/integrations/bets/views.py
+++ b/custom/enikshay/integrations/bets/views.py
@@ -1,7 +1,6 @@
 """
 https://docs.google.com/document/d/1RPPc7t9NhRjOOiedlRmtCt3wQSjAnWaj69v2g7QRzS0/edit
 """
-import datetime
 import json
 
 from dateutil import parser as date_parser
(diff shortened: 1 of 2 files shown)


custom/enikshay/integrations/bets/views.py
```diff
@@ -177,17 +176,21 @@ def _validate_updates_exist(domain, updates):
...
+            if not settings.UNIT_TESTING:
+                notify_exception(request, "BETS sent the eNikshay API a bad request.")
+            return json_response({"error": e.message}, status_code=e.status_code)
+
+        bulk_update_cases(domain, [
+            (update.case_id, update.properties, False) for update in updates
+        ], __name__ + ".payment_confirmation")
+        return json_response({'status': SUCCESS})
+    except Exception:
+        datadog_counter('commcare.enikshay.integration_500', tags=['BETS_payment_confirmation'])
```
snopoke: tags should be formatted as `name:value`

snopoke (changes requested):
Looks fine. One other option would be to add a tag to the rate limited errors metric which records which app the error was raised in: https://github.com/dimagi/commcare-hq/blob/7f462b7cc878ab57e4c72dfda99ff996b01b4aee/corehq/util/sentry.py#L42-L42

You'd have to inspect the traceback to get the app name.

PR #17839, opened 2017-09-15
SyncLogFact table

sravfeyn:
@calellowitz @gcapalbo 
cc @emord


corehq/warehouse/models/facts.py
```diff
@@ -78,7 +79,38 @@ def dependencies(cls):
...
+
+    sync_log_id = models.CharField(max_length=255)
+    sync_date = models.DateTimeField(null=True)
+
+    # these can be null per SyncLogStagingTable
+    domain = models.CharField(max_length=255, null=True)
+    user_id = models.CharField(max_length=255, null=True)
+
+    user_dim = models.ForeignKey(UserDim, on_delete=models.PROTECT)
+    domain_dim = models.ForeignKey(DomainDim, on_delete=models.PROTECT)
```
snopoke: why is it necessary to put the domain and user_id in the table as well as foreign keys to the user and domain dimensions?

sravfeyn: Yeah I had same question, Giovanni, Cal and I will [discuss this](This is what's being followed in the other Fact tables. ). This is what's being followed in the other Fact tables.

corehq/warehouse/models/facts.py
```diff
@@ -114,5 +146,6 @@ def dependencies(cls):
             USER_DIM_SLUG,
             FORM_STAGING_SLUG,
             FORM_FACT_SLUG,
+            # ToDo, should append SyncLogFact?
```
calellowitz: I think you are correct. `SYNCLOG_FACT_SLUG` does need to be here

corehq/warehouse/transforms/sql/synclog_fact.sql
```diff
@@ -0,0 +1,26 @@
...
+    domain_table.id,
+    user_table.user_id,
+    user_table.id,
+    build_id,
+    duration,
+    '{{ batch_id }}'
+FROM
+    {{ synclog_staging }} AS synclog_table
+JOIN {{ domain_dim }} AS domain_table ON synclog_table.domain = domain_table.domain
+-- Allow synclogs to be inserted that do not have users in the UserDim
```
calellowitz: why is this necessary. all users should be in the user dim. if they are not that should be seen as a failure, right?

corehq/warehouse/models/facts.py
calellowitz: I have the same question as simon, do you remember what resolution we came to? The only reason I could see for that is to allow some faster queries where all we need is the id, but that doesnt seem super userful, since rarely do we want the id and not the username.

calellowitz (changes requested):
generally looks good. just a couple comments

corehq/warehouse/transforms/sql/synclog_fact.sql
sravfeyn: I will investigate this a bit more, but I remember I had to do this because of some synclogs with null users.

sravfeyn: Comes from existing design decision https://github.com/dimagi/commcare-hq/blob/master/corehq/warehouse/transforms/sql/form_fact.sql#L35

But I think we do need a better way to handle

sravfeyn:
Mgirated this to updated code https://github.com/dimagi/commcare-hq/pull/19276

PR #17841, opened 2017-09-15
use the correct pattern for allowed_to_forward

czue:
@proteusvacuum

### DIFF
--- a/corehq/motech/repeaters/models.py
+++ b/corehq/motech/repeaters/models.py
@@ -370,10 +370,11 @@ class CreateCaseRepeater(CaseRepeater):
     Just like CaseRepeater but only create records if the case is being created.
     Used by the Zapier integration.
     """
-    # note: the logic for how this is managed is handled in signals.create_case_repeat_records
-    # so this class actually does nothing except exist to be able to differentiate the two
-    # at the repeater and repeat record level. otherwise it's exactly the same as CaseRepeater
     friendly_name = _("Forward Cases on Creation Only")
+
+    def allowed_to_forward(self, payload):
+        # assume if there's exactly 1 xform_id that modified the case it's being created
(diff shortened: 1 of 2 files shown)


snopoke (approved):

corehq/motech/repeaters/models.py
```diff
@@ -370,20 +370,23 @@ class CreateCaseRepeater(CaseRepeater):
...
     Used by the Zapier integration.
     """
-    # note: the logic for how this is managed is handled in signals.create_case_repeat_records
-    # so this class actually does nothing except exist to be able to differentiate the two
-    # at the repeater and repeat record level. otherwise it's exactly the same as CaseRepeater
     friendly_name = _("Forward Cases on Creation Only")
 
+    def allowed_to_forward(self, payload):
+        # assume if there's exactly 1 xform_id that modified the case it's being created
+        return len(payload.xform_ids) == 1
```
proteusvacuum: do forms ever come in out of order?

snopoke: alas they do but it would still create the case

proteusvacuum: yeah, just checking that this is then the desired behaviour, in order:
- "Update" form submitted -> Form creation forwarding happens
- "Create" form submitted -> Form update forwarding happens

czue: I think that's consistent behavior with how HQ would treat it (e.g. the case starts showing up in the case list when the "update" form comes in).

proteusvacuum: Hmm. There are specific things we set in "creates" that an external system might be interested in to prime their system:
- name
- case_type
- owner_id

How does Zapier behave when it receives a "create" without any of that information (i.e. when we treat the update as a create)?

PR #17842, opened 2017-09-15
Add support for field mapping where extra_pulmonary is extrapulmonary

mkangia:


mkangia:
@sheelio i suspect that for a case where site is "pulmonary_and_extrapulmonary" we would want this to be "Y" for both the cases. Let me know if thats the change needed.

proteusvacuum:
Ugh. This is annoying.

PR #17843, opened 2017-09-15
Fix infrastructure form calculation

emord:
Need to pick the greater one out of these two due to a difference between v1 and v2

http://manage.dimagi.com/default.asp?261060#

@sravfeyn


custom/icds_reports/tests/test_aggregation_script.py
```diff
@@ -161,17 +163,17 @@ def setUpClass(cls):
...
-
-        for adapter in cls.adapters:
-            if adapter.config.table_id == 'static-child_health_cases':
-                # hack because this is in a migration
-                continue
-            adapter.rebuild_table()
-
-        cls.setUpTestData()
-        move_ucr_data_into_aggregation_tables(datetime(2017, 5, 28), intervals=2)
+        # configs = StaticDataSourceConfiguration.by_domain('icds-cas')
```
sravfeyn: A comment on why these are kept commented out is useful. (unless you are planning to look at it in separate PR)

sravfeyn (approved):

custom/icds_reports/tests/test_aggregation_script.py
emord: That was unintentional. thanks for catching

PR #17845, opened 2017-09-15
prevent data modification during data / cluster migrations

snopoke:
remembered that this never got done after the eNikshay migration

### DIFF
--- a/corehq/form_processor/reprocess.py
+++ b/corehq/form_processor/reprocess.py
@@ -8,6 +8,7 @@ from casexml.apps.case.exceptions import IllegalCaseId, InvalidCaseIndex, CaseVa
 from casexml.apps.case.exceptions import UsesReferrals
 from casexml.apps.case.signals import case_post_save
 from corehq.apps.commtrack.exceptions import MissingProductId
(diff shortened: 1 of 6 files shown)


corehq/form_processor/reprocess.py
```diff
@@ -30,6 +31,10 @@ class ReprocessingError(Exception):
 
 
 def reprocess_unfinished_stub(stub, save=True):
+    if any_migrations_in_progress(stub.domain):
+        logger.info("Ignoring stub during data migration: %s", stub.xform_id)
+        return
+
```
mkangia: What does this prevent?

mkangia (approved):
Dont have context on what all should be prevented but the changes look fine.

proteusvacuum (approved):

corehq/form_processor/reprocess.py
snopoke: This prevents unifinished submission stubs from being re-processed.

mkangia: Cool, Thanks!

PR #17850, opened 2017-09-15
give enikshay-performance-test enikshay custom code

dannyroberts:
kriti asked if we can give enikshay-performance-test domain the same static data source configs as enikshay, and I think this is how we would do that? @nickpell added you to confirm or point me to the right person.


nickpell (changes requested):
You'll allow need to add this to the list of domains in each static data source and/or report that you want your domain to have

dannyroberts:
OK I'll tell kriti to have it resourced and reference this ticket if helpful.

PR #17851, opened 2017-09-15
drop obsolete tables

nickpell:
@gcapalbo @mkangia 

Keeps tables used in the locations migration for reference, and the tables used to create eNikshay cases since they still are referenced from time to time.


gcapalbo (approved):
thanks Nick. are there any foreign keys which point to any of the dropped tables?

nickpell:
> are there any foreign keys which point to any of the dropped tables?

nope

mkangia:
@nickpell i thought you needed these tables and hence we migrated. You don't need them now?

nickpell:
Some of them are still needed, but the data size is causing issues so I'm dropping the tables that aren't as likely to be needed.  We still have the dumps if they need to be recreated at some point.

PR #17853, opened 2017-09-15
Fix report builder diffs

gcapalbo:
fixes the build
introduced with https://github.com/dimagi/commcare-hq/pull/17846

@proteusvacuum


gcapalbo:
going to merge since the build is currently broken and this test passes locally

PR #17860, opened 2017-09-15
Add case type relationships, opening and closing to app summary case …

orangejenny:
…download

http://manage.dimagi.com/default.asp?251222

@proteusvacuum What do you think?

screen shot 2017-09-15 at 5 05 32 pm

### DIFF
--- a/corehq/apps/app_manager/views/app_summary.py
+++ b/corehq/apps/app_manager/views/app_summary.py
@@ -434,7 +434,6 @@ class DownloadCaseSummaryView(LoginAndDomainMixin, ApplicationViewMixin, View):
             rows.append(tuple(row))
 
         return rows
-        #return tuple((case_type.name, prop.name, prop.description) for prop in case_type.properties)
 
     def get_case_questions_rows(self, case_type, language):
         rows = []
--- a/corehq/apps/app_manager/views/app_summary.py
+++ b/corehq/apps/app_manager/views/app_summary.py
@@ -384,7 +384,8 @@ class DownloadCaseSummaryView(LoginAndDomainMixin, ApplicationViewMixin, View):
         case_metadata = self.app.get_case_metadata()
         language = request.GET.get('lang', 'en')
 
-        headers = [(_('All Case Properties'), ('case_type', 'case_property', 'description'))]
+        headers = [(_('All Case Properties'), ('case_type', 'case_property', 'description')),
+                   (_('Case Types'), ('type', 'relationships', 'opened_by', 'closed_by'))]
         headers += list((
             case_type.name,
             tuple(CASE_SUMMARY_EXPORT_HEADER_NAMES)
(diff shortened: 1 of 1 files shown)


orangejenny:
code buddies @gcapalbo / @dannyroberts

corehq/apps/app_manager/views/app_summary.py
```diff
@@ -414,6 +419,23 @@ def get(self, request, domain, app_id):
...
+                     for form_id in case_type.closed_by.keys()]
+        for i in range(max(len(relationships), len(opened_by), len(closed_by))):
+            row = [case_type.name]
+            row.append(relationships[i] if i < len(relationships) else '')
+            row.append(opened_by[i] if i < len(opened_by) else '')
+            row.append(closed_by[i] if i < len(closed_by) else '')
+            rows.append(tuple(row))
+
+        return rows
+        #return tuple((case_type.name, prop.name, prop.description) for prop in case_type.properties)
```
proteusvacuum: was this intentionally left here?

corehq/apps/app_manager/views/app_summary.py
```diff
@@ -414,6 +419,23 @@ def get(self, request, domain, app_id):
...
+    def get_case_type_rows(self, case_type, language):
+        rows = []
+        relationships = ["[{}] {}".format(r, t) for r, t in case_type.relationships.iteritems()]
+        opened_by = [_get_translated_form_name(self.app, form_id, language)
+                     for form_id in case_type.opened_by.keys()]
+        closed_by = [_get_translated_form_name(self.app, form_id, language)
+                     for form_id in case_type.closed_by.keys()]
+        for i in range(max(len(relationships), len(opened_by), len(closed_by))):
+            row = [case_type.name]
+            row.append(relationships[i] if i < len(relationships) else '')
```
proteusvacuum: do you think it would be useful to include the relationship in every row, so that you can filter the document?

corehq/apps/app_manager/views/app_summary.py
orangejenny: oops, fixed

corehq/apps/app_manager/views/app_summary.py
orangejenny: How would that work for multiple relationships? I suppose that I could only display the parent type in the row(s) containing an open/close form that's relevant to that relationship...would that work for host relationships as well?

proteusvacuum: Oh, I though this was a tuple of (relationship, opened, closed). I.e. in the case of multiple relationships, I'd want to see which form opened that particular relationship type. Not sure if that information is available anywhere. 
I.e. I'd expect:

|relationship | opened_by | closed_by|
|---------|-------|---------|
|[extension] host | host opening form | host closing form|
|[extension] host | host opening form 2 ||
|[child] parent | parent opening form| parent closing form|
|[child] parent | | parent closing form 2|

But maybe I'm overthinking it.

orangejenny: No, that makes sense and I don't think would be much more effort. I'll take a stab at it.

orangejenny: updated:

screen shot 2017-09-20 at 10 35 03 am

gcapalbo (approved):

PR #17861, opened 2017-09-15
skip check for open cases when going up in hierarchy

mkangia:
1. The check for open case is not necessary when going up in hierarchy.
2. When checking for cases below a case and relying on open cases results in failures in case of closed cases

Revisit approach [here](https://docs.google.com/document/d/1mKTCZBGCQGIvQp1mXPMbCfMl2uFhMFA1nuKTHhBrO6g/edit)

Note: Retrigger cases failing due to this.


custom/enikshay/case_utils.py
```diff
@@ -28,7 +28,6 @@
 CASE_TYPE_TEST = "test"
 CASE_TYPE_PRESCRIPTION = "prescription"
 CASE_TYPE_VOUCHER = "voucher"
-CASE_TYPE_PRESCRIPTION = "prescription"
```
mkangia: was a duplicate.

mkangia:
Checked again and looks like the issue is actually in `get_open_occurrence_case_from_person`. But this should be a safe change to do anyway.
@proteusvacuum Feel free to close this if looks like it adds no value. Would need to figure out a better way for the real issue.

mkangia:
Went ahead and made a possible way out for this to work.
it should be safe to fetch the last closed case in case no open case is found at least in case of the HIV repeater. It needs the episode case to get the nikshay id when notifying.
its failing currently with `Person with id: {person_id} exists but has no open occurrence cases`.But we would need to fix this for both episode and occurrence.
Will add further tests for this.

Other way round would be to save the latest_nikshay_id on the person case so that we don't need to refer episode case at all but that would need update for all present cases where nikshay id is present.

custom/enikshay/case_utils.py
```diff
@@ -103,32 +102,38 @@ def get_person_case_from_episode(domain, episode_case_id):
...
-
+    last_closed would give the latest closed occurrence case if no open occurrence case found
     """
     case_accessor = CaseAccessors(domain)
     occurrence_cases = case_accessor.get_reverse_indexed_cases([person_case_id])
     open_occurrence_cases = [case for case in occurrence_cases
                              if not case.closed and case.type == CASE_TYPE_OCCURRENCE]
     if not open_occurrence_cases:
+        if last_closed:
+            all_occurrence_cases = [case for case in occurrence_cases if case.type == CASE_TYPE_OCCURRENCE]
```
proteusvacuum: do you think you could just include this in the first loop conditionally? That way you don't have to loop through them twice.
```py
[case for case in occurrence_cases
if (last_closed or not case.closed) and case.type == CASE_TYPE_OCCURRENCE]
```

custom/enikshay/case_utils.py
```diff
@@ -138,6 +143,13 @@ def get_open_episode_case_from_occurrence(domain, occurrence_case_id):
     if open_episode_cases:
         return open_episode_cases[0]
     else:
+        if last_closed:
+            all_episode_cases = [case for case in episode_cases
```
proteusvacuum: same comment as above

custom/enikshay/case_utils.py
```diff
@@ -103,32 +102,38 @@ def get_person_case_from_episode(domain, episode_case_id):
     )
 
 
-def get_open_occurrence_case_from_person(domain, person_case_id):
+def get_open_occurrence_case_from_person(domain, person_case_id, last_closed=False):
```
proteusvacuum: do you think it would make sense to make a different version of this function called `get_any_occurrence_from_person` or something similar? It's a bit confusing to have this be `get_open...` with the option to return closed cases too.

proteusvacuum (changes requested):

custom/enikshay/case_utils.py
mkangia: Agree could be written better. Its just that we want to check for last closed only when there are no open cases so that change won't work. One way would be to fetch all cases, and then first check for any open case and if not then return the first closed case. That might be a little complex to read.

custom/enikshay/case_utils.py
mkangia: hmm..i just started to make that change and making two different methods was becoming costly.
For ex
```
def get_open_or_first_closed_occurence_case_from_person(domain, person_case_id):
    open_occurence_case = get_open_occurrence_case_from_person(domain, person_case_id)
    if not open_occurence_case:
        case_accessor = CaseAccessors(domain)
        occurrence_cases = case_accessor.get_reverse_indexed_cases([person_case_id])
        all_occurrence_cases = [case for case in occurrence_cases if case.type == CASE_TYPE_OCCURRENCE]
        if all_occurrence_cases:
            all_occurrence_cases.sort(key=lambda x: x.closed_on, reverse=True)
            return all_occurrence_cases[0]
```
Would need to fetch reverse indexes twice in case there is no open case. Plus the method call would get a little difficult [here](https://github.com/dimagi/commcare-hq/pull/17861/files#diff-f9aa24b679e7af67fd33a5d7a75f1485R187).

I think just renaming this method itself would be fine. Renaming this to `get_occurrence_case_from_person` with that argument defaulting to False should be fine i guess.

mkangia: Made the rename change for now.

mkangia:
@sheelio also wanted your feedback on this change that this PR is making.

So whenever we traverse from top to down in case hierarchy structure we look for only open cases, so for example if we are looking for an episode case under occurrence, we would search for only the open case. This is to take care of cases where there are probably multiple cases with the same case type and we want the one which is active/current. But in some cases the navigation fails in case the current has been closed. For ex: when sending notifications for a follow up test we fetch the open episode case to get the corresponding Nikshay ID to send it to Nikshay along with the notification for follow up test. But if the episode case was closed in a form earlier or in past (in case of a notification reattempt for follow up test when failed the first time), this traversal fails.
So this PR updates that logic and looks for the last closed case with that case type we are looking for in case no open episode case was found.

I am not sure if there is a better way to do handle such scenarios by linking cases better.
For ex, Is there possibly a way to link the relevant episode to a test case when it is being added so that we don't have to go through occurrence and then find the relevant episode?

custom/enikshay/case_utils.py
proteusvacuum: oh right, that makes sense then.

proteusvacuum (approved):

sheelio:
@mkangia 
At least for test cases, we now store the episode case ID that the test is linked to.  There are two case properties:
- test.episode_case_id (the episode that was active when the test result was entered)
- test.new_episode_case_id (if the test triggered creating a new episode, this is also set). 

My guess is that you can use test.episode_case_id for the workflow you want.

mkangia:
@sheelio 
That sounds really good. 
Would it be present for all the tests (new and existing old test cases)

sheelio:
More generally, for the public sector I would use:
- Get occurrence from person: Get the most recently opened occurrence (even if that occurrence is cloesd)
- Get episode from occurrence: Get the most recently opened episode (even if that episode is closed).

mkangia:
@sheelio @proteusvacuum 
Added a doc to note down the approach taken and have a better clarity
https://docs.google.com/document/d/1mKTCZBGCQGIvQp1mXPMbCfMl2uFhMFA1nuKTHhBrO6g/edit

Also made more changes in the PR after last review by Farid
1. use episode_case_id on tests first and then fall back to traversing cases

custom/enikshay/case_utils.py
```diff
@@ -294,6 +294,19 @@ def get_adherence_cases_between_dates(domain, person_case_id, start_date, end_da
...
+    test_case_properties = test_case.dynamic_case_properties()
+    episode_case_id = test_case_properties.get('episode_case_id')
+    if episode_case_id:
+        case_accessor = CaseAccessors(domain)
+        try:
+            episode_case = case_accessor.get_case(episode_case_id)
+            return episode_case
+        except ENikshayCaseNotFound:
+            pass
+    return False
```
proteusvacuum: would it be better to return `None` here?

custom/enikshay/case_utils.py
```diff
@@ -294,6 +294,19 @@ def get_adherence_cases_between_dates(domain, person_case_id, start_date, end_da
     return open_pertinent_adherence_cases
 
 
+def get_associated_episode_case_for_test(domain, test_case):
```
proteusvacuum: To follow the convention of the other names in this file, do you think this should be `get_episode_case_from_test`?

mkangia: There is a little difference in what this method does. It does not do any traversal but gets the episode case id from the test case itself so named it differently.

custom/enikshay/integrations/nikshay/repeater_generator.py
```diff
@@ -259,7 +260,10 @@ class NikshayFollowupPayloadGenerator(BaseNikshayPayloadGenerator):
 
     def get_payload(self, repeat_record, test_case):
         occurence_case = get_occurrence_case_from_test(test_case.domain, test_case.get_id)
-        episode_case = get_open_episode_case_from_occurrence(test_case.domain, occurence_case.get_id)
+        episode_case = get_associated_episode_case_for_test(test_case.domain, test_case)
+        if not episode_case:
+            episode_case = get_episode_case_from_occurrence(test_case.domain, occurence_case.get_id,
```
proteusvacuum: :+1: 
Do you think it would be useful to bundle this into the `get_associated_episode_case_for_test` method (or something similar in `case_utils`)? This seems like something we'd always want to use with the new case structure.

proteusvacuum (approved):

mkangia:
@proteusvacuum 
Extracted some of the definite changes in https://github.com/dimagi/commcare-hq/pull/17935 while we figure out the rest.

PR #17863, opened 2017-09-17
Pick disease classification from occurence/episode for pulomary, expu…

mkangia:
…lmonary according to new 2b structure

`disease_classification` was set correctly when we updated for using 2b structure but we missed fixing the related properties


mkangia:
@proteusvacuum i am going to merge this. Looks like a safe change.

PR #17865, opened 2017-09-18
Soft limit time for async UCR

emord:
@gcapalbo I think you said you were going to look into this, but I realized we could do this.

This will just stop queueing tasks based on a cron schedule, so the workers will stay up but should scale down during the day

buddy @sravfeyn


sravfeyn (approved):

gcapalbo (approved):
thanks. does the number of queued items usually stay low? if the workers get behind they would still continue to process remaining tasks during peak time, but agree this is probably the best solution.

PR #17867, opened 2017-09-18
change form error handling

snopoke:
FYI @sravfeyn @amstone326 
This is still in progress but I thought I'd put it up for review so long.

Best reviewed by commit :fish: 

Summary of changes:
* Don’t fire signals for device logs, duplicate forms, error forms
* On receiving a duplicate form:
  * Check if there is an UnfinishedSubmissionStub for the form and try to reprocess it if there is
  * Check if saved form is an error form and try re-process if it is
* Make sure all post save happens inside UnfinishedSubmissionStub block to allow retrying via async process

Full spec here: https://docs.google.com/document/d/1kSZNPNunybhrErF30JZW1q7pJ3CUlmlGm6i2DfkxmQg/edit#heading=h.qlgqu57heu8w


corehq/form_processor/reprocess.py
```diff
@@ -47,6 +47,10 @@ def reprocess_unfinished_stub(stub, save=True):
         save and stub.delete()
         return
 
+    reprocess_unfinished_stub_with_form(stub, form, save)
```
millerdev: Should this `return reprocess_unfinished...`?

millerdev (commented):
Reviewed because I want to better understand submission processing code.

corehq/form_processor/submission_post.py
```diff
@@ -277,27 +302,33 @@ def process_xforms_for_cases(xforms, case_db):
...
+    def _fire_post_save_signals(self, instance, cases):
+        from casexml.apps.case.signals import case_post_save
+        error_message = "Error occurred during form submission post save (%s)"
+        error_details = {'domain': instance.domain, 'form_id': instance.form_id}
+        results = successful_form_received.send_robust(None, xform=instance)
+        has_errors = log_signal_errors(results, error_message, error_details)
+
+        for case in cases:
+            results = case_post_save.send_robust(case.__class__, case=case)
+            has_errors &= log_signal_errors(results, error_message, error_details)
```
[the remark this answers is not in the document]
snopoke: Thanks @millerdev. Not sure if you're being sarcastic but this is clearly wrong.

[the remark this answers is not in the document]
millerdev: Not being sarcastic. Trying to give you the benefit of the doubt.

corehq/form_processor/submission_post.py
```diff
@@ -379,30 +379,23 @@ def _transform_instance_to_error(interface, e, instance):
...
+    _notify_submission_error(instance, exception, error_message)
     FormAccessors(interface.domain).save_new_form(instance)
 
 
-def _notify_submission_error(interface, instance, exception, message=None):
+def _notify_submission_error(instance, exception, message):
     from corehq.util.global_request.api import get_request
-    request = get_request()
-    error_message = u'{}: {}'.format(type(exception).__name__, unicode(exception))
-    instance = interface.xformerror_from_xform_instance(instance, error_message, with_new_id=True)
```
sravfeyn: Good call removing this from here :+1

snopoke:
rebased off master

millerdev:
> rebased off master

Rebasing after reviewers have started to comment on a PR makes it harder for them to figure out what they had already reviewed what, if any, of the issues they raised have been addressed. My preference is now to merge master rather than rebase after code comments have been made (although I haven't always done this in the past... still learning).

snopoke:
@millerdev yes I'm aware of that but for some reason the github diff got messed up so that it included about 50 commits from master. Rebasing seemed like the better option.

custom/enikshay/integrations/nikshay/tests/test_repeaters.py
```diff
@@ -1164,12 +1169,16 @@ def test_trigger(self):
...
-        ):
+        with capture_log_output('notify') as logoutput:
             self._create_nikshay_enabled_case(set_property=PRIVATE_PATIENT_EPISODE_PENDING_REGISTRATION)
+        self.assertIn(NikshayLocationNotFound.__name__, logoutput.get_output())
+        expected_message = (
+            "Location with id {location_id} not found. "
+            "This is the owner for person with id: {person_id}".format(
+                location_id=person.owner_id, person_id=self.person_id)
+        )
+        self.assertIn(expected_message, logoutput.get_output())
```
snopoke: @mkangia @proteusvacuum FYI this is changing the way errors that are raised by `case_post_save` signal handlers are treated.

Previously this would have resulted in a 500 response to the phone and it would have retried the form at a later stage but the signal would NOT have been re-fired since the second form submission would be marked as a duplicate.

With this change the phone would get a 422 response and possibly retry. If the phone retried (or the reprocess queue picked up the UnfinishedSubmissionStub) then the signals would get re-fired.

Note that we'll still get a Sentry notification from the error.

corehq/form_processor/submission_post.py
```diff
@@ -409,12 +410,16 @@ def _notify_submission_error(instance, exception, message):
...
+            xform_id=instance.form_id,
+            timestamp=datetime.datetime.utcnow(),
+            saved=False,
+            domain=instance.domain,
+        )
     yield unfinished_submission_stub
 
-    unfinished_submission_stub.delete()
+    if unfinished_submission_stub:
+        unfinished_submission_stub.delete()
```
millerdev: should this stub be deleted if the context manager exits with an error? if yes, should use `try`/`finally` around the `yield`.

snopoke: no, if it exits with an error then it should not be deleted, that's most of the point of the it.

snopoke:
@millerdev @sravfeyn I think this is done. There is more that can still be done but I think I'll leave that to a separate PR since this is getting quite big.

I'm going to leave the label on for now as I'd like to give it one more pass through tomorrow.

millerdev (approved):
A lot going on in this PR. I don't understand everything, but don't have any objections either. I like that you've tested it pretty extensively.

snopoke:
Might need to revert https://github.com/dimagi/commcare-hq/pull/17867/commits/ab0debebc0dc8af96c092a35cbb86d334eac2e6d based on our discussion about atomicity

custom/enikshay/integrations/nikshay/tests/test_repeaters.py
mkangia: Thanks for the notification @snopoke. Its really difficult to understand the impact of this change without knowing how form processing works and what all this PR changed. 
Can you share what would be the change in the outcome from the repeater perspective and not from the phone perspective? 
Like a repeater that runs though the `case_post_save` signal will error out when it is attempted first. And then what happens?

snopoke: This should improve things from a repeater perspective. If the signal handler errors then 2 things happen:
1. The UnfinishedSubmissionStub will get picked up by the async process to re-fire the signals
2. The response to the phone will be 422 and if the phone retires it will re-fire the signals (assuming they haven't already been re-fired by the async procesS)

proteusvacuum: Interesting. I didn't realize that if there was an error in the signal that the form would be retried later by the phone. Errors in the signals are often to do with some data error that wouldn't be fixed by resubmitting the form. Do you think we should raise errors in a different fashion to prevent the phone from needing to do extra work? 

I don't think that re-firing the signals here is actually desired, as that will create new (potentially duplicate) RepeatRecords, which already have their own retry mechanism.

snopoke:
Errors here aren't from firing the repeater they are from creating the
repeat record so I would assume (and it certainly should be the case) that
if an exception is raised here that the repeat record isn't created.

In v2.0 (current version) the form would be resubmitted by the phone since
a 500 response was returned but in v3.0 (this PR) the response code will be
422 and the phone will behave as follows:

When mobile receives this response from HQ, it will “quarantine” the form
> in question, meaning that it can then be viewed under a special filter
> within Saved Forms. From there, users will have the option to a) view an
> explanation for why the form was quarantined (this is where the <message>
> text would be used), and b) re-add the form to the unsent queue to try to
> submit it again.

amstone326:
@snopoke When this is ready, can you deploy it just on staging to start so that I can do some testing with my mobile PR?

snopoke:
100%. I'll deploy it to staging tomorrow

amstone326:
@snopoke Awesome, just let me know when it's there.

snopoke:
@amsagoff deployed to staging

emord:
@snopoke I'm not sure @amsagoff  is going to be checking that 😄 @amstone326 might be interested though

amstone326:
@snopoke Just tried some manual testing on staging and didn't experience the behavior I would expect. Sending you an email with details.

custom/enikshay/integrations/nikshay/tests/test_repeaters.py
mkangia: > I don't think that re-firing the signals here is actually desired, as that will create new (potentially duplicate) RepeatRecords, which already have their own retry mechanism.

was this issue addressed?
@snopoke @proteusvacuum

snopoke: @proteusvacuum @mkangia sorry I thought I had responded to this. If an error is raised when the signal is fired then no repeat record should be created at all so the repeat record retry mechanism won't kick in.

PR #17869, opened 2017-09-18
enikshay: treatment status migration: filter to only public cases

sheelio:
@nickpell 

This should fix the public / private sector bug.  Can you rerun with the sample output for us to review?


sheelio:
Easier to review with ?w=1

nickpell (approved):

PR #17872, opened 2017-09-18
add update migration for migrated tests

nickpell:
Updates location info for tests migrated from Nikshay.

If a DMC location cannot be found, the missing DMC is logged for followup.


nickpell:
Already ran this migration, but would like to get into the codebase

@orangejenny @proteusvacuum

PR #17873, opened 2017-09-18
Cast provided amounts to integer

esoergel:
They are sent down as strings of floats, like `u'400.0'`.  This is some duck typing to see if we can interpret it as an integer

https://sentry.io/dimagi/commcarehq/issues/289037330/

@proteusvacuum

### DIFF
--- a/custom/enikshay/integrations/bets/tests/test_bets_updates.py
+++ b/custom/enikshay/integrations/bets/tests/test_bets_updates.py
@@ -70,7 +70,7 @@ class TestBetsUpdates(TestCase):
         }]})
         self.assertResponseStatus(res, 200)
         self.assertDictContainsSubset(
-            {'state': 'paid', 'amount_paid': '100'},
+            {'state': 'paid', 'amount_paid': '100.0'},
             get_case(self.domain, voucher.case_id).case_json,
         )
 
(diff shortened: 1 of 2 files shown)


proteusvacuum:
@esoergel I updated it to match the spec document which says that we will accept a "decimal amount". If you have more up to date documentation than what we sent them, I'm interested in seeing it. I'm referencing the final "DataDump_Definitions" document sent on August 10.

proteusvacuum:
I'm merging this as it should be mostly backwards-compatible with your change if my change isn't right, and I want to tell them to retrigger everything they've sent so far. @esoergel

esoergel:
> @esoergel I updated it to match the spec document which says that we will accept a "decimal amount". If you have more up to date documentation than what we sent them, I'm interested in seeing it. I'm referencing the final "DataDump_Definitions" document sent on August 10.

This requirement was changed to be an integer during a call on Sept 1st.  We initially said "integer rounded up (ceil)", then that was later amended to be a standard rounded integer in an email on Sept 4th.

Sounds like we need to have a live copy of this document that's kept up-to-date - I will work on that.

esoergel:
Actually let's go ahead and assume this is canon: https://docs.google.com/document/d/1HodLa86TiSrWKu3js9moSmIkUg5kwK7uma_qRThuzrw/edit
It still needs some updates, but we should go with that, since it's the most recent doc we've agreed on.

PR #17874, opened 2017-09-18
Show feedback when (de-)releasing an app build fails

orangejenny:
http://manage.dimagi.com/default.asp?262084

screen shot 2017-09-18 at 1 08 40 pm

@gcapalbo / @dannyroberts / @biyeun

### DIFF
--- a/corehq/apps/app_manager/templates/app_manager/partials/releases_table.html
+++ b/corehq/apps/app_manager/templates/app_manager/partials/releases_table.html
@@ -107,8 +107,7 @@
               <div class="build-is-released">
                   <!--ko if: is_released() == 'error' -->
 
-                  <span class="label label-danger"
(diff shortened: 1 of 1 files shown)


corehq/apps/app_manager/templates/app_manager/partials/releases_table.html
```diff
@@ -105,6 +105,14 @@
                   </div>
               {% endif %}
               <div class="build-is-released">
+                  <!--ko if: is_released() == 'error' -->
+
+                  <span class="label label-danger"
+                        data-bind="visible: is_released() == 'error'">{% trans "Could not update" %}</span>
```
gcapalbo: is this check of `is_released() == 'error'` needed if the ko if is used outside of it?

orangejenny: oops, meant to remove that. thanks, fixed.

PR #17875, opened 2017-09-18
Update mapping according to new values

mkangia:
Update result grade to reflect
```
test.result_grade:
error_no_samples_tested: ???
negative_not_seen: 0
scanty: I think this is SC X based on the bacilli count
1plus: 11
2plus: 12
3plus: 13
error_invalid_result: ???
```

@sheelio

### DIFF
--- a/custom/enikshay/integrations/nikshay/field_mappings.py
+++ b/custom/enikshay/integrations/nikshay/field_mappings.py
@@ -105,7 +105,7 @@ purpose_of_testing = {
 }
 
 smear_result_grade = {
-    '0': 99,  # Negative
+    'negative_not_seen': 99,  # Negative
     'SC-0': 98,  # Positive
     'SC-1': 1,
     'SC-2': 2,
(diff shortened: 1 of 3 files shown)


custom/enikshay/integrations/nikshay/field_mappings.py
```diff
@@ -105,7 +105,7 @@
 }
 
 smear_result_grade = {
-    '0': 99,  # Negative
```
sheelio: @mkangia 

What's the mapping for scanty?  We have a potential result grade of "scanty" and the scanty count is in max_bacilli_count.  Can we juse that for the mapping, or should we just sent 98?

custom/enikshay/integrations/nikshay/field_mappings.py
```diff
@@ -116,9 +116,9 @@
     'SC-7': 7,
     'SC-8': 8,
     'SC-9': 9,
-    '1+': 11,
```
sheelio: Instead of changing the existing mapping, can we add new mappings? That way older data will also continue to work.

custom/enikshay/integrations/nikshay/field_mappings.py
mkangia: @sheelio 
We combine result grade scanty with bacilli count. So bacilli count can be between 0 and 9. So we combine them as "scanty-{bacilli_count}" to get the corresponding value.

custom/enikshay/integrations/nikshay/field_mappings.py
mkangia: @sheelio thank you for noting that.. adding back old mapping as well

PR #17878, opened 2017-09-19
Add script to update site setup

mkangia:
This could be run when we set up new environments

### DIFF
--- a/corehq/apps/hqadmin/management/commands/update_site_setup.py
+++ b/corehq/apps/hqadmin/management/commands/update_site_setup.py
@@ -20,11 +20,12 @@ class Command(BaseCommand):
     def handle(self, site_address, *args, **options):
         if not options['skip_checks']:
             if settings.SITE_ID != 1:
-                raise CommandError("SITE ID under settings expected to have value 1 since only one object is expected")
+                raise CommandError("SITE ID under settings expected to have value 1 since only one object is "
+                                   "expected")
             sites_count = Site.objects.count()
             if sites_count != 1:
-                raise CommandError("Expected to have only one object added by Site during setup but currently its %s "%
-                                   Site.objects.count())
+                raise CommandError("Expected to have only one object added by Site during setup but currently "
+                                   "its %s " % Site.objects.count())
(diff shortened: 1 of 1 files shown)


corehq/apps/hqadmin/management/commands/update_site_setup.py
```diff
@@ -0,0 +1,51 @@
...
+        if not options['skip_checks']:
+            if settings.SITE_ID != 1:
+                raise CommandError("SITE ID under settings expected to have value 1 since only one object is "
+                                   "expected")
+            sites_count = Site.objects.count()
+            if sites_count != 1:
+                raise CommandError("Expected to have only one object added by Site during setup but currently "
+                                   "its %s " % Site.objects.count())
+            site_object = Site.objects.first()
+            if site_object.name != "example.com" and site_object.domain != "example.com":
```
esoergel: You could also make this idempotent - check if `name` and `domain` are already right, and then just return.

esoergel (approved):
Nice! Know what it'd look like to add an ansible task to call this on setup?

corehq/apps/hqadmin/management/commands/update_site_setup.py
mkangia: Added a check for it as well.

mkangia:
> Nice! Know what it'd look like to add an ansible task to call this on setup?

Not sure. Would need to check.

corehq/apps/hqadmin/management/commands/update_site_setup.py
```diff
@@ -36,6 +36,10 @@ def handle(self, site_address, *args, **options):
                     """.format(name=site_object.name, domain=site_object.domain)
                 )
 
+            if site_object.name == site_object.domain == site_address:
+                print('Site object with the expect name and domain already set.')
+                print('No update made.')
```
esoergel: This should `return`

mkangia: oops. Adding.

esoergel (approved):

PR #17879, opened 2017-09-19
properly check case types in create/update case repeaters

czue:
found this :beetle: while eating my own :dog: :shallow_pan_of_food:


snopoke (approved):
:poodle:

PR #17882, opened 2017-09-19
proper default sorting for excel downloads

calellowitz:
@millerdev 
default sorting was not changed when the new nested sort was introduced


millerdev (approved):
Don't have much context here. Didn't see any syntax errors 😛

PR #17884, opened 2017-09-19
migrate a few things from pl proxy to django

czue:
@snopoke just starting to tentatively look into this. this moves `CaseAccessorSQL.get_cases` and `FormAccessorSQL.get_forms` to use the new python shard stuff instead of pl_proxy.

Is there anything else I should do to test apart from running the tests or are we pretty confident in the coverage and underlying functions?

fyi @dimagi/scale-team 

sorry if i screwed something up. it's been a while. :older_man:


corehq/sql_accessors/migrations/0055_drop_forms_cases_by_id_accessors.py
```diff
@@ -0,0 +1,18 @@
...
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('sql_accessors', '0054_drop_reindexa_accessor_functions'),
+    ]
+
+    operations = [
+        HqRunSQL("DROP FUNCTION IF EXISTS get_cases_by_id(TEXT[])"),
+        HqRunSQL("DROP FUNCTION IF EXISTS get_forms_by_id(TEXT[])"),
```
emord: This should be split up into a separate PR so that we don't drop a function that could be used by old webworkers

corehq/sql_proxy_accessors/migrations/0039_drop_get_forms_cases_by_id_accessor.py
```diff
@@ -0,0 +1,18 @@
...
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('sql_proxy_accessors', '0038_drop_unused_sql'),
+    ]
+
+    operations = [
+        HqRunSQL("DROP FUNCTION IF EXISTS get_cases_by_id(TEXT[])"),
```
emord: Should also be split to separate PR

emord (changes requested):

corehq/sql_proxy_accessors/README.md
```diff
@@ -1,6 +1,7 @@
 # Sharding postgresql
 
-We use [PL/Proxy](https://plproxy.github.io/) for sharding.
+We use [PL/Proxy](https://plproxy.github.io/) for sharding, although as of September 2017
+are in the process of phasing it out in favor of sharding in the application layer.
```
millerdev: Not very excited about this. One unfortunate consequence of sharding in the application layer is that queries to multiple shards will almost certainly be executed serially (because of the GIL) rather than in parallel unless we do something like write a C extension for sharded queries (non-trivial). I'm not sure if PL/Proxy does that, but I think [Citus has features like parallel execution](https://docs.citusdata.com/en/v6.2/dist_tables/querying.html#querying-distributed-tables-sql).

czue: Hmm, this seems worth talking to @snopoke about. I was under the impression the plan was to move everything to the application layer to make future operations like shard migrations easier to manage, but not sure if parallellization was discussed/considered. feel free to hit me up if there's a meeting I should join to listen in on or if I should jump on slack.

corehq/sql_accessors/migrations/0055_drop_forms_cases_by_id_accessors.py
czue: makes sense, thanks.

corehq/sql_proxy_accessors/README.md
czue: (will follow up on email)

czue: > I'm not sure if PL/Proxy does that, but Citus has features like parallel execution.

according to https://wiki.postgresql.org/wiki/PL/Proxy it does execute in parallel

snopoke: Any suggestions @millerdev?

czue:
thanks for the feedback on this so far. I'm going to hold off on making fixes / looking at the test error until we get consensus on whether this is a good direction to go.

corehq/sql_accessors/migrations/0055_drop_forms_cases_by_id_accessors.py
snopoke: Just to confirm we have to split this out since migration happen while the old webworkers are running.

corehq/sql_db/util.py
```diff
@@ -49,6 +50,34 @@ def run_query_across_partitioned_databases(model_class, q_expression, values=Non
...
+    :param model_ids: The list of IDs to retreive
+
+    :param q_expression: An instance of django.db.models.Q representing any additional
+    filters to apply (optional)
+
+    :return: A generator with the results
+    """
+    assert issubclass(model_class, PartitionedModel)
+    assert model_class.partition_attr
+    query_param = '{}__in'.format(model_class.partition_attr)
```
snopoke: Could use `Q` objects here instead of a dict. That would also enable passing arbitrary other filters in (similar to `run_query_across_partitioned_databases`)

corehq/sql_db/util.py
```diff
@@ -49,6 +50,34 @@ def run_query_across_partitioned_databases(model_class, q_expression, values=Non
             yield result
 
 
+def get_objects_by_id_across_partitioned_databases(model_class, model_ids, additional_query_params=None):
```
snopoke: Could be nice to have this callable from the model manager:
```
CommCareCaseSQL.objects.get_by_ids(case_ids)
```

snopoke (commented):
Looks good but needs to be split.

corehq/sql_db/util.py
gcapalbo: we also have to keep in mind that `partition_attr` is just the attribute we use to partition the object, but not necessarily the primary key we use to do lookups.

for example, some models may be sharded by case_id but have a different primary key that we use to lookup the actual objects.

might be good to rename `model_ids` to `partition_ids`, rename the function to `get_objects_by_partition_id_across_partitioned_databases`, and update the docstring to explain this difference

corehq/sql_proxy_accessors/README.md
millerdev: I know @snopoke and I had discussed this briefly in the past, and I came away thinking the work he was doing looked like good architectural cleanup without many far-reaching effects (that is, easy to switch back if we wanted). Maybe I'm forgetting what I reviewed before, but this work that Cory is doing looks more invasive.

I have had reservations about moving sharding logic from PL/Proxy into Python ever since I first heard about it, and I'm still not on board with the idea. I have not devoted a lot of time to making a case so far because I haven't been working on it directly and it's felt like I'd have to drop other things to find enough time to make a solid case. I'd like to do that if I can. The place I'd like to start is analyzing our current performance bottleneck(s) to see if we can really pin the blame on PL/Proxy and answer the question will moving sharding logic into Python improve performance? (my gut feeling says no way, but that's all I've got right now.) I'm going to make some time today to look into this a bit more.

snopoke:
I agree that we won't improve on performance but that's never been the
issue. Please review the doc I drew up for a list of pros / cons
(Architecture Review: Sharding in Django vs pl_proxy)

czue:
closing in favor of other possible options

PR #17885, opened 2017-09-19
don't fail hard when hooks raise exception

nickpell:


proteusvacuum (approved):
I'm curious which hook was failing?

nickpell:
@proteusvacuum this:
https://sentry.io/dimagi/commcarehq/issues/352504254/

proteusvacuum:
Thanks @nickpell : https://github.com/dimagi/commcare-hq/pull/17887
(the change in this PR is still fine, but that check should never be happening for these cases)

PR #17886, opened 2017-09-19
don't raise exception when owner_id is _archive_

nickpell:


nickpell:
https://sentry.io/dimagi/commcarehq/issues/352504254/

proteusvacuum:
This should be solved by: https://github.com/dimagi/commcare-hq/pull/17887

PR #17892, opened 2017-09-20
Added fully immunized column to the child health monthly view.

emord:
@sheelio https://manage.dimagi.com/default.asp?262264


sheelio:
@emord 

This won't provide the date of the immunization, just a 1 or 0 correct?  I think its fine if that's all we need.

emord:
That's correct. this PR isn't complete, it will also need a change to the dashboard code that I can hand off to sol dev tomorrow. I verified that we don't need the date for any reports at the moment, but will make a note that a rebuild is required if we want it in the future

emord:
merged lukasz's changes here. Will merge for deploy once tests pass

PR #17899, opened 2017-09-20
Properly calculate covered sites.

emord:
This possibly only happens during QA, but we noticed that some supervisors were marked as launch even though there weren't AWCs that were launched under them (also effected districts and blocks up the chain). This occurred because there were no household registrations for the AWC, but there was some other data in that AWC. To fix that I changed all hardcoded 1 to a CASE statement

There was also an issue when viewing at supervisor level the state, district and block would always be 0 because we weren't populating that. This populates those queries with a case statement as well

https://manage.dimagi.com/default.asp?261061#1395265

@sheelio @sravfeyn


custom/icds_reports/migrations/sql_templates/create_functions.sql
```diff
@@ -1186,6 +1186,10 @@ BEGIN
 		'usage_num_thr = ut.usage_num_thr, ' ||
 		'usage_num_hh_reg = ut.usage_num_hh_reg, ' ||
 		'is_launched = ut.is_launched, ' ||
+		'num_launched_states = ut.num_launched_awcs, ' ||
```
sheelio: @emord 
I had set num_launched_states / num_launched_districts / num_launched_blocks / num_launched_supervisors to null here since those values weren't applicable when viewing data per AWC.  Is it preferable to keep that, or was there a reason we want to set the value to 0?

custom/icds_reports/migrations/sql_templates/create_functions.sql
```diff
@@ -1504,10 +1508,10 @@ BEGIN
 		'month, ' ||
 		_rollup_text ||
 		'4, ' ||
-		quote_nullable(_null_value) || ', ' ||
-		quote_nullable(_null_value) || ', ' ||
-		quote_nullable(_null_value) || ', ' ||
-		'1, ' ||
```
sheelio: Based on whether we want to keep these values as null or not, it might just cleaner to just replace the "1" line with "CASE WHEN sum(num_launched_X) > 0 ...." for each level of aggregation.

sheelio: Ah nevermind, I see the reason for setting these explicitly to zero.  This looks good to me then.

PR #17900, opened 2017-09-20
ICDS: Clean Up Data Sources

sheelio:
This PR moves the iterator count to inside the data source (by moving all the month related expressions into the data source).  This allows us to more easily change the iteration count.


sheelio:
@emord 

This PR prepares the data sources to change the iteration count.   Once committed, I'll create another PR that actually changes the iteration count.

custom/icds_reports/ucr/data_sources/child_health_cases_monthly_tableau2.json
```diff
@@ -16,15 +16,65 @@
...
+        "filters": [
+          {
+            "operator": "eq",
+            "type": "boolean_expression",
+            "expression": {
+              "type": "root_doc",
+              "expression": {
+                "datatype": null,
+                "type": "property_name",
+                "property_name": "closed"
```
emord: Do you forsee any issues with the aggregation from not having rows for closed cases?

custom/icds_reports/ucr/data_sources/child_health_cases_monthly_tableau2.json
```diff
@@ -2432,6 +2434,24 @@
           }
         }
       },
+      "child_age_in_days": {
```
emord: Is there a benefit to moving this here instead of keeping it in the custom python expressions file?

emord (commented):
The test failure looks real. Excited to be able to speed these up.

My comments look like thye're hidden at the moment, but the questions I have for this are:

1) Do you see any issues with the aggregation from not having rows for closed cases? Does the aggregation depend on these in any ways? (should only be applicable to your next PR that changes the iteration count)
2) What do we gain by taking the expression out of icds_reports/ucr/expressions.py?

I'm curious how confident you are in making this change live. The changes look fairly copy-paste, so it's likely similar but I wonder if it's worth using the following process:

1) making this a V3 (or v2_test etc) UCR data source for the QA domain
2) rebuilding the QA domain's data
3) doing some comparison of the v2 vs the v3 data source
4) if all goes well, copy these changes over to the v2 datasource
5) deploy it to ICDS

sheelio:
Happy to make this a v3 data source as a test (I was using our current tests to validate that it works the same). 

The main reason I'm moving stuff out of expressions.py is that the expressions there hard-code the iteration period to -3 to 3 month (since they reference icds_month_start / icds_month_end expressions).  I replaced the icds_month_start / icds_month_end expressions with expressions inside the data source that can have their own iteration period.   (When we originally built these data sources named expressions couldn't reference each other, so we created these external expressions). 

Will look at the test failure.

emord:
Thanks that all makes sense. as long as we're confident the tests are working correctly, i'm fine with merging without creating a v3

sheelio:
Alright this should be set to merge. 

The tests definitely cover the expressions I moved into the main UCR files (month start / end, age, open in month, alive in month).

PR #17901, opened 2017-09-20
Enable ledgers for non-Supply projects

esoergel:
This allows you to add in transaction question types without turning on commtrack, and tells HQ to include ledgers in the restore.  Everything else seemed to just work.  I made an app locally using this and it went fine - I was able to add a transaction question, submit data against that, see the ledger data on the associated case, and reference it in subsequent forms.  Notably, HQ doesn't seem to care whether the specified entity id is a supply-point case (any old case works), nor whether the entry id is a product or just some arbitrary string.  I thought there were some checks against that sort of thing, but I couldn't find anything.  If anyone else has suggestions for things to test, lemme know.

@dannyroberts this is what I was talking about today
@mkangia code buddy
@snopoke can you think of anything else that could potentially go awry with this change?

### DIFF
--- a/corehq/apps/app_manager/views/formdesigner.py
+++ b/corehq/apps/app_manager/views/formdesigner.py
@@ -269,7 +269,8 @@ def _get_vellum_plugins(domain, form, module):
     privileges.
     """
     vellum_plugins = ["modeliteration", "itemset", "atwho"]
-    if toggles.COMMTRACK.enabled(domain):
+    if (toggles.COMMTRACK.enabled(domain)
+            or toggles.NON_COMMTRACK_LEDGERS.enabled(domain)):
         vellum_plugins.append("commtrack")
     if toggles.VELLUM_SAVE_TO_CASE.enabled(domain):
         vellum_plugins.append("saveToCase")
--- a/corehq/toggles.py
+++ b/corehq/toggles.py
@@ -739,6 +739,13 @@ COMMTRACK = StaticToggle(
     save_fn=_commtrackify,
 )
 
+NON_COMMTRACK_LEDGERS = StaticToggle(
+    'non_commtrack_ledgers',
+    "Enable ledgers for projects which don't use Supply.",
+    TAG_EXPERIMENTAL,
+    namespaces=[NAMESPACE_DOMAIN],
+)
+
 CUSTOM_INSTANCES = StaticToggle(
     'custom_instances',
     'Inject custom instance declarations',
--- a/corehq/toggles.py
+++ b/corehq/toggles.py
@@ -741,8 +741,12 @@ COMMTRACK = StaticToggle(
 
 NON_COMMTRACK_LEDGERS = StaticToggle(
     'non_commtrack_ledgers',
-    "Enable ledgers for projects which don't use Supply.",
+    "Enable ledgers for projects not using Supply.",
     TAG_EXPERIMENTAL,
+    description=(
+        'Turns on the ledger fixture and ledger transaction question types in '
+        'the form builder. ONLY WORKS ON SQL DOMAINS!'
+    ),
     namespaces=[NAMESPACE_DOMAIN],
 )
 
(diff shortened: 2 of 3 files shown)


corehq/apps/app_manager/views/formdesigner.py
```diff
@@ -269,7 +269,8 @@ def _get_vellum_plugins(domain, form, module):
     privileges.
     """
     vellum_plugins = ["modeliteration", "itemset", "atwho"]
-    if toggles.COMMTRACK.enabled(domain):
+    if (toggles.COMMTRACK.enabled(domain)
+            or toggles.NON_COMMTRACK_LEDGERS.enabled(domain)):
         vellum_plugins.append("commtrack")
```
esoergel: @dimagi/team-form-designer 
This seemed to do the trick for enabling transaction question types in Vellum - do you know if there are any possibly unintended side-effects?  Are there other commtrack features besides transaction questions?  Should we split those up and make a `"ledgers"` vellum plugin?

emord: @esoergel interesting. I've never seen that team name. looks like it's just cory and clayton on there.Tagging @millerdev since he would know the most about this.

I think this is fine and we don't need a ledgers vellum plugin as it only adds the questions.

It does seem unfortunate that we have two feature flags controlling this. Is there a release path for the non commtrack ledgers? could it be made into an app manager add on for advanced plans ( @orangejenny ?) ? also is there a reason the previous check wasn't `domain.commtrack_enabled`?

snopoke (approved):
One note is that this only works on domains using the SQL backend. Ledgers on the couch backend do have a dependency on product ID and will error during submission. 

Not sure if it's worth adding a check for that since this is a feature flag.

Approving but agree with @emord's comments.

corehq/apps/app_manager/views/formdesigner.py
millerdev: The `commtrack` plugin adds four question types: Balance, Transfer, Dispense, and Receive. All are closely related, so I'm assuming that's what you wanted. LGTM

corehq/toggles.py
```diff
@@ -739,6 +739,13 @@ def _commtrackify(domain_name, toggle_is_enabled):
     save_fn=_commtrackify,
 )
 
+NON_COMMTRACK_LEDGERS = StaticToggle(
+    'non_commtrack_ledgers',
+    "Enable ledgers for projects which don't use Supply.",
```
millerdev: Grammar police here: should say "...projects _that_ don't use Supply." or maybe even "...projects _that do not_ use Supply."

corehq/apps/app_manager/views/formdesigner.py
orangejenny: > could it be made into an app manager add on for advanced plans

That'd make it visible to anyone (with an advanced plan). It's fine technically, if we aren't concerned about random SaaS users turning it on.

esoergel: I think this would be a valuable tool for complicated enterprise apps, and would very much like to see it in product.  That said, I'm guessing we'd want product to weigh in and maybe revisit what the fields are named and how we expect this to be used before releasing this - the language is focused around stock updates, so it is pretty nonsensical as a generic tool.

This flag is used in two places - to decide whether to display the question types, and to decide whether to sync ledgers.  In the former case, an addon seems fine once this is released, and in the latter case, I guess we could instead do a check for stock states on the domain or something.  I suppose my ideal state for this would be to consider ledgers a feature independent from CommCare Supply, but enabled when you turn on Supply.  Not sure that's worth fully decoupling at this point, but agree that it'd be useful if/when we push this beyond eNikshay.

esoergel:
> One note is that this only works on domains using the SQL backend. Ledgers on the couch backend do have a dependency on product ID and will error during submission.

Good point!  I'll update the description to make a note of that.

esoergel:
(canceled build)

ctsims:
Wooo!!!!!! Ledgers are unshackled ⛓ 🔨 
🎊 🎊 🎊 🎊 🎊 🎊 🎊 🎊 🎊 🎊 🎊 🎊 🎊 🎊 🎊 🎊 🎊 🎊 🎊 🎊

corehq/apps/app_manager/views/formdesigner.py
ctsims: We definitely don't want any new non-dimagi's using this for now.

PR #17902, opened 2017-09-20
fix js error on case_data page

dannyroberts:
I checked that the error in http://manage.dimagi.com/default.asp?223100 (referenced in the comment) is not showing itself.


gcapalbo (approved):

orangejenny (approved):
It's a little weird that the js block now contains stylesheets, but I'm not going to be ideological about that.

kaapstorm (approved):

PR #17903, opened 2017-09-20
requirejs_main tag

millerdev:
@orangejenny

### DIFF
--- a/corehq/apps/hqwebapp/templatetags/hq_shared_tags.py
+++ b/corehq/apps/hqwebapp/templatetags/hq_shared_tags.py
@@ -717,11 +717,9 @@ class RequireJSMainNode(template.Node):
         return "<RequireJSMain Node: %r>" % (self.value,)
 
     def render(self, context):
-        if self.name in context:
-            raise TemplateSyntaxError(
-                "multiple '%s' tags not allowed (%s)" % (self.name, self.value))
-        # set name in block parent context
-        context.dicts[-2][self.name] = self.value
+        if self.name not in context:
+            # set name in block parent context
+            context.dicts[-2][self.name] = self.value
         return ''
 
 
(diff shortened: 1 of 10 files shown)


corehq/apps/hqwebapp/templatetags/tests/templates/requirejs_base.html
```diff
@@ -0,0 +1,13 @@
+{% load hq_shared_tags %}
+{% if requirejs_main %}
+unexpected {{ requirejs_main }} before block
+{% endif %}
+
+{% requirejs_main "requirejs/base" %}
+{% block requirejs_main %}{% endblock %}
```
millerdev: oops, this line should be removed (it has no effect and is not needed)

corehq/apps/hqwebapp/templatetags/hq_shared_tags.py
```diff
@@ -674,14 +673,54 @@ def render(self, context):
...
+
+
+class RequireJSMainNode(template.Node):
+
+    def __init__(self, name, value):
+        self.name = name
+        self.value = value
+
+    def __repr__(self):
+        return "<RequireJSMain Node: %r>" % (self.value,)
```
orangejenny: is the trailing comma intentional?

millerdev: Yes, it's a workaround for a common pitfall with `%` string formatting in Python. The comma makes it a single-item tuple containing `self.value`, which guarantees that the string formatting expression will always succeed (even if `self.value` is a multi-item tuple, for example, although I guess that's not much of a concern here).

orangejenny: aha, interesting

orangejenny (approved):

PR #17904, opened 2017-09-20
update bacilli count to be 9 in case its above 9

mkangia:
update bacilli count to be 9 in case its above 9 and include its count in the error message in case the notification fails

cc: buddy @esoergel

### DIFF
--- a/custom/enikshay/integrations/nikshay/repeater_generator.py
+++ b/custom/enikshay/integrations/nikshay/repeater_generator.py
@@ -299,13 +299,21 @@ class NikshayFollowupPayloadGenerator(BaseNikshayPayloadGenerator):
         lab_serial_number = test_case_properties.get('lab_serial_number')
         test_result_grade = test_case_properties.get('result_grade')
         bacilli_count = test_case_properties.get('max_bacilli_count')
+        try:
+            # Since 9 is the max value supported by Nikshay, notify 9 in case 9+
+            if bacilli_count and int(bacilli_count) > 9:
+                bacilli_count = '9'
+        except ValueError:
+            pass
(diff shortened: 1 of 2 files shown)


custom/enikshay/integrations/nikshay/repeater_generator.py
```diff
@@ -299,13 +299,21 @@ def _get_mandatory_fields(self, test_case, test_case_properties, occurence_case,
         lab_serial_number = test_case_properties.get('lab_serial_number')
         test_result_grade = test_case_properties.get('result_grade')
         bacilli_count = test_case_properties.get('max_bacilli_count')
+        try:
+            # Since 9 is the max value supported by Nikshay, notify 9 in case 9+
+            if bacilli_count and int(bacilli_count) > 9:
+                bacilli_count = '9'
+        except ValueError:
```
emord: Is this exception in case bacilli_count is a string or non-int? what are the expected values for this?

mkangia: @emord it would always be a string only and would mostly be an integer value only. But just added a safe in case its unable to parse the string value as an integer.

emord: if it's a non integer value is it safe to keep it as a non integer value? if it's safe, then I approve

mkangia: ya if its a non integer value, it would fail later with `NikshayRequiredValueMissing ` and would get listed as a failed repeat record where we can then track it.

esoergel (approved):

PR #17905, opened 2017-09-21
Keep questionMap for repeat contexts updated when refreshing questions

orangejenny:
https://manage.dimagi.com/default.asp?262193 and https://manage.dimagi.com/default.asp?261963

`questionMap` wasn't being updated when `questions` changed, resulting in repeat context getting set incorrectly.

@proteusvacuum 
cc @nickpell

### DIFF
--- a/corehq/apps/app_manager/static/app_manager/js/forms/case_config_ui.js
+++ b/corehq/apps/app_manager/static/app_manager/js/forms/case_config_ui.js
@@ -98,13 +98,19 @@ hqDefine('app_manager/js/forms/case_config_ui', function () {
             }
         });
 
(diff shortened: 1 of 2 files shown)


snopoke (commented):
Wow, I did not expect this to be a JS bug. 

Does this mean that it only occurred when uses use the 'refresh' button?

orangejenny:
Yeah, this was a weird one. This is definitely what was happening with the debug app in skelly - Jen was moving questions around in form builder in one tab, then using the refresh button in form settings in another tab and that was killing the repeat context. But she's still seeing issues with the L10K app, so there might be something else going on, too.

PR #17907, opened 2017-09-21
Mobile UCR v2

emord:
@snopoke @ctsims This has most of the code for apps to begin referencing the new mobile UCR fixtures.

todo list:
- [x] can select v1 vs v2 vs both on the app level
- [x] restores can selectively send the different formats in an app aware sync
- [x] non graph mobile UCRs refernce v2 fixtures
- [x] fixtures include indexed attribute
- [ ] change restore format in the UI
- [x] graphs reference fixtures
- [ ] purging old fixtures (updated the doc with a possible solution for this)

buddy @sravfeyn


corehq/apps/app_manager/fixtures/mobile_ucr.py
```diff
@@ -163,6 +202,59 @@ def report_config_to_fixture(report_config, restore_user):
...
+                random.randint(0, MOBILE_UCR_RANDOM_THRESHOLD) == MOBILE_UCR_RANDOM_THRESHOLD):
+            compare_ucr_dbs.delay(domain, report_config.report_id, filter_values)
+
+        report_elem = E.fixture(
+            id=report_fixture_id, user_id=restore_user.user_id,
+            report_id=report_config.report_id, last_sync=_utcnow().isoformat(),
+            indexed='true'
+        )
+        report_elem.append(rows_elem)
+        return [filters_elem, report_elem]
```
snopoke: Lots of duplicated code with v1

corehq/apps/app_manager/fixtures/mobile_ucr.py
```diff
@@ -215,5 +307,38 @@ def _row_to_row_elem(row, index, is_total_row=False):
...
+                dict(
+                    zip(
+                        map(lambda column_config: column_config.column_id, data_source.top_level_columns),
+                        map(str, total_row)
+                    )
+                ),
+                data_source.get_total_records(),
+                is_total_row=True,
+            ))
+        return rows_elem
```
snopoke: potential also for refactoring out duplicate code

corehq/apps/app_manager/suite_xml/features/mobile_ucr.py
```diff
@@ -30,9 +31,21 @@ class ReportModuleSuiteHelper(object):
...
         self.report_module = report_module
-        self.domain = self.report_module.get_app().domain
+        self.domain = self.app.domain
         self._loaded = None
 
+    @cached_property
+    def app(self):
+        return self.report_module.get_app()
+
+    @cached_property
```
snopoke: is caching necessary here? I don't think these (`mobile_ucr_restore_version` and `app`) are expensive in any way

custom/icds/messaging/indicators.py
```diff
@@ -63,7 +63,7 @@ def _get_report_fixture_for_user(domain, report_id, ota_user):
     :param report_id: the index to the result from get_report_configs()
     :param ota_user: the OTARestoreCommCareUser for which to get the report fixture
     """
-    xml = ReportFixturesProvider.report_config_to_fixture(
+    xml = ReportFixturesProvider.report_config_to_v1_fixture(
```
snopoke: fine to leave for separate PR but looks like this could quite easily use v2 fixtures without much code change

snopoke (commented):
Looking good. Do you know whether any apps reference the fixtures in forms that would need to be migrated?

corehq/apps/app_manager/fixtures/mobile_ucr.py
emord: Agreed. When initially writing this I copied it over, wrote a test and worked through differences. I can take a pass on being more DRY

corehq/apps/app_manager/suite_xml/features/mobile_ucr.py
emord: I've conditioned myself to view methods like `get_app()` to be expensive, but it appears to already be loaded. Will fix

emord:
> Do you know whether any apps reference the fixtures in forms that would need to be migrated?

I verified that there aren't any in the AWW or LS app. I haven't gone through each form yet. I would like to make an automated script to find those since I'd like to have this be the default in the future and there are a non trivial amount of these apps on prod

corehq/apps/app_manager/tests/data/fixtures/expected_v2_report.xml
```diff
@@ -0,0 +1,17 @@
+<restore>
+  <filters id="commcare-reports:deadbeef" />
+  <fixture id="commcare-reports:deadbeef" user_id="mock-user-id" report_id="deadbeef" last_sync="2017-09-11T06:35:20" indexed="true">
+    <rows>
+      <row index="0" is_total_row="False">
+        <bar>2</bar>
```
ctsims: Just to be clear: Is this expected to pass or fail?

On the phone this will fail out, since the structure isn't homogenous at the first level, I believe.

emord: Ignore the first <restore> tag. That is there because it is needed for this to be a valid XML tag. It'll actually be:

```
<OpenRosaResponse>
  <filters>...</filters>
  <fixture>
    <rows>
      <row>...</row>
    </rows>
  </fixture>
</OpenRosaResponse>

Which should be OK right?

Comment (ctsims) [reply corehq/apps/app_manager/tests/data/fixtures/expected_v2_report.xml]: that'll be just fine

Comment (ctsims) [reply corehq/apps/app_manager/tests/data/fixtures/expected_v2_report.xml]: Er, sorry, I think the point of confusion was actually what's _within_ row. That will be a homogenous set, right? Do you have a link to the actual output I can read?

Comment (emord) [reply corehq/apps/app_manager/tests/data/fixtures/expected_v2_report.xml]: @ctsims check out https://staging.commcarehq.org/hq/admin/phone/restore/?as=emord@jemordtest.commcarehq.org&version=2.0 look for fixtures with `commcare-reports`

Comment (ctsims) [reply corehq/apps/app_manager/tests/data/fixtures/expected_v2_report.xml]: Hey, what is

```
  <filters id="commcare-reports:a1cdf0edec45b8e842eca56e7eba465a"/>
```

in that document? Those aren't something the phone knows how to process...

This fixture from the report
```
  <fixture id="commcare-reports:c73dfed246044bcb8fceda0ba81ff87c" indexed="true" last_sync="2017-10-05T20:09:05.228426" report_id="c73dfed246044bcb8fceda0ba81ff87c" user_id="6ad3ce182215183a4a295b6ef1dffac8">
    <rows>
      <row index="0" is_total_row="False">
        <column_0>emord</column_0>
        <column_1>second test</column_1>
      </row>
      <row index="1" is_total_row="False">
        <column_0>emord</column_0>
        <column_1>test</column_1>
      </row>
    </rows>
  </fixture>
```

seems totally right though.

emord:
Going offline for the next week. This should be ready to merge once tests are fixed (but not ready to release). it's "feature flagged" on the app.

I'm having issues figuring out what xpath to use to reference these new fixtures. using a plan `instance('commcare-reports:uuid')` does not work. error was "parse error in xpath path [instance('commcare-reports:uuid')]" will likely need to offline with @ctsims or someone on mobile to figure out what i'm doing wrong.

Things left to do:

- [ ] keep track of deleted report configs and always send an empty fixture down for those
- [ ] provide way in the UI to select between v1, v1.5, v2 for the mobile UCR versions. (v1.5 provides both v1 and v2 fixutres)

after that we should test with enikshay to change all of their custom references (in a test space of course).

emord:
closing for now. will reopen soon

PR #17908, opened 2017-09-21
only update fields that have changed

snopoke:
https://manage.dimagi.com/default.asp?262252

Finding that cases are being updated even if there are no changes. Hopefully this will allow us to see what's causing the update.

### DIFF
--- a/corehq/apps/callcenter/utils.py
+++ b/corehq/apps/callcenter/utils.py
@@ -62,7 +62,6 @@ class _UserCaseHelper(object):
             transaction.form.archive()
 
     def create_user_case(self, commcare_user, fields):
-        fields['hq_user_id'] = commcare_user._id
         caseblock = CaseBlock(
             create=True,
             case_id=uuid.uuid4().hex,
@@ -173,6 +172,7 @@ def _get_user_case_fields(commcare_user, case_type, owner_id):
         'last_device_id_used': commcare_user.devices[0].device_id if commcare_user.devices else '',
         'owner_id': owner_id,
         'case_type': case_type,
+        'hq_user_id': commcare_user.get_id
     })
 
     return fields
--- a/corehq/apps/callcenter/utils.py
+++ b/corehq/apps/callcenter/utils.py
@@ -183,10 +183,10 @@ def _get_changed_fields(case, case_type, owner_id, fields):
         changed_fields['name'] = fields['name']
 
     if case.type != case_type:
-        fields['case_type'] = case_type
+        changed_fields['case_type'] = case_type
 
     if case.owner_id != owner_id:
-        fields['owner_id'] = owner_id
+        changed_fields['owner_id'] = owner_id
 
     props = case.dynamic_case_properties()
     for field, value in fields.items():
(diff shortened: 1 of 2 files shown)


corehq/apps/callcenter/utils.py
```diff
@@ -189,18 +178,23 @@ def valid_element_name(name):
...
 
 
-def _user_case_changed(case, case_type, close, fields, owner_id):
+def _get_changed_fields(case, case_type, owner_id, fields):
+    changed_fields = {}
+    if case.name != fields['name']:
+        changed_fields['name'] = fields['name']
+
+    if case.type != case_type:
+        fields['case_type'] = case_type
```
millerdev: This mutates an argument that was passed in, which could have unexpected side-effects elsewhere. I think you can change this to `changed_fields['case_type'] = case_type` and also change line 195 to `changed_fields.setdefault(field, value)` to avoid that (or use another approach if preferred).

corehq/apps/callcenter/utils.py
```diff
@@ -189,18 +178,23 @@ def valid_element_name(name):
...
+
+    if case.owner_id != owner_id:
+        fields['owner_id'] = owner_id
```
millerdev: same as previous comment: mutating passed-in argument

corehq/apps/callcenter/utils.py
```diff
@@ -189,18 +178,23 @@ def valid_element_name(name):
...
-    changed = changed or case.name != fields['name']
-    changed = changed or case.owner_id != owner_id
-    if not changed:
-        for field, value in fields.items():
-            if field != 'name' and props.get(field) != value:
-                changed = True
-                break
-    return changed
+    for field, value in fields.items():
+        if field != 'name' and props.get(field) != value:
```
millerdev: I think `field != 'name'` could be removed if the next line is changed to use `changed_fields.setdefault(...)`.

Also, is it ever possible that a field that is not in `props` will be intentionally set to `None`? If yes, might want to change this to `if field in not props or props.get(field) != value:`

snopoke: no, I don't think that would be desirable

snopoke:
@millerdev thanks for the review, some sloppy mistakes there.

corehq/apps/callcenter/utils.py
```diff
@@ -185,22 +172,28 @@ def valid_element_name(name):
...
+            changed_fields[field] = value
```
millerdev: This will overwrite `case_type` and `owner_id` with values from `fields` (if they exist) even if they were set above on lines 186 and 189. Is that intentional? (that's why I suggested using `setdefault`)

As is, I think this function has confusing semantics. It is not obvious what will be returned when called like
```py
changed = _get_changed_fields(case, "type_one", "bob", {
    "case_type": "type_two",
    "owner_id": "frank",
})
```
Maybe it should throw an exception if `"case_type"` or `"owner_id"` are present in `fields`? Or maybe those arguments should be removed from this function so the only way to set them is by passing them in the `fields` dict (this could result in a cleaner implementation too, both here and above in `sync_user_case`)?

snopoke: `case_type` and `owner_id` should never be in fields (though that's not obvious from the context of this function.

I've made and update to add `case_type` and `owner_id` into fields which simplifies things somewhat

millerdev (approved):

PR #17910, opened 2017-09-21
Mk add new data sources

mkangia:
1. Update test data sources to include notification related fields for new report to be added
2. Update/Add person data source to check for hiv status updates

best reviewed 🐟 / 🐟


custom/enikshay/ucr/data_sources/test_2b_v2.json
```diff
@@ -0,0 +1,3377 @@
+{
+    "domains": [
+        "enikshay",
```
emord: this is called "test_2b_v2" If it's in testing does it need to be in the enikshay domain, or can it be kept in the QA domain for now and added to enikshay later?

custom/enikshay/ucr/data_sources/test_2b_v2.json
```diff
@@ -0,0 +1,3377 @@
...
+        "sheel-enikshay",
+        "enikshay-reports-qa",
+        "enikshay-performance-test"
+    ],
+    "server_environment": [
+        "enikshay",
+        "softlayer"
```
emord: I don't think these need to be added to softlayer anymore. Is there any testing going on in that cluster

emord (commented):
I don't have any context for the calculations. You'll also need to add this file path to settings.py

custom/enikshay/ucr/data_sources/test_2b_v2.json
mkangia: sorry. removing this.

custom/enikshay/ucr/data_sources/test_2b_v2.json
mkangia: its needs to run on the real data. so should be under enikshay?

proteusvacuum: it's the `test` case type :smile:

proteusvacuum (changes requested):
I think you need to add this to `settings.py`?

custom/enikshay/ucr/data_sources/test_2b_v2.json
mkangia: oh ya.. test is the case type :)

mkangia:
@proteusvacuum thank you for the settings miss. Added.

nickpell (approved):

proteusvacuum (approved):

PR #17913, opened 2017-09-21
Remove redundant location_id validation from agency form

esoergel:
https://manage.dimagi.com/default.asp?258965
@mkangia

### DIFF
--- a/corehq/apps/locations/forms.py
+++ b/corehq/apps/locations/forms.py
@@ -334,11 +334,17 @@ class LocationForm(forms.Form):
         return location
 
 
+class LocationUserForm(NewMobileWorkerForm):
+    def clean_location_id(self):
+        # This is handled separately
+        return None
+
+
 class LocationFormSet(object):
     """Ties together the forms for location, location data, user, and user data."""
     _location_form_class = LocationForm
     _location_data_editor = CustomDataEditor
-    _user_form_class = NewMobileWorkerForm
+    _user_form_class = LocationUserForm
     _user_data_editor = CustomDataEditor
 
     def __init__(self, location, request_user, is_new, bound_data=None, *args, **kwargs):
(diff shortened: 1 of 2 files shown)


corehq/apps/locations/forms.py
```diff
@@ -334,11 +334,17 @@ def save(self, metadata):
         return location
 
 
+class LocationUserForm(NewMobileWorkerForm):
+    def clean_location_id(self):
+        # This is handled separately
```
mkangia: Can we include the reason why we are doing this a little in detail? That would help understand the need of this change quickly and better when someone sees it.

custom/enikshay/user_setup.py
```diff
@@ -266,12 +266,6 @@ def connect_signals():
     commcare_user_post_save.connect(save_user_callback, dispatch_uid="save_user_callback")
 
 
-class ENikshayNewMobileWorkerForm(NewMobileWorkerForm):
-    """Mobile worker list view create modal"""
-    def __init__(self, *args, **kwargs):
-        super(ENikshayNewMobileWorkerForm, self).__init__(*args, **kwargs)
-
-
```
mkangia: Why was  `ENikshayNewMobileWorkerForm ` even needed?
You removed it because it was not doing anything special?

mkangia (approved):
Don't understand the overall impact of the change but otherwise looks good. No blocker, just couple of comments

custom/enikshay/user_setup.py
esoergel: I think it used to do something which later got moved elsewhere.  At this point it's not doing anything, though.

corehq/apps/locations/forms.py
esoergel: updated

PR #17914, opened 2017-09-21
add fb pixel code to base registration template

biyeun:
buddy @proteusvacuum 

needs to be merged today

### DIFF
--- a/corehq/apps/registration/templates/registration/base.html
+++ b/corehq/apps/registration/templates/registration/base.html
@@ -14,7 +14,7 @@
     fbq('track', 'PageView');
     </script>
     <noscript><img height="1" width="1" style="display:none"
-    src="https://www.facebook.com/tr?id=301081893674207&ev=PageView&noscript=1"
+    src="https://www.facebook.com/tr?id={{ FACEBOOK_PIXEL_ID }}&ev=PageView&noscript=1"
     /></noscript>
     <!-- DO NOT MODIFY -->
     <!-- End Facebook Pixel Code -->
--- a/corehq/apps/registration/templates/registration/base.html
+++ b/corehq/apps/registration/templates/registration/base.html
@@ -2,7 +2,7 @@
 
 {% block head %}{{ block.super }}
     {% include 'analytics/fullstory.html' %}
-    {% if FACEBOOK_PIXEL_ID %}
+    {% if is_production %}
     <!-- Facebook Pixel Code -->
     <script>
     !function(f,b,e,v,n,t,s){if(f.fbq)return;n=f.fbq=function(){n.callMethod?
@@ -10,11 +10,11 @@
     n.push=n;n.loaded=!0;n.version='2.0';n.queue=[];t=b.createElement(e);t.async=!0;
     t.src=v;s=b.getElementsByTagName(e)[0];s.parentNode.insertBefore(t,s)}(window,
     document,'script','https://connect.facebook.net/en_US/fbevents.js');
-    fbq('init', '{{ FACEBOOK_PIXEL_ID }}');
+    fbq('init', '301081893674207');
     fbq('track', 'PageView');
     </script>
     <noscript><img height="1" width="1" style="display:none"
-    src="https://www.facebook.com/tr?id={{ FACEBOOK_PIXEL_ID }}&ev=PageView&noscript=1"
+    src="https://www.facebook.com/tr?id=301081893674207&ev=PageView&noscript=1"
     /></noscript>
     <!-- DO NOT MODIFY -->
     <!-- End Facebook Pixel Code -->
--- a/corehq/apps/registration/templates/registration/base.html
+++ b/corehq/apps/registration/templates/registration/base.html
@@ -6,7 +6,7 @@
     <!-- Facebook Pixel Code -->
     <script>
     !function(f,b,e,v,n,t,s){if(f.fbq)return;n=f.fbq=function(){n.callMethod?
-    n.callMeyythod.apply(n,arguments):n.queue.push(arguments)};if(!f._fbq)f._fbq=n;
+    n.callMethod.apply(n,arguments):n.queue.push(arguments)};if(!f._fbq)f._fbq=n;
     n.push=n;n.loaded=!0;n.version='2.0';n.queue=[];t=b.createElement(e);t.async=!0;
     t.src=v;s=b.getElementsByTagName(e)[0];s.parentNode.insertBefore(t,s)}(window,
     document,'script','https://connect.facebook.net/en_US/fbevents.js');
--- a/corehq/apps/registration/templates/registration/base.html
+++ b/corehq/apps/registration/templates/registration/base.html
@@ -2,6 +2,23 @@
 
 {% block head %}{{ block.super }}
     {% include 'analytics/fullstory.html' %}
+    {% if FACEBOOK_PIXEL_ID %}
+    <!-- Facebook Pixel Code -->
+    <script>
+    !function(f,b,e,v,n,t,s){if(f.fbq)return;n=f.fbq=function(){n.callMethod?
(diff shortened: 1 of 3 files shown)


corehq/apps/registration/templates/registration/base.html
```diff
@@ -2,6 +2,21 @@
...
 {% block head %}{{ block.super }}
     {% include 'analytics/fullstory.html' %}
+    <!-- Facebook Pixel Code -->
+    <script>
+    !function(f,b,e,v,n,t,s){if(f.fbq)return;n=f.fbq=function(){n.callMethod?
+    n.callMeyythod.apply(n,arguments):n.queue.push(arguments)};if(!f._fbq)f._fbq=n;
+    n.push=n;n.loaded=!0;n.version='2.0';n.queue=[];t=b.createElement(e);t.async=!0;
+    t.src=v;s=b.getElementsByTagName(e)[0];s.parentNode.insertBefore(t,s)}(window,
+    document,'script','https://connect.facebook.net/en_US/fbevents.js');
+    fbq('init', '301081893674207');
```
proteusvacuum: Is this a user id / password? If so, it should probably go into a settings file.
We also shouldn't be including this on enikshay / icds / other enterprise sites, right?

proteusvacuum (changes requested):

corehq/apps/registration/templates/registration/base.html
biyeun: yep. I just copied the code from an email. I'll go ahead and update. good catch

proteusvacuum: should probably get a new ID now that this will be in our git history.

biyeun:
@proteusvacuum amended for git history erasure 😳

corehq/apps/registration/templates/registration/base.html
```diff
@@ -2,6 +2,23 @@
...
+    !function(f,b,e,v,n,t,s){if(f.fbq)return;n=f.fbq=function(){n.callMethod?
+    n.callMeyythod.apply(n,arguments):n.queue.push(arguments)};if(!f._fbq)f._fbq=n;
+    n.push=n;n.loaded=!0;n.version='2.0';n.queue=[];t=b.createElement(e);t.async=!0;
+    t.src=v;s=b.getElementsByTagName(e)[0];s.parentNode.insertBefore(t,s)}(window,
+    document,'script','https://connect.facebook.net/en_US/fbevents.js');
+    fbq('init', '{{ FACEBOOK_PIXEL_ID }}');
+    fbq('track', 'PageView');
+    </script>
+    <noscript><img height="1" width="1" style="display:none"
+    src="https://www.facebook.com/tr?id=301081893674207&ev=PageView&noscript=1"
```
proteusvacuum: This looks like the same ID that was in the `init` bit before

proteusvacuum (changes requested):
Unfortunately, you can't just `git amend` this out, as the commit still exists: https://github.com/dimagi/commcare-hq/commit/0bcba4096211767e92939ee89b1ed53ba1132382

I think you might have to actually change the ID in facebook.

corehq/apps/registration/templates/registration/base.html
```diff
@@ -2,6 +2,23 @@
 
 {% block head %}{{ block.super }}
     {% include 'analytics/fullstory.html' %}
+    {% if FACEBOOK_PIXEL_ID %}
```
gcapalbo: this needs to be loaded into the template context right?

corehq/apps/registration/templates/registration/base.html
```diff
@@ -2,6 +2,23 @@
...
+    <!-- Facebook Pixel Code -->
+    <script>
+    !function(f,b,e,v,n,t,s){if(f.fbq)return;n=f.fbq=function(){n.callMethod?
+    n.callMeyythod.apply(n,arguments):n.queue.push(arguments)};if(!f._fbq)f._fbq=n;
```
proteusvacuum: this is weird, but is this actually `callMeyythod`? 
The pixel docs seem to say this should be `callMethod`.

corehq/apps/registration/templates/registration/base.html
proteusvacuum: Does this need to be defined somewhere?

proteusvacuum (changes requested):

corehq/apps/registration/templates/registration/base.html
biyeun: there is middleware that handles this. I tested it locally

corehq/apps/registration/templates/registration/base.html
biyeun: I copied and pasted this code from comms. what about `n.callMethod?` right before? Looks like legacy code

biyeun: ...as in. I didn't consult the docs for this. I assume Comms got it from a generator.

corehq/apps/registration/templates/registration/base.html
biyeun: middleware takes care of it

biyeun:
@proteusvacuum also this https://github.com/dimagi/commcarehq-ansible/pull/955

corehq/apps/registration/templates/registration/base.html
biyeun: https://github.com/dimagi/commcarehq-ansible/pull/955

biyeun: there appears to be middleware taking care of this, as `GOOGLE_ANALYTICS_API_ID` is not defined anywhere in a template context, yet is accessed the same way from a template.

biyeun: I also tested locally and it works.

corehq/apps/registration/templates/registration/base.html
calellowitz: @biyeun I think Farid is correct. That looks like a copy paste error on my part. Not sure how it got there, but its not in the original version carrie sent

biyeun: added it to the default blank localsettings.py dict so people know it's available.

gcapalbo:
isn't this FACEBOOK_PIXEL_ID going to be loaded on the registration page (which is public facing)? how come we need to add it to localsettings?

proteusvacuum:
sorry for all the comments @biyeun ... did you mean to do a submodule update in the last commit?

biyeun:
@gcapalbo that's a great question :joy: that's why I didn't think much about it when I just copy pasted the code...
I guess for best practices. but having the ID in there allows you to selectively show it only when it's present in localsettings, so enikshay for instance won't see this.

biyeun:
@proteusvacuum omg 😳

biyeun:
@proteusvacuum i'm so done with this PR. this change should have taken 5 mins, and now i'm diving into ansible :joy: 🙄

gcapalbo:
i think we should get into the habit of using settings.ENTERPRISE_MODE to filter this stuff out

biyeun:
@proteusvacuum @calellowitz confirmed that somewhere down the line someone may have mistakenly typed some `yy`s.....

biyeun:
@gcapalbo but what if enikshay wanted some facebook tracking? 🙄

biyeun:
note: decided to go with @gcapalbo suggestion to use `settings.SERVER_ENVIRONMENT == 'production'` and skip the ansible shenanigans. THANK YOU. 🙏

proteusvacuum (approved):
sorry for the back and forth.

biyeun:
@proteusvacuum all valid points tho, and you caught the `callMeyythod` :joy: :rofl: maybe it should have been typoed to `calMeyythod` 😜  @calellowitz

PR #17917, opened 2017-09-21
Duplicate person and voucher id report

esoergel:
Simple report to get some visibility into what cases have duplicate ids, and how many such cases there are.
@mkangia

### DIFF
--- a/custom/enikshay/reports/views.py
+++ b/custom/enikshay/reports/views.py
@@ -10,8 +10,8 @@ from corehq.apps.domain.decorators import domain_admin_required
 from corehq.apps.locations.permissions import location_safe
 from corehq.apps.userreports.reports.filters.choice_providers import ChoiceQueryContext, LocationChoiceProvider
 from corehq.form_processor.interfaces.dbaccessors import CaseAccessors
-from custom.enikshay.case_utils import CASE_TYPE_VOUCHER
-from custom.enikshay.const import VOUCHER_ID
+from custom.enikshay.case_utils import CASE_TYPE_VOUCHER, CASE_TYPE_PERSON
+from custom.enikshay.const import VOUCHER_ID, ENIKSHAY_ID
 from custom.enikshay.reports.utils import StubReport
 
 
@@ -49,24 +49,28 @@ class LocationsView(View):
 
 
 @domain_admin_required
-def duplicate_ids_report(request, domain):
+def duplicate_ids_report(request, domain, case_type):
+    case_type = {'voucher': CASE_TYPE_VOUCHER, 'person': CASE_TYPE_PERSON}[case_type]
+    id_property = {'voucher': VOUCHER_ID, 'person': ENIKSHAY_ID}[case_type]
+
     accessor = CaseAccessors(domain)
-    voucher_ids = accessor.get_case_ids_in_domain(CASE_TYPE_VOUCHER)
-    all_vouchers = list(accessor.iter_cases(voucher_ids))
-    counts = Counter(voucher.get_case_property(VOUCHER_ID) for voucher in all_vouchers)
-    bad_vouchers = [
+    case_ids = accessor.get_case_ids_in_domain(case_type)
+    all_cases = list(accessor.iter_cases(case_ids))
+    counts = Counter(case.get_case_property(id_property) for case in all_cases)
+    bad_cases = [
         {
-            'case_id': voucher.case_id,
-            'readable_id': voucher.get_case_property(VOUCHER_ID)
+            'case_id': case.case_id,
+            'readable_id': case.get_case_property(id_property)
         }
-        for voucher in all_vouchers
-        if counts[voucher.get_case_property(VOUCHER_ID)] > 1
+        for case in all_cases
+        if counts[case.get_case_property(id_property)] > 1
     ]
 
     context = {
-        'num_bad_vouchers': len(bad_vouchers),
-        'num_total_vouchers': len(all_vouchers),
-        'num_good_vouchers': len(all_vouchers) - len(bad_vouchers),
-        'bad_vouchers': bad_vouchers,
+        'case_type': case_type,
+        'num_bad_cases': len(bad_cases),
+        'num_total_cases': len(all_cases),
+        'num_good_cases': len(all_cases) - len(bad_cases),
+        'bad_cases': bad_cases,
     }
     return render(request, 'enikshay/duplicate_ids_report.html', context)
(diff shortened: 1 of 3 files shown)


custom/enikshay/reports/views.py
```diff
@@ -39,3 +46,31 @@ def get(self, request, domain, *args, **kwargs):
                 'total': location_choice_provider.query_count(query_context.query, user)
             }
         )
+
+
+@domain_admin_required
+def duplicate_ids_report(request, domain, case_type):
```
mkangia: Will this `domain` argument ever have value other than "enikshay"? If not then this can be a constant only.

custom/enikshay/reports/views.py
```diff
@@ -39,3 +46,31 @@ def get(self, request, domain, *args, **kwargs):
...
+    accessor = CaseAccessors(domain)
+    case_ids = accessor.get_case_ids_in_domain(case_type)
+    all_cases = [
+        {
+            'case_id': case.case_id,
+            'readable_id': case.get_case_property(id_property),
+            'opened_on': case.opened_on,
+        }
+        for case in accessor.iter_cases(case_ids)
+    ]
```
mkangia: Would this eventually become a very long list in case of many records. 
Storing all cases and then iterating on them twice could hurt. 

You probably can store the corresponding case_id(s) for a readable_id and just load those cases to find the bad cases, this would save one/two iterations over all cases. May be something like:
```
bad_case_ids = []
id_property_with_case_ids = {}
id_property_with_case_ids = defaultdict{lambda:[], id_property_with_case_ids}
for case in accessor.iter_cases(case_ids):
    id_prop_value = case.get_case_property(id_property)
    if id_prop_value:  # to avoid empty values
        id_property_with_case_ids[case.get_case_property(id_property)].append(case.case_id)
for id_prop_value in id_property_with_case_ids:  # iterate over just non-empty values and not all the cases
    if len(id_property_with_case_ids[id_prop_value]) > 2:
        bad_case_ids.append(id_property_case_ids[id_prop_value])

bad_cases = accessor.get_cases(bad_case_ids)
```

mkangia (changes requested):

custom/enikshay/reports/views.py
```diff
@@ -39,3 +46,31 @@ def get(self, request, domain, *args, **kwargs):
...
         )
+
+
+@domain_admin_required
+def duplicate_ids_report(request, domain, case_type):
+    case_type = {'voucher': CASE_TYPE_VOUCHER, 'person': CASE_TYPE_PERSON}[case_type]
+    id_property = {'voucher': VOUCHER_ID, 'person': ENIKSHAY_ID}[case_type]
+
+    accessor = CaseAccessors(domain)
+    case_ids = accessor.get_case_ids_in_domain(case_type)
```
snopoke: getting all case_ids in the domain isn't a great idea. I would suggest modifying the `SqlDomainCaseChangeProvider` (or writing something similar) to return the cases directly.

esoergel: Question about `SqlDomainCaseChangeProvider`:
```python
    def iter_all_changes(self, start_from=None):
        accessor = CaseReindexAccessor(self.domain, limit_db_aliases=self.limit_db_aliases)
        for db_alias in accessor.sql_db_aliases:
            cases = accessor.get_docs(db_alias, start_from)
            while cases:
                for case in cases:
                    yield _sql_case_to_change(case)

                start_from = case.server_modified_on
                last_id = case.id
                cases = accessor.get_docs(db_alias, start_from, last_doc_pk=last_id)
```
Why does it overwrite the `start_from` param inside the `for` loop?  Are the cases sharded by date, where all cases in a subsequent shard come after all cases in the previous shard?  If cases are assigned to shards arbitrarily, then I'd expect shards after the first to be accessed with a `start_from` equal to the most recent case modification of the previous shard.

This is also an internal-only view used for debugging, so I wasn't thinking too much about performance.  If you're concerned about the load on the server, I could also write this to instead use an ES terms aggregation with [minimum_document_count = 2](https://www.elastic.co/guide/en/elasticsearch/reference/master/search-aggregations-bucket-terms-aggregation.html#_minimum_document_count_4).  I wrote it this way to be sure of the exactness of the results, and because with the current case load it completes in just a few seconds.

esoergel: Also, I'm restricting results by case type, which looks like that function doesn't provide for.  I'd expect this to be on the order of 100k cases, do you think that's an acceptable size to load all IDs at once?

custom/enikshay/reports/views.py
esoergel: It transforms each case into a 3 element dict and then takes two more passes over that, so there's never a list of the full cases in memory.  Think that's fine?

esoergel: Also, for scale, this should be on the order of 100k elements total, and I think a few thousand with duplicates.

mkangia: `all_cases` has dict for all cases in domain right? I didnt mean all cases in memory but just a big object anyway.
Memory was one point but also the number of iterations.

Plus if there is a way to reduce the number of iterations to speed this up, we should do it anyway?
Also, Do we need to skip blank values?

custom/enikshay/reports/views.py
mkangia: oh okay! Cool.
Thank you.

custom/enikshay/reports/views.py
esoergel: I'm not convinced.  With python primitives, this should be quite small.  I just did some back-of-the envelope testing to validate:
```python
In [1]: import timeit

In [2]: a = [{'foo': 'bar', 'baz': 'bang', 'bip': 'fuzz'} for i in range(100000)]

In [3]: a.__sizeof__()
Out[3]: 824440

In [4]: def walk():
   ...:     for x in a:
   ...:         pass
   ...:     

In [5]: timeit.timeit(walk, number=100)
Out[5]: 0.32859110832214355
```
That is, this list of 100k dicts of 3 elements each has a size less than a MB.  Iterating over that list 100 times takes about a third of a second.  For a debugging report that will be used only infrequently, I think that is acceptable.  For something like this, I'd heavily weight code clarity over marginal perfomance gains.

esoergel: Regarding blank values - I think it makes sense to keep those in, as that's also an error AFAIK, so it couldn't hurt to have some visibility into those.

mkangia: Thank you Ethan. 🙇

mkangia (approved):

custom/enikshay/reports/views.py
snopoke: Sorry I missed the `case_type` filter. Seems like you've thought through the scale of this so fine to go ahead as is.

Also thanks for the bug spotting: https://github.com/dimagi/commcare-hq/pull/17964

PR #17918, opened 2017-09-21
Add name to location fixtures index

esoergel:
https://trello.com/c/lgmg71xz/222-locations-fixture-index-updates

### DIFF
--- a/corehq/apps/locations/fixtures.py
+++ b/corehq/apps/locations/fixtures.py
@@ -120,7 +120,7 @@ class FlatLocationSerializer(object):
         )
         location_type_attrs = ['{}_id'.format(t) for t in all_types if t is not None]
         attrs_to_index = ['@{}'.format(attr) for attr in location_type_attrs]
-        attrs_to_index.extend(['@id', '@type'])
+        attrs_to_index.extend(['@id', '@type', 'name'])
 
         return [get_index_schema_node(fixture_id, attrs_to_index),
                 self._get_fixture_node(fixture_id, restore_user, locations_queryset, location_type_attrs, data_fields)]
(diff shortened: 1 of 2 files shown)


esoergel:
@mkangia @ctsims

mkangia:
The build failure was not related.. retriggered to confirm if that was a temp error

corehq/apps/locations/fixtures.py
```diff
@@ -120,7 +120,7 @@ def get_xml_nodes(self, fixture_id, restore_user, locations_queryset, data_field
         )
         location_type_attrs = ['{}_id'.format(t) for t in all_types if t is not None]
         attrs_to_index = ['@{}'.format(attr) for attr in location_type_attrs]
-        attrs_to_index.extend(['@id', '@type'])
+        attrs_to_index.extend(['@id', '@type', 'name'])
```
mkangia: What decides if we need '@' with the name or not?

mkangia (approved):
Looks good. Just one minor point to clarify

corehq/apps/locations/fixtures.py
ctsims: These are acccessors into the xml element, so anything with an @ is an attribute of the root element, and things with no @ are child elements

esoergel: Whether it's an xml attribute or a child node.  If you expand the example test file, you'll see that id, type, and the others are attrs, and name is a child node.

esoergel:
Gonna close in favor of https://github.com/dimagi/commcare-hq/pull/17921, which builds off of this

corehq/apps/locations/fixtures.py
mkangia: Much Thanks! 👍

PR #17921, opened 2017-09-21
Add name and some custom data to location fixture

esoergel:
There's not yet a UI to turn this on for the custom data fields, but this will let us test it out, at least.
https://trello.com/c/lgmg71xz/222-locations-fixture-index-updates
@mkangia @ctsims

### DIFF
--- a/corehq/apps/custom_data_fields/models.py
+++ b/corehq/apps/custom_data_fields/models.py
@@ -40,6 +40,8 @@ class CustomDataField(JsonObject):
     label = StringProperty()
     choices = StringListProperty()
     is_multiple_choice = BooleanProperty(default=False)
+    # Currently only relevant for location fields
+    index_in_fixture = BooleanProperty(default=False)
 
 
 class CustomDataFieldsDefinition(Document):
--- a/corehq/apps/custom_data_fields/edit_model.py
+++ b/corehq/apps/custom_data_fields/edit_model.py
@@ -99,6 +99,7 @@ class CustomDataModelMixin(object):
     template_name = "custom_data_fields/custom_data_fields.html"
     field_type = None
     show_purge_existing = False
+    show_index_in_fixture = False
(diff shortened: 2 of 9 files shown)


corehq/apps/custom_data_fields/models.py
```diff
@@ -40,6 +40,8 @@ class CustomDataField(JsonObject):
     label = StringProperty()
     choices = StringListProperty()
     is_multiple_choice = BooleanProperty(default=False)
+    # Currently only relevant for location fields
+    index_in_fixture = BooleanProperty(default=False)
```
mkangia: nit: Can we add comment to share what happen when this is set to true. I don't think index_in_fixture is able to tell exactly what this would do unless someone understands fixtures well.

mkangia (approved):
Looks good. Nice tests.. Just a minor suggestion.

corehq/apps/custom_data_fields/models.py
esoergel: I just added some descriptions to the feature flag and the UI

esoergel:
@mkangia just added a feature flagged UI to modify this - was planning on PRing that separately later, but it was actually pretty quick.

corehq/apps/custom_data_fields/models.py
mkangia: Cool. We should write a note to refer to the FF then. Worried though in case the FF is removed in future, the context would go.

mkangia:
@esoergel Looks good.

corehq/apps/custom_data_fields/models.py
esoergel: What kind of context are you looking for?  This flag tells whether to index the field in the restore, so I thought `index_in_fixture` describes it pretty well.  The UI where this flag is turned on also has help text regarding the broader implications of toggling this.

PR #17925, opened 2017-09-22
Diagnosis Test Migration: Include yes_private and filter private patients

sheelio:
@nickpell 

Based on feedback from Jordan, this updates the diagnosis test information for patients initiated in private (who were registered in public sector).


sheelio:
@nickpell 

This should probably re-run after you testing facility name migration.

PR #17927, opened 2017-09-22
Limit enikshay location uploads for the time being

gcapalbo:
@calellowitz


calellowitz:
50 seems a little low. Based on when these uploads exploded in the past, shouldn't 1k be safe. I would have defaulted to 5k but I see how that might be riskier. My understanding is that 50 is pretty much not usable.

gcapalbo:
on Sep 12, a location import of about 500 locations caused a 25% jump in disk usage and it took about 4 hours from the start of the process before it was back to normal. since the issue only gets worse as more locations are added, and since this is only temporary until a permanent solution is in place, i think it's important to be cautious here. for now i put the limit at 100 every 4 hours

calellowitz (approved):

PR #17930, opened 2017-09-22
Enable livequery for subset of ICDS users

millerdev:
Does this seem like a good way to roll it out incrementally? I tried to pick a number high enough that we'd see restore times/errors jump if there are issues, but low enough to not cause total mayhem if things go wrong (I don't expect that, just being cautious). Should we start with a lower percentage?

@snopoke


snopoke (approved):

snopoke:
not merging due to label

PR #17931, opened 2017-09-22
ICDS: Reduce Iteration Count

sheelio:
@emord 

This PR does a few things:

- Reduces the iteration count for the tableau data sources.  It was -3 -> +3 months, its now just -1->+1 month.  It also updates the iteration for the aggregation.  
- Updates the data sources to iterate based on today's date instead of modified_on
- Removes the tests (since they're really tied to the # of iteration counts and would be a nightmare to update). 

I tested this by comparing the output of this version of the data source to the existing data sources.  I filtered the existing data source's iteration counts and also filtered them to rows with open_in_month = 1 (and they matched exactly).


emord:
Generally happy about moving in this direction, but I'm a bit hesitant on removing these tests. I'm going to think about this one a bit more

sheelio:
Are you hoping the tests can help catch UCR core bugs, or bugs in the data sources?  

(If its UCR core, we could potentially still keep them running against the old data sources, but not have those data sources deployed for ICDS).

emord:
I was hoping it could catch bugs in the data sources

custom/icds_reports/tasks.py
```diff
@@ -29,7 +29,7 @@ def run_move_ucr_data_into_aggregation_tables_task(date=None):
 
 
 @serial_task('move-ucr-data-into-aggregate-tables', timeout=30 * 60, queue='background_queue')
-def move_ucr_data_into_aggregation_tables(date=None, intervals=3):
+def move_ucr_data_into_aggregation_tables(date=None, intervals=2):
```
emord: This could be an unfortunate consequence. May need to flag with the dashboard team to see how much historical data is required when writing new indicators

sheelio: Hmm.  It would be easy to iterate from -2 months -> +1 months if that's preferable.

sheelio:
@emord 
Made a similar set of changes for the MPR data sources (so they only iterate from -3 months to +1 months).

emord:
@sheelio I took a pass at getting these tests to work with the 3 months. I fixed a few. it does look annoying to fix, but I don't think it'll be too difficult to do. some may need to slightly change the dates (i would need to go back and check my work. was just seeing how feasible it is)

I probably would have been immediately ok with removing these ~1 month ago, but the QA on the dashboards has been pretty painful, so I'm reluctant to get rid of tests that could catch something there

sheelio:
Ok.  I'll take a pass on the tests now.

emord:
Also I am not sure how this will effect the ongoing QA so I don't think it's urgent. the QA is currently relying on the fact that these calculate based on modified_on (they are submitting forms for back in May/June/July) I'll offline with the dashboard team to see what is left on this plan

sheelio:
@emord 
All the tests should be passing now.

sheelio:
@emord 

is this good to merge now that QA is done?

emord:
> QA is done

@sheelio what gave you that idea? I don't think its done

sheelio:
Ah, I might just be mistaken then (I thought we shared the new dashboard with stakeholders last week).

emord:
Yeah. that happened, but QA is still ongoing unfortunately

PR #17932, opened 2017-09-22
Set expected doses taken even if before purge date

proteusvacuum:
@calellowitz 
I hadn't written this in the spec, but this property needed to be set always, not just when the purge date was passed. 
Also this was using the wrong date (`date.today()` instead of the date in India which led to inconsistencies in the data)

cc: @biyeun

### DIFF
--- a/custom/enikshay/tasks.py
+++ b/custom/enikshay/tasks.py
@@ -389,6 +389,7 @@ class EpisodeAdherenceUpdate(object):
                 'one_week_adherence_score': 0,
                 'two_week_adherence_score': 0,
                 'month_adherence_score': 0,
+                'total_expected_doses_taken': self.get_total_expected_doses_taken(adherence_schedule_date_start),
             })
 
         adherence_cases = self.get_valid_adherence_cases()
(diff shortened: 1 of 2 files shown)


custom/enikshay/tasks.py
```diff
@@ -483,25 +484,34 @@ def get_aggregated_scores(self, latest_adherence_date, adherence_schedule_date_s
...
         doses_per_week = dose_data.get(adherence_schedule_id)
-        if doses_per_week:
-            update['expected_doses_taken'] = int(((
-                (update['aggregated_score_date_calculated'] - adherence_schedule_date_start)).days / 7.0
-            ) * doses_per_week)
-        else:
-            update['expected_doses_taken'] = 0
+        if not doses_per_week:
             soft_assert(notify_admins=True)(
                 True,
```
gcapalbo: i realize this didn't change in this pr, but setting this True means the soft assert never fires?

custom/enikshay/tasks.py
```diff
@@ -483,25 +484,34 @@ def get_aggregated_scores(self, latest_adherence_date, adherence_schedule_date_s
             end_date=update["aggregated_score_date_calculated"]
         )
 
+        doses_per_week = self.get_doses_per_week()
+        update['expected_doses_taken'] = int(((
+            (update['aggregated_score_date_calculated'] - adherence_schedule_date_start)).days / 7.0
```
gcapalbo: does this date range need to be inclusive? if so it should add 1

custom/enikshay/tasks.py
```diff
@@ -483,25 +484,34 @@ def get_aggregated_scores(self, latest_adherence_date, adherence_schedule_date_s
...
-        update['total_expected_doses_taken'] = total_expected_doses_taken
+            return 0
+        return doses_per_week
 
-        return update
+    def get_total_expected_doses_taken(self, adherence_schedule_date_start):
+        doses_per_week = self.get_doses_per_week()
+        today = self.date_today_in_india
+        total_expected_doses_taken = int(
+            round(((today - adherence_schedule_date_start).days / 7.0) * doses_per_week))
```
gcapalbo: this date range is also not inclusive, but maybe that's intentional

proteusvacuum: Good question. I think this is OK as is. The task runs first thing in the morning ("midnight").
If `adherence_schedule_date_start` was yesterday, we assume that you should have only taken 1 dose at this point. i.e. `today-adherence_schedule_date_start` should = 1

custom/enikshay/tasks.py
proteusvacuum: good catch: https://github.com/dimagi/commcare-hq/pull/17932/commits/cdea40b9a95c59b267cee3be8b2b789cac1205d6

custom/enikshay/tasks.py
gcapalbo: makes sense. does the other date range i pointed out need to be inclusive?

proteusvacuum: hmm, I didn't actually change that one in this PR (just moved it). I think the different cases of `aggregated_score_date_calculated` should actually do different things here, but I'm going to check and will update in a separate PR if needbe.

gcapalbo (approved):

custom/enikshay/tasks.py
proteusvacuum: This `expected_doses_taken` property is a count of all of the doses that should have been taken up until the date that cases are automatically closed (i.e. the "purge date"), or the date of the last adherence case, whichever is later. We don't want to count cases that remain open, as this would ruin the calculations on the phone. 

**aggregated_score_date_calculated = purge date**
- Today is `2017-01-31`. The purge_date is then `2017-01-01`
- Cases get closed at midnight of the 30th day after their `adherence_date` property here: https://enikshay.in/a/enikshay/data/edit/automatic_updates/edit/152/
    - If today is `2017-01-31`, then all cases where the adherence_date property is `2017-01-01` and earlier will be closed. i.e. all cases where `adherence_date` = `purge_date` are closed. 
- If the `adherence_schedule_date_start` is `2016-01-01`, then we want to count the number of days between `[2016-01-01, 2017-01-01]` = 366 (since 2016 was a leap year). 
- i.e. This should be inclusive. 

**aggregate_score_date_calculated = latest_adherence_date**
- Today is `2017-01-31`.
- The latest adherence case we received is `2016-12-31`
- The `adherence_schedule_date_start` is `2016-01-01`
- We want to count all the cases within the window [2016-01-01, 2016-12-31]. i.e. we also count the cases that occur on the `latest_adherence_date`. i.e. 31 days.
- Conclusion: this should also be inclusive.

proteusvacuum: https://github.com/dimagi/commcare-hq/pull/18144

PR #17935, opened 2017-09-23
Mk update follow up test and Hiv Test forwarders

mkangia:
1. Skip associated episode related checks when checking for forwarding to avoid silently failing
2. Use associated episode case for tests if available
3. Dont check for open occurrence case for test since its not needed and blocks notifications in case the occurrence has been closed with the same form as submitted for test update, [example case](https://enikshay.in/a/enikshay/reports/case_data/39381ac3-f5fc-4813-8050-0cc93c63937a/#!history?form_id=23242cd1-608b-48ef-8132-d00b8fc8d300)

cc: buddy @esoergel 

best review 🐟 / 🐟

### DIFF
--- a/custom/enikshay/case_utils.py
+++ b/custom/enikshay/case_utils.py
@@ -122,6 +122,24 @@ def get_open_occurrence_case_from_person(domain, person_case_id):
     return open_occurrence_cases[0]
 
 
+def get_associated_episode_case_for_test(test_case, occurrence_case_id):
+    """
+    get associated episode case set on the test case for new structure
+    fallback to finding the open episode case for old submissions
+    """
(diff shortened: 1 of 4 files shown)


custom/enikshay/case_utils.py
```diff
@@ -122,6 +122,24 @@ def get_open_occurrence_case_from_person(domain, person_case_id):
     return open_occurrence_cases[0]
 
 
+def get_associated_episode_case_for_test(test_case, occurrence_case_id):
+    """
+    get associated episode case set on the test case for new structure
+    fallback to finding the open episode case for old submissions
```
proteusvacuum: 👍

proteusvacuum (approved):
Thanks for the detailed commit messages!

PR #17937, opened 2017-09-23
Skip where owner is - but case is yet no closed

mkangia:

Possibly the change needed for that error.
v2 is not used anywhere now but looks like the failing indicators are just retrying and failing but still adding that change in v2 as well. 

Note: This would also need a rebuild. Added label for the same

cc: Buddy @esoergel

### DIFF
--- a/custom/enikshay/ucr/data_sources/test_tasklist_v2.json
+++ b/custom/enikshay/ucr/data_sources/test_tasklist_v2.json
@@ -66,7 +66,8 @@
             "comment": null,
             "property_value": [
               null,
-              ""
+              "",
+              "-"
             ]
           },
           "comment": null,
--- a/custom/enikshay/ucr/data_sources/test_tasklist_v3.json
+++ b/custom/enikshay/ucr/data_sources/test_tasklist_v3.json
@@ -61,7 +61,8 @@
                       "operator": "in",
                       "property_value": [
                           null,
-                          ""
+                          "",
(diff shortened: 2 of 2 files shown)


nickpell (approved):

mkangia:
Just to get some more people to confirm this looks okay to do.
@emord @proteusvacuum

proteusvacuum:
Won't this filter out ~almost all `test` cases? This is an extension case in both the public and private sector apps, so the default owner_id is `-`

mkangia:
@proteusvacuum but isn't that is what is causing the error? Guessing these test cases are not making it to the report anyway. I am not saying this is the best possible fix. Open to better options but would at least silent errors in sentry till we have a better solution.

proteusvacuum:
Maybe I'm missing context. Is the `test` tasklist only supposed to show owned tests? If so, then this is the right change, but if it is supposed to show anything else, then this change will filter out basically every test in the system.

mkangia:
Sorry @proteusvacuum 
There is not much to the context for this. I saw this error on sentry coming up quite a lot of times.. and raised this PR for the possible fix.

>  Is the test tasklist only supposed to show owned tests? If so, then this is the right change

Will try to confirm that.

mkangia:
> Won't this filter out ~almost all test cases? This is an extension case in both the public and private sector apps, so the default owner_id is -

is true.

PR #17940, opened 2017-09-23
DRTB 2 -> 2.1 Migration

mkangia:
https://dimagi.teamwork.com/#tasks/10794501

### DIFF
--- a/custom/enikshay/management/commands/drtb_2_to_2_dot_1_migration.py
+++ b/custom/enikshay/management/commands/drtb_2_to_2_dot_1_migration.py
@@ -106,22 +106,21 @@ class Command(BaseCommand):
                                 case_status = self.update_case(test_case.get_id, case_status)
                                 writer.writerow(case_status)
 
-    def update_case(self, case_id, update_props):
-        case_updates = update_props
+    def update_case(self, case_id, case_status):
+        """
+        :param case_id: id of case to be updated
+        :param case_status: case update status which has updates dict and other information about case update
+        :return: update case status
+        """
+        case_updates = case_status.get('updates')
         if case_updates:
             case_updates['datamigration_20_to_21'] = 'yes'
-            update_props['datamigration_20_to_21'] = 'yes'
             if not self.dry_run:
-                real_case_updates = case_updates.copy()
-                # remove any debug information stored for confirming changes and retain them for
-                # logging in the csv file
-                if 'debug' in real_case_updates:
-                    del real_case_updates['debug']
-                update_case(DOMAIN, case_id, real_case_updates)
-            update_props['updated'] = 'yes'
+                update_case(DOMAIN, case_id, case_updates)
+            case_status['updated'] = 'yes'
         else:
-            update_props['updated'] = 'no'
-        return update_props
+            case_status['updated'] = 'no'
+        return case_status
 
     def _get_headers(self, case_types):
         headers = ['case_type', 'case_id']
@@ -131,7 +130,7 @@ class Command(BaseCommand):
             headers += self.episode_case_relevant_props
         if 'test' in case_types:
             headers += self.test_case_relevant_props
-        headers += ['updates', 'updated', 'datamigration_20_to_21']
+        headers += ['updates', 'updated']
         return headers
 
     def get_case_status(self, case, episode_cases=None, recently_opened_episode_case=None, occurrence_case=None):
--- a/custom/enikshay/management/commands/drtb_2_to_2_dot_1_migration.py
+++ b/custom/enikshay/management/commands/drtb_2_to_2_dot_1_migration.py
@@ -164,7 +164,7 @@ class Command(BaseCommand):
         # if person.age is blank/null and person.age_entered != blank/null
         # then set person.age = person.age_entered.
         # else if person.age = '' and person.dob != '', person.age = int((today() - person.dob) / 365.25)
-        if not person_case_props.get('age') and not person_case_props.get('age_entered'):
+        if not person_case_props.get('age') and person_case_props.get('age_entered'):
             props_to_update['age'] = person_case_props.get('age_entered')
         elif not person_case_props.get('age') and person_case_props.get('dob'):
             props_to_update['age'] = int(
--- a/custom/enikshay/management/commands/drtb_2_to_2_dot_1_migration.py
+++ b/custom/enikshay/management/commands/drtb_2_to_2_dot_1_migration.py
@@ -103,7 +103,7 @@ class Command(BaseCommand):
     def _get_headers(self, case_types):
         headers = ['case_type', 'case_id']
         if 'person' in case_types:
-            headers += self.person_case_relevant_props.append('owner_id')
+            headers += (self.person_case_relevant_props + ['owner_id'])
         if 'episode' in case_types:
             headers += self.episode_case_relevant_props
         if 'test' in case_types:
--- a/custom/enikshay/management/commands/drtb_2_to_2_dot_1_migration.py
+++ b/custom/enikshay/management/commands/drtb_2_to_2_dot_1_migration.py
@@ -76,17 +76,20 @@ class Command(BaseCommand):
                         occurrence_cases = [case for case in accessor.get_reverse_indexed_cases([person_case_id])
                                             if case.type == CASE_TYPE_OCCURRENCE]
                         for occurrence_case in occurrence_cases:
-                            episode_cases = [case for case in accessor.get_reverse_indexed_cases([occurrence_case.get_id])
+                            episode_cases = [case for case in
+                                             accessor.get_reverse_indexed_cases([occurrence_case.get_id])
                                              if case.type == CASE_TYPE_EPISODE]
                             for episode_case in episode_cases:
                                 writer.writerow(self.get_case_status(episode_case, episode_cases=episode_cases))
 
                     if "test" in case_types:
                         if not occurrence_cases:
-                            occurrence_cases = [case for case in accessor.get_reverse_indexed_cases([person_case_id])
+                            occurrence_cases = [case for case in
+                                                accessor.get_reverse_indexed_cases([person_case_id])
                                                 if case.type == CASE_TYPE_OCCURRENCE]
                         for occurrence_case in occurrence_cases:
-                            test_cases = [case for case in accessor.get_reverse_indexed_cases([occurrence_case.get_id])
+                            test_cases = [case for case in
+                                          accessor.get_reverse_indexed_cases([occurrence_case.get_id])
                                           if case.type == CASE_TYPE_TEST]
                             recently_opened_episode_case = None
                             for test_case in test_cases:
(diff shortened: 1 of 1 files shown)


custom/enikshay/management/commands/drtb_2_to_2_dot_1_migration.py
```diff
@@ -0,0 +1,262 @@
...
+            'case_minor_version': '21',
+        }
+
+        person_case_props = person_case.dynamic_case_properties()
+
+        # If person.aadhaar_number != '', save person.aadhaar_number_obfuscated which is
+        # person.aadhaar number masked with *s except the last 4 digits
+        aadhar_num = person_case_props.get('aadhaar_number')
+        if aadhar_num:
+            props_to_update['aadhaar_number_obfuscated'] = '*' * 8 + aadhar_num[-4:]
```
sheelio: @mkangia 

I think the app allows aadhaar numbers that are both 12 and 13 characters.  Does this code work for both?

custom/enikshay/management/commands/drtb_2_to_2_dot_1_migration.py
```diff
@@ -0,0 +1,262 @@
...
+
+        # if person.age is blank/null and person.age_entered != blank/null
+        # then set person.age = person.age_entered.
+        # else if person.age = '' and person.dob != '', person.age = int((today() - person.dob) / 365.25)
+        if not person_case_props.get('age') and not person_case_props.get('age_entered'):
```
sheelio: @mkangia 

Should this logic be:
if not person_case_props.get('age') and ~~not~~ person_case_props.get('age_entered'):

custom/enikshay/management/commands/drtb_2_to_2_dot_1_migration.py
```diff
@@ -0,0 +1,262 @@
...
+                    if a_case.dynamic_case_properties().get('episode_type') == 'confirmed_drtb':
+                        return True
+
+        props_to_update = {}
+        episode_case_props = episode_case.dynamic_case_properties()
+        # if treatment_outcome = duplicate or treatment_outcome = invalid_registration we should blank the
+        # treatment_outcome and instead set episode.reason_for_closure = the value of treatment_outcome
+        # (Reverse the treatment_status / outcome change from Remove Person)
+        if episode_case_props.get('treatment_outcome') in ['duplicate', 'invalid_registration']:
+            props_to_update['reason_for_closure'] = episode_case_props.get('treatment_outcome')
```
sheelio: @mkangia 

The app team changed this property and now its called "close_reason" not "reason_for_closure"

custom/enikshay/management/commands/drtb_2_to_2_dot_1_migration.py
```diff
@@ -0,0 +1,262 @@
...
+            props_to_update['treatment_outcome'] = ""
+
+        # Set episode.case_definition = microbiological if episode_type = confirmed_drtb
+        if episode_case_props.get('episode_type') == "confirmed_drtb":
+            props_to_update['case_definition'] = 'microbiological'
+
+            # episode.dstb_to_drtb_transition_no_initiation = yes
+            # IF episode_type = confirmed_tb AND (treatment_initiated = 'no' or treatment_initiated = '')
+            # AND there exists another episode with episode_type=confirmed_drtb for the same occurrence
+            if (episode_case_props.get('treatment_initiated', "") in ["no", ""] and
```
sheelio: We should add a check on episode_type = confirmed_tb

custom/enikshay/management/commands/drtb_2_to_2_dot_1_migration.py
mkangia: No. This works for 12 digits because a valid aadhar number has 12 digits.

custom/enikshay/management/commands/drtb_2_to_2_dot_1_migration.py
mkangia: Right. Updating.

custom/enikshay/management/commands/drtb_2_to_2_dot_1_migration.py
mkangia: sorry. bad miss. I nested this under confirmed_drtb.

custom/enikshay/management/commands/drtb_2_to_2_dot_1_migration.py
sheelio: We should update this code to handle 12 or 13 digits (since the app allows you to enter 13 digit aadhaars).

mkangia: Thats weird.. a 13 digit aadhar number? thats not even valid.

Will update though

mkangia:
@sheelio updated. Let me know if i should get a dry run report again.

sheelio:
Yup, can you do another dry run.

sheelio:
@mkangia 
Can we mark each case that is being updated with a case property that identifies its migrated?  In other migrations we've used something like:
datamigration_20_to_21 = yes on each case that is being changed

mkangia:
@sheelio updated.

mkangia:
@sheelio this is for both public and private sector app?

sheelio:
This should only be for public sector patients (that said the case_version = 20 case property is only set on public person cases)

custom/enikshay/management/commands/drtb_2_to_2_dot_1_migration.py
```diff
@@ -0,0 +1,287 @@
...
+        return headers
+
+    def get_case_status(self, case, episode_cases=None, recently_opened_episode_case=None, occurrence_case=None):
+        if case.type == "person":
+            return self.get_person_case_status(case)
+        elif case.type == "episode":
+            return self.get_episode_case_status(case, episode_cases)
+        elif case.type == "test":
+            return recently_opened_episode_case, self.get_test_case_status(
+                case, recently_opened_episode_case, occurrence_case)
```
esoergel: You could also just call these functions directly in the `handle_method`, since you always know what type of case you're dealing with.

esoergel: Also, I swear I already reviewed this PR - was there another one recently that shared a lot of code, or did I just not submit the review?

custom/enikshay/management/commands/drtb_2_to_2_dot_1_migration.py
```diff
@@ -0,0 +1,287 @@
...
+                                             .get('treatment_initiation_date'))
+                props_to_update['debug']['treatment_initiation_date'] = treatment_initiation_date
+                date_tested = test_case_props.get('date_tested')
+                props_to_update['debug']['date_tested'] = date_tested
+                if treatment_initiation_date and date_tested:
+                    d1 = datetime.strptime(date_tested, "%Y-%m-%d")
+                    d2 = datetime.strptime(treatment_initiation_date, "%Y-%m-%d")
+                    props_to_update['rft_dstb_follow_up_treatment_month'] = (
+                        (d1.year - d2.year) * 12 + d1.month - d2.month)
+        return recently_opened_episode_case, props_to_update
```
esoergel: Looks like `recently_opened_episode_case` is potentially set in this method, which I assume is why it's being returned, but it's being thrown away (the update doesn't make it back to the `handle` method).

custom/enikshay/management/commands/drtb_2_to_2_dot_1_migration.py
```diff
@@ -0,0 +1,287 @@
...
+
+        # Set person.referred_outside_enikshay_by_id = person.referred_by_id if person.referred_by_id exists
+        if person_case_props.get('referred_by_id'):
+            props_to_update['referred_outside_enikshay_by_id'] = person_case_props.get('referred_by_id')
+
+        # if the owner id is not _archive_ set person.referred_outside_enikshay_date and
+        # person.referred_outside_enikshay_by_id to blank
+        if person_case.owner_id != ARCHIVED_CASE_OWNER_ID:
+            props_to_update['referred_outside_enikshay_date'] = ""
+            props_to_update['referred_outside_enikshay_by_id'] = ""
```
esoergel: @sheelio this seems kinda weird - `referred_outside_enikshay_by_id` is sometimes updated to reference another value, and also sometimes wiped?  Looks like it's implemented as per the instructions, just wondering if those instructions are correct.

custom/enikshay/management/commands/drtb_2_to_2_dot_1_migration.py
```diff
@@ -0,0 +1,287 @@
...
+                                          if case.type == CASE_TYPE_TEST]
+                            recently_opened_episode_case = None
+                            for test_case in test_cases:
+                                # For tests that are DSTB follow up tests, rft_general = follow_up_dstb
+                                # and result_recorded = yes
+                                recently_opened_episode_case, case_status = self.get_case_status(
+                                    test_case, recently_opened_episode_case=recently_opened_episode_case,
+                                    occurrence_case=occurrence_case)
+                                case_status = self.update_case(test_case.get_id, case_status)
+                                writer.writerow(case_status)
```
esoergel: How does `writer.writerow` handle nested dicts?  Does the `debug` information just show up as a json blob in a cell somewhere?

esoergel (approved):

custom/enikshay/management/commands/drtb_2_to_2_dot_1_migration.py
sheelio: Yea, its a bit weird but this is correct. 

The first situation is just moving it from an old case property to a new case property (which happens for almost all referred outside of enikshay cases)

The second situation handles making sure its cleared / empty for the situation where you referred outside of enikshay and then it was brought back (by referring it back into enikshay).

custom/enikshay/management/commands/drtb_2_to_2_dot_1_migration.py
mkangia: Ya. It just shows as a dict. For ex:
```
{'debug':   {'treatment_initiation_date': u'2016-12-19', 'date_tested': u'2017-02-07',   'episode_case_id': u'82ed063b618b4f65bb193852f2caff63'},   'rft_dstb_follow_up_treatment_month': 2, 'datamigration_20_to_21': 'yes'}
--
```
Is a cell.

custom/enikshay/management/commands/drtb_2_to_2_dot_1_migration.py
mkangia: The other PR is https://github.com/dimagi/commcare-hq/pull/18027.
This is but one time and the other one would eventually become a reconciliation task that runs more regularly.

mkangia: > You could also just call these functions directly in the handle_method, since you always know what type of case you're dealing with.

Just wanted to extract this logic out so kept it in a method

custom/enikshay/management/commands/drtb_2_to_2_dot_1_migration.py
mkangia: oh i see.. Ya need to pass that back up. Updating.

mkangia: Updated [here](https://github.com/dimagi/commcare-hq/pull/17940/commits/a6900c84f57cb32c96eeee2d8816708a086cadce)

esoergel (approved):

mkangia:
Ran on server.

PR #17942, opened 2017-09-25
Changes in Enikshay private reports

kkrampa:


kkrampa:
@calellowitz

custom/enikshay/ucr/data_sources/voucher_v2.json
```diff
@@ -0,0 +1,1184 @@
+{
+    "domains": [
+        "enikshay",
+        "sheel-enikshay",
+        "enikshay-reports-qa",
+        "enikshay-performance-test"
+    ],
+    "server_environment": [
+        "enikshay",
+        "softlayer"
```
proteusvacuum: we don't need to add new datasources to softlayer any more

proteusvacuum (approved):
@kkrampa when adding new datasources would you be able to copy the original file in one commit, then make the changes in a second commit? This will make it easier to review the changes in github next time.

custom/enikshay/ucr/data_sources/voucher_v2.json
kkrampa: @proteusvacuum sure will do this in two commits next time.

PR #17943, opened 2017-09-25
Update patient type mapping according to new TOG

mkangia:
cc: buddy @esoergel


esoergel (approved):

proteusvacuum (approved):
Do we have to retroactively do anything for cases that were already submitted with the wrong mappings?

mkangia:
@proteusvacuum 

> Do we have to retroactively do anything for cases that were already submitted with the wrong mappings?

i don't think we can't do anything. Will check with EY.

PR #17955, opened 2017-09-25
move SessionDetailsView out of hqadmin

dannyroberts:
All views inside hqadmin are audited by auditcare, so every auth request
from formplayer makes 2 couchdb saves. This change should take 80-200ms off of
every request to formplayer

### DIFF
--- a/corehq/apps/hqadmin/tests/test_views.py
+++ b/corehq/apps/hqadmin/tests/test_views.py
@@ -1,20 +1,8 @@
-import base64
-import hashlib
-import hmac
-import json
-
-from django.conf import settings
 from django.http import HttpResponse
 from django.test import SimpleTestCase
-from django.test import TestCase
-from django.test.client import Client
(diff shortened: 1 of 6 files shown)


dannyroberts:
@ctsims / @wspride FYI

snopoke (approved):
Do we want to keep it at `hq/admin/session_details`?

dannyroberts:
Fair question; since formplayer depends on it, I thought it'd be simplest to keep it the same. An alternative that has the benefit of not being mutually-dependent on the formplayer change (making rollback painful/error-prone) would be to just support both urls for a period of time. I don't feel that strongly about keeping it or changing it, open to suggestions.

kaapstorm (approved):

snopoke:
:boot: :construction_worker_man:

dannyroberts:
Thanks for kicking the build!

PR #17956, opened 2017-09-25
check test's person is public

nickpell:
@sheelio


sheelio:
@nickpell 
Not blocking, but I think the is_person_public function may not be robust to different case types.  I think it will work in this situation since its a "test" case but not sure it will work in others. 

You could update the is_person_public to use this function:
https://github.com/dimagi/commcare-hq/blob/master/custom/enikshay/case_utils.py#L433

nickpell:
Thanks for flagging.  If I end up writing a migration for another case type I'll edit ```is_person_public```

PR #17958, opened 2017-09-25
Let a user search for exact matches by name and locations under specific location

mkangia:
Picking this up from conversation on [slack](https://dimagi.slack.com/archives/C0K5U3V2B/p1505979321000263) as a possible workaround for better location searching

In a tree of locations where it can get annoying searching for a location with a specific name,
This lets a user
1. Search for the location with exact match to the name and then also shows all locations under it. This can be done by using "/Massachusetts" or "/Massachusetts/"
2. Search for locations under a specific location with a fuzzy match. This can be useful when there is a long list of child locations "/Massachusetts/Bos" to look for Boston under Massachusetts

[Spec](https://docs.google.com/document/d/1Jh0wPCNk0N0noqrH5c_ZDDDEmmRcwIGNkKzVv0g4ZF4/edit#)

### DIFF
--- a/corehq/apps/locations/models.py
+++ b/corehq/apps/locations/models.py
@@ -344,10 +344,7 @@ class LocationManager(LocationQueriesMixin, TreeManager):
         """
         Takes a set of location ids and returns a django queryset of their children.
         """
-        return self.get_queryset_descendants(
-            self.filter(location_id__in=location_ids),
-            include_self=False
-        ).location_ids()
+        return self.get_children(location_ids).location_ids()
 
     def get_locations_and_children_ids(self, location_ids):
         return list(self.get_locations_and_children(location_ids).location_ids())
--- a/corehq/apps/locations/models.py
+++ b/corehq/apps/locations/models.py
@@ -303,7 +303,7 @@ class LocationManager(LocationQueriesMixin, TreeManager):
         except self.model.DoesNotExist:
             return self.get(domain=domain, name__iexact=user_input)
 
-    def filter_path_by_user_input(self, domain, user_input, from_location_ids=None):
+    def filter_path_by_user_input(self, domain, user_input):
         """
         Returns a queryset including all locations matching the user input
         and their children. This means "Middlesex" will match:
@@ -313,8 +313,6 @@ class LocationManager(LocationQueriesMixin, TreeManager):
         It matches by name or site-code
         """
         direct_matches = self.filter_by_user_input(domain, user_input)
-        if from_location_ids is not None:
-            direct_matches = direct_matches.filter(location_id__in=from_location_ids)
         return self.get_queryset_descendants(direct_matches, include_self=True)
 
     def get_locations(self, location_ids):
@@ -325,20 +323,10 @@ class LocationManager(LocationQueriesMixin, TreeManager):
         Takes a set of location ids and returns a django queryset of those
         locations and their children.
         """
-        return self.get_descendants(location_ids, include_self=True)
-
-    def get_descendants(self, location_ids, include_self=False):
-        """
-        Takes a set of location ids and returns a django queryset of their descendants.
-        :param include_self: include the locations with location_ids in the final queryset as well
-        """
         return self.get_queryset_descendants(
             self.filter(location_id__in=location_ids),
-            include_self=include_self
+            include_self=True
         )
-
-    def get_descendants_ids(self, location_ids):
-        return list(self.get_descendants(location_ids).location_ids())
 
     def get_locations_and_children_ids(self, location_ids):
         return list(self.get_locations_and_children(location_ids).location_ids())
--- a/corehq/apps/locations/models.py
+++ b/corehq/apps/locations/models.py
@@ -330,7 +330,7 @@ class LocationManager(LocationQueriesMixin, TreeManager):
             include_self=True
         )
 
-    def get_children(self, location_ids):
+    def get_descendants(self, location_ids):
         """
         Takes a set of location ids and returns a django queryset of their children.
         """
@@ -339,11 +339,11 @@ class LocationManager(LocationQueriesMixin, TreeManager):
             include_self=False
         )
 
-    def get_children_ids(self, location_ids):
+    def get_descendants_ids(self, location_ids):
         """
         Takes a set of location ids and returns a django queryset of their children.
         """
-        return self.get_children(location_ids).location_ids()
+        return self.get_descendants(location_ids).location_ids()
 
     def get_locations_and_children_ids(self, location_ids):
         return list(self.get_locations_and_children(location_ids).location_ids())
--- a/corehq/apps/locations/models.py
+++ b/corehq/apps/locations/models.py
@@ -295,16 +295,13 @@ class LocationManager(LocationQueriesMixin, TreeManager):
         except self.model.DoesNotExist:
             return self.get(domain=domain, name__iexact=user_input)
 
-    def filter_by_user_input(self, domain, user_input, from_location_ids=None):
+    def filter_by_user_input(self, domain, user_input):
         """
         Accepts partial matches, matches against name and site_code.
         """
-        locations = (self.filter(domain=domain)
-                     .filter(models.Q(name__icontains=user_input) |
-                             models.Q(site_code__icontains=user_input)))
-        if from_location_ids is not None:
-            locations = locations.filter(location_id__in=from_location_ids)
-        return locations
+        return (self.filter(domain=domain)
+                    .filter(models.Q(name__icontains=user_input) |
+                            models.Q(site_code__icontains=user_input)))
 
     def filter_path_by_user_input(self, domain, user_input, from_location_ids=None):
         """
@@ -315,7 +312,9 @@ class LocationManager(LocationQueriesMixin, TreeManager):
             Massachusetts/Middlesex/Cambridge
         It matches by name or site-code
         """
-        direct_matches = self.filter_by_user_input(domain, user_input, from_location_ids)
+        direct_matches = self.filter_by_user_input(domain, user_input)
+        if from_location_ids is not None:
+            direct_matches = direct_matches.filter(location_id__in=from_location_ids)
         return self.get_queryset_descendants(direct_matches, include_self=True)
 
     def get_locations(self, location_ids):
--- a/corehq/apps/locations/models.py
+++ b/corehq/apps/locations/models.py
@@ -295,15 +295,18 @@ class LocationManager(LocationQueriesMixin, TreeManager):
         except self.model.DoesNotExist:
             return self.get(domain=domain, name__iexact=user_input)
 
-    def filter_by_user_input(self, domain, user_input):
+    def filter_by_user_input(self, domain, user_input, from_location_ids=None):
         """
         Accepts partial matches, matches against name and site_code.
         """
-        return (self.filter(domain=domain)
-                    .filter(models.Q(name__icontains=user_input) |
-                            models.Q(site_code__icontains=user_input)))
+        locations = (self.filter(domain=domain)
+                     .filter(models.Q(name__icontains=user_input) |
+                             models.Q(site_code__icontains=user_input)))
+        if from_location_ids is not None:
+            locations = locations.filter(location_id__in=from_location_ids)
+        return locations
 
-    def filter_path_by_user_input(self, domain, user_input):
+    def filter_path_by_user_input(self, domain, user_input, from_location_ids=None):
         """
         Returns a queryset including all locations matching the user input
         and their children. This means "Middlesex" will match:
@@ -312,7 +315,7 @@ class LocationManager(LocationQueriesMixin, TreeManager):
             Massachusetts/Middlesex/Cambridge
         It matches by name or site-code
         """
-        direct_matches = self.filter_by_user_input(domain, user_input)
+        direct_matches = self.filter_by_user_input(domain, user_input, from_location_ids)
         return self.get_queryset_descendants(direct_matches, include_self=True)
 
     def get_locations(self, location_ids):
@@ -327,6 +330,24 @@ class LocationManager(LocationQueriesMixin, TreeManager):
             self.filter(location_id__in=location_ids),
             include_self=True
         )
+
+    def get_children(self, location_ids):
+        """
+        Takes a set of location ids and returns a django queryset of their children.
+        """
+        return self.get_queryset_descendants(
+            self.filter(location_id__in=location_ids),
+            include_self=False
+        )
+
+    def get_children_ids(self, location_ids):
+        """
+        Takes a set of location ids and returns a django queryset of their children.
+        """
+        return self.get_queryset_descendants(
+            self.filter(location_id__in=location_ids),
+            include_self=False
+        ).location_ids()
 
     def get_locations_and_children_ids(self, location_ids):
         return list(self.get_locations_and_children(location_ids).location_ids())
--- a/corehq/apps/locations/models.py
+++ b/corehq/apps/locations/models.py
@@ -325,18 +325,16 @@ class LocationManager(LocationQueriesMixin, TreeManager):
         Takes a set of location ids and returns a django queryset of those
         locations and their children.
         """
-        return self.get_queryset_descendants(
-            self.filter(location_id__in=location_ids),
-            include_self=True
-        )
+        return self.get_descendants(location_ids, include_self=True)
 
-    def get_descendants(self, location_ids):
+    def get_descendants(self, location_ids, include_self=False):
         """
-        Takes a set of location ids and returns a django queryset of their children.
+        Takes a set of location ids and returns a django queryset of their descendants.
+        :param include_self: include the locations with location_ids in the final queryset as well
         """
         return self.get_queryset_descendants(
             self.filter(location_id__in=location_ids),
-            include_self=False
+            include_self=include_self
         )
 
     def get_descendants_ids(self, location_ids):
--- a/corehq/apps/locations/models.py
+++ b/corehq/apps/locations/models.py
@@ -265,6 +265,14 @@ class LocationQueriesMixin(object):
             publish_location_saved(domain, location_id, is_deletion=True)
         return super(LocationQueriesMixin, self).delete(*args, **kwargs)
 
+    def filter_by_user_input(self, domain, user_input):
+        """
+        Accepts partial matches, matches against name and site_code.
+        """
+        return (self.filter(domain=domain)
+                    .filter(models.Q(name__icontains=user_input) |
+                            models.Q(site_code__icontains=user_input)))
+
 
 class LocationQuerySet(LocationQueriesMixin, models.query.QuerySet):
     pass
@@ -294,14 +302,6 @@ class LocationManager(LocationQueriesMixin, TreeManager):
             return self.get(domain=domain, site_code=user_input)
         except self.model.DoesNotExist:
             return self.get(domain=domain, name__iexact=user_input)
-
-    def filter_by_user_input(self, domain, user_input):
-        """
-        Accepts partial matches, matches against name and site_code.
-        """
-        return (self.filter(domain=domain)
-                    .filter(models.Q(name__icontains=user_input) |
-                            models.Q(site_code__icontains=user_input)))
 
     def filter_path_by_user_input(self, domain, user_input, from_location_ids=None):
         """
--- a/corehq/apps/locations/models.py
+++ b/corehq/apps/locations/models.py
@@ -338,10 +338,7 @@ class LocationManager(LocationQueriesMixin, TreeManager):
         )
 
     def get_descendants_ids(self, location_ids):
-        """
-        Takes a set of location ids and returns a django queryset of their children.
-        """
-        return self.get_descendants(location_ids).location_ids()
+        return list(self.get_descendants(location_ids).location_ids())
 
     def get_locations_and_children_ids(self, location_ids):
         return list(self.get_locations_and_children(location_ids).location_ids())
--- a/corehq/apps/reports/filters/api.py
+++ b/corehq/apps/reports/filters/api.py
@@ -157,17 +157,17 @@ class LocationRestrictedEmwfOptionsMixin(object):
     def custom_locations_search(self):
         """
         When the query is specifically searching for just locations and not any other entity like user, group.
-        For ex: enter "/Bihar/patn" would match child locations under locations named Bihar, having name like
+        For ex: enter "Bihar/patn" would match child locations under locations named Bihar, having name like
         patn.
         """
-        return self.q.startswith('/')
+        return '/' in self.q
 
     @staticmethod
     def _get_location_specific_custom_filters(query):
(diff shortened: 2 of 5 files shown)


corehq/apps/locations/models.py
```diff
@@ -295,15 +295,18 @@ def get_from_user_input(self, domain, user_input):
...
-        return (self.filter(domain=domain)
-                    .filter(models.Q(name__icontains=user_input) |
-                            models.Q(site_code__icontains=user_input)))
-
-    def filter_path_by_user_input(self, domain, user_input):
+        locations = (self.filter(domain=domain)
+                     .filter(models.Q(name__icontains=user_input) |
+                             models.Q(site_code__icontains=user_input)))
+        if from_location_ids is not None:
+            locations = locations.filter(location_id__in=from_location_ids)
```
esoergel: I'd prefer just doing this as a separate filter - it feels like a second thing.  Instead of:
```python
SQLLocation.active_objects.filter_by_user_input(domain, user_input, from_location_ids)
```
You could just do:
```python
(SQLLocation.active_objects.filter_by_user_input(domain, user_input)
                           .filter(location_id__in=from_location_ids))
```

corehq/apps/locations/models.py
```diff
@@ -328,6 +331,21 @@ def get_locations_and_children(self, location_ids):
...
+            self.filter(location_id__in=location_ids),
+            include_self=False
+        )
+
+    def get_children_ids(self, location_ids):
+        """
+        Takes a set of location ids and returns a django queryset of their children.
+        """
+        return self.get_children(location_ids).location_ids()
+
```
esoergel: These two also feel very specific - I'm wary of making these models too large with things that will only be used in one place.  Also, "children" generally refers to locations whose direct parent is the location in question, whereas these two methods return all descendants.

corehq/apps/reports/filters/api.py
```diff
@@ -154,9 +154,40 @@ def extra_data_sources(self):
...
+        """
+        When the query is specifically searching for just locations and not any other entity like user, group.
+        For ex: enter "/Bihar/patn" would match child locations under locations named Bihar, having name like
+        patn.
+        """
+        return self.q.startswith('/')
+
+    @staticmethod
+    def _get_location_specific_custom_filters(query):
+        query_sections = query.split("/")
```
esoergel: I like this paradigm a lot - we usually display locations as "parent/child", so I'd be in favor of making this the default behavior of `filter_path_by_user_input`, so typing "mass/bos" would match "USA/Massachusetts/Suffolk/Boston"

esoergel (commented):
I like the general idea very much.  I think the primary issue with the current implementation is that you can only match against one node at a time.  We might want to talk to product about it, though - are users expected to know to prefix their queries with a "/"?  What if instead, we just always did this sort of lookup when a slash is present, so users can just type what they see (the path display in the format of "Massachusetts/Suffolk/Boston").  Here's one approach which should work:
```python
query = SQLLocation.active_objects
for part in user_input.split('/'):
    query = (SQLLocation.active_objects
             .get_queryset_descendants(query, include_self=False)
             .filter_by_user_input(domain, part))
```
This will have the same behavior as now when there is no slash, and it'll support an arbitrary number of components with partial matches.  Also, by passing around unevaluated querysets rather than location ids, it will only do one db hit.

corehq/apps/locations/models.py
mkangia: Making that "descendants" sounds right. I was kind of surprised that there was not a method like this already available. Find descendants of a location and not including itself would be a common use case for having such a hierarchical structure.

Can move that if that looks like bloating it, but it is bound to this class so there won't be any better place to put it and would end up just having that function definition everywhere its needed.

mkangia:
Thanks for the review @esoergel. The problem i was trying to solve was different.

> I think the primary issue with the current implementation is that you can only match against one node at a time

Thats actually the positive of this. One thing about fuzzy search is that its fuzzy. The issue i am trying to address here is when there are a lot of matches for a certain string and one gets too many results. For ex: if you go [here](https://enikshay.in/a/enikshay/reports/submit_history/) and search for "patna" you would see that you get everything except the location patna itself. So thats addressed by the aim of this PR.

> What if instead, we just always did this sort of lookup when a slash is present, so users can just type what they see (the path display in the format of "Massachusetts/Suffolk/Boston")
I don't think someone would want to type that long a string and would just type Ma/Su/Bo and it would find it. I like that but it has the same issue i highlighted above. It would match lot of locations with "Ma" and then would find "Su" and "Bo" under all possible descendants. 
So, If i need Boston, i can just type /boston/ according to the changes i made and it would give me straight away boston. Or i can type /Suffolk/ and get a list of locations under Suffolk to find boston in it.

mkangia:
> We might want to talk to product about it, though - are users expected to know to prefix their queries with a "/"?

Absolutely not. I was planning to add a help not once i get a positive response from review and it looks like something that is worth adding to the search.

corehq/apps/locations/models.py
mkangia: Yup! that looks better.

mkangia: checked this, if from_locations_ids is None/[] in case we don't want to pick from specific set of locations, this logic would return no locations. So looks like the check for `from_location_ids` is still needed.

corehq/apps/locations/models.py
mkangia: Tried to trim that down a bit with some refactor.

corehq/apps/reports/filters/api.py
mkangia: Explained in PR why i chose an exact name search instead of fuzzy search so this change would not help address the issue we are trying to solve here. Though can make that change but it won't be used anywhere as of now.

esoergel:
That's a fair point - I see what you're getting at.  The `/` thing sounds fine to me then.  Do you know of any other places in HQ where we specify exact matches through some syntax?

Regarding the queryset methods, I think my quibble with them is chiefly that they encourage passing around lists of `location_ids` a lot, when the power of Django querysets comes from the ability to chain and nest queries.  We need to use a list of location_ids when interacting with other data stores like ES, or to send to the client, but it shouldn't be necessary to assemble complex queries.

```python
# Here's an approximation of one of the queries used here:
parent_ids = SQLLocation.active_objects.filter(name=parent_name).location_ids()
from_location_ids = SQLLocation.active_objects.get_descendants_ids(parent_ids)
return (SQLLocation.active_objects
        .filter_path_by_user_input(self.domain, query, from_location_ids)
        .accessible_to_user(self.request.domain, self.request.couch_user))


# But if you expand it out, this is what's happening behind the scenes:
parent_ids = SQLLocation.active_objects.filter(name=parent_name).location_ids()
from_location_ids = list(
    SQLLocation.active_objects
    .get_queryset_descendants(
        # this is essentially the same queryset used just above:
        SQLLocation.active_objects.filter(location_id__in=parent_ids)
    )
    .location_ids()
)
direct_matches = (
    SQLLocation.active_objects
    .filter_by_user_input(domain, user_input)
    # here we're reassembling the queryset in `from_location_ids`
    .filter(location_id__in=from_location_ids)
)
return (SQLLocation.active_objects
        .get_queryset_descendants(direct_matches)
        .accessible_to_user(self.request.domain, self.request.couch_user))


# When instead, if you use MPTT builtins and pass around querysets, that could be
# simplified to this:
parents = SQLLocation.active_objects.filter(name=parent_name)
descendants = SQLLocation.active_objects.get_queryset_descendants(parents)
direct_matches = descendants.filter_by_user_input(domain, user_input)
return (SQLLocation.active_objects
        .get_queryset_descendants(direct_matches)
        .accessible_to_user(self.request.domain, self.request.couch_user))

# (although note that `filter_by_user_input would need to be moved to `LocationQueriesMixin`)
```
From an API standpoint, I'd also prefer to chain two methods rather than have separate methods for each permutation.  With couch dbaccessors we have to have separate functions for `get_released_apps`, `get_released_app_docs`, and `get_released_app_ids`, and similarly for app builds.  With queryset chaining, we can instead do the equivalent of `get_apps().released().ids()`, where we build up the logic we want without needing a method for each combination of parameters (The exception being when that compound method encapsulates an important concept).

mkangia:
@esoergel 

> That's a fair point - I see what you're getting at. The / thing sounds fine to me then. Do you know of any other places in HQ where we specify exact matches through some syntax?

Dont know about any such implementations anywhere.

Agree about the queryset. Let me work on this further. Thank you for reminding me the importance of doing this.

mkangia:
@dimagi/product 
Trying to just give some power to the user when searching for locations in exports and reports.
This idea came to me when Dev Patel mentioned how painful it was to search for a location(patna) in enikshay since it matched many locations that had "patna" in their name and eventually he got everything but not the location he was looking for.

This change gives user the ability to search a location(s) exactly by a name and also fuzzy search for locations under this particular set of location(s). The use case is when user knows already what to search and can reach to it straight away.

So in the example above one can type
"/patna" or "/patna/" to get locations with name patna and list of locations under these locations. Ideally there should be only one location with a name so this would be just location patna and locations under it. 
"/patna/ab" to search for locations under patna with name like ab.

I plan to add a help note for the user to know that "/" can be used if this change looks worth it. Open to ideas on how to add that and content that would make this easier to understand.

djmore9:
@orangejenny  Can the problem Manish is addressing with this request be solved with a UX paradigm  familiar to our SaaS users?

sheelio:
@djmore9 
The location list already displays results in the format Mass/Middlesex/Somerville if I type Somerville or if I type Middle.  I think users would expect to type the slashes to search (since that is what is displayed).  

@mkangia: If I type something like Mass/Somerv will that return a result (in my above example?)

@esoergel @mkangia 
I think the last time we considered this, we were concerned about performance of searching the full location tree in that way.  Is that addressed?

orangejenny:
@djmore9 This general problem also comes up when you're searching for a question; so, in app builder (the case management dropdowns) and form builder (bubbles). We haven't tried to address it in this fashion - if we do go ahead with this, it'd be interesting to do something similar in app/form builder - I think we have just tweaked fuzzy search params.

As far as special syntax for exact matching, the only thing that comes to mind that people would be familiar with is quote marks. Thoughts on having users type `"patna"` instead of `/patna`?

mkangia:
@sheelio 

> @mkangia: If I type something like Mass/Somerv will that return a result (in my above example?)

If the location is exactly `Mass/Middlesex/Somerville` in terms of the name, `Mass/Somerv` will return that location as well. but if Mass is actually `Massachusetts` then it won't.
So reiterating what it does is
searches for location with name exactly as the one after the first "/" and fuzzy search of whats after the second "/"
so if i type `Massachusetts/ome`, i am looking for locations like name `ome` in descendants of `Massachusetts`. So in this would match something like `Mass/Middlesex/Somerville`.

@orangejenny 
> As far as special syntax for exact matching, the only thing that comes to mind that people would be familiar with is quote marks. Thoughts on having users type "patna" instead of /patna?

That sounds fine for a single word search like probably a case property but since this is hierarchy structure i think a "/" sounds good. Its already used in locations in the way they are represented so should not be difficult to remember.

mkangia:
> @esoergel @mkangia
I think the last time we considered this, we were concerned about performance of searching the full location tree in that way. Is that addressed?

@sheelio this is not searching the full location tree that but just limits to locations under a certain location. The approach that was discussed earlier was i think the one @esoergel  mentioned [above](https://github.com/dimagi/commcare-hq/pull/17958#discussion_r141126566).

sheelio:
@esoergel  I'm ok with this pattern / syntax as long as it doesn't lead to unintended search results.

djmore9:
@mkangia Let's get this tested by QA before moving ahead.

orangejenny:
> Let's get this tested by QA before moving ahead.

This would also be a good UAT candidate. @nsclark

millerdev:
Weighing in here because as I read the discussion I also thought, like @orangejenny, that quotes would make more sense than slash (/) as an exact match delimiter. Quotes have a precedent (Google search works that way), while slashes do not (at least not as search term delimiters) that I know of.

@esoergel wrote:
> we could do both Manish's proposal (a special syntax for exact matching) and mine (fuzzy matching across multiple nodes).

It's not clear from how you said that, but were you thinking there would be a different delimiter for exact match than for multi-part fuzzy matching?

mkangia:
> Weighing in here because as I read the discussion I also thought, like @orangejenny, that quotes would make more sense than slash (/) as an exact match delimiter. Quotes have a precedent (Google search works that way), while slashes do not (at least not as search term delimiters) that I know of.

@millerdev 
using "/" because of the way locations are displayed.
So if the locations are displayed as
"/ABC/DEF/GHI", if the user wants to search within locations DEF and below, they type
"/DEF" for showing all locations under DEF
and 
"/DEF/H" for filtering locations that have name like H under DEF.

esoergel:
regarding
> we could do both Manish's proposal (a special syntax for exact matching) and mine (fuzzy matching across multiple nodes).

@millerdev wrote:
> It's not clear from how you said that, but were you thinking there would be a different delimiter for exact match than for multi-part fuzzy matching?

Same delimiter.  We've wanted for a while the ability to match against multiple levels in the tree, but have punted due to technical limitations.  Manish's idea to use slashes as semantic delimiters gave me an idea for how to get around that.  I'm proposing changing the default `filter_path_from_user_input` method to something which will recognize a slash as a delimiter (currently it tries to string match it, which fails).  Manish's idea is to use the presence of a _leading_ slash (or other such syntax) to do a special type of exact query.

Right now, `middle` matches:
`Massachusetts/Middlesex`
`Massachusetts/Middlesex/Boston`
`Massachusetts/Middlesex/Cambridge`
and `cambr` matches
`United Kingdom/England/Cambridge`
`Massachusetts/Middlesex/Cambridge`

My proposal is for you to be able to type `mass/cambr` to get:
`Massachusetts/Middlesex/Cambridge`
Which feels like a natural extension of the current behavior to me.  Ideally, `masscambr` would work too, but that's much harder to query, and we'd ideally have these full strings to query against.

Manish's proposal is for you to be able to type `/cambridge` (exact matches only) to get
`United Kingdom/England/Cambridge`
`Massachusetts/Middlesex/Cambridge`
or  `/massachusetts/cambr` (exact match on the first, fuzzy match on the second) to get:
`Massachusetts/Middlesex/Cambridge`
This is useful in the example of `patna`, where there are many locations with `patna` in the name, so it's a way to say _precisely_ "patna" - no more, no less.

@mkangia would you say that's a correct characterization?

corehq/apps/locations/models.py
```diff
@@ -265,6 +265,14 @@ def delete(self, *args, **kwargs):
             publish_location_saved(domain, location_id, is_deletion=True)
         return super(LocationQueriesMixin, self).delete(*args, **kwargs)
 
+    def filter_by_user_input(self, domain, user_input):
+        """
+        Accepts partial matches, matches against name and site_code.
+        """
+        return (self.filter(domain=domain)
+                    .filter(models.Q(name__icontains=user_input) |
+                            models.Q(site_code__icontains=user_input)))
```
esoergel: :+1:

corehq/apps/reports/filters/api.py
```diff
@@ -154,10 +154,36 @@ def extra_data_sources(self):
...
+            search_query = ""
+        return parent_name, search_query
+
     def get_locations_query(self, query):
-        return (SQLLocation.active_objects
-                .filter_path_by_user_input(self.domain, query)
-                .accessible_to_user(self.request.domain, self.request.couch_user))
+        if self.custom_locations_search():
+            parent_name, search_query = self._get_location_specific_custom_filters(query)
+            parents = SQLLocation.active_objects.filter(name=parent_name, domain=self.domain)
```
esoergel: one last thought - you might want this to use `name__iexact` so you don't match case.

esoergel (approved):

corehq/apps/reports/filters/api.py
mkangia: Do you mean it should match the exact case? Its doing that already, its case sensitive.

mkangia:
re: https://github.com/dimagi/commcare-hq/pull/17958#issuecomment-334236046
@esoergel Yup that sounds good. The only thing is when you mention 
> Which feels like a natural extension of the current behavior to me.

I am not sure how you plan to extend this by keeping this exact search functionality intact by using the same delimiter but just hope that you have considered keeping both
1. exact match when needed(this PR)
2. fuzzy match(in future PR)

corehq/apps/reports/filters/api.py
esoergel: I was thinking a case insensitive match would be more user friendly, but don't feel strongly about it.

mkangia:
@millerdev 
adding spec doc [here](https://docs.google.com/document/d/1Jh0wPCNk0N0noqrH5c_ZDDDEmmRcwIGNkKzVv0g4ZF4/edit#)

> If leading slash implies exact (case insensitive) match: /cambridge matches Cambridge but not Cambridgeport

According to changes in PR this would be a case sensitive search so it would match just cambridge. Also if make it match `Cambridgeport` or `cambridgeport` it would be same as the current implementation which is a fuzzy search.

corehq/apps/reports/filters/api.py
orangejenny: I agree with case insensitive search and feel more strongly than Ethan does, both to be more user friendly and to be more consistent with the rest of HQ.

mkangia: Updated [here](https://github.com/dimagi/commcare-hq/pull/17958/commits/13526f9bdc2d646428d700c1d26cc4a1022f8c4a)

orangejenny: thanks!

corehq/apps/reports/filters/api.py
```diff
@@ -154,10 +154,36 @@ def extra_data_sources(self):
...
+        """
+        When the query is specifically searching for just locations and not any other entity like user, group.
+        For ex: enter "Bihar/patn" would match child locations under locations named Bihar, having name like
+        patn.
+        """
+        return '/' in self.q
+
+    @staticmethod
+    def _get_location_specific_custom_filters(query):
+        query_sections = query.split("/")
```
millerdev: Looks like text after the _second_ slash will be discarded. Not sure if you want to support arbitrary parent/child/grandchild searches, or if you just want to handle the first slash? If the latter, consider changing this to `query.split("/", 1)`, which will only split on the first `/`.

mkangia: Sorry not sure i understand. Sharing what i intended.
Just expecting one "/" in the query. So "bihar/patna" would give parent_name as bihar and patna as the search query for the descendants.

millerdev: but what happens if someone types "bihar/patna/danapur"? I believe currently "danapur" would be discarded/ignored.

mkangia: ya thats fine.
There is no reason for someone to search like this.
to search `danapur` one can just type `bihar/danapur`

millerdev: some people may not realize that, and I don't think it's a good idea to be silently ignoring part of what they typed

esoergel: What about something like this for `filter_path_by_user_input`:
```python
def filter_path_by_user_input(self, domain, user_input):
    query = self
    for part in filter(None, user_input.split('/')):
        if part.startswith('"') and part.endswith('"'):
            direct_matches = query.filter(name__iexact=part[1:-1])
        else:
            direct_matches = query.filter_by_user_input(domain, part)
        query = self.get_queryset_descendants(direct_matches, include_self=True)
    return query
```
This is an expansion of my earlier suggestion - it does partial matches on as many nodes as are entered, and if you wrap any node in quotes, it'll do an exact match on that node.  That will allow you to do exact matches also on the terminal node.  It sounds like we still all disagree on how to specify exact matches, so I'm just throwing my hat into the ring.

mkangia: Thank you @esoergel. That worked well in terms of search.

For some edge case it *might* show duplicates. I am not sure but saw some duplicates when searching on local system. Could not replicate.
But,
At the back ES Search complaints about `"` and '\' in search so would lead to errors whenever the search is done for input while the user is typing the complete string.

millerdev: > At the back ES Search complaints about `"` and `'` in search

Are they being escaped properly? If yes, and those characters are truly not allowed in ES search strings (that would be surprising to me) we could strip unbalanced quotes.

millerdev: Actually, on second though, I think the way Google handles it is if there is an unbalanced quote character then everything from that character to the end of the string is matched exactly. In other words, an implicit balancing quote is added at the end of the string. We should probably do that.

esoergel: I like the idea of an implicit balancing quote - that sounds like a good idea.  Manish is describing ES errors, which are probably occurring in the other data sources that are searched in this filter.  If there's a slash, it's clear that the user is searching for a location, so otherwise we can't tell.  On the ES-backed data sources, quotes get interpreted as "advanced" search queries [here](https://github.com/dimagi/commcare-hq/blob/master/corehq/apps/es/queries.py#L56-L85).

I just took a look at the docs for [query string queries](https://www.elastic.co/guide/en/elasticsearch/reference/2.4/query-dsl-query-string-query.html#query-dsl-query-string-query) (which we use to support things like quotes and wildcards), and I noticed an alternative, the [simple query string query](https://www.elastic.co/guide/en/elasticsearch/reference/2.4/query-dsl-simple-query-string-query.html#query-dsl-simple-query-string-query):
> A query that uses the SimpleQueryParser to parse its context. Unlike the regular query_string query, the simple_query_string query will never throw an exception, and discards invalid parts of the query.

That sounds like it would solve the problem, and it seems like we should probably be using that for user-inputted searches anyways.

Manish, I'm sorry for how much this feature and discussion has ballooned, feel free to define the scope wherever seems appropriate.

mkangia: Sure Ethan.
I am going to keep it the way it is as of now with the original idea.
Feel free to make improvements later.

esoergel (approved):

mkangia:
QA finished https://manage.dimagi.com/default.asp?263951#1419774.
This is ready for live now.

@dimagi/product good to go?

sheelio:
@mkangia 

Can you outline the new behaviour of this?  Hard to follow with the different changes we made. Might be helpful to jump on a chat too if you're available to chat through this.

mkangia:
@sheelio 
Ben summarised everything [here](https://manage.dimagi.com/default.asp?263951#BugEvent.1416214)

mkangia:
@sheelio updated as discussed in [commit](https://github.com/dimagi/commcare-hq/pull/17958/commits/90c801c87e0cecc8888611686f83abcb3d1b4587)

sheelio:
Sounds good - will let Ethan sign off on the code change.

mkangia:
> Sounds good - will let Ethan sign off on the code change.
FYI @esoergel

esoergel (approved):

PR #17960, opened 2017-09-25
[enikshay] don't fix left column in worker activity report

biyeun:
@proteusvacuum 
https://manage.dimagi.com/default.asp?256191


biyeun:
@proteusvacuum bump

PR #17961, opened 2017-09-25
kill SoftwareProduct

nickpell:
This has been on staging for while - ready to merge before tomorrow's deploy.

@orangejenny


corehq/apps/accounting/forms.py
```diff
@@ -1372,26 +1360,22 @@ def _retrieve_feature_rate(self, rate_form):
...
             return new_rate
-        if product.id not in self.current_products_to_rates:
-            # the plan does not have this rate yet, compare any changes to the feature's current latest rate
-            # also mark the form as updated
-            current_rate = product.get_rate(default_instance=False)
-            if current_rate is None:
+        try:
+            current_rate = SoftwareProductRate.objects.get(id=rate_form['rate_id'].value())
+            # note: custom implementation of SoftwareProductRate.__eq__ here...
+            if not current_rate == new_rate:
```
orangejenny: Should this be `!=`? Isn't this equivalent to `(not current_rate) == new_rate`, which is probably not what you want?

corehq/apps/accounting/templates/accounting/version_summary_tab.html
```diff
@@ -16,7 +16,7 @@
                 </thead>
                 <tbody>
                     <tr>
-                        <td>{{ plan_version.product_rate.product.name }}</td>
+                        <td>{{ plan_version.product_rate }}</td>
```
orangejenny: why isn't this `plane_version.product_rate.name`?

corehq/apps/accounting/forms.py
nickpell: ```===``` is being used because ```SoftwareProductRate``` has a custom implementation of ```__eq__```

This expression is safe because ```===``` takes precedence over ```not``` in Python: http://www.informit.com/articles/article.aspx?p=459269&seqNum=11

Agree though that this is not clear and I'll introduce a followup PR with parentheses.

corehq/apps/accounting/templates/accounting/version_summary_tab.html
nickpell: ```SoftwareProductRate``` has a custom implementation of ```__str__``` which gets used here.  It shows both the name and the monthly fee.

orangejenny (approved):

corehq/apps/accounting/forms.py
orangejenny: Thanks for the link - yeah, parens would be nice for clarity. I've been doing more js than python lately, where `!` has precedence over `==`.

PR #17962, opened 2017-09-25
App builder: display notifications within form builder

orangejenny:
https://manage.dimagi.com/default.asp?260647

Depends on https://github.com/dimagi/Vellum/pull/921

@millerdev / @emord 
cc @nickpell

### DIFF
--- a/corehq/apps/app_manager/static/app_manager/js/forms/app_notifications.js
+++ b/corehq/apps/app_manager/static/app_manager/js/forms/app_notifications.js
@@ -3,34 +3,22 @@ hqDefine('app_manager/js/forms/app_notifications', function () {
     var getMessage = function(redisMessage, userId) {
         var msgObj = JSON.parse(redisMessage);
         // only show notifications from other users
-console.log("in getMessage, ids = " + msgObj.user_id + ", " + userId);
-debugger;
-        //if (msgObj.user_id !== userId) {
+        if (msgObj.user_id !== userId) {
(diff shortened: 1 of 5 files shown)


corehq/apps/app_manager/static/app_manager/js/forms/app_notifications.js
```diff
@@ -1,14 +1,27 @@
...
-    function NotifyFunction(userId) {
-        return function(msg) {
-            var msgObj = JSON.parse(msg);
-            // only show notifcations from other users
-            if (msgObj.user_id !== userId) {
-                var message = moment(msgObj.timestamp).format('h:mm:ss a') + ': ' + msgObj.text;
-                hqImport("hqwebapp/js/alert_user").alert_user(message, 'info', true);
+    var getMessage = function(redisMessage, userId) {
+        var msgObj = JSON.parse(redisMessage);
+        // only show notifications from other users
```
emord: Curious why this was implemented. Does this mean that you could have the form opened in multiple tabs and potentially screw yourself over that way?

no action necessary. I see it was this way before

emord (approved):

corehq/apps/app_manager/static/app_manager/js/forms/app_notifications.js
orangejenny: fwiw, it's because every act of opening the form triggers a notification, so without this you'd always see a notification that references yourself and comes from the current tab - but yes, it also means that if you open the form in multiple tabs you're on your own

millerdev (approved):

PR #17964, opened 2017-09-26
don't change original start_from date

snopoke:


esoergel (approved):

esoergel:
Looks like this is used for the case search and report case indices - do you know if this will have resulted in incomplete ES indices on any real environments?  Wondering if there are any necessary rebuilds.

snopoke:
@esoergel thankfully we don't. The only reindex that this might have affected would be the recent eNikshay one but since we ran a separate reindex process for each SQL DB this bug wouldn't have had any effect. We did also check doc counts after the reindex which matched up so all good.

PR #17965, opened 2017-09-26
How not to do it

kaapstorm:
Re: [FB 262139](http://manage.dimagi.com/default.asp?262139): "Potential Data Leakage Vulnerability (tt-patient-tracker)"

Figured I'd PR this for the sake of discussion. It seems that checking location safety after basic auth is going to be more painful than I thought, because the `is_location_safe` attribute on the view has been wrapped (and wrapped, and maybe wrapped again). It was late. Maybe I missed something obvious. But last night it seemed like following the code paths through corehq.apps.domain.decorators to find out where the original view function eventually pops out, and setting `is_location_safe` every step of the way, seemed like the wrong approach.

:blowfish:

### DIFF
--- a/corehq/apps/domain/auth.py
+++ b/corehq/apps/domain/auth.py
@@ -6,6 +6,7 @@ from functools import wraps
 from django.contrib.auth import authenticate
 from django.http import HttpResponse
 from tastypie.authentication import ApiKeyAuthentication
+from corehq.apps.users.models import CouchUser
 from corehq.toggles import ANONYMOUS_WEB_APPS_USAGE
 
 
@@ -84,12 +85,18 @@ def basicauth(realm=''):
     # stolen and modified from: https://djangosnippets.org/snippets/243/
     def real_decorator(view):
         def wrapper(request, *args, **kwargs):
+            from corehq.apps.locations.middleware import LocationAccessMiddleware
+
             uname, passwd = get_username_and_password_from_request(request)
             if uname and passwd:
                 user = authenticate(username=uname, password=passwd)
                 if user is not None and user.is_active:
                     request.user = user
-                    return view(request, *args, **kwargs)
+                    request.couch_user = CouchUser.from_django_user(user)
+                    # Check LocationAccessMiddleware again, now that the user is authenticated (FB 262139)
+                    location_middleware = LocationAccessMiddleware()
+                    response = location_middleware.process_view(request, view, args, kwargs)
+                    return view(request, *args, **kwargs) if response is None else response
 
             # Either they did not provide an authorization header or
             # something in the authorization attempt failed. Send a 401
--- a/corehq/apps/domain/auth.py
+++ b/corehq/apps/domain/auth.py
@@ -104,6 +104,10 @@ def basicauth(realm=''):
             response = HttpResponse(status=401)
             response['WWW-Authenticate'] = 'Basic realm="%s"' % realm
             return response
+
+        if hasattr(view, 'is_location_safe'):
+            wrapper.is_location_safe = view.is_location_safe
+            # TODO: ^^^ Nice try. Not good enough. See corehq.apps.domain.decorators
         return wrapper
     return real_decorator
 
--- a/corehq/apps/locations/middleware.py
+++ b/corehq/apps/locations/middleware.py
@@ -42,6 +42,11 @@ class LocationAccessMiddleware(MiddlewareMixin):
         else:
             request.can_access_all_locations = False
             if not is_location_safe(view_fn, view_args, view_kwargs):
+                # is_location_safe() is returning False because the wrapped (wrapped, wrapped (wrapped?)) view_fn
+                # no longer has is_location_safe attribute set on it.
+                # TODO: Find where it's being dropped, and set it on the wrapper function from the wrapped function
+                #       ^^^ For *every* code path? See corehq.apps.domain.decorators. This doesn't seem like such
+                #       a good idea any more.
                 return location_restricted_response(request)
             elif not user.get_sql_location(domain):
                 return no_permissions(request, message=RESTRICTED_USER_UNASSIGNED_MSG)
--- a/corehq/apps/domain/tests/test_auth.py
+++ b/corehq/apps/domain/tests/test_auth.py
@@ -1,0 +1,47 @@
+from base64 import b64encode
+from django.http import HttpResponse
+from mock import Mock
+from corehq.apps.domain.decorators import basic_auth
+from corehq.apps.locations.permissions import location_safe
+from corehq.apps.users.models import WebUser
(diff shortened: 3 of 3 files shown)


corehq/apps/domain/auth.py
```diff
@@ -104,6 +104,10 @@ def wrapper(request, *args, **kwargs):
             response = HttpResponse(status=401)
             response['WWW-Authenticate'] = 'Basic realm="%s"' % realm
             return response
+
+        if hasattr(view, 'is_location_safe'):
+            wrapper.is_location_safe = view.is_location_safe
+            # TODO: ^^^ Nice try. Not good enough. See corehq.apps.domain.decorators
```
esoergel: I think decorating `wrapper` with `@wraps(view)` will preserve attributes set on the view function (plus things will be a lot easier to debug if you ever try to interact with that view in a shell session)

esoergel (commented):
Another possible approach would be to do something like this in the location middleware:
```python
if not user and is_definitely_trying_to_use_basic_auth(request):
    user = get_user_in_the_same_way_basic_auth_does(request)
```
Then the location middleware can process it as normal, and if the auth is bad, it'll fail when it hits the auth decorator.

If preferable, this same approach could also be done using the auth middleware - that is:
```python
# Auth Middleware
if is_definitely_trying_to_use_basic_auth(request):
    username, _ = get_username_and_password_from_request(request)
    # check auth info here if you like, and maybe save it somewhere, but don't set the
    # authenticated django user on the request. I don't think this will actually be
    # considered "authenticated" unless
    #     request.user.is_authenticated == True
    request.couch_user = get_couch_user_from_username(username)
    request.using_basic_auth = True
else:
    request.using_basic_auth = False
    # do normal auth process


# Location Middleware
if not user and request.using_basic_auth:
    user = request.couch_user
...


# basic auth decorator
if not request.using_basic_auth:
    raise  # or return or however it fails
# I'm not concerned about the overhead of pulling this info twice:
username, password = get_username_and_password_from_request(request)
# if this DID auth in the middleware, check whatever that set and auth for real here
```

That approach might need a little more fleshing out, but it could let us have our cake and eat it too.

corehq/apps/domain/auth.py
```diff
@@ -84,19 +85,29 @@ def basicauth(realm=''):
...
             if uname and passwd:
                 user = authenticate(username=uname, password=passwd)
                 if user is not None and user.is_active:
                     request.user = user
-                    return view(request, *args, **kwargs)
+                    request.couch_user = CouchUser.from_django_user(user)
+                    # Check LocationAccessMiddleware again, now that the user is authenticated (FB 262139)
+                    location_middleware = LocationAccessMiddleware()
+                    response = location_middleware.process_view(request, view, args, kwargs)
+                    return view(request, *args, **kwargs) if response is None else response
```
esoergel: Are there any other auth types which similarly need this restriction?

esoergel:
One last thought - if the location restriction middleware is ever called with `domain` and not `user`, then we should probably either identify it as a known exception (such as basic auth), or fail hard, or at least notify as a possible exception so we can investigate.  That would let us proactively catch any future issues along these lines.

corehq/apps/domain/auth.py
snopoke: `@wraps` will only copy ` __name__, __module__ and __doc__` (https://docs.python.org/2/library/functools.html#functools.update_wrapper)

corehq/apps/domain/auth.py
calellowitz: i believe we need this for api key auth and digest auth as well

corehq/apps/domain/auth.py
calellowitz: I think this has the same issue as the initial code, this is called after the middleware finishes

corehq/apps/locations/middleware.py
```diff
@@ -42,6 +42,11 @@ def process_view(self, request, view_fn, view_args, view_kwargs):
         else:
             request.can_access_all_locations = False
             if not is_location_safe(view_fn, view_args, view_kwargs):
+                # is_location_safe() is returning False because the wrapped (wrapped, wrapped (wrapped?)) view_fn
+                # no longer has is_location_safe attribute set on it.
+                # TODO: Find where it's being dropped, and set it on the wrapper function from the wrapped function
+                #       ^^^ For *every* code path? See corehq.apps.domain.decorators. This doesn't seem like such
+                #       a good idea any more.
```
calellowitz: i am not sure its getting dropped, i think it hasnt been set yet because the view executes (with the wrapping decorator) after the middleware finishes.

millerdev:
I may not be following everything that's going on there, but if the problem is that `wrap(wrap(wrap(view))).is_location_safe` is raising `AttributeError` due to wrappers not preserving the attribute, would it work to set `request.is_location_safe = safety_value` at the point when location safety is determined during the request process? Then other code can check `getattr(request, 'is_location_safe', False)` rather than `maybe_wrapped_view.is_location_safe`. This would mean that any middleware that executes before location safety is determined during request processing would not have access to this attribute, and maybe that's the problem?

corehq/apps/domain/auth.py
esoergel: I think it also copies `__dict__` by default as part of `WRAPPER_UPDATES`:
```python
In [1]: from functools import wraps

In [2]: def my_decorator(f):
   ...:     @wraps(f)
   ...:     def wrapper(*args, **kwargs):
   ...:         return f(*args, **kwargs)
   ...:     return wrapper
   ...: 

In [3]: def my_bad_decorator(f):
   ...:     def wrapper(*args, **kwargs):
   ...:         return f(*args, **kwargs)
   ...:     return wrapper
   ...: 

In [4]: def set_foo(f):
   ...:     f.foo = 'bar'
   ...:     return f
   ...: 

In [5]: @my_decorator
   ...: @set_foo
   ...: def my_fn():
   ...:     return 3
   ...: 

In [6]: @my_bad_decorator
   ...: @set_foo
   ...: def my_bad_fn():
   ...:     return 3
   ...: 

In [7]: my_fn.foo
Out[7]: 'bar'

In [8]: my_bad_fn.foo
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-8-936c6937c7d3> in <module>()
----> 1 my_bad_fn.foo

AttributeError: 'function' object has no attribute 'foo'
```
Not sure it solves the problem here anyways, as Cal said.

esoergel:
@kaapstorm FYI I just ran across this function which may come in useful here (not sure what the current status is):
https://github.com/dimagi/commcare-hq/blob/master/corehq/apps/domain/auth.py#L38-L52
That's used in some auth decorators to determine which authentication method to use.

kaapstorm:
Closing to open a new PR

PR #17970, opened 2017-09-26
fix auto rebuild check logic

snopoke:


corehq/apps/export/export.py
```diff
@@ -465,25 +465,29 @@ def _get_base_query(export_instance):
...
-    if _should_not_rebuild_export(export_instance, last_access_cutoff):
-        return
     filters = filters or export_instance.get_filters()
     export_file = get_export_file([export_instance], filters or [])
     with export_file as payload:
         save_export_payload(export_instance, payload)
 
 
-def _should_not_rebuild_export(export, last_access_cutoff):
+def _should_rebuild_export(export, last_access_cutoff=None):
```
millerdev: nit: should remove leading `_` since this is now imported from other modules (no longer private).

millerdev: curious why the default was added (`last_access_cutoff=None`) since this function is never called with just a single argument?

corehq/apps/export/export.py
```diff
@@ -465,25 +465,29 @@ def _get_base_query(export_instance):
...
+    :return: True if export should be rebuilt
+    """
     # Don't rebuild exports that haven't been accessed since last_access_cutoff or aren't enabled
-    return (
-        last_access_cutoff
-        and export.auto_rebuild_enabled
-        and export.last_accessed
-        and export.last_accessed < last_access_cutoff
+    is_auto_rebuild = last_access_cutoff is not None
+    return not is_auto_rebuild or (
```
millerdev: The above two lines are a confusing. It seems to be saying "rebuild if not auto rebuild," which seems like a contradiction. Maybe I'm not understanding the context? Is it true that a rebuild should always happen if `last_access_cutoff is None`? If so, maybe simplify this to
```py
return last_access_cutoff is None or (
```

snopoke: yes, that is true. I introduced the variable to make it clearer what `last_access_cutoff is None` means. `last_access_cutoff` is only `None` if the function is called directly from a user request as opposed to via the scheduled task.

Fine to change it back, just thought it made this easier to read.

corehq/apps/export/export.py
snopoke: Not sure, will remove.

millerdev (approved):

PR #17974, opened 2017-09-26
fixing consumption download js

calellowitz:
@esoergel @sravfeyn cc: @gcapalbo 
http://manage.dimagi.com/default.asp?261079#1397638
looks like when we removed the checkbox to download location ids we didnt quite finish cleaning up the js


esoergel (approved):
Oh, and was this just causing a JS exception?

calellowitz:
Yep, just a js error, when you hit that checkbox

PR #17977, opened 2017-09-26
enikshay: Easily filter public private cases

sheelio:
@nickpell 

This change removes the private cases from the CSV outputs (making it easier to parse and sign-off on those outputs).


nickpell:
nice

PR #17983, opened 2017-09-26
add parens for clarity

nickpell:
@orangejenny


orangejenny (approved):
such attention to detail - thanks!

PR #17986, opened 2017-09-27
Changes in TB Notification Register

kkrampa:
* Health facility for treatment row moved to be after diagnosing facility
* Treating PHI and Treating TU to be renamed as Currently Treating PHI and Currently Treating TU


kkrampa:
@calellowitz

PR #17995, opened 2017-09-27
Add commands to deduplicate and send repeat records

esoergel:
@proteusvacuum @dannyroberts
Thoughts on the general approach here?  I discussed the deduplication with Danny yesterday and we could also use an approach which doesn't delete any records of attempts, such as either save to another bucket that's not "CANCELLED", or to add those attempts to the repeat record that we'll be keeping.  I'm not sure either gives us much actual value, so I decided to keep it simple unless someone makes a case for keeping them.

buddy: @mkangia

### DIFF
--- a/corehq/motech/repeaters/management/commands/delete_duplicate_cancelled_records.py
+++ b/corehq/motech/repeaters/management/commands/delete_duplicate_cancelled_records.py
@@ -1,0 +1,61 @@
+import csv
+import datetime
+from collections import defaultdict
+
+from django.core.management.base import BaseCommand
+
+from corehq.util.couch import IterDB
+from corehq.motech.repeaters.const import RECORD_CANCELLED_STATE
+from corehq.motech.repeaters.models import RepeatRecord
+from corehq.motech.repeaters.dbaccessors import iter_repeat_records_by_domain
+
+
+class Command(BaseCommand):
+    help = """
+    If there are multiple cancelled repeat records for a given payload id, this
+    will delete all but one for each payload, reducing the number of requests
+    that must be made.
+    """
+
+    def add_arguments(self, parser):
+        parser.add_argument(
+            'domain',
+        )
+        parser.add_argument(
+            'repeater_id',
+        )
+
+    def handle(self, domain, repeater_id, *args, **options):
+        records_by_payload_id = defaultdict(list)
+        records = iter_repeat_records_by_domain(domain, repeater_id=repeater_id, state=RECORD_CANCELLED_STATE)
+        total_records = 0
+        for record in records:
+            records_by_payload_id[record.payload_id].append(record)
+            total_records += 1
+
+        unique_payloads = len(records_by_payload_id)
+        print ("There are {} total records and {} unique payload ids."
+               .format(total_records, unique_payloads))
+        print "Delete {} duplicate records?".format(total_records - unique_payloads)
+        if not raw_input("(y/n)") == 'y':
+            print "Aborting"
+            return
+
+        log = resolve_duplicates(records_by_payload_id)
+        filename = "cancelled_repeat_records-{}.csv".format(datetime.datetime.utcnow().isoformat())
+        print "Writing log of changes to {}".format(filename)
+        with open(filename, 'w') as f:
+            writer = csv.writer(f)
+            writer.writerows(log)
+
+
+def resolve_duplicates(records_by_payload_id):
+    log = [('RepeatRecord ID', 'Payload ID', 'Deleted?')]
+    with IterDB(RepeatRecord.get_db()) as db:
+        for payload_id, records in records_by_payload_id.items():
+            log.append((records[0]._id, payload_id, records[0].failure_reason, 'No'))
+            if len(records) > 1:
+                for record in records[1:]:
+                    db.delete(record)
+                    log.append((record._id, payload_id, record.failure_reason, 'Yes'))
+    return log
--- a/corehq/motech/repeaters/management/commands/delete_duplicate_cancelled_records.py
+++ b/corehq/motech/repeaters/management/commands/delete_duplicate_cancelled_records.py
@@ -51,11 +51,11 @@ class Command(BaseCommand):
 
 def resolve_duplicates(records_by_payload_id):
     log = [('RepeatRecord ID', 'Payload ID', 'Deleted?')]
-    with IterDB(RepeatRecord.get_db()) as db:
+    with IterDB(RepeatRecord.get_db()) as iter_db:
         for payload_id, records in records_by_payload_id.items():
             log.append((records[0]._id, payload_id, records[0].failure_reason, 'No'))
             if len(records) > 1:
                 for record in records[1:]:
-                    db.delete(record)
+                    iter_db.delete(record)
                     log.append((record._id, payload_id, record.failure_reason, 'Yes'))
     return log
--- a/corehq/motech/repeaters/management/commands/send_cancelled_records.py
+++ b/corehq/motech/repeaters/management/commands/send_cancelled_records.py
@@ -1,0 +1,85 @@
+import csv
+import datetime
+import re
+import time
+
+from django.core.management.base import BaseCommand
+
+from corehq.motech.repeaters.const import RECORD_CANCELLED_STATE
+from corehq.motech.repeaters.dbaccessors import iter_repeat_records_by_domain
+
+
+class Command(BaseCommand):
+    help = """
+    Send cancelled repeat records. You may optionally specify a regex to
+    filter records using --include or --exclude, an a sleep time with --sleep
+    """
+
+    def add_arguments(self, parser):
+        parser.add_argument('domain')
+        parser.add_argument('repeater_id')
+        parser.add_argument(
+            '--include',
+            dest='include_regex',
+            help=("Regex that will be applied to a record's 'failure_reason' to "
+                  "determine whether to include it."),
+        )
+        parser.add_argument(
+            '--exclude',
+            dest='exclude_regex',
+            help=("Regex that will be applied to a record's 'failure_reason' to "
+                  "determine whether to exclude it."),
+        )
+        parser.add_argument(
+            '--sleep',
+            dest='sleep_time',
+            help="Time in seconds to sleep between each request.",
+        )
+
+    def handle(self, domain, repeater_id, *args, **options):
+        sleep_time = options.get('sleep_time')
+        include_regex = options.get('include_regex')
+        exclude_regex = options.get('exclude_regex')
+        if include_regex and exclude_regex:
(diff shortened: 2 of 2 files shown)


corehq/motech/repeaters/management/commands/delete_duplicate_cancelled_records.py
```diff
@@ -0,0 +1,61 @@
...
+        print "Writing log of changes to {}".format(filename)
+        with open(filename, 'w') as f:
+            writer = csv.writer(f)
+            writer.writerows(log)
+
+
+def resolve_duplicates(records_by_payload_id):
+    log = [('RepeatRecord ID', 'Payload ID', 'Deleted?')]
+    with IterDB(RepeatRecord.get_db()) as db:
+        for payload_id, records in records_by_payload_id.items():
```
mkangia: Not sure if this makes a difference but would it be better to keep the latest fired/created repeat record and delete others instead?

corehq/motech/repeaters/management/commands/delete_duplicate_cancelled_records.py
```diff
@@ -0,0 +1,61 @@
...
+            log.append((records[0]._id, payload_id, records[0].failure_reason, 'No'))
+            if len(records) > 1:
+                for record in records[1:]:
+                    db.delete(record)
```
mkangia: Whats the reason for using db.delete here?  Isn't there an option to use record.delete()?

corehq/motech/repeaters/management/commands/send_cancelled_records.py
```diff
@@ -0,0 +1,85 @@
...
+    """
+
+    def add_arguments(self, parser):
+        parser.add_argument('domain')
+        parser.add_argument('repeater_id')
+        parser.add_argument(
+            '--include',
+            dest='include_regex',
+            help=("Regex that will be applied to a record's 'failure_reason' to "
+                  "determine whether to include it."),
```
mkangia: 👍

corehq/motech/repeaters/management/commands/send_cancelled_records.py
```diff
@@ -0,0 +1,85 @@
...
+        total_records = len(records)
+        print "Found {} matching records.  Requeue them?".format(total_records)
+        if not raw_input("(y/n)") == 'y':
+            print "Aborting"
+            return
+
+        log = [('record_id', 'payload_id', 'state', 'failure_reason')]
+        for i, record in enumerate(records):
+            try:
+                record.fire(force_send=True)
```
mkangia: Not sure if you have tested this but this failed for me when retriggering cancelled records due to next check on the record being blank. I am not 100% sure but i think it fails [here](https://github.com/dimagi/commcare-hq/blob/master/corehq/motech/repeaters/models.py#L574)
you might have to add [this](https://github.com/dimagi/commcare-hq/blob/master/custom/enikshay/management/commands/retrigger_private_nikshay_records.py#L147)

corehq/motech/repeaters/management/commands/send_cancelled_records.py
```diff
@@ -0,0 +1,85 @@
...
+                return not bool(re.search(exclude_regex, record.failure_reason))
+            return True  # No filter applied
+
+        records = filter(
+            meets_filter,
+            iter_repeat_records_by_domain(domain, repeater_id=repeater_id, state=RECORD_CANCELLED_STATE)
+        )
+
+        total_records = len(records)
+        print "Found {} matching records.  Requeue them?".format(total_records)
```
mkangia: This does not do a requeue but instead fires them in place. So this should be "Trigger them?"?

corehq/motech/repeaters/management/commands/send_cancelled_records.py
```diff
@@ -0,0 +1,85 @@
...
+                print "{}/{}: {}".format(i, total_records, record.state)
+                log.append((record._id, record.payload_id, record.state, record.failure_reason))
+            if sleep_time:
+                time.sleep(float(sleep_time))
+
+        filename = "sent_repeat_records-{}.csv".format(datetime.datetime.utcnow().isoformat())
+        print "Writing log of changes to {}".format(filename)
+        with open(filename, 'w') as f:
+            writer = csv.writer(f)
+            writer.writerows(log)
```
mkangia: Writing logs right in the end might result in losing all the logs in case there is an error. I know you have used a general exception when firing, probably the only places where things would fail. Might be safer to just write in the csv file during the integration itself.

corehq/motech/repeaters/management/commands/send_cancelled_records.py
```diff
@@ -0,0 +1,85 @@
...
+            except Exception as e:
+                print "{}/{}: {} {}".format(i, total_records, 'EXCEPTION', repr(e))
+                log.append((record._id, record.payload_id, record.state, repr(e)))
+            else:
+                print "{}/{}: {}".format(i, total_records, record.state)
+                log.append((record._id, record.payload_id, record.state, record.failure_reason))
+            if sleep_time:
+                time.sleep(float(sleep_time))
+
+        filename = "sent_repeat_records-{}.csv".format(datetime.datetime.utcnow().isoformat())
```
mkangia: `2017-09-28T09:59:07.518073` in the name might cause issues when copying this file to local machine. I am unable to set that in the file name directly on my machine, though not sure if its okay in case the file was already created somewhere else (on the server?) and then transferred.

mkangia (approved):
Nothing blocking.

corehq/motech/repeaters/management/commands/delete_duplicate_cancelled_records.py
esoergel: That's a good question - I'm not sure, although I think if that IS preferable, we'd probably want to keep some sort of record of all attempts anyways.

corehq/motech/repeaters/management/commands/delete_duplicate_cancelled_records.py
esoergel: `db` here is an instance of `IterDB` (I probably should have called it `iter_db`), so by using this over `record.delete()` it commits the deletions chunked in bulk.

corehq/motech/repeaters/management/commands/send_cancelled_records.py
esoergel: :+1:

corehq/motech/repeaters/management/commands/send_cancelled_records.py
esoergel: Thanks, I'll also look over that second file - looks pretty relevant to this.

corehq/motech/repeaters/management/commands/send_cancelled_records.py
esoergel: Good point - it works fine on linux, but I suppose not mac.  Will update.

corehq/motech/repeaters/management/commands/delete_duplicate_cancelled_records.py
mkangia: Isn't that risky here? it the chunk fails in between we would not know what went through and what did not. Even if this did a complete roll back, we have logged them as deleted already.
You can skip changing it here because it looks safe to delete these. We can re-run this script anyway for deleting duplicates.

corehq/motech/repeaters/management/commands/delete_duplicate_cancelled_records.py
mkangia: Just clearing what i meant in case you misunderstood. What i meant was that you have picked the first in the list and deleted others. What i meant is that if we want to keep the latest fired/created record and delete the older ones instead to have when the last time the repeater fired and failed instead.

corehq/motech/repeaters/management/commands/delete_duplicate_cancelled_records.py
esoergel: That's true.  `IterDB` does keep track of which documents saved or errored, so this could use that, although it'd be somewhat tedious to link it back up with the original payload ID and failure reason.  I do think in this instance that re-running the script as you suggested would be an acceptable failure mode.

corehq/motech/repeaters/management/commands/delete_duplicate_cancelled_records.py
esoergel: I wasn't sure whether we'd want to preserve the newest, the oldest, or all of the attempts (or whether we care at all).  Along those lines, my preference would be to take the `attempts` from the deleted records, and add them on to the `attempts` from the record we're going to keep, so it looks like one record with a lot of attempts.  You could also make a case for saving the "extra" records in a separate bucket that's neither CANCELLED nor SUCCESS, and just for record keeping.  My inclination is to leave it simple until we know we do care, and then modify it to do one of the above.

esoergel: Relatedly, in your work with Nikshay, what behavior would have been most useful?

mkangia: The payload generated by a repeat record would be the same in case of Nikshay repeaters for a particular payload id. In that case i would keep the latest fired repeat record in case there has been no successful attempt for that payload id.
This reminds me, what if for a payload there are both cancelled and successful attempts. Currently this wold just delete all cancelled records except 1. 
Just a suggestion, Should we just get rid of all cancelled records in that case in case they come after the successful attempt? Can ignore it if thats not in the scope of what this PR wants to do.

esoergel:
@mkangia added a couple commits responding to feedback.

corehq/motech/repeaters/management/commands/delete_duplicate_cancelled_records.py
esoergel: For the BETS use-case that's not desirable, since the cancelled records may represent a real update where we DO want a second record sent.  Feels like there are a lot of different ways this sort of thing may be useful, I'm thinking of keeping that out of scope and maybe extending and parameterizing this as needed.

corehq/motech/repeaters/management/commands/send_cancelled_records.py
```diff
@@ -60,14 +60,16 @@ def meets_filter(record):
...
+        print "Found {} matching records.  Send them?".format(total_records)
         if not raw_input("(y/n)") == 'y':
             print "Aborting"
             return
 
         log = [('record_id', 'payload_id', 'state', 'failure_reason')]
         for i, record in enumerate(records):
             try:
+                if record.next_check is None:
+                    record.next_check = datetime.datetime.utcnow()
```
mkangia: 👍

mkangia (approved):
Looks good.
Thanks for considering the suggestions.

PR #17997, opened 2017-09-28
update application api with build and version info

snopoke:


corehq/apps/api/resources/v0_4.py
```diff
@@ -371,7 +371,28 @@ class ApplicationResource(CouchResourceMixin, HqBaseResource, DomainSpecificReso
...
+    built_on = fields.DateTimeField(attribute='built_on', null=True)
+    build_comment = fields.CharField(attribute='build_comment', null=True)
+    built_from_app_id = fields.CharField(attribute='copy_of', null=True)
     modules = fields.ListField()
+    versions = fields.ListField(use_in='detail')
+
+    def dehydrate_versions(self, bundle):
+        app = bundle.obj
+        if app.copy_of:
+            return None
```
millerdev: Could this raise an exception? IOW, is it a programmer error to call this method for an app with `bool(copy_of) == True`? Otherwise, it seems that all callers of this method will need to test the result to see if it is None, and then try to handle that situation as best as they can (and maybe they shouldn't if it's a programmer error).

Alternately, could this just `return []` here, which at least means the function always returns a list. Not sure if that would make sense or not in this context.

corehq/apps/api/resources/v0_4.py
```diff
@@ -371,7 +371,28 @@ class ApplicationResource(CouchResourceMixin, HqBaseResource, DomainSpecificReso
...
     name = fields.CharField(attribute='name')
     version = fields.IntegerField(attribute='version')
+    is_released = fields.BooleanField(attribute='is_released', null=True)
+    built_on = fields.DateTimeField(attribute='built_on', null=True)
+    build_comment = fields.CharField(attribute='build_comment', null=True)
+    built_from_app_id = fields.CharField(attribute='copy_of', null=True)
     modules = fields.ListField()
+    versions = fields.ListField(use_in='detail')
+
+    def dehydrate_versions(self, bundle):
```
millerdev: Looks like this could be a `staticmethod` since `self` is never referenced.

millerdev (approved):
Approving, but might make sense to change some things (see comments).

corehq/apps/api/resources/v0_4.py
snopoke: I did originally have it as a list but changed it to None because I thought it made more sense. App builds (`bool(app.copy_of) == True`) can't have builds (versions) of their own so `None` seemed like the right thing rather than an empty list which indicates that there just aren't any builds.

This shouldn't raise an exception since it will get called with app builds.

I'm on the fence about the `None` vs `[]` thing. Let me know if my explanation makes sense.

millerdev: `None` vs `[]`: does the caller need to distinguish between an app that has no versions (signified by `[]`) and an app build (`bool(copy_of) == True` -> `None`)? And is that even the correct way they should be making that distinction? If not, then I would return `[]`.

Generally it's much easier to reason about calling code when the function can only return a single type of result. If it can return `None` or `[]` then every caller has to handle `if versions is None: ...` or have a deeper understanding of this method's implementation to know that the value they're passing in will never produce a result of `None`. Always returning a list means they can simply do `for version in versions: ...` or whatever without worrying about special cases.

snopoke: Thanks for the lesson. I did consider that but since there's isn't an API endpoint that includes apps and builds that didn't seem like an issue i.e. the user would always know whether or not they're pulling an app build.

PR #17999, opened 2017-09-28
Fix ICDS web apps link

emord:
The dashboard is supposed to link to web apps. Before we were using an old build id which broke when upgrading that app.

This will get the latest released build id for the issue tracker app.

buddies @gcapalbo @calellowitz

### DIFF
--- a/custom/icds_reports/const.py
+++ b/custom/icds_reports/const.py
@@ -1,6 +1,6 @@
 from django.conf import settings
 
-APP_ID = '361a135293a84427afc947beee1acdfe'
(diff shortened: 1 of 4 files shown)


custom/icds_reports/const.py
```diff
@@ -1,6 +1,6 @@
 from django.conf import settings
 
-APP_ID = '361a135293a84427afc947beee1acdfe'
+APP_ID = '48cc1709b7f62ffea24cc6634a005734'
```
gcapalbo: i realize this was already here but seems like this should have a better name

PR #18002, opened 2017-09-28
Search All Cases: Match exact and then fuzzy

kaapstorm:
Context: [FB 261367](https://manage.dimagi.com/default.asp?261367)

Preserves current behaviour for non-fuzzy searches. For fuzzy searches, uses a "should" clause instead of a "must" clause, and searches for exact matches before fuzzy matches.

You can see the new query for fuzzy searches in the [Update test](1523a0b) commit.

**DISCLAIMER**: I did NOT repro the original issue. I am relying on unit tests to prove that this change does what it is meant to do. Feel free to tell me to go back and do more homework. :grimacing: 

@proteusvacuum @esoergel cc @dannyroberts

### DIFF
--- a/corehq/apps/es/case_search.py
+++ b/corehq/apps/es/case_search.py
@@ -36,7 +36,7 @@ class CaseSearchES(CaseES):
         except (KeyError, TypeError):
             return []
 
-    def case_property_query(self, key, value, clause=queries.MUST, fuzzy=False):
+    def case_property_query(self, key, value, clause=None, fuzzy=False):
         """
         Search for a case property.
         Usage: (CaseSearchES()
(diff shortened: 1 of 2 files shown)


esoergel:
I just tried this query out, it has the general structure of:
```
Query
    Must
        filter 1
        filter 2
    Should
        filter 3
        filter 3 fuzzy
```
This makes things matching filter 3 appear at the top, and weights exact matches above fuzzy matches, which we want, however, after that it includes results which _don't_ match filter 3 at all.

Setting "minimum_should_match" to 1 (or I guess the number of fuzzy queries?) prevents this.  Another approach worth considering is:
```
Query
    Must
        filter 1
        filter 2
        filter 3 fuzzy
    Should
        filter 3
```
That makes the filter 3 exact matches rate higher, and excludes anything that doesn't match the fuzzy filter.  Not sure which of those two is preferable, but they both seem to work fine.

proteusvacuum:
I think we should go with the second option: If someone has entered a search query for filter 3, we would want it to return nothing if it doesn't find any matches.

corehq/apps/es/case_search.py
```diff
@@ -49,16 +49,29 @@ def case_property_query(self, key, value, clause=queries.MUST, fuzzy=False):
...
             PATH,
             queries.filtered(
-                queries.match(value, "{}.value".format(PATH), fuzziness),
-                filters.term("{}.key".format(PATH), key)
+                queries.match(value, '{}.value'.format(PATH), fuzziness='0'),
+                filters.term('{}.key'.format(PATH), key),
             )
         )
-        return self._add_query(new_query, clause)
+        return result._add_query(exact_query, clause)
```
esoergel: This all looks correct.  From a stylistic perspective, I find it a little hard to follow the fuzzy and clause flags, would be nice to see something very explicit, like:
```python
if fuzzy:
    result._add_query(get_query(fuzziness='AUTO'), queries.MUST)
    result._add_query(get_query(fuzziness='0'), queries.SHOULD)
else:
    result._add_query(get_query(fuzziness='0'), clause)
```
Also, passing in both `fuzzy=True` negates whatever you specified for `clause`, which I could see being confusing.  You could add a check against that, or not, since I could see that being parameterized.  Actually, what about just making a `fuzzy_case_property_query` that's separate from `case_property_query`?  I haven't looked into where this is used - that might not be straightforward, so feel free to ignore.

esoergel (approved):
Looks good, thanks for the updates!  I had some optional stylistic feedback, but this is g2m by me.

PR #18009, opened 2017-09-28
migration for nikshay patient type

nickpell:
https://trello.com/c/YQ6NoNjt/216-migration-for-nikshay-patient-type

Adds a datamigration to update existing data and also updates the nikshay migration script for reference and for future roll-outs.

@sheelio


sheelio:
i'm not too familiar with this migration, so may want to loop in Manish or Farid.

custom/enikshay/management/commands/run_current_patient_type_choice_migration.py
```diff
@@ -0,0 +1,39 @@
+from custom.enikshay.management.commands.utils import BaseEnikshayCaseMigration
```
mkangia: 👍

mkangia (approved):
Looks good in terms of the migration. 
Assuming sheel/you confirmed the changes otherwise,

PR #18010, opened 2017-09-28
Data dictionary: don't error out for domains without cases

orangejenny:
js error on staging was tripping me up

@calellowitz


orangejenny:
or code buddy @nickpell

nickpell (approved):

PR #18012, opened 2017-09-28
Add BETS Voucher report

proteusvacuum:
@esoergel 
This adds a report with the following info:

Voucher Case ID | Voucher Readable ID | Voucher Status | Voucher District ID | Voucher Type | Voucher Approved Amount | Voucher Beneficiary ID | BETS Sent Date | Forwading Status | BETS Response Message
-- | -- | -- | -- | -- | -- | -- | -- | -- | --

So that EY can start to debug BETS voucher issues on their own. 

@biyeun

### DIFF
--- a/custom/enikshay/reports/choice_providers.py
+++ b/custom/enikshay/reports/choice_providers.py
@@ -1,0 +1,19 @@
+from corehq.apps.locations.models import SQLLocation
+from corehq.apps.userreports.reports.filters.choice_providers import LocationChoiceProvider
+
+
+class DistrictChoiceProvider(LocationChoiceProvider):
+    def _locations_query(self, query_text, user):
+        active_locations = SQLLocation.active_objects
+        if query_text:
+            return active_locations.filter_path_by_user_input(
+                domain=self.domain,
+                location_type__code__exact='dto',
+                user_input=query_text
+            ).accessible_to_user(self.domain, user)
+        return active_locations.accessible_to_user(self.domain, user).filter(
+            domain=self.domain, location_type__code__exact='dto',)
+
+    def get_choices_for_known_values(self, values, user):
+        selected_locations = self._locations_query(None, user).filter(location_id__in=values)
+        return self._locations_to_choices(selected_locations)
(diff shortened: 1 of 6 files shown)


custom/enikshay/reports/choice_providers.py
```diff
@@ -0,0 +1,19 @@
...
+from corehq.apps.userreports.reports.filters.choice_providers import LocationChoiceProvider
+
+
+class DistrictChoiceProvider(LocationChoiceProvider):
+    def _locations_query(self, query_text, user):
+        active_locations = SQLLocation.active_objects
+        if query_text:
+            return active_locations.filter_path_by_user_input(
+                domain=self.domain,
+                location_type__code__exact='dto',
```
esoergel: I don't think this is a valid parameter for `filter_path_by_user_input` - you may just want `active_locations.filter_by_user_input(...).filter(location_type__code__exact='dto')`
(assuming that what you want is for users to type only dto locations)

esoergel: Actually, this is being filtered again later anyways, so you can probably just drop this line.  Switch to `filter_by_user_input` if what you want is for typing to match against `dto` names only, otherwise this will match against `dto`s or their parents.

custom/enikshay/reports/repeaters.py
```diff
@@ -112,3 +122,89 @@ def _get_case_id_link(self, case_id):
...
+                ]
+            )
+        return rows
+
+    def _get_voucher_cases(self, district_id, voucher_state):
+        cs = CaseSearchES()
+        cs = cs.domain('enikshay')
+        cs = cs.case_type('voucher')
+        if district_id:
+            cs = cs.case_property_query('voucher_district_id', district_id)
```
esoergel: Should `district` be required?  Without it, this could be a very large report.

proteusvacuum:
Gah. So `voucher_district_id` is actually the district where the voucher was _created_, not where it was fulfilled, so this isn't the correct property to use for this report. I think this might actually be the reason that we have mismatches between the payment register (what they were pulling from) and what we send to BETS through the API.

PR #18018, opened 2017-09-29
Referral report

kkrampa:
http://manage.dimagi.com/default.asp?262732#1398937
@calellowitz


kkrampa:
@calellowitz Could you merge this PR?

custom/enikshay/ucr/data_sources/person_for_referral_report_v2.json
```diff
@@ -1323,6 +1322,254 @@
...
+            },
+            {
+                "display_name": "episode_type",
+                "transform": {},
+                "datatype": "string",
+                "type": "expression",
+                "is_primary_key": false,
+                "is_nullable": true,
+                "column_id": "episode_type",
+                "expression": {
```
calellowitz: i dont think this meets the spec. it looks like you should display `Confirmed DSTB (Patient) ` for `confirmed_tb` and `Confirmed DRTB (Patient)` for `confirmed_drtb`. additionally, i spoke with sarvesh and it seems there should be a third option `Presumptive TB` for `presumptive_tb`

custom/enikshay/ucr/data_sources/person_for_referral_report_v2.json
```diff
@@ -1323,6 +1322,254 @@
...
+                        }
+                    },
+                    "value_expression": {
+                        "type": "property_name",
+                        "property_name": "episode_type"
+                    }
+                }
+            },
+            {
+                "display_name": "adherence_schedule_id",
```
calellowitz: same as above. i dont think this meets the spec. It should be `schedule_daily` -> `Daily`, `schedule_mwf` -> `Intermittent (MWF)`, and `schedule_trs` -> `Intermittent (TTS)`

calellowitz (changes requested):
You appear to have missed the case definition column, as well as the separate columns for TU and PHI.
Additionally, these columns should be added to `referral_report.json`. Since the report is live right now, it would be good to create a v2 of that as well. It is also possible to do the display transforms I commented on, in the report itself if you prefer but it seems a bit cleaner to store them in the data source.

kkrampa:
@calellowitz I prefer to do this on the report side since when a new value will be added we don't need to rebuild data source. Any benefits from keeping this in the data source?

calellowitz:
@kkrampa report is fine

kkrampa:
@calellowitz Updated

calellowitz:
I think you are still missing the case definition column

kkrampa:
Added

calellowitz:
@kkrampa looking to push this out today. can we fix the merge conflicts

calellowitz:
@kkrampa looks good once you fix those lint issues

PR #18020, opened 2017-09-29
add command to soft delete extention cases without person

nickpell:
https://trello.com/c/XtU2jLmX/218-cleanup-orphaned-cases

@proteusvacuum @mkangia cc: @orangejenny

### DIFF
--- a/custom/enikshay/management/commands/soft_delete_cases_with_no_person.py
+++ b/custom/enikshay/management/commands/soft_delete_cases_with_no_person.py
@@ -3,6 +3,7 @@ import csv
 from django.core.management import BaseCommand
 
 from corehq.form_processor.interfaces.dbaccessors import CaseAccessors
+from corehq.util.log import with_progress_bar
 
 from custom.enikshay.case_utils import get_person_case
 from custom.enikshay.exceptions import ENikshayCaseNotFound
@@ -23,7 +24,7 @@ class Command(BaseCommand):
 
         with open(log_file_name, 'w') as log_file:
             logger = self.get_logger(log_file)
-            for case_id in self.get_case_ids(domain, case_type):
+            for case_id in with_progress_bar(self.get_case_ids(domain, case_type)):
                 if self.should_delete(domain, case_id):
                     self.delete_case(case_id, commit, deletion_id, domain, logger, case_type)
 
--- a/custom/enikshay/management/commands/soft_delete_cases_with_no_person.py
+++ b/custom/enikshay/management/commands/soft_delete_cases_with_no_person.py
@@ -1,0 +1,51 @@
+import csv
+
+from django.core.management import BaseCommand
+
+from corehq.form_processor.interfaces.dbaccessors import CaseAccessors
+
+from custom.enikshay.case_utils import get_person_case
+from custom.enikshay.exceptions import ENikshayCaseNotFound
+
+
+class Command(BaseCommand):
+
+    def add_arguments(self, parser):
+        parser.add_argument('domain')
+        parser.add_argument('case_type')
+        parser.add_argument('log_file_name')
+        parser.add_argument('--commit', action='store_true')
+        parser.add_argument('--deletion_id', dest='deletion_id', default='case_without_person')
+
+    def handle(self, domain, case_type, log_file_name, **options):
+        commit = options['commit']
+        deletion_id = options['deletion_id']
+
+        with open(log_file_name, 'w') as log_file:
+            logger = self.get_logger(log_file)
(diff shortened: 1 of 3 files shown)


custom/enikshay/management/commands/soft_delete_cases_with_no_person.py
```diff
@@ -0,0 +1,51 @@
...
+        try:
+            return not get_person_case(domain, case_id)
+        except ENikshayCaseNotFound:
+            return True
+
+    @staticmethod
+    def delete_case(case_id, commit, deletion_id, domain, logger, case_type):
+        logger.writerow([case_id, deletion_id, case_type])
+        if commit:
+            CaseAccessors(domain).soft_delete_cases([case_id], deletion_id=deletion_id)
```
proteusvacuum: Would you want to do this in bulk potentially? Not sure how many of these there are.

custom/enikshay/management/commands/soft_delete_cases_with_no_person.py
```diff
@@ -0,0 +1,51 @@
...
+        parser.add_argument('--commit', action='store_true')
+        parser.add_argument('--deletion_id', dest='deletion_id', default='case_without_person')
+
+    def handle(self, domain, case_type, log_file_name, **options):
+        commit = options['commit']
+        deletion_id = options['deletion_id']
+
+        with open(log_file_name, 'w') as log_file:
+            logger = self.get_logger(log_file)
+            for case_id in self.get_case_ids(domain, case_type):
```
proteusvacuum: would you want to use a `with_progress_bar` here?

custom/enikshay/management/commands/soft_delete_cases_with_no_person.py
```diff
@@ -0,0 +1,51 @@
...
+    @staticmethod
+    def get_case_ids(domain, case_type):
+        return CaseAccessors(domain).get_case_ids_in_domain(type=case_type)
+
+    @staticmethod
+    def should_delete(domain, case_id):
+        try:
+            return not get_person_case(domain, case_id)
+        except ENikshayCaseNotFound:
+            return True
```
proteusvacuum: FYI this method doesn't include `referral` cases, which I think was part of the requirements. (i.e. it will just delete them all 😨 , since the failure method of `get_person_case` is to throw an `ENikshayCaseNotFound`)

proteusvacuum (commented):
A few questions

mkangia:
👍 thanks for taking this on. Hope these cases get actually removed soon.

custom/enikshay/management/commands/soft_delete_cases_with_no_person.py
nickpell: Would it be ok to throw a different type of error instead of ```ENikshayCaseNotFound``` - something like ```ENikshayCaseTypeNotFound```?

custom/enikshay/management/commands/soft_delete_cases_with_no_person.py
nickpell: Probably at some point - looking to get this right first, optimize later (there's a lot of cases to remove)

custom/enikshay/management/commands/soft_delete_cases_with_no_person.py
nickpell: added

custom/enikshay/management/commands/soft_delete_cases_with_no_person.py
proteusvacuum: Yeah, that makes sense. I think you might need to catch that error here, too:
https://github.com/dimagi/commcare-hq/blob/fa1415cccc56f8a7b7d641cd284fa42cfea5ae1a/custom/enikshay/management/commands/utils.py#L84-L90

proteusvacuum: You might also want to add a method for referral cases if that is something that is required

nickpell: > Yeah, that makes sense. I think you might need to catch that error here, too:

I'm ok failing hard here, since it probably means someone messed up the arguments to a management command, or is trying to access a person from an unsupported case type for ```get_person_case```

nickpell: > You might also want to add a method for referral cases if that is something that is required

Yeah, I'll end up adding referral support and other case types at some point.

proteusvacuum (approved):

PR #18022, opened 2017-09-29
Small refactor

kaapstorm:
Small refactor to not use a temp file (considering we were reading it into memory anyway) and to set the content type to match the content. 

Builds on #18021 (only the last two commits are in this branch).

:blowfish: if you'd like to follow my thought process, although only the last commit is relevant.

### DIFF
--- a/corehq/apps/userreports/reports/view.py
+++ b/corehq/apps/userreports/reports/view.py
@@ -575,9 +575,11 @@ class ConfigurableReport(JSONResponseMixin, BaseDomainView):
         with os.fdopen(fd, 'wb') as temp:
             export_from_tables(self.export_table, temp, Format.HTML)
         with open(path) as f:
-            return HttpResponse(json.dumps({
-                'report': f.read(),
-            }))
+            html = f.read()
+        os.unlink(path)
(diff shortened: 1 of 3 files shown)


corehq/apps/userreports/reports/view.py
```diff
@@ -571,15 +571,11 @@ def export_table(self):
     @property
     @memoized
     def email_response(self):
-        fd, path = tempfile.mkstemp()
-        with os.fdopen(fd, 'wb') as temp:
+        with closing(StringIO()) as temp:
```
gcapalbo: this means it will write the entire file to memory, right? that might be a problem for large files

kaapstorm: That is true, but originally it was reading the entire file into memory with 
```
    return HttpResponse(json.dumps({
        'report': f.read(),
    }))
```
If using StringIO is worse than using `f.read()` then I'm happy to drop the last commit. The second-last commit cleans up old temp files, so I think it's still useful.

gcapalbo: i see, didn't realize it was already doing that

PR #18025, opened 2017-09-29
Add SerialID impelementation

esoergel:
@mkangia
https://docs.google.com/document/d/1x1xz2N7I_kTUjM4dFx31vt9EuzPrOWzSojnWD062XDs/edit?ts=59cbe3bc#


corehq/apps/ota/views.py
```diff
@@ -245,3 +247,14 @@ def heartbeat(request, domain, hq_app_id):
...
 
     return JsonResponse(info)
+
+
+@location_safe
+@login_or_digest_or_basic_or_apikey()
+@require_GET
+def get_next_id(request, domain):
+    bucket_id = request.GET.get('pool_id')
+    session_id = request.GET.get('request_id')
```
ctsims: @esoergel We're going with "Session_id" here, right? I'll update my spec.

corehq/apps/ota/models.py
```diff
@@ -89,3 +90,40 @@ def _delete_restore_blob(self):
...
+    bucket_id = models.CharField(max_length=255)
+    current_value = models.IntegerField(default=-1)
+
+    class Meta:
+        index_together = ('domain', 'bucket_id',)
+        unique_together = ('domain', 'bucket_id',)
+
+    @classmethod
+    def get_next(cls, domain, bucket_id, session_id=None):
+        if session_id is not None:
```
mkangia: Whats the reason for using this instead of just `if session_id:`?

corehq/apps/ota/models.py
```diff
@@ -89,3 +90,40 @@ def _delete_restore_blob(self):
...
+            return cls._get_next_cached(domain, bucket_id, session_id)
+        else:
+            return cls._get_next(domain, bucket_id)
+
+    @classmethod
+    @quickcache(['domain', 'bucket_id', 'session_id'])
```
mkangia: Is there is a way to combine this with `_get_next` and use `quickcache` with `skip_args` instead? 
Absolutely fine if you want to intentionally keep this separate to make it clearer.

mkangia (approved):
Looks good. No blocker just minor queries/suggestions.

corehq/apps/ota/views.py
esoergel: Yup, will update.

corehq/apps/ota/models.py
esoergel: The problem is that it'll still write to the cache if session id isn't provided, which is useless, so this approach just circumvents the cache altogether.  It felt like the clearest way to do that.

corehq/apps/ota/models.py
esoergel: I was thinking of it as being more explicitly - that is, if no argument was passed in for `session_id`, skip the cache.  However, I suppose providing an empty string would lead to more confusing behavior, so I'll switch it.

PR #18030, opened 2017-10-02
async download

calellowitz:
This implements async download for locations and UCRs. It includes a small refactor to increase code reuse.
@gcapalbo


gcapalbo (approved):
does this need qa?

calellowitz:
I tested on staging (all 3 types of export) so I don't think it needs QA.

PR #18037, opened 2017-10-02
replace CreditLine product_type with bool flag is_product

nickpell:
@orangejenny


corehq/apps/accounting/templates/accounting/credits_tab.html
```diff
@@ -33,7 +33,7 @@
                     </td>
                     {% endif %}
                     <td>
-                        {% if credit.product_type == '' or credit.product_type == None and credit.feature_type == None %}
+                        {% if credit.is_product or credit.feature_type == None %}
```
orangejenny: Double checking this condition: it looks like a credit that had product_type blank used to always evaluate to true but can now evaluate to either true or false, depending on if feature_type is present - is that intended?

nickpell: Currently, either credit.product_type == '' or credit.product_type == None.  So, I decided to simplify the boolean expression because if credit.product_type != '', then it must be None, so we don't need to check for it.  After migrating to is_product, then we can just check that credit.is_product is True (used to be credit.product_type == '').

orangejenny (approved):

PR #18038, opened 2017-10-02
Filtered Mobile worker download

sravfeyn:
@emord


corehq/apps/users/templates/users/filter_and_download.html
```diff
@@ -0,0 +1,42 @@
+{% extends 'users/base_template.html' %}
+{% load hq_shared_tags %}
+{% load i18n %}
+{% load crispy_forms_tags %}
+
+{% block js-inline %}{{ block.super }}
+<script>
```
emord: I think that this should be in a separate javascript file and you should use initial_page_data to pass any data to it.

corehq/apps/users/views/mobile/users.py
```diff
@@ -1155,10 +1162,48 @@ def page_url(self):
...
 
 
+class FilteredUserDownload(BaseManageCommCareUserView):
+    urlname = 'filter_and_download_commcare_users'
+    page_title = ugettext_noop('Filter and Download')
+
+    @method_decorator(require_can_edit_commcare_users)
+    def get(self, request, domain, *args, **kwargs):
+        form = CommCareUserFilterForm(request.GET, domain=domain)
+        context = super(FilteredUserDownload, self).main_context
```
emord: I think you could just use `self.main_context` here right?

emord (changes requested):
lgtm once you change the inline js code.

sravfeyn:
@emord updated

emord (approved):

PR #18039, opened 2017-10-02
Compress more JavaScript in app builder

orangejenny:
Noticed we aren't doing a whole lot of compressing. Deployed to staging to verify no compression errors.

`w=1`

@nickpell


nickpell (approved):
nice find

PR #18041, opened 2017-10-02
eNikshay: PMDT Migration Changes

sheelio:
@proteusvacuum 

Bugs that came out of the sanity testing in the project.  Once thisis merged, can you reverse then re-run the migration in sheel-enikshay?

### DIFF
--- a/custom/enikshay/two_b_datamigration/management/commands/import_drtb_cases.py
+++ b/custom/enikshay/two_b_datamigration/management/commands/import_drtb_cases.py
@@ -1743,6 +1743,7 @@ def clean_marital_status(value):
             "married": "married",
             "widowed": "widowed",
             "separated": "separated",
+            "": None,
         }[clean_value]
     except KeyError:
         raise FieldValidationFailure(value, "Marital Status")
--- a/custom/enikshay/two_b_datamigration/management/commands/import_drtb_cases.py
+++ b/custom/enikshay/two_b_datamigration/management/commands/import_drtb_cases.py
@@ -1481,6 +1481,7 @@ def get_follow_up_test_case_properties(column_mapping, row, treatment_initiation
                 "rft_general": "follow_up_drtb",
                 "drug_resistance_list": '',
                 "drug_sensitive_list": '',
+                "result_recorded": "yes",
             }
             properties["rft_drtb_follow_up_treatment_month"] = get_follow_up_month(
                 follow_up, properties['date_tested'], treatment_initiation_date
@@ -1511,6 +1512,7 @@ def get_follow_up_test_case_properties(column_mapping, row, treatment_initiation
                         "result": clean_result(result),
                         "drug_resistance_list": '',
                         "drug_sensitive_list": '',
+                        "result_recorded": "yes",
                     }
                     properties["result_summary_label"] = result_label(properties['result'])
                     properties_list.append(properties)
--- a/custom/enikshay/two_b_datamigration/management/commands/import_drtb_cases.py
+++ b/custom/enikshay/two_b_datamigration/management/commands/import_drtb_cases.py
@@ -936,7 +936,7 @@ def get_disease_site_properties(column_mapping, row):
     value = xlsx_value.replace('EP ', 'extra pulmonary ').lower().strip()
 
     try:
-        classification, site, site_choice = {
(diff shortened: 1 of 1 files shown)


custom/enikshay/two_b_datamigration/management/commands/import_drtb_cases.py
```diff
@@ -1662,6 +1665,20 @@ def clean_socioeconomic_status(value):
...
+    if not value:
+        return None
+    clean_value = value.lower().strip()
+    try:
+        return {
+            "unmarried": "unmarried",
+            "married": "married",
+            "widowed": "widowed",
+            "separated": "separated",
+        }[clean_value]
```
proteusvacuum: why do you need to do this? It looks like a 1-1 relationship. 
You could instead have done: 
```py
if clean_value in ["unmarried", "married", "widowed", "separated"]: 
    return clean_value
else:
    raise FieldValidationFailure(value, "Marital Status")
```

custom/enikshay/two_b_datamigration/management/commands/import_drtb_cases.py
```diff
@@ -1673,6 +1677,53 @@ def clean_socioeconomic_status(value):
...
+    if not value:
+        return None
+    clean_value = value.lower().strip()
+    try:
+        return {
+            "office clerk": "office_clerk",
+            "other craft or related trader or worker": "other_craft_and_related",
+            "corporate manager": "corporate_manager",
+            "legislators or senior official": "legislators_or_senior_official",
+            "legistrator or senior official": "legislators_or_senior_official",
```
proteusvacuum: 🤦‍♀️

custom/enikshay/two_b_datamigration/management/commands/import_drtb_cases.py
```diff
@@ -1079,15 +1079,19 @@ def get_mehsana_resistance_properties(column_mapping, row):
...
+    # value = "{}: MDR/RR".format(episode_treatment_initiation_date)
+    # if put_on_treatment:
+    #    value += "\n{}: {}".format(
+    #        put_on_treatment,
+    #        column_mapping.get_value("type_of_treatment_initiated", row)
+    #    )
+    #return value
+    current_treatment_regimen = column_mapping.get_value("treatment_regimen", row)
+    if current_treatment_regimen:
+        return ("{}: " + current_treatment_regimen).format(episode_treatment_initiation_date)
```
proteusvacuum: This will work, but is a bit confusing. A clearer way of writing it would have been:
```py
"{}: {}".format(episode_treatment_initiation_date, current_treatment_regimen)
```

proteusvacuum (approved):
Couple of nit-picky comments, but I think everything looks good.

sheelio:
@proteusvacuum 
Sorry, had to refactor one other area of the code.

custom/enikshay/two_b_datamigration/management/commands/import_drtb_cases.py
```diff
@@ -571,8 +571,16 @@ def get_case_structures_from_row(commit, domain, migration_id, column_mapping, c
     )
 
     # Close the occurrence if we have a treatment outcome recorded
-    close_occurrence = ("treatment_outcome" in episode_case_properties and
-                        episode_case_properties["treatment_outcome"])
+    if (
+        "treatment_outcome" in episode_case_properties
+        and episode_case_properties["treatment_outcome"]
+    ):
+        close_occurrence = True
```
proteusvacuum: do you need to add a close reason or something?. Also, do you need to close the episode?

sheelio: The app doesn't add a close reason to the occurrence for this.  

I was assuming that since episode is an extension, it would get auto-closed as well.  Is that not true?

proteusvacuum: it looks like you are opening these cases at the same time as closing them, is that true? 

If so, you might be creating the closed occurrence before you create the episode case, in which case the episode case will remain open.

sheelio: Ok, I'll manually go through and close out the other cases then too.

PR #18042, opened 2017-10-02
Decrease blob db access on restore

millerdev:
Accumulating response content in a temporary file is now managed by a separate class than serving that content in the response. Response content is no longer stored in the blob db if the content is not cached. This results in less blob db access.

Same as before:
- accumulate response elements in a temp file.
- put elements in second temp file wrapped in OpenRosaResponse element and delete first temp file.

Before:
- send content to blob db.
- delete second temp file.
- stream content from blob db (response).

After:
- if caching: send content to blob db.
- serve content (response).
- delete second temp file.

Best reviewed by commit 🐠 

@snopoke @proteusvacuum

### DIFF
--- a/corehq/ex-submodules/casexml/apps/phone/restore.py
+++ b/corehq/ex-submodules/casexml/apps/phone/restore.py
@@ -510,7 +510,7 @@ class RestoreConfig(object):
             case_sync=case_sync,
         )
 
-        self.force_cache = self.cache_settings.force_cache or self.async
+        self.force_cache = self.cache_settings.force_cache
         self.cache_timeout = self.cache_settings.cache_timeout
         self.overwrite_cache = self.cache_settings.overwrite_cache
 
--- a/corehq/ex-submodules/casexml/apps/phone/restore.py
+++ b/corehq/ex-submodules/casexml/apps/phone/restore.py
@@ -100,7 +100,7 @@ class StockSettings(object):
         self.sync_consumption_ledger = sync_consumption_ledger
 
 
-class RestoreResponse(object):
+class RestoreContent(object):
     start_tag_template = (
         b'<OpenRosaResponse xmlns="http://openrosa.org/http/response"%(items)s>'
         b'<message nature="%(nature)s">Successfully restored account %(username)s!</message>'
@@ -118,10 +118,7 @@ class RestoreResponse(object):
         return self
 
     def __exit__(self, exc_type, exc_val, exc_tb):
-        try:
-            self._finalize()
-        finally:
-            self.response_body.close()
+        self.response_body.close()
 
     def append(self, xml_element):
         self.num_items += 1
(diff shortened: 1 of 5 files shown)


corehq/ex-submodules/casexml/apps/phone/restore.py
```diff
@@ -706,31 +666,30 @@ def _get_task_timeout(self, new_task):
...
+    def set_cached_payload_if_necessary(self, fileobj, duration, async):
         # on initial sync, only cache if the duration was longer than the threshold
         is_long_restore = duration > timedelta(seconds=INITIAL_SYNC_CACHE_THRESHOLD)
 
-        if self.force_cache or is_long_restore or self.sync_log:
-            self._set_cache_in_redis(cache_payload_path)
-
-    def _set_cache_in_redis(self, cache_payload_path):
-        self.restore_payload_path_cache.set_value(cache_payload_path, self.cache_timeout)
+        if async or self.force_cache or is_long_restore or self.sync_log:
```
proteusvacuum: I think `self.force_cache` is currently always set to `True` if `self.async = True`. 

https://github.com/dimagi/commcare-hq/blob/f9d71d0c13857ea5342ca7343a359859d9a775e6/corehq/ex-submodules/casexml/apps/phone/restore.py#L513

Was that not getting set correctly when the task is running?

I think if you want to add the async check explictly here, it would be good to remove it from the force_cache, otherwise it might get a little confusing.

proteusvacuum (commented):
LGTM! Much clearer and happy you were able to remove one write / read cycle.

corehq/ex-submodules/casexml/apps/phone/restore.py
```diff
@@ -167,6 +167,7 @@ def as_file(self):
     def as_string(self):
         """Get content as utf8-encoded bytes
 
+        NOTE: This method is only used in tests.
```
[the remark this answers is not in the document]
millerdev: Maybe. I don't really like that decorator—test code should live in test modules, not in production source modules. I would have removed this method, except that there are quite a few tests that depend on it, and I couldn't think of an easy way to move this code into a test module. Maybe I'll take a stab at that in another PR.

corehq/ex-submodules/casexml/apps/phone/restore.py
millerdev: > Was that not getting set correctly when the task is running?

No, I just didn't realize the two were related.

> I think if you want to add the async check explictly here, it would be good to remove it from the force_cache, otherwise it might get a little confusing.

Good point. dc91b42c436b483a930eae005d1d4bc0f5aefb7a

millerdev:
✅

proteusvacuum (approved):

snopoke (commented):
Sorry for the slow response. This is a nice cleanup and separation of the pieces.

PR #18047, opened 2017-10-02
catch nginx timeout in preindex

dannyroberts:
for couchdb2, where we're proxying with nginx

### DIFF
--- a/corehq/ex-submodules/dimagi/utils/couch/sync_docs.py
+++ b/corehq/ex-submodules/dimagi/utils/couch/sync_docs.py
@@ -35,7 +35,7 @@ def index_design_docs(db, docid, design_name):
             try:
                 list(db.view(view, limit=0))
             except RequestFailed as e:
-                if 'timeout' not in e.message:
+                if 'timeout' not in e.message and 'Time-out' not in e.message:
                     raise
             else:
                 break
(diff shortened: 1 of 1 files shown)


corehq/ex-submodules/dimagi/utils/couch/sync_docs.py
```diff
@@ -35,7 +35,7 @@ def index_design_docs(db, docid, design_name):
             try:
                 list(db.view(view, limit=0))
             except RequestFailed as e:
-                if 'timeout' not in e.message:
```
emord: is it possible to check if the status code is 504?

kaapstorm (approved):

corehq/ex-submodules/dimagi/utils/couch/sync_docs.py
kaapstorm: You could try checking `e.status_int` ... but I'm not sure whether it'll be set when the exception is raised. 

(I also noticed that a RequestTimeout exception exists. Interesting that it doesn't get raised. It extends `Exception` and doesn't have a `status_int` property.)

emord (approved):

kaapstorm (approved):

PR #18048, opened 2017-10-02
Don't specify form id

emord:
This doesn't specify the form id and it will take the user to the case select screen where they can select "report a new issue"

It looks like the web apps URL includes a `steps": [0, "action 0"]`. @dannyroberts or @wspride is that "action" a new addition to the URL. it's possible I missed that before, but I thought it was supposed to be [0, 0] ?

buddy @sravfeyn


emord:
new buddy @millerdev

PR #18050, opened 2017-10-02
Make sure to update device for enikshay users

esoergel:
https://manage.dimagi.com/default.asp?262962
https://manage.dimagi.com/default.asp?261797
eNikshay relies on this signal being called during restore.
@mkangia

### DIFF
--- a/corehq/apps/ota/utils.py
+++ b/corehq/apps/ota/utils.py
@@ -3,6 +3,7 @@ from casexml.apps.case.xml import V2
 from casexml.apps.phone.restore import RestoreConfig, RestoreParams
 from couchdbkit import ResourceConflict
 
+from corehq import toggles
 from corehq.apps.domain.models import Domain
 from corehq.apps.users.util import format_username
 from corehq.apps.users.models import CommCareUser
@@ -212,4 +213,4 @@ def update_device_id(user, device_id):
         if not user.is_demo_user:
             updated = user.update_device_id_last_used(device_id)
             if updated:
-                user.save(fire_signals=False)
+                user.save(fire_signals=toggles.ENIKSHAY.enabled(user.domain))
--- a/custom/enikshay/tests/two_b_datamigration/test_logging.py
+++ b/custom/enikshay/tests/two_b_datamigration/test_logging.py
@@ -48,6 +48,7 @@ class ImportDRTBTestMixin(object):
         get_users_path = \
             "custom.enikshay.two_b_datamigration.management.commands.import_drtb_cases.get_users_by_location_id"
         mock_user = MagicMock(
+            domain=self.domain,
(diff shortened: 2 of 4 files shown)


custom/enikshay/user_setup.py
```diff
@@ -155,7 +155,7 @@ def save_user_callback(sender, couch_user, **kwargs):
         changed = set_issuer_id(commcare_user.domain, commcare_user) or changed
         changed = add_drtb_hiv_to_dto(commcare_user.domain, commcare_user) or changed
         if changed:
-            commcare_user.save()
+            commcare_user.save(fire_signals=False)
```
mkangia: Dont know much about how this would effect but do we need any signals that should be triggered?

mkangia (approved):
If its safe to avoid any/all signals being fired then safe to merge.

custom/enikshay/tests/two_b_datamigration/test_logging.py
```diff
@@ -48,6 +48,7 @@ def drtb_import(self, import_rows, format, commit=False):
         get_users_path = \
             "custom.enikshay.two_b_datamigration.management.commands.import_drtb_cases.get_users_by_location_id"
         mock_user = MagicMock(
+            domain=self.domain,
```
proteusvacuum: did you mean to add this in this PR?

custom/enikshay/user_setup.py
proteusvacuum: I think this will stop it recursively calling this method again, as this is the one called by the signal. 
👍 I think this could have been a better fix to the "device id keeps updating and getting sent to BETS" issue that I fixed rather convolutedly before.

corehq/apps/ota/utils.py
```diff
@@ -212,4 +213,4 @@ def update_device_id(user, device_id):
         if not user.is_demo_user:
             updated = user.update_device_id_last_used(device_id)
             if updated:
-                user.save(fire_signals=False)
+                user.save(fire_signals=toggles.ENIKSHAY.enabled(user.domain))
```
proteusvacuum: will this toggle get queried every time any user syncs now? (I know `enikshay` is an "always on" domain for this, but on other envs / projects).

esoergel: That's true - it will get called for every sync where the device id isn't the same as the previous one for that user.  I've generally been treating `toggles.X.enabled` as a ~free lookup, but if that's not the case, it should be.

custom/enikshay/tests/two_b_datamigration/test_logging.py
esoergel: Yeah, this broke tests - `toggles.ENIKSHAY.enabled(user.domain)` fails if `user` is a `MagicMock`, since `user.domain` is expected to be a string.  That was not easy to track down :/

PR #18060, opened 2017-10-03
Filter data sources server-side

kaapstorm:
Context: [FB 262457](https://manage.dimagi.com/default.asp?262457)

The user is looking for a case-sharing group. HQ is returning 10 results, but case-sharing group is (say) the 11th result. The data sources are being filtered client-side, so the case-sharing group the user is looking for is not returned.

If data sources are filtered server-side, what was result 11 will now appear in the top 10, and visible client-side.

@esoergel buddy @calellowitz

### DIFF
--- a/corehq/apps/data_interfaces/templates/data_interfaces/interfaces/case_management.html
+++ b/corehq/apps/data_interfaces/templates/data_interfaces/interfaces/case_management.html
@@ -58,7 +58,8 @@
                 url: '{% url "case_list_options" domain %}',
                 data: function (term, page) {
                     return {
-                        q: term
+                        q: term,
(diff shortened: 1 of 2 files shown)


esoergel:
The `EmwfOptionsView` has several subclasses (looks like 6 versions of that floating around now!), which generally override the `data_sources` property as a way to restrict which types of results are sent to the client.  This introduces a second way to restrict those results.  My preference would be for there to be only one way to do that.  Have you looked into making a subclass and pointing this filter to that URL instead?

kaapstorm:
Thanks @esoergel that's a very good call. I'll do that.

PR #18071, opened 2017-10-04
Move analytics document onready logic into main.js

orangejenny:
Bandaid to allow RequireJS to work without modularizing all of the analytics code right now.

Verified that prelogin and formplayer, though they include `analytics_all.html`, don't use `.track-usage-link`.

@millerdev / @proteusvacuum


millerdev (approved):
Will you be monitoring if this works after it's deployed? How will we know if it is broken or breaks something?

proteusvacuum (approved):

orangejenny:
@millerdev I test analytics changes locally but don't monitor them in prod. I don't think we have a good system for knowing when analytics are broken in production (see https://github.com/dimagi/commcare-hq/pull/18070 ...) and this is indeed a concern for the growth team in particular. I think @calellowitz has thought a bit about the problem but don't know that we have plans to deal with it any time soon.

PR #18075, opened 2017-10-04
ReassignCaseOptions only returns potential case owners

kaapstorm:
Context: [FB 262457](https://manage.dimagi.com/default.asp?262457), #18060 

(A definite improvement on previous PR. Thanks @esoergel )


esoergel (approved):
Nice, thanks!

PR #18076, opened 2017-10-04
[wip] Health Establishment API fixes

mkangia:
Update Health Establishment Notification:
1. Update to send only `pcp` and `plc`
2. Send only private sector locations
3. Send dummy values where missing
4. Update to fetch values for location user where needed
5. Do not trigger when there is no location user

buddy @kaapstorm


mkangia:
This is waiting on quite some locations fixes so closing this for now. Will revive it when needed.

kaapstorm:
Don't have context, but lgtm.

PR #18077, opened 2017-10-04
Revert "Decrease blob db access on restore"

snopoke:
Potentially causing issues with restore.
Reverts dimagi/commcare-hq#18042


snopoke:
@millerdev FYI

millerdev:
What kind of issues? I couldn't find anything obvious in sentry. Is this bug related: https://manage.dimagi.com/default.asp?263058 Are there any others?

calellowitz:
That is the right ticket but the `|` issue was a red herring. The PR caused all restores from web apps and the phone (but not the admin restore page it seems) on all environments it was deployed to to result in a 500. What is still a mystery is where that 500 was. It showed up in the nginx logs, but appears to have happened before it reached a django process, because there is no evidence in the django logs or sentry of a python exception.

PR #18078, opened 2017-10-04
Revert "Revert "Make sure to update device for enikshay users""

calellowitz:
Reverts dimagi/commcare-hq#18073
@snopoke

### DIFF
--- a/corehq/apps/ota/utils.py
+++ b/corehq/apps/ota/utils.py
@@ -3,6 +3,7 @@ from casexml.apps.case.xml import V2
 from casexml.apps.phone.restore import RestoreConfig, RestoreParams
 from couchdbkit import ResourceConflict
 
+from corehq import toggles
 from corehq.apps.domain.models import Domain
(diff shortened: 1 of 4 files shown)


corehq/apps/ota/utils.py
```diff
@@ -212,4 +213,4 @@ def update_device_id(user, device_id):
         if not user.is_demo_user:
             updated = user.update_device_id_last_used(device_id)
             if updated:
-                user.save(fire_signals=False)
+                user.save(fire_signals=toggles.ENIKSHAY.enabled(user.domain))
```
emord: @esoergel I think you did this original PR. Is there any reason to not use `settings.SERVER_ENVIRONMENT == 'enikshay'` That would save a couch/redis request everywhere

esoergel: That seems reasonable - I think we could also make this call the `device_id` update thing directly and not handle the other user signals.  Lemme take a crack at that.

PR #18079, opened 2017-10-04
add fast_index_couch_views

dannyroberts:
(replacement for prime_views)


emord (approved):
Does this fix the issues you were seeing yesterday?

dannyroberts:
@emord it circumvents them by not messing with -tmp views at all. I ran this in a loop overnight and it seems to be working without crashing.

dannyroberts:
By the way @emord this is the task you mentioned: https://github.com/dimagi/commcare-hq/blob/f9caf06ae9247a2a1312c635fdbcf57cad501ba5/corehq/preindex/tasks.py#L9, and it does what you said it does. Feeling better about deploying knowing that this exists and is run regularly in all production instances.

PR #18080, opened 2017-10-04
Added month filter to all views that use month aggregated data

kkrampa:
@emord


custom/icds_reports/static/js/directives/adhaar-beneficiary/adhaar-beneficiary.directive.js
```diff
@@ -24,7 +24,7 @@ function AdhaarController($scope, $routeParams, $location, $filter, demographics
     vm.bottom_five = [];
     vm.location_type = null;
     vm.loaded = false;
-    vm.filters = ['month', 'age', 'gender'];
+    vm.filters = ['age', 'gender'];
```
kkrampa: @emord 
I don't know why this variable is called ```vm.filters``` but it contains filters that are excluded from the page. We should rename it to ```vm.excludedFilters``` but I don't want to broke something by doing refactor at this point.

emord (approved):

PR #18082, opened 2017-10-04
PredictablyRandomToggle should check namespace

kaapstorm:
Previously `PredictablyRandomToggle.enabled()` ignored namespaces, and they were only checked in the parent method if `deterministic_random(self._get_identifier(item)) < self.randomness`. This meant that a user-only toggle could be enabled for a domain.

More context: [FB 262625](https://manage.dimagi.com/default.asp?262625)

@millerdev, buddy @calellowitz

### DIFF
--- a/corehq/ex-submodules/toggle/tests.py
+++ b/corehq/ex-submodules/toggle/tests.py
@@ -2,8 +2,16 @@ import uuid
 
 from couchdbkit import ResourceConflict
 from couchdbkit.exceptions import ResourceNotFound
-from django.test import TestCase
+from django.conf import settings
+from django.test import TestCase, SimpleTestCase
 
+from corehq.toggles import (
+    NAMESPACE_USER,
+    NAMESPACE_DOMAIN,
+    TAG_EXPERIMENTAL,
+    PredictablyRandomToggle,
+    deterministic_random,
+)
 from toggle.shortcuts import update_toggle_cache, namespaced_item, clear_toggle_cache
 from .models import generate_toggle_id, Toggle
 from .shortcuts import toggle_enabled, set_toggle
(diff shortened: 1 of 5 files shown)


kaapstorm:
:recycle: :construction_worker_woman: (npm failed to install packages)

kaapstorm:
:white_check_mark:

corehq/toggles.py
```diff
@@ -220,14 +221,18 @@ def randomness_percent(self):
     def _get_identifier(self, item):
         return '{}:{}:{}'.format(self.namespaces, self.slug, item)
 
-    def enabled(self, item, **kwargs):
+    def enabled(self, item, namespace=Ellipsis):
+        if namespace is Ellipsis:
+            logging.warning('PredictablyRandomToggle.enabled() may be invalid if namespace is not set')
```
millerdev: I think most violations of this will never be noticed. Consider raising an exception instead.

corehq/toggles.py
```diff
@@ -220,14 +221,18 @@ def randomness_percent(self):
     def _get_identifier(self, item):
         return '{}:{}:{}'.format(self.namespaces, self.slug, item)
 
-    def enabled(self, item, **kwargs):
+    def enabled(self, item, namespace=Ellipsis):
```
millerdev: What do you think of changing the signature to `toggle.enabled("icds", domain=True)` or `toggle.enabled("dmiller@dimagi.com", user=True)` rather than requiring a `namespace` constant, which must be imported anywhere `toggle.enabled(...)` is called? We could require that the `domain` parameter is always specified as a keyword argument (using `**kwargs`) so we don't end up with ambiguous calls like `toggle.enabled("icds", True)`. If neither `domain` nor `user` is passed, then it would fall back to the existing behavior of checking if the flag is enabled for a (properly namespaced) user or domain with the given value.

This feedback can be ignored until later since it is less critical than the other comments on this PR.

corehq/toggles.py
```diff
@@ -220,14 +221,18 @@ def randomness_percent(self):
...
+        if namespace is Ellipsis:
+            logging.warning('PredictablyRandomToggle.enabled() may be invalid if namespace is not set')
         if settings.UNIT_TESTING:
             return False
         elif item in self.always_disabled:
             return False
+        elif namespace is not Ellipsis and namespace not in self.namespaces:
+            return False
         return (
             (item and deterministic_random(self._get_identifier(item)) < self.randomness)
```
millerdev: I think the real problem is on this line, not the one below it. If this line evaluates to true the `super.enabled(...)` call will not be evaluated. I think we cannot evaluate the randomness check here unless the toggle applies to both users and domains or the namespace parameter is specified.

millerdev (changes requested):
Should have requested changes on my previous review.

corehq/toggles.py
kaapstorm: I really like this idea. I agree that its less pressing than the other changes, but let's not let it slide.

corehq/toggles.py
kaapstorm: This class currently only gets used in one place at the moment, so I think that's a good suggestion. (And if throwing exceptions is potentially a problem, a soft assert might be a reasonable compromise.)

corehq/toggles.py
```diff
@@ -220,14 +221,18 @@ def randomness_percent(self):
...
 
-    def enabled(self, item, **kwargs):
+    def enabled(self, item, namespace=Ellipsis):
+        if namespace is Ellipsis:
+            logging.warning('PredictablyRandomToggle.enabled() may be invalid if namespace is not set')
         if settings.UNIT_TESTING:
             return False
         elif item in self.always_disabled:
             return False
+        elif namespace is not Ellipsis and namespace not in self.namespaces:
```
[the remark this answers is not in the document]
kaapstorm: Interesting :upside_down_face:

corehq/toggles.py
kaapstorm: Agreed.

corehq/toggles.py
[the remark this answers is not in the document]
kaapstorm: ... does that mean that [`super().enable()` will fail](https://github.com/dimagi/commcare-hq/blob/fc0e8c3450f8d9d9b5ef61cb13344229940877c8/corehq/toggles.py#L81) if passed `NAMESPACE_USER`?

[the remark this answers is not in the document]
millerdev: That appears to be correct.

corehq/toggles.py
```diff
@@ -220,14 +220,29 @@ def randomness_percent(self):
     def _get_identifier(self, item):
         return '{}:{}:{}'.format(self.namespaces, self.slug, item)
 
-    def enabled(self, item, **kwargs):
+    def enabled(self, item, namespace=Ellipsis):
+        if namespace == NAMESPACE_USER:
+            namespace = None  # because:
+            # StaticToggle.__init__(): self.namespaces = [None if n == NAMESPACE_USER else n for n in namespaces]
+
+        if set(self.namespaces) != set(ALL_NAMESPACES) and namespace is Ellipsis:
```
millerdev: nitpick: it would be more efficient to reverse these conditions so the sets are only created/compared if `namespace is Ellipsis`.

millerdev (approved):
Would be nice to have some tests that show how this failed before and that it now works as expected. I'll let it up to you if you want to implement them in this PR or not.

corehq/ex-submodules/toggle/tests.py
```diff
@@ -120,3 +128,118 @@ def test_toggle_cache(self):
...
+    @classmethod
+    def setUpClass(cls):
+        cls.user_toggle = Toggle(
+            slug='user_toggle',
+            enabled_users=['arthur', 'diane'])
+        cls.user_toggle.save()
+        cls.domain_toggle = Toggle(
+            slug='domain_toggle',
+            enabled_users=[namespaced_item('dc', NAMESPACE_DOMAIN)])
+        cls.domain_toggle.save()
```
millerdev: I don't see these toggles being referenced anywhere. Are they needed?

millerdev (approved):
Tests! Yay!! 🎉

corehq/ex-submodules/toggle/tests.py
kaapstorm: Well spotted.

:roll_eyes: Urgh. The toggles are not enabled for "jessica" and "marvel", not because they aren't members of `user_toggle` and `domain_toggle` (as this is supposed to test), but because "test_toggle" is not actually the slug of a persisted `Toggle` instance. ... i.e. They *should* be needed and the tests are passing for the wrong reason.

millerdev (approved):

PR #18084, opened 2017-10-04
use progress bar

nickpell:
Slightly less efficient because it gets cases individually rather than in bulk, but will save my time by keeping track of how far along the migrations are.  Also, fetching cases is a small amount of the runtime, so I don't expect much of a difference.

@snopoke

### DIFF
--- a/custom/enikshay/management/commands/utils.py
+++ b/custom/enikshay/management/commands/utils.py
@@ -9,6 +9,7 @@ from casexml.apps.case.mock import CaseFactory
 from casexml.apps.case.xform import get_case_updates
 
 from corehq.form_processor.interfaces.dbaccessors import CaseAccessors
+from corehq.util.log import with_progress_bar
 
 from custom.enikshay.case_utils import (
     get_person_case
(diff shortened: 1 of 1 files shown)


custom/enikshay/management/commands/utils.py
```diff
@@ -80,9 +85,9 @@ def handle(self, domain, log_file_name, case_ids, **options):
 
             print("Finished at {}".format(datetime.datetime.utcnow()))
 
-    def is_valid_case(self, domain, case):
+    def is_valid_case(self, domain, case_id):
         try:
-            person_case_id = get_person_case(domain, case.case_id)
+            person_case_id = get_person_case(domain, case_id)
```
mkangia: this method should be named `get_person_case_id`.

mkangia (approved):
looks good. Nothing blocking.

custom/enikshay/management/commands/utils.py
nickpell: good observation

PR #18087, opened 2017-10-04
Beneficiaries always use migrated property

emord:
https://manage.dimagi.com/default.asp?262928

Changes a few indicators to look for migration status / resident. There will be a rebuild necessary for this. I'm going to write a custom thing later that will only recalcualte ones where seeking services is false (seeking services implies that the person is a resident, so these columns will already be correct)

@sheelio buddy @millerdev

### DIFF
--- a/custom/icds/management/commands/rebuild_for_migration.py
+++ b/custom/icds/management/commands/rebuild_for_migration.py
@@ -37,18 +37,30 @@ class Command(BaseCommand):
             assert is_static
             adapter = get_indicator_adapter(data_source)
             table = adapter.get_table()
-            for case_id in self._get_case_ids_to_process(adapter, table):
+            for case_id in self._get_case_ids_to_process(adapter, table, data_source_id):
                 change = FakeChange(case_id, fake_change_doc)
-                AsyncIndicator.update_indicators(change, [data_source])
+                AsyncIndicator.update_indicators(change, [data_source_id])
 
-    def _get_case_ids_to_process(self, adapter, table):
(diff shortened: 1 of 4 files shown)


millerdev (approved):
Don't feel like I have a lot of context here, so I'll let you merge when you're ready.

sheelio:
@emord 
Was this a request from the GS team that we should also remove migrated beneficiaries from counts of "all beneficiaries"?

(Is it cleaner to add a new set of columns / indicators for this instead of modifying the existing definition?

emord:
@sheelio let me know what you think of hte mgmt command i just posted. hoping to kick it off tomorrow

custom/icds/management/commands/rebuild_for_migration.py
```diff
@@ -0,0 +1,70 @@
...
+            adapter = get_indicator_adapter(data_source)
+            table = adapter.get_table()
+            for case_id in self._get_case_ids_to_process(adapter, table, data_source_id):
+                change = FakeChange(case_id, fake_change_doc)
+                AsyncIndicator.update_indicators(change, [data_source_id])
+
+    def _add_filters(self, query, table, data_source_id):
+        if data_source_id == 'static-icds-cas-static-child_cases_monthly_tableau_v2':
+            return query.filter(
+                table.columns.valid_in_month_all = 1,
```
sheelio: I realize the column name for this is actually
valid_all_registered_in_month

emord:
only changes since approval have been to the the management command, which i offlined with sheel. merging

PR #18089, opened 2017-10-04
Repeat Record report

esoergel:
First commit turns repeat record counts into links to the filtered report for that state.
Second commit adds the option to search by payload ID in the report.  That's to make it easier to track down whether a particular entity has been sent.
@gcapalbo
@dannyroberts @proteusvacuum FYI

### DIFF
--- a/corehq/apps/domain/templates/domain/admin/repeater_row.html
+++ b/corehq/apps/domain/templates/domain/admin/repeater_row.html
@@ -1,10 +1,26 @@
 {% load i18n %}
 <tr>
     <td>{{ repeater.url }}</td>
-    <td>{{ repeater.get_pending_record_count }}</td>
-    <td>{{ repeater.get_failure_record_count }}</td>
-    <td>{{ repeater.get_cancelled_record_count }}</td>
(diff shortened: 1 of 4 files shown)


corehq/apps/domain/templates/domain/admin/repeater_row.html
```diff
@@ -1,10 +1,26 @@
 {% load i18n %}
 <tr>
     <td>{{ repeater.url }}</td>
-    <td>{{ repeater.get_pending_record_count }}</td>
-    <td>{{ repeater.get_failure_record_count }}</td>
-    <td>{{ repeater.get_cancelled_record_count }}</td>
-    <td>{{ repeater.get_success_record_count }}</td>
+    <td>
+        <a href="{% url 'domain_report_dispatcher' domain 'repeat_record_report' %}?repeater={{ repeater.get_id }}&record_state=PENDING">
```
gcapalbo: this will probably work but technically i think `&` should be `&amp;`

gcapalbo (approved):

proteusvacuum (approved):
Yay! Have been meaning to do this for a while.

PR #18100, opened 2017-10-05
Only check toggle if it's on the enikshay env

esoergel:
from https://github.com/dimagi/commcare-hq/pull/18078#discussion_r142657984
@emord

### DIFF
--- a/corehq/toggles.py
+++ b/corehq/toggles.py
@@ -53,7 +53,7 @@ class StaticToggle(object):
     def __init__(self, slug, label, tag, namespaces=None, help_link=None,
                  description=None, save_fn=None, always_enabled=None,
                  always_disabled=None, enabled_for_new_domains_after=None,
-                 enabled_for_new_users_after=None):
+                 enabled_for_new_users_after=None, relevant_environments=None):
         self.slug = slug
         self.label = label
         self.tag = tag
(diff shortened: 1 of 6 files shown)


emord (approved):
This seems ok, but what do you think about adding a parameter to toggles called `enabled_environments` or something similar and having a static check here: https://github.com/dimagi/commcare-hq/blob/master/corehq/toggles.py#L75

That will solve the issue everywhere and gives us an easy way to only enable certain flags on certain server environments

esoergel:
@emord updated.  Lemme know what you think.  I'm sure there are other feature flags that will benefit from this param, but I'd like to start with this and add the others after.

corehq/toggles.py
```diff
@@ -67,12 +67,20 @@ def __init__(self, slug, label, tag, namespaces=None, help_link=None,
         self.always_disabled = always_disabled or set()
         self.enabled_for_new_domains_after = enabled_for_new_domains_after
         self.enabled_for_new_users_after = enabled_for_new_users_after
+        # pass in a set of environments where this toggle applies
+        self.relevant_environments = relevant_environments
```
emord: you could initialize this as `relevant_environments or []` and get rid of the null check later

emord (approved):

corehq/toggles.py
esoergel: Actually I need the null check - if it's not specified (or empty), then we need to treat it as if all environments are listed.

PR #18102, opened 2017-10-05
add run_result_summary_display_migration

nickpell:
@sheelio Could you check that I've implemented this logic right?

cc: @snopoke


snopoke (approved):
Pending logic review from @sheelio

custom/enikshay/management/commands/run_result_summary_display_migration.py
```diff
@@ -0,0 +1,37 @@
...
+                test.get_case_property('result_recorded')
+                and not test.get_case_property(RESULT_SUMMARY_DISPLAY)
+                and not test.get_case_property('test_type_value')
+            )
+        ):
+            return {}
+
+        return {
+            RESULT_SUMMARY_DISPLAY: '\n'.join(filter(None, [
+                'TB Detected' if test.get_case_property('result') == 'tb_detected' else 'TB Not Detected',
```
sheelio: @nickpell 
I think we want to actually have it be:
if test.get_case_property('result') == 'tb_detected':
    value = 'TB Detected'
elif test.get_case_property('result') == 'tb_not_detected':
    value = 'TB Not Detected'
else:
    value = None

sheelio:
One comment, otherwise looks good.

nickpell:
@sheelio updated

custom/enikshay/management/commands/run_result_summary_display_migration.py
```diff
@@ -0,0 +1,40 @@
...
+    include_public_cases = True
+    include_private_cases = False
+
+    @staticmethod
+    def get_case_property_updates(test, domain):
+        if (
+            test.get_case_property(DATAMIGRATION_CASE_PROPERTY) == 'yes'
+            or not (
+                test.get_case_property('result_recorded') == 'yes'
+                and not test.get_case_property(RESULT_SUMMARY_DISPLAY)
```
sheelio: I looked at the sample output and it doesn't appear very useful. 

I'm wondering if we should switch up the logic so that:
- For all tests where result_recorded = yes and test_type_value != ''/null
- We calculate the new result_summary_display
- If its different than the current result_summary_display, we update the test with the new result_summary_display

sheelio (changes requested):

custom/enikshay/management/commands/run_result_summary_display_migration.py
```diff
@@ -0,0 +1,40 @@
...
+            )
+        ):
+            return {}
+
+        return {
+            RESULT_SUMMARY_DISPLAY: '\n'.join(filter(None, [
+                {
+                    'tb_detected': 'TB Detected',
+                    'tb_not_detected': 'TB Not Detected',
+                }.get(test.get_case_property('result')),
```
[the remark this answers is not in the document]
nickpell: It will return ```None```, which then will be filtered out

custom/enikshay/management/commands/run_result_summary_display_migration.py
nickpell: updated, doing a dry run now

sheelio (approved):

custom/enikshay/management/commands/run_result_summary_display_migration.py
```diff
@@ -0,0 +1,43 @@
...
+                and not test.get_case_property('test_type_value')
+            )
+        ):
+            return {}
+
+        new_result_display_summary = '\n'.join(filter(None, [
+            {
+                'tb_detected': 'TB Detected',
+                'tb_not_detected': 'TB Not Detected',
+            }.get(test.get_case_property('result')),
```
sheelio: Lets add add a line for test.result_grade (below tb detected / not detected).  Here's the mapping:

negative_not_seen: None/blank
1plus: Result Grade: 1+
2plus: Result Grade: 2+
3plus: Result Grade: 3+
error_invalid_result: Result Grade: Error Invalid Result
0: None/blank
n/a: None/blank
1+: Result Grade: 1+
2+: Result Grade: 2+
3+: Result Grade: 3+

nickpell: There are a lot of different values for ```result_grade``` - sent you a case export on Skype.

How should I handle these other values?

sheelio: @nickpell 

Those might be for private cases?  Is this case export all test cases? I did a review of result_grade for valid private sector cases and I think this is all I got?

sheelio: For those that don't match in our mapping list, can we just use
Result Grade: <test.result_grade>

nickpell: updated

PR #18107, opened 2017-10-05
Rebuild form table command

sheelio:
@emord 
I think this should let you add to a form table over a particular date range (Currently rebuilding table isn't rebuilding for large projects).


corehq/apps/userreports/management/commands/rebuild_form_table.py
```diff
@@ -0,0 +1,33 @@
...
+
+        time_filter = form_es.submitted
+        query = (form_es.FormES()
+                 .domain(domain)
+                 .filter(time_filter(gte=start_date, lte=end_date)))
+        if xmlns_list:
+            query = query.OR(*[form_es.xmlns(x) for x in xmlns_list])
+
+        form_ids = query.get_ids()
+        rebuild_indicators_in_place(indicator_config_id, None, form_ids)
```
emord: This is going to run in the foreground. if you want it in the background, you can use `rebuild_indicators_in_place.delay(args)`

also that would pass what could potentially be a large list of ids to celery.  I think that's fine if it's going to be something small, but if it's in the thousands you might want to create an intermediary class that provides iteration

snopoke (commented):
Feels like we should build a full reindexer for these large tables that use the same setup as our other reindexers.

emord:
@snopoke do you mean like our pillowtop reindexers? That does seem like a pretty good idea. although we don't currently have an accessor to filter by xmlns (which is why this sort of functionality hasn't been written for forms yet)

PR #18109, opened 2017-10-05
Revert "Revert "Decrease blob db access on restore""

millerdev:
Reverts dimagi/commcare-hq#18077

Going to test this on staging before moving forward.

### DIFF
--- a/corehq/ex-submodules/casexml/apps/phone/restore.py
+++ b/corehq/ex-submodules/casexml/apps/phone/restore.py
@@ -9,13 +9,14 @@ from cStringIO import StringIO
 from uuid import uuid4
 from distutils.version import LooseVersion
 from datetime import datetime, timedelta
+from wsgiref.util import FileWrapper
 from xml.etree import cElementTree as ElementTree
 
 import six
 from celery.exceptions import TimeoutError
 from celery.result import AsyncResult
 from couchdbkit import ResourceNotFound
-from django.http import HttpResponse, FileResponse
+from django.http import HttpResponse, StreamingHttpResponse
 from django.conf import settings
 
 from casexml.apps.phone.data_providers import get_element_providers, get_async_providers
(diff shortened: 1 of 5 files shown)


millerdev:
Tested on staging
- before applying cbea782779c604df1eeb98ce9cd1c4d77973a41c I was able to reproduce `upstream prematurely closed connection while reading upstream` error in NGINX, and there was no error in Python. Phone said there was a sync error and did not allow me to login to the app.
- after applying cbea782779c604df1eeb98ce9cd1c4d77973a41c I was able to login and sync successfully.

I don't know why [`FileResponse`](https://docs.djangoproject.com/en/1.11/ref/request-response/#fileresponse-objects) did not work, but maybe it has something to do with "It uses [wsgi.file_wrapper](https://www.python.org/dev/peps/pep-3333/#optional-platform-specific-file-handling) if provided by the wsgi server, otherwise it streams the file out in small chunks." and the fact that I was passing a [`TemporaryFile`](https://docs.python.org/2/library/tempfile.html#tempfile.TemporaryFile) to it? That would be surprising, but I can't think of anything else.

snopoke (approved):

snopoke:
Could it be that the file get's closed / garbage collected before the transfer is complete?

emord:
My guess is that this is hitting an nginx error where the restore is being moved from memory to the file system because it's larger than the default for `client_body_buffer_size`

It would be interesting to see if bumping https://github.com/dimagi/commcarehq-ansible/blob/master/ansible/roles/nginx/vars/couchdb2.yml#L21 for whatever endpoint this hits would solve that as well (shouldn't go in that file. just linking as an exapmle)

snopoke:
@emord If it was that surely we would see it regularly? Also how would a change in the django code cause that to manifest/

emord:
Hmm. yeah I don't think that should be it either

PR #18110, opened 2017-10-05
Resize image to be parent dimensions

wpride:
fix for corehq/apps/cloudcare/templates/formplayer/case_list.html

Maybe @orangejenny or @biyeun can weigh in - essentially I want the image to fill (but not overflow) the parent `div` while maintaining the aspect ratio. 

Before:
screen shot 2017-10-05 at 4 56 20 pm
After:
screen shot 2017-10-05 at 4 55 52 pm

### DIFF
--- a/corehq/apps/cloudcare/templates/formplayer/case_list.html
+++ b/corehq/apps/cloudcare/templates/formplayer/case_list.html
@@ -125,11 +125,13 @@
     <% _.each(data, function(datum, index) { %>
     <div class="<%= prefix %>-grid-style-<%= index %> box">
         <% if (styles[index].displayFormat === 'Image') {
-        if(resolveUri(datum)) { %>
-        <img class="module-icon" style="max-width:100%; max-height:100%;" src="<%= resolveUri(datum) %>"/>
-        <% } %>
+            if(resolveUri(datum)) { %>
+                <img class="module-icon" style="max-width:100%; max-height:100%;" src="<%= resolveUri(datum) %>"/>
+            <% } %>
+        <% } else if(styles[index].widthHint === 0) { %>
+            <div style="display:none;"><%= datum %></div>
         <% } else { %>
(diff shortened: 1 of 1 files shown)


orangejenny:
Is the image guaranteed to be bigger than the allotted space in at least one dimension? I don't think setting max-width and max-height alone will work on small images; you'd need to also set either width or height to 100% to stretch the image.

If you changed the markup to use a div with a background image, instead of an actual image element, you could use `background-size: cover; background-repeat: no-repeat;` to do this (see https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Backgrounds_and_Borders/Scaling_background_images#Special_values.3A_.22contain.22_and_.22cover.22)

biyeun:
@wpride I second @orangejenny's suggestion of the `background-size` and `background-repeat` use.

wpride:
@orangejenny @biyeun Thanks guys! I'm not too worried about the image being smaller than the parent since people tend to use overly large icons and the allotted space is rather small. Given this would you still recommend the change?

If so, follow-up question: when using the `img` tag I can pass in a relative path like `/hq/multimedia/file/CommCareImage/f003c6a84f0905beac60789fba75313b/high_risk_flag.png` as the `src` and some middleware figures out how to transform this into the full path. `background-image` isn't doing the same - any ideas how to work around that?

orangejenny:
Doesn't make much difference whether your use the image or the background approach, your call.

I'm surprised the url doesn't work, though. Not sure if it'll be intelligible, but attached is a screenshot of me playing around on prod (I just added some inline styling to the `.list-grid-style-6.box`, to produce the clock in the blue box), and it seems to be fine with the same `/hq/multimedia...` path the original image is using.

screen shot 2017-10-09 at 9 11 50 am

dannyroberts (approved):

PR #18112, opened 2017-10-05
Log the datetime a repeat record was created

esoergel:
@proteusvacuum @gcapalbo


gcapalbo (approved):

proteusvacuum:
what are you using this for? do you want to add it to a report?

esoergel:
@proteusvacuum probably eventually - but for now I'm happy just to have it available.  `record.attempts[0].datetime` is a reasonable proxy, but this seemed like a low-effort way to know for sure when something was first triggered.

I found a record that looked like it shouldn't have been sent, so I wanted to see when it was queued to help figure out how it might have been in a triggerable state.

PR #17824, opened 2017-09-13
RequireJS proof of concept

orangejenny:
Replacement for https://github.com/dimagi/commcare-hq/pull/17395, ignore for now.

PR #18116, opened 2017-10-06
RequireJS proof of concept

orangejenny:
Replacement for https://github.com/dimagi/commcare-hq/pull/17824 after rebasing and doing some cleanup.

@millerdev https://github.com/dimagi/commcare-hq/compare/dc66724a961702e25738b2e958d59272cdf13309...jls/fixtures-require-js-2017-qa is the diff starting after the last commit you commented on. It's probably easier to review the full diff. The major changes are getting RequireJS optimization working with the build process, handled by [build_requirejs.py](https://github.com/dimagi/commcare-hq/pull/18116/files#diff-b892e9227f07ab50fedf93f979fd955b), which runs after `collectstatic`. (There are some corresponding changes to https://github.com/dimagi/commcare-hq-deploy/pull/389.)

I'm now generating the optimized `bundle.js` files automatically instead of having them in source control. `build_requirejs` walks through the static files directories (only the ones in `corehq`, not `bower_components`, and it also excludes `formdesigner`) and for every directory that contains javascript, it writes out a no-op `bundle.js` file and adds a [module](https://github.com/requirejs/r.js/blob/master/build/example.build.js#L355) entry to the r.js build config. It then writes the completed build config out to a js file and runs `r.js`.

The r.js build config has [bundlesConfigOutFile](https://github.com/requirejs/r.js/blob/master/build/example.build.js#L657) set, which makes r.js add a `bundles` config to the requirejs config that's included in base.html. That [bundles](http://requirejs.org/docs/api.html#config-bundles) config lists out each bundle and its contents, telling HQ to request the bundles rather than individual files. *This* then means that developers can require individual modules, not bundles, which should be a better experience. (Previously I'd said the convention should be to usually require the bundle, but now the convention can be to *always* require the individual module, which I think is also more intuitive.)

It's also possible to configure a custom bundle in the r.js build config for when the one-bundle-per-dir isn't ideal. I'm currently doing this for app manager (to avoid form designer) and for reports (because only a few js files in reports, namely the ones that are used by the view lookup table page, are actually RequireJS-friendly, so I need to exclude the others).

Lastly, `build_requirejs` sets up a [paths](http://requirejs.org/docs/api.html#config-paths) config that maps each js file (including bundles) to its proper location & version on the CDN.

This is just about ready for QA. Known issues:
- [x] select2 isn't working
- [x] the one TODO (externalizing the analytics for the global horizontal nav, which I'll do in a -
separate PR against master): https://github.com/dimagi/commcare-hq/pull/18139
- [x] live preview doesn't work on staging, which I'm looking into but doesn't need to block QA

QA ticket: http://manage.dimagi.com/default.asp?263535


corehq/apps/hqwebapp/static/hqwebapp/js/hqModules.js
```diff
@@ -39,14 +40,50 @@
...
-function hqDefine(path, moduleAccessor) {
-    if (typeof COMMCAREHQ_MODULES[path] !== 'undefined') {
-        throw new Error("The module '" + path + "' has already been defined elsewhere.");
-    }
-    if (path.match(/\.js$/)) {
-        throw new Error("Error in '" + path + "': module names should not end in .js.");
+/*
+ * Transitional version of "define" to handle both RequireJS and non-RequireJS pages.
+ * Signature deliberately matches that of "define". On non-RequireJS pages, the dependencies
+ * argument is optional, and moduleAccessor gets passed jQuery, knockout, and underscore, in that order.
```
millerdev: > moduleAccessor gets passed jQuery, knockout, and underscore, in that order.

Is that true? I don't see where it happens.

corehq/apps/hqwebapp/management/commands/build_requirejs.py
```diff
@@ -0,0 +1,87 @@
...
+
+            # Write .js files to staticfiles
+            for module in config['modules']:
+                with open(os.path.join(self.root_dir, 'staticfiles', module['name'] + ".js"), 'w') as fout:
+                    fout.write("define([], function() {});")
+
+            with open(os.path.join(self.root_dir, 'build.js'), 'w') as fout:
+                fout.write("({});".format(json.dumps(config, indent=4)))
+
+        call(["node", "bower_components/r.js/dist/r.js", "-o", "build.js"])
```
millerdev: where is this `build.js` file created? guessing it's the root of the repository which is already littered with junk. Could we put it in `staticfiles` or somewhere else?

corehq/apps/hqwebapp/static/hqwebapp/yaml/requirejs.yaml
```diff
@@ -0,0 +1,56 @@
+mainConfigFile: corehq/apps/hqwebapp/static/hqwebapp/js/requirejs_config.js
+bundlesConfigOutFile: hqwebapp/js/requirejs_config.js
+baseUrl: staticfiles
+dir: staticfiles
+allowSourceOverwrites: true
+keepBuildDir: true
+skipDirOptimize: true   # could turn this off to minify everything
+fileExclusionRegExp: /(^\.)|(\.css$)|(CACHE)|(^formdesigner)/,
```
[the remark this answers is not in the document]
millerdev: Should `formdesigner` match the entire filename? i.e., `^formdesigner$`

millerdev (commented):
Few comments, nothing too big.

corehq/apps/hqwebapp/static/hqwebapp/js/hqModules.js
orangejenny: yep, this is outdated, will fix

corehq/apps/hqwebapp/management/commands/build_requirejs.py
orangejenny: moved to `staticfiles`

corehq/apps/hqwebapp/static/hqwebapp/yaml/requirejs.yaml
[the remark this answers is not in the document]
orangejenny: Did some testing and found the regex is only matching the file name, not the full path.  So the only  practical thing to exclude is css files, since I'm not doing css optimizations. Updated.

orangejenny:
@millerdev When you're back online, could you take a look at the commits made during QA? https://github.com/dimagi/commcare-hq/compare/43a47a1aecd5070ab1034b55fa16d1eca80ea0aa...jls/fixtures-require-js-2017-qa?w=1

I need to integrate this with the new analytics code, but otherwise it's good to go. I'm hoping to release Tuesday, the 28th.

orangejenny:
@millerdev link in previous comment, to the partial diff to review, has been updated

orangejenny:
Link for review updated again, to https://github.com/dimagi/commcare-hq/compare/46e5e20f58ee5cd600df367a219c8068b50d82f3...jls/fixtures-require-js-2017-qa?w=1 after fixing today's merge conflicts and then rebasing.

millerdev (approved):

corehq/apps/hqwebapp/static/hqwebapp/js/hqModules.js
```diff
@@ -70,7 +70,7 @@ function hqDefine(path, dependencies, moduleAccessor) {
                     args[i] = hqImport(dependency);
                 }
             }
-            if (!COMMCAREHQ_MODULES.hasOwnProperty(path)) {
+            if (typeof COMMCAREHQ_MODULES[path] === 'undefined') {
```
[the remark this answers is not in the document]
orangejenny: fixed in ef98c89

millerdev (approved):

millerdev (approved):

orangejenny:
@millerdev One more commit, which removes `compress js` from requirejs-related scripts. I was seeing some weirdness yesterday related to the CDN's caching (the code that lives in `staticfiles/CACHE`) and js files not getting updated on deploy. Since r.js itself minifies, I think it's fine to just bypass django compression altogether.

millerdev (approved):

millerdev:
# 🏅 🏆 🏆 🏆 🏅

PR #18119, opened 2017-10-06
Changes in CBNAAT Lab Register

kkrampa:
* Changed logic of follow up month
* Changed way of pulling referring facility type


kkrampa:
@calellowitz

PR #18121, opened 2017-10-06
 delete forms with missing attachments on duplicate submission

snopoke:
https://sentry.io/dimagi/commcarehq/issues/261297321

This doesn't address the root cause but should unblock the re-submissions.

### DIFF
--- a/corehq/form_processor/parsers/form.py
+++ b/corehq/form_processor/parsers/form.py
@@ -1,5 +1,7 @@
 import datetime
 
+from couchdbkit import ResourceNotFound
+
 from corehq.form_processor.interfaces.dbaccessors import FormAccessors
 from corehq.form_processor.interfaces.processor import FormProcessorInterface
 from corehq.form_processor.models import Attachment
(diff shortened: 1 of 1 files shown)


nickpell (approved):

corehq/form_processor/parsers/form.py
```diff
@@ -165,7 +167,16 @@ def _handle_duplicate(new_doc):
     """
     interface = FormProcessorInterface(new_doc.domain)
     conflict_id = new_doc.form_id
-    existing_doc = FormAccessors(new_doc.domain).get_with_attachments(conflict_id)
+    try:
+        existing_doc = FormAccessors(new_doc.domain).get_with_attachments(conflict_id)
+    except ResourceNotFound:
```
millerdev: Is it possible that some issue other than missing attachments would cause this error? It's a generic couchdbkit error that could be raised if the doc is not found, right? Does that matter?

millerdev (approved):

PR #18124, opened 2017-10-06
Fix find_users_with_toggle_enabled

kaapstorm:
So `find_users_with_toggle_enabled` couldn't work, because toggles' enabled users aren't namespaced. 

Good thing it isn't used anywhere. :upside_down_face: 

And YAGNI. ... But ... I decided I wouldn't kill it after all, because `find_domains_with_toggle_enabled` exists, and future developers would expect `find_users_with_toggle_enabled` to exist too.

Yep, more toggle stuff, @millerdev, @calellowitz

### DIFF
--- a/corehq/ex-submodules/toggle/shortcuts.py
+++ b/corehq/ex-submodules/toggle/shortcuts.py
@@ -72,23 +72,18 @@ def parse_toggle(entry):
     return namespace, entry
 
 
-def find_with_toggle_enabled(toggle, namespace):
-    from corehq.toggles import ALL_NAMESPACES
-
-    if namespace not in ALL_NAMESPACES:
-        raise ValueError('Unknown toggle namespace "{}"'.format(namespace))
+def find_users_with_toggle_enabled(toggle):
     try:
         doc = Toggle.get(toggle.slug)
     except ResourceNotFound:
         return []
-    return [user[len(namespace) + 1:] for user in doc.enabled_users if user.startswith(namespace)]
-
-
-def find_users_with_toggle_enabled(toggle):
-    from corehq.toggles import NAMESPACE_USER
-    return find_with_toggle_enabled(toggle, namespace=NAMESPACE_USER)
+    return filter(lambda user: ':' not in user, doc.enabled_users)
 
 
 def find_domains_with_toggle_enabled(toggle):
     from corehq.toggles import NAMESPACE_DOMAIN
-    return find_with_toggle_enabled(toggle, namespace=NAMESPACE_DOMAIN)
+    try:
+        doc = Toggle.get(toggle.slug)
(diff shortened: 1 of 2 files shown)


corehq/ex-submodules/toggle/shortcuts.py
```diff
@@ -72,23 +72,18 @@ def parse_toggle(entry):
...
         doc = Toggle.get(toggle.slug)
     except ResourceNotFound:
         return []
-    return [user[len(namespace) + 1:] for user in doc.enabled_users if user.startswith(namespace)]
-
-
-def find_users_with_toggle_enabled(toggle):
-    from corehq.toggles import NAMESPACE_USER
-    return find_with_toggle_enabled(toggle, namespace=NAMESPACE_USER)
+    return filter(lambda user: ':' not in user, doc.enabled_users)
```
millerdev: Is it possible that a username could contain `:`? I think it would be safer to do something like
```py
prefixes = tuple(ns + ":" for ns in ALL_NAMESPACES if ns is not None)
return [u for u in doc.enabled_users if not u.startswith(prefixes)]
```
Any username that starts with `domain:` is doomed (is that possible?), but there have always been issues like that because of how the namespacing was implemented way back when.

millerdev: Actually, just tested this and it appears we do not allow usernames containing `:` so maybe it doesn't matter?

corehq/ex-submodules/toggle/shortcuts.py
```diff
@@ -72,23 +72,18 @@ def parse_toggle(entry):
...
 
 
 def find_domains_with_toggle_enabled(toggle):
     from corehq.toggles import NAMESPACE_DOMAIN
-    return find_with_toggle_enabled(toggle, namespace=NAMESPACE_DOMAIN)
+    try:
+        doc = Toggle.get(toggle.slug)
+    except ResourceNotFound:
+        return []
+    return [user[len(NAMESPACE_DOMAIN) + 1:] for user in doc.enabled_users if user.startswith(NAMESPACE_DOMAIN)]
```
millerdev: I think this will enable my (user, not domain) account "dmiller@dimagi.com" if I create another account named "domain_dmiller@dimagi.com". Suggested change:
```py
prefix = NAMESPACE_DOMAIN + ":"
return [user[len(prefix):] for user in doc.enabled_users if user.startswith(prefix)]
```

corehq/ex-submodules/toggle/shortcuts.py
kaapstorm: Yeah, but I didn't know `startswith` can take a tuple. So now I have to change it. :grin:

millerdev (approved):

PR #18128, opened 2017-10-06
re-fetch the user so we don't get a document update conflict

esoergel:
This is a pretty blunt approach, but there is already a request to the BETS
server between when this user is pulled and when it's saved, so it's not
unreasonable to anticipate race conditions.
@proteusvacuum


proteusvacuum (approved):

proteusvacuum:
Looks fine. I guess this can be called up to 5 minutes after the `RepeatRecord` is called, so this seems reasonable. I think we might need to do this with the Location repeater too.

esoergel:
Good point - updated.

proteusvacuum (approved):

PR #18131, opened 2017-10-06
Pass around login-as user rather than re-fetching from the DB

esoergel:
I think this might be causing issues with device ID syncing for eNikshay.  These functions feel very messy, but I'm resisting the urge to refactor due to the high-traffic nature of this code.
@gcapalbo

### DIFF
--- a/corehq/apps/ota/tests/test_utils.py
+++ b/corehq/apps/ota/tests/test_utils.py
@@ -114,33 +114,15 @@ class RestorePermissionsTest(LocationHierarchyTestCase):
         is_permitted, message = is_permitted_to_restore(
             self.domain,
             self.web_user,
-            self.commcare_user.username,
+            self.commcare_user,
         )
         self.assertTrue(is_permitted)
-
-    def test_web_user_as_user_bad_username(self):
-        is_permitted, message = is_permitted_to_restore(
-            self.domain,
-            self.web_user,
-            self.commcare_user.raw_username,  # Malformed, should include domain
-        )
-        self.assertFalse(is_permitted)
-        self.assertRegexpMatches(message, 'Invalid restore as user')
-
-    def test_web_user_as_user_bad_domain(self):
-        is_permitted, message = is_permitted_to_restore(
-            self.domain,
-            self.web_user,
-            u'{}@wrong-domain'.format(self.commcare_user.raw_username),
-        )
-        self.assertFalse(is_permitted)
-        self.assertRegexpMatches(message, 'Invalid restore as user')
 
     def test_web_user_as_other_web_user(self):
         is_permitted, message = is_permitted_to_restore(
             self.domain,
             self.web_user,
-            self.web_user.username,
+            self.web_user,
         )
         self.assertTrue(is_permitted)
         self.assertIsNone(message)
@@ -149,7 +131,7 @@ class RestorePermissionsTest(LocationHierarchyTestCase):
         is_permitted, message = is_permitted_to_restore(
             self.domain,
             self.commcare_user,
-            self.commcare_user.username,
+            self.commcare_user,
         )
         self.assertTrue(is_permitted)
         self.assertIsNone(message)
@@ -158,7 +140,7 @@ class RestorePermissionsTest(LocationHierarchyTestCase):
         is_permitted, message = is_permitted_to_restore(
             self.domain,
             self.web_user,
-            self.web_user.username,
+            self.web_user,
         )
         self.assertTrue(is_permitted)
         self.assertIsNone(message)
@@ -170,7 +152,7 @@ class RestorePermissionsTest(LocationHierarchyTestCase):
         is_permitted, message = is_permitted_to_restore(
             self.domain,
             self.super_user,
-            self.commcare_user.username,
+            self.commcare_user,
         )
         self.assertTrue(is_permitted)
         self.assertIsNone(message)
@@ -179,7 +161,7 @@ class RestorePermissionsTest(LocationHierarchyTestCase):
         is_permitted, message = is_permitted_to_restore(
             self.domain,
             self.no_edit_commcare_user,
-            self.location_user.username,
+            self.location_user,
         )
         self.assertFalse(is_permitted)
         self.assertRegexpMatches(message, 'does not have permission')
@@ -188,7 +170,7 @@ class RestorePermissionsTest(LocationHierarchyTestCase):
         is_permitted, message = is_permitted_to_restore(
             self.domain,
             self.commcare_user,
-            self.location_user.username,
+            self.location_user,
         )
         self.assertTrue(is_permitted)
         self.assertIsNone(message)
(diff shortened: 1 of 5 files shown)


corehq/apps/ota/views.py
```diff
@@ -168,26 +169,29 @@ def get_restore_response(domain, couch_user, app_id=None, since=None, version='1
...
     )
     if not is_permitted:
         return HttpResponse(message, status=401), None
 
-    is_demo_restore = couch_user.is_commcare_user() and couch_user.is_demo_user
-    couch_restore_user = couch_user
-    if not is_demo_restore and as_user is not None:
-        couch_restore_user = CouchUser.get_by_username(as_user)
+    couch_restore_user = as_user_obj if uses_login_as else couch_user
     update_device_id(couch_restore_user, device_id)
```
esoergel: This function updates properties in `user_data` that need to be current for the restore.  I think that the calls to `CouchUser.get_by_username(as_user)` might've been hitting a cache or something and getting out-of-date information.  This PR passes around the same user object to try to avoid that.

corehq/apps/ota/tests/test_utils.py
```diff
@@ -292,23 +274,15 @@ def test_get_restore_user_as_super_user(self):
...
         )
 
-    def test_get_restore_user_not_found(self):
-        with self.assertRaises(Exception):
-            get_restore_user(
-                self.domain,
-                self.web_user,
-                '{}@wrong-domain'.format(self.commcare_user.raw_username, self.domain)
-            )
-
```
gcapalbo: how come we're getting rid of this test and the other two?

custom/enikshay/user_setup.py
```diff
@@ -226,18 +226,11 @@ def compress_id(serial_id, growth_symbols, lead_symbols, body_symbols, body_digi
...
-
-
-def set_enikshay_device_id(user):
-    device_number = get_last_used_device_number(user)
-    if device_number and user.user_data.get('id_device_number', None) != device_number:
-        user.user_data['id_device_number'] = device_number
-        user.user_data['id_device_body'] = compress_nikshay_id(device_number, 0)
+def set_enikshay_device_id(user, device_id):
+    # device_id was JUST set, so it must be in there
+    device_number = [device.device_id for device in user.devices].index(device_id) + 1
```
gcapalbo: does the order of user.devices never change?

corehq/apps/ota/views.py
gcapalbo: is this the reason for this PR? the `get_by_username` cache is cleared on user save

corehq/apps/ota/tests/test_utils.py
esoergel: They were testing behavior related to passing in bad usernames to `get_restore_user`, which is no longer possible, since they now accept user objects.  There's a separate test which tests that the user is associated with the correct domain.  Part of that logic was moved [here](https://github.com/dimagi/commcare-hq/pull/18131/files#diff-291dfd3802bf0daa51e598788d5ca13bR179) and that is now untested.  (It's pretty straightforward though, so I'm not concerned about that).

corehq/apps/ota/views.py
esoergel: It is.  I could see this being a race condition with multiple sessions hitting this at the same time.  Passing around the same user object should make sure that the version used in restore generation is the one updated here.

custom/enikshay/user_setup.py
esoergel: It doesn't - it's modified here using append: https://github.com/dimagi/commcare-hq/blob/master/corehq/apps/users/models.py#L2123-L2123

corehq/apps/ota/views.py
gcapalbo: (we discussed offline but a lighter-weight way to fix race conditions is to just use `CriticalSection`, and identify the smallest block of code necessary to be the critical section)

esoergel: @ctsims I think this should solve the issue, since the device id info is being set on the this user object, which is passed around and eventually used in the restore generation (ie, it's not later refreshed from the DB).  That still feels a little sketchy though.  @gcapalbo is proposing locking the restore so there can be only one restore at a time for a given username.  Do you think that would be an acceptable approach?  If restores are still slow, that could mean waiting to aquire a lock.  Any sense of how big an issue that might be?

ctsims: I think locking per user should be fine for eNikshay for now if it's hard to fix more generally. Currently our restores are quite fast, and the web apps users almost never login-as the same user at once for logistical reasons.

Locking per user would be way rougher on some other projects which use a very limited number of user accounts, though

esoergel: My preference would be to merge this PR and then separately look into locking this part of the restore (since I'd want this change regardless).  @gcapalbo how does that sound to you?

gcapalbo: i'm fine with the change though i don't have much general context about the restore process. it might be worth getting sign off from someone more familiar with restores as well.

if we do release it, it should be tested - has it been?

custom/enikshay/user_setup.py
gcapalbo: isn't this part still going to be an issue if there's no locking, since editing two user objects at the same time would lead to a doc update conflict?

esoergel:
@gcapalbo after more source diving, I found something that's almost certainly the root cause.  This was modified to only update once per device per day, as an optimization.  This seems far more likely to cause the observed issue than the overlapping syncs I was addressing here originally (although both are possible).  I fixed this issue and added a test to catch this issue should the code change again.

I've tested this locally, and the device ID generation works as expected.  I haven't tested out concurrent restores for the same user.  I'll reach out to get a second opinion from someone more familiar with restores.

proteusvacuum (approved):

corehq/apps/ota/utils.py
```diff
@@ -212,8 +201,9 @@ def _inner(request, domain, *args, **kwargs):
 def update_device_id(user, device_id):
     if device_id and isinstance(user, CommCareUser):
         if not user.is_demo_user:
+            # this only updates once per day for each device
             updated = user.update_device_id_last_used(device_id)
+            if toggles.ENIKSHAY.enabled(user.domain):
+                updated = set_enikshay_device_id(user, device_id) or updated
```
esoergel: @snopoke the eNikshay custom stuff is linked in here now, which makes the dependency very explicit (it used to happen in a signal).  We've identified another approach which moves this stuff out of the restore code more into custom land (a custom fixture generator).  If having this here is a headache for ICDS related optimizations, we can bump the priority of that change and drop this.  Lemme know if that's the case.

custom/enikshay/user_setup.py
esoergel: Hey, sorry, I didn't notice this comment until just now.  I now believe that race conditions are not the primary issue (although they are a possibility).  My last two commits address the more probable cause.  I think this change should address the immediate issue, and then medium term we'll switch off of this approach altogether onto something more robust (that doesn't involve saving this to the user), particularly if we keep having issues.

PR #18133, opened 2017-10-06
BETS Voucher Report

proteusvacuum:
@esoergel 

This pulls from the `voucher_fulfilled_by_location_id` property, but you can still search by district (by looking up all the pertinent child location ids and pulling vouchers that were fulfilled there). 

If you don't enter a district, it just returns no values (this is probably only going to be used by us / Abhishek, so figure it's fine)

### DIFF
--- a/custom/enikshay/reports/__init__.py
+++ b/custom/enikshay/reports/__init__.py
@@ -3,7 +3,7 @@ from custom.enikshay.reports.case_finding import CaseFindingReport
 from custom.enikshay.reports.historical_adherence import HistoricalAdherenceReport
 from custom.enikshay.reports.treatment_outcome import TreatmentOutcomeReport
 from custom.enikshay.reports.web_dashboard import WebDashboardReport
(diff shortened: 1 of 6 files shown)


orangejenny (approved):

proteusvacuum:
@gcapalbo Is there a way I can add an "SMS sent" status to this report?

gcapalbo:
@proteusvacuum it might be a little difficult to tie the voucher to the sms after the fact but we might be able to do it. we send 3 sms alerts for vouchers, did you want to include all of them here?

proteusvacuum:
@gcapalbo yes, if it is possible, I think it would be useful to show which vouchers have had which SMS messages sent against them, and if possible, the status of this SMS from our end. I think this would help in debugging issues and creating trust that the system is working as it should be.

gcapalbo:
@proteusvacuum added here: https://github.com/dimagi/commcare-hq/pull/18198

PR #18141, opened 2017-10-09
Show error when rearranging forms

orangejenny:
https://manage.dimagi.com/default.asp?259749

@proteusvacuum

### DIFF
--- a/corehq/apps/app_manager/models.py
+++ b/corehq/apps/app_manager/models.py
@@ -2895,7 +2895,8 @@ class AdvancedForm(IndexedFormBase, NavMenuItemMediaMixin):
                     None)
 
     def disable_schedule(self):
-        self.schedule.enabled = False
+        if self.schedule:
+            self.schedule.enabled = False
         phase = self.get_phase()
         if phase:
             phase.remove_form(self)
--- a/corehq/apps/app_manager/models.py
+++ b/corehq/apps/app_manager/models.py
@@ -2695,7 +2695,9 @@ class Module(ModuleBase, ModuleDetailsMixin):
             if with_source:
                 new_form.source = form.source
         else:
-            raise IncompatibleFormTypeException()
+            raise IncompatibleFormTypeException(_('''
+                Cannot move an advanced form with actions into a basic module.
+            '''))
 
         if index is not None:
             self.forms.insert(index, new_form)
--- a/corehq/apps/app_manager/models.py
+++ b/corehq/apps/app_manager/models.py
@@ -2696,7 +2696,7 @@ class Module(ModuleBase, ModuleDetailsMixin):
                 new_form.source = form.source
         else:
             raise IncompatibleFormTypeException(_('''
-                Cannot move an advanced form with actions into a basic module.
+                Cannot move an advanced form with actions into a basic menu.
             '''))
 
         if index is not None:
--- a/corehq/apps/app_manager/views/apps.py
+++ b/corehq/apps/app_manager/views/apps.py
@@ -779,8 +779,8 @@ def rearrange(request, domain, app_id, key):
             app.rearrange_forms(to_module_id, from_module_id, i, j)
         elif "modules" == key:
             app.rearrange_modules(i, j)
-    except IncompatibleFormTypeException:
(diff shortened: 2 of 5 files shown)


corehq/apps/app_manager/views/apps.py
```diff
@@ -781,14 +780,20 @@ def rearrange(request, domain, app_id, key):
         elif "modules" == key:
             app.rearrange_modules(i, j)
     except IncompatibleFormTypeException:
-        messages.error(request, _('The form can not be moved into the desired menu.'))
+        error = _('The form is incompatible with the destination menu and was not moved.')
```
proteusvacuum: This would be a really frustrating error message to receive... Why is it incompatible? What can I do about it?

corehq/apps/hqwebapp/static/hqwebapp/js/alert_user.js
```diff
@@ -38,6 +38,9 @@ hqDefine("hqwebapp/js/alert_user", function() {
                 alert.alert_class(alert.alert_class() + ' ' + tags);
             }
         }
+
+        // Jump to top of page so user sees alert
+        $("html").scrollTop(0);
```
proteusvacuum: I wish we had hovering alerts :(... 
Having the browser scroll so that I can't see what I just did is quite annoying and adds friction to making changes like this (e.g. I moved this form, but I have no feedback on what the current state is now, because it is out of the viewport. If the top of the screen is already visible, then now my browser just randomly scrolled up a few lines)

corehq/apps/app_manager/models.py
```diff
@@ -2895,7 +2895,8 @@ def get_phase(self):
                     None)
 
     def disable_schedule(self):
-        self.schedule.enabled = False
+        if self.schedule:
+            self.schedule.enabled = False
         phase = self.get_phase()
```
proteusvacuum: Curious why this was necessary... I think all advanced forms always have the `schedule` property, and I think this is only ever called on advanced forms?

proteusvacuum (approved):

corehq/apps/app_manager/models.py
orangejenny: `schedule` is there, but it defaults to `None`

corehq/apps/app_manager/views/apps.py
orangejenny: You can't do anything about it, unfortunately. I can make the "why" aspect clearer, though.

corehq/apps/app_manager/models.py
```diff
@@ -2695,7 +2695,9 @@ def add_insert_form(self, from_module, form, index=None, with_source=False):
             if with_source:
                 new_form.source = form.source
         else:
-            raise IncompatibleFormTypeException()
+            raise IncompatibleFormTypeException(_('''
+                Cannot move an advanced form with actions into a basic module.
```
proteusvacuum: should `module` actually say `menu`?

corehq/apps/app_manager/views/apps.py
```diff
@@ -779,8 +779,8 @@ def rearrange(request, domain, app_id, key):
             app.rearrange_forms(to_module_id, from_module_id, i, j)
         elif "modules" == key:
             app.rearrange_modules(i, j)
-    except IncompatibleFormTypeException:
-        error = _('The form is incompatible with the destination menu and was not moved.')
+    except IncompatibleFormTypeException as e:
+        error = "{}{}".format(_('The form is incompatible with the destination menu and was not moved.'), str(e))
```
proteusvacuum: This won't have a space between the error coming from the exception and this text.

corehq/apps/app_manager/models.py
orangejenny: oops, thanks

corehq/apps/app_manager/views/apps.py
orangejenny: coincidentally it works (because the exception's error message has a newline) but I can't say that it's working by design. will update.

proteusvacuum (approved):

PR #18144, opened 2017-10-09
Update expected_doses_taken to be inclusive

proteusvacuum:
Code buddy: @orangejenny

@ctsims @sravfeyn : this was a bug with the adherence task before. Does my logic below make sense? 

https://github.com/dimagi/commcare-hq/pull/17932#discussion_r140597462
https://trello.com/c/CpN3oKlT/228-adherence-task-date-issues

FYI @gcapalbo 

This `expected_doses_taken` property is a count of all of the doses that should
have been taken up until the date that cases are automatically closed (i.e. the
"purge date"), or the date of the last adherence case, whichever is later. We
don't want to count cases that remain open, as this would ruin the calculations
on the phone.

**aggregated_score_date_calculated = purge date**
- Today is `2017-01-31`. The purge_date is then `2017-01-01`
- Cases get closed at midnight of the 30th day after their `adherence_date` property here: https://enikshay.in/a/enikshay/data/edit/automatic_updates/edit/152/
    - If today is `2017-01-31`, then all cases where the adherence_date property is `2017-01-01` and earlier will be closed. i.e. all cases where `adherence_date` = `purge_date` are closed.
- If the `adherence_schedule_date_start` is `2016-01-01`, then we want to count the number of days between `[2016-01-01, 2017-01-01]` = 366 (since 2016 was a leap year).
- i.e. This should be inclusive.

**aggregate_score_date_calculated = latest_adherence_date**
- Today is `2017-01-31`.
- The latest adherence case we received is `2016-12-31`
- The `adherence_schedule_date_start` is `2016-01-01`
- We want to count all the cases within the window [2016-01-01, 2016-12-31]. i.e. we also count the cases that occur on the `latest_adherence_date`. i.e. 31 days.
- Conclusion: this should also be inclusive.

### DIFF
--- a/custom/enikshay/tasks.py
+++ b/custom/enikshay/tasks.py
@@ -283,10 +283,10 @@ class EpisodeAdherenceUpdate(object):
         self.domain = domain
         self.episode = episode_case
         self.adherence_data_store = get_datastore(self.domain)
-        # set purge_date to 30 days back
+        self.date_today_in_india = datetime.datetime.now(pytz.timezone(ENIKSHAY_TIMEZONE)).date()
         self.purge_date = datetime.datetime.now(
             pytz.timezone(ENIKSHAY_TIMEZONE)).date() - datetime.timedelta(days=30)
-        self.date_today_in_india = datetime.datetime.now(pytz.timezone(ENIKSHAY_TIMEZONE)).date()
+
 
         self._cache_dose_taken_by_date = False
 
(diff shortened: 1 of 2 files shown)


sravfeyn:
Code looks good, I will let @ctsims to sign off on spec

ctsims:
@proteusvacuum your logic on inclusive v. exclusive checks out for me

PR #18150, opened 2017-10-10
Private Nikshay notification date

proteusvacuum:
This adds the date that Nikshay notification happened for private sector cases and a migration to add this to old cases. 

@orangejenny

### DIFF
--- a/custom/enikshay/model_migration_sets/private_nikshay_notifications.py
+++ b/custom/enikshay/model_migration_sets/private_nikshay_notifications.py
@@ -1,8 +1,5 @@
 from casexml.apps.case.util import get_datetime_case_property_changed
-from custom.enikshay.const import (
-    ENROLLED_IN_PRIVATE,
-    REAL_DATASET_PROPERTY_VALUE,
-)
+from custom.enikshay.const import ENROLLED_IN_PRIVATE
 
 
 class PrivateNikshayNotifiedDateSetter(object):
@@ -37,9 +34,6 @@ class PrivateNikshayNotifiedDateSetter(object):
             return False
 
         if self.episode.get_case_property(ENROLLED_IN_PRIVATE) != 'true':
-            return False
-
-        if self.person.get_case_property('dataset') != REAL_DATASET_PROPERTY_VALUE:
             return False
 
         return True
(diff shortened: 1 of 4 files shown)


custom/enikshay/model_migration_sets/private_nikshay_notifications.py
```diff
@@ -0,0 +1,45 @@
...
+        if self.episode.get_case_property('date_private_nikshay_notification') is not None:
+            return False
+
+        if self.episode.get_case_property('private_nikshay_registered') != 'true':
+            return False
+
+        if self.episode.get_case_property(ENROLLED_IN_PRIVATE) != 'true':
+            return False
+
+        if self.person.get_case_property('dataset') != REAL_DATASET_PROPERTY_VALUE:
```
mkangia: Is this available for private app as well? @sheelio mentioned this was just for the public app during the change we made [here](https://github.com/dimagi/commcare-hq/blob/master/custom/enikshay/integrations/utils.py#L34)
If available,
Is it there for all cases and not only for ones after this was added?

mkangia (approved):
Must cross check about real dataset value otherwise no blocker.
Also curious to understand the use case of this? There is a form submission anyway that tells us the date of the notification so how is this change useful?

orangejenny (approved):
Code looks fine; I don't have context on the business logic.

custom/enikshay/model_migration_sets/private_nikshay_notifications.py
proteusvacuum: This is probably a redundant check here, but yes, I believe that case property is available in the private sector app.

- The Beneficiary Register (scoped to the CTD location) lists 160,003 entries
- Case search with "enrolled_in_private = true" --> 352,358
-  "enrolled_in_private" = "true" and "dataset = real" --> 158,440 entries, 
- "enrolled_in_private = true" and "dataset = test" --> 1,366 entries

i.e. private sector patients with dataset property = 159,806. 

However, the numbers don't match completely. Either there is some test data that's made it's way into the beneficiary register (which doesn't check the dataset property), or some cases were created before this property existed... 

Regardless, I'll remove this check as it sounds like it is inconsistent with our data, and any Nikshay Registered patients have a more robust check anyway.

proteusvacuum:
@mkangia this is to be used in reports, which would be very hard to get from the form submission itself as the xmlns for these system forms is unfortunately not unique.

PR #18151, opened 2017-10-10
Allow downloading full voucher report

proteusvacuum:
Adds ability to download the full report in excel. This happens async. 

@orangejenny 
FYI @esoergel

### DIFF
--- a/custom/enikshay/reports/repeaters.py
+++ b/custom/enikshay/reports/repeaters.py
@@ -134,6 +134,7 @@ class ENikshayVoucherReport(GenericTabularReport):
     base_template = 'reports/base_template.html'
     dispatcher = CustomProjectReportDispatcher
     exportable = True
+    exportable_all = True
 
     asynchronous = True
     ajax_pagination = True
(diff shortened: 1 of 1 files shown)


custom/enikshay/reports/repeaters.py
```diff
@@ -183,19 +184,27 @@ def headers(self):
...
+        return self.get_rows(paged=True)
+
+    @property
+    def get_all_rows(self):
+        return self.get_rows(paged=False)
+
+    def get_rows(self, paged=True):
+        vouchers = [
+            CommCareCase.wrap(flatten_result(result))
+            for result in self._search_results(paged=True).raw_hits
```
orangejenny: Shouldn't this pass through the given value of `paged`?

custom/enikshay/reports/repeaters.py
```diff
@@ -273,7 +285,7 @@ def _get_voucher_repeater_ids(self):
 
     @property
     def total_records(self):
-        return self._search_results().total
+        return self._search_results(paged=True).total
```
orangejenny: That's counterintuitive, I'd have expected `paged=False` to return the full count..

proteusvacuum: yeah, this is a bit confusing. the HQES `total` property returns the count of all results for a specific query, not just the number returned. I'm using the memoized decorator so that it doesn't run the query twice. 

I don't think this gets called when a report is run against all rows. I've removed the argument to make it a bit less confusing while reading this.

custom/enikshay/reports/repeaters.py
proteusvacuum: welp. don't know how I missed that. 🤦‍♂️

esoergel (approved):
Nice, this is clean.

orangejenny (approved):

PR #18152, opened 2017-10-10
Fetch associated cases only when needed and possible

mkangia:
Resolves: https://sentry.io/dimagi/commcarehq/issues/364809492/

Introduced: https://github.com/dimagi/commcare-hq/pull/18135

cc: @sravfeyn


proteusvacuum (approved):
ouch

PR #18153, opened 2017-10-10
Add ability to succeed cancelled records to command

proteusvacuum:
@esoergel 
I wanted to do a batch "succeed" action for some 99DOTS records. This updates the command you wrote to allow you also add this new attempt. 

@orangejenny

### DIFF
--- a/corehq/motech/repeaters/management/commands/update_cancelled_records.py
+++ b/corehq/motech/repeaters/management/commands/update_cancelled_records.py
@@ -2,24 +2,44 @@ import csv
 import datetime
 import re
 import time
+from mock import MagicMock
 
 from django.core.management.base import BaseCommand
 
 from corehq.motech.repeaters.const import RECORD_CANCELLED_STATE
 from corehq.motech.repeaters.dbaccessors import iter_repeat_records_by_domain
-from corehq.motech.repeaters.models import Repeater
+from corehq.motech.repeaters.models import Repeater, RepeatRecordAttempt
 
 
 class Command(BaseCommand):
     help = """
-    Send cancelled repeat records. You may optionally specify a regex to
+    Perform an action on cancelled repeat records. You may optionally specify a regex to
     filter records using --include or --exclude, an a sleep time with --sleep.
     Specify multiple include or exclude params by space separating them.
+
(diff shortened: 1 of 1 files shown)


corehq/motech/repeaters/management/commands/update_cancelled_records.py
```diff
@@ -106,3 +134,19 @@ def meets_filter(record):
...
+            datetime=datetime.datetime.utcnow(),
+            success_response=success_message,
+            succeeded=True,
+        )
+        record.add_attempt(success_attempt)
+
+        mock_response = MagicMock()
+        mock_response.status_code = response_status
+        repeater = record.repeater
+        repeater.generator.handle_success(mock_response, repeater.payload_doc(record), record)
```
esoergel: This could be confusing - `handle_success` expects a real server response.  Custom repeaters could be defining their own `handle_success` methods which are meant to deal with a structured response.  For instance, I just found [this repeater](https://github.com/dimagi/commcare-hq/blob/master/custom/enikshay/integrations/nikshay/repeater_generator.py#L133-L149) which expects `response` to have a JSON description with extra information.

I can see that there may be circumstances when we want it called, but I'm not sure this can adequately mock a request.

esoergel (approved):
Awesome, this will be very useful!  I had one bit of feedback about calling `handle_success`, but consider it non-blocking food-for-thought.

esoergel:
jeez Stickler, pull yourself together

corehq/motech/repeaters/management/commands/update_cancelled_records.py
proteusvacuum: Yeah, I had thought about this. For the repeaters I wanted to run this on, it only cared about the status code, so I went for a YAGNI approach here. Not sure if you think I should make this clearer in the code somehow that this would need to be extended if it is run against other custom repeaters?

esoergel: That's definitely fair.  I think this is fine as long as it breaks in an obvious way if `handle_success` expects something else.  You could also pass in a `--handle_success` flag or something to call it conditionally.  I'm fine leaving it as-is until we need it another way, though.

PR #18162, opened 2017-10-10
Make Nikshay ID for PCP locations optional

proteusvacuum:
@mkangia 
@orangejenny 

https://trello.com/c/O7LSHDdU/236-99dots-registrations-failing-because-of-missing-nikshay-he-ids

TODO:
- [ ] Retrigger all failed records that failed with the "Nikshay codes not found" error

### DIFF
--- a/custom/enikshay/case_utils.py
+++ b/custom/enikshay/case_utils.py
@@ -422,7 +422,8 @@ def _get_private_locations(person_case):
         return PrivatePersonLocationHierarchy(
             sto=state_location.metadata['nikshay_code'],
             dto=district_location.metadata['nikshay_code'],
-            pcp=pcp_location.metadata['nikshay_code'],
+            # HACK: remove this when we have all of the "HE ids" imported from Nikshay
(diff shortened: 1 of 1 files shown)


custom/enikshay/case_utils.py
```diff
@@ -422,7 +422,8 @@ def _get_private_locations(person_case):
         return PrivatePersonLocationHierarchy(
             sto=state_location.metadata['nikshay_code'],
             dto=district_location.metadata['nikshay_code'],
-            pcp=pcp_location.metadata['nikshay_code'],
+            # HACK: remove this when we have all of the "HE ids" imported from Nikshay
+            pcp=pcp_location.metadata.get('nikshay_code') or None,
```
proteusvacuum: @mkangia will this break the Nikshay integration? Instead of failing hard, I think it will attempt to send registrations even when there is no pcp Nikshay ID.

mkangia: This is fine.. the notification would fail at attempt.

mkangia (approved):

PR #18167, opened 2017-10-10
don't truncate password on invite

dannyroberts:
This has been here since before I joined Dimagi...

When accepting an invitation to a domain and creating the new user in that form, the chosen password has always been silently truncated to 20 characters. Now it will allow any length, just like all the other interfaces for setting a password.


snopoke (approved):
:beetle:

biyeun (approved):

PR #18175, opened 2017-10-11
Allow adding search engine for multiple environments

proteusvacuum:
@orangejenny 

This allows you to add multiple custom search engines to your browser for the different commcare environments. Prior to this, if you'd added a search engine for prod, you weren't then able to add another custom search engine for india or enikshay.


orangejenny (approved):
sweet!

PR #18179, opened 2017-10-11
Translate message, and make it more appropriate

kaapstorm:
While looking at [FB 263070](https://manage.dimagi.com/default.asp?263070) I noticed a user-facing message that seems to be addresses to Dimagi staff, and not end users. 

This change translates the message, and tries to make it more appropriate for non-Dimagi users.


orangejenny:
We say "Please report an issue" in a few places; is that what people should be doing here?

orangejenny:
(if you do end up saying "report an issue", bonus points if you also add the question mark icon that people need to click to report an issue...i think we do that in one or two places, but not consistently)

orangejenny (approved):

PR #18180, opened 2017-10-11
temporarily skip certificate verification

mkangia:
Temporarily remove ssl certificate check to support Nikshay SOAP repeaters


proteusvacuum (approved):

dannyroberts (approved):
i want to flag here that you wrote this as a generic function, but now are editing it for the corner case of a particular use. Since I assume it's being used only in that one place, that's fine, but I would advocate changing the name of the function to reflect that, maybe to something like `get_insecure_SOAP_client`.

mkangia:
@dannyroberts just wanted to make the safest possible change for this for now in favour of time and not breaking anything.
Will add a follow up PR to add an argument `verify` which can be passed to this method.

dannyroberts:
👍

PR #18184, opened 2017-10-12
Updates to Start Enterprises Gateway

gcapalbo:
- always uses unicode encoding for all messages
- adds support for capturing delivery receipts

@esoergel


corehq/messaging/smsbackends/start_enterprise/views.py
```diff
@@ -0,0 +1,36 @@
...
+
+        if not message_id:
+            return HttpResponseBadRequest("Missing 'msgId'")
+
+        message_id = message_id.strip()
+
+        try:
+            dlr = StartEnterpriseDeliveryReceipt.objects.get(message_id=message_id)
+        except StartEnterpriseDeliveryReceipt.DoesNotExist:
+            dlr = None
```
esoergel: If the object doesn't exist, then this won't log anything but still return a "success" response, is that correct?  Can you elaborate a little on when this endpoint is hit vs the call to `record_message_ids`?  Is the latter called when we send the messages, and is this called later when the provider updates the statuses?  I guess in that case, they'd be telling us they successfully sent a record we don't recognize (possibly because it was sent before this change).

esoergel (approved):

corehq/messaging/smsbackends/start_enterprise/views.py
gcapalbo: yeah it should always return a success response. we share this gateway with others so we might end up getting delivery receipts not pertinent to our effort, so if we don't find the record we just ignore it.

record_message_ids happens right after we send the message - the gateway responds with the message ids and we create the record. later they give us the delivery report in this view where we just update the record with the delivery info.

esoergel: Makes sense - thanks for the clarification!

PR #18189, opened 2017-10-12
Add debugging info to duplicate ids report

esoergel:
@ctsims
I was just looking to see if we're still getting duplicates (there are new ones, but they may be from people who haven't synced).  This adds some info to the report to make that easier to discern - does the form which opened the case have the correct device ID?

@gcapalbo

### DIFF
--- a/custom/enikshay/reports/views.py
+++ b/custom/enikshay/reports/views.py
@@ -8,7 +8,9 @@ from django.views.generic.base import View
 from corehq.apps.domain.decorators import login_and_domain_required
 from corehq.apps.domain.decorators import domain_admin_required
 from corehq.apps.locations.permissions import location_safe
+from corehq.apps.users.models import CommCareUser
 from corehq.apps.userreports.reports.filters.choice_providers import ChoiceQueryContext, LocationChoiceProvider
+from corehq.form_processor.backends.sql.dbaccessors import CaseAccessorSQL
 from corehq.form_processor.interfaces.dbaccessors import CaseAccessors
 from custom.enikshay.case_utils import CASE_TYPE_VOUCHER, CASE_TYPE_PERSON
 from custom.enikshay.const import VOUCHER_ID, ENIKSHAY_ID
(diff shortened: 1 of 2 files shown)


custom/enikshay/reports/views.py
```diff
@@ -73,6 +75,23 @@ def duplicate_ids_report(request, domain, case_type):
     counts = Counter(case['readable_id'] for case in all_cases)
     bad_cases = [case for case in all_cases if counts[case['readable_id']] > 1]
 
+
+    for case in bad_cases:
+        form = CaseAccessorSQL.get_transactions(case['case_id'])[0].form
```
esoergel: This doesn't need to be correct 100% of the time - it's accurate for the typical use case (I checked against the production data), and that's really all I'm concerned with for the purposes of this.

custom/enikshay/reports/views.py
```diff
@@ -73,6 +75,23 @@ def duplicate_ids_report(request, domain, case_type):
...
+        if form:
+            case['form_name'] = form.form_data.get('@name', 'NA')
+            form_device_number = form.form_data.get('serial_id', {}).get('outputs', {}).get('device_number')
+            case['device_number_in_form'] = form_device_number
+
+            user = CommCareUser.get(form.user_id) if form.user_id else None
```
gcapalbo: is this a temporary report or something that needs to scale? these one-off lookups will be expensive at scale and would benefit from in-memory caching

esoergel: This is a temporary report used only internally (for identifying issues with the eNikshay ID generation scheme).  It is fairly slow, but that's fine as long as it doesn't negatively impact other HQ functionality.

gcapalbo (approved):

ctsims:
@esoergel I updated the ticket with a quick forensic investigation for one user and unlike the last couple of times I did that, it did not illustrate a clear existing problem (IE: the users hadn't received the new data they'd need to get the fix yet). 

I think this is a good thing to add. The next question I always have after answering the question you posed is "What was the last sync time (based on the last sync token) of the device wehn the create form was submitted" which would be increidbly helpful if it's possible to add.

PR #18192, opened 2017-10-12
Rebuild Form Command Synchronously

sheelio:
@dannyroberts 

This is synchronous (and also uses couch to pull list of forms).


dannyroberts:
if it's pulling form ids from couch it won't work for SQL domains, right?

sheelio:
Yup it won't - I do just need this for a couch domain (I could modify the
formaccessors but didn't want to make more risky changes).

dannyroberts (changes requested):
Ah okay. Sorry to be a bummer, but someone's inevitably going to want to use this later and be super confused/disappointed when it doesn't work on SQL. Can you change the name to `rebuild_form_table_couch`?

sheelio:
Made changes.

dannyroberts (approved):

PR #18196, opened 2017-10-13
Mk skip cert check followup

mkangia:
This builds on the quick fix done earlier https://github.com/dimagi/commcare-hq/pull/18180

1. Make skip ssl check configurable and not for all
2. Add skip ssl check for JSON requests

### DIFF
--- a/corehq/ex-submodules/dimagi/utils/post.py
+++ b/corehq/ex-submodules/dimagi/utils/post.py
@@ -13,7 +13,7 @@ def tmpfile(*args, **kwargs):
     return (os.fdopen(fd, 'w'), path)
 
 
-def simple_post(data, url, content_type="text/xml", timeout=60, headers=None, auth=None):
+def simple_post(data, url, content_type="text/xml", timeout=60, headers=None, auth=None, verify=None):
     """
     POST with a cleaner API, and return the actual HTTPResponse object, so
     that error codes can be interpreted.
@@ -32,6 +32,9 @@ def simple_post(data, url, content_type="text/xml", timeout=60, headers=None, au
     }
     if auth:
         kwargs["auth"] = auth
+
+    if verify:
+        kwargs["verify"] = verify
 
     return requests.post(url, data, **kwargs)
 
--- a/corehq/ex-submodules/dimagi/utils/post.py
+++ b/corehq/ex-submodules/dimagi/utils/post.py
@@ -36,17 +36,17 @@ def simple_post(data, url, content_type="text/xml", timeout=60, headers=None, au
     return requests.post(url, data, **kwargs)
 
 
-def get_SOAP_client(url):
+def get_SOAP_client(url, verify=True):
     # http://docs.python-zeep.org/en/master/transport.html#ssl-verification
     # Disable SSL cert verification meanwhile a better way is found
     session = requests.Session()
-    session.verify = False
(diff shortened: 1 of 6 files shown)


corehq/motech/repeaters/models.py
```diff
@@ -242,11 +242,17 @@ def get_auth(self):
             return HTTPDigestAuth(self.username, self.password)
         return None
 
-    def send_request(self, repeat_record, payload):
+    @property
+    def verify(self):
+        # overwrite to allow/skip certificate verification when sending request
+        # to https urls
+        return None
```
proteusvacuum: should this default to `True`? This seems like it will set the default to not verify SSL certificates, which seems bad.

custom/enikshay/integrations/nikshay/repeaters.py
```diff
@@ -52,6 +52,10 @@
 
 
 class BaseNikshayRepeater(CaseRepeater):
+    @property
+    def verify(self):
+        return True
```
proteusvacuum: Did you mean to set this to `False`? I though the issue was that the certificate from Nikshay was failing the verification.

proteusvacuum (changes requested):

corehq/motech/repeaters/models.py
mkangia: Passed `None` to use the default value or skip passing as argument..
https://github.com/dimagi/commcare-hq/pull/18196/files#diff-e6650b37cfd2fbc815ba0d96d2476082R36

custom/enikshay/integrations/nikshay/repeaters.py
mkangia: Sorry. Yes..

corehq/ex-submodules/dimagi/utils/post.py
```diff
@@ -33,30 +33,33 @@ def simple_post(data, url, content_type="text/xml", timeout=60, headers=None, au
     if auth:
         kwargs["auth"] = auth
 
+    if verify:
```
proteusvacuum: I think you want this to be:
```py
if verify is not None:
``` 
otherwise, this won't pass the parameter on when `verify = False`.

proteusvacuum (changes requested):

corehq/motech/repeaters/models.py
proteusvacuum: https://github.com/dimagi/commcare-hq/pull/18196#discussion_r144617076

corehq/ex-submodules/dimagi/utils/post.py
mkangia: Right. updated.

corehq/motech/repeaters/models.py
proteusvacuum: I still think this is a little confusing to have this have 3 states (`True`, `False`, and `None`) when the resulting behavior of `True` and `None` are the same. That might just be stylistic though, so up to you if you want to change it.

proteusvacuum (approved):

corehq/motech/repeaters/models.py
mkangia: Agree. Was just scared if this would start verifying certificates where not needed. 
Did a quick check and looks like a safe change so updated to `True`

PR #18197, opened 2017-10-13
Add index on MessagingSubEvent.case_id

gcapalbo:
@esoergel


gcapalbo:
@proteusvacuum this is also needed for https://github.com/dimagi/commcare-hq/pull/18198

corehq/apps/sms/migrations/0026_add_messagingsubevent_case_id_index.py
```diff
@@ -0,0 +1,29 @@
...
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('sms', '0025_add_custom_metadata'),
+    ]
+
+    operations = [
+        migrations.CreateModel(
+            name='StartEnterpriseBackend',
```
proteusvacuum: is this supposed to be here?

corehq/apps/sms/models.py
```diff
@@ -1275,7 +1275,7 @@ class MessagingSubEvent(models.Model, MessagingStatusMixin):
     xforms_session = models.ForeignKey('smsforms.SQLXFormsSession', null=True, on_delete=models.PROTECT)
 
     # If this was a reminder that spawned off of a case, this is the case's id
-    case_id = models.CharField(max_length=126, null=True)
+    case_id = models.CharField(max_length=126, null=True, db_index=True)
```
proteusvacuum: 👍

corehq/apps/sms/migrations/0026_add_messagingsubevent_case_id_index.py
esoergel: @gcapalbo I think this might've gotten pulled in from other changes you had locally.

esoergel (changes requested):

corehq/apps/sms/migrations/0026_add_messagingsubevent_case_id_index.py
gcapalbo: nope, django added it with makemigrations. it's basically a no-op migration that was never added when the start enterprises backend was added (which resides in a different app).

it's an unfortunate consequence of having proxy models in django - it doesn't change the db in any way but it has to keep track of it since SQLMobileBackend (the proxy model) resides in the sms app.

gcapalbo: yes, see https://github.com/dimagi/commcare-hq/pull/18197#discussion_r144570091

esoergel (approved):

PR #18198, opened 2017-10-13
Add links to messaging event detail records in voucher report

gcapalbo:
@proteusvacuum @esoergel

### DIFF
--- a/custom/enikshay/reports/repeaters.py
+++ b/custom/enikshay/reports/repeaters.py
@@ -8,6 +8,7 @@ from dimagi.utils.decorators.memoized import memoized
 
 from corehq.apps.es.case_search import flatten_result
 from corehq.apps.locations.models import SQLLocation
+from corehq.apps.sms.models import MessagingEvent
 from casexml.apps.case.models import CommCareCase
 from corehq.motech.repeaters.dbaccessors import (
     iter_repeat_records_by_domain,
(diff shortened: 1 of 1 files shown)


custom/enikshay/reports/repeaters.py
```diff
@@ -233,6 +235,24 @@ def _get_voucher_location_ids(self):
...
+            '<a target="_blank" href="/a/%s/reports/message_event_detail/?id=%s">[%s]</a>' %
+            (self.domain, messaging_event_id, messaging_event_id)
+        )
+
+    def get_messaging_event_links(self, voucher_case_id):
+        event_pks = (
+            MessagingEvent
+            .objects
+            .filter(domain=self.domain, messagingsubevent__case_id=voucher_case_id)
+            .values_list('pk', flat=True)
```
proteusvacuum: Do you think it would be valuable to also pull out the `status` here and display that in the link too, so it is easy to see at a glance what the status of the particular message is? I guess these events don't have readable names (like "Voucher Payment Confirmation"), so this might be more confusing. We can add this later if we think it will be useful.

proteusvacuum (approved):
Thanks! I've already added the permission for the Messaging Event report to the pertinent users.

custom/enikshay/reports/repeaters.py
gcapalbo: status is actually a little complex to calculate because there are three levels where it can be affected - at the MessagingEvent level, the MessagingSubEvent level, and at the SMS delivery level. so i just linked to the messaging event detail record because that computes the status taking all of those into account

proteusvacuum: ok, thanks

PR #18200, opened 2017-10-13
removing resident filter from agg script

calellowitz:
@sheelio cc: @kaapstorm @emord 
http://manage.dimagi.com/default.asp?263566
The resident filter was being applied to the column representing all women but not those registered in ICDS which made the percent registered over 100% since there were registered women that were not residents. This fixes the data in the short term, since the percents will be more reasonable, while we correctly implement the filter that was intended (migration status).


kaapstorm:
lgtm, although my only context is your PR comment.

sheelio:
Looks good to me as well (surprised this didn't cause any issues in the aggregation tests)

PR #18204, opened 2017-10-13
Make BETS payment confirmations chunked and async

esoergel:
The partner is reporting mysterious errors with larger requests.  I'm not convinced this is the problem, but this certainly seems like something we should do anyways if this endpoint is to handle thousands of case updates at a time.
@proteusvacuum @gcapalbo

### DIFF
--- a/custom/enikshay/integrations/bets/tasks.py
+++ b/custom/enikshay/integrations/bets/tasks.py
@@ -1,0 +1,12 @@
+from celery.task import task
+from corehq.apps.hqcase.utils import bulk_update_cases
+from dimagi.utils.chunked import chunked
+
+
+@task(queue='background_queue', ignore_result=True)
+def process_payment_confirmations(domain, payment_confirmations):
+    for chunk in chunked(payment_confirmations, 100):
+        bulk_update_cases(domain, [
+            (update.case_id, update.properties, False)
+            for update in chunk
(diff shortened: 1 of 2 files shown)


custom/enikshay/integrations/bets/tasks.py
```diff
@@ -0,0 +1,12 @@
+from celery.task import task
+from corehq.apps.hqcase.utils import bulk_update_cases
+from dimagi.utils.chunked import chunked
+
+
+@task(queue='background_queue', ignore_result=True)
```
esoergel: @gcapalbo do these look like the right params to you?

esoergel:
Gonna add in some logging.

esoergel:
@proteusvacuum I just added in some logging in the form of a notify exception.  It will dump all of the payment confirmations which contributed to the case updates to a json string so we can easily reconstruct that chunk of updates.  I'm also making an email filter to make sure I get notified if ever this occurs.
For reference, [here's a similar notify exception](https://github.com/dimagi/commcare-hq/blob/19969355f348374093fde50adfc0c153535e7bcf/corehq/apps/userreports/adapter.py#L62-L69) I found (one with details), and [here](https://sentry.io/dimagi/commcarehq/issues/311047779/) is what it looks like on sentry.  You'll also find that exception in your email.

proteusvacuum (approved):
Looks good. Thanks for updating.

PR #18211, opened 2017-10-16
vary restore cache on loadtest_factor

dannyroberts:
to fix https://manage.dimagi.com/default.asp?263060

### DIFF
--- a/corehq/ex-submodules/casexml/apps/phone/restore_caching.py
+++ b/corehq/ex-submodules/casexml/apps/phone/restore_caching.py
@@ -31,7 +31,7 @@ class _CacheAccessor(object):
 @quickcache(['domain', 'user_id'])
 def get_loadtest_factor_for_user(domain, user_id):
     from corehq.apps.users.models import CouchUser, CommCareUser
-    if ENABLE_LOADTEST_USERS.enabled(domain):
+    if ENABLE_LOADTEST_USERS.enabled(domain) and user_id:
         user = CouchUser.get_by_user_id(user_id, domain=domain)
         if isinstance(user, CommCareUser):
             return user.loadtest_factor or 1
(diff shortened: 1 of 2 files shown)


corehq/ex-submodules/casexml/apps/phone/restore_caching.py
```diff
@@ -26,6 +28,16 @@ def invalidate(self):
         get_redis_default_cache().delete(self.cache_key)
 
 
+@quickcache(['domain', 'user_id'])
```
emord: Should this be quickcached for longer?

dannyroberts:
Didn't want it merged right before we all went offline for the summit.

Main hesitations here are that this adds something to every restore request, but I've tried to make that very minimal and only affect domains with the `ENABLE_LOADTEST_USERS` flag set. Am I correct in assuming that on environments where it matters (e.g. ICDS), flag lookups are hard-coded so there's ~zero performance hit to checking one, and that for other environments adding a cache lookup is totally fine? The only other hesitation is that cache invalidation is often wrong in some counter-intuitive way, but I've tried to work this out logically and believe it should work.

corehq/ex-submodules/casexml/apps/phone/restore_caching.py
dannyroberts: The default is 5 hours. What do you think would be a reasonable timeout? 24 hours?

emord: I think the default is 5 minutes? not 5 hours?

emord:
> flag lookups are hard-coded so there's ~zero performance hit to checking one

This is only true if they use `always_enabled` or `always_disabled` or some environment string that we added for enikshay recently in corehq/toggles.py

corehq/ex-submodules/casexml/apps/phone/restore_caching.py
dannyroberts: Yeah silly me you're right. I'll make it the same timeout as restore cache

emord (approved):
This change looks good to me. Will let you decide if that performance is good enough, or if we need to change toggles

PR #18212, opened 2017-10-16
Change repeat record attempt from a <pre/> to a <div/>

esoergel:
When the response from the server doesn't contain newlines, you have to scroll
all the way over to read the error message and it's kinda annoying.  This keeps
it visually separated, but will break lines.


esoergel:
@proteusvacuum

PR #18226, opened 2017-10-18
Mk add outcome

mkangia:
cc: @proteusvacuum 

1. Add new treatment outcomes
2. Send '0' as phone number where not available

- [x] Re-trigger failed cases

### DIFF
--- a/custom/enikshay/integrations/nikshay/repeater_generator.py
+++ b/custom/enikshay/integrations/nikshay/repeater_generator.py
@@ -568,7 +568,8 @@ def _get_person_case_properties(episode_case, person_case, person_case_propertie
         # 2B is currently setting age_entered but we are in the short term moving it to use age instead
         "page": _get_person_age(person_case_properties),
         "paddress": person_case_properties.get('current_address', ''),
-        "pmob": person_case_properties.get(PRIMARY_PHONE_NUMBER, ''),
+        # send 0 since that is accepted by Nikshay for this mandatory field
+        "pmob": person_case_properties.get(PRIMARY_PHONE_NUMBER, '0'),
         "cname": person_case_properties.get('secondary_contact_name_address', ''),
         "caddress": person_case_properties.get('secondary_contact_name_address', ''),
         "cmob": person_case_properties.get(BACKUP_PHONE_NUMBER, ''),
--- a/custom/enikshay/integrations/nikshay/field_mappings.py
+++ b/custom/enikshay/integrations/nikshay/field_mappings.py
@@ -80,13 +80,13 @@ dcexpulmonory = {
 }
 
 treatment_outcome = {
-    'cured': '1',
-    'treatment_complete': '2',
-    'died': '3',
-    'failure': '4',
-    'regimen_changed': '7',
-    'loss_to_follow_up': '5',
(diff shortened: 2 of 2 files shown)


emord (approved):

custom/enikshay/integrations/nikshay/repeater_generator.py
```diff
@@ -568,7 +568,8 @@ def _get_person_case_properties(episode_case, person_case, person_case_propertie
         # 2B is currently setting age_entered but we are in the short term moving it to use age instead
         "page": _get_person_age(person_case_properties),
         "paddress": person_case_properties.get('current_address', ''),
-        "pmob": person_case_properties.get(PRIMARY_PHONE_NUMBER, ''),
+        # send 0 since that is accepted by Nikshay for this mandatory field
+        "pmob": person_case_properties.get(PRIMARY_PHONE_NUMBER, '0'),
```
emord: Is this related to the treatment_outcome? I don't see a 0 there?

emord (approved):
comment is non blocking assuming you don't need the 0 there

custom/enikshay/integrations/nikshay/repeater_generator.py
mkangia: Sneaked in an unrelated small fix that came in yesterday for missing phone number.

`send 0 since that is accepted by Nikshay for this mandatory field`

sravfeyn (approved):

custom/enikshay/integrations/nikshay/repeater_generator.py
proteusvacuum: should we be trying to send any of the alternate phone numbers here if this specific property isn't set? We collect all of these:
```
PRIMARY_PHONE_NUMBER = 'phone_number'
BACKUP_PHONE_NUMBER = 'secondary_contact_phone_number'
OTHER_NUMBER = "phone_number_other"
ALTERNATE_NUMBER_1 = "phone_number_alternate_1"
ALTERNATE_NUMBER_2 = "phone_number_alternate_2"
ALTERNATE_NUMBER_3 = "phone_number_alternate_3"
```

mkangia: We already send `secondary_contact_phone_number` in the payload.
The reason for sending `0` was to notify that we actually don't have the phone number i.e notifying the actual scenario. Probably would have fallen back to other numbers if it was mandatory.

proteusvacuum: Yeah. I mean if `phone_number` isn't set, maybe one of the alternate numbers _is_ and we should be sending that one as the "primary" number instead?

mkangia: hmm, 
i checked two cases which had phone number missing, these were missing all alternate numbers as well. Looks like its highly unlikely someone would add an alternate number but not add the primary phone number unless the apps allows something funky.
Happy to add the checks if you think it would be of value.

proteusvacuum: That makes sense. I guess fine to leave as is if you think it doesn't add anything.

PR #18227, opened 2017-10-18
Reflect Public Sector app structure changes in HQ code

mkangia:
https://trello.com/c/O2juZsuC/221-reflect-public-sector-app-structure-changes-in-hq-code

best review 🐟 / 🐟 

cc: @nickpell 
buddy: @biyeun


custom/enikshay/case_utils.py
```diff
@@ -169,7 +172,7 @@ def get_all_episode_cases_from_person(domain, person_case_id):
     return episode_cases
 
 
-def get_open_episode_case_from_occurrence(domain, occurrence_case_id):
+def get_open_active_episode_case_from_occurrence(domain, occurrence_case_id):
     """
     Gets the first open 'episode' case for the occurrence
```
sheelio: Lets update the comment to indicate that this is for public sector DSTB episodes

custom/enikshay/integrations/utils.py
```diff
@@ -61,11 +71,18 @@ def is_valid_voucher_submission(voucher_case):
...
+    """
+    person_case = get_person_case_from_test(test_case.domain, test_case.case_id)
+    if is_invalid_person_submission(person_case):
+        return False
+
     try:
         lab_referral_case = get_lab_referral_from_test(test_case.domain, test_case.get_id)
     except ENikshayCaseNotFound:
-        return False
+        return True
```
sheelio: Why is this reversed?

custom/enikshay/tasks.py
```diff
@@ -201,13 +204,16 @@ def _get_open_episode_cases(self, case_ids):
...
+                continue
+
             # if this episode is part of a deleted or archived person, don't update
             try:
                 person_case = get_person_case_from_episode(self.domain, episode_case.case_id)
             except ENikshayCaseNotFound:
                 continue
 
-            if person_case.owner_id == ARCHIVED_CASE_OWNER_ID:
+            if person_case.owner_id in [ARCHIVED_CASE_OWNER_ID, INVALID_CASE_OWNER_ID]:
```
sheelio: Same comment as before - depending on how this utility is used, we may want to include archived cases

custom/enikshay/case_utils.py
```diff
@@ -169,7 +172,7 @@ def get_all_episode_cases_from_person(domain, person_case_id):
     return episode_cases
 
 
-def get_open_episode_case_from_occurrence(domain, occurrence_case_id):
+def get_open_active_episode_case_from_occurrence(domain, occurrence_case_id):
```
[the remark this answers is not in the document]
mkangia: Makes sense. Hope this does not cause lint error :)

custom/enikshay/integrations/utils.py
```diff
@@ -31,8 +36,13 @@ def _is_submission_from_test_location(case_id, owner_id):
     return phi_location.metadata.get('is_test', "yes") == "yes"
 
 
+def is_invalid_person_submission(person_case):
+    return person_case.owner_id == INVALID_CASE_OWNER_ID
+
+
 def is_valid_person_submission(person_case):
-    if person_case.owner_id == ARCHIVED_CASE_OWNER_ID:
+    if (person_case.owner_id == ARCHIVED_CASE_OWNER_ID or
```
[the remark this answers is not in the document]
mkangia: hmm.. 
We can rearrange this now as
```
def is_valid_person_submission(person_case):
    if person_case.dynamic_case_properties().get('case_version') == PERSON_CASE_2B_VERSION:
        return person_case.dynamic_case_properties().get('dataset') == REAL_DATASET_PROPERTY_VALUE
    owner_id = person_case.owner_id
    if person_case.owner_id == ARCHIVED_CASE_OWNER_ID:
        owner_id = person_case.dynamic_case_properties().get('last_owner', None)
    if owner_id:
        return not _is_submission_from_test_location(person_case.case_id, owner_id)
    return False
```

custom/enikshay/integrations/utils.py
mkangia: Reverted the logic of method and its name. It was pretty confusing. 
If you see the [commit](https://github.com/dimagi/commcare-hq/pull/18227/commits/245e92790329ebedfc32771761901467060e51a7)

custom/enikshay/tasks.py
mkangia: I am not sure about the enikshay task. Will defer to @proteusvacuum if he knows why archived cases were ignored here.

custom/enikshay/integrations/utils.py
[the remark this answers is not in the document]
sheelio: Where is the "last_owner" case property coming from?  I think it might be called "referred_outside_enikshay_by_id"

custom/enikshay/tasks.py
proteusvacuum: this filters out archived cases so that we don't update them every night. If a case becomes unarchived, it will start getting updated again.

custom/enikshay/case_utils.py
```diff
@@ -568,6 +577,9 @@ def iter_all_active_person_episode_cases(domain, case_ids):
         if episode_case.closed:
```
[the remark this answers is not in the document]
mkangia: @sheelio This method `iter_all_active_person_episode_cases` does not rely on it being public app. It just receives a set of case ids it iterates over. So looks the name and comment for what the method does is alright.

custom/enikshay/tasks.py
sheelio: @mkangia 

We may want to rename the function so its clearer what it does or provide an option to include or not include archived cases?

custom/enikshay/case_utils.py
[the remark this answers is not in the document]
sheelio: @mkangia 

If we're adding code about is_active = true below, that would make this function only work in the public sector

custom/enikshay/integrations/utils.py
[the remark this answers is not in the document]
mkangia: @sheelio i think last_owner is set when ever a case is archived. For ex form here, https://enikshay.in/a/enikshay/reports/case_data/3b3b20b98dcf4335b15c8861b26790f0/#!history?form_id=ed7d598953dc4bdfa6e00d9845fa46c4

sheelio:
This generally looks good (from a logic perspective).  Will let Farid provide input on code style / efficiency (I.e. I don't know how many of the referral related changes should live in the case utils file vs. in the normal referral expression file).

custom/enikshay/case_utils.py
```diff
@@ -169,44 +172,52 @@ def get_all_episode_cases_from_person(domain, person_case_id):
...
+    if open_active_episode_cases:
+        return open_active_episode_cases[0]
     else:
         raise ENikshayCaseNotFound(
             "Occurrence with id: {} exists but has no open episode cases".format(occurrence_case_id)
         )
 
 
-def get_open_drtb_hiv_case_from_episode(domain, episode_case_id):
+def get_open_drtb_hiv_case_from_occurrence(domain, occurrence_case_id):
```
mkangia: @sheelio this would now return a `secondary_owner` case. Earlier this was returning `CASE_TYPE_DRTB_HIV_REFERRAL`.

sheelio: That sounds right - where is it used?

corehq/ex-submodules/casexml/apps/case/const.py
```diff
@@ -45,3 +45,4 @@
 UNOWNED_EXTENSION_OWNER_ID = '-'
 
 ARCHIVED_CASE_OWNER_ID = '_archive_'
+INVALID_CASE_OWNER_ID = '_invalid_'
```
proteusvacuum: Should also add this here: https://github.com/dimagi/commcare-hq/blob/master/corehq/apps/users/util.py#L21-L28

custom/enikshay/case_utils.py
```diff
@@ -108,6 +110,9 @@ def get_all_occurrence_cases_from_person(domain, person_case_id):
     all_cases = case_accessor.get_reverse_indexed_cases([person_case_id])
     occurrence_cases = [case for case in all_cases
                         if case.type == CASE_TYPE_OCCURRENCE]
+    open_occurrence_cases = [case for case in occurrence_cases if not case.closed]
+    if len(open_occurrence_cases) > 1:
+        raise EnikshayBadAppState("Multiple open occurrences found for a person")
```
proteusvacuum: Would be useful to have the person id here.

custom/enikshay/case_utils.py
```diff
@@ -169,44 +172,52 @@ def get_all_episode_cases_from_person(domain, person_case_id):
...
     episode_cases = case_accessor.get_reverse_indexed_cases([occurrence_case_id])
-    open_episode_cases = [case for case in episode_cases
-                          if not case.closed and case.type == CASE_TYPE_EPISODE and
-                          case.dynamic_case_properties().get('episode_type') == "confirmed_tb"]
-    if open_episode_cases:
-        return open_episode_cases[0]
+    open_active_episode_cases = [case for case in episode_cases
+                                 if not case.closed and
+                                 case.type == CASE_TYPE_EPISODE and
+                                 case.get_case_property().get('episode_type') == "confirmed_tb" and
```
proteusvacuum: I don't think this syntax works. You need either `get_case_property('episode_type')` or `dynamic_case_properties().get('episode_type')`

custom/enikshay/case_utils.py
```diff
@@ -169,44 +172,52 @@ def get_all_episode_cases_from_person(domain, person_case_id):
...
+                                 case.get_case_property('is_active') == 'true'
+                                 ]
+    if len(open_active_episode_cases) > 1:
+        raise EnikshayBadAppState("Multiple open active episode cases found for occurrence")
```
proteusvacuum: would be good to have ids here to debug with.

custom/enikshay/case_utils.py
```diff
@@ -169,44 +172,52 @@ def get_all_episode_cases_from_person(domain, person_case_id):
...
+    open_drtb_hivsecondary_owner_case = [
+        case for case in case_accessor.get_reverse_indexed_cases([occurrence_case_id])
+        if not case.closed and
+        case.type == CASE_TYPE_SECONDARY_OWNER and
+        case.get_case_property('secondary_owner_type') == 'drtb-hiv'
     ]
-    if open_drtb_cases:
-        return open_drtb_cases[0]
+    if open_drtb_hivsecondary_owner_case:
+        return open_drtb_hivsecondary_owner_case[0]
```
proteusvacuum: what if there is more than one open one here? Is that an issue?

custom/enikshay/case_utils.py
```diff
@@ -218,27 +229,38 @@ def get_open_episode_case_from_person(domain, person_case_id):
...
-    occurrence_cases = [
-        case.case_id for case in reverse_indexed_cases
-        if not case.closed and case.type == CASE_TYPE_OCCURRENCE
-    ]
-    reversed_indexed_occurrence = case_accessor.get_reverse_indexed_cases(occurrence_cases)
-    open_referral_cases.extend(
-        case for case in reversed_indexed_occurrence
-        if not case.closed and case.type == CASE_TYPE_REFERRAL
-    )
+    if closed_cases:
```
proteusvacuum: I think you could prevent this duplication of code if you use this as your conditional
```(closed_cases and case.closed) or (not closed_cases and not case.closed)```

File: custom/enikshay/case_utils.py [hunk 14]
```diff
@@ -249,6 +271,16 @@ def get_open_referral_case_from_person(domain, person_case_id):
         )
 
 
+def get_latest_closed_referral_case_from_person(domain, person_case_id, with_status=None):
+    closed_referral_cases = get_referral_cases_from_person(domain, person_case_id, closed_cases=True)
+    if with_status:
+        closed_referral_cases = filter(
```
proteusvacuum: this is fine, but curious why you used a `filter` here instead of a list comprehension like we've been using elsewhere in this file.

custom/enikshay/integrations/nikshay/field_mappings.py
```diff
@@ -21,13 +21,15 @@
...
-    'ngo_volunteer': '5',
-    'private_medical_pracitioner': '6',
-    'other_community_volunteer': '7',
+    'health_worker': '1',  # Health Worker
+    'tbhv': '2',  # TBHV
+    'asha': '3',  # ASHA
+    'asha_or_other_phi_hw': '3',  # ASHA
+    'aww': '4',  # AWW
+    'ngo_volunteer': '5',  # NGO VOLUNTEER
+    'private_medical_pracitioner': '6',  # PRIVATE PRACTICENER
```
proteusvacuum: nit: sp in the comment: `practitioner` (and the key, but I'm assuming that's intentional)

custom/enikshay/integrations/utils.py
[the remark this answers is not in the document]
proteusvacuum: This is used by BETS, 99DOTS and Nikshay for various things. Cases get _archived_ when they are referred out? I thought _archive_ was only for when there is a treatment outcome. If not, then we should probably be sending those cases to BETS and 99DOTS too, in which case, I think Manish's proposed change makes sense.

custom/enikshay/integrations/utils.py
proteusvacuum: I think this should still be `False`? I think the logic before might have been wrong.

custom/enikshay/tasks.py
proteusvacuum: This is only used in the task (and is a private method of the task, so shouldn't be used by anything other than that).

custom/enikshay/integrations/utils.py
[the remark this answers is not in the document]
sheelio: Sorry, the person owner stuff is hard to follow.  I verified the following behavior:
- After the occurrence is closed: _archive_
- If referred outside of eNikshay: _archive_
- If the PHI owner is not yet known:  -
- if the occurrence is open and the PHI owner is known: the UUID of the PHI

custom/enikshay/case_utils.py
mkangia: oh no.. intended `get_case_property('episode_type')`.

custom/enikshay/case_utils.py
mkangia: https://github.com/dimagi/commcare-hq/blob/e635e762a9d0e6d48a5695a571965d3835948e56/custom/enikshay/nikshay_datamigration/factory.py#L120

and then in `get_drtb_hiv_referral_case_structure` -> `get_case_structures_to_create`

custom/enikshay/case_utils.py
mkangia: @sheelio

custom/enikshay/case_utils.py
mkangia: oh thats nice.

how about 
```
        referral_cases = [case for case in reverse_indexed_cases
                          if bool(case.closed) == closed_cases and case.type == CASE_TYPE_REFERRAL]
        occurrence_cases = [case.case_id for case in reverse_indexed_cases
                            if case.type == CASE_TYPE_OCCURRENCE]
        reversed_indexed_occurrence = case_accessor.get_reverse_indexed_cases(occurrence_cases)
        referral_cases.extend(
            case for case in reversed_indexed_occurrence
            if bool(case.closed) == closed_cases and case.type == CASE_TYPE_REFERRAL
        )
```

proteusvacuum: that works too! I don't think you need the `bool` cast though.

custom/enikshay/case_utils.py
mkangia: no preference. Will update it to keep it similar to what is being used.

also checked that `filter` is probably not favoured. thanks for flagging.

custom/enikshay/integrations/nikshay/field_mappings.py
mkangia: didnt realise.. copied it from the [form](https://enikshay.in/a/enikshay/apps/view/3c005532d11e463680b3f9124ea5b5a8/form/bf2473304010ef7518ad99b24eb6eafc3bcfab6d/source/) and its actually `private_medical_pracitioner` only.

custom/enikshay/integrations/utils.py
mkangia: In case the lab referral case was not found the logic earlier would return False, which means that its not a test case i.e its a real case. That probably was done to not stop processing in case the method failed to determine if its a test case or not.

So now since the expectation on return value is reversed, `False` -> `True`

custom/enikshay/case_utils.py
mkangia: @sheelio please confirm this.

custom/enikshay/case_utils.py
mkangia: @sheelio please confirm.

custom/enikshay/case_utils.py
sheelio: I'm not sure how to best follow the code, but the the underlying change is that instead of DRTB HIV cases we have secondary owner cases.  As long as we update all the places that refer to getting a DRTB HIV case to using a secondary_owner case, that's fine.

custom/enikshay/case_utils.py
sheelio: There should only ever be one open secondary owner case that is for the DRTB HIV coordinator. 

(IF there are multiple, we may want to have a reconciliation that handles this).

custom/enikshay/case_utils.py
mkangia: adding just makes it explicit that it should be a bool value.. Agreed that it is probably not needed here..

custom/enikshay/integrations/utils.py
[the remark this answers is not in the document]
mkangia: @sheelio @proteusvacuum 
the change i said makes sense then?

custom/enikshay/case_utils.py
mkangia: This is used for data migration. Usage [here](https://github.com/dimagi/commcare-hq/blob/master/custom/enikshay/nikshay_datamigration/factory.py#L114).
Hey @nickpell can you suggest what the corresponding change here would be for this update that we are doing in this PR for this method to return a `secondary_owner` case and not `drtb-hiv-referral` case.

custom/enikshay/integrations/utils.py
```diff
@@ -75,7 +92,7 @@ def is_valid_test_submission(test_case):
             {lab_referral_id}"
             .format(location_id=lab_referral_case.owner_id, lab_referral_id=lab_referral_case.case_id)
         )
-    return dmc_location.metadata.get('is_test', "yes") == "yes"
+    return dmc_location.metadata.get('is_test', "yes") == "no"
```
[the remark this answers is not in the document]
mkangia: replied above.
https://github.com/dimagi/commcare-hq/pull/18227/files#r149471001

mkangia:
:(

proteusvacuum:
😢

PR #18230, opened 2017-10-18
fix for unicode in email ucrs

czue:
https://manage.dimagi.com/default.asp?263803

Introduced by https://github.com/dimagi/commcare-hq/pull/18022/

I believe all email UCRs involving unicode have been broken for 2+ weeks, but perhaps there aren't so many of those...

### DIFF
--- a/corehq/apps/userreports/reports/view.py
+++ b/corehq/apps/userreports/reports/view.py
@@ -573,9 +573,9 @@ class ConfigurableReport(JSONResponseMixin, BaseDomainView):
     def email_response(self):
         with closing(StringIO()) as temp:
             export_from_tables(self.export_table, temp, Format.HTML)
-            return self.render_json_response({
+            return HttpResponse(json.dumps({
                 'report': temp.getvalue(),
-            })
+            }), content_type='application/json')
 
     @property
     @memoized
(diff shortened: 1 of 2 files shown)


corehq/apps/userreports/reports/view.py
```diff
@@ -573,9 +573,9 @@ def export_table(self):
     def email_response(self):
         with closing(StringIO()) as temp:
             export_from_tables(self.export_table, temp, Format.HTML)
-            return self.render_json_response({
+            return HttpResponse(json.dumps({
                 'report': temp.getvalue(),
-            })
+            }), content_type='application/json')
```
mkangia: For my understanding, how does this solve the underlying issue?

czue: good question. `render_json_response` calls `json.dumps` with [a few extra arguments](https://github.com/brack3t/django-braces/blob/master/braces/views/_ajax.py#L41-L44) including [`ensure_ascii=False`](https://github.com/brack3t/django-braces/blob/master/braces/views/_ajax.py#L33) and then reencodes it. 

This apparently chokes if the data is previously encoded ([which it is](https://github.com/dimagi/commcare-hq/pull/18022/files#diff-5fe4ce2e18624b1a325fb22b1b0a01b7R264)). 

Just some obscure coincidence that this broke us basically, which was why I wrote the high-level test.

mkangia: Thanks.

kaapstorm (approved):
Urgh. Thanks @czue.

czue:
> Urgh

I wouldn't feel bad - it would have been basically impossible to have expected / predicted / known this would break anything without already having a test in place. Such odd behavior.

PR #18233, opened 2017-10-18
add package_monitor

nickpell:
@kaapstorm

### DIFF
--- a/localsettings.example.py
+++ b/localsettings.example.py
@@ -201,6 +201,7 @@ LOCAL_APPS = (
 #    'testapps.test_pillowtop',
 #    'django_fsm', # Adds the ability to generate state diagrams for models using django-fsm
 #    'kombu.transport.django', # required for celery
+#    'package_monitor',
 )
 
 LOCAL_MIDDLEWARE = [
--- a/requirements/requirements.txt
+++ b/requirements/requirements.txt
@@ -103,3 +103,4 @@ contextlib2==0.5.4
 hiredis==0.2.0
 xlwt==1.3.0
 ddtrace==0.9.2
+git+git://github.com/nickpell/django-package-monitor@np/handle-no-semver
--- a/requirements/requirements.txt
+++ b/requirements/requirements.txt
@@ -103,4 +103,3 @@ contextlib2==0.5.4
 hiredis==0.2.0
 xlwt==1.3.0
 ddtrace==0.9.2
-git+git://github.com/nickpell/django-package-monitor@np/handle-no-semver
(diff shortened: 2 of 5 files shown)


nickpell:
https://github.com/yunojuno/django-package-monitor

settings.py
```diff
@@ -413,8 +413,7 @@
     'custom.m4change',
 )
 
-INSTALLED_APPS = DEFAULT_APPS + HQ_APPS + ENIKSHAY_APPS
-
+INSTALLED_APPS = DEFAULT_APPS + HQ_APPS + ENIKSHAY_APPS + ('package_monitor',)
```
kaapstorm: Should we maybe add `'package_monitor'` to `HQ_APPS` instead?

kaapstorm (approved):

requirements/requirements.txt
```diff
@@ -103,3 +103,4 @@ contextlib2==0.5.4
 hiredis==0.2.0
 xlwt==1.3.0
 ddtrace==0.9.2
+git+git://github.com/nickpell/django-package-monitor@np/handle-no-semver
```
snopoke: do we need this on prod?

nickpell: Good point.  I've updated the PR so the app can only be used in dev environments.

localsettings.example.py
```diff
@@ -201,6 +201,7 @@
 #    'testapps.test_pillowtop',
 #    'django_fsm', # Adds the ability to generate state diagrams for models using django-fsm
 #    'kombu.transport.django', # required for celery
+#    'package_monitor',
```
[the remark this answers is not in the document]
nickpell: ignoring

settings.py
```diff
@@ -2222,3 +2221,5 @@ def _determine_couch_databases(couch_databases):
             'NAME': 'custom.nic_compliance.password_validation.UsedPasswordValidator',
         }
     ]
+
+PACKAGE_MONITOR_REQUIREMENTS_FILE = os.path.join(FILEPATH, 'requirements', 'requirements.txt')
```
emord: I think this could go in localsettings as well

urls.py
```diff
@@ -172,6 +172,13 @@
     except ImportError:
         pass
 
+    try:
+        urlpatterns += [
```
emord: You could also use `if settings.DEBUG`

nickpell: there's one already (see above)

settings.py
nickpell: so the thing is this relies on FILEPATH which is defined in settings.  I didn't see any example of how you could access FILEPATH from localsettings

emord: I think it should auto add it because earlier in the file we have `from localsettings import *`

nickpell: (resolved)

kaapstorm (approved):

PR #18235, opened 2017-10-18
Add labels and body for specific labels to email

sheelio:
This adds more useful info to the deploy emails

### DIFF
--- a/corehq/apps/hqadmin/management/utils.py
+++ b/corehq/apps/hqadmin/management/utils.py
@@ -1,5 +1,6 @@
 from __future__ import print_function
 import re
+from collections import defaultdict
 from github3 import GitHub
 from django.template.loader import render_to_string
 import requests
@@ -94,11 +95,11 @@ def _get_pr_labels(pr_number):
 
 
 def _get_prs_by_label(pr_infos):
-    prs_by_label = {}
-    for label in LABELS_TO_EXPAND:
-        prs_by_label[label] = [pr for pr in pr_infos
-           if label in [lbl['name'] for lbl in pr['labels']]
-        ]
+    prs_by_label = defaultdict(list)
+    for pr in pr_infos:
+        for label in pr['labels']:
+            if label['name'] in LABELS_TO_EXPAND:
+                prs_by_label[label['name']].append(pr)
     return prs_by_label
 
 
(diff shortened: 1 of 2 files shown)


corehq/apps/hqadmin/management/utils.py
```diff
@@ -54,18 +62,46 @@ def _get_pr_info(pr_number):
...
+
+    return [
+        {'name': label['name'], 'color': label['color']}
+        for label in json_response['labels']
+    ]
+
+
+def _get_prs_by_label(pr_infos):
+    prs_by_label = {}
+    for label in LABELS_TO_EXPAND:
```
proteusvacuum: nit: this is fine, but you run through each PR multiple times here.
another way of doing this could be:
```py
prs_by_label = defaultdict(list)
for pr in pr_infos:
    for label in pr['labels']:
        if label['name'] in EXPAND_LABELS:
            prs_by_label[label['name']].append(pr)
```

corehq/apps/hqadmin/management/utils.py
```diff
@@ -54,18 +62,46 @@ def _get_pr_info(pr_number):
...
         'line_changes': line_changes,
+        'labels': labels,
     }
 
+
+def _get_pr_labels(pr_number):
+    url = 'https://api.github.com/repos/dimagi/commcare-hq/issues/{}'.format(pr_number)
+    json_response = requests.get(url).json()
+    if not json_response.get('number'):
+        # We've probably exceed our rate limit for unauthenticated requests
```
proteusvacuum: How often does this happen? I thought we authenticated with GH when we deployed.

proteusvacuum (approved):
Seems fine. Will let you merge when there is sign off from tech-func.

corehq/apps/hqadmin/management/utils.py
sheelio: I just used the same logic that we used for authenticating with Github in _get_pr_info( https://github.com/dimagi/commcare-hq/blob/ee22dfb04d6ce2879c6f3575e87a14e8b010e626/corehq/apps/hqadmin/management/utils.py#L53)

proteusvacuum: ah I see. you could have pulled that out into a function to avoid the duplicate code, but that makes more sense.

PR #18237, opened 2017-10-19
enforce two-factor auth for all dimagi users

snopoke:
Will need to announce and give users a chance to setup 2FA on their own schedule before merging.


nickpell (approved):

snopoke:
This can now be merged since it's behind a toggle (to allow controlled rollout)

PR #18241, opened 2017-10-19
Log superuser project entry

dannyroberts:
This just logs superuser access (every 6 hours per project). What do you think?


corehq/apps/domain/project_access/models.py
```diff
@@ -0,0 +1,26 @@
+import datetime
+from django.db import models
+
+ENTRY_RECORD_FREQUENCY = datetime.timedelta(hours=6)
+
+
+class SuperuserProjectEntryRecord(models.Model):
```
emord: This is probably going to end up being a long list as HQ grows. should we partition this table now? (maybe by week or month?) http://architect.readthedocs.io/features/partition/postgresql.html

corehq/apps/domain/project_access/models.py
```diff
@@ -0,0 +1,26 @@
...
+    @classmethod
+    def record_entry(cls, username, domain):
+        record = cls(username=username, project=domain)
+        record.save()
+
+    @classmethod
+    def entry_recently_recorded(cls, username, domain):
+        return cls.objects.filter(
+            username=username,
+            project=domain,
```
emord: Should we be more consistent on project/domain here?

corehq/apps/domain/project_access/middleware.py
```diff
@@ -0,0 +1,18 @@
...
+from corehq.util.quickcache import quickcache
+
+
+class ProjectAccessMiddleware(MiddlewareMixin):
+    def process_view(self, request, view_func, view_args, view_kwargs):
+        if hasattr(request, 'couch_user') and request.couch_user.is_superuser \
+                and hasattr(request, 'domain'):
+            return self.record_entry(request.domain, request.couch_user.username)
+
+    @atomic
```
snopoke: Will this not start a transaction even if it's a cache hit? Might be better to put `atomic` as the first decorator.

corehq/apps/domain/project_access/models.py
```diff
@@ -0,0 +1,26 @@
...
+            last_login__gt=datetime.datetime.utcnow() - ENTRY_RECORD_FREQUENCY,
+        ).count() > 0
```
snopoke: It's more efficient to do a `.exists()` query

corehq/apps/domain/project_access/models.py
```diff
@@ -0,0 +1,26 @@
+import datetime
+from django.db import models
+
+ENTRY_RECORD_FREQUENCY = datetime.timedelta(hours=6)
+
+
+class SuperuserProjectEntryRecord(models.Model):
+    username = models.EmailField()
```
emord: I think this should have an index on it. It may be useful to look at all of someone's actions if their account is compromised. or maybe even a proactive email that says "this is what you did today"

snopoke (approved):

emord (approved):

dannyroberts:
Going ahead and merging, having judged that the potential for performance issues due to the extra db checks and writes will be minimal, since they are only done after checking that `request.couch_user.is_superuser`.

emord:
>  judged that the potential for performance issues due to the extra db checks and writes will be minimal, since they are only done after checking that request.couch_user.is_superuser.

no change necessary, but it seems that logging like this could be done asynchronously through kafka/celery in the future (there's probably a lot more places that we could follow that practice. don't htink it's a big deal here)

PR #18248, opened 2017-10-23
Only update fields that aren't null.

emord:
@orangejenny guessing that this may have been caused due to some hqimport changes, but no one noticed a while ago. little scary to change this because lack of tests/hasn't been touched since 2013

buddy @mkangia


corehq/apps/reports/static/reports/js/saved_reports.js
```diff
@@ -37,13 +37,13 @@ var ReportConfig = function (data) {
     self.unwrap = function () {
         var data = ko.mapping.toJS(self);
         var standardHQReport = hqImport("reports/js/standard_hq_report").getStandardHQReport();
-        if (null !== standardHQReport.slug) {
+        if (standardHQReport.slug) {
```
mkangia: Not sure if this was done intentionally but to allow `""` and restrict just `null` value. If you have checked that its a safe change then cool.

mkangia (approved):

PR #18249, opened 2017-10-23
Increase repeater post timeout to 75 seconds

proteusvacuum:
99DOTS responses sometimes take ~60s to process, which leads us to mark those
records as failed

@gcapalbo 
FYI @mkangia @dannyroberts


dannyroberts (approved):

mkangia:
`~60s to process` 🎏

PR #18252, opened 2017-10-24
use python3 compatible version of tropo library

nickpell:


nickpell:
Closing because Tropo is shutting down

PR #18255, opened 2017-10-24
fix emails containing unicode content

nickpell:
https://sentry.io/dimagi/commcarehq/issues/390195253/


corehq/util/log.py
```diff
@@ -304,4 +304,10 @@ def draw(position):
 
 def send_HTML_email(subject, recipient, html_content, *args, **kwargs):
     kwargs['ga_track'] = kwargs.get('ga_track', False) and analytics_enabled_for_email(recipient)
-    return _send_HTML_email(subject, recipient, html_content, *args, **kwargs)
+    unicode_html_content = html_content if isinstance(html_content, unicode) else html_content.decode('utf-8')
```
emord: could use six.string_type (or whatever it is)

emord (approved):
@millerdev probably knows a cleaner way to do this, but it looks ok to me

PR #18259, opened 2017-10-24
Fixed product name in stock export.

emord:
@esoergel I think you're in charge of commtrack/exports now

https://sentry.io/dimagi/commcarehq/issues/374434653/
https://manage.dimagi.com/default.asp?263876#1410694

@mkangia

### DIFF
--- a/corehq/apps/export/models/new.py
+++ b/corehq/apps/export/models/new.py
@@ -2508,7 +2508,7 @@ class StockExportColumn(ExportColumn):
 
     def _get_product_name(self, product_id):
         try:
-            return SQLProduct.objects.values_list('name', flat=True).get(product_id=product_id)
+            return SQLProduct.objects.values_list('name', flat=True).get(product_id=product_id, domain=self.domain)
         except SQLProduct.DoesNotExist:
             return product_id
 
--- a/corehq/apps/export/models/new.py
+++ b/corehq/apps/export/models/new.py
@@ -47,7 +47,7 @@ from corehq.apps.app_manager.models import (
     Application, AdvancedFormActions, RemoteApp, OpenSubCaseAction, CaseIndex
 )
 from corehq.apps.domain.models import Domain
-from corehq.apps.products.models import Product
+from corehq.apps.products.models import SQLProduct
 from corehq.apps.reports.display import xmlns_to_name
 from corehq.blobs.mixin import BlobMixin
 from corehq.form_processor.interfaces.dbaccessors import LedgerAccessors
@@ -2507,7 +2507,10 @@ class StockExportColumn(ExportColumn):
         return section_and_product_ids
 
     def _get_product_name(self, product_id):
-        return Product.get(product_id).name
+        try:
+            return SQLProduct.objects.values_list('name', flat=True).get(product_id=product_id)
+        except SQLProduct.DoesNotExist:
+            return product_id
 
     def get_headers(self, **kwargs):
         for product_id, section in self._column_tuples:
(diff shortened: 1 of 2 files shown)


mkangia (commented):
Why did this change now? 
looks like product_id and code are two different things.

emord:
@mkangia My guess was that this never actually worked. @esoergel may have more context as he's area owner for this

esoergel:
Ugh, okay, so there are a couple things going on here, I think.
* products have unique IDs (`product._id`) which are in some places are called "product_id"
* products also have a unique in the domain slug value called `product.code` which is confusingly referred to in the UI as "Product ID"
* This bit of functionality operates on a ledger section `entry_id`, which is usually a product UUID, but doesn't actually have to be.  The old behavior was expecting that.
* Looks like whoever set up this project saved ledgers to the product code instead of UUID, which I'd expect to break some other CommTrack functionality that expects a product UUID, such as reports.

I think I'd change this to check if there is a product with that ID (maybe also check that the domain matches), and if not, just return the `entry_id` itself the string representation.  That is - use the product name when you can, otherwise just whatever they put in `entry_id`.

You could also reach out to the project and tell them that they've probably misconfigured their app.

esoergel:
@emord I just found this precendent for the thing I was describing:
https://github.com/dimagi/commcare-hq/blob/master/corehq/ex-submodules/casexml/apps/case/templatetags/case_tags.py#L257-L261
(although I'd still advocate for restricting by domain)

That at least identifies that it's not a normal product.

mkangia:
Oops! thats messy.
Will defer to Ethan for review on this.

emord:
@esoergel updated and now using SQLProduct

corehq/apps/export/models/new.py
```diff
@@ -2507,7 +2507,10 @@ def _column_tuples(self):
         return section_and_product_ids
 
     def _get_product_name(self, product_id):
-        return Product.get(product_id).name
+        try:
+            return SQLProduct.objects.values_list('name', flat=True).get(product_id=product_id)
```
esoergel: Can you throw in a `domain=self.domain` here too?

esoergel: Hmm, actually I guess since it's checking product ID, it's pretty unlikely we'll get any conflicts, whatevs

esoergel (approved):

corehq/apps/export/models/new.py
emord: Nah. I like that idea. Especially since this may not actualy be a product id

esoergel (approved):

mkangia:
Build failure looks related

```
corehq.apps.export.tests.test_export_column:StockExportColumnTest.test_get_headers ... ERROR
ERROR: corehq.apps.export.tests.test_export_column:StockExportColumnTest.test_get_headers
  File "/vendor/lib/python2.7/site-packages/mock.py", line 1216, in patched
    arg = patching.__enter__()
  File "/vendor/lib/python2.7/site-packages/mock.py", line 1273, in __enter__
    self.target = self.getter()
  File "/vendor/lib/python2.7/site-packages/mock.py", line 1393, in <lambda>
    getter = lambda: _importer(target)
  File "/vendor/lib/python2.7/site-packages/mock.py", line 1126, in _importer
    thing = _dot_lookup(thing, comp, import_path)
  File "/vendor/lib/python2.7/site-packages/mock.py", line 1115, in _dot_lookup
    __import__(import_path)
ImportError: No module named Product
Ran 303 tests in 18.908s
```

PR #18264, opened 2017-10-24
Updated pillows to always retry on DocumentMissingError.

emord:
https://manage.dimagi.com/default.asp?263644#1408989

@sheelio @czue This should fix it for the future. Will still need to go back and resave the forms that were missing for whatever reason

@mkangia

### DIFF
--- a/corehq/ex-submodules/pillowtop/pillow/interface.py
+++ b/corehq/ex-submodules/pillowtop/pillow/interface.py
@@ -8,6 +8,7 @@ from corehq.util.timer import TimingContext
 from dimagi.utils.logging import notify_exception
 from kafka.common import TopicAndPartition
 from pillowtop.const import CHECKPOINT_MIN_WAIT
+from pillowtop.dao.exceptions import DocumentMissingError
 from pillowtop.utils import force_seq_int
 from pillowtop.exceptions import PillowtopCheckpointReset
 from pillowtop.logger import pillow_logging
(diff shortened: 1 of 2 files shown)


corehq/ex-submodules/pillowtop/pillow/interface.py
```diff
@@ -264,7 +265,8 @@ def fire_change_processed_event(self, change, context):
 def handle_pillow_error(pillow, change, exception):
     from pillow_retry.models import PillowError
     error_id = None
-    if pillow.retry_errors:
+    # always retry document missing errors, because the error is likely with couch
+    if pillow.retry_errors or isinstance(exception, DocumentMissingError):
```
mkangia: Is it possible to set a limit to this? 
Considering that if a document is actually missing (if even possible for such pillow processes) it would just keep retrying it?

mkangia (approved):
Thanks Emord.
Left one minor flag.

sheelio:
@emord 

Cool - at least for pradan-mis we can run the management command I wrote

czue (approved):
:clap: looks good to me! extra credit for incorporating into the test.

corehq/ex-submodules/pillowtop/pillow/interface.py
mkangia: Just meant that if a document is actually missing and not because of the situation we have here, this doc that needs to be processed will not be endlessly attempted.

snopoke: that's true but also very unlikely since the change feed comes from couchdb

PR #18267, opened 2017-10-24
more unicode bugs

nickpell:


emord (approved):

corehq/util/string_utils.py
```diff
@@ -5,3 +5,10 @@
 def random_string(n=6):
     # http://stackoverflow.com/a/23728630/835696
     return ''.join(random.SystemRandom().choice(string.ascii_uppercase + string.digits) for _ in range(n))
+
+
+def ensure_unicode(s):
+    if not s or isinstance(s, unicode):
```
millerdev: should use `six.text_type` instead of `unicode` for python 3 compatibility

nickpell: will follow up in a second PR

PR #18268, opened 2017-10-24
revert to django 1.10

nickpell:
@emord 

django-digest submodule can stay as is because its now backwards compatible.

Going to deploy to staging to confirm this is safe before merging.


nickpell:
successfully deployed to staging

emord (approved):

PR #18272, opened 2017-10-24
making ucr download location safe

calellowitz:
@esoergel

### DIFF
--- a/corehq/apps/userreports/reports/view.py
+++ b/corehq/apps/userreports/reports/view.py
@@ -27,7 +27,6 @@ from dimagi.utils.dates import DateSpan
 from dimagi.utils.modules import to_function
 from django.conf import settings
 from django.contrib import messages
-from django.urls import reverse
 from django.http import HttpRequest, HttpResponse, Http404, HttpResponseBadRequest, HttpResponseRedirect
 from django.utils.translation import ugettext as _, ugettext_noop
 from braces.views import JSONResponseMixin
@@ -63,6 +62,7 @@ from corehq.apps.userreports.util import (
     get_ucr_class_name)
 from corehq.util.couch import get_document_or_404, get_document_or_not_found, \
     DocumentNotFound
+from corehq.util.view_utils import reverse
 from couchexport.export import export_from_tables
 from couchexport.models import Format
 from dimagi.utils.couch.pagination import DatatablesParams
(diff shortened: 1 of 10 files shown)


corehq/apps/userreports/reports/view.py
```diff
@@ -668,18 +664,27 @@ def section_url(self):
...
+                'next_url': reverse(ConfigurableReport.slug, args=[self.domain, self.report_config_id]),
+                'next_url_text': _("Go back to report"),
+            })
+            return render(request, 'hqwebapp/soil_status_full.html', context)
+        else:
+            raise Http403()
+
+    @conditionally_location_safe(has_location_filter)
+    def dispatch(self, *args, **kwargs):
+        return super(DownloadUCRStatusView, self).dispatch(*args, **kwargs)
```
esoergel: I think you should be able to decorate the class itself, yeah?

esoergel: Hmm, actually it looks like `@conditionally_location_safe` doesn't do that - it would be pretty simple to add (see how `@location_safe` does it), but I'll leave it up to you.

corehq/apps/userreports/reports/view.py
```diff
@@ -668,18 +664,27 @@ def section_url(self):
...
+                                    params={'config_id': self.report_config_id}),
+                'title': _("Download Report Status"),
+                'progress_text': _("Preparing report download."),
+                'error_text': _("There was an unexpected error! Please try again or report an issue."),
+                'next_url': reverse(ConfigurableReport.slug, args=[self.domain, self.report_config_id]),
+                'next_url_text': _("Go back to report"),
+            })
+            return render(request, 'hqwebapp/soil_status_full.html', context)
+        else:
+            raise Http403()
```
esoergel: totally a style thing, but I'd do this at the top of the function - `if not has_permission: raise`, that way it's a special case that escapes early rather than a fork in the logic.  Same with the other one down below.

esoergel (approved):
Had a couple bits of non-blocking feedback, but feel free to ignore or deal with in a second PR

calellowitz:
Merging this now and will get to follow up in a separate pr to ensure this gets into today's deploy, especially in light of recent travis issues

esoergel (approved):

PR #18273, opened 2017-10-24
Mobile UCR v2 - Attempt 2

emord:
Cleaned up https://github.com/dimagi/commcare-hq/pull/17907 I rebased it on master and removed "fix" commits so that it's easier to go through commit by commit.

It exposes 3 types of mobile ucr restore versions. v1, v1.5, v2. v1 is the same as now. v2 is the schema from the spec. v1.5 sends down both. if it's an app aware sync it will send down whatever the app is configured to use (default 1.0). if it's not, it looks through the apps in the domain and will send v1 or v2 if all are the same. otherwise it sends both. The setting is hidden next to mobile ucr sync frequency in advanced settings -> advanced

This will send an empty fixture for "commcare:reports" if it's a v2 report. The only thing that isn't done, is purging deleted v2 reports on the phone. I don't think we need to block on that because we can always go back through released builds and figure out what old uuids existed.

@ctsims @snopoke This is on staging now. Happy to offline about any rollout plans.

buddy @mkangia


corehq/apps/app_manager/suite_xml/features/mobile_ucr.py
```diff
@@ -371,7 +374,7 @@ def _get_word_eval(word_translations, default_value):
         title=Text(
             locale=Locale(id=id_strings.report_data_table()),
         ),
-        fields=[_column_to_field(c) for c in config.report(domain).report_columns]
+        fields=[_column_to_field(c) for c in config.report(domain).report_columns if c.type != 'expanded']
```
emord: this just displays an empty data column on mobile v1, but it hard failed (xpath error) on v2. It didn't seem to be of much use in v1, so I am just removing it from the table

corehq/apps/app_manager/templates/app_manager/partials/mobile_report_configs.html
```diff
@@ -87,6 +88,9 @@
                             </div>
                         </td>
                         <td>
+                            <input type="number" class="form-control" data-bind="value: syncDelay, visible: $root.supportSyncDelay">
```
snopoke: could do with some help text

corehq/apps/app_manager/static/app_manager/json/commcare-app-settings.yaml
```diff
@@ -173,3 +173,11 @@
   values: ['none', '0', '1', '2', '4', '6', '8', '12', '24']
   value_names: ['Use project level setting', 'Update on every sync', '1 hour', '2 hours', '4 hours', '6 hours', '8 hours', '12 hours', '24 hours']
   default: 'none'
+
+- id: mobile_ucr_restore_version
+  name: Mobile UCR Restore Version
+  description: Version of mobile UCRs to use.
```
snopoke: Link to docs with more info would be handy for future

corehq/apps/app_manager/fixtures/mobile_ucr.py
```diff
@@ -49,25 +58,29 @@ def _should_sync(restore_state):
...
         restore_user = restore_state.restore_user
         if not toggles.MOBILE_UCR.enabled(restore_user.domain) or not _should_sync(restore_state):
-            return []
+            return False
 
         if toggles.PREVENT_MOBILE_UCR_SYNC.enabled(restore_user.domain):
-            return []
+            return False
+
+        apps = self._get_apps(restore_state, restore_user)
```
snopoke: Any concern that this will get called twice for each restore?

snopoke (approved):
Didn't review the XPath changes closely. Main comment is the need for some basic docs to ease future support burden.

corehq/blobs/tasks.py
```diff
@@ -8,7 +8,7 @@
 from corehq.blobs import get_blob_db
 
 
-@periodic_task(run_every=crontab(minute=0, hour='0,12'))
+@periodic_task(run_every=crontab(minute='*/5'))
```
millerdev: Curious why this was changed? I think it was deliberately set to be infrequent to keep load to a minimum. The guarantees on blob expiry are loose enough that this should not have been a problem, and nothing should depend on a blob actually being deleted on a strict schedule due to expiry.

emord: This adds caching to mobile UCRs for ICDS. I used the blobs because they seemed like the obvious way to cache XML.

> nothing should depend on a blob actually being deleted on a strict schedule due to expiry.

Is this a technical limitation with riak, or is it a limitiation that we have set up blob expirations? I could use set up some other model, but it seemed better to reuse this table.

millerdev: > I used the blobs because they seemed like the obvious way to cache XML.

That's a fine place to cache XML, but presence of a key in riak should not be used to tell if the cached value is valid or not.

There are some technicalities related to which version of a blob may be returned after its value has been changed or deleted. This is due to the eventual consistency design of riak. It's a performance trade-off.

The [`AbstractBlobDB.put` `timeout`](https://github.com/dimagi/commcare-hq/blob/0196adebab4c610e338e60055b0ba77d579fe27d/corehq/blobs/interface.py#L31-L33) parameter is documented as: "There are no guarantees for the maximum time it will live in blob storage."

emord: Thanks for the background. Sounds like I'll need to setup a new model or caching via redis or something to get caching working properly.

I've reverted those changes to this PR, so that this can move forward for enikshay. Can always follow up with caching later

emord:
Going to make one more change today or tomorrow to expose the report_uuid in the app manager. Will deploy this to enikshay on Monday so that the team can start thinking about refactoring hte app

corehq/apps/app_manager/fixtures/mobile_ucr.py
emord: With app-aware sync, this will get it from the restore_state https://github.com/dimagi/commcare-hq/blob/b1b4a5636cacba02a7925799deb22307a72afe74/corehq/apps/app_manager/fixtures/mobile_ucr.py#L78 

Clayton said that web apps now supports app aware sync (at least with enikshay), so I think this should be fine

corehq/apps/app_manager/static/app_manager/json/commcare-app-settings.yaml
emord: Writing up some docs for GS/enikshay team tomorrow about this. Will link once those are complete

emord:
@snopoke Resolved a conflict with your other branch https://github.com/dimagi/commcare-hq/pull/18273/commits/ec10a2b591d28e09729d00cf982baf1159ad7909

I think this is good to go. (it doesn't have the caching anymore and I'll update that in a separate PR)

corehq/apps/app_manager/static/app_manager/json/commcare-app-settings.yaml
emord: linked confluence page https://github.com/dimagi/commcare-hq/pull/18273/commits/ddfcff2dff97931b267df32064db62042cd48ebb

corehq/apps/app_manager/fixtures/mobile_ucr.py
```diff
@@ -163,37 +229,136 @@ def report_config_to_fixture(report_config, restore_user):
...
-            filters_elem.append(filter_elem)
-        return filters_elem
+    def report_config_to_v2_fixture(report_config, restore_user):
+        domain = restore_user.domain
+        report, data_source = BaseReportFixturesProvider._get_report_and_data_source(
+            report_config.report_id, domain)
+        report_fixture_id = 'commcare-reports:' + report_config.uuid
+        report_filter_id = 'commcare-reports-filters:' + report_config.uuid
+
+        # TODO: Convert to be compatible with restore_user
```
snopoke: Is this TODO still valid? Doesn't seem like it.

corehq/apps/app_manager/suite_xml/post_process/instances.py
```diff
@@ -146,6 +146,18 @@ def commcare_fixture_instances(domain, instance_name):
         return Instance(id=instance_name, src='jr://fixture/{}'.format(instance_name))
 
 
+@register_factory('commcare-reports')
+def commcare_reports_fixture_instances(domain, instance_name):
+    if instance_name.startswith('commcare-reports:') and toggles.MOBILE_UCR.enabled(domain):
```
snopoke: Why do we check the toggle here? Doesn't this only get called if the instance name is referenced in the app? If that's the case won't this break the app by not returning the instance?

I mean the app won't really work anyway since the fixture won't exists if the toggle isn't enabled but just thinking that might be an easier to debug failure mode.

corehq/apps/app_manager/templates/app_manager/partials/mobile_report_configs.html
```diff
@@ -60,6 +60,8 @@
                             <select class="form-control"
                                     data-bind="options: availableReportIds, optionsText: $root.defaultReportTitle, value: reportId">
                             </select>
+                            <label>Report UUID to use in custom expressions: </label>
+                            <span data-bind="text: uuid"></span>
```
snopoke: Nice.

corehq/apps/ota/tests/test_app_aware_restore.py
```diff
@@ -134,7 +134,7 @@ def test_report_fixtures_provider_with_app_that_doesnt_have_reports(self):
         with patch.object(ConfigurableReportDataSource, 'get_data') as get_data_mock:
             get_data_mock.return_value = self.rows
             fixtures = call_fixture_generator(report_fixture_generator, self.user, app=self.app3)
-        self.assertEqual(len(fixtures), 0)
+        self.assertEqual(len(fixtures), 1)
```
snopoke: How come this is now 1?

snopoke (approved):
Nothing blocking. Just a few questions.

snopoke:
For the caching I was wondering if instead of caching we should keep track of when each report was last sent down and just send it if it doesn't need to be refreshed?

That would save us caching a lot of data and would also reduce the payload size. It does mean we'd need to keep track of each report though (probably on the synclog I guess).

corehq/apps/app_manager/suite_xml/post_process/instances.py
emord: Good question. I think it could possibly be removed. I originally just made it hte same as https://github.com/dimagi/commcare-hq/blob/314fe154ff8bf1e2f15184f5911054e45e89ad17/corehq/apps/app_manager/suite_xml/post_process/instances.py#L144

corehq/apps/ota/tests/test_app_aware_restore.py
emord: Whenever we don't have a v1 report we send down an empty fixture. This is to purge the old v1 fixtures from phone

https://github.com/dimagi/commcare-hq/pull/18273/files/314fe154ff8bf1e2f15184f5911054e45e89ad17#diff-b00a086d998daf07fcf62372491b3ff3R157

PR #18274, opened 2017-10-24
Add tests w/ more standard headers

esoergel:
I was hoping to be able to use external libraries to assemble all the headers, then have Django parse those requests itself.  This isn't quite that, but it's close.
@proteusvacuum


corehq/apps/domain/tests/test_auth_type.py
```diff
@@ -174,3 +180,52 @@ def test_user_agent_beats_header(self):
...
+        )
+
+    def test_basic_auth(self):
+        request = self.get_django_request(auth=requests.auth.HTTPBasicAuth('foo', 'bar'))
+        self.assertEqual('basic', determine_authtype_from_request(request))
+
+    def test_digest_auth(self):
+        request = self.get_django_request(auth=requests.auth.HTTPDigestAuth('foo', 'bar'))
+        # FIXME This is wrong - this should return digest auth
+        self.assertEqual('basic', determine_authtype_from_request(request))
```
esoergel: @snopoke Any thoughts on this?  Yesterday @proteusvacuum and I figured out that anything using `login_or_digest_or_basic_or_apikey` or `login_or_digest_or_basic_or_apikey_or_token` (with the current default parameter of `BASIC`) doesn't actually work with digest auth.  Requests from a J2ME device and tastypie endpoints are special-cased to default to digest auth, so they do work fine, as far as we know.

The reason this doesn't work is that to use digest auth, the client first sends an unauthenticated request, and the server responds with a 401 status code and some parameters needed to construct a digest auth request.  Thus we have no way of knowing which auth method the client intends to use - we can't distinguish between an initial digest auth request and a normal unauthenticated request.

Defaulting to digest should work, as a basic auth request does include parameters on the first request.  Still, I'm not confident that there aren't odd use-cases that would break with a change like that.

esoergel: @emord FYI in case this falls under your security umbrella

snopoke: Interesting, that all makes sense but then how does this work for Android when it makes requests to endpoints like `search`, `claim`, `fetch_key_records` etc?

esoergel: `determine_authtype_from_request` explicitly returns `BASIC` when `"Android"` is in the user agent string, which I'm guessing means Android is always expected to use basic auth.

proteusvacuum:
@esoergel should we put this on staging and ask QA to make sure that it doesn't break anything? Or what do you see as the next steps here?

esoergel:
@proteusvacuum we could do that, but what kind of tests would you have in mind?  I guess we could just try accessing with various auth methods (basically what we did locally).  Phone requests shouldn't touch this logic, since J2ME and Android are special-cased.  I'd expect the remainder to be API hits from external clients, which wouldn't be pinging staging.

proteusvacuum:
I would at least want to test that a few different android devices do actually have the string `Android` in their useragent.

orangejenny:
Hey, this is breaking staging deploy because [this line](https://github.com/dimagi/commcare-hq/blob/master/corehq/apps/reports/views.py#L532) and [this line](https://github.com/dimagi/commcare-hq/blob/master/corehq/apps/reports/views.py#L542) pass an argument. I'm going to pull it off of staging.

esoergel:
This has been on staging for a while, and Ben went through some basic mobile test plan stuff with several different devices.  Should be ready to go.

gcapalbo (approved):

PR #18279, opened 2017-10-24
Use TEMP TABLE instead of CTE, also anti-join xid

millerdev:
After profiling it was determined that the performance of this solution is similar to the previous solution. When profiled on ICDS, the expensive part was the index scans on `form_processor_commcarecaseindexsql` (`domain = 'icds-cas' AND case_id = cid`; `domain = 'icds-cas' AND referenced_id = cid`; 93% of query time), which did not change with this optimization.

The second-most expensive part of the query is the join of `form_processor_commcarecaseindexsql` with `form_processor_commcarecasesql` (6% of query time), also not changed by this optimization.

It's possible that the anti-join (`LEFT JOIN ... WHERE xid IS NULL`) gives a small performance gain, although it was insignificant when other factors were considered.

Tool used for query analysis: http://tatiyants.com/pev/#/about

@snopoke

### DIFF
--- a/corehq/sql_accessors/sql_templates/get_related_indices.sql
+++ b/corehq/sql_accessors/sql_templates/get_related_indices.sql
@@ -6,20 +6,24 @@ CREATE FUNCTION get_related_indices(
     exclude_indices_array TEXT[]
 ) RETURNS SETOF form_processor_commcarecaseindexsql AS $$
 BEGIN
-    RETURN QUERY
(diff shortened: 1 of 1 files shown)


millerdev:
Query plans generated with http://tatiyants.com/pev/

Before:
screenshot 2017-10-25 15 53 09

After:
screenshot 2017-10-25 15 53 49

PR #18281, opened 2017-10-25
replace unicode with six.text_type

nickpell:


kaapstorm:
:recycle: :construction_worker_man: (timed out)

kaapstorm (approved):

corehq/util/string_utils.py
```diff
@@ -8,7 +9,7 @@ def random_string(n=6):
 
 
 def ensure_unicode(s):
-    if not s or isinstance(s, unicode):
+    if not s or isinstance(s, six.text_type):
```
millerdev: got carried away with Python 3 compatibility last time and didn't notice this `not s` part. That could cause unexpected things to be returned from this function since many things evaluate to false, but this function, according to its name, should always return unicode. What do you think of this instead:
```py
if isinstance(s, six.text_type):
    return s
elif isinstance(s, six.binary_type):
    return s.decode('utf-8')
return six.text_type(s)
```

nickpell: good idea - updated

kaapstorm (approved):

corehq/util/string_utils.py
```diff
@@ -12,5 +12,9 @@ def ensure_unicode(s):
     if isinstance(s, six.text_type):
         return s
     elif isinstance(s, six.binary_type):
-        return s.decode('utf-8')
+        try:
+            return s.decode('utf-8')
+        except UnicodeDecodeError:
+            # https://sentry.io/dimagi/commcarehq/issues/391378081/
+            return s.decode('latin1')
```
millerdev: Hmm, guessing at the encoding is not ideal.

A better fix for that bug would be to have [`get_username_and_password_from_request`](https://github.com/dimagi/commcare-hq/blob/c888c470f18661f8c5fede684c7545619a17cd2e/corehq/apps/domain/auth.py#L76) return correctly decoded strings as specified by the standard. Unfortunately, there is no standard character encoding due to the way the HTTP Basic Authentication standard was originally defined. [RFC-7617](https://tools.ietf.org/html/rfc7617#section-2.1) outlines a way for servers to request clients to use UTF-8, but there is no guarantee that this will be followed.

In any case, I think it would be better move this try/except guessing thing into `get_username_and_password_from_request` rather than putting it here in a general utility function that will get used all over the place.

millerdev (changes requested):
See comment in the code. In fact, this is leading me to doubt that we should have a function like `ensure_unicode` at all. There is no standard way to convert bytes to unicode without knowing the encoding. In most places, RFC-7617 notwithstanding, the character encoding is known if you trace the source of the bytes far enough, and that should always be preferred.

[In the face of ambiguity, refuse the temptation to guess.](https://www.python.org/dev/peps/pep-0020/)

PR #18287, opened 2017-10-25
Add option to include only specified loctypes in the restore

esoergel:
https://trello.com/c/unplaE7R/237-location-filtering-for-by-location-type
@proteusvacuum @calellowitz


corehq/apps/locations/views.py
```diff
@@ -373,8 +377,30 @@ def _attach_sync_boundaries_to_location_type(loc_type_data, loc_type_db):
...
         except KeyError:        # include_without_expanding location type was deleted
             loc_type.include_without_expanding = None
+        include_only = LocationTypesView._get_include_only(include_only_ids, loc_type_db)
+        loc_type.include_only.set(include_only)
         loc_type.save()
 
+    @staticmethod
+    def _get_include_only(include_only_ids, loc_type_db):
+        """The user specified that we include loc types `include_only_ids`, but
+        we need to insert any parent location types"""
```
esoergel: It would arguably be a nicer UX to do this in javascript, but this feels safer, and this is an advanced-user-only workflow

corehq/apps/locations/templates/locations/location_types.html
```diff
@@ -96,6 +96,16 @@
                                           {% endblocktrans %}">
                                     </span>
                                 </th>
+                                <th data-bind="visible: advanced_mode">
+                                    {% trans "Include only" %}
+                                    <span class="hq-help-template"
+                                          data-title="{% trans "Include only" %}"
+                                          data-content="{% blocktrans %}
+                                              <p>Include only locations of the specified types.</p>
+                                              <p>Every type you select must also have its parent included.</p>
```
proteusvacuum: Is there some way we should enforce this? 
Not having the parents included would also mean that we don't prefetch all the parents when compiling the XML, and will then do multiple queries to get the whole hierarchy of the location tree, making this pretty slow I think.

corehq/apps/locations/tests/test_location_fixtures.py
```diff
@@ -79,12 +79,16 @@ def _assemble_expected_fixture(self, xml_name, desired_locations):
...
     # and wont interfere with flat fixture generation
     @flag_enabled('HIERARCHICAL_LOCATION_FIXTURE')
-    def _assert_fixture_has_locations(self, xml_name, desired_locations, flat=False):
+    def _assert_fixture_matches_file(self, xml_name, desired_locations, flat=False):
         generator = flat_location_fixture_generator if flat else location_fixture_generator
         fixture = ElementTree.tostring(call_fixture_generator(generator, self.user)[-1])
         desired_fixture = self._assemble_expected_fixture(xml_name, desired_locations)
         self.assertXmlEqual(desired_fixture, fixture)
 
+    def assert_fixture_queryset_equals_locations(self, desired_locations):
```
proteusvacuum: nice.

corehq/apps/locations/templates/locations/location_types.html
esoergel: The server-side validation of this page automatically inserts the parent types if they're not specified.

PR #18289, opened 2017-10-25
Improve eNikshay duplicate IDs report

esoergel:
@calellowitz
I'd read commit-by-commit.  The first is a switch to bulk lookups for performance.  The second infers the 'real' device ID since it isn't provided in webapps.

### DIFF
--- a/custom/enikshay/reports/views.py
+++ b/custom/enikshay/reports/views.py
@@ -78,7 +78,7 @@ class DuplicateIdsReport(TemplateView):
                 case['device_number_in_form'] = form_device_number
                 case['form_device_id'] = form.metadata.deviceID
                 case['form_user_id'] = form.user_id
-                case['auth_user_id'] = form.auth_context.user_id
+                case['auth_user_id'] = form.auth_context.get('user_id')
 
         self.add_user_info_to_cases(bad_cases)
         context = {
@@ -96,8 +96,8 @@ class DuplicateIdsReport(TemplateView):
 
         auth_user_ids = [case['auth_user_id'] for case in bad_cases
                          if 'auth_user_id' in case]
-        auth_usernames = [user_doc['username'] for user_doc in
-                          iter_docs(CommCareUser.get_db(), auth_user_ids)]
+        auth_usernames = {user_doc['_id']: user_doc['username'] for user_doc in
+                          iter_docs(CommCareUser.get_db(), auth_user_ids)}
         for case in bad_cases:
             user_dict = user_info.get(case.get('form_user_id'))
             if user_dict:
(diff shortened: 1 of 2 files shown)


esoergel:
@calellowitz bump

esoergel:
Oh yeah, and troll: @proteusvacuum

custom/enikshay/reports/views.py
```diff
@@ -57,45 +59,65 @@ class DistrictLocationsView(LocationsView):
...
+
+    def get_cases_with_duplicate_ids(self, all_case_ids):
+        id_property = {'voucher': VOUCHER_ID, 'person': ENIKSHAY_ID}[self.case_type]
+        all_cases = [
+            {
+                'case_id': case.case_id,
+                'readable_id': case.get_case_property(id_property),
+                'opened_on': case.opened_on,
+            }
+            for case in self.accessor.iter_cases(all_case_ids)
```
proteusvacuum: wowza. Could you have used a document aggregation / top hits aggregation on the case search index for this instead? It might be a bit harder with the nested documents, but might mean that you don't have to pull every single case into memory.

custom/enikshay/reports/views.py
```diff
@@ -100,6 +90,29 @@ def get(self, request, domain, case_type, *args, **kwargs):
...
+        auth_user_ids = [case['auth_user_id'] for case in bad_cases
+                         if 'auth_user_id' in case]
+        auth_usernames = [user_doc['username'] for user_doc in
+                          iter_docs(CommCareUser.get_db(), auth_user_ids)]
+        for case in bad_cases:
+            user_dict = user_info.get(case.get('form_user_id'))
+            if user_dict:
+                case['username'] = user_dict['username']
+                device_id = case['form_device_id']
+                if device_id == 'Formplayer':
```
proteusvacuum: huh, does formplayer use different device ids for submissions / restores? I thought they were of the form `WebAppsLogin|{username}|as|{user}`. (which is kind of what you are creating when checking this value here, but with * instead of |).

esoergel: Formplayer does use different device IDs, unfortunately.  It used to be `|`, but that's been switched to `*`, I think because some place where that gets serialized plays nicer with `*`

custom/enikshay/reports/views.py
esoergel: Probably - this was purely a ease-of-implementation decision.  I will note that it doesn't pull them all into memory at once, just in chunks.  The speed it takes to generate this report doesn't matter at all (and the rest of this is also pretty slow), so I just haven't prioritized figuring out how to do that.

custom/enikshay/reports/views.py
proteusvacuum: huh, really? this doesn't actually break anything, but this testcase is no longer indicative of reality: 
https://github.com/dimagi/commcare-hq/blob/master/corehq/apps/ota/tests/test_app_aware_restore.py#L126

custom/enikshay/reports/views.py
proteusvacuum: ok

proteusvacuum (approved):

custom/enikshay/reports/views.py
esoergel: Yeah, and unfortunately making that switch essentially doubled the number of devices for eNikshay users (we keep track of those serially for ID generation)

PR #18290, opened 2017-10-26
Fix restore num items count with cached fixtures

millerdev:
https://manage.dimagi.com/default.asp?264086

@emord cc @snopoke

### DIFF
--- a/corehq/ex-submodules/casexml/apps/phone/utils.py
+++ b/corehq/ex-submodules/casexml/apps/phone/utils.py
@@ -12,7 +12,8 @@ def delete_sync_logs(before_date, limit=1000, num_tries=10):
     for i in range(num_tries):
         try:
             sync_log_ids = get_synclog_ids_before_date(before_date, limit)
-            return iter_bulk_delete_with_doc_type_verification(SyncLog.get_db(), sync_log_ids, 'SyncLog', chunksize=25)
+            return iter_bulk_delete_with_doc_type_verification(
+                SyncLog.get_db(), sync_log_ids, 'SyncLog', chunksize=25)
         except BulkSaveError:
             pass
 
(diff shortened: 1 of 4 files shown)


corehq/ex-submodules/casexml/apps/phone/utils.py
```diff
@@ -15,3 +15,35 @@ def delete_sync_logs(before_date, limit=1000, num_tries=10):
...
+        DIGITS = b'0123456789'
+        digits = []
+        i = len(ITEMS_COMMENT_PREFIX)
+        total_bytes = len(cached_bytes)
+        while i < total_bytes and cached_bytes[i] in DIGITS:
+            digits.append(cached_bytes[i])
+            i += 1
+        if digits and cached_bytes[i:i + 3] == b'-->':
+            num_items = int(b"".join(digits))
+            return cached_bytes[i + 3:], num_items
```
snopoke: Wouldn't it be easier to use a regex to get this out?

snopoke (commented):
Interesting approach. Is it not possible to store metadata with a blob in the BlobDB?

corehq/ex-submodules/casexml/apps/phone/utils.py
millerdev: I could do that, although regex engines are a bit of a black box and I didn't want it to end up doing something unexpected like searching backward from the end of the string. The time characteristics of this code are well defined and bounded for sane inputs.

On the other hand, I just remembered that Python has [`regex.match`](https://docs.python.org/3/library/re.html#re.regex.match) (I always forget that one), which appears to have been made for just this use case. I'll use that instead since it does make the code simpler. 

https://github.com/dimagi/commcare-hq/pull/18290/commits/815a163f63d113004914b673aa5bb5f463ff7487

millerdev:
> Is it not possible to store metadata with a blob in the BlobDB?

We don't have an API for that. I do see that S3 has support for user metadata, and it would be nice to use that intead, although then I also need to add metadata to fsdb... Seems like a lot of work for this one silly use case, not to mention changing the way cached fixtures are returned.

snopoke (approved):

emord (approved):

PR #18296, opened 2017-10-26
move already compressed block from include outside of compress

snopoke:
Not sure how this has been working during deploy. When running compress locally I get an error.

https://manage.dimagi.com/default.asp?263523


snopoke:
https://sentry.io/dimagi/commcarehq/issues/367953758/

millerdev (approved):

PR #18297, opened 2017-10-26
More tests

kkrampa:
@emord


custom/icds_reports/tests/reports/test_awc_infrastructure.py
```diff
@@ -0,0 +1,83 @@
+from django.test import TestCase
+
+from custom.icds_reports.reports.awc_infrastracture import get_awc_infrastructure_data
+
+
+class TestAWCInfrastructure(TestCase):
+
+    def test_data(self):
+        self.assertDictEqual(
+            get_awc_infrastructure_data(
```
emord: These tests would likely be easier to maintain if we just included important columns.

For example filtering the result to only include the values/percents and label. That way we don't need to change these columns for simple text changes

emord (approved):

PR #18298, opened 2017-10-26
improvements to two-factor reset workflow

snopoke:
* add warning on the UI with link to wiki
* add fields for auditing
* send email to admins and user


snopoke:
@dimagi/product marked this with `invisible` flag since it's HQ admins only. Let me know if that's not correct.

millerdev (approved):

sheelio:
@snopoke 
Invisible sounds right

PR #18299, opened 2017-10-26
Use month_end instead of month_start in alive_in_month

emord:
@sheelio Spec says this should use the month end for alive_in_month. https://manage.dimagi.com/default.asp?263652#1411633

Will need to do a partial rebuild. I think it should be sufficient to rebuild any indicators that have alive_in_month "no" for any month so that the previous months are calculated.

@mkangia


emord:
Closing after offlining with sheel that htis isn't actually needed

PR #18300, opened 2017-10-26
handle latin1 password in get_username_and_password_from_request

nickpell:
https://github.com/dimagi/commcare-hq/pull/18281/files#r147154523


corehq/apps/domain/auth.py
```diff
@@ -77,6 +76,12 @@ def get_username_and_password_from_request(request):
                 # decode password submitted from mobile app login
                 password = decode_password(password)
 
+                try:
+                    password = password.decode('utf-8')
```
millerdev: Should also decode the username here too. The encoding will be the same for both.

nickpell: updated

millerdev (approved):

PR #18304, opened 2017-10-26
define context

snopoke:


snopoke:
https://sentry.io/dimagi/commcarehq/issues/393040974/

PR #18307, opened 2017-10-26
Use new URL parameters for view_form

esoergel:
https://manage.dimagi.com/default.asp?264290
It looks like this was using the parameters provided to "view_form_legacy", so
the `reverse` call was failing.
@orangejenny @proteusvacuum @calellowitz


orangejenny (approved):

orangejenny:
Thanks, Ethan.

Introduced in https://github.com/dimagi/commcare-hq/pull/17076, which based on the date (and on this bug) I'm guessing was during the stretch where the module unique ids PR was in QA.

PR #18318, opened 2017-10-27
add get_person_case_from_prescription_item

nickpell:


kaapstorm (approved):

custom/enikshay/tests/utils.py
```diff
@@ -195,6 +195,29 @@ def get_prescription_case_structure(case_id, indexed_episode_id, extra_update=No
...
+    update.update(extra_update)
+    return CaseStructure(
+        case_id=case_id,
+        attrs={
+            "case_type": "prescription_item",
+            "create": True,
+            "update": update
+        },
+        indices=[CaseIndex(
+            CaseStructure(case_id=indexed_episode_id, attrs={"create": False}),
```
proteusvacuum: should this be the `prescription_id`?

proteusvacuum (changes requested):

custom/enikshay/tests/utils.py
nickpell: I'm following the pattern in ```get_prescription_case_structure``` where the host case id is passed as a function argument.  Although I should fix the argument name to be clear.

proteusvacuum (approved):

PR #18326, opened 2017-10-30
Refactored ICDS aggregation script for smaller transactions.

emord:
@sheelio This keeps the aggregation steps in the same order, but separates them out into different transactions. This should allow for much less lockingof the database when the aggregation is happening (currently it's pretty much useless overnight because so many tables are dropped/created. Eventually it would be good to break these out into separate tasks and have some of them work in parallel (like inserting from UCR tables)

buddy @mkangia


sheelio:
Looks good to me (guessing the smaller transactions are better for locking?)

emord:
Yeah. when it's all one transaction all tables will be locked for the duration of the transaction (because they are dropped/recreated) With this change they should only be locked for the duration that it takes to recreate them and only one at a time

mkangia (approved):

PR #18330, opened 2017-10-30
Don't check if the 99DOTS registration actually succeeded

proteusvacuum:
99DOTS sends a lot of timeout errors, but the case actually succeeds
registration. However, when this happens, Treatment Outcomes and Adherences
aren't sent along to 99D

@esoergel @calellowitz


esoergel (approved):
Yuck, that's annoying.  Do you think it's worth adding something in to at least check whether registration was attempted?

proteusvacuum:
Partially solved by running: https://github.com/dimagi/commcare-hq/pull/18428
Working w/ 99DOTS to resolve this more conclusively on their side

PR #18332, opened 2017-10-30
Undelete system forms on user.unretire()

esoergel:
https://manage.dimagi.com/default.asp?264402
@calellowitz @snopoke

Opening for discussion

### DIFF
--- a/corehq/apps/users/models.py
+++ b/corehq/apps/users/models.py
@@ -49,7 +49,8 @@ from corehq.apps.users.util import (
     username_to_user_id,
 )
 from corehq.apps.users.tasks import tag_forms_as_deleted_rebuild_associated_cases, \
-    tag_cases_as_deleted_and_remove_indices, tag_system_forms_as_deleted
+    tag_cases_as_deleted_and_remove_indices, tag_system_forms_as_deleted, \
+    undelete_system_forms
 from corehq.apps.sms.mixin import CommCareMobileContactMixin, apply_leniency
 from dimagi.utils.couch.undo import DeleteRecord, DELETED_SUFFIX
 from corehq.apps.hqwebapp.tasks import send_html_email_async
@@ -1673,6 +1674,8 @@ class CommCareUser(CouchUser, SingleMembershipMixin, CommCareMobileContactMixin)
 
         deleted_case_ids = self._get_deleted_case_ids()
         CaseAccessors(self.domain).soft_undelete_cases(deleted_case_ids)
+
+        undelete_system_forms(self.domain, set(deleted_form_ids), set(deleted_case_ids))
         self.save()
 
     def retire(self):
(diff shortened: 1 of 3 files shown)


corehq/apps/users/models.py
```diff
@@ -1673,6 +1674,8 @@ def unretire(self):
 
         deleted_case_ids = self._get_deleted_case_ids()
         CaseAccessors(self.domain).soft_undelete_cases(deleted_case_ids)
+
+        undelete_system_forms(self.domain, set(deleted_form_ids), set(deleted_case_ids))
```
esoergel: Could this have unintended behavior?  Could there be forms which were intentionally deleted that will now be reinstated?

corehq/apps/users/tasks.py
```diff
@@ -123,33 +123,49 @@ def tag_forms_as_deleted_rebuild_associated_cases(user_id, domain, form_id_list,
...
-    for case_id in deleted_cases:
+def _get_forms_to_modify(domain, modified_forms, modified_cases, is_deletion):
+    """Used on user.retire() and user.unretire()
+
+    Returns a list of IDs of forms which only modify the cases passed in and
+    which aren't already listed in `modified_forms`.
+    """
+    form_ids_to_modify = set()
+    for case_id in modified_cases:
         xform_ids = CaseAccessors(domain).get_case_xform_ids(case_id)
```
esoergel: The tests I wrote currently fail on the couch backend because the couch version of this method doesn't return the system forms.  Any ideas on how to get those for couch backends?

corehq/apps/users/models.py
snopoke: I don't think so.

should this be called with `.delay()`?

corehq/apps/users/tasks.py
snopoke: Strange, not sure why that would be. The list of forms is pulled from a couch view so I'd check the view to see what's going on (`form_case_index/form_case_index`)

snopoke (approved):

corehq/apps/users/models.py
esoergel: doh, yes it should, updated.

corehq/apps/users/tasks.py
esoergel: Looks like that view excludes `XFormInstance-Deleted` documents, whereas the corresponding SQL accessor _does_ include deleted results.  That's an unfortunate discrepancy.  I'm not sure if it's possible to get the deleted forms which updated a given case in a performant fashion.  That would probably require a new couch view (or a modification to this one).  I'm inclined to merge this change as-is, so it'll work for SQL domains, at least, although I could be talked out of that if you think the lack of parity is a blocker.

snopoke:
Can you not use `case.xform_ids`

esoergel:
Awesome, that works, thanks @snopoke

PR #18334, opened 2017-10-30
Mobile UCR v2 caching

emord:
This will allow UCRs to be cached independently of each other. It'll be specified on the reports module when ucr v2 is configured

image

This is going into my other branch. Will move this back to master once that's merged
buddy @mkangia

### DIFF
--- a/corehq/apps/app_manager/fixtures/mobile_ucr.py
+++ b/corehq/apps/app_manager/fixtures/mobile_ucr.py
@@ -2,10 +2,13 @@ from collections import defaultdict
 from datetime import datetime
 import logging
 import random
+from xml.etree import cElementTree as ElementTree
 
 from couchdbkit.exceptions import NoResultFound
 from django.conf import settings
+from django.core.cache import cache
 from lxml.builder import E
+from six.moves import cStringIO as StringIO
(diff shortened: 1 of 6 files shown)


mkangia (approved):
Looks good.
Do you plan to add more tests for the changes?

emord:
> For the caching I was wondering if instead of caching we should keep track of when each report was last sent down and just send it if it doesn't need to be refreshed?
>
> That would save us caching a lot of data and would also reduce the payload size. It does mean we'd need to keep track of each report though (probably on the synclog I guess).

Agreed that this could be a better approach. When I was thinking about this I was trying to figure out the best way to do this and I think we'd need to keep a dictionary of {ucr: date last synced} in the synclog and copy that over to the new synclog every sync. Does that seem better than using redis/riak?

snopoke:
@emord I think it makes sense to do it that way (using the synclogs):
* reduces load on riak
* less data to cleanup
* less work to do during restore (don't have to put / pull data out of riak)
* reduces amount of data sent to phone
* added complexity is fairly minimal

emord:
👍 closing this out and will re-work it

PR #18336, opened 2017-10-30
Update paywall text for report builder

biyeun:
@esoergel

### DIFF
--- a/corehq/apps/userreports/views.py
+++ b/corehq/apps/userreports/views.py
@@ -388,15 +388,15 @@ class ReportBuilderPaywallPricing(ReportBuilderPaywallBase):
     @property
     def page_context(self):
         context = super(ReportBuilderPaywallPricing, self).page_context
-        if has_report_builder_access(self.request):
-            max_allowed_reports = allowed_report_builder_reports(self.request)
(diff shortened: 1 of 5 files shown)


corehq/apps/userreports/views.py
```diff
@@ -266,12 +266,16 @@ def dispatch(self, request, *args, **kwargs):
...
-            'at_report_limit':
-                number_of_report_builder_reports(self.domain) >= allowed_report_builder_reports(self.request),
-            'report_limit': allowed_report_builder_reports(self.request),
+            'at_report_limit': (
+                number_of_report_builder_reports(self.domain) >= allowed_num_reports
+                and allowed_num_reports is not None
+            ),
+            'report_limit': allowed_num_reports,
             'paywall_url': paywall_home(self.domain),
+            'pricing_page_url': reverse('public_pricing') if settings.ENABLE_PRELOGIN_SITE else "",
```
esoergel: Won't this just be a dead link if `not settings.ENABLE_PRELOGIN_SITE`?  I guess that only happens on single-domain environments.

biyeun: yeah good point. i should just hide that link

esoergel (approved):

PR #18337, opened 2017-10-30
Rollout V2 of report builder to everyone

biyeun:
@esoergel


corehq/toggles.py
```diff
@@ -497,13 +497,13 @@ def toggle_values_by_name(username=None, domain=None):
...
+REPORT_BUILDER_V1 = StaticToggle(
+    "report_builder_v1",
+    "Report builder V1",
     TAG_PRODUCT,
     [NAMESPACE_DOMAIN],
     description=(
-        'Enables the "new" report builder. Note that the project must already have access to report builder'
+        'Enables the old report builder. Note that the project must already have access to report builder.'
     )
 )
```
esoergel: Is there some accompanying migration to put some domains on report builder v1?  What's the use-case for keeping it around?

biyeun: working on the migration right now. use case is that it can be an emergency fallback in case some projects have an issue with v2, and then we delete v2 in two weeks once we resolve any of those issues. similar to how the app builder migration happened

esoergel: offlined with Biyeun, the migration is something else.  Keeping this around for 2 weeks makes a lot of sense, I agree with your reasoning

esoergel (approved):
Now that's how you get your LOC numbers into the red!

biyeun:
> Now that's how you get your LOC numbers into the red! 

🎉

PR #18338, opened 2017-10-30
log error before creating PillowError

nickpell:
https://manage.dimagi.com/default.asp?260368#BugEvent.1390660

@proteusvacuum @emord

### DIFF
--- a/corehq/ex-submodules/pillowtop/pillow/interface.py
+++ b/corehq/ex-submodules/pillowtop/pillow/interface.py
@@ -265,6 +265,16 @@ class ConstructedPillow(PillowBase):
 def handle_pillow_error(pillow, change, exception):
     from pillow_retry.models import PillowError
     error_id = None
+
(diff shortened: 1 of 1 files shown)


emord (commented):
I think that this sort of error should crash the pillow, and we shouldn't make any progress on the kafka queue until the main db is in a good state again.

proteusvacuum (commented):
Won't this log the same error twice? What if you just caught the exception, and set `error_id` to "error creating PillowError" or something.

nickpell:
> What if you just caught the exception, and set error_id to "error creating PillowError" or something.

updated

nickpell:
> I think that this sort of error should crash the pillow, and we shouldn't make any progress on the kafka queue until the main db is in a good state again.

I think so too.  I'd like to log the doc id just to help understand if the bug referenced above is somehow related to these failures or not.

kaapstorm (approved):

PR #18340, opened 2017-10-30
add test tasklist datasource with nullable tu_id and dto_id columns

nickpell:

Same as test_tasklist_v3.json except tu_id and dto_id are marked as nullable.

code buddy @kaapstorm

### DIFF
--- a/custom/enikshay/ucr/data_sources/test_tasklist_v4.json
+++ b/custom/enikshay/ucr/data_sources/test_tasklist_v4.json
@@ -1,0 +1,1284 @@
+{
+    "domains": [
+        "enikshay",
+        "enikshay-performance-test"
+    ],
+    "server_environment": [
+        "enikshay"
+    ],
+    "config": {
+        "referenced_doc_type": "CommCareCase",
+        "asynchronous": true,
+        "engine_id": "ucr",
+        "description": "updated to include combined diagnostic and followup TLs",
+        "base_item_expression": {},
+        "table_id": "test_tasklist_v4",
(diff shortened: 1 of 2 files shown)


calellowitz (changes requested):
I think you will want to talk to the project team about this. They may be purposefully excluding null values, in which case we should be changing the filters to properly exclude them, not making the column more permissive.

kaapstorm:
Pending product team feedback, seems fine to me otherwise. 

:recycle: :construction_worker_man: Travis got stuck.

nickpell:
@calellowitz I wrote this data source, and wasn't intentional about setting is_nullable to true or false.  This is only used by the task list reports, which always filter by location, so none of the cases that error will ever appear to an end user.  So this change should fix the error and speed up rebuild times without affecting the end user.

emord:
fwiw it's usually easier to view these changes if you have one commit that copies an existing report to a new file and then make changes in a second commit.

Generally agree about adding better filters, but I also think it would just be good to have these data sources not erroring.

If you don't want to do a rebuild, once this is deployed I think you could just do a `insert into <new_table> select *  from <old_table>)` or something similar. 

Also you need to add this to settings.py

nickpell:
@emord Thanks for flagging settings.py, updated

nickpell:
> fwiw it's usually easier to view these changes if you have one commit that copies an existing report to a new file and then make changes in a second commit.

good to know, thanks

calellowitz (approved):

emord (approved):

PR #18344, opened 2017-10-31
Respect the `wrap` parameter

kaapstorm:
@nickpell @snopoke


kaapstorm:
For a little context @nickpell this fixes a bug where `FormProcessorCouch.hard_rebuild_case()` was expecting `FormProcessorCouch.get_case_with_lock()` to return a `CommCareCase` instance, but instead was getting a dict.

snopoke:
I think you still need to pass in `wrap=True` to the call in `FormProcessorCouch.hard_rebuild_case()`

kaapstorm:
:white_check_mark:

snopoke (commented):
Since I wrote some of this I'll let @nickpell approve

PR #18352, opened 2017-10-31
Report Builder Analytics

biyeun:
@mkangia 
cc @esoergel 

fyi @sheelio

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -45,6 +45,7 @@
     "knockout": "3.3.0",
     "knockout-validation": "2.0.3",
     "underscore": "1.8.3",
+    "underscore.string": "3.3.4",
     "leaflet": "0.7.7",
     "select2-3.4.5-legacy": "select2#3.4.5",
     "select2-3.5.2-legacy": "select2#3.5.2",
--- a/bower.json
+++ b/bower.json
@@ -45,7 +45,6 @@
     "knockout": "3.3.0",
     "knockout-validation": "2.0.3",
     "underscore": "1.8.3",
-    "underscore.string": "3.3.4",
     "leaflet": "0.7.7",
     "select2-3.4.5-legacy": "select2#3.4.5",
     "select2-3.5.2-legacy": "select2#3.5.2",
--- a/corehq/apps/hqwebapp/static/hqwebapp/js/main.js
+++ b/corehq/apps/hqwebapp/static/hqwebapp/js/main.js
@@ -280,6 +280,10 @@ hqDefine("hqwebapp/js/main", function() {
         });
     });
 
+    var capitalize = function(string) {
+        return string.charAt(0).toUpperCase() + string.substring(1).toLowerCase();
+    };
+
     return {
         beforeUnloadCallback: beforeUnloadCallback,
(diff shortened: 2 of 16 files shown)


biyeun:
Note: the analytics libs we have here are super disorganized. I'm going to take a pass at creating a better utility for it once I get all the basic changes in for report builder. FYI @orangejenny

orangejenny:
> Note: the analytics libs we have here are super disorganized. I'm going to take a pass at creating a better utility for it once I get all the basic changes in for report builder.

sounds great

bower.json
```diff
@@ -45,6 +45,7 @@
     "knockout": "3.3.0",
     "knockout-validation": "2.0.3",
     "underscore": "1.8.3",
+    "underscore.string": "3.3.4",
```
orangejenny: Are you using this for anything other than `capitalize`? I just skimmed the PR. We have a [local version](https://github.com/dimagi/commcare-hq/blob/master/corehq/apps/export/static/export/js/utils.js#L34).

biyeun: 🤔  not sure if that's the most intuitive place for it lol... exports?

orangejenny: Exports is the only place it was being used. Would make sense to move it someplace like main.js and use it here.

biyeun: Also, this doesn't do quite the same thing. This assumes that a word is passed in, but `s.capitalize` handles multiple words separated by a space. That case doesn't really apply to the report builder analytics, so I moved capitalize into `main.js` and removed `underscore.string` anyway. But the `underscore.string` library is way more powerful. I would prefer to have that available, but maybe later.

orangejenny: Yeah, let's talk about it when there's a stronger definite need for it. It might make sense to replace underscore with lodash at that point rather than adding underscore.string.

biyeun:
@orangejenny I don't want to wrestle with the breaking tests for moving capitalize outside of the exports util. That seems out of scope for this PR and could potentially affect a different part of the system.

biyeun:
I'M COMING AT YOU WITH A NEW WORLD OF HQ ANALYTICS. WAIT FOR IT.

biyeun:
@orangejenny I went ahead and just moved capitalize into `main.js` while i wait on the other analytics changes so that this gets in

corehq/apps/hqwebapp/static/hqwebapp/js/main.js
```diff
@@ -280,6 +280,10 @@ hqDefine("hqwebapp/js/main", function() {
         });
     });
 
+    var capitalize = function(string) {
+        return string.charAt(0).toUpperCase() + string.substring(1).toLowerCase();
```
orangejenny: Might want to check for `string` being blank.

orangejenny: nevermind, just tested - I thought `charAt` would fail but it's fine

orangejenny:
@biyeun thanks!

mkangia (approved):
build failed due to Couch fetch fail. restarted failed jobs.

PR #18355, opened 2017-10-31
Add transifex-client to dev requirements

esoergel:
@dannyroberts

### DIFF
--- a/requirements/dev-requirements.txt
+++ b/requirements/dev-requirements.txt
@@ -19,6 +19,6 @@ ipdb==0.9.4
 wheel==0.29.0
 graphviz==0.7
 git+git://github.com/nickpell/django-package-monitor@np/handle-no-semver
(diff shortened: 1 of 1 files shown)


requirements/dev-requirements.txt
```diff
@@ -19,5 +19,6 @@ ipdb==0.9.4
 wheel==0.29.0
 graphviz==0.7
 git+git://github.com/nickpell/django-package-monitor@np/handle-no-semver
+transifex-client>=0.12.4
```
dannyroberts: worth pinning so we know the same thing's on everyone's machine?

dannyroberts (approved):

requirements/dev-requirements.txt
esoergel: Hmm, I guess it IS scripted, so that might be wise.

PR #18357, opened 2017-10-31
Fix Deploy Emails

sheelio:
@proteusvacuum 
This fixes the colour of the labels in the email, and properly displays PRs by label (django templates can't iterate a defaultdict).  

We also want to send out the email to something like daily-announce@dimagi.com.  Is that easy to do? From looking at the code, it looks like its sent to the site admins only?

### DIFF
--- a/corehq/apps/hqadmin/management/commands/record_deploy_success.py
+++ b/corehq/apps/hqadmin/management/commands/record_deploy_success.py
@@ -8,6 +8,7 @@ from django.core.management.base import BaseCommand
 from corehq.apps.hqadmin.models import HqDeploy
 from datetime import datetime, timedelta
 from django.conf import settings
+from corehq.util.log import send_HTML_email
 
 from dimagi.utils.parsing import json_format_datetime
 from pillow_retry.models import PillowError
@@ -133,6 +134,14 @@ class Command(BaseCommand):
                 environment=options['environment'], user=options['user'],
                 compare_url=compare_url)
             call_command('mail_admins', message_body, **{'subject': 'Deploy successful', 'html': True})
+            try:
+                recipient = settings.DAILY_DEPLOY_EMAIL
+                subject = '[%] Deploy Successful'.format(options['environment'])
+                send_HTML_email(subject=subject,
+                                recipient=recipient,
+                                html_content=message_body)
+            except Exception:
+                pass
 
         if settings.SENTRY_CONFIGURED and settings.SENTRY_API_KEY:
             create_update_sentry_release()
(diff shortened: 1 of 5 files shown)


corehq/apps/hqadmin/management/utils.py
```diff
@@ -100,7 +100,7 @@ def _get_prs_by_label(pr_infos):
         for label in pr['labels']:
             if label['name'] in LABELS_TO_EXPAND:
                 prs_by_label[label['name']].append(pr)
-    return prs_by_label
+    return dict(prs_by_label)
```
snopoke: what's the need for this? It's already a dict

snopoke:
To change who it's sent to you use `send_mail` instead of calling the `mail_admins` management command here: https://github.com/dimagi/commcare-hq/blob/754ae7d1c7d8e8862c589989f8a7ba91ce4938e9/corehq/apps/hqadmin/management/commands/record_deploy_success.py#L135

Not sure why that's calling a management command instead of just using `mail_admins` function.

corehq/apps/hqadmin/management/utils.py
sheelio: Django templates can't iterate over a defaultdict (we display this in the email), so had to switch to a normal dict

snopoke: interesting, didn't know that

proteusvacuum (approved):

sheelio:
@proteusvacuum 
I've also setup the ability to email tech-announce-daily@dimagi.com (through the DAILY_DEPLOY_EMAIL localsetting).  Are you able to add that same setting to our different environments?  I'm not sure how to do that.

corehq/apps/hqadmin/management/commands/record_deploy_success.py
```diff
@@ -133,6 +134,14 @@ def handle(self, **options):
                 environment=options['environment'], user=options['user'],
                 compare_url=compare_url)
             call_command('mail_admins', message_body, **{'subject': 'Deploy successful', 'html': True})
+            try:
+                recipient = settings.DAILY_DEPLOY_EMAIL
+                subject = '[%] Deploy Successful'.format(options['environment'])
+                send_HTML_email(subject=subject,
+                                recipient=recipient,
+                                html_content=message_body)
+            except Exception:
```
proteusvacuum: This seems like a very large hammer. I would have expected something like `except AttributeError` to be the only permissible error here (if the setting is not configured)

localsettings.example.py
```diff
@@ -99,6 +99,7 @@
 SERVER_EMAIL = 'commcarehq-noreply@example.com'  # the physical server emailing - differentiate if needed
 DEFAULT_FROM_EMAIL = 'commcarehq-noreply@example.com'
 SUPPORT_EMAIL = "commcarehq-support@example.com"
+DAILY_DEPLOY_EMAIL = "tech-announce-daily@example.com"
```
proteusvacuum: We currently have all our other email addresses in `settings.py`, which gets automatically added to all the environments. I don't think that's necessarily the best idea, but I think good to follow convention for now so that everything is in one place for when that actually gets refactored.

sheelio:
@proteusvacuum 
Thanks for the feedback - updated.

PR #18358, opened 2017-11-01
Add py3k linter

millerdev:
Also purposely made a change to a non-py3k compatible file. Will fix once I've verified that the py3k linter is working.

### DIFF
--- a/corehq/ex-submodules/auditcare/utils/show_urls.py
+++ b/corehq/ex-submodules/auditcare/utils/show_urls.py
@@ -1,5 +1,6 @@
 # from http://code.activestate.com/recipes/576974-show-all-url-patterns-in-django/
 # prints a tree of all the urls in the project
+from __future__ import print_function
 from types import FunctionType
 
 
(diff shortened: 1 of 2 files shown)


corehq/ex-submodules/auditcare/utils/show_urls.py
```diff
@@ -1,5 +1,6 @@
 # from http://code.activestate.com/recipes/576974-show-all-url-patterns-in-django/
 # prints a tree of all the urls in the project
+from __future__ import print_function
```
millerdev: @dimagi/team-commcare-hq do this if you want to use `print(...)` and don't want @stickler-ci to complain about it.

kaapstorm (approved):
Nice! I really like where this is going!

emord:
@dimagi/team-commcare-hq here's a list of the py3k warnings you may get and a quick blurb about each one https://pylint.readthedocs.io/en/latest/technical_reference/features.html#id25

PR #15876, opened 2017-04-25
stop checking main DB for synclogs

snopoke:
buddy @nickpell 
@dimagi/scale-team 

I think it's been log enough that there won't be any active synclogs in the main DB. 

related to: https://trello.com/c/acH1iJxe/21-start-saving-synclogs-to-a-new-db-but-continue-checking-the-old-one

PR #18360, opened 2017-11-01
support multiple synclog DBs

snopoke:
Will require creating the new DB prior to deploy.


corehq/ex-submodules/pillowtop/migrations/0004_rename_synclog_checkpoint.py
```diff
@@ -0,0 +1,23 @@
...
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('pillowtop', '0003_auto_20170411_1957'),
+    ]
+
+    operations = [
+        # in case there's a rollback we want to override any existing value
+        RunSQL("DELETE FROM pillowtop_djangopillowcheckpoint WHERE checkpoint_id = 'synclogs'"),
```
emord: I think you'll want synclogs_2017-11-01 as the checkpoint id.

I think there are ways to use the models instead of raw sql. any reason you chose to use raw sql here?

emord (commented):
I think this looks ok except for my one comment.

Are you concerned at how manual this is? From what I can tell the steps here are:

for every environment:
1) create a new database with a particular name
2) change the localsettings to point to the new db
3) restart services
4) wait til the other db is empty and remove it

Do you think we could change this code to support x number of synclog dbs, and then have them roll over on a schedule (once a month or something)? Then we could create a script that cleans up old ones later (similar to pruning indices in ES)

Also is there a reason you didn't go with filtered replication? I don't think it would have required any code changes, just an occasional change of the db's name in localsettings. something like https://eclipsesource.com/blogs/2015/04/20/how-to-finally-delete-documents-in-couchdb/

corehq/ex-submodules/pillowtop/migrations/0004_rename_synclog_checkpoint.py
snopoke: I was already in SQL mode from writing the query below. No other reason.

This migration isn't strictly necessary since the new checkpoint will get created automatically, this is more or a backup but not that I think about it more I don't think it's that helpful in any situation.

snopoke:
@emord I'm open to discussion on this but responses below:

> Do you think we could change this code to support x number of synclog dbs, and then have them roll over on a schedule (once a month or something)? Then we could create a script that cleans up old ones later (similar to pruning indices in ES)

Any process we do is going to be manual (at least on prod since we can't auto-create databases on Cloudant). Creating X DBs up front is an option I suppose but then you'd have to check to make sure that the next DB existed before rolling over. Creating the DB's would still be manual. Also having more DB's means doing more requests for each couch query.

> Also is there a reason you didn't go with filtered replication? I don't think it would have required any code changes, just an occasional change of the db's name in localsettings. something like https://eclipsesource.com/blogs/2015/04/20/how-to-finally-delete-documents-in-couchdb/

We used to do this exact thing when we moved synclogs out of the main DB (https://github.com/dimagi/commcare-hq/pull/15876). I figured it would be easy enough to add back.

The nice thing about the replication is that there's no query performance hit. But you do need to be able to guarantee that the flip to the new DB happens when the new DB is up to date with the old one. 

It would still be a manual process and one that requires more intervention from us I think.

But like I said, I'm keen to chat about this since it is quite a big change.

emord:
> Any process we do is going to be manual (at least on prod since we can't auto-create databases on Cloudant)

There's no api to create dbs? not even `PUT /commcarehq__synclogs_new` ?

> But you do need to be able to guarantee that the flip to the new DB happens when the new DB is up to date with the old one.

That's true. I think we'd also need to be understand our disk space issues more. Like how much more space would storing that data twice be. That might be a non-starter for ICDS until we get the new hardware

Seems like you've thought this through, and I think it would be better to have this as you've implemented than having no way to cycle these DBs. Was mostly hoping we could just make this much more automated

snopoke:
I'm a little hesitant to roll this out in case it has adverse performance effects on ICDS. There's no real way to do a soft rollout either.

emord:
Agree this change is scary, but I think overall it may be worth it right? It should significantly help in the long run to get rid of the deleted docs

emord:
Maybe this is a crazy thought, but could we use this as a jumping off point to move this to postgres?

I think we'd just need to create the model and overwrite gets and saves to prefer the postgres before couch. This could go away after 2 months since we prune the synclogs

We could put this in a separate DB so that we can scale it to not have it with the main DB (should probably do that for device logs). I'm not sure how we'd fit this on current ICDS infrastructure though.

We could use architect to partition the table. (likely by date, but we could also partition it by the UUID if we want to prioritize fast reads)

We'd need to do something a little more complicated than synclog_view, but it'd be feasible.

It would remove the need to create these DBs in the future.

Separately maybe we should start logging the date of the synclog we're getting? Maybe we don't need to keep 60 days of synclogs around if no one is using them? Could make it a localsetting

snopoke:
That's an interesting idea. I we'd need to think that through some more but we could use a similar approach to this. 

This is ready to go, I created the DB on cloudant. The only thing holding me back from merging is that there's no easy way to roll back (without causing a hole bunch of 412s).

emord:
> The only thing holding me back from merging is that there's no easy way to roll back (without causing a hole bunch of 412s).

We wouldn't have 412s if we replicated from the new db back to the old db right? Although automating that to all of our clusters would be a pain.

Do we also need to manually create the DB for all the other environments?

snopoke:
I'm 99.9% sure all other envs will auto create the DB (also tested that theory on Softlayer which worked fine).

I think I'll make this configurable from localsettings and then I can roll it out in a more controlled fashion.

snopoke:
Feeling very uninspired today. Not touching this for now.

snopoke:
Closing this in favor of replication to a new DB

PR #18361, opened 2017-11-01
better handling of conflicts in app file download

snopoke:
fixes COMMCAREHQ-5H8 (https://sentry.io/dimagi/commcarehq/issues/396462129/)

### DIFF
--- a/corehq/apps/app_manager/views/download.py
+++ b/corehq/apps/app_manager/views/download.py
@@ -238,6 +238,19 @@ def download_file(request, domain, app_id, path):
         return RegexURLResolver(
             r'^', 'corehq.apps.app_manager.download_urls').resolve(path)
 
+    def create_build_files_if_necessary_handling_conflicts(is_retry=False):
+        try:
+            try:
+                # look for file guaranteed to exist if profile is created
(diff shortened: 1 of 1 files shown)


corehq/apps/app_manager/views/download.py
```diff
@@ -238,23 +238,28 @@ def resolve_path(path):
...
-                try:
-                    # look for file guaranteed to exist if profile is created
-                    request.app.fetch_attachment('files/{id}/profile.xml'.format(id=build_profile))
-                except ResourceNotFound:
-                    request.app.create_build_files(save=True, build_profile_id=build_profile)
-                    request.app.save()
-                    payload = request.app.fetch_attachment(full_path)
-                else:
-                    # if profile.xml is found the profile has been built and its a bad request
-                    raise
```
emord: Is it ok that this isn't handled anymore?

snopoke: If this situation arises the `ResourceNotFound` error will still get raised from line 262 (2 lines down). It will just be one extra request to Riak that we could have avoided but I don't think that's an issue since it's a real edge case

PR #18364, opened 2017-11-01
remove test_tasklist_v2

nickpell:

There are no more reports pointed at this data source

@calellowitz @emord


emord (approved):

calellowitz (changes requested):
I dont think that is true. And if it is, it shouldnt be. There were two reports pointed to that that did not work on v3. Has something changed? Were those reports switched?

calellowitz (approved):
offlined

PR #18370, opened 2017-11-01
mobile ucr caching

emord:
@snopoke you were right about oging this route. It ended up making the code easier to read I think. It also makes purging old reports that were deleted much easier

### DIFF
--- a/corehq/apps/app_manager/fixtures/mobile_ucr.py
+++ b/corehq/apps/app_manager/fixtures/mobile_ucr.py
@@ -1,3 +1,5 @@
+from __future__ import absolute_import
+
 from collections import defaultdict
 from datetime import datetime, timedelta
 import logging
--- a/corehq/apps/app_manager/fixtures/mobile_ucr.py
+++ b/corehq/apps/app_manager/fixtures/mobile_ucr.py
@@ -278,23 +278,25 @@ class ReportFixturesProviderV2(BaseReportFixturesProvider):
         }
 
         if needed_versions.intersection({MOBILE_UCR_MIGRATING_TO_2, MOBILE_UCR_VERSION_2}):
-            synced_fixtures, purged_fixture_ids = self._relevant_report_configs(restore_state, apps)
+            report_configs = self._get_report_configs(apps).values()
+            synced_fixtures, purged_fixture_ids = self._relevant_report_configs(restore_state, report_configs)
             fixtures.extend(self._v2_fixtures(restore_user, synced_fixtures))
             for report_uuid in purged_fixture_ids:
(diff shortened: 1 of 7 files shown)


corehq/apps/app_manager/fixtures/mobile_ucr.py
```diff
@@ -277,10 +278,67 @@ def __call__(self, restore_state):
...
+                fixtures.extend(self._empty_v2_fixtures(report_uuid))
 
         return fixtures
 
+    def _relevant_report_configs(self, restore_state, report_configs):
+        """
+        Filter out any UCRs that are already synced. This can't exist in V1,
+        because in V1 we send all reports as one fixture.
+
+        Returns a list of full ReportConfigs to sync and a set of report ids to purge
```
emord: This feels a bit weird of a return value, but we need the full report config later and if an reports were deleted we only have the ID in the synclog

corehq/apps/app_manager/views/modules.py
```diff
@@ -242,9 +243,10 @@ def _report_to_config(report):
             'moduleFilter': module.module_filter,
             'availableReports': [_report_to_config(r) for r in all_reports],  # structure for all reports
             'currentReports': [r.to_json() for r in module.report_configs],  # config data for app reports
-            'columnXpathTemplate': COLUMN_XPATH_CLIENT_TEMPLATE,
+            'columnXpathTemplate': get_column_xpath_client_template(app.mobile_ucr_restore_version),
```
emord: I think this may have accidentally been left out of my earlier PR

snopoke (approved):
Nice

corehq/apps/app_manager/fixtures/mobile_ucr.py
```diff
@@ -277,10 +280,67 @@ def __call__(self, restore_state):
...
+
+        config_uuids = {config.uuid for config in report_configs}
+        extra_configs_on_phone = set(last_ucr_syncs.keys()).difference(config_uuids)
+
+        return configs_to_sync, extra_configs_on_phone
+
+    def _empty_v2_fixtures(self, report_uuid):
+        return [
+            E.fixture(id=self._report_fixture_id(report_uuid)),
+            E.fixture(id=self._report_fixture_id(report_uuid))
```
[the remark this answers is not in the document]
emord: Thanks I fixed it here https://github.com/dimagi/commcare-hq/pull/18370/commits/223e8c61c0604ba00d26aff407edb331bc8b7e7a started to write a test, but it looked like a refactor would be necessary before that unfortunately. (there are a lot of mocks and it doesn't use the db)

PR #18373, opened 2017-11-01
py3 - userreports

nickpell:
Recommend 🐡 

This is the result of running 
python-modernize corehq/apps/userreports/ -w -f libfuturize.fixes.fix_absolute_import
python-modernize corehq/apps/userreports/ -w --future-unicode

I also undid a mistake of the script, where it treated a local variable ```filter``` like the python built-in function.  Overriding filter, map, etc doesn't play well with these scripts, and we may want to discourage that in the codebase.

That's enough changes for its own PR.

@dimagi/py3


esoergel:
Dayum, Nick - you're on a roll!  I agree on overriding builtins, we should do that regardless of upgrading, and we could enforce that if we added full pylint support to stickler:
```
$ pylint --errors-only --disable all --enable W0622 corehq/apps/reports/views.py
************* Module corehq.apps.reports.views
W:348, 8: Redefining built-in 'filter' (redefined-builtin)
W:384, 4: Redefining built-in 'format' (redefined-builtin)
W:388, 4: Redefining built-in 'filter' (redefined-builtin)
W:439, 4: Redefining built-in 'format' (redefined-builtin)
W:446, 4: Redefining built-in 'filter' (redefined-builtin)
W:581,50: Redefining built-in 'format' (redefined-builtin)
W:619, 4: Redefining built-in 'format' (redefined-builtin)
W:641, 4: Redefining built-in 'format' (redefined-builtin)
W:1173, 4: Redefining built-in 'format' (redefined-builtin)
W:1338, 4: Redefining built-in 'slice' (redefined-builtin)
W:1489,57: Redefining built-in 'format' (redefined-builtin)
W:1524,32: Redefining built-in 'file' (redefined-builtin)
W:1554, 8: Redefining built-in 'format' (redefined-builtin)
W:1610,29: Redefining built-in 'id' (redefined-builtin)
W:2014,74: Redefining built-in 'format' (redefined-builtin)
W:2026,48: Redefining built-in 'format' (redefined-builtin)
W:2035,12: Redefining built-in 'file' (redefined-builtin)
```

corehq/apps/userreports/adapter.py
```diff
@@ -1,4 +1,5 @@
 from __future__ import absolute_import
+from __future__ import unicode_literals
```
emord: For some reason I recall that you should either use this in none of your code base, or all of it. Though I suppose that we already use it in some of it, so adding it file by file shouldn't be a terrible idea.

I bet @millerdev has an opinion/knowledge of this and could weigh in on it

corehq/apps/userreports/indicators/factory.py
```diff
@@ -156,6 +157,6 @@ def validate_spec(self, spec):
             raise BadSpecError(
                 _('Illegal indicator type: "{0}", must be one of the following choice: ({1})'.format(
                     spec['type'],
-                    ', '.join(self.constructor_map.keys())
+                    ', '.join(list(self.constructor_map.keys()))
```
emord: Assuming constructor_map is a dictionary, I think that you could just remove `.keys()` here. 

python2:

```
>>> ', '.join({'key1': 'value1', 'key2': 'value2'})
'key2, key1'
```
py3
```
>>> ', '.join({'key1': 'value1', 'key2': 'value2'})
'key1, key2'
```
@millerdev mentioned this to me when we first started thinking about this

corehq/apps/userreports/reports/builder/forms.py
```diff
@@ -84,7 +86,7 @@ class FilterField(JsonField):
     def validate(self, value):
         super(FilterField, self).validate(value)
         for filter_conf in value:
-            if filter_conf.get('format', None) not in REPORT_BUILDER_FILTER_TYPE_MAP.keys() + [""]:
+            if filter_conf.get('format', None) not in list(REPORT_BUILDER_FILTER_TYPE_MAP.keys()) + [""]:
```
emord: same comment about keys

corehq/apps/userreports/reports/filters/choice_providers.py
```diff
@@ -84,7 +84,7 @@ def get_choices_for_values(self, values, user):
         used_values = {value for value, _ in choices}
         for value in values:
             if value not in used_values:
-                choices.add(Choice(value, unicode(value) if value is not None else ''))
+                choices.add(Choice(value, str(value) if value is not None else ''))
```
emord: shouldn't this be six.text_type? or does six fix that?

corehq/apps/userreports/reports/specs.py
```diff
@@ -216,7 +218,7 @@ def aggregations(self, data_source_config, lang):
             aggregation = aggregation.order('_term', order=order)
         else:
             aggregation = ES_AGG_MAP[self.aggregation](self.column_id, self.field)
-        return filter(None, [aggregation])
```
emord: I kind of prefer the filter syntax for this, but understand that py3 changes it's behavior. would it be worth using six.moves.filter when possible? (also sucks that we won't be able to lint for this)

emord (commented):
There are  a few list(dict.keys()) instances that I didn't comment on.

Thoughts on limiting these commits to one fixer at a time? might make the commits easier if there's only one type of change to review at once

corehq/apps/userreports/reports/specs.py
nickpell: There are a few alternatives here: http://python-future.org/compatible_idioms.html#map
I like option 3 (```lmap```) because it is explicit in returning a list whether you're using python 2 or 3.

This could also be re-written as ```[aggregation] if aggregation else []```

corehq/apps/userreports/indicators/factory.py
nickpell: Not sure why ```python-modernize corehq/apps/userreports/ -w --future-unicode``` picked up on these changes.  Agree ````.keys()``` should be removed.  Is there a script setting to make that change, or do we have to do this one manually?

corehq/apps/userreports/reports/filters/choice_providers.py
nickpell: On closer look this looks wrong to me, not sure why modernize would have done that.

nickpell:
Agree that it's generally easiest to limit to one fixer.  Btw modernize automatically does a lot of six conversions unless you pass the flag ```--no-six```.

corehq/apps/userreports/adapter.py
millerdev: Ha, just brought it up over here: https://github.com/dimagi/xml2json/pull/5#pullrequestreview-73802902

Dhe drawbacks listed in [Should I import unicode_literals?](http://python-future.org/unicode_literals.html) (required reading if you want to weigh in on this topic) seem compelling too. I'm leaning against using `unicode_literals` for three main reasons:

- adding it to some files but not others is likely to cause confusion because strings will not behave the same way everywhere.
- adding it to all files in the (HQ) codebase seems infeasible at this time.
- it can introduce hard-to detect bugs in Python 2.

emord:
> Agree that it's generally easiest to limit to one fixer. Btw modernize automatically does a lot of six conversions unless you pass the flag --no-six.

Yeah, when I"ve done this before I've done something like `python-modernize -f lib2to3.fixes.fix_ne` or something similar. can get the list from `python-modernize --list-fixes`. there's something similar for futurize

corehq/apps/userreports/reports/specs.py
millerdev: fwiw, I prefer list comprehension over `filter(None, ...)`

+1 to @nickpell's final suggestion of rewriting with `if aggregation else`

corehq/apps/userreports/adapter.py
esoergel: Would you then push for us to decorate all strings with either `u`, `b`, or `r`, and be explicit about it?

corehq/apps/userreports/admin.py
```diff
@@ -1,3 +1,4 @@
+from __future__ import absolute_import
```
millerdev: Did you use a tool to do this, or did you do it manually? If a tool, please share the command as I have a feeling others will find that useful.

corehq/apps/userreports/adapter.py
millerdev: For `u` and `b`, yes, I think so. The downside to that is we'll have unnecessary `u` prefixes littered everywhere when we switch to Python 3, and removing them would be disruptive to blame history.

Note that `ur` is a valid prefix in Python 2 but not on Python 3, so that does make use of the `r` prefix more complicated in some cases. `r"xyz"` in Python 3 results in a Unicode literal while in Python 2 it's a byte string. `br"xyz"` results in a byte string on both. Good explanation [here](https://stackoverflow.com/a/26063975/10840).

I do like to omit the `u` prefix if the code "just works" without it on both Python 2 and 3, but that's tricky to verify, so maybe not recommended?

One more note here: I don't like that python-modernize is removing the `u` prefix on strings. It's unnecessary (we don't need to support Python 3.0-3.2) and just creates more noise in the patch.

corehq/apps/userreports/reports/builder/v1/forms.py
```diff
@@ -71,7 +73,7 @@ class FilterField(JsonField):
     def validate(self, value):
         super(FilterField, self).validate(value)
         for filter_conf in value:
-            if filter_conf.get('format', None) not in REPORT_BUILDER_FILTER_TYPE_MAP.keys() + [""]:
+            if filter_conf.get('format', None) not in list(REPORT_BUILDER_FILTER_TYPE_MAP.keys()) + [""]:
```
millerdev: another case of keys

corehq/apps/userreports/reports/view.py
```diff
@@ -93,7 +95,7 @@ def get_filter_values(filters, request_dict, user=None):
             for filter in filters
         }
     except FilterException as e:
-        raise UserReportsFilterError(unicode(e))
+        raise UserReportsFilterError(str(e))
```
millerdev: should use `six.text_type`? also others below

corehq/apps/userreports/ui/forms.py
```diff
@@ -278,7 +279,7 @@ def __init__(self, domain, *args, **kwargs):
         self.helper.layout = crispy.Layout(
             crispy.Fieldset(
                 _("Create Data Source from Application"),
-                *report_source_fields.keys()
+                *list(report_source_fields.keys())
```
millerdev: can remove `.keys()` but not `list()`

corehq/apps/userreports/v1/views.py
```diff
@@ -420,7 +421,7 @@ def page_context(self):
                     'is_static': self.existing_report.is_static,
                 })
             return self._handle_exception(error_response, e)
-        field_names = report_form.fields.keys()
+        field_names = list(report_form.fields.keys())
```
millerdev: would be better to keep this as a dict (hash lookup should be better than sequential scan of a list)

```py
field_names = report_form.fields
```

corehq/apps/userreports/adapter.py
millerdev: Another thought to entertain the other side of the argument. Maybe we should go with the [approach Django took](https://docs.djangoproject.com/en/1.11/topics/python3/). It's well articulated and I'm assuming fairly complete in addressing all the topics we'd care about. It's also a big project, so their approach is probably doable for us since they were successful with it.

One big counterargument is that they were aiming to have their code run on both Python 2 and Python 3. I'm assuming that we'll switch to Python 3 immediately and not try to maintain backward compatibility as soon as all of our code supports that to some reasonable degree of stability.

corehq/apps/userreports/admin.py
nickpell: ```python-modernize corehq/apps/userreports/ -w --future-unicode```

I'd recommend adding a ```--no-six``` in the future

corehq/apps/userreports/adapter.py
nickpell: From a process perspective, getting compatible code via ```unicode_literals``` is as simple as running:
```python-modernize corehq/apps/userreports/ -w --future-unicode --no-six```
I could imagine us rolling this out across the entire codebase in a relatively short timeframe.

I don't know of an automated way to add ```u``` in all of needed places.

millerdev: I'm not sure I agree. Check out the second item under [Drawbacks](http://python-future.org/unicode_literals.html#drawbacks) for an example.

nickpell:
closing to tackle in separate PRs

corehq/apps/userreports/adapter.py
millerdev: @dimagi/team-commcare-hq After thinking it over for a while, I'm going to make a pronouncement so we can move forward with porting to Python 3 with as little ambiguity as possible.

I propose that we use [Django's Python 3 porting guide](https://docs.djangoproject.com/en/1.11/topics/python3/).

This means we will use `unicode_literals` as described there. Since we are targeting Python ≥ 3.3, removal of `u` prefixes is optional (although still recommended). The section on [string handling](https://docs.djangoproject.com/en/1.11/topics/python3/#string-handling) is particularly important, especially in clarifying the naming switch from `str` and `unicode` to `bytes` and `text`. We should follow that convention and rename functions in our code to avoid ambiguity.

To reduce the burden of conversion and review, instead of converting the entire code base at once, we will attempt to convert entire Django apps at once (or top-level packages if not part of a Django app). Code reviewers should be vigilant for old-style string assumptions sneaking into parts of the codebase that have been ported to use `unicode_literals`.

I'm still open to being persuaded in a different direction, so if any of this seems like not a great idea, then please speak up.

orangejenny: Late to the conversation, but going with django's approach sounds good. To this point:
> One big counterargument is that they were aiming to have their code run on both Python 2 and Python 3. I'm assuming that we'll switch to Python 3 immediately and not try to maintain backward compatibility as soon as all of our code supports that to some reasonable degree of stability.
Might be nitpicking, but unless it's feasible to switch to Python 3 very soon - like, within a month - I think we're better off thinking of our code as supporting both 2 and 3, and considering doing a later cleanup pass after the switch to 3. Which is to say, that counterargument doesn't seem like a big deal.

PR #18375, opened 2017-11-02
[Report Builder] Fix Default Filter > Date Bug

biyeun:
When Default Filter > Date is selected, actually show date range options.

@orangejenny 
cc @mkangia 

fyi @sheelio


orangejenny:
leaving for @mkangia , i don't have any special context here

millerdev (approved):

PR #18376, opened 2017-11-02
[Report Builder] limit initial list of questions to 10 (including username)

biyeun:
@orangejenny 
cc @mkangia 
fyi @sheelio

### DIFF
--- a/corehq/apps/userreports/reports/builder/forms.py
+++ b/corehq/apps/userreports/reports/builder/forms.py
@@ -1337,8 +1337,8 @@ class ConfigureListReportForm(ConfigureNewReportBase):
             data_source_field=prop.to_report_column_option().get_indicator(COUNT_PER_CHOICE)['column_id'],
             calculation=COUNT_PER_CHOICE
         ))
-        questions = filter(lambda x: x.get_type() == PROPERTY_TYPE_QUESTION,
-            self.data_source_properties.values())
+        questions = [prop for prop in self.data_source_properties.values()
+                     if prop.get_type() == PROPERTY_TYPE_QUESTION]
         if len(questions) > 9:
             questions = questions[:9]
         for q in questions:
--- a/corehq/apps/userreports/reports/builder/forms.py
+++ b/corehq/apps/userreports/reports/builder/forms.py
@@ -1337,19 +1337,20 @@ class ConfigureListReportForm(ConfigureNewReportBase):
             data_source_field=prop.to_report_column_option().get_indicator(COUNT_PER_CHOICE)['column_id'],
             calculation=COUNT_PER_CHOICE
         ))
-        for prop in self.data_source_properties.values():
-            questions_found = 0
-            if prop.get_type() == PROPERTY_TYPE_QUESTION:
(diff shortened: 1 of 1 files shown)


corehq/apps/userreports/reports/builder/forms.py
```diff
@@ -1337,19 +1337,18 @@ def _get_default_form_report_columns(self):
...
-                    exists_in_current_version=True,
-                    property=prop.get_id(),
-                    data_source_field=prop.get_id(),
-                    calculation=COUNT_PER_CHOICE,
-                ))
-                if questions_found == 4:
-                    break
+        questions = filter(lambda x: x.get_type() == PROPERTY_TYPE_QUESTION,
+            self.data_source_properties.values())
+        if len(questions) > 9:
```
orangejenny: don't need the if statement

corehq/apps/userreports/reports/builder/forms.py
```diff
@@ -1337,19 +1337,18 @@ def _get_default_form_report_columns(self):
...
-            if prop.get_type() == PROPERTY_TYPE_QUESTION:
-                questions_found += 1
-                cols.append(ColumnViewModel(
-                    display_text=prop.get_text(),
-                    exists_in_current_version=True,
-                    property=prop.get_id(),
-                    data_source_field=prop.get_id(),
-                    calculation=COUNT_PER_CHOICE,
-                ))
-                if questions_found == 4:
```
orangejenny: Curious, what was this about?

orangejenny (approved):
looks fine, just one nitpick

.stickler.yml
```diff
@@ -8,7 +8,7 @@ files:
 linters:
   flake8:
     max-line-length: 115
-    ignore: E128,E265,F403,W503  # http://pep8.readthedocs.org/en/latest/intro.html#error-codes
+    ignore: E128,E265,F403,W503,W1639  # http://pep8.readthedocs.org/en/latest/intro.html#error-codes
```
emord: This won't do anything. it's coming from pylint (py3k)

corehq/apps/userreports/reports/builder/forms.py
```diff
@@ -1337,19 +1337,18 @@ def _get_default_form_report_columns(self):
...
-                    break
+        questions = filter(lambda x: x.get_type() == PROPERTY_TYPE_QUESTION,
```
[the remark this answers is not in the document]
emord: This is coming from the py3k check. it means that it would break in py3 because filter returns an iterator, not a list. You could change this to `[prop for prop in self.data_source_properties.values() if x.get_type() == PROPERTY_TYPE_QUESTION]`

corehq/apps/userreports/reports/builder/forms.py
biyeun: it was trying to do the question limit and failing

biyeun:
@orangejenny was the nitpick addressed?

orangejenny:
no - it was the if statement before `questions = questions[:9]` - but it isn't blocking

PR #18378, opened 2017-11-02
[Report Builder] Fix false group by error on load of Case Summary report with chart

biyeun:
@orangejenny 
cc @mkangia 
fyi @sheelio


orangejenny:
leaving for @mkangia , i don't have any special context here

millerdev (approved):

PR #18385, opened 2017-11-02
Shadow forms: use parent's form validation cache

orangejenny:
L10K is having an issue where the shadow form is displaying an old error from the parent form because the shadow form still has that error cached. Circumvent this by using the parent form's validation cache directly.

@sravfeyn / @emord / @dannyroberts

### DIFF
--- a/corehq/apps/app_manager/models.py
+++ b/corehq/apps/app_manager/models.py
@@ -3125,6 +3125,13 @@ class ShadowForm(AdvancedForm):
         return get_blank_form_xml("")
 
     @property
+    def validation_cache(self):
+        if not self.shadow_parent_form:
+            return None
+        else:
+            return self.shadow_parent_form.validation_cache
+
+    @property
     def xmlns(self):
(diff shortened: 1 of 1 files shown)


emord (commented):
Does this mean if you have an error in the shadow form, that it will be stored in the parent form? In that case does the cache get cleared if you modify the shadow form without modifying the parent?

> shadow form is displaying an old error from the parent form because the shadow form still has that error cached

Why does a shadow form contain it's own errors and the parent forms?

orangejenny:
The validation cache only stores xform validation errors (see https://github.com/dimagi/commcare-hq/blob/master/corehq/apps/app_manager/models.py#L1022-L1028). Shadow forms don't have their own xml, they use the parent's (see https://github.com/dimagi/commcare-hq/blob/master/corehq/apps/app_manager/models.py#L3121) so the parent's errors and the shadow's errors are one and the same.

emord (approved):

dannyroberts:
Nice work

PR #18389, opened 2017-11-02
lmap

nickpell:
Manually updated.  Keeps py2 like behavior in py3 without any ```list(...)```

@dimagi/py3

### DIFF
--- a/corehq/apps/userreports/indicators/__init__.py
+++ b/corehq/apps/userreports/indicators/__init__.py
@@ -1,8 +1,6 @@
 from __future__ import absolute_import
 from collections import defaultdict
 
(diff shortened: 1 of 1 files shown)


millerdev:
Test failure looks real: `ImportError: No module named future.utils`

emord:
I think that comes from python-futurize's own version of six. I haven't looked into that in a while, but was hesitant to add another py2/py3 compatibility library. thoughts on that?

nickpell:
I don't think there's a python 3 equivalent of ```lmap```, so I'd say to use ```list(map(...))``` or a list comprehension

millerdev (approved):

PR #18390, opened 2017-11-02
Report Builder: Allow for Chart / Summary with no Grouping Column

sheelio:
@biyeun 
Feedback from FMs after testing out report builder

Product Note:  Allows creating a summary report that does not have a grouping column.  This is useful for cascade style reports.


biyeun (approved):

biyeun:
@sheelio nice!

PR #18394, opened 2017-11-03
kill get_parent_of_case

nickpell:
Changes behavior of ```get_episode_case_from_adherence``` and ```get_prescription_from_voucher```

@proteusvacuum


proteusvacuum:
@mkangia any objections to this?

custom/enikshay/two_b_release_1/tests.py
```diff
@@ -6,7 +6,7 @@
 from corehq.apps.receiverwrapper.util import submit_form_locally
 from corehq.form_processor.interfaces.dbaccessors import CaseAccessors
 
-from custom.enikshay.case_utils import get_parent_of_case, CASE_TYPE_DRTB_HIV_REFERRAL
+from custom.enikshay.case_utils import get_first_parent_of_case, CASE_TYPE_DRTB_HIV_REFERRAL
```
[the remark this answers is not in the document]
nickpell: going to follow this up in a second PR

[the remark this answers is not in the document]
nickpell: https://github.com/dimagi/commcare-hq/pull/18396

custom/enikshay/case_utils.py
```diff
@@ -288,7 +273,7 @@ def get_episode_case_from_adherence(domain, adherence_case_id):
     Assumes the following case structure:
     Episode <--ext-- Adherence
     """
-    return get_parent_of_case(domain, adherence_case_id, CASE_TYPE_EPISODE)
+    return get_first_parent_of_case(domain, adherence_case_id, CASE_TYPE_EPISODE)
```
mkangia: `get_parent_of_case` gets open parent case and the `get_first_parent_of_case` gets first parent irrespective of status. 

I am not sure how safe a change that it. Its used at two places and i don't have context on either. If you have considered how it effects then its fine.

nickpell: ```get_episode_case_from_adherence``` gets used in repeaters and ```get_prescription_from_voucher``` gets used in the reconciliation task - maybe @proteusvacuum can weigh in on those two uses?

proteusvacuum: - For the task, it looks like we don't send an update if the prescription is closed. We could change the `try/except` there to a if `prescription.closed`.
- Similarly, in the `NinetyNineDotsAdherenceRepeater` we should check for a closed episode instead.

custom/enikshay/integrations/ninetyninedots/repeater_generators.py
```diff
@@ -263,9 +263,12 @@ class AdherencePayloadGenerator(NinetyNineDotsBasePayloadGenerator):
 
     def get_payload(self, repeat_record, adherence_case):
         domain = adherence_case.domain
+        episode_case = get_episode_case_from_adherence(domain, adherence_case.case_id)
+        if episode_case.closed:
```
nickpell: This seems awkward but mimics the current behavior of ```get_payload```.  Let me know if you have suggestions. @proteusvacuum

custom/enikshay/two_b_release_1/tests.py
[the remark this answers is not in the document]
nickpell: fixed in master, ignoring

proteusvacuum (approved):
looks fine

PR #18396, opened 2017-11-03
add "from __future__ import absolute_import" to custom/enikshay

nickpell:
python-modernize custom/enikshay/ -f libmodernize.fixes.fix_import -w

@dimagi/py3


custom/enikshay/nikshay_datamigration/migrations/0010_followup_changes.py
```diff
@@ -2,6 +2,7 @@
 # Generated by Django 1.10.7 on 2017-04-27 18:33
 from __future__ import unicode_literals
 
+from __future__ import absolute_import
```
proteusvacuum: nit, but could these be combined with the previous `from __future__` import?

nickpell: yeah, but it's much easier to run the script to add them and leave as is.  if every file in the codebase ends up having both I'll get carpal tunnel from combining them all.

nickpell: in the long-run they'll get removed anyway

proteusvacuum: lol, ok.

proteusvacuum (approved):

PR #18400, opened 2017-11-04
No DB for doc_type

kaapstorm:
Loading is failing because there is a doc_type on the destination server that does not have a database.

This change throws an exception so that we know which doc_type it is.


snopoke (approved):

kaapstorm:
(Rebased to remove the commented-out loaders. I'll accommodate stickler.)

PR #18401, opened 2017-11-04
Refactoring Analytics on HQ pt 1

biyeun:
@orangejenny 

Here we go. Currently in a state to support the old analytics and new analytics. Using the library is controlled by a feature flag to help with QAing on production to make sure the new library doesn't throw any JS errors.

The debug levels of the new library are controlled by feature flags to enable analytics debugging output on production.

Still need to add support for
- drift
- hubspot

### DIFF
--- a/corehq/apps/analytics/static/analytics/js/deprecated.js
+++ b/corehq/apps/analytics/static/analytics/js/deprecated.js
@@ -69,7 +69,7 @@ hqDefine('analytics/js/deprecated', function () {
         return args;
     };
 
-    ga_track_event = _makeLegacyFn(_google, 'ga_track_event', _google.track.event, _addCallbackToArgs);
+    ga_track_event = _makeLegacyFn(_google, 'ga_track_event', _google.track.event, _addCallbackToArgs); // eslint-disable-line no-global-assign
 
     // Debug gtag's use of the ga function vs. the old usages
     var _oldGA = ga;
@@ -106,7 +106,7 @@ hqDefine('analytics/js/deprecated', function () {
                 userId = null;
             } else {
                 // uh-oh... can't tell what the first argument was intended to be.
-                throw "Unexpected argument types"
+                throw "Unexpected argument types";
             }
         }
         // Record identity
--- a/corehq/apps/analytics/static/analytics/js/deprecated.js
+++ b/corehq/apps/analytics/static/analytics/js/deprecated.js
@@ -1,11 +1,11 @@
 /* globals _, _kmq */
 
 // Deprecated analytics globals
-var ga_track_event,
-    trackLinkHelper,
-    gaTrackLink,
-    kmqPushSafe,
-    analytics;
+var ga_track_event, // eslint-disable-line no-unused-vars
+    trackLinkHelper, // eslint-disable-line no-unused-vars
+    gaTrackLink, // eslint-disable-line no-unused-vars
+    kmqPushSafe, // eslint-disable-line no-unused-vars
+    analytics; // eslint-disable-line no-unused-vars
 
 window.analytics = {};
 
--- a/corehq/apps/analytics/static/analytics/js/deprecated.js
+++ b/corehq/apps/analytics/static/analytics/js/deprecated.js
@@ -1,0 +1,133 @@
+/* globals _, _kmq */
+
+// Deprecated analytics globals
+var ga_track_event,
+    trackLinkHelper,
+    gaTrackLink,
+    kmqPushSafe,
+    analytics;
+
+window.analytics = {};
+
+// to support deprecated direct ga() calls
+var ga = window.ga || function () {
+    hqImport('analytics/js/google').logger.warning.log(arguments, 'no global ga function was defined');
+    if (arguments.length) {
+        var lastArg = arguments[arguments.length - 1];
+        if (lastArg) {
+            var callback = lastArg.hitCallback;
+            if (callback) {
+                callback();
+            }
+        }
+    }
+};
+
+/**
+ * This creates wrappers with warnings around legacy global functions to help with refactoring of HQ's analytics.
+ * Eventually this will be phased out and all the old globals replaced.
+ */
+hqDefine('analytics/js/deprecated', function () {
+    'use strict';
+    var _utils = hqImport('analytics/js/utils'),
+        _kissmetrics = hqImport('analytics/js/kissmetrics'),
+        _google = hqImport('analytics/js/google');
+
+    /**
+     * Helper function for wrapping the legacy functions.
+     * @param {object} api - the analytics api
+     * @param {string} fnName - Deprecated Function Name
+     * @param {function} fallbackFn - Function to fallback to.
+     * @param {function} formatArgs - (optional) Function that takes an array of arguments and reformats the arguments for fallback function.
+     * @returns {Function} A function that takes the same paramers as the legacy function, but processes it in the new world and warns of its usage.
+     * @private
+     */
+    var _makeLegacyFn = function(api, fnName, fallbackFn, formatArgs) {
+        return function () {
+            var args = Array.from(arguments);
+            api.logger.deprecated.log(arguments, fnName);
+            if (_.isFunction(formatArgs)) {
+                args = formatArgs(args);
+            }
+            return fallbackFn.apply(null, args);
+        };
+    };
+
+    /**
+     * Gets the callback frunction (from Google Analytics' hitCallback) from the last argument, if available,
+     * and re-adds to arguments
+     * @param args - Array of arguments
+     * @returns {array} - Array of arguments
+     * @private
+     */
+    var _addCallbackToArgs = function (args) {
+        var lastArg = _.last(args);
+        if (lastArg && _.isFunction(lastArg.hitCallback)) {
+            delete lastArg.hitCallback;
+            args = _.union(args, [lastArg.hitCallback]);
+        }
+        return args;
+    };
+
+    ga_track_event = _makeLegacyFn(_google, 'ga_track_event', _google.track.event, _addCallbackToArgs);
+
+    // Debug gtag's use of the ga function vs. the old usages
+    var _oldGA = ga;
+    ga = function () {
+        var args = Array.from(arguments);
+        if (args.length > 1 && args[0].startsWith('gtag')) {
+            _google.logger.debug.log(arguments, 'direct gtag > ga call');
+        } else if (args.length > 2 && args[1] === 'event') {
+            args = args.splice(2);
+            args = _addCallbackToArgs(args);
+            _google.logger.deprecated.log(arguments, 'ga');
+            return _google.track.event.apply(null, args);
+        } else {
+            _google.logger.debug.log(arguments, 'unknown ga call');
+        }
+        return _oldGA.apply(null, arguments);
+    };
+    trackLinkHelper = _makeLegacyFn(_google, 'trackLinkHelper', _utils.trackClickHelper);
+    gaTrackLink = _makeLegacyFn(_google, 'gaTrackLink', _google.track.click);
+    kmqPushSafe = _makeLegacyFn(_kissmetrics, 'kmqPushSafe', function (args, timeout) {
+        var lastArg = _.last(args);
+        args[args.length - 1] = _utils.createSafeCallback(lastArg, timeout);
+        _kmq.push(args);
+    });
+
+    window.analytics.usage = _makeLegacyFn(_google, 'window.analytics.usage', _google.track.event, _addCallbackToArgs);
+    window.analytics.trackUsageLink = _makeLegacyFn(_google, 'window.analytics.trackUsageLink', _google.track.click);
+
+    window.analytics.identify = _makeLegacyFn(_kissmetrics, 'window.analytics.identify', function (userId, traits) {
+        if (typeof userId === 'object'){
+            if (traits === undefined){
+                // the first and only argument passed was traits
+                traits = userId;
+                userId = null;
+            } else {
+                // uh-oh... can't tell what the first argument was intended to be.
+                throw "Unexpected argument types"
+            }
+        }
+        // Record identity
+        if (userId !== null) {
+            _kissmetrics.identify(userId);
+        }
+        if (traits !== undefined){
+            _kissmetrics.identifyTraits(traits);
+        }
+    });
+
+    window.analytics.workflow = _makeLegacyFn(_kissmetrics, 'window.analytics.workflow', _kissmetrics.track.event);
+    window.analytics.track = window.analytics.workflow;
+
+    window.analytics.trackWorkflowLink = _makeLegacyFn(_kissmetrics, ' window.analytics.trackWorkflowLink', function(element, event, properties) {
+        _utils.trackClickHelper(element, function (cb) {
+            _kissmetrics.track.event(event, properties, cb);
+        });
+    });
+
+    analytics = window.analytics;
+
+    return {};
+});
--- a/corehq/apps/analytics/static/analytics/js/google.js
+++ b/corehq/apps/analytics/static/analytics/js/google.js
@@ -27,7 +27,7 @@ hqDefine('analytics/js/google', function () {
 
     window.dataLayer = window.dataLayer || [];
     _gtag = function () {
-        dataLayer.push(arguments);
+        window.dataLayer.push(arguments);
         logger.verbose.log(arguments, 'gtag');
     };
     _gtag('js', new Date());
@@ -56,14 +56,6 @@ hqDefine('analytics/js/google', function () {
 
     // Configure Gtag with User Info
     _gtag('config', _init.apiId, _init.user);
-
-    var _fmtEvent = function (action, params) {
-        var messages = [];
-        messages.push({title: "Action", message: action, logGroup: true});
-        _.each(params, function (val, key) {
-            messages.push({title: key, message: val, logGroup: true});
-        });
-    };
 
     /**
      * Helper function for sending a Google Analytics Event.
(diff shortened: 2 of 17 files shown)


corehq/apps/analytics/static/analytics/js/logging.js
```diff
@@ -0,0 +1,176 @@
...
+            _grp.isRaw = true;
+            _grp.print();
+        } else {
+            message.print();
+        }
+    };
+
+    var _getStyle = function (level) {
+        var levelOptions = _LEVELS[level];
+        return (levelOptions) ? levelOptions[1] : "";
```
orangejenny: minor: this would be more readable if `_LEVELS` used hashes so you could say `levelOptions.style` instead of `levelOptions[1]`

corehq/apps/analytics/static/analytics/js/logging.js
```diff
@@ -0,0 +1,176 @@
+/* globals _, JSON */
```
orangejenny: Would it be worth having some overall docs? There's a lot of functionality in here, might be helpful to have a couple of examples to demonstrate using different logging levels and groups (even after reading the full PR, I don't quite follow whether groups are something that the calling code uses, or something used internally to format data/arguments in messages better).

corehq/apps/analytics/static/analytics/js/logging.js
```diff
@@ -0,0 +1,176 @@
...
+    var Group = function (title, message) {
+        var grp = {};
+        grp.title = title;
+        grp.level = message.level;
+        grp.message = message;
+        grp.isCollapsed = false;
+        grp.isRaw = false;
+        grp.isGroup = true;
+        grp.print = function () {
+            var _printGrp = (grp.isCollapsed) ? console.groupCollapsed : console.group;  // eslint-disable-line no-console
```
orangejenny: This is cool, I didn't know `console.group[Collapsed]` existed.

corehq/apps/analytics/static/analytics/js/logging.js
```diff
@@ -0,0 +1,176 @@
...
+ *
+ * To enable logging, enable the following INTERNAL feature flags to see each level:
+ * DEBUG: ANALYTICS_DEBUG
+ * VERBOSE: ANALYTICS_VERBOSE
+ * WARNING, DEPRECATED: ANALYTICS_DEPRECATED
+ *
+ */
+hqDefine('analytics/js/logging', function () {
+    'use strict';
+    var _LEVELS = {
```
orangejenny: I like the styles for different levels. Do the levels work as expected with the web consoles in chrome and firefox? That is, can people use the web console to turn on/off different levels of logging?

screen shot 2017-11-07 at 11 05 59 am

corehq/apps/analytics/static/analytics/js/logging.js
```diff
@@ -0,0 +1,176 @@
...
+        var levelOptions = _LEVELS[level];
+        return (levelOptions) ? levelOptions[1] : "";
+    };
+
+    var Message = function (value, level) {
+        var msg = {};
+        msg.level = level;
+        msg.value = value;
+        msg.print = function () {
+            console.log(msg.value);  // eslint-disable-line no-console
```
orangejenny: Mentioned this elsewhere, but after looking a bit at the [MDN docs](https://developer.mozilla.org/en-US/Add-ons/SDK/Tools/console), I'm even more in favor of taking advantage of the built-in `debug`, `info`, etc. That would also allow you to get rid of the `ANALYTICS_DEBUG`, `ANALYTICS_WARNING`, and `ANALYTICS_VERBOSE` feature flags; people could just interact normally with the browser's web console logging controls.

corehq/apps/analytics/static/analytics/js/kissmetrics.js
```diff
@@ -0,0 +1,152 @@
...
+        _kmqPushCommand('trackClickOnOutboundLink', properties, undefined, name);
+    };
+
+    return {
+        logger: logger,
+        identify: identify,
+        identifyTraits: identifyTraits,
+        track: {
+            event: trackEvent,
+            click: trackClick,
```
orangejenny: What would you think of renaming this to something like `internalClick`? I worry that people won't realize `outboundLink` exists.

corehq/apps/hqwebapp/templates/hqwebapp/base.html
```diff
@@ -407,19 +408,41 @@
...
         <script src="{% statici18n LANGUAGE_CODE %}"></script> {# DO NOT COMPRESS #}
+
+        {# Core dependencies for analytics & HQ's libraries #}
+        {% compress js %}
+        <script src="{% static 'hqwebapp/js/hqModules.js' %}"></script>
+        <script src="{% static 'hqwebapp/js/initial_page_data.js' %}"></script>
+        {% endcompress %}
+
+        {% if request|toggle_enabled:"ANALYTICS_NEW" or USE_NEW_ANALYTICS %}
+        {% include 'analytics/analytics_js.html' %}
```
orangejenny: Can this just be included after the compress block below, around line 424? Or are there scripts in that block that depend on analytics?

I'm a little worried about what we define as "core" HQ libraries and about the dependencies among them being fragile (this has come up in the RequireJS work, which also makes some changes to this set of script tags).

corehq/apps/analytics/static/analytics/js/kissmetrics.js
```diff
@@ -0,0 +1,152 @@
...
+        window.dataLayer.push(data);
+        logger.verbose.log(command, ['window._kmq.push', 'window.dataLayer.push', '_kmqPushCommand', commandName]);
+    };
+
+    /**
+     * Initialization function for kissmetrics libraries
+     * @param srcUrl
+     * @private
+     */
+    var _kms = function (srcUrl) {
```
orangejenny: Can `_kms` get a more readable name?

corehq/apps/analytics/static/analytics/js/kissmetrics.js
```diff
@@ -0,0 +1,152 @@
...
+     * @param {object} properties
+     * @param {function|undefined} callbackFn - optional
+     * @param {string|undefined} eventName - optional
+     */
+    var _kmqPushCommand = function (commandName, properties, callbackFn, eventName) {
+        var command, data;
+        command = _.filter([commandName, eventName, properties, callbackFn], function (a) {
+            return !_.isUndefined(a);
+        });
+        _kmq.pushNew(command);
```
orangejenny: What difference does it make if the undefined arguments are filtered out before calling `pushNew`?

corehq/apps/analytics/static/analytics/js/kissmetrics.js
```diff
@@ -0,0 +1,152 @@
...
+        logger = hqImport('analytics/js/logging').getLoggerForApi('Kissmetrics'),
+        _utils = hqImport('analytics/js/utils'),
+        _init = {};
+
+    logger.verbose.addCategory('data', 'DATA');
+    logger.debug.addCategory('ab', 'AB TEST');
+
+    window.dataLayer = window.dataLayer || [];
+
+    var KmqWrapper = function (originalObject) {
```
orangejenny: I'm not totally following the purpose of `KmqWrapper`, is it just that it logs when deprecated code is used?

corehq/apps/analytics/static/analytics/js/kissmetrics.js
```diff
@@ -0,0 +1,152 @@
+/* globals _, _kmq */
+
+var _kmq = _kmq || [];
```
orangejenny: I know `_kmq` comes from kiss, but could it get a reasonable name once it's in our code? `KissMetricsQueue`? Naming is one of the things that's irritating about working with kiss.

corehq/apps/analytics/static/analytics/js/kissmetrics.js
```diff
@@ -0,0 +1,152 @@
...
+        _utils.insertAsyncScript(srcUrl);
+    };
+
+    _init.apiId = _get('apiId');
+    logger.verbose.data(_init.apiId || "NONE SET", "API ID");
+
+    // Initialize Kissmetrics
+    if (_init.apiId) {
+        _kms('//i.kissmetrics.com/i.js');
+        _kms('//doug1izaerwt3.cloudfront.net/' + _init.apiId + '.1.js');
```
orangejenny: It's really weird to see a cloudfront url hardcoded in here, but it looks like that's what we do now in prod.

corehq/apps/hqwebapp/templates/hqwebapp/base.html
```diff
@@ -407,19 +408,41 @@
             {# do not override this block, use registerurl template tag to populate #}
         {% endblock %}
         </div>
+        {% if request|toggle_enabled:"ANALYTICS_NEW" or USE_NEW_ANALYTICS %}
```
orangejenny: Happy to see namespacing here; I've always felt a bit guilty that initial page data is a global namespace.

corehq/apps/analytics/static/analytics/js/google.js
```diff
@@ -0,0 +1,148 @@
...
+    var _addGoogleScript = function (srcUrl) {
+        logger.verbose.log(srcUrl, "Added Script");
+        _utils.insertAsyncScript(srcUrl);
+    };
+
+    if (_init.apiId) {
+        _addGoogleScript('//www.googletagmanager.com/gtag/js?id=' + _init.apiId);
+    }
+
+    window.dataLayer = window.dataLayer || [];
```
orangejenny: I don't understand exactly what `window.dataLayer` does - looks like it's storing all of the kiss and GA events, but also like we only push data to it, never read from it...is that right? Why do these need to be stored in the window?

corehq/apps/analytics/static/analytics/js/deprecated.js
```diff
@@ -0,0 +1,133 @@
+/* globals _, _kmq */
```
orangejenny: As things get deprecated, it'd be good to update the [analytics README](https://github.com/dimagi/commcare-hq/blob/master/corehq/apps/analytics/README.md).

orangejenny (commented):
This is a great step forward! Analytics code has gotten to be quite the...plate of spaghetti.

The main change that I think would be worthwhile before merging this is integrating the logging more tightly with browser tools.

Are you thinking of passive or active QA? I'd be pretty excited about active QA, having occasionally stumbled across analytics that appeared to have never worked in the first place, but I'm not sure if we have a complete list of expected analytics to turn over to QA.

Can you say a little about your plan for deprecation? We don't change the core analytics code that often, but there are bug fixes here and there so there's definitely risk of the forked code getting out of sync. I think there's little enough analytics code (based on grepping for things like `analytics.workflow`) that we could migrate relatively quickly, with a bit of dedicated effort. Maybe by the end of the year? I'm willing to help with that migration.

corehq/apps/analytics/static/analytics/js/google.js
biyeun: `window.dataLayer` is what `gtag` is reading from. There are integrations for google analytics with kissmetrics data now as well.
https://developers.google.com/analytics/devguides/collection/gtagjs/

corehq/apps/analytics/static/analytics/js/deprecated.js
biyeun: yes, absolutely!

corehq/apps/analytics/static/analytics/js/kissmetrics.js
biyeun: the global `_kmq` variable is what kissmetrics interacts with. I tried looking for a way to rename it and let kissmetrics know with no luck. Outside of this analytics library we shouldn't *ever* be calling `_kmq` directly anywhere on HQ. `_kmq.push` should only ever be seen in `analytics/js/kissmetrics.js`

biyeun:
@orangejenny since dimagi.com will have similar analytics requirements to HQ it might be worth externalizing the hq analytics django app for this purpose. what do you think?

corehq/apps/analytics/static/analytics/js/logging.js
biyeun: me too! I also didn't know you could use CSS to style `console.log` statements!! mind blown 💥

corehq/apps/hqwebapp/templates/hqwebapp/base.html
biyeun: there are indeed scripts in that block that depend on analytics.
`hq.helpers`
`main.js`

corehq/apps/analytics/static/analytics/js/logging.js
biyeun: It's nice to have at least one feature flag to disable all `console` statements from analytics altogether

corehq/apps/analytics/static/analytics/js/logging.js
biyeun: yes, i agree. updating

corehq/apps/analytics/static/analytics/js/logging.js
```diff
@@ -0,0 +1,176 @@
...
+    'use strict';
+    var _LEVELS = {
+        warning: ['WARNING', 'color: #994f00;'],
+        verbose: ['VERBOSE', 'color: #685c53;'],
+        debug: ['DEBUG', 'color: #004ebc;'],
+        deprecated: ['DEPRECATED', 'color: #e53e30;'],
+    };
+
+    var _printPretty = function (message) {
+        var _title, _value, _grp;
```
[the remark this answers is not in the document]
biyeun: agree. updating

corehq/apps/analytics/static/analytics/js/kissmetrics.js
biyeun: it logs when `_kmq.push` is called directly

corehq/apps/analytics/static/analytics/js/initial.js
```diff
@@ -0,0 +1,98 @@
...
+        if (_.isUndefined(_abTestsByApi[apiName])) {
+            var tests = _.map(_abTests, function (val, key) {
+                if (key.startsWith(apiName)) {
+                    return {
+                        slug: _.last(key.split('.')),
+                        context: val,
+                    };
+                }
+            });
+            tests = _.filter(tests, function (t) {
```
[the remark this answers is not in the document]
biyeun: yes

corehq/apps/analytics/static/analytics/js/kissmetrics.js
biyeun: yes

corehq/apps/analytics/static/analytics/js/kissmetrics.js
biyeun: yea...that's how it's initialized

corehq/apps/analytics/static/analytics/js/kissmetrics.js
biyeun: great point

corehq/apps/analytics/static/analytics/js/logging.js
biyeun: unfortunately you can't apply the console levels to groups. also it feels like a good idea to completely turn off any console output altogether

corehq/apps/analytics/static/analytics/js/logging.js
biyeun: I said this above, but I've discovered that unfortunately you can't apply the console levels to groups. also it feels like a good idea to completely turn off any console output altogether

corehq/apps/analytics/static/analytics/js/logging.js
biyeun: groups are something internal to format data/argument messages better

biyeun:
> Are you thinking of passive or active QA? I'd be pretty excited about active QA, having occasionally stumbled across analytics that appeared to have never worked in the first place, but I'm not sure if we have a complete list of expected analytics to turn over to QA.

Yeah I think active QA wouldn't be very possible. I think the QA I'm envisioning is just making sure that no javascript errors are thrown as a result of the new library before fully switching over.

> Can you say a little about your plan for deprecation? We don't change the core analytics code that often, but there are bug fixes here and there so there's definitely risk of the forked code getting out of sync. I think there's little enough analytics code (based on grepping for things like analytics.workflow) that we could migrate relatively quickly, with a bit of dedicated effort. Maybe by the end of the year? I'm willing to help with that migration.

I think analytics can get migrated easily in a day. That was my intention. The reason for having this behind a feature flag to begin with is just to see first that it behaves as intended in the production environment and that the data is being sent to google and kissmetrics properly.

I think this would be a great thing for me to work on during that hackathon, no? ;-)

biyeun:
@orangejenny thank you for your very thorough review! I pushed some changes that hopefully addressed most of the outstanding issues.

biyeun:
> The main change that I think would be worthwhile before merging this is integrating the logging more tightly with browser tools.

I wouldn't be able to disable the output of group headers without the use of the feature flags. The still output regardless of the type of `console` statement inside of them.

biyeun:
@calellowitz

corehq/apps/analytics/static/analytics/js/kissmetrics.js
orangejenny: I don't understand - couldn't this line be `var kissMetricsQueue = _kmq || [];`? If it can't, yeah, at least it's localized.

corehq/apps/analytics/static/analytics/js/logging.js
orangejenny: Argh, that's too bad about levels & groups. What do you mean about turning off console output altogether? There are other places, like formplayer, that log to console (granted, maybe they shouldn't).

orangejenny: oh, nm, just saw comment below, looks like you mean turning it off within analytics

corehq/apps/hqwebapp/templates/hqwebapp/base.html
orangejenny: Can you comment those dependencies in this file?

orangejenny:
Hey, thanks for the quick updates. Not being able to integrate more with the browser is pretty unfortunate, but so be it. Migration plan sounds great.

> externalizing the hq analytics django app

Not sure what this means - pulling the analytics app out of HQ?

orangejenny (approved):

biyeun:
> Not sure what this means - pulling the analytics app out of HQ?

Yes to make it reusable to dimagi.com and potentially other dimagi services? Just so that there is some consistency? My gut is to build out separately because the ecosystem is still in flux and consolidate later...

orangejenny:
@biyeun That sounds good to me (though my opinion isn't terribly strong).

orangejenny:
Should formplayer and prelogin now include `analytics_js.hmtl` or have any other changes?

PR #18406, opened 2017-11-06
Reindex all indices on reset

kaapstorm:
Resolves the following:
```
$ ./manage.py ptop_preindex --reset
Nothing needs to be reindexed
```

Labelling "hold off" because this change results in `reset=True` to be passed to `__init__` methods of Reindexer subclasses, and they don't all take `reset` as a param. e.g. ElasticPillowReindexer [doesn't](https://github.com/dimagi/commcare-hq/blob/0bc070384366bf3674f112c32cf0be1baec958c0/corehq/ex-submodules/pillowtop/reindexer/reindexer.py#L150-L150) but ResumableBulkElasticPillowReindexer [does](https://github.com/dimagi/commcare-hq/blob/0bc070384366bf3674f112c32cf0be1baec958c0/corehq/ex-submodules/pillowtop/reindexer/reindexer.py#L243-L243).

@snopoke cc @javierwilson buddy @millerdev

### DIFF
--- a/corehq/pillows/domain.py
+++ b/corehq/pillows/domain.py
@@ -59,6 +59,9 @@ class DomainReindexerFactory(ReindexerFactory):
     ]
 
     def build(self):
+        args = ElasticPillowReindexer.__init__.__code__.co_varnames
+        # Drop options that are not kwargs of the reindexer (like "reset")
+        options = {k: v for k, v in self.options.items() if k in args}
         return ElasticPillowReindexer(
             pillow=get_domain_kafka_to_elasticsearch_pillow(),
             change_provider=CouchViewChangeProvider(
@@ -72,5 +75,5 @@ class DomainReindexerFactory(ReindexerFactory):
             ),
             elasticsearch=get_es_new(),
             index_info=DOMAIN_INDEX_INFO,
-            **self.options
+            **options
         )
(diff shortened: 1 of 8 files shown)


millerdev (approved):
I don't have much context here

corehq/pillows/domain.py
```diff
@@ -59,6 +59,9 @@ class DomainReindexerFactory(ReindexerFactory):
     ]
 
     def build(self):
+        args = ElasticPillowReindexer.__init__.__code__.co_varnames
+        # Drop options that are not kwargs of the reindexer (like "reset")
+        options = {k: v for k, v in self.options.items() if k in args}
```
kaapstorm: This filters out invalid keyword arguments before trying to pass them to `ElasticPillowReindexer.__init__()`.

Another option would be to use  `options.pop('reset')`. I chose against that because it felt too specific and hacky. But conversely this approach feels too generic.

Alternatively, all reindexers' `__init__()` methods could take **kwargs, and ignore them. ... Hmmm ... maybe that's best.

millerdev: > all reindexers' `__init__()` methods could take **kwargs, and ignore them.

I'm not a fan of that approach because it's an easy source of bugs: someone passes in an argument they think is important when it's actually being ignored, which leads to unexpected program behavior. My preference is to pop the ones you expect but explicitly want to ignore, and raise an error for any other unexpected arguments.

As [the saying](https://www.python.org/dev/peps/pep-0020/) goes: _Errors should never pass silently._ (_Unless explicitly silenced._)

PR #18407, opened 2017-11-06
Increase contrast of trend lines on all chart views

kkrampa:
https://manage.dimagi.com/default.asp?264984

@emord


kkrampa:
@emord Can we ignore py3k errors for now? I will fix them for the whole module in the separate PR.

emord (approved):

PR #18409, opened 2017-11-06
a few shard utils

czue:
A few utilities to view and inspect shard data, including a command to see all unexpected/invalid shard data in the entire system.

hopefully this doesn't reproduce too many things that already exist elsewhere.

the shard queries won't run properly until pghash gets installed on all the shards. I still need to update ansible to make that change (though this is safe since it's only called in a management command).

Probably easiest reviewed by commit, though is pretty small


corehq/form_processor/management/commands/list_data_by_shard.py
```diff
@@ -0,0 +1,42 @@
...
+        parser.add_argument('domain', help='Domain to create data in.')
+
+    def handle(self, domain, **options):
+        forms_by_shard = Counter()
+        forms_by_db = Counter()
+        cases_by_shard = Counter()
+        cases_by_db = Counter()
+        print('======================== forms ========================')
+        print('id\t\t\t\t\tshard\tdatabase')
+        for form_id in sorted(FormAccessors(domain=domain).get_all_form_ids_in_domain()):
```
snopoke: Maybe not a good idea to load ~70M IDs into memory? (I think it's about 5GB)

corehq/sql_db/util.py
```diff
@@ -127,3 +128,23 @@ def _inner(*args, **kwargs):
...
+def get_all_sharded_models():
+    for subclass in _get_all_nested_subclasses(PartitionedModel):
+        if not subclass._meta.abstract:
+            yield subclass
+
+
+def _get_all_nested_subclasses(cls):
+    seen = set()
+    for subclass in cls.__subclasses__():
+        print subclass
```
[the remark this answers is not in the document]
snopoke: print

snopoke (commented):
Nice, one comment

corehq/form_processor/management/commands/list_data_by_shard.py
czue: haha, this is just for local testing, but i agree. was doing a small domain-level analysis

czue: I think this command will eventually be deleted/superseded by other things. let me know if you want me to do that now though am tempted to leave it around until I'm a little further along

snopoke:
Makes sense. Fine to go ahead as it's for now.

corehq/sql_db/shard_data_management.py
```diff
@@ -0,0 +1,44 @@
...
+    # syntax of this query is a bit weird because of a couple django / postgres ARRAY oddities
+    # https://stackoverflow.com/a/22008870/8207
+    # https://stackoverflow.com/a/11730789/8207
+    return """
+       select * from (
+           select {shard_id_function} as shard_id, count(*)
+           from {table_name}
+           group by shard_id
+        ) as countsByShard
+        where not shard_id = ANY(%s);
```
gcapalbo: not blocking, but could have also used `<>` here instead of `not =`

corehq/sql_db/shard_data_management.py
```diff
@@ -0,0 +1,44 @@
...
+    query = _get_shard_count_query(model)
+    valid_shards = partition_config.get_shards_on_db(database)
+    cursor.execute(query, [valid_shards])
+    results = cursor.fetchall()
+    return results
+
+
+def _get_shard_count_query(model):
+    # have to cast to varchar because some tables have uuid types
+    shard_id_function = "hash_string({id_field}::varchar, 'siphash24') & {total_shard_count}".format(
```
gcapalbo: did the casting to varchar work as expected for the uuid types? i remember having to cast to byte array: https://github.com/dimagi/commcare-hq/blob/b9c453bf26d6b0267b0e801ad9a3ba9c61119a39/corehq/form_processor/backends/sql/dbaccessors.py#L104

czue: it ran but am not sure it worked properly because I don't have any local messaging data. 

had been debating whether it was worth the overhead of getting a test setup up and running for these functions. it's a lot of bootstrapping code but is probably worth it if we're going to be relying on them to validate/delete data.

corehq/sql_db/shard_data_management.py
czue: good call, updated

corehq/sql_db/shard_data_management.py
gcapalbo: yeah we'll want to make sure we test for this use case if we're going to rely on it for validating or deleting data.

casting the uuid to varchar is not going to yield the right answer - we'll need to find a way to cast the uuid type to bytea to match what this function is doing: https://github.com/dimagi/commcare-hq/blob/a40ce735e7dd8f9d96bc69b273e44035cde928b1/corehq/form_processor/backends/sql/dbaccessors.py#L117-L128

czue: thanks for flagging this. probably saved me loads of future time trying to figure out what was going wrong down the line: https://github.com/dimagi/commcare-hq/pull/18409/commits/5684b9bada9070e6e1d01eb503306c9c9b63fbc5

czue:
(tests are going to fail until i sort out with @snopoke whether it's fine to add migrations to create the hashlib functions)

corehq/form_processor/tests/utils.py
```diff
@@ -107,7 +107,7 @@ def delete_all_sync_logs(cls):
 
     @staticmethod
     @unit_testing_only
-    def _delete_all_sql_sharded_modesl(model_class, domain=None):
+    def _delete_all_sql_sharded_models(model_class, domain=None):
```
snopoke: :eyes:

corehq/sql_db/shard_data_management.py
snopoke: Seems like you reverted this back. How come?

snopoke (approved):

corehq/sql_db/shard_data_management.py
czue: see comment here: https://github.com/dimagi/commcare-hq/pull/18409#commitcomment-25475893

corehq/form_processor/tests/utils.py
czue: :mag:

corehq/sql_db/shard_data_management.py
gcapalbo: fine to leave as `not =`, see my comment here: https://github.com/dimagi/commcare-hq/pull/18409#commitcomment-25478044

czue:
ok @gcapalbo @snopoke I think I've responded to all feedback. I also added one more utility which you can see in the latest commits.

would be nice to get this in once tests pass so I can move on (and also potentially run it on some environments after deploy)

snopoke (approved):

emord:
Going to revert this PR because the migration causes this error (on enikhsay):

`[172.25.3.5] out: django.db.utils.ProgrammingError: permission denied for language c`

Is it possible some ansible work needs to be done first?

czue:
shoot - I didn't think so. the pghashlib extension should be [installed by ansible](https://github.com/dimagi/commcarehq-ansible/blob/master/ansible/roles/postgresql/tasks/main.yml#L225-L250) on all db servers which I had thought was the only dependency (and @snopoke thought so too). I'll see if I can figure out what's going on though. Do you have any more details / stack trace info?

snopoke:
Strange. I know we make `pl_proxy` a [trusted language](https://github.com/dimagi/commcarehq-ansible/blob/14210653f6ec0e8acc226918f9c1949a2adcb8e3/ansible/roles/postgresql/tasks/main.yml#L260-L260) but I don't think we've ever had to do anything for `pghashlib`

Perhaps installing `pl_proxy` somehow makes `c` a trusted language. I guess in ansible we could just make `c` a trusted language in the same way we do for `pl_proxy`

emord:
Sorry I didn't grab hte stack trace. just the SQL error. do we not have plproxy on enikhay?

PR #18410, opened 2017-11-06
fix rebuildstaging

nickpell:
Fixes
```
  File "scripts/rebuildstaging.py", line 40, in <module>
    from .sh_verbose import ShVerbose
ValueError: Attempted relative import in non-package
```

https://stackoverflow.com/questions/11536764/how-to-fix-attempted-relative-import-in-non-package-even-with-init-py

@orangejenny

### DIFF
--- a/scripts/rebuildstaging
+++ b/scripts/rebuildstaging
@@ -46,7 +46,7 @@ do
 done
 
 function rebuildstaging() {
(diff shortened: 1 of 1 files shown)


orangejenny (approved):

orangejenny:
no test coverage, merging

PR #18411, opened 2017-11-06
Fix issues with shadow forms' validation cache

orangejenny:
Introduced by https://github.com/dimagi/commcare-hq/pull/18385

Likely need to hotfix because this is breaking one of enikshay's apps.

@emord / @dannyroberts

### DIFF
--- a/corehq/apps/app_manager/models.py
+++ b/corehq/apps/app_manager/models.py
@@ -1034,10 +1034,10 @@ class FormBase(DocumentSchema):
                     "validation_problems": e.validation_problems,
                     "version": e.version,
                 }
-                self.set_validation_cache(json.dumps(validation_dict))
-                vc = self.get_validation_cache()
+                vc = json.dumps(validation_dict)
             else:
-                vc = self.set_validation_cache("")
+                vc = ""
+            self.set_validation_cache(vc)
         if vc:
             try:
                 raise XFormValidationError(**json.loads(vc))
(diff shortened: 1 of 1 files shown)


nickpell (approved):

emord (approved):

corehq/apps/app_manager/models.py
```diff
@@ -1025,14 +1034,14 @@ def validate_form(self):
                     "validation_problems": e.validation_problems,
                     "version": e.version,
                 }
-                vc = self.validation_cache = json.dumps(validation_dict)
+                vc = self.set_validation_cache(json.dumps(validation_dict))
             else:
-                vc = self.validation_cache = ""
+                vc = self.set_validation_cache("")
```
mkangia: Probably very simple but I am confused how this is resolving the issue. Can you share how this is working now?

orangejenny: Previously, this line would fail for shadow forms because `validation_cache` wasn't a string but was instead a function that returned the parent's cache. Now `validation_cache` is always a string, but instead of code directly accessing it, it's accessed through `get_validation_cache`, `set_validation_cache`, and `clear_validation_cache`, which determine whether to get/set the form's own cache (for normal forms) or the parent's cache (for shadow forms).

mkangia: Thank you. 👍

PR #18415, opened 2017-11-06
handle multiple open referrals

nickpell:
@sheelio


sheelio:
Looks good! Does it default to an ascending sort?

nickpell:
> Does it default to an ascending sort?

yes

PR #18419, opened 2017-11-06
add absolute_import across repo

nickpell:
```python-modernize . -w -f libmodernize.fixes.fix_import```

@dimagi/py3


orangejenny:
bless you

esoergel (changes requested):
Looks like this PR rmoves `custom/ucr_ext/tests.py`, guessing that's a mistake.

esoergel:
Actually maybe not - run `$ git show 3e88dbe4554316981e7f9a43fde34aa8870702c9 custom/ucr_ext/tests.py`, it shows it as every line removed, but then added back with the future import.  Might just be a fluke in the diff algorithm.

nickpell:
yeah, it looks like a fluke to me

esoergel:
Ah, I see it now - looks like the editing process converted that file from carriage returns to normal line breaks.  Must've been edited by someone on Windows in a weird editor.

esoergel (approved):

millerdev:
woohoo 🎉

PR #18425, opened 2017-11-07
make icds ucrs available on softlayer

snopoke:
required for remote linked apps


emord (approved):
I thought these were already there. guess they were removed when v2 was created

PR #18428, opened 2017-11-07
Send updates for 99DOTS successful records wrongly marked as failed

proteusvacuum:
When registering patients in 99DOTS, we sometimes get 500 errors followed by a "This patient already exists" message. Since both of these count as "errors", we never mark these registrations as successful, and therefore never update that particular patient. I manually marked a lot of these as "successful" already. 

This grabs all of those records and sends all pertinent information to 99DOTS. 

@calellowitz 
@mkangia @esoergel

### DIFF
--- a/custom/enikshay/management/commands/mark_successfull_99dots_records.py
+++ b/custom/enikshay/management/commands/mark_successfull_99dots_records.py
@@ -6,12 +6,14 @@ from datetime import datetime
 from django.core.management.base import BaseCommand
 
 from corehq.apps.hqcase.utils import update_case
+from corehq.form_processor.exceptions import CaseNotFound
 from corehq.form_processor.interfaces.dbaccessors import CaseAccessors
 from corehq.motech.repeaters.dbaccessors import (get_repeat_record_count,
                                                  iter_repeat_records_by_domain)
 from corehq.motech.repeaters.models import RepeatRecord
 from corehq.util.log import with_progress_bar
-from custom.enikshay.case_utils import get_person_case, get_adherence_cases_from_episode
+from custom.enikshay.case_utils import (get_adherence_cases_from_episode,
+                                        get_person_case_from_episode)
 
 
 class Command(BaseCommand):
@@ -38,14 +40,17 @@ class Command(BaseCommand):
         cases_to_update = set()
         print("Filtering successful cases")
         for repeat_record in with_progress_bar(records, length=count):
-            if repeat_record.attempts and any(
-                    (existing_message in attempt.message if attempt.message is not None else "")
-                    for attempt in repeat_record.attempts):
-                episode = accessor.get_case(repeat_record.payload_id)
+            if any((existing_message in attempt.message if attempt.message is not None else "")
+                   for attempt in repeat_record.attempts):
+                try:
+                    episode = accessor.get_case(repeat_record.payload_id)
+                except CaseNotFound:
+                    continue
                 if episode.get_case_property('dots_99_registered') != 'true':
                     cases_to_update.add(episode)
 
-        with open('set_99dots_to_registered.csv', 'w') as f:
+        timestamp = datetime.utcnow().strftime("%Y-%m-%d-%H-%M-%S")
+        with open('{}_set_99dots_to_registered.csv'.format(timestamp), 'w') as f:
             writer = csv.writer(f)
             writer.writerow([
                 'beneficiary_id',
@@ -57,7 +62,7 @@ class Command(BaseCommand):
             print("Updating {} successful cases in 99DOTS".format(len(cases_to_update)))
             for case in with_progress_bar(cases_to_update):
                 writer.writerow([
-                    get_person_case(domain, case.case_id).case_id,
+                    get_person_case_from_episode(domain, case.case_id).case_id,
                     case.case_id,
                     self.update_registered_status(domain, case),
                     self.update_patients(domain, case),
@@ -66,8 +71,12 @@ class Command(BaseCommand):
                 ])
 
     def update_registered_status(self, domain, case):
-        if self.commit:
-            update_case(domain, case.case_id, {'dots_99_registered': 'true'})
+        try:
+            if self.commit:
+                update_case(domain, case.case_id, {'dots_99_registered': 'true'})
+            return "dots_99_registered updated in eNikshay"
+        except Exception as e:
+            return "failure updating case in enikshay: {}".format(e)
 
     def update_patients(self, domain, case):
         repeater_id = 'b4e19fd859f852871703e8e32a1764a9'
@@ -81,7 +90,7 @@ class Command(BaseCommand):
         adherence_cases = get_adherence_cases_from_episode(domain, case.case_id)
         if adherence_cases:
             return {
-                adherence_case.case_id: self.send_repeat_record(domain, case.case_id, repeater_id, repeater_type).state
+                adherence_case.case_id: self.send_repeat_record(domain, case.case_id, repeater_id, repeater_type)
                 for adherence_case in adherence_cases
             }
         else:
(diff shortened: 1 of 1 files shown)


custom/enikshay/management/commands/mark_successfull_99dots_records.py
```diff
@@ -0,0 +1,110 @@
...
+    """
+    def add_arguments(self, parser):
+        parser.add_argument('domain')
+        parser.add_argument('--commit', action='store_true')
+
+    def handle(self, domain, **options):
+        # For all successful registration records
+        # If any have an attempt that id "A patient with this beneficiary_id already exists"
+        # Check the episode case. If this doesn't have "dots_99_registered" then set this property to "true"
+        self.commit = options['commit']
```
mkangia: Nice notation.

custom/enikshay/management/commands/mark_successfull_99dots_records.py
```diff
@@ -0,0 +1,110 @@
...
+        repeater_id = 'dc73c3da43d42acd964d80b287926833'  # 99dots register
+        accessor = CaseAccessors(domain)
+        existing_message = "A patient with this beneficiary_id already exists"
+        count = get_repeat_record_count(domain, repeater_id, state="SUCCESS")
+        records = iter_repeat_records_by_domain(domain, repeater_id, state="SUCCESS")
+
+        cases_to_update = set()
+        print("Filtering successful cases")
+        for repeat_record in with_progress_bar(records, length=count):
+            if repeat_record.attempts and any(
```
mkangia: Whats the use of checking `repeat_record.attempts` first here? I guess `any` would return false in case there are no attempts.

custom/enikshay/management/commands/mark_successfull_99dots_records.py
```diff
@@ -0,0 +1,110 @@
...
+                    (existing_message in attempt.message if attempt.message is not None else "")
+                    for attempt in repeat_record.attempts):
+                episode = accessor.get_case(repeat_record.payload_id)
+                if episode.get_case_property('dots_99_registered') != 'true':
+                    cases_to_update.add(episode)
+
+        with open('set_99dots_to_registered.csv', 'w') as f:
```
mkangia: nit: I usually find it helpful to include timestamp in the file name. Helps to differentiate multiple runs if done with `commit` and without `commit`. Also keeps a store of when the command was run.

custom/enikshay/management/commands/mark_successfull_99dots_records.py
```diff
@@ -0,0 +1,110 @@
...
+                'UpdatePatient Status',
+                'Adherence Status',
+                'TreatmentOutcome Status'
+            ])
+            print("Updating {} successful cases in 99DOTS".format(len(cases_to_update)))
+            for case in with_progress_bar(cases_to_update):
+                writer.writerow([
+                    get_person_case(domain, case.case_id).case_id,
+                    case.case_id,
+                    self.update_registered_status(domain, case),
```
mkangia: `update_registered_status` does not return any value. You probably want something here.

custom/enikshay/management/commands/mark_successfull_99dots_records.py
```diff
@@ -0,0 +1,110 @@
...
+                    self.send_adherence(domain, case),
+                    self.send_treatment_outcome(domain, case),
+                ])
+
+    def update_registered_status(self, domain, case):
+        if self.commit:
+            update_case(domain, case.case_id, {'dots_99_registered': 'true'})
+
+    def update_patients(self, domain, case):
+        repeater_id = 'b4e19fd859f852871703e8e32a1764a9'
```
mkangia: Probably not very expensive but better to just have these as constants above and just send them in arguments instead of making a variable each time method is called.
same for repeater_type

custom/enikshay/management/commands/mark_successfull_99dots_records.py
```diff
@@ -0,0 +1,110 @@
...
+            repeater_type=repeater_type,
+            domain=domain,
+            next_check=datetime.utcnow(),
+            payload_id=case_id
+        )
+
+        if self.commit:
+            repeat_record.fire()
+
+        return repeat_record
```
mkangia: Might want to consider returning status only since thats the only things thats needed from repeat_record anyway.

mkangia (approved):
Looks good. Nothing blocking.

custom/enikshay/management/commands/mark_successfull_99dots_records.py
```diff
@@ -0,0 +1,110 @@
...
+
+        cases_to_update = set()
+        print("Filtering successful cases")
+        for repeat_record in with_progress_bar(records, length=count):
+            if repeat_record.attempts and any(
+                    (existing_message in attempt.message if attempt.message is not None else "")
+                    for attempt in repeat_record.attempts):
+                episode = accessor.get_case(repeat_record.payload_id)
+                if episode.get_case_property('dots_99_registered') != 'true':
+                    cases_to_update.add(episode)
```
esoergel: You may want to consider that `accessor.get_case` could fail if deleted test data made it on to this list somehow.

custom/enikshay/management/commands/mark_successfull_99dots_records.py
```diff
@@ -0,0 +1,110 @@
...
+                'beneficiary_id',
+                'episode_id',
+                'UpdatePatient Status',
+                'Adherence Status',
+                'TreatmentOutcome Status'
+            ])
+            print("Updating {} successful cases in 99DOTS".format(len(cases_to_update)))
+            for case in with_progress_bar(cases_to_update):
+                writer.writerow([
+                    get_person_case(domain, case.case_id).case_id,
```
esoergel: Using `get_person_case_from_episode` instead will save you from re-fetching the episode case each time.

custom/enikshay/management/commands/mark_successfull_99dots_records.py
```diff
@@ -0,0 +1,110 @@
...
+    def send_treatment_outcome(self, domain, case):
+        repeater_id = '3d912c5d00c73e9de5eb6a11c52a7301'
+        repeater_type = 'NinetyNineDotsTreatmentOutcomeRepeater'
+        if case.get_case_property('treatment_outcome', '') != '':
+            return self.send_repeat_record(domain, case.case_id, repeater_id, repeater_type).state
+        else:
+            return "No Treatment Outcome"
+
+    def send_repeat_record(self, domain, case_id, repeater_id, repeater_type):
+        repeat_record = RepeatRecord(
```
esoergel: Why create a new repeat record rather than a new attempt for the current one?

esoergel (approved):
Couple minor comments, nothing blocking.

custom/enikshay/management/commands/mark_successfull_99dots_records.py
mkangia: hmm.. nice catch.
I think probably because the records are already `SUCCESS` ones or to track these retriggers separately?

custom/enikshay/management/commands/mark_successfull_99dots_records.py
mkangia: One thing i found useful in Nikshay retriggers was to consider duplicates. If having `cases_to_update` already handles that, great!.

custom/enikshay/management/commands/mark_successfull_99dots_records.py
proteusvacuum: I think in most cases we don't actually have a repeat record to trigger from (these cases were never "successful" for the first repeater, which bubbles down to mean the other repeaters were never triggered at all). I could check if any of these exist first, I suppose.

custom/enikshay/management/commands/mark_successfull_99dots_records.py
proteusvacuum: yeah... there's only about 100 cases that fall into this category, but that's a good point for the future.

custom/enikshay/management/commands/mark_successfull_99dots_records.py
esoergel: Makes sense - so the top-level one has a record, but the others likely don't?  In that case, I'd only bother special-casing the top-level one if you think it'll help your auditability down the line.  This seems reasonable as-is.

proteusvacuum: The top level one isn't resent at all, we just check that it exists.

PR #18434, opened 2017-11-07
track pillow error queue size

snopoke:


corehq/ex-submodules/pillow_retry/tasks.py
```diff
@@ -75,3 +81,15 @@ def process_pillow_retry(error_doc_id):
...
         finally:
             release_lock(lock, True)
+
+
+@periodic_task(
+    run_every=crontab(minute="*/15"),
+    queue=settings.CELERY_PERIODIC_QUEUE,
+)
+def record_pillow_error_queue_size():
+    data = PillowError.objects.values('pillow').annotate(num_errors=Count('id'))
```
emord: `error_type` might also be interesting to log.

emord (approved):

corehq/ex-submodules/pillow_retry/tasks.py
snopoke: I did consider it but I think there are too many. Also it's easy enough to get that from the HQ UI for a specific pillow. Don't think it's that interesting to track over time (open to a persuation though)

emord: that's reasonable. I think the most useful distinction would be between ES infrastructure errors and everything else

PR #18438, opened 2017-11-07
Remove case history export

emord:
https://manage.dimagi.com/default.asp?262739

@sheelio I don't thing you're allowed to use google analytics on the server side. I could add a soft assert for you, or a datadog event for this, but not sure what will be good for your workflow

### DIFF
--- a/corehq/apps/export/forms.py
+++ b/corehq/apps/export/forms.py
@@ -409,6 +409,7 @@ class BaseFilterExportDownloadForm(forms.Form):
             'export_type': self._export_type,
             'name': export.name,
             'edit_url': self.get_edit_url(export),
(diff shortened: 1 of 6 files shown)


sheelio:
@emord 
Can we add the check in the UX when the user clicks to download the export, or do we not know that the export has case history?

(I think also fine to put it in datadog).

emord:
@sheelio Figured out how to pass it into the UI. Let me know if this analytics works for you, or if you have a better one in mind https://github.com/dimagi/commcare-hq/pull/18438/commits/8a93fd5167870e5a7f2e82a44967603749141769#diff-8ea4e9bb87e321b8b73978d24ed108f3R125

sheelio:
@emord 
Looks good

PR #18440, opened 2017-11-07
Increase warn period

wpride:
Product Note: This changes the banner on Web Apps Incomplete Forms and changes the message on when we will start removing incomplete forms (to Nov 20).  
For @sheelio and his precious users


sheelio:
My users are going to love you :)

PR #18444, opened 2017-11-08
Case Exploder with user sync

proteusvacuum:
Best 🐡 / 🐟 

This updates the case exploder to explode all the cases in a user's sync, not just the cases owned by that user. It also adds the ability to easily delete all cases that were created by a particular explosion event. 

The change I was most unsure of is using the `MockDevice` (and hence, moving that out of being a test util), but its API was too nice to not take advantage of... 

@calellowitz 

@millerdev (for the `MockDevice` changes)
@czue @snopoke (because you weighed in on the architecture discussion)

screenshot-2017-11-7 explode cases project settings - commcare hq


corehq/apps/hqcase/tasks.py
```diff
@@ -1,93 +1,145 @@
...
+                    S.append(m)
+        return L
+
+
+@task
+def delete_exploded_case_task(domain, explosion_id):
+    return delete_exploded_cases(domain, explosion_id, delete_exploded_case_task)
+
+
+def delete_exploded_cases(domain, explosion_id, task=None):
```
snopoke: You should probably delete the forms as well.

corehq/apps/hqcase/tests/test_explode_cases.py
```diff
@@ -8,12 +8,14 @@
...
 
-from casexml.apps.case.mock import CaseBlock
+from casexml.apps.phone.tests.test_sync_mode import BaseSyncTest
+from casexml.apps.case.mock import CaseBlock, CaseStructure, CaseIndex
 from casexml.apps.case.models import CommCareCase
 from casexml.apps.case.sharedmodels import CommCareCaseAttachment
 from casexml.apps.case.tests.util import delete_all_cases, delete_all_xforms
 from corehq.apps.app_manager.tests.util import TestXmlMixin
-from corehq.apps.hqcase.tasks import explode_cases
+from corehq.apps.hqcase.tasks import explode_cases_2 as explode_cases
```
snopoke: Should just be `explode_cases`

corehq/apps/hqcase/tests/test_explode_cases.py
```diff
@@ -212,3 +214,84 @@ def test_parent_child(self):
...
+
+        C = CaseStructure(
+            case_id='child',
+            attrs={'create': True, 'owner_id': '-'},
+            indices=[CaseIndex(
+                E1,
+                identifier='extension_1',
+                relationship='extension',
+                related_type=case_type,
+            )]
```
snopoke: This index is pointing in the wrong direction. It should go from E1 -> C. Same applies for others.

corehq/apps/hqcase/tests/test_explode_cases.py
```diff
@@ -212,3 +214,84 @@ def test_parent_child(self):
...
+                identifier='extension',
+                relationship='extension',
+                related_type=case_type,
+            )]
+        )
+        self.device.post_changes([O, E2])
+
+    def test_case_graph(self):
+        cases = self.device.restore().cases
+        self.accessor.get_case_ids_in_domain
```
snopoke: Unused line

corehq/apps/hqcase/tests/test_explode_cases.py
```diff
@@ -212,3 +214,84 @@ def test_parent_child(self):
...
+                related_type=case_type,
+            )]
+        )
+
+        E2 = CaseStructure(
+            case_id='extension_2',
+            attrs={'create': True, 'owner_id': '-'},
+            indices=[CaseIndex(
+                C,
+                identifier='extension',
```
snopoke: should be 'host'. Could also leave it out since `CaseIndex` will set it based on `relationship`

corehq/apps/hqcase/tests/test_explode_cases.py
```diff
@@ -212,3 +214,84 @@ def test_parent_child(self):
...
+                related_type=case_type,
+            )]
+        )
+
+        O = CaseStructure(
+            case_id='owned',
+            attrs={'create': True},
+            indices=[CaseIndex(
+                C,
+                identifier='child',
```
snopoke: should be 'parent'

corehq/apps/hqcase/tests/test_explode_cases.py
```diff
@@ -212,3 +214,84 @@ def test_parent_child(self):
...
+                C,
+                identifier='extension',
+                relationship='extension',
+                related_type=case_type,
+            )]
+        )
+        self.device.post_changes([O, E2])
+
+    def test_case_graph(self):
+        cases = self.device.restore().cases
```
snopoke: Will probably need `flag_enabled('EXTENSION_CASES_SYNC_ENABLED')` to get extensions to sync properly once relationships are fixed

corehq/apps/hqcase/tests/test_explode_cases.py
```diff
@@ -212,3 +214,84 @@ def test_parent_child(self):
...
+        self.accessor.get_case_ids_in_domain
+        graph = CaseGraph(cases)
+        self.assertEqual(
+            ['extension_1', 'child', 'extension_2', 'owned'],
```
snopoke: Correct order for the graph in the diagram is `['owned', 'child', 'extension_2', 'extension_1']` (last 2 are interchangeable)

corehq/apps/hqcase/tasks.py
```diff
@@ -1,93 +1,145 @@
...
+        for case_id, case in cases.iteritems():
+            indices = [idx.case_id for idx in case.index.itervalues()]
+            self.graph[case.case_id] = indices
+            for index in indices:
+                self.inverse_graph[index].append(case.case_id)
+
+            if not indices:
+                self.roots.append(case.case_id)
+
+    def topological_sort(self):
```
snopoke: will this be deterministic where there are multiple cases on the same level? Think the values in the inverse graph should be sorted.

snopoke (changes requested):
Very cool. Tests need some fixing up.

The parts that use the `case_search` index will only work on domains that have that enabled. Since this is an 'internal' feature you could just give that warning up front.

czue:
wow this looks like quite an effort! :clap: 

I think @snopoke did a more thorough/quality review than I'll be able to so will not review in detail unless you want me to

corehq/apps/hqcase/tests/test_explode_cases.py
proteusvacuum: I think I just named these cases in a very confusing manner (copied from a different test in the extension cases sync code). I've updated the names to make things clearer.

corehq/apps/hqcase/tests/test_explode_cases.py
proteusvacuum: This test actually works without that flag on since there's no ownership cleanliness flag for this domain.

corehq/apps/hqcase/tests/test_explode_cases.py
proteusvacuum: oops.

corehq/apps/hqcase/tasks.py
```diff
@@ -102,33 +102,109 @@ def submit_case(case, new_case_id, domain, source, new_parent_ids=dict()):
...
+
+        return ancestors
+
+    def topological_sort(self):
+        """returns all cases in topological order
+
+        Using Kahn's algorithm from here:
+        https://en.wikipedia.org/wiki/Topological_sorting
+
+        Assumes there are no cycles in the graph
```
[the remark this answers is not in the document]
proteusvacuum: I was debating this. This is called with cases that come out of a sync operation, and I don't think we allow you to have cycles there. I've added a check for now, but I'm not sure it will ever catch anything, and has to loop through each of the cases again. What do you think?

corehq/apps/hqcase/tasks.py
```diff
@@ -1,93 +1,145 @@
...
+        L = []
+        S = copy(self.roots)
+        while len(S) > 0:
+            n = S.pop()
+            L.append(n)
+            for m in inverse_graph[n]:
+                graph[m].remove(n)
+                if len(graph[m]) == 0:
+                    S.append(m)
+        return L
```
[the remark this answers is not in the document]
proteusvacuum: good point. originally I'd had a bunch of other CaseGraph methods, that I deleted since I didn't use them. This feels much better now!

proteusvacuum:
Thanks for the thorough review @snopoke and @millerdev! I've tried to address all your points in the new commits.

proteusvacuum:
@millerdev I was also wondering if you think it would make sense for the restore to use `livequery`?

corehq/apps/hqcase/tasks.py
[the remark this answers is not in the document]
millerdev: I think it's a good conservative move, even if it's unlikely to ever catch anything. The run time expense is likely not prohibitive, and it will be nice to know that it fails predictably if it finds unexpected cycles (rather than infinite recursion or loop forever or whatever).

corehq/apps/hqcase/tasks.py
```diff
@@ -1,93 +1,164 @@
...
+            roots.append(case.case_id)
+
+    # sort graph
+    sorted_ids = []
+    while len(roots) > 0:
+        root_id = roots.pop()
+        sorted_ids.append(root_id)
+        for case_id in sorted(inverse_graph[root_id]):
+            graph[case_id].remove(root_id)
+            if len(graph[case_id]) == 0:
```
millerdev: nit: `if not graph[case_id]:`

millerdev (approved):
> wondering if you think it would make sense for the restore to use livequery?

Sure, seems great to me.

All other feedback is nitpicking. You can ignore if you want.

snopoke (approved):

sheelio:
@proteusvacuum 
I'm going to retag this as product/feature-flag (I think this is a change that will affect users who are using the case exploder?)

proteusvacuum:
@sheelio technically, case exploder isn't feature flagged... It's just hidden to everyone except super users.

sheelio:
Ah - I got confused between this and the load testing users functionality.

PR #18447, opened 2017-11-08
Added command to export non-aadhar records

kkrampa:


kkrampa:
@snopoke 
I tested this locally on 1 million rows where ~40% were non-aadhar records. Size of the list in Python was about 3MB in memory. We can roughly calculate that for 16M records it gives us 120MB. I'm curious if we can handle that at once on ICDS or I should chunk results on the SQL side.

kkrampa:
@snopoke 
Also, there is a second way to solve this. COPY command has a possibility to export to CSV so we can run that directly in SQL but then the export file will be saved on the postgres machine and table will be locked at the export time.

snopoke:
Thanks, I'll test this out tomorrow.

PR #18448, opened 2017-11-08
 ignore user and domain sync_delay for v2

snopoke:
When looking at this I was also wondering if we need a way to override this for specific users e.g. during training or for 'test' users.

### DIFF
--- a/corehq/apps/app_manager/fixtures/mobile_ucr.py
+++ b/corehq/apps/app_manager/fixtures/mobile_ucr.py
@@ -34,28 +34,12 @@ from corehq.apps.app_manager.dbaccessors import (
 MOBILE_UCR_RANDOM_THRESHOLD = 1000
 
 
-def _should_sync(restore_state, is_v2):
+def _should_sync(restore_state):
     last_sync_log = restore_state.last_sync_log
     if not last_sync_log or restore_state.overwrite_cache:
         return True
 
-    sync_interval = None
-    if not is_v2:
-        # in v2 sync delay is handled per report in the app
-        sync_interval = restore_state.restore_user.get_mobile_ucr_sync_interval()
-        if sync_interval is None and restore_state.params.app:
-            app = restore_state.params.app
-            if restore_state.params.app.copy_of:
-                # get sync interval from latest app version so that we don't have to deploy a new version
-                # to make changes to the sync interval
-                try:
-                    app = get_brief_app(restore_state.domain, restore_state.params.app.copy_of)
-                except NoResultFound:
-                    pass
-            sync_interval = app.mobile_ucr_sync_interval
-        if sync_interval is None:
-            sync_interval = restore_state.project.default_mobile_ucr_sync_interval
-
+    sync_interval = restore_state.project.default_mobile_ucr_sync_interval
     sync_interval = sync_interval and sync_interval * 3600  # convert to seconds
     return (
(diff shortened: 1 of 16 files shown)


corehq/apps/app_manager/fixtures/mobile_ucr.py
```diff
@@ -34,24 +34,27 @@
...
-            try:
-                app = get_brief_app(restore_state.domain, restore_state.params.app.copy_of)
-            except NoResultFound:
-                pass
-        sync_interval = app.mobile_ucr_sync_interval
-    if sync_interval is None:
-        sync_interval = restore_state.project.default_mobile_ucr_sync_interval
+    sync_interval = None
+    if not is_v2:
+        # in v2 sync delay is handled per report in the app
```
emord: I intentionally left this in case we wanted a hard minimum for an app, but I suppose multiple configurations could be confusing. Is that your thought for removing this?

emord (commented):
Yeah. I could see a usecase for overriding this for trainings, although i think for trianings they just use a demo user right? so it's a static restore anyways?

corehq/apps/app_manager/fixtures/mobile_ucr.py
snopoke: yea I think it's a bit confusing but I guess having a hard min could be useful. I'll think about it some more.

emord: Cool. I'm not against this change, so happy to merge once tests pass.

we could also use the `<input min>` attribute to specify a minimum on each input to make it less confusing, although we may want some warning that the minimum is due to an app /domain level setting

snopoke: Went back and forth on this for ages and decided to remove the user and app settings so that it's just the domain and individual report settings.

Nobody is using the user / app settings currently anyway.

snopoke: Still need to make it possible to ignore the sync delay for test users.

corehq/apps/app_manager/models.py
```diff
@@ -5284,8 +5284,6 @@ class Application(ApplicationBase, TranslationMixin, HQMediaMixin):
     use_grid_menus = BooleanProperty(default=False)
     grid_form_menus = StringProperty(default='none',
                                      choices=['none', 'all', 'some'])
-    mobile_ucr_sync_interval = IntegerProperty()
```
emord: I've always been hesitant to remove properties like this. If we ever make a new property that has the same name, then these apps could have a value that means something different in the future.

emord (commented):
Is this no longer used on ICDS?

corehq/apps/app_manager/models.py
snopoke: True, though I don't think it's worth keeping the code just for that (especially in this case since there are only 3 apps that I know of that have values set for this property).

snopoke:
This isn't needed for ICDS. I checked and all apps either have no value set or the same value as the domain value.

emord (approved):
Looks good to me. Do you think we will never need this again even with the coming up rollouts before we get servers?

snopoke:
I don't think so. We will need a way to ignore the sync delay for test
users but I don't think we'll need the app level settings.

PR #18449, opened 2017-11-08
Add update-code.sh and hammer.sh to README

orangejenny:
@amstone326


amstone326:
Thanks @orangejenny ! (merging since this is just a readme change)

amstone326:
@orangejenny Sorry this didn't occur to me until after I merged it and started thinking about using the instructions myself, but- Is there a brief (1-2 sentence) explanation that could be added to describe in what instances/how frequently these 2 scripts might be used? Specifically like when you would want to use one over the other, since they seem similar.

PR #18451, opened 2017-11-08
Add maintenance scripts to repo

amstone326:
These have been living in separate gists only; seems silly not to just keep them in the scripts/ dir, which already has some other stuff like this.


orangejenny (approved):

orangejenny:
thanks

PR #18452, opened 2017-11-08
Update LESS sections of README

orangejenny:
@biyeun Does this look reasonable? For all practical purposes, we advise everyone to just use option 1 unless you're debugging compression issues. And we don't need two versions of LESS anymore!


biyeun:
Beautiful.

biyeun:
thank you for taking care of the LESS removal 🙏

PR #18454, opened 2017-11-08
Removed es_head from docker file.

emord:
@snopoke this won't work when we upgrade (due to changes in plugins) and it can be installed as a chrome plugin now https://chrome.google.com/webstore/detail/elasticsearch-head/ffmkiejjmecolpfloofpjologoblkegm/

### DIFF
--- a/docker/Dockerfile_es
+++ b/docker/Dockerfile_es
@@ -1,7 +1,0 @@
-FROM elasticsearch:1.7.4
-
-# Install HEAD plugin
-# https://mobz.github.io/elasticsearch-head/
-RUN plugin -i mobz/elasticsearch-head
-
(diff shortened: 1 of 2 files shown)


snopoke (approved):

emord:
@dimagi/team-commcare-hq FYI If you rebuild your ES container, it won't have es head anymore. You can download a chrome plugin to do the same thing https://chrome.google.com/webstore/detail/elasticsearch-head/ffmkiejjmecolpfloofpjologoblkegm/

esoergel:
Or if you don't use chrome:
```
$ git clone git://github.com/mobz/elasticsearch-head.git
```
Then open `index.html` from that repo

PR #18455, opened 2017-11-08
make xml optional in admin restore

nickpell:
https://trello.com/c/Lq73nOLU/250-restore-page-with-stats-only-no-xml
Product Note: The page that lets you download a User Restore can be configured to just show you how long the restore takes and other restore stats (instead of the actual restore).  This is useful if your restore is massive and crashes the browser.

### DIFF
--- a/corehq/apps/hqadmin/templates/hqadmin/admin_restore.html
+++ b/corehq/apps/hqadmin/templates/hqadmin/admin_restore.html
@@ -34,7 +34,7 @@
                         Number of locations in this restore: <strong>{{num_locations}}</strong>
                     </div>
 
-                    {% if show_xml %}
+                    {% if not hide_xml %}
                     <div>
                         <ul class="nav nav-tabs" role="tablist">
                             <li role="presentation" class="active"><a href="#payload-tab" role="tab" data-toggle="tab">Payload</a></li>
@@ -67,7 +67,7 @@
                                         {% endfor %}
                                     </tbody>
                                 </table>
-                    {% if show_xml %}
+                    {% if not hide_xml %}
                             </div>
                         </div>
                     </div>
--- a/corehq/apps/hqadmin/templates/hqadmin/admin_restore.html
+++ b/corehq/apps/hqadmin/templates/hqadmin/admin_restore.html
@@ -34,14 +34,14 @@
                         <div>
                             Number of cases by case type
                         </div>
-                    {% endif %}
                     <div>
-                        {% for case_type, count in case_type_counts.items %}
-                            <div>
-                                &emsp;{{case_type}}: {{count}}
-                            </div>
-                        {% endfor %}
+                        <ul>
+                            {% for case_type, count in case_type_counts.items %}
+                                <li>{{case_type}}: {{count}}</li>
+                            {% endfor %}
+                        </ul>
                     </div>
+                    {% endif %}
                     <div>
                         Number of locations in this restore: <strong>{{num_locations}}</strong>
                     </div>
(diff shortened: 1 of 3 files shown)


nickpell:
@sravfeyn bump

corehq/apps/hqadmin/views.py
```diff
@@ -531,13 +531,15 @@ def get_context_data(self, **kwargs):
             num_cases = 0
             num_locations = 0
         formatted_payload = etree.tostring(xml_payload, pretty_print=True)
+        hide_xml = self.request.GET.get('hide_xml') == 'true'
```
esoergel: Is there any substantial work that can be skipped if the xml isn't needed?  I'm guessing most of it needs to be run for the timings to be accurate, huh?

corehq/apps/hqadmin/templates/hqadmin/admin_restore.html
```diff
@@ -64,9 +67,11 @@
                                         {% endfor %}
                                     </tbody>
                                 </table>
+                    {% if not hide_xml %}
                             </div>
                         </div>
                     </div>
+                    {% endif %}
```
esoergel: This is kinda weird.  Could you instead just hide the xml part (and replace it with "xml hidden" or something), rather than hiding the wrapper stuff?  Just thinking that this will make it hard to modify the html of this page in the future.

proteusvacuum:
Were you also going to add the "restore stats" part? I think that was a big part of the ask here, if I understood why Kriti wanted this. Should be easy to grab from the XML itself, like we do with the num locations and num cases.

corehq/apps/hqadmin/templates/hqadmin/admin_restore.html
nickpell: good call, updated

corehq/apps/hqadmin/views.py
nickpell: No, we actually need to get the restore for timing purposes

corehq/apps/hqadmin/templates/hqadmin/admin_restore.html
```diff
@@ -40,13 +45,33 @@
...
+                        Number of reports in this restore: <strong>{{num_reports}}</strong>
+                    </div>
+                    {% if report_row_counts %}
+                        <div>
+                            Number of rows by report
+                        </div>
+                    {% endif %}
+                    <div>
+                        {% for report_id, row_count in report_row_counts.items %}
+                            <div>
```
proteusvacuum: is there a reason you didn't use a `<ul><li>` for these?

nickpell: Nope, changed

nickpell:
✅

corehq/apps/hqadmin/templates/hqadmin/admin_restore.html
```diff
@@ -41,7 +81,11 @@
 
                         <div class="tab-content">
                             <div role="tabpanel" class="tab-pane active" id="payload-tab">
-                                <pre id="payload">{{ payload }}</pre>
+                                {% if not hide_xml %}
+                                    <pre id="payload">{{ payload }}</pre>
+                                {% else %}
+                                    XML is hidden
+                                {% endif %}
```
esoergel: :+1:

corehq/apps/hqadmin/views.py
```diff
@@ -511,10 +511,27 @@ def get_context_data(self, **kwargs):
...
+            num_locations = len(locations)
+            location_type_counts = dict(Counter(location.attrib['type'] for location in locations))
+            reports = xml_payload.findall(
+                "{{{0}}}fixture[@id='commcare:reports']/{{{0}}}reports/".format(RESPONSE_XMLNS)
+            )
+            num_reports = len(reports)
+            report_row_counts = {
+                report.attrib['report_id']: len(report.findall('{{{0}}}rows/{{{0}}}row'.format(RESPONSE_XMLNS)))
+                for report in reports
+            }
```
esoergel: This method is getting pretty big, might be time to pull out these lookups into a helper

esoergel (approved):

PR #18456, opened 2017-11-08
Include attachments in form repeaters for SQL domains

esoergel:
https://manage.dimagi.com/default.asp?265044
Product Note: Data forwarding with forms will also include any attachments (like pictures). This didn't used to work on newer SQL based projects

### DIFF
--- a/corehq/apps/api/util.py
+++ b/corehq/apps/api/util.py
@@ -48,11 +48,17 @@ def get_obj(bundle_or_obj):
         return bundle_or_obj
 
 
-def form_to_es_form(xform_instance):
+def form_to_es_form(xform_instance, include_attachments=False):
+    # include_attachments is only relevant for SQL domains; they're always
+    # included for Couch domains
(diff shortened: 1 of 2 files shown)


corehq/apps/api/util.py
```diff
@@ -48,11 +48,17 @@ def get_obj(bundle_or_obj):
...
-def form_to_es_form(xform_instance):
+def form_to_es_form(xform_instance, include_attachments=False):
+    # include_attachments is only relevant for SQL domains; they're always
+    # included for Couch domains
     from corehq.pillows.xform import transform_xform_for_elasticsearch, xform_pillow_filter
     from corehq.apps.api.models import ESXFormInstance
+    from corehq.form_processor.models import XFormInstanceSQL
 
-    json_form = xform_instance.to_json()
+    if include_attachments and isinstance(xform_instance, XFormInstanceSQL):
```
esoergel: Is there a better way to tell if this is a sql form?  I didn't see any `is_sql` type things on the instance.

snopoke: I don't think there's a better way

PR #18459, opened 2017-11-09
added ga tracking action for marketing on new user creation

biyeun:
@orangejenny @calellowitz

### DIFF
--- a/corehq/apps/app_manager/static/app_manager/js/releases/app_view_release_manager.js
+++ b/corehq/apps/app_manager/static/app_manager/js/releases/app_view_release_manager.js
@@ -51,6 +51,7 @@ hqDefine("app_manager/js/releases/app_view_release_manager", function() {
         analytics.workflow('Visited the Release Manager');
         if (initial_page_data('confirm')) {
             analytics.usage('User actions', 'User created login', window.location.pathname);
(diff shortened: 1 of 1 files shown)


corehq/apps/app_manager/static/app_manager/js/releases/app_view_release_manager.js
```diff
@@ -51,6 +51,7 @@ hqDefine("app_manager/js/releases/app_view_release_manager", function() {
         analytics.workflow('Visited the Release Manager');
         if (initial_page_data('confirm')) {
             analytics.usage('User actions', 'User created login', window.location.pathname);
+            analytics.usage('User actions', 'Forms', 'Name Your First Project');
```
calellowitz: do we need to ga actions here for the same event? It seems these could be consolidated (either by having marketing use the existing one, or whoever is using that use the new one)

calellowitz (changes requested):

corehq/apps/app_manager/static/app_manager/js/releases/app_view_release_manager.js
biyeun: I think the idea here is that the value is different for `window.location`, but perhaps there is a separate event "Forms" and there are other "Forms" events to compare to? worthwhile to ask if the first one is still necessary

PR #18460, opened 2017-11-09
async xform rebuild

emord:
https://manage.dimagi.com/default.asp?264481#1419104

commit by commit. This does a few things.

* adds an interface iterate over form ids by a particular xmlns. this isn't indexed in SQL (other than the one by domain), but before it was only filtering by domain anyways, so it should still be better
* Interprets the xmlns/case types in a data source
* changes the resume to be resumed by xmlns and case type instead of by last id. before last id only worked on couch anyways. it was never implemented for sql
* removed recalculate indicators (was added as a workaround https://github.com/dimagi/commcare-hq/pull/17368)
* really large async tables (enikshay) can now be rebuilt via teh rebuild button instead of the hacky management command ( @calellowitz )

This is probably going to break tests and I haven't fully tested it yet but going to look into that in the morning.

buddy @snopoke


corehq/apps/userreports/models.py
```diff
@@ -351,14 +351,22 @@ def get_es_index_settings(self):
...
+
+            if config_filter['operator'] not in ('eq', 'in'):
+                return [None]
+
             expression = config_filter['expression']
             if expression['type'] == 'property_name' and expression['property_name'] == prop_name:
-                return config_filter['property_value']
-            return None
+                multiple = config_filter['operator'] == 'in'
+                return config_filter['property_value'] if multiple else config_filter['property_value']
```
snopoke: Both sides of this `if` are the same.

corehq/apps/userreports/tasks.py
```diff
@@ -120,51 +119,46 @@ def resume_building_indicators(indicator_config_id, initiated_by=None):
...
+    for case_type_or_xmlns in case_type_or_xmlns_list:
+        relevant_ids = []
+        document_store = get_document_store(
+            config.domain, config.referenced_doc_type, case_type_or_xmlns=case_type_or_xmlns
+        )
 
-    if not doc_id_provider:
-        doc_id_provider = document_store.iter_document_ids(last_id)
+        if not doc_id_provider:
+            doc_id_provider = document_store.iter_document_ids()
```
snopoke: I think this will only ever happen on the first loop since after that it will always have a value. That seems like it will be an issue since each loop will have a different `case_type` / `xmlns` filter.

corehq/form_processor/backends/sql/dbaccessors.py
```diff
@@ -380,6 +380,18 @@ def iter_forms_by_last_modified(start_datetime, end_datetime):
         )
 
     @staticmethod
+    def iter_form_ids_by_xmlns(domain, xmlns=None):
+        from corehq.sql_db.util import run_query_across_partitioned_databases
+
+        q_expr = Q(domain=domain)
```
snopoke: Should also filter by `state = XFormInstanceSQL.NORMAL`

snopoke (commented):
Not very familiar with this code but looks reasonable. 

Interesting that we have multiple 'resumable' iteration methods.

Just confirming that this won't be able to resume from the middle of an iteration for a particular filter value correct?

emord:
> Interesting that we have multiple 'resumable' iteration methods.

Do you mean by id and by case type/xmlns?  Or do you mean the difference between rebuilding the table and rebuilding in place?

> Just confirming that this won't be able to resume from the middle of an iteration for a particular filter value correct?

That's correct. And it was never implemented on SQL domains anyways so resume would always just restart a build there. The large UCRs from couch that I've seen have multiple xmlns so there would still be at least some form of resume available, and we could also change it to an async data source so that it's handled by the async queues if necessary.

In the future for sql I'd like to have a resume structure that keeps track of the finished database partitions and also sorts by the primary key, so that it's resumable from an xmlns/case type and id perspective. To do that efficiently we'd need an index on [domain, uuid] though and I'd rather not add more indexes to forms/cases for now

snopoke:
I was just meaning different ways we do resumable things like this in HQ. I
was thinking specifically of the `DocumentProvider` classes which are used
by the reindexers the blob db migrations. I think it should be possible to
add `xmlns` and `case_type` filtering to the `ReindexAccessors` which would
make it possible to do proper resumable rebuilds. We'd still need a tool
like the `DataSourceResumeHelper` to keep track of the DB / XMLNS / case
types.

emord:
Oh. I see now. It's possible that the UCR rebuild code was written before that was a thing? I think that UCRs were written before the SQL backend so it could have just been tacked on later and not rewritten?

corehq/form_processor/document_stores.py
```diff
@@ -21,14 +22,17 @@ def __init__(self, domain, xmlns=None):
 
     def get_document(self, doc_id):
         try:
-            return self.form_accessors.get_form(doc_id).to_json(include_attachments=True)
+            form = self.form_accessors.get_form(doc_id)
+            if isinstance(form, XFormInstanceSQL):
+                return form.to_json(include_attachments=True)
+            else:
+                return form.to_json()
```
emord: This is a little nasty. pulled it from https://github.com/dimagi/commcare-hq/pull/18456/files#diff-551bb43505398f33e4901de767d67560R56

@esoergel let me know if you came up with a utility to do it or decide on a better way

corehq/apps/userreports/models.py
emord: https://github.com/dimagi/commcare-hq/pull/18460/commits/d580ddfff0d45ef615b924e43e98d7661006ecf9

corehq/apps/userreports/tasks.py
emord: It was only used in one place (custom mgmt command) and it's no longer needed

https://github.com/dimagi/commcare-hq/pull/18460/commits/5f1769cce65cccef77e8d70bb4ce1ec9d8ace27c

corehq/form_processor/backends/sql/dbaccessors.py
emord: https://github.com/dimagi/commcare-hq/pull/18460/commits/4a69360c141b703cf2883ebecd860e69703dd2b3

custom/icds_reports/ucr/tests/base_test.py
```diff
@@ -84,6 +84,7 @@ def _get_query_object(self):
 
     def _run_iterative_monthly_test(self, case_id, cases, start_date=date(2015, 12, 1)):
         _iteratively_build_table(self.datasource)
```
emord: Wanted to change this to use rebuild_indicators, but it kept failing for some reason

corehq/form_processor/document_stores.py
esoergel: Nope, just left it like that.  I did have two thoughts that I'd be in favor of, though:

We currently insert a field called `backend_id` (either `'couch'` or `'sql'`) before sending to to elasticsearch.  This could be added as a class attribute to `XFormInstance` and `XFormInstanceSQL`.

More specific to this issue, we could also override `to_json` on `XFormInstance` (right now I believe it's just the default `Document` implementation) and have it accept and ignore `include_attachments`.  It's kinda smelly to have a parameter that doesn't do anything, but I think it's less smelly than presenting a semi-opaque interface with params that only work sometimes (We similarly add and  ignore the `domain` parameter to make `CommCareUser` act like `WebUser`).

PR #18464, opened 2017-11-09
record metric for sync interval

snopoke:


emord (approved):
Is this for figuring out if we can keep less than 60 days?

snopoke:
@emord yes

PR #18468, opened 2017-11-09
Changed clause for Result of DST

kkrampa:
@calellowitz


calellowitz (approved):
kicked the build, but looks good to me

PR #18473, opened 2017-11-09
Remove LESS_FOR_BOOTSTRAP_3_BINARY

orangejenny:
@emord 

after https://github.com/dimagi/commcarehq-ansible/pull/1019 is deployed


orangejenny:
I think I can delete more code here, but for now just going to get rid of the setting.

emord:
deploying that ansible change now

PR #18474, opened 2017-11-09
apply KO bindings

nickpell:
Makes sure that the import button for the data dictionary import is disabled before a file is uploaded.

https://manage.dimagi.com/default.asp?264926


nickpell:
bump @sravfeyn

PR #18477, opened 2017-11-10
handle missing synclog

snopoke:
Also add bucket for initial sync


snopoke:
https://sentry.io/dimagi/commcarehq/issues/402078494/

emord:
Curious if you have a threshold for making this decision, or how long you want to run before making a final decision?

Looking at it now and it's a couple of orders of magnitude difference between last 2 days and last 7 days and there's very few past that. Dropping down to keeping these for a month would halve our largest couchdb and down to a week would be almost a tenth (though that would be an aggressive cut to do at once). could give us much better performance?

PR #18348, opened 2017-10-31
add notification time report

mkangia:
https://trello.com/c/CKBqy6CG/252-gather-stats-to-check-for-time-take-for-nikshay-notifications

buddy @emord

PR #18481, opened 2017-11-10
Fix email for report

mkangia:
fixing issue in #18348 

cc: @proteusvacuum

### DIFF
--- a/custom/enikshay/management/commands/nikshay_registration_notification_time_report.py
+++ b/custom/enikshay/management/commands/nikshay_registration_notification_time_report.py
@@ -1,3 +1,5 @@
+from __future__ import absolute_import
+
 import datetime
 import csv
 import pytz
--- a/custom/enikshay/management/commands/nikshay_registration_notification_time_report.py
+++ b/custom/enikshay/management/commands/nikshay_registration_notification_time_report.py
@@ -4,8 +4,9 @@ import pytz
 
 from django.core.management.base import BaseCommand
 from django.utils.dateparse import parse_datetime
+from django.core.mail import EmailMessage
+from django.conf import settings
 
-from corehq.apps.hqwebapp.tasks import send_html_email_async
 from corehq.form_processor.interfaces.dbaccessors import CaseAccessors
 from corehq.motech.repeaters.dbaccessors import iter_repeat_records_by_domain
 from casexml.apps.case.util import get_case_property_changed_info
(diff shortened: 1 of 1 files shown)


custom/enikshay/management/commands/nikshay_registration_notification_time_report.py
```diff
@@ -76,17 +77,14 @@ def handle(self, days, *args, **options):
...
             if email_to:
-                attachment = {
-                    'title': result_file_name,
-                    'mimetype': 'text/csv',
-                    'file_obj': csvfile,
-                }
-
-                message = "Report for time taken for registration notifications yesterday till now"
-                send_html_email_async(
+                email_to = list(email_to) if not isinstance(email_to, basestring) else [email_to]
```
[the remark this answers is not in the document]
nickpell: ```basestring``` should be ```six.string_types```

https://pythonhosted.org/six/#six.string_types

custom/enikshay/management/commands/nikshay_registration_notification_time_report.py
```diff
@@ -76,17 +79,14 @@ def handle(self, days, *args, **options):
...
                     subject="Nikshay Registration Notification Time Report",
-                    recipient=email_to,
-                    html_content=message,
-                    text_content=message,
-                    file_attachments=[attachment]
+                    body="Report for time taken for registration notifications for %s day(s)" % self.days,
+                    to=email_to,
+                    from_email=settings.DEFAULT_FROM_EMAIL
                 )
+                email.attach(filename=result_file_name, content=csvfile.read())
```
proteusvacuum: Does this need a mimetype? `text/csv`

custom/enikshay/management/commands/nikshay_registration_notification_time_report.py
```diff
@@ -76,17 +79,14 @@ def handle(self, days, *args, **options):
...
-                    'title': result_file_name,
-                    'mimetype': 'text/csv',
-                    'file_obj': csvfile,
-                }
-
-                message = "Report for time taken for registration notifications yesterday till now"
-                send_html_email_async(
+                email_to = list(email_to) if not isinstance(email_to, basestring) else [email_to]
+                csvfile = open(result_file_name)
+                email = EmailMessage(
```
proteusvacuum: curious what prompted this change?

mkangia: the issue i was trying to address was `send_html_email_async` expects a StringIO object on which it would get_value later. Here we had a csv file so i thought it was better to just attach the file here then.

custom/enikshay/management/commands/nikshay_registration_notification_time_report.py
mkangia: I think its determined automatically when not passed.

from [docs](https://docs.djangoproject.com/en/dev/topics/email/)
```
 mimetype is the optional MIME type for the attachment. If you omit mimetype, the MIME content type will be guessed from the filename of the attachment.
```

PR #18482, opened 2017-11-10
Enikshay changes

kkrampa:
@calellowitz


calellowitz (changes requested):
why include the `multiple_value_string_translation` function if it is not used?

kkrampa:
@calellowitz We need that kind of functionality in couple of properties, I'm going to test this in non-static reports which Jordan is currently testing.

calellowitz:
@kkrampa makes sense

calellowitz (approved):

PR #18483, opened 2017-11-10
Encode all utf-8 strings

kkrampa:


custom/icds_reports/management/commands/sql_scripts/select_non_aadhar.sql
```diff
@@ -1,10 +1,10 @@
 SELECT
-  name,
+  convert_to(name, 'SQL_ASCII'),
```
kkrampa: @snopoke 
I know that we have utils for unicode writer in ```dimagi.utils``` but I think it can be a little bit faster to do that in the db. I tested this and it correctly converts all non-ascii characters to code representation and leaves ascii characters untouched.

kkrampa:
output for test data:
ascii

custom/icds_reports/management/commands/sql_scripts/select_non_aadhar.sql
snopoke: any reason not to use `utf-8`?

kkrampa: no, I just misunderstood postgres documentation. Tested and it works fine.

PR #18485, opened 2017-11-10
Start Enterprise: Don't fail hard on duplicate message ids

gcapalbo:
@orangejenny

### DIFF
--- a/corehq/messaging/smsbackends/start_enterprise/models.py
+++ b/corehq/messaging/smsbackends/start_enterprise/models.py
@@ -4,6 +4,7 @@ import re
 import requests
 from dimagi.utils.logging import notify_exception
 from django.db import models
+from django.db import IntegrityError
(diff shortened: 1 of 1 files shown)


orangejenny:
This doesn't look like it changes functionality - is it just for code clarity, or am I missing something?

gcapalbo:
@orangejenny it now catches the IntegrityError raised when a duplicate message id is attempted to be inserted into the table. see the comment i left in the code for the explanation

orangejenny:
But wouldn't the `set` previously there have prevented there from being any duplicate message ids?

gcapalbo:
@orangejenny it turns out the duplicate message ids are getting returned from different requests - i had used `set()` because i thought they were coming from the same request, but it turns out that's not the case

orangejenny:
Got it - thanks for explaining.

orangejenny (approved):

PR #18491, opened 2017-11-10
python-modernize . -w -f libmodernize.fixes.fix_input_six

nickpell:
@dimagi/py3

### DIFF
--- a/scripts/couchdb/utils.py
+++ b/scripts/couchdb/utils.py
@@ -5,6 +5,7 @@ from collections import namedtuple
 
 import requests
 import time
+from six.moves import input
 
 NodeDetails = namedtuple('NodeDetails', 'ip port node_local_port username password')
 
@@ -34,7 +35,7 @@ def get_membership(node_details):
 
 
 def confirm(msg):
-    return raw_input(msg + "\n(y/n)") == 'y'
+    return input(msg + "\n(y/n)") == 'y'
 
 
 def get_arg_parser(command_description):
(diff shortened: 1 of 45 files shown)


scripts/couchdb/utils.py
```diff
@@ -34,7 +35,7 @@ def get_membership(node_details):
 
 
 def confirm(msg):
-    return raw_input(msg + "\n(y/n)") == 'y'
+    return input(msg + "\n(y/n)") == 'y'
```
emord: @dannyroberts should scripts/couchdb get removed from HQ since you created that other repo?

emord (approved):
not familiar with the input vs raw_input changes on py3, but this is only mgmt commands so i'm not going to worry about it

emord (approved):

scripts/couchdb/utils.py
dannyroberts: Oh. Yeah, totally. Man I've been forgetting things like it's my job. I'll go clean that up

dannyroberts:
> not familiar with the input vs raw_input changes on py3, but this is only mgmt commands so i'm not going to worry about it

I believe in python2 `input` is basically `lambda msg: eval(raw_input(msg))` or something horrible like that. They changed it in python3 so that the innocuous-sounding `input` function just receives user input and doesn't also evaluate it as python. Since this is what `raw_input` does in python2, that basically means `input` => ø and `raw_input` => `input`.

scripts/couchdb/utils.py
dannyroberts: https://github.com/dimagi/commcare-hq/pull/18498

PR #18493, opened 2017-11-10
Update feature flag settings

sheelio:
Makes the following feature flags project only (no per user setting):
- Case List Tile
- Case List Lookup
- Multimedia Case Properties
- UCR

Change categories for:
- Mobile Privileges (now Internal)
- Mobile Worker Self Registration (now Custom)
- SMS Performance Feedback (now Custom)
- Anonymous Web Apps Usage (now Custom)

### DIFF
--- a/corehq/toggles.py
+++ b/corehq/toggles.py
@@ -380,7 +380,7 @@ CASE_LIST_TILE = StaticToggle(
     'case_list_tile',
     'Allow configuration of case list tiles',
     TAG_SOLUTIONS,
-    [NAMESPACE_DOMAIN, NAMESPACE_USER]
+    [NAMESPACE_DOMAIN]
 )
 
 SHOW_PERSIST_CASE_CONTEXT_SETTING = StaticToggle(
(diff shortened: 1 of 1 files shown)


sheelio:
@kaapstorm 
Can you take a look at this?

kaapstorm (approved):

kaapstorm:
LGTM

corehq/toggles.py
```diff
@@ -477,7 +477,7 @@ def toggle_values_by_name(username=None, domain=None):
     'user_reports',
     'User configurable reports UI',
     TAG_SOLUTIONS,
-    [NAMESPACE_DOMAIN, NAMESPACE_USER],
+    [NAMESPACE_DOMAIN],
```
emord: @sheelio Why did you change this? I think it's going to make it much harder for us to debug UCRs if we can't go look at them

sheelio: If the project is in the feature flag, you would still have access right?

emord: I'm generally pro domain namespace only, but I think this can also be helpful to debug report builder reports, and being able to rebuild tables manually on any domain.

If we give the access to the domain, then the "edit report" button would show up for all users in that domain (likely bad for enikshay/icds right?)

PR #18497, opened 2017-11-10
allow an env to disable daily deploy email

dannyroberts:
by setting DAILY_DEPLOY_EMAIL to None


dannyroberts:
I added product/feature-flag, which isn't quite right, so that it shows up as something to announce internally. Not sure how much it matters, but seems like people might want to know this change (and when it got rolled out); specifically that we'll be only sending deploy announcements out for prod, enikshay, and icds https://github.com/dimagi/commcarehq-ansible/pull/1027/files.

esoergel (approved):

PR #18501, opened 2017-11-10
Add command to reassign enikshay beneficiaries

esoergel:
https://trello.com/c/d0nwlgwi/258-run-migration-for-misassigned-beneficiaries
[spec here](https://docs.google.com/document/d/1MYCFxwCry9hhZgrYMtoF0ZYBJ0TYNxnshfb3jcoacDo/edit#heading=h.b91kw8f57ekv)
@proetusvacuum @dannyroberts

### DIFF
--- a/custom/enikshay/management/commands/reassign_enikshay_beneficiaries.py
+++ b/custom/enikshay/management/commands/reassign_enikshay_beneficiaries.py
@@ -1,0 +1,128 @@
+from __future__ import absolute_import
+from __future__ import print_function
+from __future__ import unicode_literals
+
+import csv
+import datetime
+from django.core.management.base import BaseCommand
+
+from dimagi.utils.chunked import chunked
+from dimagi.utils.decorators.memoized import memoized
+
+from casexml.apps.case.mock import CaseBlock
+from corehq.apps.es import CaseSearchES
+from corehq.apps.hqcase.utils import submit_case_blocks
+from corehq.apps.locations.models import SQLLocation
+from corehq.form_processor.interfaces.dbaccessors import CaseAccessors
+from corehq.util.log import with_progress_bar
+
+from custom.enikshay.const import ENROLLED_IN_PRIVATE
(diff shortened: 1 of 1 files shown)


custom/enikshay/management/commands/reassign_enikshay_beneficiaries.py
```diff
@@ -0,0 +1,128 @@
...
+
+    def get_person_cases(self, person_ids):
+        return CaseAccessors(self.domain).iter_cases(person_ids)
+
+    def get_person_ids(self):
+        return (CaseSearchES()
+                .domain(self.domain)
+                .case_type(CASE_TYPE_PERSON)
+                .case_property_query(ENROLLED_IN_PRIVATE, 'true')
+                .get_ids())
```
proteusvacuum: I didn't know about this method. 👍

custom/enikshay/management/commands/reassign_enikshay_beneficiaries.py
```diff
@@ -0,0 +1,128 @@
...
+            owner_id=data.get('new owner_id', CaseBlock.undefined),
+            update=(
+                {'facility_assigned_to': data['new owner_id']}
+                if 'new owner_id' in data else None
+            ),
+        )
+
+    def save_case_blocks(self, case_blocks):
+        for chunk in chunked(case_blocks, 100):
+            submit_case_blocks(
```
proteusvacuum: you could have used `bulk_update_cases` to save you the step of building the case block, but this works too.

proteusvacuum (approved):

custom/enikshay/management/commands/reassign_enikshay_beneficiaries.py
esoergel: It looked like `bulk_update_cases` doesn't let you pass in `owner_id`.  Looking at it more closely now, you can apparently also pass in `owner_id` as an element in the `update` dict - I assumed it had to be a top-level attribute.  Oh well.

PR #18506, opened 2017-11-13
Max columns

kaapstorm:
Context: [FB 265374][1], [FB 265429][2]

Report Builder creates a table in PostgreSQL for its Report Preview. Each form question gets 2 or 4 columns each depending on data type, and multiple choice answers get 2 or 4 columns per answer. For really big forms, the column count can exceed PostgreSQL's max. 

This PR drops the column count to a manageable number. It isn't very smart about it though. It just takes the first 500. (This value is arbitrary and open for discussion.) Other options might be to drop multiple choice answers first, or columns for less-frequently-used aggregations, maybe. But as a first shot, this should allow users to proceed. This code is only called by [`_build_temp_data_source()`][3].

When reproing this bug, I noticed that Report Builder's initial indicator selection doesn't always result in a valid report config. We should deal with that too, but I'll do that in a separate PR.

@sheelio (Would it be rude to tag Noah for a review? :upside_down_face: ) cc @millerdev 

  [1]: https://manage.dimagi.com/default.asp?265374
  [2]: https://manage.dimagi.com/default.asp?265429
  [3]: https://github.com/dimagi/commcare-hq/blob/ef569e8e883131b294e2986dbe195b31b671a3ff/corehq/apps/userreports/views.py#L294

### DIFF
--- a/corehq/apps/userreports/reports/builder/forms.py
+++ b/corehq/apps/userreports/reports/builder/forms.py
@@ -74,6 +74,9 @@ STATIC_CASE_PROPS = [
     "user_id",
 ]
 
+# PostgreSQL limit = 1600. Sane limit = 500?
+MAX_COLUMNS = 500
+
 
 class FilterField(JsonField):
     """
@@ -432,7 +435,7 @@ class DataSourceBuilder(object):
                 return_list.append(indicator)
                 return_list_set.add(as_hashable)
 
-        return return_list
+        return return_list[:MAX_COLUMNS]
 
     @property
     @memoized
--- a/corehq/apps/userreports/reports/builder/forms.py
+++ b/corehq/apps/userreports/reports/builder/forms.py
@@ -398,27 +398,22 @@ class DataSourceBuilder(object):
             Each object has a "property" and "aggregation" key
         :param filters: A list of filter configuration objects
         """
-        indicators = []
+
+        def get_key(i):
+            return i['column_id'], i['type']
+
+        indicators = {}
         for column in columns:
             column_option = self.report_column_options[column['property']]
-            indicators.extend(column_option.get_indicators(column['aggregation'], is_multiselect_chart_report))
+            for indicator in column_option.get_indicators(column['aggregation'], is_multiselect_chart_report):
+                indicators.setdefault(get_key(indicator), indicator)
 
         for filter in filters:
             property = self.data_source_properties[filter['property']]
             indicator = property.to_report_filter_indicator(filter)
-            indicators.append(indicator)
+            indicators.setdefault(get_key(indicator), indicator)
 
-        # remove duplicates
-        # There can be duplicates because filters and columns could be based on the same property
-        indicators_without_dups = []
-        seen_indicator_ids = set()
-        for i in indicators:
-            if (i['column_id'], i['type']) not in seen_indicator_ids:
-                indicators_without_dups.append(i)
-                seen_indicator_ids.add((i['column_id'], i['type']))
-        indicators = indicators_without_dups
-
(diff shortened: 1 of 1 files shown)


sheelio:
I'm not too familiar with this code so would defer to Daniel.  

I noticed you pulled out some code around removing duplicates?  Did you change the code to account for that?

kaapstorm:
@sheelio the code that I changed around removing duplicates has the same effect. The previous code was adding everything to a list, and then looping over it a second time to find duplicates. My change uses the keys of a dictionary to avoid duplicates, so you only need to loop once instead of twice. The return value is the same.

corehq/apps/userreports/reports/builder/forms.py
```diff
@@ -395,44 +398,31 @@ def indicators(self, columns, filters, is_multiselect_chart_report=False):
...
-        for indicator in indicators:
-            as_hashable = str(indicator)
-            if as_hashable not in return_list_set:
-                return_list.append(indicator)
-                return_list_set.add(as_hashable)
+                for indicator in column_option.get_indicators(agg):
+                    indicators.setdefault(str(indicator), indicator)
 
-        return return_list
+        return list(indicators.values())[:MAX_COLUMNS]
```
millerdev: This may not always grab the same columns because dict is an unordered data type. Could probably use [`OrderedDict`](https://docs.python.org/2/library/collections.html#collections.OrderedDict).

kaapstorm: Good point @millerdev . I'm not sure they need to be in the same order, but I do agree it's best to reproduce the previous behaviour exactly. I'll switch to `OrderedDict`.

snopoke (commented):
What happens if the user selects one of the columns that was excluded?

kaapstorm:
HQ won't be able to build the report preview. It'll give an error that says "There was an error rendering your report preview. Please adjust your configuration to try again. If the issue persists, please report an issue."

That's very misleading if the user created a legitimate report config. I'll see how we can tell the difference between a bad report config, and a good one that refers to indicators that can't be previewed.

millerdev:
If the user has told (or tells) us which columns they want we should be able to make sure those columns are available, no? Is there a reason why we can't (or it's hard to) prioritize those?

kaapstorm:
The idea was to create a data source with all possible indicators, so that when the user adds columns we can update the report preview within a reasonable timeframe. If we have to create a new data source each time the user adds a column, then it becomes too slow to be useful.

kaapstorm:
I chatted to @sheelio about this. 

If a user makes a report using only the indicators that the temp data source does have, everything will work fine. If they add an indicator that the data source doesn't have a column for, currently (i.e. as this PR stands) the preview will break. 

Ideally we should add the column to the data source. It's worth testing this to see how slow it would be if we have 100 rows (the temp data source will never have more than 100 rows) and 501 columns.

If it's too slow to be useful, we should show the user a warning that we can't show a preview, but their report looks valid, and should work after they have saved it.

snopoke:
That sounds about right. How hard would it be to give them the option of
'refreshing' the preview data if this situation arises?

sheelio:
Agree with Simon - if rebuilding is too slow, we shouldn't too it automatically and instead provide a button / link to "Refresh Preview"

kaapstorm:
While this PR is not yet ideal, it would unblock the two users who are experiencing the original bug. Can I suggest we merge, and I'll work on refreshing the preview / rebuilding the data source in a separate PR?

sheelio:
@kaapstorm 
I think thats fine (unblocking these users for now).

PR #18514, opened 2017-11-14
Various 99DOTS fixes

proteusvacuum:
- Allow adding adherence cases to previously closed cases: [trello](https://trello.com/c/iwsaCvxh/14-patient-exists-but-no-occurrence-cases). Happens if the case has a treatment outcome, but 99DOTS hasn't sent all past adherences yet
- Allow sending adherence cases to 99DOTS even if the episode is closed: [trello](https://trello.com/c/DklzzvkA/18-sending-adherence-cases-for-cases-closed-immediately-after-fails). This happens if a treatment outcome is recorded immediately after an adherence case is sent, but hasn't yet been updated.
- Also adds a test for [this](https://trello.com/c/SFQCmFHH/5-treatment-outcomes-recorded-in-public-sector-not-reflected-in-99dots), which I thought was happening because of the multiple case blocks updating the same case, but was actually because the cases in question had failed to be registered properly in 99DOTS. 

@calellowitz

### DIFF
--- a/custom/enikshay/integrations/tests/test_case_properties_changed.py
+++ b/custom/enikshay/integrations/tests/test_case_properties_changed.py
@@ -1,0 +1,39 @@
+from __future__ import absolute_import
+
+import uuid
+
+from django.test import TestCase, override_settings
+
+from casexml.apps.case.mock import CaseBlock
+from corehq.apps.domain.shortcuts import create_domain
+from corehq.apps.hqcase.utils import submit_case_blocks
+from corehq.form_processor.interfaces.dbaccessors import CaseAccessors
+from custom.enikshay.integrations.utils import case_properties_changed
+
+
+@override_settings(TESTS_SHOULD_USE_SQL_BACKEND=True)
+class TestCasePropertiesChanged(TestCase):
+    def setUp(self):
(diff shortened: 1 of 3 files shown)


custom/enikshay/integrations/tests/test_case_properties_changed.py
```diff
@@ -0,0 +1,39 @@
...
+            case_type=case_type,
+            update={'property_1': 'updated'},
+        )
+        caseblock2 = CaseBlock(
+            self.case_id,
+            case_type=case_type,
+            update={'property_2': 'updated'},
+        )
+        blocks = [caseblock1.as_string(), caseblock2.as_string()]
+        submit_case_blocks(blocks, self.domain)[1][0]
```
calellowitz: what are those indexes doing on the end? looks like you aren't doing anything with the return value.

calellowitz (approved):
looks good. one non blocking question on the test

custom/enikshay/integrations/tests/test_case_properties_changed.py
proteusvacuum: oops, good catch. thanks

PR #18522, opened 2017-11-14
always preserve form IDs in linked apps

snopoke:
This won't have any impact on existing linked apps (which have the same form IDs as their master app). Any new linked apps, or new forms added to existing linked apps will have different IDs from the master app.

Can review by commit :fish:


emord (approved):
LGTM will let @calellowitz review since he'll have more context

calellowitz (approved):
looks good. thanks for doing this

PR #18525, opened 2017-11-14
improve how we create linked apps

snopoke:
Issue with the ticket is that `master_domain` wasn't being set when creating linked apps.
https://manage.dimagi.com/default.asp?265345

### DIFF
--- a/corehq/apps/app_manager/views/apps.py
+++ b/corehq/apps/app_manager/views/apps.py
@@ -353,6 +353,13 @@ def copy_app(request, domain):
                     set_toggle(slug, link_domain, True, namespace=toggles.NAMESPACE_DOMAIN)
             linked = data.get('linked')
             if linked:
+                for module in app.modules:
+                    if isinstance(module, ReportModule):
+                        messages.error(request, _('This linked application uses mobile UCRs which '
+                                                  'are currently not supported. For this application to '
+                                                  'function correctly, you will need to remove those modules.'))
+                        return HttpResponseRedirect(reverse_util('app_settings', params={}, args=[domain, app_id]))
+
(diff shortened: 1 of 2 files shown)


corehq/apps/app_manager/views/apps.py
```diff
@@ -346,28 +346,43 @@ def copy_app(request, domain):
...
+                    domain=link_domain,
+                    master=app_id,
+                    master_domain=master_domain,
+                )
+                linked_app.save()
+                return pull_master_app(request, link_domain, linked_app.get_id)
+            else:
+                extra_properties = {'name': data['name']}
+                app_copy = import_app_util(app_id_or_source, link_domain, extra_properties)
+                if linked:
```
calellowitz: i could be misreading this code but it looks like this clause is impossible since this is in the else clause (line 375) of the `if linked` in line 355, whcih would also mean we never verify a lack of UCRs (unless that part was on purpose)

calellowitz (changes requested):
this also needs to modify the `link_apps` management command (corehq/apps/app_manager/management/commands/link_apps.py), which is what we use to link existing apps

corehq/apps/app_manager/views/apps.py
snopoke: 7ee88be

emord (approved):

calellowitz (approved):

calellowitz (approved):
Changes to the management command mentioned above aren't blocking but probably should be in a followup.
It also does occur to me now that we do support linked apps with static UCRs (or at least could), but not sure any project that would want to use that uses this view to link apps, only the command.

PR #18533, opened 2017-11-14
Case search details

proteusvacuum:
This adds new case search specific "short" details with id `m{x}_search_short` for modules that have case search enabled, serving 2 purposes:
- If "cache and index" is enabled on the "base" detail (`m{x}_case_short`), we remove this from the case search specific detail, as the mobile doesn't know how to break the cache when search results are returned. 
- The mobile can now trust that the actions included on the case search short detail are "real", i.e. the case search detail never includes the "Search All Cases" action, but inherits other actions from its "case_short" detail. 

https://trello.com/c/qG0wBM7U/207-cache-and-index-updates-on-mobile

Curious whether it is worth copying the `case_long` details too, or better to leave that as is, since I don't think that is currently causing any issues. @ctsims what do you think?

### DIFF
--- a/corehq/apps/app_manager/models.py
+++ b/corehq/apps/app_manager/models.py
@@ -124,7 +124,9 @@ from corehq.apps.app_manager.util import (
     get_correct_app_class,
     get_and_assert_practice_user_in_domain,
     LatestAppInfo,
-    update_report_module_ids)
+    update_report_module_ids,
+    module_offers_search,
+)
 from corehq.apps.app_manager.xform import XForm, parse_xml as _parse_xml, \
     validate_xform
 from corehq.apps.app_manager.templatetags.xforms_extras import trans
@@ -2547,6 +2549,10 @@ class ModuleDetailsMixin():
         except Exception:
             return []
 
+    @property
+    def search_detail(self):
+        return deepcopy(self.case_details.short)
+
     def rename_lang(self, old_lang, new_lang):
         super(Module, self).rename_lang(old_lang, new_lang)
         for case_list in (self.case_list, self.referral_list):
@@ -2560,12 +2566,15 @@ class ModuleDetailsMixin():
         return json.dumps(source) if dump_json else source
 
     def get_details(self):
-        return (
+        details = [
             ('case_short', self.case_details.short, True),
             ('case_long', self.case_details.long, True),
             ('ref_short', self.ref_details.short, False),
             ('ref_long', self.ref_details.long, False),
-        )
+        ]
+        if module_offers_search(self):
+            details.append(('search_short', self.search_detail, True))
+        return tuple(details)
 
     def validate_details_for_build(self):
         errors = []
--- a/corehq/apps/app_manager/suite_xml/sections/details.py
+++ b/corehq/apps/app_manager/suite_xml/sections/details.py
@@ -44,6 +44,10 @@ from corehq import toggles
 from dimagi.utils.decorators.memoized import memoized
 
 
+DetailColumnInfo = namedtuple('DetailColumnInfo',
+                              'column sort_element order')
+
+
 class DetailContributor(SectionContributor):
     section_name = 'details'
 
@@ -184,6 +188,9 @@ class DetailContributor(SectionContributor):
                 #   column_info.column: an instance of app_manager.models.DetailColumn
                 #   column_info.sort_element: an instance of app_manager.models.SortElement
                 #   column_info.order: an integer
+                if "search" in id and column_info.sort_element and column_info.sort_element.type == 'index':
+                    # search details can't have index elements, so remove the sort element
+                    column_info = DetailColumnInfo(column_info.column, None, None)
                 fields = get_column_generator(
                     self.app, module, detail, parent_tab_nodeset=nodeset,
                     detail_type=detail_type, *column_info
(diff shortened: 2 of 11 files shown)


corehq/apps/app_manager/suite_xml/sections/details.py
```diff
@@ -167,20 +188,25 @@ def build_detail(self, module, detail_type, detail, detail_column_infos, tabs=No
                 #   column_info.column: an instance of app_manager.models.DetailColumn
                 #   column_info.sort_element: an instance of app_manager.models.SortElement
                 #   column_info.order: an integer
+                if "search" in id and column_info.sort_element and column_info.sort_element.type == 'index':
```
orangejenny: `if "search" in id` seems fragile - is it possible to keep track of search detail ids somewhere rather than pattern matching on the id?

corehq/apps/app_manager/suite_xml/sections/details.py
```diff
@@ -95,6 +99,23 @@ def include_sort(detail_type, detail):
                                     )
                                     if d:
                                         r.append(d)
+                                    if (module_offers_search(module)
+                                       and detail_type.endswith('short')
+                                       and not module.put_in_root):
```
orangejenny: why the `not module.put_in_root` constraint?

corehq/apps/app_manager/suite_xml/sections/details.py
orangejenny: I do see that we're already doing `detail_type.endswith('short')` elsewhere, which is the same kind of pattern, but still, might be better not to replicate that.

corehq/apps/app_manager/suite_xml/sections/details.py
```diff
@@ -419,8 +445,6 @@ def get_detail_column_infos(detail_type, detail, include_sort):
     This is not intented to be a widely used format
     just a packaging of column info into a form most convenient for rendering
     """
-    DetailColumnInfo = namedtuple('DetailColumnInfo',
```
orangejenny: Should move this comment along with the `DetailColumnInfo` definition.

corehq/apps/app_manager/suite_xml/sections/remote_requests.py
```diff
@@ -102,14 +102,13 @@ def _build_remote_request_queries(self):
...
-        details_helper = DetailsHelper(self.app)
         return [SessionDatum(
             id='case_id',
             nodeset=(CaseTypeXpath(self.module.case_type)
                      .case(instance_name=RESULTS_INSTANCE)),
             value='./@case_id',
-            detail_select=details_helper.get_detail_id_safe(self.module, 'case_short'),
-            detail_confirm=details_helper.get_detail_id_safe(self.module, 'case_long'),
+            detail_select=id_strings.detail(self.module, 'search_short'),
+            detail_confirm=id_strings.detail(self.module, 'case_long'),
```
orangejenny: This looks suspicious - is it definitely okay that the long detail id in particular loses the checks in `get_detail_id_safe`?

ctsims:
@proteusvacuum No changes needed for long details, they do not support cache and index nor the actions framework at this time.

corehq/apps/app_manager/suite_xml/sections/details.py
proteusvacuum: good point. I've updated this PR so that the search detail is a "first class" detail, but I've still followed the `endswith` pattern, because that's the way all of these are checked in this code.

corehq/apps/app_manager/suite_xml/sections/remote_requests.py
proteusvacuum: updated in the newest commit: https://github.com/dimagi/commcare-hq/pull/18533/commits/347fc2c67a199fb5ceb373dcf7abb427493019a9

corehq/apps/app_manager/suite_xml/sections/details.py
proteusvacuum: moved this back to it's original place.

corehq/apps/app_manager/tests/data/suite/search_command_detail.xml
```diff
@@ -95,12 +95,12 @@
...
       <header>
         <text>
-          <locale id="m0.case_long.case_whatever_2.header"/>
+          <locale id="m0.search_short.case_whatever_2.header"/>
         </text>
       </header>
       <template>
         <text>
-          <xpath function="instance('ledgerdb')/ledgers/name/name"/>
+          <xpath function="instance('reports')/report[1]/name"/>
```
orangejenny: I'm idly curious why this changed in this PR...?

corehq/apps/app_manager/models.py
```diff
@@ -2547,6 +2549,10 @@ def detail_sort_elements(self):
         except Exception:
             return []
 
+    @property
+    def search_detail(self):
```
orangejenny: What would you think of asserting `module_offers_search` here to be extra defensive?

orangejenny (approved):

corehq/apps/app_manager/tests/data/suite/search_command_detail.xml
proteusvacuum: the diff for this commit is weird. If you check the diff of the whole range you'll see that I added `case_long` and `search_short` sections to the detail.

corehq/apps/app_manager/models.py
proteusvacuum: I don't think anything bad would happen if this was called even if there was no search action. This property also doesn't have any side effects, so I think it should be left up to the caller to determine whether they want to use the value or not.

PR #18535, opened 2017-11-15
only convert @date to datetime.date

nickpell:
https://manage.dimagi.com/default.asp?265466

@snopoke Are there any assumptions I can make about the format of ```@date``` which might simplify this code?

### DIFF
--- a/corehq/form_processor/parsers/ledgers/form.py
+++ b/corehq/form_processor/parsers/ledgers/form.py
@@ -145,7 +145,8 @@ def get_all_stock_report_helpers_from_form(xform):
 
     for elem in _extract_ledger_nodes_from_xml(form_xml):
         report_type, ledger_json = convert_xml_to_json(elem, last_xmlns=COMMTRACK_REPORT_XMLNS)
-        ledger_json['@date'] = adjust_text_to_datetime(ledger_json['@date'])
+        if '@date' in ledger_json:
+            ledger_json['@date'] = adjust_text_to_datetime(ledger_json['@date'])
         yield _ledger_json_to_stock_report_helper(xform, report_type, ledger_json)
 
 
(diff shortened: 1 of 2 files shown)


snopoke (commented):
I think we can assume it's always going to be an ISO date so you should be able to use `iso8601.parse_date` but then you'll need to handle the timezone (probably in a similar way to this: https://github.com/dimagi/commcare-hq/blob/3e88dbe4554316981e7f9a43fde34aa8870702c9/corehq/form_processor/utils/xform.py#L141-L152. Perhaps you can extract a function out of that which returns a datetime.

Another thing is that if `@date` contains time information we shouldn't discard it (by calling `.date()`) otherwise it will later get replaced with the current time (https://github.com/dimagi/commcare-hq/blob/3e88dbe4554316981e7f9a43fde34aa8870702c9/corehq/form_processor/parsers/ledgers/form.py#L292-L292).

corehq/form_processor/parsers/ledgers/form.py
```diff
@@ -145,11 +145,8 @@ def _extract_ledger_nodes_from_xml(node):
 
     for elem in _extract_ledger_nodes_from_xml(form_xml):
         report_type, ledger_json = convert_xml_to_json(elem, last_xmlns=COMMTRACK_REPORT_XMLNS)
-
-        # apply the same datetime & string conversions
-        # that would be applied to XFormInstance.form
-        adjust_datetimes(ledger_json)
-        ledger_json = XFormInstance({'form': ledger_json}).form
+        if ledger_json.get('@date'):
+            ledger_json['@date'] = adjust_text_to_datetime(ledger_json['@date'])
```
snopoke: what happens if this raises a `iso8601.ParseError`?

snopoke (commented):
Looks good, just one question

corehq/form_processor/parsers/ledgers/form.py
nickpell: It would result in a 500.  I suppose you could have get there if a user submitted a date like April 31.

snopoke:
Changed label since this is just a bug and basically 'invisible' to users.

PR #18537, opened 2017-11-15
fixing latest saved update for build profiles

calellowitz:
@orangejenny

Due to our conversation it occurred to me that the issue is likely just generating each of the files (when there is no build) specifically for that profile, because that is the update path taken when updating to latest saved. It should have nothing to do with the build process, so I kept that the same and updated when each file downloaded. This appears to work but i noted my testing steps below in case it did not cover the situation where you saw it break.

Testing steps (app version noted in parentheses):
1. installed one profile (v1)
2. made a change
3. updated (v2)
4. confirmed update was the same profile
5. reinstalled app in different profile (v2)
6. new change
7. update to latest saved (v3)
8. confirmed that update maintained the profile

### DIFF
--- a/corehq/apps/app_manager/views/download.py
+++ b/corehq/apps/app_manager/views/download.py
@@ -42,8 +42,9 @@ def download_odk_profile(request, domain, app_id):
         make_async_build.delay(request.app, username)
     else:
         request._always_allow_browser_caching = True
+    profile = request.GET.get('profile')
     return HttpResponse(
-        request.app.create_profile(is_odk=True),
+        request.app.create_profile(is_odk=True, build_profile_id=profile),
         content_type="commcare/profile"
     )
 
@@ -55,8 +56,9 @@ def download_odk_media_profile(request, domain, app_id):
         make_async_build.delay(request.app, username)
     else:
         request._always_allow_browser_caching = True
+    profile = request.GET.get('profile')
     return HttpResponse(
-        request.app.create_profile(is_odk=True, with_media=True),
+        request.app.create_profile(is_odk=True, with_media=True, build_profile_id=profile),
         content_type="commcare/profile"
     )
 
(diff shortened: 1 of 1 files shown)


orangejenny:
fyi @amstone326

calellowitz:
@orangejenny is this good to go?

orangejenny:
I wanted to test it out on staging, which hasn't been stable enough to do that.

orangejenny:
The code looks plausible, but when testing, your step 5 is breaking for me. Maybe a difference in exactly what the test app & its profiles are?

Specifically, I've been testing [this app](https://staging.commcarehq.org/a/jennytraining/apps/view/7e2dcf8c7a648866df46164b4de913b2/), which has two build profiles - one that's just English and one that's just French - as follows:
1. Make change in app so there are unbuilt changes.
2. Install an old version, using the english-only profile.
3. Update the app via latest saved state, which creates a new version of the app.
4. Uninstall the app, then install the most recent, automatically created, build, using the french-only profile. That installation fails because `Form does not contain any translations for any of the build languages` (https://sentry.io/dimagi/commcarehq/issues/409659756/). It's the same issue I was hitting with https://github.com/dimagi/commcare-hq/compare/jls/update-app-profile

Thoughts?

calellowitz:
do you have french translations for that whole app? I think you may be running into a bug/feature, where you have don't have a translation for all strings in any of the languages of the language profile, and it does not fall back to default because the default language isn't included in the profile.  More specifically, independent of update to latest saved can you install that app in that profile (for example if you made a new build the normal way).

It is up for debate whether that is reasonable behavior app profiles wont include strings defined in languages not included in the profile.

orangejenny:
You're right: even for builds created normally, installing the french profile fails with the same error. Glorious.

orangejenny (approved):

PR #18538, opened 2017-11-15
Revert "Revert "a few shard utils""

czue:
Reverts dimagi/commcare-hq#18530 so I can investigate what's going wrong.


czue:
ok, my new plan if this makes sense is to just get this PR in without the migrations, then figure out the migrations and add the dependent tests back after the fact. this will simplify the ability to use some of the tools now for what is still a bit of a forward looking, one-off set of functionality. that way i don't have to worry about code conflicts and other issues with the rest of the changes.

let me know if any concerns with that approach. @snopoke @emord

snopoke (approved):

PR #18540, opened 2017-11-15
keep attachments around after saving

snopoke:
https://sentry.io/dimagi/commcarehq/issues/264603632/

Not 100% sure what happened in this event but it would be better if we didn't have to fetch the attachments back out of Riak immediately after we've just written them in.

### DIFF
--- a/corehq/form_processor/tests/test_form_dbaccessor.py
+++ b/corehq/form_processor/tests/test_form_dbaccessor.py
@@ -91,6 +91,10 @@ class FormAccessorTestsSQL(TestCase):
 
     def test_get_with_attachments(self):
         form = create_form_for_test(DOMAIN)
+        form = FormAccessorSQL.get_form(form.form_id)  # refetch to clear cached attachments
+        with self.assertNumQueries(1, using=db_for_read_write(XFormAttachmentSQL)):
+            form.get_attachment_meta('form.xml')
+
         with self.assertNumQueries(1, using=db_for_read_write(XFormAttachmentSQL)):
             form.get_attachment_meta('form.xml')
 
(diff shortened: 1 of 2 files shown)


emord:
This looks good to me. The test that failed, looks like it was expecting this query, there may be more though. @millerdev any reason we should fetch a form from riak in this case instead of keeping it around?

```
FAIL: corehq.form_processor.tests.test_form_dbaccessor:FormAccessorTestsSQL.test_get_with_attachments
  File "/mnt/commcare-hq-ro/corehq/form_processor/tests/test_form_dbaccessor.py", line 95, in test_get_with_attachments
    form.get_attachment_meta('form.xml')
  File "/vendor/lib/python2.7/site-packages/django/test/testcases.py", line 88, in __exit__
    query['sql'] for query in self.captured_queries
AssertionError: 0 queries executed, 1 expected
Captured queries were:
```

millerdev:
@emord don't think I have enough context to answer that question right now

corehq/form_processor/tests/test_form_dbaccessor.py
```diff
@@ -91,6 +91,10 @@ def test_get_forms_by_last_modified(self):
 
     def test_get_with_attachments(self):
         form = create_form_for_test(DOMAIN)
+        form = FormAccessorSQL.get_form(form.form_id)  # refetch to clear cached attachments
+        with self.assertNumQueries(1, using=db_for_read_write(XFormAttachmentSQL)):
```
emord: Why does this happen twice? I was originally thinking we should just change it to 0 queries expected

snopoke: in the original version it should have been changed to 0 since it would use the 'cached' attachments but now that I've re-fetched the form from the DB it needs to do another query to get the attachments.

The reason I did that instead of changing it to `0` is that it the test is making sure that it doesn't query the DB if you call `get_form_with_attachments`. So asserting that it will query the DB if you just call `get_form` is part of the test.

emord (approved):

PR #18081, opened 2017-10-04
Send location to kafka after it was saved.

emord:
@esoergel as discussed earlier

@millerdev

PR #18541, opened 2017-11-15
publish changes from inside atomic block

snopoke:
https://sentry.io/dimagi/commcarehq/issues/234393735/


emord:
I don't understand why this would help and I think its the wrong thing to do. I had to fix the opposite bug here: https://github.com/dimagi/commcare-hq/pull/18081

If we send it before the transaction has completed then any listener will try to read it from the database and there will be a race to see if the transaction can complete before the form is attempted to be read

snopoke:
Good point, thanks. I'll have to figure out the error handling here in another way.

PR #18544, opened 2017-11-15
Location-safe standard reports

esoergel:
This PR makes the following reports location-safe:

- Case Activity
- Daily Form Stats
- Form Completion Time
- Form Completion vs Submission Trends
- Submissions by Form
- Worker Activity Times

It also refactors separate, location-safe EMWF back in to the primary one, consolidates the utility methods for getting the selected users, and makes the remaining one (`user_es_query`) location-safe.

:fish: :heavy_multiplication_x: :tropical_fish: should make some sense.


corehq/apps/reports/filters/users.py
```diff
@@ -369,6 +363,14 @@ def user_es_query(cls, domain, mobile_user_and_group_slugs, request_user=None):
             else:
                 return q.filter(id_filter)
 
+    @staticmethod
+    def _verify_users_are_accessible(domain, request_user, user_ids):
+        # This function would be very slow if called with many user ids
+        for user_id in user_ids:
+            other_user = CommCareUser.get(user_id)
+            if not user_can_access_other_user(domain, request_user, other_user):
+                raise PermissionDenied("One or more users are not accessible")
```
esoergel: `user_ids` is the list of users explicitly selected by the user, so I expect it'll generally be pretty small.

corehq/apps/reports/filters/users.py
```diff
@@ -328,7 +330,24 @@ def user_es_query(cls, domain, mobile_user_and_group_slugs):
...
+            location_ids = (SQLLocation.active_objects
+                            .get_locations(location_ids)
+                            .accessible_to_user(domain, request_user)
+                            .location_ids())
+            if HQUserType.REGISTERED in user_types:
+                all_accessible_loc_ids = (SQLLocation.active_objects
+                                          .accessible_to_user(domain, request_user)
+                                          .location_ids())
+                q = q.location(all_accessible_loc_ids)
+                return q.OR(
```
gcapalbo: what's the meaning of this OR - does this mean it ORs the args passed to it and ANDs the result to the main query?

corehq/apps/reports/filters/users.py
```diff
@@ -331,22 +333,14 @@ def user_es_query(cls, domain, mobile_user_and_group_slugs, request_user=None):
...
-                                          .location_ids())
-                q = q.location(all_accessible_loc_ids)
-                return q.OR(
-                    filters.term("_id", user_ids),
-                    user_es.location(location_ids),
-                )
-            else:
-                return q.location(location_ids),
+            cls._verify_users_are_accessible(domain, request_user, user_ids)
+            return q.OR(
```
gcapalbo: how come this one is an OR, doesn't that mean that it will bypass the location restriction if the user's id is in user_ids?

gcapalbo: oh nevermind - i see that the user always has access to those users because of `_verify_users_are_accessible`

gcapalbo (approved):

corehq/apps/reports/filters/users.py
esoergel: That's correct

corehq/apps/reports/filters/api.py
```diff
@@ -123,12 +124,18 @@ def data_sources(self):
...
+                (self.get_static_options_size, self.get_static_options),
+                (self.get_groups_size, self.get_groups),
+                (self.get_locations_size, self.get_locations),
+                (self.get_users_size, self.get_users),
+            ]
+        else:
+            return [
+                (self.get_locations_size, self.get_locations),
+                (self.get_users_size, self.get_users),
+            ]
```
mkangia: nit: i preferred how it was done earlier with appending whats needed according to access. One need not compare between two lists to figure out which data sources are different.

```
        sources = []
        if self.custom_locations_search():
            return [(self.get_locations_size, self.get_locations)]
        elif self.request.can_access_all_locations:
            sources.append((self.get_static_options_size, self.get_static_options))
            sources.append((self.get_groups_size, self.get_groups))
            
        sources.append((self.get_locations_size, self.get_locations))
       # appending this in the end to avoid long list of users delaying
       # locations, groups etc in the list on pagination
       sources.append((self.get_users_size, self.get_users))
       return sources
```

corehq/apps/reports/filters/api.py
```diff
@@ -298,21 +266,28 @@ def user_es_query(self, query):
...
+                (self.get_groups_size, self.get_groups),
+                (self.get_sharing_groups_size, self.get_sharing_groups),
+                (self.get_locations_size, self.get_locations),
+                (self.get_users_size, self.get_users),
+            ]
+        else:
+            return [
+                (self.get_locations_size, self.get_locations),
+                (self.get_users_size, self.get_users),
+            ]
```
mkangia: Might want to still retain `extra_data_sources` to avoid repeating this.

corehq/apps/reports/filters/users.py
```diff
@@ -313,8 +311,8 @@ def _selected_location_entries(self, mobile_user_and_group_slugs):
         location_ids = self.selected_location_ids(mobile_user_and_group_slugs)
         if not location_ids:
             return []
-        return map(self.utils.location_tuple,
-                   SQLLocation.objects.filter(location_id__in=location_ids))
+        return list(map(self.utils.location_tuple,
+                        SQLLocation.objects.filter(location_id__in=location_ids)))
```
mkangia: why was this change needed?

corehq/apps/reports/filters/users.py
```diff
@@ -326,7 +324,9 @@ def filter_context(self):
         return context
 
     @classmethod
-    def user_es_query(cls, domain, mobile_user_and_group_slugs):
+    def user_es_query(cls, domain, mobile_user_and_group_slugs, request_user):
+        # The queryset returned by this method is location-safe
+        can_access_all_locations = request_user.has_permission(domain, 'access_all_locations')
```
mkangia: Dont see a use of this variable `can_access_all_locations` except at one place. Might want to just replace that with the expression.

corehq/apps/reports/filters/users.py
```diff
@@ -364,60 +373,13 @@ def user_es_query(cls, domain, mobile_user_and_group_slugs):
             else:
                 return q.filter(id_filter)
 
-    @classmethod
-    def pull_users_and_groups(cls, domain, mobile_user_and_group_slugs,
```
mkangia: Might want to check this again.
1. I vaguely remember that we used to pull groups for selected users as well and then the users under those groups to then fetch cases that are owned by the group/users under it. Is that still happening somewhere? this was probably because for case sharing between groups? Sorry i don't remember exactly where this was.
Related to this the check now is just for selected users using `_verify_users_are_accessible ` and not for user_ids generated/added by other filters. We should confirm if user_ids are not getting added by other filter like group. Because if thats the case one would be able to view cases/forms submitted by inaccessible user under a group.

mkangia (approved):
Overall looks good. Thanks for doing this Mr. Hunt.

Two name changes if you can include in this PR, if it looks worth it. 
the name `ExpandedMobileWorkerFilter`. Probably a better name would be something like `FilterDataSourcesOptions` or something better you can think of.
Second was the slug `emw` which can be changed to a better name.

corehq/apps/reports/filters/users.py
esoergel: in python3, `map` returns an iterable, so this makes sure that it's a list.

corehq/apps/reports/filters/api.py
esoergel: it's a bit of repetition, but I like this way more.  I think of it as a declarative spec - "restricted users get these data sources, regular users get these ones".  It could certainly get messy if there were a bunch of different conditionals being applied, but as-is it kinda just lays it all out in one place.

corehq/apps/reports/filters/api.py
esoergel: Just commented on the above one, but I prefer to lay it out all in one place, so you can easily tell which sources are being used.  Otherwise you'd have to jump to the super class to and then figure out what's being added or removed from that to determine which sources this uses.  This is kinda like "The case list filter uses these ones".

corehq/apps/reports/filters/users.py
esoergel: Good point, will update

corehq/apps/reports/filters/users.py
esoergel: Restricted users can't filter by group, so I don't think that should be an issue.  They only see users and locations.  Selected groups are taken care of here: https://github.com/dimagi/commcare-hq/pull/18544/files#diff-07031c94878b608626b437e634d152ddR365

This method returns a bunch of different things, but only one of them was in-use, so I swapped those usages out with `user_es_query` for simplicity.

corehq/apps/reports/filters/api.py
esoergel: Although I just realized that this will need to account for the special locations query too...

esoergel: @mkangia check out my latest two commits - I made it so the ES queries don't fail when special characters are provided, which means we can remove the restriction for special locations queries.  We _do_ want users to be able to search for users using quotes too, I think.

esoergel:
QA passed, should be g2m

gcapalbo (approved):

esoergel:
I think github is broken, I clicked "merge", but it said there was a problem, now it shows that the branch has conflicts, however, the merge appears to have gone through: https://github.com/dimagi/commcare-hq/commit/95be9777ec9c9d407b66368234d41f28bdffcc93
I'll give it some time then close this out.

PR #18542, opened 2017-11-15
Add Ivory Coast MTN Gateway

gcapalbo:
Adds connectivity to an Ivory Coast gateway for a custom project. The process for adding a gateway backend to HQ is documented [here](http://commcare-hq.readthedocs.io/messaging/sms_backends.html).

This connectivity is outbound only, so no inbound SMS views needed to be added.

@esoergel

PR #18546, opened 2017-11-15
#18542 Feedback

gcapalbo:
Incorporates feedback from https://github.com/dimagi/commcare-hq/pull/18542
@esoergel


esoergel (approved):
Awesome, thanks

PR #18561, opened 2017-11-16
python-modernize . -w -f libfuturize.fixes.fix_print_with_import

nickpell:
@dimagi/py3


corehq/apps/cleanup/management/commands/copy_form_attachments.py
```diff
@@ -32,7 +33,7 @@ def handle(self, *args, **options):
                 seen_forms.add(form_id)
                 seen_forms.add(deprecated_form.orig_id)
 
-                print('checking form, deprecated_form_id, ', form_id, 'orig_form_id, ', deprecated_form.orig_id)
+                print(('checking form, deprecated_form_id, ', form_id, 'orig_form_id, ', deprecated_form.orig_id))
```
millerdev: print functions in this file were python3 ready, no conversion needed.

corehq/blobs/management/commands/check_blob_logs.py
```diff
@@ -64,7 +65,7 @@ def handle(self, files, migrate=False, **options):
                     try:
                         rec = json.loads(line)
                     except ValueError:
-                        print("Ignore {}", line)
+                        print(("Ignore {}", line))
```
millerdev: was python3 ready before

corehq/apps/cleanup/management/commands/copy_form_attachments.py
nickpell: I think the fixer added parens because in python2 ```print(1, 2)``` prints ```(1, 2)``` so it tries to maintain that behavior in python3.

Since the original author was ok with that behavior, I didn't feel a need to change the existing behavior of these print statements.

millerdev (approved):

PR #18563, opened 2017-11-16
python-modernize . -w -f lib2to3.fixes.fix_types

nickpell:
@dimagi/py3


corehq/ex-submodules/casexml/apps/case/templatetags/case_tags.py
```diff
@@ -301,7 +301,7 @@ def get_inverse(val):
         return datetime.datetime.max - val
     elif isinstance(val, numbers.Number):
         return 10 ** 20
-    elif isinstance(val, (types.NoneType, bool)):
+    elif isinstance(val, (type(None), bool)):
```
emord: Could this be changed to `elif val is None or isinstance(val, bool)`

nickpell: clever

emord (approved):

PR #18567, opened 2017-11-16
don't fail to export UCR when there are / in the title

nickpell:
https://manage.dimagi.com/default.asp?265680

Product Note: Currently, UCRs fail to export if there is a "/" in the title.  This fixes the bug by replacing "/" with "?".


orangejenny (approved):
Looks fine. Is there a reason these are especially susceptible to `/`, or should we do something like replace all non-word characters?

nickpell:
In this case it's only the forward slash character used in file paths, because we attempt to save a file containing the report title in the filename.

PR #18568, opened 2017-11-16
make scripts/inventory/* respect port

dannyroberts:
so that you can run `scripts/inventory/ssh pna control`

merge after https://github.com/dimagi/commcare-hq-deploy/pull/420


scripts/inventory/getinventory.py
```diff
@@ -22,23 +22,22 @@ def print_help(self, stream=sys.stderr):
...
 
     """
-    from ansible.inventory import Inventory
-    from ansible.parsing.dataloader import DataLoader
-    from ansible.vars import VariableManager
-
-    inventory = Inventory(loader=DataLoader(), variable_manager=VariableManager(), host_list=filename)
-    return inventory.get_group_dict()
+    sys.path.append(os.path.join(ROOT, 'deployment', 'commcare-hq-deploy'))
+    from fab.utils import read_inventory_file
```
kaapstorm: Yeah, I agree this is better than copy-pasting. But I'd feel a little less awkward about this if we maybe caught ImportError and gave the poor dev who accidentally breaks this a hint about what they did wrong, maybe something like ...
```
try:
    from fab.utils import read_inventory_file
except ImportError:
    print('Ooops. It looks like the commcare-hq-deploy repo is no longer '
          'in "{}". Please update this script with its new location, or '
          'resolve the dependency.'.format(os.path.join(ROOT, 'deployment', 
          'commcare-hq-deploy')))
    raise
```

kaapstorm (approved):
Thanks Danny

dannyroberts:
conflict resolution/merge commit includes blowing over the `getinventory.py` part of https://github.com/dimagi/commcare-hq/pull/18689, since this PR includes the change to call the fab file directly. Confident about the change, just wanted to note that.

PR #18570, opened 2017-11-16
 python-modernize . -w -f lib2to3.fixes.fix_numliterals

nickpell:
@dimagi/py3

### DIFF
--- a/corehq/apps/hqwebapp/uglify.py
+++ b/corehq/apps/hqwebapp/uglify.py
@@ -84,7 +84,7 @@ class JsUglifySourcemapCompressor(JsCompressor):
                 settings.STATIC_ROOT, self.output_dir, self.output_prefix
             )
             if not os.path.exists(js_compress_dir):
-                os.makedirs(js_compress_dir, 0775)
+                os.makedirs(js_compress_dir, 0o775)
             filepath = self.get_filepath(concatenated_content, basename=None)
 
             # UglifySourcemapFilter writes the file directly, as it needs to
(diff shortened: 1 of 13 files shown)


corehq/ex-submodules/fluff/tests/tests.py
```diff
@@ -319,16 +319,16 @@ def test_indicator_diff(self):
         for cls in [MockIndicators, MockIndicatorsWithGetters]:
             current = cls(domain="mock",
                                      owner_id="123",
-                                     value_week=dict(date=[[date(2012, 02, 23), 1]],
+                                     value_week=dict(date=[[date(2012, 2, 23), 1]],
```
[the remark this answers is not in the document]
nickpell: ignoring

corehq/apps/hqwebapp/uglify.py
```diff
@@ -84,7 +84,7 @@ def output(self, mode='file', forced=False):
                 settings.STATIC_ROOT, self.output_dir, self.output_prefix
             )
             if not os.path.exists(js_compress_dir):
-                os.makedirs(js_compress_dir, 0775)
+                os.makedirs(js_compress_dir, 775)
```
emord: makedirs is supposed to be an octal https://docs.python.org/2/library/os.html https://docs.python.org/3/library/os.html

emord (changes requested):

corehq/apps/hqwebapp/uglify.py
nickpell: 👍 updated

emord (approved):

PR #18575, opened 2017-11-17
Fetch users asynchronously in case exploder

proteusvacuum:
This uses select2 to fetch users asynchronously when searching for them as the page takes a long time to load otherwise. 

@mkangia

### DIFF
--- a/corehq/apps/hqcase/templates/hqcase/explode_cases.html
+++ b/corehq/apps/hqcase/templates/hqcase/explode_cases.html
@@ -7,7 +7,30 @@
              factor: ko.observable(''),
              user_id: ko.observable(''),
          });
-         $('#explode-user_id').select2();
+         $('#explode-user_id').select2({
+             ajax: {
+                 url: "{% url 'users_select2_options' domain %}",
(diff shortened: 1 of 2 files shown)


corehq/apps/hqcase/templates/hqcase/explode_cases.html
```diff
@@ -7,7 +7,30 @@
              factor: ko.observable(''),
              user_id: ko.observable(''),
          });
-         $('#explode-user_id').select2();
+         $('#explode-user_id').select2({
+             ajax: {
+                 url: "{% url 'users_select2_options' domain %}",
+                 type: 'POST',
+                 datatype: 'json',
+                 quietMills: 250,
```
mkangia: nit: is 250 ms a very short duration? Should this be made like a second?

mkangia (approved):

corehq/apps/hqcase/templates/hqcase/explode_cases.html
proteusvacuum: that's what we use on other async select2 inputs

PR #18576, opened 2017-11-17
Remove deprecation warning

proteusvacuum:
When accessing any user filter (e.g. in the case list report, user sync history, etc.), you get a `DeprecationWarning: namedtuple fields should be accessed as attributes`, which is raised here: https://github.com/dimagi/commcare-hq/blob/master/corehq/apps/reports/util.py#L173-L176

The `SimplifiedUserInfo` class is basically the only use we have for this `namedtupledict`. This class is accessed in a bunch of places through the `util._report_user_dict` function... where in most cases the "dict" is accessed like a "dict" normally is. 

Instead of changing all the references, I just deleted the `DeprecationWarning`, since this `namedtupledict` doesn't seem like it gets much use. 

@mkangia

### DIFF
--- a/corehq/apps/reports/util.py
+++ b/corehq/apps/reports/util.py
@@ -170,18 +170,10 @@ def namedtupledict(name, fields):
 
     def __getitem__(self, item):
         if isinstance(item, basestring):
-            warnings.warn(
-                "namedtuple fields should be accessed as attributes",
-                DeprecationWarning,
-            )
             return getattr(self, item)
         return cls.__getitem__(self, item)
 
     def get(self, item, default=None):
-        warnings.warn(
-            "namedtuple fields should be accessed as attributes",
-            DeprecationWarning,
-        )
         return getattr(self, item, default)
     # return a subclass of cls that has the above __getitem__
     return type(name, (cls,), {


mkangia:
@dannyroberts added this way long back [here](https://github.com/dimagi/commcare-hq/pull/3431/commits/ba541723d6a3a7f7f3dad56bc6e4739eaa0069bb) 
Checking with him if this is okay to remove now?

millerdev (changes requested):
~I don't think we should merge this PR. Those deprecation warnings are there for a reason. If they're being triggered then the code that is triggering them needs to be changed, not this code.~

millerdev (approved):
Just re-read your explanation and I'm coming around to your point of view. Maybe we should deprecate `namedtupledict` or at least make it private so more places don't use it?

proteusvacuum:
Yeah, I'm a bit curious why that warning was even there? It seems like the whole point of this `namedtupledict` is to be able to access the attributes as though they were keys. Maybe the plan was to switch this to a regular `namedtuple`. @dannyroberts, thoughts?

dannyroberts:
Yeah I wrote `namedtupledict` as a bridge from taking something that was a dict and turning it into a `namedtuple`. When I did that, I believe I also changed all usages I could find of accessed as a dict, but I wanted to keep the bridge because it was hard to determine for sure that I'd gotten everything. I never went ahead with the next steps of that plan, so people probably saw that it worked as a dict and just kept using it that way.

I've been out of the game for a while when it comes to this code, so I'm going to let you do whatever you need to do to make the job easier. My intention was definitely that this little package of data should be a well-defined namedtuple (eventually) rather than a dict, for readability and inferability.

proteusvacuum:
going to close this since I don't think it's really worth the effort.

PR #18577, opened 2017-11-17
Ce/more hs

calellowitz:
@orangejenny 
Some more hubspot requests from the marketing team.


corehq/apps/export/views.py
```diff
@@ -172,6 +172,12 @@ def _get_form_exports_by_domain(domain, has_deid_permissions):
     return _get_saved_exports(domain, has_deid_permissions, old_exports_getter, new_exports_getter)
 
 
+def get_export_count_by_domain(domain):
```
orangejenny: Weird that this is in views.py - seems like this and `_get_whatever_exports_by_domain` should all move to dbaccessors.

orangejenny (approved):

corehq/apps/export/views.py
orangejenny: Nevermind, I'm just going to do this in https://github.com/dimagi/commcare-hq/pull/18606

PR #18581, opened 2017-11-17
[hackathon] add form.modified_on date for xforms

snopoke:
Doesn't set values for existing data. That's still to come

Product: this adds a new field to forms 'modified_on'. This is displayed in the form metadata table and is also available in API output and filtering.


corehq/sql_accessors/migrations/0055_set_form_modified_on.py
```diff
@@ -0,0 +1,30 @@
...
+    'FORM_STATE_NORMAL': XFormInstanceSQL.NORMAL,
+    'FORM_OPERATION_ARCHIVE': XFormOperationSQL.ARCHIVE,
+    'FORM_OPERATION_UNARCHIVE': XFormOperationSQL.UNARCHIVE,
+})
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('sql_accessors', '0054_drop_reindexa_accessor_functions'),
```
emord: I think we'll need to figure out something with https://github.com/dimagi/commcare-hq/pull/18538 since it'll have a conflicting migration. Was that successfully deployed to staging?

emord (commented):
Looks good to me but had a question about conflicting migration

corehq/sql_accessors/migrations/0055_set_form_modified_on.py
czue: > Was that successfully deployed to staging?

it was not. I can reconcile migrations before I do that.

snopoke:
@emord  :white_check_mark:

emord (approved):

PR #18584, opened 2017-11-17
Add FormPlayer to docker script

kaapstorm:
Somewhere in here there is a bug related to networking; FormPlayer fails to run because to can't connect to Postgres. The host and port seem correct, but I'm guessing the FormPlayer container can't reach the Postrges container.

I have been testing with 
```bash
$ scripts/docker formplayer
```

My next attempts to debug will be not to start FormPlayer, and just to shell into the container and test connectivity.

@millerdev cc @snopoke 

(Also unsure whether `WEB:8000` should be included in `DEPENDENT_SERVICES`)

### DIFF
--- a/docker/hq-compose-runserver.yml
+++ b/docker/hq-compose-runserver.yml
@@ -33,9 +33,9 @@ formplayer:
     - couch
     - redis
   expose:
-    - 8080
+    - 8010
   ports:
-    - "8080:8080"
+    - "8010:8010"
 
 postgres:
   extends:
--- a/docker/hq-compose-runserver.yml
+++ b/docker/hq-compose-runserver.yml
@@ -21,6 +21,22 @@ web:
     - ..:/mnt/commcare-hq-ro:ro
     - hqservice-lib:/mnt/lib
 
+formplayer:
+  extends:
+    file: hq-compose.yml
+    service: formplayer
+  environment:
+    DEPENDENT_SERVICES: "COUCH:5984 POSTGRES:5432 REDIS:6379"
+    WEB_HOST: "web"
+  links:
+    - postgres
+    - couch
+    - redis
+  expose:
+    - 8080
+  ports:
+    - "8080:8080"
+
 postgres:
   extends:
     file: hq-compose-services.yml
--- a/docker/hq-compose-services.yml
+++ b/docker/hq-compose-services.yml
@@ -10,9 +10,9 @@ formplayer:
     - couch
     - redis
   expose:
-    - 8080
+    - 8010
   ports:
-    - "8080:8080"
+    - "8010:8010"
 
 postgres:
   extends:
--- a/docker/hq-compose-services.yml
+++ b/docker/hq-compose-services.yml
@@ -1,3 +1,19 @@
+formplayer:
+  extends:
+    file: hq-compose.yml
+    service: formplayer
+  environment:
+    DEPENDENT_SERVICES: "COUCH:5984 POSTGRES:5432 REDIS:6379"
+    WEB_HOST: "dockerhost"
+  links:
+    - postgres
+    - couch
+    - redis
+  expose:
+    - 8080
+  ports:
+    - "8080:8080"
+
 postgres:
   extends:
     file: hq-compose.yml
--- a/docker/hq-compose.yml
+++ b/docker/hq-compose.yml
@@ -30,7 +30,7 @@ formplayer:
     DEPENDENT_SERVICES: "COUCH:5984 POSTGRES:5432 REDIS:6379"
     WEB_HOST: "dockerhost"
   expose:
-    - "8080"
+    - "8010"
 
 postgres:
   image: dimagi/docker-postgresql
--- a/docker/hq-compose.yml
+++ b/docker/hq-compose.yml
@@ -24,6 +24,14 @@ web:
     - ..:/mnt/commcare-hq-ro${RO}
     - ${VOLUME_PREFIX}lib:/mnt/lib
 
+formplayer:
+  image: dimagi/formplayer
+  environment:
+    DEPENDENT_SERVICES: "COUCH:5984 POSTGRES:5432 REDIS:6379"
+    WEB_HOST: "dockerhost"
+  expose:
+    - "8080"
+
 postgres:
   image: dimagi/docker-postgresql
   environment:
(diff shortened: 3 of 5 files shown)


docker/hq-compose-formplayer.yml
```diff
@@ -0,0 +1,23 @@
+formplayer:
+  extends:
+    file: hq-compose.yml
+    service: formplayer
+  environment:
+    DEPENDENT_SERVICES: "WEB:8000 COUCH:5984 POSTGRES:5432"
```
snopoke: I don't think listing `WEB:8000` here is going to work unless you also list it under `links` since this relies on docker environment variables.

kaapstorm:
# How to run FormPlayer in Docker

If you run HQ locally, start Docker services as usual with 
```bash
    $ scripts/docker up
```

Or if you run HQ in Docker, start HQ as usual with
```bash
    $ scripts/docker runserver
```

You will see FormPlayer starting with the other containers.

The URL that you use for HQ must match the URL that FormPlayer uses. FormPlayer expects to find HQ at `http://webhost:8000`. 

If you run HQ in Docker, you will need to find the IP address of its container. I wrote a script to find the IP address that FormPlayer uses:
```bash
    $ scripts/get_webhost
```

Add that IP address to your local `/etc/hosts` file, and give it the name "webhost". 

If you run Django locally, and you start Django with 
```bash
    $ ./manage.py runserver 0.0.0.0:8000
```
then you can use the same method. Or you can set "webhost" as an alias of "localhost" in your `/etc/hosts` file.

Once you have configured `/etc/hosts` and started Django, you should be able to get to HQ at *http://webhost:8000/a/demo/*.

Check everything is working correctly by opening App Preview.

See also [formplayer PR 544](https://github.com/dimagi/formplayer/pull/544) for details on FormPlayer in Docker.

kaapstorm:
cc code buddy @dannyroberts, this is the other half of my Hackathon project.

kaapstorm:
Dropping "hold off". I've been running FormPlayer like this for a couple of days, and I don't want to go back.

docker/hq-compose-runserver.yml
```diff
@@ -21,6 +21,22 @@ web:
...
+    service: formplayer
+  environment:
+    DEPENDENT_SERVICES: "COUCH:5984 POSTGRES:5432 REDIS:6379"
+    WEB_HOST: "web"
+  links:
+    - postgres
+    - couch
+    - redis
+  expose:
+    - 8080
```
millerdev: would it be possible/acceptable to use a different port? I use 8080 locally for nginx, and I'd rather not change it if I don't have to.

millerdev (commented):
> The URL that you use for HQ must match the URL that FormPlayer uses. ...

Can you explain the reason for this? It seems inconvenient to have to edit /etc/hosts, and I wonder if there's a way that we could make it just work everywhere without that?

docker/hq-compose-runserver.yml
kaapstorm: I think first prize would be to have it configurable using an environment variable, but this port is determined by FormPlayer's configuration set in (or used by) its Dockerfile, so buildtime not runtime. So I think we'll have to set a fixed port number. (And if there is a better way, I'd love to know it.)

8081? 9080? 

fwiw, I vote 8081.

kaapstorm:
@millerdev re: editing `/etc/hosts`, @wpride will have to confirm, but I suspect that (for App Preview at least) FormPlayer runs in an iframe and needs to query HQ directly. So if the hostname it has for HQ doesn't match the hostname your browser is using, the request will be blocked.

There was a point at which running FormPlayer locally stopped working for me, and I started running it from a VM. For my browser's hostname for HQ ("localhost:8000") to match FormPlayer's hostname for HQ, I used a reverse tunnel. That might work for Docker, but a user on the container would have to authenticate on the host, and that's going to be more of a hassle than editing `/etc/hosts`.

All that said, I agree that adding an entry to the hosts's `hosts` file is ugly, and I'm totally open to better ideas.

docker/hq-compose-runserver.yml
millerdev: haha, I use 8081 locally as well. I pretty much assume 808X is reserved for my personal workspace, so something outside of that range? How about 8010? It [seems to be free](https://en.wikipedia.org/wiki/List_of_TCP_and_UDP_port_numbers) and the HQ codebase doesn't seem to reference it anywhere that I can see.

millerdev:
@kaapstorm 
> FormPlayer runs in an iframe and needs to query HQ directly.

This is exactly what I was wondering about, and that's all the explanation I need. In that case, carry on as planned.

kaapstorm:
Commit 2737464 moves FormPlayer to port 8010. You will need to update `localsettings.py` to: 
```python
FORMPLAYER_URL = 'http://localhost:8010'
```

dannyroberts (approved):

millerdev (approved):

millerdev (approved):

kaapstorm:
@dimagi/team-commcare-hq and maybe @dimagi/commcare-mobile too, this is a heads-up for those who run HQ or HQ's services in Docker. This PR now adds FormPlayer to those services. 

No action is required if you want to keep using FormPlayer whichever way you use it at the moment, or if you never use App Preview or Web Apps or build CommCare apps.

But if you want to use FormPlayer in Docker, there is some stuff you need to do:

  1. Start Docker as usual, either with `scripts/docker up` (services only) or `scripts/docker runserver` (HQ and services).
  2. Run `scripts/get_webhost` and append its output to `/etc/hosts`. (Or if you run Linux, `scripts/get_webhost | sudo tee -a /etc/hosts`.)
  3. If you run Django locally:
    3.1. Update `localsettings.py` with `FORMPLAYER_URL = 'http://localhost:8010'`
    3.2. Start Django with `./manage.py runserver 0.0.0.0:8000`.
  4. From now on, open your local HQ instance at [http://webhost:8000/](http://webhost:8000/) (This is so that FormPlayer can use the same URL from inside its Docker container.)

That should be it. If it doesn't Just Work, please yell at me on Slack.

czue:
yes!!! :clap: :clap: :clap: 

could be nice to get the instructions in this PR into the README?

ctsims:
Amazing! Thanks, Norman.

PR #18585, opened 2017-11-17
Aggregate User Status Report

czue:
## Problem :scream: 

With large projects (and particularly ICDS) it can be hard to get great visibility into how regularly your users are able to interact with the application. Most of our reports are either per-user and therefore challenging to analyze at scale, or don't show actionable data in an easy-to-understand way.

## Solution? :white_check_mark: 

A new report that shows you when your user's last interactions (submissions/syncs) with the site were.

### Current Version
image

### Old version showing graph behavior

out

## Details :mag: 

- Currently behind a product-labeled feature flag (to be confirmed with product team)
- Lives in Deployments section
- Can filter by user/group/location (though users are pretty silly)
- The two charts are last submission date and last sync date
- The toggles are for a daily breakdown versus "% of total since that date"

## Tech :computer: 

- Standard HQ report framework stuff + ES aggregations + nvd3. :bar_chart: 
- Am optimistic it should scale fine.  :crossed_fingers: 

## For Reviewer :bust_in_silhouette: 

- Probably easiest to review all at once :cityscape: 
- Probably a fair amount of lint errors that I'll look at :upside_down_face:  
- Could probably use some copy updates :pencil2: 

@dimagi/product would love your input on this

giphy


orangejenny:
This is fun!

Based on what you said while demoing, it sounds like the main value of the percentage view is to get evidence that most users have synced/submitted relatively recently. If that's the case, it might be clearer to do a high-level summary rather than a graph - something like "X% of users have submitted within the past 24 hour, Y% in the last week, Z% in the last month". That information also wouldn't then be a click away (if we want to emphasize that things are basically fine, might be nice to demonstrate that immediately), and with that graph gone, you could use the traditional past => future direction for the daily breakdown's x axis.

Are the top and bottom charts always going to line up? Haven't read the code, just looking at the image, can't tell if it's coincidence that they both stop when they hit 100% and that happened the same day for both sync and submit.

czue:
thanks!

> it might be clearer to do a high-level summary rather than a graph - something like "X% of users have submitted within the past 24 hour, Y% in the last week, Z% in the last month".

I really like this idea, will play with it a bit.

> Are the top and bottom charts always going to line up? 

Yeah they should. Currently buckets the last 60 days and then dumps the remaining people into the last bucket.

czue:
@orangejenny what do you think about this updated screenshot?

image

orangejenny:
I like it! Gets the point across. And I like the "have ever" stats at the top.

Super-picky, but I might drop the decimal from the percentages to give a tiny bit more space.

czue:
> I might drop the decimal from the percentages to give a tiny bit more space.

good call. https://github.com/dimagi/commcare-hq/pull/18585/commits/5ba0501b06cb0afb4ee2aac56e1d39efa5c162b5

czue:
i'm buddyless. anyone able to help with a review?

snopoke:
@czue looking now

snopoke (approved):

PR #18598, opened 2017-11-17
Case Property Changes Modal

proteusvacuum:
Tired of searching for the form that updated a specific case property? Want to see the progression of how a specific thing changed over time? Then this PR is for you! 

This shows a modal when you click on the case property name on the case_data report which shows you the history of a particular case property. Really useful if you have a lot of forms updating a small set of case properties. 

screenshot from 2017-11-17 17-21-18

screenshot from 2017-11-17 17-14-18

Looking for UI feedback here (@biyeun, @orangejenny). I'm not happy with adding link styling to every case property on this page, as it is a page that is used all the time, and all the blue might get overwhelming... 

@mkangia code buddy

### DIFF
--- a/corehq/apps/reports/views.py
+++ b/corehq/apps/reports/views.py
@@ -1299,6 +1299,7 @@ class CaseDetailsView(BaseProjectReportSectionView):
                     self.urlname, args=[self.domain, case_id]),
                 "show_transaction_export": toggles.COMMTRACK.enabled(
                     self.request.user.username),
+                "property_details_enabled": toggles.SUPPORT.enabled(self.request.user.username),
             },
             "show_case_rebuild": toggles.SUPPORT.enabled(self.request.user.username),
             "can_edit_data": self.request.couch_user.can_edit_data,
--- a/corehq/apps/reports/views.py
+++ b/corehq/apps/reports/views.py
@@ -17,6 +17,7 @@ from corehq.apps.users.permissions import FORM_EXPORT_PERMISSION, CASE_EXPORT_PE
     DEID_EXPORT_PERMISSION
 from corehq.form_processor.utils.general import use_sqlite_backend
 from corehq.tabs.tabclasses import ProjectReportsTab
+from corehq.util.timezones.conversions import ServerTime
 import langcodes
 import os
 import pytz
@@ -1315,9 +1316,7 @@ def form_to_json(domain, form, timezone=None):
     if timezone is None:
         received_on = json_format_datetime(form.received_on)
     else:
-        offset = timezone.utcoffset(datetime.utcnow())
-        local_received_on = form.received_on + offset
-        received_on = local_received_on.strftime("%Y-%m-%d %H:%M")
+        received_on = ServerTime(form.received_on).user_time(timezone).done().strftime("%Y-%m-%d %H:%M")
 
     return {
         'id': form.form_id,
--- a/corehq/apps/reports/views.py
+++ b/corehq/apps/reports/views.py
@@ -1359,7 +1359,7 @@ def case_forms(request, domain, case_id):
 def case_property_changes(request, domain, case_id, case_property_name):
     """Returns all changes to a case property
     """
-    case = CaseAccessors(domain).get_case(case_id)
+    case = _get_case_or_404(domain, case_id)
     changes = []
     timezone = get_timezone_for_user(request.couch_user, domain)
     for change in reversed(get_all_changes_to_case_property(case, case_property_name)):
--- a/corehq/apps/reports/views.py
+++ b/corehq/apps/reports/views.py
@@ -1369,7 +1369,6 @@ def case_property_changes(request, domain, case_id, case_property_name):
 
     context = {
         'domain': domain,
-        'case_id': case_id,
         'timezone': timezone.localize(datetime.utcnow()).tzname(),
         'property_name': case_property_name,
         'changes': changes,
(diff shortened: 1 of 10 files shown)


orangejenny:
> I'm not happy with adding link styling to every case property on this page, as it is a page that is used all the time, and all the blue might get overwhelming...

I don't think it's terrible. But since you asked, possible alternatives:
- Instead of making the property name a link, put a history-type icon next to the name. This has the advantage of being more discoverable, but it may well look as overwhelming as a sea of blue text, or even worse.
- Instead of setting this up as a modal for each case property, make an additional "Case Property History" tab on the page and then on that tab, have a select2 dropdown of case properties. You pick a property, it loads the history. If your code is already doing an ajax request for each case property's history (I haven't read the code but am guessing it does) it probably/hopefully isn't that big of a change.

proteusvacuum:
@orangejenny yeah, I tried the icon, and that was even more overwhelming. If you think this isn't terrible, then I'm inclined to leave it as is.

orangejenny:
@proteusvacuum works for me

corehq/apps/reports/views.py
```diff
@@ -1304,6 +1305,31 @@ def page_context(self):
...
+        form.xmlns,
+        app_id=form.app_id,
+        lang=get_language(),
+    )
+    if timezone is None:
+        received_on = json_format_datetime(form.received_on)
+    else:
+        offset = timezone.utcoffset(datetime.utcnow())
+        local_received_on = form.received_on + offset
+        received_on = local_received_on.strftime("%Y-%m-%d %H:%M")
```
snopoke: why not use `json_format_datetime` to be consistent with above?

corehq/apps/reports/views.py
```diff
@@ -1304,6 +1305,31 @@ def page_context(self):
...
+        domain,
+        form.xmlns,
+        app_id=form.app_id,
+        lang=get_language(),
+    )
+    if timezone is None:
+        received_on = json_format_datetime(form.received_on)
+    else:
+        offset = timezone.utcoffset(datetime.utcnow())
+        local_received_on = form.received_on + offset
```
snopoke: Could do `ServerTime(form.received_on).user_time(timezone)`

snopoke (commented):
Didn't review closely but looks good. How long does this take to load for a case with 100 forms?

May need to paginate the list. Some cases have >1000 forms.

corehq/apps/reports/views.py
proteusvacuum: Oh, didn't know about this. Will update.

corehq/apps/reports/views.py
proteusvacuum: The "forms" page that uses the `json_format` formats the datetime for display using javascript which ends up outputting the same thing. I wasn't passing any of my values through js, and it seemed easier to just do this in python. 

https://github.com/dimagi/commcare-hq/blob/master/corehq/ex-submodules/casexml/apps/case/static/case/js/case_details.js#L14-L42

proteusvacuum:
@snopoke 

It takes about 7 seconds to run a similar command for a case with 108 submissions against it. That seems pretty high, so I'll probably also have to add a loading screen. 
I think I can make pagination help with this somewhat, although since we have to pull every form for every property, the timing per page can be inconsistent (e.g. if there are 1000 form submissions, but only the first 10 of these change the case property you are looking at, we'd need to loop through all 1000 submissions before displaying anything). Not sure if you have any other suggestions on how I can do this in bulk, or by using some other approach? 

```
In [36]: c = CaseAccessors('enikshay').get_case('33c657e3-3730-4d6a-b1fa-b5af90b297a3')

In [37]: %time ci = get_case_property_changed_info(c, 'age_entered', 3)
CPU times: user 3.16 s, sys: 64 ms, total: 3.22 s
Wall time: 7.2 s
```

snopoke:
@proteusvacuum the only other approach I can think of is to paginate it by date range. Not sure if that's going to cause usability issues thouhg.

sheelio:
@proteusvacuum 
I think this looks good.  GIven the performance concerns and UX impact, my preference would be to either:
a) Put it behind an internal feature flag 
b) Create the case property history tab that Jenny suggested

proteusvacuum:
I'm putting this behind the support toggle for now. 

I've sunk about 5 hours into getting this paginated and the property table UI code to play nicely with javascript and am giving up for now since the deadline for our hackathon entries was on Friday. 

My WIP branch for that work lives here: https://github.com/dimagi/commcare-hq/compare/fr/property-changes-paginated 

I got the pagination working in the backend, but was unable to hook it into the UI in a reasonable amount of time, which I'm a little annoyed about. @orangejenny or @biyeun could we sit down together sometime to go through what I was attempting here? This page seems like it is looking for a refactor, as it was pretty much impossible for me to add anything to it in a reasonable manner, but I might have just misunderstood the point of all the indirection.

orangejenny:
To my knowledge, there isn't a standard pagination pattern in HQ other than what's available via datatables. There's bootstrap's `.pagination` for styling, but the intricate part is the behavior. I've been thinking about making a knockout component for this purpose, just haven't gotten there yet.

snopoke:
I'm 100% in favor of rewriting these pages to be 'normal' templates. All the indirection was to allow domains to customize the way the page was displayed. We could check to see if any domains are using that feature but I don't think there are (certainly none that have been setup in the last 4 years)

PR #18601, opened 2017-11-17
Kill report builder v1

orangejenny:
@calellowitz / @biyeun 

@sheelio approved offline, just marking "don't merge" because I like to be around for the deploy when large chunks of code are deleted.


biyeun (approved):
🎉

PR #18604, opened 2017-11-18
Truncate Debugger Output

ctsims:
Contains commits from:
https://github.com/dimagi/commcare-hq/pull/18593

makes more sense to review after that is pulled

Changes the way that debug info is presented in the debugger pane. By default only 50 lines of any result are shown (5 lines in the "previous queries" pane), along with a link you can click to copy the remaining results to the clipboard.

image

Previously it was quite easy to make the ~entire pane unusable by typing one heinous query, since the code formatting lib takes what seems like a polynomial time based on input size (and tanks the browser memory). This should be a no-op for basic users, and a drastic improvement to anyone doing realistic chunks of work.

Also removes the interior scroll pane (since the output is now self-limiting) of the results window, so there's no more nested scrollbox!


corehq/apps/cloudcare/static/cloudcare/js/debugger/debugger.js
```diff
@@ -410,6 +496,16 @@ hqDefine('cloudcare/js/debugger/debugger', function () {
                 }
             },
         };
+        ko.bindingHandlers.clipboardButton = {
+            init: function(element, valueAccessor) {
+                new Clipboard('.clickable',
```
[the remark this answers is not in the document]
ctsims: This is the library's mechanism as documented here:
https://clipboardjs.com/

not sure if there's a way around that.

corehq/apps/cloudcare/static/cloudcare/js/debugger/debugger.js
```diff
@@ -225,10 +283,36 @@ hqDefine('cloudcare/js/debugger/debugger', function () {
...
+            }
+            return [first, second];
+        }
+
         self.formatResult = function (output) {
             return output.replace(resultRegex, "$1");
         };
 
+        self.truncateResult = function(output, maxLines, addElipsis) {
+            var items = output.split(RegExp("\r?\n"));
```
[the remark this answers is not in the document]
ctsims: Need these.

corehq/apps/cloudcare/static/cloudcare/js/debugger/debugger.js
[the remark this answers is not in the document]
orangejenny: You can fix this with a `globals` declaration, like [this](https://github.com/dimagi/commcare-hq/blob/master/corehq/apps/app_manager/static/app_manager/js/settings/commcare_settings.js#L1).

corehq/apps/cloudcare/static/cloudcare/js/debugger/debugger.js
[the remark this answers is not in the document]
orangejenny: You can silence stickler by adding a comment with `eslint-disable-line no-control-regex`, see [here](https://eslint.org/docs/user-guide/configuring).

ctsims:
@orangejenny Thanks for the linter advice! PR now passes linting and can be reviewed.

corehq/apps/cloudcare/static/cloudcare/js/debugger/debugger.js
```diff
@@ -436,6 +496,17 @@ hqDefine('cloudcare/js/debugger/debugger', function () {
                 }
             },
         };
+        ko.bindingHandlers.clipboardButton = {
+            init: function(element, valueAccessor) {
+                new Clipboard('.clickable',
```
proteusvacuum: should this enable clipboard on the element that was clicked? I think this will enable it for all elements with the `.clickable` class.

corehq/apps/cloudcare/templates/formplayer/debugger.html
```diff
@@ -89,18 +89,34 @@
     <!--ko if: xPathQuery.errorResult()-->
     <div data-bind="text: xPathQuery.errorResult(), css: 'text-danger'"></div>
     <!--/ko-->
-    <!--ko if: options.useCodeMirror && xPathQuery.successResult()-->
+    <!--ko if: xPathQuery.isSuccessTruncated() && options.showInFull-->
+    <div>
+        <i>Showing the first </i><i data-bind="text: xPathQuery.getMaxLines()"/><i> lines...</i>
```
proteusvacuum: Is this meant to be _italics_?  You should use the `<em>` tag instead. 
```html
<em>Showing the first <span data-bind="text: xPathQuery.getMaxLines()"/></span> lines...</em>
```

corehq/apps/cloudcare/templates/formplayer/debugger.html
```diff
@@ -89,18 +89,34 @@
...
+    <!--/ko-->
+    <!--ko if: options.showInFull && xPathQuery.successResult()-->
     <textarea type="text" data-bind="
         codeMirror: xPathQuery.successResult()
     "></textarea>
     <!--/ko-->
-    <!--ko if: options.useCodeMirror && xPathQuery.hasTrace()-->
+    <!--ko if: xPathQuery.isTraceTruncated() && options.showInFull-->
+    <div>
+        <i>Showing the first </i><i data-bind="text: xPathQuery.getMaxLines()"/><i> lines...</i>
```
proteusvacuum: same `<em>` comment

corehq/apps/cloudcare/templates/formplayer/debugger.html
```diff
@@ -89,18 +89,34 @@
...
+        <a class="clickable pull-right" data-clipboard-action="copy" data-bind="clipboardButton: xPathQuery.getFullTraceResult()">
```
proteusvacuum: `clickable` is a confusing name for this. I think `clipboard` or something similar might be a better name.

corehq/apps/cloudcare/static/cloudcare/js/debugger/debugger.js
```diff
@@ -426,7 +486,7 @@ hqDefine('cloudcare/js/debugger/debugger', function () {
                 };
                 options.value = ko.unwrap(valueAccessor());
                 var editor = CodeMirror.fromTextArea(element, options);
-                editor.setSize(null, 200);  // hard-coded right now;
+                editor.setSize(null, 'auto');  // hard-coded right now;
```
proteusvacuum: is the `// hard-coded right now` comment still valid now that this is `'auto'`?

corehq/apps/cloudcare/templates/formplayer/debugger.html
```diff
@@ -89,18 +89,34 @@
     <!--ko if: xPathQuery.errorResult()-->
     <div data-bind="text: xPathQuery.errorResult(), css: 'text-danger'"></div>
     <!--/ko-->
-    <!--ko if: options.useCodeMirror && xPathQuery.successResult()-->
+    <!--ko if: xPathQuery.isSuccessTruncated() && options.showInFull-->
+    <div>
+        <i>Showing the first </i><i data-bind="text: xPathQuery.getMaxLines()"/><i> lines...</i>
+        <a class="clickable pull-right" data-clipboard-action="copy" data-bind="clipboardButton: xPathQuery.getFullSuccessResult()">
+            Copy Full Evaluation to Clipboard
```
proteusvacuum: does the copied value contain valid XML? I thought that we removed whatever wrapper node was here. Probably fine, but might be weird if you paste it into a text editor and it gives you syntax errors.

corehq/apps/cloudcare/static/cloudcare/js/debugger/debugger.js
```diff
@@ -249,10 +283,36 @@ hqDefine('cloudcare/js/debugger/debugger', function () {
             '^<[?]xml version="1.0" encoding="UTF-8"[?]>\\s*<result>\n*([\\s\\S]*?)\\s*</result>\\s*|' +
             '^<[?]xml version="1.0" encoding="UTF-8"[?]>\\s*<result/>()\\s*$');
 
+        self.getBody = function(output) {
+            if (!output) {
+                return ['',''];
+            }
+            var first = self.formatResult(output);
```
proteusvacuum: what is `first` and `second`?

corehq/apps/cloudcare/templates/formplayer/debugger.html
orangejenny: yes, <3 semantic tags

corehq/apps/cloudcare/templates/formplayer/debugger.html
ctsims: kind of, it's XML but not reliably a doc (Similar to what's there now).

ctsims:
@proteusvacuum Feedback Addressed

corehq/apps/cloudcare/templates/formplayer/debugger.html
```diff
@@ -104,9 +104,9 @@
     <!--/ko-->
     <!--ko if: xPathQuery.isTraceTruncated() && options.showInFull-->
     <div>
-        <i>Showing the first </i><i data-bind="text: xPathQuery.getMaxLines()"/><i> lines...</i>
+        <em>{% trans "Showing the first" %} <span data-bind="text: xPathQuery.getMaxLines()"/>{% trans " lines..." %}</em>
```
proteusvacuum: nit, but having the leading space and the ellipsis in the translate tag will mean that the translator will have to add those into the translation too which is a bit awkward.

proteusvacuum (approved):

corehq/apps/cloudcare/templates/formplayer/debugger.html
ctsims: Fixed. I left the "..." because I think that's part of the translated string, but agree WRT the spaces.

PR #18605, opened 2017-11-18
Add Missing Frameworks for new analytics

biyeun:
@orangejenny / @calellowitz 

Adds
- drift
- hubspot

Also clears up analytics not loading with adblockers and issues with kissmetrics


corehq/apps/analytics/static/analytics/js/drift.js
```diff
@@ -0,0 +1,56 @@
...
+            _drift.factory = function (methodName) {
+               return function() {
+                    var methodFn = Array.prototype.slice.call(arguments);
+                    methodFn.unshift(methodName);
+                    _drift.push(methodFn);
+                    return _drift;
+                };
+            };
+            _.each(_drift.methods, function (methodName) {
+                _drift[methodName] = _drift.factory(methodName);
```
orangejenny: Picky, but I think it'd be clearer to put the guts of `factory` in here rather than pulling them into a separate function.

orangejenny (approved):

PR #18606, opened 2017-11-19
Show new exports in dashboard 'Data' tile

orangejenny:
Dashboard only shows old exports. The fact that no one has complained about this perhaps says something about the utility of the data dashboard tile. Easy enough to fix, though.

Also moved some dbaccessors-type code from views to dbaccessors.

@calellowitz / @esoergel


corehq/apps/export/dbaccessors.py
```diff
@@ -87,7 +87,7 @@ def _get_saved_exports(domain, has_deid_permissions, old_exports_getter, new_exp
         from corehq.apps.export.utils import revert_new_exports
         exports += revert_new_exports(new_exports)
     if not has_deid_permissions:
-        exports = filter(lambda x: not x.is_safe, exports)
+        exports = [e for e in exports if e.is_safe]
```
orangejenny: This reverses the logic of the previous code, which **really** looks like a bug and could plausibly have gone undetected.

esoergel (approved):
Nice!

PR #18608, opened 2017-11-20
Implement recipient selection in new broadcast UI

gcapalbo:
Implements the different recipient selection mechanisms in the new broadcast UI.
spec: https://docs.google.com/document/d/11dzolEG2bc5jeuElX314NE3qoeJ4DsZomN-YerX4SZA/edit#

Screenshot:
pr-recipients

@esoergel @emord


corehq/apps/casegroups/dbaccessors.py
```diff
@@ -18,5 +18,15 @@ def get_case_group_meta_in_domain(domain):
     return [(group.get_id, group.name) for group in get_case_groups_in_domain(domain)]
 
 
+def search_case_groups_in_domain(domain, search_string, limit=10):
+    count = 0
+    for result in get_case_group_meta_in_domain(domain):
+        if search_string.lower() in result[1].lower():
```
emord: there should be very few of these right? I see it's cached for an hour, so it's probably not a big concern

corehq/messaging/scheduling/async_handlers.py
```diff
@@ -24,6 +29,49 @@ def recipients_response(self):
...
+        ]
+
+    @property
+    def user_organization_recipients_response(self):
+        domain = self.request.domain
+        query = self.data.get('searchString')
+        result = (
+            SQLLocation
+            .active_objects
+            .filter(domain=domain, name__icontains=query)
```
emord: Manish wrote a location search that also searches with / https://github.com/dimagi/commcare-hq/pull/17958/ . So that you can search "MA / Boston" or something similar. Should this input allow that as well?

Also from that PR, I think you should be able to search for site code

corehq/apps/casegroups/dbaccessors.py
gcapalbo: yes, there aren't many of these

corehq/messaging/scheduling/async_handlers.py
gcapalbo: just looked into that, and it looks like it will require a bit of work to integrate with. since the location selection mechanism in the current broadcast ui is similar to what's implemented in this pr, i'd rather leave this as is for now and add that later as an enhancement if we choose to

emord: 👍

PR #18609, opened 2017-11-20
domain stats

snopoke:
This isn't very pretty but made it much easier to get these numbers for TDH


emord (approved):
Ugh. another hosted instance?

PR #18615, opened 2017-11-20
Migrate Analytics

biyeun:
@orangejenny cc @calellowitz 

please also merge:

https://github.com/dimagi/commcarehq-prelogin/pull/146
https://github.com/dimagi/Vellum/pull/929


orangejenny:
There's a commcare-hq-deploy change attached to https://github.com/dimagi/commcare-hq/pull/18615/commits/a192e6ccf6afe4092484debd0903c69f63471c66

orangejenny:
Test failures looks real:

`TemplateDoesNotExist: hqwebapp/includes/analytics_all.html`

and

`Warning: Error: The module 'hqwebapp/js/initial_page_data' has already been defined elsewhere.` (need to remove the initial_page_data.js script tag from app manager's mocha.html)

.eslintrc.js
```diff
@@ -18,7 +18,6 @@ module.exports = {
         "$": false,
         "ko": false,
         "_": false,
-        "ga_track_event": false,
```
orangejenny: So nice to see something getting removed from this list.

corehq/apps/analytics/static/analytix/js/kissmetrix.js
```diff
@@ -60,11 +61,14 @@ hqDefine('analytics/js/kissmetrics', function () {
             testName = _.last(testName.split('.'));
             if (_.isObject(ab) && ab.version) {
                 test[ab.name || testName] = ab.version;
-            } else {
+            } else if (!_.isEmpty(ab)) {
```
orangejenny: It looks like `getAbTests` only returns objects with keys for `slug` and `context`, why does this if/else test if ab is an object and look for ab.name and ab.version, and then potentially treat ab as a string? Could use a comment explaining why this is all needed.

orangejenny:
Looks good, just some minor questions/comments.

orangejenny:
@biyeun Added a commit that should fix the js test. The python test will be fixed once the prelogin PR is merged and the submodule update is committed to this PR.

corehq/apps/analytics/static/analytix/js/kissmetrix.js
biyeun: i think this is me overengineering 😳

orangejenny (approved):

orangejenny:
Vellum PR is merged, but I'm having trouble running `vellum-to-hq`, therefore holding off on merging this.

orangejenny:
Ran `vellum-to-hq` so https://github.com/dimagi/Vellum/pull/929 is now merged into HQ (https://github.com/dimagi/commcare-hq/commit/194bd316d28a19ddb822f021e06023f1d6b5d497).

orangejenny:
Hey, I didn't notice during review, but the module names here don't match the file names: modules are `analytics/js/foo` but files are `analytix/js/foo.js`. The `analytix` workaround is weird (necessary, but weird), but I'd rather have the module names say `analytix` so we can keep the convention that module names always match filenames - long-term requirejs plan is to to use the variant of `define` where you don't specify a module name at all. Thoughts?

PR #18616, opened 2017-11-20
bulk case pillow

emord:
@snopoke Not that happy with this, but a rough outline of bulk submitting to ES. The only improvement here is submitting the changes to ES in bulk. It still grabs changes from couch/sql one by one. I also wasn't able to bulk insert PillowError because we don't have upserts in pg 9.4

### DIFF
--- a/corehq/ex-submodules/pillowtop/processors/elastic.py
+++ b/corehq/ex-submodules/pillowtop/processors/elastic.py
@@ -1,4 +1,5 @@
 from __future__ import absolute_import
+from collections import namedtuple
 from datetime import datetime
 import math
 import time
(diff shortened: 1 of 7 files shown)


corehq/ex-submodules/pillowtop/processors/elastic.py
```diff
@@ -63,6 +69,83 @@ def _delete_doc_if_exists(self, doc_id):
             self.elasticsearch.delete(self.index_info.index, self.index_info.type, doc_id)
 
 
+class BulkElasticProcessor(ElasticProcessor):
```
snopoke: Seems very similar to the `BulkPillowReindexProcessor` class.

snopoke (commented):
Looks like the right direction

corehq/ex-submodules/pillowtop/processors/elastic.py
emord: Agreed. I think at some point we can consolidate these processors, but right now reindex accessors don't actually try to handle errors from getting the docs or from inserting them into ES. It's not clear to me where or if those errors fit into the reindex workflow.

I thought it might make sense to write this out here specifically for cases, ensure it works/helps us scale and then try to generalize it to do the same for xforms/case search/whatever else

PR #18617, opened 2017-11-20
Hackathon: Dashboard: angular => knockout ...and kill template apps

orangejenny:
Easier to review by commit - there are a lot but many of them are short.

Also kills template apps.

@calellowitz / @nickpell 

QA: http://manage.dimagi.com/default.asp?265945


nickpell (approved):

snopoke (commented):
Didn't review closely but that seems like a lot less code.

PR #18620, opened 2017-11-20
make userreports filter py3 compatible

nickpell:
@dimagi/py3


orangejenny (approved):

corehq/apps/userreports/sql/data_source.py
```diff
@@ -27,7 +27,7 @@ def engine_id(self):
 
     @property
     def filters(self):
-        return filter(None, [fv.to_sql_filter() for fv in self._filter_values.values()])
+        return [_f for _f in [fv.to_sql_filter() for fv in self._filter_values.values()] if _f]
```
emord: Does this need the inner `[]`?

calellowitz: to second this, seems like it could be  `[fv.to_sql_filter() for fv in self._filter_values.values() if fv.to_sql_filter()]`

calellowitz (changes requested):

corehq/apps/userreports/sql/data_source.py
emord: that could be bad if `to_sql_filter` is expensive since that would call it twice

nickpell: I'd prefer not to cal ```to_sql_filter``` twice for that reason.  Maybe ```list(filter(None, [...]))``` would actually be more clear?

orangejenny: Brackets within brackets seems clearer than `filter(None, [...])`

nickpell: going to go with brackets within brackets as is

PR #18623, opened 2017-11-20
dict compatibility - icds_reports

nickpell:
@dimagi/py3


nickpell:
will handle stickler feedback in a following pass

PR #18624, opened 2017-11-20
use six.iteritems

nickpell:
@dimagi/py3


emord (approved):

corehq/apps/app_manager/views/app_summary.py
```diff
@@ -142,7 +143,7 @@ def _translate_name(names, language):
     try:
         return names[language]
     except KeyError:
-        first_name = names.iteritems().next()
+        first_name = six.iteritems(names)
```
calellowitz: this looks different than what was there before. i think you are missing `.next()`

corehq/util/pagination.py
```diff
@@ -195,7 +196,7 @@ def __iter__(self):
 
         retried = {}
         while self.state.retry != retried:
-            for item_id, retries in list(self.state.retry.iteritems()):
+            for item_id, retries in list(six.iteritems(self.state.retry)):
```
calellowitz: why does this need to be wrapped in `list`

calellowitz (changes requested):
just a few minor questions. there were also few files where you did not `import six`, assuming thats because those were already there but did not check each one

corehq/apps/app_manager/views/app_summary.py
nickpell: Good catch - updated

corehq/util/pagination.py
nickpell: ```list``` looks unnecessary to me.

@snopoke was there a reason ```list``` was used here?

snopoke: Can't think why I'd have added the `list`. Doesn't seem necessary.

PR #18626, opened 2017-11-21
check individual es hosts

snopoke:
In an attempt to detect when a single ES node fails invisibly

https://manage.dimagi.com/default.asp?260042

### DIFF
--- a/corehq/apps/hqadmin/service_checks.py
+++ b/corehq/apps/hqadmin/service_checks.py
@@ -26,7 +26,7 @@ from corehq.apps.change_feed.connection import get_kafka_client_or_none
 from corehq.apps.es import GroupES
 from corehq.blobs import get_blob_db
 from corehq.blobs.util import random_url_id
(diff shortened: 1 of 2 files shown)


corehq/apps/hqadmin/service_checks.py
```diff
@@ -100,11 +100,15 @@ def check_elasticsearch():
            'date': datetime.datetime.now().isoformat()}
     send_to_elasticsearch('groups', doc)
     refresh_elasticsearch_index('groups')
-    hits = GroupES().remove_default_filters().doc_id(doc['_id']).run().hits
+    bad_hosts = []
+    for host_config in es_hosts():
+        hits = GroupES(debug_host=host_config).remove_default_filters().doc_id(doc['_id']).run().hits
```
emord: I think this will only work if group has a shard on each node which could cause false positives

snopoke: hmm, good point

PR #18627, opened 2017-11-21
make use of heartbeat data

snopoke:
probably best reviewed per commit :fish: 

Still working on how best to expose this data on the UI.

I do realize that this duplicates the data in `reporting_metadata`. I think it would be possible to replace `reporting_metadata` with this data model but I haven't confirmed that yet.


snopoke:
Ignoring stickler comments which don't seem correct

corehq/apps/ota/views.py
```diff
@@ -250,27 +260,69 @@ def _log_time_since_last_sync(restore_config):
...
         info.update(LatestAppInfo(brief_app_id, domain).get_info())
+    else:
+        couch_user = request.couch_user
+        save_user = update_latest_builds(couch_user, app_id, datetime.utcnow(), app_version)
+        try:
+            last_sync = adjust_text_to_datetime(last_sync_time)
+        except iso8601.ParseError:
+            pass
+        else:
+            save_user |= update_last_sync(couch_user, app_id, last_sync, app_version)
```
snopoke: `update_last_sync` and `update_last_sync` aren't really necessary but shouldn't hurt

corehq/apps/users/util.py
```diff
@@ -258,15 +259,15 @@ def update_latest_builds(user, app_id, date, version):
...
-def filter_by_app(data_list, app_id):
+def filter_by_app(obj_list, app_id):
     """
-    returns the last sync, submission, or build for the given app id
-    :param data_list: list from user's reporting metadata (last syncs, last submissions, or last builds)
+    :param obj_list: list from objects with ``app_id`` property
+    :returns: The object with matching app_id
     """
-    last_items = filter(
+    last_items = list(filter(
```
emord: Might be easier as a list comprehension

corehq/apps/users/models.py
```diff
@@ -2125,20 +2189,33 @@ def update_device_id_last_used(self, device_id, when=None):
...
             return True
 
+    def get_last_used_device(self):
+        if not self.devices:
+            return None
+
+        return sorted(self.devices, key=lambda dev: dev.last_used)[-1]
+
+    def get_device(self, device_id):
+        matches = [
```
emord: This could return early if written like (I don't htink python has a `first` method):

```
for device in devices:
    if device.device_id == device_id:
        return device

return None
```

emord (changes requested):
I think the stickler warnings also apply (mostly changing filters to list comprehensions)

corehq/apps/users/util.py
```diff
@@ -232,10 +232,9 @@ def update_latest_builds(user, app_id, date, version):
...
     from corehq.apps.users.models import LastBuild
-    last_builds = filter(
-        lambda build: build.app_id == app_id,
-        user.reporting_metadata.last_builds,
-    )
+    last_builds = [
+        build for build in user.reporting_metadata.last_builds if build.app_id == app_id
+    ]
     if last_builds:
         assert len(last_builds) == 1, 'Must only have one last build per app'
```
emord: I think htis assert could removed if the above was a dict/set comprehension. Do we think this assert is valuable?

emord (approved):
Looks good but test failure looks real

```
ERROR: corehq.apps.users.tests.test_user_model:UserDeviceTest.test_merge_device_app_meta
  File "/mnt/commcare-hq-ro/corehq/apps/users/tests/test_user_model.py", line 140, in test_merge_device_app_meta
    m2.merge(m1)
  File "/mnt/commcare-hq-ro/corehq/apps/users/models.py", line 832, in merge
    if other.latest_request <= self.last_request:
AttributeError: 'DeviceAppMeta' object has no attribute 'latest_request'
```

corehq/apps/users/util.py
snopoke: e257257

PR #18630, opened 2017-11-21
reindex accessor refactor + mgmt command to populate form.server_modified_on

snopoke:
See also https://github.com/dimagi/commcare-hq/pull/18581

### DIFF
--- a/corehq/form_processor/backends/sql/dbaccessors.py
+++ b/corehq/form_processor/backends/sql/dbaccessors.py
@@ -73,14 +73,13 @@ def iter_all_rows(reindex_accessor):
     reindex accessor
     """
     for db_alias in reindex_accessor.sql_db_aliases:
-        docs = reindex_accessor.get_docs(db_alias, reindex_accessor.startkey_min_value)
+        docs = reindex_accessor.get_docs(db_alias)
         while docs:
             for doc in docs:
                 yield doc
 
-            start_from_for_db = getattr(doc, reindex_accessor.startkey_attribute_name)
             last_id = doc.id
-            docs = reindex_accessor.get_docs(db_alias, start_from_for_db, last_doc_pk=last_id)
+            docs = reindex_accessor.get_docs(db_alias, last_doc_pk=last_id)
 
 
 class ShardAccessor(object):
@@ -182,8 +181,6 @@ class ShardAccessor(object):
 
 
 class ReindexAccessor(six.with_metaclass(ABCMeta)):
-    startkey_min_value = datetime.min
-
     def __init__(self, limit_db_aliases=None):
         self.limit_db_aliases = limit_db_aliases
 
@@ -212,13 +209,6 @@ class ReindexAccessor(six.with_metaclass(ABCMeta)):
         """
         raise NotImplementedError
 
-    @abstractproperty
-    def startkey_attribute_name(self):
-        """
-        :return: The name of the attribute to filter successive batches by e.g. 'received_on'
-        """
-        raise NotImplementedError
-
     @abstractmethod
     def get_doc(self, doc_id):
         """
@@ -235,7 +225,7 @@ class ReindexAccessor(six.with_metaclass(ABCMeta)):
         return doc.to_json()
 
     @abstractmethod
-    def get_docs(self, from_db, startkey, last_doc_pk=None, limit=500):
+    def get_docs(self, from_db, last_doc_pk=None, limit=500):
         """Get a batch of
         :param from_db: The DB alias to query
         :param startkey: The filter value to start from e.g. form.received_on
@@ -287,10 +277,6 @@ class FormReindexAccessor(ReindexAccessor):
     def model_class(self):
         return XFormInstanceSQL
 
-    @property
-    def startkey_attribute_name(self):
-        return 'received_on'
-
     def get_doc(self, doc_id):
         try:
             return FormAccessorSQL.get_form(doc_id)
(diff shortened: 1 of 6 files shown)


corehq/form_processor/backends/sql/dbaccessors.py
```diff
@@ -73,14 +76,24 @@ def iter_all_rows(reindex_accessor):
...
-            start_from_for_db = getattr(doc, reindex_accessor.startkey_attribute_name)
-            last_id = doc.id
-            docs = reindex_accessor.get_docs(db_alias, start_from_for_db, last_doc_pk=last_id)
+            last_id = getattr(doc, reindex_accessor.primary_key_field_name)
+            docs = reindex_accessor.get_docs(db_alias, last_doc_pk=last_id)
+
+
+def iter_all_ids(reindex_accessor):
+    for db_alias in reindex_accessor.sql_db_aliases:
+        docs = reindex_accessor.get_doc_ids(db_alias)
```
emord: Should this be `get_docs`?  Looks like you're using both get_doc_ids nad get_docs here

corehq/form_processor/backends/sql/dbaccessors.py
```diff
@@ -234,46 +248,72 @@ def doc_to_json(self, doc):
...
+        if filters:
+            query = query.filter(filters)
+        return query
+
+    def get_doc_ids(self, from_db, last_doc_pk=None, limit=500):
+        """
+        :param from_db: The DB alias to query
+        :param last_doc_pk: The primary key of the last doc from the previous batch
+        :param limit: Desired batch size
+        :return: List of dict(doc_id=<document ID>, primary_key=<doc primary key>)
```
emord: generator instaed of list. also maybe this could be a named tuple?

corehq/form_processor/backends/sql/dbaccessors.py
```diff
@@ -234,46 +248,72 @@ def doc_to_json(self, doc):
...
-        else:
-            from_db = 'default' if from_db is None else from_db
-            sql_query = "SELECT reltuples FROM pg_class WHERE oid = '{}'::regclass"
-            db_cursor = connections[from_db].cursor()
-            with db_cursor as cursor:
-                cursor.execute(sql_query.format(self.model_class._meta.db_table))
-                return int(fetchone_as_namedtuple(cursor).reltuples)
+        query = self.query(from_db, for_count=True)
+        sql, params = query.query.sql_with_params()
+        explain_query = 'EXPLAIN {}'.format(sql)
```
emord: Why does this explain the query?

corehq/form_processor/backends/sql/dbaccessors.py
```diff
@@ -300,38 +340,15 @@ def get_doc(self, doc_id):
...
-
-    @property
-    def count_query(self):
-        # deletion clause left out on purpose since it throws off the count estimate
+    def extra_filters(self, for_count=False):
+        filters = []
+        if not for_count:
+            # don't inlucde in count query since the query planner can't account for it
+            # hack for django: 'state & DELETED = 0' so 'state = state + state & DELETED'
+            filters.append(Q(state=F('state').bitand(XFormInstanceSQL.DELETED) + F('state')))
```
emord: Does this just mean state != DELETED ? in that case would `~Q(state=DELETED)` be more obvious?

corehq/form_processor/backends/sql/dbaccessors.py
snopoke: It does and explain and then parses the row count out of it which is MUCH faster than doing a count query. Since these are only ever used for approximate counts it's ok if they aren't 100% accurate.

To be safe I've renamed the method to `get_approximate_doc_count`

corehq/form_processor/backends/sql/dbaccessors.py
snopoke: Unfortunately not: https://github.com/dimagi/commcare-hq/blob/6ae2006b2a823c4733054de0fc2de17107f26617/corehq/form_processor/models.py#L283-L286

emord: Ah that makes sense. Thanks

emord (approved):
Test failure looks real
```
FAIL: corehq.form_processor.tests.test_reindex_accessors:ShardedCaseReindexAccessorTests.test_get_docs
  File "/mnt/commcare-hq-ro/corehq/form_processor/tests/test_reindex_accessors.py", line 75, in test_get_docs
    self.assertEqual(8, len(docs))
AssertionError: 8 != 12
```

corehq/form_processor/backends/sql/dbaccessors.py
snopoke: the other one should have been `get_doc_ids`

corehq/util/doc_processor/sql.py
```diff
@@ -48,14 +45,10 @@ def resumable_sql_model_iterator(iteration_key, reindex_accessor, chunk_size=100
     is stopped and later resumed.
     """
 
-    def data_function(from_db, filter_value, last_id):
-        return reindex_accessor.get_docs(from_db, filter_value, last_id, limit=chunk_size)
+    def data_function(from_db, last_id):
```
millerdev: Changing this function signature made it not possible to resume iterations that were in progress (ongoing blob db migration) because [`ResumableArgsProvider`](https://github.com/dimagi/commcare-hq/blob/6143fb5b3e09a969b1a50e932550bf2360a2f27b/corehq/util/pagination.py#L131-L132) uses persisted iteration state (args passed to this function) from the old code to resume iteration.

snopoke: Apologies for that, I completely forgot about those migrations. we can revert this if necessary

PR #18631, opened 2017-11-21
ICDS: Add Map Location Name

sheelio:
@emord 
This adds the custom property map_location_name to the location UCRs, tables and corresponding views.


emord (approved):

emord:
test failure looks real

```
ProgrammingError: column awc_location.block_map_location_name does not exist
LINE 20:  awc_location.block_map_location_name,
```

sheelio:
@emord 
I moved generating the views to a nightly task - the issue was that it was using the same SQL file to generate views but this file was called from multiple migrations.  Older migrations didn't yet have the columns required for the newer version of the view generation to work.

emord:
@sheelio you should just remove that line from those migrations and if it's the only line you can make it a noop_migration https://github.com/dimagi/commcare-hq/blob/master/custom/icds_reports/migrations/0021_remove_categories.py

I don't think this should be in the daily generation code because a) that's already got a lot going on in it and b) if we make some change in the view that the code depends on there could be  intermittent 500s when webworkers are using new definitions until the aggregation is run again. and we generally deploy after aggregation has kicked off, so it'd 500 for a full day

emord (changes requested):
see above comment

sheelio:
@emord Updated.

emord (approved):

PR #18632, opened 2017-11-21
Child Beneficiary List export at State, District, Block , Sector levels

kkrampa:
@emord 
One outstanding thing here is to set the limit of rows that can be exported.


emord (approved):

kkrampa:
@emord Sriya agreed to restrict this to the district level but I found that my previous solution had an issue.
 We have only awc_id in child monthly view so we are not able to filter directly via district and supervisor.
I added this line https://github.com/dimagi/commcare-hq/pull/18632/files#diff-129264561592e648289c087ab1350b9bR614 but this can make export pretty inefficient

emord:
@kkrampa can you base this off of https://github.com/dimagi/commcare-hq/pull/18674 ? That should add the necessary ids for this to work

kkrampa:
@emord Updated

emord (approved):

kkrampa:
@emord :white_check_mark:

PR #18634, opened 2017-11-21
add forgotten migrations

emord:
ran make migrations and htese migrations were never created. (thanks @ctsims for bringing it up)

buddy @snopoke


ctsims:
Glad to hear I wasn't crazy to think this was weird.

snopoke (approved):

snopoke:
These tables already exist in our databases so it may be necessary to fake the migrations.

PR #18635, opened 2017-11-21
Domain is nullable?

emord:
@kaapstorm i found this after running makemigrations.

Looks like you made this change in the model #15404 but the migration was never created, so it's possible either this has been erroring or it's no longer needed. What do you think is the correct fix for this?

@snopoke


kaapstorm:
The model should not have `domain` as nullable. `domain` is only set [in the `log_request` decorator][1] and it's always assigned a value. Setting `timestamp` as `auto_now_add` seems right, but instead of including `domain` in the migration it would be better to fix the model. I'm happy to do that in the morning.

  [1]: https://github.com/dimagi/commcare-hq/blob/ae0c538d328eb56f0f300dee1fc16fc5b77e485f/corehq/apps/dhis2/api.py#L24

kaapstorm:
Closing in favour of #18641.

PR #18640, opened 2017-11-22
fixing case exploder status for big numbers

calellowitz:
@orangejenny @nickpell 
https://sentry.io/dimagi/commcarehq/issues/407495462/

There is no need to convert to and from a float, integer division accomplishes the same thing without breaking trying to create an intermediate representation.


proteusvacuum:
💣

PR #18641, opened 2017-11-22
domain property should not be nullable

kaapstorm:
Corrects `JsonApiLog.domain`, and adds a missing migration for defaulting `JsonApiLog.timestamp` to `now()` on add.

@emord cc @snopoke @dannyroberts


corehq/motech/dhis2/migrations/0004_auto_20171122_0827.py
```diff
@@ -0,0 +1,20 @@
...
+    dependencies = [
+        ('dhis2', '0003_jsonapilog_log_level'),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name='jsonapilog',
+            name='timestamp',
+            field=models.DateTimeField(auto_now_add=True),
+        ),
```
snopoke: This migration doesn't match up with the model change

kaapstorm: That's true. Sorry, I should have pointed to #18635 for context. 

I made a couple of changes a while ago (commits a94815d and 7b95974 in PR #15404) that I forgot to generate a migration for. When @emord ran `./manage.py makemigrations` Django wanted to set the `domain` column as nullable, which seemed wrong. So in this PR I am reverting the model's `domain` property to not be nullable (because it is never None), but I am going ahead with the rest of the missing migration (which is just for the timestamp field).

snopoke (approved):

emord:
Thanks for resolving

PR #18643, opened 2017-11-22
Also update patients on import

kaapstorm:
When importing from OpenMRS, update existing patients as well as adding unknown patients.

(Builds on #18642. Only [the last commit](https://github.com/dimagi/commcare-hq/pull/18643/commits/0cf6d22493d20be3dc7ebce22ec32e9b2b8b361a) is unique to this branch.)

@dannyroberts 

(Labelling product/custom because this is MOTECH, and specific to FGH. Please let me know if this should be labelled "product/invisible" instead.)


kaapstorm:
(Dropping "hold off")

dannyroberts (approved):

PR #18648, opened 2017-11-22
_get_pr_numbers to use old GitHub library

kaapstorm:
Reverts half of cb7650f

@nickpell @orangejenny @calellowitz cc @dannyroberts


nickpell:
is this broken?  If not it would be nice to leave as is

PR #18657, opened 2017-11-25
Move Icon Badges add on to Feature Flag

mkangia:
https://manage.dimagi.com/default.asp?265043#BugEvent.1419504

This moves the application add on Icon Badges to a feature flag. This would then be available only for enikshay. @sheelio is going to handle comms to other users if any using it.
Also adds command to find applications/domains using a particular add on or with all add ons enabled

### DIFF
--- a/corehq/apps/app_manager/management/commands/applications_with_add_ons.py
+++ b/corehq/apps/app_manager/management/commands/applications_with_add_ons.py
@@ -7,6 +7,8 @@ from corehq import toggles
 from corehq.apps.app_manager.models import Domain
 from corehq.apps.toggle_ui.utils import find_static_toggle
 from corehq.toggles import NAMESPACE_DOMAIN
+from corehq.apps.app_manager.dbaccessors import get_app_ids_in_domain
+from corehq.apps.app_manager.models import Application
 
 
 class Command(BaseCommand):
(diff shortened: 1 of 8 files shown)


corehq/apps/app_manager/management/commands/applications_with_add_ons.py
```diff
@@ -0,0 +1,54 @@
...
+                for application in domain.full_applications(include_builds=False):
+                    if not application.is_remote_app():
+                        all_add_ons_enabled = toggles.ENABLE_ALL_ADD_ONS.enabled(domain.name)
+                        if add_on_name in application.add_ons or all_add_ons_enabled:
+                            writer.writerow({
+                                'domain': domain.name,
+                                'application_id': application.get_id,
+                                'app_name': application.name,
+                                'all_add_ons_enabled': all_add_ons_enabled,
+                                'status': application.add_ons.get('custom_icon_badges')
```
orangejenny: should be `add_on_name`

orangejenny (changes requested):

corehq/apps/app_manager/management/commands/applications_with_add_ons.py
mkangia: oops, thanks for catching.

orangejenny (approved):
Updates look good, feel free to self-merge once py3 check passes.

mkangia:
@sheelio unsure of what label this should have. 
`product/feature-flag` + `product/all-users-all-environment`?

sheelio:
I would use all users-all-environments (since we're changing behaviour for potentially non feature flagged projects).

PR #18659, opened 2017-11-26
import reduce

nickpell:
https://docs.python.org/2/library/functools.html#functools.reduce


nickpell:

esoergel (approved):

PR #18661, opened 2017-11-26
fix more longs

nickpell:
get rid of more ```long```s, and rename variables named long for easy linting

### DIFF
--- a/corehq/apps/app_manager/views/modules.py
+++ b/corehq/apps/app_manager/views/modules.py
@@ -753,7 +753,7 @@ def edit_module_detail_screens(request, domain, app_id, module_unique_id):
         if case_list_lookup is not None:
             _save_case_list_lookup_params(detail.short, case_list_lookup, lang)
 
-    if int is not None:
+    if long_ is not None:
         detail.long.columns = map(DetailColumn.from_json, long_)
         if tabs is not None:
             detail.long.tabs = map(DetailTab.wrap, tabs)
--- a/corehq/apps/app_manager/views/modules.py
+++ b/corehq/apps/app_manager/views/modules.py
@@ -698,7 +698,7 @@ def edit_module_detail_screens(request, domain, app_id, module_unique_id):
     params = json_request(request.POST)
     detail_type = params.get('type')
     short = params.get('short', None)
-    long = params.get('long', None)
+    long_ = params.get('long', None)
     tabs = params.get('tabs', None)
     filter = params.get('filter', ())
     custom_xml = params.get('custom_xml', None)
@@ -753,8 +753,8 @@ def edit_module_detail_screens(request, domain, app_id, module_unique_id):
         if case_list_lookup is not None:
             _save_case_list_lookup_params(detail.short, case_list_lookup, lang)
 
-    if long is not None:
-        detail.long.columns = map(DetailColumn.from_json, long)
+    if int is not None:
+        detail.long.columns = map(DetailColumn.from_json, long_)
         if tabs is not None:
             detail.long.tabs = map(DetailTab.wrap, tabs)
         if print_template is not None:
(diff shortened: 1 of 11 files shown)


nickpell:

corehq/apps/app_manager/views/modules.py
```diff
@@ -753,8 +753,8 @@ def edit_module_detail_screens(request, domain, app_id, module_unique_id):
         if case_list_lookup is not None:
             _save_case_list_lookup_params(detail.short, case_list_lookup, lang)
 
-    if long is not None:
-        detail.long.columns = map(DetailColumn.from_json, long)
+    if int is not None:
```
esoergel: should this be `long_`?

corehq/apps/sms/handlers/form_session.py
```diff
@@ -166,7 +166,7 @@ def validate_answer(event, text, v):
     # Validate longint
     elif event.datatype == "longint":
         try:
-            long(text)
+            int(text)
```
esoergel: I'm not confident in this one - the datatype explicitly states `longint` for some reason (it's not clear to me why without context).  We might need to examine more closely if we plan to abolish the distinction between `int`s and `long`s.

esoergel (changes requested):
Don't all these files need `from builtins import int` to continue working in python 2?

@gcapalbo you might want to review the `corehq/apps/sms` changes

corehq/apps/sms/handlers/form_session.py
millerdev: might be reasonable to use `(long if six.PY2 else int)` here?

nickpell: I think ```int``` suffices:

python2:
```
>>> int('9223372036854775807')
9223372036854775807
>>> int('9223372036854775808')
9223372036854775808L
```

python3:
```
>>> int('9223372036854775807')
9223372036854775807
>>> int('9223372036854775808')
9223372036854775808
```

That said, I wonder if ```validate_answer``` is correct to succeed as-is for the text input ```'9223372036854775808'``` when ```event.datatype == "int"```. @gcapalbo thoughts?

corehq/apps/app_manager/views/modules.py
nickpell: good catch - fixed

corehq/apps/sms/handlers/form_session.py
nickpell: Although, this would fail in both python 2 and 3:
```
>>> int('1L')
  File "<stdin>", line 1, in <module>
ValueError: invalid literal for int() with base 10: '1L'
```

Whereas in python 2:
```
>>> long('1L')
1L
```

corehq/apps/app_manager/views/modules.py
```diff
@@ -753,8 +753,8 @@ def edit_module_detail_screens(request, domain, app_id, module_unique_id):
         if case_list_lookup is not None:
             _save_case_list_lookup_params(detail.short, case_list_lookup, lang)
 
-    if long is not None:
-        detail.long.columns = map(DetailColumn.from_json, long)
+    if long_ is not None:
+        detail.long.columns = map(DetailColumn.from_json, long_)
```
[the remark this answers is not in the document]
nickpell: ignoring

corehq/apps/sms/handlers/form_session.py
gcapalbo: for the purposes of this PR, it's ok to just use `int()` here. what this is referring to is when people answer sms surveys (so filling out a form) through touchforms, and it's doing a first level of validation on their input.

but there are a couple things here that i'll clean up - it seems that the form builder has actually done away with the long datatype, and also as Nick pointed out the behavior of int() is not actually restricting to 32 bits as the [integer question type will](https://confluence.dimagi.com/display/commcarepublic/Form+Builder#FormBuilder-3.NumericQuestions), so the integer type validation has been wrong.

i've got a note to myself to clean those parts up, but in the meantime this is fine as is since it will behave the same as before

gcapalbo (approved):

PR #18663, opened 2017-11-26
use __nonzero__ = __bool__

nickpell:
avoids triggering the lib2to3 fixer


nickpell:

esoergel (approved):

PR #18672, opened 2017-11-27
resolve inconsistencies in how subscription properties are set

nickpell:
Recommend 🐡


calellowitz (approved):
you can fix the lint error if you want, otherwise feel free to merge

PR #18674, opened 2017-11-27
Added parent location ids to child health view.

emord:
@sheelio was there any reason to exclude these before. Guessing it was just because we didn't need them at the time. We'll want them now for this https://github.com/dimagi/commcare-hq/pull/18632

buddy @snopoke


sheelio:
@emord Just not needed so we didn't include them.

PR #18655, opened 2017-11-24
Rename kissmetrics module to kissmetrix

orangejenny:
Same idea as https://github.com/dimagi/commcare-hq/pull/18650 - for our RequireJS setup, module name needs to match filename.

https://github.com/dimagi/commcarehq-prelogin/pull/148 needs to be deployed at the same time as this.

@calellowitz / @nickpell / @biyeun 

Apropos of nothing, this is my 1000th HQ PR. Whee.

PR #18679, opened 2017-11-27
update prelogin

orangejenny:
@nickpell


orangejenny:
Merging - there's no test coverage for prelogin javascript, and this needs to go out in the same deploy as https://github.com/dimagi/commcare-hq/pull/18655

PR #18680, opened 2017-11-27
don't count multiple 99dots adherence cases on same day as conflicting

nickpell:
https://trello.com/c/ggQfamHZ/267-historical-adherence-calendar-still-shows-duplicates

@proteusvacuum cc: @millerdev


proteusvacuum:
I actually think we just never want to show the duplicate icon. @ctsims can confirm, but I'm pretty sure this is what we do on mobile.

nickpell:
@proteusvacuum updated

proteusvacuum (approved):

PR #18683, opened 2017-11-27
Don't try to display create form for users who can't create

esoergel:
https://manage.dimagi.com/default.asp?265677
@gcapalbo


gcapalbo (approved):
@esoergel just needs a label

esoergel:
Doh, thanks G

PR #18686, opened 2017-11-28
pass right number of args

nickpell:
@proteusvacuum


proteusvacuum:
Fixed here, https://github.com/dimagi/commcare-hq/pull/18682. Can we merge that one instead?

PR #18691, opened 2017-11-28
Avoid adding None returned in case of owner not found

mkangia:
https://sentry.io/dimagi/commcarehq/issues/317554853/

https://manage.dimagi.com/default.asp?266171

@sheelio 
There was an issue if the filter has an id for something to filter on but that is been removed. In this case the filter was set for a location which was now removed.

### DIFF
--- a/corehq/apps/export/forms.py
+++ b/corehq/apps/export/forms.py
@@ -647,11 +647,10 @@ class DashboardFeedFilterForm(forms.Form):
 
             emwf_utils_class = CaseListFilterUtils if export_type is CaseExportInstance else EmwfUtils
             emwf_data = []
-            for x in selected_items:
-                choice_tuple = emwf_utils_class(domain).id_to_choice_tuple(str(x))
+            for item in selected_items:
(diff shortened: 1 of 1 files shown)


corehq/apps/export/forms.py
```diff
@@ -646,7 +646,11 @@ def get_form_data_from_export_instance_filters(cls, export_instance_filters, dom
                 )
 
             emwf_utils_class = CaseListFilterUtils if export_type is CaseExportInstance else EmwfUtils
-            emwf_data = [emwf_utils_class(domain).id_to_choice_tuple(str(x)) for x in selected_items]
+            emwf_data = []
+            for x in selected_items:
```
proteusvacuum: nit: `x` is not a good name

proteusvacuum (approved):

corehq/apps/export/forms.py
mkangia: absolutely. 
Thanks for the quick review. Updating.

proteusvacuum: 👍

PR #18692, opened 2017-11-28
ICDS: Change indicator scheduling

gcapalbo:
Switches the monthly indicators from running on the first Monday of the month to running on the first of the month,

@millerdev @emord

### DIFF
--- a/custom/icds/tasks.py
+++ b/custom/icds/tasks.py
@@ -82,9 +82,8 @@ def get_supervisor_location_ids(domain):
     return get_location_ids_with_location_type(domain, SUPERVISOR_LOCATION_TYPE_CODE)
 
 
-def is_first_week_of_month():
-    day = ServerTime(datetime.utcnow()).user_time(pytz.timezone('Asia/Kolkata')).done().day
-    return day >= 1 and day <= 7
+def get_current_date():
+    return ServerTime(datetime.utcnow()).user_time(pytz.timezone('Asia/Kolkata')).done().date()
 
 
 def get_user_ids_under_location(domain, site_code):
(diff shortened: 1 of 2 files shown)


custom/icds/tasks.py
```diff
@@ -106,16 +105,24 @@ def get_language_code(user_id, telugu_user_ids, marathi_user_ids):
...
+    run_every=crontab(hour=9, minute=0),
     queue=settings.CELERY_PERIODIC_QUEUE,
     ignore_result=True
 )
-def run_weekly_indicators(phased_rollout=True):
+def run_user_indicators(phased_rollout=True):
     """
-    Runs the weekly SMS indicators Monday at 9am IST.
-    If it's the first week of the month, also run the monthly indicators.
+    Runs the weekly / monthly user SMS indicators at 9am IST.
```
millerdev: line 108 says `hour=9`, and I believe all of our servers have their timezone set to UTC, so I think this should say "3:30am IST"

custom/icds/tasks.py
```diff
@@ -106,16 +105,24 @@ def get_language_code(user_id, telugu_user_ids, marathi_user_ids):
...
+    This task is run every day and the following logic is applied:
+        - if it's Monday, the weekly indicators are sent
+        - if it's the first of the month, the monthly indicators are sent
```
emord: Should this be true even if the first of the month is Sunday? For some reason I thought that the gov didn't like doing that kind of thing on sundays, but I could just be misremembering

custom/icds/tasks.py
millerdev: Sorry, I had my timezones switched. This should say "2:30pm IST"

emord (approved):

custom/icds/tasks.py
emord: Celery workers run on IST for enikshay and icds so that downtime tasks coorespond more appropriately with midnight. There's a CELERY_TIMEZONE setting

custom/icds/tasks.py
gcapalbo: I'll doublecheck with the project team about this point, but the new requirement is to send on the first of the month

millerdev (approved):

custom/icds/tasks.py
gcapalbo: actually, we're already sending other messages on Sundays. I'll still doublecheck whether we should be sending on Sundays or not but if we need to impose that restriction i think we'd just do so in the SMS settings for the project so that all messages would just queue up on Sundays and then get sent on Monday

PR #18693, opened 2017-11-28
python-modernize . -w -f libmodernize.fixes.fix_xrange_six

nickpell:
@dimagi/py3

### DIFF
--- a/corehq/apps/case_importer/const.py
+++ b/corehq/apps/case_importer/const.py
@@ -1,9 +1,10 @@
 from __future__ import absolute_import
 from django.utils.translation import ugettext_noop
+from six.moves import range
 
 
 class LookupErrors:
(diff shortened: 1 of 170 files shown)


corehq/apps/case_importer/const.py
```diff
@@ -1,9 +1,10 @@
 from __future__ import absolute_import
 from django.utils.translation import ugettext_noop
+from six.moves import range
 
 
 class LookupErrors:
-    NotFound, MultipleResults = range(2)
+    NotFound, MultipleResults = list(range(2))
```
kaapstorm: We could get nitpicky about this specific instance, but I think nitpicking these would be a terrible idea.

kaapstorm (approved):

PR #18694, opened 2017-11-28
gitignore .swo files

orangejenny:
@calellowitz


mkangia (approved):
checked with Jenny that swo is one of vim's temp file extensions so should be safe to add.

PR #18703, opened 2017-11-28
Run Kissmetrics callbacks even if analytics isn't enabled

orangejenny:
Similar to https://github.com/dimagi/commcare-hq/pull/18688, need to run analytics callbacks (which contain HQ code) even if analytics aren't enabled.

Removed a couple of redundant `isEnabled` checks for the sake of cleaner code. This does mean that actions will get logged even if analytics aren't enabled. At the moment, I'm more worried about the callback not getting called than about extra logging.

@calellowitz / @biyeun

### DIFF
--- a/corehq/apps/analytics/static/analytix/js/kissmetrix.js
+++ b/corehq/apps/analytics/static/analytix/js/kissmetrix.js
@@ -33,6 +33,8 @@ hqDefine('analytix/js/kissmetrix', function () {
             if (properties) data.km_property = properties;
             window.dataLayer.push(data);
             logger.verbose.log(command, ['window._kmq.push', 'window.dataLayer.push', '_kmqPushCommand', commandName]);
+        } else if (callbackFn) {
+            callbackFn();
         }
     };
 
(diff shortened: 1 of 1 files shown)


millerdev (approved):

mkangia (approved):

emord:
> need to run analytics callbacks (which contain HQ code) even if analytics aren't enabled.

Is there a more comprehensive plan to separate this? We'll have to continue keeping up environments without analytics, so it seems like this is an unfortunate dependcy to have

esoergel:
> Is there a more comprehensive plan to separate this? We'll have to continue keeping up environments without analytics, so it seems like this is an unfortunate dependcy to have

I haven't been involved with the analytics code, so forgive my ignorance, but why is the primary action of buttons passed to analytics code and called there?  Shouldn't it be the other way around?  What happens if the third party is slow or down, or not enabled for the environment, or the client blocks third party actions?

orangejenny:
@esoergel 

> why is the primary action of buttons passed to analytics code and called there?

The primary action may be a click away from the page, while the analytics event is an ajax request to a third-party site. Clicking away from the page will cancel pending ajax requests, which is why the main action ends up being dependent on the analytics call completing. [createSafeCallback](https://github.com/dimagi/commcare-hq/blob/master/corehq/apps/analytics/static/analytix/js/utils.js#L102) wraps the callback in code to get it executed even if there's some other failure.

orangejenny:
no test coverage and needs to get in for deploy => merging

PR #18649, opened 2017-11-22
Downgrade Information Pop Up 

orangejenny:
https://manage.dimagi.com/default.asp?265868

@sheelio 

HQ
screen shot 2017-11-22 at 2 41 53 pm

Email
screen shot 2017-11-22 at 2 43 06 pm

@calellowitz / @nickpell

PR #18704, opened 2017-11-28
bcc growth@ on downgrade emails

nickpell:
https://manage.dimagi.com/default.asp?264241

should merge with https://github.com/dimagi/commcarehq-ansible/pull/1079


settings.py
```diff
@@ -490,6 +490,7 @@
 INTERNAL_SUBSCRIPTION_CHANGE_EMAIL = 'accounts+subchange+internal@example.com'
 BILLING_EMAIL = 'billing-comm@example.com'
 INVOICING_CONTACT_EMAIL = 'billing-support@example.com'
+GROWTH_EMAIL = 'growth@example.com'
```
nickpell: @orangejenny re: https://github.com/dimagi/commcare-hq/pull/18649#issuecomment-347277850

I'm adding the growth email here

orangejenny: great

orangejenny (approved):

kaapstorm (approved):

settings.py
emord: Any reason these aren't empty strings? (not asking you to fix all of these, just thinking about this one for now)

If the default is `''` then we can either check to see if django's email engine will just ignore it, or only bcc if it exists. Then we don't have to worry about setting this for enterprise environments that "growth" (or these other emails) will never care about

nickpell: @emord Seems like a reasonable approach to me

PR #18705, opened 2017-11-28
Je/server env

emord:
@dimagi/scale-team 

Celery wasn't running due to:

/bin/bash: /home/cchq/www/icds/current/services/supervisor/icds_celery_bash_runner.sh: No such file or directory

### DIFF
--- a/corehq/apps/hqadmin/management/commands/make_supervisor_conf.py
+++ b/corehq/apps/hqadmin/management/commands/make_supervisor_conf.py
@@ -46,7 +46,7 @@ class SupervisorConfCommand(BaseCommand):
         self.params = options['params'] or {}
         if self.params:
             self.params = self.extend_params(json.loads(self.params))
-        self.environment = options.get('environment', settings.SERVER_ENVIRONMENT)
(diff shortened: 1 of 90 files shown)


emord:
feels like we might be creating more work for ourselves. maybe we should just change it over to icds_new or something?

snopoke:
What do you mean about changing it over?

emord:
Err. I mean in some places we have icds, and some places it's icds-new/icds_new because of https://github.com/dimagi/commcarehq-ansible/blob/master/ansible/vars/icds-new/icds-new_public.yml#L4

Seems like it might have just been easier to gone with icds-new everywhere, even if it meant re-deploying stack

emord:
deployed to cids-new and verified celery is now running

snopoke:
Oh right, should definitely all be consistant

calellowitz:
I think we can revert this and then just delete the line you are referring to. ENV would still be icds-new/icds_new while ansible and fab deploying but internally and all over the code it would just be icds. Then we would not have to change any of the code, or the UCRs, and it would be easy to switch the new environment to just being called icds when we decommission the old one.
@snopoke @emord Does that idea make sense to you
cc: @dimagi/scale-team

PR #18706, opened 2017-11-28
Conditonal Alerts UI: Step 1

gcapalbo:
spec: https://docs.google.com/document/d/1iJq6EHEYTppIx0p2UMsIr8q3qKk_bFAAfvojtMsxvH4/edit#

So far this releases basic navigation and integrates the rule criteria part of the "create" page. Screenshot:

conditional-alert-1

@millerdev

### DIFF
--- a/corehq/messaging/scheduling/views.py
+++ b/corehq/messaging/scheduling/views.py
@@ -5,6 +5,7 @@ from django.db import transaction
 from django.http import (
     Http404,
     HttpResponseRedirect,
+    HttpResponseBadRequest,
 )
 from django.urls import reverse
 from django.utils.decorators import method_decorator
--- a/corehq/messaging/scheduling/views.py
+++ b/corehq/messaging/scheduling/views.py
@@ -13,6 +13,7 @@ from django.utils.translation import ugettext_lazy as _
 
 from corehq import privileges
 from corehq.apps.accounting.decorators import requires_privilege_with_fallback
+from corehq.apps.data_interfaces.models import AutomaticUpdateRule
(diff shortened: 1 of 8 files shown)


corehq/messaging/scheduling/views.py
```diff
@@ -492,3 +494,119 @@ def schedule_form(self):
...
+        basic_info_form_valid = self.basic_info_form.is_valid()
+        criteria_form_valid = self.criteria_form.is_valid()
+
+        if self.read_only_mode:
+            # Don't allow making changes to rules that have custom
+            # criteria/actions unless the user has permission to
+            return HttpResponseBadRequest()
+
+        if basic_info_form_valid and criteria_form_valid:
+            if not self.is_system_admin and self.criteria_form.requires_system_admin_to_save:
```
millerdev: nit: could this be changed to `if self.read_only_mode:`?

millerdev (approved):

corehq/messaging/scheduling/views.py
gcapalbo: the logic in `self.read_only_mode` is slightly different - it's checking for `requires_system_admin_to_edit`, which inspects the existing rule (only relevant when editing) and decides whether it should be shown as read only in the UI.

the check that's happening here is for `requires_system_admin_to_save`, and it's checking right before save if there are options which have been enabled which require a superuser. we shouldn't ever hit this because the UI restricts it anyway, but since anything can be posted it's needed.

millerdev: Ah, I missed that subtle difference. Thanks for the clarification.

PR #18711, opened 2017-11-29
Allow clearing date spans in reports

orangejenny:
https://manage.dimagi.com/default.asp?266238

Took a suggestion from the [daterangepicker docs](http://www.daterangepicker.com/) and replaced the "Cancel" button with a "Clear" button.

@dimagi/product What do y'all think? It does get rid of the ability to cancel, but it saves adding another button.

Before
screen shot 2017-11-28 at 10 50 26 pm

After
screen shot 2017-11-28 at 10 49 52 pm

### DIFF
--- a/corehq/apps/hqwebapp/static/hqwebapp/js/daterangepicker.config.js
+++ b/corehq/apps/hqwebapp/static/hqwebapp/js/daterangepicker.config.js
@@ -49,7 +49,7 @@
 
         var $el = $(this);
         $el.daterangepicker(config);
-        $el.on('cancel.daterangepicker', function(ev, picker) {
+        $el.on('cancel.daterangepicker', function() {
             $el.val('');
         });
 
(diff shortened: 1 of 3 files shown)


orangejenny:
code buddy @calellowitz

calellowitz (approved):
looks good to me pending product approval

sheelio:
This looks good - to cancel, can i just click outside the datepicker modal?

orangejenny:
@sheelio Not exactly. If you click outside the widget, it closes, but the value doesn't revert.

Suppose the value is set to the last 7 days. You open the widget and change it to the last 30 days. If you click out of the widget, it keeps the 30 days value. If you clicked cancel, it would revert back to being the last 7 days. So this PR would get eliminate the ability to truly cancel.

sheelio:
That seems fine to me (most date pickers I looked at are not cancellable in that way).

orangejenny:
@sheelio Cool, thanks for the feedback.

PR #18723, opened 2017-11-29
python-modernize . -w -f libmodernize.fixes.fix_map

nickpell:
@dimagi/py3

### DIFF
--- a/corehq/apps/api/resources/__init__.py
+++ b/corehq/apps/api/resources/__init__.py
@@ -73,7 +73,7 @@ class CorsResourceMixin(object):
             allowed = []
 
         request_method = request.method.lower()
-        allows = ','.join(map(lambda x: x.upper() if x else '', allowed))
+        allows = ','.join([x.upper() if x else '' for x in allowed])
 
         if request_method == 'options':
             response = HttpResponse(allows)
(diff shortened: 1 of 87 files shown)


corehq/apps/api/resources/__init__.py
```diff
@@ -73,7 +73,7 @@ def method_check(self, request, allowed=None):
             allowed = []
 
         request_method = request.method.lower()
-        allows = ','.join(map(lambda x: x.upper() if x else '', allowed))
+        allows = ','.join([x.upper() if x else '' for x in allowed])
```
kaapstorm: Approving, because this is correct in that it preserves existing behaviour, but the original code doesn't do what it should. `allows` [should end up](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Allow) with a value like `"GET,POST"` but this code could result in a value like `",,,GET,,POST"`. It should rather be
```python
allows = ','.join([x.upper() for x in allowed if x])
```

kaapstorm (approved):

corehq/apps/api/resources/__init__.py
esoergel: Also, you can drop the square brackets and just pass a generator expression to `join`

PR #18726, opened 2017-11-29
Throw 404 instead of 500 when saved export is missing

orangejenny:
https://manage.dimagi.com/default.asp?266263

@calellowitz / @mkangia

### DIFF
--- a/corehq/apps/export/views.py
+++ b/corehq/apps/export/views.py
@@ -2489,7 +2489,11 @@ def can_download_daily_saved_export(export, domain, couch_user):
 @require_GET
 def download_daily_saved_export(req, domain, export_instance_id):
     with CriticalSection(['export-last-accessed-{}'.format(export_instance_id)]):
-        export_instance = get_properly_wrapped_export_instance(export_instance_id)
+        try:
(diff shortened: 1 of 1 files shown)


corehq/apps/export/views.py
```diff
@@ -2489,7 +2489,11 @@ def can_download_daily_saved_export(export, domain, couch_user):
 @require_GET
 def download_daily_saved_export(req, domain, export_instance_id):
     with CriticalSection(['export-last-accessed-{}'.format(export_instance_id)]):
-        export_instance = get_properly_wrapped_export_instance(export_instance_id)
+        try:
+            export_instance = get_properly_wrapped_export_instance(export_instance_id)
+        except ResourceNotFound:
+            raise Http404
```
mkangia: nit: Can we update this to `Http404(_(u"Export not found"))` so its clear that the url is fine but we are missing the resource?

mkangia (approved):

corehq/apps/export/views.py
orangejenny: sure - 0d6bc95

mkangia: Thank you.

PR #18727, opened 2017-11-29
Make daily saved export url consistent

orangejenny:
Minor, but I got briefly tripped up by `daily_saved` vs `dailysaved` while doing https://github.com/dimagi/commcare-hq/pull/18726

Calling this all-users-all-enviroments is kind of pedantic - the only way a user would be affected is if they have a daily saved export bookmarked, and making them update that URL seems minor enough.

@calellowitz / @mkangia

### DIFF
--- a/corehq/apps/export/urls.py
+++ b/corehq/apps/export/urls.py
@@ -117,7 +117,7 @@ urlpatterns = [
     url(r"^custom/new/case/download/(?P<export_id>[\w\-]+)/$",
         DownloadNewCaseExportView.as_view(),
         name=DownloadNewCaseExportView.urlname),
-    url(r"^custom/dailysaved/download/(?P<export_instance_id>[\w\-]+)/$",
+    url(r"^custom/daily_saved/download/(?P<export_instance_id>[\w\-]+)/$",
         download_daily_saved_export,
         name="download_daily_saved_export"),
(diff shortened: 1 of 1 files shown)


sheelio:
Hey Jenny - there may be users who do have this bookmarked. Would it do a re-direct?  
(Is there another place we use daily_saved?)

orangejenny:
It wouldn't. If you're concerned about people losing their bookmarks, I can just close this - it's a minor tech cleanup, not worth adding code to redirect.

sheelio:
Ya, unless there's a good reason for it, would want to avoid the change (people react strongly if something stops working around their data access :| )

orangejenny:
@sheelio yeah, makes sense, glad for the feedback

corehq/apps/export/urls.py
```diff
@@ -169,9 +169,6 @@
         DailySavedExportPaywall.as_view(),
         name=DailySavedExportPaywall.urlname),
 
-    url(r"^custom/dailysaved/download/(?P<export_instance_id>[\w\-]+)/$",
-        download_daily_saved_export,
-        name="download_daily_saved_export"),
```
mkangia: Looks like the bookmark/saved links was the reason someone ended up adding two links for the same end point. :|

PR #18728, opened 2017-11-30
Require user field properties to start with letter

orangejenny:
https://manage.dimagi.com/default.asp?265671

Just enforce the same validation as for case properties, except that case properties also have to support slashes (`parent/name`).

@calellowitz / @mkangia

### DIFF
--- a/corehq/apps/custom_data_fields/edit_model.py
+++ b/corehq/apps/custom_data_fields/edit_model.py
@@ -3,7 +3,7 @@ import json
 
 from django.contrib import messages
 from django.core.exceptions import ValidationError
-from django.core.validators import validate_slug
+from django.core.validators import RegexValidator, validate_slug
 from django.utils.translation import ugettext as _, ugettext_lazy
 from django import forms
 from corehq.apps.hqwebapp.decorators import use_jquery_ui
@@ -57,6 +57,7 @@ class XmlSlugField(forms.SlugField):
     default_validators = [
         validate_slug,
         validate_reserved_words,
+        RegexValidator(r'^[a-zA-Z]', ''),
     ]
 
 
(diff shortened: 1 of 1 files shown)


corehq/apps/custom_data_fields/edit_model.py
```diff
@@ -57,6 +57,7 @@ class XmlSlugField(forms.SlugField):
     default_validators = [
         validate_slug,
         validate_reserved_words,
+        RegexValidator(r'^[a-zA-Z]', ''),
```
mkangia: Do we want to allow underscores, space? 
Looking at the description [here](https://manage.dimagi.com/default.asp?265671#BugEvent.1427164)
> Unable to use numbers **only** as a user property name.

Should it be okay to check if there is any non-integer character?

mkangia (changes requested):

corehq/apps/custom_data_fields/edit_model.py
orangejenny: I'm deliberately making the validation stricter than the ticket requested. A key like `23andme` wouldn't trigger the error in the ticket, that is true. But in general, when we're accepting user input and going to use it xml, we're better off programming defensively and requiring that input to be a valid xml identifier.

This particular line is only validating the first character, so it definitely shouldn't be a space. I think underscore as first character is technically fine, I'm disallowing it here just to make this validation consistent with case property validation (one less set of validation rules for users to remember).

mkangia: Sounds good. Just one thing.
Looks like no one would have been able to use a number there since that resulted in an error, an if someone used a character like underscore they would now start to receive an error message?

orangejenny: They'll receive an error message the next time they go to the user fields page and try to make an edit. I'm inclined to say this is okay and wait and see if we get any tickets about it. 

@dimagi/product also open to your feedback on this from

sheelio: @orangejenny , I'm not quite clear what the change is.  Are we now blocking underscores in the user property name?  (That's a pretty common workflow I think for user property names).

orangejenny: The change is to require properties to start with a letter. They can still contain underscores and numbers and dashes).

sheelio: Ah ok - this is fine then.

orangejenny:
@mkangia given Sheel's comments, this is good to go, isn't it?

mkangia (approved):

PR #18729, opened 2017-11-30
Lookup table upload: row of zeroes shouldn't end the upload

orangejenny:
https://manage.dimagi.com/default.asp?265975

@calellowitz / @mkangia

### DIFF
--- a/corehq/util/workbook_json/excel.py
+++ b/corehq/util/workbook_json/excel.py
@@ -179,7 +179,7 @@ class WorksheetJSONReader(IteratorJSONReader):
                     _convert_float(cell.value)
                     for cell in row[:width]
                 ]
-                if not any(cell_values):
+                if not any(cell != '' for cell in cell_values):
                     break
                 yield cell_values
         super(WorksheetJSONReader, self).__init__(iterator())


corehq/util/workbook_json/excel.py
```diff
@@ -179,7 +179,7 @@ def _convert_float(value):
                     _convert_float(cell.value)
                     for cell in row[:width]
                 ]
-                if not any(cell_values):
+                if not any(cell != '' for cell in cell_values):
```
mkangia: :nit
I could not really understand `# Specifically check for None so that we can allow a value of 0` above.
probably if we just return value even in case its `None` you would not need to modify this.

mkangia (approved):

corehq/util/workbook_json/excel.py
orangejenny: Not sure I'm following. I think the `Specifically check for None` comment is explaining why the return statement isn't `return value if value else ''`. The `is not None` allows `_convert_float` to always return a value (never `None`). I'm not sure why it's avoiding returning `None`...but regardless, changing that return statement to `return None` wouldn't fix the reported bug. A row full of zeroes would still result in `cell_values` being full of zeroes and `any(cell_values)` incorrectly evaluating to false.

mkangia: Thanks for those specifics Jenny.
I totally missed about 0. 

So to sum what i understand,
that method is iterating rows and then parsing the values in them to int if they can be
and breaks at a point where it finds a row with no value to parse i.e all values are empty string.

Its probably out of the scope of this PR so feel free to merge this as-is.
But if we can check if there is a better way to check for the number of rows to be parsed and then break out when done, like there is max_rows or `:height, we can actually just get rid of this check and make it more reliable.

mkangia (approved):

corehq/util/workbook_json/excel.py
orangejenny: Yep, that's what's happening. I agree that having a max_rows-type value would be more reliable that iterating until there's a blank row, but don't know enough about `openpyxl` to know if such a thing exists.

PR #18730, opened 2017-11-30
reset mobile worker passwords in a domain

sravfeyn:


snopoke (approved):

corehq/apps/domain/management/commands/reset_user_passwords.py
```diff
@@ -0,0 +1,56 @@
...
+        django_users = User.objects.filter(
+            username__endswith="@{domain}.commcarehq.org".format(domain=domain_name)
+        )
+        assert len(all_users) == django_users.count(), "Something is wrong, django/couch users don't match"
+
+        num_users = len(all_users)
+        for i, users in enumerate(chunked(all_users, 1000)):
+            for user in users:
+                user.set_password(password)
+            CommCareUser.bulk_save(users)
```
calellowitz: does this work with all the code we have to update both couch and django users?

corehq/apps/domain/management/commands/reset_user_passwords.py
```diff
@@ -0,0 +1,56 @@
...
+            print("Updated {count}/{total}".format(count=(i+1)*1000, total=num_users))
+
+        for i, users in enumerate(chunked(django_users, 1000)):
+            for user in users:
+                user.set_password(password)
```
calellowitz: it looks like set_password creates a new password hash. could this be different on the couch and django user and if so does that matter? i imagine there is some sort of random salt added

calellowitz (changes requested):
few questions about couch and django users but otherwise looks fine

corehq/apps/domain/management/commands/reset_user_passwords.py
sravfeyn: It's the same, couch `set_password` actually pulls from Django

calellowitz: right but that function generates a random salt each time, so if we use it once on the couch user and once on the django user the hashed strings will be different. unless im missing something?

corehq/apps/domain/management/commands/reset_user_passwords.py
sravfeyn: Yea, there isn't extra things that needs to happen on password change. There are couch/django syncs, but that's taken care of by doing it separately. There are post-save signals that update ES, but it's not needed for password change.

corehq/apps/domain/management/commands/reset_user_passwords.py
sravfeyn: Ha, that's a good point! The `check_password` still works based on raw-password, I also tested this locally on a domain and trying to login to make sure everything works.

Alternatively, we could generate the hash once for all users and use the same for all users.

calellowitz: if it still works and its only for testing purposes here that is good enough for me. just wanted to make sure we were testing since it seemed like that could have been a problem

calellowitz (approved):

corehq/apps/domain/management/commands/reset_user_passwords.py
sravfeyn: Cool, I am not favorite of this either, but for throwaway test data, it seems okay.

sravfeyn:
I am checking with Lilianna to see if any custom user data needs to be deidentified.

PR #18736, opened 2017-11-30
Restore proper search icon size in web apps

orangejenny:
https://manage.dimagi.com/default.asp?266389

Fixes regression where search icon in web apps was appearing too large.

Directly caused by https://github.com/dimagi/commcare-hq/pull/18709/commits/ae21524652623d3dc715004464b7a843afa0d22a which turned out not to be a glyphicons-specific workaround. But it also turns out that there's a specific class being applied to this icon to make it 4 times bigger than normal, so removing that sets the icon back to normal size.

@calellowitz


esoergel (approved):

millerdev (approved):

orangejenny:
HTML only / no test coverage => merging

PR #18741, opened 2017-11-30
eNikshay: find bad episode structure

proteusvacuum:
https://manage.dimagi.com/default.asp?265996

Private sector has had a few bad case structures due to people not syncing. This will get us a list of all the places there are multiple open episodes for a single occurrence. 

@emord

### DIFF
--- a/custom/enikshay/case_utils.py
+++ b/custom/enikshay/case_utils.py
@@ -585,9 +585,13 @@ def get_all_episode_ids(domain):
     return case_ids
 
 
-def iter_all_active_person_episode_cases(domain, case_ids):
+def iter_all_active_person_episode_cases(domain, case_ids, sector="both"):
     """From a list of case_ids, return all the active episodes and associate person case
     """
+    accepted_sectors = ['both', 'private', 'public']
+    if sector not in accepted_sectors:
+        raise ValueError('sector argument should be one of {}'.format(accepted_sectors))
+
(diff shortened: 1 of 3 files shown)


custom/enikshay/case_utils.py
```diff
@@ -585,9 +585,13 @@ def get_all_episode_ids(domain):
     return case_ids
 
 
-def iter_all_active_person_episode_cases(domain, case_ids):
+def iter_all_active_person_episode_cases(domain, case_ids, sector="both"):
     """From a list of case_ids, return all the active episodes and associate person case
     """
+    accepted_sectors = ['both', 'private', 'public']
```
emord: I usually try to use a tuple or a constant with something like this so that it can't accidentally be modified later

custom/enikshay/case_utils.py
```diff
@@ -597,6 +601,11 @@ def iter_all_active_person_episode_cases(domain, case_ids):
         if episode_case.closed:
             continue
 
+        if sector == 'private' and episode_case.get_case_property('enrolled_in_private') != 'true':
```
emord: Case properties and values may also be good candidates to be a constant. Is there a place in the code that lists assumptions about app/case structure like this?

emord (approved):
Looks good just a couple of nits

emord (approved):

PR #18743, opened 2017-12-01
Standardize core js includes in icds tableau dashboard

orangejenny:
Introduced in https://github.com/dimagi/commcare-hq/pull/18116, followup for https://github.com/dimagi/commcare-hq/pull/18740

@calellowitz / @emord

### DIFF
--- a/custom/icds_reports/templates/icds_reports/tableau.html
+++ b/custom/icds_reports/templates/icds_reports/tableau.html
@@ -61,9 +61,9 @@
     {% include "hqwebapp/includes/core_libraries.html" %}
     <script type="text/javascript" src="{% static 'bootstrap/dist/js/bootstrap.min.js' %}"></script>
     <script type="text/javascript" src="{% static 'tableau.min.js' %}"></script>
-    {% include "hqwebapp/includes/ko.html" %}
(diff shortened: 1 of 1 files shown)


orangejenny:
The actual bug here is that hq.helpers.js depends on underscore and therefore needs to be included after underscore.

This used to work because hq.helpers used to use the globally defined `_`, and the logic that actually uses `_` is inside a document ready handler, so by the time it ran, the global had been defined.

Now the `_` used in hq.helpers is a local variable that ultimately comes from the global `_`, but it's defined at the time the module's own `hqDefine` runs...so if underscore hasn't yet been included when hq.helpers is defined, that local `_` gets set to undefined.

orangejenny:
@emord Libraries are now re-ordered.

custom/icds_reports/templates/icds_reports/tableau.html
```diff
@@ -57,14 +57,15 @@
             padding-bottom: 21px;
         }
     </style>
-    <script type="text/javascript" src="{% static 'jquery/dist/jquery.min.js' %}"></script>
+    {% include "hqwebapp/includes/jquery.html" %}
```
emord: Seems weird that this wouldn't be in core_libraries

emord (approved):

PR #18745, opened 2017-12-01
Better standardize "core" JavaScript inclusions

orangejenny:
Opening for review.

The general thought is to enforce dependencies among the core libraries in core_libraries.html  and then expect others templates to rely on this one instead of including individual script tags.

We're especially susceptible to problems in the the handful of pages that don't inherit from `hqwebapp/base.html` - see https://github.com/dimagi/commcare-hq/pull/18743

@millerdev 
code buddy @calellowitz


emord (approved):
Didn't review too closely but am pro this idea

millerdev (approved):

orangejenny:
Fixed some bugs, wrote some tests. Good to go now.

millerdev (approved):

PR #18746, opened 2017-12-01
warn when creating users with username of deleted users

snopoke:
https://manage.dimagi.com/default.asp?265963#1448795

@dimagi/product 
Updated this to display a warning message if the username belongs to a deleted user:

image

### DIFF
--- a/corehq/apps/users/dbaccessors/all_commcare_users.py
+++ b/corehq/apps/users/dbaccessors/all_commcare_users.py
@@ -1,10 +1,16 @@
 from __future__ import absolute_import
+
+from collections import namedtuple
+
 from corehq.apps.users.models import CommCareUser
 from corehq.apps.es import UserES
 from corehq.util.quickcache import quickcache
 from corehq.util.test_utils import unit_testing_only
 from dimagi.utils.couch.database import iter_docs, iter_bulk_delete
 from six.moves import map
+
+
+UserExists = namedtuple('UserExists', 'exists is_deleted')
 
 
 def get_all_commcare_users_by_domain(domain):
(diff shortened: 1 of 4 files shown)


czue:
My impression is that deleting and recreating users is quite a common training workflow used to "clear" data. Could be worth getting some inputs from GS before proceeding with this change (or shifting to the warning approach you mentioned).

sheelio:
@snopoke - do we know what the downsides are? This is a pretty common workflow for GS folks (but getting less common with demo mode mobile workers).  I would put in a warning if anything.

snopoke:
The original bug this came from is quite an edge case and probably better solved in other ways: https://manage.dimagi.com/default.asp?265963

snopoke:
The original bug this came from is quite an edge case and probably better solved in other ways: https://manage.dimagi.com/default.asp?265963

corehq/apps/users/dbaccessors/all_commcare_users.py
```diff
@@ -182,6 +188,30 @@ def get_deleted_user_by_username(cls, username):
...
+    :return: namedtuple(exists:bool, is_deleted:bool)
+    """
+    result = CommCareUser.get_db().view(
+        'users/by_username',
+        key=username,
+        include_docs=False,
+        reduce=False,
+    )
+    if result:
+        return UserExists(True, False)
```
mkangia: When someone adds a new user with the same username that has been deleted, it would have two couch docs with the same name but only one django user?

This returns `False` without checking if that username is present for a deleted user or not. Guess thats because if a user exists with that username checking for a deleted user is irrelevant. But in case someone adds a username which was deleted earlier, this should return True for both as well? Or that wont ever happen?

mkangia (approved):
Overall looks good. Just left one comment/query.

corehq/apps/users/dbaccessors/all_commcare_users.py
snopoke: This return value is correct. If both are True then the user will just get a warning. Having (True, False) indicates that there is a 'live' user with that username which will block any new users being created.

PR #18749, opened 2017-12-01
proper handling for no profile

calellowitz:
Fixes regression where updating to latest saved state was failing on mobile.

@orangejenny cc: @sravfeyn 
sometimes mobile sends the build profile as `'None'` the string, instead of nothing. this handles that.

### DIFF
--- a/corehq/apps/app_manager/views/download.py
+++ b/corehq/apps/app_manager/views/download.py
@@ -10,7 +10,6 @@ from django.shortcuts import render
 from django.template.loader import render_to_string
 from django.utils.translation import ugettext_lazy as _
 
-from corehq import toggles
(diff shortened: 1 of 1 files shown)


corehq/apps/app_manager/views/download.py
```diff
@@ -32,6 +31,13 @@
                       "If error persists, please contact us at commcarehq-support@dimagi.com")
 
 
+def _get_profile(request):
+    profile = request.GET.get('profile')
+    if profile not in (None, 'None'):
```
orangejenny: `if profile in app.build_profiles` would be more robust

orangejenny (approved):

orangejenny:
(changed PR description and product label)

calellowitz:
@orangejenny fixed the feedback, should be good to go

PR #18752, opened 2017-12-01
command to deidentify ICDS cases

snopoke:
@dimagi/scale-team

### DIFF
--- a/corehq/apps/cleanup/management/commands/deid_case_data.py
+++ b/corehq/apps/cleanup/management/commands/deid_case_data.py
@@ -6,6 +6,7 @@ from django.db import connections
 
 from corehq.form_processor.utils.sql import fetchone_as_namedtuple
 from corehq.sql_db.util import get_db_aliases_for_partitioned_query
+from corehq.util.log import with_progress_bar
 
 DEID_FIELDS = {
     'household': [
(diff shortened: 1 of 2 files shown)


corehq/apps/cleanup/management/commands/deid_case_data.py
```diff
@@ -0,0 +1,135 @@
...
+                        UPDATE form_processor_commcarecasesql
+                        SET case_json = json_object_set_keys(case_json::json, %s::TEXT[], %s::TEXT[])
+                        WHERE domain = %s and type = %s AND id > %s AND id <= %s;
+                    """, [keys, values, domain, case_type, min_id, max_id])
+
+                min_id += rows_to_update
+                max_id += rows_to_update
+
+
+def _create_function(db_name):
```
czue: any reason you chose to do this here instead of in a migration? 

i guess it's not needed on all envs...

czue (approved):
Quite cool you can do this. If it was simple it might be nice to have a test, though I realize it's a one-off.

corehq/apps/cleanup/management/commands/deid_case_data.py
snopoke: This is a once off command that I'm already running. Not sure we should event commit it to the repo since I don't think we'd ever use the approach on live data.

emord: +1 we should not commit this

PR #18753, opened 2017-12-01
Add More Domain Analytics

sheelio:
@nickpell 
I'm working on adding some new domain level calculations to support some of our retention / pricing analysis work. 

This does the following:
- Adds new domain level calculated properties
- Tracks Saved Dashboard / Saved Export downloads through kissmetrics
- Tracks API access through kissmetrics


nickpell:
Seems reasonable that we could have an API that retrieves these calculations.

The accounting APIs use tastypie to retrieve postgres models:
https://github.com/dimagi/commcare-hq/blob/master/corehq/apps/api/accounting.py

sheelio:
Cool - are you suggesting we should add an API instead for these? The code I modified is used here:
https://github.com/dimagi/commcare-hq/blob/194113c1d571bc6bbe206b7115d0f48f449215ca/corehq/apps/reports/tasks.py

And then exposed as a resource here:
https://github.com/dimagi/commcare-hq/blob/2ed89103410b228196fb7c9f564e3b00f3c80ef5/corehq/apps/api/domain_metadata.py

nickpell:
Using ```DomainMetadataResource``` looks right

sheelio:
@nickpell 
I think this is good to review.

sheelio:
@calellowitz   Not sure if you or Nick are best suited to review these changes.

corehq/apps/api/resources/__init__.py
```diff
@@ -104,6 +105,11 @@ def dispatch(self, request_type, request, **kwargs):
                 content_type="application/json",
                 status=401))
         if request.user.is_superuser or domain_has_privilege(request.domain, privileges.API_ACCESS):
+            if type(self) is DomainSpecificResourceMixin:
```
nickpell: should use ```isinstance```

nickpell (changes requested):

nickpell (approved):

sheelio:
@nickpell 
Tests are now passing - let me know if good to merge or if you want to review once more.

PR #18754, opened 2017-12-01
Add case property history toggle

proteusvacuum:
This should hopefully just be temporary... 👿 
https://manage.dimagi.com/default.asp?265989
@emord @sheelio

### DIFF
--- a/corehq/toggles.py
+++ b/corehq/toggles.py
@@ -923,6 +923,12 @@ SUPPORT = StaticToggle(
     help_link='https://confluence.dimagi.com/display/ccinternal/Support+Flag',
 )
 
+CASE_PROPERTY_HISTORY = StaticToggle(
+    'case_property_history',
+    'Shows a modal on the case property page allowing you to see the history of various case properties',
+    TAG_SOLUTIONS,
+)
+
 BASIC_CHILD_MODULE = StaticToggle(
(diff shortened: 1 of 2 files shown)


emord (approved):

corehq/toggles.py
```diff
@@ -923,6 +923,12 @@ def _commtrackify(domain_name, toggle_is_enabled):
     help_link='https://confluence.dimagi.com/display/ccinternal/Support+Flag',
 )
 
+CASE_PROPERTY_HISTORY = StaticToggle(
+    'case_property_history',
+    'Shows a modal on the case property page allowing you to see the history of various case properties',
+    TAG_SOLUTIONS,
```
emord: Err. Actually can you add the namespace domain here?

emord (changes requested):

corehq/toggles.py
emord: Also I think if this is going to get released, it shouldn't be solutions? @sheelio would know more

sheelio: @proteusvacuum 
Ya, I would product path it if we do intend to release it.

proteusvacuum: Updated

emord (approved):

PR #18755, opened 2017-12-01
Make management command to flag duplicate IDs

esoergel:
https://trello.com/c/Av4Y5ltP/264-duplicate-id-flagging
This PR pulls out the code which identifies duplicate IDs and uses that for a management command which flags all of those cases for later update.

@sravfeyn

### DIFF
--- a/custom/enikshay/duplicate_ids.py
+++ b/custom/enikshay/duplicate_ids.py
@@ -3,6 +3,7 @@ from __future__ import unicode_literals
 from collections import Counter
 import six
 
+from corehq.apps.es import CaseES
 from corehq.apps.users.models import CommCareUser
 from corehq.form_processor.backends.sql.dbaccessors import CaseAccessorSQL
 from corehq.form_processor.interfaces.dbaccessors import CaseAccessors
@@ -11,9 +12,10 @@ from dimagi.utils.couch.database import iter_docs
 from .const import VOUCHER_ID, ENIKSHAY_ID
 
 
-def get_cases_with_duplicate_ids(domain, case_type, all_case_ids):
+def get_cases_with_duplicate_ids(domain, case_type):
     accessor = CaseAccessors(domain)
     id_property = {'voucher': VOUCHER_ID, 'person': ENIKSHAY_ID}[case_type]
+    all_case_ids = accessor.get_case_ids_in_domain(case_type)
     all_cases = [
         {
             'case_id': case.case_id,
@@ -28,15 +30,14 @@ def get_cases_with_duplicate_ids(domain, case_type, all_case_ids):
 
 
 def get_bad_case_info(domain, case_type, full_debug_info=False):
-    accessor = CaseAccessors(domain)
-    case_ids = accessor.get_case_ids_in_domain(case_type)
-    bad_cases = get_cases_with_duplicate_ids(domain, case_type, case_ids)
+    total_cases = CaseES().domain(domain).case_type(case_type).count()
+    bad_cases = get_cases_with_duplicate_ids(domain, case_type)
     add_debug_info_to_cases(bad_cases, full_debug_info)
     context = {
         'case_type': case_type,
         'num_bad_cases': len(bad_cases),
-        'num_total_cases': len(case_ids),
-        'num_good_cases': len(case_ids) - len(bad_cases),
+        'num_total_cases': total_cases,
+        'num_good_cases': total_cases - len(bad_cases),
         'bad_cases': bad_cases,
     }
     return context
--- a/custom/enikshay/duplicate_ids.py
+++ b/custom/enikshay/duplicate_ids.py
@@ -1,14 +1,13 @@
 from __future__ import absolute_import
 from __future__ import unicode_literals
-import six
 from collections import Counter
+import six
 
 from corehq.apps.users.models import CommCareUser
 from corehq.form_processor.backends.sql.dbaccessors import CaseAccessorSQL
 from corehq.form_processor.interfaces.dbaccessors import CaseAccessors
 from dimagi.utils.couch.database import iter_docs
 
-from .case_utils import CASE_TYPE_VOUCHER, CASE_TYPE_PERSON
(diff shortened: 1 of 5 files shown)


custom/enikshay/duplicate_ids.py
```diff
@@ -12,6 +14,68 @@
 from .const import VOUCHER_ID, ENIKSHAY_ID
 
 
+class DuplicateIdsResult(AggregationResult):
```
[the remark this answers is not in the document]
proteusvacuum: Thanks for trying! I'm interested in understanding how we can do stuff like this better in the future, so this is all good to know.

custom/enikshay/duplicate_ids.py
```diff
@@ -0,0 +1,96 @@
...
+from corehq.form_processor.backends.sql.dbaccessors import CaseAccessorSQL
+from corehq.form_processor.interfaces.dbaccessors import CaseAccessors
+from dimagi.utils.couch.database import iter_docs
+
+from .const import VOUCHER_ID, ENIKSHAY_ID
+
+
+def get_cases_with_duplicate_ids(domain, case_type, all_case_ids):
+    accessor = CaseAccessors(domain)
+    id_property = {'voucher': VOUCHER_ID, 'person': ENIKSHAY_ID}[case_type]
```
sravfeyn: Isn't below be more direct and readable?
```
id_property = VOUCER_ID if case_type == 'voucher' else ENIKSHAY_ID
```

custom/enikshay/duplicate_ids.py
```diff
@@ -0,0 +1,96 @@
...
+    return context
+
+
+def add_debug_info_to_cases(bad_cases, full_debug_info):
+    _add_form_info_to_cases(bad_cases, full_debug_info)
+    _add_user_info_to_cases(bad_cases)
+
+
+def _add_form_info_to_cases(bad_cases, full_debug_info):
+    for case in bad_cases[:None if full_debug_info else 300]:
```
sravfeyn: What is 300?

custom/enikshay/duplicate_ids.py
```diff
@@ -0,0 +1,96 @@
...
+    auth_usernames = {user_doc['_id']: user_doc['username'] for user_doc in
+                      iter_docs(CommCareUser.get_db(), auth_user_ids)}
+    for case in bad_cases:
+        user_dict = user_info.get(case.get('form_user_id'))
+        if user_dict:
+            case['username'] = user_dict['username']
+            device_id = case['form_device_id']
+            if device_id == 'Formplayer':
+                if case['form_user_id'] == case['auth_user_id']:
+                    device_id = "WebAppsLogin"
```
sravfeyn: Can this be declared as a constant somewhere?

custom/enikshay/duplicate_ids.py
```diff
@@ -0,0 +1,96 @@
...
+
+from corehq.apps.users.models import CommCareUser
+from corehq.form_processor.backends.sql.dbaccessors import CaseAccessorSQL
+from corehq.form_processor.interfaces.dbaccessors import CaseAccessors
+from dimagi.utils.couch.database import iter_docs
+
+from .const import VOUCHER_ID, ENIKSHAY_ID
+
+
+def get_cases_with_duplicate_ids(domain, case_type, all_case_ids):
```
sravfeyn: Doc comments on these functions will be useful for others.

sravfeyn (approved):
Looks good, very minor comments. Feel free to self-merge after addressing them

custom/enikshay/duplicate_ids.py
snopoke: unless you want to raise an exception when the case type is neither `voucher` nor `person`

custom/enikshay/duplicate_ids.py
```diff
@@ -0,0 +1,96 @@
...
+
+
+def add_debug_info_to_cases(bad_cases, full_debug_info):
+    _add_form_info_to_cases(bad_cases, full_debug_info)
+    _add_user_info_to_cases(bad_cases)
+
+
+def _add_form_info_to_cases(bad_cases, full_debug_info):
+    for case in bad_cases[:None if full_debug_info else 300]:
+        form = CaseAccessorSQL.get_transactions(case['case_id'])[0].form
```
snopoke: If you're wanting to get the 'opening' form you could use `case.get_opening_transactions()[0]`

custom/enikshay/duplicate_ids.py
esoergel: It's an arbitrary number of cases for which to include expensive debug info.  I moved this to the view and added a comment

custom/enikshay/duplicate_ids.py
esoergel: I'm not sure that's appropriate.  This is a hack that I wouldn't want to grant that much legitimacy.  WebApps uses different device IDs in restores and in form submissions, and this code takes a guess as to what WebApps probably used in the restore.  I added a comment about it.

custom/enikshay/duplicate_ids.py
esoergel: I just added a file-level explanation of the issue that these functions address.  With that context, my hope is that the function names can be descriptive enough.

custom/enikshay/duplicate_ids.py
esoergel: these are case stubs, so unfortunately that's not available.  Would it be preferable to do something like this?
```python
CaseAccessorSQL.get_transactions_by_type(case_id, CaseTransaction.TYPE_CASE_CREATE)
```

snopoke: I think that is preferable since it will load a lot less transactions from the DB.

esoergel: Updated.  Thanks for the tip!

millerdev (approved):
Only reviewed last three commits, as requested by @esoergel

PR #18757, opened 2017-12-01
add feature flag for domain to retry failed SMS messages indefinitely

nickpell:
Product Note: Adds feature flag for enikshay so outgoing SMS messages that have reached the maximum number of failed attempts can still be retried, at a less frequent interval than the normal retry schedule, instead of never attempting the SMS again.

Set the retry interval for SMSs that have failed the maximum number of times to 6 hours.

https://trello.com/c/687nFMt2/281-configure-sms-reminders-to-dictate-whether-messages-are-sticky-which-can-be-retried

@gcapalbo @ctsims

### DIFF
--- a/corehq/apps/sms/tasks.py
+++ b/corehq/apps/sms/tasks.py
@@ -20,6 +20,7 @@ from corehq.apps.sms.change_publishers import publish_sms_saved
 from corehq.apps.sms.util import is_contact_active
 from corehq.apps.users.models import CouchUser, CommCareUser
 from corehq.form_processor.interfaces.dbaccessors import CaseAccessors
+from corehq.toggles import RETRY_SMS_INDEFINITELY
 from corehq.util.celery_utils import no_result_task
 from corehq.util.timezones.conversions import ServerTime
 from dimagi.utils.couch.cache.cache_core import get_redis_client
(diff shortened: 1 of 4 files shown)


corehq/apps/sms/tasks.py
```diff
@@ -48,6 +49,8 @@ def handle_unsuccessful_processing_attempt(msg):
     msg.num_processing_attempts += 1
     if msg.num_processing_attempts < settings.SMS_QUEUE_MAX_PROCESSING_ATTEMPTS:
         delay_processing(msg, settings.SMS_QUEUE_REPROCESS_INTERVAL)
+    elif RETRY_SMS_INDEFINITELY.enabled(msg.domain):
```
gcapalbo: could we also add a check here that msg.direction == OUTGOING? just to be explicit that this won't apply to inbound messages

corehq/toggles.py
```diff
@@ -875,6 +875,14 @@ def _commtrackify(domain_name, toggle_is_enabled):
     always_enabled={"enikshay"},
 )
 
+RETRY_SMS_INDEFINITELY = StaticToggle(
+    'retry_sms_indefinitely',
+    'Enikshay: Retry SMS indefinitely',
+    TAG_CUSTOM,
+    [NAMESPACE_DOMAIN],
+    description='Requeues an SMS that has reached the maximum number of unsuccessful attempts.',
```
gcapalbo: (technically, we're not requeueing - we're just never removing from the queue)

nickpell:
@gcapalbo updated

gcapalbo (approved):

kaapstorm (approved):

PR #18764, opened 2017-12-03
prep for running filter conversion script

nickpell:
These changes will make the PR for running ```python-modernize . -w -f libmodernize.fixes.fix_filter``` much cleaner.

Recommend 🐡 

First commit replaces a usage of ```filter``` with a cleaner implementation.

Second commit adds a ```filter``` import statement so conversion scripts won't apply to overridden usages of ```filter```.

Third commit adds a ```filter``` import statement to prevent the conversion script from replacing ```filter(None, ...)``` syntax with a messier list comprehension.  I only did this in cases where it was obviously ok to do, and will let the conversion script replace with list comprehensions in all other occurrences.

@dimagi/py3

### DIFF
--- a/corehq/apps/app_manager/app_translations/app_translations.py
+++ b/corehq/apps/app_manager/app_translations/app_translations.py
@@ -899,7 +899,7 @@ def _has_at_least_one_translation(row, prefix, langs):
     :param langs:
     :return:
     """
-    return bool(filter(None, [row.get(prefix + '_' + l) for l in langs]))
+    return any(row.get(prefix + '_' + l) for l in langs)
 
 
 def _get_col_key(translation_type, language):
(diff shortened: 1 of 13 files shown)


kaapstorm:
I'm happy with commits 1 and 3, and I know it's super nitpicky, but there is just something that feels wrong to me about importing something you don't need in order to avoid renaming a variable from `filter` to, say, `filter_`. Is there a reason you chose not to rename the variable?

kaapstorm:
... hmmm, I guess it's just temporary. ... OK.

kaapstorm (approved):

nickpell:
@kaapstorm It seemed like a slightly less risky fix

PR #18767, opened 2017-12-04
hypelink when available

sravfeyn:
When you search in HQ by doc_id, a corresponding HQ link for that doc will be displayed if available. For e.g. if you search for an application ID, a link to view the app will be displayed.

screen shot 2017-12-04 at 4 27 24 pm

### DIFF
--- a/corehq/apps/hqadmin/templates/hqadmin/doc_in_es.html
+++ b/corehq/apps/hqadmin/templates/hqadmin/doc_in_es.html
@@ -25,12 +25,7 @@
             </div>
             {% if doc_id %}
                 <h3>Doc {{ doc_id }} {{ es_info.status }}</h3>
-                <p>
-                Doc type: {{ couch_info.doc_type }}
-                {% if hq_url %}
-                    (<a href="{{hq_url}}">View on CommCareHQ</a>)
-                {% endif %}
-                </p>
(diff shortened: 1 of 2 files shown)


sravfeyn:
@esoergel

esoergel:
Have you seen that the [raw_doc](https://www.commcarehq.org/hq/admin/raw_doc/?id=83b7983f565c4b62978c86c983b405c8&db_name=) view has something like this already?  The commcare flower icon will use the existing search functionality to find the document, if possible.  I'd also recommend using that tool over `doc_in_es` unless you're specifically interested in the version stored in elasticsearch.

sravfeyn:
Hey yeah, I think it's worth keeping if we are keeping `doc_in_es` around.

esoergel:
Fair enough.  Although I'd still recommend linking instead to [`quick_find`](https://github.com/dimagi/commcare-hq/blob/master/corehq/apps/hqwebapp/views.py#L1081-L1081) (or whatever it uses under the hood), which knows how to handle a very broad range of doc types.

esoergel (approved):

sravfeyn:
Good point Ethan! Updated

esoergel:
Cool, thanks!

PR #18769, opened 2017-12-04
Conditional Alerts UI: Step 3

gcapalbo:
spec: https://docs.google.com/document/d/1iJq6EHEYTppIx0p2UMsIr8q3qKk_bFAAfvojtMsxvH4/edit#

This integrates the scheduling form into the conditional alert UI, fills in the edit conditional alert view, and fills in the conditional alert list view, allowing a full create-list-edit workflow. The rest of the scheduling and recipient options will follow in subsequent PRs.

Best reviewed commit by commit.

screen shot:
schedule-integration

Product: This feature flag isn't enabled yet for anyone as the new reminder UI is under active development.

@millerdev

### DIFF
--- a/corehq/messaging/scheduling/views.py
+++ b/corehq/messaging/scheduling/views.py
@@ -244,12 +244,47 @@ class ConditionalAlertListView(BaseMessagingSectionView, DataTablesAJAXPaginatio
     urlname = 'conditional_alert_list'
     page_title = _('Schedule a Conditional Message')
 
+    LIST_CONDITIONAL_ALERTS = 'list_conditional_alerts'
+
     @method_decorator(_requires_new_reminder_framework())
     @method_decorator(requires_privilege_with_fallback(privileges.OUTBOUND_SMS))
     @method_decorator(require_permission(Permissions.edit_data))
     @use_datatables
     def dispatch(self, *args, **kwargs):
         return super(ConditionalAlertListView, self).dispatch(*args, **kwargs)
+
+    def get_conditional_alerts_queryset(self):
+        return (
+            AutomaticUpdateRule
+            .objects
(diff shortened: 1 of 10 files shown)


corehq/messaging/scheduling/views.py
```diff
@@ -280,10 +316,22 @@ def page_context(self):
...
+        return ConditionalAlertScheduleForm(self.domain, self.schedule, self.rule)
+
+    @property
+    def schedule(self):
+        return None
+
     @property
-    def initial_rule(self):
+    def rule(self):
         return None
```
millerdev: Nit: could these be simple class attributes?
```py
schedule = None
rule = None
```

millerdev (approved):
Code looks good. Not sure if the test failure is related to these changes.
```
FAIL: casexml.apps.phone.tests.test_cleanliness:OwnerCleanlinessTest.test_hint_invalidation
  File "corehq/ex-submodules/casexml/apps/phone/tests/test_cleanliness.py", line 198, in test_hint_invalidation
    self._set_owner(self.parent.case_id, new_owner)
  File "corehq/ex-submodules/casexml/apps/phone/tests/test_cleanliness.py", line 106, in _set_owner
    self.assertEqual(owner_id, case.owner_id)
AssertionError: '888aedc624be4448b345b0ad14d82dd2' != u'5035968ffee74d81bc609b38427410ff'
```

corehq/messaging/scheduling/views.py
gcapalbo: i kind of liked keeping it with a property definition to help reinforce that it should be overridden in the subclass, and also because since they don't have a setter it's not possible to accidentally give them a value and get things mixed up in the base class

PR #18770, opened 2017-12-04
Implement help text showing via a modal instead of a popover

amstone326:
**Product Note:** When a user clicks on the help text icon for a question in Live Preview or Web Apps, the text will now show in a modal/dialog, instead of as a small popover.

Addresses https://manage.dimagi.com/default.asp?265415. It turns out that most of this was already implemented, but the popover for the help text was broken in live preview (it was working fine in Web Apps). After investigating for a while, Jenny and I determined that this was because `target=focus` attribute seems to not work in iframes (changing to a different target mode fixes it, and this unanswered SO post also lends credence to this theory: https://stackoverflow.com/questions/26910870/how-can-i-inject-a-popover-inside-an-iframe). Instead of solving by switching to another target mode, which would create an inconsistency with how the rest of HQ does popovers, we decided to implement this as a modal instead. This ultimately seemed more correct because it creates parity with how help text appears on mobile.


amstone326:
Before (in Web Apps):
screen shot 2017-12-04 at 9 26 51 am

After (in Web Apps):
screen shot 2017-12-04 at 9 27 07 am

After (in Live Preview):
screen shot 2017-12-04 at 9 26 01 am

wpride (approved):
Woo!

orangejenny:
Possibly worth directly tagging @dimagi/product so they see this well before the release (not sure when they look at labeled PRs for mobile).

corehq/apps/cloudcare/templates/form_entry/templates.html
```diff
@@ -112,15 +112,25 @@ <h1 class="title" data-bind="text: title, visible: !showInFormNavigation()"></h1
...
                     role="button"
-                    data-trigger="focus"
                     data-placement="top"
-                    data-toggle="popover"
-                    data-bind="attr: { 'data-content': help }">
+                    data-toggle="modal"
+                    data-target="#help-text-modal">
                     <i class="fa fa-info-circle"></i>
                 </a>
+                <div id="help-text-modal" class="modal fade" role="dialog" tabindex='-1'>
```
orangejenny: Does this definitely work if there are multiple questions with help text on the same page? The semantics for `id` are that it's a unique identifier, so if you try to `getElementById` (which bootstrap is presumably doing somewhere) and there are two of that id on the page, I don't know that it'll get the one you want.

Might be better to add a click handler that calls `.modal('show')` rather than using the `data-*` attributes to control these modals.

czue:
this is just an annoying comment from the :peanuts: gallery, but how do we feel about maybe centering the modal in the web view? it looks a little out of place to me up there in the upper leftish quadrant. :triangular_ruler:

amstone326:
@czue Sorry the screenshot of web apps is misleading; that modal is actually centered horizontally, I just took a screenshot that didn't capture the whole screen. (@orangejenny said she assumes you were suggesting horizontal centering only, since vertical is a lot harder)

amstone326:
@orangejenny So the tree-traversal strategy for selecting the modal to show is working as expected and was easy to set up. But I've now been struggling with the fact that doing it this way has made it so that the modal disappears immediately after it's shown. 

I did a bunch of SO reading, but none of the suggestions I found seem like they apply to us. The primary one is that if you're loading bootstrap from multiple sources, that can cause the `.modal()` function to actually get invoked twice, thus hiding it after it's shown. But I looked at the page sources and only saw `bootstrap.min.js` there, and also we use the `.modal()` pattern elsewhere in HQ, and presumably that sort of problem would be affecting those other instances too.

Does anything else come to mind for what could be causing this? If not I'll keep digging around tomorrow.

orangejenny:
If `.modal()` is getting called twice, it's likely that your click handler is being attached twice. If that were the case, the old `.popover()` call also would have been called twice, which could happen - calling `modal` multiple times will open and close the modal but don't know that `popover` behaves the same way. So that's something to investigate - when, and how many times, does `afterRender` get called?

FYI: The `click` function is a shortcut for the more generic [on](http://api.jquery.com/on/), which can attach any kind of event to an element. There's an analogous [off](http://api.jquery.com/off/) function that you can use to clear event handlers from an element. When event handling gets messy, we sometimes just use `off` to remove any existing handlers of the given type and re-attach them with `on`.

czue:
@amstone326 ah gotcha. i think vertical centering would be nicer but definitely not trying to ask you to do a bunch of work to make it happen. 

surprised to hear it's hard though, i thought it would be [as simple as this](https://stackoverflow.com/a/24829651/8207). either way, probably much better things for you to be doing with your time!

orangejenny:
@czue Heh, that's basically what I mean: that link is a good deal more complicated than the `margin: auto` that horizontally centers the modal. But yes, it's certainly doable. The comments suggest that solution doesn't allow a user to click the backdrop to close the modal, which would be inconsistent with most of our other modals. On the other hand, vertical centering in app preview would have a nice parity with mobile. Don't think it's a big deal either way, leaving it to Aliza.

amstone326:
@orangejenny Sadly I had actually already tried both of those things. `afterRender()` is only getting called once, and calling `$(".help-text-trigger").off();` before attaching the click listener doesn't change the behavior.

1 question- Nearly every post that I've found about this online ends up coming to the same conclusion as [this one](https://www.sitepoint.com/community/t/bootstrap-modal-closes-immediately/203801), i.e. that including 2 versions of bootstrap.js or including the same one twice would cause this. From what I could tell via a grep in the corehq/apps/cloudcare directory, we aren't doing that, but would you mind checking to see if you see something I missed?

amstone326:
@orangejenny Alrighty, this is ready to go now..

corehq/apps/cloudcare/templates/form_entry/templates.html
```diff
@@ -112,15 +112,25 @@ <h1 class="title" data-bind="text: title, visible: !showInFormNavigation()"></h1
...
+                    <div class="modal-dialog">
+                        <div class="modal-content">
+                            <div class="modal-body">
+                                <p align="left" data-bind="text: help"></p>
+                            </div>
+                            <div class="modal-footer">
+                                <!-- This needs to be a <span> instead of a <button> because (inexplicably) making it
+                                a button causes it to get clicked automatically when the modal opens, thus closing
+                                the modal immediately after it opens. We have yet to figure out why this is so. -->
+                                <span class="btn btn-default help-modal-close" data-dismiss="modal">Close</span>
```
orangejenny: should be translated

corehq/apps/cloudcare/static/cloudcare/js/form_entry/fullform-ui.js
```diff
@@ -277,8 +277,10 @@ function Form(json) {
     };
 
     self.afterRender = function() {
-        // Once form has finished rendering, render all popovers
-        $('.js-form-container [data-toggle="popover"]').popover();
+        $(".help-text-trigger").click(function(event) {
+            var container = $(event.target).closest(".caption");
+            container.find("#help-text-modal").modal('show');
```
orangejenny: Picky, but you still have an id that appears multiple times on the page. The code works, but it still isn't semantic. I'd change this to `.help-text-modal` or just remove `help-text-modal` altogether and select `.modal` here, since there's only going to be one modal in the `.caption`.

corehq/apps/cloudcare/static/cloudcare/js/form_entry/fullform-ui.js
```diff
@@ -277,8 +277,10 @@ function Form(json) {
     };
 
     self.afterRender = function() {
-        // Once form has finished rendering, render all popovers
-        $('.js-form-container [data-toggle="popover"]').popover();
+        $(".help-text-trigger").click(function(event) {
+            var container = $(event.target).closest(".caption");
```
orangejenny: Very picky, but it's good to get into the habit of using `currentTarget` instead of `target`. `currentTarget` is the element the handler is attached to, whereas `target` is the element that actually received the event (may be a child of `currentTarget`). This code will work either way, just a best practice.

orangejenny (approved):
Some picky comments. Feel free to self-merge after adding the translation.

corehq/apps/cloudcare/templates/form_entry/templates.html
amstone326: @orangejenny From looking at other templates, I assume you mean to use this pattern?:

`{% trans 'some text' %}`

The reason I ask is because in this particular file, it doesn't look like any other display strings are using that pattern (they're all just hard-coded). Not sure if there's a reason for that or just oversight. Also, am I intended to actually add the translations myself? I assume not but just not sure how this works in HQ.

orangejenny: Yep, `trans` for python templates (`gettext` translates JavaScript strings, it's used elsewhere in this file). You don't add translations, there's a process that picks up the strings and puts them in transifex. The whole file should be translated, it's an oversight that it's mostly untranslated.

amstone326: @orangejenny Ok cool thanks. PR is updated

orangejenny (approved):
thanks for updates, looks good!

PR #18771, opened 2017-12-04
python-modernize . -w -f libmodernize.fixes.fix_filter

nickpell:
@dimagi/py3


millerdev (approved):
Didn't review every line, but didn't see any issues on the lines I did review.

millerdev:
Would be nice to wrap some of those long lines since they were not so long before.

kaapstorm (commented):
:boom:

PR #18763, opened 2017-12-03
Add add and dob to ccs records and child health records

emord:
@sheelio @dannyroberts Adds the values to the UCR table we talked about the other day. Will still need to follow it up with a change to the aggregation script, but want to get this out sooner to get the data there.

budyd @proteusvacuum

PR #18772, opened 2017-12-04
Aggregate based on new columns

emord:
@sheelio @dannyroberts 

Follow up for #18763 Sounds like they're still waiting for sign off from the ministry so leaving this as open for review. Tests will also fail right now, but will fix those later.

buddy @proteusvacuum


dannyroberts (approved):

emord:
@czue updated this to filter out births and deliveries from 4 days ago so that the discrepancy disappears. 

These tests are frustratingly brittle so it's not tested here. I did test the aggregation script and ensure it had the correct values here: https://github.com/dimagi/commcare-hq/commit/a753bff178d9b09d0582e6c02b2227548daa34a8 but I don't think I'll get to the query tests

czue (approved):
I don't have a whole lot of context but this all looks reasonable. @emord saw you removed the label - assume this is good to go from the team's perspective? any concerns about deploying it right before you go offline?

emord:
> saw you removed the label - assume this is good to go from the team's perspective

yep

>any concerns about deploying it right before you go offline?

you have my number :) But if  you do want to revert this for whatever reason, it should be sufficient to get rid of the added where clauses that filter on add and/or dob. I just uploaded a commit on a branch that does this https://github.com/dimagi/commcare-hq/commit/5b017290c3e26704761d1fbba9b1560b2828ff53

dannyroberts:
OK cool, so basically if anything goes wrong with this while you're gone, we should PR and merge this? https://github.com/dimagi/commcare-hq/compare/je/dont-filter-dob-add?expand=1. Am I interpreting that right?

PR #18775, opened 2017-12-04
Add jQuery UI to core_libraries.html

orangejenny:
Followup for https://github.com/dimagi/commcare-hq/pull/18745

Found a new dependency: jQuery UI, if included, needs to go after jQuery but before Bootstrap. Otherwise reports break. Dependency management is fun.

@millerdev / @emord 

@nickpell this needs to go into the next deploy

### DIFF
--- a/corehq/apps/hqwebapp/templatetags/hq_shared_tags.py
+++ b/corehq/apps/hqwebapp/templatetags/hq_shared_tags.py
@@ -748,7 +748,7 @@ def javascript_libraries(context, **kwargs):
     return {
         'request': getattr(context, 'request', None),
         'underscore': kwargs.pop('underscore', False),
-        'jquery': kwargs.pop('jquery', False),
+        'jquery_ui': kwargs.pop('jquery_ui', False),
         'ko': kwargs.pop('ko', False),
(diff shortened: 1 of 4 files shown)


nickpell (approved):

corehq/apps/hqwebapp/templatetags/hq_shared_tags.py
```diff
@@ -748,7 +748,7 @@ def javascript_libraries(context, **kwargs):
     return {
         'request': getattr(context, 'request', None),
         'underscore': kwargs.pop('underscore', False),
-        'jquery': kwargs.pop('jquery', False),
```
emord: Looks like `jquery` was never used. Do you know why it was here before?

emord (approved):

corehq/apps/hqwebapp/templatetags/hq_shared_tags.py
orangejenny: Originally it was an option, then I removed it in https://github.com/dimagi/commcare-hq/pull/18745/commits/72d963c1a8ade544268377355f037f4b1993d9aa but forgot to update this file.

PR #18778, opened 2017-12-04
Cache expensive part, with less variance

esoergel:
@sravfeyn @emord
The expensive call only varies on case_id, but since this was caching the result of the whole method, it had to vary on 2 or 3 parameters.  This change will result in more cache hits.


emord (approved):
nice

esoergel:
Github doesn't show the build as completed, but it is: https://travis-ci.org/dimagi/commcare-hq/builds/311511204

PR #18781, opened 2017-12-04
Reports: Disallow future dates in date spans

orangejenny:
Dev Patel requested in https://manage.dimagi.com/default.asp?266238 that we not allow date spans in reports to reach into the future. Sounds reasonable to me. @dimagi/product is this okay with you?

@calellowitz


sheelio:
@orangejenny  
We have some workflows that require future dates in reports (i.e. if I'm trying to filter on expected date of delivery).  I think to simplify things (and deal with timezone oddities), we decided to allow dates in the future a little while ago.

orangejenny:
@sheelio Cool, thanks.

PR #18782, opened 2017-12-04
Revert "Allow clearing date spans in reports"

orangejenny:
Reverts dimagi/commcare-hq#18711
See https://manage.dimagi.com/default.asp?266238

An issue with this came up during testing: when you clear dates and Apply, somewhere along the line they get defaulted back to the report's default date range. I don't have the bandwidth to investigate (originally pulled the ticket off the interrupt queue because it looked like a very easy win), just going to revert and put it back on the queue.

@calellowitz


orangejenny:
@calellowitz Not high priority, but it's just a revert.

PR #18784, opened 2017-12-04
Add decorator to cache UCR expressions

esoergel:
There's a lot of repeated boilerplate around caching UCR expressions in the context cache.  This PR attempts to simplify that.  There are a bunch more places where this could be used, but I figure this is a decent enough starting place.
@emord @sravfeyn

@dannyroberts I copied much of the cache key stuff from quickcache - you might wanna have a looksee.

### DIFF
--- a/corehq/apps/locations/ucr_expressions.py
+++ b/corehq/apps/locations/ucr_expressions.py
@@ -2,6 +2,7 @@ from __future__ import absolute_import
 from jsonobject import DefaultProperty
 
 from corehq.apps.locations.models import SQLLocation
+from corehq.apps.userreports.decorators import ucr_context_cache
 from corehq.apps.userreports.expressions.factory import ExpressionFactory
 from corehq.apps.userreports.specs import TypeProperty
 from corehq.util.quickcache import quickcache
@@ -85,14 +86,9 @@ class AncestorLocationExpression(JsonObject):
     def __call__(self, item, context=None):
         location_id = self._location_id_expression(item, context)
         location_type = self._location_type_expression(item, context)
+        return self._get_ancestor(location_id, location_type)
 
-        cache_key = (self.__class__.__name__, location_id, location_type)
-        if context.get_cache_value(cache_key, False) is not False:
-            return context.get_cache_value(cache_key)
-        ancestor_json = self._get_ancestor(location_id, location_type)
-        context.set_cache_value(cache_key, ancestor_json)
-        return ancestor_json
-
+    @ucr_context_cache(vary_on=('location_id', 'location_type',))
     def _get_ancestor(self, location_id, location_type):
         try:
             location = SQLLocation.objects.get(location_id=location_id)
--- a/corehq/apps/locations/ucr_expressions.py
+++ b/corehq/apps/locations/ucr_expressions.py
@@ -86,10 +86,11 @@ class AncestorLocationExpression(JsonObject):
     def __call__(self, item, context=None):
         location_id = self._location_id_expression(item, context)
         location_type = self._location_type_expression(item, context)
-        return self._get_ancestor(location_id, location_type)
+        return self._get_ancestor(location_id, location_type, context)
 
+    @staticmethod
     @ucr_context_cache(vary_on=('location_id', 'location_type',))
-    def _get_ancestor(self, location_id, location_type):
+    def _get_ancestor(location_id, location_type, context):
         try:
             location = SQLLocation.objects.get(location_id=location_id)
             ancestor = location.get_ancestors(include_self=False).get(location_type__name=location_type)
--- a/corehq/apps/userreports/decorators.py
+++ b/corehq/apps/userreports/decorators.py
@@ -51,7 +51,7 @@ def ucr_context_cache(vary_on=()):
     """
     def decorator(fn):
         assert 'context' in fn.func_code.co_varnames
-        assert isinstance(vary_on, (tuple, list))
+        assert isinstance(vary_on, tuple)
 
         @wraps(fn)
         def _inner(*args, **kwargs):
--- a/corehq/apps/userreports/decorators.py
+++ b/corehq/apps/userreports/decorators.py
@@ -1,4 +1,6 @@
 from __future__ import absolute_import
+import hashlib
+import inspect
 from functools import wraps
 
 from django.conf import settings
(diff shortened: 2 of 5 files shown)


corehq/apps/userreports/decorators.py
```diff
@@ -39,3 +41,32 @@ def _inner(*args, **kwargs):
...
+        assert 'context' in fn.func_code.co_varnames
+        assert isinstance(vary_on, (tuple, list))
+
+        @wraps(fn)
+        def _inner(*args, **kwargs):
+            # shamelessly stolen from quickcache
+            callargs = inspect.getcallargs(fn, *args, **kwargs)
+            context = callargs['context']
+            prefix = '{}.{}'.format(
+                fn.__name__[:40] + (fn.__name__[40:] and '..'),
```
emord: It'd be nice to be able to override this. I could see some instances where we want to share a cache (especially in custom expressions)

corehq/apps/locations/ucr_expressions.py
```diff
@@ -85,14 +86,9 @@ def configure(self, location_id_expression, location_type_expression):
...
-        if context.get_cache_value(cache_key, False) is not False:
-            return context.get_cache_value(cache_key)
-        ancestor_json = self._get_ancestor(location_id, location_type)
-        context.set_cache_value(cache_key, ancestor_json)
-        return ancestor_json
-
+    @ucr_context_cache(vary_on=('location_id', 'location_type',))
     def _get_ancestor(self, location_id, location_type):
         try:
             location = SQLLocation.objects.get(location_id=location_id)
```
emord: Perhaps worth caching both of these calls separately (getting the location and the ancestory), out of scope for this PR

corehq/apps/userreports/decorators.py
```diff
@@ -39,3 +41,32 @@ def _inner(*args, **kwargs):
...
+
+def ucr_context_cache(vary_on=()):
+    """
+    Decorator which caches calculations performed during a UCR EvaluationContext
+    The decorated function or method must have a parameter called 'context'
+    which will be used by this decorator to store the cache.
+    """
+    def decorator(fn):
+        assert 'context' in fn.func_code.co_varnames
+        assert isinstance(vary_on, (tuple, list))
```
emord: this could just be `tuple`  unless you see an instance where lists would be useful? I think we'd always want `vary_on` to be some constant?

corehq/apps/locations/ucr_expressions.py
esoergel: Agreed.  The data source I was looking at makes a bunch of DB hits which I think come through here - I'll be looking at this next.

corehq/apps/userreports/decorators.py
esoergel: It gets coerced to a tuple below, so it doesn't really matter, as long as it's an iterable.  Are you thinking it'd be safer from mutation or something?

corehq/apps/userreports/decorators.py
esoergel: I was thinking about that as well, but I figured if we'd want to share a cache, then it'd probably also want the same implementation.  That's basically what I did here: https://github.com/dimagi/commcare-hq/pull/18778/files

Right now this only works with simple arguments, but if/when it's expanded to support complex stuff, that could definitely be useful, and I'd support adding it.
Consider two expressions which get data from a form - one starts off with a form object, and the other starts off with only a form ID.  They'd both want to share a key, but we wouldn't want the first scenario to re-fetch the form, so the implementations would differ.  For the first scenario to work, though, this decorator would have to support something like `'form.get_id'` as a key.  I think I'll punt on that until I need it.

corehq/apps/userreports/decorators.py
emord: I think of it as a code simplicity thing. I can't imagine ever not wanting a tuple here, so enforcing it seems reasonable. (tuples are slightly more performant as well, but that's likely not a real issue here)

corehq/apps/userreports/decorators.py
emord: that makes sense 👍

corehq/apps/userreports/decorators.py
esoergel: Yeah, I suppose it's no real burden on future implementers to require a tuple here.  I'll change it.

corehq/apps/locations/ucr_expressions.py
```diff
@@ -85,14 +86,9 @@ def configure(self, location_id_expression, location_type_expression):
...
 
-        cache_key = (self.__class__.__name__, location_id, location_type)
-        if context.get_cache_value(cache_key, False) is not False:
-            return context.get_cache_value(cache_key)
-        ancestor_json = self._get_ancestor(location_id, location_type)
-        context.set_cache_value(cache_key, ancestor_json)
-        return ancestor_json
-
+    @ucr_context_cache(vary_on=('location_id', 'location_type',))
     def _get_ancestor(self, location_id, location_type):
```
dannyroberts: Doesn't this need a parameter called `context` to work, according to the `ucr_context_cache` docstring below?

esoergel: You're absolutely right.  Updated.

custom/enikshay/expressions.py
```diff
@@ -153,38 +150,22 @@ def __call__(self, item, context=None):
         return None
 
     @staticmethod
+    @ucr_context_cache(vary_on=('person_id',))
     def _get_referral(context, domain, person_id):
```
dannyroberts: any chance that by not including `domain` in the cache key, you're opening up a attack vector of accessing another domain's data through the cache? (i.e. I can make a request the involves looking up a person_id, and it'll look it up in the cache without checking it's for the right domain?) Or maybe that's not a problem because the `context` variable is going to be limited to a scope smaller than domains anyway? Any downside to varying on `domain` as well? Seems overall more correct to me.

emord: I think it's fine to vary on domain, but you're correct that context is only valid in a very short scope (only the current document change, it's just a local dict, not open to other processes)

esoergel: I don't feel strongly about it, was mostly just trying to mimic the existing functionality.  Like J$ said, the context is specific to a document change, so it's already domain-scoped and most of these functions get the domain _from_ the context itself, so it'd be redundant to specify the domain.  Setting the cache would essentially be:
```python
context.cache[(context.domain, mykey)] = get_my_value(context)
```
I suppose `ucr_context_cache` could assert that the context has a domain or something like that.

esoergel:
I think this should be g2m, unless @dannyroberts wants to push for the domain key.

custom/enikshay/expressions.py
esoergel: Actually, as I think about it more, it seems reasonable to add `domain` as a required parameter for `EvaluationContext`.  We pretty consistently assume that these contexts exist within the scope of a domain (including in this PR), so it doesn't seem unreasonable to require that explicitly.  I do think that's out of scope for this PR, but I could be talked into trying that out separately.

dannyroberts: I'd encourage it if you have some good ideas, but yeah this PR certainly doesn't need to block on it :)

dannyroberts (approved):

PR #18787, opened 2017-12-05
Mk resave es trips

mkangia:
added a utility to resave xforms and cases missing in es
like https://manage.dimagi.com/default.asp?266070#BugEvent.1429438

best review 🐟 / 🐟


corehq/apps/es/management/commands/resave_failed_forms_and_cases.py
```diff
@@ -0,0 +1,66 @@
...
+def perform_resave_on_xforms(domain):
+    xform_ids_missing_in_es, _ = compare_xforms(domain, 'XFormInstance')
+    print("%s Ids found for xforms missing in ES." % len(xform_ids_missing_in_es))
+    print(xform_ids_missing_in_es)
+    ok = input("Type 'ok' to continue: ")
+    if ok != "ok":
+        print("No changes made")
+        return
+    form_accessor = FormAccessors(domain)
+    for xform_id in with_progress_bar(xform_ids_missing_in_es, length=10):
```
snopoke: `length` should be `len(xform_ids_missing_in_es` or just left out entirely since `xform_ids_missing_in_es` is just a list

corehq/apps/hqcase/utils.py
```diff
@@ -255,3 +257,15 @@ def bulk_update_cases(domain, case_changes, device_id):
...
+
+
+def resync_case_to_es(domain, case):
+    if should_use_sql_backend(domain):
+        publish_case_saved(case)
+    else:
+        CommCareCase.get_db().save_doc(case._doc)  # don't just call save to avoid signals
+
+
+def resave_case(domain, case):
```
snopoke: Why is this method necessary?

corehq/apps/es/management/commands/resave_failed_forms_and_cases.py
snopoke: Also you could use `for form in with_progress_bar(form_accessor.iter_forms(form_ids), length=len(form_ids))`

corehq/apps/es/management/commands/resave_failed_forms_and_cases.py
```diff
@@ -0,0 +1,66 @@
...
+def perform_resave_on_cases(domain):
+    case_ids_missing_in_es, _ = compare_cases(domain, 'CommCareCase')
+    print("%s Ids found for cases missing in ES." % len(case_ids_missing_in_es))
+    print(case_ids_missing_in_es)
+    ok = input("Type 'ok' to continue: ")
+    if ok != "ok":
+        print("No changes made")
+        return
+    case_accessor = CaseAccessors(domain)
+    for case_id in with_progress_bar(case_ids_missing_in_es, length=10):
```
snopoke: same comments as above

corehq/apps/es/management/commands/resave_failed_forms_and_cases.py
mkangia: Ah! wanted to modify the granularity/ number of cases per iteration.
Misinterpreted `granularity = min(50, length)` in `with_progress_bar`.
Thanks for catching. Will avoid passing it.

corehq/apps/hqcase/utils.py
mkangia: Agree could have just called `resync_case_to_es` but added this method as a single place where we want to do things when re-saving case. Also keeps the same vocab as resave_form.
Plus,
`resync_case_to_es` is one of the specific things but overall might want to club this with something like resave to UCRs in future.
Fine to remove it if you think it would never be needed.

snopoke: Fine with keeping the function name `resave_case` but I don't think we need the two separate methods. This will re-trigger any UCRs for this case as well since what this is doing is putting the case in the kafka queue again so it will get picked up by all pillows listening to the case topic.

mkangia: cool, renaming.

corehq/apps/reports/views.py
```diff
@@ -169,7 +168,9 @@
     get_group,
     group_filter,
     users_matching_filter,
-    resync_case_to_es)
+)
+from corehq.form_processor.utils.xform import resave_form as _resave_form
+from corehq.apps.hqcase.utils import resave_case as _resave_case
```
snopoke: Why have these been renamed to appear private?

mkangia: there are already methods with those names in the view. Is there a better way to avoid conflict?

snopoke: I think renaming the views would be better e.g. `resave_case_view`. That will also require changes to the urls file.

mkangia: Sounds good. updated.

corehq/apps/hqadmin/management/commands/check_case_index_ids.py
```diff
@@ -108,7 +108,7 @@ def handle(self, **options):
 
             for case_to_resync in exists:
                 # if the case actually exists resync it to fix the es search
-                resync_case_to_es(domain, case_to_resync)
+                resave_case(domain, case_to_resync)
```
mkangia: Looking at this closely now, this probably is a bug.
this wants to just resync case but is actually doing two things
1. resync es
2. also send post case save signal.
it may be does not need the latter. Should this be updated now to `resave_case(domain, case_to_resync, send_post_save_signal=False)`

snopoke: yes I think so

snopoke (commented):
Otherwise looks good

corehq/apps/hqcase/utils.py
```diff
@@ -255,3 +256,11 @@ def bulk_update_cases(domain, case_changes, device_id):
...
         case_blocks.append(case_block)
     return submit_case_blocks(case_blocks, domain, device_id=device_id)
+
+
+def resave_case(domain, case, send_post_save_signal=True):
+    from corehq.form_processor.change_publishers import publish_case_saved
+    if should_use_sql_backend(domain):
+        publish_case_saved(case, send_post_save_signal)
+    else:
+        CommCareCase.get_db().save_doc(case._doc)  # don't just call save to avoid signals
```
snopoke: Based on the comment on this line if `send_post_save_signal == True` then I think this should call `case.save()`

snopoke (commented):
Sorry, just noticed one other thing

corehq/apps/hqcase/utils.py
mkangia: Thanks for catching that. Updated.

snopoke:
kicked build :boot: :construction:

mkangia:
thank you @snopoke. Did it fail due to some reason?

snopoke:
@mkangia it was an unrelated test failure

PR #18789, opened 2017-12-05
Mk 265955 kill user sync history

mkangia:
Product Note: Sync History Report is deprecated and redirects to App Status Report Page with message:
`The Sync History report has been deprecated. You can use the Application Status report to identify the last time a user synced.`
And any schedule emails would show:
`The Sync History report has been deprecated. You can use the Application Status report to identify the last time a user synced. This saved email will stop working within the next two months. Please update your saved reports email settings if needed.`

https://manage.dimagi.com/default.asp?265955

This PR adds a generic way to deprecate reports. Once a report is flagged as deprecated  (in code, no UI), Link to that report will redirect user back to previous page or any other configured page with message:
`This report has been deprecated.` or a custom message which can be added for that report on deprecation.
Same applies for saved reports using the deprecated report. Links on those pages will redirect user with the same message as above.

Similarly for any emails sent out for this report under Scheduled reports will show a message in the mail
`This report has been deprecated.` or a custom email message which was added for that report on deprecation.

### DIFF
--- a/corehq/apps/reports/standard/deployments.py
+++ b/corehq/apps/reports/standard/deployments.py
@@ -291,14 +291,22 @@ class SyncHistoryReport(DeploymentsReport):
     # To be removed. Link deactivated on 6th Dec 2017.
     name = ugettext_noop("User Sync History")
     slug = "sync_history"
+    is_deprecated = True
+    deprecation_email_message = ugettext_lazy(
+        "The Sync History report has been deprecated. You can use the Application Status report to identify "
+        "the last time a user synced. This saved email will stop working within the next two months. "
+        "Please update your saved reports email settings if needed.")
+    deprecation_message = ugettext_lazy(
+        "The Sync History report has been deprecated. "
+        "You can use the Application Status report to identify the last time a user synced."
+    )
 
     @property
-    def view_response(self):
+    def deprecate_response(self):
         from django.contrib import messages
-        messages.success(
+        messages.warning(
             self.request,
-            _('Sync History Report is not being maintained anymore and will be removed soon. '
-              'Please report an issue if you have any concerns regarding this.')
+            self.deprecation_message
         )
         return HttpResponseRedirect(
             reverse(ApplicationStatusReport.dispatcher.name(), args=[],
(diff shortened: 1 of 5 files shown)


corehq/apps/reports/standard/deployments.py
```diff
@@ -291,123 +288,21 @@ def _choose_latest_version(*app_versions):
...
-
-    @property
-    def headers(self):
-        headers = DataTablesHeader(
-            DataTablesColumn(_("Sync Date"), sort_type=DTSortType.NUMERIC),
-            DataTablesColumn(_("# of Cases"), sort_type=DTSortType.NUMERIC),
-            DataTablesColumn(_("Sync Duration"), sort_type=DTSortType.NUMERIC)
+    def view_response(self):
+        from django.contrib import messages
+        messages.success(
```
snopoke: maybe make this a `warning`

mkangia:
@sheelio can this go under some QA?

sheelio:
Sure - is it possible to put it on staging and then flag Ben / Clark to take a look?

sheelio:
@mkangia 

One question - this PR looks like it _removes_ the existing report?  Is that true?  Does the report also get removed from the email?  I think we want to keep the data in the email until we're ready to fully remove it.

mkangia:
@sheelio 

> One question - this PR looks like it removes the existing report? Is that true? Does the report also get removed from the email? I think we want to keep the data in the email until we're ready to fully remove it.

Yup, it removes the report content from the email to reflect the same behaviour as on the web.
We can find a way to keep it around but then it would be a different behaviour from web.

sheelio:
@mkangia 
I looked into it and its only one project who emails this - so I think fine to remove and I'll reach out to the project.

sheelio:
I'm reaching out to the project now - let me confirm before we merge this.

mkangia:
Awesome, thank you @sheelio for the insight

mkangia:
> Sure - is it possible to put it on staging and then flag Ben / Clark to take a look?

@sheelio this is on staging now. Have not reached out to Ben. Can you let them know or let me know if i should sharing what needs to be QA'ed

sheelio:
@mkangia Can you follow this process?

mkangia:
@sheelio sure sheel. Submitted request.

mkangia:
@sheelio this is g2g from QA. Safe to merge this now? If yes, just merge it.

PR #18791, opened 2017-12-05
make all data analytics es use export settings

calellowitz:
these queries were timing out on icds.
@orangejenny @sravfeyn


orangejenny (commented):
Don't have context, but code looks reasonable. Browsing similar code, looks like there's a `ES_EXPORT_INSTANCE` constant that would be better practice than the `'export'`.

orangejenny (approved):

calellowitz:
@orangejenny good catch, updated

orangejenny (approved):

PR #18792, opened 2017-12-05
fixing ucrs with bad filters

calellowitz:
@emord cc: @orangejenny 
so rebuilds dont keep adding millions of cases to the queue


emord (approved):
We could also start looking inside `and` filters and trying to pull these out, but this looks good for now

PR #18794, opened 2017-12-05
Fixes bug where location filter search doesn't return more than 10 matching results

sravfeyn:
Fixes a minor bug in location filter on 'Edit Mobile Worker' page which causes users not to be able to find a location if the search matches more than 10 locations. This makes it so that the user can scroll through the search results till all matching results are exhausted (10 at a time).

@esoergel 
https://manage.dimagi.com/default.asp?266669#1430120


corehq/apps/locations/views.py
```diff
@@ -901,52 +901,33 @@ def page_url(self):
...
     query = request.GET.get('name', '').lower()
+    page = int(request.GET.get('page', 1))
     user = request.couch_user
 
     base_queryset = SQLLocation.objects.accessible_to_user(domain, user)
 
     def loc_to_payload(loc):
         return {'id': loc.location_id, 'name': loc.get_path_display()}
 
-    if location_id:
```
sravfeyn: I am pretty sure none of this is being used. When the JS controller that uses this view [discontinued](https://github.com/dimagi/commcare-hq/commit/3a1a2123c8c067e386b47736a200765e852fb516#diff-3d26b140905a47f5974408406518e0f4) using this, it didn't clean up the view itself. I have also made sure this is not used at all by grepping some prod logs
```
cchq@hqproxy0:~/www/production/log$ cat production_commcare-nginx_access.log | grep '/child_locations.*id=' 
cchq@hqproxy0:~/www/production/log$ cat production_commcare-nginx_access.log | grep '/child_locations.*ids=' 
cchq@hqproxy0:~/www/production/log$ cat production_commcare-nginx_access.log | grep '/child_locations.*name=' 
196.41.59.154 - - - - - [04/Dec/2017:12:18:05 +0000] "GET /a/tz-lad/settings/locations/child_locations/?name= HTTP/2.0" 200 428 "https://www.commcarehq.org/a/tz-lad/settings/users/commcare/" "Mozilla/5.0 (Windows NT 6.1; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/62.0.3202.94 Safari/537.36"
196.41.59.154 - - - - - 
....
```

corehq/apps/locations/views.py
```diff
@@ -901,52 +901,33 @@ def page_url(self):
 @locations_access_required
 @location_safe
 def child_locations_for_select2(request, domain):
-    location_id = request.GET.get('id')
-    ids = request.GET.get('ids')
+    from django.core.paginator import Paginator
```
esoergel: This is cool, I haven't seen it before.

corehq/apps/locations/views.py
esoergel: Nice, thanks for the clean-up!

esoergel (approved):

PR #18799, opened 2017-12-05
Persona test: update copy

orangejenny:
This option gets cut off, and rewording is easier than futzing with styles to allow two-line options.

screen shot 2017-12-04 at 9 01 02 pm

@calellowitz / @biyeun


orangejenny:
@proteusvacuum blargh, you're right. not going to deal with that today. @biyeun any idea how complex it would be to update the styling here?

biyeun:
@orangejenny i can take a look at it next week? I'll be in office

orangejenny:
@biyeun sweet, that sounds great

PR #18802, opened 2017-12-05
Ce/no updates

calellowitz:
@proteusvacuum for enikshay context
@emord for ucr context 
@orangejenny code buddy
This removes a calculation from the nightly enikshay task, which served mainly to increment a case property by one, but caused 10s of thousands of cases to be updated, backing up ucr queues for enikshay. It pushes that calculation to the UCR itself since its really just date math.

### DIFF
--- a/custom/enikshay/tasks.py
+++ b/custom/enikshay/tasks.py
@@ -383,9 +383,11 @@ class EpisodeAdherenceUpdate(object):
 
         adherence_cases = self.get_valid_adherence_cases()
         dose_status_by_date = calculate_dose_status_by_day(adherence_cases)
+        doses_per_week = self.get_doses_per_week()
 
         update = {
             'adherence_total_doses_taken': self.count_doses_taken(dose_status_by_date),
+            'doses_per_week': doses_per_week
         }
         update.update(self.get_adherence_scores(dose_status_by_date))
         update.update(self.get_aggregated_scores(
@@ -487,7 +489,6 @@ class EpisodeAdherenceUpdate(object):
         )
 
         doses_per_week = self.get_doses_per_week()
-        update['doses_per_week'] = doses_per_week
 
         # the expected number of doses taken between the time the adherence
         # schedule started and the last valid date of the score
--- a/custom/enikshay/tasks.py
+++ b/custom/enikshay/tasks.py
@@ -487,6 +487,7 @@ class EpisodeAdherenceUpdate(object):
         )
 
         doses_per_week = self.get_doses_per_week()
+        update['doses_per_week'] = doses_per_week
 
         # the expected number of doses taken between the time the adherence
         # schedule started and the last valid date of the score
--- a/custom/enikshay/tasks.py
+++ b/custom/enikshay/tasks.py
@@ -385,7 +385,6 @@ class EpisodeAdherenceUpdate(object):
         dose_status_by_date = calculate_dose_status_by_day(adherence_cases)
 
         update = {
-            'total_expected_doses_taken': self.get_total_expected_doses_taken(adherence_schedule_date_start),
             'adherence_total_doses_taken': self.count_doses_taken(dose_status_by_date),
         }
         update.update(self.get_adherence_scores(dose_status_by_date))
@@ -496,7 +495,6 @@ class EpisodeAdherenceUpdate(object):
         num_days = (update['aggregated_score_date_calculated'] - adherence_schedule_date_start).days + 1
         update['expected_doses_taken'] = int(doses_per_week * num_days / 7.0)
 
-        update['total_expected_doses_taken'] = self.get_total_expected_doses_taken(adherence_schedule_date_start)
         return update
 
     @memoized
@@ -512,13 +510,6 @@ class EpisodeAdherenceUpdate(object):
             )
             return 0
         return doses_per_week
-
-    def get_total_expected_doses_taken(self, adherence_schedule_date_start):
-        doses_per_week = self.get_doses_per_week()
-        today = self.date_today_in_india
-        total_expected_doses_taken = int(
-            round(((today - adherence_schedule_date_start).days / 7.0) * doses_per_week))
-        return total_expected_doses_taken
 
     def check_and_return(self, update_dict):
         """
--- a/custom/enikshay/ucr/data_sources/episode_v4.json
+++ b/custom/enikshay/ucr/data_sources/episode_v4.json
@@ -19,19 +19,14 @@
         "table_id": "episode_v4",
         "display_name": "Episode v4",
         "configured_filter": {
-            "type": "and",
-            "filters": [
-                {
-                    "operator": "eq",
-                    "expression": {
-                        "datatype": null,
-                        "type": "property_name",
-                        "property_name": "type"
-                    },
-                    "type": "boolean_expression",
-                    "property_value": "episode"
-                }
-            ]
+            "operator": "eq",
+            "expression": {
+                "datatype": null,
+                "type": "property_name",
+                "property_name": "type"
+            },
+            "type": "boolean_expression",
+            "property_value": "episode"
         },
         "configured_indicators": [
             {
--- a/custom/enikshay/ucr/data_sources/episode_v4.json
+++ b/custom/enikshay/ucr/data_sources/episode_v4.json
@@ -912,6 +912,21 @@
                 "column_id": "adherence_schedule_id"
             },
             {
+                "display_name": "doses_per_week",
+                "datatype": "date",
+                "type": "expression",
+                "is_primary_key": false,
+                "transform": {
+                },
+                "is_nullable": true,
+                "expression": {
+                    "datatype": "integer",
+                    "type": "property_name",
+                    "property_name": "doses_per_week"
+                },
+                "column_id": "doses_per_week"
+            },
+            {
                 "display_name": "adherence_schedule_date_start",
                 "datatype": "date",
                 "type": "expression",
--- a/custom/enikshay/ucr/data_sources/episode_v4.json
+++ b/custom/enikshay/ucr/data_sources/episode_v4.json
@@ -912,6 +912,21 @@
                 "column_id": "adherence_schedule_id"
             },
             {
+                "display_name": "adherence_schedule_date_start",
+                "datatype": "date",
+                "type": "expression",
+                "is_primary_key": false,
+                "transform": {
+                },
+                "is_nullable": true,
+                "expression": {
+                    "datatype": "date",
+                    "type": "property_name",
+                    "property_name": "adherence_schedule_date_start"
+                },
+                "column_id": "adherence_schedule_date_start"
+            },
+            {
                 "display_name": "not_microbiological_basis_of_diagnosis",
                 "datatype": "string",
                 "type": "expression",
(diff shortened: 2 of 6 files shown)


custom/enikshay/ucr/reports/beneficiary_register_v2.json
```diff
@@ -0,0 +1,751 @@
+{
+    "domains": [
+        "enikshay"
+    ],
+    "server_environment": [
+        "enikshay",
+        "softlayer"
```
proteusvacuum: don't need this on softlayer

calellowitz:
Thoughts I had about potential better ways, but was unsure so left as is:
There are a lot of calculations in the UCR that should maybe get pushed to the data source, I think only the `total_expected_doses_taken` one needs to be there. Not sure which is better.
We can change the reconciliation task to add `doses_per_week` as a case property and then add it to the data source so we don't have to calculate it at all.

custom/enikshay/ucr/reports/beneficiary_register_v2.json
```diff
@@ -0,0 +1,751 @@
...
+                "display": "Adherence Mechanism Assigned"
+            },
+            {
+                "expression": {
+                    "type": "evaluator",
+                    "context_variables": {
+                        "a": {
+                            "datatype": "integer",
+                            "type": "property_name",
+                            "property_name": "total_expected_doses_taken"
```
proteusvacuum: I don't think this will work any more,

custom/enikshay/ucr/reports/beneficiary_register_v2.json
```diff
@@ -0,0 +1,751 @@
...
+                "type": "expression",
+                "display": "total_expected_doses_taken",
+                "column_id": "total_expected_doses_taken"
+            },
+            {
+                "expression": {
+                    "type": "switch",
+                    "switch_on": {
+                        "type": "property_name",
+                        "property_name": "adherence_schedule_id"
```
proteusvacuum: @ctsims do you think this is a reasonable thing to do? @calellowitz has hardcoded the number of days for each adherence schedule for the report that was previously fetched from the lookup table. I don't forsee this changing any time soon, and we don't currently have a mechanism to fetch data from lookup tables in UCR.

custom/enikshay/ucr/reports/beneficiary_register_v2.json
```diff
@@ -0,0 +1,751 @@
+{
+    "domains": [
+        "enikshay"
+    ],
+    "server_environment": [
+        "enikshay",
+        "softlayer"
+    ],
+    "data_source_table": "episode_v3",
```
proteusvacuum: this should be `episode_v4`

custom/enikshay/ucr/data_sources/person_for_referral_report_v3.json
```diff
@@ -17,19 +17,14 @@
...
-                    "expression": {
-                        "datatype": null,
-                        "type": "property_name",
-                        "property_name": "type"
-                    },
-                    "type": "boolean_expression",
-                    "property_value": "person"
-                }
-            ]
+            "operator": "eq",
```
proteusvacuum: I think this will cause a rebuild for this table, right? I think you should leave this as is then, since it works, and that's the pattern we've been using in the other datasources anyway.

custom/enikshay/ucr/data_sources/episode_v4.json
```diff
@@ -19,19 +19,14 @@
         "table_id": "episode_v4",
         "display_name": "Episode v4",
         "configured_filter": {
-            "type": "and",
-            "filters": [
```
proteusvacuum: will changing the filter still allow this to be updated in place? Otherwise I don't think you should include this change.

proteusvacuum (changes requested):

custom/enikshay/ucr/data_sources/person_for_referral_report_v3.json
calellowitz: oh, didnt realize that would be in here. i merged my branch from https://github.com/dimagi/commcare-hq/pull/18792 to prevent a merge conflict on the other data source. I thought it was already merged, i see it was just approved. but this pattern will make rebuilds work better. @emord can explain better, but basically when there is a simple filter (ie no and) we can only add the right case types to the queue for the rebuild as opposed to going through all cases. this also shouldnt trigger a rebuild either way. I think we only do that when adding a new column. and all the new ones are in place anyway.

custom/enikshay/ucr/data_sources/episode_v4.json
calellowitz: i dont think it will trigger a rebuild, and if it does it should definitely be in place. but this data source is only used in testing anyway, so im not sure its a problem if we have to rebuild in off hours

custom/enikshay/ucr/reports/beneficiary_register_v2.json
calellowitz: this is how we do the other calculations. i think it just references other columns

custom/enikshay/ucr/reports/beneficiary_register_v2.json
calellowitz: one alternative i mention below in the other comment is to have the reconciliation task update `doses_per_week` and add that to the data source. then if the fixture changes the reconciliation task will update all the episodes. that would probably back up the queue again but it would be a one time thing

custom/enikshay/ucr/reports/beneficiary_register_v2.json
calellowitz: good catch

custom/enikshay/ucr/reports/beneficiary_register_v2.json
proteusvacuum: but this case property won't exist anymore, right? You've removed it from the task, and the datasource.

proteusvacuum: or are you saying that `property_name` actually references the `expression`, below?

calellowitz: that is what i am hoping/thinking it will do. otherwise that implies that it fetches every case and does a property name lookup on that case. i assumed it pulled from the column called that because of the error we saw a few months ago, when we noticed that you are required to have a column for each of the context variables in the expression https://github.com/dimagi/commcare-hq/pull/17835

proteusvacuum: Sounds like this can reference other expressions: https://github.com/dimagi/commcare-hq/tree/master/corehq/apps/userreports#expression-columns

custom/enikshay/ucr/data_sources/episode_v4.json
emord: Rebuilds only happen automatically whenever a column is added or removed or the type changes. Filters have no bearing on rebuilds

custom/enikshay/ucr/data_sources/person_for_referral_report_v3.json
emord: as mentioned earlier there's no rebuild for filter changes

custom/enikshay/ucr/data_sources/episode_v4.json
proteusvacuum: Wouldn't filter changes change which documents should be included in the datasource (potentially removing some that were there before, and adding others that had already been looked over)?

emord (commented):
Not a lot of enikshay context here, so just leaving some general comments

> This removes a calculation from the nightly enikshay task, which served mainly to increment a case property by one, but caused 10s of thousands of cases to be updated, backing up ucr queues for enikshay. It pushes that calculation to the UCR itself since its really just date math.

Does this actually stop any updates from happening? from tasks.py it looks like it updates another case property in the lines around it

> There are a lot of calculations in the UCR that should maybe get pushed to the data source, I think only the total_expected_doses_taken one needs to be there. Not sure which is better.

Calculations in the data source is an area that's really been hurting ICDS dashboards. I would try to keep the data sources as simple as possible so they are quicker to rebuild. The more logic and related doc lookups in the data source, the longer it will take to rebuild. This is especially true when we start getting high case loads

I would say we should prioritize placing calculations in the following:
1) case properites - obviously not always feasible, but it's by far the quickest to build in UCRs
2) reports - This would mean we do some form of calculation whenever the report is viewed. There are some limitiations in the reporting json definitions though
3) data source - Our rebuild processes are already very slow and until we do more work to make these fast I think at enikshay/icds scale it makes sense to do as few calculations as possible. The reasons for putting a calculation here, is if you require a value from the latest submitted form or some form of case history or if this will be queried a lot and it would be useful to be indexed.

Unfortunately at this point I think it's pretty easy to do something in a UCR data source that would easily get caught in code review as a "wtf" if you were writing real code. (like the fact that UCRs work on one doc at a time)

custom/enikshay/ucr/data_sources/episode_v4.json
emord: The way we decide to rebuild is to inspect the table that's currently live in the DB and the table that would be created from the UCR config. If those are the same, then we don't rebuild. 

filters are lost in that, but there's actually no way to tell what the old filter was anyways (and thus no way to compare them), because we don't keep old docs around

emord: Another thing that gets lost is if you change a calculation of a column. As long as the data type and column_id stay the same you won't get an automatic rebuild. For example if you change it from a case property lookup to a complex expresion you need to manually kick off the rebuild

calellowitz:
@proteusvacuum I made `doses_per_week` a case property updated by the reconciliation task, so if the fixture changes, those changes will get reflected. This will result in a one time update of every case by the reconciliation task now and each time that fixture changes. Based on the comments from @emord  ] I left the rest of the calculations in the UCR, to take load off the data source. I think I hit all the feedback but let me know if you see something else

calellowitz:
@emord 

> Does this actually stop any updates from happening? from tasks.py it looks like it updates another case property in the lines around it

Yeah, https://github.com/dimagi/commcare-hq/blob/master/custom/enikshay/tasks.py#L763 ensures that we only update case properties that have changed

emord:
> Yeah, https://github.com/dimagi/commcare-hq/blob/master/custom/enikshay/tasks.py#L763 ensures that we only update case properties that have changed and most of the other case properties are pretty static

Even if only one property is updated then the data source will need to be recalculated. I don't see what gains we'll get from this, unless the number of case changes (as in number of cases, not number of case properties)  are decreased.

calellowitz:
@emord 
We found that for the majority (perhaps vast majority, though it was not a particularly rigorous check), the only case property this task updated was `total_expected_doses_taken`. So by removing that property, a large portion of the cases that were being updated, will no longer be updated.

emord:
> So by removing that property, a large portion of the cases that were being updated, will no longer be updated.

Thanks. that wasn't clear to me earlier, but does make it more obvious that there are gains from this. thanks

custom/enikshay/ucr/reports/beneficiary_register_v2.json
proteusvacuum: bump

custom/enikshay/ucr/data_sources/episode_v4.json
proteusvacuum: Got it. Didn't know that before. Thanks.

custom/enikshay/tasks.py
```diff
@@ -488,6 +487,7 @@ def get_aggregated_scores(self, latest_adherence_date, adherence_schedule_date_s
         )
 
         doses_per_week = self.get_doses_per_week()
+        update['doses_per_week'] = doses_per_week
```
proteusvacuum: this doesn't logically make sense to include in the `get_aggregated_scores` method. Can you move it elsewhere? It won't update all the time if left here because of the return value above.

proteusvacuum (approved):

PR #18804, opened 2017-12-05
Fix app preview red dot not appearing when forms/modules are rearranged

orangejenny:
As reported in field-dev. The actual bug fix is just adding rearrange to the list of urls. The rest of the changes make this logic shared between app preview's red dot and app builder's  purple "Updates available to publish" banner.

This code is a little fragile, discussion [here](https://github.com/dimagi/commcare-hq/pull/16308#discussion_r121703824)

@calellowitz / @biyeun 
fyi @wpride


orangejenny:
@calellowitz not urgent, if you can review by the end of the week that's fine

calellowitz (approved):

PR #18806, opened 2017-12-05
Fix owner_ids not matching assigned facility

proteusvacuum:
Somehow we messed up the owner_ids of ~3000 cases: https://manage.dimagi.com/default.asp?266113
Luckily, we stored this data on the case in a different property.

This task checks where this property doesn't match. If the `owner_id` was updated after the `facility_assigned_to` then we assume this was intentional, and don't reset the `owner_id`.

@emord

### DIFF
--- a/corehq/ex-submodules/casexml/apps/case/util.py
+++ b/corehq/ex-submodules/casexml/apps/case/util.py
@@ -186,16 +186,25 @@ def iter_cases(case_ids, strip_history=False, wrap=True):
 def property_changed_in_action(case_transaction, case_id, case_property_name):
     from casexml.apps.case.xform import get_case_updates
     PropertyChangedInfo = namedtuple("PropertyChangedInfo", 'transaction new_value modified_on')
-    update_actions = [
-        (update.modified_on_str, update.get_update_action(), case_transaction)
-        for update in get_case_updates(case_transaction.form)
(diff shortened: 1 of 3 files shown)


corehq/ex-submodules/casexml/apps/case/util.py
```diff
@@ -186,16 +186,25 @@ def iter_cases(case_ids, strip_history=False, wrap=True):
...
-            property_changed = update_action.dynamic_properties.get(case_property_name)
+    include_create_fields = case_property_name in ['owner_id', 'name', 'external_id']
+    case_updates = get_case_updates(case_transaction.form)
+
+    actions = []
+    for update in case_updates:
+        if update.id == case_id:
+            actions.append((update.modified_on_str, update.get_update_action(), case_transaction))
+            if include_create_fields:
+                actions.append((update.modified_on_str, update.get_create_action(), case_transaction))
```
emord: Is it valid to change any of the `include_create_fields` after it's already been created? the particular scenario I'm thinking of is reassigning cases would supply the owner_id in an update action and then with this change also include a second(third,fourth...) create action

proteusvacuum: good point.

emord (approved):

PR #18810, opened 2017-12-06
property filter the forms

snopoke:


calellowitz (approved):
not sure i totally understand but looks fine by me

PR #18817, opened 2017-12-06
make StringIO imports compatible

nickpell:
@dimagi/py3


nickpell:
test failures are real

nickpell:
maybe we need to distinguish between
https://pythonhosted.org/six/#six.StringIO
and
https://pythonhosted.org/six/#six.BytesIO

kaapstorm:
Yeah, [this cheat sheet](http://python-future.org/compatible_idioms.html#stringio) seems to suggest that we should be using `io.BytesIO` for handling byte strings in Python 2.

millerdev:
This is probably going to require a case-by-case analysis to figure out which of `BytesIO` or `StringIO` should be used, and probably custom code in some situations to explicitly do the bytes/text conversion that is currently being done implicitly by Python 2.

nickpell:
closing out of date PR

PR #18819, opened 2017-12-06
Cache more enikshay expressions

esoergel:
@emord @sravfeyn

### DIFF
--- a/custom/enikshay/expressions.py
+++ b/custom/enikshay/expressions.py
@@ -296,16 +296,20 @@ class EpisodeFromPersonExpression(JsonObject):
 
     def __call__(self, item, context=None):
         person_id = self._person_id_expression(item, context)
-        domain = context.root_doc['domain']
         if not person_id:
             return None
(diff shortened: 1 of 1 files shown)


custom/enikshay/expressions.py
```diff
@@ -296,16 +296,20 @@ def configure(self, person_id_expression):
 
     def __call__(self, item, context=None):
         person_id = self._person_id_expression(item, context)
-        domain = context.root_doc['domain']
         if not person_id:
             return None
+        return self._get_open_episode_case_json(person_id, context)
+
+    @staticmethod
```
emord: Does the ucr_context_cahce decorator require these to be staticmethod?

emord (approved):
Would be cool if you have a way to gather metrics about the effects of these changes. I'd be interested to see how impactful each of these changes will be

custom/enikshay/expressions.py
esoergel: Nope, they could also be normal instance methods or module-level functions.

PR #18820, opened 2017-12-06
Use parent_location_id directly

esoergel:
Once upon a time, this wasn't available, so people used lineage as a workaround.
@sravfeyn


sravfeyn:
👍

PR #18823, opened 2017-12-07
fetch treating hospital as the pcp if its available and has nikshay code

mkangia:
Change requested:
To use treating hospital as the pcp if its present and has a nikshay code when sending nikshay registration notifications.

This modifies it for both Nikshay and 99dots.
@proteusvacuum does that sound right to change for 99dots as well?

### DIFF
--- a/custom/enikshay/case_utils.py
+++ b/custom/enikshay/case_utils.py
@@ -405,7 +405,8 @@ def _get_private_locations(person_case, episode_case=None):
                     pcp_location = None
     if not pcp_location:
         try:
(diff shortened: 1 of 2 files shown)


mkangia:
Hey @proteusvacuum, can you review this one?

proteusvacuum:
@mkangia sent an email to 99DOTS. I'm not sure we will want to do this for them, but checking.

mkangia:
@proteusvacuum looks like we are good here to merge?

proteusvacuum:
yup!

PR #18826, opened 2017-12-07
Ce/default basic

calellowitz:
@esoergel @proteusvacuum @snopoke 
https://manage.dimagi.com/default.asp?266722
This fixes a bug where old versions of commcare on android 6.0.1 cannot log in because of a bug in the digest auth libraries being used. This defaults mobile login to basic auth (which i believe is intended anyway).  It reverts some of the changes made to default everything to digest, but I do not believe it should break any functionality that relies on that since it should only affect the three phone endpoints (sync, submission, and keys).
This also revealed an issue with this check https://github.com/dimagi/commcare-hq/blob/master/corehq/apps/domain/auth.py#L58 . Notably the user agent for android phone does not always (or maybe ever) contain android.

This issue is resolved in newer mobile versions, but ICDS is stuck on the old LTS version

### DIFF
--- a/corehq/apps/domain/auth.py
+++ b/corehq/apps/domain/auth.py
@@ -19,7 +19,7 @@ API_KEY = 'api_key'
 TOKEN = 'token'
 
 
-def determine_authtype_from_header(request, default=None):
+def determine_authtype_from_header(request, default=DIGEST):
     """
     Guess the auth type, based on the headers found in the request.
     """
@@ -35,12 +35,10 @@ def determine_authtype_from_header(request, default=None):
 
     # If there is no HTTP_AUTHORIZATION header, we return a 401 along with the necessary
     # headers for digest auth
-    if default is None:
-        default = DIGEST
     return default
 
 
-def determine_authtype_from_request(request, default=None):
+def determine_authtype_from_request(request, default=DIGEST):
     """
     Guess the auth type, based on the (phone's) user agent or the
     headers found in the request.
@@ -51,8 +49,6 @@ def determine_authtype_from_request(request, default=None):
         ANDROID: BASIC,
     }
     user_type = guess_phone_type_from_user_agent(user_agent)
-    if default is None:
-        default = DIGEST
     if user_type is not None:
         return type_to_auth_map.get(user_type, default)
     else:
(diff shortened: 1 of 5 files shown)


corehq/apps/domain/auth.py
```diff
@@ -35,10 +35,10 @@ def determine_authtype_from_header(request):
 
     # If there is no HTTP_AUTHORIZATION header, we return a 401 along with the necessary
     # headers for digest auth
```
snopoke: Comment needs an update

snopoke (approved):

proteusvacuum:
👍 Do we want to remove the useragent check if it isn't doing anything? I think leaving it in might be misleading to future wanderers.

calellowitz:
I left it for the J2ME part but not sure if that is even still relevant. Either way could remove the android part and rename it to something less misleading

corehq/apps/ota/views.py
```diff
@@ -44,7 +43,7 @@
 
 @location_safe
 @handle_401_response
-@login_or_digest_or_basic_or_apikey_or_token()
+@login_or_digest_or_basic_or_apikey_or_token(default=BASIC)
```
esoergel: I believe the problem is that `@login_or_digest_or_basic_or_apikey_or_token(default=BASIC)` cannot work as described.  You can only authenticate with digest auth if it is the default when no information is provided by the client.

We could remove the default argument and make a second decorator that explicitly doesn't support digest.

calellowitz: Is this simply a question of a misleadingly named function, or do you think this breaks some existing functionality. If its the former i am happy to fix in a follow up, but I want to make sure this gets in before i head off to bed, unless you want to take it over. ICDS really needs this fixed tonight.

If you believe this will break something, then i am not sure i understand. The default will be digest except for this view, which i think is only hit by mobile and therefore we want to force basic auth.

esoergel: Cal, Farid, and I offlined about this extensively and I think we're all on the same page now, but for the sake of this PR's history:

I think it's more of a naming question (although with issues beyond just code clarity), and agree that fixing ICDS is much higher a priority than getting this perfect.

PR #18827, opened 2017-12-07
Guessing patient identity: Review / sanity check

kaapstorm:
Hey this is just a sanity check on my approach.

Some projects need a custom way to try to match OpenMRS patients to CommCare cases, if they can't make a definitive match on an ID.

I want a way to keep this custom code outside the Motech code, but still allow the OpenMRS config for a repeater to pull it in. So I'm defining a class to make these guesses, and registering it. The repeater can then select the class in its OpenMRS config.

This code is incomplete, but I hope it gives you a clear idea of where I'm heading. I just wanted to check that this approach not only makes sense, but is also the best way to do this. Please let me know what you think.

@nickpell I think @dannyroberts has a lot more context on this. Please don't feel obliged to try to wrap your head around Motech's OpenMRS repeater just for this. But if you do have an opinion on registering custom classes (and the best way to do that) and allowing users to select them, I'm interested to hear it.

@dannyroberts @nickpell cc @sheelio

### DIFF
--- a/corehq/motech/openmrs/guessers.py
+++ b/corehq/motech/openmrs/guessers.py
@@ -1,0 +1,29 @@
+from collections import defaultdict
+
+
+PATIENT_GUESSERS = []
+
+
+def register_patient_guesser(class_):
+    PATIENT_GUESSERS.append(class_)
+    return class_
+
+
+class PatientGuesserBase(object):
+    """
+    A PatientGuesser is used to guess patient identity if ID matchers
+    fail.
+    """
+
+    def guess_patients(self, requests, case, case_config):
+        """
+        Given a case, search OpenMRS for possible matches. Return the
+        best guesses. If just one guess is returned, it will be chosen.
+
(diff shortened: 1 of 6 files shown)


dannyroberts:
@kaapstorm this seems like a cool and very practical idea. If you want me to validate your thinking on the actually guessing strategy, can you walk me through the weights / threshold things in the actual `possible_health` implementation?

kaapstorm:
Hey @dannyroberts thanks for the feedback. I'm pretty confident in the strategy. (Is that a bad sign? :upside_down_face: ) The weights and thresholds will be based on the [Patient Resolution](https://docs.google.com/document/d/1BXVj6LG4HYh-ltHKKgsAQ6uvq-klbJCVKHp_rLqnAzU/edit?ts=59305082#heading=h.6a4s138rhzdw) section of the project's technical overview. I'd like the properties, weights and thresholds to be easy to tweak, but I'm happy for them to be hardcoded at this point, at least until we use a similar strategy in another project.

PR #18828, opened 2017-12-07
stop heartbeat update user

snopoke:
This seems to have caused a lot of doc update conflicts so reverting for now and adding some logging to try figure out where the conflicts are coming form:

* https://sentry.io/dimagi/commcarehq/issues/410616451/?environment=icds
* https://sentry.io/dimagi/commcarehq/issues/410593323/?environment=icds

### DIFF
--- a/corehq/apps/users/models.py
+++ b/corehq/apps/users/models.py
@@ -1,4 +1,6 @@
 from __future__ import absolute_import
+
+import inspect
 from datetime import datetime
 from dateutil.relativedelta import relativedelta
 import logging
@@ -19,6 +21,7 @@ from corehq.apps.users.permissions import EXPORT_PERMISSIONS
 from corehq.apps.users.const import ANONYMOUS_USERNAME
 from corehq.form_processor.interfaces.supply import SupplyInterface
 from corehq.form_processor.interfaces.dbaccessors import FormAccessors
+from corehq.util.global_request import get_request
 from corehq.util.soft_assert import soft_assert
 from dimagi.ext.couchdbkit import (
     StringProperty,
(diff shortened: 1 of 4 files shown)


emord:
```
ERROR: corehq.apps.ota.tests.test_heartbeat:HeartbeatTests.test_blank_last_sync
  File "/mnt/commcare-hq-ro/corehq/apps/ota/tests/test_heartbeat.py", line 91, in test_blank_last_sync
    self.assertIsNone(device.get_meta_for_app(self.app.get_id).last_sync)
AttributeError: 'NoneType' object has no attribute 'get_meta_for_app'

```
looks real

corehq/apps/users/models.py
```diff
@@ -78,11 +81,34 @@
...
 from six.moves import map
 
+debug_save_logger = logging.getLogger('debug_user_save')
+
 COUCH_USER_AUTOCREATED_STATUS = 'autocreated'
 
 MAX_LOGIN_ATTEMPTS = 5
 
 
+def log_user_save(type_, doc):
```
emord: Is there a reason to include this in master? Seems like it'll only affect local testing for a short while (until this is resolved) so it could be in a branch

emord (approved):
Looks good. just a non blocking comment about logging

corehq/apps/users/models.py
snopoke: Actually I was wanted this deployed to ICDS. I've tried to repro the issue locally and on other envs and haven't been able to.

emord: so you'll change https://github.com/dimagi/commcare-hq/pull/18828/files#diff-bdf3ecebd8379ca98cc89e545fc90899R1289 locally on a webworker on ICDS to actually get it to log? makes sense if that's the way forward

snopoke:
Oh no, that's supposed to say `icds`

PR #18805, opened 2017-12-05
Wait until document is ready to call analytics init functions

orangejenny:
https://manage.dimagi.com/default.asp?266675

This fixes a couple of regressions, though to my knowledge only the first one has been reported:
- can't save groups
- can't delete groups
- can't create a scheduled report
- on mobile workers page, can't update phone number, can't delete user, can't update user info
- can't close case from single case view
- can't restore a form from case history
- can't publish to exchange

This functionality may still be broken for users who have adblockers turned on. I'm working on addressing that.

@nickpell Can this get into today's deploy?
code buddy @calellowitz

PR #18834, opened 2017-12-07
Analytics: handle adblockers

orangejenny:
Handle script failures more gracefully. Right now, if a script doesn't load, analytics callbacks don't run, which breaks a variety of functionality (outlined in https://github.com/dimagi/commcare-hq/pull/18805).

The main change is adding `initApi` to `utils.js`, which creates a promise that will be resolved when all necessary scripts are loaded, or rejected if any one of them fails (or if analytics is disabled). The analytics tracking functions then run inside calls to `done`, with code that should always execute running both in `done` and in `fail`. This replaces the earlier logic that checked for `isEnabled` but not for script loading success.

The other changes just move code around and do some minor refactors: I made loggers private, since they aren't used outside of analytics, and got rid of the `_data` objects, which were making it harder to pull out `initApi` as its own function.

QA has been doing an analytics pass on prod (https://manage.dimagi.com/default.asp?266696); they're going to pause that and instead QA this on staging.

`w=1`

@calellowitz / @biyeun


corehq/apps/analytics/static/analytix/js/drift.js
```diff
@@ -18,53 +18,41 @@ hqDefine('analytix/js/drift', [
 ) {
     'use strict';
     var _get = initialAnalytics.getFn('drift'),
-        _global = initialAnalytics.getFn('global'),
-        logger = logging.getLoggerForApi('Drift'),
-        _data = {},
-        _drift = {};
+        _drift = {},
+        _logger,
+        _ready;
```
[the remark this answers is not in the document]
orangejenny: deliberately not fixing this, too easy for future devs to not realize the return value of `initApi` matters

corehq/apps/analytics/static/analytix/js/drift.js
```diff
@@ -18,53 +18,41 @@ hqDefine('analytix/js/drift', [
...
 
-        _drift.SNIPPET_VERSION = '0.3.1';
-        _data.apiId = _get('apiId');
+            _drift.SNIPPET_VERSION = '0.3.1';
 
-        if (_data.apiId) {
-            _data.scriptUrl = "https://js.driftt.com/include/" + utils.getDateHash() + "/" + _data.apiId + '.js';
-            logger.verbose.log(_data.scriptUrl, "Adding Script");
-            utils.insertScript(_data.scriptUrl, logger.debug.log, {
-                crossorigin: 'anonymous',
```
orangejenny: This line gets lost, which is fine because doesn't actually set the script's attribute (needs to be `crossOrigin` not `crossorigin`). This code originally came from [drift docs](https://docs.stackpile.io/docs/drift).

corehq/apps/analytics/static/analytix/js/hubspot.js
```diff
@@ -15,36 +15,28 @@ hqDefine('analytix/js/hubspot', [
...
+        _ready;
 
     var _hsq = window._hsq = window._hsq || [];
 
-    var __init__ = function () {
-        _data.apiId = _get('apiId');
-        if (_data.apiId) {
-            _data.scriptSrc = '//js.hs-analytics.net/analytics/' + utils.getDateHash() + '/' + _data.apiId + '.js';
-            utils.insertScript(_data.scriptSrc, logger.debug.log, {
-                id: 'hs-analytics',
```
orangejenny: This also gets lost in the refactor, which I'm pretty sure is okay - the id is here because it's supplied in HubSpot's example code, but we never use it and I'm skeptical that the script references its own id. Will find out in testing.

orangejenny:
In testing, realized that at least with my adblocker, the kissmetrics scripts we request are allowed, but those scripts themselves make requests that are blocked. Haven't found a way that we can detect those errors, since we don't have control over those requests, but did confirm that the kissmetrics callbacks run even if an "inner" script is blocked.

orangejenny:
@calellowitz This is the highest priority to review (and, granted, the most complicated).

calellowitz (approved):
Thanks for handling. Adblockers are a continual source of problems around our analytics code and this is definitley a step in the right direction

PR #18837, opened 2017-12-07
Domain Calculations - Handle old exports

sheelio:
@nickpell 
Was noticing an issue where this function was failing for projects with old-style exports (they don't have an is_daily_saved_export property).


nickpell (approved):
I feel like there must be cleaner way to do this but this should work.

PR #18838, opened 2017-12-07
Moved locations task from commtrack to locations.

emord:
Was looking around for the import code for something else and was surprised to see it in commtrack

@esoergel for locations
@proteusvacuum buddy


emord:
including this in a different PR

PR #18841, opened 2017-12-07
Add ANALYTICS_DEBUG setting

orangejenny:
@calellowitz / @biyeun

### DIFF
--- a/corehq/apps/analytics/templates/analytics/initial/drift.html
+++ b/corehq/apps/analytics/templates/analytics/initial/drift.html
@@ -1,9 +1,13 @@
 {% load hq_shared_tags %}
-{% if not request|toggle_enabled:"USER_TESTING_SIMPLIFY" %}
-  {% if is_saas_environment or ANALYTICS_CONFIG.DEBUG %}
-    {% if not request.user.is_authenticated or request.couch_user.days_since_created < 31 or ANALYTICS_CONFIG.DEBUG %}
-      {# This was already publicly visible, best practice: todo get into settings #}
-       {% initial_analytics_data 'drift.apiId' '7456r8nz2u34' %}
-    {% endif %}
-  {% endif %}
+
+{# TODO: API id was already publicly visible, but it should be moved into settings #}
+{% if ANALYTICS_CONFIG.DEBUG %}
+    {# Developer is working on analytics #}
+    {% initial_analytics_data 'drift.apiId' '7456r8nz2u34' %}
+{% elif not is_saas_environment %}
+    {# Don't include: this is an enterprise environment #}
+{% elif request|toggle_enabled:"USER_TESTING_SIMPLIFY" %}
+    {# Don't include: this is a user test and the Drift popup would be a distraction #}
+{% elif not request.user.is_authenticated or request.couch_user.days_since_created < 31 %}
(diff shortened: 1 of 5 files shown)


orangejenny:
I'd appreciate this one getting reviewed soon-ish (higher priority than the other analytics PR, I'm still testing that one plus it'll go through QA).

corehq/apps/analytics/templates/analytics/initial/drift.html
```diff
@@ -1,7 +1,9 @@
 {% load hq_shared_tags %}
-{% if is_saas_environment and not request|toggle_enabled:"USER_TESTING_SIMPLIFY" %}
-  {% if not request.user.is_authenticated or request.couch_user.days_since_created < 31 %}
-    {# This was already publicly visible, best practice: todo get into settings #}
-    {% initial_analytics_data 'drift.apiId' '7456r8nz2u34' %}
+{% if not request|toggle_enabled:"USER_TESTING_SIMPLIFY" %}
+  {% if is_saas_environment or ANALYTICS_CONFIG.DEBUG %}
```
emord: Why not include this in the if statement above? Also `is_saas_environment` and the debug check should come probably come before toggle_enabled since toggles are stored in the DB and these other two are in localsettings (so in local memory)

emord: Could it be rewritten as:

```
if analytics.debug:
    include 
elif not is_saas_environment:
   don't include
elif USER_TESTING_SIMPLIFY:
  don't include
elif user not authed or user created < 31 days:
   include
```

the 3 nests of if statements is getting a bit confusing for me and the django template syntax doesn't help.

orangejenny: https://github.com/dimagi/commcare-hq/pull/18841/commits/b003ac9f61dd386da6c14fa09fd5e81b8bfb5c3a better?

corehq/apps/analytics/templates/analytics/initial/drift.html
```diff
@@ -1,7 +1,13 @@
...
-{% if is_saas_environment and not request|toggle_enabled:"USER_TESTING_SIMPLIFY" %}
-  {% if not request.user.is_authenticated or request.couch_user.days_since_created < 31 %}
-    {# This was already publicly visible, best practice: todo get into settings #}
+
+{# TODO: API id was already publicly visible, but it should be moved into settings #}
+{% if ANALYTICS_CONFIG.DEBUG %}
+    {# Developer is working on analytics #}
+    {% initial_analytics_data 'drift.apiId' '7456r8nz2u34' %}
+{% elif not is_saas_environment %}
+    {# Don't include: this is an enterprise environment #}
```
emord: Inside this elif there is only a comment. Does that work well in django templates? I'm thinking of the following case where python would throw a syntax error

```
if x:
   # blah
elif y:
   do something
```

emord (approved):
Looks good. just a question about if this syntax works. guessing it does though?

Thanks for the comments as well. makes it much more readable.

corehq/apps/analytics/templates/analytics/initial/drift.html
orangejenny: yeah, templates are okay with this syntax (tested)

PR #18842, opened 2017-12-07
Fixed intrahealth pillow for the short term.

emord:
This is not code that I'm proud of, but it gets intrahealth/ipm unblocked on their deadline and it's already running on their server

pna context: @kaapstorm @javierwilson @snopoke 
buddy @proteusvacuum

### DIFF
--- a/corehq/ex-submodules/fluff/pillow.py
+++ b/corehq/ex-submodules/fluff/pillow.py
@@ -172,7 +172,7 @@ def get_multi_fluff_pillow(indicator_classes, name, kafka_topic, delete_filtered
         for cls in indicator_classes
     ]
     domains = list(set(d for cls in indicator_classes for d in cls.domains))
-    doc_types = list(set(cls.wrapper._doc_type for cls in indicator_classes))
+    doc_types = list(set(cls.document_class.doc_type for cls in indicator_classes))
     assert len(doc_types) == 1
 
     return FluffPillow(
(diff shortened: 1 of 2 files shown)


corehq/ex-submodules/fluff/pillow.py
```diff
@@ -71,6 +72,13 @@ def _is_doc_type_deleted_match(self, type):
 
     def change_transform(self, doc_dict):
         delete = False
+        if self.document_class == XFormInstanceSQL:
```
emord: This is nasty

corehq/ex-submodules/fluff/pillow.py
```diff
@@ -20,8 +21,8 @@ def __init__(self, indicator_class, delete_filtered=False):
         self.document_class = indicator_class.document_class
         self.save_direct_to_sql = indicator_class().save_direct_to_sql
         self.deleted_types = indicator_class.deleted_types
-        self.doc_type = indicator_class.document_class._doc_type
         self.wrapper = indicator_class.wrapper or indicator_class.document_class
```
emord: I did a quick grep and didn't see anything in custom pillows that ever used it. I'm guessing htat something old used it and it was never cleaned up

emord (commented):
For a bit more context, the pillow (and fluff in general) was running into errors because it assumes that hte documents live in couch. This changes it to grab the forms/cases from postgres and then it calls `to_json` on it and then it re-wraps it with the couch doc (since there's no wrap method on sql models).

Maybe we can rewrite these document processors in ucr?

kaapstorm (approved):

snopoke (commented):
Not ideal but OK. Are there tests for this in case a SQL model change breaks this?

emord:
> Not ideal but OK. Are there tests for this in case a SQL model change breaks this?

Not at hte moment. I think we need to go back through fluff and get it to use FormAccessors and CaseAccessors to ensure this works well and we can migrate the other domains. There was some deadline today so I just got this working as quickly as possible for kelly and the team

emord:
forgot about this. these aren't used anymore anyways

PR #18843, opened 2017-12-07
Comments on related doc rebuilding

emord:
@dimagi/scale-team here are some notes I came up with from the idea of rebuilding related docs with UCRs this morning. Some assumptions:

1) This is only applicable when using the asynchronous processing feature. Otherwise we assume that these are being processed real time and do not need this
2) We only rebuild related documents that are already in the queue. The reason I think we should do this is so that we don't need to grab all data sources from the domain and run the data source filter on them. We will simply grab the id and the related data source ids that are in the queue.

I've left comments/notes about two approaches, "proactive" and "reactive"

The proactive approach will proactively retrieve related documents when inserting the record into the queue and insert them into an array field. When processing the doc in the celery task it will just grab those related docs from the queue and process those as well

The reactive approach keeps the insertion process the same. It will retrieve the related doc ids when it's processed and the process those in the same task.

One problem I see with this is that there's a good chance for deadlock here. Currently the tasks are split up by the domain and data source's id. So the child health cases and ccs_record cases get processed in different celery tasks. What will happen if we are processing a ccs_record case and it's cooresponding child_health case is being processed in a separate task (and vice versa)?


custom/icds_reports/ucr/data_sources/ccs_record_cases_monthly_tableau2.json
```diff
@@ -3110,6 +3110,24 @@
...
+      "household_case_id": {
+        "comment": "parent of the person case",
+        "rebuild": false
+      },
+      "child_person_cases_id": {
+        "comment": "all person cases under household case",
+        "rebuild": false
+      },
+      "child_health_cases_id": {
+        "comment": "all child_health cases under all child_person_cases. This could be a custom expression in custom/icds_reports and remove the other definitions",
```
emord: This will probably be a nastier version of `icds_parent_parent_id`

corehq/apps/userreports/tasks.py
```diff
@@ -288,6 +289,9 @@ def save_document(doc_ids):
     timer = TimingContext()
     with CriticalSection(lock_keys):
         indicators = AsyncIndicator.objects.filter(doc_id__in=doc_ids)
+        # proactive method: also grab related doc ids from AsyncIndicator and get their locks
+        # note: these only be rebuilt if they are also in the queue
+        # question: What happens if these child docs are being processed by another queue?
```
czue: practically if they are already in another queue we can just not process them, right? since they'll be handled there? is there an easy way in our locking framework to say "try and grab this lock and if you don't succeed just return immediately"?

corehq/apps/userreports/tasks.py
```diff
@@ -305,13 +309,17 @@ def save_document(doc_ids):
 
         with timer:
             for doc in doc_store.iter_documents(doc_ids):
+                # reactive method: grab the related docs here and get their locks
+                # note: these only be rebuilt if they are also in the queue
+                # question: What happens if these child docs are being processed by another queue?
```
czue: same comment as above

custom/icds_reports/ucr/data_sources/ccs_record_cases_monthly_tableau2.json
```diff
@@ -3110,6 +3110,24 @@
         ]
       }
     },
+    "related_docs": {
+      "person_case_id": {
```
czue: these represent ids of indicators inside the config? so they'd be pulled out directly from the row?

czue (commented):
looks interesting. had some inline comments re: deadlock. also wasn't quite sure how the `related_docs` mechanism on the data source would work.

do you have an intuition between proactive and reactive? reactive seems simpler if there aren't any issues with it. one thing I was wondering was whether we could handle the "a child was created after the initial row went into the queue" thing in reactive mode. if so, and if we can pull the related docs directly out of the row without having to do any extra I/O then it seems like it's globally easier/safer to do it that way.

generally feels pretty do-able! wonder if we could :racehorse: this out

corehq/apps/userreports/tasks.py
emord: Yeah. that make sense. I think we can use `CriticalSection(blocking=False)` to do that sort of thing.

custom/icds_reports/ucr/data_sources/ccs_record_cases_monthly_tableau2.json
emord: I imagine this as an expression that calculates the the id from the root document, so it'd look something like https://github.com/dimagi/commcare-hq/blob/5298a6c541c90274cab09154e5e2b00d2b2363a4/custom/ucr_ext/expressions.py#L239-L270

We'd probably want to do something custom if there's any chance to optimize this though so that we can get the child health cases from the ccs record case and vice versa. Right now the hierarchy looks like:
```
                 household
                    /          \
              person      person
              /                     \ 
          ccs_record        child_health
````

There's are some more cases than that. also important to note that a person doesn't always have a child_health or ccs_record and there can be multiple ccs_records and child_health in a household, so we'd need to figure out how to handle possibly N docs from one doc change (this is probably the biggest flag I have here)

**Comment by emord** [comment]
> do you have an intuition between proactive and reactive? reactive seems simpler if there aren't any issues with it. 

Agreed. I think I prefer the reactive mode. 

> one thing I was wondering was whether we could handle the "a child was created after the initial row went into the queue" thing in reactive mode. 

I think that this would work well in the reactive mode. I was imagining this would happen, but let me know if i'm misunderstanding you.

1) ccs_record is put into the queue with no child_health case
2) 100k case put into the queue
3) ccs_record is updated with a child_health case. The child case is behind ccs_record by 100k cases
4) 100k cases are put into the queue
5) ccs_record is popped from the queue
6) The celery worker processing it fetches the associated child_health cases and pops those from the queue
7) both cases are  processed using their most up to date data and removed from the queue

> if so, and if we can pull the related docs directly out of the row without having to do any extra I/O then it seems like it's globally easier/safer to do it that way.

I don't think there's a way to do this without extra DB calls to get the case hierarchy. it will need to happen in either in the pillow or in the final processing. None of our current tables get the child health cases for the ccs_record or vice versa

File: custom/icds_reports/ucr/data_sources/ccs_record_cases_monthly_tableau2.json (hunk 4 above)
Comment (czue) [reply custom/icds_reports/ucr/data_sources/ccs_record_cases_monthly_tableau2.json]: > there can be multiple ccs_records and child_health in a household, so we'd need to figure out how to handle possibly N docs from one doc change (this is probably the biggest flag I have here)

Is this concern a performance / load one or an implementation one? On the implementation side, if this is a band aid on top of a band aid and the long term plan is to remove both (which I think is what I heard in the meeting) then I imagine we could go as far as to hard code it for each data source if we needed to and just return a list/generator, no? That would be if we didn't have an easy way to return a list of stuff in vanilla UCR (which we probably could as well)

Comment (emord) [reply custom/icds_reports/ucr/data_sources/ccs_record_cases_monthly_tableau2.json]: It was more of a performance/load one. This makes it harder to reason about the performance since the question of performance is no longer "how long does it take for this case to get this data?", but "how long does it take this case **and it's subcases** to get it's data **and it's subcases data**". And that is complicated because it can be a many-to many relationship (from my understanding of the app).

Another thing that this made me think of is what happens in the following case:

```
                               household 
             /                 /                  \                         \
   person           person               person              person
         /                     /                       \                            \ 
ccs_record          ccs_record        child_health          child_health

```

I don't know if that's a real, valid hierarchy that is expected, but I've seen some test cases like that during QA. So we'd need to recalculate all ccs_records and child_health records under a household to not risk discrepancies. Originally I only planned on evaluating all child_health cases connected to a ccs_record and vice versa, but I think it's possible to have a child_health case not related to the current ccs_record under the household

Comment (czue) [reply custom/icds_reports/ucr/data_sources/ccs_record_cases_monthly_tableau2.json]: Might be good to quickly sync up over slack/voice if you think this is a concern (and we're moving forward on this). I don't feel that I have a good sense of how this would cause issues. In my head this is just re-ordering work that was all going to happen anyways, but it sounds like there's a concern/constraint you have around knowing how much work you might have to do up front that I don't totally understand. Guessing related to some nuance of how the async indicators function?

**Comment by emord** [comment]
@czue I abandoned the original generic approach and it's pretty custom for ICDS. It ended up being a pretty small change (~50 lines) this way so I'm more confident handing it off. Let me know what you think.

note I'm longer trying to evaluate the related docs in the same task as the parent doc. Instead it's getting the doc and queueing it up. This means that there will still be a discrepancy, but that discrepancy should last for closer to an hour instead of days.

**Comment by emord** [comment]
@czue I think this is ready for review. Let me know what you think

File: custom/icds_reports/ucr/expressions.py [hunk 5]
```diff
@@ -781,3 +781,36 @@ def get_app_version(spec, context):
...
+                                      \---->  child_health
+
+    This could likley be optimized pretty easily to take in multiple case_ids but we
+    don't currently have an interface to do so
+    """
+    db = CaseAccessors('icds-cas')
+    current_case = db.get_case(case_id)
+    if not current_case or not current_case.indices:
+        return []
+    person_case = db.get_case(current_case.indices[0].referenced_id)
```
czue: seems like you are making assumptions about the case here that should be checked? e.g. that its type is one of the expected ones.

custom/icds_reports/ucr/expressions.py
```diff
@@ -781,3 +781,36 @@ def get_app_version(spec, context):
 
 def datetime_now(spec, context):
     return DateTimeNow.wrap(spec)
+
+
+def get_related_docs_ids(case_id):
```
czue: could potentially rename this function as well to make it more clear that it's a very specific one-off

custom/icds_reports/ucr/expressions.py
```diff
@@ -781,3 +781,36 @@ def get_app_version(spec, context):
...
+    """
+    db = CaseAccessors('icds-cas')
+    current_case = db.get_case(case_id)
+    if not current_case or not current_case.indices:
+        return []
+    person_case = db.get_case(current_case.indices[0].referenced_id)
+    if not person_case or not person_case.indices:
+        return []
+    house_case = db.get_case(person_case.indices[0].referenced_id)
+    if not house_case:
```
czue: you could probably also check the case type here to avoid doing extra work if it's not what you're expecting

custom/icds_reports/ucr/expressions.py
czue: Also could be nice to have a couple tests for this function if not too hard.

czue (approved):
This looks good - don't think there are any blockers but I did have some suggestions and questions.

emord (commented):
I think I was able to respond to everything. Added some tests and changed the function/variable names

czue:
this looks good to me. assume you'll merge whenever you're happy

PR #18844, opened 2017-12-07
Fix Kissmetrics data flow

orangejenny:
Fixes issue 1 [here](https://docs.google.com/document/d/1EJazRzP8CfGqG6SkiR9VW5bL3863MJvwopOoQhQJjB0/edit). We've been sending events, but the funnel reports are broken, looks like it's because the logic that identifies users in kissmetrics was lost during https://github.com/dimagi/commcare-hq/pull/18615 

This PR replicates the [old logic](https://github.com/dimagi/commcare-hq/blob/jls/pre-analytics-refactor/corehq/apps/hqwebapp/templates/hqwebapp/includes/analytics_all.html#L134-L142) in the new code.

This should go into the next deploy.

@biyeun / @calellowitz

### DIFF
--- a/corehq/apps/analytics/static/analytix/js/kissmetrix.js
+++ b/corehq/apps/analytics/static/analytix/js/kissmetrix.js
@@ -66,6 +66,18 @@ hqDefine('analytix/js/kissmetrix', [
             _addKissmetricsScript('//doug1izaerwt3.cloudfront.net/' + _init.apiId + '.1.js');
         }
 
+        // Identify user and HQ instance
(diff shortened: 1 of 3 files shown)


gcapalbo (approved):

gcapalbo:
is there anything here that should be environment specific?

orangejenny:
@gcapalbo Not in this PR, the `_global('isEnabled')` checks in kissmetrix.js prevent this from being executed in non-SaaS environments.

PR #18845, opened 2017-12-07
Async UCR metric for average time behind

emord:
@calellowitz you'll probably be interested. This takes the 100 earliest created docs in the queue and averages the lag. This way any random/intermittent errors that occur and increase the "oldest created indicator" metric will hopefully be averaged out and this will give a more real metric. For example, right now on ICDS the first one in the queue is from 1 AM and the 6th is from 5 PM so 5 records that had some (likely db related) error are raising that metric by 16 hours

@proteusvacuum

### DIFF
--- a/corehq/apps/userreports/tasks.py
+++ b/corehq/apps/userreports/tasks.py
@@ -1,4 +1,5 @@
 from __future__ import absolute_import
+from __future__ import division
 from collections import defaultdict
 from datetime import datetime, timedelta
 import logging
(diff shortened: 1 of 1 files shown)


proteusvacuum (approved):

corehq/apps/userreports/tasks.py
```diff
@@ -367,16 +368,25 @@ def _save_document_helper(indicator, doc):
...
     if oldest_indicator and oldest_indicator.date_queued:
-        lag = (datetime.utcnow() - oldest_indicator.date_queued).total_seconds()
+        lag = (now - oldest_indicator.date_queued).total_seconds()
         datadog_gauge('commcare.async_indicator.oldest_queued_indicator', lag)
 
-    indicator = AsyncIndicator.objects.first()
-    if indicator:
-        lag = (datetime.utcnow() - indicator.date_created).total_seconds()
+    oldest_100_indicators = AsyncIndicator.objects.all()[:100]
+    oldest_indicator = oldest_100_indicators[0]
```
dannyroberts: This is causing this error https://sentry.io/dimagi/commcarehq/issues/416664542/. PRing this fix which I think is all we want: https://github.com/dimagi/commcare-hq/pull/18969.

PR #14999, opened 2017-02-16
Revert "Revert "Anonymous Web Apps usage""

benrudolph:
Reverts dimagi/commcare-hq#14997

PR #18846, opened 2017-12-08
Auth decorators

esoergel:
This PR is an attempt to clean up the logic around authentication methods.  Context is provided in the retro email I'll send out soon.

### DIFF
--- a/corehq/apps/domain/auth.py
+++ b/corehq/apps/domain/auth.py
@@ -44,28 +44,14 @@ def determine_authtype_from_request(request, default=DIGEST):
     headers found in the request.
     """
     user_agent = request.META.get('HTTP_USER_AGENT')
-    type_to_auth_map = {
-        J2ME: DIGEST,
-        ANDROID: BASIC,
-    }
-    user_type = guess_phone_type_from_user_agent(user_agent)
-    if user_type is not None:
-        return type_to_auth_map.get(user_type, default)
-    else:
-        return determine_authtype_from_header(request, default)
+    if is_probably_j2me(user_agent):
+        return DIGEST
+    return determine_authtype_from_header(request, default)
 
 
-def guess_phone_type_from_user_agent(user_agent):
-    """
-    A really dumb utility that guesses the phone type based on the user-agent header.
-    """
+def is_probably_j2me(user_agent):
     j2me_pattern = '[Nn]okia|NOKIA|CLDC|cldc|MIDP|midp|Series60|Series40|[Ss]ymbian|SymbOS|[Mm]aemo'
-    if user_agent:
-        if re.search(j2me_pattern, user_agent):
-            return J2ME
-        elif 'Android' in user_agent:
-            return ANDROID
-    return None
+    return user_agent and re.search(j2me_pattern, user_agent)
 
 
 def get_username_and_password_from_request(request):
--- a/corehq/apps/domain/auth.py
+++ b/corehq/apps/domain/auth.py
@@ -19,15 +19,16 @@ API_KEY = 'api_key'
 TOKEN = 'token'
 
 
-def determine_authtype_from_header(request, support_non_j2me_digest=True):
+def determine_authtype_from_header(request, default=DIGEST):
     """
     Guess the auth type, based on the headers found in the request.
 
-    CommCare mobile typically sends an unauthenticated request first, and
-    expects a basic auth challenge in response. This is at odds with the spec
-    for digest auth, in which the client sends an unauthenticated response and
-    receives a digest auth challenge. This means we can't support both CommCare
-    mobile and digest auth at the same endpoint.
+    If default is set to something other than DIGEST, digest auth will not work
+
+    CommCare mobile sends an unauthenticated request first, and we need to
+    issue a basic auth challenge in response. This means we can't support both
+    CommCare mobile and digest auth at the same endpoint (since digest auth
+    requires a digest auth challenge).
 
     For non-mobile endpoints (such as APIs), we can support basic, digest,
     token, and apikey by defaulting to digest, since in all other cases, the
@@ -36,18 +37,18 @@ def determine_authtype_from_header(request, support_non_j2me_digest=True):
     auth_header = (request.META.get('HTTP_AUTHORIZATION') or '').lower()
     if auth_header.startswith('basic '):
         return BASIC
+    elif auth_header.startswith('digest '):
+        # Note: this will not identify initial, uncredentialed digest requests
+        return DIGEST
     elif auth_header.startswith('token '):
         return TOKEN
     elif all(ApiKeyAuthentication().extract_credentials(request)):
         return API_KEY
 
-    if support_non_j2me_digest:
-        # return a 401 along with the necessary headers for digest auth
-        return DIGEST
-    return BASIC
+    return default
 
 
-def determine_authtype_from_request(request, support_non_j2me_digest=True):
+def determine_authtype_from_request(request, default=DIGEST):
     """
     Guess the auth type, based on the (phone's) user agent or the
     headers found in the request.
@@ -55,7 +56,7 @@ def determine_authtype_from_request(request, support_non_j2me_digest=True):
     user_agent = request.META.get('HTTP_USER_AGENT')
     if is_probably_j2me(user_agent):
         return DIGEST
-    return determine_authtype_from_header(request, support_non_j2me_digest)
+    return determine_authtype_from_header(request, default)
 
 
 def is_probably_j2me(user_agent):
--- a/corehq/apps/domain/auth.py
+++ b/corehq/apps/domain/auth.py
@@ -19,26 +19,35 @@ API_KEY = 'api_key'
 TOKEN = 'token'
 
 
-def determine_authtype_from_header(request, default=DIGEST):
+def determine_authtype_from_header(request, support_non_j2me_digest=True):
     """
     Guess the auth type, based on the headers found in the request.
+
+    CommCare mobile typically sends an unauthenticated request first, and
+    expects a basic auth challenge in response. This is at odds with the spec
+    for digest auth, in which the client sends an unauthenticated response and
+    receives a digest auth challenge. This means we can't support both CommCare
+    mobile and digest auth at the same endpoint.
+
+    For non-mobile endpoints (such as APIs), we can support basic, digest,
+    token, and apikey by defaulting to digest, since in all other cases, the
+    client should send the initial request with an Authorization header.
     """
     auth_header = (request.META.get('HTTP_AUTHORIZATION') or '').lower()
     if auth_header.startswith('basic '):
         return BASIC
-    elif auth_header.startswith('digest '):
-        return DIGEST
     elif auth_header.startswith('token '):
         return TOKEN
     elif all(ApiKeyAuthentication().extract_credentials(request)):
         return API_KEY
 
-    # If there is no HTTP_AUTHORIZATION header, we return a 401 along with the necessary
-    # headers for digest auth
-    return default
+    if support_non_j2me_digest:
+        # return a 401 along with the necessary headers for digest auth
+        return DIGEST
+    return BASIC
 
 
-def determine_authtype_from_request(request, default=DIGEST):
+def determine_authtype_from_request(request, support_non_j2me_digest=True):
     """
     Guess the auth type, based on the (phone's) user agent or the
     headers found in the request.
@@ -46,7 +55,7 @@ def determine_authtype_from_request(request, default=DIGEST):
     user_agent = request.META.get('HTTP_USER_AGENT')
     if is_probably_j2me(user_agent):
         return DIGEST
-    return determine_authtype_from_header(request, default)
+    return determine_authtype_from_header(request, support_non_j2me_digest)
 
 
 def is_probably_j2me(user_agent):
--- a/corehq/apps/ota/views.py
+++ b/corehq/apps/ota/views.py
@@ -25,7 +25,7 @@ from corehq.apps.case_search.utils import CaseSearchCriteria
 from corehq.apps.domain.decorators import (
     login_or_digest_or_basic_or_apikey,
     check_domain_migration,
-    login_or_digest_or_basic_or_apikey_or_token,
+    login_or_mobile_or_basic_or_apikey_or_token,
 )
 from corehq.apps.domain.models import Domain
 from corehq.apps.es.case_search import flatten_result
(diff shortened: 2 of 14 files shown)


corehq/apps/domain/auth.py
```diff
@@ -19,53 +19,48 @@
...
-def determine_authtype_from_header(request, default=DIGEST):
+def determine_authtype_from_header(request, support_non_j2me_digest=True):
     """
     Guess the auth type, based on the headers found in the request.
+
+    CommCare mobile typically sends an unauthenticated request first, and
+    expects a basic auth challenge in response. This is at odds with the spec
+    for digest auth, in which the client sends an unauthenticated response and
+    receives a digest auth challenge. This means we can't support both CommCare
+    mobile and digest auth at the same endpoint.
```
czue: I'm not sure I understand how this statement can be true if the mobile endpoints support both. You're saying they don't support "real" digest? Just this bizarre commcare mobile digest frankenstein thing?

corehq/apps/domain/auth.py
```diff
@@ -19,53 +19,48 @@
...
+
+    For non-mobile endpoints (such as APIs), we can support basic, digest,
+    token, and apikey by defaulting to digest, since in all other cases, the
+    client should send the initial request with an Authorization header.
     """
     auth_header = (request.META.get('HTTP_AUTHORIZATION') or '').lower()
     if auth_header.startswith('basic '):
         return BASIC
-    elif auth_header.startswith('digest '):
-        return DIGEST
```
czue: I don't think I understand why this was removed. if the header is specifically asking for digest why would we not return it?

corehq/apps/domain/auth.py
```diff
@@ -19,53 +19,48 @@
 TOKEN = 'token'
 
 
-def determine_authtype_from_header(request, default=DIGEST):
+def determine_authtype_from_header(request, support_non_j2me_digest=True):
```
czue: Might be me, but I find this renaming of the parameter to be more confusing than not. Perhaps we could create two additional functions that call a private function with the right params but leave this as-is?

czue (commented):
thanks for this @esoergel and apologies for what was probably quite a confusing journey since I think I authored a lot of this code originally to solve one of the worst and longest running bugs of all time (the secure submissions [apache 502 issue](https://docs.google.com/document/d/1q363dxTsmSJ57kk3_fW504j3THacs0CZI2XSXoTca_Q/edit)) and probably did it haphazardly.

I'm actually not totally sure I understand the purpose of this PR. Is it fixing something or just trying to clean up the current state of affairs?

My thoughts about a future world that would be good to live in would be that we should have a very specific and limited set of well-tested decorators that clearly articulate what their purposes are. I think that good names for these would be something like:

`@support_mobile_auth`
`@support_web_auth`
`@support_mobile_and_web_auth`

These could just be aliases for `login_or_digest_or_basic_or_token_or_smoke_signals` etc. but I believe that naming them that way will:

1. Force a deliberate choice to be made about every decorated view, and make it much more intuitive what to choose.
2. Make it easier to spot mis-tagged endpoints
3. Reduce the number of different decorators used (and therefore unexpected behaviors introduced).

Of course under the hood of these things we will want a very consistent set of well-tested criteria for exactly what it means to "support mobile auth". 

I'm not necessarily saying all of the above needs to happen in the context of this PR, but I believe that if we feel this is a potential future source of high-priority-hard-to-discover errors then it might be worth taking some time/care to try and move to that world while we're changing things?

corehq/apps/domain/auth.py
calellowitz: +1 to @czue seems like we should keep it and document that so that if someone needs to use digest on an endpoint that defaults to basic auth they can

corehq/apps/domain/auth.py
calellowitz: I think this comment is also misleading. I dont think the phone does anything wrong. Sending an unauthenticated request, receiving for a 401, then attempting to authenticate in the method requested is I believe a very normal request cycle. The issue is determining what the correct response from the server is. When we receive an unauthenticated request do we prompt for basic or digest? The issue isn't really mobile or digest, it is clients that can only auth by one method and do not tell us that method. It seems as wrong to assume an unauthenticated request is really a request for digest auth as it is to assume it is a request for basic auth. The crux of the matter is that in instances when the client doesn't tell us, we can only support one type of auth and we just need to decide what that is and be explicit about it. In order to help make that decision we can use the user agent, like we do for j2me, or a header, like the authorization ones below, but I think it still makes sense to think of those as a default for when we are not sure.

corehq/apps/domain/auth.py
calellowitz: as per my comment below i think it is clearer to call this default

corehq/apps/ota/views.py
```diff
@@ -238,7 +238,7 @@ def get_restore_response(domain, couch_user, app_id=None, since=None, version='1
     return restore_config.get_response(), restore_config.timing_context
 
 
-@login_or_digest_or_basic_or_apikey()
+@login_or_digest_or_basic_or_apikey
```
calellowitz: i think these all need to support mobile (though it looks like ones i left out of my initial pr)

corehq/apps/domain/auth.py
```diff
@@ -19,53 +19,48 @@
...
+
+def is_probably_j2me(user_agent):
     j2me_pattern = '[Nn]okia|NOKIA|CLDC|cldc|MIDP|midp|Series60|Series40|[Ss]ymbian|SymbOS|[Mm]aemo'
-    if user_agent:
-        if re.search(j2me_pattern, user_agent):
-            return J2ME
-        elif 'Android' in user_agent:
-            return ANDROID
-    return None
+    return user_agent and re.search(j2me_pattern, user_agent)
```
calellowitz: thanks for cleaning this up

calellowitz (changes requested):
with the exception of user agent handling, this does not feel any clearer (or more correct) than what was there before. it is still really confusing tracking down exactly what is going on

snopoke (commented):
Generally agree with @czue and @calellowitz. I think we should try and be more explicit about this and only deviate from that under special circumstances e.g. j2me phone, specified in auth header.

For J2ME can't we use the url that the request came from (since J2ME has a separate endpoint in nginx)?

One thing I'm still not quite sure of is what precipitated the change in the first place.

corehq/apps/domain/auth.py
esoergel: I removed this because I thought it was misleading, but maybe it'd be better to keep it in there with a comment.  I'll add it back.  The first request in a digest auth transaction doesn't contain this header, so it's not a reliable way of determining that the client intends to use digest auth.  However, the second request _does_ include this header, so it seems useful to keep this here for that.

Cal, are you saying we should keep it in and ask clients to manually insert a header that matches this?  It does seem reasonable to require an explicit declaration of what type of auth they want to use, but I think we should do something separate, so as not to confuse the digest auth spec.  The `secure_post` endpoint accepts an `authtype` GET param - we could use that, or make a separate authtype header or something.

esoergel (commented):
@czue Thanks for the feedback.  My goal with this PR was to remove some of the ambiguity around what authtypes are supported where, without actually changing HQ's behavior.  In particular, `@login_or_digest_or_basic_or_apikey(default=BASIC)` looks like it supports digest auth, but it does not.  I do think this needs a followup PR to support other mobile endpoints.

I also agree with what you're saying about renaming the decorators, I did something like that initially, then walked it back because it felt too confusing to do with these other changes.  I'd like to add that in now, though.  Do you know why token auth is supported only in the restore?  Can we just add that to the other one?

Ideally we'd be able to get this down to two decorators:
* `@support_mobile_auth` - Supports basic, apikey, and token auth, and digest for j2me devices only.  Defaults to basic if no credentials are provided.
* `@support_web_auth` (or `@support_any_auth`?) - Supports basic, digest, apikey, and token auth.  Defaults to digest if no credentials are provided.
Do you agree with those two categories?  How about the names?

One last thought - these decorators only check that users have a domain login - there's no additional permissions check.  I'm not sure that's well reflected anywhere either.

emord:
note on some cleanup we could do here: it looks like token auth is only used in anonymous web apps (introduced https://github.com/dimagi/commcare-hq/pull/14999) and since that's internal I think we could make that be the same as apikey. I'm not sure why apikey wasn't used in the first place. it looks exactly the same (just replace token with apikey in the header http://www.django-rest-framework.org/api-guide/authentication/#tokenauthentication).

ApiKey comes from tastypie and it's externally documented so we'd need to keep that around

corehq/apps/domain/auth.py
czue: Ok, thanks for these clarifications - I think I have a much better sense of why this is complex/hard now (and sorry it took this long for me to get there).

Thinking about the current situation it seems like it's something like this:

| Default Challenge | Supports Digest? | Supports Basic? | Needed For |
| --- | --- | --- | --- | 
| Basic | Via non-standard header / urlparam only | Yes | Mobile Endpoints |
| Digest | Yes | Via standard auth header or nonstandard user agent header/urlparam | ??? |

And I think what you've found is that for obscure legacy support reasons we now find ourselves in a situation where we have different endpoints that need to be configured in both of the above ways due to reasons that a) live outside our control, and b) aren't particularly sensical. Does that sound right? And can you remind me what the requirement is for the bottom line where the "???"s are?

I think maybe we could just create two sensibly-named methods to represent these two configurations and then not change the inner code (and just make it private and well documented essentially)? I do agree with @calellowitz about the default argument to this particular function, despite the odd behavior that it creates.

Finally, (and sorry this probably isn't the right place for this comment) - is there any reason not to support sessions/tokens/etc. in all endpoints? Then I think we can make it so that we _only_ need these two decorators.

esoergel: That's my understanding, yes.  Thanks for the coherent summary!

The second row above is needed for API endpoints - both our standard tastypie stuff and custom/one-off code.  It's also not clear exactly which of those we do need for a given endpoint.  We could certainly require API consumers to use a particular type of auth, although I'm guessing we have a variety of auth types currently in use, so that sounds hard to change.

> Finally, (and sorry this probably isn't the right place for this comment) - is there any reason not to support sessions/tokens/etc. in all endpoints?

No idea.

This all sounds good.  I'll roll back some of my overzealous changes and rename these two decorators to something sensible.  I think I'll also keep the third one with token auth separate for now and consider that a separate piece of tech debt to clean up as @emord described.  For the two main ones, how about:
`@mobile_auth`, `@api_auth`
Other possibility:
`@require_mobile_login`, `@require_api_login`
Also open to suggestions - not really particular on the exact names.

calellowitz: > Finally, (and sorry this probably isn't the right place for this comment) - is there any reason not to support sessions/tokens/etc. in all endpoints?

I think @sravfeyn removed session auth on purpose from all the api endpoints for one of the security audits we had to do for ICDS/eNikshay. I am not sure of the rationale but I believe we do not want session auth on all endpoints. I see no reason token auth should not be.

The rest sounds great

calellowitz: >  We could certainly require API consumers to use a particular type of auth, although I'm guessing we have a variety of auth types currently in use, so that sounds hard to change.

i could be way off in this but my understanding was that most clients could handle either type of auth. Things like curl, or any programming language, should not only be able to do both, but any solid http library should handle both transparently without bothering the programmer. That in fact is why we had this problem to begin with, the phone was willing to do whatever auth we asked of it, so it switched from basic to digest without anyone knowing, it just had a bug in its digest implementation. Which is all to say, im sure we will break someone's workflow because what doesn't, but I think this may be less of a drastic break than we expect, and whatever eNikshay partner required digest and could not basic auth, may be more the outlier than the rule.

corehq/apps/domain/auth.py
calellowitz: I think it is more about giving them the option to insert a header like that, but generally yes. Either they learn to handle whatever auth we throw at them, something most code probably already can, since its likely using some http library that does auth transparently without burdening the programmer with protocol details, or if they require a certain type they can add that header, and we promise to support them. That is not an ideal world, but one that probably can make almost everyone happy.

calellowitz: > The secure_post endpoint accepts an authtype GET param - we could use that, or make a separate authtype header or something.
I like the header "Authentication" that we already use most places better than the get param, because it seems more correct for reasons I cannot quite articulate. Get params seem like something that should affect the view/response, while headers are something that affect the http details like authentication type. In my mind you shouldn't get the same response from using opposite get params (like two different types of auth) but the header should not affect the response so it is reasonable to choose auth types that way.

corehq/apps/domain/auth.py
esoergel: @calellowitz good point.  And in the case of enikshay, it's a custom endpoint anyways, so it could also support _only_ digest auth if we wanted.

esoergel:
Just made a few changes in line with the discussion here, lemme know what you think.  Also open to naming suggestions for these two decorators.

I also took the opportunity to look at the views currently using the default-to-digest decorator and switched over the ones that were clearly mobile-focused.  I'll PR that separately though, so as to not muddy the waters here.  This is that commit: https://github.com/dimagi/commcare-hq/commit/d57d95e007c888f41677671a83e63f41d8f1a5f7

corehq/apps/domain/auth.py
```diff
@@ -22,19 +22,29 @@
...
+
+    For non-mobile endpoints (such as APIs), we can support basic, digest,
+    token, and apikey by defaulting to digest, since in all other cases, the
+    client should send the initial request with an Authorization header.
     """
     auth_header = (request.META.get('HTTP_AUTHORIZATION') or '').lower()
     if auth_header.startswith('basic '):
         return BASIC
     elif auth_header.startswith('digest '):
+        # Note: this will not identify initial, uncredentialed digest requests
```
czue: :+1:

czue (approved):
Nice @esoergel this looks great to me, thanks for being patient with me.

One last thought I had would be to make all the other decorators "private" - i.e. prefix the name with an `_`. Not sure if you've already managed to grep the codebase for all of them or there are still a few corners we need to revisit?

corehq/apps/domain/auth.py
czue: I imagine there's a world in which we go through a lot of trouble to remove those requirements via proactive comms, soft asserts, and support, but wonder whether the ROI is worth it. living in a two-decorator world seems like it might not be too bad

esoergel:
They are each used in a handful of places.  I'm nervous about changing the way any of these endpoints do auth, although I guess if it's strictly additive that shouldn't be an issue.  I think I'd like to do that separately, maybe one-at-a-time so we can rollback easily if needed.

esoergel:
@calellowitz Any remaining issues with this PR?

calellowitz (approved):
looks good to me

PR #18849, opened 2017-12-08
Example of spec representation

emord:
@sheelio @esoergel you guys both have interest in this so this is what I had in mind. It'll get tricky to explain these in a good way once you start having sub expressions. I'm not sure how I would word that.

There would also need to be a view that goes through each column and prints out each of these.

```
In [1]: from corehq.apps.userreports.expressions.specs import PropertyNameGetterSpec

In [2]: prop = PropertyNameGetterSpec(type='property_name', property_name='propname', datatype='string')

In [3]: print prop
Case's propname property with a datatype of string
```

### DIFF
--- a/corehq/apps/userreports/expressions/specs.py
+++ b/corehq/apps/userreports/expressions/specs.py
@@ -57,6 +57,9 @@ class PropertyNameGetterSpec(JsonObject):
     property_name = DefaultProperty(required=True)
     datatype = DataTypeProperty(required=False)
 
+    def __str__(self):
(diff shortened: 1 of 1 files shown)


emord:
Want to add a note that this PR is just me sharing my idea for this. not signing up to take it to completion (especially since holidays are coming up). either of you (or anyone else) should feel free to throw away this implementation and replace it with something else

emord:
@sheelio started running with this so closing it out

PR #18027, opened 2017-09-29
Mk reconciliation migrations

mkangia:
https://trello.com/c/iedwBirn/204-model-reconciliation-for-public-sector

3 down/ 1 more to go.

PR #18850, opened 2017-12-08
Mk update reconciliation report

mkangia:
following up an improvement in #18027

This adds ability to process on specific cases by passing person case ids. The view ensures that the case ids are for open person cases and notifies the user in case any id is not.

### DIFF
--- a/custom/enikshay/management/commands/drug_resistance_reconciliation.py
+++ b/custom/enikshay/management/commands/drug_resistance_reconciliation.py
@@ -1,4 +1,4 @@
-from __future__ import absolute_import
+from __future__ import absolute_import, print_function
 from collections import defaultdict
 
 from django.core.management.base import CommandError
--- a/custom/enikshay/management/commands/drug_resistance_reconciliation.py
+++ b/custom/enikshay/management/commands/drug_resistance_reconciliation.py
@@ -12,6 +12,7 @@ from custom.enikshay.case_utils import (
 from custom.enikshay.management.commands.base_model_reconciliation import (
     BaseModelReconciliationCommand,
     DOMAIN,
+    get_all_occurrence_case_ids_from_person,
 )
 from custom.enikshay.exceptions import ENikshayCaseNotFound
 from custom.enikshay.const import (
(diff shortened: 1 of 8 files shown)


custom/enikshay/management/commands/drug_resistance_reconciliation.py
```diff
@@ -145,22 +148,31 @@ def reconcile_drug_resistance_cases(self, drug_resistance_cases, drug_id, occurr
...
-                .using(db)
-                .filter(domain=DOMAIN, type="occurrence", closed=False)
-                .values_list('case_id', flat=True)
-            )
-            num_case_ids = len(case_ids)
-            if self.log_progress:
-                print("processing %d docs from db %s" % (num_case_ids, db))
-            for i, case_id in enumerate(case_ids):
-                yield case_id
+        if self.person_case_ids:
```
esoergel: If person_case_ids is an empty list, should this show all or none?  Wondering if this line should be:
`if self.person_case_ids is not None:`
(or whether that's applicable at all)

custom/enikshay/management/commands/multiple_open_referrals_reconciliation.py
```diff
@@ -74,32 +77,37 @@ def get_case_to_be_retained(self, referral_cases, occurrence_case_id):
 
     def public_app_case(self, occurrence_case_id):
         try:
-            person_case = get_person_case_from_occurrence(DOMAIN, occurrence_case_id)
+            self.person_case = get_person_case_from_occurrence(DOMAIN, occurrence_case_id)
```
esoergel: If the call to `get_person_case_from_occurrence` here fails, then `self.person_case` will still have the value from the previous call, and `get_case_to_be_retained` (which references it) won't know the difference.

it's a bit opaque to set this property in this method and then reference it in another method, especially since this method is called a number of times with different arguments, since it's not clear where that came from.  Another approach would be to fetch it first higher up (in `handle` in this case) and then pass it in to each of these methods.

esoergel (changes requested):
Just one potential source of error, otherwise looks good.

custom/enikshay/management/commands/drug_resistance_reconciliation.py
mkangia: Before this command is queued its ensured that the `person_case_ids` is either None or a list of case ids. So i think it would never be an empty list.

custom/enikshay/management/commands/multiple_open_referrals_reconciliation.py
mkangia: agreed that this looks a little delicate but the way the command is written if `get_person_case_from_occurrence` fails and returns False, there is nothing further done about this case. Going to make this a little more secure by adding `self.person_case=None` before this.

> Another approach would be to fetch it first higher up (in handle in this case) and then pass it in to each of these methods.
passing this around was exactly what i was trying to avoid by assigning this here. :)

PR #18851, opened 2017-12-08
Revert "remove tb_lab_register"

proteusvacuum:
This reverts commit 4c0b7ce36a065bee02f74b89d1a7e5e0b702500f.

@emord @dannyroberts 

https://manage.dimagi.com/default.asp?266864

Don't touch things that aren't broken.

### DIFF
--- a/custom/enikshay/ucr/reports/qa/tb_lab_register.json
+++ b/custom/enikshay/ucr/reports/qa/tb_lab_register.json
@@ -1,0 +1,523 @@
+{
+    "domains": [
+        "enikshay-reports-qa",
+        "sheel-enikshay"
+    ],
+    "server_environment": [
+        "enikshay",
+        "softlayer"
+    ],
+    "data_source_table": "test_qa",
+    "report_id": "tb_lab_register_qa",
+    "config": {
+        "description": "",
+        "title": "TB Lab Register (QA)",
+        "sort_expression": [],
+        "visible": true,
+        "configured_charts": [],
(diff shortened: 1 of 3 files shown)


emord (approved):

dannyroberts:
:'( sad do see such a beautiful deletion go. Is this just easier than getting to the bottom of why deleting this caused problems, or did you realize this was needed after all?

proteusvacuum:
What I'm reading from the sentry error: https://sentry.io/dimagi/commcarehq/issues/414121205/ is that the user has a saved report of this type. I'm not sure the behaviour of having a saved version of a deleted report is defined?

dannyroberts:
When I look closer at the stack trace on sentry, I see inside the `good_configs` method:

```
        for config in all_configs:
            if config.is_configurable_report and not config.configurable_report:
                continue
            good_configs.append(config.to_complete_json(lang=self.language))
```

In other words, it's already intentionally trying to go through and throw out any invalid configurable reports. So I think the right solution isn't to bring back this thing that you'd decided to throw out, but rather to make this `configurable_report` method (it's a `@property`) or whatever it calls recognize when a static report has been deleted and return `None` there as well. Not sure how helpful that suggestion is or if you just want to leave this

proteusvacuum:
That sounds right to me in the long run. Seems out of scope for my code deletion PR though, so if you want to tackle it as part of the interrupt ticket feel free. Otherwise, I'm going to leave this as is.

PR #18854, opened 2017-12-08
Dashboard aggregation failure alerts

emord:
@snopoke from the meeting today. Going to use a soft assert and that will ping me lilianna and sriya so that either i can look at it or be forwarded to another dev.

@calellowitz how were you guys planning to handle errors in the new data warehouse? I can't think if anything other than alert and go digging.

buddy @proteusvacuum

### DIFF
--- a/custom/icds_reports/tasks.py
+++ b/custom/icds_reports/tasks.py
@@ -24,6 +24,8 @@ celery_task_logger = logging.getLogger('celery.task')
 
 UCRAggregationTask = namedtuple("UCRAggregationTask", ['type', 'date'])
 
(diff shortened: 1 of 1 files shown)


snopoke:
@emord for the data warehouse there are 2 stages that occur:

1. Copy data to staging table
2. Update warehouse tables from staging

If an error occurred in (1) we could just start over.
If an error occurred in (2) we should be able remove all data from the warehouse tables based on the `batch_id` (which is unique to each batch run) and then retry loading the data from staging. If that still fails we'd have to revert to the 'notify and dig' approach

PR #18855, opened 2017-12-08
Remove v1 ICDS UCRs

emord:
@sheelio you said that maybe we can do this soon/now?

@proteusvacuum


sheelio:
I think we need a confirmation from Lilianna that we can trigger this .  I'll email her about it again.  Cal / Saket / Srikant have okayed it.

emord:
@sheelio just added a couple more commits for more cleanup. Lilianna just approved this, so I think we can merge/deploy today?

@czue FYI that this should help out with lag/fewer db queries

sheelio:
Yup, I think good to merge / deploy today.

czue:
nice!! do async indicators just ignore references to deleted data sources?

emord:
@czue yeah. they catch it and move on https://github.com/dimagi/commcare-hq/blob/master/corehq/apps/userreports/tasks.py#L338-L341

emord:
@czue fwiw i think that exception was causing those jobs to take longer. I've since run some manual commands to remove these. so there's a huge dropoff on the async ucr dashboard

PR #18856, opened 2017-12-08
add LogLongRequestMiddleware

dannyroberts:
to log request details to sentry when a request takes longer than 2 minutes.
This is so I can look into http://manage.dimagi.com/default.asp?266770, but is useful for such cases in general.

I don't expect it to be a performance hit at all, as nothing (other than two calls to utcnow and calculating the difference) is done in the vast majority of cases, and it just adds one sentry to already very long requests.

We've done a similar thing in formplayer and it's been suuuper useful

### DIFF
--- a/corehq/middleware.py
+++ b/corehq/middleware.py
@@ -13,6 +13,7 @@ from django.utils.deprecation import MiddlewareMixin
 
 from corehq.apps.domain.models import Domain
 from corehq.const import OPENROSA_DEFAULT_VERSION
+from dimagi.utils.logging import notify_exception
 
 from dimagi.utils.parsing import json_format_datetime, string_to_utc_datetime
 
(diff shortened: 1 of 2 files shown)


esoergel (approved):
Seems simple enough.  I assume this won't catch timeouts, right (I think that's fine, not sure we can do much there)?  Will `process_response` be called if an exception is raised?  Looks like 1.10 had [some changes](https://docs.djangoproject.com/en/1.11/topics/http/middleware/) around how that sort of thing is handled.  Not sure of the answer, just wanted to double check that you'd considered raised exceptions.

dannyroberts:
@esoergel yeah I'm not sure about either of those (possibly very common) "edge" cases, but the thing I'm currently working on definitely completes (at least often) without a timeout—it just takes a really long time. Now, I forget how timeouts work, but I believe it's just nginx saying "screw this I never wanted your stupid response anyway"; not sure if the actual processing of the request gets halted or not. (If you kill/restart gunicorn, then anything past the soft timeout will definitely just get hard killed, but that's a different situation.) As for exceptions/500s, I don't expect that this will get called for those, but I'm not banking on the behavior one way or another at this point.

PR #18857, opened 2017-12-08
Fix enikshay location edit page

esoergel:
https://manage.dimagi.com/default.asp?266738
@sravfeyn


orangejenny (approved):
Not a ton of context on this, but reading the ticket & the code it looks reasonable.

PR #18858, opened 2017-12-08
Remove a bunch of unused enikshay user_setup code

esoergel:
https://github.com/dimagi/commcare-hq/pull/18857 inspired me to clean this up at least a little

This work was started, then there were some spec changes and it was deprioritized, and eventually I just ended up turning much of it off.  It's been sitting around long enough that we should just remove it, and add some back in if/when it becomes relevant again.

This should have no functional changes, note that `skip_custom_setup` currently returns `True`, so I went through and removed everything that only happened if that was `False`

@sravfeyn

### DIFF
--- a/custom/enikshay/user_setup.py
+++ b/custom/enikshay/user_setup.py
@@ -1,17 +1,9 @@
-"""
-This uses signals to hook in to user and location forms for custom validation
-and some autogeneration. These additions are turned on by a feature flag, but
-domain and HQ admins are excepted, in case we ever need to violate the
-assumptions laid out here.
-"""
 from __future__ import absolute_import
 import math
-import re
 import uuid
 from crispy_forms import layout as crispy
 from django import forms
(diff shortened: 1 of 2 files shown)


custom/enikshay/user_setup.py
```diff
@@ -29,125 +21,6 @@
...
-
-
-def skip_custom_setup(domain, request_user):
-    return True  # We're gonna revisit this, turning off for now
-    return not toggles.ENIKSHAY.enabled(domain) or request_user.is_domain_admin(domain)
-
-
-def clean_user_callback(sender, domain, request_user, user, forms, **kwargs):
-    if skip_custom_setup(domain, request_user):
-        return
```
esoergel: Given that `skip_custom_setup` returns `True`, this is never called, nor is anything only used inside it.

dannyroberts (approved):

PR #18862, opened 2017-12-11
doses are an integer not a date

calellowitz:
oops. the dose data isnt being updated in the new data source.
@sravfeyn @mkangia 
fyi @proteusvacuum


proteusvacuum:
👍

PR #18864, opened 2017-12-11
wip: get form json from ES if available

czue:
@emord curious to get your input on this. In my [initial profiling exploration](https://gist.github.com/czue/6580a94b37e18d3e587ded5767922b96) this lops of about 1/3 of the processing time (admittedly for a random case / datasource pair).

I think there are some significant potential problems/assumptions here:

- Would getting the full doc from ES cause too much load on the cluster? I know it has been a source of instability
- Would switching to this model break all the shared caching that's been setup because the docs would be too big?
- Does the form coming out of ES actually match perfectly what's in Riak?
- Probably others I'm overlooking

### DIFF
--- a/custom/icds_reports/ucr/expressions.py
+++ b/custom/icds_reports/ucr/expressions.py
@@ -171,9 +171,10 @@ class FormsInDateExpressionSpec(JsonObject):
             context.set_cache_value(cache_key, count)
             return count
 
-        form_ids = [x['_id'] for x in xforms]
-        xforms = FormAccessors(domain).get_forms(form_ids)
-        xforms = [self._get_form_json(f, context) for f in xforms if f.domain == domain]
+        es_forms_by_id = {x['_id']: x for x in xforms}
+        xforms = FormAccessors(domain).get_forms(es_forms_by_id.keys())
+
+        xforms = [self._get_form_json(f, context, es_forms_by_id) for f in xforms if f.domain == domain]
 
         context.set_cache_value(cache_key, xforms)
         return xforms
@@ -195,7 +196,7 @@ class FormsInDateExpressionSpec(JsonObject):
             xform['timeEnd'] = datetime.strptime(time, '%Y-%m-%dT%H:%M:%S.%fZ').date()
             return xform
 
-        forms = mget_query('forms', xform_ids, ['form.meta.timeEnd', 'xmlns', '_id'])
+        forms = mget_query('forms', xform_ids, source=True)
         forms = list(filter(None, map(_transform_time_end, forms)))
         context.set_cache_value(cache_key, forms)
         return forms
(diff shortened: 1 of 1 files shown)


emord:
Apologies if this isn't organized well, but I think this will get across my thoughts/answers to your questions.

It's been a while since I looked into this. So it's hard to say why I didn't go down this path. I think this is fine to push as long as tests pass (i think ccsrecord and child health tests in [this directory](https://github.com/dimagi/commcare-hq/tree/master/custom/icds_reports/ucr/tests) will fail) 

From what I remember, the main benefit of doing it this way is that timeEnd is a way we do a lot of the filtering and it is not stored in SQL, so to filter before we had to get the entire form from riak. 

> Would getting the full doc from ES cause too much load on the cluster? I know it has been a source of instability

Yeah we've often had a lot of timeouts. A lot of them are related to scroll queries (exports). I'm not sure what load mget would have on the cluster. Also we share the server with riak and postgres standby and ES is the first to kill itself from OOM, so I would expect some errors.

If this does fail more now due to the extra load of getting the entire source, it might result in much worse performance (due to more retries) since this doc would totally fail to update (since this is how we do our filtering).

> Would switching to this model break all the shared caching that's been setup because the docs would be too big?

The caching is just a dictionary in the evalation context so I don't think there's a hard size limit. What exactly do you mean by this?

maybe we could just make the caching better? for example [this](https://github.com/dimagi/commcare-hq/blob/master/custom/icds_reports/ucr/data_sources/ccs_record_cases_monthly_tableau2.json#L2998-L3000) and [this](https://github.com/dimagi/commcare-hq/blob/master/custom/icds_reports/ucr/data_sources/ccs_record_cases_monthly_tableau2.json#L1579-L1582) both look for form xmlns start with d4a, but because one has another XMLNS it will result in two queries. Could this be changed to share at least part of the cache/would that be helpful?

Another issue with the cache is that it does a separate query for every month (based on the iteration). Maybe it would be better to get all of the case's form metadata (like form.meta.timeEnd, xmlns etc) once, cache it and then do the filtering in python? That means more python code, but fewer db queries.

> Does the form coming out of ES actually match perfectly what's in SQL?

I think it should be a superset of it? https://github.com/dimagi/commcare-hq/blob/master/corehq/pillows/xform.py#L66-L132 doesn't look like it overwrites anything

> Probably others I'm overlooking

Maybe this should catch timeouts and try sql if that happens?

I've also seen some pretty wide variation based on what time of day these scripts happen. These async UCRs are only run at night though. My guess is that it's due to what's cached, so forms that have been recently submitted will be in memory, but anything in the history will be on disk. It might perform worse when there's 100 processes using this and they're all hitting the disk. Maybe a hybrid approach (half hitting riak, half ES?) would work best so that we use more server's ram?

Also whatever changes we make now, should get re-profiled when this is moved to the new cluster. I expect wildly different(better) performance when we aren't putting riak on the SQL and ES servers, and we can actually use the postgres standbys for better performance

czue:
Thanks for responses @emord . Sounds like there might be something interesting/promising in here so I'll keep poking around at it. A few quick responses to your comments.

> The caching is just a dictionary in the evalation context so I don't think there's a hard size limit. What exactly do you mean by this?

Ah I was confused about this. I thought we had done something that could handle cross-process caching via e.g. redis. This makes me much more excited about this plan, actually.

> maybe we could just make the caching better?

I do think there could be gains here. But my intuition (backed by limited profile data) is that if we can pull the forms from ES we'll get better ROI than optimizing the caching and still having to get every form out of Riak once. Still I'll definitely look closer into this.

> Maybe this should catch timeouts and try sql if that happens?

This is a good idea, though might be a challenge, since the query/filtering itself is done in ES in the same query. A less impactful approach would be to change nothing, but just try a second query against ES to get the docs out. I can profile this later and see if it's a viable first-pass solution (and keep a SQL fallback).

Will report more soon. Also might attempt to see how hard it would be to setup laboratory for this.

czue:
this is superceded by https://github.com/dimagi/commcare-hq/pull/18886

PR #18865, opened 2017-12-11
ucr profiling updates

czue:
@emord I found these defaults to be more helpful in trying to get a big picture of where bottlenecks were. curious what you think.

### DIFF
--- a/corehq/apps/userreports/management/commands/profile_data_source.py
+++ b/corehq/apps/userreports/management/commands/profile_data_source.py
@@ -27,9 +27,9 @@ class Command(BaseCommand):
 
         cProfile.runctx('config.get_all_values(doc)', {}, local_variables, 'ucr_stats.log')
         p = pstats.Stats('ucr_stats.log')
-        p.sort_stats('time')
+        p.sort_stats('cumulative')
 
-        print("Top 10 functions by time\n")
+        print("Top 10 functions by cumulative time\n")
         p.print_stats(10)
 
         print("Specs timing\n")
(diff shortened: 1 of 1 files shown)


emord:
This seems fine. Can you update this with the differences in output? I'm not sure what these two options do. @esoergel may be interested

czue:
this just changes the sort order. `time` is time spent _actually in the function itself_ whereas cumulative time is time spent in the function _and all other functions it calls inside itself_. so it's a good indicator of where the heavy lifting is happening even if that lifting happens several calls down the chain.

because of this you end up with profile outputs that have a lot of similar timings in them (as a function calls a child that calls another child that eventually does something slow). This is why it's helpful to show more lines of profile data.

[These profiles](https://gist.github.com/czue/6580a94b37e18d3e587ded5767922b96) were generated with the new configuration.

emord:
Ah yeah that makes sense. I think these might have two different uses for profiling though. cumulative time is good for figuring out "this particular expression is really expensive" and time is good for figuring out "this part of an expression is expensive" right?

Originally I think my goal was figuring out how expensive things like get_subcases or get_case_forms were. 

I think this change is good from my perspective. @esoergel would this change have helped you out recently?

esoergel:
Cool, yeah, seems useful.  I'll have to play around with it once I get back on that.

czue:
@emord @esoergel alright since there didn't appear to be consensus on this I just made the sort order configurable. I'll keep my opinions about the default to myself. ¯\_(ツ)_/¯

PR #18866, opened 2017-12-11
Use is_case_create instead of transaction.action_type

proteusvacuum:
@emord 
https://sentry.io/dimagi/commcarehq/issues/416481336/

### DIFF
--- a/corehq/ex-submodules/casexml/apps/case/tests/test_property_changed_date.py
+++ b/corehq/ex-submodules/casexml/apps/case/tests/test_property_changed_date.py
@@ -1,12 +1,18 @@
 from __future__ import absolute_import
-from datetime import datetime
-import pytz
 
+from datetime import datetime
+
+import pytz
 from django.test import TestCase
+
 from casexml.apps.case.mock import CaseFactory, CaseStructure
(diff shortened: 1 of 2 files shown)


corehq/ex-submodules/casexml/apps/case/tests/test_property_changed_date.py
```diff
@@ -1,12 +1,18 @@
 from __future__ import absolute_import
+
 from datetime import datetime
-import pytz
 
+import pytz
```
emord: nit: you changed the imports, but didn't make them alphabetical

emord (approved):

corehq/ex-submodules/casexml/apps/case/tests/test_property_changed_date.py
proteusvacuum: I used https://github.com/timothycrosley/isort, but I guess it didn't do the right thing...

emord: weird. Maybe it puts imports before from .. import ?

PR #18867, opened 2017-12-11
Ce/ben reg backport

calellowitz:
@proteusvacuum 
after a long discussion with the enikshay reports team, it looks like the data source i wanted to use for the updated version of this report is not yet tested and does not support the existing report. they also mentioned it would be best to make changes to the live beneficiary register as opposed to a new report (as i believe @proteusvacuum suggested as well). I did some qa on v2 today and after fixing some bugs the adherence score, this PR backports the data source and report changes to the report and data source. it goes a bit on a tangent where i was going to create a new data source, but i think this is better. the only report that uses the one listed is this report and i am confident the data source can rebuild fully overnight.


custom/enikshay/ucr/data_sources/episode_v3_1.json
```diff
@@ -16,22 +16,20 @@
...
-                    "expression": {
-                        "datatype": null,
-                        "type": "property_name",
-                        "property_name": "type"
-                    },
-                    "type": "boolean_expression",
-                    "property_value": "episode"
-                }
-            ]
+            "filters": {
```
esoergel: Flagging that you changed `filters` from a list to a dict, but this is an `and` filter.  Does that work?

custom/enikshay/ucr/data_sources/episode_v3.json
```diff
@@ -6,8 +6,7 @@
         "enikshay-performance-test"
     ],
     "server_environment": [
-        "enikshay",
-        "softlayer"
```
esoergel: Nice catch

custom/enikshay/ucr/data_sources/episode_v3.json
```diff
@@ -18,17 +18,14 @@
...
-                "property_value": "episode"
-            }
+            "operator": "eq",
+            "expression": {
+                "datatype": null,
+                "type": "property_name",
+                "property_name": "type"
+            },
+            "type": "boolean_expression",
+            "property_value": "episode"
```
esoergel: 👍

esoergel (approved):
Seemed reasonable enough.  You're sure this is best done in-place rather than as a separate report?

custom/enikshay/ucr/data_sources/episode_v3_1.json
esoergel: (saw that you resolved this later)

custom/enikshay/ucr/reports/beneficiary_register.json
```diff
@@ -6,7 +6,7 @@
         "enikshay",
         "softlayer"
     ],
-    "data_source_table": "episode_v3",
+    "data_source_table": "episode_v3_1",
```
proteusvacuum: where is this defined?

PR #18873, opened 2017-12-11
Added more specific error logging.

emord:
I think that I originally did this to keep the error logs small and parseable for intermittent db errors on ICDS. I think it will be useful to know which of these failed though

@proteusvacuum


emord:
new buddy @czue

czue (approved):
:spiral_notepad:

PR #18874, opened 2017-12-11
Fixture restore cache invalidation

dannyroberts:
This is a "prefactor" for https://manage.dimagi.com/default.asp?266045. Right now it basically does nothing (except globally bust restore caches due to a change in the key structure). But it exposes a function called `invalidate_restore_cache` that lets you invalidate caches for a single user or an entire domain.

This lets us implement invalidation workflows such as (a) invalidate restore for users affected by a fixture update (b) invalidate restores for a whole domain if it's too hard to figure out from a fixture upload what users' caches to invalidate or there are too many of them (?) (c) when a user's location or group assignment changes, invalidate their restore cache, etc.

Does this seem like a reasonable approach?

### DIFF
--- a/corehq/ex-submodules/casexml/apps/phone/restore_caching.py
+++ b/corehq/ex-submodules/casexml/apps/phone/restore_caching.py
@@ -54,16 +54,15 @@ class _RestoreCache(_CacheAccessor):
 
     @classmethod
     def _make_cache_key(cls, domain, user_id, sync_log_id, device_id):
-        # to invalidate all restore cache keys, increment the number below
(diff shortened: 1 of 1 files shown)


proteusvacuum (approved):

proteusvacuum:
So fresh.

mkangia (approved):

snopoke:
@dannyroberts was there ever a followup to this PR that actually uses the function? I don't see it being used in the current code.

dannyroberts:
Huh, yeah I guess so. `gl -S invalidate_restore_cache` is telling me that this is the only time that function's name ever shows up in our git history

PR #18875, opened 2017-12-11
Refactor 99DOTS

proteusvacuum:
This re-writes how the 99DOTS integration compiles its payload, and checks which case properties to send. I'm about to add ~50 case properties to the integration, and didn't want to add these individually as constants referenced in multiple places. We're also going to add a new endpoint that allows 99DOTS to update these properties, and only these properties.

Probably best 🐡 / 🐡 

In theory this PR should maintain all current behaviour. 

@emord buddy
@dannyroberts since we talked about this
@mkangia, @esoergel  FYI


mkangia:
Soo good! 

nit: can we add some documentation in there so that its just quicker to understand whats happening within those magical methods.

proteusvacuum:
New buddy @sravfeyn

proteusvacuum:
@mkangia good point, I'll add some docs.

custom/enikshay/integrations/ninetyninedots/api_spec.py
```diff
@@ -59,24 +94,170 @@ def load_api_spec():
...
+
+class BasePatientPayload(StrictJsonObject):
+    sector = jsonobject.StringProperty(required=True)
+    state_code = jsonobject.StringProperty(required=False)
+    district_code = jsonobject.StringProperty(required=False)
+    tu_code = jsonobject.StringProperty(required=False)
+    phone_numbers = jsonobject.StringProperty(required=False)
+    treatment_supporter_name = jsonobject.StringProperty(required=False)
+    treatment_supporter_phone_number = jsonobject.StringProperty(required=False)
+    merm_params = jsonobject.ObjectProperty(MermParams, exclude_if_none=True)
```
emord: what's the difference between `ObjectProperty` and `SchemaProperty`

emord (commented):
I don't think I have enough context to review the implementation details but looks good to me.

I like the YAML definition of the API. Is this something that you've followed other places in enikshay? I'm curious if you've felt that it's been helpful/think if it would be good if we had more documentation/definitions as YAML?

custom/enikshay/integrations/ninetyninedots/api_spec.py
proteusvacuum: I don't think jsonobject has a `SchemaProperty`: https://github.com/dimagi/jsonobject#jsonobjectobjectpropertyitem_type

emord: Oops. yeah that's a couchdbkit thing

proteusvacuum:
@emord We haven't done this elsewhere in eNikshay yet, but I feel this is a much clearer definition of what is going on than what we had before. In follow up PRs I'm going to be modifying this a little as we also need to start _accepting_ these properties from 99DOTS. Having this all defined in one place will be really nice, I think. However, I've already found that this can be a little brittle, as there are so many special cases. 

Once I'm done with the whole thing it could be cool to explore other places patterns like this might be useful.

custom/enikshay/integrations/ninetyninedots/tests/test_api_spec.py
```diff
@@ -0,0 +1,11 @@
...
+
+from django.test import SimpleTestCase
+
+from custom.enikshay.integrations.ninetyninedots.api_spec import load_api_spec
+
+
+class NinetyNineDotsSpecLoaderTest(SimpleTestCase):
+    def test_loading_spec(self):
+        # if the yaml spec file is unable to be loaded, this will throw an error
+        load_api_spec()
```
dannyroberts: good work on this commit, Farid.

custom/enikshay/integrations/ninetyninedots/api_spec.py
```diff
@@ -59,24 +94,170 @@ def load_api_spec():
...
+        required=False,
+        exclude_if_none=True
+    )  # yy/MM/dd HH:mm:ss
+    RT_hours = jsonobject.StringProperty(
+        required=False,
+        exclude_if_none=True
+    )  # 1 = 12 hours; i.e. for 3 days - RT_hours = 6
+
+
+class BasePatientPayload(StrictJsonObject):
```
dannyroberts: Curious how you decided what to put in code (here) versus in the spec (yaml file)

dannyroberts (approved):

PR #13420, opened 2016-10-01
add expression column type to UCR reports

czue:
this is the other thing I demo'd tuesday. the whole hierarchy of `Column` and `ColumnConfig` objects is super confusing in these reports and I would like to clean it up a bit more but decided to just put this up for now since I wasn't quite sure how to do that and there's messy dependencies on `SqlData`.

@emord might be a good person to review since he's been in this code recently. cc buddy @snopoke . probably easiest by commit.

PR #18880, opened 2017-12-12
allowing expression columns to be invisible and have descriptions

calellowitz:
@orangejenny @proteusvacuum @emord 
This is breaking an enikshay report. Not sure why, but for some reason until now it has been impossible to make expression columns invisible (or give them a help text). This fixes that.

### DIFF
--- a/corehq/apps/userreports/reports/specs.py
+++ b/corehq/apps/userreports/reports/specs.py
@@ -455,11 +455,12 @@ def _add_column_id_if_missing(obj):
         obj['column_id'] = obj.get('alias') or obj['field']
 
 
-class CalculatedColumn(namedtuple('CalculatedColumn', ['header', 'slug'])):
+class CalculatedColumn(namedtuple('CalculatedColumn', ['header', 'slug', 'visible', 'help_text'])):
 
     @property
     def data_tables_column(self):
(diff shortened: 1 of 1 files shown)


emord (approved):
Looks good as long as these default to false

I don't see any red flags from the original implementation https://github.com/dimagi/commcare-hq/pull/13420

corehq/apps/userreports/reports/specs.py
```diff
@@ -481,9 +482,10 @@ def get_column_config(self, data_source_config, lang):
             CalculatedColumn(
                 header=self.get_header(lang),
                 slug=self.column_id,
+                visible=self.visible,
                 # todo: are these needed?
```
proteusvacuum: nit: this comment is no longer accurate

calellowitz:
> Looks good as long as these default to false

Visible actually defaults to true (as it does for all other column types), but yeah description defaults empty.

corehq/apps/userreports/reports/specs.py
calellowitz: I left that there for the one that is still commented out. I can remove it, I just did not know how the other property worked so did not feel confident testing it.

calellowitz:
merging now, to fix live reports.
@proteusvacuum happy to follow up with fixes for that comment if you would like

PR #18881, opened 2017-12-12
add dynamically configurable random toggle

czue:
I thought it might be useful to have a toggle that we could configure dynamically and turned out to be quite easy to implement.

The specific use case is that I want to PoC something in ICDS UCR processing and be able to ratchet it up and down depending on whether it starts to cause problems. But I imagine it could be useful in other future places as well? basically anything where the toggle is controlling a system thing instead of a user thing.

any thoughts as to whether there's any reason this is a bad idea to exist? if approved i'll look into adding to the UI as well.

### DIFF
--- a/corehq/toggles.py
+++ b/corehq/toggles.py
@@ -228,8 +228,8 @@ class PredictablyRandomToggle(StaticToggle):
         super(PredictablyRandomToggle, self).__init__(slug, label, tag, list(namespaces),
                                                       help_link=help_link, description=description,
                                                       always_disabled=always_disabled)
-        assert namespaces, 'namespaces must be defined!'
-        assert 0 <= randomness <= 1, 'randomness must be between 0 and 1!'
+        _ensure_valid_namespaces(namespaces)
+        _ensure_valid_randomness(randomness)
         self.randomness = randomness
 
     @property
@@ -286,8 +286,8 @@ class DynamicallyPredictablyRandomToggle(PredictablyRandomToggle):
         super(PredictablyRandomToggle, self).__init__(slug, label, tag, list(namespaces),
                                                       help_link=help_link, description=description,
                                                       always_disabled=always_disabled)
-        assert namespaces, 'namespaces must be defined!'
-        assert 0 <= default_randomness <= 1, 'default randomness must be between 0 and 1!'
+        _ensure_valid_namespaces(namespaces)
+        _ensure_valid_randomness(default_randomness)
         self.default_randomness = default_randomness
 
     @property
@@ -370,6 +370,16 @@ def toggle_values_by_name(username=None, domain=None):
     return {toggle_name: (toggle.enabled(username, NAMESPACE_USER) or
                           toggle.enabled(domain, NAMESPACE_DOMAIN))
             for toggle_name, toggle in all_toggles_by_name().items()}
+
+
+def _ensure_valid_namespaces(namespaces):
+    if not namespaces:
+        raise Exception('namespaces must be defined!')
+
+
+def _ensure_valid_randomness(randomness):
+    if 0 <= randomness <= 1:
+        raise Exception('randomness must be between 0 and 1!')
 
 
 APP_BUILDER_CUSTOM_PARENT_REF = StaticToggle(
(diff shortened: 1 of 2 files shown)


corehq/toggles.py
```diff
@@ -265,6 +265,43 @@ def enabled(self, item, namespace=Ellipsis):
...
+        namespaces,
+        default_randomness=0,
+        help_link=None,
+        description=None,
+        always_disabled=None
+    ):
+        super(PredictablyRandomToggle, self).__init__(slug, label, tag, list(namespaces),
+                                                      help_link=help_link, description=description,
+                                                      always_disabled=always_disabled)
+        assert namespaces, 'namespaces must be defined!'
```
emord: not blocking. I think you sent us a talk about instagram scaling up and they started using python -O which gets rid of asserts. Should we stop using these and start throwing exceptions?

corehq/toggles.py
```diff
@@ -265,6 +265,43 @@ def enabled(self, item, namespace=Ellipsis):
             or super(PredictablyRandomToggle, self).enabled(item, namespace)
         )
 
+
+class DynamicallyPredictablyRandomToggle(PredictablyRandomToggle):
+    """
+    A PredictablyRandomToggle whose randomness can be configured via the database/UI.
```
emord: There's no UI component at the moment though right?

corehq/toggles.py
```diff
@@ -265,6 +265,43 @@ def enabled(self, item, namespace=Ellipsis):
...
+                                                      help_link=help_link, description=description,
+                                                      always_disabled=always_disabled)
+        assert namespaces, 'namespaces must be defined!'
+        assert 0 <= default_randomness <= 1, 'default randomness must be between 0 and 1!'
+        self.default_randomness = default_randomness
+
+    @property
+    @quickcache(vary_on=['self.slug'])
+    def randomness(self):
+        from toggle.models import Toggle
```
emord: nit: can this be at the top of the file with other imports?

corehq/toggles.py
snopoke: +1

snopoke (commented):
No big issues with this but I think updating the value in the UI should be restricted to `staff` users.

corehq/toggles.py
emord: @snopoke didn't realize that. Should this be communicated to the team to no longer use `assert` ? I don't see a way to enforce this using stickler (flake8/pep8)

snopoke: I think it's bad practice to use `assert` for anything that get's relied on in production. In this case though I actually think it's fine since these classes are statically defined in code so the `assert` would fail locally or during build. If this `assert` got removed in production by using `-O` that won't affect anything.

corehq/toggles.py
czue: correct, going to see how hard it is to add. we could also just manage via CLI if we wanted to really restrict it to developers.

corehq/toggles.py
czue: good call. there was another inline import elsewhere in the file and i thought maybe someone had had to do that for a circular import situation, but looks like that's probably not the case.

corehq/toggles.py
czue: from a performance perspective, the reason you compile asserts out is just to avoid the branching logic right? so if we had `if not namespaces: raise Exception()` would that not be the same? 

I'm happy to swap it (and the code it was copied from) to use `Exception`, but agree with @emord that we should communicate out to the team if that's the suggested standard moving forwards.

czue:
@emord @snopoke updated. thoughts on handling the UI / CLI bit in a separate PR?

corehq/toggles.py
snopoke: @czue correct re branching logic but what I was saying is that you shouldn't really use asserts for runtime validation since they can get compiled out. In this case it's fine since they aren't really runtime validations.

snopoke:
@czue fine to put UI / CLI bit in separate PR

czue:
@snopoke before I spend any time - any thoughts on which would be better? UI seems nicer if we're not concerned with people messing things up. i don't have a good sense of how big a risk that is

snopoke:
I don't think it's a big risk but it is quite a bit more work. I'm ok with either.

PR #18883, opened 2017-12-12
swallow unexpected string instead of dict in form export

dannyroberts:
https://manage.dimagi.com/default.asp?264884

### DIFF
--- a/corehq/apps/export/models/new.py
+++ b/corehq/apps/export/models/new.py
@@ -548,10 +548,11 @@ class TableConfiguration(DocumentSchema):
             doc = row_doc.doc
             row_index = row_doc.row
 
-            try:
+            if isinstance(doc, dict):
                 next_doc = doc.get(path[0].name, {})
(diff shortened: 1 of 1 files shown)


corehq/apps/export/models/new.py
```diff
@@ -548,10 +548,11 @@ def _get_sub_documents_helper(path, row_docs):
             doc = row_doc.doc
             row_index = row_doc.row
 
-            try:
+            if isinstance(doc, dict):
                 next_doc = doc.get(path[0].name, {})
-            except AttributeError:
-                raise PossibleHeterogenousDataException(doc, path[0].name)
```
esoergel: This is the only place this exception is raised, so you could also drop the exception class and the code which catches it (the soft_assert).

dannyroberts: Good call. I'll move the soft assert here.

esoergel (approved):
Thanks!

PR #18886, opened 2017-12-13
pull UCR forms from ES, take 2

czue:
a more comprehensive version of https://github.com/dimagi/commcare-hq/pull/18864

### DIFF
--- a/custom/icds_reports/tests/__init__.py
+++ b/custom/icds_reports/tests/__init__.py
@@ -36,6 +36,7 @@ FILE_NAME_TO_TABLE_MAPPING = {
 
 def setUpModule():
     if settings.USE_PARTITIONED_DATABASE:
+        print('============= WARNING: not running test setup because settings.USE_PARTITIONED_DATABASE is True.')
         return
 
     _call_center_domain_mock = mock.patch(
--- a/custom/icds_reports/ucr/expressions.py
+++ b/custom/icds_reports/ucr/expressions.py
@@ -188,8 +188,7 @@ class FormsInDateExpressionSpec(JsonObject):
         if context.get_cache_value(cache_key) is not None:
             return context.get_cache_value(cache_key)
 
-        forms = mget_query('forms', xform_ids, ['form.meta.timeEnd', 'xmlns', '_id'])
-        forms = list(filter(None, map(self._transform_time_end_and_filter_bad_data, forms)))
+        forms = self._bulk_get_forms_from_elasticsearch(xform_ids, ['form.meta.timeEnd', 'xmlns', '_id'])
         context.set_cache_value(cache_key, forms)
         return forms
 
@@ -228,6 +227,13 @@ class FormsInDateExpressionSpec(JsonObject):
     @staticmethod
     def _get_form_json_cache_key(form):
         return (XFORM_CACHE_KEY_PREFIX, form.form_id)
+
+    @staticmethod
+    def _bulk_get_forms_from_elasticsearch(form_ids, source):
+        forms = mget_query('forms', form_ids, source)
+        return list(filter(None, [
+            FormsInDateExpressionSpec._transform_time_end_and_filter_bad_data(f) for f in forms
+        ]))
 
     @staticmethod
     def _transform_time_end_and_filter_bad_data(xform):
--- a/custom/icds_reports/ucr/expressions.py
+++ b/custom/icds_reports/ucr/expressions.py
@@ -177,9 +177,16 @@ class FormsInDateExpressionSpec(JsonObject):
             context.set_cache_value(cache_key, count)
             return count
 
-        form_ids = [x['_id'] for x in xforms]
-        xforms = FormAccessors(domain).get_forms(form_ids)
-        xforms = FormsInDateExpressionSpec._get_form_json_list(case_id, xforms, context, domain)
+        if ICDS_UCR_ELASTICSEARCH_DOC_LOADING.enabled(case_id, NAMESPACE_OTHER):
+            def _clean(f):
+                del f['timeEnd']  # for parity with what comes out of Riak
+                return f
+
+            xforms = list(map(_clean, xforms))
+        else:
+            form_ids = [x['_id'] for x in xforms]
+            xforms = FormAccessors(domain).get_forms(form_ids)
+            xforms = FormsInDateExpressionSpec._get_form_json_list(case_id, xforms, context, domain)
 
         context.set_cache_value(cache_key, xforms)
         return xforms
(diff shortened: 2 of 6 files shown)


czue:
iniital profiling seems to indicate that ES is marginally faster, but with a very high variance. not sure whether this is worth it or not.

czue:
(will look at tests as well)

emord (commented):
This looks reasonable to me. I don't see where you are gathering data here. Are you planning to use logs, datadog or something else?

Also interesting idea to create an "other" namespace for toggles. Any reason to not go with a localsetting? I think you still have to use the command line since there's not a UI to change this percentage right?

czue:
> Are you planning to use logs, datadog or something else?

For the time being I was just running the UCR profiling commands manually in a release and looking at the output. Not exactly sure the long term plan yet.

> Also interesting idea to create an "other" namespace for toggles. Any reason to not go with a localsetting? I think you still have to use the command line since there's not a UI to change this percentage right?

I'm planning on adding a UI soon and localsettings would require an ansible deploy which is just one more bit of friction I'd like to avoid.

emord:
> For the time being I was just running the UCR profiling commands manually in a release and looking at the output. Not exactly sure the long term plan yet.

FWIW because of the variances and inconsistencies we've seen in these profiling, I'm pretty hesitant to promise/report anything timing-based just based on the data source profile command. I'm doubtful that our ES cluster is going to handle a 100 processes hitting it for the entire docs (plus whatever other things hit it overnight) so what works for this profile may not work at scale. I think we'll also have different(better) performance once the migration is done.

I could see some usefulness in the medium term to keep this at an intermediate value. Maybe there's some local maximum where if x% of our calls are going to ES and y% are going to riak then it could optimize our resources. Long term that would be great to get rid of.

Do you have any other db calls you're looking at? I could see it being useful having multiple levers to pull while testing this if there will be other changes.

> I'm planning on adding a UI soon and localsettings would require an ansible deploy which is just one more bit of friction I'd like to avoid. I also think the pattern is potentially useful so good to build out some infrastructure around it

agreed that I think this could prove useful in the future

czue:
> FWIW because of the variances and inconsistencies we've seen in these profiling, I'm pretty hesitant to promise/report anything timing-based just based on the data source profile command. I'm doubtful that our ES cluster is going to handle a 100 processes hitting it for the entire docs (plus whatever other things hit it overnight) so what works for this profile may not work at scale.

Totally agree with this. That's why my plan had been to tweak the toggle value and look at datadog. Still planning on doing this, though I think I'm going to convert it back to the version that gets the initial docs out. Just haven't had a chance to finish that up yet. Will work with @snopoke and others on this while you're out.

czue:
@snopoke no rush, but I think this will be ready to go soon and then once new environment is up and running my plan is to play with the toggle and see if there is any significant impact on processing times.

snopoke (approved):
Kicked build :boot: :construction:

czue:
I think it was real. https://github.com/dimagi/commcare-hq/pull/18886/commits/3717e0a890e81fefbe04acaa43248bbdf9afccbe

custom/icds_reports/tests/__init__.py
```diff
@@ -36,6 +36,7 @@
 
 def setUpModule():
     if settings.USE_PARTITIONED_DATABASE:
+        print('============= WARNING: not running test setup because settings.USE_PARTITIONED_DATABASE is True.')
```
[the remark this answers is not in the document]
czue: is there a better way to do this? logging? or just ignore @stickler-ci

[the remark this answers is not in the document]
snopoke: I ignore those when it's not in prod code. The other way to do it would be to use `logger.warn`

czue:
:pray:

sravfeyn:
Hey @czue looks like this is still a random toggle. Do you remember any datadog profiling notebooks that you can link me to? I am looking into this related ES issue https://dimagi-dev.atlassian.net/browse/IIO-461

czue:
It's not random, it's always enabled: https://www.icds-cas.gov.in/hq/flags/edit/icds_ucr_elasticsearch_doc_loading/

I think it's been always enabled since basically when it was merged. Will leave a comment on JIRA

czue:
Er... I don't actually have much to weigh in beyond what Emord already said on that JIRA ticket though definitely agree removing this dependency is the best long term solution. 

I don't think I have any datadog notebooks or anything you can refer too, but this change has been live for probably 18 months at least.

sravfeyn:
Cool. If it's not random, then there's nothing to compare, thanks!

custom/icds_reports/ucr/expressions.py
```diff
@@ -215,15 +215,64 @@ def _get_case_form_ids(self, case_id, context):
...
+    def _set_cached_form_json(form, form_json, context):
+        context.set_cache_value(FormsInDateExpressionSpec._get_form_json_cache_key(form), form_json)
+
+    @staticmethod
+    def _get_form_json_cache_key(form):
+        return (XFORM_CACHE_KEY_PREFIX, form.form_id)
+
+    @staticmethod
+    def _bulk_get_forms_from_elasticsearch(form_ids, source):
+        forms = mget_query('forms', form_ids, source)
```
sravfeyn: Hey @czue Looks like this expression `icds_get_case_forms_in_date` is used only in one ICDS UCR with `count: True`. Was that not the case when you wrote this? I am thinking to make a PR to just get the counts and meta fields explicitly and remove full form loading. 

Also was the performance gain over shard DBs or over the blobdb that's needed for loading the full form?

czue: Hard to remember offhand, but I'm pretty sure it was more broadly used before that. Should be trivial to checkout the last commit of this PR and grep again I would think.

Performance was relative to blobdb.

@emord might remember why we were doing this in memory versus pushing the query into ES on the first go around. I think it had something to do with caching and re-use.

czue: But sorry, just to be more clear:

> I am thinking to make a PR to just get the counts and meta fields explicitly and remove full form loading.

I don't see any reason why this wouldn't be a good idea at this point, but @emord might.

emord: These are used much less than they were when initially, however as pointed out in the scale meeting these are still used for more than just counts https://github.com/dimagi/commcare-hq/blob/dcaf9aebd6f22d09b002e53f8bc5ea75e77d0f54/custom/icds_reports/ucr/data_sources/child_cases_monthly_v2.json#L1830-L1857

Cory is right that we originally didn't do rely on ES counts so that the these cached queries could be reused.

(FWIW I think your idea to create a new expression to pull these from a UCR data source is pretty good. It is not an ideal solution, but it sounds like it hits a good balance of LOE vs better place than we are now)

PR #18887, opened 2017-12-13
change DeviceReportEntry id to bigint

snopoke:
This can only be merged after https://github.com/dimagi/commcare-hq/pull/18888 has been run on all envs.


sravfeyn (approved):

millerdev (approved):

corehq/ex-submodules/phonelog/models.py
```diff
@@ -7,6 +7,7 @@
 
 @architect.install('partition', type='range', subtype='date', constraint='day', column='server_date')
 class DeviceReportEntry(models.Model):
+    id = models.BigAutoField(primary_key=True)
```
emord: I haven't looked deeply at https://code.djangoproject.com/ticket/373 but it looks like what I'm going to suggest is unsupported in django. There's not a lot of use for this model, so it may not be completely crazy to port it to sqlalchemy?

Can we just drop this id completely? We could have a composite primary key for xform_id and i (already unique_togther) and it means this table is slightly smaller and we don't rely on some field autoincrementing.

snopoke: yea, that would be great

snopoke:
@emord what do you think about just resetting the ID sequence generator back to 0? Since we only keep 30 days of records I don't think it should be an issue.

emord:
Seems fine to me. We could also use UUID's right? That should work indefinitely I think?

PR #18888, opened 2017-12-13
command to change int 'id' field to bigint

snopoke:
INT ID field on DeviceReportEntry needs to be changed to BIGINT

See https://github.com/dimagi/commcare-hq/pull/18887

### DIFF
--- a/corehq/apps/cleanup/management/commands/int_id_to_bigint.py
+++ b/corehq/apps/cleanup/management/commands/int_id_to_bigint.py
@@ -1,0 +1,116 @@
+from __future__ import absolute_import
+from __future__ import print_function
+
+from django.apps import apps
+from django.core.management.base import BaseCommand, CommandError
+from django.db import connections
+
+from corehq.form_processor.utils.sql import fetchone_as_namedtuple
+from corehq.util.django_migrations import add_if_not_exists_raw
+from corehq.util.log import with_progress_bar
+
+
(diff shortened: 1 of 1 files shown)


corehq/apps/cleanup/management/commands/int_id_to_bigint.py
```diff
@@ -0,0 +1,117 @@
...
+from corehq.util.django_migrations import add_if_not_exists_raw
+from corehq.util.log import with_progress_bar
+from six.moves import input
+
+
+def confirm(table_name):
+    return input(
+        u"""
+        Are you sure you want to modify the {} table? [y/N]
+        """.format(table_name)
```
millerdev: nit: this is going to be kinda ugly because of extra whitespace in the string. Consider adding `.strip()`

millerdev (approved):
I'm assuming this is based on https://stackoverflow.com/a/33509181/10840 given how closely it mirrors that solution. Maybe add a comment crediting that source?

corehq/apps/cleanup/management/commands/int_id_to_bigint.py
snopoke: I actually quite like the way this dispalys on the console.

emord:
@snopoke I think the answer to this is that yes it works, but did you verify this works on the child tables since we use architect to do split the tables up by days?

PR #11745, opened 2016-05-25
better warning when a UCR rebuild is going to do nothing

czue:
and a refactor and tests. the actual change is just the last commit.

related to http://manage.dimagi.com/default.asp?225930

@emord @calellowitz I think this is mostly your code cc @biyeun

PR #18889, opened 2017-12-13
clear resume info for static ucrs as well

calellowitz:
@emord @czue cc: @dannyroberts 
the resume info was not being cleared for static reports upon completion which made attempts to rebuild them exit immediately.  appears to have been the case since https://github.com/dimagi/commcare-hq/pull/11745

Did you purposefully exclude those @czue or is this an oversight. want to make sure this will not break anything

### DIFF
--- a/corehq/apps/userreports/tasks.py
+++ b/corehq/apps/userreports/tasks.py
@@ -155,8 +155,8 @@ def _iteratively_build_table(config, resume_helper=None, in_place=False, limit=-
 
         resume_helper.add_completed_case_type_or_xmlns(case_type_or_xmlns)
 
+    resume_helper.clear_resume_info()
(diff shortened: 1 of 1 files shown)


czue (approved):
I have no memory of doing this at all, let alone intentionally. I also think the change predates the PR you listed (which just refactored stuff out into the resume helper)

My strong intuition is that intention was just to not call any code that attempts to call `.save()` on a static data sources so I think this is fine/correct. Also explains that rebuild bug!

emord (approved):
So the next time we need to rebuild one of these we'll need to rebuild it twice right? Once to clear the resume info, the next to actually start it up?

dannyroberts (approved):

PR #18890, opened 2017-12-13
prevent admin model view timeout

sravfeyn:
https://docs.djangoproject.com/en/1.11/ref/contrib/admin/#django.contrib.admin.ModelAdmin.show_full_result_count

Fix for https://manage.dimagi.com/default.asp?266178#1431557

@proteusvacuum


sravfeyn:
@snopoke

snopoke (approved):

PR #18891, opened 2017-12-13
release hearbeat for all apps

sravfeyn:
Product Note: This adds a heartbeat URL to all apps.  This enables using the force update functionality for applications in the future.
https://manage.dimagi.com/default.asp?266442#1433770
@proteusvacuum

### DIFF
--- a/corehq/apps/app_manager/models.py
+++ b/corehq/apps/app_manager/models.py
@@ -5534,7 +5534,10 @@ class Application(ApplicationBase, TranslationMixin, HQMediaMixin):
         if toggles.CUSTOM_PROPERTIES.enabled(self.domain) and "custom_properties" in self__profile:
             app_profile['custom_properties'].update(self__profile['custom_properties'])
 
-        apk_heartbeat_url = self.heartbeat_url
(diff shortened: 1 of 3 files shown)


corehq/apps/app_manager/models.py
```diff
@@ -5534,10 +5534,7 @@ def create_profile(self, is_odk=False, with_media=False,
         if toggles.CUSTOM_PROPERTIES.enabled(self.domain) and "custom_properties" in self__profile:
             app_profile['custom_properties'].update(self__profile['custom_properties'])
 
-        if toggles.PHONE_HEARTBEAT.enabled(self.domain):
-            apk_heartbeat_url = self.heartbeat_url
-        else:
-            apk_heartbeat_url = None
+        apk_heartbeat_url = self.heartbeat_url
```
sheelio: @sravfeyn 

I think this is the only change we need - we still want to hide the UX for "enable update prompts" behind the feature flag.

sravfeyn: Cool, updated

proteusvacuum (approved):

PR #18895, opened 2017-12-13
Add Summary to Data Sources

sheelio:
Product Note: This adds a summary page to each data source which provides a more readable output of each indicator, named expression and named filter. 

image


corehq/apps/userreports/filters/__init__.py
```diff
@@ -11,6 +14,9 @@ class Filter(object):
     def __call__(self, item, context=None):
         return True
 
+    def __str__(self):
```
sheelio: I added this method on the main filter class but wondering why we don't combine the Filter object and the FilterSpec objects and make the filter specs callable (similar to expressions).

emord:
Looks like a great first pass! I think it will take some back and forth with GS and other members to get the wording and formatting right, but this does seem easier to me.

Is there a section for "configured_filter" to show how the docs are filtered?

Also might be good to add a header with the name and description: https://github.com/dimagi/commcare-hq/blob/master/custom/icds_reports/ucr/data_sources/it_report_follow_issue.json#L15-L16

sheelio:
@emord 

Cool, updated.  My guess is if the code looks good, we can deploy this and get feedback from the GS team as they use this.

emord (approved):

corehq/apps/userreports/expressions/specs.py
```diff
@@ -290,6 +355,22 @@ def get_variables(self, item, context):
...
 
+    def __str__(self):
+        # context_variables = "\n".join(
+        #     ["{}:{}".format(name, str(exp)) for name, exp in self._context_variables.items()]
+        # )
+        # value = "{} with (\n{}\n)".format(self.statement,
+        #                                   context_variables)
+        # if self.datatype:
+        #     value =  "({}){}".format(self.datatype, value)
+        # return value
```
esoergel: Did you mean to leave this comment here?

corehq/apps/userreports/filters/__init__.py
```diff
@@ -70,3 +90,20 @@ def __init__(self, expression, operator, reference_expression):
...
+class NamedFilter(Filter):
+    def __init__(self, filter_name, filter):
+        self.filter_name = filter_name
+        self.filter = filter
+
+    def __call__(self, item, context=None):
+        return self.filter(item, context)
+
+    def __str__(self):
+        return "NamedF:{}".format(self.filter_name)
```
esoergel: `NamedF`?  You could probably just go with something like:
`'"{}" filter` or `'named filter: "{}"'`

corehq/apps/userreports/models.py
```diff
@@ -255,9 +255,52 @@ def indicators(self):
...
+                "comment": self.named_filters[name].get('comment'),
+                "readable_output": str(filter)
+            }
+            for name, filter in self.named_filter_objects.items()
+        ]
+
+    @property
+    def configured_filter_summary(self):
+        return str(FilterFactory.from_spec(self.configured_filter, context=self.get_factory_context()))
+
```
esoergel: These 4 methods might be better off as utilities, rather than model methods.  They look specific to the template where this is going to be rendered, as opposed to something that will be referenced frequently.

corehq/apps/userreports/indicators/specs.py
```diff
@@ -41,6 +42,9 @@ def wrap(cls, obj):
             wrapped.display_name = wrapped.column_id
         return wrapped
 
+    def readable_output(self, context):
+        raise NotImplementedError
+
```
esoergel: Out of curiosity, why is this `readable_output` rather than `__str__`?

esoergel: Oh nm, it's because you need to pass in `context`

custom/icds_reports/ucr/expressions.py
```diff
@@ -393,55 +417,36 @@ def get_case_forms_by_date(spec, context):
...
+            "property_path": [
+                "form",
+                "meta",
+                "timeEnd"
+            ],
+            "datatype": "date"
+        },
+        "items_expression": items_expression
+    }
     return ExpressionFactory.from_spec(spec, context)
```
esoergel: Clever change.  Took me a minute to figure out how this is the same, but I'm with you now!

esoergel (commented):
Looks good!  I'm happy to see how small the code for this is if you discount the `__str__` functions.  I'm guessing we'll probably iterate on those as this gets used, but this is a great start!

PR #18896, opened 2017-12-13
update/create ledger for an adherence at 99dots and nightly reconciliation

mkangia:
1. [99DOTS API updates](https://docs.google.com/document/d/1VduUNl11KHKIvpK19HgicX1l9FlTp_Sp-JyYN1dVaps/edit#heading=h.witeiq3l8hgd)
2. [Nightly Model Reconciliation Process updates](https://docs.google.com/document/d/1VduUNl11KHKIvpK19HgicX1l9FlTp_Sp-JyYN1dVaps/edit#heading=h.33k33fhi5pp0)


custom/enikshay/tasks.py
```diff
@@ -185,14 +187,26 @@ def run_batch(self, case_ids):
...
+            self._update_ledger_values(episode_case_ids)
 
         return BatchStatus(update_count, noupdate_count, success_count, errors, case_batches, t.interval)
 
+    def _update_ledger_values(self, episode_case_ids):
+        accessor = CaseAccessors(self.domain)
+        for episode_case_id in episode_case_ids:
+            # fetch episode case again to get updated case
+            episode_case = accessor.get_case(episode_case_id)
+            adherence_cases = get_adherence_cases_from_episode(episode_case)
```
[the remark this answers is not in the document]
sravfeyn: You can use `AdherenceDatastore` which is already initialized in this class (see, `adherence_data_store`)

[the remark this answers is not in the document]
sravfeyn: You will be able to avoid these calculations if you move this logic to `EpisodeAdherenceUpdate`, calculate ledger updates, and then commit them in bulk [here](https://github.com/dimagi/commcare-hq/pull/18896/files#diff-7ede24022ad113f7bd1a083192d23896R199)

custom/enikshay/integrations/ninetyninedots/views.py
```diff
@@ -74,6 +76,16 @@ def update_patient_adherence(request, domain):
             message=("An error occurred updating the episode case after receiving a 99DOTS"
                      "adherence case for beneficiary {}. {}").format(beneficiary_id, e))
 
+    try:
+        for adherence_case in adherence_cases:
```
proteusvacuum: I think the `factory.update_episode_adherence_properties()` will already take care of this (since that calls the reconciliation task). Otherwise, this should be fine.

custom/enikshay/tasks.py
```diff
@@ -163,10 +165,13 @@ def run_batch(self, case_ids):
             updates = []
             for episode in self._get_open_episode_cases(case_ids):
                 did_error = False
+                adherence_cases = None
                 update_json = {}
                 for updater in self.updaters:
                     try:
                         potential_update = updater(self.domain, episode).update_json()
+                        if isinstance(updater, EpisodeAdherenceUpdate):
+                            adherence_cases = updater.get_all_adherence_cases()
```
proteusvacuum: I think we only want to update the ledger values for the "winning" case. I've added some info in the spec.

proteusvacuum (changes requested):

custom/enikshay/integrations/ninetyninedots/views.py
mkangia: Oh i see, `update_single_episode` -> `EpisodeAdherenceUpdate.get_json()`

This is just getting the updates to be done on the episode case but does not run the `EpisodeAdherenceUpdate` updater/reconciliation. I am not sure if there is a way to merge the ledger update with `EpisodeAdherenceUpdate` and even if there is, it would be more efficient than doing ledger update as a follow up after the adherence update.

custom/enikshay/utils.py
```diff
@@ -0,0 +1,117 @@
...
+        return True
+
+    domain = episode_case.domain
+    if adherence_source and adherence_value:
+        fixture_id = get_id_of_fixture_tagged_adherence_ledger_values(domain)
+        if fixture_id:
+            ledger_value = get_all_fixture_items(
+                domain,
+                fixture_id
+            ).get(adherence_source, {}).get(adherence_value)
```
proteusvacuum: nit, since you are using a `defaultdict` for this, you should actually be able to do 
```py
get_all_fixture_items(...)[adherence_source].get(adherence_value)
```
since the `defaultdict` already creates the empty dict for you if the key doesn't exist.

custom/enikshay/utils.py
```diff
@@ -0,0 +1,117 @@
...
+                    balance = Balance()
+                    balance.entity_id = episode_case.case_id
+                    balance.date = datetime.utcnow()
+                    balance.section_id = "adherence"
+                    entry = Entry()
+                    entry.id = entry_id
+                    entry.quantity = ledger_value
+                    balance.entry = entry
+                    return submit_case_blocks([balance.as_string()],
+                                              episode_case.domain,
```
proteusvacuum: nit, you stored this as the variable `domain` above

custom/enikshay/utils.py
```diff
@@ -0,0 +1,117 @@
...
+    """
+    ledger_accessor = LedgerAccessors(domain)
+    try:
+        return ledger_accessor.get_ledger_value(episode_case_id, "adherence", entry_id)
+    except LedgerValueNotFound:
+        return None
+
+
+@quickcache(['domain'], memoize_timeout=7 * 24 * 60 * 60, timeout=7 * 24 * 60 * 60)
+def get_id_of_fixture_tagged_adherence_ledger_values(domain):
```
proteusvacuum: nit: this name is very long (almost as long as the function itself)! could rename it as `get_adherence_fixture_id` or something.

custom/enikshay/utils.py
```diff
@@ -0,0 +1,117 @@
...
+    fixture = FixtureDataType.by_domain_tag(domain, "adherence_ledger_values").first()
```
proteusvacuum: nit: would make this name a const.

custom/enikshay/utils.py
```diff
@@ -0,0 +1,117 @@
...
+    Entry,
+)
+from corehq.apps.hqcase.utils import submit_case_blocks
+from corehq.apps.users.util import SYSTEM_USER_ID
+
+
+def get_adherence_cases_by_date(adherence_cases):
+    adherence_cases_by_date = defaultdict(list)
+    for case in adherence_cases:
+        adherence_date = parse_date(case['adherence_date']) or parse_datetime(case['adherence_date']).date()
```
proteusvacuum: is there the risk that the `case` doesn't have an `adherence_date` property? What happens then?

custom/enikshay/tasks.py
```diff
@@ -795,28 +813,26 @@ def calculate_dose_status_by_day(adherence_cases):
...
-            status_by_day[day] = DoseStatus(taken=False, missed=True, unknown=False, source=False)
+            status_by_day[day] = DoseStatus(taken=False, missed=True, unknown=False, source=False,
+                                            value=case.get('adherence_value'))
         else:
             pass  # unknown
     return status_by_day
 
 
-def _get_relevent_case(cases):
+def get_relevent_case(cases):
```
proteusvacuum: I'd rename this now that it is used outside of just this file. There's also a spelling mistake in this name. 
`get_primary_adherence_case` or something similar would be more appropriate.

custom/enikshay/tasks.py
```diff
@@ -163,10 +167,13 @@ def run_batch(self, case_ids):
             updates = []
             for episode in self._get_open_episode_cases(case_ids):
                 did_error = False
+                dose_status_by_date = None
                 update_json = {}
                 for updater in self.updaters:
                     try:
                         potential_update = updater(self.domain, episode).update_json()
+                        if isinstance(updater, EpisodeAdherenceUpdate):
```
proteusvacuum: You should put this into the `EpisodeAdherenceUpdate` class instead.

custom/enikshay/tasks.py
```diff
@@ -193,6 +206,11 @@ def run_batch(self, case_ids):
 
         return BatchStatus(update_count, noupdate_count, success_count, errors, case_batches, t.interval)
 
+    @staticmethod
+    def _update_ledger_values(episode_case, adherence_date, adherence_source, adherence_value):
```
proteusvacuum: this should go in the `EpisodeAdherenceUpdate` class, not here.

custom/enikshay/tasks.py
```diff
@@ -183,6 +190,12 @@ def run_batch(self, case_ids):
                     update_count += 1
                 else:
                     noupdate_count += 1
+                if dose_status_by_date:
```
proteusvacuum: this should go in the `EpisodeAdherenceUpdate` class, not here.

custom/enikshay/tasks.py
```diff
@@ -193,6 +206,11 @@ def run_batch(self, case_ids):
 
         return BatchStatus(update_count, noupdate_count, success_count, errors, case_batches, t.interval)
 
+    @staticmethod
+    def _update_ledger_values(episode_case, adherence_date, adherence_source, adherence_value):
+        if adherence_date and adherence_source and adherence_value:
+            update_ledger_for_adherence(episode_case, adherence_date, adherence_source, adherence_value)
```
proteusvacuum: currently, this will make a single form submission per episode case that needs to be updated, which was causing us issues before. this should be batched as we do with the case updates.

custom/enikshay/utils.py
```diff
@@ -0,0 +1,117 @@
...
+from corehq.util.quickcache import quickcache
+from casexml.apps.stock.mock import (
+    Balance,
+    Entry,
+)
+from corehq.apps.hqcase.utils import submit_case_blocks
+from corehq.apps.users.util import SYSTEM_USER_ID
+
+
+def get_adherence_cases_by_date(adherence_cases):
```
proteusvacuum: I think I would class this as a `case_util` and it could go in that file.

custom/enikshay/utils.py
```diff
@@ -0,0 +1,117 @@
+from __future__ import absolute_import
```
proteusvacuum: Would you want to call this file `ledger_utils` instead?

custom/enikshay/tasks.py
proteusvacuum: you can do something like 
```
ledger_updates += getattr(updater(self.domain, episode), ledger_updates)
``` 
which will then allow you to batch them in the same way as we do for case updates. (and then rename the `update_json` in all the updaters to `case_update_json` or something more descriptive. ).

proteusvacuum (changes requested):

custom/enikshay/tasks.py
mkangia: @proteusvacuum what issues did you face earlier with single form/episode?

proteusvacuum: we update ~20,000 episode cases a night and there is a time overhead in submitting a form, even if it only contains a single update. By batching the form submissions you can reduce the amount of time the task takes. If you don't batch them, I don't think this would complete in a reasonable amount of time.

custom/enikshay/utils.py
```diff
@@ -0,0 +1,117 @@
...
+                    entry.quantity = ledger_value
+                    balance.entry = entry
+                    return submit_case_blocks([balance.as_string()],
+                                              episode_case.domain,
+                                              SYSTEM_USER_ID)
+
+
+def update_ledger_for_adherence(episode_case, adherence_date, adherence_source, adherence_value):
+    entry_id = "date_%s" % adherence_date
+    update_ledger_with_episode(episode_case, entry_id, adherence_source, adherence_value)
```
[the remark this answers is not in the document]
mkangia: i just wanted to extract the logic of how the adherence date is mapped to an entry id so that `entry_id = "date_%s" % adherence_date` is not spilled out everywhere this update is needed.

renamed it to `update_episode_ledger_for_adherence`

custom/enikshay/utils.py
mkangia: not sure about this. Extracted this method out of the updater. Probably something not expected to happen.
will still add this.

custom/enikshay/tasks.py
mkangia: the one drawback i have seen to that is loading this form on case detail page. But i guess we ignored that considering the benefits of batching here.

mkangia:
@proteusvacuum Thanks for the review. Made all the changes.
One tricky situation here is tracking ledger updates and case updates separately. Not sure if there is a way to combine then into one.

custom/enikshay/const.py
```diff
@@ -184,3 +184,5 @@
 }
 
 LOCATION_SITE_CODE_MEHSANA = 'dto_gu_msn'
+FIXTURE_FOR_EPISODE_LEDGERS = "adherence_ledger_values"
```
proteusvacuum: nit: would be nice to keep these 2 names grammatically consistent, so this one would be `EPISODE_LEDGER_FIXTURE`

custom/enikshay/ledger_utils.py
```diff
@@ -0,0 +1,111 @@
...
+        all_items = FixtureDataItem.by_data_type(domain, fixture_id)
+        result = defaultdict(dict)
+        for item in all_items:
+            source = item.fields['adherence_source'].field_list[0].field_value
+            value = item.fields['adherence_value'].field_list[0].field_value
+            ledger_value = item.fields['ledger_value'].field_list[0].field_value
+            result[source][value] = ledger_value
+        return result
+    else:
+        return {}
```
proteusvacuum: should this also return a `defaultdict`? or do you want the call to `_get_all_fixture_items` to fail when there is no `fixture_id`? If so, might be worth just raising an exception here instead.

custom/enikshay/ledger_utils.py
```diff
@@ -0,0 +1,111 @@
...
+    entry.quantity = ledger_value
+    balance.entry = entry
+    return balance
+
+
+def bulk_update_ledger_cases(domain, ledger_updates):
+    case_blocks = []
+    for episode_case_id, entry_id, balance in ledger_updates:
+        balance = _ledger_update_xml(episode_case_id, entry_id, balance)
+        case_blocks.append(ElementTree.tostring(balance.as_xml()))
```
proteusvacuum: I think you can use `balance.as_string` here

custom/enikshay/ledger_utils.py
```diff
@@ -0,0 +1,111 @@
...
+    fixture_id = _adherence_values_fixture_id(domain)
+    if fixture_id:
+        return (_get_all_fixture_items(domain, fixture_id)
+                [adherence_source]
+                .get(adherence_value))
+
+
+def _ledger_update_xml(episode_case_id, entry_id, ledger_value):
+    balance = Balance()
+    balance.entity_id = episode_case_id
```
proteusvacuum: nit: Is there a reason you define the `Balance` and `Entry` nodes this way and not like:
```py
Balance(
    entity_id=episode_case_id,
    date=datetime.utcnow(),
    entry=Entry(id=entry_id, quantity=ledger_value),
    ...,
)
```

custom/enikshay/tasks.py
```diff
@@ -156,17 +164,24 @@ def run_batch(self, case_ids):
...
+            ledger_updates = []
             for episode in self._get_open_episode_cases(case_ids):
                 did_error = False
                 update_json = {}
                 for updater in self.updaters:
                     try:
-                        potential_update = updater(self.domain, episode).update_json()
+                        _updater = updater(self.domain, episode)
+                        potential_update = _updater.update_json()
+                        if isinstance(_updater, EpisodeAdherenceUpdate):
```
proteusvacuum: I still think you shouldn't have to reference the `EpisodeAdherenceUpdate` here at all by either checking if the updater that you are currently calling has the ledger_updates() method, or by adding that as a noop as the base class.

custom/enikshay/tasks.py
```diff
@@ -156,17 +164,24 @@ def run_batch(self, case_ids):
...
+            case_updates = []
+            ledger_updates = []
             for episode in self._get_open_episode_cases(case_ids):
                 did_error = False
                 update_json = {}
                 for updater in self.updaters:
                     try:
-                        potential_update = updater(self.domain, episode).update_json()
+                        _updater = updater(self.domain, episode)
+                        potential_update = _updater.update_json()
```
proteusvacuum: can you call this `case_update` since now `potential_update` is a misleading name

custom/enikshay/tasks.py
```diff
@@ -156,17 +164,24 @@ def run_batch(self, case_ids):
...
+                        if isinstance(_updater, EpisodeAdherenceUpdate):
+                            episode_ledger_updates = _updater.ledger_updates()
+                            if episode_ledger_updates:
```
proteusvacuum: I think you can just call `extend` here directly without checking if there are updates since:
```py
In [1]: x = ['a']

In [2]: x.extend([])

In [3]: x
Out[3]: ['a']
```

custom/enikshay/tasks.py
```diff
@@ -624,6 +646,16 @@ def check_and_return(self, update_dict):
...
             return None
 
+    def ledger_updates(self):
+        _updates = []
+        for dose_date, dose_status in self.dose_status_by_date.items():
+            if dose_status.source and dose_status.value:
+                expected_value = get_expected_fixture_value(self.domain, dose_status.source, dose_status.value)
+                entry_id = ledger_entry_id_for_adherence(dose_date)
+                if ledger_needs_update(self.domain, self.episode.case_id, entry_id, expected_value):
+                    _updates.append((self.episode.case_id, entry_id, expected_value))
```
proteusvacuum: I'm curious why you chose to pass this through as a tuple instead of just creating the `Balance` xml directly here?

mkangia: just following the same process that is being used for case updates.

custom/enikshay/tasks.py
mkangia: Updated to
```
if hasattr(_updater, 'ledger_updates'):
    ledger_updates.extend(_updater.ledger_updates())
```

Unclear though why that makes it better?

custom/enikshay/ledger_utils.py
mkangia: i think its better to avoid exception here and cause the enikshay task to fail or otherwise the overhead to handle the exception. Does that sound right?
Updated to return default dict to avoid exception later. Also added the check to skip update if the expected value is not present in the fixture.

mkangia:
@proteusvacuum Made the required changes and replied where needed.

custom/enikshay/tasks.py
```diff
@@ -177,12 +177,10 @@ def run_batch(self, case_ids):
...
-                        potential_update = _updater.update_json()
-                        if isinstance(_updater, EpisodeAdherenceUpdate):
-                            episode_ledger_updates = _updater.ledger_updates()
-                            if episode_ledger_updates:
-                                ledger_updates.extend(episode_ledger_updates)
-                        update_json.update(get_updated_fields(episode.dynamic_case_properties(), potential_update))
+                        case_update = _updater.update_json()
+                        if hasattr(_updater, 'ledger_updates'):
+                            ledger_updates.extend(_updater.ledger_updates())
+                        update_json.update(get_updated_fields(episode.dynamic_case_properties(), case_update))
```
proteusvacuum: nit: I think this should be called `case_update_json`

proteusvacuum (approved):
Thanks for the changes!

PR #18897, opened 2017-12-13
make threshold for LogLongRequestMiddleware 10 minutes

dannyroberts:
instead of 2, in order to reduce the number of sentry events we get


snopoke (approved):
Wow, 10mins!

dannyroberts:
Ha, yeah. If we get very few of these, we can ratchet it down, but right now we're getting like 20k per day which isn't that helpful.

PR #18899, opened 2017-12-13
Temporary: initialize analytics loggers on module definition

orangejenny:
App preview is breaking for enikshay, I think because of the order in which document ready handlers are called. Haven't been able to reproduce locally or on prod.

Loggers shouldn't be created until the document is ready, because they depend on initial page data, so https://github.com/dimagi/commcare-hq/pull/18859/commits/edac51e393570466231dbbcf9b64e7f683d3a9d8 moved them to a document ready handler.

However, the ready handler that creates the loggers might not execute before other handlers that try to call them. The right way to fix this is probably to treat logger initialization as a promise. This PR is a band-aid that will initialize the loggers immediately (and possibly incorrectly, that is, logs may not display when they should be displaying...but that's not a user-facing problem).

Also turning off the assertion added in https://github.com/dimagi/commcare-hq/pull/18860 since this re-introduces code that tries to access initial page data before the document is ready, which would mean a deluge of email.

code buddy @gcapalbo

### DIFF
--- a/corehq/apps/analytics/static/analytix/js/drift.js
+++ b/corehq/apps/analytics/static/analytix/js/drift.js
@@ -21,7 +21,7 @@ hqDefine('analytix/js/drift', [
         _global = initialAnalytics.getFn('global'),
         _data = {},
         _drift = {},
-        logger;
+        logger = logging.getLoggerForApi('Drift');
 
     var __init__ = function () {
         logger = logging.getLoggerForApi('Drift');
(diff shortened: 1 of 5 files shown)


millerdev (approved):
Syntax looks fine. Don't really understand what's being done here.

orangejenny:
offlined with @millerdev , don't have this issue totally nailed down but think there's a good chance this PR will fix it and don't think it'll make things worse

orangejenny:
no test coverage

PR #18900, opened 2017-12-13
Switch mobile endpoints to @mobile_auth

esoergel:
Followup from https://github.com/dimagi/commcare-hq/pull/18846

I believe these endpoints are all intended to be accessed from mobile and should be decorated thusly.  Note that this drops support for digest auth for these endpoints.  Essentially, this reverts the behavior change from https://github.com/dimagi/commcare-hq/pull/18274 (they [defaulted to basic auth](https://github.com/dimagi/commcare-hq/blob/bba48ddd9a54ee6ff6b4789e2954a52a94d48c10/corehq/apps/ota/views.py), since no alternate default was provided).  Cal switched `restore` and `fetch_key_records` in https://github.com/dimagi/commcare-hq/pull/18826, this similarly switches these other endpoints.

@calellowitz @czue


calellowitz (approved):

czue (approved):
:clap:

PR #18904, opened 2017-12-14
PatientFinders find patients in OpenMRS that match cases in CommCare

kaapstorm:
Getting there, @dannyroberts 

cc @sheelio buddy @mkangia 

:blowfish:


corehq/motech/openmrs/finders.py
```diff
@@ -0,0 +1,165 @@
...
+            for patient in response.json()['results']:
+                score = self.get_score(patient, case)
+                if score >= self.threshold:
+                    candidates[patient['uuid']] = PatientScore(patient, score)
+        if not candidates:
+            return []
+        patients_scores = sorted(candidates.values(), key=lambda cand: cand.score, reverse=True)
+        if len(patients_scores) == 1:
+            return [patients_scores[0].patient]
+        if float(patients_scores[0].score) / patients_scores[1].score > 1 + self.confidence_margin:
```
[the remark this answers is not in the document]
kaapstorm: Yeah but no, Mr Stickler. That numerator is a float, so that's obviously [true division](https://www.python.org/dev/peps/pep-0238/).

Do you have any strong feelings about "future division" vs casting, reviewers?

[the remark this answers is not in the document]
dannyroberts: I think it's just saying you should include

```
from __future__ import division
```
https://github.com/janjur/readable-pylint-messages/blob/master/README.md#old-division

as a way to ensure that this remains python3 compatible. Even though in this case what you wrote here I think would be python3-compatible, not casting and using the `__future__` import marks this file's division as python3-compatible so when we make the switch no one has to think about whether this particular line is or isn't compatible. Does that make sense? I think in general that's why we turned on stricter py3-compat linter errors.

corehq/motech/openmrs/finders.py
```diff
@@ -0,0 +1,166 @@
...
+
+from jsonpath_rw import parse
+
+
+PATIENT_FINDERS = []
+
+
+def register_patient_finder(class_):
+    PATIENT_FINDERS.append(class_)
+    return class_
```
mkangia: nit: is this return required because this a decorator and should return the `class_`?

corehq/motech/openmrs/finders.py
```diff
@@ -0,0 +1,166 @@
...
+        #         "birthdate": {
+        #           "doc_type": "CaseProperty",
+        #           "case_property": "dob"
+        #         }
+        #       },
+        #       // ...
+        #     }
+        #
+
+        for person_prop, value_source in case_config['person_properties'].items():
```
mkangia: Should we use `get` here `case_config['person_properties']` instead to avoid failure in case the key is missing? Similarly for other 3 `property_map` iteration after this..

corehq/motech/openmrs/finders.py
```diff
@@ -0,0 +1,166 @@
...
+        Matches cases to patients. Returns a list of patients, each
+        with a confidence score >= THRESHOLD
+        """
+        PatientScore = namedtuple('PatientScore', ['patient', 'score'])
+        self.set_property_map(case_config)
+
+        candidates = {}  # key on OpenMRS UUID to filter duplicates
+        for prop in self.searchable_properties:
+            value = case.get_case_property(prop)
+            response = requests.get('/ws/rest/v1/patient', {'q': value, 'v': 'full'})
```
mkangia: `/ws/rest/v1/patient` looks like a constant? Can just extract it out.

corehq/motech/openmrs/repeater_helpers.py
```diff
@@ -226,7 +227,20 @@ def get_subresource_instances(requests, person_uuid, subresource):
     )).json()['results']
 
 
-def get_patient(requests, info, openmrs_config, problem_log):
+def find_patient(domain, requests, case_id, openmrs_config):
+    PatientFinder = to_function(openmrs_config.patient_finder)
+    if not PatientFinder:
+        return None
```
mkangia: Can this return a message/exception that the `finder` was not found? Combining this with `None` when no patients are found would not make it clear why there were no records found.

mkangia (approved):
Overall looks good.. Didn't understand the use case and overall purpose so reviewed till the extent i could.

sheelio:
@kaapstorm 
Still trying to figure out exactly how this is supposed to work.  My understanding was:
- If there is a match on case.openmrs_id, this is 100% trustworthy and we should just use it
- If there is a match on something like bahmni_id or household_id, we should do a secondary validation on something like DOB, first_name+last_name
- Once a match happens, we update the case with the OpenMRS ID (so in the future we can just rely on it)

Is this what the code is doing?

kaapstorm:
@sheelio, It does the first two steps, but not yet the last one. 

The first step, "If there is a match on case.openmrs_id", happens in the [`get_patient` function](https://github.com/dimagi/commcare-hq/pull/18904/commits/600d41ab303e41c60a41905d9ac973f594c9d980#diff-ad1809f9bda087777a6c93629dda7a67R248). It uses the ID matchers that are part of the normal OpenMRS case config. If "openmrs_id" (or any other 100% certain ID matcher that might have been configured) does match, then we use that patient. 

Otherwise we go to Step 2, which tries to find the patient. [Properties are compared in order](https://github.com/dimagi/commcare-hq/pull/18904/commits/8fb3bb9a55982f00bb06aa9bd05ac1baf487d558#diff-b2bb772d97ec070579013d8ebebf68feR16) and we validate all of them. First, `bahmni_id`, then `household_id`, then `dob`, `first_name`, `last_name` and finally `municipality`. 

It's a little more complicated than what we discussed because we can't search on all properties, only IDs and last name, and searches are expensive but comparisons are cheap. So we search against each of the searchable properties, and compare all the weighted properties. Found patients are stored in a dictionary keyed on their OpenMRS ID to prevent duplicate search results.

Instead of just accepting all patients that score higher than 1 (the default but configurable value of `threshold`), we have a final comparison. Imagine there are 2 John Smith's from Somerville who are born on Jan 1 1980. They will both score 1 for DOB + full name + municipality. But one of them has a matching household_id and the other doesn't. So his score is 1.9 and the other is 1.0. If one patient's score is more that 10% (by default but configurable) better than another, we pick him. If there is less than a 10% difference we return both (and neither will be selected, but they will be logged).

We will log all these results, so we can see whether our weights and thresholds make sense. 

Let me know if that answers your question, and if you have any concerns with that approach.

If we can get a real or typical sample dataset, and ignore the OpenMRS IDs, and run these comparisons, and then compare our results with known matches, we could refine (programmatically?) our weights and 10% margin, until we get as close as possible to the known matches, while minimising type 2 / false positive errors.

corehq/motech/openmrs/finders.py
```diff
@@ -135,6 +138,25 @@ def weights():
...
         return sum(weights())
 
+    def save_match_id(self, case, case_config, patient):
+        """
+        If we are confident of the patient matched to a case, save
+        the patient's ID to the case.
+        """
+        id_map = {m['identifier_type_id']: m['case_property'] for m in case_config['id_matchers']}
+        case_update = {}
+        for identifier in patient['identifiers']:
```
kaapstorm: Example value for `patient['identifiers']`:
```json
  "identifiers": [
    {
      "display": "Patient Identifier = GAN200036",
      "uuid": "c0d20f06-e3c9-406e-95dd-ca55cc2e3818",
      "identifier": "GAN200036",
      "identifierType": {
        "uuid": "81433852-3f10-11e4-adec-0800271c1b75",
        "display": "Patient Identifier",
        "links": [ ,,, ]
      },
      "location": null,
      "preferred": true,
      "voided": false,
      "links": [ ,,, ],
      "resourceVersion": "1.8"
    }
  ],
```

corehq/motech/openmrs/finders.py
```diff
@@ -155,12 +177,16 @@ def find_patients(self, requests, case, case_config):
             return []
         patients_scores = sorted(candidates.values(), key=lambda cand: cand.score, reverse=True)
         if len(patients_scores) == 1:
-            return [patients_scores[0].patient]
+            patient = patients_scores[0].patient
+            self.save_match_id(case, case_config, patient)
```
kaapstorm: `self.save_match_id` will save the patient's value of the identifier(s) listed in the `OpenmrsCaseConfig`'s "id_matchers" to the case, so that next time the case's patient will be matched immediately. An example value of OpenmrsCaseConfig.id_matchers would look like:
```json

  "id_matchers": [
    {
      "case_property": "openmrs_id", 
      "identifier_type_id": "8d793bee-c2cc-11de-8d13-0010c6dffd0f", 
      "doc_type": "IdMatcher"
    }
  ], 
```

corehq/motech/openmrs/finders.py
kaapstorm: `case_config` is an instance of [OpenmrsCaseConfig](https://github.com/dimagi/commcare-hq/blob/master/corehq/motech/openmrs/openmrs_config.py#L90) so it will always have `person_properties` ... but you make a good point, because that is not obvious. Maybe it would be better to use it as an attribute, `case_config.person_properties`, instead of a key.

corehq/motech/openmrs/finders.py
kaapstorm: Hmmm. That's true, [but][1] [there][2] [are][3] [a][4] [lot][5] [of][6] [endpoints][7] [like][8] [it][9] ... [and][10] [we][11] [only][12] [refer][13] to each of them once ... except that one, now that I'm going through them! I'll reuse [search_patients()][14], so at least all the endpoints are in the same module.


  [1]: https://github.com/dimagi/commcare-hq/blob/master/corehq/motech/openmrs/repeater_helpers.py#L89
  [2]: https://github.com/dimagi/commcare-hq/blob/master/corehq/motech/openmrs/repeater_helpers.py#L98
  [3]: https://github.com/dimagi/commcare-hq/blob/master/corehq/motech/openmrs/repeater_helpers.py#L129
  [4]: https://github.com/dimagi/commcare-hq/blob/master/corehq/motech/openmrs/repeater_helpers.py#L141
  [5]: https://github.com/dimagi/commcare-hq/blob/master/corehq/motech/openmrs/repeater_helpers.py#L145
  [6]: https://github.com/dimagi/commcare-hq/blob/master/corehq/motech/openmrs/repeater_helpers.py#L160
  [7]: https://github.com/dimagi/commcare-hq/blob/master/corehq/motech/openmrs/repeater_helpers.py#L167
  [8]: https://github.com/dimagi/commcare-hq/blob/master/corehq/motech/openmrs/repeater_helpers.py#L184
  [9]: https://github.com/dimagi/commcare-hq/blob/master/corehq/motech/openmrs/repeater_helpers.py#L200
  [10]: https://github.com/dimagi/commcare-hq/blob/master/corehq/motech/openmrs/repeater_helpers.py#L213
  [11]: https://github.com/dimagi/commcare-hq/blob/master/corehq/motech/openmrs/repeater_helpers.py#L223
  [12]: https://github.com/dimagi/commcare-hq/blob/master/corehq/motech/openmrs/repeater_helpers.py#L359
  [13]: https://github.com/dimagi/commcare-hq/blob/master/corehq/motech/openmrs/repeater_helpers.py#L363
  [14]: https://github.com/dimagi/commcare-hq/blob/master/corehq/motech/openmrs/repeater_helpers.py#L166

corehq/motech/openmrs/finders.py
kaapstorm: Yes, exactly.

```python
@foo
class Bar:
    pass
```
is [the same](https://www.python.org/dev/peps/pep-3129/#id11) as
```python
class Bar:
    pass
Bar = foo(Bar)
```

mkangia:
@kaapstorm is this waiting on me for review?

kaapstorm:
@mkangia no. I just want a bit more testing, which keeps getting pushed back by other things. -- Hopefully tomorrow. 🤞

mkangia:
cool. @kaapstorm please let me know when and if you want me to take another look

kaapstorm:
@mkangia I'll rebase off #19525 ... and keep testing. (Nice to finally be making progress on this.)

corehq/motech/openmrs/jsonpath.py
```diff
@@ -0,0 +1,35 @@
+from __future__ import absolute_import
+
+from jsonpath_rw import JSONPath
+
+
+class WhereCmp(JSONPath):
```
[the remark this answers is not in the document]
kaapstorm: Ignoring. Implemented `__eq__` to conform to other JSONPath subclasses, but none of them implement `__hash__`, so I'm not going to either.

dannyroberts:
@kaapstorm looks like there was some recent work here, but not sure whether you're still adding things or you're just waiting for a final review. Just let me know if/when you need a review and I'll spring into action

kaapstorm:
Hey @dannyroberts the conversation seems to have migrated to PR #19836 which builds on this. I think we're pretty close to merging, so if you have a chance, it would be great if you could eyeball it.

PR #18906, opened 2017-12-14
Allow date opened to be updated with case importer

proteusvacuum:
https://manage.dimagi.com/default.asp?266615#1434384

To update the `date_opened` (or `opened_on` or `opened_date`) through the importer, turn on that flag and add a column called `date_opened`. 

I'm not sure if we should add some messaging to the UI for when you are trying to do this without the flag enabled, because I think right now it will just ignore that property (which maintains the current behaviour).

@sravfeyn


sravfeyn:
If the flag is not enabled you could append `date_opened` to `EXCLUDED_FIELDS`, that probably will explicitly warn user.

sravfeyn:
This seems it will lead to some unforeseen incosistencies. Is this needed for eNikshay, isn't there a better way to get what is needed without doing this. Just curious.

PR #18907, opened 2017-12-14
ICDS: Aggregate Performance Indicator Updates

gcapalbo:
https://manage.dimagi.com/default.asp?251016

Fixes an issue where an `IndicatorError` was being raised when it shouldn't be.

@emord

### DIFF
--- a/custom/icds/messaging/indicators.py
+++ b/custom/icds/messaging/indicators.py
@@ -403,7 +403,7 @@ class LSAggregatePerformanceIndicator(LSIndicator):
         num_weigh = self.get_value_from_fixture(self.weighed_fixture, 'open_weighed')
         num_weigh_avail = self.get_value_from_fixture(self.weighed_fixture, 'open_count')
         num_days_open = int(self.get_value_from_fixture(self.days_open_fixture, 'awc_opened_count'))
-        num_awc_locations = len(self.days_open_fixture.findall('./rows/row[@is_total_row="False"]'))
+        num_awc_locations = len(self.awc_locations)
         if num_awc_locations:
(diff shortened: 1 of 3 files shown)


custom/icds/messaging/indicators.py
```diff
@@ -201,7 +201,7 @@ def get_value_from_fixture(self, fixture, attribute):
                         "Attribute {} not found in restore for AWC {}".format(attribute, location_name)
                     )
 
-        raise IndicatorError("AWC {} not found in the restore".format(location_name))
+        return 0
```
gcapalbo: from what i can tell, for all four of the ucrs being used, when the AWC id is not found it means that there were no forms or cases for that AWC that match the ucr filters, so the aggregation for these fields should be represented with a 0 instead of raising an exception

custom/icds/messaging/indicators.py
```diff
@@ -403,7 +403,7 @@ def get_messages(self, language_code=None):
         num_weigh = self.get_value_from_fixture(self.weighed_fixture, 'open_weighed')
         num_weigh_avail = self.get_value_from_fixture(self.weighed_fixture, 'open_count')
         num_days_open = int(self.get_value_from_fixture(self.days_open_fixture, 'awc_opened_count'))
-        num_awc_locations = len(self.days_open_fixture.findall('./rows/row[@is_total_row="False"]'))
+        num_awc_locations = len(self.awc_locations)
```
gcapalbo: using `len(self.awc_locations)` as the denominator here is better for the same reason - AWCs could drop out of the actual report

emord (approved):

PR #18915, opened 2017-12-15
eNikshay: Add custom reminder recipients

gcapalbo:
Includes the ability to tie in additional recipients for two alerts in one specific district.


custom/enikshay/messaging/custom_recipients.py
```diff
@@ -40,3 +44,58 @@ def agency_user_case_from_voucher_fulfilled_by_id(handler, reminder):
 
     # get_usercase() just returns None if the user case does not exist; no exceptions raised
     return user.get_usercase()
+
+
+@quickcache(['location.location_id'], timeout=4 * 60 * 60)
+def located_in_mehsana(location):
+    return location.get_ancestors().filter(site_code=LOCATION_SITE_CODE_MEHSANA).count() > 0
```
proteusvacuum: probably not a big deal, but since I think this only refers to a single location (https://enikshay.in/a/enikshay/settings/locations/fa7472fe0c9751e5d14595c1a08bf499/), you could have used the indexed `location_id` property here instead of `site_code`, which I don't think is indexed.

custom/enikshay/messaging/custom_recipients.py
```diff
@@ -40,3 +44,58 @@ def agency_user_case_from_voucher_fulfilled_by_id(handler, reminder):
...
+def get_user_cases_at_location(location):
+    users = get_all_users_by_location(location.domain, location.location_id)
+    user_cases = [u.memoized_usercase for u in users if isinstance(u, CommCareUser) and u.is_active]
+    return [case for case in user_cases if case]
+
+
+def beneficiary_registration_recipients(handler, reminder):
+    beneficiary = reminder.case
+
+    owner_location = SQLLocation.by_location_id(beneficiary.owner_id)
```
proteusvacuum: The `by_location_id` method doesn't use the `active_objects` manager, which is fine in this case, however, I think we might want to change how that method works. If you used `SQLLocation.active_objects.get()` you could also do the domain filter right in the query.

proteusvacuum (approved):

custom/enikshay/messaging/custom_recipients.py
gcapalbo: agreed, updated: https://github.com/dimagi/commcare-hq/pull/18915/commits/3974cb3e7f746e2e4c7363aa5bc53fdd9956b408

custom/enikshay/messaging/custom_recipients.py
gcapalbo: it actually won't matter in this case since the heart of the query is going to be the `.get_ancestors()` part, which retrieves the ancestors of the current location. so from there, it's just filtering a handful of locations using site code. but i chose to use site code since i try to allow all constants to be used cross-domain or cross-environment in case this is ever used in a test environment

proteusvacuum: 👍

PR #18916, opened 2017-12-15
Don't filter based on add or dob

emord:
> OK cool, so basically if anything goes wrong with this while you're gone, we should PR and merge this? 

That's correct. it puts the filtering back to it's previous state and basically means that  #18772 just added a couple of unused columns

@dannyroberts @czue


czue:
:+1: :pray:

emord:
Since we've solved the discrepancy in another way with related doc rebuilding: https://github.com/dimagi/commcare-hq/pull/18843 We should be able to remove these filters.

PR #18918, opened 2017-12-15
Make drift ID secret

proteusvacuum:
@orangejenny @biyeun 
This previously would have meant that anyone installing HQ themselves would have access to Mackenzie for help. Which is cool, but I don't think was our intention.

Depends on: https://github.com/dimagi/commcarehq-ansible/pull/1150
@sravfeyn


dannyroberts (approved):

orangejenny:
thanks!

PR #18920, opened 2017-12-15
Don't filter on users if there are none

proteusvacuum:
http://manage.dimagi.com/default.asp?266935#1432559
@sravfeyn


dannyroberts (approved):
sweet change!

proteusvacuum:
🍬

PR #18922, opened 2017-12-15
Added soft assert for non dimagi superusers.

emord:
@krsdimagi 🔐 

@czue

### DIFF
--- a/corehq/apps/hqadmin/tasks.py
+++ b/corehq/apps/hqadmin/tasks.py
@@ -3,11 +3,16 @@ from datetime import date, timedelta
 
 from celery.schedules import crontab
 from celery.task.base import periodic_task
+from django.contrib.auth import get_user_model
+from django.db.models import Q
 
 from corehq.apps.hqadmin.models import HistoricalPillowCheckpoint
+from corehq.util.soft_assert import soft_assert
 from dimagi.utils.logging import notify_error
 from pillowtop.utils import get_couch_pillow_instances
 from .utils import check_for_rewind
+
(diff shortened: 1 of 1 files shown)


krsdimagi:
Nice, thanks!

corehq/apps/hqadmin/tasks.py
```diff
@@ -25,9 +30,22 @@ def check_pillows_for_rewind():
...
+
 @periodic_task(run_every=crontab(hour=0, minute=0), queue='background_queue')
 def create_historical_checkpoints():
     today = date.today()
     thirty_days_ago = today - timedelta(days=30)
     HistoricalPillowCheckpoint.create_pillow_checkpoint_snapshots()
     HistoricalPillowCheckpoint.objects.filter(date_updated__lt=thirty_days_ago).delete()
+
+
+@periodic_task(run_every=crontab(minute=0), queue='background_queue')
```
esoergel: This might be more frequently than needed (we'll get an email every hour until that user is removed), but we can always change it if it becomes annoying.  Have you checked the live environments to see if there are currently non-dimagi users?

esoergel (approved):

corehq/apps/hqadmin/tasks.py
emord: I'm ok with (maybe even pro)  annoying people for a potential security breach

czue (approved):
Not relevant for this PR, but generally another scenario that is more likely and not caught by this would be a former employee retaining superuser status.

PR #18929, opened 2017-12-17
UCR Summary PR Feedback

sheelio:
@esoergel 
I merged https://github.com/dimagi/commcare-hq/pull/18895 without responding to your feedback.  Here are the requested changes

### DIFF
--- a/corehq/apps/userreports/templates/userreports/summary_data_source.html
+++ b/corehq/apps/userreports/templates/userreports/summary_data_source.html
@@ -54,7 +54,7 @@
     </tr>
 {% for expression in named_expression_summary %}
     <tr>
-        <td id="NamedE:{{ expression.name }}">{{ expression.name }}</td>
+        <td id="{{ named_expression_prefix }}:{{ expression.name }}">{{ expression.name }}</td>
         <td>{{ expression.comment }}</td>
         <td><pre>{{ expression.readable_output|safe }}</pre></td>
     </tr>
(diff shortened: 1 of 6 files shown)


esoergel (approved):

esoergel (approved):

corehq/apps/userreports/templates/userreports/summary_data_source.html
```diff
@@ -54,7 +54,7 @@ <h2>Named Expressions</h2>
     </tr>
 {% for expression in named_expression_summary %}
     <tr>
-        <td id="NamedE:{{ expression.name }}">{{ expression.name }}</td>
+        <td id="{{ named_expression_prefix }}:{{ expression.name }}">{{ expression.name }}</td>
```
esoergel: Couldn't this also just use `{{ expression }}` or something to use the implementation in `__str__`?  It's unfortunate to have this duplicated, because if `__str__` changes in the future, this link would break.

esoergel: actually I guess this is duplicated in the link generation thing anyways.  Hmm, that's tricky.  I suppose you could do something else for link so the string display of the expression and the HTML id are different things.  This seems fine, no strong opinions.

sheelio: Cool - I guess we could generate the link in the View and pass that through to use here if we want it less brittle.

PR #18935, opened 2017-12-18
Clean up location model methods

esoergel:
@mkangia @millerdev
Just ran across these two methods that looked like they could be dropped.

`self.path` calls `self.get_ancestors()`, so `.parents` was doing a query to get
all ancestor IDs, then passing those IDs back to the DB in a second query.

`get_parent_of_type` can be replaced by the `get` queryset method, which will
raise an exception if multiple results are returned.

### DIFF
--- a/custom/enikshay/location_utils.py
+++ b/custom/enikshay/location_utils.py
@@ -1,18 +1,19 @@
 from __future__ import absolute_import
 from collections import namedtuple
 
+from corehq.apps.locations.models import SQLLocation
 from custom.enikshay.exceptions import NikshayLocationNotFound, NikshayCodeNotFound
 
 
 def get_health_establishment_hierarchy_codes(location):
     def get_parent(parent_type):
-        parent = location.get_parent_of_type(parent_type)
-        if not parent:
+        try:
+            return location.get_ancestors().get(location_type__name=parent_type)
+        except SQLLocation.DoesNotExist:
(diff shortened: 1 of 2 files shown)


millerdev (approved):
nice refactor/cleanup

custom/enikshay/location_utils.py
```diff
@@ -1,18 +1,19 @@
...
+from corehq.apps.locations.models import SQLLocation
 from custom.enikshay.exceptions import NikshayLocationNotFound, NikshayCodeNotFound
 
 
 def get_health_establishment_hierarchy_codes(location):
     def get_parent(parent_type):
-        parent = location.get_parent_of_type(parent_type)
-        if not parent:
+        try:
+            return location.get_ancestors().get(location_type__name=parent_type)
```
mkangia: Thanks @esoergel. Quick question. i checked that get returned a single record here. Is that because it finds a single match and in case of multiple(though not possible in case of locations) would return the first?

esoergel: Actually `.get` will throw an error if there's not precisely one match.  If there's no match, you `DoesNotExist` is raised, if there are more than one matches, `MultipleObjectsReturned` is raised.  As you said, that shouldn't be possible with that query, and since it was already raising an exception if that happened (`ValueError`), I figured it'd be safe enough to just rely on Django's `.get` behavior.

PR #18937, opened 2017-12-18
Revert "Mk 265955 kill user sync history"

sheelio:
Reverts dimagi/commcare-hq#18789
This restores the user sync history report.  IT might be useful to some enterprise projects and want to figure out what its being used for before we either remove it or move it to the support feature flag.


sheelio:
Talked to Clayton and he's ok just killing the report.

PR #18943, opened 2017-12-19
manually 404 on tableau

sravfeyn:


custom/icds_reports/views.py
```diff
@@ -116,6 +117,9 @@ def couch_user(self):
         return self.request.couch_user
 
     def get_context_data(self, **kwargs):
+        if settings.SERVER_ENVIRONMENT == 'icds-new':
```
dannyroberts: Can you use the new ICDS_ENVS setting variable as introduced here: https://github.com/dimagi/commcare-hq/pull/18934/files?

custom/icds_reports/reports/reports.py
```diff
@@ -1,6 +1,6 @@
 from __future__ import absolute_import
 from django.urls import reverse
-
+from django.http import Http404
```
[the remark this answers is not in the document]
dannyroberts: I think this is real/easily addressable.

dannyroberts (changes requested):

custom/icds_reports/views.py
sravfeyn: I thought it would be nice to keep the old one ON, in case GS wants to have a look at old one for any reason.

dannyroberts: oh... but isn't the old one just totally down now? But okay I suppose that's fine.

dannyroberts (approved):

PR #18948, opened 2017-12-19
ICDS: Simplify Expression Output

sheelio:


sheelio:
Urgh, got things in a bad broken state

PR #18950, opened 2017-12-19
ICDS: Improve UCR output

sheelio:


sheelio:
I think there's a cleaner way to do this.

PR #18952, opened 2017-12-19
fix subtree-merge for the edge case of a blank path

dannyroberts:


dannyroberts:
noticed this is still open. Minor change we can merge anytime. @millerdev buddy

PR #18953, opened 2017-12-19
Use new decorator

proteusvacuum:
@esoergel 
This is currently breaking tests.


custom/enikshay/integrations/ninetyninedots/views.py
```diff
@@ -98,7 +98,7 @@ def update_patient_details(request, domain):
 
 
 @toggles.NINETYNINE_DOTS.required_decorator()
-@login_or_digest_or_basic_or_apikey()
+@api_auth()
```
esoergel: This shouldn't have the `()`

esoergel (approved):

PR #18956, opened 2017-12-19
Case migrations work-in-progress

esoergel:
@anyone this is all feature flagged, so it should be able to be merged without risk.  It'll be quite a while before this is actually used, just want to get it in to source control for development.

@wpride

### DIFF
--- a/corehq/apps/case_migrations/views.py
+++ b/corehq/apps/case_migrations/views.py
@@ -1,0 +1,35 @@
+from __future__ import absolute_import
+from __future__ import unicode_literals
+
+from django.contrib import messages
+from django.http import HttpResponseRedirect
+from django.utils.decorators import method_decorator
+from django.utils.translation import ugettext as _
+from django.views.generic import FormView
+
+from corehq.apps.domain.decorators import domain_admin_required
+from corehq.apps.domain.views import BaseDomainView
+from corehq.util import reverse
+
+from .forms import MigrationForm
+
+
+@method_decorator(domain_admin_required, name='dispatch')
+class BaseMigrationView(BaseDomainView):
+    section_name = "Case Migrations"
+
+    @property
+    def section_url(self):
+        return reverse(MigrationView.urlname, args=(self.domain,))
+
+
+# TODO feature flag
+class MigrationView(BaseMigrationView, FormView):
+    urlname = 'case_migration'
+    template_name = 'case_migrations/migration.html'
+    form_class = MigrationForm
+
+    def form_valid(self, form):
+        messages.add_message(self.request, messages.SUCCESS,
+                             _('Migration submitted successfully!'))
+        return HttpResponseRedirect(self.page_url)
--- a/corehq/apps/case_migrations/views.py
+++ b/corehq/apps/case_migrations/views.py
@@ -15,6 +15,7 @@ from corehq.apps.domain.decorators import domain_admin_required
 from corehq.apps.domain.models import Domain
 from corehq.apps.domain.views import BaseDomainView
 from corehq.form_processor.interfaces.dbaccessors import CaseAccessors
+from corehq.toggles import WEBAPPS_CASE_MIGRATION
 from corehq.util.timer import TimingContext
 from corehq.util import reverse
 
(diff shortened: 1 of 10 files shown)


esoergel:
@sravfeyn code buddy ping

esoergel:
@millerdev @mkangia code buddies ping

corehq/apps/case_migrations/views.py
```diff
@@ -0,0 +1,86 @@
...
+        messages.add_message(self.request, messages.SUCCESS,
+                             _('Migration submitted successfully!'))
+        return HttpResponseRedirect(self.page_url)
+
+
+def get_related_case_ids(domain, case_id):
+    from casexml.apps.case.templatetags.case_tags import get_case_hierarchy
+    case = CaseAccessors(domain).get_case(case_id)
+    child_cases = {c.case_id for c in get_case_hierarchy(case, {})['case_list']}
+    indexed_cases = {index_info.referenced_id for index_info in case.indices}
```
millerdev: Looks like `get_case_hierarchy` pulls full cases out of the db and uses `case.reverse_indices` to load related cases. Might be more efficient to use [`get_related_indices`](https://github.com/dimagi/commcare-hq/blob/9bb1858e60025df1cecd05dbf775ff1731ea0dea/corehq/form_processor/interfaces/dbaccessors.py#L344), which only loads case_ids and does the hierarchy traversal in the db (non-recursively, multiple calls are necessary). See [livequery](https://github.com/dimagi/commcare-hq/blob/master/corehq/ex-submodules/casexml/apps/phone/data_providers/case/livequery.py#L198-L212) for more details.

millerdev (approved):
One comment. Looks powerful.

corehq/apps/case_migrations/views.py
esoergel: Nice, thanks for the tip, I'll switch it over.  Know if it's possible to also get the indexed cases in the same query?  Right now this is in the proof-of-concept phase, so I'm not concerned about performance, but if/when this gets used at all seriously this functionality will need to be highly optimized.

millerdev: > Know if it's possible to also get the indexed cases in the same query?

Yes, although as we already discovered, this only loads extension cases in the one direction (descendants). `get_related_indices` was created to traverse the [graph of live cases for a device](https://github.com/dimagi/commcare-hq/blob/master/docs/restore-logic.rst#device-restore-optimization), which starts from owned/open cases on a device and finds all other live cases. I'm not sure what your requirements are, but guessing it's broader than that (all cases that have not been deleted? or possibly all cases in the domain that exist in any state?).

millerdev: Unless we only want descendant extension cases, which I think we do not, these two statements seem incompatible:

> ...eventually this aspect of the restore should use the same code as the actual restore.

> ...it seems reasonable to say that we only traverse descendant cases.

The restore code only traverses to descendants if they are extension cases. This is because of how it picks its starting point: open, owned cases that are not extension cases.

It sounds like you want to start from an arbitrary starting point, which may or may not (?) be a "live" case, and traverse child and extension cases from there. I think this will require a different query than `get_related_indices`.

esoergel: I see, are you saying that because live child cases will be included in the restore anyways, they're not included in `get_related_indices`?  That makes sense. It's premature to write a custom query for this though, so I think I'll stick with `get_case_hierarchy` for now.

esoergel: @millerdev I just dropped the WIP commit and removed the code that traversed to cases "above" the primary one, so now it should only see descendant cases.  That code will certainly need revisiting before this is ready to go, but I'm happy with this for the POC phase.  Would you mind checking out that last commit?

millerdev: > are you saying that because live child cases will be included in the restore anyways, they're not included in `get_related_indices`?

Almost. Any live child case will either be open and directly owned by the device owner, or a parent of another such case. None of the liveness rules keep a child alive because its parent is alive, but a parent will stay alive as long as its child is alive.

millerdev (approved):

PR #18957, opened 2017-12-19
Ensure you are on latest master before running update translations

proteusvacuum:
I wasn't on latest master when running this and ran into issues. 

@dannyroberts 
@sravfeyn

### DIFF
--- a/scripts/update-translations.sh
+++ b/scripts/update-translations.sh
@@ -22,6 +22,14 @@ then
     abort "You must be on master to run this command."
 fi
 
+git fetch
+tracking_branch=`git rev-parse --abbrev-ref --symbolic-full-name @{u}`
+upstream_changes=`git rev-list HEAD...$tracking_branch --count`
+if [[ $upstream_changes > 0 ]]
+then
+    abort "There are changes upstream that you should pull with 'git pull' before running this command."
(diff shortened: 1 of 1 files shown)


kaapstorm (approved):
Nice. This has caught me before too.

scripts/update-translations.sh
```diff
@@ -22,6 +22,14 @@ then
     abort "You must be on master to run this command."
 fi
 
+git fetch
+tracking_branch=`git rev-parse --abbrev-ref --symbolic-full-name @{u}`
+upstream_changes=`git rev-list HEAD...$tracking_branch --count`
```
dannyroberts: interesting, so this checks whether there are upstream changes on whatever branch you're on, but not that you're on branch master. Any particular reason for that?

dannyroberts (approved):
Nice, very pro pushing any kind of best-practice process into our tools like this!

scripts/update-translations.sh
proteusvacuum: The previous check aborts if you are not on master.

proteusvacuum: I did it this way in case anyone has the upstream remote not set called `origin`

PR #18958, opened 2017-12-19
Allow selecting parent cases in shadow forms with their own details

proteusvacuum:
https://manage.dimagi.com/default.asp?267251

If you select "Select Parent First" and chose a different case list, shadow modules just ignore what you selected. This looks like it was intentional to do that. https://github.com/dimagi/commcare-hq/commit/1c24973ca602758c470dbc5c52e5aa84a05b0b2b#diff-a86df6df8b6c62195748605648bce855R374

However, I'm not sure if what I've done here is bad from a shadow forms perspoective.Pinging @orangejenny if you have some other insights that I might be missing as to why this was set up the way it was previously?

@dannyroberts interrupt buddy
@sravfeyn code buddy

### DIFF
--- a/corehq/apps/app_manager/suite_xml/sections/entries.py
+++ b/corehq/apps/app_manager/suite_xml/sections/entries.py
@@ -358,7 +358,7 @@ class EntriesHelper(object):
             else:
                 parent_filter = ''
 
-            detail_module = module if module.module_type == 'shadow' else datum['module']
+            detail_module = datum['module']
             detail_persistent = self.get_detail_persistent_attr(datum['module'], detail_module, "case_short")
             detail_inline = self.get_detail_inline_attr(datum['module'], detail_module, "case_short")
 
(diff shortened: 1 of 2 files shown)


orangejenny:
@proteusvacuum The original code was indeed intentional: when you select a datum in a shadow module, you should use the shadow module's detail, not the datum's associated module (which will be the source module). The oversight was that this didn't support parent-child selection, it just used the shadow module's detail for all `detail-select` screens.

What should happen is the shadow module's detail should only be used to select the (child) case, whereas the parent cases's `detail-select` should be the one associated with that parent case's datum (which ultimately comes from the module's `parent_select` property).

orangejenny:
Closing in favor of https://github.com/dimagi/commcare-hq/pull/18971

proteusvacuum:
Thanks @orangejenny !

PR #18967, opened 2017-12-20
Update analytics soft asserts recipients

orangejenny:
@biyeun Probably makes sense for us to get these going forward.

@calellowitz Left you on, feel free to remove yourself if that makes sense.


orangejenny:
or new code buddy @gcapalbo

gcapalbo (approved):

PR #18969, opened 2017-12-20
fix async_indicators_metrics IndexError

dannyroberts:
noticed this on sentry: https://sentry.io/dimagi/commcarehq/issues/416664542/
not sure whether it's causing anyone any problems

### DIFF
--- a/corehq/apps/userreports/tasks.py
+++ b/corehq/apps/userreports/tasks.py
@@ -384,7 +384,7 @@ def async_indicators_metrics():
         datadog_gauge('commcare.async_indicator.oldest_queued_indicator', lag)
 
     oldest_100_indicators = AsyncIndicator.objects.all()[:100]
-    oldest_indicator = oldest_100_indicators[0]
+    oldest_indicator = oldest_100_indicators[0] if oldest_100_indicators else None
     if oldest_indicator:
         lag = (now - oldest_indicator.date_created).total_seconds()
         datadog_gauge('commcare.async_indicator.oldest_created_indicator', lag)
--- a/corehq/apps/userreports/tasks.py
+++ b/corehq/apps/userreports/tasks.py
@@ -384,8 +384,8 @@ def async_indicators_metrics():
         datadog_gauge('commcare.async_indicator.oldest_queued_indicator', lag)
 
     oldest_100_indicators = AsyncIndicator.objects.all()[:100]
-    oldest_indicator = oldest_100_indicators[0] if oldest_100_indicators else None
-    if oldest_indicator:
+    if oldest_100_indicators.exists():
+        oldest_indicator = oldest_100_indicators[0]
         lag = (now - oldest_indicator.date_created).total_seconds()
         datadog_gauge('commcare.async_indicator.oldest_created_indicator', lag)
 


corehq/apps/userreports/tasks.py
```diff
@@ -384,7 +384,7 @@ def async_indicators_metrics():
         datadog_gauge('commcare.async_indicator.oldest_queued_indicator', lag)
 
     oldest_100_indicators = AsyncIndicator.objects.all()[:100]
-    oldest_indicator = oldest_100_indicators[0]
+    oldest_indicator = oldest_100_indicators[0] if oldest_100_indicators else None
```
esoergel: You could also move this inside the `if` block, since I don't think an `AsyncIndicator` instance can ever be falsy
```python
if oldest_100_indicators:
    oldest_indicator = oldest_100_indicators[0]
    ...
```
Also, how does this db evaluation work?  Would wrapping this in `list()` prevent any db calls?  I'm wondering specifically if the boolean check on `oldest_100_indicators` does an `.exists()` call and then a full bulk when it's evaluated.

esoergel (approved):

corehq/apps/userreports/tasks.py
dannyroberts: I thought the db was called when during the `[:100]` above, but maybe it's even lazier than that. Regardless, it would be insane for `if oldest_100_indicators:` to mean anything other than "if that query returned anything" when put right after a list-like slice, so I'm pretty sure that's fine to do.

But yeah I can move it into the if block like you suggested.

esoergel: The db is actually that lazy.  Slicing a queryset is just syntactic sugar for returning a new queryset with some additional parameters, it doesn't evaluate anything.  I just did some testing with the `django.db.backends` log turned up:
```python
In [1]: qs = SQLLocation.objects.filter(domain='esoergel')[:3]

In [2]: if qs:
   ...:     print "hi"
   ...:  
2017-12-20 21:27:35,801 DEBUG (0.002) SELECT "locations_sqllocation"."id", "locations_sqllocation"."domain", "locations_sqllocation"."name", "locations_sqllocation"."location_id", "locations_sqllocation"."location_type_id", "locations_sqllocation"."site_code", "locations_sqllocation"."external_id", "locations_sqllocation"."metadata", "locations_sqllocation"."created_at", "locations_sqllocation"."last_modified", "locations_sqllocation"."is_archived", "locations_sqllocation"."latitude", "locations_sqllocation"."longitude", "locations_sqllocation"."parent_id", "locations_sqllocation"."stocks_all_products", "locations_sqllocation"."supply_point_id", "locations_sqllocation"."user_id", "locations_sqllocation"."lft", "locations_sqllocation"."rght", "locations_sqllocation"."tree_id", "locations_sqllocation"."level" FROM "locations_sqllocation" WHERE "locations_sqllocation"."domain" = 'esoergel' ORDER BY "locations_sqllocation"."tree_id" ASC, "locations_sqllocation"."lft" ASC LIMIT 3; args=('esoergel',)
hi

In [3]: len(list(qs))
Out[3]: 3
```
That is, `bool(qs)` evaluates the full query, which means in this case it behaves like we want, since that query is cacheable and we don't need to query for the full list.  It does, however, illustrate the importance of using `.exists()` when possible, since Django won't add it for you (justifiably, I'd say), and it'll happily evaluate the full queryset.

dannyroberts:
Oh, I see. Thanks

corehq/apps/userreports/tasks.py
dannyroberts: Thanks, @esoergel. I went with the more explicit `.exists()` so that no one else reading it has to think to themselves "is this by mistake executing a larger query than it needs to—oh I see, doesn't matter because the same query is used below and gets cached".

snopoke:
Merging despite build failure (unrelated)

PR #18970, opened 2017-12-20
Report Builder: Cleanup

sheelio:
This PR does a few things:
- It cleans up a lot of old v1 related code
- It removes the special "group_by" field from the Table report to simplify the code
- It reduces the frequency that the report is refreshed
- It makes a number of changes to the preview report temp data source functionality:
  - It cleans up the code for building / managing the temp data sources
  - It forces the temp data source to include any columns that were chosen in the report
  - It optionally rebuilds the temp data source when the report preview is refreshed (if required)


kaapstorm (approved):
This is great!

PR #18930, opened 2017-12-18
Added UI tests to icds reports

ghost:
Last update of UI tests of icds reports for this ticket: 
https://manage.dimagi.com/default.asp?263218

@lwyszomi

PR #18975, opened 2017-12-21
Revert "Added UI tests to icds reports"

snopoke:
Reverts dimagi/commcare-hq#18930

FYI @pstudzinski https://github.com/dimagi/commcare-hq/pull/18930#issuecomment-353207325


czue:
@lwyszomi @pstudzinski as you can see we had to revert https://github.com/dimagi/commcare-hq/pull/18930 due to test failures. could you please re-open that PR and fix the test failures?

PR #18977, opened 2017-12-21
reduce retry threshold to an hour

czue:


czue:
@emord fyi for when you get back in case you have any concerns

emord:
@czue I haven't been keeping up with these and haven't gone through all the PRs but curious why this is needed?

This is really only needed if we have a very unexpected error in the celery worker which should be rare. 

https://github.com/dimagi/commcare-hq/blob/3c2c05d0557b1d41dd9f5b046e96de427e59f555/corehq/apps/userreports/tasks.py#L322 & https://github.com/dimagi/commcare-hq/blob/3c2c05d0557b1d41dd9f5b046e96de427e59f555/corehq/apps/userreports/models.py#L833 resets the datae queued when they failed, so they should get queued again immediately after a failure.

If it takes more than one hour to go through the initial documents that are queued then this will double queue some documents (and it looks like you've increased the number of documents that gets queued in other PRs)

czue:
Hey - welcome back. 

Yeah discussed with @snopoke and came to the same conclusion you have. However, for reasons we couldn't understand things were getting stuck in an unqueued state with a recent `date_queued` and so everything was stopping running because the same 30,000 documents just kept getting loaded and skipped. You can see this happening [here](https://app.datadoghq.com/dash/319634/asynchronous-ucr-dashboard?live=false&page=0&is_auto=false&from_ts=1513807200000&to_ts=1513893599999&tile_size=m&tpl_var_env=icds-new&fullscreen=false)

We deployed this as a workaround believing that double-queueing was better than stalling and not having the time/capacity to get to the bottom of the root cause. Would be great to figure it out and get rid of this change once we have a bit more time/capacity.

PR #18982, opened 2017-12-21
yaml has special words for true and false

proteusvacuum:
@sravfeyn


esoergel:
Merging 'cause Farid ran locally

PR #18989, opened 2017-12-22
Filter get_reverse_indexed_cases by type and 'closed'

esoergel:
As you can see, eNikshay uses this accessor method a lot, and then pretty much every time filters in memory at least by case type and often by open/closed status.  This means we're needlessly fetching a bunch of cases all the time, particularly when getting cases reverse indexed from episodes, which currently pull all adherence cases.  I'm not entirely sure how to get good numbers on the improvements this does, but I'm open to suggestions.

I'd like to wait until after break to merge, but opening for review.  Commit-by-commit should make the most sense.


corehq/form_processor/interfaces/dbaccessors.py
```diff
@@ -382,8 +382,12 @@ def get_closed_case_ids_for_owner(self, owner_id):
     def get_all_reverse_indices_info(self, case_ids):
         return self.db_accessor.get_all_reverse_indices_info(self.domain, case_ids)
 
-    def get_reverse_indexed_cases(self, case_ids):
-        return self.db_accessor.get_reverse_indexed_cases(self.domain, case_ids)
+    def get_reverse_indexed_cases(self, case_ids, case_types=None, is_closed=None):
+        # TODO only the SQL version accepts these args, how should we handle?
+        # assert is_sql or case_types is None and is_closed is None
+        # Mimic the functionality in memory for couch?
+        # Change couch view and reindex?
```
esoergel: This PR changes the method signature for only the SQL version.  Do we have a convention for how to handle that?  I laid out some options here.  I don't think this has much utility outside of custom dev, so the couch version may never get used.

corehq/sql_accessors/sql_templates/get_reverse_indexed_cases_3.sql
```diff
@@ -0,0 +1,33 @@
...
+) RETURNS SETOF form_processor_commcarecasesql AS $$
+DECLARE
+    query_expr      TEXT := '
+    SELECT case_table.* FROM form_processor_commcarecasesql as case_table
+        INNER JOIN form_processor_commcarecaseindexsql as case_index_table
+            ON ( case_table.case_id = case_index_table.case_id AND case_index_table.domain = $1 )
+            JOIN (SELECT UNNEST($2) AS referenced_id) AS cx USING (referenced_id)
+    WHERE
+        case_table.domain = $1
+        AND case_table.deleted = FALSE';
```
esoergel: I'm not crazy about sticking this big a query expression in a string, any idea if I can keep it in the body somehow?  Was mostly just trying to mimic the structure used in the other SQL functions.

custom/enikshay/case_utils.py
```diff
@@ -150,15 +145,12 @@ def get_associated_episode_case_for_test(test_case, occurrence_case_id):
...
 def get_all_episode_cases_from_person(domain, person_case_id):
-    case_accessor = CaseAccessors(domain)
-    episode_cases = []
     occurrence_cases = get_all_occurrence_cases_from_person(domain, person_case_id)
-    for occurrence_case in occurrence_cases:
-        all_cases = case_accessor.get_reverse_indexed_cases([occurrence_case.get_id])
-        episode_cases += [case for case in all_cases
-                          if case.type == CASE_TYPE_EPISODE and
-                          case.dynamic_case_properties().get('episode_type') == "confirmed_tb"]
-    return episode_cases
```
esoergel: Note that this change also switches from a query per occurrence_case to one query across all occurrence_cases under a person.

custom/enikshay/case_utils.py
```diff
@@ -692,44 +672,38 @@ def get_most_recent_referral_case_from_person(domain, person_case_id):
...
+            if case.type == CASE_TYPE_TEST:
+                for voucher_case in accessor.get_reverse_indexed_cases(
+                        [case.case_id], case_types=[CASE_TYPE_VOUCHER]):
+                    yield voucher_case
+            if case.type == CASE_TYPE_EPISODE:
+                for prescription_case in accessor.get_reverse_indexed_cases(
+                        [case.case_id], case_types=[CASE_TYPE_PRESCRIPTION]):
+                    for voucher_case in accessor.get_reverse_indexed_cases(
+                            [prescription_case.case_id], case_types=[CASE_TYPE_VOUCHER]):
+                        yield voucher_case
```
esoergel: I can't wait for python 3's `yield from`

proteusvacuum:
Build failed: https://travis-ci.org/dimagi/commcare-hq/jobs/320355643#L4408
```======================================================================
ERROR: corehq.apps.userreports.tests.test_expressions:TestGetSubcasesExpression.test_no_subcases
  File "/mnt/commcare-hq-ro/corehq/util/test_utils.py", line 269, in inner
  File "/mnt/commcare-hq-ro/corehq/util/test_utils.py", line 241, in __call__
  File "/mnt/commcare-hq-ro/corehq/util/test_utils.py", line 255, in call_with_settings
    fn(*args, **kwargs)
  File "/mnt/commcare-hq-ro/corehq/util/test_utils.py", line 237, in fn_with_pre_and_post
    self.fn(*args, **kwargs)
  File "/mnt/commcare-hq-ro/corehq/apps/userreports/tests/test_expressions.py", line 1159, in test_no_subcases
    subcases = self.expression(case.to_json(), self.context)
  File "/mnt/commcare-hq-ro/corehq/apps/userreports/expressions/specs.py", line 436, in __call__
    return self._get_subcases(case_id, context)
  File "/mnt/commcare-hq-ro/corehq/apps/userreports/decorators.py", line 68, in _inner
    res = fn(*args, **kwargs)
  File "/mnt/commcare-hq-ro/corehq/apps/userreports/expressions/specs.py", line 441, in _get_subcases
    return [c.to_json() for c in CaseAccessors(domain).get_reverse_indexed_cases([case_id])]
  File "/mnt/commcare-hq-ro/corehq/form_processor/interfaces/dbaccessors.py", line 390, in get_reverse_indexed_cases
    return self.db_accessor.get_reverse_indexed_cases(self.domain, case_ids, case_types, is_closed)
TypeError: get_reverse_indexed_cases() takes exactly 2 arguments (4 given)```

File: corehq/form_processor/interfaces/dbaccessors.py (hunk 1 above)
Comment (esoergel) [reply corehq/form_processor/interfaces/dbaccessors.py]: @snopoke Do you have any suggestions for resolving this?

File: corehq/sql_accessors/sql_templates/get_reverse_indexed_cases_3.sql [hunk 5]
```diff
@@ -0,0 +1,33 @@
+DROP FUNCTION IF EXISTS get_reverse_indexed_cases(TEXT, TEXT[]);
```
emord: It looks like this changes the function quite a bit and uses the same function name. Can you change this to add a new function with a different name and then in a different PR dropping the old name? Doing it that way means we won't have interrmittent errors and it's possible to rollback

corehq/form_processor/interfaces/dbaccessors.py
emord: I think that changing couch views on forms/cases would take a long time to reindex.

It seems reasonable to do it in memory for couch. That's what it's doing now anyways right? If so then there shouldn't be any negative effects

emord (changes requested):
I think the main blocking change here is to change the function signatures so that we can revert/not have intermittent errors

corehq/sql_accessors/sql_templates/get_reverse_indexed_cases_3.sql
esoergel: Makes sense.  I'm guessing that's something we'd want to do for any SQL function changes that affect the parameters or return value.  Would be good to add to https://github.com/dimagi/commcare-hq/blob/master/corehq/form_processor/README.md#updating-a-sql-function .  I'll do that if no one else gets to it before I wrap up this PR.

corehq/form_processor/backends/sql/dbaccessors.py
```diff
@@ -867,14 +867,15 @@ def get_indexed_case_ids(domain, case_ids):
...
+    def get_reverse_indexed_cases(domain, case_ids, case_types=None, is_closed=None):
         assert isinstance(case_ids, list)
+        assert case_types is None or isinstance(case_types, list)
         if not case_ids:
             return []
 
         cases = list(CommCareCaseSQL.objects.raw(
-            'SELECT * FROM get_reverse_indexed_cases(%s, %s)',
-            [domain, case_ids])
+            'SELECT * FROM get_reverse_indexed_cases_3(%s, %s, %s, %s)',
```
emord: Why `_3` instead of `2`?

corehq/sql_accessors/sql_templates/get_reverse_indexed_cases_3.sql
emord: I agree that this  seems weird. It's possible that plpgsql doesn't have a way to add filters other than string concatenation though.

@snopoke is that why our other functions do that?

emord (approved):
Didn't really look at the changes in enikshay logic. Thanks for updating the documentation on updating a sql function

corehq/form_processor/backends/sql/dbaccessors.py
esoergel: There's the sql function name, the name of the `.sql` file containing the function, and the analog of both of those in `sql_proxy_accessors`.  I figured it'd be a little saner to use the same number for all four of those (the file name needed to be 3, since the last version was 2).

corehq/sql_accessors/sql_templates/get_reverse_indexed_cases_3.sql
esoergel: aside: "plpgsql" is such a letter salad - "we use plpgsql to interact with the postgresql ordbms, which replaces our internal tool, asdfghjkl;"

snopoke: If you want to make the query dynamic you have to either to string concat like you're doing now or string formatting like this: https://github.com/dimagi/commcare-hq/blob/b927357119edb2f92fe02cd3257434586f06f14d/corehq/sql_accessors/sql_templates/get_forms_by_state.sql#L19

snopoke (approved):

PR #18991, opened 2017-12-27
task to delete old images

snopoke:
@dimagi/scale-team

### DIFF
--- a/custom/icds/tasks.py
+++ b/custom/icds/tasks.py
@@ -1,7 +1,12 @@
 from __future__ import absolute_import
+
+from datetime import timedelta, datetime
+
 import pytz
 from celery.schedules import crontab
 from celery.task import task, periodic_task
+from django.conf import settings
+
 from corehq.apps.locations.dbaccessors import (
     generate_user_ids_from_primary_location_ids_from_couch,
     get_location_ids_with_location_type,
@@ -11,6 +16,9 @@ from corehq.apps.reminders.tasks import CELERY_REMINDERS_QUEUE
 from corehq.apps.reminders.util import get_one_way_number_for_recipient
 from corehq.apps.sms.api import send_sms, MessageMetadata
 from corehq.apps.users.models import CommCareUser
+from corehq.blobs import get_blob_db
+from corehq.form_processor.models import XFormAttachmentSQL
+from corehq.sql_db.util import get_db_aliases_for_partitioned_query
 from corehq.util.timezones.conversions import ServerTime
 from custom.icds.const import (
     AWC_LOCATION_TYPE_CODE,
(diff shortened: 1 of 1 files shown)


custom/icds/tasks.py
```diff
@@ -159,3 +165,37 @@ def run_user_indicators(phased_rollout=True):
...
+
+    start = datetime.utcnow()
+    max_age = start - timedelta(days=90)
+    db = get_blob_db()
+    paths = []
+    deleted_attachments = []
+
+    def _get_query(db_name, max_age):
+        return XFormAttachmentSQL.objects.using(db_name).filter(
+            content_type='image/jpeg',
```
czue: are all images jpegs?

czue (commented):
not "approving" just cause I don't really know how the blobdb works but looks good.

custom/icds/tasks.py
snopoke: For ICDS they are

custom/icds/tasks.py
```diff
@@ -159,3 +165,37 @@ def run_user_indicators(phased_rollout=True):
             if is_monday:
                 run_indicator.delay(domain, user_id, LSSubmissionPerformanceIndicator, language_code)
                 run_indicator.delay(domain, user_id, LSVHNDSurveyIndicator, language_code)
+
+
+@periodic_task(run_every=crontab(minute=0, hour='22'))
+def delete_old_images():
+    if settings.SERVER_ENVIRONMENT not in settings.ICDS_ENVS:
+        return
```
dannyroberts: Is it possible to _only define_ this periodic task if `settings.SERVER_ENVIRONMENT not in settings.ICDS_ENVS`?

Not a huge deal, but I think that would prevent a no-op task from being enqueued once a day on other environments, which is much more of a clutter/principle issue than a performance one.

I'm thinking either

```
if settings.SERVER_ENVIRONMENT not in settings.ICDS_ENVS:
    @periodic_task(run_every=crontab(minute=0, hour='22'))
    def delete_old_images():
...
```

or

```
def delete_old_images():
...
if settings.SERVER_ENVIRONMENT not in settings.ICDS_ENVS:
    delete_old_images = periodic_task(run_every=crontab(minute=0, hour='22'))(delete_old_images)
```

custom/icds/tasks.py
```diff
@@ -159,3 +165,37 @@ def run_user_indicators(phased_rollout=True):
...
+        return XFormAttachmentSQL.objects.using(db_name).filter(
+            content_type='image/jpeg',
+            form__domain='icds-cas',
+            form__received_on__lt=max_age
+        )
+
+    for db_name in get_db_aliases_for_partitioned_query():
+        attachments = _get_query(db_name, max_age)
+        while attachments.exists():
+            for attachment in attachments[10000]:
```
millerdev: Was `attachments[10000]` meant to be a slice (limit)? Something like `attachments[:10000]` As is, it appears to reference the 10,000th attachment, and I don't understand why you'd only want to that here.

snopoke:
Updated

millerdev (approved):

PR #18992, opened 2017-12-27
Add episode id as the source doc id [skip ci]

mkangia:
Add episode id on the report to map them to the episode.
Will be useful when finding cases that need to be retriggered for notification.


calellowitz (changes requested):
seems good to me from the technical, have you talked to the project team about this. adding columns to user facing reports, particularly columns that wont have meaning to the users seems like it might not be a good idea.

mkangia:
@calellowitz 
thanks for flagging. Checked with the team and confirmed it was fine to add this.

calellowitz (approved):

PR #18993, opened 2017-12-27
Various fixes for health establishment integration

mkangia:
1. add verify false to avoid failure when checking for certificate for https url
2. sanitise text for xml
3. add response parse in case of pre-existing locations
4. fix check for node by using "is not None" otherwise its considered as falsy even when present

@proteusvacuum


calellowitz (approved):
looks fine to me but i do not really have any context on the code

mkangia:
Thanks for the review @calellowitz.
@proteusvacuum can you give a look?

PR #18995, opened 2017-12-29
add new data source for episode task list

mkangia:
This adds this [data source](https://enikshay.in/a/enikshay/configurable_reports/data_sources/edit/ab2aeeadb0bffed38ecdf65305529412/) as static 
we can remove the one already present before deploying unless it has been removed already. and also kill the celery task building it if it still is... celerymachine/task/9c25e234-d709-4a33-9bd2-6e55923a2bcc

### DIFF
--- a/custom/enikshay/ucr/data_sources/episode_tasklist_v3.json
+++ b/custom/enikshay/ucr/data_sources/episode_tasklist_v3.json
@@ -1,0 +1,2733 @@
+{
+  "domains": [
+    "enikshay",
+    "sheel-enikshay",
+    "enikshay-reports-qa",
+    "enikshay-test",
+    "enikshay-performance-test"
+  ],
+  "server_environment": [
+    "enikshay"
+  ],
+  "config": {
+    "referenced_doc_type": "CommCareCase",
+    "asynchronous": true,
+    "engine_id": "ucr",
+    "description": "",
+    "base_item_expression": {},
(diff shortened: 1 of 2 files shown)


ctsims (approved):

custom/enikshay/ucr/data_sources/episode_tasklist_v3.json
```diff
@@ -0,0 +1,2733 @@
...
+          }
+        },
+        "transform": {},
+        "is_primary_key": false,
+        "is_nullable": true,
+        "type": "expression",
+        "column_id": "owner_name"
+      },
+      {
+        "comment": "TODO: REMOVE?",
```
czue: should this be removed?

czue: (same question applies to all "todo" comments in here)

czue (approved):
i'm assuming this is just moving an already functional data source to static yeah? if so this seems fine (I obviously didn't review the indicators in detail).

there are loads of todos in the comments, many of which say that things should be removed. would we want to address any of those prior to building the data source?

custom/enikshay/ucr/data_sources/episode_tasklist_v3.json
mkangia: Its present in the original data source as well `episode_task_listv2` so not changing it.

PR #18997, opened 2018-01-02
Gateway error handling improvements

gcapalbo:
https://manage.dimagi.com/default.asp?267665

We've recently had some instances where there is a large volume of queued messages belonging to one gateway which is experiencing downtime. This causes a few problems which this PR solves by:

- using a more reasonable timeout value so that these messages don't block processing of other messages for so long

- only notifying exceptions about a down gateway once an hour per gateway so that we don't use up so many sentry events

- enforces a maximum number of 8 simultaneous connections to the push backend, which was previously enforced by setting the concurrency on the sms celery worker to 8. now that this will happen in the code, we can bump the number of sms workers we use to keep messages flowing when one gateway is down.

@orangejenny

### DIFF
--- a/corehq/apps/sms/tasks.py
+++ b/corehq/apps/sms/tasks.py
@@ -134,18 +134,25 @@ def connection_slot_key_base(backend):
 
 
 def reserve_connection_slot(backend, max_simultaneous_connections):
-    with CriticalSection(['reserve-connection-slot-for-%' % backend.couch_id]):
+    """
+    There is one redis key per connection slot, numbered from 1 to
+    max_simultaneous_connections.
+    A slot is considered taken if the corresponding redis key exists,
+    or is considered free if it does not exist.
+    """
+    with CriticalSection(['reserve-connection-slot-for-%s' % backend.couch_id]):
         client = get_redis_client()
         key_base = connection_slot_key_base(backend)
         slot_keys = client.keys(key_base + '*')
-        slot_numbers = [int(slot_key.replace(key_base, '')) for slot_key in slot_keys]
+        reserved_slots = [slot_key.replace(key_base, '') for slot_key in slot_keys]
 
         for slot_number in range(1, max_simultaneous_connections + 1):
-            if slot_number not in slot_numbers:
-                key = key_base + slot_number
+            slot_string = str(slot_number)
+            if slot_string not in reserved_slots:
+                key = key_base + slot_string
                 client.set(key, 1)
                 client.expire(key, 60)
-                return slot_number
+                return slot_string
 
     return None
 
--- a/corehq/apps/sms/tasks.py
+++ b/corehq/apps/sms/tasks.py
@@ -129,6 +129,32 @@ def message_is_stale(msg, utcnow):
         return True
 
 
+def connection_slot_key_base(backend):
+    return 'backend-%s-connection-slot-' % backend.couch_id
+
+
+def reserve_connection_slot(backend, max_simultaneous_connections):
+    with CriticalSection(['reserve-connection-slot-for-%' % backend.couch_id]):
+        client = get_redis_client()
+        key_base = connection_slot_key_base(backend)
+        slot_keys = client.keys(key_base + '*')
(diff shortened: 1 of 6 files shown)


corehq/apps/sms/tasks.py
```diff
@@ -129,6 +129,32 @@ def message_is_stale(msg, utcnow):
...
+        client = get_redis_client()
+        key_base = connection_slot_key_base(backend)
+        slot_keys = client.keys(key_base + '*')
+        slot_numbers = [int(slot_key.replace(key_base, '')) for slot_key in slot_keys]
+
+        for slot_number in range(1, max_simultaneous_connections + 1):
+            if slot_number not in slot_numbers:
+                key = key_base + slot_number
+                client.set(key, 1)
+                client.expire(key, 60)
```
esoergel: Is this 60 seconds?  Was it just chosen arbitrarily?

corehq/apps/sms/tasks.py
```diff
@@ -154,12 +181,21 @@ def handle_outgoing(msg):
...
+            return True
+
     result = send_message_via_backend(
         msg,
         backend=backend,
         orig_phone_number=orig_phone_number
     )
 
+    if max_simultaneous_connections:
+        free_connection_slot(backend, connection_slot)
```
esoergel: Flagging that if there's an exception in sending the message, the slot won't be freed up until it hits the timeout.  You could try/catch the in between calls or use a construct like `with reserve_connection_slot(...)` to guarantee that this gets called.

esoergel: Actually I guess `send_message_via_backend` is itself wrapped in a big try/except, huh?  Given that, this is probably safe enough to to do as-is.

esoergel (approved):
Looks solid

corehq/apps/sms/tasks.py
```diff
@@ -154,12 +181,21 @@ def handle_outgoing(msg):
             # Requeue the message and try it again shortly
             return True
 
+    if max_simultaneous_connections:
```
orangejenny: Thoughts on explicitly testing for `is not None` here (and below), or else noting in `get_max_simultaneous_connections` that a return value of zero will also be ignored?

orangejenny (approved):

corehq/apps/sms/tasks.py
gcapalbo: @orangejenny i don't think we'd be able to support a value of 0 since that would mean that we'd never have any connections open

corehq/apps/sms/tasks.py
gcapalbo: yeah 60 seconds. it should be large enough to cover an http request to a gateway, but small enough that waiting for it to timeout if a task drops half-way through doesn't take too long, so 1 minute seemed appropriate.

corehq/apps/sms/tasks.py
```diff
@@ -129,6 +129,39 @@ def message_is_stale(msg, utcnow):
...
+    """
+    There is one redis key per connection slot, numbered from 1 to
+    max_simultaneous_connections.
+    A slot is considered taken if the corresponding redis key exists,
+    or is considered free if it does not exist.
+    """
+    with CriticalSection(['reserve-connection-slot-for-%s' % backend.couch_id]):
+        client = get_redis_client()
+        key_base = connection_slot_key_base(backend)
+        slot_keys = client.keys(key_base + '*')
```
snopoke: @gcapalbo I think this is causing issues with redis. Most of the entries in the redis slow log are from this:
```
128) 1) (integer) 52935
     2) (integer) 1516023372
     3) (integer) 957970  (microseconds = 0.95 s)
     4) 1) "KEYS"
        2) ":1:backend-7d74ce121f9d4d379d084278e4987cbf-connection-slot-*"
     5) "172.24.32.14:37935"
     6) ""
```

Also this is what the slowlog metric looks like which roughly coincides with when this was deployed:
image

Could you not use a redis set to do this?

PR #19001, opened 2018-01-02
preventDefault action when submitting case claim query

wpride:
The default behavior of the submit button was causing the screen to reload after submitting  a case claim query when we just want Marionette to update the page asynchronously.


ctsims:
I don't think I have the context to review this, honestly, but I don't see any obvious problems.

PR #19004, opened 2018-01-02
 end_date should be earliest date_start of future subscriptions

nickpell:
Fixes a bug, and removes some obsolete query constraints (now enforced by a table constraint).

### DIFF
--- a/corehq/apps/accounting/models.py
+++ b/corehq/apps/accounting/models.py
@@ -76,8 +76,6 @@ SMALL_INVOICE_THRESHOLD = 100
 
 UNLIMITED_FEATURE_USAGE = -1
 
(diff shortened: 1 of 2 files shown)


nickpell:
@dannyroberts

dannyroberts (approved):
@nickpell can you give me a quick description of what makes this check obsolete? Also, is it a _change to something better_ or are you removing it because it's currently a no-op?

nickpell:
@dannyroberts 

This migration adds a table constraint to ensure that the subscription's end date is always greater than or equal to the start date: https://github.com/dimagi/commcare-hq/blob/3e368f969a2f1f1c623bf9c1425b70d8acb533ce/corehq/apps/accounting/migrations/0013_subscription_dates_check.py#L19

nickpell:
(Which makes it a no-op)

dannyroberts:
Aha! Thanks, makes sense!

PR #19007, opened 2018-01-03
Mk private sector fixes 3

mkangia:
hopefully last set of fixes

best review 🐟 / 🐟

### DIFF
--- a/custom/enikshay/integrations/nikshay/field_mappings.py
+++ b/custom/enikshay/integrations/nikshay/field_mappings.py
@@ -155,10 +155,10 @@ basis_of_diagnosis = {
 }
 
 health_establishment_type = {
-    'hospital': 'H',
-    'lab': 'L',
+    'Hospital': 'H',
+    'Lab': 'L',
     'pharmacy': 'P',
-    'doctor': 'P'
+    'Doctor': 'P'
 }
 
 health_establishment_sector = {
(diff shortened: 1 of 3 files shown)


custom/enikshay/integrations/nikshay/field_mappings.py
```diff
@@ -150,9 +155,10 @@
 }
 
 health_establishment_type = {
-    'hospital': 'H',
-    'lab': 'L',
+    'Hospital': 'H',
+    'Lab': 'L',
     'pharmacy': 'P',
```
calellowitz: is this supposed to not be capitalized. Its the only one

mkangia:
@calellowitz i need this in deploy and its a safe + independent change so i am merging this. Let me know if there is anything i should fix in this.

custom/enikshay/integrations/nikshay/field_mappings.py
```diff
@@ -150,9 +155,10 @@
...
+    'Doctor': 'P'
```
calellowitz: should this have a different letter (same as pharmacy now)

custom/enikshay/integrations/nikshay/field_mappings.py
mkangia: there is actually no use for that value.. we can remove it later.

custom/enikshay/integrations/nikshay/field_mappings.py
mkangia: Ya it should be the same.

PR #19010, opened 2018-01-03
Fix broken icds ui test

ghost:
Use fake date in tests in: kpi.directive.spec.js and month-filter.directive.spec.js files.
@esoergel 
@lwyszomi


esoergel:
thanks!

PR #19013, opened 2018-01-03
only try to process docs that have associated indicators

czue:
https://sentry.io/dimagi/commcarehq/issues/309028178/?environment=icds-new

I don't have a good model for when this error would trigger, but it seems to happen a lot. Basically some of the `doc_ids` that get passed to the `save_document` task no longer have an associated indicator and are causing the whole lot to fail. I assume due to parallel processing? Not sure if the locks are meant to help but they are not preventing all errors.

Fyi @emord in case you wanted to look closer into this when you get back. Seems to have been coming up consistently for a long time.

### DIFF
--- a/corehq/apps/userreports/tasks.py
+++ b/corehq/apps/userreports/tasks.py
@@ -306,7 +306,7 @@ def save_document(doc_ids):
         indicator_config_ids = first_indicator.indicator_config_ids
 
         with timer:
-            for doc in doc_store.iter_documents(doc_ids):
+            for doc in doc_store.iter_documents(indicator_by_doc_id.keys()):
                 indicator = indicator_by_doc_id[doc['_id']]
                 successfully_processed, to_remove = _save_document_helper(indicator, doc)
                 if successfully_processed:


emord:
I think this is because what I described here: https://github.com/dimagi/commcare-hq/pull/18977#issuecomment-354175876

This seems fine, but it's going to mess with some of the metrics (really an issue with metrics that should be fixed). Right now some of the graphs assume that it's processing 100 docs in these so it'll make the processing time look like less than it actually is.

This change seems fine to me, but I think the real fix will be moving that threshold back to 24 hours (or something much longer than an hour). and figuring out the issue you guys were running into:

> However, for reasons we couldn't understand things were getting stuck in an unqueued state with a recent date_queued and so everything was stopping running because the same 30,000 documents just kept getting loaded and skipped

czue:
Ah right that makes sense. I think this is probably better than the alternative even if the metrics get a little messed up, since at the moment if it fails here the entire batch gets dropped.

Definitely want to look back into that other issue - though we're still playing catch-up from the various problems we ran into so might hold off on that until things have stabilized.

PR #19015, opened 2018-01-03
Update v1 filter fields in addition to columns

esoergel:
I typed out some debugging details in the ticket:
https://manage.dimagi.com/default.asp?267201

I think this should fix the issue, although one thing I'm unsure of is how the data source gets updated.  I assume that happens when this is saved, otherwise the columns would break.  Would be great to get some confirmation on that though.

@sravfeyn

### DIFF
--- a/corehq/apps/userreports/reports/builder/forms.py
+++ b/corehq/apps/userreports/reports/builder/forms.py
@@ -1032,14 +1032,15 @@ class ConfigureNewReportBase(forms.Form):
         the knockout view model representing this filter in the report builder.
 
         """
-        exists = self._column_exists(filter['field'])
+        field = filter['field']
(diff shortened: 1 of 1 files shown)


esoergel:
@kaapstorm looks like you're area owner for report builder, can you validate that this change makes sense?

sravfeyn (approved):

esoergel:
@sheelio looks like you've done a fair bit of work in this area as well.  Does this change make sense?

sheelio:
I think this looks good (the change from v1 -> v2 was appending a data type to the possible options, but we didn't handle that for filters?)

kaapstorm (approved):

kaapstorm:
This looks right. +1 for improved docs.

PR #19016, opened 2018-01-03
Fix supply map reports

orangejenny:
https://manage.dimagi.com/default.asp?267551

Immediate cause of bug was having script tags in async partials: when the "Apply" button is clicked, those scripts get reloaded, which errors.

Also replaced the usage of `initial_page_data` in `base_maps.html` - the replacement code uses the same general idea, but doesn't use the actual initial-page-data code because that doesn't play well with ajax reports.

Overall this PR is a band-aid. The reports front end doesn't work especially well with more modern js, I'll need to deal with that eventually but not today.

@gcapalbo / @esoergel


orangejenny:
(supply is a feature flag, right?)

orangejenny:
ignoring all stickler complaints

gcapalbo (approved):

gcapalbo:
are the console.log statements there because you want to temporarily debug in production?

orangejenny:
@gcapalbo They were statements I moved, didn't add them...and I just realized I didn't commit a deleted file. https://github.com/dimagi/commcare-hq/pull/19016/commits/2b8d3e159409df089524fab275cc02ea08b9c4ff

gcapalbo:
ah, i see

gcapalbo (approved):

esoergel:
Nice, thanks Jenny!

PR #19018, opened 2018-01-03
disable default Subscription Manager

nickpell:
This has been requested in the past to make it more explicit that suppressed subscriptions are hidden from the default queryset.

### DIFF
--- a/corehq/apps/accounting/models.py
+++ b/corehq/apps/accounting/models.py
@@ -981,11 +981,16 @@ class Subscriber(models.Model):
             raise SubscriptionChangeError("The upgrade was not successful.")
 
 
-class SubscriptionManager(models.Manager):
+class VisibleSubscriptionManager(models.Manager):
 
     def get_queryset(self):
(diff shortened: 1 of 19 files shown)


corehq/apps/accounting/models.py
```diff
@@ -1026,7 +1031,9 @@ class Subscription(models.Model):
     is_hidden_to_ops = models.BooleanField(default=False)
     skip_auto_downgrade = models.BooleanField(default=False)
 
-    objects = SubscriptionManager()
+    visible_objects = VisibleSubscriptionManager()
+    visible_and_suppressed_objects = models.Manager()
```
dannyroberts: This basically renames `objects` to `visible_and_suppressed_objects`, right?

dannyroberts (approved):

corehq/apps/accounting/models.py
nickpell: yeah

PR #19021, opened 2018-01-03
python-modernize . -w -f libmodernize.fixes.fix_dict_six

nickpell:
@dimagi/py3

### DIFF
--- a/corehq/apps/app_manager/fixtures/mobile_ucr.py
+++ b/corehq/apps/app_manager/fixtures/mobile_ucr.py
@@ -60,7 +60,7 @@ class BaseReportFixturesProvider(FixtureProvider):
             return False
 
         apps = self._get_apps(restore_state, restore_user)
-        report_configs = self._get_report_configs(apps).values()
+        report_configs = list(self._get_report_configs(apps).values())
         if not report_configs:
             return False
 
@@ -144,7 +144,7 @@ class ReportFixturesProvider(BaseReportFixturesProvider):
         }
 
         if needed_versions.intersection({MOBILE_UCR_VERSION_1, MOBILE_UCR_MIGRATING_TO_2}):
-            fixtures.extend(self._v1_fixture(restore_user, self._get_report_configs(apps).values()))
+            fixtures.extend(self._v1_fixture(restore_user, list(self._get_report_configs(apps).values())))
         else:
             fixtures.extend(self._empty_v1_fixture(restore_user))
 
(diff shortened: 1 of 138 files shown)


corehq/apps/app_manager/fixtures/mobile_ucr.py
```diff
@@ -60,7 +60,7 @@ def uses_reports(self, restore_state):
             return False
 
         apps = self._get_apps(restore_state, restore_user)
-        report_configs = self._get_report_configs(apps).values()
+        report_configs = list(self._get_report_configs(apps).values())
         if not report_configs:
             return False
```
millerdev: Probably not something you want to fix now, but this could be simplified to
```py
return bool(self._get_report_configs(apps))
```
 The `.values()` call is not necessary for how the result is being used.

corehq/apps/app_manager/models.py
```diff
@@ -415,10 +415,10 @@ class FormActions(DocumentSchema):
 
     def all_property_names(self):
         names = set()
-        names.update(self.update_case.update.keys())
-        names.update(self.case_preload.preload.values())
+        names.update(list(self.update_case.update.keys()))
```
millerdev: I think most of these can be changed to eliminate `.keys()` (and sometimes even `list(...)`), so this would become
```py
names.update(self.update_case.update)
```
In general `list(dict_instance)` is equivalent to `list(dict_instance.keys())`

Understood if this is not something you want to address now since these are automated changes.

corehq/apps/app_manager/suite_xml/generator.py
```diff
@@ -111,7 +111,7 @@ def media_resources(self):
             for lang in self.build_profile.langs:
                 media_list += self.app.media_language_map[lang].media_refs
             requested_media = set(media_list)
-        for path, m in sorted(self.app.multimedia_map.items(), key=lambda item: item[0]):
+        for path, m in sorted(list(self.app.multimedia_map.items()), key=lambda item: item[0]):
```
millerdev: unnecessary change

corehq/apps/app_manager/suite_xml/post_process/instances.py
```diff
@@ -123,7 +123,7 @@ def __call__(self, fn):
 }
 
 
-@register_factory(*INSTANCE_KWARGS_BY_ID.keys())
+@register_factory(*list(INSTANCE_KWARGS_BY_ID.keys()))
```
millerdev: how does this even work (even before this change)? the args would be passed in a non-deterministic order. Is there only ever one arg?

millerdev (approved):
Didn't review every line. I think fine to merge if you don't want to visit and clean up each change manually.

Edit: however, it might be nice to eliminate the obviously unnecessary changes.

corehq/apps/app_manager/suite_xml/post_process/instances.py
nickpell: it looks like the ```__init__``` for ```register_factory``` could take in the args in any order and ```__call__``` would have the same result.

corehq/apps/app_manager/fixtures/mobile_ucr.py
nickpell: https://github.com/dimagi/commcare-hq/pull/19033

nickpell:
@millerdev I'm going to leave most of the automated changes as-is for now.  I may return to do some clean up if I'm bored later.

PR #19024, opened 2018-01-04
fix `timeEnd` `KeyError` on elasticsearch ICDS UCRs

czue:
This fixes [this error](https://sentry.io/dimagi/commcarehq/issues/427615729/) as well as adds a test for it. It also converts the test to use real elasticsearch

@emord I *think* I prefer using real elasticsearch to the mocks, but am open to an opposing viewpoint. I'll just leave this here for you to look at when you get back since it's not urgent.

### DIFF
--- a/custom/icds_reports/ucr/expressions.py
+++ b/custom/icds_reports/ucr/expressions.py
@@ -179,13 +179,7 @@ class FormsInDateExpressionSpec(JsonObject):
             context.set_cache_value(cache_key, count)
             return count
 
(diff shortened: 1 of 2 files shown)


emord:
Fine with me to use the real ES instead of a mock. I may have just done a mock because historically ES was a pain to setup in a test

czue:
@emord finally had a chance to look back into this and got to the bottom of it. since the cache is just in-memory, mutating the `xform` objects was causing the cached value to also lose the `timeEnd`. I could copy the forms before putting them in the cache, but that felt like needless extra work, so we just end up storing a version of the doc that also has the `timeEnd` populated in the cache. not sure if you have any concerns with doing that.

PR #19031, opened 2018-01-04
UCRs don't have an is_deprecated property

esoergel:
@mkangia
https://manage.dimagi.com/default.asp?267453
https://sentry.io/dimagi/commcarehq/issues/422090706/
Introduced https://github.com/dimagi/commcare-hq/pull/18789
Only seen a couple times, which kinda surprises me - I would've thought more UCRs used scheduled reports

### DIFF
--- a/corehq/apps/reports/models.py
+++ b/corehq/apps/reports/models.py
@@ -482,7 +482,7 @@ class ReportConfig(CachedCouchDocumentMixin, Document):
         except Exception:
             pass
 
-        if self.report.is_deprecated:
+        if getattr(self.report, 'is_deprecated', False):
             return ReportContent(
                 self.report.deprecation_email_message or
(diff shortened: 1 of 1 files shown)


esoergel:
Oh also @sravfeyn ping

orangejenny (approved):

corehq/apps/reports/models.py
```diff
@@ -482,7 +482,7 @@ def get_report_content(self, lang, attach_excel=False):
         except Exception:
             pass
 
-        if self.report.is_deprecated:
+        if getattr(self.report, 'is_deprecated', False):
```
gcapalbo: this is probably fine, but would it be better to only check `is_deprecated` if the object is an instance of a class that has the attribute?

PR #19032, opened 2018-01-04
Report Builder: Handle missing columns in current version

sheelio:
Product Note: This fixes a bug in report builder which would cause an error if  it was being edited if the report has a case property or form question that no longer existed in the application.  

@esoergel 
This should better handle missing columns (though I don't love how report builder generally handles form questions / case properties that used to exist but no longer exist).


sheelio:
Hold off on merging - this doesn't work perfectly.

esoergel (approved):

PR #19034, opened 2018-01-04
Fix tabclasses edge case

esoergel:
I'm pretty sure this is the issue behind https://manage.dimagi.com/default.asp?267185
I don't know why those requests wouldn't have a domain, could have something to do with the state of the user's auth, idk.  At any rate, as-is this is pretty clearly wrong.

First commit is minor clean-up, second commit is the fix.

@sravfeyn

### DIFF
--- a/corehq/apps/reports/dispatcher.py
+++ b/corehq/apps/reports/dispatcher.py
@@ -87,8 +87,8 @@ class ReportDispatcher(View):
             project = None
 
         def process(reports):
-            if project and callable(reports):
-                reports = reports(project)
+            if callable(reports):
+                reports = reports(project) if project else tuple()
             return tuple(reports)
 
         corehq_reports = process(getattr(reports, attr_name, ()))
--- a/corehq/apps/reports/dispatcher.py
+++ b/corehq/apps/reports/dispatcher.py
@@ -78,7 +78,7 @@ class ReportDispatcher(View):
         """
         return True
 
-    def get_reports(self, domain=None):
+    def get_reports(self, domain):
         attr_name = self.map_name
         from corehq import reports
         if domain:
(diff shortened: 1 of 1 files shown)


corehq/apps/reports/dispatcher.py
```diff
@@ -87,8 +87,8 @@ def get_reports(self, domain=None):
             project = None
 
         def process(reports):
-            if project and callable(reports):
-                reports = reports(project)
+            if callable(reports):
+                reports = reports(project) if project else tuple()
```
orangejenny: Today I learned that `tuple(tuple())` (which gets returned if project is falsy) is equivalent to `tuple()`. Interesting.

corehq/apps/reports/dispatcher.py
```diff
@@ -111,18 +111,15 @@ def process(reports):
...
-                    for report in group)
-
     def get_report(self, domain, report_slug, *args):
         """
         Returns the report class for `report_slug`, or None if no report is
         found.
-
         """
-        return self.get_reports_dict(domain).get(report_slug, None)
+        for name, group in self.get_reports(domain):
```
orangejenny: nice bit of cleanup

orangejenny (approved):

gcapalbo (approved):

corehq/apps/reports/dispatcher.py
esoergel: yeah, that bugged me

corehq/apps/reports/dispatcher.py
esoergel: haha, yeah, `tuple()` is a constructor , so it's roughly equivalent to a tuple literal `()`.  I just figured `tuple()` was more intuitive than `()` since you never really see the latter for some reason.  The second invocation (line 92) performs type coercion if `reports` isn't already a tuple.

This whole business about not knowing whether reports is even a callable or not is really code-smelly, but I didn't wanna go too deep and start refactoring all this.

PR #19035, opened 2018-01-04
Fix old exports

esoergel:
@dannyroberts @emord
https://manage.dimagi.com/default.asp?266678
New exports have a check for a table with the label 'Case History'.  This was causing an [exception](https://sentry.io/dimagi/commcarehq/issues/412148968/events/10179986964/) when old exports were accessed, as they didn't have the method being checked for.  The fix was to add that method also to old exports, but the logic at least (if not also the spirit) was specific to new exports, so that [errored](https://sentry.io/dimagi/commcarehq/issues/412148968/) too.  This PR just assumes old exports don't have case history, which is true at least of the export in question.  @emord do you know whether any old exports may have case history tables, and if so, how to identify them?

### DIFF
--- a/corehq/apps/reports/models.py
+++ b/corehq/apps/reports/models.py
@@ -1024,12 +1024,7 @@ class CaseExportSchema(HQExportSchema):
 
     @property
     def has_case_history_table(self):
-        case_history_table = [table for table in self.tables if table.label == 'Case History']
-        return any(
-            column.selected
-            for table in case_history_table
-            for column in table.columns
-        )
+        return False  # This check is only for new exports
 
 
 class DefaultFormExportSchema(DefaultExportSchema):


orangejenny:
Aren't we killing case history anyway?

orangejenny (approved):

orangejenny:
...Not to say this shouldn't be done, rather to say if we're killing it anyway, doesn't seem worth worrying about old exports that have case history.

dannyroberts (approved):

esoergel:
@orangejenny unfortunately under the current behavior, old exports 500 even if they don't have case history.

orangejenny:
Yeah, I meant that this assumption

> This PR just assumes old exports don't have case history

sounds fine, and even if it's incorrect, that seems like not a big deal if case history is going away.

PR #19037, opened 2018-01-04
remove dependency on guppy

nickpell:
https://trello.com/c/QqCeFGsB/26-guppy

@dimagi/py3

### DIFF
--- a/corehq/apps/app_manager/management/commands/build_apps.py
+++ b/corehq/apps/app_manager/management/commands/build_apps.py
@@ -8,13 +8,6 @@ from lxml import etree
 import os
 from corehq.apps.app_manager.models import Application, RemoteApp
 
-try:
-    from guppy import hpy
-    track_perf = True
-except ImportError:
-    track_perf = False
-
-
 _parser = etree.XMLParser(remove_blank_text=True)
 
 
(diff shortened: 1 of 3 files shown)


corehq/apps/app_manager/management/commands/build_apps.py
```diff
@@ -8,13 +8,6 @@
 import os
 from corehq.apps.app_manager.models import Application, RemoteApp
 
-try:
-    from guppy import hpy
-    track_perf = True
-except ImportError:
-    track_perf = False
```
millerdev: I think better to not modify this file. The command will still work if guppy is not installed. I believe when I checked last guppy did not officially support Python3 (so we should remove it from our requirements files), but I think it can be installed on Python3.

requirements/uninstall-requirements.txt
```diff
@@ -1,2 +1,3 @@
+guppy
```
millerdev: This is probably unnecessary, maybe even undesirable for devs who have it installed and want to use it?

requirements/dev-requirements.txt
```diff
@@ -1,5 +1,4 @@
 django-debug-toolbar==1.8
-guppy==0.1.10
```
millerdev: 👍

nickpell:
@millerdev All good points - updated

nickpell:
✅

PR #19038, opened 2018-01-05
Revert "Analytics: fix A/B tests"

czue:
Reverts dimagi/commcare-hq#19020

Suspect this might have caused http://manage.dimagi.com/default.asp?268059


czue:
confirming this fixed the bug. @mkangia @calellowitz

PR #19047, opened 2018-01-05
Report Builder: Add Parent Case Properties to List

sheelio:
Product Note: This also allows displaying parent case properties in the report builder for a case-based report.

@kaapstorm for report builder changes
@emord for the UCR changes


kaapstorm (approved):
This is nice to have.

corehq/apps/userreports/expressions/specs.py
```diff
@@ -291,6 +291,70 @@ def __str__(self):
...
+    case_expression = DictProperty(required=True)
+    index = StringProperty(required=False)
+
+    def configure(self, case_expression, context):
+        from corehq.apps.userreports.expressions import ExpressionFactory
+        self._case_expression = case_expression
+
+        index = self.index or 'parent'
+        spec = {
+            'type': 'related_doc',
```
emord: Since this is doesn't have any extra python code and it's just a shorthand expression for other UCR expression can it be added to https://github.com/dimagi/commcare-hq/blob/master/custom/ucr_ext/expressions.py instead?

sheelio:
@emord 

This is used by the Report Builder so doesn't feel like something that's custom.  I'll create a new code file for these types of expressions?

emord (approved):
Looks good once stickler stuff is fixed. Makes sense to move those out of custom/ucr_ext.

emord (approved):

emord (approved):

PR #19049, opened 2018-01-06
Remove IVR Functionality

gcapalbo:
https://manage.dimagi.com/default.asp?266165

Product: This removes IVR-related functionality from CommCareHQ as discussed in the ticket. It also removes the sms/callback workflow support since that's related. I'm labeling this with product/feature-flag because only superusers could enable IVR or sms/callback workflows before.

@orangejenny


gcapalbo:
no need to review, am going to open another pr with a slightly different approach for this

PR #19052, opened 2018-01-08
adding upsert and unique indices to simple dims

calellowitz:
@sravfeyn @gcapalbo @czue cc: @mkangia 
Now that ICDS (and hopefully the rest soon) is on 9.6 i switched the code that loads data into the dimension tables to properly use upsert, instead of a naive insertion. I also added a unique index on the relevant ids to make that possible.


czue:
build is failing. guessing we might need to bump the PG version on travis?

```
ERROR: corehq.warehouse.tests.test_dimensions:TestAppDim.test_app_dim
  File "/mnt/commcare-hq-ro/corehq/warehouse/tests/test_dimensions.py", line 278, in test_app_dim
    ApplicationDim.commit(self.batch)
  File "/mnt/commcare-hq-ro/corehq/warehouse/models/dimensions.py", line 41, in commit
    cls.load(batch)
  File "/mnt/commcare-hq-ro/corehq/warehouse/etl.py", line 48, in load
    cursor.execute(cls._sql_query_template(cls.slug, batch))
  File "/vendor/lib/python2.7/site-packages/raven/contrib/django/client.py", line 114, in execute
    return real_execute(self, sql, params)
  File "/vendor/lib/python2.7/site-packages/django/db/backends/utils.py", line 62, in execute
    return self.cursor.execute(sql)
  File "/vendor/lib/python2.7/site-packages/django/db/utils.py", line 94, in __exit__
  File "/vendor/lib/python2.7/site-packages/django/db/backends/utils.py", line 62, in execute
    return self.cursor.execute(sql)
ProgrammingError: syntax error at or near "ON"
LINE 28: ON CONFLICT (application_id) DO UPDATE
```

calellowitz:
We probably should not do that until all of our live clusters are upgraded. I added a do not merge tag so we can decide what the right way forward is while our clusters live in different states. Since this code is not used, I am not sure it matters but it would be best to upgrade first and then deploy this.

czue (approved):
seems good once we sort out the CI situation.

emord:
> We probably should not do that until all of our live clusters are upgraded

@calellowitz It seems fine to upgrade postgres since we have to support it on ICDS anwyays. In fact it seems kind of bad that we didn't upgrade it on our tests before doing it on prod.

I think we can just change postgres to 9.6 here https://github.com/dimagi/docker-postgresql but I haven't touched that before

corehq/warehouse/transforms/sql/application_dim.sql
```diff
@@ -25,3 +25,12 @@ SELECT
...
     '{{ batch_id }}'
 FROM {{ application_staging }}
+ON CONFLICT (application_id) DO UPDATE
+SET application_id = EXCLUDED.application_id
+    domain = EXCLUDED.domain,
+    name = EXCLUDED.name,
+    application_last_modified = EXCLUDED.application_last_modified,
+    deleted = EXCLUDED.deleted,
+    dim_last_modified = EXCLUDED.dim_last_modified,
+    dim_created_on = EXCLUDED.dim_created_on,
```
gcapalbo: we'll want to leave out created on from the update part

corehq/warehouse/transforms/sql/application_dim.sql
```diff
@@ -25,3 +25,12 @@ SELECT
     now(),
     '{{ batch_id }}'
 FROM {{ application_staging }}
+ON CONFLICT (application_id) DO UPDATE
+SET application_id = EXCLUDED.application_id
```
gcapalbo: we probably don't need to update application id since we already know it matches

corehq/warehouse/transforms/sql/application_dim.sql
```diff
@@ -25,3 +25,12 @@ SELECT
...
+    domain = EXCLUDED.domain,
+    name = EXCLUDED.name,
+    application_last_modified = EXCLUDED.application_last_modified,
+    deleted = EXCLUDED.deleted,
```
gcapalbo: does referencing `EXCLUDED.deleted` work? if so then we're ok, but if not we may need to alias the column in the `SELECT` query. i'm not familiar with the upsert syntax so i don't know if it's using the column names from the select or from the insert

gcapalbo:
@calellowitz looks good. would be nice to also add tests for these in the respective dim tests

calellowitz:
> In fact it seems kind of bad that we didn't upgrade it on our tests before doing it on prod.
I was more concerned about the breaking changes the opposite way. 9.6 should support anything that works in 9.4 but the inverse is not true. I do not expect anyone else to be playing around with postgres features but it seems more likely that someone could accidentally use a 9.6 only feature and break our other clusters than that something that passes on 9.4 does not work on 9.6

corehq/warehouse/transforms/sql/application_dim.sql
calellowitz: my understanding is that this should work. from the pg documentation
> The SET and WHERE clauses in ON CONFLICT DO UPDATE have access to the existing row using the table's name (or an alias), and to rows proposed for insertion using the special excluded table.

I interpreted deleted as part of the "row proposed for insertion", but let me know if that does not seem right to you. I also played around with sql fiddle and it appeared to work.
https://www.postgresql.org/docs/9.6/static/sql-insert.html

calellowitz:
closing this in favor of https://github.com/dimagi/commcare-hq/pull/19054
I merged all the commits from this branch there and have been adding enough new upsert work there that it did not seem worth separating them

PR #19054, opened 2018-01-08
Ce/auto batches

calellowitz:
This changes batches to have an auto incrementing id and start at the date of the last successful batch. It also prints the ID, to help airflow.

@gcapalbo @sravfeyn @czue cc: @mkangia 

includes commits from https://github.com/dimagi/commcare-hq/pull/19052 to avoid merge conflicts


corehq/warehouse/management/commands/create_batch.py
```diff
@@ -1,52 +1,26 @@
...
+    Example: ./manage.py create_batch
     """
     help = USAGE
 
-    def add_arguments(self, parser):
-        parser.add_argument('batch_id')
+    def handle(self, **options):
+        last_batch = Batch.objects.filter(completed_on__isnull=False).order_by('-end_datetime').first()
+        start = last_batch.end_datetime
+        end = date.today()
```
czue: Not knowing much about how this works - I assume it's fine if this ends up being quite a large window? We'd never want to do anything explicitly?

corehq/warehouse/management/commands/create_batch.py
```diff
@@ -1,52 +1,26 @@
...
     """
-    Example: ./manage.py create_batch 222617b9-8cf0-40a2-8462-7f872e1f1344 -s 2017-05-01 -e 2017-06-01
+    Example: ./manage.py create_batch
     """
     help = USAGE
 
-    def add_arguments(self, parser):
-        parser.add_argument('batch_id')
+    def handle(self, **options):
+        last_batch = Batch.objects.filter(completed_on__isnull=False).order_by('-end_datetime').first()
```
czue: This makes sense in steady state, but how does the first batch get into the DB? Does this need to check for that scenario?

czue (commented):
Don't really feel I have enough context to approve this, but threw in a few questions. Seems logical/fine generally.

corehq/warehouse/management/commands/create_batch.py
gcapalbo: yeah if it doesn't find one we'll need to set the start_datetime to a date that's far enough in the past to capture all source data

corehq/warehouse/management/commands/create_batch.py
gcapalbo: we should allow windows to be any size - from a couple hours to a couple weeks, but we should be setting this to datetime.utcnow().

in order for this range method to work, the following needs to be true:

- we need to store utc timestamps in the batch table
- we need to be comparing to last modified utc timestamps in the source data when we pull, which are set from server processes (not any timestamps that come from the phones)
- we need to be comparing last_modified with > start_datetime and <= end_datetime

corehq/warehouse/models/meta.py
```diff
@@ -3,8 +3,6 @@
 
 
 class Batch(models.Model):
-    batch_id = models.UUIDField(unique=True, db_index=True, primary_key=True)
-
```
gcapalbo: we should probably make all batch id foreign key columns be ints now too. not sure if we'll be able to use an actual foreign key if any facts get sharded. but if we ever have the capability to replicate certain tables across all shards, the batch table would be one to do that with

gcapalbo:
@calellowitz looks good. i think we're going to need to track batches on a per-dim and per-fact basis though. so we'll need to add the dim/fact slug to the batch table, and the process will be something like:

start batch for app dim >> load app staging >> load app dim >> complete batch for app dim

and for facts or any dims that have dependencies, we'll need to set the end_datetime of the batch to be the oldest end_datetime of all its dependencies' batches.

there's a couple reasons we're going to need it this way:

- when we add new dims, we'll need to load the entire history in the first batch. if we just use one shared batch it would only load an incremental on its first run.

- by splitting the batch per dim and fact, failures in lower priority dims / facts that nothing else depends on won't hold up loading other dims and facts on their regularly scheduled times

corehq/warehouse/management/commands/create_batch.py
```diff
@@ -1,26 +1,32 @@
...
-        last_batch = Batch.objects.filter(completed_on__isnull=False).order_by('-end_datetime').first()
-        start = last_batch.end_datetime
-        end = date.today()
+    def add_arguments(self, parser):
+        parser.add_argument('dag_slug')
+
+    def handle(self, dag_slug, **options):
+        last_batch = Batch.objects.filter(dag_slug=dag_slug, completed_on__isnull=False).order_by('-end_datetime').first()
+        start = last_batch.end_datetime if last_batch else '2000-01-01'
+        end = datetime.utcnow()
```
gcapalbo: as i mentioned in [my comment](https://github.com/dimagi/commcare-hq/pull/19054#issuecomment-356097907) we're going to need to inspect if this dim or fact has dependencies and if so use the oldest end_datetime from all dependencies as the end time.

each dim and fact class should have the dependencies listed in the code

corehq/warehouse/models/meta.py
```diff
@@ -8,6 +8,7 @@ class Batch(models.Model):
 
     created_on = models.DateTimeField(auto_now_add=True)
     completed_on = models.DateTimeField(null=True)
+    dag_slug = models.CharField(max_length=100)
```
gcapalbo: they might be the same but this should be the slug defined on the warehouse model class

corehq/warehouse/models/meta.py
calellowitz: do you know how to do this in django? All of the column specs in the migrations just say foreign key (it must determine the column type automatically). Perhaps adding a new migration with the same specification will force it to redetermine what type it is.

corehq/warehouse/management/commands/create_batch.py
calellowitz: I think we should try to push as much of the dependency logic to airflow as possible. If we make the batches self contained (one per dim for example), we can have airflow only run each batch when those dependencies have been fulfilled. Then we would not need to track dependencies in both airflow and our own code

corehq/warehouse/models/meta.py
gcapalbo: hmm, when you run the migration in this pr does it update the foreign key columns too? if not a manual migration might be needed

corehq/warehouse/management/commands/create_batch.py
gcapalbo: i think the only requirement here is to set the end datetime based on the oldest of the dependencies' batches end times because if you're loading a fact which links to data from a few different dims, then you need to make sure all your pulled fact data will have corresponding dim entries so that the foreign key references won't be blank. whether we do that here or within airflow doesn't really matter, but i don't think we can avoid doing this.

calellowitz:
Based on all the conversation on this PR, and the linked airflow one, I am going to flesh out this proof of concept more, since I think it is hard to have these conversations in the abstract. Over the next few days, I will continue to push to these PRs and any and all feedback while I am working is appreciated, but do not feel obligated to continue reviewing until I ping to say I think its done.

calellowitz:
@gcapalbo 
I think this is ready for a real review now. Beyond finalizing the batch methods, the main thing i changed was to make loading a staging table idempotent. If a later dim needed the same staging table as an earlier one, the load method will now determine if the later batch needs earlier data, and if so, add that data, otherwise it will leave the table as it was. This does mean we could write to the staging tables more than once per run of the DAG, but that seemed like an ok tradeoff to make, and would only happen if the multiple dependencies were not run at the same time in the past (likely only the first run of a new dim or fact that depends upon that staging table)

corehq/warehouse/etl.py
```diff
@@ -44,6 +46,14 @@ def load(cls, batch):
...
         assert issubclass(cls, WarehouseTable)
         database = db_for_read_write(cls)
+        if append:
+            batches = cls.objects.distinct('batch').values_list('batch', flat=True)
+            if batches:
+                oldest = Batch.objects.filter(pk__in=batches).order_by('start_datetime').first()
+                if batch.start_datetime < oldest.start_datetime:
+                    batch.end_datetime = oldest.start_datetime
+                else:
+                    return
```
gcapalbo: i'm not completely following what append is used for and what is happening when it's True - what's this for?

__in queries also don't perform that well at high scale

corehq/warehouse/etl.py
```diff
@@ -103,13 +113,22 @@ def record_iter(cls, start_datetime, end_datetime):
...
 
         assert issubclass(cls, WarehouseTable)
+        if append:
+            batches = cls.objects.distinct('batch').values_list('batch', flat=True)
+            if batches:
+                oldest = Batch.objects.filter(pk__in=batches).order_by('start_datetime').first()
+                if batch.start_datetime < oldest.start_datetime:
+                    batch.end_datetime = oldest.start_datetime
+                else:
+                    return
```
gcapalbo: same comment about append

gcapalbo:
On Wed, Jan 17, 2018 at 6:30 AM, Cal Ellowitz <notifications@github.com>
wrote:

> @gcapalbo <https://github.com/gcapalbo>
> I think this is ready for a real review now. Beyond finalizing the batch
> methods, the main thing i changed was to make loading a staging table
> idempotent. If a later dim needed the same staging table as an earlier one,
> the load method will now determine if the later batch needs earlier data,
> and if so, add that data, otherwise it will leave the table as it was. This
> does mean we could write to the staging tables more than once per run of
> the DAG, but that seemed like an ok tradeoff to make, and would only happen
> if the multiple dependencies were not run at the same time in the past
> (likely only the first run of a new dim or fact that depends upon that
> staging table)
>
what are the scenarios in which we'd need to do this? i think having
multiple processes update the same staging tables within a single run of
the warehouse is something we should try to avoid

calellowitz:
> what are the scenarios in which we'd need to do this? i think having
multiple processes update the same staging tables within a single run of
the warehouse is something we should try to avoid

The idea was that when there are multiple dims that depend on the same staging table, each dim would do its own load operation, by checking whether the batches that are already in the staging table cover the timeframe needed by the current batch. If not, it would load just the needed extra data, otherwise the load does nothing. That is the append logic above.

gcapalbo:
i don't think we have any dims or facts where we need this today though right? will defer to your judgement whether we need this or not, but it might introduce some complexity that we don't require

corehq/warehouse/models/meta.py
calellowitz: looks like i never responded to this comment but I have added migrations to do that in future commits

calellowitz:
> i don't think we have any dims or facts where we need this today though right? will defer to your judgement whether we need this or not, but it might introduce some complexity that we don't require

I was picturing dims like the `UserGroupDim` or `UserLocationDim` both of which depend on staging tables that are also used in other dims. So for example the `GroupStaging` table is used by both the `UserGroupDim` and the `GroupDim`. The work I did makes it so that if those two dims require different different dates worth of data it can happen naturally as opposed to precomputing the oldest date. If a dim that updates later needs data from an earlier date, it will load those dates into the staging table. If you think precomputing dates is simpler, I think we can make that change, it just requires us to change how we track dependencies (likely in the hq models themselves).

gcapalbo:
i still have a feeling that having multiple processes updating the same staging tables might get us into trouble. i think the preferred way of doing this is to design the airflow process in such a way that the dims with these dependencies always run together and always use the same dates and staging data.

if we ever add a new dim that also has the same staging dependencies as an existing one, we'd just do a one-time historical update, and then put it on the same schedule as the others so it can just share the same staging data as the rest going forward.

i think that might simplify things a bit - or do you think we would still need the append functionality?

calellowitz:
@gcapalbo That concern makes a lot of sense to me. I do have a couple questions about that alternative.

> i think the preferred way of doing this is to design the airflow process in such a way that the dims with these dependencies always run together and always use the same dates and staging data.

Does this mean rolling back one dim if a load into the other one fails so that they remain at the same time (possibly with a transaction you dont close until both succeed)? Or is there a simpler way.

> we'd just do a one-time historical update

Is this like a db migration/management command that we would manually do for each one, or do you see this as something that should be easy to handle automatically, either in airflow or our codebase.

gcapalbo:
On Mon, Jan 22, 2018 at 11:21 AM, Cal Ellowitz <notifications@github.com>
wrote:

> @gcapalbo <https://github.com/gcapalbo> That concern makes a lot of sense
> to me. I do have a couple questions about that alternative.
>
> i think the preferred way of doing this is to design the airflow process
> in such a way that the dims with these dependencies always run together and
> always use the same dates and staging data.
>
> Does this mean rolling back one dim if a load into the other one fails so
> that they remain at the same time (possibly with a transaction you dont
> close until both succeed)? Or is there a simpler way.
>
I don't think we'd need to worry about rolling anything back, I think it
would probably be enough to just rely on airflow for this - mainly we'd
just need to make sure that none of the jobs for any of these dims with
shared staging tables would start again until all of them have completed
successfully.

So that even if one of them has a failure, then none of them start again.
This, along with the change in the other PR around using the same end date
for all batches should be enough

> we'd just do a one-time historical update
>
> Is this like a db migration/management command that we would manually do
> for each one, or do you see this as something that should be easy to handle
> automatically, either in airflow or our codebase.
>
Usually these are one-off scripts like a management command, but depending
on what data is being brought in an alternative could be to just
temporarily introduce a new staging table for the first run, and then use
the shared staging tables after the batches are synchronized.

calellowitz:
Thanks for all the clarifications. I will look into implementing that tomorrow

calellowitz:
@gcapalbo I think this is finally good to go.

corehq/warehouse/management/commands/create_batch.py
```diff
@@ -16,11 +20,12 @@ class Command(BaseCommand):
 
     def add_arguments(self, parser):
         parser.add_argument('dag_slug')
+        parser.add_argument('end_date', nargs='?', default=None)
```
gcapalbo: should we just make this be required or is there somewhere it's not used?

also we could just pass in type=[utc_timestamp](https://github.com/dimagi/commcare-hq/blob/master/corehq/util/argparse_types.py#L6) and it will automatically handle validation

calellowitz: fixed

gcapalbo (approved):

corehq/warehouse/migrations/0023_auto_20180306_1258.py
```diff
@@ -0,0 +1,56 @@
+# -*- coding: utf-8 -*-
```
nickpell: There are two migrations that are prefixed with ```0023``` - can you please fix?  Is blocking staging deploy.

nickpell: bump @calellowitz @sravfeyn

sravfeyn: https://github.com/dimagi/commcare-hq/pull/19725

nickpell:
Heads up - another migrations conflict: 

```CommandError: Conflicting migrations detected; multiple leaf nodes in the migration graph: (0012_adding_indices_for_warehouse, 0012_get_location_fixture_ids in locations).```

@calellowitz

File: corehq/apps/reports/standard/deployments.py [hunk 10]
```diff
@@ -57,27 +60,38 @@ class ApplicationStatusReport(GetParamsMixin, PaginatedReportMixin, DeploymentsR
     ]
     primary_sort_prop = None
 
+    @property
+    def warehouse(self):
+        return settings.SERVER_ENVIRONMENT in settings.ICDS_ENVS \
+            or settings.SERVER_ENVIRONMENT == 'staging'
```
snopoke: consider making this a feature flag to allow easier rollout?

nickpell:
Staging deploy fails with:

```
[hqdjango0-staging.internal-va.commcarehq.org] out: CommandError: Conflicting migrations detected; multiple leaf nodes in the migration graph: (0013_adding_indices_for_warehouse, 0013_rm_mptt_fields in locations).
[hqdjango0-staging.internal-va.commcarehq.org] out: To fix them run 'python manage.py makemigrations --merge'
```

@calellowitz Can you fix?

calellowitz:
Tests are now passing. I think it is finally time to merge the warehouse master pr.

@snopoke @gcapalbo

corehq/apps/locations/models.py
```diff
@@ -379,12 +379,12 @@ class SQLLocation(AdjListModel):
     domain = models.CharField(max_length=255, db_index=True)
     name = models.CharField(max_length=255, null=True)
     location_id = models.CharField(max_length=100, db_index=True, unique=True)
-    location_type = models.ForeignKey(LocationType, on_delete=models.CASCADE)
+    location_type = models.ForeignKey(LocationType, on_delete=models.CASCADE, db_index=True)
```
snopoke: I don't see a migration for this but I also think that foreign keys are automatically indexed so this shouldn't be necessary

corehq/warehouse/models/facts.py
```diff
@@ -80,41 +80,64 @@ def dependencies(cls):
...
+
+    sync_log_id = models.CharField(max_length=255)
+    sync_date = models.DateTimeField(null=True)
+
+    # these can be null per SyncLogStagingTable
+    domain = models.CharField(max_length=255, null=True)
+
+    user_dim = models.ForeignKey(UserDim, on_delete=models.PROTECT)
+    # not all synclogs have domains, added in 11/2016
+    domain_dim = models.ForeignKey(DomainDim, on_delete=models.PROTECT, null=True)
```
snopoke: can we not infer this from the user and populate it anyway?

corehq/warehouse/models/meta.py
snopoke: @calellowitz ?

snopoke (commented):
Phew! Big PR! I haven't been following very closely only noticed some things now. Added a few comments but nothing blocking.

calellowitz:
> Phew! Big PR!

haha yeah. the original pr was approved a while ago. the rest is from merging in other PRs that were separately approved so nothing should be too new, but always good to have another set of eyes

corehq/warehouse/models/facts.py
calellowitz: I do not think we can for web users using webapps, but all new synclogs should have this anyway so we would not expect that to be an issue

corehq/apps/locations/models.py
calellowitz: hm ok. i will just delete that addition

PR #19058, opened 2018-01-08
convert more IO imports to python 3

nickpell:
These changes have test coverage and therefore should be relatively safe.

@dimagi/py3


emord (approved):
These kind of changes are like the most terrifying thing in python 3. Thanks for look at them!

PR #19059, opened 2018-01-08
Fetch all objects when we call api global/subscription

ghost:
Changed SubscriptionResource to get all subscriptions - no "is_hidden_to_ops" filter 
 https://manage.dimagi.com/default.asp?252556#1432373


fyniac:
@nickpell can you please review this and then merge if ok?

PR #19064, opened 2018-01-08
Don't count, just pick a big number

esoergel:
https://manage.dimagi.com/default.asp?266178
Currently this admin report is unusable due to the size of the table.  A lot of this comes from Django trying to do a full count for nice pagination.  There's a built-in option to not do a full count, but that only applies to the filtered version of the report.  This is a shameless hack that should hopefully make it so @amstone326 can at least use the report.
@sravfeyn

### DIFF
--- a/corehq/ex-submodules/phonelog/admin.py
+++ b/corehq/ex-submodules/phonelog/admin.py
@@ -1,10 +1,21 @@
 from __future__ import absolute_import
 from django.contrib import admin
+from django.core.paginator import Paginator
+from django.utils.functional import cached_property
 from .models import DeviceReportEntry, UserErrorEntry, UserEntry, ForceCloseEntry
+
+
+class TableIsTooBigPaginator(Paginator):
+    @cached_property
+    def count(self):
+        # This is supposed to return the matching count, but that's wicked slow
+        # for massive tables, so just pick a big number
+        return 10000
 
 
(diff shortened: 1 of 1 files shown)


corehq/ex-submodules/phonelog/admin.py
```diff
@@ -20,8 +31,6 @@ class DeviceReportEntryAdmin(admin.ModelAdmin):
         'date',
         'username',
     ]
-    # postgres SQL count query takes long time on big tables
-    show_full_result_count = False
```
sravfeyn: Surprised that this didn't work.

esoergel: Yeah, I was surprised too - the [docs](https://docs.djangoproject.com/en/1.11/ref/contrib/admin/#django.contrib.admin.ModelAdmin.show_full_result_count) say:
>  Set show_full_result_count to control whether the full count of objects should be displayed on a filtered admin page (e.g. 99 results (103 total)). If this option is set to False, a text like 99 results (Show all) is displayed instead.
> The default of show_full_result_count=True generates a query to perform a full count on the table which can be expensive if the table contains a large number of rows.

The problem was that if the admin page wasn't filtered, it'd still show the full count, and even filtered it would show _a_ count.  The filtered version of the report actually worked after adding that though.

PR #19066, opened 2018-01-09
Add tasks to run non-commit tasks

mkangia:
Just queue tasks to be run daily to email reconciliations non-commit runs.


calellowitz (approved):
dont have any context on this but it looks fine. why are these so late in the morning? these look like they will run during the work day

mkangia:
@calellowitz .. timed them to run morning IST preferably after the enikshay task has finished because they each take around 30 mins to finish up.

PR #19068, opened 2018-01-09
py3: non-DB tests only

nickpell:
opening to run tests only


.travis.yml
```diff
@@ -1,4 +1,7 @@
 language: python
+python:
+  - "2.7"
+  - "3.6"
```
millerdev: I think you want to remove one of these. Otherwise it will run the tests twice, but they'll both use the same python version: the one listed in the dockerfile.

